Issue #7591, opened 2025-07-07
[Improvement] fix index existence check


justinmclean:
### What would you like to be improved?

Issue occurs in deleteIndexDefinition in :
catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/operation/MysqlTableOperations.java
catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
catalogs/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/operation/OceanBaseTableOperations.java

### How should we improve?

The index existence check should only verify existence when ifExists is false, preventing failures when dropping non-existent indexes.

senlizishi:
Hi @justinmclean I looked at the code and was a bit confused. 

Firstly, the comment for isIfExists is` /* * @ return If true, silence the error if index does not exist during drop. */`This is different from its own meaning and looks more like a switch than whether it exists or not. 


Next, If anyMatch has an Index., throw Index does not exist. 
 Is it a logical error?  Could you give me some guidance?  Thx
```
if (Arrays.stream(lazyLoadTable.index())
          .anyMatch(index -> index.name().equals(deleteIndex.getName()))) {
        throw new IllegalArgumentException("Index does not exist");
      }
```

justinmclean:
If ifExists is set to true then the code should not throw an exception if the index doesn't exist.

senlizishi:
> If ifExists is set to true then the code should not throw an exception if the index doesn't exist.

OK. Please assign to me, I will try to fix it.

PR #7621, opened 2025-07-09
improvement(catalogs): fix index existence check

senlizishi:

fix index existence check

Fix: #7591 

No

tests added.


senlizishi:
@justinmclean PTAL

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -774,7 +774,7 @@ static String addIndexDefinition(TableChange.AddIndex addIndex) {
 
   static String deleteIndexDefinition(
       JdbcTable lazyLoadTable, TableChange.DeleteIndex deleteIndex) {
-    if (deleteIndex.isIfExists()) {
+    if (!deleteIndex.isIfExists()) {
       Preconditions.checkArgument(
           Arrays.stream(lazyLoadTable.index())
               .anyMatch(index -> index.name().equals(deleteIndex.getName())),
```
yuqi1129, line 780: I'm puzzled that since you use `!deleteIndex.isIfExists()` in L777,  should this place be `nonMatch` NOT `anyMatch` here?

senlizishi: > I'm puzzled that since you use `!deleteIndex.isIfExists()` in L777, should this place be `nonMatch` NOT `anyMatch` here?

Thank you for your review. My understanding is that when deleteIndex.isIfExists() is false, the code block will be entered. Next, check whether the name of any index in lazyLoadTable is the same as deleteIndex.getName(). If it is the same, it will run normally. Otherwise, it will report an Index does not exist exception. I don't know if my understanding is correct ?

justinmclean: My understanding is that if isIfExists returns true, then it will ignore any warnings about non-existent indexes. If it returns false, then you do want to warn about non-existent indexes.

yuqi1129: I see. It's my mistake.

justinmclean: The method doesn't have the best name, which can be a little confusing.

Issue #7382, opened 2025-06-12
[Subtask] Implement operations for model version with multiple URIs


xiaozcy:
### Describe the subtask

Implement operations for model version with multiple URIs.

### Parent issue

#7363

Issue #7436, opened 2025-06-19
[Subtask] Implement REST APIs for model version with multiple URIs


xiaozcy:
### Describe the subtask

Implement REST APIs for model version with multiple URIs.

### Parent issue

#7363

joonseolee:
@xiaozcy 

I’ve been handling only simple issues so far, but I wanted to focus on something more challenging, so I was looking for a ticket.  

Should I implement the REST APIs using the already developed features from ticket #7382 ?
I'm not sure which REST APIs need to be implemented — should I refer to ticket [#6893](https://github.com/apache/gravitino/issues/6893) for guidance?

xiaozcy:
Hi @joonseolee, thanks for following this issue, I'm already woking on it and will finish it soon, maybe you can start with some other issues.

PR #7622, opened 2025-07-09
[#7382] feat(model): Implement operations for model version with multiple URIs

xiaozcy:

This PR implements operations for model version with multiple URIs.

Fix: #7382 

No.

UT.

### DIFF
--- a/catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
+++ b/catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
@@ -28,8 +28,8 @@ import java.time.Instant;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Catalog;
 import org.apache.gravitino.Entity;
 import org.apache.gravitino.EntityAlreadyExistsException;
@@ -235,19 +235,6 @@ public class ModelCatalogOperations extends ManagedSchemaOperations
       throws NoSuchModelException, ModelVersionAliasesAlreadyExistException {
     NameIdentifierUtil.checkModel(ident);
 
-    if (uris == null || uris.isEmpty()) {
-      throw new IllegalArgumentException("At least one URI needs to be set for the model version");
-    }
-    uris.forEach(
-        (name, uri) -> {
-          if (StringUtils.isBlank(name)) {
-            throw new IllegalArgumentException("URI name must not be blank");
-          }
-          if (StringUtils.isBlank(uri)) {
-            throw new IllegalArgumentException("URI must not be blank for name: " + name);
-          }
-        });
-
     StringIdentifier stringId = StringIdentifier.fromProperties(properties);
     Preconditions.checkArgument(stringId != null, "Property string identifier should not be null");
 
@@ -288,7 +275,7 @@ public class ModelCatalogOperations extends ManagedSchemaOperations
     NameIdentifierUtil.checkModel(ident);
     NameIdentifier modelVersionIdent = NameIdentifierUtil.toModelVersionIdentifier(ident, version);
 
-    return internalGetModelVersionUri(ident, modelVersionIdent, uriName);
+    return internalGetModelVersionUri(ident, modelVersionIdent, Optional.ofNullable(uriName));
   }
 
   @Override
@@ -297,7 +284,7 @@ public class ModelCatalogOperations extends ManagedSchemaOperations
     NameIdentifierUtil.checkModel(ident);
     NameIdentifier modelVersionIdent = NameIdentifierUtil.toModelVersionIdentifier(ident, alias);
 
-    return internalGetModelVersionUri(ident, modelVersionIdent, uriName);
+    return internalGetModelVersionUri(ident, modelVersionIdent, Optional.ofNullable(uriName));
   }
 
   @Override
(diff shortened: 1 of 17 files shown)


jerryshao:
Hi @xiaozcy this PR is too big to review, I would suggest to split into some small PRs, like moving event related codes to another PR, which will be easy to review.

xiaozcy:
@jerryshao OK, I'll split it up later.

xiaozcy:
@jerryshao Please help review #7674 first, I will rebase this branch after it is merged.

jerryshao:
@xiaozcy I have left some comments in #7674 , please check it, thanks.

xiaozcy:
Hi @jerryshao, this PR is ready for review, thanks.

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
```diff
@@ -226,13 +228,26 @@ public ModelVersion getModelVersion(NameIdentifier ident, String alias)
...
+    uris.forEach(
+        (name, uri) -> {
+          if (StringUtils.isBlank(name)) {
+            throw new IllegalArgumentException("URI name must not be blank");
+          }
+          if (StringUtils.isBlank(uri)) {
+            throw new IllegalArgumentException("URI must not be blank for name: " + name);
+          }
+        });
+
```
jerryshao: I think we should do an such check when we received the request, not postpone to here.

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
```diff
@@ -524,6 +557,48 @@ private ModelVersion internalGetModelVersion(NameIdentifier ident) {
...
+    // If the default uri name is not set for the model version, try to get the default uri name
+    // from the model properties
+    Map<String, String> modelProperties = getModel(modelIdent).properties();
+    if (modelProperties.containsKey(ModelVersion.PROPERTY_DEFAULT_URI_NAME)) {
+      String defaultUriName = modelProperties.get(ModelVersion.PROPERTY_DEFAULT_URI_NAME);
+      return getUriByName(uris, defaultUriName, modelVersionIdent);
+    }
+
+    throw new IllegalArgumentException("Either uri name of default uri name should be provided");
```
jerryshao: “Either...or...”

jerryshao: Can you please also add the uri name to the exception message.

catalogs/catalog-model/src/test/java/org/apache/gravtitino/catalog/model/TestModelCatalogOperations.java
```diff
@@ -637,6 +630,171 @@ public void testLinkAndListModelVersionInfos() {
...
+    Assertions.assertEquals("u1", ops.getModelVersionUri(modelIdent, "alias2", "n1"));
+    Assertions.assertEquals("u2", ops.getModelVersionUri(modelIdent, "alias2", "n2"));
+
+    // Test get uri without uri name
+    Assertions.assertThrows(
+        IllegalArgumentException.class, () -> ops.getModelVersionUri(modelIdent, 0, null));
+    Assertions.assertThrows(
+        IllegalArgumentException.class, () -> ops.getModelVersionUri(modelIdent, "alias1", null));
+    Assertions.assertThrows(
+        IllegalArgumentException.class, () -> ops.getModelVersionUri(modelIdent, "alias2", null));
```
jerryshao, line 676: Do we support “null” uri name?

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
```diff
@@ -524,6 +557,48 @@ private ModelVersion internalGetModelVersion(NameIdentifier ident) {
     }
   }
 
+  private String internalGetModelVersionUri(
+      NameIdentifier modelIdent, NameIdentifier modelVersionIdent, String uriName) {
```
jerryshao: Do we support "null" uriName, if so, I would suggest using `Optional<String>` to define this argument to make the code more clear.

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
xiaozcy: OK, I'll remove these checks for now add them when REST API is supported later(#7436).

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
xiaozcy: The exception here will only be thrown when the uri name is not specified, how about change the message to "The default uri name needs to be set when the uri name is not specified"?

catalogs/catalog-model/src/test/java/org/apache/gravtitino/catalog/model/TestModelCatalogOperations.java
xiaozcy: Yes, the default uri name is not set in this test case, so `IllegalArgumentException` will be thrown here.

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
xiaozcy: Done.

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
jerryshao: OK.

xiaozcy:
Hi @jerryshao, CI seems to have been cancelled for unknown reasons, could you please help re-trigger it?

Issue #7623, opened 2025-07-09
[Subtask] Add core logic implementation for policy management (part-1)


mchades:
### Describe the subtask

This issue tracks the work of core logic implementation for policy management. As the part 1 task, it will only implement basic policy operations.

### Parent issue

#7139

PR #7625, opened 2025-07-09
[#7623] feat(policy): Add core logic implementation for policy management (part-1)

mchades:

Add core logic implementation for policy management

Fix: #7623 

no

tests added

### DIFF
--- a/core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
+++ b/core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
@@ -182,11 +182,9 @@ public class PolicyManager implements PolicyDispatcher {
                 "Policy with name %s under metalake %s does not exist", policyName, metalake);
           } catch (EntityAlreadyExistsException e) {
             throw new RuntimeException(
-                "Policy with name "
-                    + policyName
-                    + " under metalake "
-                    + metalake
-                    + " already exists");
+                String.format(
+                    "Trying to alter policy %s under metalake %s, but the new name already exists",
+                    policyName, metalake));
           } catch (IOException ioe) {
             LOG.error("Failed to alter policy {} under metalake {}", policyName, metalake, ioe);
             throw new RuntimeException(ioe);
--- a/core/src/main/java/org/apache/gravitino/tag/TagManager.java
+++ b/core/src/main/java/org/apache/gravitino/tag/TagManager.java
@@ -172,7 +172,9 @@ public class TagManager implements TagDispatcher {
                 "Tag with name %s under metalake %s does not exist", name, metalake);
           } catch (EntityAlreadyExistsException e) {
             throw new RuntimeException(
-                "Tag with name " + name + " under metalake " + metalake + " already exists");
+                String.format(
+                    "Trying to alter tag %s under metalake %s, but the new name already exists",
+                    name, metalake));
           } catch (IOException ioe) {
             LOG.error("Failed to alter tag {} under metalake {}", name, metalake, ioe);
             throw new RuntimeException(ioe);


mchades:
@jerryshao plz help to review when you have time, thanks!

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
```diff
@@ -51,17 +68,51 @@ public PolicyManager(IdGenerator idGenerator, EntityStore entityStore) {
...
 
   @Override
   public Policy[] listPolicyInfos(String metalake) {
-    throw new UnsupportedOperationException("Not implemented yet");
+    NameIdentifier metalakeIdent = NameIdentifierUtil.ofMetalake(metalake);
+    return TreeLockUtils.doWithTreeLock(
+        NameIdentifier.of(NamespaceUtil.ofPolicy(metalake).levels()),
+        LockType.READ,
+        () -> {
+          checkMetalake(metalakeIdent, entityStore);
```
jerryshao: I think we can move the `checkMetalake` out of lock, it is a low probablitiy that metalake will not exist or not in use when we do policy operations. Moving this out can improve the performance. WDYT?

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
```diff
@@ -74,28 +125,106 @@ public Policy createPolicy(
...
       boolean inheritable,
       Set<MetadataObject.Type> supportedObjectTypes,
-      PolicyContent content) {
-    throw new UnsupportedOperationException("Not implemented yet");
+      PolicyContent content)
+      throws PolicyAlreadyExistsException {
+    NameIdentifier metalakeIdent = NameIdentifierUtil.ofMetalake(metalake);
+    return TreeLockUtils.doWithTreeLock(
+        NameIdentifier.of(NamespaceUtil.ofPolicy(metalake).levels()),
+        LockType.WRITE,
```
jerryshao, line 132: I would suggest to scope the lock area as small as possible to avoid unnecessary perfomance issue.

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
```diff
@@ -74,28 +125,106 @@ public Policy createPolicy(
...
+        });
   }
 
   @Override
   public Policy alterPolicy(String metalake, String policyName, PolicyChange... changes) {
-    throw new UnsupportedOperationException("Not implemented yet");
+    NameIdentifier metalakeIdent = NameIdentifierUtil.ofMetalake(metalake);
+    return TreeLockUtils.doWithTreeLock(
+        NameIdentifier.of(NamespaceUtil.ofPolicy(metalake).levels()),
+        LockType.WRITE,
```
jerryshao, line 172: Lock on the policy itself should be enough.

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
mchades: ok, updated

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
mchades: fixed

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
```diff
@@ -127,4 +251,100 @@ public Policy getPolicyForMetadataObject(
...
   }
+
+  private void changePolicyEnabledState(
+      String metalake, String policyName, boolean expectedEnabledState) {
+    NameIdentifier metalakeIdent = NameIdentifierUtil.ofMetalake(metalake);
+    TreeLockUtils.doWithTreeLock(
+        NameIdentifierUtil.ofPolicy(metalake, policyName),
+        LockType.WRITE,
+        () -> {
+          checkMetalake(metalakeIdent, entityStore);
```
jerryshao: Also here, you can move this out of the lock.

mchades: fixed

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
```diff
@@ -74,28 +123,103 @@ public Policy createPolicy(
...
+                NameIdentifierUtil.ofPolicy(metalake, policyName),
+                PolicyEntity.class,
+                Entity.EntityType.POLICY,
+                policyEntity -> updatePolicyEntity(policyEntity, changes));
+          } catch (NoSuchEntityException e) {
+            throw new NoSuchPolicyException(
+                "Policy with name %s under metalake %s does not exist", policyName, metalake);
+          } catch (EntityAlreadyExistsException e) {
+            throw new RuntimeException(
+                "Policy with name "
```
yuqi1129: The message is incorrect. The `policyName` is the value that lies in `PolicyChange` NOT itself.

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
```diff
@@ -127,4 +251,100 @@ public Policy getPolicyForMetadataObject(
...
+                "Failed to change policy {} enabled state under metalake {}",
+                policyName,
+                metalake,
+                ioe);
+            throw new RuntimeException(ioe);
+          }
+        });
+  }
+
+  private PolicyEntity.Builder newPolicyBuilder(PolicyEntity policyEntity) {
```
yuqi1129, line 287: You'd better place it in class `PolicyEntity` and the name can be `toBuilder()`.

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
mchades: fixed

core/src/main/java/org/apache/gravitino/policy/PolicyManager.java
mchades: I would prefer to keep it as is, since this is not a common operation for PolicyEntity and is only used in this class.

jerryshao:
I don't have further comments, @yuqi1129 can you please review again?

yuqi1129:
> I don't have further comments, @yuqi1129 can you please review again?

I also have no comments anymore.

yuqi1129 (approved):

Issue #7583, opened 2025-07-07
[Improvement] duplicate log handlers in python client


justinmclean:
### What would you like to be improved?

In clients/client-python/gravitino/utils/http_client.py  only configure logging once to prevent duplicate log handlers

### How should we improve?

as above

yunchipang:
i'll pick this up

PR #7627, opened 2025-07-09
[#7583] remove unused logger in python client

yunchipang:

Logging configuration moved to module level, handler added only once when module is imported.

The previous logger implementation in `http_client.py` creates a log handler every time a `Response` object is instantiated, which may cause log message duplication and consume unnecessary memory over time.
Fix: #7583 

No.

Existing tests for `client-python`.

### DIFF
--- a/clients/client-python/gravitino/utils/http_client.py
+++ b/clients/client-python/gravitino/utils/http_client.py
@@ -42,6 +42,11 @@ from gravitino.exceptions.base import RESTException, UnknownError
 from gravitino.exceptions.handlers.error_handler import ErrorHandler
 
 logger = logging.getLogger(__name__)
+if not logger.handlers:
+    logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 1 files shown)


jerryshao:
I found that there's no code actually using this logger, maybe we can remove the code.

yunchipang:
> I found that there's no code actually using this logger, maybe we can remove the code.

@jerryshao thanks for the review! removed.

jerryshao:
Thanks @yunchipang ,  my suggestion is don't blindly fix things as the issue described, it is better to think if there's something more to improve, so that we can do big things. Thanks.

yunchipang:
> Thanks @yunchipang , my suggestion is don't blindly fix things as the issue described, it is better to think if there's something more to improve, so that we can do big things. Thanks.

@jerryshao I’ll definitely be more mindful of this in future changes. Appreciate the reminder!

Issue #7628, opened 2025-07-10
[Improvement] Remove Trino version check


hqbhoho:
### What would you like to be improved?

Now `CatalogRegister#checkTrinoSpiVersion` will throw exception if Trino version not from 435 to 439. This may restrict access for users of other versions, but some features may still function normally in incompatible versions.​

### How should we improve?

A warn log instead of  throw exception.

PR #7629, opened 2025-07-10
[#7628] refactor(trino-connector): Remove Trino version check

hqbhoho:

A warn log instead of throw exception.

Now `CatalogRegister#checkTrinoSpiVersion` will throw exception if Trino version not from 435 to 439. This may restrict access for users of other versions, but some features may still function normally in incompatible versions.​

Fix: #7628 

no

no


trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogRegister.java
```diff
@@ -65,11 +65,9 @@ private void checkTrinoSpiVersion(ConnectorContext context) {
 
     int version = Integer.parseInt(context.getSpiVersion());
     if (version < MIN_SUPPORT_TRINO_SPI_VERSION || version > MAX_SUPPORT_TRINO_SPI_VERSION) {
-      String errmsg =
-          String.format(
-              "Unsupported Trino-%s version. The Supported version for the Gravitino-Trino-connector from Trino-%d to Trino-%d",
-              trinoVersion, MIN_SUPPORT_TRINO_SPI_VERSION, MAX_SUPPORT_TRINO_SPI_VERSION);
-      throw new TrinoException(GravitinoErrorCode.GRAVITINO_UNSUPPORTED_TRINO_VERSION, errmsg);
+      LOG.warn(
+          "Incompatible Trino-%s version. The Compatible version for the Gravitino-Trino-connector from Trino-%d to Trino-%d",
```
yuqi1129: The message can be 'The version %s has not been fully tested with Gravitino, there may be compatiablity problem....'

hqbhoho: done

yuqi1129:
It's fine with me, @diqiu50 , please also take a look.

diqiu50:
I think it would be better to make this configurable.

hqbhoho:
> I think it would be better to make this configurable.

Yeah, Maybe it can  add `gravitino.trino.skip-version-validation` to skip check. So user will know there may be compatiablity problem if enable.

hqbhoho:
@diqiu50 @yuqi1129 PTAL when you have a time. Thanks!

docs/trino-connector/configuration.md
```diff
@@ -13,3 +13,4 @@ license: "This software is licensed under the Apache License version 2."
...
 | gravitino.metadata.refresh-interval-seconds | integer | 10                    | The `gravitino.metadata.refresh-interval-seconds` defines the interval in seconds to refresh metadata from Gravitino server, the default value is 10 seconds.                                                                                                                                                       | No       | 0.9.0         |
+| gravitino.trino.skip-version-validation     | boolean | false                 | The `gravitino.trino.skip-version-validation` defines whether skip Trino version validation or not. Note other versions of Trino have not undergone thorough testing with Gravitino, so there may be compatiablity problem if true.                                                                                 | No       | 1.0.0         |
```
yuqi1129: "Note that Gravitino only supports Trino 425 to 439 and other versions of Trino..."

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogRegister.java
```diff
@@ -64,12 +64,20 @@ private void checkTrinoSpiVersion(ConnectorContext context) {
...
+    if (!skipTrinoVersionValidation) {
+      if (version < MIN_SUPPORT_TRINO_SPI_VERSION || version > MAX_SUPPORT_TRINO_SPI_VERSION) {
+        String errmsg =
+            String.format(
+                "Unsupported Trino-%s version. The Supported version for the Gravitino-Trino-connector from Trino-%d to Trino-%d",
+                trinoVersion, MIN_SUPPORT_TRINO_SPI_VERSION, MAX_SUPPORT_TRINO_SPI_VERSION);
+        throw new TrinoException(GravitinoErrorCode.GRAVITINO_UNSUPPORTED_TRINO_VERSION, errmsg);
+      }
+    } else {
+      LOG.warn(
```
yuqi1129: The message is incorrect. How do you know the version is not between 435 and 439 here?

docs/trino-connector/configuration.md
hqbhoho: done

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogRegister.java
hqbhoho: > The message is incorrect. How do you know the version is not between 435 and 439 here?

The version depends on Trino runtime environment.  `this.trinoVersion = context.getSpiVersion()` wiil return current  Trino version which is running.

yuqi1129: So, when `skipTrinoVersionValidation` is true and the current version is between `435` and `439`, will you still output the warning message 'The version 435 has not undergone thorough testing with Gravitino'?

hqbhoho: > So, when `skipTrinoVersionValidation` is true and the current version is between `435` and `439`, will you still output the warning message 'The version 435 has not undergone thorough testing with Gravitino'?

Oh, I see, you are right!

hqbhoho: done

yuqi1129:
LGTM, @diqiu50 , would you like to take a look?

hqbhoho:
@yuqi1129 @diqiu50 Could you merge it if no issues? Thanks!

trino-connector/trino-connector/src/test/java/org/apache/gravitino/trino/connector/TestGravitinoConnector.java
```diff
@@ -66,6 +66,7 @@ protected QueryRunner createQueryRunner() throws Exception {
       properties.put(
           "catalog.config-dir", queryRunner.getCoordinator().getBaseDataDir().toString());
       properties.put("discovery.uri", queryRunner.getCoordinator().getBaseUrl().toString());
+      properties.put("gravitino.trino.skip-version-validation", "true");
```
diqiu50: Why do we add this line？

diqiu50:
LGTM ，except for the above problem.

trino-connector/trino-connector/src/test/java/org/apache/gravitino/trino/connector/TestGravitinoConnector.java
hqbhoho: No real use. I have remove it.

Issue #7585, opened 2025-07-07
[Improvement] Secure createDatabase


justinmclean:
### What would you like to be improved?

In integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/MySQLContainer.java change the create database SQL to be more secure.

### How should we improve?

One way would be to escape backticks in the database name before executing the CREATE statement. Another would be to check that the name matches `[A-Za-z0-9_]+` or similar.

georgereuben:
I can take this issue

PR #7654, opened 2025-07-10
[#7585] feat: add DB name validation to secure DB creation

georgereuben:
Added DB name validation to the createDatabase method to secure DB creation.

Fix: #7585 

No

Tested with unit tests to validate if invalid names are rejected and the names in the TestDatabaseName enum are accepted


justinmclean (approved):
Thanks for fixing this

Issue #7640, opened 2025-07-10
[Improvement] Missing validate in build method in SparkJobTemplate.java


justinmclean:
### What would you like to be improved?

In api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java add a validation call before constructing the SparkJobTemplate in the build method so that builder parameters are checked.

### How should we improve?

see above

PR #7657, opened 2025-07-10
[#7640] add missing validate call to build method in SparkJobTemplate

yunchipang:

Add a call of the existing `validation` method before object construction in `build`.

`validate` method exists with proper validation logic but is never called during object construction.
Fix: #7640

No.

Existing tests.


justinmclean (approved):
Thanks for fixing this

Issue #7638, opened 2025-07-10
[Improvement] Correct URL encode path in alterModelVersion in GenericModelCatalog.java


justinmclean:
### What would you like to be improved?

In alterModelVersion in clients/client-java/src/main/java/org/apache/gravitino/client/GenericModelCatalog.java the path needs to be URL encoded
```
    ModelVersionResponse resp =
        restClient.put(
            formatModelVersionRequestPath(modelFullIdent) + "/aliases/" + alias,
            req,
            ModelVersionResponse.class,
            Collections.emptyMap(),
            ErrorHandlers.modelErrorHandler());
```

### How should we improve?

alias needs to be URL encoded

su8y:
Hi👋, can i take this issue ? Thanks

PR #7662, opened 2025-07-11
[#7638] improvement(client-java): Correct URL encode path in alterModelVersion in GenericModelCatalog.java

su8y:

Added Missing `RESTUtils.encodeString()` surrounds alias

Fix: #7638

None

Execute existing `cleints:client-java` module test


jerryshao:
Please check all the related codes that may miss encoding for alias in the code, and also not only the java client, but also the python client.

jerryshao:
Besides, when you fix something, it is better to add a UT to cover the fix.

justinmclean (dismissed):
Thanks for fixing this

jerryshao (changes requested):
Before merging this, please check all the related codes to see if there is no similar issue. Besides, you need to add the UT to cover the fix.

justinmclean:
The other code (e.g getModelVersion/deleteModelVersion) already uses RESTUtils encodeString, it was just missing it in this code.

justinmclean:
The Python code also encodes the alias, so there is nothing to be done there.

su8y:
> Before merging this, please check all the related codes to see if there is no similar issue. Besides, you need to add the UT to cover the fix.

Thanks for your time @jerryshao. Sorry, I hadn't considered the Python client as it was outside the initial scope of the issue.

I also realized that while unit tests exist for the code I've worked on, there aren't any specifically for encoding. I'll correct this and upload the changes soon.

@justinmclean I've observed that, in addition to `alterModelVersion()`, other methods within `GenericModelCatalog.java` also lack UT verifying proper `RESTUtils.encodString()`. Should these areas be addressed concurrently?

jerryshao:
Yeah, it is better if you can also add UTs to cover other interfaces.

su8y:
@jerryshao May I ask a question?, I've got unit tests written for the current issue, but I hit some significant code duplication with the existing update model tests. I've worked around it for now by changing the target string used in the test, but I'm wondering if there's a better, more idiomatic way to handle this in our codebase. Any thoughts on that? 🤔

Also, as I was digging in, I noticed quite a few methods using `RESTUtils.encodeString()` don't have dedicated encoding tests, not just in `GenericModelCatalog.java` but elsewhere too. For this PR, should I go ahead and add comprehensive encoding tests for all of those methods, or would it be better to just include the test directly relevant to this issue and then open a separate issue for broader encoding test coverage?

Thanks!!

justinmclean:
Duplicate code in unit tests is usually is not a big issue - see WET vs DRY.

su8y:
> Duplicate code in unit tests is usually is not a big issue - see WET vs DRY.

Got it, thanks for the insight! 🙏

jerryshao:
> @jerryshao May I ask a question?, I've got unit tests written for the current issue, but I hit some significant code duplication with the existing update model tests. I've worked around it for now by changing the target string used in the test, but I'm wondering if there's a better, more idiomatic way to handle this in our codebase. Any thoughts on that? 🤔
> 
> Also, as I was digging in, I noticed quite a few methods using `RESTUtils.encodeString()` don't have dedicated encoding tests, not just in `GenericModelCatalog.java` but elsewhere too. For this PR, should I go ahead and add comprehensive encoding tests for all of those methods, or would it be better to just include the test directly relevant to this issue and then open a separate issue for broader encoding test coverage?
> 
> Thanks!!

I think you can only focus on the model-related interfaces, no need to cover all the other code.

su8y:
I applied `@ParameterizedTest` to the UT to better verify the expected encoded values.
Please let me know if further modifications are needed. Thanks

justinmclean (approved):
Thanks for fixing and adding all of those tests

justinmclean (approved):

justinmclean:
@jerryshao can you approve your requested change

jerryshao:
Hi @su8y can you please confirm do we have similar issue in Python client counterpart?

justinmclean:
There are no similar  issues in the Pythion code that I can see, for example:
```
      resp = self.rest_client.get(
            f"{self._format_model_version_request_path(model_full_ident)}/aliases/"
            f"{encode_string(alias)}",
            error_handler=MODEL_ERROR_HANDLER,
        )
```

Issue #7673, opened 2025-07-11
[Subtask] Add event when getting URI of model version


xiaozcy:
### Describe the subtask

Add event when getting URI of model version.

### Parent issue

#7363

PR #7674, opened 2025-07-11
[#7673] feat(model): Add event when getting URI of model version

xiaozcy:

Add event when getting URI of model version.

Fix: #7673 

No.

UT.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/api/event/RegisterAndLinkModelEvent.java
+++ b/core/src/main/java/org/apache/gravitino/listener/api/event/RegisterAndLinkModelEvent.java
@@ -19,9 +19,11 @@
 
 package org.apache.gravitino.listener.api.event;
 
+import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.listener.api.info.ModelInfo;
+import org.apache.gravitino.model.ModelVersion;
 
 /**
  * Represents an event that is generated after a model is successfully registered and a model
@@ -30,6 +32,24 @@ import org.apache.gravitino.listener.api.info.ModelInfo;
 public class RegisterAndLinkModelEvent extends ModelEvent {
   private final ModelInfo registeredModelInfo;
   private final Map<String, String> uris;
+
+  /**
+   * Constructs a new instance of {@link RegisterAndLinkModelEvent}, capturing the user, identifier,
+   * registered model information, and linked model version information.
+   *
+   * @param user The user responsible for triggering the model operation.
+   * @param identifier The identifier of the Model involved in the operation. This encapsulates some
+   *     information.
+   * @param registeredModelInfo The final state of the model post-creation.
+   * @param uri The URI of the linked model version.
+   */
+  public RegisterAndLinkModelEvent(
+      String user, NameIdentifier identifier, ModelInfo registeredModelInfo, String uri) {
+    super(user, identifier);
+
+    this.registeredModelInfo = registeredModelInfo;
+    this.uris = ImmutableMap.of(ModelVersion.URI_NAME_UNKNOWN, uri);
+  }
 
   /**
    * Constructs a new instance of {@link RegisterAndLinkModelEvent}, capturing the user, identifier,
@@ -59,6 +79,15 @@ public class RegisterAndLinkModelEvent extends ModelEvent {
    */
   public ModelInfo registeredModelInfo() {
     return registeredModelInfo;
+  }
+
+  /**
+   * Retrieves the unknown URI of the linked model version.
+   *
+   * @return the unknown URI of the linked model version
+   */
+  public String uri() {
+    return uris.get(ModelVersion.URI_NAME_UNKNOWN);
   }
 
   /**
--- a/core/src/main/java/org/apache/gravitino/listener/api/info/ModelVersionInfo.java
+++ b/core/src/main/java/org/apache/gravitino/listener/api/info/ModelVersionInfo.java
@@ -20,6 +20,7 @@
 package org.apache.gravitino.listener.api.info;
 
 import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.gravitino.Audit;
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/gravitino/catalog/ModelNormalizeDispatcher.java
```diff
@@ -125,6 +126,18 @@ public void linkModelVersion(
...
+      throws NoSuchModelVersionException, NoSuchModelVersionURINameException {
+    return dispatcher.getModelVersionUri(normalizeCaseSensitive(ident), version, uriName);
+  }
+
+  @Override
+  public String getModelVersionUri(NameIdentifier ident, String alias, String uriName)
+      throws NoSuchModelVersionException, NoSuchModelVersionURINameException {
+    return dispatcher.getModelVersionUri(normalizeCaseSensitive(ident), alias, uriName);
+  }
+
```
jerryshao, line 140: I think the `getModelVersionUri` can be a client-side call, client can call `getModelVersion` to get a model version, these two interfaces are just the wrapper of `getModelVersion`, am I right?

core/src/main/java/org/apache/gravitino/listener/api/event/RegisterAndLinkModelEvent.java
```diff
@@ -58,12 +62,12 @@ public ModelInfo registeredModelInfo() {
...
-   * Retrieves the uri of the linked model version.
+   * Retrieves the URIs of the linked model version.
    *
-   * @return the uri of the linked model version
+   * @return the URIs of the linked model version
    */
-  public String uri() {
-    return uri;
+  public Map<String, String> uris() {
+    return uris;
```
jerryshao, line 99: We should keep the compatibility of thIs class, this class is user-facing. So we should not change this API, we can keep and add a new API, you can refer to fileset's change.

core/src/main/java/org/apache/gravitino/listener/api/info/ModelVersionInfo.java
```diff
@@ -92,12 +92,12 @@ public ModelVersionInfo(
...
-   * Returns the URI of the model version.
+   * Returns the URIs of the model version.
    *
-   * @return the URI of the model version.
+   * @return the URIs of the model version.
    */
-  public String uri() {
-    return uri;
+  public Map<String, String> uris() {
+    return uris;
```
jerryshao, line 133: Also for this class, we should keep the compability of this class.

core/src/main/java/org/apache/gravitino/catalog/ModelNormalizeDispatcher.java
xiaozcy: Yes.

core/src/main/java/org/apache/gravitino/listener/api/event/RegisterAndLinkModelEvent.java
xiaozcy: Got it, I've modified these classes to keep the compatibility.

core/src/main/java/org/apache/gravitino/catalog/ModelNormalizeDispatcher.java
jerryshao: So, why do you still keep this code, I think you don't need to add these two methods in the server side.

xiaozcy: You mean implement it in the client side, or just let users implement it by their own?

I think it's best to implement it in the server side, because this implementation would be universal and doing so will simplify the usage. Besides, it is also convenient to conduct audits on the server side. If the implementation does not meet the requirements, users can call `getModelVersion` and choose the URI by their own.

Could you please tell me your concerns about this?

jerryshao: As I mentioned, these two methods are just the helper method for `getModelVersion`, to help users to get URI conveniently. The user can `getModelVersion` and get URI by name by themselves, or we can provide client-side helper method to get URI simply, but it is not necessary to have a new client-server API for this requirement.

jerryshao: The logic is the same as `getModelVersion` in the server-side, so why do we need to add lots of similar logic in the server side for these two methods? These can be achieved in the client-side.

xiaozcy: There are several disadvantages to do it in the client side:
1. Audits cannot be conducted, but we may need to analyze the access frequency of different URIs based on the audits.
2. It may require two client calls to achieve this, because we support set the `default-uri-name` at the model level, clients may need to call both `getModelVersion` and `getModel` at once.
3. The same logic needs to be implemented in clients of different languages, which leads to repetitive work.

So I still think it's better to do it in the server side, like what we did for fileset.

jerryshao: The 2nd reason seems valid. 

For the first one, if it is only for the audit purpose, the performance problem should also be considered. As we saw the performance penalty of using Fileset with the remote call to translate the location (just to achieve the audit).

Anyway, I'm OK with the current solution, we can go with it.

core/src/main/java/org/apache/gravitino/listener/api/info/ModelVersionInfo.java
```diff
@@ -83,21 +84,53 @@ public ModelVersionInfo(
...
+   * @param aliases The aliases of the model version.
+   * @param auditInfo The audit information of the model version.
+   */
+  public ModelVersionInfo(
+      Map<String, String> uris,
+      String comment,
+      Map<String, String> properties,
+      String[] aliases,
+      Audit auditInfo) {
+    this.uris = uris;
```
jerryshao, line 111: I think you should use this constructor for model version link and other places, right? Will you do it in the next PR?

xiaozcy: Yes, it will be done in the next PR.

jerryshao (approved):

Issue #7584, opened 2025-07-07
[Improvement]  Add rollback logic to the in-memory test store


justinmclean:
### What would you like to be improved?

Implement rollback logic in the in-memory test store so any failed transaction will restore the previous state, thus ensuring tests don’t leave residual entities behind. Code can be found in core/src/test/java/org/apache/gravitino/storage/memory/TestMemoryEntityStore.java.

### How should we improve?

Take a snapshot of the entity map and restore it if an exception occurs.

georgereuben:
I can take this up

justinmclean:
Hi @georgereuben how are you going with this? Do you need any help?

georgereuben:
> Hi @georgereuben how are you going with this? Do you need any help

Done with it @justinmclean , raising a PR in a bit

PR #7687, opened 2025-07-14
[#7584] improvement: Add rollback logic to the in-memory test store

georgereuben:

Add rollback logic to the in-memory test store. A snapshot will be created during transaction execution and will be restored if an exception occurs

Fix: #7584 

No

Tested with unit tests for rollback logic


core/src/test/java/org/apache/gravitino/storage/memory/TestMemoryEntityStore.java
```diff
@@ -166,6 +174,20 @@ public <R, E extends Exception> R executeInTransaction(Executable<R, E> executab
     public void close() throws IOException {
       entityMap.clear();
     }
+
+    public Map<NameIdentifier, Entity> createSnapshot() {
+      try {
+        lock.lock();
```
yuqi1129: L159 and here should be moved out of try block.

georgereuben: updated this, thanks

justinmclean (approved):
Thanks for fixing this

Issue #7631, opened 2025-07-10
[Improvement]  Building multiple AuditInfo objects alters previously built instances


justinmclean:
### What would you like to be improved?

This test (added to TestAuditInfo.java) will fail:
```
  @Test
  public void testBuilderReuseDoesNotAffectBuiltInstance() {
    AuditInfo.Builder builder = AuditInfo.builder().withCreator("first");
    AuditInfo first = builder.build();
    builder.withCreator("second");
    AuditInfo second = builder.build();

    Assertions.assertEquals("first", first.creator());
    Assertions.assertEquals("second", second.creator());
  }
```

### How should we improve?

Create a new AuditInfo instance during the build call.

georgereuben:
I can take this up

PR #7689, opened 2025-07-14
[#7631] improvement: Update AuditInfo builder to create a new instance for each build call

georgereuben:
Updated the AuditInfo builder to create a new AuditInfo instance for each build call. Added a unit test to test this

Building multiple AuditInfo objects alters previously built instances.
Fix: #7631 

No

Added a unit test.


core/src/main/java/org/apache/gravitino/meta/AuditInfo.java
```diff
@@ -225,6 +226,12 @@ public Builder withLastModifiedTime(Instant lastModifiedTime) {
      * @return the constructed and validated {@link AuditInfo} instance.
      */
     public AuditInfo build() {
+      AuditInfo auditInfo = new AuditInfo();
+      auditInfo.creator = this.creator;
+      auditInfo.createTime = this.createTime;
+      auditInfo.lastModifier = this.lastModifier;
+      auditInfo.lastModifiedTime = this.lastModifiedTime;
+
```
justinmclean, line 234: Another approach is to utilise the auditInfo instance and create a new one after the assignment. That would be less code/changes, but I'm not sure that's a better approach. What do you think?

georgereuben: > Another approach is to utilise the auditInfo instance and create a new one after the assignment. That would be less code/changes, but I'm not sure that's a better approach. What do you think?

Thanks for the feedback! I considered that approach initially, but I think the current implementation is better for thread safet yand follows standard builder patterns. Each `build()` call would create a completely independent instance without side effects on the builder’s state. While it’s a few more lines of code, I think it would be better in the long run.

justinmclean: Thanks for that.

Issue #7504, opened 2025-06-30
[Subtask] Framework for job management


jerryshao:
### Describe the subtask

This task focuses on adding the job management framework in core, which is used to mange all the registered jobs in Gravitino.

### Parent issue

#7154

PR #7695, opened 2025-07-14
[#7504] feat(core): Add the job framework for Job System (part-1)

jerryshao:

This is the first part of work to add the job framework for Job System in Gravitino.

Fix: #7504

No.

Add new UTs to cover the code.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
+++ b/api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
@@ -30,6 +30,21 @@ import org.apache.commons.lang3.StringUtils;
  * Represents a job template for executing Spark applications. This class extends the JobTemplate
  * class and provides functionality specific to Spark job templates, including the class name, jars,
  * files, archives, and configurations required for the Spark job.
+ *
+ * <p>Take Spark word count job as an example:
+ *
+ * <p>className: "org.apache.spark.examples.JavaWordCount" executable
+ * "https://example.com/spark-examples.jar" arguments: ["{{input_path}}", "{{output_path}}"]
+ * configs: {"spark.master": "local[*]", "spark.app.name": "WordCount"}
+ *
+ * <p>configs is a map of configuration parameters that will be used by the Spark application. It
+ * can be templated by using placeholders like "{{foo_value}}" and "{{bar_value}}". These
+ * placeholders will be replaced with actual values when the job is executed.
+ *
+ * <p>jars, files, and archives are lists of resources that will be used by the Spark application .
+ * These resources must be accessible to the Gravitino server, and can be located in the local file
+ * system, on a web server (e.g., HTTP, HTTPS, FTP). Distributed file systems like HDFS or S3 will
+ * be supported in the future.
  */
 public class SparkJobTemplate extends JobTemplate {
 
--- a/core/src/main/java/org/apache/gravitino/Configs.java
+++ b/core/src/main/java/org/apache/gravitino/Configs.java
@@ -399,4 +399,11 @@ public class Configs {
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .createWithDefault("caffeine");
+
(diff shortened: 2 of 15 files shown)


core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -0,0 +1,233 @@
...
+
+    String stagingDirPath = config.get(Configs.JOB_STAGING_DIR);
+    this.stagingDir = new File(stagingDirPath);
+    if (stagingDir.exists()) {
+      if (!stagingDir.isDirectory()) {
+        throw new IllegalArgumentException(
+            String.format("Staging directory %s exists but is not a directory", stagingDirPath));
+      }
+
+      if (!stagingDir.canExecute() || !stagingDir.canRead() || !stagingDir.canWrite()) {
```
yuqi1129: !(stagingDir.canExecute() && stagingDir.canRead() && stagingDir.canWrite()) will be more readable.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -0,0 +1,233 @@
...
+import org.apache.gravitino.meta.JobEntity;
+import org.apache.gravitino.meta.JobTemplateEntity;
+import org.apache.gravitino.utils.NameIdentifierUtil;
+import org.apache.gravitino.utils.NamespaceUtil;
+
+public class JobManager implements JobOperationDispatcher, Closeable {
+
+  private final EntityStore entityStore;
+
+  private final File stagingDir;
```
yuqi1129, line 53: So, when will we use this `stagingDir`?

jerryshao: `stagingDir` will be used when running a job, it will be used to store the artifacts, as well as the output for the specific running job. This will be leveraged in the next PR.

jerryshao:
@yuqi1129 @mchades @FANNG1 @jerqi please help to review when you have time, thanks.

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -399,4 +399,11 @@ private Configs() {}
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .createWithDefault("caffeine");
+
+  public static final ConfigEntry<String> JOB_STAGING_DIR =
```
mchades, line 403: So this configuration is set at the server level? Should we consider supporting it at the metalake level? (That means each metalake has its own staging dir)

api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
```diff
@@ -30,6 +30,21 @@
  * Represents a job template for executing Spark applications. This class extends the JobTemplate
  * class and provides functionality specific to Spark job templates, including the class name, jars,
  * files, archives, and configurations required for the Spark job.
+ *
+ * <p>Take Spark word count job as an example:
+ *
+ * <p>className: "org.apache.spark.examples.JavaWordCount" executable
+ * "https://example.com/spark-examples.jar" arguments: ["{{input_path}}", "{{output_path}}"]
```
mchades, line 37: so the `executable` is the actual job resource, not the engine (such as spark, bash, etc.) to execute the job?

If the answer is yes, how does the user specify engine information, such as the Spark version or Spark home?

core/src/main/java/org/apache/gravitino/Configs.java
jerryshao: My design is that: this is a root dir, each metalake can have a sub-directory. Making this a metalake-level configuration can only be achieved in the metalake property, and also makes things complicated.

api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
jerryshao: That needs to be configured in the server side, like specifying the Spark home. I will do this when I implement the Spark executor.

Issue #7706, opened 2025-07-15
[Improvement] Reuse token in OAuth2TokenProvider’s getTokenData method instead of calling getAccessToken twice


justinmclean:
### What would you like to be improved?

Reuse token in clients/client-java/src/main/java/org/apache/gravitino/client/OAuth2TokenProvider.java getTokenData method instead of calling getAccessToken twice

### How should we improve?

see above

arjun7579:
can i work on this?

justinmclean:
Sure if you have any questions just ask.

arjun7579:
@justinmclean can you take a look at it now, i fixed it.

PR #7709, opened 2025-07-15
[#7706] improvement: Reuse token in OAuth2TokenProvider’s getTokenData method instead of calling getAccessToken twice

arjun7579:

Instead of calling the `getAccessToken()` function twice, we store the returned value in a variable and reuse it within the method.

The previous approach was inefficient because it invoked the same function twice. With this change, the function is called only once, and the result is cached in a local variable, improving performance and avoiding redundant method calls.

Fix: #7706


arjun7579:
> Looks good, thanks for looking into this.

always happy to help :smile:

Issue #7705, opened 2025-07-15
[Improvement] Add check for null/empty for partition field names to ensure validation fails early


justinmclean:
### What would you like to be improved?

In common/src/main/java/org/apache/gravitino/dto/rel/PartitionUtils.java the validateFieldExistence method can be improved by adding a Preconditions checkArgument to check for null/empty values for partition field names.

### How should we improve?

see above

raveendra11:
I can contribute on this.

justinmclean:
If you have any questions just ask.

PR #7714, opened 2025-07-15
#7705 - [Improvement] Add check for null/empty for partition field na…

raveendra11:
This PR adds a validation check for the fieldName argument in the PartitionUtils.validateFieldExistence method to ensure it is not null, empty, or blank.
This prevents invalid inputs from causing unexpected behavior later in the validation logic.

Previously, the method assumed fieldName and its first element were valid, which could result in a NullPointerException or invalid behavior if passed a null, empty, or whitespace-only value.

Fix: #7705

No, this is an internal utility method enhancement and does not introduce any user-facing API or configuration change.


common/src/main/java/org/apache/gravitino/dto/rel/PartitionUtils.java
```diff
@@ -38,7 +38,12 @@ public class PartitionUtils {
   public static void validateFieldExistence(ColumnDTO[] columns, String[] fieldName)
       throws IllegalArgumentException {
     Preconditions.checkArgument(ArrayUtils.isNotEmpty(columns), "columns cannot be null or empty");
-
+    Preconditions.checkArgument(
+        fieldName != null
+            && fieldName.length > 0
+            && fieldName[0] != null
+            && !fieldName[0].trim().isEmpty(),
+        "fieldName[0] cannot be null, empty or blank");
```
yuqi1129: Maybe the following is sufficient.
```
Preconditions.checkArgument(ArrayUtils.isNotEmpty(fieldName), "fieldName cannot be null or empty");
```

Anyway,  the current implementation is more strict and I find it acceptable .

yuqi1129:
Please change the PR title and description. If you have any questions, you can refer to other PR materials.
@raveendra11

raveendra11:
@yuqi1129 Could you please check now.

common/src/main/java/org/apache/gravitino/dto/rel/PartitionUtils.java
yuqi1129: I  still suggest you use  
`ArrayUtils.isNotEmpty(fieldName) && StringUtils.isNotBlank(fieldName[0])`

raveendra11: Okay.

raveendra11: I made as suggested. Kindly check again?

common/src/main/java/org/apache/gravitino/dto/rel/PartitionUtils.java
```diff
@@ -39,7 +39,10 @@ public static void validateFieldExistence(ColumnDTO[] columns, String[] fieldNam
       throws IllegalArgumentException {
     Preconditions.checkArgument(ArrayUtils.isNotEmpty(columns), "columns cannot be null or empty");
     Preconditions.checkArgument(
-        ArrayUtils.isNotEmpty(fieldName), "fieldName cannot be null or empty");
+        ArrayUtils.isNotEmpty(fieldName) && StringUtils.isNotBlank(fieldName[0]),
+        "fieldName cannot be null or empty");
+
+    ArrayUtils.isNotEmpty((fieldName), "fieldName cannot be null or empty");
```
yuqi1129: Why do you need to check `fieldName` again, as you have done it in L42

yuqi1129 (approved):

raveendra11:
Hi @yuqi1129 Could you pls run CI again?

Issue #7538, opened 2025-07-01
[FEATURE] Support Azure AD login for OAuth authentication


bharos:
### Describe the feature

Currently, Gravitino in OAuth mode only supports login by passing the client ID, secret and scope.
But we should be able to login via OAuth providers, like Azure AD. 
For example, UI should say login via Azure, and should allow user to login through Azure login and go back to Gravitino UI after successful auth.

### Motivation

Currently users need to use plain username and password than one click login through OAuth providers like Azure

### Describe the solution

We need to support multiple login within OAuth itself. For example, the CLI clients like spark-sql could still use secret based auth, but for WebUI users should be able to use Azure AD authentication login

bharos:
I will take a stab at this. 
Seems like it needs some frontend and backend changes at auth layer. Will try sending a PR.

PR #7727, opened 2025-07-15
[7538] feat(OAuth) : Adding configs for Azure OAuth login

bharos:

Adding configs required for Azure based oAuth login
These will not cause any problems with the current oAuth as these are just additional configs exposed in /configs endpoint

These changes will be used in future PRs for Azure MSAL based oAuth support in the front end code, through /configs endpoint

Fix: #7538 

This does not introduce user facing changes. Only changing the /configs API response for now

Tested in local deployment.
Enabled Azure oauth by
```
gravitino.authenticators = oauth
gravitino.authenticator.oauth.provider = azure
```
and set
gravitino.authenticator.oauth.azure.client-id , gravitino.authenticator.oauth.azure.authority, gravitino.authenticator.oauth.azure.redirect-uri, gravitino.authenticator.oauth.azure.jwks-uri, gravitino.authenticator.oauth.azure.scope

along with existing oauth required parameters (we can modify this in future to make it such that if provider is set, the current oauth params need not be "required")
Configs endpoint returns:
Screenshot 2025-07-22 at 12 05 23 PM

### DIFF
--- a/common/src/main/java/org/apache/gravitino/auth/ProviderType.java
+++ b/common/src/main/java/org/apache/gravitino/auth/ProviderType.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.auth;
+
+/** The type of OAuth provider. */
+public enum ProviderType {
+  /** Default provider. */
+  DEFAULT,
+  /** Azure AD provider. */
+  AZURE
+}
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
@@ -73,4 +73,40 @@ public interface OAuthConfig {
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .create();
+
+  // OAuth provider configs
+  ConfigEntry<String> PROVIDER =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "provider")
+          .doc("The OAuth provider to use (e.g., azure)")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .create();
+
+  ConfigEntry<String> CLIENT_ID =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "client-id")
+          .doc("OAuth client ID used for Web UI authentication")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .create();
+
+  ConfigEntry<String> AUTHORITY =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "authority")
+          .doc("OAuth authority URL (authorization server)")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .create();
+
+  ConfigEntry<String> SCOPE =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "scope")
+          .doc("OAuth scopes (space-separated)")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .create();
+
+  ConfigEntry<String> JWKS_URI =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "jwks-uri")
+          .doc("JWKS URI for token validation")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .create();
 }
--- a/dev/charts/gravitino/resources/config/gravitino.conf
+++ b/dev/charts/gravitino/resources/config/gravitino.conf
@@ -69,6 +69,17 @@ gravitino.authenticator.oauth.defaultSignKey = {{ .Values.authenticator.oauth.de
 gravitino.authenticator.oauth.serverUri = {{ .Values.authenticator.oauth.serverUri }}
 gravitino.authenticator.oauth.tokenPath = {{ .Values.authenticator.oauth.tokenPath }}
 
+gravitino.authenticator.oauth.provider = {{ .Values.authenticator.oauth.provider }}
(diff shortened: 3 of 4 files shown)


bharos:
@jerqi I have this PR to add Azure related configs to the /configs endpoint, when gravitino.authenticator.oauth.provider is set to azure

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -73,4 +73,48 @@ public interface OAuthConfig {
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .create();
+
+  // Extensible OAuth provider configs
+  ConfigEntry<String> PROVIDER =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "provider")
+          .doc("The OAuth provider to use (e.g., azure)")
+          .version(ConfigConstants.VERSION_0_3_0)
```
roryqi: The version should be 1_0_0.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -73,4 +73,48 @@ public interface OAuthConfig {
...
+          .stringConf()
+          .create();
+
+  // Azure-specific configs
+  ConfigEntry<String> AZURE_CLIENT_ID =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "azure.client-id")
```
roryqi: Why do we need `client-id`? I think it may be input by the user.

common/src/main/java/org/apache/gravitino/auth/ProviderType.java
```diff
@@ -0,0 +1,26 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.auth;
+
+/** The type of OAuth provider. */
+public enum ProviderType {
```
roryqi, line 23: Do we need default OAuth provider?

roryqi:
Sorry, we should submit new feature to main branch first. If necessary, I will label the `branch-0.9`, the CI will cherry-pick the commit to branch 0.9 after merging the main branch automatically.

roryqi:
Another consideration point is that we should make Azure OAuth pluggable. We should define interface in the `common` module and add implementation in the `bundles/azure` module.

common/src/main/java/org/apache/gravitino/auth/ProviderType.java
bharos: Added a DEFAULT type in enum

bharos:
> Another consideration point is that we should make Azure OAuth pluggable. We should define interface in the `common` module and add implementation in the `bundles/azure` module.

Yes will consider doing this in future PR where I will add the token validation logic for Azure

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
bharos: We need client ID for Web UI login. I agree that for machine users like spark-sql we don't need this, but here, we expose client-id through /configs API for front end to use it while creating login

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
bharos: Thanks, updated

bharos:
I have simplified the configs. I am thinking that the configs doesn't need to be Azure specific, as other providers also typically use similar configs like client-id, authority etc.. so at least for now, we can keep it generic.
I might add a backend/server side logic for Azure JWT validation if necessary, but for frontend, for now, we can just expose parameters without "Azure" in the config names

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -73,4 +73,40 @@ public interface OAuthConfig {
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .create();
+
+  // OAuth provider configs
+  ConfigEntry<String> PROVIDER =
```
roryqi, line 78: If we add new configurations, we should add documents, too.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
roryqi: If so, Web UI operations will be executed by a user. Is it ok?

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
bharos: @jerqi 
Is it better to add the documentation when we have the server implementation? As adding it now in the doc means those configs won't do anything and it is confusing ?
But if it's ok, I could add it with this PR itself. Or I can add it after the next PR.

My plan for the main PRs is:
- Add configs (this PR)
- Add server side logic for JWKS validation (currently server only supports OAuth based on defaultSignKey, I will add logic to get public key from JWKS_URI)
- Add front-end logic to use MSAL based logic for Web UI (if provider=azure)
- Add documentation

roryqi: OK for me. We can add documents later.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
roryqi: @bharos Gently ping

jerqi (approved):
LGTM.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -73,4 +73,40 @@ public interface OAuthConfig {
...
+
+  ConfigEntry<String> SCOPE =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "scope")
+          .doc("OAuth scopes (space-separated)")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .create();
+
+  ConfigEntry<String> JWKS_URI =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "jwks-uri")
```
roryqi, line 107: Could you use camel style name? Gravitino uses camel style although Iceberg may use `-` style name .

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -73,4 +73,40 @@ public interface OAuthConfig {
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .create();
+
+  // OAuth provider configs
+  ConfigEntry<String> PROVIDER =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "provider")
+          .doc("The OAuth provider to use (e.g., azure)")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
```
jerqi, line 82: What's the default value?

Issue #7701, opened 2025-07-15
[Improvement] Support primary key in MySQL with Trino connector


hqbhoho:
### What would you like to be improved?

`show create table` would display the primary key info of mysql table.

PR #7733, opened 2025-07-16
[#7701] feat(trino-connector): Support show index info with MySQL

hqbhoho:

Support show index info with MySQL

`show create table`  would display the index info of MySQL table.
```sql
-- In MySQL
CREATE TABLE gt_mysql_test_index.demo_with_primary_key_and_unique_key
(
    key1 INT,
    key2 INT,
    col1 INT,
    primary key (key1),
    unique key unique_key1 (key2)
);
-- In Trino 
-- show create table
CREATE TABLE gt_mysql.gt_mysql_test_index.demo_with_primary_key_and_unique_key (
   key1 integer NOT NULL,
   key2 integer,
   col1 integer
)
COMMENT ''
WITH (
   engine = 'InnoDB',
   primary_key = ARRAY['key1'],
   unique_key = ARRAY['unique_key1:key2']
)
```

Fix: #7701

no

local tests


hqbhoho:
@diqiu50 @yuqi1129 Could you PTAL at this PR? Thanks!

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/mysql/MySQLMetadataAdapter.java
```diff
@@ -92,6 +101,53 @@ public GravitinoTable createTable(ConnectorTableMetadata tableMetadata) {
...
+            List<String> columns =
+                Arrays.stream(index.fieldNames())
+                    .flatMap(Arrays::stream)
+                    .collect(Collectors.toUnmodifiableList());
+            uniqueKeys.add(String.format("%s:%s", index.name(), Strings.join(columns, ',')));
+            break;
+        }
+      }
+      if (!primaryKeys.isEmpty()) {
+        properties.put(TABLE_PRIMARY_KEY, primaryKeys);
```
yuqi1129: You'd better make a copy of `properties` as we do not know whether it is an immutable map or not.

yuqi1129 (commented):
Well done. It's okay with me generally except a minor one. 

@diqiu50 , please also help to take a look.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/mysql/MySQLMetadataAdapter.java
hqbhoho: > You'd better make a copy of `properties` as we do not know whether it is an immutable map or not.

the `porperties` come from `CatalogConnectorMetadataAdapter#normalizeProperties` will be a HashMap.
```
  private Map<String, Object> normalizeProperties(
      Map<String, String> properties, List<PropertyMetadata<?>> propertyTemplate) {
    Map<String, Object> validProperties = new HashMap<>();
    for (PropertyMetadata<?> propertyMetadata : propertyTemplate) {
      String name = propertyMetadata.getName();
      if (properties.containsKey(name)) {
        validProperties.put(name, properties.get(name));
      } else {
        LOG.warn("Property {} is not defined in Trino, we will ignore it", name);
      }
    }
    return validProperties;
  }
```

yuqi1129: You can't guarantee that the result will not be modified to an immutable map as the method signature is just `Map<String, String>`

yuqi1129: Another point is that you should not directly change the value of an object from a method unless you are aware of the side effects.

hqbhoho: okk，a copy of properties is more stricter. I will change.

hqbhoho: done

Jackeyzhe:
It's a good work!

hqbhoho:
@yuqi1129 @diqiu50 Could you merge it if no issues? Thanks!

yuqi1129:
@hqbhoho 
Thanks for your contribution.

Issue #7703, opened 2025-07-15
[Improvement]  Equality logic in the Decimal class should check precision


justinmclean:
### What would you like to be improved?

In api/src/main/java/org/apache/gravitino/rel/types/Decimal.java the equals equals method only checks scale. 

### How should we improve?

It should check precision as well. Also add some unit tests to check that this is working correctly.

tarun2684:
Hi @justinmclean, I’d like to tackle it.

I’ll draft a PR shortly with these changes. Please let me know if you had any additional edge cases in mind!
and can you assign this issue to me ?

tarun2684:
Thanks for assigning the issue to me , I'll reach out once I make all the changes .

josephC45:
Hi @justinmclean I noticed this issue was assigned to @tarun2684 a couple of days ago. Since I’ve already implemented a fix, would it be okay if I take this one and open a PR?

tarun2684:
Hi @josephC45 ,
Thanks for the message! I saw this issue was assigned to me recently, and I’d actually like to proceed with it if that’s okay. Appreciate that you've already worked on a fix—if it's alright, I’d prefer to continue with the implementation from my side and open the PR right away.

josephC45:
@tarun2684 Sure thing, no problem! I’ll hang onto my forked changes in case anything shifts, but all good for you to move ahead.

tarun2684:
Hey @josephC45,
Thanks a ton for being so understanding and generous .
Really appreciate you stepping back and giving a beginner like me the chance to contribute — means a lot!
Hope to collaborate again on something soon .

PR #7748, opened 2025-07-17
[#7703] fix(types): Add precision check to Decimal.equals() and unit tests

tarun2684:

This PR modifies the `equals()` method in the `Decimal` class (`Decimal.java`) to include a comparison of the `precision` field in addition to `scale` and `value`.  
This change ensures that two `Decimal` instances are only considered equal if their precision, scale, and value match.  

Additionally, a unit test has been added to `TestTypes.java` to validate the correct behavior of the updated `equals()` method.

Previously, the `equals()` method only compared the `scale` and `value`, ignoring the `precision`.  
This could result in incorrect equality results for two decimal types with the same value and scale, but different precision.  

Fixes: #7703

No, this PR does not introduce any user-facing API or configuration change. It only improves internal logic correctness.

- Added a unit test in `TestTypes.java`:
  - Confirmed that two decimal types with same scale and precision are equal.
  - Confirmed that two decimal types with different precision are not equal.
- If the Gradle build fails due to JDK 21, code correctness was verified manually.


tarun2684:
Thanks for the review @mchades ! I've updated the PR title as requested.
Looking into the CI failures now—I'll push fixes or updates soon once I’ve debugged them.

justinmclean:
You need to run `./gradlew :api:spotlessApply` and push the changes

tarun2684:
Hi @justinmclean ,
I’ve run the Gradle build locally and everything is working fine without any issues. Let me know if there's anything else you'd like me to update.

Thanks!

mchades (approved):
LGTM.

Issue #7702, opened 2025-07-15
[Improvement]  Empty strings should yield empty lists instead of a list containing a blank value


justinmclean:
### What would you like to be improved?

strToSeq in /Users/justinmclean/gravitino/core/src/main/java/org/apache/gravitino/config/ConfigEntry.java will convert empty strings to a list containing a blank value, an empty list would be better.

This test in core/src/test/java/org/apache/gravitino/config/TestConfigEntry.java should help:
```
  @Test
  public void testStrToSeqEmptyString() {
    ConfigEntry<String> conf = new ConfigBuilder("gravitino.test.seq").stringConf().create();
    List<String> result = conf.strToSeq("", s -> s);
    Assertions.assertTrue(result.isEmpty());
  }
```

### How should we improve?

see above

georgereuben:
I can take this up

PR #7755, opened 2025-07-18
[#7702] improvement(core) : Update strToSeq to return empty lists for empty strings

georgereuben:
Update strToSeq to return empty lists for empty strings

Fix: #7702 

No

Added a unit test


yuqi1129:
@georgereuben 
Is this PR ready for review?

core/src/test/java/org/apache/gravitino/config/TestConfigEntry.java
```diff
@@ -168,4 +169,11 @@ public void testCheckValue() {
     Assertions.assertThrows(
         IllegalArgumentException.class, () -> testConfOptional.readFrom(configMap));
   }
+
+  @Test
+  public void testStrToSeqEmptyString() {
+    ConfigEntry<String> conf = new ConfigBuilder("gravitino.test.seq").stringConf().create();
+    List<String> result = conf.strToSeq("", s -> s);
```
yuqi1129, line 176: You may need to test a `null` value.

georgereuben:
> @georgereuben Is this PR ready for review?

Yes, I was facing issues with the `testCheckValue` test, so I have made some changes to make it work with the new strToSeq behavior.

The original test used a list containing only an empty string [""], which serialized to an empty string "". With the new behavior, `strToSeq("") ` returns an empty list [] instead of [""], so validation never runs.

Changed the test to use ["valid", "", "another"] which serializes to "valid,,another". This still tests validation on empty elements while  working with the updated strToSeq behavior.

I have updated the test in the [latest commit](https://github.com/apache/gravitino/pull/7755/commits/bd3d3f1b8ea734513fe8d4c08bd3bb89d98b5d49) . Please let me know if this change looks good

core/src/test/java/org/apache/gravitino/config/TestConfigEntry.java
georgereuben: Added a test to check that null strings return empty lists as well

yuqi1129 (approved):
LGTM

Issue #7766, opened 2025-07-21
[Improvement] Chart: use the sql scripts in the image instead of in the chart folder


danhuawang:
### What would you like to be improved?

Currently, the sql scripts which are used to init the MySQL deployed pod are maintained in chard folder. This way will cause init failed if the image is replaced with the enterprise image.

Use the sql scripts in the image will solve the problem.

### How should we improve?

Use the sql scripts in the image instead of in the chart folder.

PR #7767, opened 2025-07-21
[#7766] Use the sql scripts in the image instead of in the chart

danhuawang:

Launch an init container named sqlfile to mount the scripts path to pod shared volumes.
Launch another init container named init-sql to init the table that Gravitino required.
Remove the scripts folder in chart.

Fix: #7766

N/A

CI


danhuawang:
Wait for rebuilding the `apache/gravitino:1.0.0-SNAPSHOT` image @jerqi   Otherwise the chart CI will failed because of missing `schema-1.0.0-mysql.sql` in the old testing image.

roryqi:
> Wait for rebuilding the `apache/gravitino:1.0.0-SNAPSHOT` image @jerqi Otherwise the chart CI will failed because of missing `schema-1.0.0-mysql.sql` in the old testing image.

Done.

dev/charts/gravitino/values.yaml
```diff
@@ -44,9 +49,7 @@ mysql:
   ##
   image:
     tag: 8.0.36-debian-12-r12
-  ## @param initdbScriptsConfigMap ConfigMap with the initdb scripts
-  ##
-  initdbScriptsConfigMap: gravitino-mysql-init
+  sqlScriptVersion: 1.0.0
```
jerryshao: Developers will easily miss this configuration, which will lead to errors, also it is hard to configure without any instructions. It is better we can figure out the script version automatically.

danhuawang: Fixed.

dev/charts/gravitino/templates/deployment.yaml
```diff
@@ -46,13 +46,81 @@ spec:
...
+              echo "Waiting for Gravitino to be ready..."
+              sleep 2
+          done
+          VERSION=$(/root/gravitino/bin/gcli.sh -s  | grep -oP '[0-9]+\.[0-9]+\.[0-9]+' | head -1)
+          echo $VERSION > /tmp/scripts/version.txt
+        resources:
+          {{- toYaml .Values.initResources | nindent 10 }}
+        volumeMounts:
+          - mountPath: /tmp/scripts/
+            name: scripts-emptydir
```
jerryshao: Can you please explain the meaning of these code, how do you get the version? I'm confused that you need a version to get mysql script. On the contrary, you need to start the gravitino to get the version?

danhuawang: I thought we don't have a static file in the image that record the gravitino server version, especially if I replace the image with the enterprise one. 
Additionally, I attempt to use `GRAVITINO_VERSION` in "gravitino-env.sh" file, however this environment variable in enterprise image is a PLACEHOLDER. 
In order to cover the enterprise scenario, I finally use an init container to temporarily start the gravitino (h2 embedded database) to get the version and save the version to a shared pod volume. This init container will terminated once get version finished.

jerryshao: Let me think if there's a better solution.

danhuawang: @yuqi1129  Could you help take a look at this issue?

yuqi1129: I see.

danhuawang: Fixed. Now use the name of gravitino server jar in libs directory to get the version.

jerryshao (approved):

Issue #7505, opened 2025-06-30
[Subtask] Add job execution interface in Gravitino


jerryshao:
### Describe the subtask

This task focuses on adding the job execution interface in Gravitino.

### Parent issue

#7154

PR #7772, opened 2025-07-21
[#7505] feat(core): Add job execution logic and job framework (part-2)

jerryshao:

This is the second part of the job framework PR to add the job executor interface and complete the job framework.

Fix: #7505 

No.

Add UTs to cover the code changes.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/JobManager.java
+++ b/core/src/main/java/org/apache/gravitino/job/JobManager.java
@@ -278,7 +278,14 @@ public class JobManager implements JobOperationDispatcher, Closeable {
     // Create staging directory.
     // TODO(jerry). The job staging directory will be deleted using a background thread.
     long jobId = idGenerator.nextId();
-    File jobStagingDir = new File(stagingDir, JobHandle.JOB_ID_PREFIX + jobId);
+    String jobStagingPath =
+        stagingDir.getAbsolutePath()
+            + File.separator
+            + metalake
+            + File.separator
+            + JobHandle.JOB_ID_PREFIX
+            + jobId;
+    File jobStagingDir = new File(jobStagingPath);
     if (!jobStagingDir.mkdirs()) {
       throw new RuntimeException(
           String.format("Failed to create staging directory %s for job %s", jobStagingDir, jobId));


core/src/main/java/org/apache/gravitino/connector/job/JobExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @param jobId The unique identifier of the job.
+   * @return The status of the job.
+   * @throws NoSuchJobException If the job with the given identifier does not exist.
+   */
+  JobHandle.Status getJobStatus(String jobId) throws NoSuchJobException;
+
+  /**
+   * Cancel a job by its unique identifier. The job runner should stop the job if it is currently
+   * running. If the job is already completed, it should return directory without any error. If the
```
yuqi1129: directory -> directly.

core/src/main/java/org/apache/gravitino/connector/job/JobExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+   * job does not exist, it should throw a {@link NoSuchJobException}.
+   *
+   * <p>This method should not block on the job cancellation, it should return immediately, and
```
yuqi1129: `This method should not block on the job cancellation` is quite confusing. Should it be `This method should not be blocked...`

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -226,8 +267,243 @@ public JobEntity getJob(String metalake, String jobId) throws NoSuchJobException
...
+                AuditInfo.builder()
+                    .withCreator(PrincipalUtils.getCurrentPrincipal().getName())
+                    .withCreateTime(Instant.now())
+                    .build())
+            .build();
+
+    try {
+      entityStore.put(jobEntity, false /* overwrite */);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to register the job entity " + jobEntity, e);
```
yuqi1129, line 325: Do we need to cancel the job if there are any exceptions while storing to the entity store?

jerryshao: It is hard to keep the consistency for different cases, I will think about it.

jerryshao: I would leave as it is. if the entity store is failed, then it may have some serious issues, trying to keep the consistency in such case seems worthless and makes things more complicated.

jerryshao:
@yuqi1129 @mchades @FANNG1 can you please help to review when you have time?

core/src/main/java/org/apache/gravitino/job/JobExecutorFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private JobExecutorFactory() {
+    // Private constructor to prevent instantiation
+  }
+
+  public static JobExecutor create(Config config) {
+    String jobExecutorName = config.get(Configs.JOB_EXECUTOR);
+    String jobExecutorClassKey =
+        JOB_EXECUTOR_CONF_PREFIX + jobExecutorName + JOB_EXECUTOR_CLASS_SUFFIX;
+    String clzName = config.getRawString(jobExecutorClassKey);
```
mchades, line 42: what's the class name for `local` executor?  You want to add it in another PR, or add an empty class in this PR?

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -226,8 +267,243 @@ public JobEntity getJob(String metalake, String jobId) throws NoSuchJobException
...
+      throws NoSuchJobTemplateException {
+    checkMetalake(NameIdentifierUtil.ofMetalake(metalake), entityStore);
+
+    // Check if the job template exists, will throw NoSuchJobTemplateException if it does not exist.
+    JobTemplateEntity jobTemplateEntity = getJobTemplate(metalake, jobTemplateName);
+
+    // Create staging directory.
+    // TODO(jerry). The job staging directory will be deleted using a background thread.
+    long jobId = idGenerator.nextId();
+    File jobStagingDir = new File(stagingDir, JobHandle.JOB_ID_PREFIX + jobId);
```
mchades: Should the metalake directory be added to the path?  As mentioned in https://github.com/apache/gravitino/pull/7695#discussion_r2218125427

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -226,8 +267,243 @@ public JobEntity getJob(String metalake, String jobId) throws NoSuchJobException
...
+            .withNamespace(NamespaceUtil.ofJob(metalake))
+            .withAuditInfo(
+                AuditInfo.builder()
+                    .withCreator(PrincipalUtils.getCurrentPrincipal().getName())
+                    .withCreateTime(Instant.now())
+                    .build())
+            .build();
+
+    try {
+      entityStore.put(jobEntity, false /* overwrite */);
```
mchades, line 323: It seems that the job and template relation operation is unnecessary?https://github.com/apache/gravitino/blob/f06649a359a8e3bda30f68c571b01cc6374db504/core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java#L40-L41

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -226,8 +267,243 @@ public JobEntity getJob(String metalake, String jobId) throws NoSuchJobException
...
+      throw new RuntimeException(
+          String.format("Failed to create staging directory %s for job %s", jobStagingDir, jobId));
+    }
+
+    // Create a JobTemplate by replacing the template parameters with the jobConf values, and
+    // also downloading any necessary files from the URIs specified in the job template.
+    JobTemplate jobTemplate = createRuntimeJobTemplate(jobTemplateEntity, jobConf, jobStagingDir);
+
+    // Submit the job template to the job executor
+    String jobExecutionId;
```
mchades, line 299: What is the difference between `jobExecutionId` and `jobId`? If `jobExecutionId` is used for tracking, should it also be used as the `jobStagingDir` instead of using `jobId`?

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -226,8 +267,243 @@ public JobEntity getJob(String metalake, String jobId) throws NoSuchJobException
...
+            .collect(
+                Collectors.toMap(
+                    entry -> replacePlaceholder(entry.getKey(), jobConf),
+                    entry -> replacePlaceholder(entry.getValue(), jobConf)));
+
+    // For shell job template
+    if (content.jobType() == JobTemplate.JobType.SHELL) {
+      List<String> scripts = fetchFilesFromUri(content.scripts(), stagingDir, TIMEOUT_IN_MS);
+
+      return new ShellJobTemplate.Builder()
```
mchades: Should we use `ShellJobTemplate.builder()` to align with the project's code style?

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -413,4 +413,14 @@ private Configs() {}
...
           .createWithDefault("/tmp/gravitino/jobs/staging");
+
+  public static final ConfigEntry<String> JOB_EXECUTOR =
+      new ConfigBuilder("gravitino.job.executor")
+          .doc(
+              "The executor to run jobs, by default it is 'local', user can implement their own "
+                  + "executor and set it here.")
+          .version(ConfigConstants.VERSION_1_0_0)
+          .stringConf()
+          .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
```
FANNG1, line 425: remove nullable check here because there is a default `local` option ?

core/src/main/java/org/apache/gravitino/job/JobExecutorFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private static final String JOB_EXECUTOR_CONF_PREFIX = "gravitino.executor.";
+
+  private static final String JOB_EXECUTOR_CLASS_SUFFIX = ".class";
+
+  private JobExecutorFactory() {
+    // Private constructor to prevent instantiation
+  }
+
+  public static JobExecutor create(Config config) {
```
FANNG1, line 38: Should we use an isolated classloader for the job executor to avoid class conflict with Gravitino server?

core/src/main/java/org/apache/gravitino/job/JobExecutorFactory.java
jerryshao: This will be done in the next PR.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: I feel like metalake directory is not so necessary, and the job id is globally unqiue, so I think it is fine not adding the metalake directory. What is your concern about this?

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: This will be used in job listing by job template.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: The `jobExecutionId` is a job id returned by external job executor, which is unique to the external job executor (like Airflow). The `jobExecutionId` is used to get the status from job executor, whereas the `jobId` is a unique id in Gravitino.

core/src/main/java/org/apache/gravitino/job/JobExecutorFactory.java
jerryshao: JobExecutor typically is an executor connecting to the remote job scheduler like Airflow, or local executor (isolated processor), which doesn't have classloader problem. So, we don't have to use isolated classloader.

core/src/main/java/org/apache/gravitino/job/JobManager.java
mchades: In a multi-tenant mode, using the metalake directory can make later maintenance more convenient.

core/src/main/java/org/apache/gravitino/connector/job/JobExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Initialize the job executor with the given configurations.
+   *
+   * @param configs A map of configuration key-value pairs.
+   */
+  void initialize(Map<String, String> configs);
+
+  /**
+   * Submit a job with the given name and job template to the external job runner.
```
yuqi1129, line 46: What's the definition of `external job runner`? Does this include the local runner mentioned in the class comments?

api/src/main/java/org/apache/gravitino/job/JobHandle.java
```diff
@@ -42,6 +42,9 @@ enum Status {
     /** The job has completed successfully. */
     SUCCEEDED,
 
+    /** The job is being cancelled. */
+    CANCELLING,
+
     /** The job has been cancelled. */
     CANCELLED;
```
yuqi1129, line 49: Do we support job statuses like `pause` or `resume`?

core/src/main/java/org/apache/gravitino/connector/job/JobExecutor.java
jerryshao: Yes, local runner would be the one used for test and light-weighted jobs running on the server of the Gravitino. For production use, user can implement the external job runner like Livy, Airflow, etc.

api/src/main/java/org/apache/gravitino/job/JobHandle.java
jerryshao: No, it is not necessary. For most of the jobs like Spark, they don't support `pause` and `resume`, we don't have to handle these two statuses.

core/src/main/java/org/apache/gravitino/Configs.java
jerryshao: I think we still need this check to avoid passing any meaningless configuration value.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: OK, I will modify it.

mchades (commented):
I have no more comments.

jerryshao:
Can you please approve this PR if you have no further comment @yuqi1129 @mchades

mchades (approved):

PR #7776, opened 2025-07-22
fix(iceberg): Fix OAuth user identity propagation in Iceberg create table REST operation

bharos:

Fix IcebergTableOperationExecutor to execute catalog operations within authenticated user's security context using PrincipalUtils.doAs()
- This ensures that when clients (like Spark SQL) connect with OAuth tokens, table creation operations properly record the authenticated user as the creator instead of showing 'N/A'

Fixes issue where OAuth-authenticated table creation showed incorrect creator in audit metadata.

No

Tested by creating tables through Gravitino Web UI, curl command and spark-sql client, all with OAuth enabled.
Created by field now shows up on the UI on the table details

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
@@ -19,8 +19,11 @@
 
 package org.apache.gravitino.iceberg.service.dispatcher;
 
+import org.apache.gravitino.UserPrincipal;
+import org.apache.gravitino.auth.AuthConstants;
 import org.apache.gravitino.iceberg.service.IcebergCatalogWrapperManager;
 import org.apache.gravitino.listener.api.event.IcebergRequestContext;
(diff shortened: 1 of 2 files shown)


FANNG1:
https://github.com/apache/gravitino/pull/7731 could this works for you?

bharos:
> #7731 could this works for you?

@FANNG1 Thanks for pointing to the PR, looks similar. But I guess my branch already has this change and it was not working still.
Anyways, I think this change is required on IcebergRequestContext constructor to get OAuth user from HTTP request attribute

Also, probably the change on IcebergTableOperationExecutor is also needed, as the outer Utils.doAs() in IcebergTableOperations is set on the main thread possibly doesn't propagate to catalog worker threads/connection pools?

FANNG1:
> > #7731 could this works for you?
> 
> @FANNG1 Thanks for pointing to the PR, looks similar. But I guess my branch already has this change and it was not working still. Anyways, I think this change is required on IcebergRequestContext constructor to get OAuth user from HTTP request attribute
> 
> Also, probably the change on IcebergTableOperationExecutor is also needed, as the outer Utils.doAs() in IcebergTableOperations is set on the main thread possibly doesn't propagate to catalog worker threads/connection pools?

Could you confirm #7731 is included in your branch already. and enable auditing `gravitino.audit.enabled = true` to check whether it works.

Issue #7754, opened 2025-07-18
[Improvement] Add a configuration to disable entity store gc


FANNG1:
### What would you like to be improved?

If multiple servers share the same RDMS,  we better provide the ability to disable entity store GC in some server to make only one server start the GC.

### How should we improve?

1. add a config item in `Configs.java` like 
```java
  public static final ConfigEntry<Boolean> ENTITY_STORE_GC_ENABLED =
      new ConfigBuilder(ENTITY_RELATIONAL_JDBC_BACKEND_GC_ENABLED)
          .doc("Whether to enable the GC for Entity store.")
          .version(ConfigConstants.VERSION_1_0_0)
          .booleanConf()
          .createWithDefault(true);
```
2.  enable the gc only when the config is enabled in `RelationalEntityStore.java` or create a mock gc.
```java
    if (config.get(Configs.ENTITY_STORE_GC_ENABLED)) {
      this.garbageCollector = new RelationalGarbageCollector(backend, config);
      this.garbageCollector.start();
    }
```

3. close the gc when startted
```
    if (garbageCollector == null) {
      garbageCollector.close();
    }
```

georgereuben:
I can take this up

PR #7777, opened 2025-07-22
[#7754] improvement(storage) : Add a configuration to disable entity store GC for RelationalEntityStore

georgereuben:
Added a configuration to disable GC for relational entity stores

To provide an option to disable entity store GC in some server to make only one server start the GC in case multiple servers use the same RDMS.

Fix: #7754 

No

N/A

### DIFF
--- a/core/src/main/java/org/apache/gravitino/Configs.java
+++ b/core/src/main/java/org/apache/gravitino/Configs.java
@@ -176,11 +176,11 @@ public class Configs {
           .createWithDefault(DEFAULT_RELATIONAL_JDBC_BACKEND_PATH);
 
   public static final ConfigEntry<Boolean> ENTITY_STORE_GC_ENABLED =
-          new ConfigBuilder(ENTITY_RELATIONAL_JDBC_BACKEND_GC_ENABLED)
-                  .doc("Whether to enable the GC for Entity store.")
-                  .version(ConfigConstants.VERSION_1_0_0)
-                  .booleanConf()
-                  .createWithDefault(true);
+      new ConfigBuilder(ENTITY_RELATIONAL_JDBC_BACKEND_GC_ENABLED)
+          .doc("Whether to enable the GC for Entity store.")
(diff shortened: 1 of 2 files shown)


yuqi1129:
@FANNG1 
Have you reached an agreement that a switch was necessary to address this issue?

FANNG1:
> @FANNG1 Have you reached an agreement that a switch was necessary to address this issue?

Not seeing opponent options,  do you have a better suggestion?

yuqi1129:
> > @FANNG1 Have you reached an agreement that a switch was necessary to address this issue?
> 
> Not seeing opponent options, do you have a better suggestion?

No, other components like statistic collectors also face the same problem, so I wonder if there is a unified solution.

FANNG1:
> > > @FANNG1 Have you reached an agreement that a switch was necessary to address this issue?
> > 
> > 
> > Not seeing opponent options, do you have a better suggestion?
> 
> No, other components like statistic collectors also face the same problem, so I wonder if there is a unified solution.

This may need a long-term, complicated solution, this could be a short-term solution to avoid unexpected results and errrors.

jerryshao:
What's the problem here? Will another GC lead to a wrong result?

FANNG1:
> What's the problem here? Will another GC lead to a wrong result?

I'm not sure whether there will be error results, but I see many error & warning logs

jerryshao:
I agree with @yuqi1129; adding a configuration is easy. However, we should have a thorough solution for multi-server deployment, not a band-aid fix. At least we should identify the errors and figure out reasons.

We haven't claimed that we support multi-active deployment, such problem should be treated as a whole to address this problem.

georgereuben:
Hi @yuqi1129 @FANNG1   , just wanted to confirm if I should continue working on this issue or are we looking for long term solutions? Please let me know

FANNG1:
> Hi @yuqi1129 @FANNG1 , just wanted to confirm if I should continue working on this issue or are we looking for long term solutions? Please let me know

We could wait for long-term solutions.

Issue #7771, opened 2025-07-21
[Improvement] Optimize fileset catalog server side operations


yuqi1129:
### What would you like to be improved?

Currently, there are the following issues present in fileset catalog:

- Unclosed file system instances. Repeatedly creating file system instances will take much resources, we'd better find a way to solve it
- Redundant operation of the single-file logic check. 

### How should we improve?

We can take the following steps to address the issue mentioned above.

-  Split `FileSystemProvider` and do not share it between Gravitino client and Gravitino server
-  Introduce a cache system to reuse `FileSystem` instances. 
-  Remove single-file check logic.

PR #7782, opened 2025-07-22
[#7771] improvement(fileset-catalog): Introduce the cache system for filesystem in FilesetCatalogOperations

yuqi1129:

Add Caffeine to cache file system instance to avoid repeatedly create and close file system.

Creating and closing a file system instance is a heavy operation, and we need to avoid such behaviour.

Fix: #7771 

N/A.

Existing tests.


yuqi1129:
This PR only introduces the cache system and will not remove the check operation for the single file.

yuqi1129:
@jerqi @mchades 
Please help to review this PR, thanks.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -129,8 +130,71 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
...
+            .scheduler(Scheduler.forScheduledExecutorService(scheduler))
+            .build();
+  }
+
+  static class FileSystemCacheKey {
+    private final NameIdentifier ident;
+    private final Map<String, String> conf;
+    private final String currentUser;
+
+    FileSystemCacheKey(NameIdentifier ident, Map<String, String> conf) {
```
mchades: You should take multiple locations into account.

yuqi1129: Do you mean that the files under a fileset are different types or different clusters?
```
fileset1:
location1: hdfs://ip1/path
location2: hdfs://ip2/path
location3: hdfs://ip3/path
```

fileset2
```
location1: file://ip1/path
location2: hdfs://ip2/path
location3: s3://bucket/path
```

mchades: yes

yuqi1129: let me think a bit.

yuqi1129: @mchades 
please review it again.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -129,8 +129,75 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
...
+        new ScheduledThreadPoolExecutor(
+            1,
+            new ThreadFactoryBuilder()
+                .setDaemon(true)
+                .setNameFormat("file-system-cache-for-fileset" + "-%d")
+                .build());
+
+    this.fileSystemCache =
+        Caffeine.newBuilder()
+            .expireAfterAccess(1000 * 60 * 60 /* 1 hour */, TimeUnit.MILLISECONDS)
```
diqiu50: Extract a constant for this value.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -1248,10 +1302,59 @@ private boolean hasCallerContext() {
...
+
+  /**
+   * Extracts the prefix from the given path. The prefix is defined as the scheme and the first
+   * slash after the scheme.
+   *
+   * @param path the path from which to extract the prefix.
+   * @return the prefix of the path, or an empty string if the path is null or empty.
+   */
+  @VisibleForTesting
+  String extractPrefix(String path) {
```
diqiu50: Add some testers for this function

yuqi1129: I added, please see L1823 in `TestFilesetCatalogOperations`

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: OK

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -1248,10 +1302,59 @@ private boolean hasCallerContext() {
...
+    if (path.startsWith("file:/")) {
+      return "file:///";
+    }
+
+    int protocolEnd = path.indexOf("://");
+    if (protocolEnd == -1) {
+      return path;
+    }
+
+    int firstSlash = path.indexOf('/', protocolEnd + 3);
```
mchades: what's the meaning of the 3?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -129,8 +129,75 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
...
+                })
+            .scheduler(Scheduler.forScheduledExecutorService(scheduler))
+            .build();
+  }
+
+  static class FileSystemCacheKey {
+    private final NameIdentifier ident;
+    private final Map<String, String> conf;
+    private final String pathPrefix;
+    private final String currentUser;
```
mchades: Is it necessary to include NameIdentifier in the cache key, since you've already added `conf` to it?

yuqi1129: It looks weird to make a file system instance shared between different filesets, so I added the name identfier to it. Removing it won't make a difference and it's logically correct.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: 3 is the lenght of `://`. For example

for path string 'hdfs://dir1/dir2/file1', protocolEnd will be 4 and 'd' in 'dir1' will be 4 + 3, 7.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: I have no preference. @diqiu50  do you have any suggestions?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
mchades: you'd better use `"://".length`

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
mchades: If there are no side effects, I would suggest removing the NameIdentifier from the cache key to avoid a new FS instance.

Another question: might a closed FS be retrieved from the cache?

yuqi1129: > Another question: might a closed FS be retrieved from the cache?

The connection for a cache instance may timeout due to long-term idleness, however, the instance itself will not close until we manually close it. When an instance with a closed connection is reused, I believe it will try to connect to the server side automatically.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: sure

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: fixed

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -1248,10 +1298,59 @@ private boolean hasCallerContext() {
...
+    if (protocolStart == -1) {
+      return path;
+    }
+
+    int firstSlash = path.indexOf('/', protocolStart + protocolSlash.length());
+    if (firstSlash == -1) {
+      return path + "/";
+    }
+
+    return path.substring(0, firstSlash + 1);
```
mchades: can we use `path.toUri().getAuthority()` to simplify the function logic?

yuqi1129: No,  `path.toUri().getAuthority()` will only get the machine or bucket name. For example:
- `hdfs://127.0.0.1:9000/dir1/dir2/file1`, the result is 127.0.0.1:9000.
- `s3://bucket1/dir1/dir2/file2`, the result is bucket1.
- `gs://bucket1/dir1/dir2/file3`, the result is also bucket1.

yuqi1129: @mchades 
Please help to take a look again.

mchades: I mean the cache key is actually `scheme`, `authority`, `user`, `conf`, which is more meaningful.

yuqi1129: let me think a bit.

yuqi1129: > I mean the cache key is actually `scheme`, `authority`, `user`, `conf`, which is more meaningful.

done

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -129,8 +130,76 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
...
+                    LOG.warn("Failed to close FileSystem instance in cache", e);
+                  }
+                })
+            .scheduler(Scheduler.forScheduledExecutorService(scheduler))
+            .build();
+  }
+
+  static class FileSystemCacheKey {
+    // When the path is a local path without prefixes 'file', the scheme and authority are both null
+    @Nullable private final String scheme;
```
mchades, line 164: Will the formalized local file path also have a null scheme?

yuqi1129: No, a formalized path contains intact `scheme` and `authority`.  This is a protective action, and in some cases, the path has not been formalized.

mchades: what I mean is that the `scheme` should never be null

yuqi1129: Do you mean I need to manually extract the schema from a path that has not been formalized?

yuqi1129: > Do you mean I need to manually extract the schema from a path that has not been formalized?

This could be a tricky and time-consuming process, as formulating a path requires retrieving a file system first.

mchades: So, when the scheme and authority are null, it means a default scheme and authority, not a local file, right?

yuqi1129: Yeah, we support setting the default file system (scheme), and the default one( If it's not customized by the user) is local.

mchades: I see. But the field comment appears to be inaccurate.

yuqi1129: I have fixed the comment.

Issue #7788, opened 2025-07-23
[Bug report] Fix duplicate column when loading tables from external catalogs


TEOTEO520:
### Version

main branch

## Problem

  When loading tables from external catalogs (e.g., Hive), duplicate columns can cause `IllegalStateException: Duplicate key` in
  `TableOperationDispatcher.updateColumnsIfNecessary()`.

  ## Root Cause

  This issue typically occurs when:
  1. In `HiveTableConverter.getColumns()`, `table.getSd().getCols()` contains a partition field (e.g., `log_date`)
  2. The same field also exists in `table.getPartitionKeys()`
  3. This results in duplicate columns in the final `table.columns`

  This situation can arise when different engines (Hive, Spark) operate on the same Iceberg table, potentially causing HMS metadata inconsistencies.

<img width="1425" height="802" alt="Image" src="https://github.com/user-attachments/assets/40418ef8-1d2d-490b-a27b-f7517020e070" />

<img width="1239" height="690" alt="Image" src="https://github.com/user-attachments/assets/724e2499-14c4-4542-9c1a-e6a32216dae7" />

### Error message and/or stacktrace

"type": "RuntimeException",
  "message": "Failed to operate object [new_version_export_bcut_d] operation [LOAD] under [bi_art], reason [Duplicate key (51,BaseColumn(name=log_date, comment=日期分区, dataType=org.apache.gravitino.rel.types.Types$StringType@69a72836, nullable=true, autoIncrement=false, defaultValue=org.apache.gravitino.rel.Column$$Lambda$723/554189346@405ccc23))]",
  "stack": [
    "java.lang.IllegalStateException: Duplicate key (51,BaseColumn(name=log_date, comment=日期分区, dataType=org.apache.gravitino.rel.types.Types$StringType@69a72836, nullable=true, autoIncrement=false, defaultValue=org.apache.gravitino.rel.Column$$Lambda$723/554189346@405ccc23))",
    "\tat java.util.stream.Collectors.lambda$throwingMerger$0(Collectors.java:133)",
    "\tat java.util.HashMap.merge(HashMap.java:1254)",
    "\tat java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)",
    "\tat java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)",
    "\tat java.util.stream.IntPipeline$4$1.accept(IntPipeline.java:250)",
[...]

PR #7789, opened 2025-07-23
[#7788] fix(core): fix duplicate column when load table

TEOTEO520:

  1. Added a merge function when collecting columns from catalog table
  2. Added a merge function when collecting columns from table entity

  When loading tables from external catalogs (e.g., Hive), duplicate columns can cause IllegalStateException: Duplicate key crashes. This commonly occurs when:

  - HiveTableConverter.getColumns() includes a partition field (e.g., log_date) from table.getSd().getCols()
  - The same field also exists in table.getPartitionKeys()
  - This results in duplicate columns in the final table schema
  
Fix: #7788

No

Test added

### DIFF
--- a/catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveTableConverter.java
+++ b/catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveTableConverter.java
@@ -88,6 +88,10 @@ public class HiveTableConverter {
           BUILDER extends BaseColumn.BaseColumnBuilder<BUILDER, COLUMN>, COLUMN extends BaseColumn>
       Column[] getColumns(Table table, BUILDER columnBuilder) {
     StorageDescriptor sd = table.getSd();
+    // Collect column names from sd.getCols() to check for duplicates
+    java.util.Set<String> columnNames =
+        sd.getCols().stream().map(f -> f.getName()).collect(java.util.stream.Collectors.toSet());
+
     return Stream.concat(
             sd.getCols().stream()
                 .map(
@@ -98,6 +102,8 @@ public class HiveTableConverter {
                             .withComment(f.getComment())
                             .build()),
             table.getPartitionKeys().stream()
+                // Filter out partition keys that already exist in sd.getCols()
+                .filter(p -> !columnNames.contains(p.getName()))
                 .map(
                     p ->
                         columnBuilder
--- a/catalogs/hive-metastore-common/src/test/java/org/apache/gravitino/hive/converter/TestHiveTableConverter.java
+++ b/catalogs/hive-metastore-common/src/test/java/org/apache/gravitino/hive/converter/TestHiveTableConverter.java
@@ -1,0 +1,212 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.hive.converter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Arrays;
+import java.util.List;
+import org.apache.gravitino.connector.BaseColumn;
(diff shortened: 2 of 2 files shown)


yuqi1129:
@mchades 
Please help to take a look, I see no problem with it.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -624,12 +624,32 @@ private Pair<Boolean, List<ColumnEntity>> updateColumnsIfNecessary(
...
+                    Collectors.toMap(
+                        p -> p.getRight().name(),
+                        Function.identity(),
+                        (existing, replacement) -> {
+                          LOG.warn(
+                              "Duplicate column name '{}' found at position {} and {}, using the first occurrence",
+                              existing.getRight().name(),
+                              existing.getLeft(),
+                              replacement.getLeft());
+                          return existing;
```
mchades: >  This commonly occurs when:
> 
> HiveTableConverter.getColumns() includes a partition field (e.g., log_date) from table.getSd().getCols()
> The same field also exists in table.getPartitionKeys()
> This results in duplicate columns in the final table schema

As in the case you mentioned, we should fix the logic of the Hive catalog instead of allowing duplicate column names.

TEOTEO520: It's not always the best practice to completely fix the table metadata, for the following reasons:

1. This situation occurs in HMS (Hive Metastore) where, due to certain special table creation circumstances, a field exists in both table.getSd().getCols() and table.getPartitionKeys(). This is not entirely a case of "duplicate fields" in the true sense.

2. Tables in this situation can be used normally - they can be queried successfully with engines like Spark, and metadata information can be retrieved normally using SQL commands like SHOW CREATE TABLE and DESC FORMATTED. As a metadata proxy service, Gravitino should reasonably display the table's metadata information rather than directly reporting an error.

3. In our production practice, there are indeed dozens of online production tables in this situation, and they are being widely used. We prefer Gravitino to be compatible with this situation rather than directly reporting an error.

mchades: > It's not always the best practice to completely fix the table metadata

The point is not to fix the table metadata, but to correct the logic of converting a HiveTable to a GravitinoTable in order to prevent column duplication.

mchades: @TEOTEO520 plz update the PR when you have time, thanks!

TEOTEO520: OK, I'll update this PR today ~

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveTableConverter.java
```diff
@@ -98,6 +102,8 @@ Column[] getColumns(Table table, BUILDER columnBuilder) {
                             .withComment(f.getComment())
                             .build()),
             table.getPartitionKeys().stream()
+                // Filter out partition keys that already exist in sd.getCols()
+                .filter(p -> !columnNames.contains(p.getName()))
```
mchades, line 109: I'm thinking about this: if we filter out duplicated partition keys, the partitioning information becomes incomplete. So, should we filter the duplicated field in cols?

mchades: Think twice—it's the same result, since `getPartitioning` is used to retrieve partitioning information.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveTableConverter.java
```diff
@@ -88,6 +88,10 @@ public static Transform[] getPartitioning(Table table) {
           BUILDER extends BaseColumn.BaseColumnBuilder<BUILDER, COLUMN>, COLUMN extends BaseColumn>
       Column[] getColumns(Table table, BUILDER columnBuilder) {
     StorageDescriptor sd = table.getSd();
+    // Collect column names from sd.getCols() to check for duplicates
+    java.util.Set<String> columnNames =
+        sd.getCols().stream().map(f -> f.getName()).collect(java.util.stream.Collectors.toSet());
```
mchades: plz remove the package name. And the map function can be simplified to `FieldSchema::getName`

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveTableConverter.java
TEOTEO520: The current implementation (prioritizing regular columns) is reasonable because:
1. Regular columns contain the actual data
2. Partitioning information can be obtained separately via getPartitioning() without loss
3. It aligns with Hive's design principles

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveTableConverter.java
TEOTEO520: done, thanks

Issue #3302, opened 2024-05-08
[EPIC] Supports StarRocks catalog


yuqi1129:
### Describe the proposal

Currently, we have made the Doris catalog accessible in Gravitino, as StarRocks is also used by many users, we need to support it as well. 
Since they share the same code bases, most of the features are the same, I think we can 
- Make Doris catalog connect to the StarRocks data source.
-Or we need to provide support for the StarRocks catalog. 

### Task list

- [ ] Support StarRocks catalog in web UI
- [ ] Make StarRocks table-related operation ready to use
- [ ] Check the StarRocks and Doris column type system.

yuqi1129:
> Has this issue started development yet?

Not yet. If you are interested in it, you can give it a try or share your thoughts on it.

zhangxuan183:
OK,thanks

liunaijie:
Hi @yuqi1129, I am interested on this, could you assign it to me, thx~

yuqi1129:
@liunaijie 
Thanks for your participation. The community has also been working on it recently. Please keep me informed about the progress if you have been working on it. Thank you.

liunaijie:
> [@liunaijie](https://github.com/liunaijie) Thanks for your participation. The community has also been working on it recently. Please keep me informed about the progress if you have been working on it. Thank you.

@yuqi1129 Thanks for your replay, I am not started yet. If the community is working on this, I am looking forward on this feature.

 In my company, we use starrocks, hive, mysql, kafka. 
And I find the starrocks is not support, and starrocks is similar with Doris. So I want implement this catalog refer the Doris implement.

yuqi1129:
Yeah, we have not start yet and welcome your contribution.

yuqi1129:
@liunaijie 
I have raised a skeleton PR for StarRocks catalog. Would you like to handle the rest? You can refer to Doris catalog.

yuqi1129:
@taylor12805, @liunaijie, do you guys have time to start working on https://github.com/apache/gravitino/issues/7675 within this week? If you don't have free time, I will do it myself, as we plan to release this feature in 1.0.0 and it's approaching.

liunaijie:
[...]

PR #7792, opened 2025-07-24
[#3302][Sub-Task] StarRocks catalog E2E test

liunaijie:

add StarRocks Catalog Implement

To support StarRocks Catalog.

No

By E2E test

### DIFF
--- a/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/config/JdbcConfig.java
+++ b/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/config/JdbcConfig.java
@@ -20,6 +20,7 @@
 package org.apache.gravitino.catalog.jdbc.config;
 
 import java.util.Map;
+import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Config;
 import org.apache.gravitino.config.ConfigBuilder;
@@ -65,7 +66,7 @@ public class JdbcConfig extends Config {
           .doc("The password of the Jdbc connection")
           .version(ConfigConstants.VERSION_0_3_0)
           .stringConf()
-          .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
+          .checkValue(Objects::nonNull, ConfigConstants.NOT_NULL_ERROR_MSG)
           .create();
 
   public static final ConfigEntry<Integer> POOL_MIN_SIZE =
--- a/core/src/main/java/org/apache/gravitino/Config.java
+++ b/core/src/main/java/org/apache/gravitino/Config.java
@@ -198,7 +198,7 @@ public abstract class Config {
         (k, v) -> {
           String trimmedK = k.trim();
           String trimmedV = v.trim();
-          if (!trimmedK.isEmpty() && !trimmedV.isEmpty()) {
+          if (!trimmedK.isEmpty()) {
             if (predicate.test(trimmedK)) {
               configMap.put(trimmedK, trimmedV);
             }
--- a/integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/StarRocksContainer.java
+++ b/integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/StarRocksContainer.java
@@ -1,0 +1,135 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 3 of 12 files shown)


yuqi1129:
@liunaijie 
Please resolve the conflict if you have free time. Thanks.

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/config/JdbcConfig.java
```diff
@@ -65,7 +66,7 @@ public class JdbcConfig extends Config {
           .doc("The password of the Jdbc connection")
           .version(ConfigConstants.VERSION_0_3_0)
           .stringConf()
-          .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
+          .checkValue(Objects::nonNull, ConfigConstants.NOT_BLANK_ERROR_MSG)
```
liunaijie: The alert message will update to `NOT_NULL_ERROR_MSG`

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/StarRocksContainer.java
```diff
@@ -0,0 +1,179 @@
...
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.Network;
+
+public class StarRocksContainer extends BaseContainer {
+  public static final Logger LOG = LoggerFactory.getLogger(StarRocksContainer.class);
+
+  public static final String DEFAULT_IMAGE = "starrocks/allin1-ubuntu:3.3-latest";
+  public static final String HOST_NAME = "gravitino-ci-starrocks";
+  public static final String USER_NAME = "root";
+  public static final String PASSWORD = "";
```
yuqi1129, line 45: Why not use a meaningful password instead of an empty one? is that any other consideration?

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/StarRocksContainer.java
```diff
@@ -0,0 +1,179 @@
...
+  public static final int FE_HTTP_PORT = 8030;
+  public static final int FE_MYSQL_PORT = 9030;
+
+  private static final String STARROCKS_FE_PATH = "/opt/apache-starrocks/fe/log";
```
yuqi1129: Starrocks is not an Apache project, you can change it to other name.

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/config/JdbcConfig.java
yuqi1129: I am very confused about why we need to make this change.

core/src/main/java/org/apache/gravitino/Config.java
```diff
@@ -198,7 +198,7 @@ public void loadFromMap(Map<String, String> map, Predicate<String> predicate) {
         (k, v) -> {
           String trimmedK = k.trim();
           String trimmedV = v.trim();
-          if (!trimmedK.isEmpty() && !trimmedV.isEmpty()) {
+          if (!trimmedK.isEmpty()) {
```
yuqi1129, line 201: Also here.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/StarRocksContainer.java
liunaijie: Done

core/src/main/java/org/apache/gravitino/Config.java
liunaijie: When set property, empty key is meanless, we can filter it, but maybe the empty value is valid is sometime.  I think we don't need filter it in the framework.

For example I want take a spike for Gravitino. I download it and want to connect my dev mysql, the password is empty string. 
Then I can't connect it as I just use an empty password.

yuqi1129: OK

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/StarRocksContainer.java
yuqi1129: @liunaijie 
Please leave a comment here.

liunaijie: it's default password.  
image

Issue #7793, opened 2025-07-24
[Improvement] Remove unnecessary check for Gravitino fileset operations


yuqi1129:
### What would you like to be improved?

This is the follow-up step for #7771. In the Gravitino fileset operations, some checks, such as single file checks,k are not necessarily performed.

PR #7794, opened 2025-07-24
[#7793] improvement(fileset-catalog): Remove single file check in FilesetCatalogOperations

yuqi1129:

Remove single file check logic in FilesetCatalogOperations.

It's not needed.

Fix: #7793

N/A.

Existing tests.

### DIFF
--- a/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGVFSOperations.java
+++ b/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGVFSOperations.java
@@ -63,14 +63,6 @@ public class DefaultGVFSOperations extends BaseGVFSOperations {
     FileSystem actualFs = getActualFileSystem(gvfsPath, currentLocationName());
     Path actualFilePath =
         getActualFilePath(gvfsPath, currentLocationName(), FilesetDataOperation.OPEN);
-    if (actualFs.getFileStatus(actualFilePath).isDirectory()) {
-      String message =
-          "Cannot open a directory path: "
-              + gvfsPath
-              + ". Please use listStatus or getFileStatus to access directory.";
-      throw new IOException(message);
-    }
-
     return actualFs.open(actualFilePath, bufferSize);
   }
 


jerryshao:
The fix is not complete, you should also fix the gvfs code. Also, you should not allow user to create a fileset with file.

yuqi1129:
> you should also fix the gvfs code

I have added code to disable creating a fileset with a single file. I have no idea what changes should be made to the GVFS and why, regarding the GVFS. From my understanding, GVFS will only be able to read the metadata of the files, since we have disabled single-file fileset, user can free to create sub directories or files under this fileset.

Please let me know if there are any mistakes.

jerryshao:
Check the code of gvfs, it has some speicial logic to handle the single file logic, you should check the code.

yuqi1129:
> Check the code of gvfs, it has some speicial logic to handle the single file logic, you should check the code.

OK, I have checked and it seems that all single-file related logic was in `getFileLocation` and it has been modified.

yuqi1129:
@mchades 
Please help to reivew this PR, thanks.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -447,6 +446,11 @@ public Fileset createMultipleLocationFileset(
 
           filesetPathsBuilder.put(entry.getKey(), formalizePath);
           FileSystem fs = getFileSystemWithCache(formalizePath, conf);
+
+          if (fs.exists(formalizePath) && fs.getFileStatus(formalizePath).isFile()) {
+            throw new RuntimeException("Fileset location cannot be a file: " + formalizePath);
+          }
```
mchades: should we also check the location of catalog/schema when creating a catalog/schema

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -715,10 +701,6 @@ public String getFileLocation(NameIdentifier ident, String subPath, String locat
           case RENAME:
             // Fileset only mounts a single file, the storage location of the fileset cannot be
             // renamed; Otherwise the metadata in the Gravitino server may be inconsistent.
```
mchades: the code comments also need to be removed?

BTW, this change breaks compatibility. We should also update the user documentation.

mchades (commented):
since we add the limitation: **Fileset location cannot be a file**

We can also add some validation in GVFS before performing actual file system operations. For example, if a user tries to open `gvfs://fileset/catalog/schema/fileset`, we can throw an error immediately.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: > BTW, this change breaks compatibility. We should also update the user documentation.

Yeah, we need to update the doc and explicitly disable create fileset with a file.

yuqi1129:
> For example, if a user tries to open gvfs://fileset/catalog/schema/fileset, we can throw an error immediately.

This part has been checked in `getFileLocation` in the open method of `FileSystem`

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: ok

yuqi1129: Since we have supported a placeholder, we won't be able to get the detailed path in the schema/catalog level. 
Is it acceptable that we only check the path if it doesn't have a placeholder?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: done

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -1057,6 +1044,20 @@ private Map<String, Path> getAndCheckCatalogStorageLocations(Map<String, String>
             }
 
             checkPlaceholderValue(v);
+
+            if (!containsPlaceholder(v)) {
+              Path path = new Path(v);
+              FileSystem fs = getFileSystemWithCache(path, conf);
+              try {
+                if (fs.exists(path) && fs.getFileStatus(path).isFile()) {
+                  throw new RuntimeException("Fileset catalog location cannot be a file: " + v);
```
mchades: This check for catalog location seems unnecessary, as the schema will also perform the check, and the catalog will not create a directory for this location. Instead, it will introduce additional FS operations during the catalog's creation.

mchades (commented):
Missing changes to GVFS？

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
mchades: Not done. Code comments were not removed.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: > This check for catalog location seems unnecessary, as the schema will also perform the check, and the catalog will not create a directory for this location.

We can make exceptions in advance if the location for a catalog exists and it's a file.

yuqi1129:
> Missing changes to GVFS？

What do you mean by this? GVFS will only use `getFileLocation`  to get file location and there seems not need further changes for GVFS.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
mchades: if you want to keep these code, you need to check `disableFSOps` first

mchades:
> > Missing changes to GVFS？
> 
> What do you mean by this? GVFS will only use `getFileLocation` to get file location and there seems not need further changes for GVFS.

see https://github.com/apache/gravitino/pull/7794#pullrequestreview-3090288712

yuqi1129:
> > > Missing changes to GVFS？
> > 
> > 
> > What do you mean by this? GVFS will only use `getFileLocation` to get file location and there seems not need further changes for GVFS.
> 
> see [#7794 (review)](https://github.com/apache/gravitino/pull/7794#pullrequestreview-3090288712)

I'm not so sure that whether the point you address is here:

image

What's the meaning of adding this check here since `open` will so do such check here.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGVFSOperations.java
```diff
@@ -63,6 +63,9 @@ public FSDataInputStream open(Path gvfsPath, int bufferSize) throws IOException
     FileSystem actualFs = getActualFileSystem(gvfsPath, currentLocationName());
     Path actualFilePath =
         getActualFilePath(gvfsPath, currentLocationName(), FilesetDataOperation.OPEN);
+    if (actualFs.exists(actualFilePath) && !actualFs.getFileStatus(actualFilePath).isFile()) {
```
mchades: I mean that we can add a check before L63:
if the gvfsPath matches the pattern `[gvfs://fileset]/*/*/*[/]`, we can throw an error in advance

yuqi1129: It can throw exceptions and solve the problem partially. 
For example, If the path `gvfs://filest/*/*/directory` and without `/` in suffix, we can't tell whether it a file or directory, of course if it ends with `/`, it's definitely a directory.

If you believe it's necessary, I can make the necessary changes.

mchades: `open`, `create`, and `append` can only be performed on a file, due to the limitation that only a directory can be mounted on a fileset. Therefore, if the `gvfsPath` points to a fileset location, we should throw an error early.

yuqi1129: Cloud storage will behave differently compared to HDFS. For example, OSS

image



So, how can we add such limitations to GVFS?  What if the detail storage have different behavior?

mchades: Sorry, I was wrong before. We should handle `open` and `append`, not `create`

yuqi1129: I'm hesitant to do it as I'm not 100% sure whether all storage that supports HCFS shares the same properties, that is: 

It's prohibited to open a directory and append a directory 

Another point is that I think the check needs to be passed by, and let the specific storage handle it as we can't tell the difference between different storage.

mchades: @jerryshao WDYT on this?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -447,6 +446,11 @@ public Fileset createMultipleLocationFileset(
 
           filesetPathsBuilder.put(entry.getKey(), formalizePath);
           FileSystem fs = getFileSystemWithCache(formalizePath, conf);
+
+          if (fs.exists(formalizePath) && fs.getFileStatus(formalizePath).isFile()) {
+            throw new RuntimeException("Fileset location cannot be a file: " + formalizePath);
```
jerryshao: You'd better throw an "IllegalArgumentException", as it is mainly due to user pass in a wrong location.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -769,6 +748,11 @@ public Schema createSchema(NameIdentifier ident, String comment, Map<String, Str
           if (schemaPath != null && !containsPlaceholder(schemaPath.toString())) {
             try {
               FileSystem fs = getFileSystemWithCache(schemaPath, conf);
+              if (fs.exists(schemaPath) && fs.getFileStatus(schemaPath).isFile()) {
+                throw new RuntimeException(
```
jerryshao: Also here.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGVFSOperations.java
```diff
@@ -60,6 +60,14 @@ public void close() throws IOException {
 
   @Override
   public FSDataInputStream open(Path gvfsPath, int bufferSize) throws IOException {
+    if (gvfsPath.toString().endsWith("/")) {
+      String message =
+          "Cannot open a directory path: "
+              + gvfsPath
+              + ". Please use listStatus or getFileStatus to access directory.";
+      throw new IOException(message);
+    }
```
jerryshao: Use "/" to check if it is a folder seems unstable?

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGVFSOperations.java
jerryshao: I think wrong operation will be failed finally, we don't have to do the strict check beforehand. My concern is like @yuqi1129 mentioned, we should not rely on the path to do the strict check, because the behavior of different storages are different (and we don't know the expected behavior). We don't have to the over-check, we can let the underlying storage to check.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGVFSOperations.java
yuqi1129: It's hard to use raw string to identify whether it is a directory or not, and I will use `fs.getFileStatus().isDirectory` instead.

mchades: as discussed https://github.com/apache/gravitino/pull/7794#discussion_r2269309728, we can remove this check too

yuqi1129: removed

mchades (approved):

Issue #7807, opened 2025-07-26
[Bug report] Incorrect column not null limitation in unique index


hqbhoho:
### Version

main branch

### Describe what's wrong

As #5721 mentioned, `autoIncrement`  column can not be null rather than `unique index` column.
```shell
mysql> CREATE TABLE t3 (  c1 INT NULL , UNIQUE KEY (c1) );
Query OK, 0 rows affected (0.00 sec)

mysql> show create table t3;
+-------+------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                             |
+-------+------------------------------------------------------------------------------------------------------------------------------------------+
| t3    | CREATE TABLE `t3` (
  `c1` int DEFAULT NULL,
  UNIQUE KEY `c1` (`c1`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> CREATE TABLE t4 (  c1 INT NULL AUTO_INCREMENT, UNIQUE KEY (c1) );
Query OK, 0 rows affected (0.01 sec)

mysql> show create table t4;
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                        |
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------+
| t4    | CREATE TABLE `t4` (
  `c1` int NOT NULL AUTO_INCREMENT,
  UNIQUE KEY `c1` (`c1`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+-----------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)


```

### Error message and/or stacktrace

N/A

### How to reproduce

N/A

### Additional context

N/A

PR #7808, opened 2025-07-26
[#7807] improvement(mysql-catalog): Add index column not null limitation

hqbhoho:
- The `primary key` must be `not null`
-  The auto increment `unique key` must be `not null`

Fix: #7807

no

local tests


hqbhoho:
@yuqi1129 @mchades  PTAL when you have a time. Thanks!

yuqi1129:
@hqbhoho 
Do we need to modify other JDBC catalogs like PG?

hqbhoho:
> @hqbhoho Do we need to modify other JDBC catalogs like PG?

@yuqi1129  Yes, we need do, I think  we should implement abstract methods to validate all attributes of columns.​ Like
```
validateIndexes(indexes)
validateIncrementCol(columns, indexes);
validatePartitioning();
validateDistribution();
```
 Maybe different catalog have different constraint.

I'll review and refactor the code later when I have time.​

hqbhoho:
@yuqi1129 @mchades 
Hi, is there anything else that needs to be modified? 
Currently, create mysql table with unique index column which must be `not null`.

yuqi1129:
> @yuqi1129 @mchades Hi, is there anything else that needs to be modified? Currently, create mysql table with unique index column which must be `not null`.

No, the current is generally okay with me, and you can use another PR to refine it later.

yuqi1129 (approved):

PR #7812, opened 2025-07-26
[#7823] feat(server): upgrade Gravitino server and Iceberg REST server mini JDK version to 17

FANNG1:
1. Upgrade Gravitino server and Iceberg REST server minimal JDK version to 17
2. Adjust the Integration test for the current module:
    a.  JDK17  module: support JDK17 for embedded and deploy mode. 
    b.  Non-JDK17 module:  support JDK17 for deploy mode only.
3. Move integration test from `catalog-model` and `client-java` to `integration-test` module, which supports JDK17 in embedded mode and JDK 7,11,17 in deploy mode.

JDK17 modules: server-related modules:  `:server`, `:iceberg`, `:catalogs`, `:lineage`, `authorizations`.  
`:catalogs:catalogs-fileset` `:catalogs:hadoop-common` is the special case because `bundles` modle is JDK 8 compatible and relies on them, so they are JDK 8, not 17.

Fix: #7823 

adding ITs


FANNG1:
As disscussed offline, I propose a new PR #7827  to [upgrade_jdk](https://github.com/apache/gravitino/tree/upgrade_jdk) branch not main branch.

Issue #7743, opened 2025-07-17
[Subtask] Add local job executor to the job system


jerryshao:
### Describe the subtask

To support running the light-weighted job locally, also for the convenience of testing, this task will introduce the local job executor framework to Gravitino's job system.

### Parent issue

#7154

PR #7814, opened 2025-07-28
[#7743]feat(core): Add Local job executor and shell processor builder

jerryshao:

Add the local job executor and shell process builder for Gravitino's job system.

Fix: #7743 

No.

Add UTs to cover the code.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
+++ b/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
@@ -217,7 +217,7 @@ public class LocalJobExecutor implements JobExecutor {
   public void close() throws IOException {
     // Mark the executor as finished to stop processing jobs
     this.finished = true;
-    jobExecutorService.shutdownNow();
+    jobPollingExecutorService.shutdownNow();
 
     if (!waitingQueue.isEmpty()) {
       LOG.warn(
@@ -225,8 +225,6 @@ public class LocalJobExecutor implements JobExecutor {
           waitingQueue.size());
       waitingQueue.clear();
     }
-
-    jobPollingExecutorService.shutdownNow();
 
     // Stop the running jobs
     runningProcesses.forEach(
@@ -236,6 +234,8 @@ public class LocalJobExecutor implements JobExecutor {
           }
         });
     runningProcesses.clear();
+
+    jobExecutorService.shutdownNow();
 
     // Stop the job status cleanup executor
     jobStatusCleanupExecutor.shutdownNow();
@@ -280,10 +280,8 @@ public class LocalJobExecutor implements JobExecutor {
       LOG.error("Error while executing job", e);
       // If an error occurs, we should mark the job as failed
       synchronized (lock) {
-        if (jobPair != null) {
-          String jobId = jobPair.getLeft();
-          jobStatus.put(jobId, Pair.of(JobHandle.Status.FAILED, System.currentTimeMillis()));
-        }
+        String jobId = jobPair.getLeft();
+        jobStatus.put(jobId, Pair.of(JobHandle.Status.FAILED, System.currentTimeMillis()));
       }
     }
   }
--- a/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
+++ b/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
@@ -37,6 +37,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
```diff
@@ -0,0 +1,300 @@
...
+              Thread thread = new Thread(runnable);
+              thread.setName("LocalJobStatusCleanup-" + thread.getId());
+              thread.setDaemon(true);
+              return thread;
+            });
+    jobStatusCleanupExecutor.scheduleAtFixedRate(
+        this::cleanupJobStatus,
+        jobStatusCleanupIntervalInMs,
+        jobStatusCleanupIntervalInMs,
+        java.util.concurrent.TimeUnit.MILLISECONDS);
```
yuqi1129: Remove `java.util.concurrent.`

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
```diff
@@ -0,0 +1,300 @@
...
+            maxRunningJobs,
+            runnable -> {
+              Thread thread = new Thread(runnable);
+              thread.setName("LocalJobExecutor-" + thread.getId());
+              thread.setDaemon(true);
+              return thread;
+            });
+    // Start the job executor service to run jobs
+    for (int i = 0; i < maxRunningJobs; i++) {
+      jobExecutorService.submit(this::runJob);
```
yuqi1129: In this way, there will always `maxRunningJobs` running threads even there is no jobs at all.

I suggest we use a thread to monitor the queue and use the thread pool the handle the job

// one thread
```java
while (isRunnning) {
     try {
        jobPair = waitingQueue.poll(3000, TimeUnit.MILLISECONDS);
        if (jobPair == null) {
          // If no job is available, continue to the next iteration
          continue;
        }
         
        threadpool.submit(jobPair)
      }
}

```

yuqi1129 (approved):

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
jerryshao: Let me think about it.

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
```diff
@@ -0,0 +1,320 @@
...
+            runnable -> {
+              Thread thread = new Thread(runnable);
+              thread.setName("LocalJobExecutor-" + thread.getId());
+              thread.setDaemon(true);
+              return thread;
+            });
+
+    this.jobStatus = Maps.newHashMap();
+
+    this.jobStatusKeepTimInMs =
```
yuqi1129: jobStatusKeepTimInMs -> jobStatusKeepTimeInMs

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutorConfigs.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public static final String LOCAL_JOB_EXECUTOR_NAME = "local";
+
+  public static final String WAITING_QUEUE_SIZE = "waitingQueueSize";
+  public static final int DEFAULT_WAITING_QUEUE_SIZE = 100;
+
+  public static final String MAX_RUNNING_JOBS = "maxRunningJobs";
+  public static final int DEFAULT_MAX_RUNNING_JOBS =
+      Math.min(Runtime.getRuntime().availableProcessors() / 2, 10);
```
yuqi1129: The value of `Runtime.getRuntime().availableProcessors()/ 2` will be 0 if we running Gravitino server in single core machine like container, do we need to handle it?

core/src/main/java/org/apache/gravitino/job/local/ShellProcessBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    List<String> commandList = Lists.newArrayList(shellJobTemplate.executable());
+    commandList.addAll(shellJobTemplate.arguments());
+
+    ProcessBuilder builder = new ProcessBuilder(commandList);
+    builder.directory(workingDirectory);
+    builder.environment().putAll(shellJobTemplate.environments());
+
+    File outputFile = new File(workingDirectory, "output.log");
```
yuqi1129, line 49: Do we need to add a method like `clear` to remove those files after the process is done successfully.

jerryshao: I will do this in the separate PR.

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutorConfigs.java
jerryshao: I will fix it.

yuqi1129 (approved):

Issue #7145, opened 2025-05-09
[Subtask] Server Implementation for Policy (part-1)


mchades:
### Describe the subtask

Integrate policy management features into the server-side framework.

### Parent issue

#7139

LindaSummer:
Hi @mchades ,

I'm very interested in this epic.😊

Could this issue be assigned to me?

It seems that there are some dependecies need to finish.

If you do't mind, please let me know if this issue is ready for me to get involved in.

Best Regards,
Edward

mchades:
> It seems that there are some dependecies need to finish.

yes.



> If you do't mind, please let me know if this issue is ready for me to get involved in.

sure. I will ping you at the appropriate time. Thanks for your passion！

mchades:
@LindaSummer Sorry, I'll take this part myself since there are some dependent PRs under review, and I want to catch up on the progress.

LindaSummer:
> [@LindaSummer](https://github.com/LindaSummer) Sorry, I'll take this part myself since there are some dependent PRs under review, and I want to catch up on the progress.

Hi @mchades ,

Thanks for your kind reminder.
No problem, and I will try to take on some other tasks. 😄

Best Regards,
Edward

PR #7821, opened 2025-07-28
[#7145] feat(policy): implement the policy management on server-side (part-1)

mchades:

implement the policy management on server-side

Fix: #7145 

yes, RESTFul APIs added

tests addded

### DIFF
--- a/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
@@ -64,40 +64,4 @@ public interface PolicyContentDTO extends PolicyContent {
       // no validation needed for custom content
     }
   }
-
-  /** Represents a data compaction policy content DTO. */
-  @EqualsAndHashCode
-  @ToString
-  @Builder(setterPrefix = "with")
-  @AllArgsConstructor(access = lombok.AccessLevel.PRIVATE)
-  class DataCompactionContentDTO implements PolicyContentDTO {
-
-    @JsonProperty("targetSizeBytes")
-    private long targetSizeBytes;
-
-    @JsonProperty("properties")
-    private Map<String, String> properties;
-
-    // Default constructor for Jackson deserialization only.
-    private DataCompactionContentDTO() {}
-
-    /**
-     * Returns the target size in bytes for data compaction.
-     *
-     * @return the target size in bytes.
-     */
-    public long targetSizeBytes() {
-      return targetSizeBytes;
-    }
-
-    @Override
-    public Map<String, String> properties() {
-      return properties;
-    }
-
-    @Override
-    public void validate() throws IllegalArgumentException {
-      // no validation needed for custom content
-    }
-  }
 }
--- a/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
@@ -20,6 +20,7 @@ package org.apache.gravitino.dto.policy;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Map;
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
@@ -32,6 +33,7 @@ public interface PolicyContentDTO extends PolicyContent {
   @EqualsAndHashCode
   @ToString
   @Builder(setterPrefix = "with")
+  @AllArgsConstructor(access = lombok.AccessLevel.PRIVATE)
   class CustomContentDTO implements PolicyContentDTO {
 
     @JsonProperty("customRules")
@@ -50,6 +52,42 @@ public interface PolicyContentDTO extends PolicyContent {
      */
     public Map<String, Object> customRules() {
       return customRules;
+    }
+
+    @Override
+    public Map<String, String> properties() {
+      return properties;
+    }
+
+    @Override
+    public void validate() throws IllegalArgumentException {
+      // no validation needed for custom content
+    }
+  }
+
+  /** Represents a data compaction policy content DTO. */
+  @EqualsAndHashCode
+  @ToString
+  @Builder(setterPrefix = "with")
+  @AllArgsConstructor(access = lombok.AccessLevel.PRIVATE)
+  class DataCompactionContentDTO implements PolicyContentDTO {
+
+    @JsonProperty("targetSizeBytes")
+    private long targetSizeBytes;
+
+    @JsonProperty("properties")
+    private Map<String, String> properties;
+
+    // Default constructor for Jackson deserialization only.
+    private DataCompactionContentDTO() {}
+
+    /**
+     * Returns the target size in bytes for data compaction.
+     *
+     * @return the target size in bytes.
+     */
+    public long targetSizeBytes() {
+      return targetSizeBytes;
     }
 
     @Override
--- a/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
@@ -20,7 +20,6 @@ package org.apache.gravitino.dto.policy;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Map;
-import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
@@ -33,7 +32,6 @@ public interface PolicyContentDTO extends PolicyContent {
   @EqualsAndHashCode
   @ToString
   @Builder(setterPrefix = "with")
-  @AllArgsConstructor
   class CustomContentDTO implements PolicyContentDTO {
 
     @JsonProperty("customRules")
--- a/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.dto.policy;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import org.apache.gravitino.policy.PolicyContent;
+
+/** Represents a Policy Content Data Transfer Object (DTO). */
+public interface PolicyContentDTO extends PolicyContent {
+
+  /** Represents a custom policy content DTO. */
+  @EqualsAndHashCode
+  @ToString
+  @Builder(setterPrefix = "with")
+  @AllArgsConstructor
+  class CustomContentDTO implements PolicyContentDTO {
+
+    @JsonProperty("customRules")
+    private Map<String, Object> customRules;
+
+    @JsonProperty("properties")
+    private Map<String, String> properties;
+
+    // Default constructor for Jackson deserialization only.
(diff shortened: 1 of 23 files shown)


mchades:
@jerryshao @yuqi1129 Could you please review this PR when you have time? Thanks!

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.gravitino.policy.PolicyContent;
+
+/** Represents a Policy Content Data Transfer Object (DTO). */
+public interface PolicyContentDTO extends PolicyContent {
+
+  /** Represents a custom policy content DTO. */
+  @EqualsAndHashCode
+  @ToString
+  @Builder(setterPrefix = "with")
+  @AllArgsConstructor
```
jerryshao: Do we need this `@AllArgsConstructor`? I don't think we need this since we already have a builder.

common/src/main/java/org/apache/gravitino/dto/policy/PolicyDTO.java
```diff
@@ -0,0 +1,298 @@
...
+  @JsonTypeInfo(
+      use = JsonTypeInfo.Id.NAME,
+      include = JsonTypeInfo.As.EXTERNAL_PROPERTY,
+      property = "policyType",
+      defaultImpl = PolicyContentDTO.CustomContentDTO.class)
+  @JsonSubTypes({
+    // add mappings for built-in types here
+    // For example: @JsonSubTypes.Type(value = DataCompactionContent.class, name =
+    // "system_data_compaction")
+  })
```
jerryshao, line 71: Does it work as we don't have built-in policy for now?

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -0,0 +1,273 @@
...
+      @QueryParam("details") @DefaultValue("false") boolean verbose) {
+    LOG.info(
+        "Received list policy {} request for metalake: {}", verbose ? "infos" : "names", metalake);
+
+    try {
+      return Utils.doAs(
+          httpRequest,
+          () -> {
+            if (verbose) {
+              Policy[] policies = policyDispatcher.listPolicyInfos(metalake);
```
yuqi1129, line 89: `policies` will never be null, I think we can simplify the code below.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -0,0 +1,273 @@
...
+      return ExceptionHandlers.handlePolicyException(OperationType.GET, name, metalake, e);
+    }
+  }
+
+  @PUT
+  @Path("{policy}")
+  @Produces("application/vnd.gravitino.v1+json")
+  @Timed(name = "alter-policy." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "alter-policy", absolute = true)
+  public Response alterPolicy(
```
yuqi1129, line 181: There are two methods name `alterPolicy`, I believe it's more suitble to differentate them.

common/src/main/java/org/apache/gravitino/dto/util/DTOConverters.java
```diff
@@ -1188,4 +1244,25 @@ public static JobTemplate fromDTO(JobTemplateDTO jobTemplateDTO) {
...
+      return null;
+    }
+
+    if (policyContentDTO instanceof PolicyContentDTO.CustomContentDTO) {
+      PolicyContentDTO.CustomContentDTO customContentDTO =
+          (PolicyContentDTO.CustomContentDTO) policyContentDTO;
+      return PolicyContents.custom(customContentDTO.customRules(), customContentDTO.properties());
+    }
+
+    throw new IllegalArgumentException(
```
yuqi1129, line 1265: Why does `toDTO` supports two cases `PolicyContentDTO` and `olicyContents.CustomContent` and one here?

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
```diff
@@ -0,0 +1,170 @@
...
+      defaultImpl = PolicyContentDTO.CustomContentDTO.class)
+  @JsonSubTypes({
+    // add mappings for built-in types here
+    // For example: @JsonSubTypes.Type(value = DataCompactionContent.class, name =
+    // "system_data_compaction")
+  })
+  private final PolicyContentDTO policyContent;
+
+  @JsonProperty("exclusive")
+  private final Boolean exclusive;
```
yuqi1129, line 69: Can't we use primitive type `exclusive` here?

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
```diff
@@ -0,0 +1,170 @@
...
+  @JsonTypeInfo(
+      use = JsonTypeInfo.Id.NAME,
+      include = JsonTypeInfo.As.EXTERNAL_PROPERTY,
+      property = "policyType",
+      defaultImpl = PolicyContentDTO.CustomContentDTO.class)
+  @JsonSubTypes({
+    // add mappings for built-in types here
+    // For example: @JsonSubTypes.Type(value = DataCompactionContent.class, name =
+    // "system_data_compaction")
+  })
```
jerryshao, line 65: Can you please add the test to verify the polymorphism of the `PolicyContent`.

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Validates the request.
+   *
+   * @throws IllegalArgumentException If the request is invalid, this exception is thrown.
+   */
+  @Override
+  public void validate() throws IllegalArgumentException {
+    Preconditions.checkArgument(
+        StringUtils.isNotBlank(name), "\"name\" is required and cannot be empty");
```
yuqi1129, line 121: You'd better unify the workd `can't be empty` is `null or empty` or just `empty`. I see you use `cannot be null or empty` in L287.

common/src/main/java/org/apache/gravitino/dto/requests/PolicyUpdateRequest.java
```diff
@@ -0,0 +1,175 @@
...
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = PolicyUpdateRequest.RenamePolicyRequest.class, name = "rename"),
+  @JsonSubTypes.Type(
+      value = PolicyUpdateRequest.UpdatePolicyCommentRequest.class,
+      name = "updateComment"),
+  @JsonSubTypes.Type(
+      value = PolicyUpdateRequest.UpdatePolicyContentRequest.class,
+      name = "updateContent")
+})
+public interface PolicyUpdateRequest extends RESTRequest {
```
jerryshao, line 48: Can you also add a test to verify the polymorphism of the `PolicyUpdateRequest`.

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
```diff
@@ -0,0 +1,170 @@
...
+  private PolicyCreateRequest() {
+    this.name = null;
+    this.policyType = null;
+    this.comment = null;
+    this.enabled = null;
+    this.exclusive = null;
+    this.inheritable = null;
+    this.supportedObjectTypes = null;
+    this.policyContent = null;
+  }
```
jerryshao, line 111: Using `this(xxx, xxx...)`.

common/src/main/java/org/apache/gravitino/dto/requests/PolicyUpdateRequest.java
```diff
@@ -0,0 +1,175 @@
...
+     * Creates a new RenamePolicyRequest.
+     *
+     * @param newName The new name of the policy.
+     */
+    public RenamePolicyRequest(String newName) {
+      this.newName = newName;
+    }
+
+    /** This is the constructor that is used by Jackson deserializer */
+    public RenamePolicyRequest() {
```
jerryshao: Make all the jackson default constructor private, also using `this(xxx, xxx...)` to replace the code.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -0,0 +1,273 @@
...
+      return ExceptionHandlers.handlePolicyException(OperationType.ALTER, name, metalake, e);
+    }
+  }
+
+  @PATCH
+  @Path("{policy}")
+  @Produces("application/vnd.gravitino.v1+json")
+  @Timed(name = "set-policy." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "set-policy", absolute = true)
+  public Response alterPolicy(
```
jerryshao: Maybe you should use a different method name to differentiate.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -0,0 +1,273 @@
...
+      @PathParam("metalake") String metalake, @PathParam("policy") String name) {
+    LOG.info("Received delete policy request for policy: {} under metalake: {}", name, metalake);
+
+    try {
+      return Utils.doAs(
+          httpRequest,
+          () -> {
+            boolean deleted = policyDispatcher.deletePolicy(metalake, name);
+            if (!deleted) {
+              LOG.warn("Failed to delete policy {} under metalake {}", name, metalake);
```
jerryshao: It is actually "Not found", not "Failed to delete", right?

common/src/main/java/org/apache/gravitino/dto/util/DTOConverters.java
mchades: `PolicyContentDTO` is the base class

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
mchades: Because we need to distinguish the case where the user has not specified a value for it.

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
mchades: where is the L287? the max number line is L170 in this file

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
mchades: If the `@AllArgsConstructor` is removed, an error will occur during compile:

>  error: constructor CustomContentDTO in class CustomContentDTO cannot be applied to given types;
>   @Builder(setterPrefix = "with")
>   ^
>   required: no arguments
>   found:    Map<String,Object>,Map<String,String>
>   reason: actual and formal argument lists differ in length

common/src/main/java/org/apache/gravitino/dto/policy/PolicyDTO.java
mchades: It doesn't work yet. That's why I added the comment here as a placeholder/reminder.

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
mchades: Since there is no built-in type policy, we cannot test polymorphism now. However, I have tested it manually locally. Please check the test code.https://github.com/apache/gravitino/pull/7821/commits/116422790d4f99e341377a8b372448da3b4485db

mchades:
@jerryshao @yuqi1129 all comments resolved, plz review again, thanks!

common/src/main/java/org/apache/gravitino/dto/requests/PolicyCreateRequest.java
jerryshao: Can we add some mock policy in UT and test it?

jerryshao: Also, we can keep those tests, if we can support polymorphism, then we can update those tests.

mchades: yeah, I don't remove the test class and only remove the polymorphism tests

jerryshao (dismissed):

jerryshao:
@yuqi1129 can you please take a look again?

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -0,0 +1,267 @@
...
+                  Arrays.stream(policies)
+                      .map(p -> DTOConverters.toDTO(p, Optional.empty()))
+                      .toArray(PolicyDTO[]::new);
+
+              LOG.info("List {} policies info under metalake: {}", policyDTOs.length, metalake);
+              return Utils.ok(new PolicyListResponse(policyDTOs));
+
+            } else {
+              String[] policyNames = policyDispatcher.listPolicies(metalake);
+              policyNames = policyNames == null ? new String[0] : policyNames;
```
yuqi1129, line 100: Resolve this one.

Issue #7823, opened 2025-07-29
[Subtask] Upgrade Gravitino server and Iceberg REST server minimal JDK version to 17


FANNG1:
### Describe the subtask

Iceberg 1.7 drops the JDK8 support, to use the latest Iceberg version,   upgrade Gravitino server and Iceberg REST server minimal JDK version to 17

### Parent issue

#6719

PR #7827, opened 2025-07-29
[#7823] feat(server): upgrade Gravitino server and Iceberg REST server mini JDK version to 17

FANNG1:
1. Upgrade Gravitino server and Iceberg REST server minimal JDK version to 17
2. Adjust the Integration test for the current module:
    a.  JDK17  module: support JDK17 for embedded and deploy mode. 
    b.  Non-JDK17 module:  support JDK17 for deploy mode only.
3. Move integration test from `catalog-model` and `client-java` to `integration-test` module, which supports JDK17 in embedded mode and JDK 7,11,17 in deploy mode.

JDK17 modules: server-related modules:  `:server`, `:iceberg`, `:catalogs`, `:lineage`, `authorizations`.  
`:catalogs:catalogs-fileset` `:catalogs:hadoop-common` is the special case because `bundles` modle is JDK 8 compatible and relies on them, so they are JDK 8, not 17.

Fix: #7823 

adding ITs


.github/workflows/backend-integration-test-action.yml
```diff
@@ -62,6 +62,12 @@ jobs:
           -x :spark-connector:spark-runtime-3.3:test -x :spark-connector:spark-runtime-3.4:test -x :spark-connector:spark-runtime-3.5:test
           -x :trino-connector:integration-test:test -x :trino-connector:trino-connector:test
           -x :authorizations:authorization-chain:test -x :authorizations:authorization-ranger:test 
+          -x :integration-test:test -x :catalogs:catalog-fileset:test
```
diqiu50, line 65: Why we remove the module of `:integration-test:test`

.github/workflows/flink-integration-test-action.yml
```diff
@@ -46,9 +46,8 @@ jobs:
 
       - name: Flink Integration Test
         id: integrationTest
-        # run embedded mode and deploy mode integration tests
+        # run deploy mode integration tests
         run: |
-          ./gradlew -PskipTests -PtestMode=embedded -PjdkVersion=${{ inputs.java-version }} -PskipDockerTests=false :flink-connector:flink:test --tests "org.apache.gravitino.flink.connector.integration.test.**"
           ./gradlew -PskipTests -PtestMode=deploy -PjdkVersion=${{ inputs.java-version }} -PskipDockerTests=false :flink-connector:flink:test --tests "org.apache.gravitino.flink.connector.integration.test.**"
```
diqiu50, line 52: Does the Flink integration test not run in deploy mode?

.github/workflows/integration-test-action.yml
```diff
@@ -0,0 +1,73 @@
+name: New Integration Test Action
```
diqiu50, line 1: Is this integration test running on the client side?

.github/workflows/integration-test-action.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Check required command
+        run: |
+          dev/ci/check_commands.sh
+
+      - name: Package Gravitino
+        if: ${{ inputs.test-mode == 'deploy' }}
```
diqiu50, line 48: Can we validate this earlier in the process? It currently takes a long time to fail at this stage.

build.gradle.kts
```diff
@@ -105,6 +106,53 @@ project.extra["extraJvmArgs"] = if (extra["jdkVersion"] in listOf("8", "11")) {
   )
 }
 
+fun useHighVersionJDK(project: Project): Boolean {
```
diqiu50: is useJDK17  better?

build.gradle.kts
```diff
@@ -105,6 +106,53 @@ project.extra["extraJvmArgs"] = if (extra["jdkVersion"] in listOf("8", "11")) {
...
+    return true
+  }
+
+  return false
+}
+
+// Use JDK 17 for embedded mode, use the JDK from `jdkVersion` for deploy mode
+fun getJdkVersionForTest(project: Project): JavaLanguageVersion {
+  if (rootProject.extra["isTestModeEmbedded"] == true) {
+    return JavaLanguageVersion.of(17)
```
diqiu50: It would be better to extract the JDK version into a variable, such as defaultJdkVersion or minSupportedJdkVersion.

clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/GravitinoVirtualFileSystemIT.java
```diff
@@ -54,10 +54,12 @@
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIf;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @Tag("gravitino-docker-test")
+@DisabledIf("org.apache.gravitino.integration.test.util.ITUtils#isEmbedded")
 public class GravitinoVirtualFileSystemIT extends BaseIT {
```
diqiu50, line 63: introduce a base class named ClientIT to avoid adding @DisabledIf annotations to multiple classes.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/integration/test/IcebergRestKerberosHiveCatalogIT.java
```diff
@@ -157,4 +158,20 @@ private static boolean isEmbedded() {
...
+      } else {
+        classRef = Class.forName("sun.security.krb5.Config");
+      }
+
+      Method refershMethod = classRef.getMethod("refresh");
+      refershMethod.invoke(null);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
```
diqiu50, line 176: What problem does this change aim to solve?

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/GravitinoITUtils.java
```diff
@@ -18,12 +18,15 @@
...
 
+import com.google.common.base.Preconditions;
 import java.io.File;
 import java.nio.charset.StandardCharsets;
 import java.util.UUID;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
```
diqiu50: Don't use the shaded library

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/GravitinoITUtils.java
```diff
@@ -34,6 +37,9 @@ private GravitinoITUtils() {
 
   public static void startGravitinoServer() {
     String gravitinoStartShell = System.getenv("GRAVITINO_HOME") + "/bin/gravitino.sh";
```
diqiu50, line 39: Check that it is not null.

.github/workflows/backend-integration-test-action.yml
FANNG1: `:integration-test:test` is the new module which test in seperate actions.

.github/workflows/flink-integration-test-action.yml
FANNG1: all client tests doesn't run in embedde mode because the `server` module is JDK 17, the target compatibility of client test code is JDK 8

.github/workflows/integration-test-action.yml
FANNG1: I couldn't get your point, could you provide more information?

.github/workflows/integration-test-action.yml
FANNG1: It doesn't separate clients and servers,  for the special client tests like spark&flink tests, we could change the workflow yaml when changing these modules.

build.gradle.kts
FANNG1: You mean use a variable to represent `JavaLanguageVersion.of(17)` here?  I didn't see too much benefit since it only use once, WDYT?

build.gradle.kts
FANNG1: ok

clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/GravitinoVirtualFileSystemIT.java
FANNG1: I didn't figure out the reason, in some classes adding the `DisabledIf` to base class doesn't take affect to the sub class, so I add to `GravitinoVirtualFileSystemS3IT`  and `GravitinoVirtualFilesystemWithFSOpsDisabledIT ` too.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/integration/test/IcebergRestKerberosHiveCatalogIT.java
FANNG1: seems caused by upgrading the target compatibility from 8 to 17, the `sun.security.krb5` is not visible anymore.
```
error: package sun.security.krb5 is not visible
      sun.security.krb5.Config.refresh();
                  ^
  (package sun.security.krb5 is declared in module java.security.jgss, which does not export it to the unnamed module)
```

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/GravitinoITUtils.java
FANNG1: `GRAVITINO_HOME` couldn't be null, it's used everywhere.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/GravitinoITUtils.java
FANNG1: ok

FANNG1:
@diqiu50 all comments addressed, please help to review again, thx

.github/workflows/backend-integration-test-action.yml
diqiu50: ok

.github/workflows/flink-integration-test-action.yml
diqiu50: Did Flink integration tests previously only run in embedded mode?

build.gradle.kts
diqiu50: I noticed that JavaLanguageVersion.of(17) is used in multiple places

.github/workflows/integration-test-action.yml
diqiu50: What I mean is that if it is in deploy mode, we should skip running the integration-test-action.yml workflow.

.github/workflows/integration-test-action.yml
diqiu50: Sorry, I misunderstood

build.gradle.kts
FANNG1: ok

FANNG1:
use new branch `branch-jdk` , so close the PR

Issue #7688, opened 2025-07-14
[Bug report] Duplicate tags for matadata object


danhuawang:
### Version

main branch

### Describe what's wrong

There's a tag named "high". 
Add this tag to the schema `catalog_hive_rename.arriel_schema`, then add this tag to the catalog `catalog_hive_rename` .

Get tags for metalake object `catalog_hive_rename.arriel_schema.arriel_table`. The table has duplicate tags as following:

```
{
    "code": 0,
    "tags": [
        {
            "name": "high",
            "comment": "",
            "properties": {
                "color": "#a11303"
            },
            "audit": {
                "creator": "service_admin",
                "createTime": "2025-05-30T06:48:48.002340774Z"
            },
            "inherited": true
        },
        {
            "name": "successfully",
            "comment": "",
            "properties": {
                "color": "#7aef6a"
            },
            "audit": {
                "creator": "service_admin",
                "createTime": "2025-06-12T08:34:24.212329812Z",
                "lastModifier": "service_admin",
                "lastModifiedTime": "2025-06-17T07:52:42.949961823Z"
            },
            "inherited": true
        },
        {
            "name": "high",
            "comment": "",
            "properties": {
                "color": "#a11303"
            },
            "audit": {
                "creator": "service_admin",
                "createTime": "2025-05-30T06:48:48.002340774Z"
            },
            "inherited": true
        }
    ]
}
```

### Error message and/or stacktrace

N/A

### How to reproduce

See above

danhuawang:
@jerryshao

PR #7830, opened 2025-07-29
[#7688] fix(server): Fix duplicated tags when list tags for a metadata object.

yuqi1129:

Use set to remove duplicated tags when list all tags for a metadata object.

We need to eliminate duplicates.

Fix: #7688 

N/A

UT

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
@@ -20,12 +20,11 @@ package org.apache.gravitino.server.web.rest;
 
 import com.codahale.metrics.annotation.ResponseMetered;
 import com.codahale.metrics.annotation.Timed;
-import com.google.common.collect.Lists;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.Set;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.DefaultValue;
@@ -51,6 +50,7 @@ import org.apache.gravitino.metrics.MetricNames;
 import org.apache.gravitino.server.web.Utils;
 import org.apache.gravitino.tag.Tag;
 import org.apache.gravitino.tag.TagDispatcher;
+import org.glassfish.jersey.internal.guava.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
```diff
@@ -163,7 +163,7 @@ public Response listTagsForMetadataObject(
                 MetadataObjects.parse(
                     fullName, MetadataObject.Type.valueOf(type.toUpperCase(Locale.ROOT)));
 
-            List<TagDTO> tags = Lists.newArrayList();
+            Set<TagDTO> tags = Sets.newHashSet();
```
mchades, line 166: 202 line also need to be updated

yuqi1129: No, 
```
 String[] tagNames = tags.stream().map(TagDTO::name).distinct().toArray(String[]::new);

```
It already uses `distinct` here.

yuqi1129: But there seems to be another problem.  The size of the result may not be consistent with different value `verbose` here.

mchades: > No,
> 
> ```
>  String[] tagNames = tags.stream().map(TagDTO::name).distinct().toArray(String[]::new);
> ```
> 
> It already uses `distinct` here.

`tags` is a set now and does not need to distinct

yuqi1129: The problem is that `equals` method in TagDTO is 
```java
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    }
    if (!(o instanceof TagDTO)) {
      return false;
    }

    TagDTO tagDTO = (TagDTO) o;
    return Objects.equal(name, tagDTO.name)
        && Objects.equal(comment, tagDTO.comment)
        && Objects.equal(properties, tagDTO.properties)
        && Objects.equal(audit, tagDTO.audit);
  }
```

So there can be two different `TagDTO` with the same `name`, that is why I say `The size of the result may not be consistent with different value verbose here.`

mchades: the 200 line code comment also needs to be removed since the `tags` is a set

mchades (approved):
lgtm

Issue #7841, opened 2025-07-30
[Subtask] Server Implementation for Policy (part-2)


mchades:
### Describe the subtask

this issue will track the work of policy relational operations implementation on server-side

### Parent issue

#7139

PR #7842, opened 2025-07-30
[#7841] feat(policy): implement the policy management on server-side (part-2)

mchades:

policy relational operations implementation on server-side

Fix: #7841 

yes, new APIs are implemented

tests added


mchades:
@jerryshao @yuqi1129 Please help review this PR when you have time, thanks!

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/PolicyMetaBaseSQLProvider.java
```diff
@@ -73,6 +73,27 @@ public String listPolicyPOsByMetalakeAndPolicyNames(
         + "</script>";
   }
 
+  public String listDiffPolicyPOsByMetalakeIdAndPolicy(
```
jerryshao, line 86: What is the meaning and usage of "DiffPolicy"?

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/PolicyMetaBaseSQLProvider.java
```diff
@@ -73,6 +73,27 @@ public String listPolicyPOsByMetalakeAndPolicyNames(
...
+        + " pvi ON pm.policy_id = pvi.policy_id"
+        + " AND pm.current_version = pvi.version"
+        + " WHERE pm.metalake_id = #{metalakeId}"
+        + " AND pm.policy_id != #{policy.policyId}"
+        + " AND pm.policy_type = #{policy.policyType}"
+        + " AND (pm.inheritable != #{policy.inheritable} OR"
+        + " pm.exclusive != #{policy.exclusive} OR"
+        + " pm.supported_object_types != #{policy.supportedObjectTypes})"
+        + " AND pm.deleted_at = 0 AND pvi.deleted_at = 0";
+  }
```
jerryshao, line 105: Can you explain more about this SQL query?

core/src/main/java/org/apache/gravitino/storage/relational/service/PolicyMetaService.java
```diff
@@ -86,6 +91,24 @@ public void insertPolicy(PolicyEntity policyEntity, boolean overwritten) throws
...
+                policyPO.getPolicyName(),
+                diffPolicyPOs.get(0).getPolicyName(),
+                policyPO.getPolicyType(),
+                diffPolicyPOs.get(0).isExclusive(),
+                diffPolicyPOs.get(0).isInheritable(),
+                diffPolicyPOs.get(0).getSupportedObjectTypes(),
+                policyPO.isExclusive(),
+                policyPO.isInheritable(),
+                policyPO.getSupportedObjectTypes()));
+      }
```
jerryshao, line 110: Shall we move this kind of check logic and below to `PolicyManager`? I think we involve too much logic in the storage layer.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/PolicyMetaBaseSQLProvider.java
mchades: When creating a policy, if the policy with the same type already exists, we should ensure that the specified immutable attributes (`exclusive`, `inheritable`, `supportedObjectTypes`) match those of the existing policy. The `listDiffPolicyPOsByMetalakeIdAndPolicy` method is used to check the differences between policies.

common/src/main/java/org/apache/gravitino/dto/requests/PoliciesAssociateRequest.java
```diff
@@ -0,0 +1,84 @@
...
+  public void validate() throws IllegalArgumentException {
+    Preconditions.checkArgument(
+        policiesToAdd != null || policiesToRemove != null,
+        "policiesToAdd and policiesToRemove cannot both be null");
+
+    if (policiesToAdd != null) {
+      for (String Policy : policiesToAdd) {
+        Preconditions.checkArgument(
+            StringUtils.isNotBlank(Policy),
+            "policiesToAdd must not contain null or empty Policy names");
```
yuqi1129: `Policy` should not be capitalized unless it's a technical term.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,348 @@
...
+            MetadataObject parentObject = MetadataObjects.parent(object);
+            while (parentObject != null) {
+              if (!targetPolicy.supportedObjectTypes().contains(parentObject.type())) {
+                // If the parent object type is not supported by the target policy, we skip it.
+                parentObject = MetadataObjects.parent(parentObject);
+                continue;
+              }
+
+              Policy[] parentPolicies =
+                  policyDispatcher.listPolicyInfosForMetadataObject(metalake, parentObject);
```
jerryshao, line 134: Why do you use "list" rather than "get" for policy?

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,348 @@
...
+                  metalake);
+              return Utils.ok(
+                  new PolicyResponse(DTOConverters.toDTO(policy.get(), Optional.of(false))));
+            }
+
+            // ensure the policy exists and is inheritable
+            Policy targetPolicy = policyDispatcher.getPolicy(metalake, policyName);
+            if (!targetPolicy.inheritable()
+                || !targetPolicy.supportedObjectTypes().contains(object.type())) {
+              return logNotFoundPolicy(metalake, policyName, type, fullName);
```
yuqi1129: The name `logNotFoundPolicy` is not very proper, based on the code, it can be `getResponseAndLogNotFoundPolicy`, you can simplify it as you like.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,348 @@
...
+                  policyName,
+                  type,
+                  fullName,
+                  metalake);
+              return Utils.ok(
+                  new PolicyResponse(DTOConverters.toDTO(policy.get(), Optional.of(false))));
+            }
+
+            // ensure the policy exists and is inheritable
+            Policy targetPolicy = policyDispatcher.getPolicy(metalake, policyName);
```
yuqi1129, line 109: What if `targetPolicy` does not exist here? Does it throw an exception directly?

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,348 @@
...
+            MetadataObject object =
+                MetadataObjects.parse(
+                    fullName, MetadataObject.Type.valueOf(type.toUpperCase(Locale.ROOT)));
+
+            Set<PolicyDTO> policies = new HashSet<>();
+            Set<String> exclusivePolicyTypes = new HashSet<>();
+            Policy[] nonInheritedPolicies =
+                policyDispatcher.listPolicyInfosForMetadataObject(metalake, object);
+            for (Policy policy : nonInheritedPolicies) {
+              if (policy.exclusive()) {
```
yuqi1129, line 205: Do we check the property `exclusive`  in the association operation? Why do we need to check it again?

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,348 @@
...
+              Policy[] inheritedPolicies =
+                  policyDispatcher.listPolicyInfosForMetadataObject(metalake, parentObject);
+              for (Policy policy : inheritedPolicies) {
+                if (!policy.supportedObjectTypes().contains(object.type())
+                    || !policy.inheritable()) {
+                  // If the policy is not inheritable or does not support the object type,
+                  // we skip it.
+                  continue;
+                }
+                if (policy.exclusive()) {
```
yuqi1129, line 229: ditto

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,348 @@
...
+            + ", full name: "
+            + fullName
+            + " under metalake: "
+            + metalakeName);
+  }
+
+  private Optional<Policy> getPolicyForObject(
+      String metalake, MetadataObject object, String policyName) {
+    try {
+      return Optional.ofNullable(
```
yuqi1129, line 368: Since it will throw a `NoSuchPolicyException`, why do you use `Optional` here?

core/src/main/java/org/apache/gravitino/storage/relational/service/PolicyMetaService.java
mchades: Moving this logic to `PolicyManager` would expose unnecessary methods to `policyMetaService` and also require adding a new interface to `JDBCBackend`, which is not a standard operation for `RelationalBackend`. So I prefer to keep it where it is.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: Because the same type of policy may have different names, so we should list all policies for the object first

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: throws NoSuchPolicyException

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: it's used to calculate the inherited value

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: Optional means that the policy exists but is not associated with the object.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/PolicyMetaBaseSQLProvider.java
mchades: JavaDoc comment added

mchades:
@jerryshao @yuqi1129 all comments resolved, please help to review again when you have time, thanks very much!

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
yuqi1129: So, Will `Optional<Policy>` never be empty,  and the metadata in the policy can be empty?

mchades: I don't understand your question. I've added a JavaDoc for the method. Please let me know if you still have any questions.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,357 @@
...
+    if (!targetPolicy.exclusive() || ArrayUtils.isEmpty(policies)) {
+      return false;
+    }
+    return Arrays.stream(policies)
+        .anyMatch(
+            p ->
+                p.exclusive()
+                    && p.policyType().equalsIgnoreCase(targetPolicy.policyType())
+                    && !p.name().equals(targetPolicy.name()));
+  }
```
jerryshao: What's the difference of these two methods? AFAIK, we can move the difference out of this method, so they can be combined as one.

What's more, the logic here in this file to check the different status is too complicated, you should add more comments. Which also makes me think that this policy system is a bit hard to understand and explain, because there're too many properties.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,357 @@
...
+            // ensure the policy exists and is inheritable
+            Policy targetPolicy = policyDispatcher.getPolicy(metalake, policyName);
+            if (!targetPolicy.inheritable()
+                || !targetPolicy.supportedObjectTypes().contains(object.type())) {
+              return getPolicyNotFoundResponse(metalake, policyName, type, fullName);
+            }
+
+            if (hasConflictExclusivePolicy(metalake, object, targetPolicy)) {
+              return getPolicyNotFoundResponse(metalake, policyName, type, fullName);
+            }
```
jerryshao: How could this be happened? Please explain more.

mchades: code comments added

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: fixed and added more comments

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,375 @@
...
+                  new PolicyResponse(DTOConverters.toDTO(policy.get(), Optional.of(false))));
+            }
+
+            // 2. If not found, try to find an inheritable policy from its parents.
+            // First, ensure the target policy exists and is inheritable for this object type.
+            Policy targetPolicy = policyDispatcher.getPolicy(metalake, policyName);
+            if (!targetPolicy.inheritable()
+                || !targetPolicy.supportedObjectTypes().contains(object.type())) {
+              return getPolicyNotFoundResponse(metalake, policyName, type, fullName);
+            }
```
jerryshao: Where is the logic to get policy from its parent? You just randomly get a policy by its name. I'm super confused about your logic here. You need to explain more.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,375 @@
...
+                    continue;
+                  }
+                  // Record the exclusive policy type to block any further policies of the same
+                  // type from higher-level ancestors.
+                  exclusivePolicyTypes.add(policy.policyType());
+                }
+                policies.add(DTOConverters.toDTO(policy, Optional.of(true)));
+              }
+              parentObject = MetadataObjects.parent(parentObject);
+            }
```
jerryshao, line 240: The "exclusive", "inherited", and "supportedTypes" make things too complicated, I suggest we revisit the design and have a discussion offline. I don't think it can be easily explained to the users.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,375 @@
...
+          httpRequest,
+          () -> {
+            request.validate();
+            MetadataObject object =
+                MetadataObjects.parse(
+                    fullName, MetadataObject.Type.valueOf(type.toUpperCase(Locale.ROOT)));
+            String[] policyNames =
+                policyDispatcher.associatePoliciesForMetadataObject(
+                    metalake, object, request.getPoliciesToAdd(), request.getPoliciesToRemove());
+            policyNames = policyNames == null ? new String[0] : policyNames;
```
jerryshao, line 296: Why can't we do the complicated validation in association, rather than in the list/get operation?

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: In fact, getting the policy from its parent node starts from the following `while (parentObject != null)`, and the logic here is `//First, make sure the target policy exists and can be inherited by this object type.`

I will remove this inaccurate comment:`//2. If not found, try to find an inheritable policy from its parents.`

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
mchades: The association operation validates the exclusive policy to ensure that the same type of exclusive policy cannot be associated with one metadata object (the logic is in `associatePoliciesWithMetadataObject.associatePoliciesForObject`), while the list and get operations perform the calculation (determine the inheritance relationship), not validation.

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -0,0 +1,375 @@
...
+                  new PolicyResponse(DTOConverters.toDTO(policy.get(), Optional.of(false))));
+            }
+
+            // 2. If not found, first ensure the target policy exists and is inheritable for this
+            // object type.
+            Policy targetPolicy = policyDispatcher.getPolicy(metalake, policyName);
+            if (!targetPolicy.inheritable()
+                || !targetPolicy.supportedObjectTypes().contains(object.type())) {
+              return getPolicyNotFoundResponse(metalake, policyName, type, fullName);
+            }
```
jerryshao, line 113: Why "is inheritable for this object type"? Based on the code, it is the policy itself inheritable, and happens that it can support the type this object has. How do you know the relationship between this policy with the specified name and the object?

mchades: Honestly, I didn't understand your question. 

The code logic reaches here, which means this policy has **no direct** association with this object. There are two possible results: one is that this policy may have been inherited from the parent object; the other is that this policy does not apply to this object. By judging the `inheritable` and `supportedObjectTypes` properties of the policy, we can quickly determine whether this policy can apply to this object.

jerryshao: The policy itself is inheritable, which doesn't mean that this object can inherit this policy. In you above code, when you check the `inheritable`, you don't check if it is inheritable to this "object", how do you know this policy is actually associated to this object?

mchades: > In you above code, when you check the inheritable, you don't check if it is inheritable to this "object"

the codes do the check `!targetPolicy.supportedObjectTypes().contains(object.type())`

jerryshao: It is just the type, objects can have the same type, but they may have no relations. How do you say that if policy is associated to fileset A that can be inherited to fileset B (though this policy can support fileset)?

mchades: The code logic has never mentioned filesetB, and the metadata object has always been fixed as filesetA.

mchades:
plz review #8096

Issue #7836, opened 2025-07-30
[Subtask] Add partition stats storage API


jerqi:
### Describe the subtask

Add partition stats storage API

### Parent issue

#7268

PR #7690, opened 2025-07-14
[#7269] feat(storage): Add storage for statistics

jerqi:
### What changes were proposed in this pull request?

Add storage for statistics

### Why are the changes needed?

Fix: #7269

### Does this PR introduce _any_ user-facing change?

Add storage for statistics

### How was this patch tested?

Add UT

PR #7850, opened 2025-07-30
[#7836] feat(storage): Add partition statistics storage interface

roryqi:

Add partition statistics storage interface

Fix: #7836

Developer API, not for users.

Just an interface.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/stats/SupportsPartitionStatistics.java
+++ b/api/src/main/java/org/apache/gravitino/stats/SupportsPartitionStatistics.java
@@ -24,7 +24,7 @@ import org.apache.gravitino.exceptions.UnmodifiableStatisticException;
 
 /** SupportsPartitionStatistics provides methods to list and update statistics for partitions. */
 @Unstable
-interface SupportsPartitionStatistics {
+public interface SupportsPartitionStatistics {
 
   /**
    * Lists statistics for partitions from one partition name to another partition name.


core/src/main/java/org/apache/gravitino/storage/external/stats/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,117 @@
...
+   * @return the name of the metalake
+   */
+  String metalake();
+
+  /**
+   * Returns the properties of this storage.
+   *
+   * @return a map of properties associated with this storage.
+   */
+  Map<String, String> properties();
```
jerryshao: Why do you need this two interface, can you provide a mock implementation, which can also used for test.

roryqi: OK. I have implemented Memory storage and LanceDB storage. I can submit memory storage first.

Actually, I added these two interfaces to constrain the storage field. We should have metalake and properties as the class fields..
Metalake field: In other interfaces, because we don't need to pass metalake as the paramter.
properties: Upper layer can get storage property, such as location, etc.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,117 @@
...
+   * @return the name of the metalake
+   */
+  String metalake();
+
+  /**
+   * Returns the properties of this storage.
+   *
+   * @return a map of properties associated with this storage.
+   */
+  Map<String, String> properties();
```
jerryshao: It is weird you want the implementor to provide you a metalake and properties. If you're an implementor, how do you provide these? This is anti-intuitive.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,117 @@
...
+                Collectors.toMap(
+                    Map.Entry::getKey,
+                    e ->
+                        e.getValue().stream()
+                            .map(
+                                entry ->
+                                    PartitionStatisticsDrop.of(
+                                        entry.partitionName(),
+                                        Lists.newArrayList(entry.statistics().keySet())))
+                            .collect(Collectors.toList())));
```
jerryshao: This code is hard to review using stream. Also, how do you guarantee the thread safe thing?

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: Removed.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: Add lock and add more data structure.

core/src/main/java/org/apache/gravitino/stats/storage/MemoryPartitionStatsStorageFactory.java
```diff
@@ -0,0 +1,218 @@
...
+      }
+
+      synchronized (tableStats) {
+        Map<String, Map<String, StatisticValue<?>>> resultStats = Maps.newHashMap();
+        for (PersistedPartitionStatistics partitionStat :
+            tableStats.partitionStatistics().values()) {
+          String partitionName = partitionStat.partitionName();
+          boolean lowerBoundSatisfied =
+              range
+                  .lowerPartitionName()
```
roryqi: Extract a method to improve code readable.

core/src/main/java/org/apache/gravitino/stats/storage/MetadataObjectStatisticsUpdate.java
```diff
@@ -0,0 +1,71 @@
...
+ * under the License.
+ */
+package org.apache.gravitino.stats.storage;
+
+import java.util.List;
+import org.apache.gravitino.MetadataObject;
+import org.apache.gravitino.stats.PartitionStatisticsUpdate;
+
+/**
+ * MetadataObjectStatisticsUpdate represents a collection of statistics updates for a specific
```
roryqi, line 26: Add  comment.

jerqi:
@Copilot

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorageFactory.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface PartitionStatisticStorageFactory {
+
+  /**
+   * Creates an instance of {@link org.apache.gravitino.stats.storage.PartitionStatisticStorage}.
+   *
+   * @param properties additional properties for the storage configuration
+   * @return an instance of {@link org.apache.gravitino.stats.storage.PartitionStatisticStorage}
+   */
+  PartitionStatisticStorage open(Map<String, String> properties);
```
mchades: In factory class, `create` is usually used as the method name

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,112 @@
...
+   * implementation should guarantee the thread safe.
+   *
+   * @param metalake the name of the metalake
+   * @param metadataObject the metadata object for which statistics are being listed
+   * @param partitionRange the range of partition names for which statistics are being listed
+   * @return a list of {@link PersistedPartitionStatistics} objects, each containing the partition
+   *     name
+   */
+  List<PersistedPartitionStatistics> listStatistics(
+      String metalake, MetadataObject metadataObject, PartitionRange partitionRange);
```
mchades, line 42: Should we use `Entity` instead of `MetadataObject`, as discussed in https://github.com/apache/gravitino/pull/7690#discussion_r2262665448?

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Appends statistics to the storage for a given metadata object. The implementation should
+   * guarantee the thread safe.
+   *
+   * @param metalake the name of the metalake
+   * @param statisticsToAppend a list of {@link MetadataObjectStatisticsUpdate} objects, each
+   *     containing the metadata object and its associated statistics updates.
+   */
+  void appendStatistics(String metalake, List<MetadataObjectStatisticsUpdate> statisticsToAppend);
```
mchades: What's the difference from `updateStatistics`?

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: I prefer using metadata object here. Because we have defined many classes which contained the metadata object. If we uses entity, we need to convert them and create many extra classes. I don't think it's necessary. Only the entityStore prefer using entity. Other interfaces, we can use metadata objects. But we should avoid mixing metadata object and entities in interfaces of the same class.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: The difference is how to process the same key data:
The origin data is 
```
{
"k1": "v1"
}
```
Append k1:v2
is
``` 
{
"k1": "v2",
"k1": "v1"
}
```
Update k1:v2
is
```
{
"k1": "v2"
}
```

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorageFactory.java
roryqi: OK.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
mchades: what if append is 
```
{
"k1": "v2"
}
```
and update is
```
{
"k2": "v1"
}
```

roryqi: append is 
```
{
"k1": "v2"
}
```
is
```
{
"k1": "v2",
"k1": "v2"
}
```
and update is
```
{
"k2": "v1"
}
```
is
```
{
"k1": "v1",
"k2": "v1",
}
```

jerryshao: Why do we need "append" semantic?

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,119 @@
...
+                                    PartitionStatisticsDrop.of(
+                                        partitionUpdate.partitionName(),
+                                        Lists.newArrayList(partitionUpdate.statistics().keySet())))
+                            .collect(Collectors.toList())))
+            .collect(Collectors.toList());
+
+    dropStatistics(metalake, statisticsToDrop);
+    appendStatistics(metalake, statisticsToUpdate);
+  }
+}
```
jerryshao, line 91: Why can we provide a "update" interface, let user to decide how to implement the update semantic, I think here the default operation is low-efficient.

Again, why do we provide a "append" API, and use "drop" + "append" to achieve "update"? We can always use "update" to achieve upsert semantic.

roryqi: Some storage doesn't support upsert sematics. For example, Lance format doesn't support upsert in Java's API. We need to delete first and append later.  The design is mainly considered the storage which doesn't support to upsert , but supports delete and append.

jerryshao: You don't have to do such an assumption in the interface layer. If one store cannot support upsert, or needs a different way to support it, the implementor should figure out a way. As an interface provider, you need to bring out the "requirement", the implementor will pick the right store based on the "requirement" (interface spec).

roryqi: OK, I removed the interfaces and polish the comment.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: OK, I removed the interfaces and polish the comment.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -86,34 +73,13 @@ void dropStatistics(
...
 
   /**
-   * Updates statistics for a given metadata object. The default implementation is to first drop and
-   * then append the statistics. Developer can override this logic if needed. The implementation
-   * should guarantee the thread safe. The upper layer will add the parent object level write lock.
-   * For example, if the metadata object is a table, the write lock of the schema level will be
-   * held.
+   * Updates statistics for a given metadata object. The implementation should guarantee the thread
+   * safe. The upper layer will add the parent object level write lock. For example, if the metadata
+   * object is a table, the write lock of the schema level will be held.
```
jerryshao: You should describe the behavior, is it pure update, or upsert.

core/src/main/java/org/apache/gravitino/stats/storage/MemoryPartitionStatsStorageFactory.java
```diff
@@ -0,0 +1,226 @@
...
+        String metalake, MetadataObject metadataObject, PartitionRange range) {
+      MetadataObjectStatisticsContainer tableStats =
+          totalStatistics.get(new MetadataContainerKey(metalake, metadataObject));
+
+      if (tableStats == null) {
+        return Lists.newArrayList();
+      }
+
+      Map<String, Map<String, StatisticValue<?>>> resultStats = Maps.newHashMap();
+      for (PersistedPartitionStatistics partitionStat : tableStats.partitionStatistics().values()) {
```
jerryshao, line 67: Do you need a lock here for iteration?

core/src/main/java/org/apache/gravitino/stats/storage/MemoryPartitionStatsStorageFactory.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        if (tableStats.partitionStatistics().isEmpty()) {
+          totalStatistics.remove(new MetadataContainerKey(metalake, metadataObject));
+        }
+      }
+    }
+
+    @Override
+    public void close() throws IOException {}
```
jerryshao: clean the map explicitly when closed.

core/src/main/java/org/apache/gravitino/stats/storage/MemoryPartitionStatsStorageFactory.java
roryqi: /**
   * Updates statistics for a given metadata object. Locking guarantee: The upper layer will acquire
   * a write lock at the metadata object level. For example, if the metadata object is a table, the
   * write lock of the table level will be held.
   *
   * @param metalake the name of the metalake
   * @param statisticsToUpdate a list of {@link MetadataObjectStatisticsUpdate} objects, each
   *     containing the metadata object and its associated statistics updates.
   */
We will acquire the write lock. So we don't accquire lock here.

core/src/main/java/org/apache/gravitino/stats/storage/MemoryPartitionStatsStorageFactory.java
roryqi: Fixed.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: Added the comment. `If the statistic exists, it will be updated; If the statistic doesn't exist, it will be created`

core/src/main/java/org/apache/gravitino/stats/storage/PersistedPartitionStatistics.java
```diff
@@ -0,0 +1,73 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.stats.storage;
+
+import java.util.Map;
+import org.apache.gravitino.stats.StatisticValue;
+
+/** Represents a collection of statistics for a specific partition in a metadata object. */
+public class PersistedPartitionStatistics {
```
mchades, line 25: do you need to implemente the `equals` and `hashCode` method for this class?

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
mchades: What if a partial statistics update fails?

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
```diff
@@ -0,0 +1,86 @@
...
+   * Drops statistics for specified partitions of a metadata object. Locking guarantee: The upper
+   * layer will acquire a write lock at the metadata object level. For example, if the metadata
+   * object is a table, the write lock of the table level will be held.
+   *
+   * @param metalake the name of the metalake
+   * @param partitionStatisticsToDrop a map where the key is a {@link MetadataObject} and the value
+   *     is a list of {@link PartitionStatisticsDrop}
+   */
+  void dropStatistics(
+      String metalake, List<MetadataObjectStatisticsDrop> partitionStatisticsToDrop);
```
mchades: What if a partial statistics drop fails?

mchades: BTW, the return value is void, how can we distinguish between the absence of a statistic (even partition) and a successful deletion?

The parameter is `List<MetadataObjectStatisticsDrop>`. Does this mean that users can drop statistics across tables, schemas, or even catalogs in a request?

core/src/main/java/org/apache/gravitino/stats/storage/PersistedPartitionStatistics.java
roryqi: I don't have the requirement to judge whether the value is equal.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: It depends on the underlying storage system.  Some storage can provide transactions. However, some storage can't provide transactions.

core/src/main/java/org/apache/gravitino/stats/storage/PartitionStatisticStorage.java
roryqi: 1. It depends on the underlying storage.
 2. Because the some underlying systems can't provide information of absence of a statistic.
 3. Yes, they are.

mchades (approved):
lgtm

Issue #7848, opened 2025-07-30
[Improvement] Support create mysql table with column properties in Trino


hqbhoho:
### What would you like to be improved?

Support create mysql table with column properties in Trino like this:
```sql
CREATE TABLE gt_mysql.gt_mysql_test_index.demo_with_primary_key_and_unique_key (
   key1 integer NOT NULL WITH ( auto_increment = true ),
   col1 integer
)
COMMENT ''
WITH (
   engine = 'InnoDB',
   primary_key = ARRAY['key1']
)
```
Maybe support  default values and auto-increment.

PR #7852, opened 2025-07-30
[#7848] feat(trino-connector): Support create table with column properties in MySQL

hqbhoho:
Support create MySQL table with default value and auto-increment

Fix: #7848

no

local tests

### DIFF
--- a/docs/trino-connector/catalog-mysql.md
+++ b/docs/trino-connector/catalog-mysql.md
@@ -16,7 +16,7 @@ To connect to MySQL, you need:
 
 ## Create table
 
-At present, the Apache Gravitino Trino connector only supports basic MySQL table creation statements, which involve fields, null allowances, comments, primary keys and indexes. However, it does not support advanced features like default values and auto-increment.
+At present, the Apache Gravitino Trino connector only supports basic MySQL table creation statements, which involve fields, null allowances, comments, primary keys, indexes, default values and auto-increment.
 
 The Gravitino Trino connector does not support `CREATE TABLE AS SELECT`.
 
@@ -46,6 +46,32 @@ The following are supported MySQL table properties:
 | auto_increment_offset              | string | (none)         | The auto increment offset for the table.                                                                                                | No       | 0.4.0         |
 | primary_key                        | list   | (none)         | The primary keys for the table, can choose multi columns as the table primary key. All key columns must be defined as `NOT NULL`.       | No       | 1.0.0         |
 | unique_key                         | list   | (none)         | The unique keys for the table, can choose multi columns for multi unique key. Each unique key should be defined as `keyName:col1,col2`. | No       | 1.0.0         |
+
+The following are supported MySQL column properties:
+
+| Property name                      | Type    | Default Value | Description                                       | Required | Since Version |
+|------------------------------------|---------|---------------|---------------------------------------------------|----------|---------------|
+| auto_increment                     | boolean | false         | The auto increment column.                        | No       | 1.0.0         |
+| default                            | string  | (none)        | The default value for column.                     | No       | 1.0.0         |
+
+**Note:** Currently, creating tables only supports constant default values and does not support expression default values, 
+and `show create table` also exclusively renders constant default values in its output.
+The following are Trino type which support configuration of default values:
+
+| Type name | Default Value example                   |
+|-----------|-----------------------------------------|
+| TINYINT   | 1                                       |
+| SMALLINT  | 1                                       | 
+| INT       | 1                                       | 
+| BIGINT    | 1                                       | 
+| REAL      | 1.0                                     | 
+| DOUBLE    | 1.0                                     | 
+| DECIMAL   | 1.0                                     | 
+| VARCHAR   | abc                                     | 
+| CHAR      | abc                                     | 
+| DATE      | 2025-08-07                              | 
+| TIME      | 01:01:01                                | 
+| TIMESTAMP | 2025-08-07 01:01:01 (CURRENT_TIMESTAMP) | 
 
 ## Basic usage examples
 
@@ -141,6 +167,32 @@ WITH (
    engine = 'InnoDB',
    primary_key = ARRAY['key5','key1'],
    unique_key = ARRAY['unique_key1:key2','unique_key2:key4,key3']
+);
+```
+
+Create a new table named `table_column_properties` in schema `mysql_test.database_01` with auto_increment and default.
+
+```sql
+CREATE TABLE mysql_test.database_01.table_column_properties(
+    key1 INT NOT NULL WITH (auto_increment=true),
+    f1 VARCHAR(200) WITH (default='VARCHAR'),
+    f2 CHAR(20) WITH (default='CHAR') ,
+    f4 DECIMAL(10, 3) WITH (default='0.3') ,
+    f5 REAL WITH (default='0.3') ,
+    f6 DOUBLE WITH (default='0.3') ,
+    f8 TINYINT WITH (default='1') ,
+    f9 SMALLINT WITH (default='1') ,
+    f10 INT WITH (default='1') ,
+    f11 INTEGER WITH (default='1') ,
+    f12 BIGINT WITH (default='1'),
+    f13 DATE WITH (default='2024-04-01'),
+    f14 TIME WITH (default='08:00:00'),
+    f15 TIMESTAMP WITH (default='2012-12-31 11:30:45'),
+    f16 TIMESTAMP WITH TIME ZONE WITH (default='2012-12-31 11:30:45'),
+    f17 TIMESTAMP WITH TIME ZONE WITH (default='CURRENT_TIMESTAMP')
+)
+WITH (
+   primary_key = ARRAY['key1']
 );
 ```
 
--- a/trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/JdbcColumnDefaultValueConverter.java
+++ b/trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/JdbcColumnDefaultValueConverter.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.trino.connector.catalog.jdbc;
+
+import static org.apache.gravitino.rel.Column.DEFAULT_VALUE_NOT_SET;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import org.apache.gravitino.rel.expressions.Expression;
+import org.apache.gravitino.rel.expressions.FunctionExpression;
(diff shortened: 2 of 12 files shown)


hqbhoho:
@yuqi1129 @diqiu50 ​​I've developed a rough version.​​ PTAL when you have time, thanks!

Here are two issues:
- ​​UnparsedExpression​​: This may occur because the server only supports parsing default values for specific data types or special default value expressions.

- Boolean-type constants​​: The server currently does not support them. It uses the BIT type for table creation, and the default value should be either the number 1 or its binary representation b'1'. However, the server fails to pass these values correctly and erroneously adds quotes (e.g. '1') in generated table creation statements.

Please give me some suggestions.

diqiu50:
When creating a table, if an UnparsedExpression occurs, throwing an exception should be sufficient.
For SHOW CREATE TABLE, simply display the column names and types along with other basic information to ensure the query works properly.

Boolean-type constants​​，  think performing some conversions should be sufficient.

hqbhoho:
@diqiu50  In current version, I will warp with `UnparseExpression` like this
```
   int_col_2 integer WITH ( default = 'UnparsedExpression(rand())' )
```
So it is better to support `SHOW CREATE TABLE` and `CREATE TABLE` only for constant default values,  even if it differs from the actual definition in MySQL, we can document this limitation.

Another question, if default value is null，`SHOW CREATE TABLE` whether to display null value like this
```
int_col_2 integer WITH ( default = 'NULL' )
```
I think this is bloated code.

Only for constant default values maybe like this:
```
    # MySQL  create table 
    col_int int default (rand()),
    col_tinyint TINYINT DEFAULT 0,
    col_smallint SMALLINT DEFAULT NULL,
    
    # Trino show table
    col_int int,
    col_tinyint tinyint WITH ( default = '0' ),
    col_smallint smallint,    
```

yuqi1129:
> @yuqi1129 @diqiu50 ​​I've developed a rough version.​​ PTAL when you have time, thanks!
> 
> Here are two issues:
> 
> * ​​UnparsedExpression​​: This may occur because the server only supports parsing default values for specific data types or special default value expressions.
> * Boolean-type constants​​: The server currently does not support them. It uses the BIT type for table creation, and the default value should be either the number 1 or its binary representation b'1'. However, the server fails to pass these values correctly and erroneously adds quotes (e.g. '1') in generated table creation statements.
> 
> Please give me some suggestions.

- We only need to make exceptions and do not support this type of default value.
- Is there big difficulty to convert string `1` to boolean true? Can you provide more details about this problem?

hqbhoho:
> @diqiu50 In current version, I will warp with `UnparseExpression` like this
> 
> ```
>    int_col_2 integer WITH ( default = 'UnparsedExpression(rand())' )
> ```
> 
> So it is better to support `SHOW CREATE TABLE` and `CREATE TABLE` only for constant default values, even if it differs from the actual definition in MySQL, we can document this limitation.
> 
> Another question, if default value is null，`SHOW CREATE TABLE` whether to display null value like this
> 
> ```
> int_col_2 integer WITH ( default = 'NULL' )
> ```
> 
> I think this is bloated code.
> 
> Only for constant default values maybe like this:
> 
> ```
>     # MySQL  create table 
>     col_int int default (rand()),
>     col_tinyint TINYINT DEFAULT 0,
>     col_smallint SMALLINT DEFAULT NULL,
>     
>     # Trino show table
>     col_int int,
>     col_tinyint tinyint WITH ( default = '0' ),
>     col_smallint smallint,    
> ```
> 

@yuqi1129  Is the above solution reasonable?

Boolean value will be resloved by passing a defaule value like `Literals.of(1, BooleanType.get())`

diqiu50:
> > Only for constant default values maybe like this:
> > ```
> >     # MySQL  create table 
> >     col_int int default (rand()),
> >     col_tinyint TINYINT DEFAULT 0,
> >     col_smallint SMALLINT DEFAULT NULL,
> >     
> >     # Trino show table
> >     col_int int,
> >     col_tinyint tinyint WITH ( default = '0' ),
> >     col_smallint smallint,    
> > ```
> 
I think this is fine. The result of SHOW CREATE TABLE is consistent with what Trino’s native MySQL catalog shows. Could you please explain your concern?

hqbhoho:
> I think this is fine. The result of SHOW CREATE TABLE is consistent with what Trino’s native MySQL catalog shows. Could you please explain your concern?

Currently, Trino’s native MySQL catalog not support column properties. I agree with this approach which only display constant default values. We can document this limitation, otherwise, users may question why default values are displayed for some fields but not others.

@diqiu50 Thank you for feedback!

hqbhoho:
@yuqi1129 @diqiu50 PTAL when you have time, thanks!

docs/trino-connector/catalog-mysql.md
```diff
@@ -16,7 +16,7 @@ To connect to MySQL, you need:
 
 ## Create table
 
-At present, the Apache Gravitino Trino connector only supports basic MySQL table creation statements, which involve fields, null allowances, comments, primary keys and indexes. However, it does not support advanced features like default values and auto-increment.
+At present, the Apache Gravitino Trino connector only supports basic MySQL table creation statements, which involve fields, null allowances, comments, primary keys, indexes, default values and auto-increment.
```
diqiu50, line 20: Please remove the unnecessary whitespace.

trino-connector/integration-test/src/test/resources/trino-ci-testset/testsets/jdbc-mysql/00009_datatype_mapping.sql
```diff
@@ -2,3 +2,7 @@
 show create table gt_mysql.gt_mysql_test_all_type.demo;
 
 select * from gt_mysql.gt_mysql_test_all_type.demo;
+
+drop table gt_mysql.gt_mysql_test_all_type.demo;
+
+drop schema gt_mysql.gt_mysql_test_all_type
```
diqiu50: Why is there a table deletion here, but table creation is not placed in this file?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/JdbcColumnDefaultValueConverter.java
```diff
@@ -0,0 +1,85 @@
...
+        return literal.value().toString();
+      }
+    }
+
+    // todo support UnparseExpression default value convert
+    return null;
+  }
+
+  public Expression toGravitino(Type columnType, String columnDefaultValue, boolean nullable) {
+    return DEFAULT_VALUE_NOT_SET;
```
diqiu50, line 83: Why does this always return DEFAULT_VALUE_NOT_SET here? It was previously mentioned that constants are supported.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/mysql/MySQLMetadataAdapter.java
```diff
@@ -174,6 +197,38 @@ private static String[][] convertIndexFieldNames(Set<String> fieldNames) {
...
+   * The auto-increment column will be verified. There can only be one auto-increment column and it
+   * must be the primary key or unique index.
+   *
+   * @param columns table auto increment columns
+   * @param indexes table indexes
+   */
+  protected static void validateIncrementCol(List<GravitinoColumn> columns, Index[] indexes) {
+    if (columns.size() > 1) {
+      throw new TrinoException(NOT_SUPPORTED, "Only one column can be auto-incremented in MySQL");
+    }
```
diqiu50, line 212: Add comment to explain the logic

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/mysql/MysqlColumnDefaultValueConverter.java
```diff
@@ -0,0 +1,102 @@
...
+        case DECIMAL:
+          Types.DecimalType decimalType = (Types.DecimalType) type;
+          return Literals.decimalLiteral(
+              Decimal.of(columnDefaultValue, decimalType.precision(), decimalType.scale()));
+        case DATE:
+          return Literals.dateLiteral(LocalDate.parse(columnDefaultValue, DATE_FORMATTER));
+        case TIME:
+          return Literals.timeLiteral(LocalTime.parse(columnDefaultValue, TIME_FORMATTER));
+        case TIMESTAMP:
+          if (columnDefaultValue.toUpperCase().startsWith(CURRENT_TIMESTAMP)) {
```
diqiu50: Why are we using startsWith(CURRENT_TIMESTAMP)? What‘s the actual value?

trino-connector/integration-test/src/test/resources/trino-ci-testset/testsets/jdbc-mysql/00009_datatype_mapping.sql
hqbhoho: Table create by integration-test-common/docker-script/init/mysql/init.sql
 
 Trino create table can not cover all mysql data type.  This table only test all mysql data type mapping, so it can be dropped after this test.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/JdbcColumnDefaultValueConverter.java
hqbhoho: `MysqlColumnDefaultValueConverter` extends `JdbcColumnDefaultValueConverter` and will overwrite `toGravitino `
Here only a default implementation, if all jdbc catalog have same implementation, we can overwrite this method in  `JdbcColumnDefaultValueConverter` . This PR only check and support MySQL.

Alternatively, this time we can implement it in `JdbcColumnDefaultValueConverter` instead of `MysqlColumnDefaultValueConverter`. What do you think?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/mysql/MysqlColumnDefaultValueConverter.java
hqbhoho: In MySQL
```
+-------+--------------+------+-----+----------------------+-------------------+
| Field | Type         | Null | Key | Default              | Extra             |
+-------+--------------+------+-----+----------------------+-------------------+
| f1    | timestamp    | NO   |     | CURRENT_TIMESTAMP    | DEFAULT_GENERATED |
| f2    | timestamp(6) | NO   |     | CURRENT_TIMESTAMP(6) | DEFAULT_GENERATED |
+-------+--------------+------+-----+----------------------+-------------------+
```
In Gravitino Server support TimestampType with default value 'CURRENT_TIMESTAMP'

Here should use `equal`. Currently, TimestampType is without precision.

docs/trino-connector/catalog-mysql.md
hqbhoho: done

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/mysql/MySQLMetadataAdapter.java
hqbhoho: done

trino-connector/integration-test/src/test/resources/trino-ci-testset/testsets/jdbc-mysql/00009_datatype_mapping.sql
diqiu50: In that case, you just don’t need to delete it.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/jdbc/JdbcColumnDefaultValueConverter.java
diqiu50: OK

trino-connector/integration-test/src/test/resources/trino-ci-testset/testsets/jdbc-mysql/00009_datatype_mapping.sql
hqbhoho: okk I will remove later.

PR #7877, opened 2025-07-31
[#7922] feat (server):  Client-related releases jars compatible with JDK 8

diqiu50:

Make client-related releases jars compatible with JDK 8

Fix: #7922

Yes, Client and common modules support JDK8 runtime

UT and IT


build.gradle.kts
```diff
@@ -284,10 +346,12 @@ subprojects {
           vendor.set(JvmVendorSpec.AMAZON)
         }
         languageVersion.set(JavaLanguageVersion.of(17))
-      } else {
-        languageVersion.set(JavaLanguageVersion.of(extra["jdkVersion"].toString().toInt()))
+      } else if (CompatibleWithJDK8(project)) {
+        languageVersion.set(JavaLanguageVersion.of(17))
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
```
FANNG1, line 334: this will make the client IT failures for the client refers to JDK 17 server module

diqiu50: My idea is to use JDK 17 for normal builds and running tests, without generating JDK 8 compatible code. When the release8 parameter is added, JDK 8 compatible code will be generated for the client build, and tests will be skipped in this case.


Build & Release Process Summary
The build process includes two steps:
1. Build with Java 17 (full build with tests)
```
./gradlew build
```

Runs all tests
Packages all modules
Produces Java 17 bytecode

2. Build Java 8-compatible client (skip tests)
```
./gradlew build -Drelease8 -x test
```
Skips tests
Builds only the client module
Produces Java 8-compatible artifacts

FANNG1: when publish to Maven, we should build server packages too. suggest to use one command to build server and client packages to maven.

diqiu50: Without -Drelease8, all modules are built (including tests). When -Drelease8 is added, only client-related modules are built with JDK 8 compatibility, and tests are skipped.

gradle.properties
```diff
@@ -31,7 +31,7 @@ SONATYPE_PASSWORD = password
 
 # jdkVersion is used to specify the version of JDK to build and test Gravitino, current
 # supported version is 8, 11, and 17.
-jdkVersion = 8
```
FANNG1: jdkVersion is not used anymore, how about removing it?

build.gradle.kts
```diff
@@ -273,21 +273,78 @@ subprojects {
...
+    }
+  }
+
   java {
     toolchain {
       // Some JDK vendors like Homebrew installed OpenJDK 17 have problems in building trino-connector:
       // It will cause tests of Trino-connector hanging forever on macOS, to avoid this issue and
       // other vendor-related problems, Gravitino will use the specified AMAZON OpenJDK 17 to build
       // Trino-connector on macOS.
+      val isRelease8 = project.hasProperty("release8")
```
FANNG1: `release8` is something like we release all packages compatibility with JDK8, but the fact is we release server pacakge to JDK17 and client packages to JDK8, we need a more meaning name.

FANNG1:
Could you add release parameter before publish to maven in release script?  And add an extra workflow to ensure the release mode is workable?  https://github.com/apache/gravitino/blob/0f010bd26d0db23691f407e30a2005e6da70e06e/dev/release/release-build.sh#L339-L340

build.gradle.kts
diqiu50: fixed

FANNG1:
`clientCompatibleWithJDK8` seems complicated, others LGTM, @jerryshao do you have time to take another look?

jerryshao:
This configuration can be a built-in configuration, we don't have to let the user know or set it explicitly.

build.gradle.kts
```diff
@@ -64,21 +62,12 @@ plugins {
...
-
 val scalaVersion: String = project.properties["scalaVersion"] as? String ?: extra["defaultScalaVersion"].toString()
 if (scalaVersion !in listOf("2.12", "2.13")) {
   throw GradleException("Scala version $scalaVersion is not supported.")
 }
 
-project.extra["extraJvmArgs"] = if (extra["jdkVersion"] in listOf("8", "11")) {
-  listOf()
-} else {
+project.extra["extraJvmArgs"] =
```
FANNG1, line 68: In `clientCompatibleWithJDK8 ` mode,  the client modules should not use the following JDK17 vm options.

FANNG1:
> This configuration can be a built-in configuration, we don't have to let the user know or set it explicitly.

To clarify the meaning of the configuration you referred here,  is it `clientCompatibleWithJDK8` or `jdkVersion`?

build.gradle.kts
diqiu50: These parameters are not used at compile time. they are needed during test execution.

diqiu50:
> should you update the codes:
> 
> https://github.com/apache/gravitino/blob/2b88148375a80a7e1034cc7d8aa8f2793293c3d0/build.gradle.kts#L132-L135

After modifying this, many files will be reformatted. It’s better to change it in a later PR.

jerryshao:
> > should you update the codes:
> > https://github.com/apache/gravitino/blob/2b88148375a80a7e1034cc7d8aa8f2793293c3d0/build.gradle.kts#L132-L135
> 
> After modifying this, many files will be reformatted. It’s better to change it in a later PR.

This will affect lots of code, we don't have to upgrade it until necessary.

build.gradle.kts
```diff
@@ -44,14 +44,12 @@ plugins {
...
-  if (JavaVersion.current() >= JavaVersion.VERSION_1_8 &&
-    JavaVersion.current() <= JavaVersion.VERSION_17
-  ) {
+  if (JavaVersion.current() >= JavaVersion.VERSION_17) {
     alias(libs.plugins.spotless)
   } else {
     throw GradleException(
       "The Gravitino Gradle toolchain currently does not support " +
-        "Java version ${JavaVersion.current()}. Please use JDK versions 8 through 17."
+        "Java version ${JavaVersion.current()}. Please use JDK version 17 or higher."
```
mchades: Is JDK 11 also not allowed?

.github/workflows/build.yml
```diff
@@ -63,7 +63,9 @@ jobs:
           cache: 'gradle'
 
       - name: Build with Gradle
-        run: ./gradlew build -x test -PjdkVersion=8
+        run: |
+          ./gradlew build -x test
+          ./gradlew build -PclientCompatibleWithJDK8 -x test
```
mchades: Does JDK11 need to be tested?

catalogs/catalog-common/src/main/java/org/apache/gravitino/credential/config/AzureCredentialConfig.java
```diff
@@ -84,32 +83,26 @@ public AzureCredentialConfig(Map<String, String> properties) {
     loadFromMap(properties, k -> true);
   }
 
-  @NotNull
```
mchades: why remove the annotation?

bundles/aliyun/build.gradle.kts
```diff
@@ -26,13 +26,13 @@ plugins {
...
 dependencies {
   compileOnly(project(":api"))
-  compileOnly(project(":catalogs:catalog-common"))
-  compileOnly(project(":catalogs:catalog-fileset"))
-  compileOnly(project(":core"))
   compileOnly(libs.hadoop3.client.api)
   compileOnly(libs.hadoop3.client.runtime)
   compileOnly(libs.hadoop3.oss)
 
+  implementation(project(":common")) {
```
mchades, line 33: They used `compileOnly` before. Why do we need `implementation` now?

docs/how-to-build.md
```diff
@@ -12,14 +12,14 @@ license: "This software is licensed under the Apache License version 2."
 
 + Linux or macOS operating system
 + Git
-+ A Java Development Kit, version 8 to 17, installed in your environment to launch Gradle
++ A Java Development Kit, version 17 or higher, installed in your environment to launch Gradle
```
mchades: Do we support a JDK version higher than JDK 17?

The code comment in `build.gradle.kts` states that Spotless does not work with JDK 21:

>  // Spotless version < 6.19.0 (https://github.com/diffplug/spotless/issues/1819) has an issue
>   // running against JDK21, but we cannot upgrade the spotless to 6.19.0 or later since it only
>   // support JDK11+. So we don't support JDK21 and thrown an exception for now.

docs/how-to-build.md
```diff
@@ -12,14 +12,14 @@ license: "This software is licensed under the Apache License version 2."
...
 + Python 3.8, 3.9, 3.10, 3.11, or 3.12 to build the Gravitino Python client
 + Optionally, Docker to run integration tests
 
 :::info Please read the following notes before trying to build Gravitino.
 
-+ Gravitino requires a minimum of JDK8 and supports up to JDK17 to run Gradle, so you need to install a JDK, version 8 to 17, to launch the build environment.
-+ Gravitino itself supports using JDK 8, 11, or 17 to build. The Gravitino Trino connector uses JDK17 to build (to avoid vendor-related issues on some platforms, Gravitino uses the specified Amazon Corretto OpenJDK 17 to build the Trino connector on macOS).
++ Gravitino requires a minimum of JDK17 to run Gradle, so you need to install a JDK17 to launch the build environment.
```
mchades, line 21: What is the exact meaning of "Gravitino"? Does it refer to both the client and the server? Can JDK 11 be used to set up the server?

build.gradle.kts
diqiu50: Yes

docs/how-to-build.md
diqiu50: JDK 17 is minimum

.github/workflows/build.yml
diqiu50: No, it's not supported

docs/how-to-build.md
diqiu50: Yes, but using a newer version of the JDK may lead to compatibility issues. and upgrade Spotless need to reformat code.
I removed some  comments

bundles/aliyun/build.gradle.kts
diqiu50: @yuqi1129 Is it more appropriate to use compileOnly or implementation here

catalogs/catalog-common/src/main/java/org/apache/gravitino/credential/config/AzureCredentialConfig.java
diqiu50: I prefer not to introduce too many dependencies into the common module. But if this one is necessary, I’ll add it

docs/how-to-build.md
```diff
@@ -45,16 +45,15 @@ license: "This software is licensed under the Apache License version 2."
    ./gradlew build
    ```
 
-   The default specified JDK version is 8, but if you want to use JDK 11 or 17 to build, modify the property `jdkVersion` to 11 or 17 in the `gradle.properties` file, or specify the version with `-P`, like:
-
+   The default specified JDK version is 17, but if you want to use a different JDK version , update the `jdkVersion` property in gradle.properties, or pass it via the -P option, for example:
    ```shell
    ./gradlew build -PjdkVersion=11
```
mchades: you have said we don't support JDK11 - https://github.com/apache/gravitino/pull/7877#discussion_r2262003742

diqiu50: remove it

docs/how-to-build.md
yuqi1129: Yeah, you should test whether JDK21 or JDK25 is better.

docs/how-to-build.md
```diff
@@ -12,14 +12,14 @@ license: "This software is licensed under the Apache License version 2."
...
++ A Java Development Kit, version 17 or higher, installed in your environment to launch Gradle
 + Python 3.8, 3.9, 3.10, 3.11, or 3.12 to build the Gravitino Python client
 + Optionally, Docker to run integration tests
 
 :::info Please read the following notes before trying to build Gravitino.
 
-+ Gravitino requires a minimum of JDK8 and supports up to JDK17 to run Gradle, so you need to install a JDK, version 8 to 17, to launch the build environment.
-+ Gravitino itself supports using JDK 8, 11, or 17 to build. The Gravitino Trino connector uses JDK17 to build (to avoid vendor-related issues on some platforms, Gravitino uses the specified Amazon Corretto OpenJDK 17 to build the Trino connector on macOS).
++ Gravitino requires a minimum of JDK17 to run Gradle, so you need to install a JDK17 to launch the build environment.
++ Gravitino itself supports using JDK 17 to build. The Gravitino Trino connector uses JDK17 to build (to avoid vendor-related issues on some platforms, Gravitino uses the specified Amazon Corretto OpenJDK 17 to build the Trino connector on macOS).
```
yuqi1129: >  Gravitino uses the specified Amazon Corretto OpenJDK 17 to build the Trino connector on macOS

Should the build and release task be on the Linux release?

diqiu50: Not necessarily, but recommended.

catalogs/catalog-common/src/main/java/org/apache/gravitino/credential/config/AzureCredentialConfig.java
mchades: @FANNG1 cc

common/build.gradle.kts
```diff
@@ -36,6 +36,7 @@ dependencies {
   implementation(libs.jackson.datatype.jdk8)
   implementation(libs.jackson.datatype.jsr310)
   implementation(libs.jackson.databind)
+  implementation(libs.bundles.log4j)
```
mchades: is it necessary?

docs/how-to-build.md
diqiu50: I have added a restriction to allow only JDK 17.

common/build.gradle.kts
diqiu50: Yes, the Config need this libs

.github/workflows/build.yml
jerryshao: So we need to build twice to generate jdk 8 compilant package?

build.gradle.kts
```diff
@@ -273,19 +259,78 @@ subprojects {
...
       // It will cause tests of Trino-connector hanging forever on macOS, to avoid this issue and
       // other vendor-related problems, Gravitino will use the specified AMAZON OpenJDK 17 to build
       // Trino-connector on macOS.
+      val clientCompatibleWithJDK8 = project.hasProperty("clientCompatibleWithJDK8")
       if (project.name == "trino-connector") {
         if (OperatingSystem.current().isMacOsX) {
           vendor.set(JvmVendorSpec.AMAZON)
         }
         languageVersion.set(JavaLanguageVersion.of(17))
+      } else if (clientCompatibleWithJDK8 && CompatibleWithJDK8(project)) {
```
jerryshao: Do we need a "clientCompatibleWithJDK8" to control whether to generate jdk8 compatible package?

bundles/aliyun/build.gradle.kts
jerryshao: CC @yuqi1129

common/build.gradle.kts
jerryshao: We need to exclude all the log4j related dependencies in the client side, include clients and connectors. Otherwise, it will be conflicted with different engines and make the log not work.

bundles/aliyun/build.gradle.kts
yuqi1129: `compileOnly` should be fine as vendor-specific jars always come with `hadoop3-client-runtime`, which already contains `common.`

yuqi1129: `compileOnly` should be fine as vendor-specific jars always come with `hadoop3-client-runtime`, which already contains `common.`

.github/workflows/build.yml
diqiu50: This is used to verify whether the generated JDK 8-compliant package can run, in order to prevent users from using features in these modules that require a JDK higher than 8.

build.gradle.kts
diqiu50: Yes, in clientCompatibleWithJDK8 mode, integration tests for some modules cannot be compiled. This option can be removed once the issue is resolved, making JDK 8 compatibility the default behavior for the client module.

bundles/aliyun/build.gradle.kts
diqiu50: `hadoop3-client-runtime` should not contain common (Gravitino).

common/build.gradle.kts
diqiu50: Use `implementation(libs.slf4j.api)` instead of  `implementation(libs.bundles.log4j)`

build.gradle.kts
jerryshao: Why don't you fix those issues, I don't think it needs additional property to control.

catalogs/catalog-common/src/main/java/org/apache/gravitino/credential/config/AzureCredentialConfig.java
FANNG1: I perfer to keep it, as the dependence is lightweight.

docs/how-to-build.md
```diff
@@ -177,7 +171,7 @@ sudo apt install apt-transport-https ca-certificates curl software-properties-co
 
 Updating the package list ensures you have the latest information on the newest versions of packages and dependencies. Installing the necessary packages lets your system download and manage additional software securely.
 
-### Download and setup Java SDK 17 (11 or 8 also works)
+### Download and setup Java SDK 17
 
 **On Ubuntu (WSL):**
```
yuqi1129, line 171: L184 also needs to be revised.

build.gradle.kts
diqiu50: The issue is that the `embedded` mode in integration tests depends on the server and core modules. This causes all modules to require the target to be set to JDK 8. To remove this requirement, the integration tests need to be modified according to the approach @FANNG1  used previously.

diqiu50: I think the changes to the integration tests can be postponed until later.

docs/how-to-build.md
diqiu50: fixed

build.gradle.kts
diqiu50: I think we cannot remove the `clientCompatibleWithJDK8` parameter. How do you want to handle it ? @jerryshao  Do you want to add a `./gradlew release` command to automatically set this parameter for compilation and packaging?

FANNG1: `./gradlew release` seems clearer to build the release package

FANNG1: @jerryshao ,  what do you think about adding a special task like `./gradlew release` to build JDK8 for clients and JDK17 for server modules?

jerryshao: Yeah, that's better. For example, if we compile/assemble the package, or if we publish the maven artifact, we should ship the jdk8-compatible client package by default. If we run the test/build, we can build all with jdk17. We don't need a flag to control the behavior.

catalogs/catalog-common/build.gradle.kts
```diff
@@ -23,8 +23,10 @@ plugins {
 
 // try to avoid adding extra dependencies because it is used by catalogs and connectors.
 dependencies {
+  implementation(project(":common"))
```
FANNG1, line 26: any reason about adding the dependencies to `common`?

catalogs/catalog-common/build.gradle.kts
```diff
@@ -23,8 +23,10 @@ plugins {
...
   implementation(libs.commons.lang3)
   implementation(libs.guava)
+  implementation(libs.jakarta.validation.api)
   implementation(libs.slf4j.api)
```
FANNG1, line 30: Why moving `CredentialConfig` to `catalog-common`?

common/build.gradle.kts
```diff
@@ -36,6 +36,8 @@ dependencies {
   implementation(libs.jackson.datatype.jdk8)
   implementation(libs.jackson.datatype.jsr310)
   implementation(libs.jackson.databind)
+  implementation(libs.jakarta.validation.api)
```
FANNG1, line 39: why adding `libs.jakarta.validation.api` and `libs.slf4j.api`?

.github/workflows/build.yml
```diff
@@ -63,7 +63,9 @@ jobs:
           cache: 'gradle'
 
       - name: Build with Gradle
-        run: ./gradlew build -x test -PjdkVersion=8
+        run: |
+          ./gradlew build -x test
+          ./gradlew release -x test
```
FANNG1, line 67: Is it possible to skip the test automatically when running `./gradlew release`, since the test couldn't run in release mode.

FANNG1, line 67: Could we use a seperate step like follow?
```
- name: Build release jar
  run: | 
     ./gradlew release
```

mchades: does `release` already includes `build`?

FANNG1: There is something different, `build` build all packages with JDK17, while `release` builds client modules for JDK8

common/build.gradle.kts
diqiu50: The `Config.java` need The `libs.slf4j.api`, The `@NotNull` need `jakarta.validation`

.github/workflows/build.yml
diqiu50: I tried it, but it didn’t work.

catalogs/catalog-common/build.gradle.kts
diqiu50: Some subclass like `OSSCredentialConfig` denpen on the `Config` in the `common`

catalogs/catalog-common/build.gradle.kts
diqiu50: XXCredentialConfig should belong to `catalog-common`

FANNG1:
LGTM to the main part, except some code replacements

diqiu50:
@jerryshao  Please take a review

build.gradle.kts
```diff
@@ -64,21 +60,12 @@ plugins {
   alias(libs.plugins.errorprone)
 }
 
-if (extra["jdkVersion"] !in listOf("8", "11", "17")) {
```
jerryshao, line 67: Did you remove this `jdkVersion` property.

build.gradle.kts
```diff
@@ -273,6 +259,65 @@ subprojects {
     mavenLocal()
   }
 
+  fun CompatibleWithJDK8(project: Project): Boolean {
```
jerryshao: Why do you capitalize "C" for this method?

build.gradle.kts
```diff
@@ -273,6 +259,65 @@ subprojects {
...
+
+    return false
+  }
+
+  tasks.register("printJvm") {
+    group = "help"
+    description = "print JVM information"
+
+    doLast {
+
```
jerryshao: Remove this empty line.

build.gradle.kts
```diff
@@ -1022,3 +1071,17 @@ fun checkOrbStackStatus() {
...
+tasks.register("release") {
+  group = "release"
+  description = "Builds and package a release version."
+  doFirst {
+    println("Releasing project...")
+  }
+
+  doLast {
+    println("Release build completed")
+  }
```
jerryshao: Remove unnecessary logs.

docs/how-to-build.md
```diff
@@ -45,16 +45,10 @@ license: "This software is licensed under the Apache License version 2."
    ./gradlew build
    ```
 
-   The default specified JDK version is 8, but if you want to use JDK 11 or 17 to build, modify the property `jdkVersion` to 11 or 17 in the `gradle.properties` file, or specify the version with `-P`, like:
+   If you want to use the client module with JDK 8, you can specify the version using -PclientCompatibleWithJDK8, for example:
```
jerryshao: Do you still have this "clientCompatibleWithJDK8" property?

build.gradle.kts
```diff
@@ -1022,3 +1071,17 @@ fun checkOrbStackStatus() {
 }
 
 printDockerCheckInfo()
+
+tasks.register("release") {
```
jerryshao, line 1074: What's the usage of this task, and how do we use it?

.github/workflows/build.yml
```diff
@@ -63,7 +63,12 @@ jobs:
           cache: 'gradle'
 
       - name: Build with Gradle
-        run: ./gradlew build -x test -PjdkVersion=8
+        run: |
+          ./gradlew build -x test
+
+      - name: Build release jars
+        run: |
+          ./gradlew release -x test
```
jerryshao: Why do we need "build" first, then "release", is "release" alone enough?

build.gradle.kts
diqiu50: Yes

build.gradle.kts
diqiu50: Build the project compatible with JDK 8. Run the command ./gradlew release.

.github/workflows/build.yml
diqiu50: It seems that it can be removed.

docs/how-to-build.md
diqiu50: removed

docs/how-to-build.md
yuqi1129: > Gravitino uses the specified Amazon Corretto OpenJDK 17 to build the Trino connector on macOS

It's recommended to use Amazon Corretto OpenJDK 17 to build Gravitino.

yuqi1129 (commented):
I have no more comments except a minor one.

jerryshao:
@yuqi1129 @mchades @FANNG1 please review again.

.github/workflows/build.yml
```diff
@@ -63,7 +63,8 @@ jobs:
           cache: 'gradle'
 
       - name: Build with Gradle
```
FANNG1, line 65: `Build with Gradle` -> Build Gravitino package in release mode?

FANNG1 (approved):

.github/workflows/build.yml
diqiu50: I think `Build with Gradle` is OK

Issue #7440, opened 2025-06-19
[Subtask] Support for altering model version with multiple URIs


xiaozcy:
### Describe the subtask

Support for altering model version with multiple URIs

### Parent issue

#7363

PR #7881, opened 2025-07-31
[#7440] feat(model): Support for altering model version with multiple URIs

xiaozcy:

Support for altering model version with multiple URIs.

Fix: #7440

No.

UT.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/model/ModelVersionChange.java
+++ b/api/src/main/java/org/apache/gravitino/model/ModelVersionChange.java
@@ -335,8 +335,7 @@ public interface ModelVersionChange {
      * @param newUri The new uri to be set for the model version.
      */
     public UpdateUri(String newUri) {
-      this.uriName = ModelVersion.URI_NAME_UNKNOWN;
-      this.newUri = newUri;
+      this(ModelVersion.URI_NAME_UNKNOWN, newUri);
     }
 
     /**


xiaozcy:
Hi @jerryshao , please help review it when you are free, thx.

api/src/main/java/org/apache/gravitino/model/ModelVersionChange.java
```diff
@@ -302,6 +335,18 @@ final class UpdateUri implements ModelVersionChange {
...
+
+    /**
+     * Creates a new {@link UpdateUri} instance with the specified new uri and its name.
+     *
+     * @param uriName The name of the uri to be updated for the model version.
+     * @param newUri The new uri to be set for the model version.
+     */
+    public UpdateUri(String uriName, String newUri) {
+      this.uriName = uriName;
       this.newUri = newUri;
```
jerryshao, line 349: This can be implemented as `this(xxx, xxx)`

catalogs/catalog-model/src/main/java/org/apache/gravitino/catalog/model/ModelCatalogOperations.java
```diff
@@ -614,6 +625,18 @@ private void doRemoveProperty(
...
+    entityUris.replace(change.uriName(), change.newUri());
+  }
+
+  private void doAddUri(Map<String, String> entityUris, ModelVersionChange.AddUri change) {
+    entityUris.putIfAbsent(change.uriName(), change.uri());
+  }
+
+  private void doRemoveUri(Map<String, String> entityUris, ModelVersionChange.RemoveUri change) {
+    entityUris.remove(change.uriName());
+  }
```
jerryshao, line 638: What are the behaviors if:

1. The update URI name doesn't exist.
2. The added URI name already exists.
3. The removed URI doesn't exist.

Shall we throw an exception, or do it silently, what do you think?

xiaozcy: I tend to do it silently, and that's exactly how it is implemented at present:
1. use `Map#replace`  to update URI
2. use `Map#putIfAbsent` to add URI
3. use `Map#remove` to remove URI

jerryshao: You'd better give a reason. From my point, people may misuse these APIs and lead to an unexpected behavior.

jerryshao: Maybe we'd better throw the related exception if "replace" has no existing uri name, or "put" has existing name, and "delete" has no existing name.

xiaozcy: I checked similar logic, like removing the property of an entity, and none of them threw an exception in this case, maybe we should keep consistent?

jerryshao: OK, I see.

jerryshao (approved):

PR #7883, opened 2025-07-31
[#7926] feat(auth): Add OAuth support for JWKS URLs

bharos:
# [#7926] feat(auth): Add OAuth support for JWKS URLs 

This PR introduces comprehensive OAuth 2.0 JWT authentication support for Gravitino server components:
**Core OAuth Authentication Framework**:
   - `OAuth2TokenAuthenticator` - Main authenticator for JWT token validation with Bearer token support
   - `OAuthTokenValidator` interface with two implementations:
     - `StaticSignKeyValidator ` - Static key-based JWT validation (RS256/HS256)
     - `JwksTokenValidator` - Dynamic JWKS endpoint-based validation
   - `OAuthTokenValidatorFactory` - Factory pattern for validator creation based on configuration

**Enterprise Authentication Requirements**: Modern enterprise environments require OAuth 2.0/JWT authentication through JWKS URIs, for example Azure AD auth.

Fix: #7926

**Yes**, this PR introduces the following user-facing changes:

#### **New Authentication validation Method**:
This PR adds support for token validation with JWKS URIs.
- **JWT Token Requirements**: Tokens must contain:
  - Valid `sub` (subject) claim for user identification
  - Valid `aud` (audience) claim matching configured service audience
  - Valid signature using configured algorithm and key
  - Non-expired `exp` (expiration) claim

#### **API Compatibility**:
- **Backward Compatible**: Existing authentication method (with defaultSignKey) continue to work unchanged
- **Additive**: OAuth authentication is an additional option which uses JWKS URI, not a replacement

-  **Unit Tests**
- **Manual Testing**: Validated against real OAuth providers (Azure AD, custom OAuth servers)


gradle/libs.versions.toml
```diff
@@ -71,6 +71,7 @@ scala-java-compat = "1.0.2"
 sqlite-jdbc = "3.42.0.0"
 testcontainers = "1.20.6"
 jwt = "0.11.1"
+nimbus-jose-jwt = "9.37.3"
```
roryqi, line 74: You should add NOTICE and LICENSE of the jar to License.bin and NOTICE.bin files, too.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthTokenValidatorFactory.java
```diff
@@ -0,0 +1,60 @@
...
+   * @return An initialized OAuthTokenValidator
+   * @throws IllegalArgumentException if configuration is invalid
+   */
+  public static OAuthTokenValidator createValidator(Config config) {
+    if (config == null) {
+      throw new IllegalArgumentException("Configuration cannot be null");
+    }
+
+    OAuthTokenValidator validator;
+    if (StringUtils.isNotBlank(config.get(OAuthConfig.JWKS_URI))) {
```
roryqi: Maybe is it suitable to provide a configuration like `gravitino.authenticator.oauth.tokenValidatorClass`?
Another point is what difference between tokenValidator and OAuthProvider? Is `tokenValidator` enough?

jerqi (approved):

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -109,4 +109,11 @@ public interface OAuthConfig {
           .version(ConfigConstants.VERSION_1_0_0)
           .stringConf()
           .create();
+
+  ConfigEntry<String> PRINCIPAL_FIELD =
```
roryqi, line 113: Could you use camel style name? Gravitino uses camel style although Iceberg may use - style name .

roryqi: You can modify the `jwks-uri`, too.

gradle/libs.versions.toml
bharos: I already see the nimbus+jwt in here : https://github.com/apache/gravitino/blob/main/LICENSE.bin#L352
Does that mean I can skip adding this

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
bharos: Sure, I updated the other configs I added too

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthTokenValidatorFactory.java
bharos: >Maybe is it suitable to provide a configuration like gravitino.authenticator.oauth.tokenValidatorClass?

Ok, added

>Another point is what difference between tokenValidator and OAuthProvider? Is tokenValidator enough?

What do you mean by OAuthProvider ? Do you mean the OAuth2TokenAuthenticator class ?
I currently only updated the validation logic, to make it pluggable through the factory and added additional validator that supports the JWKS URI, and moved existing validation to DefaultJwtTokenValidator
I think the OAuth2TokenAuthenticator is still needed.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
```diff
@@ -0,0 +1,170 @@
...
+      if (StringUtils.isNotBlank(expectedIssuer)) {
+        expectedClaimsBuilder.issuer(expectedIssuer);
+      }
+
+      // Set expected audience if provided
+      if (StringUtils.isNotBlank(serviceAudience)) {
+        expectedClaimsBuilder.audience(serviceAudience);
+      }
+
+      com.nimbusds.jwt.proc.DefaultJWTClaimsVerifier<SecurityContext> claimsVerifier =
```
roryqi: Could we avoid using the long package name?

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
```diff
@@ -0,0 +1,170 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Generic JWKS-based OAuth token validator that uses Nimbus JOSE + JWT library to validate JWT
+ * tokens from any OAuth provider that exposes a JWKS endpoint.
+ *
+ * <p>This validator supports OAuth providers that publish their public keys via JWKS (JSON Web Key
+ * Set) endpoints
+ */
+public class JwksTokenValidator implements OAuthTokenValidator {
```
roryqi, line 48: Could we give a better name for this? I can't get the difference from the default validator by the name.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
```diff
@@ -0,0 +1,170 @@
...
+    // Try the configured principal field first
+    if (StringUtils.isNotBlank(principalField)) {
+      String principal = (String) validatedClaims.getClaim(principalField);
+      if (principal != null) {
+        return principal;
+      }
+    }
+
+    // Fallback to common user identity fields
+    String[] fallbackFields = {"unique_name", "upn", "email", "preferred_username", "sub"};
```
roryqi, line 162: I'm not sure whether to make these configurable.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
bharos: Fixed

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
bharos: Hmm ok.
I named the other class better now. So we have 2 cases for validation: one using DefaultSignKey and the other is using JWKS URI for tokens to validate. Based on this I have renamed the classes as
DefaultSignKeyValidator and JwksTokenValidator

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
bharos: Maybe I do it in next PR ? I will keep this in mind to change

gradle/libs.versions.toml
roryqi: Yes.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
roryqi: You can add a `todo` comment here.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthTokenValidatorFactory.java
roryqi: In the last PR, we have defined ` ProviderType`. Every ProviderType should have its `OAuthProvider`.  What's the relationship between `ProviderType` and `TokenValidator`? How do we use `ProviderType`?

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
roryqi: `StatisticSignKeyValidator` may be better?

roryqi:
UTs are necessary, too.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthTokenValidatorFactory.java
bharos: So, ProviderType I am planning to use only for the Web UI.
I realize that for the token validation, we don't need provider specific logic.
For web UI we can use it to give provider specific login interfaces (like MSAL for Azure Login). But backend logic can for now be either defaultSignKey or JWKS

roryqi: Maybe you should provide a config option similar to `redirect url` instead of `ProviderType` according to your words. Is it enough for your cases? You can see OAuth2TokenAuthenticator. We support enums and custom at the same time, you can see the code
```
private static final Logger LOG = LoggerFactory.getLogger(AuthenticatorFactory.class);

  public static final ImmutableMap<String, String> AUTHENTICATORS =
      ImmutableMap.of(
          AuthenticatorType.SIMPLE.name().toLowerCase(),
          SimpleAuthenticator.class.getCanonicalName(),
          AuthenticatorType.OAUTH.name().toLowerCase(),
          OAuth2TokenAuthenticator.class.getCanonicalName(),
          AuthenticatorType.KERBEROS.name().toLowerCase(),
          KerberosAuthenticator.class.getCanonicalName());

  private AuthenticatorFactory() {}

  public static List<Authenticator> createAuthenticators(Config config) {
    List<String> authenticatorNames = config.get(Configs.AUTHENTICATORS);

    List<Authenticator> authenticators = Lists.newArrayList();
    for (String name : authenticatorNames) {
      String className = AUTHENTICATORS.getOrDefault(name, name);
      try {
        Authenticator authenticator =
            (Authenticator) Class.forName(className).getDeclaredConstructor().newInstance();
        authenticators.add(authenticator);
      } catch (Exception e) {
        LOG.error("Failed to create and initialize Authenticator by name {}.", name, e);
        throw new RuntimeException("Failed to create and initialize Authenticator: " + name, e);
      }
    }
    return authenticators;
  }
```
Maybe we should provide `JWKS` UI and add some redirect url to finish the feature.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
bharos: Done

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
bharos: Done

bharos:
> UTs are necessary, too.

@jerqi Added unit tests. PTAL

jerqi (approved):
LGTM.

roryqi:
Could you create some subtasks issues? We shouldn't reuse an issue.

bharos:
> Could you create some subtasks issues? We shouldn't reuse an issue.

Thanks for the review!
Added subtask #7926 , updated the PR description

roryqi:
Thanks, merged.

PR #7905, opened 2025-08-03
[SIP]  Support Gravitino MCP server

FANNG1:

Fix: #(issue)


FANNG1:
@jerryshao @shaofengshi PTAL, thx

FANNG1:
[uv.lock](https://github.com/apache/gravitino/pull/7905/files#diff-e55a83edf870925730735751614e4a26912314804cc14555ce069bb77533ce9f) and license header files  eat up too much code line, the real code is several hundreds

mcp-server/mcp_server/server.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @asynccontextmanager
+    async def app_lifespan(server: FastMCP) -> AsyncIterator[GravitinoContext]:
+        logging.info(f"Add Gravitino context: {gravitino_context}")
+        yield gravitino_context
+
+    return app_lifespan
+
+
+def create_gravition_mcp(setting: Setting) -> FastMCP:
```
yuqi1129: Typo gravition

mcp-server/mcp_server/tools/table.py
```diff
@@ -0,0 +1,227 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from fastmcp import Context, FastMCP
+
+
+def load_table_tools(mcp: FastMCP):
+    @mcp.tool(tags={"table"})
```
yuqi1129, line 22: I'm not entirely sure if this type of division is correct. We need to support more tags like model, fileset, topic, etc.

yuqi1129:
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':mcp-server:checkSystemPython'.
> Python version 3.9 is too old. Requires Python >= 3.10. Found: Python 3.9.6


The module depends on the Python version installed on the compiling machine physicall, which is not user-friendly, I don't think it's easy to upgrade the python version of my Mac.

mcp-server/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. Python 3.10+
+2. uv is installed. You can install uv by following the [official guide](https://docs.astral.sh/uv/getting-started/installation/).
+
+### How to run:
+
+1. Clone the code from GitHub, and change to `mcp-server` directory
+2. Create virtual environment, `uv venv`
+3. Install the required Python packages. `uv pip install -e .`
+4. Add Gravitino MCP server to corresponding LLM tools, use local MCP server like:
```
yuqi1129: You should include detailed steps for how users can use it, for example, how we use it in cursor, goole, .etc.

mcp-server/uv.lock
```diff
@@ -0,0 +1,938 @@
+version = 1
+revision = 2
```
yuqi1129, line 2: Is this file created manually? it's too big to mantain

FANNG1: It's created automatically.  Most Python projects include this file, it's requried.

mcp-server/README.md
FANNG1: Different LLM tools may have different ways to  add MCP server configuration, but they all use the same mcp server configurations as my test for cursor, claud, trae.  Adding such detailed steps for diverse LLMs is not a good way and hard to maintance, may be we could use one LLM tools as example, WDYT?

mcp-server/mcp_server/tools/table.py
FANNG1: It dependes on how should we export tools,  in some senerios , the user may doesn't want to use table tools but use policy, tag tools.

mcp-server/mcp_server/server.py
FANNG1: updated

mcp-server/README.md
yuqi1129: Yes, that is what I want to express. It will lower the threshold to use our MCP server for new comers.

FANNG1:
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
>   Execution failed for task ':mcp-server:checkSystemPython'.
> 
> > Python version 3.9 is too old. Requires Python >= 3.10. Found: Python 3.9.6
> 
> The module depends on the Python version installed on the compiling machine physicall, which is not user-friendly, I don't think it's easy to upgrade the python version of my Mac.

I had removed the python check, and uv venv will automatically download the requried python, could you check it?

mcp-server/build.gradle.kts
```diff
@@ -0,0 +1,277 @@
...
+
+      val blackExitCode = exec {
+        workingDir = pythonProjectDir
+        commandLine(venvPython, "-m", "black", "--check", "mcp_server", "tests")
+        isIgnoreExitValue = false
+      }.exitValue
+
+      if (blackExitCode != 0) {
+        throw GradleException("Python black formatting check failed")
+        anyFailed = true
```
yuqi1129: `anyFailed = true` anyFailed is never used.

yuqi1129:
> could

It's fine on my machine.

mcp-server/mcp_server/connector/rest/rest_client_schema_operation.py
```diff
@@ -0,0 +1,33 @@
...
+class RESTClientSchemaOperation(SchemaOperation):
+    def __init__(self, metalake_name: str, rest_client: AsyncClient):
+        self.metalake_name = metalake_name
+        self.rest_client = rest_client
+
+    async def get_list_of_schemas(self, catalog_name: str) -> str:
+        response = await self.rest_client.get(
+            f"/api/metalakes/{self.metalake_name}/catalogs/{catalog_name}/schemas"
+        )
+        return extract_content_from_response(response, "identifiers", [])
```
yuqi1129: For MCP server,  is one field `identifiers ` enough? Why don't you return all the information here? do you have any further consideration?

FANNG1: The response only has `identifiers` and `code` fields

mcp-server/build.gradle.kts
FANNG1: removed

docs/gravitino-mcp-server.md
```diff
@@ -0,0 +1,74 @@
...
+}
+```
+
+Or start a HTTP MCP server by `uv run mcp_server --metalake test --uri http://127.0.0.1:8090 --transport http --mcp-url http://localhost:8000/mcp`, and use the configuration:
+
+```json
+{
+  "mcpServers": {
+    "gravitino": {
+      "url": "http://localhost:1234/mcp1"
```
shaofengshi: it it better to make this url be consistent with the above one "http://localhost:8000/mcp"?

mcp-server/README.md
```diff
@@ -0,0 +1,64 @@
...
+}
+```
+
+Or start HTTP MCP server by `uv run mcp_server --metalake test --uri http://127.0.0.1:8090 --transport http --mcp-url http://localhost:8000/mcp`, and use the configuration:
+
+```json
+{
+  "mcpServers": {
+    "gravitino": {
+      "url": "http://localhost:1234/mcp1"
```
shaofengshi: the same comment as the above one. besides, keep two documents seems be odd, when we make a change, need to update in two places.

docs/gravitino-mcp-server.md
jerryshao: How to configure the port of mcp server?

mcp-server/README.md
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+}
+```
```
jerryshao: We can remove this readme doc, and use the one above.

docs/gravitino-mcp-server.md
shaofengshi: it is configured with `--mcp-url` parameter

mcp-server/build.gradle.kts
```diff
@@ -0,0 +1,274 @@
...
+        commandLine(uvExecutable.absolutePath, "--version")
+        isIgnoreExitValue = true
+      }
+      if (uvCheck.exitValue != 0) {
+        throw GradleException("UV installation failed")
+      }
+
+      project.extensions.extraProperties.set("uvExecutable", uvExecutable.absolutePath)
+    }
+  }
```
jerryshao, line 110: What if I already installed a UV, will this install another one for me? Also, how about let user to install UV beforehand. My concern is that a): Gradle may not have the permission to install; b) User may want a different version. 

I don't have a strong preference on it, just some thoughts about it.

mcp-server/mcp_server/connector/factory.py
```diff
@@ -0,0 +1,52 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from mcp_server.connector.rest.rest_client_operation import RESTClientOperation
+
+
+class ConnectorFactory:
```
jerryshao: The concept of `connector` is a bit strange, do we have a better name?

mcp-server/build.gradle.kts
yuqi1129: > My concern is that a): Gradle may not have the permission to install;

This should not be a problem, the script will install `uv` with path `{mcp-server}/.uv` and the  permission should be ok

 b) User may want a different version.

Accordingly to the scripts, it seems that mcp-server will always use the latest version in the first compiling process.
I also have no preference about it.

mcp-server/mcp_server/connector/factory.py
```diff
@@ -18,35 +18,35 @@
...
 
-class ConnectorFactory:
+class RESTClientFactory:
     """
-    Factory for creating and configuring Gravitino connector instances.
+    Factory for creating and configuring Gravitino rest client instances.
     """
 
-    _connector_class = RESTClientOperation
+    _rest_client_class = RESTClientOperation
```
jerryshao: Should you also change the package name?

mcp-server/mcp_server/connector/rest/rest_client_operation.py
```diff
@@ -0,0 +1,55 @@
...
+)
+from mcp_server.connector.rest.rest_client_schema_operation import (
+    RESTClientSchemaOperation,
+)
+from mcp_server.connector.rest.rest_client_table_operation import (
+    RESTClientTableOperation,
+)
+
+
+class RESTClientOperation(GravitinoOperation):
```
jerryshao: This package uses plain rest protocol, can we rename this package to "plain", and all the class names to `PlainREST...`, `plain_rest_client...`

yuqi1129: Classes like `RESTClientSchemaOperation`  are in the package `rest`, I'm going to rename it to `plain` and then rename package `connector` to `clients` , the current package structure is as follows:
image


Would that be acceptable to you?

jerryshao: OK

mcp-server/mcp_server/client/plain/rest_client_catalog_operation.py
```diff
@@ -0,0 +1,33 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+from httpx import AsyncClient
+
+from mcp_server.client import CatalogOperation
+from mcp_server.client.plain.utils import extract_content_from_response
+
+
+class PlainRESTClientCatalogOperation(CatalogOperation):
```
jerryshao: Can you also rename the file to align with the class name here and below.

jerryshao (approved):

jerryshao:
@shaofengshi do you want to take another look?

shaofengshi:
> @shaofengshi do you want to take another look?

LGTM, thanks!

shaofengshi (approved):

Issue #7906, opened 2025-08-04
[Improvement] ​​Increase the priority of configuration which is defined by Gravitino in Trino


hqbhoho:
### What would you like to be improved?

`trino.bypass` properties will be skipped when the catalog properties is defined by Gravitino

PR #7907, opened 2025-08-04
[#7906] improvement(trino-connector): Increase the priority of configuration which is defined by Gravitino

hqbhoho:
`trino.bypass` properties will be skipped when the catalog properties is defined by Gravitino

Fix: #7906 

No

local tests

### DIFF
--- a/docs/trino-connector/supported-catalog.md
+++ b/docs/trino-connector/supported-catalog.md
@@ -137,6 +137,7 @@ call gravitino.system.create_catalog(
 ```
 
 A prefix with `trino.bypass.` in the configuration key is used to indicate Gravitino Trino connector to pass the Trino connector configuration to the Gravitino catalog in the Trino runtime.
+Note that it is when a configuration property is explicitly defined in Gravitino that the `trino.bypass.`properties which for the same property will be skipped.
 
 More Trino connector configurations can refer to:
 - [Hive catalog](https://trino.io/docs/current/connector/hive.html#hive-general-configuration-properties)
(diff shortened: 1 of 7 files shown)


docs/trino-connector/supported-catalog.md
```diff
@@ -137,6 +137,7 @@ call gravitino.system.create_catalog(
 ```
 
 A prefix with `trino.bypass.` in the configuration key is used to indicate Gravitino Trino connector to pass the Trino connector configuration to the Gravitino catalog in the Trino runtime.
+Note that it is when a configuration property is explicitly defined in Gravitino that the `trino.bypass.`properties which for the same property will be skipped.
```
yuqi1129, line 140: @diqiu50 
Please check whether this content is clear enough.

yuqi1129:
The code parts are LGTM, @diqiu50, please help to review it.

docs/trino-connector/supported-catalog.md
hqbhoho: For example, the Trino MySQL connector properties `connection-url`, `connection-user` and `connection-password` ​​inherit values directly​​ from the `jdbc-url`, `jdbc-user` and `jdbc-password` defined in the Gravitino MySQL catalog. As a result, these configurations ​​cannot be overridden​​ by any `trino.bypass.*` properties.

Maybe we can add an example or we list the configuration which should not defined by `trino.bypass.*` properties?

yuqi1129: What I want to convey is that the sentence is quite complex and hard to understand. Can you polish and make it more readable?  By the way, an example is always appreciated.

hqbhoho: Note​​ that if Trino connector properties ​​directly inherit values​​ from Gravitino catalog, these configurations ​​cannot be overridden through​​ any `trino.bypass.*` properties. For example, the Trino MySQL connector properties `connection-url`, `connection-user` and `connection-password` ​​directly inherit​​ the `jdbc-url`, `jdbc-user` and `jdbc-password` values defined in the Gravitino MySQL catalog. Therefore, defining `trino.bypass.connection-url`, `trino.bypass.connection-user` or `trino.bypass.connection-password` ​​is invalid and must be avoided​​.

How about changing it like this?

yuqi1129: It generally looks good to me.  `is invalid and must be avoided` -> `will be ommitted and does not take effect`.​​

PR #7913, opened 2025-08-04
[#7823] feat(server): upgrade Gravitino server mini JDK version to 17

diqiu50:
1. Upgrade Gravitino server server minimal JDK version to 17
2. Remove other version of JDK in CI pipe lines.

Fix: #7823 

UTs and ITs


FANNG1 (approved):

FANNG1:
This PR is to upgrade JDK versions. #7877 will add a release option to build server modules for JDK17 and client modules for JDK8

FANNG1:
@diqiu50 could you fix the CI failures?

Issue #6719, opened 2025-03-20
[FEATURE] Update Iceberg version to latest


metalshanked:
### Describe the feature

Since the iceberg versions have a lot of breaking changes (eg:- Oauth token endpoints deprecated etc)
 many OSS projects are updating the core iceberg frequently.

Would be great to have gravitino also map to the newer iceberg versions

jerryshao:
BTW, is there any particular feature you're pursuing for the newer version?

metalshanked:
There were some issues related to view queries from Trino. Being tracked under a separate issue. thx

Issue #7413, opened 2025-06-17
Support Iceberg v3 spec


metalshanked:
Will Gravitino work with the newly released iceberg v3 spec. Are there REST catalog implications?

Thanks!

FANNG1:
I'm working on upgrade Iceberg to high version which supports V3 spec recently, the draft PR is https://github.com/apache/gravitino/pull/7418

dyrnq:
stare

Issue #6337, opened 2025-01-20
[Subtask] upgrade Iceberg version  to 1.7 in Iceberg REST server version


FANNG1:
### Describe the subtask

upgrade Iceberg REST server version from 1.5 to 1.7 to support more endpoints

### Parent issue

#6336

FANNG1:
Normally we only need to change Iceberg version in `gradle/libs.versions.toml` from `1.5.2` to `1.7.2`,  but Iceberg drop java8 support in `1.7`, I'm not sure whether the Iceberg jar could works on Gravitino in Java 1.8 enviroment.

FANNG1:
after updating the iceberg version to `1.7.1`, there are some compile errors

```
      > No matching variant of org.apache.iceberg:iceberg-gcp:1.7.1 was found. The consumer was configured to find a library for use during compile-time, compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally but:
          - Variant 'apiElements' capability org.apache.iceberg:iceberg-gcp:1.7.1 declares a library for use during compile-time, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component, compatible with Java 11 and the consumer needed a component, compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
          - Variant 'runtimeElements' capability org.apache.iceberg:iceberg-gcp:1.7.1 declares a library for use during runtime, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component, compatible with Java 11 and the consumer needed a component, compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
```

zzzk1:
@FANNG1 Are you working on this? I would like to help. 😊

FANNG1:
> [@FANNG1](https://github.com/FANNG1) Are you working on this? I would like to help. 😊

thanks, but I'm not sure is there any way to make Iceberg 1.7 work on Java8, since Gravitino still supports Java8

metalshanked:
Agree, since future updates to iceberg is important, no option but to drop Java 8 support i think

FANNG1:
[...]

PR #7917, opened 2025-08-04
[#6719][#6337][#7413]  feat(iceberg): upgrade Iceberg version to 1.9

FANNG1:

* Upgrade Iceberg version from 1.6 to 1.9
* Add support for Endpoints in the config response for better compatibility with Trino.

Fix: https://github.com/apache/gravitino/issues/6719 https://github.com/apache/gravitino/issues/6337 https://github.com/apache/gravitino/issues/7413

No

compiled and check Iceberg version, test whether it works.


FANNG1:
@jerryshao @jerqi PTAL, thx

gradle/libs.versions.toml
```diff
@@ -55,8 +55,8 @@ commons-collections3 = "3.2.2"
 commons-configuration1 = "1.6"
 commons-dbcp2 = "2.11.0"
 caffeine = "2.9.3"
-iceberg = '1.6.1' # used for Gravitino Iceberg catalog and Iceberg REST service
-iceberg4spark = "1.4.1" # used for compile spark connector
+iceberg = '1.9.2' # used for Gravitino Iceberg catalog and Iceberg REST service
+iceberg4connector = "1.6.1" # used for compile spark connector
```
jerryshao: It is not only used for Spark, right? I think you should change a word.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestIcebergCatalogWrapperManagerForREST.java
```diff
@@ -55,7 +55,7 @@ public void testValidGetOps(String rawPrefix) {
   }
 
   @ParameterizedTest
-  @ValueSource(strings = {"hello", "\\\n\t\\\'", "\u0024", "\100", "[_~", "default_catalog/"})
+  @ValueSource(strings = {"hello", "\\\n\t\\\'", "\u0024", "\100", "[_~"})
```
jerryshao, line 58: Why do we do such change?

FANNG1: `default_catalog` is the default catalog name if the user doesn't specify the catalog name in `prefix`.  After this PR , we add `ENDPOINTS` to the config interface which represent the capacity of the underlying catalog backend.  if user doesn't specifing the catalog name, we will response the catalog capacibilities from `default_catalog` and set `prefix` to the catalog name of `default_catalog`.  so `default_catalog` is a valid catalog identifier.

gradle/libs.versions.toml
FANNG1: Yes,  updated

Issue #7840, opened 2025-07-30
[Subtask] Add  a background thread to poll the status periodically


jerryshao:
### Describe the subtask

This issue tracks the work of polling the status automatically.

### Parent issue

#7154

PR #7936, opened 2025-08-05
[#7840] feat(core): Add job status puller and orphan file cleaner

jerryshao:

Add the job status pulling thread and orphan file cleaner thread for job manager.

Fix: #7840 

No.

Add UTs.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/JobManager.java
+++ b/core/src/main/java/org/apache/gravitino/job/JobManager.java
@@ -476,7 +476,8 @@ public class JobManager implements JobOperationDispatcher {
     cleanUpExecutor.shutdownNow();
   }
 
-  public void pullAndUpdateJobStatus() {
+  @VisibleForTesting
+  void pullAndUpdateJobStatus() {
     List<String> metalakes = listInUseMetalakes(entityStore);
     for (String metalake : metalakes) {
       // This unnecessary list all the jobs, we need to improve the code to only list the active
@@ -531,7 +532,8 @@ public class JobManager implements JobOperationDispatcher {
     }
   }
 
-  public void cleanUpStagingDirs() {
+  @VisibleForTesting
+  void cleanUpStagingDirs() {
     List<String> metalakes = listInUseMetalakes(entityStore);
 
     for (String metalake : metalakes) {


core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -103,6 +127,32 @@ public JobManager(Config config, EntityStore entityStore, IdGenerator idGenerato
...
+
+    this.jobStagingDirKeepTimeInMs = config.get(Configs.JOB_STAGING_DIR_KEEP_TIME_IN_MS);
+    this.cleanUpExecutor =
+        Executors.newSingleThreadScheduledExecutor(
+            runnable -> {
+              Thread thread = new Thread(runnable, "job-staging-dir-cleanup");
+              thread.setDaemon(true);
+              return thread;
+            });
+    long scheduleInterval = jobStagingDirKeepTimeInMs / 10;
```
yuqi1129, line 152: You'd better add a check that `jobStagingDirKeepTimeInMs` is larger than 10;

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -470,7 +620,7 @@ static String replacePlaceholder(String inputString, Map<String, String> replace
       return inputString; // Return as is if the input string is blank
     }
 
-    StringBuffer result = new StringBuffer();
+    StringBuilder result = new StringBuilder();
```
yuqi1129, line 653: Is there any thread-race problem for `StringBuffer` here?

jerryshao: I don't know. This is recommended by intelij idea, so I change to use `StringBuilder`.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: OK, I will, normally this should not be happened.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -388,7 +472,102 @@ public JobEntity cancelJob(String metalake, String jobId) throws NoSuchJobExcept
...
+      List<JobEntity> finishedJobs =
+          listJobs(metalake, Optional.empty()).stream()
+              .filter(
+                  job ->
+                      job.status() == JobHandle.Status.CANCELLED
+                          || job.status() == JobHandle.Status.SUCCEEDED
+                          || job.status() == JobHandle.Status.FAILED)
+              .filter(
+                  job ->
+                      job.finishedAt() > 0
```
yuqi1129, line 549: When will a job status is canceled, succeeded or failed and the finished at is `0`?

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -388,7 +472,102 @@ public JobEntity cancelJob(String metalake, String jobId) throws NoSuchJobExcept
   @Override
   public void close() throws IOException {
     jobExecutor.close();
-    // TODO(jerry). Implement any necessary cleanup logic for the JobManager.
+    statusPullExecutor.shutdownNow();
+    cleanUpExecutor.shutdownNow();
+  }
+
+  public void pullAndUpdateJobStatus() {
```
yuqi1129: Can be set as a package access level.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: No, the finished time of the completed job is the current timestamp.

Issue #7758, opened 2025-07-18
[Subtask] Add the server-side REST interface for the job system


jerryshao:
### Describe the subtask

This issue tracks the work of adding the server-side REST interface for the job system.

### Parent issue

#7154

PR #7939, opened 2025-08-05
[#7758] feat(server): Add the REST interface for job system - part2

jerryshao:

Add the job related rest interface for Gravitino.

Fix: #7758 

No.

Added UTs.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java
@@ -279,7 +279,7 @@ public class JobOperations {
     }
   }
 
-  @DELETE
+  @POST
   @Path("runs/{jobId}")
   @Produces("application/vnd.gravitino.v1+json")
   @Timed(name = "cancel-job." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java
```diff
@@ -191,7 +199,111 @@ public Response deleteJobTemplate(
...
+            LOG.info("Run job {} in metalake: {}", jobEntity.name(), metalake);
+            return Utils.ok(new JobResponse(toDTO(jobEntity)));
+          });
+
+    } catch (Exception e) {
+      return ExceptionHandlers.handleJobException(OperationType.RUN, "", metalake, e);
+    }
+  }
+
+  @DELETE
```
yuqi1129: It seems that post is more proper here as `cancelJob` only changes the status and will not do a real removal.

jerryshao: I asked the chatgpt, and it recommends me to use `@DELETE`. I don't have a strong preference on it.

jerryshao: Think about it again, I will change to use POST.

Issue #7745, opened 2025-07-17
[Bug report] Create role for `file_catalog2.file_schema1` failed


danhuawang:
### Version

main branch

### Describe what's wrong

<img width="1172" height="855" alt="Image" src="https://github.com/user-attachments/assets/120bd0be-a1f8-4320-9ee8-7ff8601ecead" />

### Error message and/or stacktrace

POST http://a4295d83b4f084bc7905efef7298e117-45919047.ap-northeast-1.elb.amazonaws.com:8090/api/metalakes/test/roles/
Response
URL: http://a4295d83b4f084bc7905efef7298e117-45919047.ap-northeast-1.elb.amazonaws.com:8090/api/metalakes/test/roles/

Status code: 500

Content Type: application/json

Request Headers
Accept=application/vnd.gravitino.v1+json
Authorization=Basic TGlzYTpEQHRhJHRyYXQw
Content-Type=application/json
Content Body
{
    "name": "RoleSchemaEditor",
    "properties": {
        "k1": "v1"
    },
    "securableObjects": [
        {
            "type": "SCHEMA",
            "privileges": [
                {
                    "name": "USE_SCHEMA",
                    "condition": "ALLOW"
                },
                {
                    "name": "CREATE_FILESET",
                    "condition": "ALLOW"
                },
                {
                    "name": "READ_FILESET",
                    "condition": "ALLOW"
                },
                {
                    "name": "WRITE_FILESET",
                    "condition": "ALLOW"
                }
            ],
            "fullName": "file_catalog2.file_schema1"
        }
    ]
}
Response Headers
Date: Thu, 17 Jul 2025 11:04:46 GMT
Content-Type: application/json
Transfer-Encoding: chunked
Server: Jetty(9.4.51.v20230217)
Response Body
{
    "code": 1002,
    "type": "RuntimeException",
    "message": "Failed to operate object [RoleSchemaEditor] operation [CREATE] under [test], reason [Failed to get current user]",
    "stack": [
        "java.lang.RuntimeException: Failed to get current user",
        "\tat org.apache.gravitino.catalog.fileset.authentication.UserContext.getUserContext(UserContext.java:115)",
        "\tat org.apache.gravitino.catalog.fileset.SecureFilesetCatalogOperations.initialize(SecureFilesetCatalogOperations.java:105)",
        "\tat org.apache.gravitino.connector.BaseCatalog.ops(BaseCatalog.java:178)",
[...]

PR #7940, opened 2025-08-05
[#7745] fix(fs-bundles): avoid shade jdk class

mchades:

remove shade "com.sun" to avoid shade jdk class in bundle jar

Fix: #7745 

no

test locally


bundles/aliyun-bundle/build.gradle.kts
```diff
@@ -46,7 +46,10 @@ tasks.withType(ShadowJar::class.java) {
   // Relocate dependencies to avoid conflicts
   relocate("com.fasterxml.jackson", "org.apache.gravitino.aliyun.shaded.com.fasterxml.jackson")
   relocate("com.google", "org.apache.gravitino.aliyun.shaded.com.google")
-  relocate("com.sun", "org.apache.gravitino.aliyun.shaded.com.sun")
+  relocate("com.sun.activation", "org.apache.gravitino.aliyun.shaded.com.sun.activation")
+  relocate("com.sun.istack", "org.apache.gravitino.aliyun.shaded.com.sun.istack")
+  relocate("com.sun.jersey", "org.apache.gravitino.aliyun.shaded.com.sun.jersey")
+  relocate("com.sun.xml", "org.apache.gravitino.aliyun.shaded.com.sun.xml")
```
jerryshao, line 52: So do you want to shade these jars, they're also mostly provided by jdk, right?

mchades: they are not provided by jdk

jerryshao: Which jars included these classes, AFAICK, they're provided by Sun as unsafe APIs.

mchades: they are included by `hadoop-client-runtime-3.3.1.jar`.

We should note that these are third-party libraries that use the `com.sun.*` namespace but are not part of the JDK.

yuqi1129: I could not find any clear issues here. If you have tested it locally and with spark, we can merge it first.

jerryshao (approved):

Issue #7759, opened 2025-07-18
[Subtask] Add Java client for the job system


jerryshao:
### Describe the subtask

This issue tracks the work of adding the Java client for the job system.

### Parent issue

#7154

PR #7944, opened 2025-08-06
[#7759]feat(client-java): Add Java client for Job System

jerryshao:

This PR adds the java client support for job system.

Fix: #7759 

NO.

Add UTs to cover the code.


jerryshao:
@mchades @yuqi1129 please help me to review, thanks.

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoMetalake.java
```diff
@@ -1189,6 +1204,162 @@ public String[] listBindingRoleNames() {
...
+        .map(org.apache.gravitino.dto.util.DTOConverters::fromDTO)
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public void registerJobTemplate(JobTemplate jobTemplate)
+      throws JobTemplateAlreadyExistsException {
+    JobTemplateRegisterRequest req =
+        new JobTemplateRegisterRequest(DTOConverters.toJobTemplateDTO(jobTemplate));
+    req.validate();
```
mchades: Should the request be validated by the server instead of the client?

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoMetalake.java
```diff
@@ -1189,6 +1204,162 @@ public String[] listBindingRoleNames() {
...
+        restClient.get(
+            String.format(API_METALAKES_JOB_TEMPLATES_PATH, RESTUtils.encodeString(this.name()))
+                + "/"
+                + RESTUtils.encodeString(jobTemplateName),
+            JobTemplateResponse.class,
+            Collections.emptyMap(),
+            ErrorHandlers.jobErrorHandler());
+    resp.validate();
+
+    return org.apache.gravitino.dto.util.DTOConverters.fromDTO(resp.getJobTemplate());
```
mchades, line 1254: Redundant package name?

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoMetalake.java
jerryshao: This will also be validated in the server side. To vaildate in the client side is to avoid unncessary remote call.

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoMetalake.java
jerryshao: There's also another client DTOConverters in the same package, so adding a full name to avoid the conflicts.

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoMetalake.java
mchades: There is a case: after users upgrade the server to change certain request limitations, if we validate the request on the client side, they will also need to upgrade the client.

jerryshao: I see, I will remove it.

jerryshao: Actually, that is an incompatible change, we should avoid it. Anyway, I will remove it.

mchades (approved):

Issue #7945, opened 2025-08-06
refactor: Remove unused ProviderType enum and clarify OAuth provider config supports 'default' and 'oidc'


bharos:
### Describe the subtask

We don't need providerType as we now use OIDC so we won't have to mention specific OAuth provider like Azure/Google etc.. as OIDC provides common interface for all the oauth providers

### Parent issue

[7538](https://github.com/apache/gravitino/issues/7538)

bharos:
@jerqi Since we are using OIDC for web UI login, we no longer need "ProviderType" because OIDC can support all standard providers, so the good thing is we won't need any provider specific implementation for Azure/Google/OAuth0 etc..

So the config just needs to say provider = default (to use the existing kind of login workflow) or oidc(to use the oidc login workflow)

PR #7946, opened 2025-08-06
[#7945] refactor: Validate Provider type and clarify OAuth provider config supports 'default' and 'oidc'

bharos:

Move ProviderType from common to server-common, and provide the supported values (default, oidc). Use it to validate the OAuthConfig provider

Minor refactor and validate the config

Fix: #7945

NA
Added unit tests

Tested manually as well:
Screenshot 2025-08-06 at 2 54 51 PM

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
@@ -77,10 +77,22 @@ public interface OAuthConfig {
   // OAuth provider configs
   ConfigEntry<String> PROVIDER =
       new ConfigBuilder(OAUTH_CONFIG_PREFIX + "provider")
-          .doc("The OAuth provider to use (e.g., azure)")
+          .doc(
+              "The OAuth provider to use. This will be used in the Gravitino Web UI to determine the authentication flow.")
           .version(ConfigConstants.VERSION_1_0_0)
           .stringConf()
-          .create();
+          .checkValue(
+              value -> {
+                if (value == null) return false;
+                for (ProviderType type : ProviderType.values()) {
+                  if (type.name().equalsIgnoreCase(value)) {
+                    return true;
+                  }
+                }
+                return false;
+              },
+              "Invalid OAuth provider type. Supported values: 'default', 'oidc'")
+          .createWithDefault(ProviderType.DEFAULT.name().toLowerCase());
 
   ConfigEntry<String> CLIENT_ID =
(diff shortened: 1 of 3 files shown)


bharos:
@jerqi can you please review

Since we are using OIDC for web UI login, we no longer need "ProviderType" because OIDC can support all standard providers, so the good thing is we won't need any provider specific implementation for Azure/Google/OAuth0 etc..

So the config just needs to say provider = default (to use the existing kind of login workflow) or oidc(to use the oidc login workflow)

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -77,10 +77,10 @@ public interface OAuthConfig {
   // OAuth provider configs
   ConfigEntry<String> PROVIDER =
       new ConfigBuilder(OAUTH_CONFIG_PREFIX + "provider")
-          .doc("The OAuth provider to use (e.g., azure)")
+          .doc("The OAuth provider to use. Supported values: 'default', 'oidc'")
```
roryqi: You would better to say that the config option will be used only for UI.

common/src/main/java/org/apache/gravitino/auth/ProviderType.java
```diff
@@ -1,28 +0,0 @@
...
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-
-package org.apache.gravitino.auth;
-
-/** The type of OAuth provider. */
-public enum ProviderType {
```
roryqi, line 23: Maybe we don't need to remove enum. We can use enum to validate the config option to avoid using a wrong parameter.

roryqi: Maybe we don't need this to common module. server-common may be enough.

bharos: Done

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
bharos: Done

bharos:
@jerqi PTAL
Added some unit tests as well

roryqi:
Thanks @bharos , merged.

Issue #7949, opened 2025-08-06
[FEATURE] Support tag operation for MCP Server


yuqi1129:
### Describe the feature

Provide support for the tag-related API for the MCP server.

PR #7963, opened 2025-08-07
[#7949] feat(mcp-server): Support tag operations for MCP server

yuqi1129:

Support tag operations for MCP server

To support more feature.

Fix: #7949

N/A

image

image

image

image

### DIFF
--- a/mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
+++ b/mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
@@ -86,7 +86,7 @@ class PlainRESTClientTagOperation(TagOperation):
         self, metadata_full_name: str, metadata_type: str
     ) -> str:
         response = await self.rest_client.get(
-            f"/api/metalakes/{self.metalake_name}/objects/{metadata_type}/{metadata_full_name}/tags"
+            f"/api/metalakes/{self.metalake_name}/objects/{metadata_type}/{metadata_full_name}/tags?details=true"
         )
         return extract_content_from_response(response, "names", [])
 
--- a/mcp-server/mcp_server/tools/tag.py
+++ b/mcp-server/mcp_server/tools/tag.py
@@ -305,8 +305,16 @@ def load_tag_tool(mcp: FastMCP):
 
         Example Return Value:
             [
-              "tag1",
-              "tag2"
+              {
+                "name": "import",
+                "comment": "Import tag for data governance",
+                "properties": {},
+                "audit": {
+                  "creator": "anonymous",
+                  "createTime": "2025-08-12T11:41:09.257235Z"
+                },
+                "inherited": false
+              }
             ]
         """
         client = ctx.request_context.lifespan_context.rest_client()


yuqi1129:
This PR depends on #7905 and is not ready to review until #7905 is merged.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
```diff
@@ -0,0 +1,97 @@
...
+    async def alter_tag(self, tag_name: str, updates: list) -> str:
+        response = await self.rest_client.put(
+            f"/api/metalakes/{self.metalake_name}/tags/{tag_name}",
+            json={
+                "updates": updates,
+            },
+        )
+        return extract_content_from_response(response, "tag", {})
+
+    async def delete_tag(self, name: str) -> None:
```
shaofengshi, line 61: Delete operation is dangerous, and not idempotent; In my opinion, we don't need to expose such operation as MCP tool at this moment.

yuqi1129: OK, I will raise another PR to temporarily disable it.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
```diff
@@ -0,0 +1,97 @@
...
+        response = await self.rest_client.delete(
+            f"/api/metalakes/{self.metalake_name}/tags/{name}"
+        )
+        if response.status_code != 200:
+            raise Exception(f"Failed to delete tag {name}: {response.text}")
+        return None
+
+    async def association_tag_with_metadata(
```
shaofengshi: "association" is a noun, not a verb, please use "associate".

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
```diff
@@ -0,0 +1,97 @@
...
+        self,
+        metadata_full_name: str,
+        metadata_type: str,
+        tags_to_associate: list,
+        tags_to_disassociate: list,
```
shaofengshi, line 74: This is confusing that have disassociate here. I would suggest to use another method/tool for disassociate. Or you'd better give another name.

yuqi1129: This is an API provided by Gravitino. If it's not well designed, I can split it into two.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
yuqi1129: sure

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
jerryshao, line 61: We don't have to support create, alter and delete for now.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
jerryshao: Gravitino is designed to be leveraged by UI easily. But when used in LLM, you should keep it simple and self-explanatory.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
yuqi1129: These APIs have been disabled and can be enabled if needed.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
yuqi1129: I see.

yuqi1129: I've divided it into two.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
yuqi1129: @jerryshao 
API like `associate/disassociate` tags with metadata can also make changes to Gravitino, do we need to disable them at the same time?

jerryshao: No, that is mainly user people and UI, not for LLM.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
jerryshao: Where is the code?

yuqi1129:
```python
    @mcp.tool(tags={"tag"}, enabled=False)
    async def disassociate_tag_from_metadata(
        ctx: Context,
        metadata_full_name: str,
        metadata_type: str,
        tags_to_disassociate: list,
    ) -> str:

    @mcp.tool(tags={"tag"}, enabled=False)
    async def associate_tag_with_metadata(
        ctx: Context,
        metadata_full_name: str,
        metadata_type: str,
        tags_to_associate: list,
    ) -> str:
```

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
yuqi1129: I have disabled tag association and disassociation operations.

mcp-server/mcp_server/tools/tag.py
```diff
@@ -0,0 +1,353 @@
...
+        Args:
+            ctx (Context): The request context object containing lifespan context
+                           and connector information.
+            metadata_full_name (str): Full name of the metadata object to associate tags with.
+            It's typically in the format "catalog.schema.table" or "catalog.schema" or "catalog".
+            or "catalog.schema.fileset". The "model", "topic" are also supported and the format
+            is the same as for "catalog.schema.table".
+            metadata_type (str): Type of the metadata object (e.g., "table", "schema", "catalog",
+             "fileset", "model", "topic").
+            tags_to_associate (list): List of tag names to associate with the metadata.
```
jerryshao, line 219: Here it is the same question as previously proposed by @FANNG1 . Using metadata object name "xx.xx.xx" may not be straightforward for LLM. @FANNG1 uses different parameters for catalog / schema / table when loading a table. I was thinking that for tag related operations, can we also make it like before, which will make LLM easy to understand.

mcp-server/mcp_server/tools/tag.py
```diff
@@ -0,0 +1,353 @@
...
+        ctx: Context, metadata_full_name: str, metadata_type: str
+    ) -> str:
+        """
+        List all tags associated with a specific metadata item.
+
+        Args:
+            ctx (Context): The request context object containing lifespan context
+                           and connector information.
+            metadata_full_name (str): Full name of the metadata item (e.g., table, column).
+            metadata_type (str): Type of the metadata (e.g., "table", "column").
```
jerryshao: With this simple description, I don't think LLM can easily figure out what is the full name and what is the supported type.

jerryshao: You should list all the supported metadata object type, and also the examples.

yuqi1129: I plan to add the following tool to inform LLM about the details of metadata type and full name, thus LLM can know the type of metadata it supports. Would that be acceptable to you? 

```python
    @mcp.tool(tags={"metadata"})
    async def list_all_metadata_types(
        ctx: Context
    ) -> list:
        """
        List all metadata types available in the MCP server.

        Args:
            ctx: Context object containing request metadata.

        Returns:
            str: JSON formatted string containing list of metadata types.
        """
        return [
            "catalog",
            "schema",
            "table",
            "column",
            "model",
            "topic",
            "fileset"
        ]

    @mcp.tool(tags={"metadata"})
    async def get_metadata_fullname_formats(
        ctx: Context) -> dict:
        """
        Get the formats for metadata full names.
        Args:
            ctx: Context object containing request metadata.
        Returns:
            dict: Dictionary containing formats for different metadata types. Key
            is the metadata type and value is the format string.
        """
        return {
            "catalog": "{catalog}",
            "schema": "{catalog}.{schema}",
            "table": "{catalog}.{schema}.{table}",
            "column": "{catalog}.{schema}.{table}.{column}",
            "model": "{catalog}.{model}",
            "topic": "{catalog}.{topic}",
            "fileset": "{catalog}.{fileset}"
        }

```

jerryshao: Did you try if it works by LLM?

yuqi1129: Yes, I have tried it

image
image
image
image
image

jerryshao: Maybe you can merge this two APIs into one.

yuqi1129: I see.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        return extract_content_from_response(response, "names", [])
+
+    async def list_tags_for_metadata(
+        self, metadata_full_name: str, metadata_type: str
+    ) -> str:
+        response = await self.rest_client.get(
+            f"/api/metalakes/{self.metalake_name}/objects/{metadata_type}/{metadata_full_name}/tags"
+        )
+        return extract_content_from_response(response, "names", [])
```
jerryshao: If the list operations can support listing the details, we'd better always listing the details, so we can give more information to LLM.

jerryshao: Not just for this PR, but also for others.

mcp-server/mcp_server/tools/metadata.py
```diff
@@ -0,0 +1,30 @@
...
+from fastmcp import Context, FastMCP
+
+
+def load_metadata_tool(mcp: FastMCP):
+    @mcp.tool(tags={"metadata"})
+    async def metadata_type_to_fullname_formats(
+        ctx: Context) -> dict:
+
+        """
+        Get metadata type to full name formats.
```
jerryshao, line 25: Please be aware that if you're using metadata (rather than metadata object), then all the places should be unified, otherwise it will be confused to LLM. The terminology should be consistent.

yuqi1129: I see.

mcp-server/mcp_server/client/plain/plain_rest_client_tag_operation.py
yuqi1129: Let me try to resolve it.

yuqi1129: Fix all list operations in this PR.

jerryshao (approved):

jerryshao:
I'm going to close this PR. You'd better fix the python style issue in the following issue.

Issue #7958, opened 2025-08-07
[Improvement] Create a Names from just white space shodul not be allowed


justinmclean:
### What would you like to be improved?

Update this test to teh following and the test will fail.
```
  @Test
  public void testFromStringInvalidArgs() {
    Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString(null));
    Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString(".a"));
    Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString("a."));
    Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString("a..b"));
    Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString(" "));
  }
```

### How should we improve?

Update Namespace.fromString to throw a IllegalArgumentException if given a namespace that is just whitespace

ibrahim307le:
can you assign this issue to me, i can able to fix this with my knowledge and learnings. kindly assign this for my contribution to your project.

ibrahim307le:
Hi @justinmclean ,

I’ve reviewed this issue and would love to contribute a fix. Would it be possible to assign this to me?

Thanks in advance!

yossev:
Hello! 

a PR with the Fix will be submitted in a few minutes.

Thank you!

yossev:
This PR is ready for review.
Please let me know if any further changes or clarifications are needed.


Edit:  I might've missed adding one of the test cases. if thats fine, i can fix this further.

justinmclean:
Hi @ibrahim307le, thanks for offering to help. @yossev, it would have been best to pick an issue that someone else hadn't asked to do, but thanks for the PR.

yossev:
> Hi [@ibrahim307le](https://github.com/ibrahim307le), thanks for offering to help. [@yossev](https://github.com/yossev), it would have been best to pick an issue that someone else hadn't asked to do, but thanks for the PR.

Thanks for the feedback! I noticed the issue wasn’t officially assigned, so I went ahead with a solution, but I totally understand the need to avoid overlap. I’ll be more mindful about checking for ongoing interest next time.

Please let me know if there’s anything you'd like me to adjust in the PR. Appreciate the chance to contribute!

justinmclean:
@yossev I think you've made the fix too wide and it will now not allow namespace names containing a space.

yossev:
[...]

PR #7965, opened 2025-08-07
[#7958] fix(namespace): reject blank or whitespace-only input in fromString()

yossev:

This PR updates the `Namespace.fromString()` method to reject inputs that are blank or consist only of whitespace (e.g., `" "`, `"\t"`, `"\n"`).  
A new precondition was added using `StringUtils.isBlank(...)`, and placed early in the validation chain to ensure the method fails fast on invalid input.

Previously, passing a blank or whitespace-only string would bypass structural validations and return a namespace unexpectedly (e.g., via `.split()` producing an empty array).  
This could lead to invalid namespace objects or hidden bugs in downstream components.

Yes.  
Calling `Namespace.fromString(" ")` or similar will now throw an `IllegalArgumentException` instead of silently creating an empty namespace.

- a test case in `testFromStringInvalidArgs` exists to verify the behavior:
  ```java
  Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString(" "));

### DIFF
--- a/api/src/main/java/org/apache/gravitino/Namespace.java
+++ b/api/src/main/java/org/apache/gravitino/Namespace.java
@@ -75,14 +75,13 @@ public class Namespace {
    */
   public static Namespace fromString(String namespace) {
     Preconditions.checkArgument(namespace != null, "Cannot create a namespace with null input");
+    // Reject blank or whitespace-only namespaces (e.g., " ", "\t", "\n")
+    Preconditions.checkArgument(!StringUtils.isBlank(namespace), "Cannot create a namespace with blank input");
     Preconditions.checkArgument(!namespace.endsWith("."), "Cannot create a namespace end with dot");
     Preconditions.checkArgument(
         !namespace.startsWith("."), "Cannot create a namespace starting with a dot");
     Preconditions.checkArgument(
         !namespace.contains(".."), "Cannot create a namespace with an empty level");
-    if (StringUtils.isBlank(namespace)) {
-      return empty();
-    }
     return Namespace.of(namespace.split("\\."));
   }
 


api/src/main/java/org/apache/gravitino/Namespace.java
```diff
@@ -75,14 +75,13 @@ public static Namespace of(String... levels) {
    */
   public static Namespace fromString(String namespace) {
     Preconditions.checkArgument(namespace != null, "Cannot create a namespace with null input");
+    // Reject blank or whitespace-only namespaces (e.g., " ", "\t", "\n")
```
justinmclean: There is no real need for a comment like this

justinmclean:
@yossev you need to add the test case as well.

yossev:
Hello, @justinmclean 

Working on adding the test case and fixing the comment issue.

Thank you!

yossev:
Hello, @justinmclean , Hope you're having a great day so far!!

Please check latest commit, and let me know if you have any suggestions.

Thank you.

yossev:
> Hi @yossev, please run ./gradlew spotlessApply for the formatting fixes needed

Hello, @georgereuben. Working on it

api/src/main/java/org/apache/gravitino/Namespace.java
```diff
@@ -75,14 +75,12 @@ public static Namespace of(String... levels) {
    */
   public static Namespace fromString(String namespace) {
     Preconditions.checkArgument(namespace != null, "Cannot create a namespace with null input");
+    Preconditions.checkArgument(!StringUtils.isBlank(namespace), "Cannot create a namespace with blank input");
```
georgereuben: I think the issue requires that strings with whitespaces should throw the `IllegalArgumentException` and that empty strings should return `Namespace.empty` as before.

Your changes would throw the IllegalArgumentException for empty strings as well. Please check with @justinmclean if this is what's needed?

justinmclean: Yes that's correct

justinmclean:
The CI is still failing, as you need to correctly format the run by running:
`./gradlew :api:spotlessApply`
and commit the changes.

yossev:
Hello , @georgereuben , @justinmclean 

As discussed before , i fixed the problem that @georgereuben pointed out.
strings with whitespaces should now throw the `IllegalArgumentException` and empty strings should return `Namespace.empty` as before.

The only challenge I’m facing now is that `./gradlew :api:spotlessApply` is POSIX-compliant, and I’m having trouble running it on my Windows PC.

Any suggestions?

justinmclean:
I'll see how this CI goes and if there is an issue, fix it for you.

api/src/main/java/org/apache/gravitino/Namespace.java
```diff
@@ -75,7 +75,8 @@ public static Namespace of(String... levels) {
    */
   public static Namespace fromString(String namespace) {
     Preconditions.checkArgument(namespace != null, "Cannot create a namespace with null input");
-    Preconditions.checkArgument(!namespace.endsWith("."), "Cannot create a namespace end with dot");
+    Preconditions.checkArgument(!namespace.contains(" "), "Cannot create a namespace containing whitespace");
```
georgereuben: I think this would invalidate legitimate namespaces which might have a space. a possible alternative could be to alter line 83 to be `if(namespace.isEmpty()) {return empty()}` and to add a precondition check using `StringUtils.isBlank()` after this check so that empty namespaces return `Namespace.empty` and only purely whitespace namespaces are caught by the precondition check.

@justinmclean , what would you suggest?

yossev: I have no idea how i messed up this bad and didnt see this through.
apologies, working on a fix

justinmclean:
The CI is failing as you need to run `./gradlew :api:spotlessApply' and push the changes to format the code correctly, but I did fix this for you and pushed the changes to your repo.

yossev:
> The CI is failing as you need to run `./gradlew :api:spotlessApply' and push the changes to format the code correctly, but I did fix this for you and pushed the changes to your repo.

The only challenge I’m facing now is that ./gradlew :api:spotlessApply is POSIX-compliant, and I’m having trouble running it on my Windows PC.

Is there a way i can run it on my Windows PC?

justinmclean:
It should be able to run on the Windows Subsystem for Linux (WSL).  See https://learn.microsoft.com/en-us/windows/wsl/install. I'm sorry, I don't have access to a Windows machine at the moment, so I can't test that. You can also look at the CI output, and it will show you the changes that are needed.

yossev:
> It should be able to run on the Windows Subsystem for Linux (WSL). See https://learn.microsoft.com/en-us/windows/wsl/install. I'm sorry, I don't have access to a Windows machine at the moment, so I can't test that. You can also look at the CI output, and it will show you the changes that are needed.

Hello, @justinmclean , i used WSL to run `./gradlew :api:spotlessApply` as we discussed.
Hopefully the CI doesn't fail this time.

api/src/main/java/org/apache/gravitino/Namespace.java
```diff
@@ -75,7 +75,10 @@ public static Namespace of(String... levels) {
    */
   public static Namespace fromString(String namespace) {
     Preconditions.checkArgument(namespace != null, "Cannot create a namespace with null input");
-    Preconditions.checkArgument(!namespace.endsWith("."), "Cannot create a namespace end with dot");
+    Preconditions.checkArgument(
+        !namespace.contains(" "), "Cannot create a namespace containing whitespace");
```
justinmclean, line 79: This is still not quite correct, we want to reject blank or whitespace-only input, not anything that contains a space.

yossev: Working on a fix.

Abyss-lord: I think it's better to use `StringUtils.containsOnly()` to check whether namespace only contains whitespace.
`Preconditions.checkArgument(StringUtils.containsOnly(namespace, " "));`

api/src/test/java/org/apache/gravitino/TestNamespace.java
```diff
@@ -60,6 +60,7 @@ public void testFromStringInvalidArgs() {
     Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString(".a"));
     Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString("a."));
     Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString("a..b"));
+    Assertions.assertThrows(IllegalArgumentException.class, () -> Namespace.fromString(" "));
```
Abyss-lord, line 63: I think we can add more cases, e.g. `""`、`"\r\n"`

PR #7968, opened 2025-08-07
[MINOR] fix(core): Fix the race issue in `LocalJobExecutor`

jerryshao:

Fix the race issue in `LocalJobExecutor`.

Fix the bug.

N/A

N/A

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
+++ b/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
@@ -174,7 +174,7 @@ public class LocalJobExecutor implements JobExecutor {
       if (!jobStatus.containsKey(jobId)) {
         throw new NoSuchJobException("No job found with ID: %s", jobId);
       }
-      LOG.trace(
+      LOG.debug(
           "Get status {} and finished time {} for job {}",
           jobStatus.get(jobId).getLeft(),
           jobStatus.get(jobId).getRight(),
--- a/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
+++ b/core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
@@ -258,11 +258,17 @@ public class LocalJobExecutor implements JobExecutor {
 
       Process process;
       synchronized (lock) {
-        jobStatus.put(jobId, Pair.of(JobHandle.Status.STARTED, UNEXPIRED_TIME_IN_MS));
+        // This happens when the job is cancelled before it starts.
+        Pair<JobHandle.Status, Long> statusPair = jobStatus.get(jobId);
+        if (statusPair == null || statusPair.getLeft() != JobHandle.Status.QUEUED) {
+          LOG.warn("Job {} is not in QUEUED state, cannot start it", jobId);
+          return;
+        }
 
         LocalProcessBuilder processBuilder = LocalProcessBuilder.create(jobTemplate, configs);
         process = processBuilder.start();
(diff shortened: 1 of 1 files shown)


jerryshao:
@yuqi1129 this is the fix for flaky test, please help to review.

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
```diff
@@ -156,12 +156,12 @@ public String submitJob(JobTemplate jobTemplate) {
     String newJobId = LOCAL_JOB_PREFIX + UUID.randomUUID();
     Pair<String, JobTemplate> jobPair = Pair.of(newJobId, jobTemplate);
 
-    // Add the job template to the waiting queue
-    if (!waitingQueue.offer(jobPair)) {
-      throw new IllegalStateException("Waiting queue is full, cannot submit job: " + jobTemplate);
-    }
-
     synchronized (lock) {
```
yuqi1129, line 159: Does this operation need to be protected by the lock?

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
```diff
@@ -255,15 +255,18 @@ public void runJob(Pair<String, JobTemplate> jobPair) {
...
       synchronized (lock) {
         jobStatus.put(jobId, Pair.of(JobHandle.Status.STARTED, UNEXPIRED_TIME_IN_MS));
-      }
 
-      LocalProcessBuilder processBuilder = LocalProcessBuilder.create(jobTemplate, configs);
-      Process process = processBuilder.start();
-      runningProcesses.put(jobId, process);
+        LocalProcessBuilder processBuilder = LocalProcessBuilder.create(jobTemplate, configs);
+        process = processBuilder.start();
+        runningProcesses.put(jobId, process);
```
yuqi1129, line 270: So, what is the root cause of CI failure? Based on your changes, the real process and status changes are not within a lock.

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
jerryshao: Which operation?

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
jerryshao: The main reason is the process is started before the status is changed to "queue", so the fetched status is always "queue".

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
yuqi1129: I have not problem about this point.

core/src/main/java/org/apache/gravitino/job/local/LocalJobExecutor.java
yuqi1129: Then there is no need to put the process start logic into the lock as you have done it in `submitJob`.  The current main logic can guarantee the status will be changed `RUNNING` before processes is started.

jerryshao: There's a problem if the `cancel` happens before the process start, it will lead to a wrong status.

yuqi1129: I see, good fix.

Issue #7797, opened 2025-07-24
[Bug report] role: catalog's fullName in securableObjects didn't update immediately after the catalog renamed


danhuawang:
### Version

main branch

### Describe what's wrong

 The catalog had been renamed to new name `new_auth_mysql_catalog1`,but fullName in role is still the old name.
Then if grant the role to a user immediately, it would fail due to NoSuchCatalogException.

<img width="776" height="793" alt="Image" src="https://github.com/user-attachments/assets/2414eb8b-ab7a-4a90-aad2-9bd14064d8b5" />

<img width="622" height="706" alt="Image" src="https://github.com/user-attachments/assets/8348c4a6-3b3e-4efa-8459-9dcca7953294" />

### Error message and/or stacktrace

N/A

### How to reproduce

1. Create a role as following
```
{
    "code": 0,
    "role": {
        "name": "tableRole2",
        "audit": {
            "creator": "anonymous",
            "createTime": "2025-07-24T09:41:32.518001927Z"
        },
        "properties": {
            "k1": "v1"
        },
        "securableObjects": [
            {
                "type": "catalog",
                "privileges": [
                    {
                        "name": "use_catalog",
                        "condition": "deny"
                    }
                ],
                "fullName": "auth_mysql_catalog1"
            }
        ]
    }
}
```
2. Rename catalog `auth_mysql_catalog1` to `new_auth_mysql_catalog1`
3. Get the role check fullName 
4. Grant the role to a user, it would fail due to NoSuchCatalogException.

danhuawang:
@jerqi Do you know any one can help check this issue?

roryqi:
@Abyss-lord Could you help about this issue?

Abyss-lord:
> [@Abyss-lord](https://github.com/Abyss-lord) Could you help about this issue?

sure.

LindaSummer:
Hi @xunliu ,

I am interested in this topic.😊
Could this issue be assigned to me?

Best Regards,
Edward

Issue #7787, opened 2025-07-23
[Bug report] EntityCache may cache wrong result when listEntitiesByRelation


mchades:
### Version

main branch

### Describe what's wrong

https://github.com/apache/gravitino/blob/b52b8d42515c1ccf7df32661a7fee820b54c59a7/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java#L207-L225

For the above cache logic, if a result with `allFields=false` is cached, a request with `allFields=true` will retrieve the incorrect cached result because the cache key does not include the `allFields` parameter.

### Error message and/or stacktrace

see above description

### How to reproduce

see above description

LindaSummer:
Hi @mchades ,

I want to work on this issue. 😄
Could it be assigned to me?

Best Regards,
Edward

jerryshao:
This implementation has another issue. If the entity is updated by `put` operation, but it is still cached here, so calling this method will get the old entity from cache and lead to error:

```
         Optional<List<E>> entities = cache.getIfPresent(relType, nameIdentifier, identType); 
         if (entities.isPresent()) { 
           return entities.get(); 
         } 
```

@Abyss-lord please take a look at this cache issue.

LindaSummer:
> This implementation has another issue. If the entity is updated by `put` operation, but it is still cached here, so calling this method will get the old entity from cache and lead to error:
> 
> ```
>          Optional<List<E>> entities = cache.getIfPresent(relType, nameIdentifier, identType); 
>          if (entities.isPresent()) { 
>            return entities.get(); 
>          } 
> ```
> 
> [@Abyss-lord](https://github.com/Abyss-lord) please take a look at this cache issue.

Hi @jerryshao ,

Please correct me if I misunderstand. 😊
Should we always delete the cache without updating it until the first loading to keep it idempotent?

Best Regards,
Edward

jerryshao:
The key problem here is that the same entity is cached twice by different keys, one using non-relation entity operation, one using relation entity operation. This will lead to inconsistent data.

If one is updated, then another is still old, so if updated using non-relation entity operation, and fetched using relation entity operation, there will be issue.

PR #7969, opened 2025-08-07
[#7797] fix(core): Fix get expired security objects when load role entity.

yuqi1129:

When a security object changed, other entity that associated with it should also updated.

It's bug introudced by entity store cache system.

Fix: #7797 

N/A.

UTs

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
@@ -20,14 +20,17 @@ package org.apache.gravitino.storage.relational;
 
 import static org.apache.gravitino.Configs.ENTITY_RELATIONAL_STORE;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -120,9 +158,39 @@ public <E extends Entity & HasIdentifier> E update(
...
       throws IOException, NoSuchEntityException, EntityAlreadyExistsException {
     cache.invalidate(ident, entityType);
+
+    // Invalid all related cache entries by relation.
+    invalidateCacheByRelation(ident, entityType);
+
     return backend.update(ident, entityType, updater);
   }
 
+  private void invalidateCacheByRelation(NameIdentifier ident, Entity.EntityType entityType)
```
xunliu, line 168: hi @yuqi1129 
Can we use this `invalidateCacheByRelation ` to fix https://github.com/apache/gravitino/issues/7787#issuecomment-3157777461?

jerryshao:
Fixed by another PR, I'm going to close this PR.

PR #7971, opened 2025-08-07
[#7972] fix(authz): Fix NPE when create table failed by not auth

hdygxsj:

Fix NPE when create table failed by not auth

Fix: #7972

None

Existing test

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
@@ -168,16 +168,20 @@ public class GravitinoInterceptionService implements InterceptionService {
         String currentUser,
         String methodName) {
       String contextualMessage;
+      String accessMetadataMessage =
+          accessMetadataName != null
+              ? String.format("on metadata '%s'", accessMetadataName.name())
+              : "";
       if (StringUtils.isNotBlank(errorMessage)) {
         contextualMessage =
             String.format(
-                "User '%s' is not authorized to perform operation '%s' on metadata '%s': %s",
-                currentUser, methodName, accessMetadataName.name(), errorMessage);
+                "User '%s' is not authorized to perform operation '%s' %s: %s",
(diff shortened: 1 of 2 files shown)


roryqi:
Could u add new test cases for this fix? Existing test cases can't cover it.

clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
```diff
@@ -233,7 +233,9 @@ private void throwFailure(
     }
 
     errorHandler.accept(errorResponse);
-
+    if (response.getCode() == 403) {
```
roryqi: You should add this to client `ErrorHandlers`.

hdygxsj: fixed

clients/client-java/src/main/java/org/apache/gravitino/client/ErrorHandlers.java
```diff
@@ -1085,6 +1085,9 @@ public ErrorResponse parseResponse(int code, String json, ObjectMapper mapper) {
 
     @Override
     public void accept(ErrorResponse errorResponse) {
+      if (errorResponse.getCode() == ErrorConstants.FORBIDDEN_CODE) {
```
roryqi, line 1088: The error message may be formatted in every type Handler. Do we need formatted error message?

hdygxsj: I don't understand what you mean. What is a formatted error message?

We need to convert 403 into a specific Exception instead of using RuntimeException, otherwise, throwing exceptions due to non-403 errors might cause the IT to fail to achieve the expected results.

roryqi: Client handler doesn't need to format the error message. Sorry, I don't have other comments.

jerqi (approved):

PR #7973, opened 2025-08-08
[#7787] fix(cache): add `allFields` check in caffeine cache

LindaSummer:

Add `allFields` check in cache entity.

Refer to #7787 , the cache entity misses the `allFields` part and may cause an inconsistent cache query.

Fix: #7787

No.

Unit tests have passed.


core/src/main/java/org/apache/gravitino/cache/EntityCacheKey.java
```diff
@@ -30,6 +30,24 @@ public class EntityCacheKey {
...
+   * @param relationType The type of the relation, it can be null.
+   * @param allFields Whether to retrieve all fields of the entity.
+   * @return A new instance of {@link EntityCacheKey}.
+   */
+  public static EntityCacheKey of(
+      NameIdentifier ident,
+      Entity.EntityType type,
+      SupportsRelationOperations.Type relationType,
+      boolean allFields) {
+    return new EntityCacheKey(ident, type, relationType, allFields);
```
yuqi1129: If `allFields` is false, then the entities retrieved are the same. If I call this API several times? How many fields are fetched from storage if `allFields` is false.  Are these fields constant?

core/src/main/java/org/apache/gravitino/cache/EntityCacheKey.java
```diff
@@ -41,7 +59,7 @@ public class EntityCacheKey {
    */
   public static EntityCacheKey of(
       NameIdentifier ident, Entity.EntityType type, SupportsRelationOperations.Type relationType) {
-    return new EntityCacheKey(ident, type, relationType);
+    return new EntityCacheKey(ident, type, relationType, null);
```
yuqi1129: What's the purpose of setting a Boolean field to null?

core/src/main/java/org/apache/gravitino/cache/EntityCacheKey.java
LindaSummer: Hi @yuqi1129 ,

I understand your concern now. Sorry for misunderstanding it as a three-state `true`, `false`, and `none` state.

I plan to remove the cache of `allFields == false` and only cache the value when `allFields` is `true`.
Could this logic match our design? 😊

Best Regards,
Edward

core/src/main/java/org/apache/gravitino/cache/EntityCacheKey.java
LindaSummer: Hi @yuqi1129 ,

Sorry that I thought we just needed to cache for three states. 😊
Please let me know whether the plan in the first comment matches our design.

Best Regards,
Edward

core/src/main/java/org/apache/gravitino/cache/EntityCacheKey.java
yuqi1129: @xunliu 
I'm okay with it. Do you have any opinions on this?

mchades: why not cache when `allFields==true`?

LindaSummer: Hi @mchades ,

In the current PR, the `allFields==true` will be cached, and the `allFields==false` will directly fall back to storage.

I think skipping the incomplete result in `allFields==false` could solve the inconsistency problem in the cache. 😊

Best Regards,
Edward

LindaSummer:
Hi @yuqi1129 , @mchades ,

Sorry to bother you.
Could you help take a look and give me some more suggestions? 😊

Best Regards,
Edward

mchades:
@xunliu Did you also work on fixing this?

jerryshao:
Fixed by another PR. Thanks a lot for your contribution @LindaSummer .

Issue #7974, opened 2025-08-08
[Improvement] Removing an id from a comment can leave trailing whitespace


justinmclean:
### What would you like to be improved?

In removeIdFromComment in core/src/main/java/org/apache/gravitino/StringIdentifier.java.

This test shows the issue:
```
  @Test
  public void testRemoveIdFromCommentTrimsTrailingSpaces() {
    StringIdentifier identifier = StringIdentifier.fromId(42L);
    String commentWithSpace = "This is a comment ";
    String commentWithId = StringIdentifier.addToComment(identifier, commentWithSpace);
    Assertions.assertEquals(
        commentWithSpace.trim(), StringIdentifier.removeIdFromComment(commentWithId));
  }
```

### How should we improve?

Trim the comment.

Sujal2114:
Please assign this to me

justinmclean:
Done. If you have any questions, just ask.

Auxilus08:
Hi! @justinmclean  I'd like to work on this issue. I've already identified the problem and have a fix ready. 

The issue is in the `removeIdFromComment` method where the fallback case doesn't trim the comment. The fix is simple - change `return comment;` to `return comment.trim();` on line 226.

Can this issue be assigned to me? I can submit a PR with the fix and test case.

justinmclean:
@Auxilus08 Thanks for the PR, but it’s best to check in before starting work on an issue, and to avoid working on issues that are already assigned to someone else.

Auxilus08:
@justinmclean Got it, thanks for the feedback! I’ll make sure to check in before picking up an issue next time and avoid ones that are already assigned

justinmclean:
@Sujal2114 Are you still working on this?

Sujal2114:
I am Sorry for not completing the assigned issue as I was busy with some other work. You can assign this issue to @Auxilus08 as he has already worked on this.

justinmclean:
No problem @Sujal2114 thanks for that, I've assigned it to @Auxilus08.

Auxilus08:
I will begin working on the issue again shortly.

PR #7988, opened 2025-08-08
[#7974] fix: trim trailing spaces in StringIdentifier.removeIdFromComment

Auxilus08:

Fix `StringIdentifier.removeIdFromComment` method to properly trim trailing spaces from comments when no identifier is found.

- Modified removeIdFromComment method to return comment.trim() when no identifier is found
- Added test case testRemoveIdFromCommentTrimsTrailingSpaces to verify the fix
- Ensures consistent behavior whether identifier is present or not

The current implementation inconsistently handles trailing spaces:
- When an identifier is found and removed: returns `comment.substring(0, indexOf).trim()` (trimmed)
- When no identifier is found: returns `comment` (not trimmed)

This causes the test case `testRemoveIdFromCommentTrimsTrailingSpaces` to fail because comments with trailing spaces are not properly normalized.

Fix: #7974

No. This is a bug fix that makes the behavior more consistent and predictable. No user-facing APIs are changed.

- Added new test case `testRemoveIdFromCommentTrimsTrailingSpaces` that verifies the fix
- Verified existing tests still pass
- The fix ensures both code paths consistently trim whitespace

Test case:
```java
@Test
public void testRemoveIdFromCommentTrimsTrailingSpaces() {
  StringIdentifier identifier = StringIdentifier.fromId(42L);
  String commentWithSpace = "This is a comment ";
  String commentWithId = StringIdentifier.addToComment(identifier, commentWithSpace);
  Assertions.assertEquals(
      commentWithSpace.trim(), StringIdentifier.removeIdFromComment(commentWithId));
}

### DIFF
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/ErrorHandlers.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/ErrorHandlers.java
@@ -1150,6 +1150,9 @@ public class ErrorHandlers {
 
     @Override
     public void accept(ErrorResponse errorResponse) {
+      if (errorResponse.getCode() == ErrorConstants.FORBIDDEN_CODE) {
+        throw new ForbiddenException("Forbidden error :%s", errorResponse.getMessage());
+      }
       throw new RESTException("Unable to process: %s", formatErrorMessage(errorResponse));
     }
   }
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
@@ -204,7 +204,6 @@ public class HTTPClient implements RESTClient {
   private void throwFailure(
       CloseableHttpResponse response, String responseBody, Consumer<ErrorResponse> errorHandler) {
     ErrorResponse errorResponse = null;
-
     if (responseBody != null) {
       try {
         if (errorHandler instanceof ErrorHandler) {
@@ -233,7 +232,6 @@ public class HTTPClient implements RESTClient {
     }
 
     errorHandler.accept(errorResponse);
-
     // Throw an exception in case the provided error handler does not throw.
     throw new RESTException("Unhandled error: %s", errorResponse);
   }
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

File: core/src/main/java/org/apache/gravitino/StringIdentifier.java [hunk 1]
```diff
@@ -219,9 +219,9 @@ public static String removeIdFromComment(String comment) {
       }
 
       if (indexOf != -1) {
-        return comment.substring(0, indexOf);
+        return comment.substring(0, indexOf).trim();
       }
     }
-    return comment;
+    return comment.trim();
```
justinmclean, line 225: This would change the comment even if it didn't have an identifier in it, I'm not sure we want to do that.

Auxilus08:
For comments without identifiers, we should preserve them exactly as they are.

Thanks for the review. I'll revert that change

core/src/main/java/org/apache/gravitino/StringIdentifier.java
Auxilus08: For comments without identifiers, we should preserve them exactly as they are.

Thanks for the review. I'll revert that change

justinmclean:
@Auxilus08 I notice you closed this issue without making the suggested change - was that intentional?

Auxilus08:
> @Auxilus08 I notice you closed this issue without making the suggested change - was that intentional?
I want to explain that I locked my PR but not the issue. I think there is a mistake because I assumed @Sujal2114 was actually taking this issue on since for some reason it was assigned. I would happily take it back on if it is still open so just let me know.

Auxilus08:
> @Auxilus08 I notice you closed this issue without making the suggested change - was that intentional?

I’d like to clarify that I closed my PR but not the issue itself. I believe there may have been a misunderstanding, as I assumed @Sujal2114 was actively working on it given the assignment. If the issue is still available, I would be happy to resume work on it. Please let me know if that would be helpful.
@justinmclean

justinmclean:
Yes, right you are. I'll ping @Sujal2114 and see if he is still working on it.

Auxilus08:
Sounds good, thank you! I’ll wait for your confirmation before proceeding

PR #7991, opened 2025-08-09
[#7958] fix(authz): Include recursive flag in hashCode(), equals(), toString()

leesj8116:

Updated the PathBasedMetadataObject class to:
- Compare the `recursive` flag in `equals()` and `hashCode()` methods.
- Include the `recursive` flag in the `toString()` output.

Previously, the `recursive` flag was not considered in equality checks or hash code generation,
which could lead to incorrect comparisons and inconsistent behavior in hash-based collections.
Additionally, the `toString()` output did not reflect the `recursive` state, making debugging harder.

Fix: #7958

No user-facing API changes.  
However, objects with different `recursive` flag values will now be treated as unequal.

- Added `testRecursiveFlagAffectsEquality()` to the `TestPathBasedMetadataObject` class.
- Added a case to `testToString()`.


leesj8116:
Sorry, I add wrong issue number. this PR is closed.

PR #7996, opened 2025-08-09
fix: trim trailing spaces in removeIdFromComment

Sujal2114:
Updated the removeIdFromComment method in StringIdentifier to trim the resulting comment after removing the identifier. This ensures that any trailing spaces left behind are removed, resulting in a clean comment string.

Why are the changes needed?
Previously, when removing an identifier from a comment, the method sometimes left an extra space at the end.
Example:
"This is a comment  (From Gravitino, DO NOT EDIT: gravitino.v1.uid42)"
would become:

"This is a comment " // trailing space remains
This caused unit test failures and returned inconsistent results. The change fixes this by trimming the comment.

Yes.
removeIdFromComment will now always return a trimmed comment string, with no leading or trailing whitespace.

How was this patch tested?
Verified with the unit test testRemoveIdFromCommentTrimsTrailingSpaces that the comment is properly trimmed after identifier removal.

Ran the full test suite to confirm no regressions.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/StringIdentifier.java
+++ b/core/src/main/java/org/apache/gravitino/StringIdentifier.java
@@ -219,7 +219,7 @@ public class StringIdentifier {
       }
 
       if (indexOf != -1) {
-        return comment.substring(0, indexOf);
+        return comment.substring(0, indexOf).trim();
       }
     }
     return comment;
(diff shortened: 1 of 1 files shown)


justinmclean:
I'm not sure why the CI is failing, but given there's no changes, I guess we'll wait until you add those changes again?

Issue #7977, opened 2025-08-08
[Improvement] Fix poossble race condition in TestMemoryEntityStore.java


justinmclean:
### What would you like to be improved?

The snapshot should be created inside the lock to avoid a race condition.

### How should we improve?

See above.

justinmclean:
Note it looks like the original issue with Spotless has been fixed:
https://github.com/diffplug/spotless/issues/1819

georgereuben:
I can take this up

justinmclean:
Go for it!

PR #7997, opened 2025-08-09
[#7977] fix: resolve race condition in snapshot creation by adjusting lock usage

georgereuben:
Previously, the snapshot for rollback was created outside the lock, which could lead to inconsistent state and race conditions in concurrent scenarios.
This change ensures the snapshot is always created while holding the lock, preventing data corruption and ensuring reliable transaction rollback.

Fix: #7977

No

With UTs added earlier

### DIFF
--- a/core/src/test/java/org/apache/gravitino/storage/memory/TestMemoryEntityStore.java
+++ b/core/src/test/java/org/apache/gravitino/storage/memory/TestMemoryEntityStore.java
@@ -154,8 +154,8 @@ public class TestMemoryEntityStore {
     @Override
     public <R, E extends Exception> R executeInTransaction(Executable<R, E> executable)
         throws E, IOException {
-      Map<NameIdentifier, Entity> snapshot = createSnapshot();
       lock.lock();
+      Map<NameIdentifier, Entity> snapshot = createSnapshot();
       try {
         return executable.execute();
       } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/apache/gravitino/storage/memory/TestMemoryEntityStore.java
```diff
@@ -176,17 +176,10 @@ public void close() throws IOException {
...
-        return entityMap.entrySet().stream()
-            .collect(
-                Collectors.toMap(
-                    Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, Maps::newHashMap));
-      } catch (Exception e) {
-        throw new RuntimeException("Failed to create snapshot of entity store", e);
-      } finally {
-        lock.unlock();
-      }
+      return entityMap.entrySet().stream()
```
justinmclean, line 179: Any reason teh try/catch was removed?

georgereuben: I had added this method earlier. Just noticed that try catch blocks were not being used in other sections of the file so I'd removed it for consistency since I didn't think this one was important. Should I add it back?

justinmclean: No problem, I think that's good.

PR #7998, opened 2025-08-09
Used passed in lineSeparator in writeHorizontalLine

### LINKED ISSUE #7978: [Improvement] In TableFormat.java the passed in lineSeparator is not used

In clients/cli/src/main/java/org/apache/gravitino/cli/outputs/TableFormat.java the passed in lineSeparator towriteHorizontalLine is not used.

### How should we improve?

Use it

Spiritedswordsman:
In clients/cli/src/main/java/org/apache/gravitino/cli/outputs/TableFormat.java 

This update ensures that the lineSeparator argument passed to writeHorizontalLine is actually used instead of being ignored.
Previously, the method always used System.lineSeparator(). Now it directly writes the provided `lineSeparator`.

Closes #7978

### DIFF
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/outputs/TableFormat.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/outputs/TableFormat.java
@@ -431,7 +431,7 @@ public abstract class TableFormat<T> extends BaseOutputFormat<T> {
     if (right != null) {
       osw.write(right);
     }
-    
+
     osw.write(lineSeparator != null ? lineSeparator : System.lineSeparator());
   }
 


clients/cli/src/main/java/org/apache/gravitino/cli/outputs/TableFormat.java
```diff
@@ -351,9 +351,7 @@ private static void writeHorizontalLine(
       osw.write(right);
     }
 
-    if (lineSeparator != null) {
-      osw.write(System.lineSeparator());
-    }
+    osw.write(lineSeparator);
```
justinmclean: I think we still need to use the null check

Spiritedswordsman: Sure, 
But then the function right below "writeRow" Didn't had any null check so I followed the same format.

Soo shall I add a null check for that too?

justinmclean: It probably wouldn't hurt to have one, so yes add it. Sorry I didn't notice the inconsistency.

Spiritedswordsman: Yea, no issues

Spiritedswordsman: Heyy, how about using `System.lineSeparator()` if lineSeparator is null?

justinmclean: I would guess that is null, then the intent is that no separator should be, but you could also do that. I don't really have a strong feeling either way.

Spiritedswordsman:
Heyy, Im done with this task, let me know if you want any other changes :v: 
also I would visit other issues I can work on.

justinmclean (dismissed):
Thanks for fixing this

justinmclean (approved):
Thanks for fixing this

Issue #7979, opened 2025-08-08
[Improvement] Add reserved words check for MySQL/Postgres


justinmclean:
### What would you like to be improved?

In:
catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/MysqlCatalogCapability.java
catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/PostgreSqlCatalogCapability.java

A TODO exists to validate the name against reserved words.

A good starting point would be to have these as reserved words for MySQL:
mysql
information_schema
performance_schema
sys

For Postgres:
 pg_catalog
 information_schema

### How should we improve?

See above

georgereuben:
I can take this up

PR #8002, opened 2025-08-10
[#7979] Improvement(catalogs): Add reserved words check for MySQL and Postgres

georgereuben:
- Added validation for reserved words in PostgreSQL catalog capability.
- Reserved words (like pg_catalog, information_schema) are now checked for both schema and table scopes to prevent naming conflicts.

Prevents users from creating schemas or tables with names that conflict with system catalogs or schemas.

Fix: #7979 

Users will receive an error if they attempt to use reserved names for schemas or tables in PostgreSQL catalogs.

UTs and ITs

### DIFF
--- a/catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/MysqlCatalogCapability.java
+++ b/catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/MysqlCatalogCapability.java
@@ -18,6 +18,7 @@
  */
 package org.apache.gravitino.catalog.mysql;
 
+import java.util.Set;
 import org.apache.gravitino.connector.capability.Capability;
 import org.apache.gravitino.connector.capability.CapabilityResult;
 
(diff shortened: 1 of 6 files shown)


yuqi1129:
@georgereuben , could you please add some UTs or ITs to cover the changes?

georgereuben:
> @georgereuben , could you please add some UTs or ITs to cover the changes?

Added UTs and ITs, please let me know if any changes are needed

catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/MysqlCatalogCapability.java
```diff
@@ -39,13 +40,23 @@ public class MysqlCatalogCapability implements Capability {
...
   public CapabilityResult specificationOnName(Scope scope, String name) {
-    // TODO: Validate the name against reserved words
     if (!name.matches(MYSQL_NAME_PATTERN)) {
       return CapabilityResult.unsupported(
           String.format("The %s name '%s' is illegal.", scope, name));
     }
+
+    if ((scope == Scope.SCHEMA || scope == Scope.TABLE)
+        && MYSQL_RESERVED_SCHEMAS.contains(name.toLowerCase())) {
+      return CapabilityResult.unsupported(
```
yuqi1129, line 55: I can't get why tables name like `performance_schema` and `sys` are also not supported.

georgereuben: Thanks for pointing this out. My initial intention was to prevent conflicts by restricting both schema and table names, but I see now that restricting reserved schema names like the ones above won't be needed for tables. I'll update the validation so that only schema names are checked against MYSQL_RESERVED_SCHEMAS.

yuqi1129:
Since those names are protected, I believe that MySQL & PG will throw exceptions if we attempt to create a schema with that name. So why do we check it again? Is it just to throw an exception early?

georgereuben:
> Since those names are protected, I believe that MySQL & PG will throw exceptions if we attempt to create a schema with that name. So why do we check it again? Is it just to throw an exception early?

Yes, it would provide early validation and fail-fast before db operations, and clearer error logs for users that we can make consistent across the different db catalogs.

yuqi1129:
> > Since those names are protected, I believe that MySQL & PG will throw exceptions if we attempt to create a schema with that name. So why do we check it again? Is it just to throw an exception early?
> 
> Yes, it would provide early validation and fail-fast before db operations, and clearer error logs for users that we can make consistent across the different db catalogs.

I see. This PR is acceptable to me, and I'm going to merge it today. By the way, could you also raise other PRs to solve similar problems in other catalogs , such as StarRocks, Doris, Hive , etc.

georgereuben:
> > > Since those names are protected, I believe that MySQL & PG will throw exceptions if we attempt to create a schema with that name. So why do we check it again? Is it just to throw an exception early?
> > 
> > 
> > Yes, it would provide early validation and fail-fast before db operations, and clearer error logs for users that we can make consistent across the different db catalogs.
> 
> I see. This PR is acceptable to me, and I'm going to merge it today. By the way, could you also raise other PRs to solve similar problems in other catalogs , such as StarRocks, Doris, Hive , etc.

Sure, I'll look into it

Issue #7937, opened 2025-08-05
[Subtask] Add the integration test for the job system


jerryshao:
### Describe the subtask

This issue tracks the work of adding the integration test for the job system.

### Parent issue

#7154

PR #8015, opened 2025-08-11
[#7937] test(core): Add ITs for job system

jerryshao:

Add the integration tests to the job system, also fix the issues during the tests.

Fix: #7937 

No.

Add the tests.


jerryshao:
@yuqi1129 @mchades please help to review this PR, thanks.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
```diff
@@ -38,8 +38,6 @@ enum Type {
     ROLE_USER_REL,
     /** Role and group relationship */
     ROLE_GROUP_REL,
-    /** Job template and job relationship */
-    JOB_TEMPLATE_JOB_REL,
```
yuqi1129, line 42: Is it no longer in use?

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/JobIT.java
```diff
@@ -0,0 +1,356 @@
...
+    Assertions.assertEquals(jobHandle.jobId(), retrievedJob.jobId());
+    Assertions.assertEquals(JobHandle.Status.CANCELLED, retrievedJob.jobStatus());
+
+    // Test cancel a non-existent job
+    Assertions.assertThrows(
+        NoSuchJobException.class, () -> metalake.cancelJob("non_existent_job_id"));
+  }
+
+  private String generateTestEntryScript() {
+    String content =
```
yuqi1129, line 312: Since only JDK17 is allowed to compile the project, I believe we can use the following JDK17 feature
```
"""
Content here
"""
```

to improve the code below.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -313,19 +321,25 @@ public List<JobEntity> listJobs(String metalake, Optional<String> jobTemplateNam
               NameIdentifier jobTemplateIdent =
                   NameIdentifierUtil.ofJobTemplate(metalake, jobTemplateName.get());
 
+              // If jobTemplateName is present, we need to list the jobs associated with the job.
+              // Using a mock namespace from job template identifier to get the jobs associated
+              // with job template.
+              String[] elements =
+                  Stream.concat(
```
yuqi1129: ArrayUtils.add(jobTemplateIdent.namespace().levels(), jobTemplateIdent.name());

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
jerryshao: Yes, because of the bug in entity cache, I use a different way to implement the job and job template store, not using this interface.

yuqi1129:
There is a problem in `JobOperations`, you can fix it by the way
https://github.com/apache/gravitino/blob/0ac8ac2a81c524c1b00e7a0b779ea73cecb9edac/server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java#L202-L209

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/JobIT.java
jerryshao: The client module needs to be compatible with jdk8, I think it is fine leave as it is.

yuqi1129: I see.  This module will be built with the target `JDK8` and will use `JDK17` for source compatibility, allowing us to use JDK17 syntax.  Anyway, this is not a major issue; leaving it as is or not does not matter too much.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
yuqi1129: I see.

jerryshao:
@yuqi1129 do you have further comments? @mchades please also take a look.

yuqi1129:
> @yuqi1129 do you have further comments? @mchades please also take a look.

I have no more valueable comments.

jerryshao:
Can you please approve this PR? Thanks.

yuqi1129:
> Can you please approve this PR? Thanks.

OK, let's merge it first, @mchades. Maybe he is busy with his tasks.

Issue #8012, opened 2025-08-11
[Improvement] fix hashcode in TableEntity.java


justinmclean:
### What would you like to be improved?

Currently, hashcode doesn't include namespace as this test shows:
```
  @Test
  public void testHashCodeIncludesNamespace() {
    AuditInfo audit =
        AuditInfo.builder().withCreator("test").withCreateTime(Instant.now()).build();

    TableEntity table1 =
        TableEntity.builder()
            .withId(1L)
            .withName("t")
            .withNamespace(Namespace.of("catalog", "schema1"))
            .withColumns(Collections.emptyList())
            .withAuditInfo(audit)
            .build();

    TableEntity table2 =
        TableEntity.builder()
            .withId(1L)
            .withName("t")
            .withNamespace(Namespace.of("catalog", "schema2"))
            .withColumns(Collections.emptyList())
            .withAuditInfo(audit)
            .build();

    Assertions.assertNotEquals(
        table1.hashCode(), table2.hashCode(), "hashCode should include namespace");
  }
```

### How should we improve?

hashcode is missing namespace, so it should be added.

joonseolee:
Can I take this ticket?

justinmclean:
It's yours. If you need any help ask.

PR #8019, opened 2025-08-11
[#8012] fix: include namespace in TableEntity hashCode

joonseolee:

This PR updates the TableEntity.hashCode() implementation to include the namespace field in the hash calculation.  
It also updates related test cases to validate the change and applies Spotless formatting to comply with the project code style.

The current hashCode() implementation does not include the namespace field, causing objects with different namespaces to have identical hash values.
This could lead to incorrect behavior when TableEntity objects are used in hash-based collections (e.g., HashSet, HashMap).
By including the namespace, equality and hash consistency are ensured.

Fix: #8012 

No. This change only affects internal equality/hash logic of TableEntity and does not modify any user-facing API or property keys.

Updated existing unit tests in TestEntity.java


core/src/main/java/org/apache/gravitino/meta/TableEntity.java
```diff
@@ -141,7 +141,7 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(id, name, auditInfo, columns);
+    return Objects.hashCode(id, name, auditInfo, columns, namespace);
```
justinmclean, line 144: This may still not be correct; try testing with two different namespaces that have the same names.

joonseolee: @justinmclean 

Thank you for your comment :)  
Just to confirm, are you suggesting that I should add test cases to verify the scenario where TableEntities have the same name but different namespaces??

```java
  @Test
  public void testHashCodeWithDifferentNamespaceHavingSameValues() {
    AuditInfo audit = AuditInfo.builder().withCreator("test").withCreateTime(Instant.now()).build();

    TableEntity table1 =
        TableEntity.builder()
            .withId(2L)
            .withName("t")
            .withNamespace(Namespace.of("catalog", "schema1"))
            .withColumns(Collections.emptyList())
            .withAuditInfo(audit)
            .build();

    TableEntity table2 =
        TableEntity.builder()
            .withId(2L)
            .withName("t")
            .withNamespace(Namespace.of("catalog", "schema1"))
            .withColumns(Collections.emptyList())
            .withAuditInfo(audit)
            .build();

    Assertions.assertEquals(table1.hashCode(), table2.hashCode(), "hashCode should be the same");
  }
```

justinmclean: Yes please add that test as well

joonseolee: @justinmclean 

Thank you !  
I have added the tests.

justinmclean (approved):
Thanks for that

Issue #8007, opened 2025-08-11
[Improvement] Fix hashcode for IndexDTO.java


justinmclean:
### What would you like to be improved?

There is an issue with the hashcode of common/src/main/java/org/apache/gravitino/dto/rel/indexes/IndexDTO.java

Running this test shows the issue:
```
  @Test
  public void testHashCodeConsistentWithEquals() {
    String[][] fields1 = new String[][] {{"a"}, {"b"}};
    String[][] fields2 = new String[][] {{"a"}, {"b"}};

    IndexDTO index1 = IndexDTO.builder()
        .withIndexType(IndexDTO.IndexType.PRIMARY_KEY)
        .withName("idx")
        .withFieldNames(fields1)
        .build();

    IndexDTO index2 = IndexDTO.builder()
        .withIndexType(IndexDTO.IndexType.PRIMARY_KEY)
        .withName("idx")
        .withFieldNames(fields2)
        .build();

    Assertions.assertEquals(index1, index2);
    Assertions.assertEquals(index1.hashCode(), index2.hashCode());
  }
```

### How should we improve?

fix the code in hashCode

iinow:
can I take this issue?

justinmclean:
Sure, if you need help just ask.

PR #8020, opened 2025-08-11
[#8007] fix: include fieldnames in hashCode

iinow:

Modified IndexDTO's hashCode() method to include fieldNames in the hash calculation.
The fieldNames is a two-dimensional String[][] array, and each sub-array’s hash is incorporated to ensure more accurate hash codes.

This fix ensures that fieldNames contributes to the hash code, maintaining the contract between equals() and hashCode().

Fix: #8007

No

Added a new test class org.apache.gravitino.dto.rel.indexes.TestIndexDTO to verify the updated hashCode() behavior.


justinmclean:
Hi the CI failed as the file TestIndexDTO.java is missing an ASF license header.

iinow:
@justinmclean 

I added ASF license header

justinmclean (approved):
thanks for fixing this

Issue #8009, opened 2025-08-11
[Improvement] Posssible NPE in ModelVersionUpdatesRequest.java


justinmclean:
### What would you like to be improved?

In common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionUpdatesRequest.java a null pointer exception will happen if update is null when validate() is call.

### How should we improve?

Check if it is null and throw an IllegalArgumentException

joonseolee:
Can I take this issue?

justinmclean:
Yes you can. If you need help, just ask.

PR #8021, opened 2025-08-11
[#8009] fix: prevent NPE in validate() when ModelVersionUpdatesRequest updates is null

joonseolee:

This PR adds a null check for the updates field in ModelVersionUpdateRequest.validate().
If updates is null, the method now throws an IllegalArgumentException with a clear message instead of triggering a NullPointerException.
This ensures that validation fails gracefully when no updates are provided.

Previously, when updates was null, calling updates.forEach(...) would result in a NullPointerException, which is less informative and harder to debug.
By explicitly checking for null and throwing IllegalArgumentException, the code provides a clear error message and prevents unexpected runtime failures.

Fix: #8009 

No

Added a test to verify that validate() throws IllegalArgumentException when updates is null.

### DIFF
--- a/common/src/test/java/org/apache/gravitino/dto/requests/TestModelVersionUpdatesRequest.java
+++ b/common/src/test/java/org/apache/gravitino/dto/requests/TestModelVersionUpdatesRequest.java
@@ -16,35 +16,20 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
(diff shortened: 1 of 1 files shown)


justinmclean:
The CI is failing as the file TestModelVersionUpdatesRequest.java is misisng an ASF license header.

joonseolee:
@justinmclean 

Thank you!  
I have added the header :)

Issue #7981, opened 2025-08-08
[Improvement] In StatisticValues.java order object statistics


justinmclean:
### What would you like to be improved?

In api/src/main/java/org/apache/gravitino/stats/StatisticValues.java ensure that there is a deterministic field order for object statistics by sorting map entries before building the StructType, preventing inconsistent ordering across runs.

### How should we improve?

See above.

bcoke-scottlogic:
Hi Justin, I'm interested in working on this issue. Would it be possible to assign it to me?

justinmclean:
It's yours if you have any questions just ask.

PR #8022, opened 2025-08-11
[#7981] Improvement (api): Sort object statistics to ensure deterministic field order

bcoke-scottlogic:

Sort object statistics with a treemap in StatisticValues.

Sorting object statistics prevents inconsistent order of fields across runs.

Fix: #7981

N/A

Added a unit test that checks that the object value map is sorted when the object value is created.


api/src/main/java/org/apache/gravitino/stats/StatisticValues.java
```diff
@@ -291,7 +292,7 @@ public static class ObjectValue implements StatisticValue<Map<String, StatisticV
     private ObjectValue(Map<String, StatisticValue<?>> valueMap) {
       Preconditions.checkArgument(
           valueMap != null && !valueMap.isEmpty(), "Values cannot be null or empty");
-      this.valueMap = valueMap;
+      this.valueMap = new TreeMap<>(valueMap);
```
diqiu50: Should we check whether the passed-in map is a TreeMap before copying?

bcoke-scottlogic: The TreeMap is for sorting the passed in map, it shouldn't matter what kind of map is passed in. Is the suggestion that if the passed map is already a TreeMap we set valueMap without the TreeMap constructor? E.g. 
```
      if (valueMap instanceof TreeMap) {
        this.valueMap = valueMap;
      } else {
        this.valueMap = new TreeMap<>(valueMap);
      }
```

justinmclean: That looks good to me.

bcoke-scottlogic: Thanks, I've put the change in.

Issue #8017, opened 2025-08-11
[Improvement] Use SQL prepared statements in JdbcDatabaseOperations.java and PostgreSqlSchemaOperations.java


justinmclean:
### What would you like to be improved?

In the exist method the SQL string can be changed to use safer prepared statements.

### How should we improve?

generateDatabaseExistSql woudl need to change as well.

ankamde:
@justinmclean I would like to start working on this issue. Please assign it to me.

justinmclean:
Done. If you have any questions please ask.

ankamde:
@justinmclean You can un-assign me. Looks like @Manjunatha-h was faster.

justinmclean:
Hi @Manjunatha-h, thank you for your PR. However, please avoid working on tickets that have already been assigned to others in the future. This helps ensure clear ownership and avoids duplication of effort.

Manjunatha-h:
ok sir @justinmclean .

K-Sqr:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. Before I start, is there a preferred approach for identifier positions (like database names in DDL) that can't be bound as parameters? Any pointers would be appreciated!

PR #8025, opened 2025-08-11
updated SQL prepared statements in JdbcDatabaseOperations.java and Po…

Manjunatha-h:
Title:
[#8017] fix: use SQL prepared statements in JdbcDatabaseOperations and PostgreSqlSchemaOperations

Replaced direct SQL statement execution with prepared statements in
JdbcDatabaseOperations.java and PostgreSqlSchemaOperations.java to improve security and prevent SQL injection vulnerabilities.

Why are the changes needed?
Direct SQL execution can lead to SQL injection risks.

Prepared statements ensure that query parameters are safely bound, improving both security and reliability.

This change aligns with best practices for database access in Java.

Fix: #8017

No. This is an internal improvement that does not change the public API or user-facing behavior.

How was this patch tested?
Verified that database operations still function as expected using prepared statements.

Ran existing unit tests to ensure no regressions were introduced.

Checked that queries execute correctly with different parameter values.

### DIFF
--- a/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcDatabaseOperations.java
+++ b/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcDatabaseOperations.java
@@ -22,6 +22,7 @@ package org.apache.gravitino.catalog.jdbc.operation;
 import com.google.common.collect.ImmutableMap;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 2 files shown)


Manjunatha-h:
@justinmclean here it is sir.   #8017

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcDatabaseOperations.java
```diff
@@ -123,8 +125,12 @@ public boolean exist(String databaseName) {
 
   protected String generateDatabaseExistSql(String databaseName) {
```
georgereuben, line 126: Can we add a deprecated label to this method if not for safe deleting it, since it's been replaced by `generateDatabaseExistPreparedSql()`?

justinmclean: I'd keep the method name the same and alter it rather than add a new method.

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlSchemaOperations.java
```diff
@@ -179,9 +179,18 @@ private String getShowSchemaCommentSql(String schema) {
         schema);
   }
 
+  //UPDATED VERSION OF GETSHOWSCHEMACOMMENTSQL
```
justinmclean, line 182: There is no need for comments like this

justinmclean:
The CI is failing because we use spotless to format the code. You'll need to run, this to fix:
```
./gradlew :catalogs:catalog-jdbc-common:spotlessApply
./gradlew :catalogs:catalog-jdbc-postgresql:spotlessApply
```
and push the changes.

justinmclean:
@Manjunatha-h are you still working on this?

Issue #8008, opened 2025-08-11
[Improvement] Impove exception message in AuthenticationFactory.java


justinmclean:
### What would you like to be improved?

In lineage/src/main/java/org/apache/gravitino/lineage/auth/AuthenticationFactory.java:
```
 throw new BadRequestException("Unsupported authentication ");
```

### How should we improve?

The space at the end probably meant that the code should be:
```
throw new BadRequestException("Unsupported authentication " + authType);
```

viralkachhadiya:
please assign me.

justinmclean:
Done. if you need help just ask, but I think you should be fine with this one.

justinmclean:
This has been fixed

PR #8027, opened 2025-08-11
[#8008] Improved exception message

viralkachhadiya:
1. [#8008 ] Improvement(AuthenticationFactory): Improved exception message
 
- Improved exception message on AuthenticationFactory.java

  1. current exception messages in AuthenticationFactory.java are too generic and incomplete, which makes debugging harder

Fix: #(8008)

  1. Users/developers who trigger these errors will now receive more descriptive and actionable messages.
  2. No functional behavior change — only the exception text changes.

Not required

### DIFF
--- a/lineage/src/main/java/org/apache/gravitino/lineage/auth/AuthenticationFactory.java
+++ b/lineage/src/main/java/org/apache/gravitino/lineage/auth/AuthenticationFactory.java
@@ -32,7 +32,7 @@ public class AuthenticationFactory {
       case "none":
         return new NoAuthStrategy();
       default:
-        throw new BadRequestException("Unsupported authentication " + String.valueOf(authType));
+        throw new BadRequestException("Unsupported authentication " + authType);
     }
   }
 }


lineage/src/main/java/org/apache/gravitino/lineage/auth/AuthenticationFactory.java
```diff
@@ -32,7 +32,7 @@ public static LineageServerAuthenticationStrategy createStrategy(String authType
       case "none":
         return new NoAuthStrategy();
       default:
-        throw new BadRequestException("Unsupported authentication ");
+        throw new BadRequestException("Unsupported authentication " + String.valueOf(authType));
```
justinmclean: I don't think String.valueOf is needed hare as authType is a string?

viralkachhadiya: okay. I thought whether it will be happen with nullable/empty string . let me do it like as you said.

justinmclean (approved):
Thanks for that

viralkachhadiya:
CC @justinmclean

jerryshao:
Please fix the PR title to follow others, thanks.

Issue #8011, opened 2025-08-11
[Improvement] Code in LockManager.java is incorrect


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/lock/LockManager.java, the deadlock logging in LockManager attempts to format the entire map instead of the actual timestamp, leading to an IllegalArgumentException when checking for long-held locks.

Here's a test to show this:
```
  @Test
  void testCheckDeadLockDoesNotThrow() {
    LockManager lockManager = new LockManager(getConfig());
    TreeLockNode node = new TreeLockNode("test-node");
    node.addHoldingThreadTimestamp(
        Thread.currentThread(),
        NameIdentifier.of("test"),
        System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(40));
    Assertions.assertDoesNotThrow(() -> lockManager.checkDeadLock(node));
  }
```

### How should we improve?

just format the timestamp

TheYorouzoya:
Can I work on this one, please?

roryqi:
cc @yuqi1129

justinmclean:
@TheYorouzoya  it's yours. If you need a hand just ask.

PR #8043, opened 2025-08-12
[#8011] Improvement(LockManager): Corrected code in LockManager.java

TheYorouzoya:

This PR fixes a bug in the checkDeadLock() function inside core/src/main/java/org/apache/gravitino/lock/LockManager.java.
The code is modified to correctly format each thread's timestamp inside the forEach loop.

Previously, the deadlock logging in LockManager was trying to format the entire map instead of the actual timestamp, which was leading to an IllegalArgumentException when checking for long-held locks.

Fix: #8011

No.

Locally, with a unit test provided in the original issue. Since the fix basically amounts to fixing a typo rather than changing any core functionality, I've opted to not include the test with the PR.


yuqi1129:
Well done.

Issue #8041, opened 2025-08-12
[Bug report] Unable to use spark to access fileset with S3


yuqi1129:
### Version

main branch

### Describe what's wrong

25/08/12 16:23:19 WARN MetricsConfig: Cannot locate configuration: tried hadoop-metrics2-s3a-file-system.properties,hadoop-metrics2.properties
25/08/12 16:23:24 WARN FileSystem: Failed to initialize fileystem s3a://xiaoyu-bucket/test/example_with_dir: java.io.IOException: Class class org.apache.gravitino.aws.shaded.org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider does not implement AWSCredentialsProvider
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/yuqi/venv/lib/python3.9/site-packages/pyspark/sql/readwriter.py", line 1864, in csv
    self._jwrite.csv(path)
  File "/Users/yuqi/venv/lib/python3.9/site-packages/py4j/java_gateway.py", line 1322, in __call__
    return_value = get_return_value(
  File "/Users/yuqi/venv/lib/python3.9/site-packages/pyspark/errors/exceptions/captured.py", line 179, in deco
    return f(*a, **kw)
  File "/Users/yuqi/venv/lib/python3.9/site-packages/py4j/protocol.py", line 326, in get_return_value
    raise Py4JJavaError(
py4j.protocol.Py4JJavaError: An error occurred while calling o66.csv.
: org.apache.gravitino.exceptions.GravitinoRuntimeException: Exception occurs when create new FileSystem for fileset: test.s3_credential_catalog.schema.example_with_dir, location: null, msg: Class class org.apache.gravitino.aws.shaded.org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider does not implement AWSCredentialsProvider
	at org.apache.gravitino.filesystem.hadoop.BaseGVFSOperations.lambda$getActualFileSystemByLocationName$1(BaseGVFSOperations.java:598)
	at org.apache.gravitino.shaded.com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2406)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1916)
[...]

PR #8050, opened 2025-08-13
[#8041] fix(fileset-fs): fix dependency conflicts

mchades:

fix dependency conflicts

Fix: #8041 

no

local tested


jerryshao:
@yuqi1129 please help to review.

yuqi1129:
Do you have time to test the other three cloud storage systems?  The current changes have already been verified by myself and others. The only thing that worries me are whether others like GCS share similar problems.

mchades:
After tracking the source code, I have another idea that needs to be tested first. I won't make this PR ready for review until I validate the changes.

docs/fileset-catalog-with-s3.md
mchades, line 1: I think you can first review the new clarified target and the shading strategy, while I will conduct the tests at the same time. @jerryshao @yuqi1129 @FANNG1

bundles/aws/build.gradle.kts
```diff
@@ -40,15 +43,12 @@ dependencies {
     exclude("*")
   }
 
-  implementation(libs.aws.iam)
-  implementation(libs.aws.policy)
```
FANNG1, line 44: s3 token provider relies on these packages.

bundles/aws/build.gradle.kts
```diff
@@ -63,19 +63,10 @@ tasks.withType(ShadowJar::class.java) {
     exclude(dependency("org.slf4j:slf4j-api"))
   }
 
-  relocate("com.amazonaws", "org.apache.gravitino.aws.shaded.com.amazonaws")
```
FANNG1, line 66: if not shading `software.amazon`, the package may conflict with the amazon package in Iceberg aws bundle jar

bundles/aws/build.gradle.kts
mchades: When using this `gravitino-aws.jar`, we assume that the environment already has all AWS dependencies.

bundles/aws/build.gradle.kts
mchades: `gravitino-aws.jar` does not introduce `software.amazon` dependency

bundles/aws/build.gradle.kts
FANNG1: According to my understanding, there is no such assumption.  the bundle jars is used to provide the dependencies jars. cc @yuqi1129

FANNG1: For the user cases which doesn't need hadoop and aws jars, another solution is we did't provide any bundle jars.  I guess user may use some code in bundle, we could move this part out of bundle.

mchades: please see the updated doc for `gravitino-aws.jar` and `gravitino-aws-bundle.jar` in this PR:

> - [`gravitino-aws-bundle-${gravitino-version}.jar`](https://mvnrepository.com/artifact/org.apache.gravitino/gravitino-aws-bundle) is a bundle that 
>   contains all the dependencies needed to access S3. It is recommended to use this JAR when your environment does not include cloud storage dependencies. 
>   You can use this bundle to access cloud storage through Gravitino directly without downloading extra dependencies.
> - [`gravitino-aws-${gravitino-version}.jar`](https://mvnrepository.com/artifact/org.apache.gravitino/gravitino-aws) is a condensed JAR. 
>   It is recommended to use this JAR when your environment can already access cloud storage. It only contains Gravitino's own code and shaded basic dependencies,
>   which can ensure that there will be no conflicts with the dependencies of cloud storage.

yuqi1129: Since lacking so many dependencies, how can we assure that it's easy for users to use the jar. IFAIK, the difference between `aws` and `aws-bundle` is that the latter contains dependencies `hadoop-aws` and `hadoop`, so is it reasonable to exclude dependencies like `libs.aws.iam`, `libs.aws.policy`?

mchades: > Since lacking so many dependencies, how can we assure that it's easy for users to use the jar. 

It is because we assume that the user environment already has access to cloud storage, so these users find it easy to use. If not, they should use `gravitino-aws-bundle.jar`.

> is it reasonable to exclude dependencies like libs.aws.iam, libs.aws.policy?

`libs.aws.iam` and `libs.aws.policy` are also a part of aws dependency, so I think it's reasonable to exclude them from `gravitino-aws.jar`

yuqi1129: > It is because we assume that the user environment already has access to cloud storage, so these users find it easy to use. If not, they should use gravitino-aws-bundle.jar.

Assuming we are going to access S3 through GVFS and only need to access the files, and there is already `hadoop-aws` and `hadoop` dependencies there, based on your changes, I also need to put `libs.aws.policy` and `libs.aws.iam` in the class path to make it runs, however, I would never going to use `IAM` and `Policy`.

Another point is that adding `aws.jar` to the classpath of Hadoop command line should be fine to make Hadoop works, after this change, we needs more extra packages.

mchades: > I also need to put `libs.aws.policy` and `libs.aws.iam` in the class path to make it runs

Are you sure? If it's just simple file access, I think there's no need to introduce these two dependencies. (the test `TestCredentialProvider` in this module runs well and we only use `testImplementation(libs.aws.s3)`)

> Another point is that adding aws.jar to the classpath of Hadoop command line should be fine to make Hadoop works, after this change, we needs more extra packages.

we needs more extra packages like what?

yuqi1129: @FANNG1

FANNG1: @mchades @yuqi1129 , let's propose a offline meet to dissuss it

yuqi1129: Sure. cc @mchades

mchades:
#8357 fix the issue

PR #8059, opened 2025-08-13
fix(docs): fix a url link in the authorization doc.

zhongyujiang:
Correct a link in the authorization doc.


jerryshao (approved):
LGTM, thanks for the fix.

Issue #8061, opened 2025-08-13
[Improvement] Simply build task in github CI about changes in mcp-server


yuqi1129:
### What would you like to be improved?

If there are changes in the module `mcp-server`,  there is no need to build the whole project, and only mcp-server is needed to compile.

PR #8062, opened 2025-08-13
[#8061] improvement(mcp-server): Add a CI pipeline to build module `mcp` specifically.

yuqi1129:

Add a CI pipeline to build module `mcp` specifically.

To optimize CI resources

Fix: #8061 

N/A.

CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,11 +43,16 @@ jobs:
               - '!**/*.md'
               - '!**/*.png'
               - '!**/*.svg'
+              - '!mcp-server/**'
             spark_connector_changes:
               - spark-connector/**
+            
+            mcp_server_changes:
+              - mcp-server/**
     outputs:
       source_changes: ${{ steps.filter.outputs.source_changes }}
       spark_connector_changes: ${{ steps.filter.outputs.spark_connector_changes }}
+      mcp_server_changes: ${{ steps.filter.outputs.mcp_server_changes }}
 
   compile-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


yuqi1129:
Commit [fc7ca2b](https://github.com/apache/gravitino/pull/8062/commits/fc7ca2b48470a8203c2af62906977c3f1d48ead0)  has been test and the CI pipelines are as expected.

.github/workflows/build.yml
```diff
@@ -23,31 +23,39 @@ jobs:
...
-              - 'web/**'
-              - '**/*.java'
-              - '**/*.kts'
-              - '**/*.py'
-              - '**/*.rs'
-              - '**/resources/**'
-              - '**/src/**'
-              - '!**/*.md'
-              - '!**/*.png'
-              - '!**/*.svg'
```
yuqi1129, line 45: The prefix `!` does not take effect, and we should use the `excluded` keyword to exclude some files.

FANNG1: This is the reason why using `include` and `exclude` for all yml files?

.github/workflows/build.yml
```diff
@@ -98,6 +106,28 @@ jobs:
...
+
+      - uses: actions/setup-java@v4
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Build MCP Server
+        run: |
+          ./gradlew -PjdkVersion=${{ matrix.java-version }} -PpythonVersion=${{ matrix.python-version }} -PskipDockerTests=false :mcp-server:build
```
FANNG1: it's useless to pass `pythonVersion` and `java-version` for `:mcp-server:build`

.github/workflows/build.yml
yuqi1129: Yeah

FANNG1:
LGTM

.github/workflows/build.yml
FANNG1: do we need to exclude `mcp-server` for most yml files? since `mcp-server` is not in include list

.github/workflows/build.yml
```diff
@@ -98,6 +106,28 @@ jobs:
...
             spark-connector/**/*.log
 
+  mcp-server-build:
+    needs: changes
+    if: needs.changes.outputs.mcp_server_changes == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        java-version: [ 17 ]
```
FANNG1: could you remove this matrix too?

.github/workflows/build.yml
yuqi1129: I have tried, and it seems that as long as `mcp-server` is not in the `excluded` list, the CI will be triggered even though `mcp-server` is also not in the `include` list when there are changes under `mcp-server` folder.

.github/workflows/build.yml
yuqi1129: Yes, let me simplify it.

.github/workflows/build.yml
FANNG1: This may make the configuraiton complicated, Take `spark-integration-test` for example,  we should exclude not only `mcp-server` but also `python-clients` etcs.

yuqi1129: > do we need to exclude mcp-server for most yml files? since mcp-server is not in include list

The fact is: `'**/*.py'` may include most files in `mcp-server`, so we need to exclude `mcp-server/**`.

yuqi1129: Fixed

Issue #8064, opened 2025-08-13
[Subtask] simplify the definition of policy


mchades:
### Describe the subtask

remove `inheritable`, `exclusive` and `supported_object_types` from policy to simplify the policy usage.

### Parent issue

#7139

PR #8065, opened 2025-08-13
[#8064] improvement(policy): simplify the definition of policy

mchades:

- remove `inheritable`, `exclusive` from policy
- move `supported_object_types` from policy to content
- only validate the `supported_object_types` for built-in policy

Fix: #8064

no

tests updated


jerryshao:
For built-in policy, we should also validate other rules?

> only validate the supported_object_types for built-in policy

mchades:
> For built-in policy, we should also validate other rules?
> 
> > only validate the supported_object_types for built-in policy

Yes, when we add the built-in policy implementation, we should also implement the `validate()` method in the corresponding built-in content interface.

api/src/main/java/org/apache/gravitino/policy/PolicyOperations.java
```diff
@@ -74,38 +71,6 @@ Policy createPolicy(
       String name, String type, String comment, boolean enabled, PolicyContent content)
```
jerryshao, line 71: Shall we use enum type? I think for custom policy, the type should always be "CUSTOM", what do you think? cc @yuqi1129

core/src/main/java/org/apache/gravitino/storage/relational/service/PolicyMetaService.java
```diff
@@ -305,19 +305,6 @@ public List<PolicyEntity> associatePoliciesWithMetadataObject(
               ? Collections.emptyList()
               : getPolicyPOsByMetalakeAndNames(metalake, policyNamesToAdd);
```
jerryshao, line 306: I remembered that you have several checking logic in the storage layer, do you need to remove that logic?

api/src/main/java/org/apache/gravitino/policy/PolicyOperations.java
mchades: I have fixed the type value of the custom policy to "custom", but still use string as the field type for `policyType`, this is to minimize the impact on clients for future changes. You can review again.

core/src/main/java/org/apache/gravitino/storage/relational/service/PolicyMetaService.java
mchades: The relevant check logic has been removed, and other check logic is in another [PR](https://github.com/apache/gravitino/pull/7842/files#diff-1d6a9bb4111963dfdd2e5dcb2690acdad1d7e1ccecc2f48428f2af3f90dc54f5) that has not been merged yet.

core/src/test/java/org/apache/gravitino/meta/TestPolicyEntity.java
```diff
@@ -50,7 +50,7 @@ public void testPolicyEntityFields() {
             .withName("test")
             .withNamespace(namespace)
             .withComment("test comment")
-            .withPolicyType("my_compaction")
+            .withPolicyType("custom")
```
jerryshao: Why don't we make this field enum?

api/src/main/java/org/apache/gravitino/policy/Policy.java
```diff
@@ -103,8 +76,10 @@ public static BuiltInType fromPolicyType(String policyType) {
             String.format("Unknown built-in policy type: %s", policyType));
       }
```
jerryshao, line 77: For "custom", this logic is correct?

api/src/main/java/org/apache/gravitino/policy/PolicyOperations.java
jerryshao: OK.

core/src/main/java/org/apache/gravitino/meta/PolicyEntity.java
```diff
@@ -73,9 +64,6 @@ public static Builder builder() {
   private String policyType;
```
jerryshao: I think we'd use enum internally, to avoid unexpected type and force the type check on serde.

api/src/main/java/org/apache/gravitino/policy/Policy.java
mchades: It's correct. if the policy is custom, the method will be return in above L70:
```
 for (BuiltInType type : BuiltInType.values()) {
        if (type.policyType.equalsIgnoreCase(policyType)) {
          return type;
        }
      }
```

core/src/main/java/org/apache/gravitino/meta/PolicyEntity.java
mchades: ok, fixed. Please help to review again, thanks!

jerryshao (approved):

Issue #8063, opened 2025-08-13
[Subtask] Support read operations for Gravitino job in the MCP server


yuqi1129:
### Describe the subtask

Support all read operations of the Gravitino job.

### Parent issue

#7483

PR #8067, opened 2025-08-13
[#8063] feat(mcp-server): Support job related read operations for MCP server.

yuqi1129:

Support job related read operations for MCP server.

It's a key feature for mcp server.

Fix: #8063

N/A.

Test locally. 

image
image
image

image

### DIFF
--- a/mcp-server/mcp_server/client/plain/plain_rest_client_operation.py
+++ b/mcp-server/mcp_server/client/plain/plain_rest_client_operation.py
@@ -24,8 +24,12 @@ from mcp_server.client import (
     TableOperation,
     TagOperation,
 )
+from mcp_server.client.job_operation import JobOperation
 from mcp_server.client.plain.plain_rest_client_catalog_operation import (
     PlainRESTClientCatalogOperation,
+)
+from mcp_server.client.plain.plain_rest_client_job_operation import (
+    PlainRESTClientJobOperation,
 )
 from mcp_server.client.plain.plain_rest_client_schema_operation import (
     PlainRESTClientSchemaOperation,
@@ -60,5 +64,10 @@ class PlainRESTClientOperation(GravitinoOperation):
 
     def as_tag_operation(self) -> TagOperation:
         return PlainRESTClientTagOperation(
+            metalake_name=self.metalake_name, rest_client=self.rest_client
+        )
+
+    def as_job_operation(self) -> JobOperation:
+        return PlainRESTClientJobOperation(
             metalake_name=self.metalake_name, rest_client=self.rest_client
         )
--- a/mcp-server/mcp_server/tools/job.py
+++ b/mcp-server/mcp_server/tools/job.py
@@ -1,0 +1,146 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from fastmcp import Context, FastMCP
+
+
+def load_job_tool(mcp: FastMCP):
+    @mcp.tool(tags={"tag"})
+    async def get_list_of_jobs(
+        ctx: Context,
+        job_template_name: str = "",
+    ) -> str:
+        """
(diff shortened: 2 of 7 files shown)


mcp-server/tests/unit/tools/mock_operation.py
```diff
@@ -21,10 +21,11 @@
...
     TableOperation,
+    TagOperation,
     TopicOperation,
 )
 from mcp_server.client.fileset_operation import FilesetOperation
-from mcp_server.client.tag_operation import TagOperation
+from mcp_server.client.job_operation import JobOperation
 
 
 class MockOperation(GravitinoOperation):
```
FANNG1, line 31: Could you add the test for job operation?

mcp-server/mcp_server/client/plain/plain_rest_client_operation.py
```diff
@@ -24,12 +24,16 @@
     TableOperation,
     TagOperation,
 )
+from mcp_server.client.job_operation import JobOperation
```
FANNG1, line 27: How about moving `JobOperation` to the above xxxOperations?

mcp-server/mcp_server/tools/job.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Parameters:
+            ctx (Context): The context object containing Gravitino context.
+
+        Returns:
+            str: A JSON string representing an array of job template objects with the following structure:
+                [
+                  {
+                    "jobType": "shell",
+                    "name": "shell_test",
```
FANNG1, line 110: Could LLM understand that the `name` here is `job_template_name` to filter the jobs? maybe we could add some description for the name.

jerryshao:
We should also support run and cancel job, otherwise, it is not so useful.

yuqi1129:
> We should also support run and cancel job, otherwise, it is not so useful.

I see. It will modify the metadata(job) in Gravitino, If it's OK, I can change to support it. 

By the way, I am unsure which modification operation we should support or not. Could you help me clear it out?

mcp-server/mcp_server/tools/job.py
```diff
@@ -0,0 +1,146 @@
...
+# under the License.
+
+from fastmcp import Context, FastMCP
+
+
+def load_job_tool(mcp: FastMCP):
+    @mcp.tool(tags={"job"})
+    async def get_list_of_jobs(
+        ctx: Context,
+        job_template_name: str = "",
```
jerryshao: `job_template_name` can be `None`, not empty. `None` means list all the jobs, empty will lead to error.

mcp-server/mcp_server/client/plain/plain_rest_client_operation.py
yuqi1129: The current style check tool can't detect the problem,  I'm working on how to make it work automatically. Manually adjusting them is time-consuming.

mcp-server/tests/unit/tools/mock_operation.py
yuqi1129: I see.

jerryshao:
> > We should also support run and cancel job, otherwise, it is not so useful.
> 
> I see. It will modify the metadata(job) in Gravitino, If it's OK, I can change to support it.
> 
> By the way, I am unsure which modification operation we should support or not. Could you help me clear it out?

Think of as a user who wants to build a metadata-driven workflow by LLM, without this interface, how can they achieve this?

mcp-server/mcp_server/tools/job.py
yuqi1129: added.

mcp-server/mcp_server/client/plain/plain_rest_client_operation.py
yuqi1129: This https://github.com/apache/gravitino/pull/8094 PR is specialized to solve this problem, let's merge it first.

FANNG1: ok

mcp-server/mcp_server/tools/job.py
```diff
@@ -0,0 +1,253 @@
...
+                    "customFields": {},
+                    "audit": {
+                      "creator": "anonymous",
+                      "createTime": "2025-08-13T11:45:50.124843Z"
+                    },
+                    "scripts": []
+                  }
+                ]
+
+                ame: The name of the job template to get.
```
FANNG1: ame -> name

mcp-server/mcp_server/tools/job.py
```diff
@@ -0,0 +1,253 @@
...
+        job_template_name: str,
+        job_config: dict,
+    ) -> str:
+        """
+        Run a job using a specified job template and configuration.
+
+        Parameters:
+            ctx (Context): The context object containing Gravitino context.
+            job_template_name (str): The name of the job template to use for running the job.
+            job_config (dict): A dictionary containing the configuration for the job.
```
FANNG1, line 191: Could LLM submit a job with proper `job_config`?

yuqi1129: Yes. Please see the picture below:
image

jerryshao (approved):

Issue #8057, opened 2025-08-13
[Improvement] Incorrect logic in check in JdbcAuthorizationProperties.java


justinmclean:
### What would you like to be improved?

This line contains incorrect logic:
```
if (!properties.containsKey(key) && properties.get(key) != null) {
```

Here are some unit tests to show that:
```
  @Test
  void testJdbcProperties() {
    Map<String, String> properties = Maps.newHashMap();
    properties.put(JdbcAuthorizationProperties.JDBC_URL, "jdbc:test://localhost");
    properties.put(JdbcAuthorizationProperties.JDBC_USERNAME, "user");
    properties.put(JdbcAuthorizationProperties.JDBC_PASSWORD, "pwd");
    properties.put(JdbcAuthorizationProperties.JDBC_DRIVER, "driver");

    JdbcAuthorizationProperties jdbcProps = new JdbcAuthorizationProperties(properties);
    Assertions.assertDoesNotThrow(jdbcProps::validate);
  }

  @Test
  void testJdbcPropertiesMissingUrl() {
    Map<String, String> properties = Maps.newHashMap();
    properties.put(JdbcAuthorizationProperties.JDBC_USERNAME, "user");
    properties.put(JdbcAuthorizationProperties.JDBC_PASSWORD, "pwd");
    properties.put(JdbcAuthorizationProperties.JDBC_DRIVER, "driver");

    JdbcAuthorizationProperties jdbcProps = new JdbcAuthorizationProperties(properties);
    Assertions.assertThrows(IllegalArgumentException.class, jdbcProps::validate);
  }

  @Test
  void testJdbcPropertiesNullUsername() {
    Map<String, String> properties = Maps.newHashMap();
    properties.put(JdbcAuthorizationProperties.JDBC_URL, "jdbc:test://localhost");
    properties.put(JdbcAuthorizationProperties.JDBC_USERNAME, null);
    properties.put(JdbcAuthorizationProperties.JDBC_PASSWORD, "pwd");
    properties.put(JdbcAuthorizationProperties.JDBC_DRIVER, "driver");

    JdbcAuthorizationProperties jdbcProps = new JdbcAuthorizationProperties(properties);
    Assertions.assertThrows(IllegalArgumentException.class, jdbcProps::validate);
  }
```

### How should we improve?

Fix the logic.

ankamde:
@justinmclean Please assign this issue to me.

justinmclean:
Done, all yours.

PR #8070, opened 2025-08-13
[#8057] Improvement: Incorrect logic in check in JdbcAuthorizationProperties.java

ankamde:

Incorrect logic in check in JdbcAuthorizationProperties.java

Because JdbcAuthorizationProperties didn't check for missing properties.

Fix: #8057

No.

By running unit tests.


ankamde:
@justinmclean I modified original test - we can't put null in properties because AuthorizationProperties.java constructor throws NPE. I added unit test for this. Please have a look.

authorizations/authorization-common/src/main/java/org/apache/gravitino/authorization/jdbc/JdbcAuthorizationProperties.java
```diff
@@ -34,7 +34,7 @@ public JdbcAuthorizationProperties(Map<String, String> properties) {
   }
 
   private void check(String key, String errorMsg) {
-    if (!properties.containsKey(key) && properties.get(key) != null) {
+    if (!properties.containsKey(key)) {
```
justinmclean, line 37: This is not correct what happens if the property is null?

authorizations/authorization-common/src/test/java/org/apache/gravitino/authorization/jdbc/TestJdbcAuthorizationProperties.java
```diff
@@ -0,0 +1,61 @@
...
+  @Test
+  void testJdbcPropertiesNullUsername() {
+    Map<String, String> properties = Maps.newHashMap();
+    properties.put(JdbcAuthorizationProperties.JDBC_URL, "jdbc:test://localhost");
+    properties.put(JdbcAuthorizationProperties.JDBC_USERNAME, null);
+    properties.put(JdbcAuthorizationProperties.JDBC_PASSWORD, "pwd");
+    properties.put(JdbcAuthorizationProperties.JDBC_DRIVER, "driver");
+
+    Assertions.assertThrows(
+        NullPointerException.class, () -> new JdbcAuthorizationProperties(properties));
```
justinmclean: You changed the unit test here - please keep the tests how they were originally specified. We should not be throwing NullPointerException.

justinmclean: And while I see why you did this, the code still shouldn't throw a NPE.

authorizations/authorization-common/src/main/java/org/apache/gravitino/authorization/jdbc/JdbcAuthorizationProperties.java
justinmclean: in case like this is best you make a little chart 2x2 with the key existing and not existing vs the property being null or not, you can work out the correct logic for from that.

ankamde: In the latest version I'm filtering out null values in AuthorizationProperties - this way there is no need to check for null in check method.

ankamde:
@justinmclean Please have a look.

justinmclean (approved):
Thanks for working on this.

Issue #8037, opened 2025-08-12
[Improvement] Add support for nested field or throw an exception in PartitionUtils.java


justinmclean:
### What would you like to be improved?

In common/src/main/java/org/apache/gravitino/dto/rel/PartitionUtils.java validateFieldExistence only supports a single field name and doesn't support nested fields. Note the code only looks at fieldName[0].

### How should we improve?

Either add support for nested field or throw an exception saying this is not supported.

pandeysambhi:
@justinmclean I had the same thought while I was working on #7975 . I assumed this might have been implemented the way it is to support checking multiple fields at once. Whatever implementation of the validateFieldExistence I checked, the fieldName argument was being sent as a String array. Can you guide a little on how you want this to be improved, as if we go by throwing an exception, we would have to change the usage everywhere else as well. We can go for the nested field check as a simpler way, but I would want to understand the implications of the same. Let me know your thoughts, and I can start working on this.

justinmclean:
A simple fix would be to just throw an exception if a nested field is passed, as the code doesn't handle that case and users should be aware of it. A more complete solution would be to have the code handle it.

PaulOberg1:
Hi @justinmclean , I'm interested in contributing to this :). Could I be assigned?

justinmclean:
Hi @PaulOberg1 thanks for offering to help, but @pandeysambhi is working on this. Could you take another issue?

PR #8075, opened 2025-08-13
[#8037] improvement(validateFieldExistence): throw exception if nested fieldName in validateFieldExistence

pandeysambhi:
If the fieldName is nested, then validateFieldExistence will throw an IllegalArgumentException in PartitionUtils

As the support for nested fieldName check is not supported in validateFieldExistence function in PartitionUtils, If the fieldName is nested, then validateFieldExistence will throw an IllegalArgumentException 

Fix: #8037 

No

Unit testing


justinmclean (approved):
Thanks for fixing this

Issue #8054, opened 2025-08-13
[Improvement] Old column logic is incorrect is TableColumnMetaService.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/storage/relational/service/TableColumnMetaService.java in updateColumnPOsFromTableDiff there is an issue with.

```
      // If the column is not existed in old columns, or if the column is updated, mark it as UPDATE
      if (oldColumn == null) {
        columnPOsToInsert.add(
            POConverters.initializeColumnPO(newTablePO, newColumn, ColumnPO.ColumnOpType.CREATE));
      } else if (!oldColumn.equals(newColumn)) {
        columnPOsToInsert.add(
            POConverters.initializeColumnPO(newTablePO, newColumn, ColumnPO.ColumnOpType.UPDATE));
      }
```

### How should we improve?

The if statement should also check if the old column is not equal to the new column.

TamimEhsan:
@justinmclean can i take this one?

justinmclean:
It's all yours.

PR #5078, opened 2024-10-09
[#4867] feat(core): Add storage support for columns in Gravitino

jerryshao:
### What changes were proposed in this pull request?

This PR aims to add storage support for columns in Gravitino.

### Why are the changes needed?

With this, we can also support managing columns in Gravitno, like set tags, do column level privileges.

Fix: #4867 

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

UTs added.

PR #8079, opened 2025-08-14
[#8054] fix(TableColumnMetaService): Correct column update logic to handle new columns

TamimEhsan:

Fix the coumn update logic in `updateColumnPOsFromTableDiff` of `TableMetaService`

The previous version didn't handle CREATE column. Instead, used to handle both the CREATE and UPDATE as UPDATE

Fix: #8054 

No

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/TableColumnMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/TableColumnMetaService.java
@@ -144,7 +144,10 @@ public class TableColumnMetaService {
     for (ColumnEntity newColumn : newColumns.values()) {
       ColumnEntity oldColumn = oldColumns.get(newColumn.id());
       // If the column is not existed in old columns, or if the column is updated, mark it as UPDATE
-      if (oldColumn == null || !oldColumn.equals(newColumn)) {
+      if (oldColumn == null) {
+        columnPOsToInsert.add(
+            POConverters.initializeColumnPO(newTablePO, newColumn, ColumnPO.ColumnOpType.CREATE));
+      } else if (!oldColumn.equals(newColumn)) {
         columnPOsToInsert.add(
             POConverters.initializeColumnPO(newTablePO, newColumn, ColumnPO.ColumnOpType.UPDATE));
       }


justinmclean (dismissed):
Thanks for fixing that.

TamimEhsan:
@justinmclean should I add any tests for the change?

jerryshao:
How do you make sure the logic is correct without a test? If there's no bug, I'm against modifying the code with no reason. It might be by design (I forgot the reason).

jerryshao:
Modification in the core part will introduce unknown issues if you don't know the history. Not to say you don't have a test to cover the changes.

TamimEhsan:
@jerryshao I have written some tests for the changes. But yes, it only makes sure the requested fix works provided that the updated logics are correct.  Will push that if this PR moves forward.

justinmclean:
@tamimEhan Please add your tests. However, I do think the logic is rather apparent, i.e. if the column didn't exist before, it's a creation, if it did exist before, it's an update. If this is by design, for whatever reason, then there should be a comment explaining why it was done in a way that looks like an error.

jerryshao:
There are lot of hidden things in every project that cannot be figured out by a static code scanner. If there's a bug, we should fix it. If not, we'd better leave it as it is. Because no one remembers why it should be like this, especially those that are not apparent as you saw.

justinmclean:
I didn't use a static code scanner to find this. I looked at the code, and I believe there is a bug.

core/src/main/java/org/apache/gravitino/storage/relational/service/TableColumnMetaService.java
```diff
@@ -143,8 +143,12 @@ void updateColumnPOsFromTableDiff(
     List<ColumnPO> columnPOsToInsert = Lists.newArrayList();
     for (ColumnEntity newColumn : newColumns.values()) {
       ColumnEntity oldColumn = oldColumns.get(newColumn.id());
-      // If the column is not existed in old columns, or if the column is updated, mark it as UPDATE
-      if (oldColumn == null || !oldColumn.equals(newColumn)) {
```
yuqi1129, line 147: This depends on the real definition of `ColumnPO.ColumnOpType.CREATE` and `ColumnPO.ColumnOpType.UPDATE`.
Based on previous code, changes introduced by `JDBCBackend#update` will be marked as `UPDATE`, and `JDBCBackend#insert` will be viewed as `CREATE` operation.

TamimEhsan:
found this conversation from the original commit https://github.com/apache/gravitino/pull/5078/files#r1794721831

jerryshao:
Thanks @TamimEhsan for working on this. I'm going to close this PR as it is by design. You can work on other bugs or issues if you'd like to. Greatly appreciated!

Issue #8055, opened 2025-08-13
[Improvement] Prevented a potential NullPointerException in CompatibilityUtils.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/audit/v2/CompatibilityUtils.java a NPR could occur if operationStatus is null.

Add this line to testOperationStatus to show the issue:
```
Assertions.assertEquals(Status.UNKNOWN, CompatibilityUtils.toAuditLogStatus(null));
```

### How should we improve?

Use == instead of calling equals when comparing the enum values.

keepConcentration:
@justinmclean 

Hi, Could you please assign this issue to me?

justinmclean:
All yours.

PR #8080, opened 2025-08-14
[#8055] improvement(core): Replace equals() with == for enum comparison in CompatibilityUtils

keepConcentration:

Replace `equals()` method with `==` operator for enum comparison in the `toAuditLogStatus()` method within `CompatibilityUtils` class. Additionally, add a null safety test case to ensure proper handling of null `OperationStatus` parameters.  

1. Using `==` for enum comparison is more efficient than `equals()` since enums are singletons in Java
2. Added test coverage for null parameter handling to ensure robustness

Fix: #8055

No, this is purely an internal refactoring change. 

1. Existing unit tests continue to pass, ensuring no functional regression
2. Added a new test case `testOperationStatusNullHandling()` to verify proper null parameter handling


justinmclean (approved):
Thanks for fixing that.

Issue #8034, opened 2025-08-12
[Improvement] Fully suport nallable simpleStrings in Types.java


justinmclean:
### What would you like to be improved?

Improve api/src/main/java/org/apache/gravitino/rel/types/Types.java to fully suport nallable simpleStrings.

Some test code:
```
    Types.MapType mapTypeNotNull =
        Types.MapType.valueNotNull(Types.IntegerType.get(), Types.StringType.get());
    Assertions.assertEquals(Type.Name.MAP, mapTypeNotNull.name());
    Assertions.assertEquals(Types.IntegerType.get(), mapTypeNotNull.keyType());
    Assertions.assertEquals(Types.StringType.get(), mapTypeNotNull.valueType());
    Assertions.assertEquals("map<integer,string, NOT NULL>", mapTypeNotNull.simpleString());
    Assertions.assertEquals(
        mapTypeNotNull,
        Types.MapType.valueNotNull(Types.IntegerType.get(), Types.StringType.get()));
```

### How should we improve?

Change the simpleString method so that the returned string reflects whether the map’s values are nullable or not, appending NOT NULL when appropriate

TheYorouzoya:
@justinmclean Considering I've previously worked on a "good first issue", would it be appropriate for me to take this? If not, would you have a recommendation for an issue that I could work on?

justinmclean:
You are welcome to work on this. It's not a simple fix, but you can do it.

justinmclean:
If you get stuck or don't know what to do, just ask for help.

Sven0x14:
I've been thinking about this for a while. Aren't the values of a map always nullable, since they are always reference types? I think I've misunderstood something here.

justinmclean:
SimpleStrings can be nullable or not nullable, it's a property they have vs them having an actual null value - does that help?

PR #8092, opened 2025-08-14
[#8034] Improvement: Fully support nullable simpleStrings in Types.java

TheYorouzoya:

This PR adds full support for nullable simpleStrings in `api/src/main/java/org/apache/gravitino/rel/types/Types.java`. More specifically, for the `simpleString()` method for the complex type `MapType`.
Additionally, a couple of unit tests are added to cover the cases where `MapType` and `ListType` can have non-nullable fields in them.

It is important to indicate (in the human readable `simpleString()` output) whether an instantiated complex type like the `MapType` can or cannot have null values for its elements. In code, this is indicated by a `valueNullable` boolean field.
For example, a `MapType` which maps each `IntegerType` to a `StringType` might not want to allow an `IntegerType->Null` mapping. In that case, a call to the `simpleString()` method should return the following string `map<integer,string, NOT NULL>` to indicate this behavior.

Fix: #8034

No.

Locally, with the provided unit test in the original issue post. Additionally, since `ListType` also allows for a similar non-null behavior as indicated by the code, I've decided to add a test to cover it as well.
Since the other two complex types either do not exhibit the behavior (in case of `UnionType`) or already have the behavior covered in a test (in case of `StructType`), no additional tests were added for the other complex types.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/rel/types/Types.java
+++ b/api/src/main/java/org/apache/gravitino/rel/types/Types.java
@@ -1096,7 +1096,9 @@ public class Types {
 
     @Override
     public String simpleString() {
-      return "map<" + keyType.simpleString() + "," + valueType.simpleString() + ">";
+      return valueNullable
+          ? "map<" + keyType.simpleString() + "," + valueType.simpleString() + ">"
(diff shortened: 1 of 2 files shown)


TheYorouzoya:
@justinmclean please review

justinmclean (approved):
Thanks for working on this.

Issue #8082, opened 2025-08-14
[Improvement] In PartitionOperations.java missing call to request.validate()


justinmclean:
### What would you like to be improved?

In server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java addPartitions is misisng request.validate() 

Here's a test to show that:
```
  @Test
  public void testAddPartitionWithNullPartitions() {
    // Missing partitions should trigger request validation failure and return BAD_REQUEST.
    AddPartitionsRequest req = new AddPartitionsRequest(null);
    Response resp =
        target(partitionPath(metalake, catalog, schema, table))
            .request(MediaType.APPLICATION_JSON_TYPE)
            .accept("application/vnd.gravitino.v1+json")
            .post(Entity.entity(req, MediaType.APPLICATION_JSON_TYPE));

    Assertions.assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), resp.getStatus());

    ErrorResponse errorResp = resp.readEntity(ErrorResponse.class);
    Assertions.assertEquals(ErrorConstants.ILLEGAL_ARGUMENTS_CODE, errorResp.getCode());
  }
```

### How should we improve?

add it.

justinmclean:
It might be good to see if any other calls that need it have also forgotten to call the request.validate()

JeonDaehong:
I would like to contribute to this as well. I would appreciate it if you could assign it to me.

PR #8098, opened 2025-08-14
[#8082] fix: call request.validate() in PartitionOperations.java

JeonDaehong:

In `PartitionOperations.java`, the `addPartitions` method was missing a call to `request.validate()`.  
This PR adds `request.validate()` to ensure that invalid requests are rejected properly.

Without this validation, requests with missing or null partitions could bypass checks, potentially causing unexpected errors.  
Adding this call ensures that invalid requests return `BAD_REQUEST` as expected.

Fix: #8082(https://github.com/apache/gravitino/issues/8082)

No. Existing APIs behave the same for valid requests. Invalid requests now correctly return `BAD_REQUEST`.

- Added a unit test `testAddPartitionWithNullPartitions()` that verifies a `BAD_REQUEST` response for invalid input.  
- Verified locally and via CI that the test passes and validation works as expected.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
@@ -173,6 +173,8 @@ public class PartitionOperations {
     Preconditions.checkArgument(
         request.getPartitions().length == 1, "Only one partition is supported");
 
+    request.validate();
+
     try {
       return Utils.doAs(
           httpRequest,


JeonDaehong:
Hi @justinmclean 

I’m a first-time contributor to Apache Gravitino.

I wanted to check if submitting my changes via a PR like this is okay. I’d really appreciate any feedback you could provide.

If this approach works, I’ll also submit PRs for the other issues I’ve been assigned and resolved.

Thank you!

JeonDaehong:
Hi @justinmclean

It seems like there might be an issue in the CI/CD process.
Is there any action I need to take on my end?

justinmclean:
I'll look into the issue for you, I don't think it's related to the changes you made.

JeonDaehong:
> I'll look into the issue for you, I don't think it's related to the changes you made.

Thank you. Please let me know once it’s confirmed!

justinmclean (approved):
Thanks for fixing this.

JeonDaehong:
> Thanks for fixing this.

Thank you for the approval.
I’ll work on another issue as well and submit a PR.

Issue #8101, opened 2025-08-14
[Improvement] remove AuthorizationMetadataPrivileges


hdygxsj:
### What would you like to be improved?

AuthorizationMetadataPrivileges is useless

PR #8104, opened 2025-08-14
[#8101]Improvement(authz): Remove AuthorizationMetadataPrivileges

hdygxsj:

Remove AuthorizationMetadataPrivileges

Fix: #8101

None

Existing IT and UT


server-common/src/main/java/org/apache/gravitino/server/authorization/annotations/AuthorizationMetadataPrivileges.java
xunliu, line 1: Is this interface no longer in use?

hdygxsj: Yes, he has never been used before.

Issue #7947, opened 2025-08-06
[Bug report] UI: if authenticators is oauth, github repo api authorized failed


danhuawang:
### Version

main branch

### Describe what's wrong

<img width="1912" height="1033" alt="Image" src="https://github.com/user-attachments/assets/17bcfbe4-9fed-4960-8d1b-9bb71b9a005a" />

### Error message and/or stacktrace

```
{
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest",
  "status": "401"
}

```

### How to reproduce

1. Set authenticators to oauth
```
gravitino.authorization.enable = true
gravitino.authenticators = oauth
```
2. Go to the web page ,create catalog ,schema,fileset. Check the apache repo stars number

xunliu:
hi @danhuawang this bug is not related to Cache and metadata Authz.

bharos:
>Go to the web page ,create catalog ,schema,fileset. Check the apache repo stars number

@danhuawang are you saying that it shows the repo & stars number after login, and only after when you do create catalog ,schema,fileset  it stops showing ? Or is it always showing you 0 ?

bharos:
@LauraXia123  @jerryshao LMK if I can work on it, as I have recently made oauth related code changes related to this

bharos:
I think we should skip Github calls from OAuth entirely, because it's an external API and we don't need to attach the token for these API calls.

jerryshao:
OK, sure.

PR #8106, opened 2025-08-14
[#7947] fix(web): Prevent OAuth headers from being sent to GitHub API causing 401 errors

bharos:

Since Github APIs are external and public APIs, we don't need to attach OAuth headers for those calls.
Current code also prevents calling this API before login, which is unnecessary. Users don't need to be logged in to access this API

Fix: #7947

No
Tested on localhost. No headers attached:
Screenshot 2025-08-14 at 3 57 51 PM


roryqi:
@LauraXia123 cc

Issue #8087, opened 2025-08-14
[Improvement] Fix OAuth2 token provider to automatically fetch a new token


justinmclean:
### What would you like to be improved?

Fix OAuth2 token provider to automatically fetch a new token when the current one is missing or expired, avoiding null returns and authentication failures.

Here's a unit test that will help:
```
@Test
  public void testFetchTokenWhenNull() throws Exception {
    HTTPClient client = mock(HTTPClient.class);
    OAuth2TokenResponse response = new OAuth2TokenResponse("new_token", null, "bearer", 1, null, null);
    when(client.postForm(anyString(), anyMap(), any(), anyMap(), any())).thenReturn(response);

    DefaultOAuth2TokenProvider provider = newProvider(client);

    String token = provider.getAccessToken();
    Assertions.assertEquals("new_token", token);
    verify(client).postForm(anyString(), anyMap(), any(), anyMap(), any());
  }

  private DefaultOAuth2TokenProvider newProvider(HTTPClient client) throws Exception {
    Constructor<DefaultOAuth2TokenProvider> ctor =
        DefaultOAuth2TokenProvider.class.getDeclaredConstructor();
    ctor.setAccessible(true);
    DefaultOAuth2TokenProvider provider = ctor.newInstance();

    Field clientField = OAuth2TokenProvider.class.getDeclaredField("client");
    clientField.setAccessible(true);
    clientField.set(provider, client);

    Field credentialField = DefaultOAuth2TokenProvider.class.getDeclaredField("credential");
    credentialField.setAccessible(true);
    credentialField.set(provider, "cred");

    Field scopeField = DefaultOAuth2TokenProvider.class.getDeclaredField("scope");
    scopeField.setAccessible(true);
    scopeField.set(provider, "scope");

    Field pathField = DefaultOAuth2TokenProvider.class.getDeclaredField("path");
    pathField.setAccessible(true);
    pathField.set(provider, "/token");

    Field tokenField = DefaultOAuth2TokenProvider.class.getDeclaredField("token");
    tokenField.setAccessible(true);
    tokenField.set(provider, null);
    return provider;
  }
```

justinmclean:
Basically, if expires == null or expires <= System.currentTimeMillis() return a new token, if expires > System.currentTimeMillis() return existing token (although you may not need to do this explicitly).

pandeysambhi:
@justinmclean would like to work on this

justinmclean:
All yours

PR #8116, opened 2025-08-15
[#8087] improvement(Oauth2): OAuth2 token provider to automatically fetch a new token

pandeysambhi:
Oauth2 provider automatically fetches a new token when the old one is expired

Improve the Oauth2 token provider to fetch auth token automatically

Fix: #8087 

Unit testing

### DIFF
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/DefaultOAuth2TokenProvider.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/DefaultOAuth2TokenProvider.java
@@ -35,21 +35,12 @@ public class DefaultOAuth2TokenProvider extends OAuth2TokenProvider {
   @Override
   protected String getAccessToken() {
     synchronized (this) {
-      if (token == null) {
-        token =
-            OAuth2ClientUtil.fetchToken(client, Collections.emptyMap(), credential, scope, path)
-                .getAccessToken();
-        return token;
-      }
-
       Long expires = OAuth2ClientUtil.expiresAtMillis(token);
       if (expires == null || expires <= System.currentTimeMillis()) {
         token =
             OAuth2ClientUtil.fetchToken(client, Collections.emptyMap(), credential, scope, path)
                 .getAccessToken();
-        return token;
       }
-
       return token;
     }
   }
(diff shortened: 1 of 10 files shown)


justinmclean:
This PR contains some unnecessary changes. Please refrain from changing the formatting of existing code or adding unnecessary comments, as this is likely to fail Spotless formatting.

justinmclean:
Yes as the code is no longer correctly formatted, the CI is failing, you'll need to run `./gradlew :clients:client-java:spotlessApply` and push the changes to fix.

pandeysambhi:
> Yes as the code is no longer correctly formatted, the CI is failing, you'll need to run `./gradlew :clients:client-java:spotlessApply` and push the changes to fix.

on it

pandeysambhi:
@justinmclean cleaned up the PR, let me know if looks good

clients/client-java/src/main/java/org/apache/gravitino/client/DefaultOAuth2TokenProvider.java
```diff
@@ -35,18 +35,21 @@ private DefaultOAuth2TokenProvider() {}
...
+        token =
+            OAuth2ClientUtil.fetchToken(client, Collections.emptyMap(), credential, scope, path)
+                .getAccessToken();
         return token;
       }
 
-      token =
-          OAuth2ClientUtil.fetchToken(client, Collections.emptyMap(), credential, scope, path)
-              .getAccessToken();
       return token;
```
justinmclean, line 44: You really only need one "return token;" at the end.

clients/client-java/src/main/java/org/apache/gravitino/client/DefaultOAuth2TokenProvider.java
```diff
@@ -35,18 +35,21 @@ private DefaultOAuth2TokenProvider() {}
   @Override
   protected String getAccessToken() {
     synchronized (this) {
-      Long expires = OAuth2ClientUtil.expiresAtMillis(token);
-      if (expires == null) {
-        return null;
+      if (token == null) {
```
roryqi: The token can't be null. Because we initialize the token in the build method.

clients/client-java/src/main/java/org/apache/gravitino/client/DefaultOAuth2TokenProvider.java
pandeysambhi: As I had multiple ifs, so had the return statements to return as soon as any condition was fulfilled, after incorporating @jerqi 's suggestions, I updated to have only one.

clients/client-java/src/main/java/org/apache/gravitino/client/DefaultOAuth2TokenProvider.java
pandeysambhi: @jerqi Thanks for this, missed this piece. made the changes

Issue #8113, opened 2025-08-15
[Improvement] In AccessControlHookDispatcher.java check result of deleteRole before notifying of change


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java, you should only notify the user when a role deletion succeeds.

### How should we improve?

only notifyRoleUserRelChange if resultOfDeleteRole is true

JeonDaehong:
Please assign it to me. Thank you.

pandeysambhi:
@justinmclean please assign this to me.

justinmclean:
I think it might be best if you fix up the other PR's first.

Surya00011:
Hi @justinmclean, this issue still seems to be open. Can I take this?

PR #8120, opened 2025-08-15
[#8113] Improvement to only notify the user when a role deletion succeeds

Raj-G07:

Only` notify the user when a role deletion succeeds.

To prevent errors and only notify the user when a role deletion succeeds.

Fixes: #8113

No.

CI passed

### DIFF
--- a/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
@@ -192,8 +192,8 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
       log.debug(e.getMessage());
     }
     boolean resultOfDeleteRole = dispatcher.deleteRole(metalake, role);
-    if (oldRole != null) {
-      notifyRoleUserRelChange(((RoleEntity) oldRole).id());
+    if (resultOfDeleteRole && oldRole != null) {
+      notifyRoleUserRelChange(metalake, role);
     }
     return resultOfDeleteRole;
   }
@@ -214,7 +214,6 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
       String metalake, String role, MetadataObject object, Set<Privilege> privileges)
       throws NoSuchMetalakeException, NoSuchRoleException {
     Role grantedRole = dispatcher.grantPrivilegeToRole(metalake, role, object, privileges);
-    notifyRoleUserRelChange(metalake, role);
     return grantedRole;
   }
 
@@ -223,7 +222,6 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
       String metalake, String role, MetadataObject object, Set<Privilege> privileges)
       throws NoSuchMetalakeException, NoSuchRoleException {
     Role revokedRole = dispatcher.revokePrivilegesFromRole(metalake, role, object, privileges);
-    notifyRoleUserRelChange(metalake, role);
     return revokedRole;
   }
 
--- a/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
@@ -191,13 +191,10 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
     } catch (NoSuchRoleException e) {
       log.debug(e.getMessage());
     }
-    
     boolean resultOfDeleteRole = dispatcher.deleteRole(metalake, role);
-
     if (resultOfDeleteRole && oldRole != null) {
       notifyRoleUserRelChange(metalake, role);
     }
-
     return resultOfDeleteRole;
   }
 
--- a/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
@@ -143,6 +143,7 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
   public User revokeRolesFromUser(String metalake, List<String> roles, String user)
       throws NoSuchUserException, IllegalRoleException, NoSuchMetalakeException {
     User revokedUser = dispatcher.revokeRolesFromUser(metalake, roles, user);
+    notifyRoleUserRelChange(metalake, roles);
     return revokedUser;
   }
 
@@ -216,6 +217,7 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
       String metalake, String role, MetadataObject object, Set<Privilege> privileges)
       throws NoSuchMetalakeException, NoSuchRoleException {
     Role grantedRole = dispatcher.grantPrivilegeToRole(metalake, role, object, privileges);
+    notifyRoleUserRelChange(metalake, role);
     return grantedRole;
   }
 
@@ -224,6 +226,7 @@ public class AccessControlHookDispatcher implements AccessControlDispatcher {
       String metalake, String role, MetadataObject object, Set<Privilege> privileges)
       throws NoSuchMetalakeException, NoSuchRoleException {
     Role revokedRole = dispatcher.revokePrivilegesFromRole(metalake, role, object, privileges);
+    notifyRoleUserRelChange(metalake, role);
     return revokedRole;
   }
 
--- a/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
@@ -44,7 +44,6 @@ import org.apache.gravitino.exceptions.NoSuchRoleException;
 import org.apache.gravitino.exceptions.NoSuchUserException;
 import org.apache.gravitino.exceptions.RoleAlreadyExistsException;
 import org.apache.gravitino.exceptions.UserAlreadyExistsException;
-import org.apache.gravitino.meta.RoleEntity;
 import org.apache.gravitino.utils.NameIdentifierUtil;
 import org.apache.gravitino.utils.PrincipalUtils;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -119,7 +119,7 @@ public String[] listGroupNames(String metalake) throws NoSuchMetalakeException {
   public User grantRolesToUser(String metalake, List<String> roles, String user)
       throws NoSuchUserException, IllegalRoleException, NoSuchMetalakeException {
     User grantedUser = dispatcher.grantRolesToUser(metalake, roles, user);
-    notifyRoleUserRelChange(metalake, roles);
+
```
justinmclean: I think this needs to stay in

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -192,9 +191,10 @@ public boolean deleteRole(String metalake, String role) throws NoSuchMetalakeExc
       log.debug(e.getMessage());
     }
     boolean resultOfDeleteRole = dispatcher.deleteRole(metalake, role);
-    if (oldRole != null) {
-      notifyRoleUserRelChange(((RoleEntity) oldRole).id());
-    }
+
+    if(resultOfDeleteRole && oldRole!= null) 
+      notifyRoleUserRelChange(metalake,role)
+
```
justinmclean: The CI is failing here because we use Spotless to make sure the code is consistently formatted. You can run Spotless yourself to fix this or manually add the brackets back in.

Raj-G07:
@justinmclean Now it's ok!

justinmclean:
Sorry there is still a Stoppless issue here you need to run `./gradlew :core:spotlessApply` and commit the changes.

Raj-G07:
@justinmclean Run the command from the project root `./gradlew :core:spotlessApply`, now it's fine

roryqi:
This is not a fix, just an improvement.

Raj-G07:
@justinmclean  What am i missing?

justinmclean:
@Raj-G07, please look at the output of the CI. It shows you exactly what is incorrect, but as I said above, it's best if you run Spotless.
```
          @@ -190,11 +189,11 @@
           ····}·catch·(NoSuchRoleException·e)·{
           ······log.debug(e.getMessage());
           ····}
          -····
          +
           ····boolean·resultOfDeleteRole·=·dispatcher.deleteRole(metalake,·role);
           
          -····if(resultOfDeleteRole·&&·oldRole!=·null)·{
          -······notifyRoleUserRelChange(metalake,role);
          +····if·(resultOfDeleteRole·&&·oldRole·!=·null)·{
          +······notifyRoleUserRelChange(metalake,·role);
           ····}
           
           ····return·resultOfDeleteRole;
```

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -223,7 +224,6 @@ public Role revokePrivilegesFromRole(
       String metalake, String role, MetadataObject object, Set<Privilege> privileges)
       throws NoSuchMetalakeException, NoSuchRoleException {
     Role revokedRole = dispatcher.revokePrivilegesFromRole(metalake, role, object, privileges);
-    notifyRoleUserRelChange(metalake, role);
```
roryqi: We shouldn't delete this line.

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -214,7 +216,6 @@ public Role grantPrivilegeToRole(
       String metalake, String role, MetadataObject object, Set<Privilege> privileges)
       throws NoSuchMetalakeException, NoSuchRoleException {
     Role grantedRole = dispatcher.grantPrivilegeToRole(metalake, role, object, privileges);
-    notifyRoleUserRelChange(metalake, role);
```
roryqi: We shouldn't delete this line.

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -143,7 +143,6 @@ public Group revokeRolesFromGroup(String metalake, List<String> roles, String gr
   public User revokeRolesFromUser(String metalake, List<String> roles, String user)
       throws NoSuchUserException, IllegalRoleException, NoSuchMetalakeException {
     User revokedUser = dispatcher.revokeRolesFromUser(metalake, roles, user);
-    notifyRoleUserRelChange(metalake, roles);
```
roryqi: We shouldn't delete this line.

Raj-G07:
@jerqi Take a look on this!

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
justinmclean: I'm also not sure why you have changed the parameters

justinmclean:
@Raj-G07 Please review your changes again.

Raj-G07:
@justinmclean  I want to know dispatcher.deleteRole() delete entity.id() , if the answer is yes, then we have to undo the changes otherwise not , what i believe!

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -191,10 +191,13 @@ public boolean deleteRole(String metalake, String role) throws NoSuchMetalakeExc
     } catch (NoSuchRoleException e) {
       log.debug(e.getMessage());
     }
+    
     boolean resultOfDeleteRole = dispatcher.deleteRole(metalake, role);
-    if (oldRole != null) {
-      notifyRoleUserRelChange(((RoleEntity) oldRole).id());
+
+    if(resultOfDeleteRole && oldRole!= null) {
```
roryqi: `role == null` means that `deleted` will be false. It' a repeated judgement.

justinmclean:
Formatting is still an issue:
  ```
        @@ -191,11 +190,11 @@
           ····}·catch·(NoSuchRoleException·e)·{
           ······log.debug(e.getMessage());
           ····}
          -····
          +
           ····boolean·resultOfDeleteRole·=·dispatcher.deleteRole(metalake,·role);
           
          -····if(resultOfDeleteRole·&&·oldRole!=·null)·{
          -······notifyRoleUserRelChange(metalake,role);
          +····if·(resultOfDeleteRole·&&·oldRole·!=·null)·{
          +······notifyRoleUserRelChange(metalake,·role);
           ····}
           
           ····return·resultOfDeleteRole;
```

You need to run './gradlew :core:spotlessApply' to fix these issues.

**Comment by Raj-G07** [comment]
@justinmclean  What i am doing really , i made the changes , run `./gradlew :core:spotlessApply` , and then `git add .` , `git commit -m diofjoasifjwaeo` , `git push -u origin branch_name` , this is whole process.

**Comment by justinmclean** [comment]
Those git commands look right to me, but the code is not correctly formatted. Notice the four spaces on the blank line and the misisng space between metalake and role. How does that compare to your local code?

**Comment by Raj-G07** [comment]
@justinmclean I do it manually, considering one dot = one space , may be works this time!

**Comment by justinmclean** [comment]
You still have four spaces instead of a blank line

**Comment by Raj-G07** [comment]
@justinmclean Remove unnecessary spaces

**Comment by Raj-G07** [comment]
@justinmclean Remove RoleEntity because this can never be used.

**Comment by Raj-G07** [comment]
@justinmclean We are now so close to merge , what things i am missing?

**Comment by roryqi** [comment]
I don't think that we need this PR. Origin logic already covered the cache which delete operation is false.

**Comment by justinmclean** [comment]
I don't understand what you mean by that. Can you point me to the code?

**Comment by roryqi** [comment]
> I don't understand what you mean by that. Can you point me to the code?

This code.

```
if (oldRole != null) {
```

justinmclean:
I don't see how that applies as deleteRole could fail for other reasons other than the role being null.

roryqi:
> I don't see how that applies as deleteRole could fail for other reasons other than the role being null.

This is our system logic. If the entity doesn't exist, we return false when deleting. Other errors will throw exception.

justinmclean:
Not existing and being null are different, if you return false then you shouldn't call notifyRoleUserRelChange as nothing has changed.

justinmclean:
The documentation states, "True if the Role was successfully deleted, false only when there's no such role, otherwise it will throw an exception.". So I would expect that if an unknown role were passed in, it would return false, and in that case, we don't want to call notifyRoleUserRelChange, because nothing has changed.

roryqi:
> The documentation states, "True if the Role was successfully deleted, false only when there's no such role, otherwise it will throw an exception.". So I would expect that if an unknown role were passed in, it would return false, and in that case, we don't want to call notifyRoleUserRelChange, because nothing has changed.

If unknown role passed, the role will be null.

justinmclean:
Also note getRole can return a role, and roleDelete can then return false due to a race condition in the current code.

justinmclean:
For example, getRole() finds the role, but another thread/process deletes it before the deleteRole() call, the deleteRole() would then return false (as no such role exists now), while oldRole is still non-null, and you would call notifyRoleUserRelChange when no change was made. This fix stops that from happening.

roryqi:
Actually, we have upper layer lock to 

> For example, getRole() finds the role, but another thread/process deletes it before the deleteRole() call, the deleteRole() would then return false (as no such role exists now), while oldRole is still non-null, and you would call notifyRoleUserRelChange when no change was made. This fix stops that from happening.

OK.

justinmclean (approved):
Thanks for fixing this.

Raj-G07:
@justinmclean Happy to help.

Issue #5891, opened 2024-12-17
[Bug report] User A create schema1, then user B create schema2 in the same catalog, the creator of schema2 is incorrect.


danhuawang:
### Version

main branch

### Describe what's wrong

The creator of the schema is not the current user.

<img width="1858" alt="image" src="https://github.com/user-attachments/assets/585f5546-32c5-4012-97a3-12a266b69889" />

### Error message and/or stacktrace

N/A

### How to reproduce

1. User A create a catalog named file_catalog
2. User B create schema1 in the file_catalog ,the creator is user B
3. User A create schema2 in the file_catalog ,the creator is user B.   Expection: the creator is user A.

danhuawang:
@jerqi Can you help check this issue?

jerqi:
Maybe you should use user impersonation otherwise auditinfo  will use the user of the Gravitino service.

danhuawang:
> Maybe you should use user impersonation otherwise auditinfo will use the user of the Gravitino service.

But it also didn't use the user of the Gravitino service.

jerqi:
@yuqi1129 Could you take a look?

yuqi1129:
OK, I will take time for this.

yuqi1129:
<img width="1645" height="752" alt="Image" src="https://github.com/user-attachments/assets/007eef54-4014-460e-a6bf-be09bf41deca" />

 PrincipalUtils.getCurrentUserName() will randomly get a user in the UGI, so the problem occurs. The principals in `subject` is a set and can't guarantee the sequence: we should always use the last one.

<img width="960" height="643" alt="Image" src="https://github.com/user-attachments/assets/bd5c62ac-898c-43e6-b470-8386c2bd0177" />


@jerqi Maybe we should try using thread local as I have proposed initially.

roryqi:
Why will it have two user principals?

yuqi1129:
@jerqi 

<img width="1258" height="493" alt="Image" src="https://github.com/user-attachments/assets/03e83a65-475f-4aea-8c4e-3ab7c4b18d1b" />

Different threads will share the same `UserGroupInformation` and we need to clear the user when the method is done.

PR #8123, opened 2025-08-15
[#5891] fix(fileset-catalog): Fix bugs that caused by set User principal in fileset catalog.

yuqi1129:

unset `UserPrincipal` when operation is finished.

It's a bug..

Fix: #5891 

N/A.

Test locally.


roryqi:
Could you add a UT for this fix?

Issue #8109, opened 2025-08-15
[Improvement] Incorrect SQL syntax for Postgres with insert on conflict


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/UserRoleRelPostgreSQLProvider.java the SQL syntax is not correct. Postgres uses EXCLUDED as shown in https://www.postgresql.org/docs/current/sql-insert.html not VALUES.

### How should we improve?

Use EXCLUDED. instead of VALUES().

keepConcentration:
I’d like to work on this issue. Could you please assign it to me?

justinmclean:
It's all yours.

PR #8125, opened 2025-08-15
[#8109] improvement(core): correct PostgreSQL ON CONFLICT syntax using EXCLUDED instead of VALUES

keepConcentration:

This PR fixes the incorrect SQL syntax in `UserRoleRelPostgreSQLProvider.java` by replacing `VALUES()` function calls with `EXCLUDED` references in the `ON CONFLICT ... DO UPDATE` clause. The change ensures proper PostgreSQL compatibility for upsert operations.\

The current implementation uses MySQL-specific `VALUES()` function syntax in PostgreSQL provider, which is incorrect according to PostgreSQL documentation. PostgreSQL's `INSERT ... ON CONFLICT ... DO UPDATE` requires using `EXCLUDED` table alias to reference the values that would have been inserted, not the `VALUES()` function.

Fix: #8109

No user-facing changes. This is a bug fix that ensures the existing functionality works correctly on PostgreSQL databases.  

Tested with existing tests.


justinmclean (approved):
Thanks, that looks good to me.

Issue #8084, opened 2025-08-14
[Improvement] Missing inherited in hashCode and equals TagDTO.java


justinmclean:
### What would you like to be improved?

Missing inherited in hashCode and equals in common/src/main/java/org/apache/gravitino/dto/tag/TagDTO.java

Unit test that shows this:
```
  @Test
  public void testEqualsConsidersInherited() {
    AuditDTO audit = AuditDTO.builder().withCreator("user1").withCreateTime(Instant.now()).build();

    TagDTO ownedTag =
        TagDTO.builder()
            .withName("tag_test")
            .withComment("tag comment")
            .withAudit(audit)
            .withInherited(Optional.of(false))
            .build();

    TagDTO inheritedTag =
        TagDTO.builder()
            .withName("tag_test")
            .withComment("tag comment")
            .withAudit(audit)
            .withInherited(Optional.of(true))
            .build();

    Assertions.assertNotEquals(ownedTag, inheritedTag);
  }
```

### How should we improve?

fix hashCode and equals methods to include inherited

mxnzx:
Could you please assign this issue to me ?

justinmclean:
All yours, if you need any help just ask.

mxnzx:
Hi,

I added a property and my unit tests pass locally. However, when I run the whole unit test suite with:

`./gradlew test -PskipITs`

I get a failure in TestMetadataObjectTagOperations.testListTagsForObject():

```
expected: <4> but was: <6>
at org.apache.gravitino.server.web.rest.TestMetadataObjectTagOperations.testListTagsForObject(TestMetadataObjectTagOperations.java:367)
```


The docs seem to suggest that if the same tag is attached to both a parent and a child, the child’s tag list may include two entries with different inherited values. But this test appears to expect only one entry per tag name, effectively preferring the direct tag over the inherited one, resulting in a total of 4.

Could you clarify the intended behavior for this endpoint?

Should the API return both entries (direct and inherited) when names collide (leading to 6 in this case)?

Or is the intended spec to deduplicate by tag name and prefer direct > inherited, keeping the count at 4?

Depending on your guidance, I can either adjust the server logic (dedup by name with direct precedence) or update the test expectation/documentation accordingly.

justinmclean:
I'd say the test is correct and there needs to be some further changes to the code to filter out duplicates, so I would go with the deduplicate by tag name and prefer direct > inherited, keeping the count at 4.

mxnzx:
[...]

Issue #7933, opened 2025-08-05
[Subtask] add user doc for policy management


mchades:
### Describe the subtask

add user doc for policy management

### Parent issue

#7139

PR #8131, opened 2025-08-16
[#8084] fix: include `inherited` in TagDTO equality 

mxnzx:

- **TagDTO**: include `inherited` in `equals()` and `hashCode()`.
- **MetadataObjectTagOperations**: when listing tags (`details=true`), **de-duplicate by tag name** and **prefer direct over inherited**; names-only path also returns unique names.
- **Tests**: add a unit test to ensure `inherited` affects equality; existing REST test (`TestMetadataObjectTagOperations`) now passes with the dedup logic.

(AS-IS)
- `TagDTO.equals/hashCode` ignored `inherited`, so two tags with the same name but different inheritance were considered equal.
- The list API could return duplicate tag names (one inherited, one direct), and `inherited` on the returned tag might not reflect the stronger (direct) association.

(TO-BE)
- `TagDTO` equality/hash now considers `inherited` so comparisons are accurate.
- Listing tags de-duplicates by name and **keeps the direct association when both direct and inherited exist**. This keeps the expected count stable and the `inherited` flag correct.

Fix: #8084 

- API shape unchanged.
- Listing tags (`/metalakes/{metalake}/objects/.../tags`) now:
  - Does not return duplicate names when both direct and inherited exist.
  - For duplicate names, the returned entry is the **direct** one (`inherited=false`).
- This aligns with existing test expectations; functional behavior is clarified and made consistent.

- Local runs:
  - `./gradlew spotlessApply`
  - `./gradlew :server:test --tests "*TestMetadataObjectTagOperations.testListTagsForObject*"`
  - `./gradlew :common:test --tests "org.apache.gravitino.dto.tag.TestTagDTO"`
  - `./gradlew test -PskipITs`

### DIFF
--- a/common/src/main/java/org/apache/gravitino/dto/tag/TagDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/tag/TagDTO.java
@@ -80,16 +80,21 @@ public class TagDTO implements Tag {
     }
 
     TagDTO tagDTO = (TagDTO) o;
+
+    boolean thisInherited = this.inherited.isPresent() && this.inherited.orElse(false);
+    boolean tagDtoInherited = tagDTO.inherited.isPresent() && tagDTO.inherited.orElse(false);
+
     return Objects.equal(name, tagDTO.name)
         && Objects.equal(comment, tagDTO.comment)
         && Objects.equal(properties, tagDTO.properties)
         && Objects.equal(audit, tagDTO.audit)
-        && Objects.equal(inherited, tagDTO.inherited);
+        && Objects.equal(thisInherited, tagDtoInherited);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(name, comment, properties, audit, inherited);
+    return Objects.hashCode(
+        name, comment, properties, audit, inherited.isPresent() && inherited.orElse(false));
   }
 
   /** @return a new builder for constructing a Tag DTO. */
--- a/common/src/main/java/org/apache/gravitino/dto/tag/TagDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/tag/TagDTO.java
@@ -83,12 +83,13 @@ public class TagDTO implements Tag {
     return Objects.equal(name, tagDTO.name)
         && Objects.equal(comment, tagDTO.comment)
         && Objects.equal(properties, tagDTO.properties)
-        && Objects.equal(audit, tagDTO.audit);
+        && Objects.equal(audit, tagDTO.audit)
+        && Objects.equal(inherited, tagDTO.inherited);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(name, comment, properties, audit);
+    return Objects.hashCode(name, comment, properties, audit, inherited);
   }
 
   /** @return a new builder for constructing a Tag DTO. */
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
@@ -20,11 +20,14 @@ package org.apache.gravitino.server.web.rest;
 
 import com.codahale.metrics.annotation.ResponseMetered;
 import com.codahale.metrics.annotation.Timed;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.DefaultValue;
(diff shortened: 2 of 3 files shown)


mxnzx:
While working on deduplicating tags by name (preferring direct > inherited), I’d like to keep REST responses consistent. My proposal is to always populate the inherited field in outgoing DTOs as true or false (never empty), treating Optional.empty() as direct = false when emitting:


```java
// Normalize for REST output
boolean isInherited = tag.inherited().orElse(false); // empty -> false (direct)
TagDTO dto = DTOConverters.toDTO(tag, Optional.of(isInherited));
```


Would this normalization be acceptable across the API? Or are there cases where we should preserve Optional.empty() in responses?

Related: for getTagForObject (single tag fetch), do you prefer returning inherited=false for direct tags to align with listTagsInfo and client expectations?

Thanks!

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
```diff
@@ -187,18 +189,34 @@ public Response listTagsForMetadataObject(
...
+                byName.put(t.name(), t);
+              } else {
+                boolean prevDirect = !prev.inherited().orElse(false);
+                boolean currDirect = !t.inherited().orElse(false);
+                if (!prevDirect && currDirect) {
+                  byName.put(t.name(), t);
+                }
+              }
+            }
+
```
justinmclean, line 206: There may be a simplified way to do this. e.g. Add all direct tags, then add any inherited tags that have not all been added.

justinmclean:
We've got a couple of tag tests that are failing here. We'll need to review the tests and determine if they or the code require changes.

common/src/main/java/org/apache/gravitino/dto/tag/TagDTO.java
```diff
@@ -80,15 +80,21 @@ public boolean equals(Object o) {
     }
 
     TagDTO tagDTO = (TagDTO) o;
+
+    boolean thisInherited = this.inherited.isPresent() && this.inherited.orElse(false);
+    boolean tagDtoInherited = tagDTO.inherited.isPresent() && tagDTO.inherited.orElse(false);
```
jerryshao, line 85: Inherited can be "true", "false", and "null" by design. This changes mix the "false" and "null" together, which is not correct. Also I delibrately not add this field into "hashcode" and "equals" calculation, because this is not a property of the tag, this is calculated in runtime.

mxnzx: Just to clarify, are you saying that this issue doesn’t need to be fixed?

jerryshao: Not necessary.

mxnzx: Thank you for your feedback

justinmclean: I noticed in another PR (#7933) that the documentation has changed from:
"Same tag can be associated to both parent and child metadata objects. When you list the associated tags of a child metadata object, this tag will be included twice in the result list with different `inherited` values."

To:
"Same tag can be associated to both parent and child metadata objects. But when you list the associated tags of a child metadata object, this tag will be included only once in the result list with `inherited` value `false`."

Was the tests/code updated to reflect this?

mchades: We already have tests for this; see:
https://github.com/apache/gravitino/blob/3dea8edba46fe7951efa1c2051438fc87030101a/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/TagIT.java#L510-L523

justinmclean: Sure, but the behaviour has changed, according to the documentation change.

mchades: > but the behaviour has changed

The behavior changed from what to what?  The documentation change is intended to reflect the real behavior, the previous document had an error.

justinmclean: The original documentation was what I believed to be the intention, which is in part why I raised this as an issue.

mxnzx: I think we should correct the documentation first and then determine whether this issue is still valid. I’m sorry I don’t have full context on this feature. If the issue remains valid after the docs update, please let me know. I’d be glad to assist if there’s any way I can contribute.

mchades: > I think we should correct the documentation first

yeah, the doc will be updated in 
https://github.com/apache/gravitino/pull/8136/files#diff-b1d9307a334d34469324df471c932063b1aa82ad2a7b3bb85ade2bdee7f3e68d

jerryshao:
Close this PR. Thanks @mxnzx for your contribution.

mxnzx:
@jerryshao Glad to have worked on this issue! Looking forward to contributing to more issues.

Issue #8110, opened 2025-08-15
[Improvement] Error in equals  in FilesetCatalogOperations.java


justinmclean:
### What would you like to be improved?

The issue is here:
```
&& (scheme == null ? that.authority == null : scheme.equals(that.scheme))
&& (authority == null ? that.authority == null : authority.equals(that.authority))
```

Here's a unit test for it:
```
  @Test
  public void testEqualsWhenSchemeIsNull() {
    Map<String, String> conf = Collections.singletonMap("key", "val");
    FilesetCatalogOperations.FileSystemCacheKey key1 =
        new FilesetCatalogOperations.FileSystemCacheKey(null, "auth", conf);
    FilesetCatalogOperations.FileSystemCacheKey key2 =
        new FilesetCatalogOperations.FileSystemCacheKey(null, "auth", conf);

    assertEquals(key1, key2);
    assertEquals(key1.hashCode(), key2.hashCode());
  }
```

### How should we improve?

Use that.scheme not that.authority in the first line.

JeonDaehong:
Please assign it to me. Thank you.

justinmclean:
It's yours, as usual if you need any help, just ask.

Raj-G07:
Attempt : #8110

JeonDaehong:
@Raj-G07 
I am assigned to this issue and currently working on it.

Raj-G07:
@JeonDaehong Since it had a couple of days, I wasn’t sure if work had started yet, so I went ahead and solved this issue.
                           Could you please take a look at  #8132 instead?

justinmclean:
@Raj-G07 While I appreciate your enthusiasm, to avoid conflicts and duplicated effort, please don’t work on issues that are already assigned to others.

justinmclean:
However, if an issue has been open for a couple of weeks without any progress, please feel free to ask whether you can work on it, but please make sure to ask first before working on it.

JeonDaehong:
@justinmclean 

Thank you for letting me know.

I plan to resolve this issue within this week.

Thank you.

PR #8132, opened 2025-08-16
[#8110] Improvement: Equality check in equals method in FileSetCatalogOperations.java

Raj-G07:
  - Fixed a bug in the `equals` method where `scheme` was incorrectly compared to `that.authority`.  
  - Updated the logic to correctly compare `scheme` with `that.scheme`

   This fix ensures object equality is consistent and accurate.  

Fix: #8110

  - No user-facing changes.  
  - This fix only corrects the internal equality logic.

 Verified that objects with different `scheme` values are not considered equal.


yuqi1129:
LGTM

@Raj-G07 
Please change the title and description of the PR, if you have any problems, you can refer to those of other PRs. Thanks

Raj-G07:
Thanks for reviewing and merging it!

PR #8144, opened 2025-08-18
[#8083] fix: null-safe handling of 'managed' property in CreateFileset.java

### LINKED ISSUE #8083: [Improvement] Potential NullPointerException in parsing the managed property when creating filesets

In clients/cli/src/main/java/org/apache/gravitino/cli/commands/CreateFileset.java, this line in handle() could cause a NPE if the property doesn't exist.
```
boolean managed = properties.get("managed").equals("true");
```

Here a test showing that:
```
 @Test
  void testOperationStatusNullHandling() {
    Assertions.assertDoesNotThrow(() -> CompatibilityUtils.toAuditLogStatus(null));
    Assertions.assertEquals(Status.UNKNOWN, CompatibilityUtils.toAuditLogStatus(null));
  }
```

### How should we improve?

Defaulting to external mode if the property is missing.

JeonDaehong:
In `CreateFileset.java`, the `handle()` method directly called `properties.get("managed").equals("true")`, which could throw a `NullPointerException` if the `managed` property was missing.  
This PR updates the code to use a null-safe check and defaults to external mode (`Fileset.Type.EXTERNAL`) when the property is not set.

Without this fix, missing `managed` properties could cause NPEs, breaking fileset creation.  
Adding a null-safe check ensures that missing properties are handled gracefully and the command defaults to external mode.

Fixes #8083 

No. Valid requests behave the same. Invalid or missing `managed` properties are now safely handled without throwing exceptions.

- Manually verified locally that creating filesets without the `managed` property defaults to external mode.  
- Added unit test (or can be added) to ensure that missing `managed` properties do not cause exceptions.

### DIFF
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/commands/CreateFileset.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/commands/CreateFileset.java
@@ -20,6 +20,7 @@
 package org.apache.gravitino.cli.commands;
 
 import java.util.Map;
+import java.util.Optional;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.cli.CommandContext;
 import org.apache.gravitino.cli.ErrorMessages;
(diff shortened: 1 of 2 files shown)


JeonDaehong:
@justinmclean 

The tests have passed.
I would appreciate it if you could merge it.

justinmclean:
Coudl you add the test please.

JeonDaehong:
> Coudl you add the test please.

Got it.
I’ll add the test method today and update the PR!

JeonDaehong:
@justinmclean 

Added test code and committed the changes.

If the tests have been completed and there are no issues, please proceed with the merge.

Thank you.

JeonDaehong:
> Looks good to me. Thanks!

Thanks !! 😊

Issue #4377, opened 2024-08-05
[Improvement] Remove duplicate code of RemoveResponse and DeleteResponse


xunliu:
### What would you like to be improved?

Currently, have the same logic code `RemoveResponse` and `DeleteResponse` in the Gravitino, It's duplicate code

### How should we improve?

I think we can delete anyone, I recommend delete `RemoveResponse`.

PR #8146, opened 2025-08-18
[#4377] fix(common): Deprecate the DeleteResponse and use DropResponse instead

jerryshao:

`DeleteResponse` is duplicated with `DropResponse`, so deprecate the `DeleteResponse` and use `DropResponse` instead.

Fix: #4377

Yes

Add UTs to test the compability.


jerqi (approved):

roryqi:
One question: Should we remove this class before release 1.0?

jerryshao:
> One question: Should we remove this class before release 1.0?

I would keep it, so we can use it for backward-compatible test.

roryqi:
> > One question: Should we remove this class before release 1.0?
> 
> I would keep it, so we can use it for backward-compatible test.

OK.

Issue #7168, opened 2025-05-14
[Bug report] Creating indexes for JDBC backed DB using Gravitino scripts


mike-g-hyle:
### Version

main branch

### Describe what's wrong

In this file:

- https://github.com/apache/gravitino/blob/main/scripts/postgresql/schema-0.9.0-postgresql.sql

There are multiple creations references to a same named index on different tables. For example `idx_metalake_id`.

When creating these indexes, one first one will be created, all other will not be created because indexes should be unique to the schema. This happens with more than one index. `idx_metalake_id` is only an example.

### Error message and/or stacktrace

Index alread exists, skipping

### How to reproduce

Gravito-0.9.0-incubating

Run https://github.com/apache/gravitino/blob/main/scripts/postgresql/schema-0.9.0-postgresql.sql on a postgres DB.

yuqi1129:
Can you provide more details here? I couldn't understand what was being said.

justinmclean:
For Postgres, index names are unique within a schema, not within a table. The script contains multiple indexes with the same name, which will not be created beyond the first one.

yuqi1129:
Let me try again.

PR #8147, opened 2025-08-18
[#7168] fix(script): Fix index name uniqueness in schema for PG.

yuqi1129:

Add the table name as the prefixes for the index name in PG. 

For PG, the index name should be unique within a schema, NOT a table.

Fix: #7168 

N/A.

Locally.


jerryshao:
Can you please confirm that there's no duplicated index in PG script.

yuqi1129:
> Can you please confirm that there's no duplicated index in PG script.

What do you mean here? I have added the `table_name` as the prefix of the index name; there should no duplicated one here.

> Can you please confirm that there's no duplicated index in PG script.

Checked, no duplicateds can be found.

jerryshao (approved):

Issue #8167, opened 2025-08-19
[Improvement] MyBatis add mapper will load in lots of classes other that mapper class


jerryshao:
### What would you like to be improved?

Currently, we're using package name to load in all the mapper class to MyBatis configuration, but our package contains lots of classes other than mapper class, so they will also load in unexpected to increase the footprinter. So here slightly adjust the SPI to avoid load in the unwanted classes.

PR #8170, opened 2025-08-19
[#8167]improve(core): Improve the SPI to load in the specific mapper class

jerryshao:

Change the SPI interface to add the specific mapper list instead of general package name.

Currently, we're using package name to load in all the mapper class to MyBatis configuration, but our package contains lots of classes other than mapper class, so they will also load in unexpected classes to increase the footprint. So here slightly adjust the SPI to avoid load in the unwanted classes.

Fix: #8167 

No.

Existing tests.


yuqi1129:
I recalled that the enterprise edition already uses package names to load several extra mappers. @mchades Can you please help verify this?

mchades (approved):
lgtm.

Issue #7355, opened 2025-06-06
[Bug report] Some tables loaded in gravitino postgres catalog but can't show in Trino


danhuawang:
### Version

main branch

### Describe what's wrong

Some tables loaded in postgre catalog as following

<img width="338" alt="Image" src="https://github.com/user-attachments/assets/0e1dc42d-310e-4e4d-90f1-bbfa2cce13fd" />

But  in Trino , I can't list these tables , and this break data preview  process:

<img width="526" alt="Image" src="https://github.com/user-attachments/assets/1dd5bdfe-d41a-4a71-86c8-2d1fd848dd4c" />

### Error message and/or stacktrace

```
[
  "java.lang.RuntimeException: java.sql.SQLException: Query failed (#20250605_201352_00061_d9xb2): line 1:15: Table '\"\"\"test.hadoop_pg\"\"\".public.delegation_tokens' does not exist",
  "\tat com.datastrato.gravitino.preview.TrinoJdbcDataPreviewOperator.preview(TrinoJdbcDataPreviewOperator.java:86)",
  "\tat com.datastrato.gravitino.catalog.DatastratoTableOperationDispatcher.preview(DatastratoTableOperationDispatcher.java:71)",
  "\tat com.datastrato.gravitino.catalog.DatastratoTableHookDispatcher.preview(DatastratoTableHookDispatcher.java:34)",
  "\tat com.datastrato.gravitino.catalog.DatastratoTableNormalizeDispatcher.preview(DatastratoTableNormalizeDispatcher.java:49)",
  "\tat com.datastrato.gravitino.listener.DatastratoTableEventDispatcher.preview(DatastratoTableEventDispatcher.java:56)",
  "\tat com.datastrato.gravitino.server.web.rest.DataPreviewOperations.lambda$previewTable$0(DataPreviewOperations.java:77)",
  "\tat java.base/java.security.AccessController.doPrivileged(AccessController.java:712)",
  "\tat java.base/javax.security.auth.Subject.doAs(Subject.java:439)",
  "\tat org.apache.gravitino.utils.PrincipalUtils.doAs(PrincipalUtils.java:39)",
  "\tat org.apache.gravitino.server.web.Utils.doAs(Utils.java:198)",
  "\tat com.datastrato.gravitino.server.web.rest.DataPreviewOperations.previewTable(DataPreviewOperations.java:71)",
  "\tat jdk.internal.reflect.GeneratedMethodAccessor144.invoke(Unknown Source)",
[...]

PR #8171, opened 2025-08-19
[#7355] fix (doc): Update document for uppercase table and column names not supported in trino-connector  of PostgreSQL catalog

diqiu50:

Update document for uppercase table and column names not supported in trino-connector

Fix: #7355

Update documents

NO

### DIFF
--- a/docs/trino-connector/catalog-postgresql.md
+++ b/docs/trino-connector/catalog-postgresql.md
@@ -22,7 +22,7 @@ For example, the table name MyTable is equivalent to mytable and MYTABLE.
 However, if you create a table with quoted identifiers, such as "MyTable", it becomes case sensitive and must be referenced exactly as "MyTable".
 
 When using the Gravitino Trino connector with PostgreSQL, you must use unquoted identifiers to avoid case sensitivity issues.
-Otherwise, table names or column names containing uppercase letters may not be found.
+Otherwise, schema names, table names or column names containing uppercase letters may not be found.
 
 ## Create table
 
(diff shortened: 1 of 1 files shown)


docs/trino-connector/catalog-postgresql.md
```diff
@@ -14,6 +14,16 @@ To connect to PostgreSQL, you need:
 - PostgreSQL 10.x or higher.
 - Network access from the Trino coordinator and workers to PostgreSQL. Port 5432 is the default port.
 
+## Case sensitivity
+
+PostgreSQL treats unquoted identifiers as case insensitive.
+For example, the table name MyTable is equivalent to mytable and MYTABLE.
+
+However, if you create a table with quoted identifiers, such as "MyTable", it becomes case sensitive and must be referenced exactly as "MyTable".
```
yuqi1129, line 22: What is the case sensitivity of schema names?

diqiu50: The `schema` is the same. I fixed

yuqi1129: > PostgreSQL treats unquoted identifiers as case insensitive.
For example, the table name MyTable is equivalent to mytable and MYTABLE.

What is the final result if an unquoted table `MyTable` is created?  is it `mytable`? Please explain it in the docs

diqiu50: It seems like 'mytable', but using 'mytable', 'MyTable', or 'MYTABLE' in queries is valid.

yuqi1129: I see.

yuqi1129 (approved):

PR #8173, opened 2025-08-19
[#7436] feat(model): Implement REST APIs for model version with multiple URIs

xiaozcy:

Implement REST APIs for model version with multiple URIs.

Fix: #7436 

Yes.

UTs.

### DIFF
--- a/common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionLinkRequest.java
+++ b/common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionLinkRequest.java
@@ -94,7 +94,9 @@ public class ModelVersionLinkRequest implements RESTRequest {
     }
 
     if (uris == null || uris.isEmpty()) {
-      Preconditions.checkArgument(StringUtils.isNotBlank(uri), "URI cannot be empty");
+      Preconditions.checkArgument(
+          StringUtils.isNotBlank(uri),
+          "\"uri\" field cannot be empty if uris field is null or empty");
     } else {
       uris.forEach(
           (n, u) -> {
--- a/common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionLinkRequest.java
+++ b/common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionLinkRequest.java
@@ -56,11 +56,27 @@ public class ModelVersionLinkRequest implements RESTRequest {
   @JsonProperty("properties")
   private Map<String, String> properties;
 
+  /**
+   * Constructor for ModelVersionLinkRequest.
+   *
+   * @param uri the uri of the model version.
+   * @param aliases the aliases of the model version.
+   * @param comment the comment of the model version.
+   * @param properties the properties of the model version.
+   */
   public ModelVersionLinkRequest(
       String uri, String[] aliases, String comment, Map<String, String> properties) {
     this(uri, null, aliases, comment, properties);
   }
 
+  /**
+   * Constructor for ModelVersionLinkRequest
+   *
+   * @param uris the uris of the model version.
+   * @param aliases the aliases of the model version.
+   * @param comment the comment of the model version.
+   * @param properties the properties of the model version.
+   */
   public ModelVersionLinkRequest(
       Map<String, String> uris, String[] aliases, String comment, Map<String, String> properties) {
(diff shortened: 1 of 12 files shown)


xiaozcy:
Hi @jerryshao, please help review it when you are free, thx.

docs/open-api/models.yaml
```diff
@@ -422,13 +482,20 @@ components:
...
       required:
-        - uri
         - version
         - audit
       properties:
         uri:
           type: string
           description: The uri of the model version
+          nullable: true
+          default: {}
```
jerryshao: IIUC, `uri` is a string, not a dict.

server/src/main/java/org/apache/gravitino/server/web/rest/ModelOperations.java
```diff
@@ -662,6 +673,86 @@ public Response alterModel(
     }
   }
 
+  @GET
+  @Path("{model}/versions/{version}/uri")
```
jerryshao, line 677: I would suggest to use path parameter, rather than query parameter for uri name, like `versions/{version}/uris/{uriName}`.

xiaozcy: The `uriName` is not required for this  API. If it is not specified, we will use the `default-uri-name` property to select a URI.

docs/open-api/models.yaml
xiaozcy: Yes, it's a mistake, thanks for pointing it out.

common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionLinkRequest.java
```diff
@@ -63,5 +92,15 @@ public void validate() throws IllegalArgumentException {
             !NumberUtils.isCreatable(alias), "alias must not be a number or a number string");
       }
     }
+
+    if (uris == null || uris.isEmpty()) {
+      Preconditions.checkArgument(StringUtils.isNotBlank(uri), "URI cannot be empty");
```
jerryshao: Can you please add more description about this check, for example, "uri field cannot be empty if uris field is null or empty".

xiaozcy: Fixed.

docs/open-api/models.yaml
```diff
@@ -366,6 +366,66 @@ paths:
         "5xx":
           $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
 
+  /metalakes/{metalake}/catalogs/{catalog}/schemas/{schema}/models/{model}/versions/{version}/uri:
+    parameters:
+      - $ref: "./openapi.yaml#/components/parameters/metalake"
+      - $ref: "./openapi.yaml#/components/parameters/catalog"
+      - $ref: "./openapi.yaml#/components/parameters/schema"
+      - $ref: "./openapi.yaml#/components/parameters/model"
+      - $ref: "#/components/parameters/version"
```
jerryshao, line 375: I think you miss the query parameter definition here.

docs/open-api/models.yaml
```diff
@@ -366,6 +366,66 @@ paths:
...
+        "5xx":
+          $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
+
+  /metalakes/{metalake}/catalogs/{catalog}/schemas/{schema}/models/{model}/aliases/{alias}/uri:
+    parameters:
+      - $ref: "./openapi.yaml#/components/parameters/metalake"
+      - $ref: "./openapi.yaml#/components/parameters/catalog"
+      - $ref: "./openapi.yaml#/components/parameters/schema"
+      - $ref: "./openapi.yaml#/components/parameters/model"
+      - $ref: "#/components/parameters/alias"
```
jerryshao, line 413: Also here.

docs/open-api/models.yaml
xiaozcy: Fixed.

docs/open-api/models.yaml
xiaozcy: Fixed.

jerryshao (approved):

Issue #5199, opened 2024-10-21
[Subtask] Support Table partitioning, bucketing and sort ordering and indexes


unknowntpo:
### Describe the subtask

docs: https://gravitino.apache.org/docs/0.6.1-incubating/table-partitioning-bucketing-sort-order-indexes/
related issues: #248  #264 

### Parent issue

#5198

tsungchih:
I was wondering if I could take this one. Here's the scope and the plan for this issue. And hope my understanding is correct.

The 1st PR

- [x] PartitionDTO.java
- [x] IdentityPartitionDTO.java
- [x] ListPartitionDTO.java
- [x] RangePartitionDTO.java

PRs:
- #7902 
- #7904 

The 2nd PR

- [x] JsonUtils.java (PartitionDTOSerializer, PartitionDTODeserializer)

PRs:
- #8128 

The 3rd PR

- [x] Partitioning.java
- [x] IdentityPartitioningDTO.java
- [x] YearPartitioningDTO.java
- [x] MonthPartitioningDTO.java
- [x] DayPartitioningDTO.java
- [x] HourPartitioningDTO.java
- [x] BucketPartitioningDTO.java
- [x] ListPartitioningDTO.java
- [x] RangePartitioningDTO.java
- [x] TruncatePartitioningDTO.java
- [x] FunctionPartitioningDTO.java
 
PRs: 
- #8004 
- #8129 

The 4th PR

- [x] JsonUtils.java (PartitioningSerializer, PartitioningDeserializer)

PRs:
- #8363 

The 5th PR

- [x] DistributionDTO.java
- [x] JsonUtils.java (DistributionSerializer, DistributionDeserializer)

PRs:
- #8185 
- #8458 

The 6th PR
- [x] SortOrderDTO.java
- [x] JsonUtils.java (SortOrderSerializer, SortOrderDeserializer)

PRs:
- #8220 
- #8459 

The 7th PR
- [x] Index.java
- [x] Indexes.java

PRs:
- #8143 

The 8th PR
- [x] IndexDTO.java
- [x] JsonUtils.java (IndexSerializer, IndexDeserializer)

PRs:
- #8262 
- #8457

unknowntpo:
@tsungchih Okay.

jerryshao:
Hi @tsungchih @unknowntpo, how many PRs do we still need for this issue?

unknowntpo:
[...]

PR #8185, opened 2025-08-19
[#5199] feat(client-python): add distribution DTO

tsungchih:

- DistributionDTO

**NOTE** that an refactor of `FieldReferenceDTO` in its Builder method `with_column_name()` is embraced in this PR as well. The refactor is aimed at conforming to [the one in Java client.](https://github.com/apache/gravitino/blob/main/common/src/main/java/org/apache/gravitino/dto/rel/expressions/FieldReferenceDTO.java#L77)

We need to support table partitioning, bucketing and sort ordering and indexes

#5199

No

Unit tests


tsungchih:

Issue #8162, opened 2025-08-19
[Improvement] Improve oauth in the CLI  in GravitinoConfig.java


justinmclean:
### What would you like to be improved?

Correct clients/cli/src/main/java/org/apache/gravitino/cli/GravitinoConfig.java so that OAuth and Kerberos authentication data is detected using the parsed authType rather than the constant key.

These tests should help:
```

  @Test
  public void oauthConfigFile() throws IOException {
    File tempFile = new File(System.getProperty("java.io.tmpdir") + "/oauth_config.properties");

    Properties props = new Properties();
    props.setProperty("auth", "oauth");
    props.setProperty("serverURI", "http://oauth.example.com");
    props.setProperty("credential", "cred");
    props.setProperty("token", "token");
    props.setProperty("scope", "scope");

    try (Writer writer = Files.newBufferedWriter(tempFile.toPath(), StandardCharsets.UTF_8)) {
      props.store(writer, "OAuth Config");
    }

    GravitinoConfig config = new GravitinoConfig(tempFile.getAbsolutePath());
    config.read();
    assertEquals("oauth", config.getGravitinoAuthType());
    assertNotNull(config.getOAuth());
    assertEquals("http://oauth.example.com", config.getOAuth().getServerURI());
    assertEquals("cred", config.getOAuth().getCredential());
    assertEquals("token", config.getOAuth().getToken());
    assertEquals("scope", config.getOAuth().getScope());

    tempFile.delete();
  }

  @Test
  public void kerberosConfigFile() throws IOException {
    File tempFile =
        new File(System.getProperty("java.io.tmpdir") + "/kerberos_config.properties");

    Properties props = new Properties();
    props.setProperty("auth", "kerberos");
    props.setProperty("principal", "user@REALM");
    props.setProperty("keytabFile", "/path/to/keytab");

    try (Writer writer = Files.newBufferedWriter(tempFile.toPath(), StandardCharsets.UTF_8)) {
      props.store(writer, "Kerberos Config");
    }

    GravitinoConfig config = new GravitinoConfig(tempFile.getAbsolutePath());
    config.read();
    assertEquals("kerberos", config.getGravitinoAuthType());
    assertNotNull(config.getKerberos());
    assertEquals("user@REALM", config.getKerberos().getPrincipal());
    assertEquals("/path/to/keytab", config.getKerberos().getKeytabFile());

    tempFile.delete();
  }
```

### How should we improve?

See above

andyyngo:
[...]

PR #8187, opened 2025-08-19
fix(cli): Fix OAuth/Kerberos authentication detection in GravitinoConfig

mbaurin:
### What changes were proposed in this pull request ?

This PR fixes the OAuth/Kerberos authentication detection logic in `GravitinoConfig.java`. The changes include:

1. **Fixed authentication type comparisons**: Replace `authKey.equals("oauth")` with `"oauth".equals(authType)` and similar for Kerberos
2. **Removed duplicate OAuth initialization**: Eliminated redundant OAuth configuration block that was overwriting Kerberos settings
3. **Consolidated control flow**: Created clean if-else if structure for proper authentication type handling

The root issue was that the code was comparing against the constant `authKey` ("auth") instead of the parsed `authType` value from the configuration file.

The current implementation has critical bugs that prevent OAuth and Kerberos authentication from working correctly:

1. **Wrong variable comparison**: Code used `authKey` (constant "auth") instead of `authType` (parsed value like "oauth" or "kerberos")
2. **Configuration overwriting**: Duplicate OAuth logic was overwriting any Kerberos configuration
3. **Unreachable code paths**: Kerberos authentication could never be properly initialized

These bugs prevent the CLI from correctly detecting and configuring authentication when `auth=oauth` or `auth=kerberos` is set in the configuration file.

Fix: https://github.com/apache/gravitino/issues/8162

**No user-facing API changes**, but fixes critical authentication functionality:

- When `auth=oauth` is configured, `config.getOAuth()` now correctly returns valid OAuth configuration
- When `auth=kerberos` is configured, `config.getKerberos()` now correctly returns valid Kerberos configuration
- No changes to configuration file format or property keys

1. **Code review**: Verified the logic fixes address all identified issues
2. **Integration verification**: Confirmed the changes work with existing `Command.java` usage patterns at lines 202-223
3. **Style compliance**: Ensured code follows Apache Gravitino conventions using defensive null-safe comparisons

The existing unit tests in `TestGravitinoConfig.java` should pass without modification

### DIFF
--- a/clients/cli/src/test/java/org/apache/gravitino/cli/TestGravitinoConfig.java
+++ b/clients/cli/src/test/java/org/apache/gravitino/cli/TestGravitinoConfig.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.cli;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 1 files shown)


justinmclean:
@mbaurin thanks for the fix, can you please add the tests as well.

mbaurin:
@justinmclean tests are done 😄

justinmclean:
Thanks for the tests

mbaurin:
will fix the CI !

justinmclean:
I'm sure you're aware but the test failed because we use spotless, you can run `./gradlew :clients:cli:spotlessApply` and then commit the changes it makes to fix.

mbaurin:
> I'm sure you're aware but the test failed because we use spotless, you can run `./gradlew :clients:cli:spotlessApply` and then commit the changes it makes to fix.

thanks for the tip, it is fixed :)

justinmclean (approved):
Thanks for fixing this.

Issue #8177, opened 2025-08-19
[Bug report] impersonation doesn't take affect when creating table in Iceberg REST catalog with auth type kerberos


FANNG1:
### Version

main branch

### Describe what's wrong

`HiveBackendProxy` hooks the method of `HiveCatalog` to do the impersonation, take create table for example, it only hooks catalog.build method, but the real create table operation acts in the create method which is not hooked.

```
    Table table =
        catalog
            .buildTable(ident, request.schema())
            .withLocation(request.location())
            .withPartitionSpec(request.spec())
            .withSortOrder(request.writeOrder())
            .withProperties(request.properties())
            .create();
```

### Error message and/or stacktrace

no

### How to reproduce

apply the patch to `CatalogIcebergKerberosHiveIT#testIcebergWithKerberosAndUserImpersonation`
```java
+    kerberosHiveContainer.executeInContainer(
+        "hadoop", "fs", "-chmod", "-R", "000", "/user/hive/warehouse-catalog-iceberg");
+
     // Create table
+
     NameIdentifier tableNameIdentifier = NameIdentifier.of(SCHEMA_NAME, TABLE_NAME);
-    catalog
-        .asTableCatalog()
-        .createTable(
-            tableNameIdentifier,
-            createColumns(),
-            "",
-            ImmutableMap.of(),
-            Transforms.EMPTY_TRANSFORM,
-            Distributions.NONE,
-            SortOrders.NONE);
+    exception =
+        Assertions.assertThrows(
+            Exception.class,
+            () ->
+                catalog
+                    .asTableCatalog()
+                    .createTable(
+                        tableNameIdentifier,
+                        createColumns(),
+                        "",
+                        ImmutableMap.of(),
+                        Transforms.EMPTY_TRANSFORM,
+                        Distributions.NONE,
+                        SortOrders.NONE));
+    exceptionMessage = Throwables.getStackTraceAsString(exception);
+    System.out.println(exceptionMessage);
+    // Make sure the real user is 'gravitino_client'
+    Assertions.assertTrue(
+        exceptionMessage.contains("Permission denied: user=gravitino_client, access=WRITE"));
```

yuqi1129:
```
    Table table = catalog.buildTable(ident, request.schema()).withLocation(request.location()).withPartitionSpec(request.spec()).withSortOrder(request.writeOrder()).withProperties(request.properties()).create();
```

method `create()` is not the method of `HiveCatalog`, so the proxy can't impersonate it.

PR #8189, opened 2025-08-19
[#8177] fix(iceberg-common): Fix the impersonation problem in create iceberg table

yuqi1129:

Handle impersonation problem for Iceberg table using Hive backend.

Impersonation does not work for iceberg catalog with Hive backend

Fix: #8177

N/A.

IT


yuqi1129:
After discussing with @FANNG1 offline, the current implementation may not be extensive enough; there can be problems if we upgrade the version of Iceberg or if the method signatures change. I'm going to explore a more general solution.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -60,4 +72,64 @@ public void close() throws IOException {
...
+  public <R> R execute(Executable<R> runnable) {
+    try {
+      return runnable.run();
+    } catch (Exception e) {
+      LOGGER.error("Failed to execute runnable: {}", runnable, e);
+      throw new RuntimeException(e);
+    }
+  }
+
+  public Catalog.TableBuilder buildTable(TableIdentifier identifier, Schema schema) {
```
agnes-xinyi-lu: It will not be scalable if we only fix createTable. updateTable also has the same problem. Same with create/updateView. 
The solution could be adding another layer of impersonation on Iceberg TableOperations where it does the real load/update table. Or move the impersonation up a level to catalog wrapper. 
This is a bit off topic of this problem but still related if we impersonate the catalog wrapper, is it OK to make impersonation more generic instead of just on kerberorized HiveCatalog? Most companies have their own network and authentication mechanism, they could use whatever authenticator that security allows. And the service that are talking to Gravitino IRC server, could impersonate any user in any way they want, as long as Gravitino thinks it's a superuser that could impersonate others. Kerberos impersonation could be one of the implementations.

FANNG1: @agnes-xinyi-lu yes, this solutino is not so general, we'd better make the impersonation more extensible.  @yuqi1129 , how about closing this PR since we didn't use this solution.

yuqi1129: Agree, a more extensive solution is better than this one.

yuqi1129:
Close it temporarily as it's a not a desired solution.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
yuqi1129: @FANNG1 @agnes-xinyi-lu 
The code has been updated. Please review it again, thanks.

FANNG1:
Is this PR ready to review?

yuqi1129:
> Is this PR ready to review?

Yeah, please take a look.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+  }
+
+  @Override
+  public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy)
+      throws Throwable {
+    if (needProxy) {
+      final String finalPrincipalName;
+      String proxyKerberosPrincipalName = PrincipalUtils.getCurrentPrincipal().getName();
```
agnes-xinyi-lu: Instead of bounding this with kerberos, is it possible to add a ImpersonationPrinciplaProvider interface which lets user to customize the way to pass in user principal? For example, I'm just using whatever user name is provided in the authenticator.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -60,4 +74,120 @@ public void close() throws IOException {
...
+  //    ExtendedViewAwareTableBuilderExtended builder =
+  //        new ExtendedViewAwareTableBuilderExtended(identifier, schema);
+  //    ViewAwareTableBuilderProxy proxy =
+  //        new ViewAwareTableBuilderProxy(builder, this.proxy != null ? this.proxy : this);
+  //    ExtendedViewAwareTableBuilderExtended r =
+  //        proxy.getProxy(new Object[] {this, identifier, schema});
+  //    return r;
+  //  }
+
+  public static class ViewAwareTableBuilderProxy implements MethodInterceptor {
```
agnes-xinyi-lu: are these still needed?

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalogOperations.java
```diff
@@ -113,7 +114,8 @@ public void initialize(
     resultConf.put("catalog_uuid", info.id().toString());
     IcebergConfig icebergConfig = new IcebergConfig(resultConf);
 
-    this.icebergCatalogWrapper = new IcebergCatalogWrapper(icebergConfig);
+    IcebergCatalogWrapper rawWrapper = new IcebergCatalogWrapper(icebergConfig);
```
FANNG1, line 119: Could we add a check to determine whether we should proxy the IcebergCatalogWrapper?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
FANNG1: `PrincipalUtils.getCurrentPrincipal().getName()` is the username provided in the authenticator, is this what you expected?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
```diff
@@ -0,0 +1,143 @@
...
+  boolean needProxy = false;
+  private KerberosClient kerberosClient;
+  private ClientPool<IMetaStoreClient, TException> newClientPool;
+
+  public IcebergCatalogWrapperProxy(IcebergCatalogWrapper target) {
+    this.target = target;
+    Catalog catalog = target.catalog;
+
+    // Need special handling for HiveCatalog.
+    if (catalog instanceof HiveCatalog) {
```
FANNG1: Could you do some abstraction to remove the hive catalog checks.  we could support impersonation in custom catalog backend.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalogOperations.java
yuqi1129: Please refer to the comments mentioned above.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
agnes-xinyi-lu: +1

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
```diff
@@ -0,0 +1,125 @@
...
+    if (!authenticationConfig.isImpersonationEnabled()) {
+      return methodProxy.invoke(target, objects);
+    }
+
+    final String finalPrincipalName;
+    String proxyKerberosPrincipalName = PrincipalUtils.getCurrentPrincipal().getName();
+
+    if (!proxyKerberosPrincipalName.contains("@")) {
+      finalPrincipalName =
+          String.format("%s@%s", proxyKerberosPrincipalName, kerberosClient.getRealm());
```
agnes-xinyi-lu: kerberosClient could be null and then here it throws exception. Is it possible to make kerberosClient optional or exposed through a generic interface

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
agnes-xinyi-lu: Right, in my scenario that's the case, but I'm not asking for the exact solution, only suggesting to make it more generic or customizable so I could implement my own class to provide impersonation info instead of relying on Kerberos.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
```diff
@@ -0,0 +1,130 @@
...
+    UserGroupInformation realUser =
+        authenticationConfig.isImpersonationEnabled()
+            ? UserGroupInformation.createProxyUser(
+                finalPrincipalName, kerberosClient.getLoginUser())
+            : kerberosClient.getLoginUser();
+
+    try {
+      ClientPool<IMetaStoreClient, TException> newClientPool =
+          (ClientPool<IMetaStoreClient, TException>)
+              FieldUtils.readField(target.catalog, "clients", true);
```
FANNG1: seems a little tacky to reference `clients` directly here

yuqi1129: Would you happen to have any good suggestions?  The code is related to `HiveCatalog`, it's proper to put it here, the best way is place it in the hive catalog extension, as we intend to add a wapper there, I could not come up with a more elegant away right now.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
yuqi1129: This is not the scope of this PR.  The user `PrincipalUtils.getCurrentPrincipal().getName()` is the one who call Gravitino API to do operations such as creating tables, dropping tables, I believe `SimpleTokenProvider` is the one you need.

yuqi1129:
@FANNG1 
Could you give some advice on the way to fix the problem this PR tries to fix?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -68,6 +112,135 @@ public void close() throws IOException {
...
+          String.format("%s@%s", proxyKerberosPrincipalName, kerberosClient.getRealm());
+    } else {
+      finalPrincipalName = proxyKerberosPrincipalName;
+    }
+
+    UserGroupInformation realUser =
+        authenticationConfig.isImpersonationEnabled()
+            ? UserGroupInformation.createProxyUser(
+                finalPrincipalName, kerberosClient.getLoginUser())
+            : kerberosClient.getLoginUser();
```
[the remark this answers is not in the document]
yuqi1129: It will throw an exception if `initKerberosClient` fails, so there is no problem here.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -47,8 +67,32 @@ public void addResource(Closeable resource) {
...
+    AuthenticationConfig authenticationConfig = new AuthenticationConfig(properties);
+    if (authenticationConfig.isKerberosAuth()) {
+      this.kerberosClient = initKerberosClient();
+    }
+
+    try {
+      resetIcebergHiveClientPool();
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to reset IcebergHiveClientPool", e);
+    }
```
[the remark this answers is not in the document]
yuqi1129: `initKerberosClient` will throw an exception if it fails, so it's okay.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/KerberosAwareIcebergCatalogProxy.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public IcebergCatalogWrapper getProxy(IcebergConfig config) {
+    Enhancer e = new Enhancer();
+    e.setClassLoader(target.getClass().getClassLoader());
+    e.setSuperclass(target.getClass());
+    e.setCallback(this);
+
+    Class<?>[] argClass = new Class[] {IcebergConfig.class};
+    return (IcebergCatalogWrapper) e.create(argClass, new Object[] {config});
+  }
```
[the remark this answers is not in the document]
yuqi1129: It's no a problem

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -68,6 +112,135 @@ public void close() throws IOException {
...
+
+    UserGroupInformation realUser =
+        authenticationConfig.isImpersonationEnabled()
+            ? UserGroupInformation.createProxyUser(
+                finalPrincipalName, kerberosClient.getLoginUser())
+            : kerberosClient.getLoginUser();
+    try {
+      ClientPool<IMetaStoreClient, TException> newClientPool =
+          (ClientPool<IMetaStoreClient, TException>) FieldUtils.readField(this, "clients", true);
```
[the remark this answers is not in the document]
yuqi1129: There is no need to do so.

FANNG1:
@jerqi do you hive time to reivew this PR?

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalogOperations.java
```diff
@@ -113,7 +115,18 @@ public void initialize(
...
+    // We have replaced `UserGroupInformation#loginUserFromKeytab` with
+    // `UserGroupInformation#loginUserFromKeytabAndReturnUGI`, the former will change the current
+    // login user globally, which is not expected in Gravitino as we are going to support multiple
+    // catalogs within the same class loader. The proxy will ensure each catalog has its own
+    // `UserGroupInformation` instance and I have removed old `HiveBackendProxy`. In
+    // IcebergCatalogWrapperProxy, we will do both Kerberos access and user impersonation if needed,
+    // so please check the code in IcebergCatalogWrapperProxy for details.
+    this.icebergCatalogWrapper =
```
FANNG1, line 122: This is the entrypoint for Gravitino Iceberg catalog,   could you support IRC and add related test too?https://github.com/apache/gravitino/blob/80c9bd7502ce890c94126a6f3a558434aa8ef137/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergCatalogWrapperManager.java#L101-L104?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapperProxy.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.lang.reflect.Method;
+import net.sf.cglib.proxy.Enhancer;
+import net.sf.cglib.proxy.MethodInterceptor;
+import net.sf.cglib.proxy.MethodProxy;
+import org.apache.gravitino.iceberg.common.IcebergConfig;
+import org.apache.gravitino.iceberg.common.authentication.SupportsKerberos;
+import org.apache.iceberg.catalog.Catalog;
+
+public class IcebergCatalogWrapperProxy implements MethodInterceptor {
```
FANNG1: how about the following name?
  - KerberosAwareIcebergCatalogProxy (emphasizes why it exists)
  - ImpersonatingIcebergCatalogProxy (focus on user-impersonation role)
  - KerberosDelegatingIcebergCatalogProxy (highlights delegation token work)

yuqi1129: I will choose the first one as the concept of the last one is rather narrow, and the second one is not correct strictly.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -93,4 +274,18 @@ private void closeInternalClientPool() {
       LOGGER.warn("Failed to close HiveCatalog internal client pool", e);
     }
   }
+
+  private KerberosClient initKerberosClient() {
+    try {
+      KerberosClient kerberosClient = new KerberosClient(this.properties(), this.getConf());
+      // For Iceberg rest server, we haven't set the catalog_uuid, so we set it to 0 as there is
+      // only one catalog in the rest server, so it's okay to set it to 0.
```
FANNG1: There maybe multiple catalogs for IRC

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/SupportsKerberos.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Perform operations with Kerberos authentication.
+   *
+   * @param properties the catalog properties
+   * @param executable the operations to be performed
+   * @return the result of the operations
+   * @param <R> the return type of the operations
+   * @throws Throwable if any error occurs during the operations
+   */
+  <R> R doKerberosOperations(Map<String, String> properties, Executable<R> executable)
```
FANNG1: why adding properties to `doKerberosOperations `?, As the catalog have catalog properties already.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/KerberosAwareIcebergCatalogProxy.java
```diff
@@ -0,0 +1,83 @@
...
+import net.sf.cglib.proxy.MethodProxy;
+import org.apache.gravitino.iceberg.common.IcebergConfig;
+import org.apache.gravitino.iceberg.common.authentication.SupportsKerberos;
+import org.apache.iceberg.catalog.Catalog;
+
+/**
+ * A proxy class for IcebergCatalogWrapper to handle Kerberos authentication or other cross-cutting
+ * concerns.
+ */
+public class KerberosAwareIcebergCatalogProxy implements MethodInterceptor {
```
roryqi, line 34: If we only handle Iceberg catalog, we don't need use MethodInterceptor. I added the ProxyPlugin is to handle different kinds of catalogs. If we only handle one kind of catalog, we extend it and wrap it directly.

roryqi: Could u reuse the ProxyPlugin interface?

yuqi1129: This seems not to be a good idea as we also need to handle operations in the Iceberg REST, `ProxyPlugin` only works on REST API In Gravitino and does not cover `CatalogWrapperForREST`

roryqi: You can extract the common implementation. You can give a simple wrapper in the `ProxyPlugin`. It's not conflicted.

yuqi1129: It's not convenient enough.


The plugin in can be as the following code in Gravitino.
```
  @Override
  public Object doAs(
      Principal principal, Executable<Object, Exception> action, Map<String, String> properties)
      throws Throwable {
    IcebergCatalogWrapper wrapper = ops.getIcebergCatalogWrapper();
    if (wrapper.getCatalog() instanceof SupportsKerberos supportsKerberos) {
      return supportsKerberos.doKerberosOperations(() -> {
        try {
          return action.execute();
        } catch (Exception e) {
          throw new UndeclaredThrowableException(e);
        }
      });
    }

    return action.execute();
  }
```

However, we still need to intercept `IcebergCatalogWrapper` to support impersonation for `IcebergCatalogWrapperForREST`.

roryqi: We shouldn't break the design. You should treat the impersonation as the AOP unless we have strong reasons.

yuqi1129: What design? ` should treat the impersonation as the AOP`, In this PR, it's the AOP of `IcebergCatalogWrapper` NOT `BaseCatalog` as you did. In fact, we only need to handle iceberg catalogs with the Hive backend, do you mean we need to make it as the AOP of `BaseCatalog`?

roryqi: We should put the logic into `ProxyPlugin` if we need the impersionation. If we don't hope this, we should remove the interface. If this is the delicated to Hive backend. We should extend the class instead of dynamic proxy.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/integration/test/IcebergRestKerberosHiveWithUserImpersonationCatalogIT.java
```diff
@@ -0,0 +1,133 @@
...
+  @BeforeAll
+  void prepareSQLContext() {
+
+    // Change the ownership of /user/hive to normal user for user impersonation test. If we do not
+    // change the ownership, the normal user will not have the permission to create table in Hive
+    // as the /user/hive is owned by user `cli`, please see what's done in `initEnv` method in
+    // superclass.
+    containerSuite
+        .getKerberosHiveContainer()
+        .executeInContainer("hadoop", "fs", "-chown", "-R", NORMAL_USER, "/user/hive/");
```
FANNG1, line 55: Will this command infect other tests?

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalogOperations.java
```diff
@@ -113,7 +115,11 @@ public void initialize(
     resultConf.put("catalog_uuid", info.id().toString());
     IcebergConfig icebergConfig = new IcebergConfig(resultConf);
 
-    this.icebergCatalogWrapper = new IcebergCatalogWrapper(icebergConfig);
+    IcebergCatalogWrapper rawWrapper = new IcebergCatalogWrapper(icebergConfig);
+    this.icebergCatalogWrapper =
+        rawWrapper.getCatalog() instanceof SupportsKerberos
+            ? new KerberosAwareIcebergCatalogProxy(rawWrapper).getProxy(icebergConfig)
```
FANNG1, line 124: Could we create KerberosAwareIcebergCatalogProxy only when Kerberos is enabled? then when `doKerberosOperations` is invoked, we must do kerberos operations.
```
  public <R> R doKerberosOperations(Executable<R> executable) throws Throwable {
    Map<String, String> properties = this.properties();
    AuthenticationConfig authenticationConfig = new AuthenticationConfig(properties);
    if (!authenticationConfig.isKerberosAuth()) {
      return executable.execute();
    }
    xxx
}
```

FANNG1:
LGTM except minor comments

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/integration/test/IcebergRestKerberosHiveWithUserImpersonationCatalogIT.java
yuqi1129: I should not, as we will execute this command before each test.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalogOperations.java
yuqi1129: done

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/integration/test/IcebergRestKerberosHiveWithUserImpersonationCatalogIT.java
yuqi1129: This change only applies to the Kerberos Hive Docker image.

Issue #7804, opened 2025-07-25
[Bug report] If role is deleted, the role list in user details didn't updated.


danhuawang:
### Version

main branch

### Describe what's wrong

<img width="837" height="889" alt="Image" src="https://github.com/user-attachments/assets/fd5e1c95-c82e-474e-9afb-ac0cb65b698e" />

<img width="942" height="660" alt="Image" src="https://github.com/user-attachments/assets/b1c23864-add9-4d51-b301-1a44001a8d71" />

<img width="891" height="905" alt="Image" src="https://github.com/user-attachments/assets/cebf2657-69e4-4b04-aea2-1bff98e1edf1" />

### Error message and/or stacktrace

N/A

### How to reproduce

1. Create role1,role2,role3,role4 in metalake
2. Grant all the four roles to a user
3. Get the user info, check all roles are granted successfully.
4. Delete role3
5. Get the user info, check role3 if deleted in roles list two

PR #8190, opened 2025-08-20
[#7804] Fix(cache): Fix cache relation remove

xunliu:

Fix: #7804 

### DIFF
--- a/core/src/test/java/org/apache/gravitino/utils/TestUtil.java
+++ b/core/src/test/java/org/apache/gravitino/utils/TestUtil.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import org.apache.gravitino.Catalog;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.Namespace;
+import org.apache.gravitino.authorization.AuthorizationUtils;
 import org.apache.gravitino.authorization.Privileges;
 import org.apache.gravitino.authorization.SecurableObject;
 import org.apache.gravitino.file.Fileset;
@@ -389,6 +390,7 @@ public class TestUtil {
   public static UserEntity getTestUserEntity(
       long id, String name, String metalake, List<Long> roles) {
     return UserEntity.builder()
+        .withNamespace(AuthorizationUtils.ofUserNamespace(metalake))
         .withId(id)
         .withName(name)
         .withNamespace(NamespaceUtil.ofUser(metalake))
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -330,6 +349,32 @@ private <KEY, VALUE> Caffeine<KEY, VALUE> newBaseBuilder(Config cacheConfig) {
...
+    return nameIdent;
+  }
+
+  private EntityCacheKey makeEntityCacheKey(Entity entity) {
+    NameIdentifier nameIdent = getNameIdentifier(entity);
+    EntityCacheKey entityCacheKey = EntityCacheKey.of(nameIdent, entity.type());
+    return entityCacheKey;
+  }
+
+  public void putReverseIndex(Entity entity, EntityCacheRelationKey key) {
```
unknowntpo, line 364: All entities in cache has the type:  `<E extends Entity & HasIdentifier>`, not `Entity`.

So it should be like this:

```
  public <E extends Entity & HasIdentifier> void putReverseIndex(E entity, EntityCacheRelationKey key) {
```

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -228,7 +241,8 @@ public <E extends Entity & HasIdentifier> void put(E entity) {
         () -> {
           invalidateOnKeyChange(entity);
           NameIdentifier identifier = getIdentFromEntity(entity);
-          EntityCacheKey entityCacheKey = EntityCacheKey.of(identifier, entity.type());
+          EntityCacheRelationKey entityCacheKey =
+              EntityCacheRelationKey.of(identifier, entity.type());
```
jerryshao, line 245: How do you know that for some entity types, we need to update the entity when the related entity is updated/deleted? For example, if the role is deleted, we need to update the user entity, here I don't see the specific logic to handle this. I think we should have a specific logic to check the entity type and do some specific logic. It is hard to make it generic.

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
unknowntpo: After doing some research, I think we should keep this as is for now, because changing generic type parameter involves changing these file: `BaseEntityCache`, `CaffeineEntityCache`, `EntityCache`, `NoOpsCache`,`SupportsEntityStoreCache`, `SupportsRelationEntityCache`, `TestCaffeineEntityCache`, 

Let's focus on the correctness of cache for now.

Issue #8157, opened 2025-08-19
[Improvement] Fix exception handling in handle() in  RoleDetails.java


justinmclean:
### What would you like to be improved?

In clients/cli/src/main/java/org/apache/gravitino/cli/commands/RoleDetails.java

```
    try {
      GravitinoClient client = buildClient(metalake);
      objects = client.getRole(role).securableObjects();
    } catch (NoSuchMetalakeException err) {
      exitWithError(ErrorMessages.UNKNOWN_METALAKE);
    } catch (NoSuchUserException err) {
      exitWithError(ErrorMessages.UNKNOWN_GROUP);
    } catch (Exception exp) {
      exitWithError(exp.getMessage());
    }
```

### How should we improve?

The NoSuchUserException is exiting with error ErrorMessages.UNKNOWN_GROUP, but should be about the role.

pandeysambhi:
@justinmclean Can I take this up ?

justinmclean:
Yes, as usual, if you need a hand just ask.

PR #8191, opened 2025-08-20
[#8157]Improvement: exception handling in RoleDetails for Unknown Role

pandeysambhi:
Exception handing for NoSuchUserException for Unknown Role

The NoSuchUserException is exiting with error ErrorMessages.UNKNOWN_GROUP, but should be about the role.

Fix: #8157 

Unit testing


clients/cli/src/main/java/org/apache/gravitino/cli/commands/RoleDetails.java
```diff
@@ -56,7 +56,7 @@ public void handle() {
     } catch (NoSuchMetalakeException err) {
       exitWithError(ErrorMessages.UNKNOWN_METALAKE);
     } catch (NoSuchUserException err) {
```
justinmclean: Are you sure NoSuchUserException is correct?

pandeysambhi: I thought it should be NoSuchRoleException, but saw that issue mentioned only the message update. I should have asked to clarify though. In order to improve this, should I keep what was there and add an additional block for NoSuchRoleException or replace the user exception with role?

justinmclean: I'd double check what exception are being thrown, but I'd guess that NoSuchUserException is not needed.

pandeysambhi: sure, let me know and I can make the necessary changes

justinmclean: Sorry, I was a little unclear there. I'm asking you to double-check.

pandeysambhi: Sure, let me take a look again

roryqi: NoSuchUserException is unnecessary.

pandeysambhi: updated the PR, please have a look.

justinmclean (approved):
Thanks for fixing this.

Issue #8161, opened 2025-08-19
[Improvement] Fix removeCatalog to respect the ignoreIfNotExists flag


justinmclean:
### What would you like to be improved?

removeCatalog in flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStore.java doesn't use the ignoreIfNotExists flag

### How should we improve?

Only throw an error only when a missing catalog is not meant to be ignored. It might be a good idea to check the return of dropCatalog as well.

Surya00011:
Hi, @justinmclean, I would like to work on this

justinmclean:
all yours

justinmclean:
This looks good, but it would be great if you could add some unit tests.

Surya00011:
Thank you for letting me know, sir @justinmclean. I'll add the unit tests and raise a PR.

PR #8211, opened 2025-08-20
[#8161] fix(catalog): respect ignoreIfNotExists flag in removeCatalog

Surya00011:

- Updated `removeCatalog` implementation to respect the `ignoreIfNotExists` flag.
- If the catalog does not exist and `ignoreIfNotExists` is true, the method now exits silently.
- If the catalog does not exist and `ignoreIfNotExists` is false, a `CatalogException` is thrown.
- Wrapped underlying errors in `CatalogException` for consistency.

- The previous implementation ignored the `ignoreIfNotExists` flag.
- This led to unexpected exceptions even when users explicitly indicated they wanted missing catalogs to be ignored.
- Fixing this ensures more predictable behavior and aligns with the contract of the API.

Fix: #8161

Yes.  
- Users who set `ignoreIfNotExists = true` will no longer see unnecessary exceptions if the catalog is missing.  
- Behavior is now consistent with expected API semantics.

### DIFF
--- a/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStore.java
+++ b/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStore.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.function.Predicate;
+
 import org.apache.flink.configuration.Configuration;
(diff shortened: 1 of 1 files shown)


justinmclean:
I'm fairly certain Spotless will fail because it appears you've changed a lot of whitespace. Did you run Spotless on the code before making the PR? We use Spotless to ensure that the code is consistently formatted.

justinmclean:
Yes, the Spotless check failed. You'll need to run `./gradlew :flink-connector:flink:spotlessApply` to correctly format your changes and then commit those changes.

Issue #4874, opened 2024-09-06
[Bug report] Missing comment of the struct field when creating table


LauraXia123:
### Version

main branch

### Describe what's wrong

```
curl 'http://0.0.0.0:3002/api/metalakes/metalake_demo/catalogs/catalog_hive/schemas/sales/tables' \
  -H 'Accept: application/vnd.gravitino.v1+json' \
  -H 'Accept-Language: zh-CN,zh;q=0.9' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json;charset=UTF-8' \
  -H 'Origin: http://0.0.0.0:3002' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://0.0.0.0:3002/catalogs/catalog_hive/sales?catalogType=relational' \
  --data-raw '{"name":"table7","comment":"","columns":[{"name":"structComment","type":{"type":"struct","fields":[{"name":"field","type":"binary","nullable":true,"comment":"xxx1111"}]},"nullable":true,"comment":""}],"properties":{}}' \
  --insecure
```
get the response
`{
    "code": 0,
    "table": {
        "name": "table7",
        "comment": "",
        "columns": [
            {
                "name": "structcomment",
                "type": {
                    "type": "struct",
                    "fields": [
                        {
                            "name": "field",
                            "type": "binary",
                            "nullable": true
                        }
                    ]
                },
                "comment": "",
                "nullable": true,
                "autoIncrement": false
            }
        ],
        "properties": {
            "input-format": "org.apache.hadoop.mapred.TextInputFormat",
            "transient_lastDdlTime": "1725612072",
            "output-format": "org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat",
            "location": "hdfs://hive:9000/user/hive/warehouse/sales.db/table7",
            "table-type": "MANAGED_TABLE",
            "serde-lib": "org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe",
            "serde-name": "table7"
        },
        "audit": {
            "creator": "anonymous",
            "createTime": "2024-09-06T08:41:12.419Z"
        },
        "distribution": {
            "strategy": "none",
            "number": 0,
            "funcArgs": []
        },
        "sortOrders": [],
        "partitioning": [],
        "indexes": []
    }
}`

Missing the comment of the struct field
[...]

PR #8212, opened 2025-08-20
[#4874] improvement(hive): add validation for Hive struct data type

mchades:

add validation for Hive struct data type

Fix: #4874 

no

tests added

### DIFF
--- a/catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveDataTypeConverter.java
+++ b/catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveDataTypeConverter.java
@@ -41,6 +41,7 @@ import static org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.getUnionTyp
 import static org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.getVarcharTypeInfo;
 import static org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils.getTypeInfoFromTypeString;
 
+import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/converter/HiveDataTypeConverter.java
```diff
@@ -116,7 +117,17 @@ public TypeInfo fromGravitino(Type type) {
...
+                    t -> {
+                      // Hive does not support comments in struct fields. This is no way to preserve
+                      // comments in struct fields when converting from Gravitino to Hive.
+                      // See: https://issues.apache.org/jira/browse/HIVE-26593
+                      Preconditions.checkArgument(
+                          t.comment() == null,
+                          "Hive does not support comments in struct fields: %s",
+                          t.name());
+                      return fromGravitino(t.type());
+                    })
```
jerryshao, line 130: Compared to throwing an exception. In my opinion, it is better to add a warning log and ignore the comment field for Hive table. What do you think?

Directly throwing an exception and failing the request seems too strict.

mchades: I'm afraid this will cause confusion for the user, as the field is set without errors but doesn't work when loaded (as the issue mentioned #4874 ). Sometimes the user cannot access the server log, which will take time to explain.

jerryshao: Can you please also update the related docs?

mchades: fixed

jerryshao (approved):

Issue #8182, opened 2025-08-19
[Bug report] A user can load model, but this user can't list all the model versions


danhuawang:
### Version

main branch

### Describe what's wrong

userC can load the model
<img width="758" height="707" alt="Image" src="https://github.com/user-attachments/assets/2b8f97b5-6914-4997-804b-42a8f644519b" />

but userC list version , it return nothing
<img width="810" height="559" alt="Image" src="https://github.com/user-attachments/assets/4786f342-1252-40ae-a660-cf39391ce2a5" />

I switch to "anonymous" user, it return 4 verison

<img width="762" height="649" alt="Image" src="https://github.com/user-attachments/assets/a26847cb-2712-416c-90f8-f66d2698d1c6" />

### Error message and/or stacktrace

N/A

### How to reproduce

1. create model catalog `model_catalog_1`,schema `model_schema_1 `, model `model_070 `, version `0`,`1`,`2`,`3`
2. create roleC as following, create necessary catalogs  that the role needed.
```
{
"name": "roleC",
"properties": {
"k1": "v1"
},
"securableObjects": [
{
"type": "CATALOG",
"privileges": [
{
"name": "USE_SCHEMA",
"condition": "ALLOW"
},
{
"name": "USE_CATALOG",
"condition": "ALLOW"
},
{
"name": "READ_FILESET",
"condition": "ALLOW"
},
{
"name": "WRITE_FILESET",
"condition": "ALLOW"
}
],
"fullName": "fileset_catalog_1"
},
{
"type": "CATALOG",
"privileges": [  
{
"name": "USE_SCHEMA",
"condition": "ALLOW"
},
{
"name": "USE_CATALOG",
"condition": "ALLOW"
},
{
"name": "select_table",
"condition": "ALLOW"
}
],
"fullName": "mysql_1"
},
{
"type": "CATALOG",
"privileges": [
{
"name": "USE_SCHEMA",
"condition": "ALLOW"
},
{
"name": "USE_CATALOG",
"condition": "ALLOW"
},
{
"name": "USE_MODEL",
"condition": "ALLOW"
}
],
"fullName": "model_catalog_1"
}
]
}
```
3. add userC to metalake
4. grant roleC to userC
5. use "userC" to load model 
6. use "userC" to list model version , expect to see all the versions

danhuawang:
@hdygxsj Could you help check this issue?

hdygxsj:
[...]

PR #8223, opened 2025-08-20
[#8182] bug(authz): fix list model version bug

hdygxsj:

fix list model version bug

Fix: #8182

None

org.apache.gravitino.client.integration.test.authorization.ModelAuthorizationIT

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataFilterHelper.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataFilterHelper.java
@@ -172,6 +172,15 @@ public class MetadataFilterHelper {
         nameIdentifierMap.put(
             Entity.EntityType.CATALOG, NameIdentifierUtil.getCatalogIdentifier(nameIdentifier));
         break;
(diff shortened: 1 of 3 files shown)


hdygxsj:
@jerqi PHAL, thanks

server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataFilterHelper.java
```diff
@@ -172,6 +172,15 @@ private static Map<Entity.EntityType, NameIdentifier> spiltMetadataNames(
         nameIdentifierMap.put(
             Entity.EntityType.CATALOG, NameIdentifierUtil.getCatalogIdentifier(nameIdentifier));
         break;
+      case MODEL_VERSION:
```
roryqi, line 175: Maybe we should throw Exceptions when default cases to avoid forgetting adding new entityTypes.

hdygxsj: fixed

jerqi (approved):

Issue #8159, opened 2025-08-19
[Improvement] Extend CLI parsing to support more complex types


justinmclean:
### What would you like to be improved?

The CLI can support parsing of more complex types e.g. maps contain maps or lists.

Here are some tests to help:
```
  @Test
  public void testParseTypeNestedList() {
    Type type = ParseType.toType("list(list(integer))");
    assertThat(type, instanceOf(Types.ListType.class));
    Type innerList = ((Types.ListType) type).elementType();
    assertThat(innerList, instanceOf(Types.ListType.class));
    Type elementType = ((Types.ListType) innerList).elementType();
    assertThat(elementType, instanceOf(Types.IntegerType.class));
  }

  @Test
  public void testParseTypeMapWithNestedValue() {
    Type type = ParseType.toType("map(string,list(integer))");
    assertThat(type, instanceOf(Types.MapType.class));
    Type keyType = ((Types.MapType) type).keyType();
    Type valueType = ((Types.MapType) type).valueType();
    assertThat(keyType, instanceOf(Types.StringType.class));
    assertThat(valueType, instanceOf(Types.ListType.class));
    Type elementType = ((Types.ListType) valueType).elementType();
    assertThat(elementType, instanceOf(Types.IntegerType.class));
  }
```

### How should we improve?

Note, while this might seem at first complex, it is very simple to do and only requires changing a couple of lines of code.

keepConcentration:
@justinmclean 

Hi, Could you please assign this issue to me?

PR #8227, opened 2025-08-20
[#8159] improvement(cli): Use toType() instead of toBasicType() for support more complex types

keepConcentration:

This PR updates the `ParseType` class to support parsing of nested complex types (e.g. `list(list(integer))`, `map(string,list(integer))`).  

Currently, the CLI only supports simple types or single-level lists/maps.  
It fails for more complex definitions like `list(list(integer))` or `map(string,list(integer))`.  

Fix: #8159

Users can now define more complex schemas in the CLI, such as:
- `list(list(integer))`
- `map(string,list(integer))`

- Added new unit tests:
  - `testParseTypeNestedList`
  - `testParseTypeMapWithNestedValue`
- Executed existing unit tests


justinmclean (approved):
Thanks for fixing that.

Issue #7987, opened 2025-08-08
[Subtask] Add Lance Storage for partition statistics


jerqi:
### Describe the subtask

Add Lance Storage for partition statistics

### Parent issue

#7268

PR #8228, opened 2025-08-20
[#7987] feat(core): Add Lance Storage for partition statistics

jerqi:

Add Lance Storage for partition statistics

Fix: #7987 

I will add the document later.

Added UT.


.github/workflows/build.yml
```diff
@@ -125,6 +124,11 @@ jobs:
         run: |
           dev/ci/util_free_space.sh
 
+      - name: Install dependencies
```
roryqi, line 130: Lancedb depends on libssl1.1

roryqi:
I will rebase the PR about integration tests.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,415 @@
...
+import org.apache.gravitino.stats.PartitionStatisticsUpdate;
+import org.apache.gravitino.stats.StatisticValue;
+import org.apache.gravitino.utils.MetadataObjectUtil;
+import org.apache.gravitino.utils.PrincipalUtils;
+
+/** LancePartitionStatisticStorage is based on Lance format files. */
+public class LancePartitionStatisticStorage implements PartitionStatisticStorage {
+
+  private final String LOCATION = "location";
+  private final String DEFAULT_LOCATION = "/tmp";
```
jerryshao: Do you support remote storage like hdfs/s3?

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,415 @@
...
+  private final String MAX_ROWS_PER_FILE = "maxRowsPerFile";
+  private final int DEFAULT_MAX_ROWS_PER_FILE = 1000000; // 10M
+  private final String MAX_BYTES_PER_FILE = "maxBytesPerFile";
+  private final int DEFAULT_MAX_BYTES_PER_FILE = 100 * 1024 * 1024; // 100 MB
+  private final String MAX_ROWS_PER_GROUP = "maxRowsPerGroup";
+  private final int DEFAULT_MAX_ROWS_PER_GROUP = 1000000; // 1M
+  private final String READ_BATCH_SIZE = "readBatchSize";
+  private final int DEFAULT_READ_BATCH_SIZE = 10000; // 10K
```
jerryshao: Make all of them static.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,415 @@
...
+        Integer.parseInt(
+            properties.getOrDefault(
+                MAX_BYTES_PER_FILE, String.valueOf(DEFAULT_MAX_BYTES_PER_FILE)));
+    this.maxRowsPerGroup =
+        Integer.parseInt(
+            properties.getOrDefault(
+                MAX_ROWS_PER_GROUP, String.valueOf(DEFAULT_MAX_ROWS_PER_GROUP)));
+    this.readBatchSize =
+        Integer.parseInt(
+            properties.getOrDefault(READ_BATCH_SIZE, String.valueOf(DEFAULT_READ_BATCH_SIZE)));
```
jerryshao: You should check the validity of all these parameters.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: Lance Rust JNI has supported S3.  HDFS is supporting in the community. It will use Rust JNI, too. https://github.com/lancedb/lance/issues/2140

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: Added modifier static.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: I added check to make sure that all of them are positive.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,427 @@
...
+import org.apache.gravitino.stats.PartitionStatisticsModification;
+import org.apache.gravitino.stats.PartitionStatisticsUpdate;
+import org.apache.gravitino.stats.StatisticValue;
+import org.apache.gravitino.utils.MetadataObjectUtil;
+import org.apache.gravitino.utils.PrincipalUtils;
+
+/** LancePartitionStatisticStorage is based on Lance format files. */
+public class LancePartitionStatisticStorage implements PartitionStatisticStorage {
+
+  private static final String LOCATION = "location";
```
jerryshao, line 71: How to set such configurations from Gravitino, what is the full name of the configuration?

roryqi: The full name is `gravitino.stats.partition.option.location`.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,427 @@
...
+                + statistics.stream().map(str -> "'" + str + "'").collect(Collectors.joining(", "))
+                + ")");
+      }
+
+      if (partitionSQLs.size() == 1) {
+        dataset.delete(partitionSQLs.get(0));
+      } else if (partitionSQLs.size() > 1) {
+        String filterSQL =
+            partitionSQLs.stream().map(str -> "(" + str + ")").collect(Collectors.joining(" OR "));
+        dataset.delete(filterSQL);
```
jerryshao, line 308: Do you have code to cover this case?

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,427 @@
...
+              String partitionName = new String(partitionNameVector.get(i), StandardCharsets.UTF_8);
+              String statisticName = new String(statisticNameVector.get(i), StandardCharsets.UTF_8);
+              String statisticValueStr =
+                  new String(statisticValueVector.get(i), StandardCharsets.UTF_8);
+              String auditInoStr = new String(auditInfoNameVector.get(i), StandardCharsets.UTF_8);
+
+              StatisticValue<?> statisticValue =
+                  JsonUtils.anyFieldMapper().readValue(statisticValueStr, StatisticValue.class);
+              AuditInfo auditInfo =
+                  JsonUtils.anyFieldMapper().readValue(auditInoStr, AuditInfo.class);
```
jerryshao, line 397: Should you handle the json deserialize exception?

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -0,0 +1,427 @@
...
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  private Dataset open(String fileName) {
+    try {
+      return Dataset.open(fileName, allocator);
+    } catch (IllegalArgumentException illegalArgumentException) {
+      if (illegalArgumentException.getMessage().contains("was not found")) {
```
jerryshao, line 422: Do we have better way to check if the lance file exists or not? Using exception message is too tricky.

roryqi: Lance Java SDK doesn't provide the API to check the file exists now.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
jerryshao: You'd better change to gravitino.partitionStats.lance.xxx, "lance" is a dynamic parameter for different storages.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: The method throws `JsonProcessExeption`. It belongs to IOException. We have thrown IOException to upper layer to process. It's ok that we don't process this Exception here if you don't have other preference.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: Added new test cases for this.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: Our little concern is whether it  is weird that the middle key words uses camel style or not.

roryqi: Changed to `gravitino.stats.partition.lance.location`.

core/src/main/java/org/apache/gravitino/stats/StatisticManager.java
```diff
@@ -58,7 +58,7 @@
 
 public class StatisticManager implements Closeable {
 
-  private static final String OPTIONS_PREFIX = "gravitino.stats.partition.option.";
+  private static final String OPTIONS_PREFIX = "gravitino.partition.stats.";
```
jerryshao: I think it is better to change to `gravitino.stats.partition.*` to align with "gravitino.stats.partition.storageFactoryClass".

jerryshao: Also, it is better to fix the code in `StatisticManager` to filter out the configurations for the specific storage:

```
    String className = config.get(Configs.PARTITION_STATS_STORAGE_FACTORY_CLASS);
    Map<String, String> options = config.getConfigsWithPrefix(OPTIONS_PREFIX);
    try {
      PartitionStatisticStorageFactory factory =
          (PartitionStatisticStorageFactory)
              Class.forName(className).getDeclaredConstructor().newInstance();
      this.partitionStorage = factory.create(options);
```

So the specific storage doesn't need to take care the prefix like `lance.`. What do you think?

jerryshao: With this, you need to introduce another configuration like `gravitino.stats.partition.storage` = `lance` to specify the storage name.

roryqi: It's a little complex. We only have one storage factory. Do we really need the prefix?

jerryshao: How about changing to `gravitino.stats.partition.storageOption.`?

roryqi: OK.

jerryshao (approved):

PR #8231, opened 2025-08-20
Update README.md

SameerSalman29:

Fix: #(issue)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+[![Build Status](https://github.com/SameerSalman29/gravitino/actions/workflows/main.yml/badge.svg)](https://github.com/SameerSalman29/gravitino/actions)
+[![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](LICENSE)
+[![Open Issues](https://img.shields.io/github/issues/SameerSalman29/gravitino)](https://github.com/SameerSalman29/gravitino/issues)
+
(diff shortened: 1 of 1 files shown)


justinmclean:
Hi, thanks for the contribution.

However, my view is that the README does not need a manual Table of Contents. GitHub already auto-generates a TOC for any Markdown file with multiple headings, which covers all the navigation needs. Here's what it looks like:
image

Since the Gravitino README isn’t excessively long, the built-in navigation is most likely sufficient. Adding a manual TOC would create extra maintenance without adding much value. For longer files like [CONTRIBUTING.md](https://github.com/apache/gravitino/blob/main/CONTRIBUTING.md), it does make sense to have one.

README.md
```diff
@@ -1,3 +1,7 @@
+[![Build Status](https://github.com/SameerSalman29/gravitino/actions/workflows/main.yml/badge.svg)](https://github.com/SameerSalman29/gravitino/actions)
+[![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](LICENSE)
+[![Open Issues](https://img.shields.io/github/issues/SameerSalman29/gravitino)](https://github.com/SameerSalman29/gravitino/issues)
+
```
justinmclean, line 4: These link to SameerSalman29, we would not add those.

Issue #8203, opened 2025-08-20
[Improvement] Possible NPE in CreateTagFailureEvent.java


justinmclean:
### What would you like to be improved?

Guard against a null TagInfo in CreateTagFailureEvent by only constructing a NameIdentifier when TagInfo is present, preventing a NullPointerException.

Unit test to show the issue:
```
  @Test
  public void testNullTagInfoDoesNotThrow() {
    Exception ex = new Exception("boom");
    CreateTagFailureEvent event = new CreateTagFailureEvent("user", "metalake", null, ex);
    Assertions.assertNull(event.tagInfo());
    Assertions.assertNull(event.identifier());
    Assertions.assertEquals(ex, event.exception());
    Assertions.assertEquals(OperationType.CREATE_TAG, event.operationType());
    Assertions.assertEquals(OperationStatus.FAILURE, event.operationStatus());
  }
```

### How should we improve?

Prevent the NPE.

PCloud63514:
Hi @justinmclean Would it be alright if I work on this issue?

justinmclean:
Thanks for your interest in contributing to Apache Gravitino! 🎉 Please see our [CONTRIBUTING](https://github.com/apache/gravitino/blob/main/CONTRIBUTING.md) guide. For code contributions, it explains how to compile with Gradle, run unit tests (unit tests should cover all code changes), and format code with Spotless before submitting a PR.

PCloud63514:
> @PCloud63514 , thanks for your contribution, is generated by Gravitino internally, and the couldn't be null, so I prefer not to add an extra check for it, WDYT?`CreateTagFailureEvent``tagInfo`


@justinmclean Hello, I’m leaving a comment after reviewing this PR!

From what I can tell, there doesn’t seem to be any situation where targetInfo would be empty in the process.
Of course, I understand that even if targetInfo isn’t empty now, it could still happen due to a developer’s mistake during development or testing.

When you created this issue, was there something specific you had in mind, or any advice you could share with me?

justinmclean:
I suggest you take a look at the test, passing a null TagInfo results in an RTE when it tries tagInfo.name() in the CreateTagFailureEvent constructor.

```
  public CreateTagFailureEvent(String user, String metalake, TagInfo tagInfo, Exception exception) {
    super(user, NameIdentifierUtil.ofTag(metalake, tagInfo.name()), exception);
    this.tagInfo = tagInfo;
  }
```

[...]

PR #8233, opened 2025-08-21
[#8203] fix: prevent NPE in CreateTagFailureEvent when TagInfo is null

PCloud63514:

- Add null guard in `CreateTagFailureEvent` constructor.  
- Do not construct a `NameIdentifier` when `TagInfo` is null.  
- Ensure `tagInfo()` and `identifier()` return `null` consistently if no `TagInfo` is provided.  
- Added a unit test `testNullTagInfoDoesNotThrow` to verify expected behavior.

- Previously, passing a null `TagInfo` caused a `NullPointerException` in the constructor when creating the `NameIdentifier`.  
- This patch prevents the unexpected NPE and ensures that failure events can still be created safely without tag information.  

Fix: #8203

- No API changes.  
- Behavioral change: `CreateTagFailureEvent` no longer throws NPE if `TagInfo` is null.  
- In such cases, both `tagInfo()` and `identifier()` return `null`.

- Added a new test in `CreateTagFailureEventTest` verifying null `TagInfo` does not throw.  
- Verified that events with valid `TagInfo` still construct a proper `NameIdentifier`.  
- Ensured all existing tests continue to pass.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/api/event/CreateTagFailureEvent.java
+++ b/core/src/main/java/org/apache/gravitino/listener/api/event/CreateTagFailureEvent.java
@@ -40,7 +40,10 @@ public class CreateTagFailureEvent extends TagFailureEvent {
    *     insights into the reasons behind the operation's failure.
    */
   public CreateTagFailureEvent(String user, String metalake, TagInfo tagInfo, Exception exception) {
-    super(user, NameIdentifierUtil.ofTag(metalake, tagInfo.name()), exception);
+    super(
+        user,
+        (tagInfo == null ? null : NameIdentifierUtil.ofTag(metalake, tagInfo.name())),
+        exception);
     this.tagInfo = tagInfo;
   }
 
--- a/core/src/test/java/org/apache/gravitino/listener/api/event/TestTagEvent.java
+++ b/core/src/test/java/org/apache/gravitino/listener/api/event/TestTagEvent.java
@@ -19,14 +19,7 @@
 
 package org.apache.gravitino.listener.api.event;
 
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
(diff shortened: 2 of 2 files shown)


jerryshao:
@FANNG1 please help to review.

FANNG1:
@PCloud63514 , thanks for your contribution,  `CreateTagFailureEvent` is generated by Gravitino internally, and the `tagInfo` couldn't be null, so I prefer not to add an extra check for it, WDYT?

PCloud63514:
@FANNG1 Thanks for your feedback.
I initially added this to prevent a possible NPE, but on second thought, I realize I didn’t fully consider in what cases TagInfo might not be provided. 😅  
From what I can see, there doesn’t seem to be any case where TagInfo would be empty in the current flow. 
Could there be something I’m overlooking?

FANNG1:
> @FANNG1 Thanks for your feedback. I initially added this to prevent a possible NPE, but on second thought, I realize I didn’t fully consider in what cases TagInfo might not be provided. 😅 From what I can see, there doesn’t seem to be any case where TagInfo would be empty in the current flow. Could there be something I’m overlooking?

Sorry, I don't have other good first issues for now : (

PCloud63514:
@FANNG1 I shared these thoughts with the issue author and asked for their advice. I found the phrase “cheap insurance against hard-to-debug NPEs” very meaningful.

Also, since the NameIdentifier constructor internally checks the name, I believe it’s better to rely on the constructor’s responsibility for data integrity rather than causing an early NPE.

jerryshao:
It is the internal code (not for the users). The developers should guarantee that it will never be null. If there's a null value, something must be wrong; the fix here tries to hide the issue, which I think is not proper.

PCloud63514:
Thanks everyone for the discussion.
This PR was opened based on the issue reporter’s concern about a potential NPE.
After reviewing the feedback, the general agreement is that a null check is unnecessary.
I’ll proceed to close this PR in line with that consensus.

Issue #8205, opened 2025-08-20
[FEATURE]  Support multi location in fileset


blackfaced:
### Describe the feature

When debugging a fileset locally using the Python client as described in Gravitino documentation, I found that the s3_endpoint option in the options dictionary does not take effect. Even after setting s3_endpoint to an external address, the client still tries to connect to the internal endpoint.
For example, I set:
```python
options = {
"s3_endpoint": "https://external-s3-endpoint.com/",
...
}
```
But the error message shows the client is still trying to connect to the internal endpoint:
Connect timeout on endpoint URL: "https://internal-s3-endpoint.com/xxxxxx"

it seems that gravitino uses catalog_props first, then check for options

<img width="2396" height="866" alt="Image" src="https://github.com/user-attachments/assets/0da2a267-f82b-48b4-b83b-2df397320f99" />

I hope that there is a way to override the s3_endpoint in catalog_props, so that I can use it in different environment. 

### Motivation

on cloud environment, there is inner endpoint and public endpoint, inner endpoint is fast and cheap, public endpoint is slow and expensive, both has use cases

jerryshao:
@mchades can you please take a look?

mchades:
I think the client options should take precedence over the catalog props and server-side configurations, WDYT? @yuqi1129

yuqi1129:
> I think the client options should take precedence over the catalog props and server-side configurations, WDYT? [@yuqi1129](https://github.com/yuqi1129)

It should be fine with me and seems to be more reasonable.

yuqi1129:
<img width="1118" height="528" alt="Image" src="https://github.com/user-attachments/assets/55c8e8ab-3f84-48f4-a4e8-0e4572dfb7d3" />

Java GVFS is fine. Configurations from clients' sides will override those from the Gravitino server, so there is no need to make  any changes to Java GVFS.

PR #8236, opened 2025-08-21
[#8205] improvement(client-python): Make GVFS configuration prior to that of catalogs from Gravitino server.

yuqi1129:

Configurations from GVFS clients will override those from Gravitino server.

Let the user customize their own settings.

Fix: #8205 

N/A.

UT and existing tests.


clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
mchades, line 1: Other cloud storage options should also be considered.

yuqi1129: I have checked it. For Azure/GCS/HDFS, client configurations are always needed and there is no need to make any modifications.

mchades (approved):
lgtm.

PR #8238, opened 2025-08-21
[#8168] Fix NPE in EntityCombinedFileset by initializing hiddenProperties field

### LINKED ISSUE #8168: [Improvement] Possible Null Pointer Exception in TestEntityCombinedFileset.java

An NPE can potentially occur as hiddenProperties is not initialised.

Here is a unit test showing that:
```
  @Test
  void testPropertiesWithoutHiddenProperties() {
    Fileset fileset = Mockito.mock(Fileset.class);
    ImmutableMap<String, String> properties =
        ImmutableMap.of("propA", "valueA", "propB", "valueB");
    Mockito.when(fileset.properties()).thenReturn(properties);

    EntityCombinedFileset entityCombinedFileset = EntityCombinedFileset.of(fileset);

    Assertions.assertEquals(properties, entityCombinedFileset.properties());
  }
```

### How should we improve?

initialise hiddenProperties

JeonDaehong:
In `EntityCombinedFileset.java`, the `properties()` method directly called `hiddenProperties.contains()` without null checking, which could throw a `NullPointerException` if `hiddenProperties` was not initialized. This PR updates the code to initialize `hiddenProperties` with an empty set and adds null-safe handling in the `withHiddenProperties()` method.

Without this fix, calling `properties()` on an `EntityCombinedFileset` instance before setting hidden properties could cause NPEs, breaking fileset property retrieval. Adding proper initialization ensures that the object is always in a valid state and handles missing hidden properties gracefully.

Fixes #8168

No. Valid requests behave the same. Cases where `hiddenProperties` was not explicitly set are now safely handled without throwing exceptions.

* Added unit tests to verify that calling `properties()` without setting hidden properties does not cause exceptions
* Added tests to ensure null input to `withHiddenProperties()` is handled safely
* Added tests to verify that hidden property filtering works correctly when properties are set
* Verified that existing functionality remains unchanged

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/EntityCombinedFileset.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/EntityCombinedFileset.java
@@ -18,6 +18,7 @@
  */
 package org.apache.gravitino.catalog;
 
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -33,7 +34,7 @@ public final class EntityCombinedFileset implements Fileset {
   private final FilesetEntity filesetEntity;
 
   // Sets of properties that should be hidden from the user.
-  private Set<String> hiddenProperties;
+  private Set<String> hiddenProperties = Collections.emptySet();
 
   private EntityCombinedFileset(Fileset fileset, FilesetEntity filesetEntity) {
     this.fileset = fileset;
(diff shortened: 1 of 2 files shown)


JeonDaehong:
@justinmclean 

Hello, I’ve submitted the PR. Please review it.

JeonDaehong:
@justinmclean 
Hello! The tests have passed. If there are no issues, I kindly request a merge.

Issue #8218, opened 2025-08-20
[Bug report] Doris Bucket Not Number


an-shi-chi-fan:
### Version

0.9.0

### Describe what's wrong

DISTRIBUTED BY RANDOM BUCKETS AUTO
PROPERTIES 

if buckets is auto the function
```java
public static Distribution extractDistributionInfoFromSql(String createTableSql)
```
will throw exception

### Error message and/or stacktrace

```java

Failed to operate table(s) [orders] operation [LOAD] under schema [xxxx], reason [Cannot parse null string]","stack":["java.lang.NumberFormatException: Cannot parse null string","\tat java.base/java.lang.Integer.parseInt(Integer.java:630)","\tat java.base/java.lang.Integer.valueOf(Integer.java:999)","\tat org.apache.gravitino.catalog.doris.utils.DorisUtils.extractDistributionInfoFromSql(DorisUtils.java:208)","\tat org.apache.gravitino.catalog.doris.operation.DorisTableOperations.getDistributionInfo(DorisTableOperations.java:791)","\tat org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations.load(JdbcTableOperations.java:226)","\tat org.apache.gravitino.catalog.jdbc.JdbcCatalogOperations.loadTable(JdbcCatalogOperations.java:332)","\tat
```

### How to reproduce

support buckets is auto value

mbaurin:
Hey ! Just squashed that annoying Doris bug for you. The whole DISTRIBUTED BY RANDOM BUCKETS AUTO thing was throwing NumberFormatException: Cannot parse null string because of a silly regex issue in the parsing logic

an-shi-chi-fan:
thanks, which release version will contains this fix

PR #8240, opened 2025-08-21
[#8218] fix(doris): Fix BUCKETS AUTO parsing for RANDOM distribution

mbaurin:

Fixed the `DorisUtils.extractBucketNum()` method to properly handle `BUCKETS AUTO` parsing for `DISTRIBUTED BY RANDOM` statements. The method was incorrectly calling `matcher.find(5)` which advances the matcher position, causing the bucket value extraction to fail and resulting in a `NumberFormatException` when parsing null values.

**Changes:**
- Changed `if (matcher.find(5))` to `if (matcher.group(5) != null)` in `DorisUtils.extractBucketNum()`
- Added test case for `DISTRIBUTED BY RANDOM BUCKETS AUTO` in `TestDorisUtils.testDistributedInfoPattern()`

This fixes a critical bug where Doris table operations fail when encountering `DISTRIBUTED BY RANDOM BUCKETS AUTO` syntax. The root cause was that `matcher.find(5)` advances the matcher position after the initial match, and since there are no subsequent matches, it returns `false` and never reaches the bucket value extraction logic. This leaves `bucketValue` as null, causing `Integer.valueOf(bucketValue)` to throw `NumberFormatException: Cannot parse null string`.

The bug affects users trying to load tables with auto bucket distribution using random strategy, which is a valid Doris SQL syntax.

  Fix: #8218 

No user-facing API changes. This is a bug fix that enables proper parsing of existing Doris SQL syntax that was previously failing. Users can now successfully work with tables using `DISTRIBUTED BY RANDOM BUCKETS AUTO` without encountering parsing errors.

1. **Existing tests**: Verified all existing Doris utility tests continue to pass, ensuring no regression
2. **New test case**: Added `TestDorisUtils.testDistributedInfoPattern()` test case specifically for `DISTRIBUTED BY RANDOM BUCKETS AUTO` to prevent future regressions
3. **Manual verification**: Created and ran a standalone test to reproduce the original bug and verify the fix resolves the `NumberFormatException`
4. **Code formatting**: Applied Spotless formatting to ensure code style compliance

**Test command:**
```
./gradlew :catalogs:catalog-jdbc-doris:test --tests="*TestDorisUtils*"
```

All tests pass successfully with the fix applied.

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/converter/DorisExceptionConverter.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/converter/DorisExceptionConverter.java
@@ -49,6 +49,7 @@ public class DorisExceptionConverter extends JdbcExceptionConverter {
   static final int CODE_OTHER = 1105;
   static final int CODE_DELETE_NON_EXISTING_PARTITION = 1507;
   static final int CODE_PARTITION_ALREADY_EXISTS = 1517;
+  static final int CODE_BUCKETS_AUTO_NOT_SUPPORTED = 1064;
 
   private static final String DATABASE_ALREADY_EXISTS_PATTERN_STRING =
       ".*?detailMessage = Can't create database '.*?'; database exists";
@@ -83,6 +84,12 @@ public class DorisExceptionConverter extends JdbcExceptionConverter {
   private static final Pattern PARTITION_ALREADY_EXISTS_PARTITION =
       Pattern.compile(PARTITION_ALREADY_EXISTS_STRING);
 
+  private static final String BUCKETS_AUTO_NOT_SUPPORTED_STRING =
+      ".*?syntax error.*AUTO.*|.*?You have an error in your SQL syntax.*AUTO.*|.*?errCode = 2, detailMessage = Syntax error.*AUTO.*";
+
+  private static final Pattern BUCKETS_AUTO_NOT_SUPPORTED_PATTERN =
+      Pattern.compile(BUCKETS_AUTO_NOT_SUPPORTED_STRING, Pattern.CASE_INSENSITIVE);
+
   @SuppressWarnings("FormatStringAnnotation")
   @Override
   public GravitinoRuntimeException toGravitinoException(SQLException se) {
@@ -109,6 +116,15 @@ public class DorisExceptionConverter extends JdbcExceptionConverter {
         return new NoSuchPartitionException(se, se.getMessage());
       case CODE_PARTITION_ALREADY_EXISTS:
         return new PartitionAlreadyExistsException(se, se.getMessage());
+      case CODE_BUCKETS_AUTO_NOT_SUPPORTED:
+        String bucketsAutoMessage =
+            String.format(
+                "BUCKETS AUTO is not supported in this version of Apache Doris (requires Doris 1.2.2+). "
+                    + "BUCKETS AUTO was introduced in Doris 1.2.2. "
+                    + "Please either upgrade to Doris 1.2.2+ or specify a specific bucket number instead of AUTO. "
+                    + "Original error: %s",
+                se.getMessage());
+        return new GravitinoRuntimeException(se, bucketsAutoMessage);
       default:
         if (se.getMessage() != null && se.getMessage().contains("Access denied")) {
           return new ConnectionFailedException(se, se.getMessage());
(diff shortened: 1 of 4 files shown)


yuqi1129:
@mbaurin 
Thank you for your help. I have a question?  In which version does Doris support the keyword `AUTO`? Currently, Doris catalog is based on version `1.2.x`, I remembered that there is no `AUTO` there, please correct me if I'm wrong.

mbaurin:
> @mbaurin Thank you for your help. I have a question? In which version does Doris support the keyword `AUTO`? Currently, Doris catalog is based on version `1.2.x`, I remembered that there is no `AUTO` there, please correct me if I'm wrong.

Thanks for raising this ! I did some research and found that BUCKETS AUTO was actually introduced in Apache Doris 1.2.2, not earlier 1.2.x versions.

From the Doris 1.2.2 release notes and documentation:
- Feature was added as "Auto Bucket" in 1.2.2
- The 1.2-lts branch has regression tests for BUCKETS AUTO dating from April 2023
- It's not available in 1.2.0/1.2.1

What specific Doris version is Gravitino's catalog currently targeting? If we're supporting 1.2.2+, then this fix is valid. If we're still on 1.2.0/1.2.1, we might need to either:
1. Keep the fix - if we want forward compatibility with 1.2.2+
2. Add version validation - reject BUCKETS AUTO syntax for older versions
3. Update our target Doris version - if it's time to move to 1.2.2+

The existing codebase already had partial support for this (the HASH variant was working), so it seems like there was already some expectation of BUCKETS AUTO support.

What's your preference on how to handle this version compatibility ? Happy to adjust the approach accordingly !

yuqi1129:
Accordingly to [docs](https://gravitino.apache.org/docs/0.9.1/jdbc-doris-catalog), the version is `1.2.x`,  it hard to tell whether it is `1.2.0/1.2.1` or a higher version(1.2.2+)? So, is there any way to support both?

mbaurin:
> Accordingly to [docs](https://gravitino.apache.org/docs/0.9.1/jdbc-doris-catalog), the version is `1.2.x`, it hard to tell whether it is `1.2.0/1.2.1` or a higher version(1.2.2+)? So, is there any way to support both?

Good point, here are some approaches to support both:

Option 1: Graceful degradation
- Keep the parsing logic to handle BUCKETS AUTO
- When actually executing against Doris, if it fails due to unsupported syntax, we could catch the error and provide a
helpful message
- This way newer Doris versions work seamlessly, older versions get clear feedback

Option 2: Runtime version detection
- Add a version check against the Doris instance
- Parse BUCKETS AUTO only if version >= 1.2.2
- Requires adding version detection logic to the catalog

Option 3: Configuration-based
- Add a catalog property like doris.support.auto.buckets=true/false
- Let users configure based on their Doris version
- Default to false for backward compatibility

Option 4: Keep current approach
- The existing codebase already had partial AUTO support (HASH variant worked)
- This suggests the maintainers expected AUTO support
- We're just fixing the RANDOM variant bug

I'd lean toward Option 1, it's the least disruptive and provides the best user experience. What do you think?

yuqi1129:
> > Accordingly to [docs](https://gravitino.apache.org/docs/0.9.1/jdbc-doris-catalog), the version is `1.2.x`, it hard to tell whether it is `1.2.0/1.2.1` or a higher version(1.2.2+)? So, is there any way to support both?
> 
> Good point, here are some approaches to support both:
> 
> Option 1: Graceful degradation
> 
> * Keep the parsing logic to handle BUCKETS AUTO
> * When actually executing against Doris, if it fails due to unsupported syntax, we could catch the error and provide a
>   helpful message
> * This way newer Doris versions work seamlessly, older versions get clear feedback
> 
> Option 2: Runtime version detection
> 
> * Add a version check against the Doris instance
> * Parse BUCKETS AUTO only if version >= 1.2.2
> * Requires adding version detection logic to the catalog
> 
> Option 3: Configuration-based
> 
> * Add a catalog property like doris.support.auto.buckets=true/false
> * Let users configure based on their Doris version
> * Default to false for backward compatibility
> 
> Option 4: Keep current approach
> 
> * The existing codebase already had partial AUTO support (HASH variant worked)
> * This suggests the maintainers expected AUTO support
> * We're just fixing the RANDOM variant bug
> 
> I'd lean toward Option 1, it's the least disruptive and provides the best user experience. What do you think?

Yeah, solution one is more graceful, and the change is less intrusive.

mbaurin:
@yuqi1129 Option 1 implemented 👍

yuqi1129 (approved):

yuqi1129:
@mbaurin 

Please resolve the code format issue.

mbaurin:
> @mbaurin
> 
> Please resolve the code format issue.

done @yuqi1129

yuqi1129 (approved):

Issue #7744, opened 2025-07-17
[Improvement] nested call `SessionUtils.getWithoutCommit` will cause session closed


mchades:
### What would you like to be improved?

when such codes:
```
SessionUtils.getWithoutCommit(
            MapperA.class,
            mapper ->{
              SessionUtils.getWithoutCommit(
                  MapperB.class, m -> {
                    return somthing;
                });
                return somthing
            });
```

The error will be encountered during runtime:
```
    "Caused by: org.apache.ibatis.executor.ExecutorException: Executor was closed.",
    "\tat org.apache.ibatis.executor.BaseExecutor.createCacheKey(BaseExecutor.java:197)",
    "\tat org.apache.ibatis.executor.CachingExecutor.createCacheKey(CachingExecutor.java:146)",
    "\tat org.apache.ibatis.executor.CachingExecutor.query(CachingExecutor.java:88)",
    "\tat org.apache.ibatis.session.defaults.DefaultSqlSession.selectList(DefaultSqlSession.java:147)",
    "\t... 76 more"
```

### How should we improve?

We should add some checking mechanism to avoid nested usage or to avoid closing this session when it is nested.

PR #8242, opened 2025-08-22
[#7744] improvement(core): allow nested SqlSession call

mchades:

- add sessionCount for lifecycle management

Fix: #7744

no

tests added

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/UserMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/UserMetaService.java
@@ -285,12 +285,12 @@ public class UserMetaService {
     SessionUtils.doMultipleWithCommit(
         () ->
             userDeletedCount[0] =
-                SessionUtils.doWithoutCommitAndFetchResult(
+                SessionUtils.getWithoutCommit(
                     UserMetaMapper.class,
                     mapper -> mapper.deleteUserMetasByLegacyTimeline(legacyTimeline, limit)),
         () ->
             userRoleRelDeletedCount[0] =
-                SessionUtils.doWithoutCommitAndFetchResult(
+                SessionUtils.getWithoutCommit(
                     UserRoleRelMapper.class,
                     mapper ->
                         mapper.deleteUserRoleRelMetasByLegacyTimeline(legacyTimeline, limit)));
--- a/core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
@@ -22,6 +22,8 @@ package org.apache.gravitino.storage.relational.session;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.ibatis.session.SqlSession;
 import org.apache.ibatis.session.TransactionIsolationLevel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * SqlSessions is a utility class to maintain the MyBatis's {@link SqlSession} object. It is a
@@ -29,7 +31,9 @@ import org.apache.ibatis.session.TransactionIsolationLevel;
  * methods to commit, rollback and close the {@link SqlSession} object.
  */
 public final class SqlSessions {
+  private static final Logger LOG = LoggerFactory.getLogger(SqlSessions.class);
   private static final ThreadLocal<SqlSession> sessions = new ThreadLocal<>();
+  private static final ThreadLocal<Integer> sessionCount = ThreadLocal.withInitial(() -> 0);
 
   private SqlSessions() {}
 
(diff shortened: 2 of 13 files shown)


mchades:
@yuqi1129 could you please help to review this PR? thanks!

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
```diff
@@ -62,53 +72,65 @@ public static SqlSession getSqlSession() {
...
+          sqlSession.rollback();
+        }
+        // Always close the session when the count reaches 0
+        sqlSession.close();
+      } finally {
+        sessions.remove();
+        sessionCount.remove();
+      }
+    } else if (count < 0) {
+      // This should not happen if the session management is correct.
```
yuqi1129, line 131: When will this happen?

core/src/main/java/org/apache/gravitino/storage/relational/service/UserMetaService.java
```diff
@@ -285,12 +285,12 @@ public int deleteUserMetasByLegacyTimeline(long legacyTimeline, int limit) {
     SessionUtils.doMultipleWithCommit(
         () ->
             userDeletedCount[0] =
-                SessionUtils.doWithoutCommitAndFetchResult(
+                SessionUtils.getWithoutCommit(
```
yuqi1129, line 288: Why did you change here? `getWithoutCommit` will not return the modified count.

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
mchades: This should not happen, unless an unknown bug is introduced.

core/src/main/java/org/apache/gravitino/storage/relational/service/UserMetaService.java
mchades: > getWithoutCommit will not return the modified count.

why did you say that?

yuqi1129: OK, it's my fault.

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
```diff
@@ -62,53 +72,65 @@ public static SqlSession getSqlSession() {
...
+      // Reset the count and remove the session to avoid further issues.
+      LOG.warn(
+          "Session count is negative: {}. Resetting session count and removing session.",
+          count + 1);
```
yuqi1129: `count + 1` can be zero if `count` is negative.

yuqi1129:
Can the PR also fix the one https://github.com/apache/gravitino/pull/8210 that tries to?

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
mchades: `count` should not be negative, unless an unknown bug is introduced.

mchades: fixed

mchades:
> Can the PR also fix the one #8210 that tries to?

yes, the new commit I pushed can fixed that

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
```diff
@@ -52,8 +62,8 @@ public static SqlSession getSqlSession() {
               .getSqlSessionFactory()
               .openSession(TransactionIsolationLevel.READ_COMMITTED);
       sessions.set(sqlSession);
-      return sqlSession;
     }
+    sessionCount.set(sessionCount.get() + 1);
```
jerryshao: This is not atomic, you'd better use `AtomicInteger`.

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
```diff
@@ -62,53 +72,68 @@ public static SqlSession getSqlSession() {
...
-    SqlSession sqlSession = sessions.get();
-    if (sqlSession != null) {
-      try {
-        sqlSession.commit();
-        sqlSession.close();
-      } finally {
-        sessions.remove();
-      }
-    }
+    handleSessionClose(true, false);
```
jerryshao: Can you change all the code style to `handleSessionClose(true /** commit */, false /** rollback */);` for better review.

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessions.java
mchades: Thanks for the suggestion. While the ThreadLocal already ensures thread safety here, I agree that using `AtomicInteger` makes the intent of a counter more explicit and the code cleaner. I've made the change.

jerryshao (approved):

jerryshao:
@yuqi1129 can you please take another review?

yuqi1129:
> @yuqi1129 can you please take another review?

I don't have any more valuable comments.

Issue #6907, opened 2025-04-13
[Bug report] SparkConnector show tables in non-existed PG schema doesn't throw NoSuchNamespaceException


Xander-run:
### Version

main branch

### Describe what's wrong

SparkConnector list table in non-existed PG schema doesn't throw `NoSuchNamespaceException`

### Error message and/or stacktrace

```
spark-sql (default)> SHOW TABLES IN mock_schema;
25/04/13 18:49:44 WARN ObjectStore: Failed to get database global_temp, returning NoSuchObjectException
25/04/13 18:49:44 WARN ObjectStore: Failed to get database mock_schema, returning NoSuchObjectException
[SCHEMA_NOT_FOUND] The schema `mock_schema` cannot be found. Verify the spelling and correctness of the schema and catalog.
If you did not qualify the name with a catalog, verify the current_schema() output, or qualify the name with the correct catalog.
To tolerate the error on drop use DROP SCHEMA IF EXISTS.
spark-sql (default)> USE catalog_postgres;
Time taken: 0.068 seconds
spark-sql ()> SHOW TABLES IN mock_schema;
Time taken: 0.029 seconds
```

### How to reproduce

Use setups in `gravitino-playground` with gravitino-spark-connector built on main branch, followed steps in https://github.com/apache/gravitino/issues/6697#issuecomment-2746218039 

```
spark-sql (default)> USE catalog_postgres;
Time taken: 0.068 seconds
spark-sql ()> SHOW TABLES IN mock_schema;
Time taken: 0.029 seconds
```

Xander-run:
I'd like to work on this issue.

PR #8243, opened 2025-08-22
[#6907] feat(pg): throw exception when listing table on not exists schemas in PG catalog

FANNG1:

throw exception when listing table on not exists schemas in PG catalog

Fix: #6907 

no

IT

### DIFF
--- a/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
+++ b/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
@@ -45,7 +45,9 @@ import org.apache.gravitino.catalog.jdbc.config.JdbcConfig;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcColumnDefaultValueConverter;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcExceptionConverter;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcTypeConverter;
+import org.apache.gravitino.catalog.jdbc.operation.DatabaseOperation;
 import org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations;
+import org.apache.gravitino.catalog.jdbc.operation.RequireDatabaseOperation;
 import org.apache.gravitino.exceptions.NoSuchColumnException;
 import org.apache.gravitino.exceptions.NoSuchSchemaException;
 import org.apache.gravitino.exceptions.NoSuchTableException;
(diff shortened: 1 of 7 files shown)


FANNG1:
@yuqi1129 @jerryshao PTAL.

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
```diff
@@ -89,17 +93,27 @@ public void initialize(
...
       try (ResultSet tables = getTables(connection)) {
         while (tables.next()) {
-          if (Objects.equals(tables.getString("TABLE_SCHEM"), databaseName)) {
+          if (Objects.equals(tables.getString("TABLE_SCHEM"), schemaName)) {
             names.add(tables.getString("TABLE_NAME"));
           }
         }
       }
-      LOG.info("Finished listing tables size {} for database name {} ", names.size(), databaseName);
+      if (names.isEmpty()) {
```
yuqi1129: Why not use `schemaOperations.schemaExists` before listing all tables?

FANNG1: In most cases the schema exists when user listing tables in schema, so we could reduce on RPC call.

yuqi1129: Do you mean there will be a case: two threads access the catalog simultaneously and one has just listed all tables in a schema, and another thread then deleted the schema.

FANNG1: No,  if moving `schemaOperations.schemaExists` before,   `schemaOperations.schemaExists` is invoked every time, it will be expensive to do this because most times the user will not list tables in a empty schema.

FANNG1:
@yuqi1129 , any other comments?

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
FANNG1: updated, @yuqi1129 please help to review again

yuqi1129:
Is this a bug fix or a feature?

FANNG1:
> Is this a bug fix or a feature?

it's a bug fix,  updated the PR title.

yuqi1129:
Should it be cherry-picked to branch-0.9?

FANNG1:
> Should it be cherry-picked to branch-0.9?

no, minor changes

yuqi1129:
> > Should it be cherry-picked to branch-0.9?
> 
> no, minor changes

Got

Issue #8200, opened 2025-08-20
[Improvement] Double unlock should thow an exception in TreeLock.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/lock/TreeLock.java calling unlock with no loge should thow an exception.

Here a unit test to help:
```
  @Test
  void testDoubleUnlockThrows() {
    lock.lock(LockType.READ);
    lock.unlock();
    assertThrows(IllegalStateException.class, () -> lock.unlock());
  }
```

### How should we improve?

Fix and add tests.

keepConcentration:
@justinmclean 

Hi, can I take this issue?

justinmclean:
All yours.

PR #8244, opened 2025-08-22
[#8200] improvement(core): throw exception when unlocking without an active lock in TreeLock

keepConcentration:

Added a check in `TreeLock.unlock()` to throw `IllegalStateException` when `unlock()` is called without any active lock.  

Previously, calling `unlock()` without holding a lock (e.g., double unlock) was silently ignored, which could hide logic errors in lock handling.  

Fix: #8200

Yes.  
  1. Users calling `unlock()` without an active lock will now receive an `IllegalStateException`.

  1. Added a unit test `testDoubleUnlockThrows()` verifying that a second `unlock()` call throws `IllegalStateException`
  2. Executed existing unit tests


yuqi1129 (approved):
LGTM

Issue #8251, opened 2025-08-22
[Bug report] Gravitino will OOM when repeated load fileset and modify fileset catalog


yuqi1129:
### Version

main branch

### Describe what's wrong

```shell
while [[ 1 -eq 1 ]]
do 
curl -X PUT -H "Accept: application/vnd.gravitino.v1+json" \
-H "Content-Type: application/json" -d '{
  "updates": [
    {
      "@type": "setProperty",
      "property": "key3",
      "value": "value3"
    }
  ]
}' http://localhost:8090/api/metalakes/test/catalogs/hdfs_catalog

curl -X GET -H "Accept: application/vnd.gravitino.v1+json" \
-H "Content-Type: application/json"  \
http://localhost:8090/api/metalakes/test/catalogs/hdfs_catalog/schemas/schema/filesets/fileset01/files

sleep 1
done
```

Running the script above can reproduce it.

### Error message and/or stacktrace

2025-08-22 10:55:50.634 WARN [Gravitino-webserver-68] [org.apache.gravitino.utils.PrincipalUtils.doAs(PrincipalUtils.java:50)] - doAs method occurs an unexpected error
java.lang.OutOfMemoryError: Metaspace
        at java.base/java.lang.ClassLoader.defineClass1(Native Method) ~[?:?]
        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017) ~[?:?]
        at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150) ~[?:?]
        at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:524) ~[?:?]
        at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:427) ~[?:?]
        at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:421) ~[?:?]
        at java.base/java.security.AccessController.executePrivileged(AccessController.java:807) ~[?:?]
        at java.base/java.security.AccessController.doPrivileged(AccessController.java:712) ~[?:?]
        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:420) ~[?:?]
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:592) ~[?:?]
        at org.apache.gravitino.utils.IsolatedClassLoader$CustomURLClassLoader.doLoadClass(IsolatedClassLoader.java:204) ~[gravitino-core-1.0.0-SNAPSHOT.jar:?]
[...]

Issue #8249, opened 2025-08-22
[Bug report] Fail to browser files in Azure cloud storage


yuqi1129:
### Version

main branch

### Describe what's wrong

<img width="1615" height="705" alt="Image" src="https://github.com/user-attachments/assets/a6210653-30d0-40ac-8b07-9d4d538a7082" />

### Error message and/or stacktrace

{
    "code": 1001,
    "type": "IllegalArgumentException",
    "message": "Failed to operate fileset(s) [fileset01] operation [LOAD] under schema [schema], reason [lastModified must be a valid timestamp]",
    "stack": [
        "java.lang.IllegalArgumentException: lastModified must be a valid timestamp",
        "\tat com.google.common.base.Preconditions.checkArgument(Preconditions.java:143)",
        "\tat org.apache.gravitino.dto.file.FileInfoDTO$FileInfoDTOBuilder.build(FileInfoDTO.java:165)",
        "\tat org.apache.gravitino.catalog.fileset.FilesetCatalogOperations.lambda$listFiles$4(FilesetCatalogOperations.java:340)",
        "\tat java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)",
        "\tat java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992)",
        "\tat java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)",
        "\tat java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)",
        "\tat java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)",
        "\tat java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)",
        "\tat java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)",
        "\tat org.apache.gravitino.catalog.fileset.FilesetCatalogOperations.listFiles(FilesetCatalogOperations.java:341)",
        "\tat org.apache.gravitino.catalog.fileset.SecureFilesetCatalogOperations.listFiles(SecureFilesetCatalogOperations.java:261)",
        "\tat org.apache.gravitino.catalog.FilesetOperationDispatcher.lambda$listFiles$6(FilesetOperationDispatcher.java:113)",
[...]

PR #8252, opened 2025-08-22
[#8251][#8249] fix(fileset-catalog): Fix OOM problem for fileset catalog

yuqi1129:

1. Clear thread locals and daemon thread for cusom class loader to let GC free it.
2. Fix tiemstamp can be 0 for azure file system.

They are bugs.

Fix: #8251 
Fix: #8249 

N/A.

Test locally.

### DIFF
--- a/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
+++ b/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
@@ -53,7 +53,8 @@ public class ClassLoaderResourceCleanerUtils {
     }
 
     // Clear statics threads in FileSystem and close all FileSystem instances.
-    closeStatsDataClearerInFileSystem(classLoader);
+    executeAndCatch(
+        ClassLoaderResourceCleanerUtils::closeStatsDataClearerInFileSystem, classLoader);
 
     // Stop all threads with the current class loader and clear their threadLocal variables for
     // jetty threads that are loaded by the current class loader.
@@ -61,19 +62,20 @@ public class ClassLoaderResourceCleanerUtils {
     // within jetty thread with the current class loader. However, there are clear by
     // `catalog.close` in ForkJoinPool in CaffeineCache, in this case, the thread local variable
     // will not be cleared, so we need to clear them manually here.
-    stopThreadsAndClearThreadLocalVariables(classLoader);
+    executeAndCatch(
+        ClassLoaderResourceCleanerUtils::stopThreadsAndClearThreadLocalVariables, classLoader);
 
     // Release the LogFactory for the classloader, each classloader has its own LogFactory
     // instance.
-    releaseLogFactoryInCommonLogging(classLoader);
+    executeAndCatch(ClassLoaderResourceCleanerUtils::releaseLogFactoryInCommonLogging, classLoader);
 
-    closeResourceInAWS(classLoader);
+    executeAndCatch(ClassLoaderResourceCleanerUtils::closeResourceInAWS, classLoader);
 
-    closeResourceInGCP(classLoader);
+    executeAndCatch(ClassLoaderResourceCleanerUtils::closeResourceInGCP, classLoader);
 
-    closeResourceInAzure(classLoader);
+    executeAndCatch(ClassLoaderResourceCleanerUtils::closeResourceInAzure, classLoader);
 
-    clearShutdownHooks(classLoader);
+    executeAndCatch(ClassLoaderResourceCleanerUtils::clearShutdownHooks, classLoader);
   }
 
   /**
@@ -81,34 +83,27 @@ public class ClassLoaderResourceCleanerUtils {
    *
    * @param targetClassLoader the classloader where Hadoop FileSystem is loaded
    */
-  private static void closeStatsDataClearerInFileSystem(ClassLoader targetClassLoader) {
-    try {
-      Class<?> fileSystemClass =
-          Class.forName("org.apache.hadoop.fs.FileSystem", true, targetClassLoader);
-      MethodUtils.invokeStaticMethod(fileSystemClass, "closeAll");
+  private static void closeStatsDataClearerInFileSystem(ClassLoader targetClassLoader)
+      throws Exception {
+    Class<?> fileSystemClass =
+        Class.forName("org.apache.hadoop.fs.FileSystem", true, targetClassLoader);
+    MethodUtils.invokeStaticMethod(fileSystemClass, "closeAll");
 
-      Class<?> mutableQuantilesClass =
-          Class.forName("org.apache.hadoop.metrics2.lib.MutableQuantiles", true, targetClassLoader);
-      Class<?> statisticsClass =
-          Class.forName("org.apache.hadoop.fs.FileSystem$Statistics", true, targetClassLoader);
+    Class<?> mutableQuantilesClass =
+        Class.forName("org.apache.hadoop.metrics2.lib.MutableQuantiles", true, targetClassLoader);
+    Class<?> statisticsClass =
+        Class.forName("org.apache.hadoop.fs.FileSystem$Statistics", true, targetClassLoader);
 
-      ScheduledExecutorService scheduler =
-          (ScheduledExecutorService)
-              FieldUtils.readStaticField(mutableQuantilesClass, "scheduler", true);
-      scheduler.shutdownNow();
-      Field statisticsCleanerField =
-          FieldUtils.getField(statisticsClass, "STATS_DATA_CLEANER", true);
-      Object statisticsCleaner = statisticsCleanerField.get(null);
-      if (statisticsCleaner != null) {
-        ((Thread) statisticsCleaner).interrupt();
-        ((Thread) statisticsCleaner).setContextClassLoader(null);
-        ((Thread) statisticsCleaner).join();
-      }
-    } catch (Exception e) {
-      LOG.warn(
-          "Failed to close FileSystem and MutableQuantiles statistics cleaner in {}",
-          targetClassLoader,
-          e);
+    ScheduledExecutorService scheduler =
+        (ScheduledExecutorService)
+            FieldUtils.readStaticField(mutableQuantilesClass, "scheduler", true);
+    scheduler.shutdownNow();
+    Field statisticsCleanerField = FieldUtils.getField(statisticsClass, "STATS_DATA_CLEANER", true);
+    Object statisticsCleaner = statisticsCleanerField.get(null);
+    if (statisticsCleaner != null) {
+      ((Thread) statisticsCleaner).interrupt();
+      ((Thread) statisticsCleaner).setContextClassLoader(null);
+      ((Thread) statisticsCleaner).join();
     }
   }
 
@@ -205,23 +200,18 @@ public class ClassLoaderResourceCleanerUtils {
    *
    * @param targetClassLoader the classloader where the shutdown hooks are registered.
    */
-  private static void clearShutdownHooks(ClassLoader targetClassLoader) {
-    try {
-      Class<?> shutdownHooks = Class.forName("java.lang.ApplicationShutdownHooks");
-      IdentityHashMap<Thread, Thread> hooks =
-          (IdentityHashMap<Thread, Thread>)
-              FieldUtils.readStaticField(shutdownHooks, "hooks", true);
+  private static void clearShutdownHooks(ClassLoader targetClassLoader) throws Exception {
+    Class<?> shutdownHooks = Class.forName("java.lang.ApplicationShutdownHooks");
+    IdentityHashMap<Thread, Thread> hooks =
+        (IdentityHashMap<Thread, Thread>) FieldUtils.readStaticField(shutdownHooks, "hooks", true);
 
-      hooks
-          .entrySet()
-          .removeIf(
-              entry -> {
-                Thread thread = entry.getKey();
-                return thread.getContextClassLoader() == targetClassLoader;
-              });
-    } catch (Exception e) {
-      LOG.warn("Failed to clean shutdown hooks", e);
-    }
+    hooks
+        .entrySet()
+        .removeIf(
+            entry -> {
+              Thread thread = entry.getKey();
+              return thread.getContextClassLoader() == targetClassLoader;
+            });
   }
 
   /**
@@ -229,15 +219,12 @@ public class ClassLoaderResourceCleanerUtils {
    *
    * @param currentClassLoader the classloader where the commons-logging is loaded.
    */
-  private static void releaseLogFactoryInCommonLogging(ClassLoader currentClassLoader) {
+  private static void releaseLogFactoryInCommonLogging(ClassLoader currentClassLoader)
+      throws Exception {
     // Release the LogFactory for the FilesetCatalogOperations class loader
-    try {
-      Class<?> logFactoryClass =
-          Class.forName("org.apache.commons.logging.LogFactory", true, currentClassLoader);
-      MethodUtils.invokeStaticMethod(logFactoryClass, "release", currentClassLoader);
-    } catch (Exception e) {
-      LOG.warn("Failed to release LogFactory for class loader {}", currentClassLoader, e);
-    }
+    Class<?> logFactoryClass =
+        Class.forName("org.apache.commons.logging.LogFactory", true, currentClassLoader);
+    MethodUtils.invokeStaticMethod(logFactoryClass, "release", currentClassLoader);
   }
 
   /**
@@ -245,30 +232,21 @@ public class ClassLoaderResourceCleanerUtils {
    *
    * @param classLoader the classloader where AWS SDK is loaded
    */
-  private static void closeResourceInAWS(ClassLoader classLoader) {
+  private static void closeResourceInAWS(ClassLoader classLoader) throws Exception {
     // For Aws SDK metrics, unregister the metric admin MBean
-    try {
-      Class<?> awsSdkMetricsClass =
-          Class.forName("com.amazonaws.metrics.AwsSdkMetrics", true, classLoader);
-      MethodUtils.invokeStaticMethod(awsSdkMetricsClass, "unregisterMetricAdminMBean");
-    } catch (Exception e) {
-      LOG.warn(
-          "Failed to unregister AWS SDK metrics admin MBean from class loader {}", classLoader, e);
-    }
+    Class<?> awsSdkMetricsClass =
+        Class.forName("com.amazonaws.metrics.AwsSdkMetrics", true, classLoader);
+    MethodUtils.invokeStaticMethod(awsSdkMetricsClass, "unregisterMetricAdminMBean");
   }
 
-  private static void closeResourceInGCP(ClassLoader classLoader) {
+  private static void closeResourceInGCP(ClassLoader classLoader) throws Exception {
     // For GCS
-    try {
-      Class<?> relocatedLogFactory =
-          Class.forName(
-              "org.apache.gravitino.gcp.shaded.org.apache.commons.logging.LogFactory",
-              true,
-              classLoader);
-      MethodUtils.invokeStaticMethod(relocatedLogFactory, "release", classLoader);
-    } catch (Exception e) {
-      LOG.warn("Failed to find GCS shaded LogFactory in class loader {}", classLoader, e);
-    }
+    Class<?> relocatedLogFactory =
+        Class.forName(
+            "org.apache.gravitino.gcp.shaded.org.apache.commons.logging.LogFactory",
+            true,
+            classLoader);
+    MethodUtils.invokeStaticMethod(relocatedLogFactory, "release", classLoader);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


bundles/azure-bundle/build.gradle.kts
```diff
@@ -25,7 +25,10 @@ plugins {
 }
 
 dependencies {
-  implementation(project(":bundles:azure"))
+  implementation(project(":bundles:azure")) {
+    exclude(group = "commons-logging", module = "commons-logging")
```
mchades, line 31: Why exclude this? You left a comment saying it is needed for runtime.

https://github.com/apache/gravitino/blob/728ea2e702177cd0dc456ba5e525795d14e67f5d/bundles/gcp/build.gradle.kts#L44-L45

common/src/main/java/org/apache/gravitino/dto/file/FileInfoDTO.java
```diff
@@ -162,7 +162,8 @@ public FileInfoDTO.FileInfoDTOBuilder path(String path) {
     public FileInfoDTO build() {
       Preconditions.checkArgument(StringUtils.isNotBlank(name), "name cannot be null or empty");
       Preconditions.checkArgument(size >= 0, "size cannot be negative");
-      Preconditions.checkArgument(lastModified > 0, "lastModified must be a valid timestamp");
+      // Azure Blob Storage uses a timestamp in milliseconds, it can be 0 for newly created files.
+      Preconditions.checkArgument(lastModified >= 0, "lastModified must be a valid timestamp");
```
mchades, line 166: I don't get the relationship between `a timestamp in milliseconds` and `0 for newly created files`

yuqi1129: All in all, it can be 0.

yuqi1129: Is the unit here not 'timestamp'?

mchades: I think you can change it to:
``
// it can be 0 for newly created files in Azure Blob Storage
``

bundles/azure-bundle/build.gradle.kts
yuqi1129: There is already `commons.logging` in `hadoop-azure` for bundle, so it is no need to add it.

common/src/main/java/org/apache/gravitino/dto/file/FileInfoDTO.java
yuqi1129: OK

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,182 @@ public void close() throws IOException {
...
+          try {
+            thread.join(5000);
+          } catch (InterruptedException e) {
+            LOG.warn("Failed to join peer cache thread: {}", thread.getName(), e);
+          }
+        }
+      }
+
+      // Release the LogFactory for the FilesetCatalogOperations class loader
+      LogFactory.release(SecureFilesetCatalogOperations.class.getClassLoader());
```
FANNG1: Does other catalog has similar problems for LogFactory?

yuqi1129: Yes,  the Paimon and Iceberg catalog share the same problem, and the cases are more complicated, I'm workong on it.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,182 @@ public void close() throws IOException {
     catalogUserContext.close();
 
     UserContext.cleanAllUserContext();
+
+    boolean testEnv = System.getenv("GRAVITINO_TEST") != null;
```
FANNG1: How about keep the cleanup process in TEST if there are no bad effects, we could check whether there are exceptions in cleanup in test.

yuqi1129: No, it will lead problem if the logic is absent, as there is only one class loader in the test environment.

FANNG1: could you add the comment to the above code?

yuqi1129: I see.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,182 @@ public void close() throws IOException {
...
+            LOG.warn("Failed to join peer cache thread: {}", thread.getName(), e);
+          }
+        }
+      }
+
+      // Release the LogFactory for the FilesetCatalogOperations class loader
+      LogFactory.release(SecureFilesetCatalogOperations.class.getClassLoader());
+
+      // For Aws SDK metrics, unregister the metric admin MBean
+      try {
```
FANNG1: Could you do some abstraction like `cleanupAwsResource`?

yuqi1129: done

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,208 @@ public void close() throws IOException {
...
+      MethodUtils.invokeStaticMethod(methodUtilsClass, "unregisterMetricAdminMBean");
+    } catch (Exception e) {
+      LOG.warn("Failed to unregister AWS SDK metrics admin MBean", e);
+      // This is not critical, so we just log the warning
+    }
+  }
+
+  private static void clearThreadLocalMap(Thread thread) {
+    if (thread != null && thread.getName().startsWith("Gravitino-webserver-")) {
+      // Try to
```
FANNG1: Try to?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,208 @@ public void close() throws IOException {
...
+
+      closeResourceInAWS();
+
+      closeResourceInGCP();
+
+      closeResourceInAzure();
+
+      clearShutdownHooks();
+    } catch (Exception e) {
+      LOG.warn("Failed to clear FileSystem statistics cleaner thread", e);
```
FANNG1: Is the log content correct?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
mchades, line 1: Can you post some evidence that proves these thread locals caused OOM? I think this will also help with troubleshooting when OOM occurs again. It also helps reviewers to check this PR locally.

yuqi1129: If the comment I posted wasn't clear enough, I can't provide any evidence that is more persuasive. it can be summarized as 'Catalog related thread locals are initialized in jetty thread pool and then closed with Caffeine thread pool'

```text
      // Clear thread local map for webserver threads in the current class loader. Why do we need
      // this? Because the webserver threads are long-lived threads and will holds may thread
      // local references to the current class loader. However, they can't be cleared as
      // `Catalog.close` is called in the thread pool located in the Caffeine cache, which is not
      // in the webserver threads. So we need to manually clear the thread local map for webserver
      // threads.
```

FANNG1 (dismissed):

FANNG1:
@mchades any other comments?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,211 @@ public void close() throws IOException {
     catalogUserContext.close();
 
     UserContext.cleanAllUserContext();
+
+    boolean testEnv = System.getenv("GRAVITINO_TEST") != null;
+    if (testEnv) {
+      // In test environment, we do not need to clean up class loader related stuff
```
mchades: The comment should explain "why", and the real reason is that you mentioned in the previous comment that "there is only one class loader in the test environment."

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,211 @@ public void close() throws IOException {
...
+      // threads.
+      clearThreadLocalMap(thread);
+
+      // Close all threads that are using the FilesetCatalogOperations class loader
+      if (runningWithCurrentClassLoader(thread)) {
+        LOG.info("Interrupting thread: {}", thread.getName());
+        thread.setContextClassLoader(null);
+        thread.interrupt();
+        try {
+          thread.join(5000);
```
mchades: Does this conflict with the `Thread.sleep(500);` in the CatalogManager?

mchades (commented):
My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops. 

I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1

FANNG1:
> My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> 
> I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1

Yes, the current solution depends on the implementation logic of dependency jars. classloader reuse mentioned by @jerryshao is a more general way,  but my concern is the time for 1.0 and may introduce some unexpected behavior before the release.

yuqi1129:
> My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> 
> I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1

I admit that the current solution is not an elegant one and hard to extend, but there is no better one available for me to fix the problem. If you have any good idea, feel free to tell me. 

The biggest problem is that Gravitino provides a pluggable way to support multiple features, such as fileset, we can't guarantee that if there are similar problems in the jars provided by users, another points we can also introduce new jars as the project develops. No one can guarantee that the same problem will be triggered again.

yuqi1129:
> > My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> > I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1
> 
> Yes, the current solution depends on the implementation logic of dependency jars. classloader reuse mentioned by @jerryshao is a more general way, but my concern is the time for 1.0 and may introduce some unexpected behavior before the release.

A cache mechanism can alleviate the problem, but it cannot fix it at its source.  Since we are using Caffeine cache for the catalog manager, the class loader will be recreated if the cache expires.

yuqi1129:
> Quote reply

I have checked that the reused class loader has a problem with static values like `UserGroupInformation.loginUser`, we need to check and test, and then make corresponding changes.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -244,6 +244,9 @@ public void close() {
               catalog = null;
               return null;
             });
+        // Let hooks close first before classLoader close, so that the hook can
+        // release any resources that the catalog might be using.
+        Thread.sleep(500);
```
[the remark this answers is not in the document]
jerryshao: Using sleep is very flaky, you should use another way like the coplit suggested.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,211 @@ public void close() throws IOException {
...
+          .entrySet()
+          .removeIf(
+              entry -> {
+                Thread thread = entry.getKey();
+                return thread.getContextClassLoader()
+                    == FilesetCatalogOperations.class.getClassLoader();
+              });
+    } catch (Exception e) {
+      LOG.warn("Failed to clean shutdown hooks", e);
+    }
```
jerryshao: The code above is very very brittle. You should separate this bunch of code from the main logic, and use another util class to encapsulate all the logic. That util class can be removed as a whole once we figure out a better way.

Another thing is that you should add lots of comments to describe each line of reflection, otherwise, others cannot understand why you made such a change.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,211 @@ public void close() throws IOException {
...
+
+      closeResourceInAzure();
+
+      clearShutdownHooks();
+    } catch (Exception e) {
+      LOG.warn("Failed to clear resources(Thread, ThreadLocal variants) in the class loader", e);
+    }
+  }
+
+  private static void stopThreadsAndClearThreadLocal() {
```
jerryshao: "ThreadLocalVariables"

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/SecureFilesetCatalogOperations.java
```diff
@@ -271,6 +280,211 @@ public void close() throws IOException {
...
+
+  private void unregisterLogFactory() {
+    try {
+      LogFactory.release(SecureFilesetCatalogOperations.class.getClassLoader());
+    } catch (Exception e) {
+      LOG.warn("Failed to unregister LogFactory", e);
+    }
+  }
+
+  private static void closeResourceInAzure() {
```
jerryshao: Can we move such code into the specific bundle package?

yuqi1129: Module `Azure` already depends on `catalog-fileset`, it will lead to circular dependencies if moved there.

yuqi1129: I'm going to move those methods to `catalog-common`, so that other catalogs can reuse them.

jerryshao:
> My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> 
> I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1

I agree. My thought is that though we may have lots of catalogs, the number of providers is limited, which means the number of classloaders are limited, we don't have to create a classloader for each catalog. We can reuse the classloader for the same type of catalogs, and with this we can reduce the class-related OOM issues. 

Also, many engines like Spark/Trino that support multiple connectors will face the same problem, especially for Trino, which can also support dynamic connector register/unregister. How do they handle this problem? We'd better have a long-term solution.

For now, I think we can have a short solution, but we should clearly separate this from the main logic, so we can remove it easily. Also, we should add the comments as possible as we can.

yuqi1129:
> > My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> > I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1
> 
> I agree. My thought is that though we may have lots of catalogs, the number of providers is limited, which means the number of classloaders are limited, we don't have to create a classloader for each catalog. We can reuse the classloader for the same type of catalogs, and with this we can reduce the class-related OOM issues.
> 
> Also, many engines like Spark/Trino that support multiple connectors will face the same problem, especially for Trino, which can also support dynamic connector register/unregister. How do they handle this problem? We'd better have a long-term solution.
> 
> For now, I think we can have a short solution, but we should clearly separate this from the main logic, so we can remove it easily. Also, we should add the comments as possible as we can.

I see. 

I will try to introduce the class loader reuse mechanism in the next version, and possibly there will be static variable sharing problems like `UGI` exists, so conducting full tests is necessary before this changes.

yuqi1129:
> > > My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> > > I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1
> > 
> > 
> > I agree. My thought is that though we may have lots of catalogs, the number of providers is limited, which means the number of classloaders are limited, we don't have to create a classloader for each catalog. We can reuse the classloader for the same type of catalogs, and with this we can reduce the class-related OOM issues.
> > Also, many engines like Spark/Trino that support multiple connectors will face the same problem, especially for Trino, which can also support dynamic connector register/unregister. How do they handle this problem? We'd better have a long-term solution.
> > For now, I think we can have a short solution, but we should clearly separate this from the main logic, so we can remove it easily. Also, we should add the comments as possible as we can.
> 
> I see.
> 
> I will try to introduce the class loader reuse mechanism in the next version, and possibly there will be static variable sharing problems like `UGI` exists, so conducting full tests is necessary before this changes.

In Trino, there will be only one classloader for each type of catalog.

jerryshao:
> > > > My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> > > > I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1
> > > 
> > > 
> > > I agree. My thought is that though we may have lots of catalogs, the number of providers is limited, which means the number of classloaders are limited, we don't have to create a classloader for each catalog. We can reuse the classloader for the same type of catalogs, and with this we can reduce the class-related OOM issues.
> > > Also, many engines like Spark/Trino that support multiple connectors will face the same problem, especially for Trino, which can also support dynamic connector register/unregister. How do they handle this problem? We'd better have a long-term solution.
> > > For now, I think we can have a short solution, but we should clearly separate this from the main logic, so we can remove it easily. Also, we should add the comments as possible as we can.
> > 
> > 
> > I see.
> > I will try to introduce the class loader reuse mechanism in the next version, and possibly there will be static variable sharing problems like `UGI` exists, so conducting full tests is necessary before this changes.
> 
> In Trino, there will be only one classloader for each type of catalog.

We can also have the same solution as I mentioned above. The problem we face is the same as Trino (they also have the same UGI issue). Please investigate it and have a long-term solution.

yuqi1129:
> > > > > My biggest concern is the maintainability of the clear resource logic. It seems that we are doing some complex work that should belong to the JVM level's GC, and this workload will get out of control as the project develops.
> > > > > I don't know if we have other alternative solutions to solve the OOM problem, but the current solution seems not to be the optimal one in the long run. WDYT? @jerryshao @FANNG1
> > > > 
> > > > 
> > > > I agree. My thought is that though we may have lots of catalogs, the number of providers is limited, which means the number of classloaders are limited, we don't have to create a classloader for each catalog. We can reuse the classloader for the same type of catalogs, and with this we can reduce the class-related OOM issues.
> > > > Also, many engines like Spark/Trino that support multiple connectors will face the same problem, especially for Trino, which can also support dynamic connector register/unregister. How do they handle this problem? We'd better have a long-term solution.
> > > > For now, I think we can have a short solution, but we should clearly separate this from the main logic, so we can remove it easily. Also, we should add the comments as possible as we can.
> > > 
> > > 
> > > I see.
> > > I will try to introduce the class loader reuse mechanism in the next version, and possibly there will be static variable sharing problems like `UGI` exists, so conducting full tests is necessary before this changes.
> > 
> > 
> > In Trino, there will be only one classloader for each type of catalog.
> 
> We can also have the same solution as I mentioned above. The problem we face is the same as Trino (they also have the same UGI issue). Please investigate it and have a long-term solution.

I will. Should I move this PR or focus on a new solution? I'm afraid we won't be able to finish it within the time frame of 1.0.0, as large number of tests are needed.

jerryshao:
No, we can have a short-term solution. Just refactor the code as what I mentioned above. For the long term solution, we can move to the next release.

yuqi1129:
Polish done, please help to review it again.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderUtils.java
```diff
@@ -0,0 +1,314 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ClassLoaderUtils {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ClassLoaderUtils.class);
+
+  private ClassLoaderUtils() {}
+
+  /**
```
jerryshao: Please move the public class to the top.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderUtils.java
```diff
@@ -0,0 +1,314 @@
...
+   * Close the stats data clearer thread in Hadoop FileSystem to prevent memory leaks when using
+   *
+   * @param targetClassLoader the classloader where Hadoop FileSystem is loaded
+   */
+  private static void closeStatsDataClearerInFileSystem(ClassLoader targetClassLoader) {
+    try {
+      Class<?> FileSystem =
```
jerryshao: Can you please rename all the class variables to `xxxClass`?

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderUtils.java
```diff
@@ -0,0 +1,314 @@
...
+import java.lang.reflect.Field;
+import java.util.IdentityHashMap;
+import java.util.Timer;
+import java.util.concurrent.ScheduledExecutorService;
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.commons.lang3.reflect.MethodUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ClassLoaderUtils {
```
jerryshao: This class name cannot directly reflect the purpose, can you figure out a better name? Also, add the comments to this class.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderUtils.java
```diff
@@ -0,0 +1,314 @@
...
+
+      closeResourceInAWS(classLoader);
+
+      closeResourceInGCP(classLoader);
+
+      closeResourceInAzure(classLoader);
+
+      clearShutdownHooks(classLoader);
+    } catch (Exception e) {
+      LOG.warn("Failed to clear resources(Thread, ThreadLocal variants) in the class loader", e);
```
jerryshao: "variants" or ""variables"?

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderUtils.java
jerryshao: Seems this is not changed.

bundles/azure-bundle/build.gradle.kts
mchades: If so, could you plz add this to the code comment?

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
```diff
@@ -0,0 +1,320 @@
...
+
+      closeResourceInAWS(classLoader);
+
+      closeResourceInGCP(classLoader);
+
+      closeResourceInAzure(classLoader);
+
+      clearShutdownHooks(classLoader);
+    } catch (Exception e) {
+      LOG.warn("Failed to clear resources(threads, thread local variables) in the class loader", e);
```
mchades: Looks like if an error occurs during the call to `closeResourceInAWS`, the methods `closeResourceInGCP`, `closeResourceInAzure`, and `closeStatsDataClearerInFileSystem` will never be called? Is this the expected behavior?

yuqi1129: No, I have caught any possible exceptions in `closeResourceInAWS `.

```java
  private static void closeResourceInAWS(ClassLoader classLoader) {
    // For Aws SDK metrics, unregister the metric admin MBean
    try {
      Class<?> awsSdkMetricsClass =
          Class.forName("com.amazonaws.metrics.AwsSdkMetrics", true, classLoader);
      MethodUtils.invokeStaticMethod(awsSdkMetricsClass, "unregisterMetricAdminMBean");
    } catch (Exception e) {
      LOG.warn(
          "Failed to unregister AWS SDK metrics admin MBean from class loader {}", classLoader, e);
    }
  }
```

mchades: So, why do you need to catch the exception again here?

yuqi1129: It's okay to remove the `try/catch` block, and it's just the preventive measures.

mchades: we can use a function to call the resource cleaner functions and catch exceptions

yuqi1129: added

mchades (approved):
lgtm.

Issue #7938, opened 2025-08-05
[Improvement] metalake becomes orphan when its owner is deleted


mchades:
### What would you like to be improved?

When the owner of a Metalake deletes themselves from the Metalake, the Metalake becomes an orphan, and after that, no one can manage the Metalake anymore. We should avoid this situation.

### How should we improve?

avoid the metalake owner delete themselves

PR #8253, opened 2025-08-22
[#7938] improvement: Add check when deleting users or groups to avoid orphan metalake

jerqi:

Add check when deleting users or groups to avoid orphan metalake

Fix: #7938

No.

Added UT.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
@@ -130,19 +130,18 @@ public class GroupOperations {
       return Utils.doAs(
           httpRequest,
           () -> {
-            Owner owner =
-                ownerDispatcher
-                    .getOwner(
-                        metalake, MetadataObjects.of(null, metalake, MetadataObject.Type.METALAKE))
-                    .orElseThrow(
-                        () ->
-                            new IllegalStateException("Owner not found for metalake: " + metalake));
-            if (owner.type() == Owner.Type.GROUP && owner.name().equals(group)) {
-              throw new IllegalArgumentException(
-                  String.format(
-                      "Cannot remove group %s from metalake %s because the current user is not the owner of the metalake.",
-                      group, metalake));
-            }
+            ownerDispatcher
+                .getOwner(
+                    metalake, MetadataObjects.of(null, metalake, MetadataObject.Type.METALAKE))
+                .ifPresent(
+                    owner -> {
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
```diff
@@ -124,6 +130,20 @@ public Response removeGroup(
       return Utils.doAs(
           httpRequest,
           () -> {
+            Owner owner =
+                ownerDispatcher
+                    .getOwner(
+                        metalake, MetadataObjects.of(null, metalake, MetadataObject.Type.METALAKE))
+                    .orElseThrow(
+                        () ->
+                            new IllegalStateException("Owner not found for metalake: " + metalake));
```
mchades: Does this mean that if the metalake owner cannot be found, we are unable to delete the group as well?

server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
```diff
@@ -124,6 +130,20 @@ public Response removeGroup(
...
+            if (owner.type() == Owner.Type.GROUP && owner.name().equals(group)) {
+              throw new IllegalArgumentException(
+                  String.format(
+                      "Cannot remove group %s from metalake %s because the current user is not the owner of the metalake.",
```
mchades: the message does not match the condition

roryqi: Fixed.

server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
roryqi: Em.. you are right, I should remove the constraint.

server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
mchades: It seems not fixed.

roryqi: I have changed to
```
Cannot remove group %s from metalake %s because the group is not the owner of the metalake.
```

mchades: but your condition is `owner.type() == Owner.Type.GROUP && owner.name().equals(group)`

roryqi: My fault. Fixed.

Issue #7572, opened 2025-07-04
[Improvement] Performance Degradation When Concurrent Requests Exceed 8 in GVFS Java Client


huiqing-wang:
### What would you like to be improved?

During performance testing using the GVFS Java client, we observed a significant performance drop when the concurrency level exceeds 8 threads.
﻿
Upon checking the connection count between the GVFS client and server using netstat, we found that only 5 connections are established. The number of connections does not increase even when the concurrency level rises.
﻿
This suggests that the GVFS client may have a hard-coded or configuration-based connection limit, which could be a bottleneck for high-concurrency scenarios.

![Image](https://github.com/user-attachments/assets/a05d8b28-2db4-4ac0-80ef-090639381699)

It uses Executors.newFixedThreadPool to create a fixed-size thread pool, where each thread executes the method：
`
long fileNum = 0;
        long start = System.currentTimeMillis();
        try {
            FileSystem gravitinoFs = subPath.getFileSystem(conf);

            for (FileAttrDao fileAttrDao : files) {
                fileNum++;

                long fileStart = System.currentTimeMillis();

                org.apache.hadoop.fs.Path srcFilePath = new org.apache.hadoop.fs.Path(fileAttrDao.getFile().getAbsolutePath());

                org.apache.hadoop.fs.Path dstFilePath = new org.apache.hadoop.fs.Path(
                        String.format("%s/%s/%s", fileset, subDir, fileAttrDao.getFileName())
                );

                gravitinoFs.copyFromLocalFile(srcFilePath, dstFilePath);

                long fileEnd = System.currentTimeMillis();
                log.info("{} ， {} ms", fileAttrDao.getFileName(), (fileEnd - fileStart));
            }

        } catch (Exception e) {
            e.printStackTrace();
        } finally {
            long end = System.currentTimeMillis();
        }
`

yuqi1129:
@huiqing-wang 
Can you provide any details about the test code or test steps?

huiqing-wang:
Gravitino Client established an HTTP connection with the server using the httpcomponents-5.2.1 to create RestClient. 
Two default parameters, maxConnTotal=25 and maxConnPerRoute=5, were used to create the client connection pool;
`public static final int DEFAULT_MAX_TOTAL_CONNECTIONS = 25;
public static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 5;`

The source code and testing phenomena are completely consistent.

[...]

PR #4320, opened 2024-07-31
[#4278] feat(filesystem): Refactor the `getFileLocation` logics in hadoop GVFS

xloya:
### What changes were proposed in this pull request?

Refactor the `getFileLocation` logic in Hadoop GVFS so that when sending a request to the server to obtain the file location, the current data operation and operation path information are reported.

### Why are the changes needed?

Fix: #4278 

### How was this patch tested?

Add and refactor some UTs, and existing ITs maintain normally.

PR #8254, opened 2025-08-22
[#7572] improvement(gvfs): add fileset cache in GVFS

mchades:

add fileset cache in GVFS

Fix: #7572

no

CI pass


clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
```diff
@@ -80,6 +107,14 @@ private Cache<NameIdentifier, FilesetCatalog> newCatalogCache(
         .build();
   }
 
+  private Cache<NameIdentifier, Fileset> newFilesetCache(
+      ScheduledThreadPoolExecutor filesetCleanScheduler) {
+    return Caffeine.newBuilder()
+        .maximumSize(100)
```
yuqi1129: Why is the size so small? Is there any other consideration that I am not aware of?

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
```diff
@@ -70,6 +76,27 @@ public FilesetCatalog getFilesetCatalog(NameIdentifier catalogIdent) {
...
+        NameIdentifier.of(filesetIdent.namespace().level(0), filesetIdent.namespace().level(1));
+    FilesetCatalog filesetCatalog = getFilesetCatalog(catalogIdent);
+    Fileset fileset =
+        filesetCache.get(
+            filesetIdent,
+            ident ->
+                filesetCatalog.loadFileset(
+                    NameIdentifier.of(filesetIdent.namespace().level(2), filesetIdent.name())));
+    Preconditions.checkArgument(
+        fileset != null, String.format("Loaded fileset: %s is null.", filesetIdent));
```
yuqi1129: When will the fileset become null?

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
mchades: Just copy from catalogCache. It seems there won't be too many filesets in one GVFS instance. What value do you prefer? I'll update it accordingly.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
mchades: It shouldn't, unless a bug is introduced.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
yuqi1129: This can be slightly larger, say 1000, or 10000.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
yuqi1129: It can't be null according to the code.

mchades: So you are suggesting removing this check?

yuqi1129: This is not a big problem; it depends on you. I remember when I used the same code, jerry just suggested me  remove it as `NoSuchFilesetException` will occur.  For me both are okay.

clients/client-python/gravitino/filesystem/gvfs_base_operations.py
```diff
@@ -547,3 +549,32 @@ def _get_fileset_catalog(self, catalog_ident: NameIdentifier):
...
+                fileset_ident.namespace().level(0), fileset_ident.namespace().level(1)
+            )
+            catalog: FilesetCatalog = self._get_fileset_catalog(catalog_ident)
+            fileset = catalog.as_fileset_catalog().load_fileset(
+                NameIdentifier.of(fileset_ident.namespace().level(2), fileset_ident.name())
+            )
+            self._fileset_cache[fileset_ident] = fileset
+            return fileset
+        finally:
+            write_lock.release()
```
jerryshao, line 602: Because cache has side-effects, I suggest that we add a configuration to enable/disable the cache, by default it can be disabled.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
```diff
@@ -80,6 +107,14 @@ private Cache<NameIdentifier, FilesetCatalog> newCatalogCache(
         .build();
   }
 
+  private Cache<NameIdentifier, Fileset> newFilesetCache(
+      ScheduledThreadPoolExecutor filesetCleanScheduler) {
+    return Caffeine.newBuilder()
+        .maximumSize(100)
+        .scheduler(Scheduler.forScheduledExecutorService(filesetCleanScheduler))
```
jerryshao: What is the usage of this `scheduler`? I don't find any `Callable` or `Runnable` for this scheduler.

mchades: the Caffine will use it automatically, the same is also used elsewhere in the project

https://github.com/apache/gravitino/blob/a4d0545b54e44504ce445de3f97c218aaf233eb1/catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/CachedClientPool.java#L73-L81

yuqi1129: In the client size, there is no need to set it explicitly. The reason that I set it in `CachedClientPool` is that we need to hold the reference of the `scheduler` or JVM can't clean up the class loader in Gravitino server.

mchades: ok, if it's unnecessary, I'll remove all the scheduler from the client side.

BTW, it's first introduced by the PR #4320 in [GravitinoVirtualFileSystem.java](https://github.com/apache/gravitino/pull/4320/files#diff-64c5b87c772c66c642144ad44764c1339ad846ddc013000c52c9ab503cbce02eR141-R145)

clients/client-python/gravitino/filesystem/gvfs_base_operations.py
mchades: added

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetCatalogCache.java
mchades: fixed

mchades: after I removed this scheduler, the cache tests failed (more details see [here](https://github.com/apache/gravitino/actions/runs/17231457433/job/48886396015?pr=8254)): 

> TestSimpleClient > testInternalCache() FAILED
>     org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a Lambda expression in org.apache.gravitino.filesystem.hadoop.TestGvfsBase expected: <0> but was: <1> within 5 seconds.
>         at app//org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
>         at app//org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
>         at app//org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
>         at app//org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1006)
>         at app//org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
>         at app//org.apache.gravitino.filesystem.hadoop.TestGvfsBase.testInternalCache(TestGvfsBase.java:387)
>         at java.base@17.0.16/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>         at java.base@17.0.16/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
>         at java.base@17.0.16/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>         at java.base@17.0.16/java.lang.reflect.Method.invoke(Method.java:569)


I'll add it back and make it a local variable.

mchades:
@jerryshao @yuqi1129 All comments resolved, please help to review again, thanks!

clients/client-python/gravitino/filesystem/gvfs_config.py
```diff
@@ -76,3 +76,6 @@ class GVFSConfig:
 
     # The configuration key prefix for the client.
     GVFS_FILESYSTEM_CLIENT_CONFIG_PREFIX = "gvfs_gravitino_client_"
+
+    # The configuration key for whether to enable fileset catalog cache. The default is false.
+    GVFS_FILESYSTEM_ENABLE_FILESET_CATALOG_CACHE = "enable_fileset_catalog_cache"
```
jerryshao: This is not only for catalog cache, the configuration name is misleading, can you figure out a better name.

jerryshao: Also, you should clearly describe the side-effects in here and doc.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/BaseGVFSOperations.java
```diff
@@ -146,13 +146,15 @@ protected BaseGVFSOperations(Configuration configuration) {
...
 
-    GravitinoClient client = GravitinoVirtualFileSystemUtils.createClient(configuration);
-    this.filesetCatalogCache = new FilesetCatalogCache(client);
+    this.gravitinoClient = GravitinoVirtualFileSystemUtils.createClient(configuration);
+    this.enableFilesetCatalogCache =
+        configuration.getBoolean(
+            FS_GRAVITINO_FILESET_CATALOG_CACHE_ENABLE,
+            FS_GRAVITINO_FILESET_CATALOG_CACHE_ENABLE_DEFAULT);
+    this.filesetCatalogCache =
+        enableFilesetCatalogCache ? new FilesetCatalogCache(gravitinoClient) : null;
```
jerryshao: I don't like specifically setting `null` value in the code, it is error-prone. You'd better use `Optional` instead.

clients/client-python/gravitino/filesystem/gvfs_config.py
mchades: fixed

jerryshao (approved):
LGTM, @yuqi1129 do you want to take another look.

yuqi1129 (approved):

Issue #8241, opened 2025-08-21
[Improvement] Support SQL UPDATE/DELETE/MERGE in Trino


hqbhoho:
### What would you like to be improved?

Support SQL UPDATE/DELETE/MERGE with Hive/Iceberg catalog in Trino.

mbaurin:
Hey ! 👋 Just implemented this and opened a PR with full UPDATE/DELETE/MERGE support for Hive/Iceberg catalogs. Let me know what you think !

hqbhoho:
> Hey ! 👋 Just implemented this and opened a PR with full UPDATE/DELETE/MERGE support for Hive/Iceberg catalogs. Let me know what you think !

@mbaurin sorry,  I have do with it to support UPDATE/DELETE/MERGE, you can help review it , I have filed a PR https://github.com/apache/gravitino/pull/8188 to support with jdbc catalog.

PR #8258, opened 2025-08-22
[#8241] improvement(trino-connector): Support SQL UPDATE/DELETE/MERGE with Hive/Iceberg

hqbhoho:

Support SQL UPDATE/DELETE/MERGE with Hive/Iceberg

Fix: #8241

No

local tests


hqbhoho:
@diqiu50 @yuqi1129  PTAL, thanks!

integration-test-common/docker-script/init/hive/init.sql
```diff
@@ -0,0 +1,30 @@
...
+TBLPROPERTIES ('transactional'='true');
+CREATE TABLE gt_hive_test_acid.test_merge
+(
+    id INT,
+    name STRING,
+    salary INT
+)
+CLUSTERED BY (id) INTO 4 BUCKETS
+STORED AS ORC
+TBLPROPERTIES ('transactional'='true');
```
diqiu50, line 48: Why is the table created here instead of within the test script?

integration-test-common/docker-script/init/hive/init.sh
```diff
@@ -19,4 +19,14 @@
 IP=$(hostname -I | awk '{print $1}')
 sed -i "s|<value>hdfs://__REPLACE__HOST_NAME:9000|<value>hdfs://${IP}:9000|g" ${HIVE_TMP_CONF_DIR}/hive-site.xml
 
-/bin/bash /usr/local/sbin/start.sh
+/bin/bash /usr/local/sbin/start.sh &
+
```
diqiu50, line 38: Does it not work in `docker-comple.yaml `
```
    healthcheck:
      test: ["CMD", "/tmp/check-status.sh"]
```

integration-test-common/docker-script/init/hive/init.sql
hqbhoho: Currently, it can't create a hive ACID table like this: 
```sql
CREATE TABLE tb01 (
    n_nationkey bigint,
    n_name varchar,
    n_regionkey bigint,
    n_comment varchar,
    part_key varchar
) WITH (bucket_count = 4, bucketed_by = ARRAY['n_regionkey'], transactional = 'true', format = 'ORC');

Expect:
CREATE TABLE
Actual:
Query 20250903_094457_00270_ajycf failed: Catalog 'gt_hive' table property 'transactional' does not exist.
```

Considering that ACID tables in Hive are not widely used, I feel it's unnecessary to support their creation in Gravitino for now.

integration-test-common/docker-script/init/hive/init.sh
hqbhoho: The main purpose here is to perform the execution to create the relevant tables which define in `init.sql`.
​​Run `/bin/bash /usr/local/sbin/start.sh` will block.
`/usr/local/sbin/start.sh` define end with:
```shell
# start hive
echo "Starting Hive..."
${HIVE_HOME}/bin/schematool -initSchema -dbType mysql
${HIVE_HOME}/bin/hive --service hiveserver2 > /dev/null 2>&1 &
${HIVE_HOME}/bin/hive --service metastore > /dev/null 2>&1 &

echo "Hive started successfully."

# persist the container
tail -f /dev/null
```

diqiu50 (approved):
LGTM

Issue #8206, opened 2025-08-20
[Improvement] Possible null pointer exceptions during drop-event logging in AsyncQueueListener.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/listener/AsyncQueueListener.java, a NPE can occur.

Here's a unit test to show the issue:
```
public class TestAsyncQueueListener {
  private static class NoopListener implements EventListenerPlugin {
    @Override
    public void init(Map<String, String> properties) {}

    @Override
    public void start() {}

    @Override
    public void stop() {}
  }

  @Test
  public void testDropEventLoggingDoesNotThrow() {
    AsyncQueueListener listener =
        new AsyncQueueListener(Collections.singletonList(new NoopListener()), "test", 1, 1);
    PreEvent event = new ListMetalakePreEvent("user");
    listener.onPreEvent(event); // fill the queue
    Assertions.assertDoesNotThrow(() -> listener.onPreEvent(event));
  }
}
```

### How should we improve?

Fix by initialising lastRecordDropEventTime

hahaffan:
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.Assertions;

import java.util.Collections;
import java.util.Map;

public class TestAsyncQueueListener {
  private static class NoopListener implements EventListenerPlugin {
    @Override
    public void init(Map<String, String> properties) {}

    @Override
    public void start() {}

    @Override
    public void stop() {}
  }

  @Test
  public void testDropEventLoggingDoesNotThrow() {
    AsyncQueueListener listener =
        new AsyncQueueListener(Collections.singletonList(new NoopListener()), "test", 1, 1);
    PreEvent event = new ListMetalakePreEvent("user");

    listener.onPreEvent(event); // fill the queue
    Assertions.assertDoesNotThrow(() -> listener.onPreEvent(event));
  }
}

justinmclean:
@hahaffan, do you want to work on this issue? It's unclear why you posted the above comment.

timurturbil:
I would like to work on the issue, can you assign it to me? @justinmclean

justinmclean:
Sure it's all yours.

PR #8259, opened 2025-08-22
[#8206] fix: NPE by initializing lastRecordDropEventTime in AsyncQueueListener

timurturbil:

Initialize `lastRecordDropEventTime` in `AsyncQueueListener` to avoid a potential `NullPointerException` when logging drop events. This ensures the timestamp is always non-null before being accessed.

`AsyncQueueListener` can throw a `NullPointerException` if the queue is full and drop events are logged before `lastRecordDropEventTime` is initialized.
By setting a default value (`Instant.EPOCH`), we guarantee safe access and correct logging behavior.

Fix: #8206

No.
This change only affects the internal implementation of event logging. No user-facing APIs, configurations, or behaviors are modified.

This patch was tested with the unit test provided by the issue author (`TestAsyncQueueListener.testDropEventLoggingDoesNotThrow`), which verifies that dropping events no longer throws an exception but the test is not included in the commit.


core/src/main/java/org/apache/gravitino/listener/AsyncQueueListener.java
```diff
@@ -68,6 +68,7 @@ public AsyncQueueListener(
     this.asyncProcessor = new Thread(() -> processEvents());
     this.dispatcherJoinSeconds = dispatcherJoinSeconds;
     this.highWatermarkThreshold = (int) (queueCapacity * HIGH_WATERMARK_RATIO);
+    this.lastRecordDropEventTime = Instant.EPOCH;
```
FANNG1: Could you initialize the value like this?
```java
  private Instant lastRecordDropEventTime = Instant.EPOCH;
```

FANNG1 (approved):

FANNG1:
@timurturbil , merged to main, thanks for your contribution.

Issue #8166, opened 2025-08-19
[Improvement] Fix handling of SQL exception error codes in PostgreSQLExceptionConverter.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/storage/relational/converters/PostgreSQLExceptionConverter.java, SQL exception codes are treated as a number, which is incorrect. e.g. 28P01 means invalid password

Here are some tests to show that.
```
  @Test
  public void testConvertDuplicatedEntryException() {
    SQLException sqlException = new SQLException("duplicate", "23505");
    PostgreSQLExceptionConverter converter = new PostgreSQLExceptionConverter();
    Assertions.assertThrows(
        EntityAlreadyExistsException.class,
        () -> converter.toGravitinoException(sqlException, Entity.EntityType.METALAKE, "test"),
        String.format("The %s entity: %s already exists.", Entity.EntityType.METALAKE, "test"));
  }

  @Test
  public void testNonNumericSqlStateHandledAsIOException() {
    SQLException sqlException = new SQLException("error", "28P01");
    PostgreSQLExceptionConverter converter = new PostgreSQLExceptionConverter();
    Assertions.assertThrows(
        IOException.class,
        () -> converter.toGravitinoException(sqlException, Entity.EntityType.METALAKE, "test"));
  }
```

### How should we improve?

Treat them as strings. Note currently a java.lang.NumberFormatException will be thrown.

iinow:
hi~ can I take this issue?

justinmclean:
Sure, if you need any help, just ask.

PR #8260, opened 2025-08-22
[#8166] fix: treat PostgreSQL SQLSTATE as string

iinow:

Fix bug in `PostgreSQLExceptionConverter` where SQLSTATE codes were parsed as numbers.  
Now SQLSTATE is treated as a string to avoid `NumberFormatException` (e.g., "28P01").

PostgreSQL SQLSTATE codes can include letters. Parsing them as integers causes runtime errors.

Fix: #8166

No

1. Added unit tests for numeric code ("23505") and non-numeric code ("28P01").  
2. Verified tests pass successfully.


shaofengshi:
LGTM; @yuqi1129 Yuqi, what do you think?

Issue #7752, opened 2025-07-18
[Bug report] JcasbinAuthorizer handle exception explictly


FANNG1:
### Version

main branch

### Describe what's wrong

https://github.com/apache/gravitino/blob/794814b5ba74165a7d6f2f1bc99de00b5a3829d4/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java#L163-L182

I'm testing Iceberg auth,  `MetadataIdConverter.getID` failed for using Gravitino catalog manager,  but there are no logs or exception but telling me the user doesn't have the privilege to access metadata object xx. This misleads me to check whether the user role is correct. 

### Error message and/or stacktrace

no message

### How to reproduce

in Iceberg auth test, not easy to reproduce.

FANNG1:
cc @lsyulong

FANNG1:
There are some limitations for the current `JcasbinAuthorizer` , some entity store interface may throw exception like query roles for specify user and the external schema&table created out of Gravitino maydoesn't have meta in Gravitino, if we throw exception, it may interrupt the overall authz process(suppose the user is the owner of the catalog but doesn't assign any role), so remove 1.0 tag and we should rethink how to improve the cases properly. cc @hdygxsj @xunliu @jerryshao

roryqi:
We should authorize first. If one object doesnt, we should return having no privileges.

PR #8269, opened 2025-08-25
[#7752] fix(authz): JcasbinAuthorizer handle exception explicitly

FANNG1:

JcasbinAuthorizer handles exceptions explicitly not eat it.

Fix: #7752 

no

existing ITs

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -171,8 +171,7 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
                   UserEntity.class)
           != null;
     } catch (Exception e) {
-      LOG.warn("Can not get user {} in metalake {}", currentUserName, metalake, e);
-      return false;
+      throw new RuntimeException(e);
     }
   }
 
@@ -192,8 +191,7 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
         loadRolePrivilege(metalake, currentUserName, userId);
         return allowEnforcer.hasRoleForUser(String.valueOf(userId), String.valueOf(roleId));
       } catch (Exception e) {
-        LOG.warn("can not get user id or role id.", e);
-        return false;
+        throw new RuntimeException(e);
       }
     }
     throw new UnsupportedOperationException("Unsupported Entity Type.");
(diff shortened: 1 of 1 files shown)


FANNG1:
@hdygxsj @xunliu @jerryshao PTAL.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -309,8 +307,7 @@ private boolean loadPrivilegeAndAuthorize(
         userId = userEntity.id();
         metadataId = MetadataIdConverter.getID(metadataObject, metalake);
       } catch (Exception e) {
-        LOG.debug("Can not get entity id", e);
-        return false;
+        throw new RuntimeException(e);
```
hdygxsj, line 310: If a certain metadata does not exist in Gravitino, an exception will be thrown here. In fact, it should return `false`, because throwing an exception would interrupt the list metadata operation.

xunliu: @hdygxsj Can we try catching this exception in the list metadata operation?

FANNG1: As `loadPrivilegeAndAuthorize` return `false` means authz failed, how to handle get metadata failed?

FANNG1:
Many IT failures for the change, I'll fix the IT if the change is acceptable.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
hdygxsj: We might want to swallow some exceptions and return false, while letting the remaining exceptions propagate. 

Such as,

```
java.lang.RuntimeException: org.apache.gravitino.exceptions.NoSuchEntityException: No such schema entity: default
```

hdygxsj: > As `loadPrivilegeAndAuthorize` return `false` means authz failed, how to handle get metadata failed?

Because during the authentication process, we access a large amount of metadata that does not exist in Gravitino.

For example, if a user has the "select table" privilege on a certain table, they should be able to load it. However, the schema metadata might not exist in Gravitino. If an exception is thrown during the metadata authorization expression validation chain, it would cause the entire authorization check to fail.

FANNG1: could you provide more context about  `the schema metadata might not exist in Gravitino`?

Issue #7370, opened 2025-06-10
[Bug report] Gravitino will OOM when alter Iceberg catalog several times


yuqi1129:
### Version

main branch

### Describe what's wrong

Create an Iceberg catalog with Hive backend, then do the following operation: 

➜  [/Users/yuqi/Downloads] while [[ 1 -eq 1 ]]
do
curl -X PUT -H "Accept: application/vnd.gravitino.v1+json" \
-H "Content-Type: application/json" -d '{
  "updates": [
    {
      "@type": "setProperty",
      "property": "key3",
      "value": "value3"
    }
  ]
}' http://localhost:8090/api/metalakes/test1/catalogs/iceberg_catalog

curl -X GET -H "Accept: application/vnd.gravitino.v1+json" \
-H "Content-Type: application/json"  \
http://localhost:8090/api/metalakes/test1/catalogs/iceberg_catalog/schemas/default/tables/table

sleep 1
done

Then OOM will occur; for more details, please see below.

### Error message and/or stacktrace

Caused by: java.util.ServiceConfigurationError: org.apache.gravitino.CatalogProvider: Provider org.apache.gravitino.catalog.lakehouse.iceberg.IcebergCatalog could not be instantiated
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:586) ~[?:?]
	at java.base/java.util.ServiceLoader$ProviderImpl.newInstance(ServiceLoader.java:813) ~[?:?]
	at java.base/java.util.ServiceLoader$ProviderImpl.get(ServiceLoader.java:729) ~[?:?]
	at java.base/java.util.ServiceLoader$3.next(ServiceLoader.java:1403) ~[?:?]
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133) ~[?:?]
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
[...]

PR #8271, opened 2025-08-25
[#7370] fix(iceberag,paimon): Fix OOM problem for Iceberg and Paimon catalog

yuqi1129:

Close all daemon threads and clear all thread locals that hinder the class loader from being cleaned.

They are bugs.

Fix: #7370

N/A.

Test locally.


jerryshao:
@yuqi1129 would you please continue on this PR?

yuqi1129:
> @yuqi1129 would you please continue on this PR?

Got it.

yuqi1129:
It seems that for all catalogs, only when we close:
1. All thread locals used in catalogs related to running threads.
2. All running threads.

After a thorough test, Hadoop jars are the primary area of concern.

Issue #8274, opened 2025-08-26
[Improvement] Update Python Client to support time/timestamp precision


predator4ann:
### What would you like to be improved?

After #6657 merged, we should make the Python Client also support time precision.
- Link #7331 - Task 3

PR #8275, opened 2025-08-26
[#8274] Update Python Client to support time/timestamp precision

predator4ann:

Fix: #8274 

No

UT already added.


mchades (approved):
lgtm.

Issue #8282, opened 2025-08-26
[Improvement] Fix tests in TestModelMetaService.java


justinmclean:
### What would you like to be improved?

In core/src/test/java/org/apache/gravitino/storage/relational/service/TestModelMetaService.java:
```
    ModelEntity updatedModel =
        ModelEntity.builder()
            .withId(modelEntity.id())
            .withName(newComment)
            .withNamespace(modelEntity.namespace())
            .withLatestVersion(modelEntity.latestVersion())
            .withAuditInfo(modelEntity.auditInfo())
            .withComment(modelEntity.comment())
            .withProperties(modelEntity.properties())
            .build();
```
Is incorrectly setting the name and comment.

The test testInsertAndUpdateModelProperties has a similar issue.

### How should we improve?

Fix the tests.

JeonDaehong:
Can I take on this issue and work on resolving it?

Aamir017:
@justinmclean 

Hi, I’m interested in fixing this. Could you assign it to me?

justinmclean:
@JeonDaehong It's yours. @Aamir017 thanks again for offering to help, but @JeonDaehong asked first.

JeonDaehong:
> @JeonDaehong It's your. @Aamir017 thanks again for offering to help, but @JeonDaehong asked first.

Thanks :D

PR #8296, opened 2025-08-26
Fix incorrect name/comment assignment in TestModelMetaService tests

JeonDaehong:
**What changes were proposed in this pull request?**
In `TestModelMetaService.java`, the tests testInsertAndUpdateModelComment and testInsertAndUpdateModelProperties were incorrectly setting the name and comment fields. This PR fixes the assignments so that name and comment are updated correctly according to the test's intent.

**Why are the changes needed?**
Previously, the tests would either overwrite the name with a comment value or keep the comment unchanged, causing the tests to not properly verify updates. Fixing the assignments ensures that comment updates and property updates are correctly tested.

Fixes: #8282

**Does this PR introduce any user-facing change?**
No. This only fixes test behavior.

**How was this patch tested?**
- Corrected field assignments in the affected tests.
- Verified that `testInsertAndUpdateModelComment` correctly updates the comment without changing the name.
- Verified that `testInsertAndUpdateModelProperties` correctly maintains the name and updates properties.
- Ran all existing tests to confirm no regressions.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/storage/relational/service/TestModelMetaService.java
+++ b/core/src/test/java/org/apache/gravitino/storage/relational/service/TestModelMetaService.java
@@ -266,11 +266,11 @@ public class TestModelMetaService extends TestJDBCBackend {
     ModelEntity updatedModel =
         ModelEntity.builder()
             .withId(modelEntity.id())
-            .withName(newComment)
+            .withName(modelEntity.name())
             .withNamespace(modelEntity.namespace())
             .withLatestVersion(modelEntity.latestVersion())
             .withAuditInfo(modelEntity.auditInfo())
-            .withComment(modelEntity.comment())
+            .withComment(newComment)
             .withProperties(modelEntity.properties())
             .build();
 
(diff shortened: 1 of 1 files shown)


JeonDaehong:
@justinmclean 

It says the test failed, but I can’t see the details of the failure.
Could I get some feedback on this?

justinmclean:
I think they just took too long to run and were cancelled

JeonDaehong:
> I think they just took too long to run and were cancelled

@justinmclean 
This is the first time I’ve encountered something like this—how should I handle it?

justinmclean:
I can re-run the failed ones for you

JeonDaehong:
> I can re-run the failed ones for you

@justinmclean 

Please try running it again.
If it still doesn’t work, then it might be an issue with the changes I made to the source code, in which case I’ll take a look and fix it.

JeonDaehong:
It looks like the test is completed :D

justinmclean (approved):
Thanks for fixing

JeonDaehong:
Thanks !

Issue #8306, opened 2025-08-27
[Improvement] Rollback call when autocommit is true in ScriptRunner.java


justinmclean:
### What would you like to be improved?

In ScriptRunner, conn.rollback() will always be called (as it's in a finally block). However, this could possibly lead to issues with auto-commit connections. The rollback should only occur when auto-commit is set to false.
Also, I think the conn.rollback() should be in the exception block, otherwise, it will also be called directly after a commit, while this has no impact, it seems unnecessary.

### How should we improve?

Only rollback auto-commit is set to false and there is an exception

Surya00011:
Hi @justinmclean, I'll take this

justinmclean:
It's all yours

PR #8332, opened 2025-08-28
[#8306] Avoid rollback when autocommit is true in ScriptRunner.java

Surya00011:
---
- Updated `ScriptRunner.java` so that `conn.rollback()` is **not always called**.  
- Rollback is now only performed when:  
  - `autoCommit` is set to `false`, **and**  
  - an exception occurs during script execution.  
- Moved `conn.rollback()` logic from the `finally` block into the `catch` block.  
- Cleaned up control flow to avoid unnecessary rollback calls after successful commits.  

---

---
Previously, `conn.rollback()` was called in the `finally` block regardless of connection mode.  

- When `autoCommit = true`, rollback has no effect and is misleading.  
- When a commit was already performed successfully, rollback was still invoked unnecessarily.  

This change makes rollback behavior **cleaner and safer**, ensuring it is only triggered when actually required.  

---

---
-  No user-facing API changes.  
-  Internal improvement in the Hive Metastore test utility `ScriptRunner.java`.


Surya00011:
Hi @justinmclean, any comments?

Issue #7908, opened 2025-08-04
[Improvement] Support pass s3 config


liangyouze:
### What would you like to be improved?

We are using other object storage compatible with the S3 protocol. When accessing it with boto3, we need to pass the configuration `"addressing_style"="virtual"` like this:
```
s3_config = Config(region_name=self.region, s3={"addressing_style": "virtual"})
s3 = boto3.client('s3', endpoint_url=self.endpoint,
                                aws_access_key_id=self.ak,
                                aws_secret_access_key=self.sk, config=s3_config)
```
 However, when using the gvfs client, there is no way to pass this parameter to botocore. We hope gvfs can support passing additional configurations

yuqi1129:
Are there any volunteers available to address this issue?

yuqi1129:
With #8333, you can use `config_kwargs` to pass custom configuration. For more, please see https://s3fs.readthedocs.io/en/latest/api.html

PR #8333, opened 2025-08-28
[#7908] improvement(python-client): Support passing custom configuration for python GVFS client.

yuqi1129:

Support passing kwargs to Python GVFS client.

It's a common requirement.

Fix: #7908 

N/A.

ITs

### DIFF
--- a/clients/client-python/tests/integration/test_gvfs_with_s3.py
+++ b/clients/client-python/tests/integration/test_gvfs_with_s3.py
@@ -114,7 +114,9 @@ class TestGvfsWithS3(TestGvfsWithHDFS):
             name=cls.metalake_name, comment="", properties={}
         )
         cls.gravitino_client = GravitinoClient(
-            uri="http://localhost:8090", metalake_name=cls.metalake_name
+            uri="http://localhost:8090",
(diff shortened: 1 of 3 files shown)


clients/client-python/tests/integration/test_gvfs_with_s3.py
```diff
@@ -264,16 +266,23 @@ def test_rmdir(self):
 
         fs.rm_file(rmdir_file)
 
+    # pylint: disable=W0212
```
mchades, line 269: What does W0212 refer to? Why is it not processed but disabled? (could you please also explain in the code comment?)

mchades (commented):
one minor comment

clients/client-python/tests/integration/test_gvfs_with_s3.py
yuqi1129: > What does W0212 refer to? 

This method access private method directly. Please see ` fs.operations._get_actual_filesystem(mkdir_dir, None)`. Method `_get_actual_filesystem` is a private method and should not be accessed outside; however, this is test code.

Issue #8281, opened 2025-08-26
[Improvement] Fix handlePolicyException paramaters in PolicyOperations.java


justinmclean:
### What would you like to be improved?

This line of code has the parameters in an incorrect order:
```
return ExceptionHandlers.handlePolicyException(OperationType.LIST, "", policyName, e);
```

### How should we improve?

correct order and use matalake as parent.

pandeysambhi:
@justinmclean can I take this up?

justinmclean:
sure

PR #8334, opened 2025-08-28
[#8281]Improvement(PolicyOperation): handle policy exception parameters

pandeysambhi:
Fixed handlePolicyException paramaters in PolicyOperations.java

improvement for handlePolicyException

Fix: #8281 

No

Unit test

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
@@ -287,7 +287,7 @@ public class PolicyOperations {
           });
 
     } catch (Exception e) {
-      return ExceptionHandlers.handlePolicyException(OperationType.LIST, policyName, metalake, e);
+      return ExceptionHandlers.handlePolicyException(OperationType.LIST, "", metalake, e);
     }
   }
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -287,7 +287,7 @@ public Response listMetadataObjectsForPolicy(
           });
 
     } catch (Exception e) {
-      return ExceptionHandlers.handlePolicyException(OperationType.LIST, "", policyName, e);
+      return ExceptionHandlers.handlePolicyException(OperationType.LIST, policyName, metalake, e);
```
jerryshao: The change is not correct, it should be `ExceptionHandlers.handlePolicyException(OperationType.LIST, "", metalake, e);`

pandeysambhi: @jerryshao issue mentioned that the order of parameters are incorrect and to use metalake as the parent. as per your suggestion, policyName should be empty, do you want me to update the PR with that change.

jerryshao: Yes.

Issue #8335, opened 2025-08-28
[Bug report] REST interface unnecessary encoding issue


jerryshao:
### Version

main branch

### Describe what's wrong

The Java and Python http client will encode the query parameters automatically, doing it manually will bring in decoding issue.

### Error message and/or stacktrace

N/A

### How to reproduce

N/A

PR #8336, opened 2025-08-28
[#8335] fix(client): Fix double encoding issue and add missing URL encoding

jerryshao:

This PR mainly fixes server things:

1. Fix the double encoding issue. Because Java and Python client will automatically encode the query parameters. Doing it manually will encode the string twice, which will lead to unexpected issues.
2. Add missing encoding in the URL.
3. Add a client version number in the header to let the server check and do the action based on the version.

Fix the double encoding problem.

Fix: #8335 

Yes, the change will break the backward compatibility for fileset related interfaces.

Local test.

### DIFF
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/FilesetCatalog.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/FilesetCatalog.java
@@ -249,9 +249,9 @@ class FilesetCatalog extends BaseSchemaCatalog
       CallerContext callerContext = CallerContext.CallerContextHolder.get();
 
       Map<String, String> params = new HashMap<>();
-      params.put("sub_path", RESTUtils.encodeString(subPath));
+      params.put("sub_path", subPath);
       if (locationName != null) {
-        params.put("location_name", RESTUtils.encodeString(locationName));
+        params.put("location_name", locationName);
       }
       FileLocationResponse resp =
           restClient.get(
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
@@ -33,6 +33,8 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.Version;
 import org.apache.gravitino.auth.AuthConstants;
 import org.apache.gravitino.dto.responses.ErrorResponse;
 import org.apache.gravitino.exceptions.RESTException;
@@ -680,7 +682,14 @@ public class HTTPClient implements RESTClient {
     // Some systems require the Content-Type header to be set even for empty-bodied requests to
     // avoid failures.
     request.setHeader(HttpHeaders.CONTENT_TYPE, bodyMimeType);
+    // Set the API version header
     request.setHeader(HttpHeaders.ACCEPT, VERSION_HEADER);
+
+    // Set the client version header
+    if (StringUtils.isNotBlank(Version.getCurrentVersion().version)) {
+      request.setHeader(Version.CLIENT_VERSION_HEADER, Version.getCurrentVersion().version);
+    }
+
     if (requestHeaders != null) {
       requestHeaders.forEach(request::setHeader);
     }
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
@@ -23,9 +23,6 @@ import static org.apache.gravitino.file.Fileset.LOCATION_NAME_UNKNOWN;
 import com.codahale.metrics.annotation.ResponseMetered;
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.collect.ImmutableMap;
-import java.io.UnsupportedEncodingException;
-import java.net.URLDecoder;
-import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -44,7 +41,6 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Entity;
 import org.apache.gravitino.MetadataObject;
 import org.apache.gravitino.NameIdentifier;
@@ -234,8 +230,7 @@ public class FilesetOperations {
       @PathParam("schema") @AuthorizationMetadata(type = Entity.EntityType.SCHEMA) String schema,
       @PathParam("fileset") @AuthorizationMetadata(type = Entity.EntityType.FILESET) String fileset,
       @QueryParam("sub_path") @DefaultValue("/") String subPath,
-      @QueryParam("location_name") String locationName)
-      throws UnsupportedEncodingException {
+      @QueryParam("location_name") String locationName) {
     LOG.info(
         "Received list files request: {}.{}.{}.{}, subPath: {}, locationName:{}",
         metalake,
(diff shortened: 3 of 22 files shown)


clients/client-java/src/main/java/org/apache/gravitino/client/FilesetCatalog.java
```diff
@@ -249,9 +249,9 @@ public String getFileLocation(NameIdentifier ident, String subPath, String locat
       CallerContext callerContext = CallerContext.CallerContextHolder.get();
 
       Map<String, String> params = new HashMap<>();
-      params.put("sub_path", RESTUtils.encodeString(subPath));
```
jerryshao, line 252: Java and Python http client will help to encode the query parameters automatically, doing this manually will introduce the double encoding problem.

roryqi:
What's the upgrade steps? 
1. Update the server when enabling compatible
2. New clients and old clients can access the server at the same time
3. Remove old clients
4. Switch off  the compatible the feature

For stage 2, we should handle the new clients and old clients at the same time. Now , we can't handle the new clients and old clients at the same time from the code.

jerryshao:
> What's the upgrade steps?
> 
> 1. Update the server when enabling compatible
> 2. New clients and old clients can access the server at the same time
> 3. Remove old clients
> 4. Switch off  the compatible the feature
> 
> For stage 2, we should handle the new clients and old clients at the same time. Now , we can't handle the new clients and old clients at the same time from the code.

Yeah, seems like the step 2 is the problem. For most of the case (no % in the string), encoding once and decoding twice is fine without issue, but we cannot guarantee all the corner cases.

jerryshao:
@mchades @xloya do you have any thoughts on this?

mchades:
> > What's the upgrade steps?
> > 
> > 1. Update the server when enabling compatible
> > 2. New clients and old clients can access the server at the same time
> > 3. Remove old clients
> > 4. Switch off  the compatible the feature
> > 
> > For stage 2, we should handle the new clients and old clients at the same time. Now , we can't handle the new clients and old clients at the same time from the code.
> 
> Yeah, seems like the step 2 is the problem. For most of the case (no % in the string), encoding once and decoding twice is fine without issue, but we cannot guarantee all the corner cases.

We can add a client version to the request header so that the server can identify the client version and no longer needs to rely on the `enableCompatibility` configuration manually.

xloya:
> > > 升级步骤是什么？
> > > 
> > > 1. 启用兼容时更新服务器
> > > 2. 新客户端和老客户端可以同时访问服务器
> > > 3. 删除旧客户端
> > > 4. 关闭兼容功能
> > > 
> > > 对于第二阶段，我们应该同时处理新客户和老客户。现在，从代码上来说，我们无法同时处理新客户和老客户。
> > 
> > 
> > 是的，看起来问题出在第 2 步。大多数情况下（字符串中没有 % 符号），编码一次解码两次是没问题的，但我们不能保证所有极端情况都能正常工作。
> 
> 我们可以在请求头中添加客户端版本，这样服务器就可以识别客户端版本，不再需要依赖`enableCompatibility`手动配置。

Yeah, I agree that adding the client version number in the header is a good way to solve the compatibility problem on the server side.

jerryshao:
> > > What's the upgrade steps?
> > > 
> > > 1. Update the server when enabling compatible
> > > 2. New clients and old clients can access the server at the same time
> > > 3. Remove old clients
> > > 4. Switch off  the compatible the feature
> > > 
> > > For stage 2, we should handle the new clients and old clients at the same time. Now , we can't handle the new clients and old clients at the same time from the code.
> > 
> > 
> > Yeah, seems like the step 2 is the problem. For most of the case (no % in the string), encoding once and decoding twice is fine without issue, but we cannot guarantee all the corner cases.
> 
> We can add a client version to the request header so that the server can identify the client version and no longer needs to rely on the `enableCompatibility` configuration manually.

What is the behavior if user doesn't specify the client version in the header? Shall we use the new or old API? Normally, it should use the new API, if so, seems we cannot do the backward compatibility (because old client doesn't have the version number in the header).

mchades:
> > > > What's the upgrade steps?
> > > > 
> > > > 1. Update the server when enabling compatible
> > > > 2. New clients and old clients can access the server at the same time
> > > > 3. Remove old clients
> > > > 4. Switch off  the compatible the feature
> > > > 
> > > > For stage 2, we should handle the new clients and old clients at the same time. Now , we can't handle the new clients and old clients at the same time from the code.
> > > 
> > > 
> > > Yeah, seems like the step 2 is the problem. For most of the case (no % in the string), encoding once and decoding twice is fine without issue, but we cannot guarantee all the corner cases.
> > 
> > 
> > We can add a client version to the request header so that the server can identify the client version and no longer needs to rely on the `enableCompatibility` configuration manually.
> 
> What is the behavior if user doesn't specify the client version in the header? Shall we use the new or old API? Normally, it should use the new API, if so, seems we cannot do the backward compatibility (because old client doesn't have the version number in the header).


Good question. The key point is that requests from the new and old versions will arrive at the server simultaneously, but the server can only choose one decoding method—either decoding once or twice.

After thinking twice, to solve the problem in phase 2, we can add a request header (the client version seems not suitable for this case, we can use for example `X-Query-Parameter-Encoded`), to control the server-side decoding method, and then the behavior becomes:

- If the `X-Query-Parameter-Encoded` header does not exist, the decoding method is determined by the server configuration `gravitino.server.api.enableCompatibility`
- if `X-Query-Parameter-Encoded=true`, we decode the query parameter **twice** and ignore the server configuration `gravitino.server.api.enableCompatibility`
-  if `X-Query-Parameter-Encoded=false`, we decode the query parameter **once** and ignore the server configuration `gravitino.server.api.enableCompatibility`
- the new version client uses `X-Query-Parameter-Encoded=false` by default

in this way, the upgrade steps become:

1. Update the server with a compatible configuration enabled
2. New clients and old clients can access the server simultaneously and will be processed through different methods.
3. Remove old clients
4. Switch off the compatible feature

WDYT?

roryqi:
My opinion is that we should add client version and client kind in the headers. It will be convenient if we have some issues to process the compatible issues.

It's ok for me if we can cover the majority of cases for this fix.  It make senses if the client occurs the bug and upgrade.

jerryshao:
I added a client version to the header. If the version number exists, I will check the version number to do the decoding accordingly. But since the old client doesn't have the version number in the header, so the change here still breaks the compatibility.

From my point, it is acceptable to break the compatibility:
1. It is 1.0 version, we can do a breaking change for major version.
2. This is a bug. We should fix it even if it breaks the compatibility.

server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
```diff
@@ -374,12 +368,20 @@ public Response getFileLocation(
...
-        RESTUtils.decodeString(subPath),
-        Optional.ofNullable(locationName).map(RESTUtils::decodeString).orElse(null));
+        subPath,
+        locationName);
     try {
       return Utils.doAs(
           httpRequest,
           () -> {
+            int[] clientVersion = Utils.getClientVersion(httpRequest);
+            boolean isV1PlusClient = clientVersion == null || clientVersion[0] >= 1;
```
roryqi, line 378: `Curl` command will be decoded two times?

clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
```diff
@@ -680,7 +682,14 @@ private void addRequestHeaders(
     // Some systems require the Content-Type header to be set even for empty-bodied requests to
     // avoid failures.
     request.setHeader(HttpHeaders.CONTENT_TYPE, bodyMimeType);
+    // Set the API version header
     request.setHeader(HttpHeaders.ACCEPT, VERSION_HEADER);
+
+    // Set the client version header
```
roryqi, line 688: Do we need add the Language type?

server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
jerryshao: It will only decode once if the version number is missing.

clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
jerryshao: What is the usage scenario?

server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
roryqi: Got it.

clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
roryqi: The Java client is broken. But python client is ok. Maybe we  only need to do some fixes for Java client.

jerryshao: What is the meaning?

roryqi: > What is the meaning?

Not related to this issue.

If Java or Python has some issues, the one is ok, another has some bugs.  It's useful for to have the client kind information. We can process them using different logic.

jerryshao: We can fix it separately.

mchades (approved):
lgtm.

jerqi (approved):

PR #8351, opened 2025-08-29
[#8346] fix(server): fix misleading log message in catalog drop operation

### LINKED ISSUE #8346: [Improvement] Correct CatalogOperations.dropCatalog to log a success message only when a catalog is dropped

In server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java, if a catalog fails to be dropped, it logs both failure and success.

### How should we improve?

Only log one message

KyleLin0927:

Move success log message inside conditional block to avoid misleading logs in `CatalogOperations.java` .

Fixes #8346

No.

- Ran `./gradlew clean build`


justinmclean (approved):
Thanks for this

Issue #8343, opened 2025-08-28
[Improvement] In GroupDetails.java catch correct exception


justinmclean:
### What would you like to be improved?

In clients/cli/src/main/java/org/apache/gravitino/cli/commands/GroupDetails.java handle it's catching NoSuchUserException

### How should we improve?

Change to NoSuchGroupException

PCloud63514:
Hi! @justinmclean
Can I take this issue?

justinmclean:
All yours

PR #8362, opened 2025-08-29
[#8343] fix: correct invalid exception handling in GroupDetails

PCloud63514:
- In GroupDetails.java, updated the catch clause to handle the correct exception:
  - Changed from NoSuchUserException → NoSuchGroupException.

- GroupDetails is currently catching NoSuchUserException by mistake, when it should be handling NoSuchGroupException.

Fix: #8343 

- No.
- This change only corrects internal exception handling logic, with no modification to public APIs or user-visible behavior.

- No new tests added, since this only fixes the caught exception type.


justinmclean (approved):
Thanks

PR #8363, opened 2025-08-30
[#5199] feat(client-python): add partitioning DTO serdes

tsungchih:

JsonUtils.java

- PartitioningSerializer
- PartitioningDeserializer

We need to support table partitioning, bucketing and sort ordering and indexes

#5199

No

Unit tests

### DIFF
--- a/clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
+++ b/clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
@@ -351,3 +351,124 @@ class TestPartitioningSerdes(unittest.TestCase):
             expected_serialized[PartitioningSerdes.ASSIGNMENTS_NAME],
             serialized[PartitioningSerdes.ASSIGNMENTS_NAME],
         )
+
+    def test_serdes_range_partitioning_dto_invalid_assignments(self):
+        json_string = f"""
+        {{
+            "{PartitioningSerdes.STRATEGY}": "{Partitioning.Strategy.RANGE.value}",
+            "{PartitioningSerdes.FIELD_NAME}": {json.dumps(TestPartitioningSerdes.field_name)},
+            "{PartitioningSerdes.ASSIGNMENTS_NAME}": "invalid_assignments"
+        }}
+        """
+        with self.assertRaisesRegex(
+            IllegalArgumentException,
+            "Cannot parse range partitioning from non-array assignments",
+        ):
+            PartitioningSerdes.deserialize(json.loads(json_string))
+
+    def test_serdes_range_partitioning_dto_invalid_range_assignment(self):
+        json_string = f"""
+        {{
+            "{PartitioningSerdes.STRATEGY}": "{Partitioning.Strategy.RANGE.value}",
+            "{PartitioningSerdes.FIELD_NAME}": {json.dumps(TestPartitioningSerdes.field_name)},
+            "{PartitioningSerdes.ASSIGNMENTS_NAME}": [
+                {{
+                    "{PartitioningSerdes.PARTITION_TYPE}": "list",
+                    "{PartitioningSerdes.PARTITION_NAME}": "p202204_California",
+                    "{PartitioningSerdes.PARTITION_PROPERTIES}": {{}},
+                    "{PartitioningSerdes.LIST_PARTITION_LISTS}": [
+                        [
+                            {{
+                                "{PartitioningSerdes.TYPE}": "literal",
+                                "{PartitioningSerdes.DATA_TYPE}": "date",
+                                "{PartitioningSerdes.LITERAL_VALUE}": "2022-04-01"
+                            }},
+                            {{
+                                "{PartitioningSerdes.TYPE}": "literal",
+                                "{PartitioningSerdes.DATA_TYPE}": "string",
+                                "{PartitioningSerdes.LITERAL_VALUE}": "Los Angeles"
+                            }}
+                        ]
+                    ]
+                }}
+            ]
+        }}
+        """
+        with self.assertRaisesRegex(
+            IllegalArgumentException,
+            "Cannot parse range partitioning from non-range assignment",
+        ):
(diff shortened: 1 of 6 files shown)


tsungchih:

unknowntpo:
@tsungchih I'll take a look today.

jerryshao (dismissed):

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
```diff
@@ -0,0 +1,540 @@
...
+            self.assertListEqual(
+                partitioning_dto.field_name(), deserialized.field_name()
+            )
+
+    def test_deserialize_single_field_partitioning_dto_from_json_string(self):
+        for strategy, partitioning_dto in self.single_field_partitioning_dtos.items():
+            json_string = f"""
+            {{
+                "{PartitioningSerdes.STRATEGY}": "{strategy.value}",
+                "{PartitioningSerdes.FIELD_NAME}": {json.dumps(partitioning_dto.field_name())}
```
unknowntpo: @tsungchih  At here, I think you intended to place a plain json string and test serialization / deserialization, but for this json string, you use `json.dumps()` to construct part of the string, which makes the string not so `PLAIN` and hard to understand its actual value at first glance.

Could we avoid this by just hard-coding `partitioning_dto.field_name()` ? Will this take too much effort ?

I think the rest of the test also have this problem.

tsungchih: @unknowntpo Perhaps we can use a dictionary instead of JSON string in the unit test. Here's an example.

```python
    def test_deserialize_single_field_partitioning_dto_from_json_string(self):
        for strategy, partitioning_dto in self.single_field_partitioning_dtos.items():
            expected_serialized = {
                PartitioningSerdes.STRATEGY: strategy.value,
                PartitioningSerdes.FIELD_NAME: partitioning_dto.field_name(),
            }
            deserialized = PartitioningSerdes.deserialize(expected_serialized)

            self.assertEqual(partitioning_dto.name(), deserialized.name())
            self.assertEqual(
                partitioning_dto.strategy().value, deserialized.strategy().value
            )
            self.assertListEqual(
                partitioning_dto.field_name(), deserialized.field_name()
            )
```

I was wondering if this would be more straightforward. How do you think??

unknowntpo: `expected_serialized` should not be python data structure, it should be plain json string.

I think just hard code entire json string is the best option.

```
            {
                "strategy": "identity",
                "fieldName": ["dummy_field_0"]
            }
```

tsungchih: @unknowntpo Please take a look into the latest commit when you're available. The plain JSON strings are used as the test data in the commit.

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
```diff
@@ -0,0 +1,569 @@
...
+        ):
+            self.assertRaisesRegex(
+                IOError,
+                "Unknown partitioning strategy",
+                PartitioningSerdes.serialize,
+                mock_dto,
+            )
+
+    def test_deserialize_invalid_json(self):
+        invalid_partitioning_data = (None, "invalid_data", {})
```
unknowntpo: can we make `invalid_partitioning_data` a JSON string, and use json.load to load it ?
Just like what `test_deserialize_identity_partitioning_dto` did.

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
```diff
@@ -0,0 +1,569 @@
...
+        for invalid_data in invalid_partitioning_data:
+            with self.assertRaisesRegex(
+                IllegalArgumentException, "Cannot parse partitioning from invalid JSON"
+            ):
+                PartitioningSerdes.deserialize(invalid_data)
+
+    def test_deserialize_invalid_strategy(self):
+        """Tests missing strategy and unknown partitioning strategy."""
+
+        invalid_data_base = {
```
unknowntpo: ditto

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
```diff
@@ -0,0 +1,569 @@
...
+
+        with self.assertRaisesRegex(
+            IOError,
+            "Unknown partitioning strategy",
+        ):
+            PartitioningSerdes.deserialize(invalid_strategy_data)
+
+    def test_serialize_single_field_partitioning_dto(self):
+        for partitioning_dto in self.single_field_partitioning_dtos.values():
+            serialized = PartitioningSerdes.serialize(partitioning_dto)
```
unknowntpo, line 116: dtto.

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
tsungchih: I moved the empty JSON object from `invalid_partitioning_data` as `invalid_json_string = "{}"`.

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
tsungchih: The test data for this case have been changed as plain JSON strings. Thanks for the comments.

clients/client-python/tests/unittests/dto/rel/test_partitioning_serdes.py
tsungchih: This test case is aimed at testing if the method `serialize()` is able to derive the expected python object of dictionary for each single field partitioning DTO. Its corresponding JSON string representation will be handled by the class extending `DataClassJsonMixin` and holding `PartitioningDTO` as a field.

unknowntpo:
@jerryshao LGTM.

Issue #8340, opened 2025-08-28
[Improvement] Removed redundant validation in PolicyEntity.validate


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/meta/PolicyEntity.java, both validatePolicy() and content().validate() are called. validatePolicy() allready calls content().validate().

### How should we improve?

Remove the redundant content().validate().

Jackeyzhe:
@justinmclean Hi, I'd like to work on it. can you assign it to me ?

justinmclean:
It's yours. If you have any questions, please ask.

PR #8365, opened 2025-08-30
[#8340][Improvement]Remove the redundant content().validate()

Jackeyzhe:
Remove the redundant content().validate() in core/src/main/java/org/apache/gravitino/meta/PolicyEntity.java

Fix: #8340 

no

use exist ut


justinmclean (approved):
Thanks for that

Issue #8322, opened 2025-08-28
[Improvement] ExpressionUtil.parseSortOrder incorrectly reads the sort direction so “ASC” and “DESC” are not interpreted correctly


justinmclean:
### What would you like to be improved?

ExpressionUtil.parseSortOrder incorrectly reads the sort direction so “ASC” and “DESC” are not interpreted correctly.

You can add this code to testSortOrderFiledToExpression to show this:
```
    sortOrderFiled = List.of("f1 ASC");
    sortOrders = ExpressionUtil.sortOrderFiledToExpression(sortOrderFiled);
    Assertions.assertEquals(1, sortOrders.length);
    Assertions.assertEquals(SortOrders.ascending(NamedReference.field("f1")), sortOrders[0]);
```

### How should we improve?

In parseSortOrder in trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/iceberg/ExpressionUtil.java change the group from 1 to 2.

yonghyeonpark:
@justinmclean Hi, I’d like to take this issue.

justinmclean:
All yours, if you need any help, just ask.

PR #8366, opened 2025-08-31
[#8322] improvement: fix sort direction parsing bug and add test case

yonghyeonpark:

- Fixed incorrect group index in `ExpressionUtil.parseSortOrder()` method and added test case to verify the fix.

- The bug caused "ASC" and "DESC" keywords to be incorrectly interpreted.

Fix: #8322 

- No.

- Added test case for "f1 ASC" parsing to `testSortOrderFiledToExpression()` method.


jerryshao:
@diqiu50 please help to review this PR.

Issue #8303, opened 2025-08-27
[Improvement] OAuth2TokenResponse.java validate method is missing a super call


justinmclean:
### What would you like to be improved?

In common/src/main/java/org/apache/gravitino/dto/responses/OAuth2TokenResponse.java validate method is missing a super call

### How should we improve?

addd  super.validate() call

JeonDaehong:
I would appreciate it if you could assign it to me.

justinmclean:
Thanks again for taking this on

JeonDaehong:
> Thanks again for taking this on

Thanks !

PR #8376, opened 2025-09-01
Add super.validate() call in OAuth2TokenResponse.validate() method

JeonDaehong:
**What changes were proposed in this pull request?**
In common/src/main/java/org/apache/gravitino/dto/responses/OAuth2TokenResponse.java, the validate() method was missing a call to super.validate(). This PR adds the super.validate() call to ensure that the parent class validation logic is executed before the subclass-specific checks.

**Why are the changes needed?**
Without calling super.validate(), the validation logic defined in BaseResponse is skipped, which could allow invalid responses to pass through. Adding the call ensures both parent and subclass validations are performed correctly.

**Does this PR introduce any user-facing change?**
No. This change only affects internal validation logic.

 **Fixes:** #8303

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
+++ b/server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
@@ -198,7 +198,14 @@ public class GravitinoServer extends ResourceConfig {
 
   public static void main(String[] args) {
     LOG.info("Starting Gravitino Server");
-    String confPath = System.getenv("GRAVITINO_TEST") == null ? "" : args[0];
+    String confPath = "";
+	if (System.getenv("GRAVITINO_TEST") != null) {
+		if (args.length < 1) {
+			LOG.error("Missing argument for GRAVITINO_TEST conf path.");
+			System.exit(1);
+		}
+		confPath = args[0];
+	}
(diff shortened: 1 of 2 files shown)


justinmclean:
You'll need to run Spottless to fix teh code formatting, please run `./gradlew :server:spotlessApply` and commit the changes.

server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
```diff
@@ -198,7 +198,14 @@ public void stop() throws IOException {
...
     LOG.info("Starting Gravitino Server");
-    String confPath = System.getenv("GRAVITINO_TEST") == null ? "" : args[0];
+    String confPath = "";
+	if (System.getenv("GRAVITINO_TEST") != null) {
+		if (args.length < 1) {
+			LOG.error("Missing argument for GRAVITINO_TEST conf path.");
+			System.exit(1);
+		}
+		confPath = args[0];
+	}
```
justinmclean: I don't think this change needs to be in this PR.

JeonDaehong:
@justinmclean 

Sorry, it seems some unrelated code from another PR was mixed in. I've revised it and pushed the corrected version.

jerqi (approved):

Issue #8390, opened 2025-09-02
[Improvement] In AuthorizationPluginException.java unexpected formatting errors can accur


justinmclean:
### What would you like to be improved?

In api/src/main/java/org/apache/gravitino/exceptions/AuthorizationPluginException.java, the message is prematurely formatted.

Here's a test to show that:
```
  @Test
  public void testPercentInMessageDoesNotCauseFormatting() {
    AuthorizationPluginException ex =
        assertDoesNotThrow(() -> new AuthorizationPluginException("100% done"));
    assertEquals("100% done", ex.getMessage());
  }
```

### How should we improve?

Don't format the message.

Surya00011:
Hi @justinmclean , I will take this

justinmclean:
Thanks for helping out on this.

PR #8396, opened 2025-09-02
[#8390] fix(api): remove premature String.format in AuthorizationPluginException

Surya00011:

- Removed the direct call to `String.format` in `AuthorizationPluginException` constructors.  
- Delegated `(message, args)` directly to `GravitinoRuntimeException`.  
- This ensures messages containing `%` (like `"100% done"`) are not misinterpreted as format specifiers when no args are passed.

Currently, `AuthorizationPluginException` always formats the message string.  
This causes errors when the message includes `%` but no formatting arguments are provided (e.g., `"100% done"`).  

By removing premature formatting and letting the parent class handle it, we avoid these errors and maintain expected behavior.

Fixes: #8390

Yes.  
- Exception messages are now preserved as-is if no formatting arguments are provided.  
- Example: `"100% done"` will no longer cause a `MissingFormatArgumentException`.


justinmclean:
Can you please add the test as well.

justinmclean (approved):
Looks good to me.

PR #8398, opened 2025-09-02
[#8305] Safely handle confPath in GravitinoServer.main()

### LINKED ISSUE #8305: [Improvement] Potential ArrayIndexOutOfBoundsException in GravitinoServer when checking for command-line arguments

In the main method of server/src/main/java/org/apache/gravitino/server/GravitinoServer.java as ArrayIndexOutOfBoundsException could occur.

### How should we improve?

Make it not possible to happen.

JeonDaehong:
**What would you like to be improved?**
In server/src/main/java/org/apache/gravitino/server/GravitinoServer.java, the main method could throw an ArrayIndexOutOfBoundsException if the GRAVITINO_TEST environment variable is set but args is empty.

**How should we improve?**
Check if args.length < 1 and log an error, then exit safely.

```java
String confPath = "";
if (System.getenv("GRAVITINO_TEST") != null) {
    if (args.length < 1) {
        LOG.error("Missing argument for GRAVITINO_TEST conf path.");
        System.exit(1);
    }
    confPath = args[0];
}
```
This ensures that even if GRAVITINO_TEST is set and args is empty, no exception will be thrown.

Fixes #8305

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
+++ b/server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
@@ -18,6 +18,7 @@
  */
 package org.apache.gravitino.server;
 
+import com.google.common.base.Preconditions;
 import java.io.File;
 import java.io.IOException;
 import java.util.HashSet;
@@ -198,12 +199,10 @@ public class GravitinoServer extends ResourceConfig {
 
   public static void main(String[] args) {
     LOG.info("Starting Gravitino Server");
+
     String confPath = "";
     if (System.getenv("GRAVITINO_TEST") != null) {
-      if (args.length < 1) {
-        LOG.error("Missing argument for GRAVITINO_TEST conf path.");
-        System.exit(1);
-      }
+      Preconditions.checkArgument(args.length >= 1, "Missing conf path argument");
       confPath = args[0];
     }
     ServerConfig serverConfig = loadConfig(confPath);


server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
```diff
@@ -198,7 +198,14 @@ public void stop() throws IOException {
 
   public static void main(String[] args) {
     LOG.info("Starting Gravitino Server");
-    String confPath = System.getenv("GRAVITINO_TEST") == null ? "" : args[0];
+    String confPath = "";
+    if (System.getenv("GRAVITINO_TEST") != null) {
+      if (args.length < 1) {
+        LOG.error("Missing argument for GRAVITINO_TEST conf path.");
```
jerryshao: The log message is not correct, this is not for GRAVITINO_TEST.

JeonDaehong: I’ve checked it. I’ll make the changes. Thanks for the feedback.

jerryshao:
Please update the PR title to follow the convention.

server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
```diff
@@ -198,7 +198,14 @@ public void stop() throws IOException {
...
+        System.exit(1);
+      }
```
Abyss-lord: Using `Preconditions` might be a cleaner way to handle this. WDYT
```java
if (System.getenv("GRAVITINO_TEST") != null) {
    Preconditions.checkArgument(
        args.length >= 1,
        "Missing conf path."
    );
}
```

JeonDaehong:
@justinmclean 

I’d like to contribute more actively to Gravitino and communicate with committers or PMC members. Is there a Slack or Discord community available?

server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
JeonDaehong: I will resolve this issue within today or tomorrow.
Thank you for your comment

JeonDaehong:
@jerryshao @Abyss-lord @justinmclean 

I have completed the modifications.

Issue #8385, opened 2025-09-02
[Improvement] TestJsonCodec.java is not thread safe


justinmclean:
### What would you like to be improved?

trino-connector/trino-connector/src/test/java/org/apache/gravitino/trino/connector/util/json/TestJsonCodec.java is not thread safe.

### How should we improve?

Make thread safe by making the mapper and JSON type volatile?

Surya00011:
Hi @justinmclean sir, I'll Wrk on this.

justinmclean:
assigned to you

PR #8399, opened 2025-09-02
[#8385] Make JsonCodec.java is thread safe

Surya00011:

- Made `JsonCodec` thread-safe by marking `mapper` and `jsonType` as `volatile`.
- Applied double-checked locking in `getMapper` and `getJsonType` to ensure proper lazy initialization in concurrent scenarios.
- No functional changes to the JSON serialization/deserialization behavior.

- `JsonCodec` was not thread-safe: concurrent access to `mapper` and `jsonType` could cause race conditions or inconsistent initialization.

Fix: #8385

### Does this PR introduce  any user-facing change?

- No user-facing API changes.
- No properties or configuration keys were added or removed.
- Existing functionality behaves identically, but now safely under concurrent access.

### DIFF
--- a/docs/fileset-catalog-with-adls.md
+++ b/docs/fileset-catalog-with-adls.md
@@ -306,7 +306,7 @@ Or use the bundle jar with Hadoop environment if there is no Hadoop environment:
 
 ### Using Spark to access the fileset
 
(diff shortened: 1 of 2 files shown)


justinmclean (approved):
Thanks for fixing this

Issue #8388, opened 2025-09-02
[Improvement] Pass ignoreNotExist to dropCatalog in dropCatalog in DropCatalogStoredProcedure.java


justinmclean:
### What would you like to be improved?

In trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/system/storedprocedure/DropCatalogStoredProcedure.java ignoreNotExist is not passed on.

### How should we improve?

Change true to ignoreNotExist.

Raj-G07:
@justinmclean Can i take up this issue?

justinmclean:
Sure, if you need any help just ask.

Raj-G07:
@justinmclean Created a PR

PR #8401, opened 2025-09-02
[IMP]:Fixed the Javadoc in dropCatalog in DropCatalogStoredProcedure.java

Raj-G07:

Fix: #8388


Raj-G07:
@justinmclean What am i missing?

justinmclean:
I think this may be an issue with an existing test that is trying to drop a catalog while it is still in use. This change has changed the flag from true to false and thus we get an exception.

Raj-G07:
@justinmclean So it is okay to call a function only if ignoreNotExist is true , 
```
   if(ignoreNotExist){ 
      catalogConnector.getMetalake().dropCatalog(catalogName, ignoreNotExist);
   };

**Comment by diqiu50** [comment]
@Raj-G07 Please fix the description of the pr

File: trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/system/storedprocedure/DropCatalogStoredProcedure.java [hunk 1]
```diff
@@ -97,7 +97,7 @@ public void dropCatalog(String catalogName, boolean ignoreNotExist) {
             GravitinoErrorCode.GRAVITINO_CATALOG_NOT_EXISTS,
             "Catalog " + NameIdentifier.of(metalake, catalogName) + " not exists.");
       }
-      catalogConnector.getMetalake().dropCatalog(catalogName, true);
+      catalogConnector.getMetalake().dropCatalog(catalogName, ignoreNotExist);
```
diqiu50: The second parameter is not `ignoreExist`,  The `true` means  force drop

Raj-G07:
@diqiu50 
What’s the exact use case of the `ignoreNotExist` params here?
I want to understand in which scenarios we prefer ignoring non-existent resources instead of throwing an error.

diqiu50:
> @diqiu50 What’s the exact use case of the `ignoreNotExist` params here? I want to understand in which scenarios we prefer ignoring non-existent resources instead of throwing an error.

The above code implemented the logic of ignore non-existence

diqiu50 (approved):
LGTM

Issue #8386, opened 2025-09-02
[Improvement] Add precodition to JdbcUrlUtils.java to check driver and url not null


justinmclean:
### What would you like to be improved?

 Add precodition to common/src/test/java/org/apache/gravitino/utils/TestJdbcUrlUtils.java to check that thedriver and url parameters are not null.

### How should we improve?

as above

Abyss-lord:
I would like to work on it.

justinmclean:
All yours.

PR #8405, opened 2025-09-03
[#8386] fix(common): Add precodition to JdbcUrlUtils.java to check driver and url not null

Abyss-lord:

Add precodition to JdbcUrlUtils.java to check driver and url not null

Fix: #8386 

No.

local test.


Abyss-lord:
Hi @justinmclean , could you please review this PR when you have time? I’d really appreciate your feedback.

common/src/main/java/org/apache/gravitino/utils/JdbcUrlUtils.java
```diff
@@ -68,18 +69,19 @@ private JdbcUrlUtils() {
    * @param all the JDBC configuration properties
    */
   public static void validateJdbcConfig(String driver, String url, Map<String, String> all) {
+    Preconditions.checkArgument(driver != null, "Driver class name cannot be null");
```
yuqi1129: To be more exact, these two can't be empty, so it's better to use `StringUtils.isNotEmpty()`.

Abyss-lord: @yuqi1129 I think it’s better to use `StringUtils.isBlank()` instead of `StringUtils.isNotEmpty()`. The latter only checks for null or zero-length strings, while in this case we need to ensure that both `driver` and `url` are not blank (i.e., not null, not empty, and not just whitespace).

image

Abyss-lord: Fix

common/src/main/java/org/apache/gravitino/utils/JdbcUrlUtils.java
```diff
@@ -68,18 +70,19 @@ private JdbcUrlUtils() {
    * @param all the JDBC configuration properties
    */
   public static void validateJdbcConfig(String driver, String url, Map<String, String> all) {
+    Preconditions.checkArgument(StringUtils.isNotBlank(driver), "Driver class name cannot be null");
```
yuqi1129: Driver class name cannot be null -> can't be blank

Abyss-lord: > Driver class name cannot be null -> can't be blank

Fix

justinmclean (approved):
Looks good thanks for doing this.

Issue #8383, opened 2025-09-02
[Improvement] Fix malformed SQL in PolicyMetaPostgreSQLProvider.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/PolicyMetaPostgreSQLProvider.java the SQL in insertPolicyMetaOnDuplicateKeyUpdate is misisng a commin in teh second last row.

### How should we improve?

Add the needed comma.

Ashwil-Colaco:
hi @justinmclean please check my PR

PR #8406, opened 2025-09-03
fix/ added a missing comma in insertPolicyMetaOnDuplicateKeyUpdate in PolicyMetaPostgreSQLProvider.java

Ashwil-Colaco:
Title:

[#8383] fix(core): added missing comma in SQL of PolicyMetaPostgreSQLProvider.java

Added a missing comma in the SQL statement of insertPolicyMetaOnDuplicateKeyUpdate in PolicyMetaPostgreSQLProvider.java.

Without this comma, the SQL would fail or behave incorrectly when updating policy metadata in PostgreSQL. This fixes a small bug that can cause runtime errors.

Fix: #8383

There are no changes to APIs or configuration visible to end users. Hence, it doesn't introduce _any_ user-facing change

Applied Spotless formatting 
Ran ./gradlew build -x test.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/PolicyMetaPostgreSQLProvider.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/PolicyMetaPostgreSQLProvider.java
@@ -75,7 +75,7 @@ public class PolicyMetaPostgreSQLProvider extends PolicyMetaBaseSQLProvider {
         + " metalake_id = #{policyPO.metalakeId},"
         + " audit_info = #{policyPO.auditInfo},"
         + " current_version = #{policyPO.currentVersion},"
-        + " last_version = #{policyPO.lastVersion}"
+        + " last_version = #{policyPO.lastVersion},"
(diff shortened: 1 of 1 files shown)


jerryshao:
I don't know why this is not reflected by test before, @mchades can you review the PR? Besides, I think we need a test to cover this.

mchades:
> I don't know why this is not reflected by test before

The method `insertPolicyMetaOnDuplicateKeyUpdate` is only called when using an insert policy with `overwritten=true`, but currently, no one is using it—alter policy uses the `updatePolicyMeta` method instead. Therefore, the bug cannot be detected yet.

> Besides, I think we need a test to cover this.

Since the policy has no operation to trigger the case, the policyIT cannot cover it. By default, the Mapper's UT only tests H2. To make the Mapper UT cover different backend storages, this should be done in a separate PR.

So, the current fix looks good to me. @Ashwil-Colaco, please update the PR title to follow the correct template.

Ashwil-Colaco:
@mchades,Thanks for the review

Ashwil-Colaco:
@mchades ,please check my Title now

Ashwil-Colaco:
Thank you @mchades

PR #8413, opened 2025-09-03
[MINOR] improve(CI): Increase the CI timeout for Python and Ranger test

jerryshao:

Increase the timeout to 90 minutes for Python and Ranger CI Test.

To make the CI more stable without rerun.

No.

N/A


justinmclean:
I think the ASF infrastructure will cancel any job that runs for more than 1 hour, so this may not work.

Issue #8395, opened 2025-09-02
[Improvement] Add a distributed environment for the trino connector integration test


hqbhoho:
### What would you like to be improved?

Maybe we can add a configuration trino-worker-num (default 0) to facilitate deploying a distributed environment for testing purposes. If trino-worker-num = 0, it will use the current implementation which use a single node that functions as both a coordinator and worker. It will not change default CI behavior.

```bash trino-connector/integration-test/trino-test-tools/trino_test.sh --test_set=jdbc-mysql --trino-worker-num=3```

PR #8419, opened 2025-09-03
[#8395] improvement(trino-connector): Support integration test run in a distributed environment

hqbhoho:
Add a configuration `trino_worker_num` (default 0) to facilitate deploying a distributed environment for testing purposes.

Fix: #8395 

Run Trino integration tests with a distributed environment which have 3 workers.
`bash trino-connector/integration-test/trino-test-tools/trino_test.sh --test_set=jdbc-mysql --trino_worker_num=3`

local test

### DIFF
--- a/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlSchemaOperations.java
+++ b/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlSchemaOperations.java
@@ -76,8 +76,8 @@ public class PostgreSqlSchemaOperations extends JdbcDatabaseOperations {
   @Override
   public List<String> listDatabases() {
     List<String> result = new ArrayList<>();
-    try (Connection connection = getConnection()) {
-      ResultSet resultSet = getSchema(connection, null);
+    try (Connection connection = getConnection();
+        ResultSet resultSet = getSchema(connection, null)) {
       while (resultSet.next()) {
         String schemaName = resultSet.getString(1);
         if (!isSystemDatabase(schemaName)) {
@@ -162,10 +162,11 @@ public class PostgreSqlSchemaOperations extends JdbcDatabaseOperations {
   }
 
   public boolean schemaExists(Connection connection, String schema) throws SQLException {
-    ResultSet resultSet = getSchema(connection, schema);
-    while (resultSet.next()) {
-      if (Objects.equals(resultSet.getString(1), schema)) {
-        return true;
+    try (ResultSet resultSet = getSchema(connection, schema)) {
+      while (resultSet.next()) {
+        if (Objects.equals(resultSet.getString(1), schema)) {
+          return true;
+        }
       }
     }
     return false;
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataIdConverter.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataIdConverter.java
@@ -37,6 +37,7 @@ import org.apache.gravitino.meta.BaseMetalake;
 import org.apache.gravitino.meta.CatalogEntity;
 import org.apache.gravitino.meta.ColumnEntity;
 import org.apache.gravitino.meta.FilesetEntity;
+import org.apache.gravitino.meta.GroupEntity;
 import org.apache.gravitino.meta.ModelEntity;
 import org.apache.gravitino.meta.ModelVersionEntity;
 import org.apache.gravitino.meta.RoleEntity;
@@ -85,7 +86,7 @@ public class MetadataIdConverter {
           .put(Entity.EntityType.MODEL_VERSION, ModelVersionEntity.class)
           .put(Entity.EntityType.COLUMN, ColumnEntity.class)
           .put(Entity.EntityType.USER, UserEntity.class)
-          .put(Entity.EntityType.GROUP, Entity.class)
+          .put(Entity.EntityType.GROUP, GroupEntity.class)
           .put(Entity.EntityType.ROLE, RoleEntity.class)
           .build();
 
--- a/server-common/src/test/java/org/apache/gravitino/server/authorization/TestMetadataIdConverter.java
+++ b/server-common/src/test/java/org/apache/gravitino/server/authorization/TestMetadataIdConverter.java
@@ -46,6 +46,7 @@ import org.apache.gravitino.meta.BaseMetalake;
 import org.apache.gravitino.meta.CatalogEntity;
 import org.apache.gravitino.meta.ColumnEntity;
 import org.apache.gravitino.meta.FilesetEntity;
+import org.apache.gravitino.meta.GroupEntity;
 import org.apache.gravitino.meta.ModelEntity;
 import org.apache.gravitino.meta.SchemaEntity;
 import org.apache.gravitino.meta.SchemaVersion;
@@ -68,6 +69,7 @@ public class TestMetadataIdConverter {
   private NameIdentifier ident5;
   private NameIdentifier ident6;
   private NameIdentifier ident7;
+  private NameIdentifier ident8;
 
   // Test Entities
   private BaseMetalake entity1;
@@ -77,6 +79,7 @@ public class TestMetadataIdConverter {
   private ModelEntity entity5;
   private FilesetEntity entity6;
   private TopicEntity entity7;
+  private GroupEntity entity8;
 
   @BeforeAll
   void initTest() throws IOException {
@@ -86,7 +89,7 @@ public class TestMetadataIdConverter {
   }
 
   @Test
-  void testConvert() throws IOException, IllegalAccessException {
+  void testConvert() throws IllegalAccessException {
     CatalogManager mockCatalogManager = mock(CatalogManager.class);
     Object originalCatalogManager =
         FieldUtils.readDeclaredField(GravitinoEnv.getInstance(), "catalogManager", true);
@@ -141,6 +144,12 @@ public class TestMetadataIdConverter {
                   MetadataIdConverter.normalizeCaseSensitive(
                       eq(ident7), eq(Capability.Scope.TOPIC), eq(mockCatalogManager)))
           .thenReturn(ident7);
+      mockedStatic
+          .when(
+              () ->
+                  MetadataIdConverter.normalizeCaseSensitive(
+                      eq(ident8), eq(null), eq(mockCatalogManager)))
+          .thenReturn(ident8);
 
       Long metalakeConvertedId =
           MetadataIdConverter.getID(
@@ -198,6 +207,7 @@ public class TestMetadataIdConverter {
     ident5 = NameIdentifier.of("metalake", "catalog", "schema", "model");
     ident6 = NameIdentifier.of("metalake", "catalog", "schema", "fileset");
     ident7 = NameIdentifier.of("metalake", "catalog", "schema", "topic");
+    ident8 = NameIdentifier.of("metalake", "group");
   }
 
   private void initTestEntities() {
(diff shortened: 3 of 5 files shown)


integration-test-common/docker-script/docker-compose.yaml
```diff
@@ -106,6 +107,31 @@ services:
...
+      - ../../trino-connector/trino-connector/build/libs:/usr/lib/trino/plugin/gravitino
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    depends_on:
+      hive:
+        condition: service_healthy
+      postgresql:
+        condition: service_healthy
+      mysql:
+        condition: service_healthy
```
diqiu50, line 138: Do we need to add depends on Trino worker node  in the Trino coordinator node

integration-test-common/docker-script/init/trino/init-worker.sh
```diff
@@ -0,0 +1,50 @@
...
+mkdir -p /tmp/data
+
+# Check the number of Gravitino connector plugins present in the Trino plugin directory
+num_of_gravitino_connector=$(ls /usr/lib/trino/plugin/gravitino | grep gravitino-trino-connector-* | wc -l)
+if [[ "${num_of_gravitino_connector}" -ne 1 ]]; then
+  echo "Multiple versions of the Gravitino connector plugin found or none present."
+  exit 1
+fi
+
+/usr/lib/trino/bin/run-trino
```
diqiu50: It may be possible to reuse the init script and add an environment variable to distinguish different logic.

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryITBase.java
```diff
@@ -103,7 +113,10 @@ public void setup() throws Exception {
 
     setEnv();
     trinoQueryRunner = new TrinoQueryRunner(trinoUri);
-    createMetalake();
+
+    if (!autoStart) {
+      createMetalake();
+    }
```
diqiu50: Why did you make this change?

hqbhoho: Beacuse `loadMetalake` will call `system.exit(-1)`  if metalake not exists. So we need create metalake  before the Trino cluster is deployed.

integration-test-common/docker-script/docker-compose.yaml
hqbhoho: done

integration-test-common/docker-script/init/trino/init-worker.sh
hqbhoho: done

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryTestTool.java
```diff
@@ -85,6 +85,13 @@ public static void main(String[] args) throws Exception {
           "Additional parameters that can replace the value of ${key} in the testers contents, "
               + "example: --params=key1,v1;key2,v2");
 
+      options.addOption(
+          "trino_worker_num",
+          true,
+          "Specify the number of Trino worker, the default value is 0."
```
yuqi1129: What's the default value of `trino_worker_num`? I suggest we run in deploy mode by default, as this is what the production deploys.

hqbhoho: Currently,  it will use a single node that functions as both a coordinator and worker in default CI behavior.

In production deploy,  it will use independent worker. There is no limit to the number of Trino workers but at least one worker is required. 

More worker may need more resources, we do not prefer change default CI behavior. So I add a configuration `trino-worker-num` to facilitate deploying a distributed environment for testing purposes.  If `trino-worker-num` = 0, it will use the current implementation which use a single node that functions as both a coordinator and worker.

If run in a distributed environment is necessary, you can run  integration tests with a distributed environment which have 3 independent workers like this:
```
bash trino-connector/integration-test/trino-test-tools/trino_test.sh --test_set=jdbc-mysql --trino_worker_num=3
```

diqiu50: Is better the default value is 1 . The coordinator node is include one work in the current configuration

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryITBase.java
diqiu50: I think invoking createMetalake is mandatory and is not related to whether it starts automatically.

diqiu50: I think it is not a good practice to execute the test database initialization script when the container starts. This is because we also have other containerized test environments, and tests may fail due to the absence of these scripts.

diqiu50: Currently, in each test suite we have a catalog_xxx_prepare.sql, which is executed when testing the catalog. We can add a catalog_xxx_init.sh to be executed before catalog_xxx_prepare.sql. Correspondingly, a catalog_xxx_stop.sh can be executed after catalog_xxx_cleanup.sql. In this way, the environment and the tests are separated.

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryTestTool.java
hqbhoho: > Is better the default value is 1 . The coordinator node is indclude one work in the current configuration

- 0 is a single node that functions as both a coordinator and worker(current ci behavior). This scenario can be considered as having no independent workers, coordinator work as worker.
- 1  is two node which one is coordinator and one is independent worker.

Or change the name `trino_worker_num` to `trino_independent_worker_num`?

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryITBase.java
diqiu50: Here we may need to enhance the functionality of TrinoITContainers to send the script to the container for execution.
Could you modify it according to this approach?

hqbhoho: > I think invoking createMetalake is mandatory and is not related to whether it starts automatically.

I don't change the original logic. If autoStart, I add `createMetalake()` in `setEnv()`.
Maybe we can add all`createMetalake()` in `setEnv()` in this PR
```
private void setEnv() throws Exception {
    baseIT = new BaseIT();
    if (autoStart) {
      baseIT.startIntegrationTest();
      gravitinoClient = baseIT.getGravitinoClient();
      gravitinoUri = String.format("http://127.0.0.1:%d", baseIT.getGravitinoServerPort());
      // The creation of a metalake must be completed before the Trino cluster is deployed.
      createMetalake();

      trinoITContainers = ContainerSuite.getTrinoITContainers();
      trinoITContainers.launch(baseIT.getGravitinoServerPort(), trinoWorkerNum);

      trinoUri = trinoITContainers.getTrinoUri();
      hiveMetastoreUri = trinoITContainers.getHiveMetastoreUri();
      hdfsUri = trinoITContainers.getHdfsUri();
      mysqlUri = trinoITContainers.getMysqlUri();
      postgresqlUri = trinoITContainers.getPostgresqlUri();

    } else if (autoStartGravitino) {
      baseIT.startIntegrationTest();
      gravitinoClient = baseIT.getGravitinoClient();
      gravitinoUri = String.format("http://127.0.0.1:%d", baseIT.getGravitinoServerPort());
      createMetalake();
    } else {
      gravitinoClient = GravitinoAdminClient.builder(gravitinoUri).build();
      createMetalake();
    }
  }
```
Furthermore, `create metalake` maybe  a part of the Gravitino server-side startup for test. I think it should create a new PR for this.

Currently, I run the integration test failed sometimes in local, Trino master will exit because of metalake not exists.

diqiu50: Why does it exit? Do you have any error messages? I think they have no dependency relationship.

hqbhoho: > Why does it exit? Do you have any error messages? I think they have no dependency relationship.

`CatalogConnectorManager` will `loadMetalake` in a schedule thread. If metalake is not exists, it will exit.
```
private void loadMetalake() {
    try {
      if (!catalogRegister.isTrinoStarted()) {
        LOG.info("Waiting for the Trino started.");
        return;
      }

      Set<String> usedMetalakes = new HashSet<>();
      if (config.singleMetalakeMode()) {
        usedMetalakes.add(targetMetalake);
        metalakes.computeIfAbsent(targetMetalake, this::retrieveMetalake);
      } else {
        GravitinoMetalake[] allMetalakes = gravitinoClient.listMetalakes();
        for (GravitinoMetalake metalake : allMetalakes) {
          usedMetalakes.add(metalake.name());
          metalakes.put(metalake.name(), metalake);
        }
      }

      for (String usedMetalake : usedMetalakes) {
        try {
          GravitinoMetalake metalake = metalakes.get(usedMetalake);
          LOG.debug("Load metalake: {}", usedMetalake);
          loadCatalogs(metalake);
        } catch (Exception e) {
          LOG.error("Load Metalake {} failed.", usedMetalake, e);
        }
      }
    } catch (Exception e) {
      LOG.error("Error when loading metalake", e);
      System.exit(-1);
    }
  }
 ```
 
 Currently, `createMetalake` is called after the container starts. So Trino master will exit if `loadMetalake` is called when `createMetalake`  is not called, but the failures are very sporadic. Maybe `System.exit(-1)`should be removed.
 
 The creation of a metalake before deploying Trino cluster  is better.

diqiu50: I think we need to remove System.exit(-1). When the Gravitino server is not started, it should not cause the Trino server to exit.

hqbhoho: Yes, any connector error should not cause the Trino server to exit. Create an new PR to fix it or fix in this PR？
After remove `System.exit(-1)`, `createMetalake()` maybe not need change.

diqiu50: Fix in this PR

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryTestTool.java
diqiu50: I understand your point. By default, there is one worker, and there is no need to start a separate worker node.

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryITBase.java
hqbhoho: done

trino-connector/integration-test/src/test/java/org/apache/gravitino/trino/connector/integration/test/TrinoQueryTestTool.java
hqbhoho: I update the doc `Specify the number of Trino independent worker`.

Issue #8411, opened 2025-09-03
[Improvement] AuditInfo.merge nulls out existing values when merging with overwrite enabled


justinmclean:
### What would you like to be improved?

In core/src/test/java/org/apache/gravitino/meta/TestAuditInfo.java AuditInfo.merge nulls out existing values when merging with overwrite enabled. It should keep those values.

Here a unit test to show this:
```
  @Test
  public void testMergeDoesNotOverwriteWithNull() {
    Instant now = Instant.now();
    AuditInfo original =
        AuditInfo.builder()
            .withCreator("creator")
            .withCreateTime(now)
            .withLastModifier("modifier")
            .withLastModifiedTime(now)
            .build();
    AuditInfo other = AuditInfo.builder().withLastModifier("new").build();

    original.merge(other, true);

    Assertions.assertEquals("creator", original.creator());
    Assertions.assertEquals(now, original.createTime());
    Assertions.assertEquals("new", original.lastModifier());
    Assertions.assertEquals(now, original.lastModifiedTime());
  }
```

### How should we improve?

Don't copy across nulls.

PR #8420, opened 2025-09-03
[#8411] improvement(core): Fix AuditInfo.merge nulling out existing values

mbaurin:

Fixed the `AuditInfo.merge()` method to prevent null values from overwriting existing non-null values when `overwrite` is set to `true`. The merge logic was updated to only overwrite existing values when the source value is non-null.

The current implementation of `AuditInfo.merge()` has a bug where it nulls out existing non-null values when merging with `overwrite=true`. This happens because the method unconditionally takes values from the source object when overwrite is enabled, even if those values are null.

This bug affects multiple EntityCombined classes (Table, Schema, Fileset, Topic, Model, ModelVersion) that use this method with `overwrite=true`, potentially causing loss of audit information across the system.

Fix: [#(issue)](https://github.com/apache/gravitino/issues/8411)

No. This is an internal bug fix that preserves existing audit information correctly. The behavior change ensures that audit fields are not unexpectedly nulled out, which maintains data integrity but doesn't change any APIs.

Added a new unit test `testMergeDoesNotOverwriteWithNull()` in `TestAuditInfo.java` that verifies:
- Non-null values are preserved when merging with null values (with overwrite=true)
- Non-null values from the source correctly overwrite existing values
- The fix maintains backward compatibility for the non-overwrite case

All existing tests pass, confirming no regression in functionality.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/meta/AuditInfo.java
+++ b/core/src/main/java/org/apache/gravitino/meta/AuditInfo.java
@@ -157,11 +157,20 @@ public final class AuditInfo implements Audit, Entity {
       return this;
     }
 
-    this.creator = overwrite || this.creator == null ? other.creator : creator;
-    this.createTime = overwrite || this.createTime == null ? other.createTime : createTime;
-    this.lastModifier = overwrite || this.lastModifier == null ? other.lastModifier : lastModifier;
+    this.creator =
+        (overwrite && other.creator != null) || this.creator == null ? other.creator : creator;
+    this.createTime =
+        (overwrite && other.createTime != null) || this.createTime == null
+            ? other.createTime
+            : createTime;
+    this.lastModifier =
+        (overwrite && other.lastModifier != null) || this.lastModifier == null
+            ? other.lastModifier
+            : lastModifier;
     this.lastModifiedTime =
-        overwrite || this.lastModifiedTime == null ? other.lastModifiedTime : lastModifiedTime;
+        (overwrite && other.lastModifiedTime != null) || this.lastModifiedTime == null
+            ? other.lastModifiedTime
+            : lastModifiedTime;
 
     return this;
   }
--- a/core/src/test/java/org/apache/gravitino/meta/TestAuditInfo.java
+++ b/core/src/test/java/org/apache/gravitino/meta/TestAuditInfo.java
@@ -82,4 +82,24 @@ public class TestAuditInfo {
     Assertions.assertEquals("first", first.creator());
     Assertions.assertEquals("second", second.creator());
   }
+
+  @Test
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/gravitino/meta/AuditInfo.java
```diff
@@ -157,11 +157,20 @@ public AuditInfo merge(AuditInfo other, boolean overwrite) {
...
+            : createTime;
+    this.lastModifier =
+        (overwrite && other.lastModifier != null) || this.lastModifier == null
+            ? other.lastModifier
+            : lastModifier;
     this.lastModifiedTime =
-        overwrite || this.lastModifiedTime == null ? other.lastModifiedTime : lastModifiedTime;
+        (overwrite && other.lastModifiedTime != null) || this.lastModifiedTime == null
+            ? other.lastModifiedTime
+            : lastModifiedTime;
```
jerryshao, line 173: This break the original meaning of "overwrite", I don't think the fix is correct and necessary. This is a semantic change, I don't know why do you think this is a bug and change this.

justinmclean: No It doesn't and I think it's needed - take a look at the unit test:
```
    AuditInfo original =
        AuditInfo.builder()
            .withCreator("creator")
            .withCreateTime(now)
            .withLastModifier("modifier")
            .withLastModifiedTime(now)
            .build();
    AuditInfo other = AuditInfo.builder().withLastModifier("new").build();

    original.merge(other, true);
```

Someone would not expect all the other fields, including the creation time, to be set to null. It's quite likely we are losing audit information this way.

justinmclean: And if you set it to false, the modifier field will not be updated.

jerryshao: It is not whether the field is null or not. It is who takes the precedence. 

Beside, normally it will not happen that all the fields are null based on the current code. It only happens in your mimic case.

Anyway, this is a behavior change, and I'm not in favor of this change.

jerryshao: In the real case, one audit info is from the underlying sources, another is stored in Gravitino, if the underlying one is empty/null, we will overwrite with the one stored in Gravitino. If the underlying one is not null / empty, we will always use the underlying one.

Besides, the one stored in Gravitino will always have the necessary fields, so it will not meet the case all the fields are null.

jerryshao:
Hi @mbaurin thanks a lot for your contribution. I'm going to close this issue. You can work on other issues if you'd like to.

Issue #8431, opened 2025-09-04
[Improvement] Remove unnecessary relocate in gcp/adls bundle jars


yuqi1129:
### What would you like to be improved?

Some packages like `com.sun.xml` does not exists in `azure-bundle.jar`, so we do not add the following line 
```
  relocate("com.sun.xml", "org.apache.gravitino.azure.shaded.com.sun.xml")
```

in the azure-bundle/build.gradle.kts

Issue #8391, opened 2025-09-02
[Bug report] Can't run Spark to access azure as document describe


yuqi1129:
### Version

main branch

### Describe what's wrong

1. Remove hadoop related jars from pyspark class path.
2. do the following code with JDK17, pyspark 3.5.0 as document says: https://gravitino.apache.org/docs/0.9.1/hadoop-catalog-with-adls#using-spark-to-access-the-fileset
```java

import logging
logging.basicConfig(level=logging.INFO)

from gravitino import NameIdentifier, GravitinoClient, Catalog, Fileset, GravitinoAdminClient

gravitino_url = "http://localhost:8090"
metalake_name = "test"

catalog_name = "azure_catalog"
schema_name = "schema"
fileset_name = "fileset01"

fileset_ident = NameIdentifier.of(schema_name, fileset_name)

gravitino_admin_client = GravitinoAdminClient(uri=gravitino_url)
gravitino_client = GravitinoClient(uri=gravitino_url, metalake_name=metalake_name)
from pyspark.sql import SparkSession
import os

os.environ["PYSPARK_SUBMIT_ARGS"] = "--jars /Users/yuqi/project/graviton/bundles/azure-bundle/build/libs/gravitino-azure-bundle-1.0.0-SNAPSHOT.jar,/Users/yuqi/project/graviton/clients/filesystem-hadoop3-runtime/build/libs/gravitino-filesystem-hadoop3-runtime-1.0.0-SNAPSHOT.jar --conf \"spark.driver.extraJavaOptions=--add-opens=java.base/sun.nio.ch=ALL-UNNAMED -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005\" --conf \"spark.executor.extraJavaOptions=--add-opens=java.base/sun.nio.ch=ALL-UNNAMED -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005\" --master local[1] pyspark-shell"

os.environ["HADOOP_USER_NAME"] = "anonymous"

spark = SparkSession.builder \
    .appName("s3_fielset_test") \
    .config("spark.hadoop.fs.AbstractFileSystem.gvfs.impl", "org.apache.gravitino.filesystem.hadoop.Gvfs") \
    .config("spark.hadoop.fs.gvfs.impl", "org.apache.gravitino.filesystem.hadoop.GravitinoVirtualFileSystem") \
[...]

PR #8432, opened 2025-09-04
[#8431] improvement(bundles): Remove unnecessary relocate in fileset bundle jars.

yuqi1129:

Remove unnecessary relocate in gcp/adls bundle jars.

To avoid potential problem.

Fix: #8431 
Fix: #8391

N/A.

Existing tests.


docs/fileset-catalog-with-adls.md
```diff
@@ -328,7 +328,7 @@ schema_name = "your_adls_schema"
 fileset_name = "your_adls_fileset"
 # JDK8
 os.environ["PYSPARK_SUBMIT_ARGS"] = "--jars /path/to/gravitino-azure-{gravitino-version}.jar,/path/to/gravitino-filesystem-hadoop3-runtime-{gravitino-version}.jar,/path/to/hadoop-azure-3.2.0.jar,/path/to/azure-storage-7.0.0.jar,/path/to/wildfly-openssl-1.0.4.Final.jar --master local[1] pyspark-shell"
-# JDK17
+# JDK17, PySpark 3.1.3 with Hadoop 3.2.0 is not compatible with JDK17, we can't run Spark with JDK17 directly, but the following configuration may help if you use PySpark 3.3.3 or above
```
jerryshao: I would suggest we use a newer version of Spark for the example, 3.1.3 is too old.

yuqi1129: OK, let me change Spark 3.1.3 to 3.5.0; we have already tested it.

mchades (commented):
let me upadte aliyun doc in this PR

docs/fileset-catalog-with-adls.md
```diff
@@ -306,12 +306,12 @@ Or use the bundle jar with Hadoop environment if there is no Hadoop environment:
 
 ### Using Spark to access the fileset
 
-The following code snippet shows how to use **PySpark 3.1.3 with Hadoop environment(Hadoop 3.2.0)** and JDK8 to access the fileset:
+The following code snippet shows how to use **PySpark 3.5.0 with Hadoop environment(Hadoop 3.3.4)** and JDK8, JDK17 to access the fileset:
```
mchades: I would suggest introducing the usage with JDK 17 only, as it is the main version we want users to use. Including JDK 8 here would complicate things. WDYT?  @jerryshao

jerryshao: The thing is that why do we only pick JDK 8 and 17? If we want to list all the supported JDK versions, JDK 11 also should be listed. Otherwise, I would suggest giving an example without needing to mention the JDK version, and add a sentence mentioned about the difference of using JDK 17.

jerryshao (dismissed):

jerryshao:
@mchades, please take another look at this PR? Thanks.

mchades (approved):
lgtm.

Issue #8409, opened 2025-09-03
[Improvement] Fix delimiters in Properties.java


justinmclean:
### What would you like to be improved?

In clients/cli/src/main/java/org/apache/gravitino/cli/Properties.java correct the CLI Properties parser so that custom delimiters and key-value separators are treated as literal strings rather than regular expressions, ensuring that characters like `|` or `.` no longer break parsing.

Here's a unit tests to help:
```
  @Test
  public void testDelimiterWithRegexSpecialChar() {
    Properties properties = new Properties("|", "=");
    String[] input = {"key1=value1|key2=value2"};

    Map<String, String> result = properties.parse(input);

    assertEquals(2, result.size());
    assertEquals("value1", result.get("key1"));
    assertEquals("value2", result.get("key2"));
  }

  @Test
  public void testSeparatorWithRegexSpecialChar() {
    Properties properties = new Properties(",", "|");
    String[] input = {"key1|value1,key2|value2"};

    Map<String, String> result = properties.parse(input);

    assertEquals(2, result.size());
    assertEquals("value1", result.get("key1"));
    assertEquals("value2", result.get("key2"));
  }
```

Jackeyzhe:
Hi, @justinmclean I'd like to work on it

PR #8442, opened 2025-09-04
[#8409]Fix delimiters in Properties.java

Jackeyzhe:
Fix delimiters in Properties.java, make sure it split string correct with custom delimiter

Fix: #8409 

no

add unit tests

### DIFF
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/Properties.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/Properties.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.cli;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * A utility class to parse a delimited list of name-value pairs into a List of key-value entries.
@@ -45,8 +46,8 @@ public class Properties {
    * @param keyValueSeparator The separator used to distinguish keys from values in each pair.
    */
   public Properties(String delimiter, String keyValueSeparator) {
-    this.delimiter = delimiter;
-    this.keyValueSeparator = keyValueSeparator;
+    this.delimiter = Pattern.quote(delimiter);
+    this.keyValueSeparator = Pattern.quote(keyValueSeparator);
   }
 
   /**
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/Properties.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/Properties.java
@@ -19,6 +19,7 @@
 
 package org.apache.gravitino.cli;
 
+import com.google.common.base.Preconditions;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


clients/cli/src/main/java/org/apache/gravitino/cli/Properties.java
```diff
@@ -45,8 +46,8 @@ public Properties() {
    * @param keyValueSeparator The separator used to distinguish keys from values in each pair.
    */
   public Properties(String delimiter, String keyValueSeparator) {
-    this.delimiter = delimiter;
-    this.keyValueSeparator = keyValueSeparator;
+    this.delimiter = Pattern.quote(delimiter);
+    this.keyValueSeparator = Pattern.quote(keyValueSeparator);
   }
```
Abyss-lord: I think it would be better to add checks for null arguments and empty strings. Also, since we’re using Pattern, it makes more sense to define a Pattern argument directly instead of a raw string. For example, the code could look like this, WDYT
```java
public Properties(String delimiter, String keyValueSeparator) {
    Preconditions.checkArgument(
        delimiter != null && !delimiter.isEmpty(), "Delimiter cannot be null or empty");
    Preconditions.checkArgument(
        keyValueSeparator != null && !keyValueSeparator.isEmpty(),
        "Key-value separator cannot be null or empty");
    this.delimiterPattern = Pattern.compile(Pattern.quote(delimiter));
    this.keyValueSeparatorPattern = Pattern.compile(Pattern.quote(keyValueSeparator));
  }
```

Abyss-lord: meanwhile, I think `parse` method should refactor like follows:
```java
public Map<String, String> parse(String[] inputs) {
    HashMap<String, String> map = new HashMap<>();

    if (inputs != null) {
      for (String input : inputs) {
        if (input == null || input.isEmpty()) {
          continue;
        }

        // Split the input by the delimiter into key-value pairs
        String[] pairs = delimiterPattern.split(input);

        for (String pair : pairs) {
          if (pair.isEmpty()) {
            continue;
          }

          String[] keyValue = keyValueSeparatorPattern.split(pair, 2);
          if (keyValue.length == 2) {
            map.put(keyValue[0].trim(), keyValue[1].trim());
          }
        }
      }
    }
    return map;
  }
```

clients/cli/src/test/java/org/apache/gravitino/cli/PropertiesTest.java
```diff
@@ -181,4 +181,28 @@ public void testDuplicateKeysArray() {
...
+  public void testSeparatorWithRegexSpecialChar() {
+    Properties properties = new Properties(",", "|");
+    String[] input = {"key1|value1,key2|value2"};
+
+    Map<String, String> result = properties.parse(input);
+
+    assertEquals(2, result.size());
+    assertEquals("value1", result.get("key1"));
+    assertEquals("value2", result.get("key2"));
+  }
```
Abyss-lord, line 208: It would be better if we add some cases for null inputs.

Abyss-lord (commented):
just some small questions.

Jackeyzhe:
> @Jackeyzhe just some small questions.

thanks for your comment, I'll fix it

Issue #8339, opened 2025-08-28
[Improvement] In MetadataIdConverter.java GroupEntity is mapped to wrong type


justinmclean:
### What would you like to be improved?

In server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataIdConverter.java, Entity.EntityType.GROUP is mapped to Entity.class

### How should we improve?

It should be GroupEntity.class. Also update existing tests to include a GroupEntity test.

mxnzx:
Could you assign it to me ?

mxnzx:
Hi @justinmclean,
While fixing this issue, I noticed an incorrect description/message in `GroupEntity`.
(In core/src/main/java/org/apache/gravitino/meta/GroupEntity.java) 

It’s related to this area but not exactly the same scope. 
Do you prefer I add that minor correction to this PR, or would you rather I open a separate issue and PR for it?

justinmclean:
If minor, you can add it to your PR.

jerryshao:
@jerqi please take a look take this issue.

roryqi:
Yes, this is an issue. We should correct this.

mxnzx:
@jerqi Are there any other issues I should check?

roryqi:
> [@jerqi](https://github.com/jerqi) Are there any other issues I should check?

No other issues.

mxnzx:
Hi @justinmclean.

Is GROUP a top-level, metalake-scoped object (like CATALOG or ROLE)?
If so, should we allow names.size() == 1 for GROUP in the precondition below 
(in java/org/apache/gravitino/MetadataObjects.java)? 

If yes, I’ll add `|| type == MetadataObject.Type.GROUP` and update the tests.

```java
Preconditions.checkArgument(
    names.size() != 1
        || type == MetadataObject.Type.CATALOG
        || type == MetadataObject.Type.METALAKE
        || type == MetadataObject.Type.ROLE,
    "If the length of names is 1, it must be the CATALOG, METALAKE, or ROLE type");
```

PR #8443, opened 2025-09-04
[#8339] feat(authz): map GROUP to GroupEntity in MetadataIdConverter; add unit test

mxnzx:
- Map `Entity.EntityType.GROUP` → `GroupEntity.class` in `MetadataIdConverter`
- Add a unit test to cover the GROUP → GroupEntity code path in `getID()`

- GROUP was mapped to `Entity.class`, which is incorrect and may lead to runtime issues
  (e.g., invalid casting or resolving the wrong class in the entity store).
- The fix ensures the converter resolves GROUP to the correct entity class and prevents regressions.

Fix: #8339

- No user-facing/API changes.

- Added a unit test that verifies `getID()` returns the expected id for GROUP.
- Local runs:

```
./gradlew spotlessApply
./gradlew :server-common:test --tests 'org.apache.gravitino.server.authorization.TestMetadataIdConverter'
./gradlew test -PskipITs
```

### Notes / Follow-ups (for reviewers)
- This PR does **not yet update**:
  - `MetadataObjectUtil.checkMetadataObject()` to handle GROUP explicitly.
  - The Javadoc / comments in `MetadataObject` for the newly added GROUP type.

These will need to be clarified, and I’d like to confirm the expected behavior first.
(e.g., GROUP as a metalake-scoped object)
I’m opening this to get feedback on the current approach before finalizing those parts.


api/src/main/java/org/apache/gravitino/MetadataObject.java
```diff
@@ -61,7 +61,9 @@ enum Type {
     /** A role is an object contains specific securable objects with privileges */
     ROLE,
     /** A model is mapped to the model artifact in ML. */
-    MODEL
+    MODEL,
+    /** */
+    GROUP
```
roryqi: Group isn't a metadata object.

roryqi:
Group is an entity. But it's not the metadata object.
What's the difference between the entities and the metadata obejcts?
1. Metadata object is the object managed by the Gravitino. It's used for users.
2. Entity is the concept of the storage. It's internal concept.

Group isn't managed by Gravitino actually. It's the scope of user system.

mxnzx:
@jerqi 
Thanks for reviewing. I understand that GROUP is an auth entity (not a MetadataObject). I’ll update this PR to:
- revert the MetadataObject-related changes
- keep the GROUP → GroupEntity.class mapping fix
- add a unit test for the mapping

api/src/main/java/org/apache/gravitino/MetadataObject.java
roryqi: @mxnzx Please remove this line.

api/src/main/java/org/apache/gravitino/MetadataObjects.java
```diff
@@ -73,7 +73,8 @@ public static MetadataObject of(List<String> names, MetadataObject.Type type) {
         names.size() != 1
             || type == MetadataObject.Type.CATALOG
             || type == MetadataObject.Type.METALAKE
-            || type == MetadataObject.Type.ROLE,
+            || type == MetadataObject.Type.ROLE
+            || type == MetadataObject.Type.GROUP,
```
roryqi: Please remove this line.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -191,6 +193,8 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
           throw checkNotNull(exceptionToThrowSupplier).get();
         }
         break;
+      case GROUP:
```
roryqi: Please remove this line.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -111,6 +112,7 @@ public static NameIdentifier toEntityIdent(String metalakeName, MetadataObject m
       case FILESET:
       case COLUMN:
       case MODEL:
+      case GROUP:
```
roryqi: Please remove this line.

mxnzx: Hi all, I’m very sorry for the delay. I don’t have access to my laptop until the 18th, so I’m setting up a fresh environment and will try to push the updates by tomorrow morning. 
If anything blocks me, I will deliver the update on the 19th. Thank you for your patience, and apologies again for the delay.

mxnzx:
@jerqi  
Hi, sorry again for the delay.  

I have removed the GROUP-related `MetadataObject`.  
Since `TestConverter.java` only covers metadata object conversion, I'm not sure if a separate test for GROUP mapping is required. I did add a proper test, but I won’t be able to fully check this issue until the 18th.  

Could you please review the changes I’ve made so far?  
Once again, I apologize for the delay due to personal circumstances. Thank you for your understanding.

jerqi (approved):
LGTM.

mxnzx:
thanks for the review !

Issue #8371, opened 2025-09-01
[Improvement] Add cache for Lance storage


jerqi:
### What would you like to be improved?

 Add cache for Lance storage

### How should we improve?

 Add cache for Lance storage

PR #8450, opened 2025-09-05
[#8371] improvement(core): Add cache for Lance

roryqi:

Add cache for Lance

Fix: #8371 

Yes, I added the documents.

Add UT.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
+++ b/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
@@ -523,7 +523,7 @@ public class LancePartitionStatisticStorage implements PartitionStatisticStorage
                       });
 
               // Ensure dataset uses the latest version
-              if (newlyCreated.get()) {
+              if (!newlyCreated.get()) {
                 cachedDataset.checkoutLatest();
               }
 


core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -133,7 +153,45 @@ public LancePartitionStatisticStorage(Map<String, String> properties) {
...
+            properties.getOrDefault(INDEX_CACHE_SIZE, String.valueOf(DEFAULT_INDEX_CACHE_SIZE)));
+    Preconditions.checkArgument(
+        indexCacheSize > 0,
+        "Lance partition statistics storage indexCacheSizeBytes must be positive");
+
     this.properties = properties;
+
+    this.cache =
+        Caffeine.newBuilder()
+            .maximumSize(datasetCacheSize)
```
yuqi1129: Will all datasets be cached, and will there be no expiration time associated with them? I'm not very sure that whether 10000 file handler is a large number or not.

docs/manage-statistics-in-gravitino.md
```diff
@@ -245,13 +245,16 @@ For example, if you set an extra property `foo` to `bar` for Lance storage optio
...
-| `gravitino.stats.partition.storageOption.maxBytesPerFile` | The maximum bytes per file           | `104857600`                    | No                                              | 1.0.0         |
-| `gravitino.stats.partition.storageOption.maxRowsPerGroup` | The maximum rows per group           | `1000000`                      | No                                              | 1.0.0         |
-| `gravitino.stats.partition.storageOption.readBatchSize`   | The batch record number when reading | `10000`                        | No                                              | 1.0.0         |
+| Configuration item                                                   | Description                          | Default value                  | Required                                        | Since version |
+|----------------------------------------------------------------------|--------------------------------------|--------------------------------|-------------------------------------------------|---------------|
```
yuqi1129: Please reduce the number of `--` in column `Default value`

docs/manage-statistics-in-gravitino.md
```diff
@@ -245,13 +245,16 @@ For example, if you set an extra property `foo` to `bar` for Lance storage optio
...
+| `gravitino.stats.partition.storageOption.maxRowsPerFile`             | The maximum rows per file            | `1000000`                      | No                                              | 1.0.0         |
+| `gravitino.stats.partition.storageOption.maxBytesPerFile`            | The maximum bytes per file           | `104857600`                    | No                                              | 1.0.0         |
+| `gravitino.stats.partition.storageOption.maxRowsPerGroup`            | The maximum rows per group           | `1000000`                      | No                                              | 1.0.0         |
+| `gravitino.stats.partition.storageOption.readBatchSize`              | The batch record number when reading | `10000`                        | No                                              | 1.0.0         |
+| `gravitino.stats.partition.storageOption.datasetCacheSize`           | The dataset of Lance cache size      | `10000`                        | No                                              | 1.0.0         |
```
yuqi1129: The dataset of Lance cache size -> `size of dataset cache for Lance`

gradle/libs.versions.toml
```diff
@@ -29,7 +29,7 @@ guava = "32.1.3-jre"
 lombok = "1.18.20"
 slf4j = "2.0.9"
 log4j = "2.24.3"
-lance = "0.31.0"
+lance = "0.34.0"
```
yuqi1129, line 32: Was Lance involved so quickly?

roryqi: Yes, it is involved very quickly. I need the new features.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: The dataset will cache the metadata file and index file. If we have expiration time, we need the complex mechanism to trigger the cache to avoid the slow read.

docs/manage-statistics-in-gravitino.md
roryqi: No, I can't. It will break the style.

docs/manage-statistics-in-gravitino.md
roryqi: Fixed.

docs/manage-statistics-in-gravitino.md
yuqi1129: I mean 
image

roryqi: OK, I modify the `required` column.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
yuqi1129: When will it be closed? If the size is less than 10000, should it remain open as long as Gravitino Server is running?

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -197,123 +255,131 @@ public void updateStatistics(
...
+          String statisticName = statistic.getKey();
+          String statisticValue =
+              JsonUtils.anyFieldMapper().writeValueAsString(statistic.getValue());
+
+          UInt8Vector tableIdVector = (UInt8Vector) root.getVector(TABLE_ID_COLUMN);
+          VarCharVector partitionNameVector = (VarCharVector) root.getVector(PARTITION_NAME_COLUMN);
+          VarCharVector statisticNameVector = (VarCharVector) root.getVector(STATISTIC_NAME_COLUMN);
+          LargeVarCharVector statisticValueVector =
+              (LargeVarCharVector) root.getVector(STATISTIC_VALUE_COLUMN);
+          VarCharVector auditInfoVector = (VarCharVector) root.getVector(AUDIT_INFO_COLUMN);
```
yuqi1129: The lines can be moved out of the for loop according to the code below. Is it expected?

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
yuqi1129: @jerqi cc

roryqi: Yes, I think so.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: Fixed.

yuqi1129 (approved):
LGTM, @jerryshao Would you like to take a look?

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -133,7 +153,45 @@ public LancePartitionStatisticStorage(Map<String, String> properties) {
...
+                METADATA_FILE_CACHE_SIZE, String.valueOf(DEFAULT_METADATA_FILE_CACHE_SIZE)));
+    Preconditions.checkArgument(
+        metadataFileCacheSize > 0,
+        "Lance partition statistics storage metadataFileCacheSizeBytes must be positive");
+    this.indexCacheSize =
+        Long.parseLong(
+            properties.getOrDefault(INDEX_CACHE_SIZE, String.valueOf(DEFAULT_INDEX_CACHE_SIZE)));
+    Preconditions.checkArgument(
+        indexCacheSize > 0,
+        "Lance partition statistics storage indexCacheSizeBytes must be positive");
```
jerryshao, line 177: I would suggest to use "0" or "-1" as the default value for data, index, and metadata cache size, which means cache will be disabled by default. Cache is the advanced feature, we don't have to turn on by default.

roryqi: Fixed. Use 0 as the switch disable the cache.
IndexDatasize and metafileCacheSize is used for Lance internal parameter. We don't need to switch off.

jerryshao: What is the default value for index and metafile cache, we'd better align with Lance default value.

roryqi: The default value is 6G. I feel a little bigger for us.

jerryshao: If by default Lance doesn't turn on these two features, we should also align with Lance.

roryqi: image

roryqi: Lance will turn on this by default.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -79,13 +90,21 @@ public class LancePartitionStatisticStorage implements PartitionStatisticStorage
...
+  private static final String INDEX_CACHE_SIZE = "indexCacheSizeBytes";
+  private static final long DEFAULT_INDEX_CACHE_SIZE = 100L * 1024 * 1024; // 100MB
   // The schema is `table_id`, `partition_name`,  `statistic_name`, `statistic_value`, `audit_info`
   private static final String TABLE_ID_COLUMN = "table_id";
   private static final String PARTITION_NAME_COLUMN = "partition_name";
   private static final String STATISTIC_NAME_COLUMN = "statistic_name";
   private static final String STATISTIC_VALUE_COLUMN = "statistic_value";
   private static final String AUDIT_INFO_COLUMN = "audit_info";
 
+  private final Cache<Long, Dataset> cache;
```
jerryshao: Better renaming to `datasetCache`.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -133,7 +154,49 @@ public LancePartitionStatisticStorage(Map<String, String> properties) {
...
+              .evictionListener(
+                  (RemovalListener<Long, Dataset>)
+                      (key, value, cause) -> {
+                        if (value != null) {
+                          value.close();
+                        }
+                      })
+              .build();
+    } else {
+      cache = null;
```
jerryshao: Better to use `Optional` instead of `null` check.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -354,70 +447,100 @@ private static String getPartitionFilter(PartitionRange range) {
...
+              tableId,
+              id -> {
+                newlyCreated.set(true);
+                return open(getFilePath(id));
+              });
+
+      // ensure that the dataset is the latest version
+      if (!newlyCreated.get()) {
+        dataset.checkoutLatest();
+      }
```
jerryshao: This code has a threading issue, only one thread needs to check out the lastest snapshot.

gradle/libs.versions.toml
jerryshao, line 32: If this lance version introduce some different dependencies, we should check it.

roryqi: image
image
No different dependencies.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: Fixed.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: There are many threads to checkout, maybe we don't have a better solution.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
roryqi: OK, I replaced.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -197,95 +262,39 @@ public void updateStatistics(
...
+              .newTransactionBuilder()
+              .operation(Append.builder().fragments(fragmentMetas).build())
+              .transactionProperties(Collections.emptyMap())
+              .build();
+      newDataset = appendTxn.commit();
+
+      Dataset finalNewDataset = newDataset;
+      datasetCache.ifPresent(cache -> cache.put(tableId, finalNewDataset));
+    } finally {
+      if (!datasetCache.isPresent()) {
```
jerryshao, line 284: Can you please change to `isEmpty`? That looks more straightforward.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -354,70 +446,101 @@ private static String getPartitionFilter(PartitionRange range) {
...
+              Dataset cachedDataset =
+                  cache.get(
+                      tableId,
+                      id -> {
+                        newlyCreated.set(true);
+                        return open(getFilePath(id));
+                      });
+
+              // Ensure dataset uses the latest version
+              if (newlyCreated.get()) {
```
jerryshao: I think it should be `!newlyCreated.get()`, right?

roryqi: Yes, fixed.

Issue #8382, opened 2025-09-02
[Improvement] Fix NPE in IcebergHiveCachedClientPool.java


justinmclean:
### What would you like to be improved?

In iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/utils/IcebergHiveCachedClientPool.java, you will encounter an NPE if you close two pools.

Here's a test to show that:
```
  @Test
  void testCloseMultipleInstances() throws IOException {
    Configuration configuration = new Configuration();
    Map<String, String> properties = Maps.newHashMap();
    IcebergHiveCachedClientPool pool1 = new IcebergHiveCachedClientPool(configuration, properties);
    IcebergHiveCachedClientPool pool2 = new IcebergHiveCachedClientPool(configuration, properties);

    Assertions.assertDoesNotThrow(
        () -> {
          pool1.close();
          pool2.close();
        });
  }
```

pandeysambhi:
@justinmclean can I work on this ?

justinmclean:
Sure, if you have any questions, just ask.

pandeysambhi:
@justinmclean 
NPE was happening as scheduledExecutorService is initialized in the init() method, but only if clientPoolCache is null.
clientPoolCache is a static field, meaning it is shared across all instances of IcebergHiveCachedClientPool.
Only the first instance that calls init() will initialize both clientPoolCache and scheduledExecutorService.
Any subsequent instances will find clientPoolCache already initialized, so they will not initialize scheduledExecutorService (it remains null for those instances).When we create multiple instances of IcebergHiveCachedClientPool, only the first instance will have a non-null scheduledExecutorService.The other instances will have scheduledExecutorService as null.
When we call close() on all instances, each will attempt to call scheduledExecutorService.shutdownNow()
For instances where scheduledExecutorService is null, this results in a NullPointerException.

I am trying to get more in depth understanding of the project, from what I understood, we can either make scheduledExecutorService as static so that its shared or add a null check before closing it so that only the instance that actually initialized the executor tries to shut it down. I have opened a PR with null check. Just wanted to outline my understanding and reasoning here.

PR #8455, opened 2025-09-06
[#8382]Improvement(IcebergHiveCachedClientPool): prevent NPE in IcebergHiveCachedClientPool when closing multiple instances

pandeysambhi:
Add a null check before shutting down scheduledExecutorService in close method.

clientPoolCache is a static field, meaning it is shared across all instances of IcebergHiveCachedClientPool.
Only the first instance that calls init() will initialize both clientPoolCache and scheduledExecutorService.
Any subsequent instances will find clientPoolCache already initialized, so they will not initialize scheduledExecutorService (it remains null for those instances).When we create multiple instances of IcebergHiveCachedClientPool, only the first instance will have a non-null scheduledExecutorService.The other instances will have scheduledExecutorService as null.
When we call close() on all instances, each will attempt to call scheduledExecutorService.shutdownNow()
For instances where scheduledExecutorService is null, this results in a NullPointerException.

Fix: (#8382 )

No

unit testing


yuqi1129:
Good catch.

Issue #8381, opened 2025-09-02
[Improvement] In OceanBaseTableOperations.java in alterTable  empty SQL statments should not stop all SQL statements


justinmclean:
### What would you like to be improved?

Change alterTable so that empty SQL statements no longer stop processing subsequent changes.

### How should we improve?

Use continue instead of return. Modify it so that it doesn't throw a NPE.

pandeysambhi:
@justinmclean can I take this up?

justinmclean:
All yours. If you have any questions, just ask.

PR #8456, opened 2025-09-06
[#8381]Improvement(OceanbaseTableOperations):   empty SQL statements should not stop all SQL statements in alterTable

pandeysambhi:
Use of continue instead of return in alterTable method so that empty SQL statements no longer stop processing subsequent changes.

 empty SQL statements no longer stop processing subsequent changes in alterTable method 

Fix: #8381

No

Unit testing

### DIFF
--- a/catalogs/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/operation/OceanBaseTableOperations.java
+++ b/catalogs/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/operation/OceanBaseTableOperations.java
@@ -217,7 +217,7 @@ public class OceanBaseTableOperations extends JdbcTableOperations {
         String sql = generateAlterTableSql(databaseName, tableName, change);
         if (StringUtils.isEmpty(sql)) {
           LOG.info("No changes to alter table {} from database {}", tableName, databaseName);
-          return;
+          continue;
         }
         JdbcConnectorUtils.executeUpdate(connection, sql);
       }
--- a/catalogs/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/operation/TestOceanBaseTableOperations.java
+++ b/catalogs/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/operation/TestOceanBaseTableOperations.java
@@ -1157,16 +1157,18 @@ public class TestOceanBaseTableOperations extends TestOceanBase {
         Indexes.EMPTY_INDEXES);
 
     // Alter table with one valid change and one empty SQL change
-    TableChange validChange = TableChange.updateColumnComment(new String[] {"col_1"}, "Updated comment");
-    TableChange emptyChange = new TableChange() {
-        @Override
-        public String toString() {
+    TableChange validChange =
+        TableChange.updateColumnComment(new String[] {"col_1"}, "Updated comment");
+    TableChange emptyChange =
+        new TableChange() {
+          @Override
+          public String toString() {
             return "EmptyChange";
-        }
-    };
+          }
+        };
 
-    Assertions.assertDoesNotThrow(() ->
-        TABLE_OPERATIONS.alterTable(TEST_DB_NAME, tableName, emptyChange, validChange));
+    Assertions.assertDoesNotThrow(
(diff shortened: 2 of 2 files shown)


catalogs/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/operation/OceanBaseTableOperations.java
```diff
@@ -217,7 +217,7 @@ public void alterTable(String databaseName, String tableName, TableChange... cha
         String sql = generateAlterTableSql(databaseName, tableName, change);
         if (StringUtils.isEmpty(sql)) {
           LOG.info("No changes to alter table {} from database {}", tableName, databaseName);
-          return;
+          continue;
```
yuqi1129, line 220: I'm generally okay with the change. @mchades, please also check if we intentionally return directly if the SQL string is empty.

mchades: @pandeysambhi I'm not sure when `alterTableSql` will be empty. Could you please add tests to cover this scenario?

pandeysambhi: @mchades would you mind letting me know if tests are running in the oceanbase sub project for you. I keep getting the below error. I have tried couple of things to resolve it but nothing concrete so far.
Execution failed for task ':catalogs:catalog-jdbc-oceanbase:test'.
> No tests found for given includes: [org.apache.gravitino.catalog.oceanbase.operation.TestOceanBaseTableOperations](--tests filter)

justinmclean:
The test testAlterTableSkipsEmptySql is currently failing.

mchades (approved):
lgtm.

catalogs/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/operation/TestOceanBaseTableOperations.java
```diff
@@ -1133,4 +1133,37 @@ public boolean isMySQLDriverVersionSupported(String driverVersion) {
...
+        "test_table",
+        Collections.emptyMap(),
+        null,
+        Distributions.NONE,
+        Indexes.EMPTY_INDEXES);
+
+    // Alter table with one valid change and one empty SQL change
+    TableChange validChange =
+        TableChange.updateColumnComment(new String[] {"col_1"}, "Updated comment");
+    TableChange emptyChange = TableChange.deleteColumn(new String[] {"non_existent_col"}, true);
```
mchades, line 1162: which chang is the `empty SQL change`?

pandeysambhi: second one L1162. let me know if you would want me to explain in detail

mchades: why did you say the `deleteColumn` change is an empty SQL change? I saw it will skipe the empty generated SQL by 279L:
https://github.com/apache/gravitino/blob/f70a235227042f505cd6ec6982497b5e4c7961ca/catalogs/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/operation/OceanBaseTableOperations.java#L275-L281

pandeysambhi: @mchades From what I understood , L279 would check for empty deleteColSql and hence L280 would not be executed. then at L321 , following check would lead to an empty string return. as aletrSql is initialised with an empty arraylist.
  if (CollectionUtils.isEmpty(alterSql)) {
      return "";
    }

mchades: I see, thanks for your explanation

PR #8458, opened 2025-09-06
[#5199] feat(client-python): add distribution serdes

tsungchih:

JsonUtils.java

- DistributionSerializer
- DistributionDeserializer

We need to support table partitioning, bucketing and sort ordering and indexes.

#5199 

No

Unit tests

### DIFF
--- a/clients/client-python/tests/unittests/dto/rel/test_distribution_serdes.py
+++ b/clients/client-python/tests/unittests/dto/rel/test_distribution_serdes.py
@@ -1,0 +1,114 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import json
+import unittest
+from dataclasses import dataclass, field
+
+from dataclasses_json import DataClassJsonMixin, config
+
+from gravitino.api.expressions.distributions.strategy import Strategy
+from gravitino.dto.rel.distribution_dto import DistributionDTO
+from gravitino.dto.rel.expressions.json_serdes._helper.serdes_utils import SerdesUtils
+from gravitino.dto.rel.json_serdes.distribution_serdes import DistributionSerDes
+from gravitino.exceptions.base import IllegalArgumentException
+
+
+@dataclass
+class MockDataClass(DataClassJsonMixin):
+    distribution: DistributionDTO = field(
+        metadata=config(
+            encoder=DistributionSerDes.serialize,
+            decoder=DistributionSerDes.deserialize,
+        )
+    )
+
+
+class TestDistributionSerdes(unittest.TestCase):
+    def test_distribution_serdes_invalid_json(self):
+        json_strings = [
+            '{"distribution": {}}',
+            '{"distribution": ""}',
+        ]
+        for json_string in json_strings:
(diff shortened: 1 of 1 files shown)


tsungchih:

clients/client-python/gravitino/dto/rel/json_serdes/distribution_serdes.py
```diff
@@ -0,0 +1,55 @@
...
+from gravitino.dto.rel.expressions.json_serdes._helper.serdes_utils import SerdesUtils
+from gravitino.utils.precondition import Precondition
+from gravitino.utils.serdes import SerdesUtilsBase
+
+
+class DistributionSerDes(SerdesUtilsBase, JsonSerializable[DistributionDTO]):
+    """Custom JSON deserializer for DistributionDTO objects."""
+
+    @classmethod
+    def serialize(cls, data_type: DistributionDTO) -> dict[str, Any]:
```
jerryshao, line 33: Sorry, I forgot to leave the question in the index serdes PR. Shall we register the `serialize` and `deserialize` methods to the related DTO class, like `DistributionDTO` here? When will we call these two methods?

tsungchih: @jerryshao The signature of method `serialize` and `deserialize` conforms to the ones in Java client which are bound to `Index`. Apart from the `IndexSerializer` and `IndexDeserializer`, `PartitioningSerializer` and `PartitioningDeserializer` are similarly having signature bound to `Partitioning` instead of `PartitioningDTO`. I think there should have some reason at the time to derive the result.

The method `deserialize` could be called when [creating a table](https://gravitino.apache.org/docs/0.9.0-incubating/manage-relational-metadata-using-gravitino#create-a-table) while the method `serialize` should be called when listing tables, I think. The [commit of adding unit tests](https://github.com/apache/gravitino/pull/8458/commits/1311c3b705d3558ff992553e04258546a5b1aad3#diff-f0a7e88a43956c542630e051e39cab05e73e90b5877f4858184ef242359e71cbR31-R38) showed the corresponding usage of the two methods in the `MockDataClass`.

jerryshao: My question is that for "creating a table", we will receive a JSON string of "table" from the server, and then we will deserialize it into a "TableDTO", this "TableDTO" has fields like column, index, partitioning, etc, they will also be automatically deserialized when deserializing the "TableDTO", which means we have to register the "serialize" and "deserialize" methods to the related fields. Otherwise, how do the nested fields be deserialized automatically?

Refer to Java client, the deserialize and serialize methods are either registered by "annotation", or through Jackson mapper. I though we should have a similar way in Python?

tsungchih: @jerryshao Yes, definitely. With the "TableDTO" you mentioned, the equivalent role has shown in the unit tests, `MockDataClass`. Since we're using `dataclasses-json` for serdes, the serdes for nested fields are handled by the library if and only if the class is decorated by `@dataclass_json` or extends `DataClassJsonMixin`. Therefore, we may have the "TableDTO" be declared as follows.

```python
from dataclasses import dataclass, field

from dataclasses_json import DataClassJsonMixin, config

from gravitino.dto.rel.indexes.json_serdes.index_serdes import IndexSerdes
from gravitino.dto.rel.json_serdes.distribution_serdes import DistributionSerDes


@dataclass
class TableDTO(DataclassJsonMixin):
    name: str
    comment: str
    ...
    ...
    distribution: DistributionDTO = field(
        metadata=config(
            encoder=DistributionSerdes.serialize,
            decoder=DistributionSerdes.deserialize,
        )
    )
    indexes: list[IndexDTO] = field(
        metadata=config(
            encoder=lambda idxs: [IndexSerdes.serialize(idx) for idx in  idxs],
            decoder=lambda idxs: [IndexSerdes.deserialize(idx) for idx in idxs],
        )
    )
   ...
   ...
```

After the server has received the JSON string, say `json_string`, we should be able to deserialize the string by means of `table_dto = TableDTO.from_json(json_string)`. We can then serialize a `table_dto` using `json_string = table_dto.to_json()`.

Hope this explanation answers your question.

jerryshao: I see, thanks for the explanation.

jerryshao (approved):

PR #8472, opened 2025-09-08
[#233] Improvement(catalog): StarRocks table alter operation trigger IllegalPropertyOperation on unknown properties

playasim:

Add a new `IllegalPropertyException` which extends `GravitinoRuntimeException` to handle the unknown property issue in Starrocks.

Unknown properties in Starrocks can not be well handled. It throw an generic `GravitinoRuntimeException` when set up an unknown property to table. So raise this PR to add a new Custom Exception class to handle the case.

Fix: #(8418)

NO

run `./gradlew build` and passed all the UTs and ITs.
run `./gradlew :catalogs:catalog-jdbc-starrocks:test --tests "org.apache.gravitino.catalog.starrocks.operation.TestStarRocksTableOperations" -PskipDockerTests=false` and passed docker test in `StarRocksTableOperations`

### DIFF
--- a/catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/converter/StarRocksExceptionConverter.java
+++ b/catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/converter/StarRocksExceptionConverter.java
@@ -23,6 +23,7 @@ import java.sql.SQLException;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcExceptionConverter;
 import org.apache.gravitino.exceptions.ConnectionFailedException;
 import org.apache.gravitino.exceptions.GravitinoRuntimeException;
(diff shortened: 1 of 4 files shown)


catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/converter/StarRocksExceptionConverter.java
```diff
@@ -47,6 +48,7 @@ public class StarRocksExceptionConverter extends JdbcExceptionConverter {
   static final int CODE_NO_SUCH_COLUMN = 1054;
   static final int CODE_DELETE_NON_EXISTING_PARTITION = 1507;
   static final int CODE_PARTITION_ALREADY_EXISTS = 1517;
+  static final int CODE_GENERIC_ERROR = 5064;
```
playasim, line 51: error code: https://docs.starrocks.io/docs/3.3/sql-reference/Error_code/

Issue #8508, opened 2025-09-10
[Improvement] In GravitinoJdbcCatalog.java precvent possible NPE in createNamespace


justinmclean:
### What would you like to be improved?

In spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/jdbc/GravitinoJdbcCatalog.java if metadata is null a NPE will occur.

### How should we improve?

Check if metadata is not null

khawaja-abdullah:
Hey @justinmclean, may I take this up?

khawaja-abdullah:
Hi @justinmclean, raised a [PR](https://github.com/apache/gravitino/pull/8512) against this issue, please check when you get the chance. Thanks!

PR #8512, opened 2025-09-10
[#8508]

khawaja-abdullah:

Fix: #(issue)

### DIFF
--- a/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/jdbc/GravitinoJdbcCatalog.java
+++ b/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/jdbc/GravitinoJdbcCatalog.java
@@ -85,7 +85,7 @@ public class GravitinoJdbcCatalog extends BaseCatalog {
   public void createNamespace(String[] namespace, Map<String, String> metadata)
       throws NamespaceAlreadyExistsException {
     Map<String, String> properties = Maps.newHashMap();
-    if (!metadata.isEmpty()) {
+    if (metadata != null && !metadata.isEmpty()) {
       metadata.forEach(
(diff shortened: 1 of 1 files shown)


spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/jdbc/GravitinoJdbcCatalog.java
```diff
@@ -85,7 +85,7 @@ protected SparkTypeConverter getSparkTypeConverter() {
   public void createNamespace(String[] namespace, Map<String, String> metadata)
       throws NamespaceAlreadyExistsException {
     Map<String, String> properties = Maps.newHashMap();
-    if (!metadata.isEmpty()) {
+    if (metadata != null && !metadata.isEmpty()) {
```
Abyss-lord: we can simplify this using a utility method:
```java
if (MapUtils.isNotEmpty(metadata)) {
    ...
}
```
Benefits:

1. **Readability** – isNotEmpty makes the intent clearer at a glance.
2. **Safety** – it handles the null check internally, so we don’t need an explicit `!= null`.
3. **Consistency** – aligns with common practice of using utility classes rather than repeating the null/empty check pattern.

khawaja-abdullah: Thank you @Abyss-lord, I wasn't aware of this utility prior to making this change. I will update it right away.

khawaja-abdullah: Feedback was addressed. Thank you.

Issue #8509, opened 2025-09-10
[Improvement] Return a more specific ConnectionFailedException in PostgreSqlExceptionConverter.java


justinmclean:
### What would you like to be improved?

In catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java make PostgreSQL connection errors return a more specific ConnectionFailedException by checking for SQLState class “08”.

khawaja-abdullah:
Hey @justinmclean, may I take this up?

khawaja-abdullah:
Hi @justinmclean, raised a [PR](https://github.com/apache/gravitino/pull/8513) against this issue, please check when you get the chance. Thanks!

PR #8513, opened 2025-09-10
[#8509] improvement(catalog-postgresql): map SQLState class "08" to ConnectionFailedException

khawaja-abdullah:
Update `PostgreSqlExceptionConverter` so that SQLState codes in class "08" (connection errors) are mapped to a more specific `ConnectionFailedException` instead of the generic `GravitinoRuntimeException`.

Previously, PostgreSQL connection-related errors (SQLState class "08") were reported as generic runtime exceptions, making it harder for clients and higher layers to distinguish connection failures from other SQL errors.  
This change improves error handling by returning a more meaningful exception, allowing callers to react appropriately (e.g., retrying connections).

Fix: #8509 

Yes.  
Users interacting with Gravitino via the PostgreSQL catalog will now receive a `ConnectionFailedException` for SQLState class "08" errors instead of a generic `GravitinoRuntimeException`.

N/A

### DIFF
--- a/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
+++ b/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
@@ -35,7 +35,13 @@ public class PostgreSqlExceptionConverter extends JdbcExceptionConverter {
   private static final String INVALID_SCHEMA_NAME = "3D000";
   private static final String INVALID_SCHEMA = "3F000";
   private static final String UNDEFINED_TABLE = "42P01";
-  private static final String CONNECTION_EXCEPTION = "08"; // class code
+
+  /**
+   * SQLSTATE '08' is the class code of connection exceptions See <a
+   * href="https://www.postgresql.org/docs/current/errcodes-appendix.html#ERRCODES-TABLE">PostgreSQL
+   * errcodes appendix</a>.
+   */
+  private static final String CONNECTION_EXCEPTION = "08";
 
   @SuppressWarnings("FormatStringAnnotation")
   @Override
--- a/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
+++ b/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
@@ -44,12 +44,13 @@ public class PostgreSqlExceptionConverter extends JdbcExceptionConverter {
           return new NoSuchSchemaException(se.getMessage(), se);
         case "42P01":
           return new NoSuchTableException(se.getMessage(), se);
-        default: {
-          if (se.getSQLState().startsWith("08")) {
-            return new ConnectionFailedException(se.getMessage(), se);
+        default:
+          {
+            if (se.getSQLState().startsWith("08")) {
+              return new ConnectionFailedException(se.getMessage(), se);
+            }
+            return new GravitinoRuntimeException(se.getMessage(), se);
           }
-          return new GravitinoRuntimeException(se.getMessage(), se);
-        }
       }
     } else {
       if (se.getMessage() != null && se.getMessage().contains("password authentication failed")) {
--- a/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
+++ b/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
@@ -44,8 +44,12 @@ public class PostgreSqlExceptionConverter extends JdbcExceptionConverter {
           return new NoSuchSchemaException(se.getMessage(), se);
         case "42P01":
           return new NoSuchTableException(se.getMessage(), se);
-        default:
+        default: {
+          if (se.getSQLState().startsWith("08")) {
+            return new ConnectionFailedException(se.getMessage(), se);
+          }
           return new GravitinoRuntimeException(se.getMessage(), se);
+        }
       }
     } else {
       if (se.getMessage() != null && se.getMessage().contains("password authentication failed")) {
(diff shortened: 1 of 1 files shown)


catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
```diff
@@ -44,8 +44,12 @@ public GravitinoRuntimeException toGravitinoException(SQLException se) {
           return new NoSuchSchemaException(se.getMessage(), se);
         case "42P01":
           return new NoSuchTableException(se.getMessage(), se);
-        default:
+        default: {
+          if (se.getSQLState().startsWith("08")) {
+            return new ConnectionFailedException(se.getMessage(), se);
+          }
           return new GravitinoRuntimeException(se.getMessage(), se);
+        }
```
Abyss-lord: It would be better not to hardcode SQLState strings like "42P04" or "42P07" directly in the switch block. I suggest extracting these codes into a dedicated constants class (e.g. PostgresErrorCodes) or an enum or a static variable. That way, we can replace:
```java
case "42P07":
    return new TableAlreadyExistsException(se.getMessage(), se);
```
with something like:
```java
case PostgresErrorCodes.DUPLICATE_TABLE:
    return new TableAlreadyExistsException(se.getMessage(), se);
```
Meanwhile, you can change other variables. e.g.
```java
private static final String DUPLICATE_DATABASE = "42P04";
private blic static final String DUPLICATE_SCHEMA = "42P06";
private static final String DUPLICATE_TABLE = "42P07";
private static final String INVALID_SCHEMA_NAME = "3D000";
private static final String INVALID_SCHEMA = "3F000";
private static final String UNDEFINED_TABLE = "42P01";
```

Abyss-lord:
@khawaja-abdullah plz use spotlessApply format the code.

khawaja-abdullah:
> @khawaja-abdullah plz use spotlessApply format the code.

Sorry for this oversight @Abyss-lord - I did apply it but looks like I did it on an unrelated module - fixed now.

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
khawaja-abdullah: That's a great suggestion, let me get back to you with this change shortly.

khawaja-abdullah: Updated as per suggestion. Thank you.

jerryshao: The purpose of `ConnectionFailedException` is for catalog connection test. Here, I think RuntimeException and IOException are more suitable.

khawaja-abdullah: Thanks for your input @jerryshao - @justinmclean could you chime in here please?

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
```diff
@@ -29,23 +29,36 @@
 
 public class PostgreSqlExceptionConverter extends JdbcExceptionConverter {
 
+  private static final String DUPLICATE_DATABASE = "42P04";
+  private static final String DUPLICATE_SCHEMA = "42P06";
+  private static final String DUPLICATE_TABLE = "42P07";
+  private static final String INVALID_SCHEMA_NAME = "3D000";
+  private static final String INVALID_SCHEMA = "3F000";
+  private static final String UNDEFINED_TABLE = "42P01";
+  private static final String CONNECTION_EXCEPTION = "08"; // class code
```
yuqi1129: What is `class code`?

khawaja-abdullah: Hey @yuqi1129, good that you called it out- in SQLSTATE terms each error code is a five-character string:
- Class code = first 2 characters of SQLSTATE (error category).
- Subclass code = next 3 characters (exact error).

Let me know if you find it misleading, I will remove.

khawaja-abdullah: [Ref](https://www.postgresql.org/docs/current/errcodes-appendix.html#:~:text=Class%2008%20%E2%80%94%20Connection%20Exception)

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
justinmclean: The 08 class of errors are connection exceptions, so ConnectionFailedException seems to be the correct one to me. Please see https://www.postgresql.org/docs/current/errcodes-appendix.html:

Class 08 — Connection Exception
08000 | connection_exception
08003 | connection_does_not_exist
08006 | connection_failure
08001 | sqlclient_unable_to_establish_sqlconnection
08004 | sqlserver_rejected_establishment_of_sqlconnection
08007 | transaction_resolution_unknown
08P01 | protocol_violation

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/converter/PostgreSqlExceptionConverter.java
yuqi1129: I see, you can refine it as `'08' is the class code of connection exceptions. By the way, adding a link is always preferred.

khawaja-abdullah: Thanks for your input @yuqi1129, I have added a documentation comment with a link to source.

justinmclean (approved):
Thanks for improving this

Issue #8511, opened 2025-09-10
[Improvement] Missing space between the credential type and credential context in CredentialCacheKey.java


justinmclean:
### What would you like to be improved?

Credential type and credential context should have a space between them. Here's a unit test showing the issue:
```
  @Test
  void testToStringContainsSpaceBetweenFields() {
    PathBasedCredentialContext context =
        new PathBasedCredentialContext("user", ImmutableSet.of(), ImmutableSet.of());
    CredentialCacheKey key = new CredentialCacheKey("s3-token", context);

    Assertions.assertTrue(
        key.toString().contains("credentialType: s3-token credentialContext:"));
  }
```

khawaja-abdullah:
Hey @justinmclean, may I take this up?

khawaja-abdullah:
Hi @justinmclean, raised a [PR](https://github.com/apache/gravitino/pull/8515) against this issue, please check when you get the chance. Thanks!

PR #8515, opened 2025-09-10
[#8511] improvement(core): add missing space in CredentialCacheKey.toString()

khawaja-abdullah:
Added a missing space between the `credentialType` and `credentialContext` fields in the `toString()` output of `CredentialCacheKey`, improving its readability.

The missing space made the string representation less readable and could confuse debugging or log parsing. This fix ensures clear separation between fields and improves developer experience while interpreting the output.

Fix: #8511 

Yes—minor formatting improvement seen in log outputs, error messages, or any usage of `toString()` on `CredentialCacheKey`. No behavioral changes.

- Added unit test to verify the spaced output in `toString()` includes `"credentialType: <type> credentialContext:"`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/credential/CredentialCacheKey.java
+++ b/core/src/main/java/org/apache/gravitino/credential/CredentialCacheKey.java
@@ -57,7 +57,7 @@ public class CredentialCacheKey {
     stringBuilder
         .append("credentialType: ")
         .append(credentialType)
-        .append("credentialContext: ")
+        .append(" credentialContext: ")
         .append(credentialContext);
     return stringBuilder.toString();
   }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/credential/CredentialCacheKey.java
```diff
@@ -57,7 +57,7 @@ public String toString() {
     stringBuilder
         .append("credentialType: ")
         .append(credentialType)
-        .append("credentialContext: ")
+        .append(" credentialContext: ")
```
FANNG1, line 60: could you use `,` to replace ` `?

khawaja-abdullah: Hey, thanks for your input @FANNG1 - a missing space was explicitly requested to be added in #8511. If @justinmclean also agrees to use comma (`,`) instead of space(` `), I will update right away.

justinmclean: Space is fine by me, but it doesn't really matter.

Issue #8507, opened 2025-09-10
[Improvement] In ListTopicFailureEvent.java add precondition to check if namespace is not null


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java.

Here's a unit test showing that:
```
  @Test
  public void testNamespaceMustNotBeNull() {
    Assertions.assertThrows(
        NullPointerException.class,
        () -> new ListTopicFailureEvent("user", null, new Exception("boom")));
  }
}
```

### How should we improve?

See above

midhunweb:
Hi, I’d like to work on this issue. I’ll add a precondition in ListTopicFailureEvent to ensure namespace is not null, and update the unit test to assert on the exception message.

justinmclean:
Thanks for offering to help, all yours.

HyungGeun94:
Hi @justinmclean, 

I’d like to contribute by adding a null check in `ListTopicFailureEvent` and updating the unit tests to assert the exception message, similar to the original PR #8507. 

I noticed the previous PR was closed and the branch deleted. I have forked the repo and prepared a new branch for this improvement, following the project’s coding and formatting guidelines. 

Could I proceed with a new PR for this issue?

PR #8516, opened 2025-09-10
[#8507] Improvement(core): In ListTopicFailureEvent.java add precondition to check if namespace is not null

midhunweb:
[Improvement(core)] In ListTopicFailureEvent.java add precondition to check if namespace is not null

Added a null check in the ListTopicFailureEvent constructor using Objects.requireNonNull to ensure that the namespace is not null.  
Added a unit test `testListTopicFailureEventWithNullNamespace()` in TestTopicEvent to verify that a NullPointerException is thrown when namespace is null.

To prevent potential NullPointerExceptions when creating a ListTopicFailureEvent with a null namespace and improve code robustness.

Fix: #8507

N/A

N/A

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
+++ b/core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
@@ -19,6 +19,8 @@
 
 package org.apache.gravitino.listener.api.event;
 
+import static java.util.Objects.requireNonNull;
+
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.Namespace;
 import org.apache.gravitino.annotation.DeveloperApi;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
```diff
@@ -39,7 +41,8 @@ public final class ListTopicFailureEvent extends TopicFailureEvent {
    * @param exception The exception encountered during the attempt to list topics.
    */
   public ListTopicFailureEvent(String user, Namespace namespace, Exception exception) {
```
Abyss-lord, line 43: I suggest adding a new static factory method to NameIdentifier, for example:
```java
 public static NameIdentifier of(Namespace namespace) {
   check(namespace != null, "Cannot create a NameIdentifier with null namespace");
   return of(namespace.levels());
 }
```
This way, `ListTopicFailureEvent` can remain unchanged. WDYT

Abyss-lord:
@midhunweb you should run `./gradlew :core:spotlessApply` and `./gradlew :api:spotlessApply` to format the codes.

FANNG1:
@midhunweb , thanks for your PR, The namespace couldn't be null in list topic operations https://github.com/apache/gravitino/blob/1039bc399e4b325501d14b3fc9255af902a041fe/server/src/main/java/org/apache/gravitino/server/web/rest/TopicOperations.java#L90. It seems no requirement to add this nullable check.

justinmclean:
You need to format the code correctly by running `./gradlew :core:spotlessApply` and push the changes.

Issue #8445, opened 2025-09-05
[FEATURE] Cache and reuse catalog classloader


yuqi1129:
### Describe the feature

Currently, modifying a catalog like alter, creating, will create a new catalog and then close it after the delete options have been used. As we all know that frequently creating/dropping classloaders is very expensive and memory-consumpation; what's more, it's easy lead to memory leaks. We need to introduce a new mechanism to resolve this.

yuqi1129:
https://docs.google.com/document/d/16dQbOsscPBgzxwSsfH6yfP0A7o_CqvTfp4QCG9GpVTM/edit?tab=t.0

According to the survey, it is not feasible to share class loaders between catalogs; to be more precise, between catalogs that contain the Hadoop file system, it's okay to share class loaders for JDBC and message catalogs. However, most of the memory used is Hadoop-related catalogs, so it's not necessary to use it. The only option we have is to avoid creating a new class loader each time the catalog is modified.

yuqi1129:
Close it temporarily.

PR #8536, opened 2025-09-11
[#8445] improvement(core): Share and cache class loader instance for catalog

yuqi1129:

1. Add a reference count to show how many catalogs a class loader is shared with.
2. Do not immediately close a class loader when it is not used by any catalog.

For better performance.

Fix: #8445

N/A.

Existing tests.


yuqi1129:
Close it temporarily as it's imperfect.

Issue #8313, opened 2025-08-27
[Subtask] Standardize credentials in loadTable/loadView responses for Iceberg REST server


FANNG1:
### Describe the subtask

Standardize credentials in loadTable/loadView responses for Iceberg REST server

### Parent issue

#6336

sunxiaojian:
@FANNG1 You can assign it to me.

FANNG1:
> [@FANNG1](https://github.com/FANNG1) You can assign it to me.

done

sunxiaojian:
No need to handle, closed.

PR #8539, opened 2025-09-12
[#8313] Improvement (iceberg-rest): Standardize credentials in loadTable/loadView responses for Iceberg REST server

sunxiaojian:

Standardize credentials in loadTable/loadView responses for Iceberg REST server

Fix: #([8313](https://github.com/apache/gravitino/issues/8313))

N/A

N/A

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
@@ -45,7 +45,10 @@ import org.apache.iceberg.catalog.Namespace;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.ServiceUnavailableException;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
+import org.apache.iceberg.rest.responses.ImmutableLoadViewResponse;
 import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.apache.iceberg.rest.responses.LoadViewResponse;
+import org.apache.iceberg.view.ViewMetadata;
 
 /** Process Iceberg REST specific operations, like credential vending. */
 public class CatalogWrapperForREST extends IcebergCatalogWrapper {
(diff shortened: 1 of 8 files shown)


sunxiaojian:
@FANNG1 PTAL,thx.

FANNG1:
> @FANNG1 PTAL,thx.

Thanks for the PR, will review it recently.

FANNG1:
I conducted some tests and found that Spark could create and read the view in S3 storage without this PR. The main reason is that `ViewMetadata` contains the full metadata of the view, which seems to make it unnecessary to obtain additional credentials to access the storage for the view,  WDYT? @sunxiaojian

sunxiaojian:
> I conducted some tests and found that Spark could create and read the view in S3 storage without this PR. The main reason is that `ViewMetadata` contains the full metadata of the view, which seems to make it unnecessary to obtain additional credentials to access the storage for the view, WDYT? @sunxiaojian

@FANNG1 I checked the iceberg source code, and it seems indeed unnecessary; only when FileIO is used would credential information be required.

reference:
 [loadView](https://github.com/apache/iceberg/blob/ee90c10e39cec0ccceb9425e03a3e0b5690daf3b/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L1131 )
  [loadTable](https://github.com/apache/iceberg/blob/ee90c10e39cec0ccceb9425e03a3e0b5690daf3b/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L452)

FANNG1:
> > I conducted some tests and found that Spark could create and read the view in S3 storage without this PR. The main reason is that `ViewMetadata` contains the full metadata of the view, which seems to make it unnecessary to obtain additional credentials to access the storage for the view, WDYT? @sunxiaojian
> 
> @FANNG1 I checked the iceberg source code, and it seems indeed unnecessary; only when FileIO is used would credential information be required.
> 
> reference: [loadView](https://github.com/apache/iceberg/blob/ee90c10e39cec0ccceb9425e03a3e0b5690daf3b/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L1131) [loadTable](https://github.com/apache/iceberg/blob/ee90c10e39cec0ccceb9425e03a3e0b5690daf3b/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L452)

Sorry for wasting your time :(

sunxiaojian:
> Sorry for wasting your time :(

Not a waste at all! Thanks for verifying.   Let me close it.

PR #8542, opened 2025-09-13
[#8589] feat(client-python):  add table change to support alter table

tsungchih:

TableChange.java

- TableChange

In this PR, the approach to implementing the above class conforms to the following classes for consistent coding style.

- MetalakeChange (metalake_change.py)
- SchemaChange (schema_change.py)
- FilesetChange (fileset_change.py)

We need to support python client for table operations.

#8589 

No

Unit tests


tsungchih:

tsungchih:

unknowntpo:
LGTM

jerryshao:
Hi @tsungchih can you please a subtask issue to track the work here, do not use the epic issue.

clients/client-python/gravitino/api/rel/table_change.py
```diff
@@ -0,0 +1,942 @@
...
+        def get_new_name(self) -> str:
+            """Retrieves the new name for the table.
+
+            Returns:
+                str: The new name of the table.
+            """
+            return self._new_name
+
+        def __str__(self):
+            return f"RENAMETABLE {self._new_name}"
```
jerryshao, line 328: I think we should add `__eq__` and `__hash__` methods, what do you think?

clients/client-python/gravitino/api/rel/table_change.py
```diff
@@ -0,0 +1,942 @@
...
+
+from dataclasses_json import config
+
+from gravitino.api.column import Column
+from gravitino.api.expressions.expression import Expression
+from gravitino.api.expressions.indexes.index import Index
+from gravitino.api.types.type import Type
+
+
+class TableChange(ABC):
```
jerryshao, line 30: I found that we put this class under `api` package, I think it is better aligned with Java's implementation, what do you think? Also, some other classes are already in the `api` package, do we have a compatible solution to handle this?

tsungchih:
> Hi @tsungchih can you please a subtask issue to track the work here, do not use the epic issue.

@jerryshao I've tried to create a subtask issue as you mentioned. Hope I'm doing it right as expected.

clients/client-python/gravitino/api/rel/table_change.py
tsungchih: @jerryshao Thanks for the comments. The `dataclass` will automatically generate an `__eq__` method to compare the class as if it were a tuple of its fields, in order [ref](https://docs.python.org/3.9/library/dataclasses.html#module-level-decorators-classes-and-functions). Furthermore, `dataclass` will generate a `__hash__()` method since `eq` is `True` by default and `frozen` is explicitly set to `True`.

Here's a simple example to show the feature.

```python
>>> from gravitino.api.table_change import TableChange
>>> rename1 = TableChange.RenameTable("new_name1")
>>> rename2 = TableChange.RenameTable("new_name1")
>>> rename3 = TableChange.RenameTable("new_name2")
>>> rename1==rename2
True
>>> rename1==rename3
False
>>> hash(rename1)==hash(rename2)
True
>>> hash(rename1)==hash(rename3)
False
```

And some classes in the module `table_change` implemented `__hash__` method because they have data member of `list` type which is unhashable. We can, of course, explicitly define customized `__eq__` and `__hash__` methods. Do you prefer to explicitly define both of them rather than use the generated ones by `dataclass`?

Thanks again for your comments. I just happened to find some issues in the customized hash logics. I'll fix it in the next few days.

clients/client-python/gravitino/api/rel/table_change.py
tsungchih: @jerryshao Oh, perhaps we should put this module under `clients/client-python/gravitino/api/rel/table_change.py`. Is that what you expected to align with Java's implementation?

clients/client-python/gravitino/api/rel/table_change.py
jerryshao: I would suggest defining them explicitly. Otherwise, we may rely on the dataclass behavior, and it may be changed.

clients/client-python/gravitino/api/rel/table_change.py
jerryshao: Yeah, that's what I thought. But we already have some classes like column in the api package, we'd also move them to the rel package in a compatible way.

jerryshao:
> > Hi @tsungchih can you please a subtask issue to track the work here, do not use the epic issue.
> 
> @jerryshao I've tried to create a subtask issue as you mentioned. Hope I'm doing it right as expected.

Yes, it is suggested that each PR has a related issue to track, and that issue focuses only on one thing. And also update the PR title and PR description to associate to that issue.

clients/client-python/gravitino/api/rel/table_change.py
tsungchih: Sure, understand. Let me make the changes to explicitly define them.

clients/client-python/gravitino/api/rel/table_change.py
tsungchih: Sure, that makes sense. Would you mind if I moving the class Column in this PR when it's possible.

jerryshao: OK, sure.

tsungchih: @jerryshao I've moved the class Column into `gravitino.api.rel`. Apart from Column, I also moved package `indexes`, `partitions`, and `types` into `gravitino.api.rel` as well. It's a big change to move `expressions` compared to the aforementioned moves. I therefore decided to make the refactor in the next PR. Please take a look into this when you're available. Thanks!

clients/client-python/gravitino/api/rel/table_change.py
tsungchih: @jerryshao I've done to explicitly define the `__eq__` and `__hash__` methods. Please take a look into it when you're available. Thanks!

clients/client-python/gravitino/api/rel/table_change.py
jerryshao: OK, we can do this in a separate PR.

jerryshao (approved):

Issue #5198, opened 2024-10-21
[EPIC] Support python client for table operations


unknowntpo:
### Describe the proposal

Support relational catalog operations for Python client.

### Task list

Support relational catalog operations for Python client
- [ ] Support Column data type
  - [docs](https://gravitino.apache.org/docs/0.6.1-incubating/manage-relational-metadata-using-gravitino#apache-gravitino-table-column-type)
 - related java issue: [#599](https://github.com/apache/gravitino/issues/599) #3427
- [ ] Support  Expression system
  - [docs](https://gravitino.apache.org/docs/0.6.1-incubating/expression)
  - related java issue: #600
- [ ] Support  Table partitioning, bucketing and sort ordering and indexes
  - [docs](https://gravitino.apache.org/docs/0.6.1-incubating/table-partitioning-bucketing-sort-order-indexes)
  - related java issues: #248 #264
- [ ] Support  Column default value
 - [docs](https://gravitino.apache.org/docs/0.6.1-incubating/manage-relational-metadata-using-gravitino#table-column-default-value)
   - related java issues: #896
- [ ] Support  Table partition management
  - [docs](https://gravitino.apache.org/docs/0.6.1-incubating/manage-table-partition-using-gravitino)
  - related issues: #1575

tsungchih:
@unknowntpo Just a heads-up. I started working on support for alter table

PR #8545, opened 2025-09-14
[#5198] feat(client-python): add policy

tsungchih:

- Policy.java
- PolicyContent.java
- PolicyContents.java

We need to support python client for table operations.

#5198 

No

Unit tests

### DIFF
--- a/clients/client-python/gravitino/api/policy/policy.py
+++ b/clients/client-python/gravitino/api/policy/policy.py
@@ -1,0 +1,193 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from abc import ABC, abstractmethod
+from enum import Enum
+
+from gravitino.api.auditable import Auditable
+from gravitino.api.metadata_object import MetadataObject
+from gravitino.api.policy.policy_content import PolicyContent
+from gravitino.api.policy.policy_contents import PolicyContents
+from gravitino.exceptions.base import (
+    IllegalArgumentException,
+    UnsupportedOperationException,
+)
+from gravitino.utils.precondition import Precondition
+
+
+class Policy(Auditable):
+    """The interface of the policy.
+
+    The policy is a set of rules that can be associated with a metadata object.
+    The policy can be used for data governance and so on.
+    """
+
+    BUILT_IN_TYPE_PREFIX: str = "system_"
+    """The prefix for built-in policy types. All built-in policy types should start with this prefix."""
+
+    @abstractmethod
+    def name(self) -> str:
+        """Get the name of the policy.
+
+        Returns:
+            The name of the policy.
+        """
+
+    @abstractmethod
+    def policy_type(self) -> str:
+        """Get the type of the policy.
+
+        Returns:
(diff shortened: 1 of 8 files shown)


tsungchih:

clients/client-python/gravitino/api/policy/policy.py
```diff
@@ -0,0 +1,193 @@
...
+    @abstractmethod
+    def content(self) -> PolicyContent:
+        """Get the content of the policy.
+
+        Returns:
+            PolicyContent: The content of the policy.
+        """
+
+    @abstractmethod
+    def inherited(self) -> bool:
```
unknowntpo: in `Policy.java`, it's `Optional<Boolean> inherited();`. Do we need to align this behavior ?

FANNG1:
@mchades PTAL, thx

clients/client-python/gravitino/api/policy/policy.py
jerryshao: Yes, I think so. In the Java implementation, it will have three states: true, false and null, which represent different states. We'd also support this in Python.

jerryshao:
The policy API depends on the implementation of the relational table API, I think we can support the table operations firstly.

clients/client-python/gravitino/api/policy/policy.py
tsungchih: Thanks for the comment. The signature of the method has been changed accordingly.

tsungchih: @jerryshao @unknowntpo May I have your help with taking a look into this PR when you're available? Many thanks!

unknowntpo: LGTM.

tsungchih:
> The policy API depends on the implementation of the relational table API, I think we can support the table operations firstly.

Hi @jerryshao , sorry for having this comment missed and for the late reply. To support the table operations (I mean the EPIC #5198), implementing `RelationalTable` is one of its subtask. And to complete the implementation, we need to implement `SupportsPolicies` since `RelationalTable` implements the interface. And therefore `Policy`, `PolicyContents`, and `PolicyContent`. I have no idea what you mean as to "support the table operations firstly". I thought I'm on the path to supporting the table operations. I would appreciate it if you could elaborate more on your comment. Thanks!

mchades:
> implementing RelationalTable is one of its subtask. And to complete the implementation, we need to implement SupportsPolicies since RelationalTable implements the interface.

It's unnecessary.  We can first implement the other interfaces of `RelationalTable` to ensure its core functionality works end-to-end, and leave the implementation of the `SupportsPolicies` interface for later, as it represents an auxiliary capability rather than a fundamental one.

tsungchih:
> > implementing RelationalTable is one of its subtask. And to complete the implementation, we need to implement SupportsPolicies since RelationalTable implements the interface.
> 
> It's unnecessary. We can first implement the other interfaces of `RelationalTable` to ensure its core functionality works end-to-end, and leave the implementation of the `SupportsPolicies` interface for later, as it represents an auxiliary capability rather than a fundamental one.

@mchades Thanks for the prompt reply. Here's the ordering for the subsequent implementation according to my understanding.

- `TableDTO` and `Table`
- `DTOConverters.fromDTO()`
- `RelationalTable`
- `PartitionNameListResponse`
- `PartitionListResponse`
- `PartitionResponse`
- `AddPartitionsRequest`
- `DropResponse`
- all `SupportsXXX` and their related classes

Could you please take a look to see if my understanding is correct? Thanks!

mchades:
> Could you please take a look to see if my understanding is correct?

correct

mchades:
BTW, I would suggest adding a table-relative integration test after `RelationalTable` and before `PartitionNameListResponse`.

tsungchih:
> BTW, I would suggest adding a table-relative integration test after `RelationalTable` and before `PartitionNameListResponse`.

Thanks for your suggestion and for your kind guidance.

jerryshao:
I think we should also add relational catalog support, so we can operate the table from end to end, to make this feature complete.

tsungchih:
@jerryshao @unknowntpo I'm going to resume implementation aimed at relational table. Could you please take a review on this PR when you are available? Thanks!

jerryshao:
@tsungchih As I mentioned in another PR. The work of adding policy is not a prerequisite for a relational catalog. We still have several functions that are not supported in the relational catalog. We'd better finish that part first before moving on to the new territory.

Issue #8483, opened 2025-09-09
[Improvement] Ensure ClientPoolImpl.java run method doesn't leak unusable connections


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/utils/ClientPoolImpl.java it is possible for the run method to leak unusable connections

### How should we improve?

Make sure it properly handles reconnect failures by closing the old client, adjusting the pool size, and skipping release to prevent leaking unusable connections.

keepConcentration:
@justinmclean 

Hi, I'd like to work on this issue. Could you assign it to me?

PR #8547, opened 2025-09-15
[#8483] improvement(core): prevent leaking unusable connections in ClientPoolImpl run method

keepConcentration:

This PR updates the `ClientPoolImpl#run` method to ensure unusable connections are not leaked when a reconnection attempt fails.

Previously, if a reconnection attempt failed inside `run`, the old (unusable) client could still be released back into the pool, leading to potential connection leaks and invalid state.

Fix: #8483

No user-facing changes.

Executed existing unit tests


core/src/main/java/org/apache/gravitino/utils/ClientPoolImpl.java
```diff
@@ -55,14 +55,24 @@ public <R> R run(Action<R, C, E> action) throws E, InterruptedException {
...
     try {
       return action.run(client);
 
     } catch (Exception exc) {
       if (retry && isConnectionException(exc)) {
         try {
           client = reconnect(client);
-        } catch (Exception ignored) {
+        } catch (Exception reconnectException) {
+          shouldRelease = false;
```
yuqi1129, line 67: Since the client will not be null if the code goes here(See L57), why do we need this parameter? What is it used for?

keepConcentration: Thank you for the review. @yuqi1129 

The `shouldRelease` flag is used in the finally block to control whether a client should be returned to the pool when it becomes unusable for reasons other than a connection exception.

yuqi1129: I see.

core/src/main/java/org/apache/gravitino/utils/ClientPoolImpl.java
```diff
@@ -55,14 +55,24 @@ public <R> R run(Action<R, C, E> action) throws E, InterruptedException {
...
+          synchronized (this) {
+            close(client);
+            currentSize -= 1;
+          }
+          synchronized (signal) {
```
yuqi1129: Why do we call `signal.notifyAll();`? Only when we successfully release a client to the pool will we notify all blocked  thread(see get())

keepConcentration: Thanks for the feedback. You’re right that `notifyAll()` should only be called when a client is actually returned to the pool. I’ll remove this code.

keepConcentration:
@yuqi1129 

Thanks for the feedback. I’ve updated the code as suggested and removed the unnecessary `notifyAll()`.

yuqi1129 (approved):

Issue #8558, opened 2025-09-16
[Bug report] metrics of backend connect pool is invisible


mchades:
### Version

main branch

### Describe what's wrong

As Gravitino supports backend connection pool metrics via #6470, these metrics should be available at the `/metrics` endpoint, but they were not.

### Error message and/or stacktrace

`activeConnectionCount`, `idleConnectionCount`, and `maxConnectionCount` not found in the metrics result

### How to reproduce

1. setup server
2. curl `ip:port/metrics`

PR #8560, opened 2025-09-16
[#8558] fix(metrics): fix missing backend connection pool metrics 

mchades:

 fix missing backend connection pool metrics 

Fix: #8558 

no

CI pass

### DIFF
--- a/core/src/main/java/org/apache/gravitino/metrics/MetricNames.java
+++ b/core/src/main/java/org/apache/gravitino/metrics/MetricNames.java
@@ -22,12 +22,9 @@ package org.apache.gravitino.metrics;
 public class MetricNames {
   public static final String HTTP_PROCESS_DURATION = "http-request-duration-seconds";
   public static final String SERVER_IDLE_THREAD_NUM = "http-server.idle-thread.num";
-  public static final String ENTITY_STORE_RELATION_DATASOURCE_ACTIVE_CONNECTIONS =
-      "entity-store.relation-datasource.active-connections";
-  public static final String ENTITY_STORE_RELATION_DATASOURCE_IDLE_CONNECTIONS =
-      "entity-store.relation-datasource.idle-connections";
-  public static final String ENTITY_STORE_RELATION_DATASOURCE_MAX_CONNECTIONS =
-      "entity-store.relation-datasource.max-connections";
+  public static final String DATASOURCE_ACTIVE_CONNECTIONS = "datasource.active-connections";
+  public static final String DATASOURCE_IDLE_CONNECTIONS = "datasource.idle-connections";
+  public static final String DATASOURCE_MAX_CONNECTIONS = "datasource.max-connections";
 
   private MetricNames() {}
 }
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/gravitino/metrics/source/RelationDatasourceMetricsSource.java
```diff
@@ -26,15 +26,10 @@
 public class RelationDatasourceMetricsSource extends MetricsSource {
 
   public RelationDatasourceMetricsSource(BasicDataSource dataSource) {
-    super(MetricsSource.GRAVITINO_SERVER_METRIC_NAME);
+    super(MetricsSource.GRAVITINO_RELATIONAL_STORE_METRIC_NAME);
```
mchades, line 29: The current solution may be a workaround. We should find a way to avoid the same issue by improving the metrics framework. @FANNG1 you can also update the code based on my PR.

FANNG1: update the PR to add the valid check when registering the metricSources

core/src/main/java/org/apache/gravitino/metrics/MetricNames.java
```diff
@@ -22,12 +22,9 @@
...
   public static final String SERVER_IDLE_THREAD_NUM = "http-server.idle-thread.num";
-  public static final String ENTITY_STORE_RELATION_DATASOURCE_ACTIVE_CONNECTIONS =
-      "entity-store.relation-datasource.active-connections";
-  public static final String ENTITY_STORE_RELATION_DATASOURCE_IDLE_CONNECTIONS =
-      "entity-store.relation-datasource.idle-connections";
-  public static final String ENTITY_STORE_RELATION_DATASOURCE_MAX_CONNECTIONS =
-      "entity-store.relation-datasource.max-connections";
+  public static final String DATASOURCE_ACTIVE_CONNECTIONS = "datasource.active-connections";
+  public static final String DATASOURCE_IDLE_CONNECTIONS = "datasource.idle-connections";
+  public static final String DATASOURCE_MAX_CONNECTIONS = "datasource.max-connections";
```
hqbhoho, line 27: Whether support custom labels in metrics？Because  `connection pool` will also use for jdbc catalog, if we want expose these metrics, it can reuse?

jerryshao:
@FANNG1 please help to review.

core/src/main/java/org/apache/gravitino/metrics/MetricNames.java
mchades: the final exposed metric name is `metricSourceName.metricName`, so the metrics name here can be reused

hqbhoho: okk, I see. 
For multi jdbc catalog, the mertics maybe like:
```
gravitino-jdbc-catalog.datasource.max-connections{ catalog='mysql1'}
gravitino-jdbc-catalog.datasource.max-connections{ catalog='mysql2'}
```
or 
```
gravitino-jdbc-catalog.mysql1.datasource.max-connections
gravitino-jdbc-catalog.mysql2.datasource.max-connections
```
Maybe we should support custom label  in mertircs？

mchades: good question! I prefer the style one.

The current PR does not include catalog metrics, it will expose backend metrics such as `gravitino-relational-store.datasource.max-connections`. We will address the catalog metrics in a separate PR for the next release.

Issue #8350, opened 2025-08-28
[Improvement] Add precondition to PartitionStatisticsListResponse.java to validating each partition statistic is not null preventing NPE


justinmclean:
### What would you like to be improved?

In common/src/main/java/org/apache/gravitino/dto/responses/PartitionStatisticsListResponse.java check if each partition statistic is not null.

Here's a test to help:
```
  @Test
  void testPartitionStatisticsListResponseNullElement() {
    AuditDTO audit = AuditDTO.builder().withCreator("user1").withCreateTime(Instant.now()).build();
    StatisticDTO statistic =
        StatisticDTO.builder().withName("stat").withAudit(audit).withValue(Optional.empty()).build();
    PartitionStatisticsDTO valid =
        PartitionStatisticsDTO.of("p1", new StatisticDTO[] {statistic});
    PartitionStatisticsDTO[] stats = new PartitionStatisticsDTO[] {valid, null};
    PartitionStatisticsListResponse response = new PartitionStatisticsListResponse(stats);
    assertThrows(IllegalArgumentException.class, response::validate);
  }
```

### How should we improve?

Add precondition

SteveArseneLee:
Hi @justinmclean
Can I take this issue?

justinmclean:
Sure, if you have any questions just ask.

PR #8566, opened 2025-09-16
fix(common): add null check to prevent NPE in PartitionStatisticsList…

SteveArseneLee:

Add null check precondition to `PartitionStatisticsListResponse.validate()` method to prevent NullPointerException when the partition statistics array contains null elements. The fix ensures that each partition statistic is validated for null before calling its `validate()` method.

* The `PartitionStatisticsListResponse.validate()` method was vulnerable to NPE when iterating through partition statistics array
* If the array contained null elements, calling `partitionStat.validate()` would throw NullPointerException
* The existing code only checked if the array itself was null, but not individual elements
* This created a potential runtime failure that should be caught with proper validation

Fix: #8350

No, this is a defensive programming improvement. The change prevents potential NPE by adding proper validation, but does not change the external API or behavior for valid inputs.

* Added comprehensive test case `testPartitionStatisticsListResponseNullElement()` to verify null elements trigger `IllegalArgumentException`
* Existing unit tests continue to pass, ensuring no functional regression
* Local testing completed successfully with Gradle build and Spotless formatting
* All common module tests compile and run successfully
* RAT license audit passes without issues


roryqi:
@SteveArseneLee Merged. Thanks for your effort.

Issue #8312, opened 2025-08-27
[Subtask] Support endpoint to retrieve valid credentials for a given table


FANNG1:
### Describe the subtask

`/v1/{prefix}/namespaces/{namespace}/tables/{table}/credentials`

### Parent issue

#6336

sunxiaojian:
@FANNG1 You can assign it to me.

sunxiaojian:
@FANNG1  I noticed that although Iceberg has defined the endpoint, it hasn't been implemented yet. Should we wait until Iceberg implements it before merging the related PR?

<img width="890" height="136" alt="Image" src="https://github.com/user-attachments/assets/ad118ea2-8240-45af-a814-6b2c5e1d7537" />

FANNG1:
> [@FANNG1](https://github.com/FANNG1) I noticed that although Iceberg has defined the endpoint, it hasn't been implemented yet. Should we wait until Iceberg implements it before merging the related PR?
> 
[...]

PR #8592, opened 2025-09-17
[#8312] feat(iceberg-rest):Support endpoint to retrieve valid credentials for a given table 

sunxiaojian:

Support endpoint to retrieve valid credentials for a given table.

Fix: #8312

N/A

org.apache.gravitino.iceberg.service.rest.TestIcebergTableOperations#testGetTableCredentials


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableEventDispatcher.java
```diff
@@ -230,4 +231,33 @@ public void renameTable(IcebergRequestContext context, RenameTableRequest rename
...
+    LoadCredentialsResponse loadCredentialsResponse;
+    try {
+      loadCredentialsResponse =
+          icebergTableOperationDispatcher.getTableCredentials(context, tableIdentifier);
+    } catch (Exception e) {
+      eventBus.dispatchEvent(new IcebergLoadTableFailureEvent(context, gravitinoNameIdentifier, e));
+      throw e;
+    }
+    // Note: Currently there is no dedicated IcebergLoadCredentialsEvent event, reusing
+    // IcebergLoadTableEvent
```
mchades: Should we add a new event for getTableCredentials in this PR? @FANNG1 

or leave a todo here instead of using the wrong event

sunxiaojian: ok, I'm adding a new event.

sunxiaojian: @mchades fixed

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -99,6 +102,29 @@ public LoadTableResponse loadTable(TableIdentifier identifier, boolean requestCr
     return loadTableResponse;
   }
 
+  /**
+   * Get table credentials.
+   *
+   * @param identifier
+   * @return
+   */
```
mchades, line 110: please complete the JavaDoc

mchades (commented):
Since you're adding a new API, would you also need to update the documentation about the IRS?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
sunxiaojian: fixed

sunxiaojian:
> Since you're adding a new API, would you also need to update the documentation about the IRS?

@mchades There should be no documentation to update, and the IRC API should not be open to the public yet

mchades:
> > Since you're adding a new API, would you also need to update the documentation about the IRS?
> 
>There should be no documentation to update, and the IRC API should not be open to the public yet
 
should the doc need to be updated?
https://github.com/apache/gravitino/blob/c2fd69d11dbb1a706b70e1763fc3e719aac1014c/docs/iceberg-rest-service.md?plain=1#L20

sunxiaojian:
> > > Since you're adding a new API, would you also need to update the documentation about the IRS?
> > 
> > 
> > There should be no documentation to update, and the IRC API should not be open to the public yet
> 
> should the doc need to be updated?
> 
> https://github.com/apache/gravitino/blob/c2fd69d11dbb1a706b70e1763fc3e719aac1014c/docs/iceberg-rest-service.md?plain=1#L20

@mchades updated

mchades:
@FANNG1 please help to review when you have time, thx

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -332,6 +333,45 @@ private String SerializeUpdateTableRequest(UpdateTableRequest updateTableRequest
...
+    try {
+      return Utils.doAs(
+          httpRequest,
+          () -> {
+            // Convert Iceberg table identifier to Gravitino NameIdentifier
+            TableIdentifier tableIdentifier = TableIdentifier.of(icebergNS, table);
+            // First check if the table exists
+            IcebergRequestContext context =
+                new IcebergRequestContext(httpServletRequest(), catalogName);
+            boolean exists = tableOperationDispatcher.tableExists(context, tableIdentifier);
```
FANNG1: How about removing the table exists check since the load table will throw table not found exception

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -99,6 +102,30 @@ public LoadTableResponse loadTable(TableIdentifier identifier, boolean requestCr
...
   }
 
+  /**
+   * Get table credentials.
+   *
+   * @param identifier The table identifier for which to load credentials
+   * @return A {@link org.apache.iceberg.rest.responses.LoadCredentialsResponse} object containing
+   *     the credentials.
+   */
+  public LoadCredentialsResponse loadTableCredentials(TableIdentifier identifier) {
```
FANNG1: loadTableCredentials -> getTableCredentials to keep consistent with others?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -99,6 +102,30 @@ public LoadTableResponse loadTable(TableIdentifier identifier, boolean requestCr
...
+    try {
+      LoadTableResponse loadTableResponse = super.loadTable(identifier);
+      Credential credential = getCredential(loadTableResponse);
+      // Convert Gravitino credential to Iceberg credential.
+      ImmutableCredential icebergCredential =
+          ImmutableCredential.builder()
+              .prefix(credential.credentialType())
```
FANNG1: it's not proper to passing credential type to prefix, the prefix is supposed to be the location like `s3://xx`, we could leave it empty for the prefix.

FANNG1:
LGTM except minor comments

sunxiaojian:
> LGTM except minor comments

@FANNG1 fixed all.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
sunxiaojian: Cannot use empty, the validation will fail.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -46,6 +46,20 @@ public class CredentialPropertyUtils {
   @VisibleForTesting
   static final String ICEBERG_ADLS_ACCOUNT_KEY = "adls.auth.shared-key.account.key";
 
+  // Map from credential types to their corresponding storage system prefixes
+  // Only includes the types supported by toIcebergProperties method
+  private static final Map<Class<? extends Credential>, String> CREDENTIAL_TYPE_TO_PREFIX_MAP =
```
FANNG1: Gravitino provides pluggable credential provider to support custom credentials.  I'm afraid this may break it.

FANNG1: From personal perspective, I prefer to ignore `prefix` properties, because Gravitino doesn't support generate different credential for different locations yet, WDYT?

sunxiaojian: ok, then I'll reimplement an iceberg Credential, since ImmutableCredential doesn't allow ignoring the prefix

FANNG1: does it support passing a empty string `""`??

sunxiaojian: > ok, then I'll reimplement an iceberg Credential, since ImmutableCredential doesn't allow ignoring the prefix

If the implementation involves using the prefix in logical processing during usage, this approach would be problematic. However, it is currently believed that the client should not use it; only the Credential is needed.

sunxiaojian: > does it support passing a empty string `""`??

unsupported，I customized it，Or we can fill this value with the fixed value "gravitino";

FANNG1:
@sunxiaojian  merged to main, thanks for your work!

Issue #8596, opened 2025-09-18
[Subtask] add metrics for backend queries


mchades:
### Describe the subtask

In scenarios with high concurrency requests, the backend query is a time-consuming operation. We need metrics related to backend query functions for observation and optimization.

### Parent issue

#8594

PR #8606, opened 2025-09-18
[#8596] improvement(metrics): add metrics for backend meta service

mchades:

- introduce aspectj for metric annotation
- add metrics for backend meta service

Fix: #8596 

local tested:
image

### DIFF
--- a/core/src/main/java/org/apache/gravitino/metrics/MethodMonitorAspect.java
+++ b/core/src/main/java/org/apache/gravitino/metrics/MethodMonitorAspect.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.metrics;
+
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.GravitinoEnv;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Pointcut;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Aspect
+public class MethodMonitorAspect {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MethodMonitorAspect.class);
+  private MetricRegistry metricRegistry;
+
+  public MethodMonitorAspect() {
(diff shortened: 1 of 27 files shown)


gradle/libs.versions.toml
```diff
@@ -127,8 +127,10 @@ jcasbin = "1.81.0"
 ognl = "3.4.7"
 concurrent-trees = "2.6.0"
 jakarta-validation = "2.0.2"
+aspectj = "1.9.24"
 
 [libraries]
+aspectj-aspectjrt = { group = "org.aspectj", name = "aspectjrt", version.ref = "aspectj" }
```
jerryshao, line 133: Can you please 1. check the license of this dependency; 2. Update the license.bin to add this; 3. check the transitive dependencies of this one, and exclude the conflicted/unused ones if necessary.

mchades: Updated the license; AspectJ does not have third-party dependencies.

core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
```diff
@@ -77,14 +81,7 @@ public FilesetPO getFilesetPOBySchemaIdAndName(Long schemaId, String filesetName
     return filesetPO;
   }
 
-  // Fileset may be deleted, so the FilesetPO may be null.
-  public FilesetPO getFilesetPOById(Long filesetId) {
```
FANNG1, line 81: Not used any more?

core/src/main/java/org/apache/gravitino/metrics/MethodMonitorAspect.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Around("monitoredMethods() && @annotation(monitored)")
+  public Object monitorMethod(ProceedingJoinPoint pjp, Monitored monitored) throws Throwable {
+    if (metricRegistry == null) {
+      return pjp.proceed();
+    }
+
+    String baseName = buildBaseName(pjp, monitored);
+
+    Timer timer = metricRegistry.timer(MetricRegistry.name(baseName, "total"));
```
FANNG1: Could we use MetricsSource to manage the metrics?

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
```diff
@@ -83,22 +86,15 @@ public CatalogPO getCatalogPOByName(String metalakeName, String catalogName) {
     return catalogPO;
   }
 
+  @Monitored(
```
FANNG1, line 91: If all the method are using `GRAVITINO_RELATIONAL_STORE_METRIC_NAME+xx`, I would suggest to use two arguments for `@Monitored`

core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
mchades: I think so. I can't see its use, and if it's really needed later, we can add it back.

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
mchades: The two-parameter approaches are not general and flexible. Current style follows the existing metrics annotation style:
https://github.com/apache/gravitino/blob/c2fd69d11dbb1a706b70e1763fc3e719aac1014c/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java#L85

core/src/main/java/org/apache/gravitino/metrics/MethodMonitorAspect.java
mchades: It seems it cannot. Since the annotation can be attached to any function, metrics can come from different MetricsSources.

FANNG1: we could specify the metric source in the annotation

mchades: fixed

FANNG1 (approved):

Issue #7962, opened 2025-08-07
[Subtask] Add the Local Spark runner for LocalExecutor


jerryshao:
### Describe the subtask

Currently, we already have shell runner for LocalExecutor, this issue will also add the Spark support, so that user can run Spark applications by Gravitino's job system.

### Parent issue

#7154

PR #8607, opened 2025-09-18
[#7962]feat(core): Add spark process builder support for local job executor

jerryshao:

This PR adds the Spark process builder for local job executor, so users can run the Spark jobs in the local environment.

This is a part of the work of the job system

Fix: #7962

No.

UT and local tests.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
+++ b/api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
@@ -18,13 +18,11 @@
  */
 package org.apache.gravitino.job;
 
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import org.apache.commons.lang3.StringUtils;
 
 /**
  * Represents a job template for executing Spark applications. This class extends the JobTemplate
@@ -144,7 +142,9 @@ public class SparkJobTemplate extends JobTemplate {
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder("\nSparkJobTemplate{\n");
-    sb.append("  className='").append(className).append("',\n");
+    if (className != null) {
+      sb.append("  className='").append(className).append("',\n");
+    }
 
     if (!jars.isEmpty()) {
       sb.append("  jars=[\n");
@@ -279,9 +279,6 @@ public class SparkJobTemplate extends JobTemplate {
     @Override
     protected void validate() {
       super.validate();
-
-      Preconditions.checkArgument(
-          StringUtils.isNotBlank(className), "Class name must not be null or empty");
 
       this.jars = jars != null ? ImmutableList.copyOf(jars) : ImmutableList.of();
       this.files = files != null ? ImmutableList.copyOf(files) : ImmutableList.of();
--- a/core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
+++ b/core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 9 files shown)


api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
```diff
@@ -280,9 +280,6 @@ protected Builder self() {
     protected void validate() {
       super.validate();
 
-      Preconditions.checkArgument(
```
FANNG1, line 283: please update the document too,  https://github.com/apache/gravitino/blob/c2fd69d11dbb1a706b70e1763fc3e719aac1014c/docs/manage-jobs-in-gravitino.md?plain=1#L147  https://github.com/apache/gravitino/blob/c2fd69d11dbb1a706b70e1763fc3e719aac1014c/docs/manage-jobs-in-gravitino.md?plain=1#L171

docs/manage-jobs-in-gravitino.md
```diff
@@ -548,6 +545,7 @@ The following are the default configurations for the local job executor:
 | `gravitino.jobExecutor.local.waitingQueueSize`      | The size of the waiting queue for queued jobs in the local job executor   | `100`                                  | No       | 1.0.0         |
 | `gravitino.jobExecutor.local.maxRunningJobs`        | The maximum number of running jobs in the local job executor              | `max(1, min(available cores / 2, 10))` | No       | 1.0.0         |
 | `gravitino.jobExecutor.local.jobStatusKeepTimeInMs` | The time in milliseconds to keep the job status in the local job executor | `3600000` (1 hour)                     | No       | 1.0.0         |
+| `gravitino.jobExecutor.local.sparkHome`             | The home directory of Spark, it is required for running Spark jobs        | `None`                                 | No       | 1.1.0         |
```
FANNG1: Is this configuration used to support multiple Spark versions?  Could we add a default Spark home? then the user doesn't need to specify it every time.

core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+    builder.directory(workingDirectory);
+    builder.environment().putAll(sparkJobTemplate.environments());
+
+    File outputFile = new File(workingDirectory, "output.log");
+    File errorFile = new File(workingDirectory, "error.log");
+
+    builder.redirectOutput(outputFile);
+    builder.redirectError(errorFile);
+
+    LOG.info("Starting local Spark job with command: {}", String.join(" ", commandList));
```
FANNG1: Could we add environment vars to the logs to help debug?

core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.job.SparkJobTemplate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The SparkProcessBuilder class is responsible for constructing and starting a local Spark job
+ * process using the spark-submit command. Using local job executor to run Spark job has some
+ * limitations: 1. It uses the aliveness of the spark-submit process to determine if the job is
+ * still running, which is not accurate for cluster deploy mode. 2. It cannot support user
```
FANNG1: `cluster deploy mode` -> `the cluster deploy mode`?

FANNG1:
LGTM except minor comments

docs/manage-jobs-in-gravitino.md
jerryshao: It is a system-level configuration; users don't need to configure it at each time. Also, it is hard for Gravitino to figure out a default value. What I can think of is to read the `SPARK_HOME` environment variable if this is not configured, WDYT?

roryqi:
Can Spark job will  use the Gravitino Spark connector easily?

jerryshao:
> Can Spark job will use the Gravitino Spark connector easily?

No, that's unrelated.

FANNG1:
> > Can Spark job will use the Gravitino Spark connector easily?
> 
> No, that's unrelated.

If there are multiple catalogs,  seems we could leverage Gravitino Spark connector to simple the catalog related configuration for Spark job.

docs/manage-jobs-in-gravitino.md
FANNG1: ok

api/src/main/java/org/apache/gravitino/job/SparkJobTemplate.java
FANNG1: Seems not to be fixing the second sentence?

jerryshao: Fixed.

FANNG1 (approved):

Issue #8632, opened 2025-09-19
[Subtask] add core workflow for renameTable with new schema name


mchades:
### Describe the subtask

Support renaming a table with a new schema name, except for the catalog implementation

### Parent issue

#8375

PR #8635, opened 2025-09-19
[#8632] feature(alterTable): add core workflow for renameTable with newSchema

mchades:

dd core workflow for renameTable with newSchema

Fix: #8632 

no, it is backward compatibility

tests added


core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -199,12 +200,27 @@ public Table alterTable(NameIdentifier ident, TableChange... changes)
...
+      if (change instanceof TableChange.RenameTable) {
+        TableChange.RenameTable rename = (TableChange.RenameTable) change;
+        if (rename.getNewSchemaName().isPresent()
+            && !rename.getNewSchemaName().get().equals(schemaName)) {
+          nameIdentifierForLock = getCatalogIdentifier(ident);
+          break;
+        }
+        nameIdentifierForLock = getSchemaIdentifier(ident);
+      }
+    }
```
jerryshao, line 223: I saw it always uses the write lock, which is different from the comment mentioned above "use the read lock on the table if there does not exist TableChange.RenameTable in the changes".

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -251,20 +267,37 @@ public Table alterTable(NameIdentifier ident, TableChange... changes)
                           TableEntity.class,
                           TABLE,
                           tableEntity -> {
-                            String newName =
+                            Namespace newNamespace =
                                 Arrays.stream(changes)
-                                    .filter(c -> c instanceof TableChange.RenameTable)
-                                    .map(c -> ((TableChange.RenameTable) c).getNewName())
+                                    .filter(
```
jerryshao: Can we move this to a helper method to simplify the code here?

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
```diff
@@ -182,9 +190,9 @@ public <E extends Entity & HasIdentifier> TableEntity updateTable(
               updateResult.set(
                   SessionUtils.getWithoutCommit(
                       TableMetaMapper.class,
-                      mapper -> mapper.updateTableMeta(newTablePO, oldTablePO))),
+                      mapper -> mapper.updateTableMeta(newTablePO, oldTablePO, newSchemaId))),
           () -> {
-            if (updateResult.get() > 0 && isColumnChanged) {
+            if (updateResult.get() > 0 && (isColumnChanged || isSchemaChanged)) {
```
jerryshao, line 195: You should add the tests to cover the changes in meta service here and bove.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
mchades: fixed

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
mchades: added

mchades:
all comments resolved, please help to review again, thx! @jerryshao

Issue #8638, opened 2025-09-22
[Subtask] Add Java and Python interface for job template alteration


jerryshao:
### Describe the subtask

This task tracks the work of defining the Java and Python interface for job template alteration.

### Parent issue

#8429

PR #8644, opened 2025-09-22
[#8638] feat(api): Add Java and Python interface for job template alteration

jerryshao:

Add the Java and Python APIs for job template alteration.

This is the 1st pr of supporting job template alteration.

Fix: #8638 

Yes.

Add UTs.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/job/JobTemplateChange.java
+++ b/api/src/main/java/org/apache/gravitino/job/JobTemplateChange.java
@@ -1,0 +1,701 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.job;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * The interface for job template changes. A job template change is an operation that modifies a job
+ * template. It can be one of the following:
+ *
+ * <ul>
+ *   <li>Rename the job template.
+ *   <li>Update the comment of the job template.
+ *   <li>Update the job template details, such as executable, arguments, environments, custom
+ *       fields, etc.
+ * </ul>
+ */
+public interface JobTemplateChange {
+
+  /**
+   * Creates a new job template change to update the name of the job template.
+   *
+   * @param newName The new name of the job template.
+   * @return The job template change.
+   */
+  static JobTemplateChange rename(String newName) {
+    return new RenameJobTemplate(newName);
+  }
+
+  /**
+   * Creates a new job template change to update the comment of the job template.
+   *
+   * @param newComment The new comment of the job template.
+   * @return The job template change.
+   */
+  static JobTemplateChange updateComment(String newComment) {
+    return new UpdateJobTemplateComment(newComment);
+  }
+
+  /**
+   * Creates a new job template change to update the details of the job template.
+   *
+   * @param templateUpdate The job template update details.
+   * @return The job template change.
+   */
+  static JobTemplateChange updateTemplate(TemplateUpdate templateUpdate) {
+    return new UpdateJobTemplate(templateUpdate);
+  }
+
+  /** A job template change to rename the job template. */
+  final class RenameJobTemplate implements JobTemplateChange {
+    private final String newName;
+
+    private RenameJobTemplate(String newName) {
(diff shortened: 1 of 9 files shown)


jerryshao:
@FANNG1 @mchades can you please help to review this PR?

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -503,7 +503,32 @@ void pullAndUpdateJobStatus() {
 
       activeJobs.forEach(
           job -> {
-            JobHandle.Status newStatus = jobExecutor.getJobStatus(job.jobExecutionId());
+            JobHandle.Status newStatus = job.status();
+            try {
+              newStatus = jobExecutor.getJobStatus(job.jobExecutionId());
+            } catch (NoSuchJobException e) {
+              // If the job is not found in the external job executor, we assume the job is
```
FANNG1, line 510: Which conditions will throw `NoSuchJobException`?   why not using a new stats like `UNKNOWN`?

api/src/main/java/org/apache/gravitino/job/JobTemplateChange.java
```diff
@@ -0,0 +1,701 @@
...
+ */
+public interface JobTemplateChange {
+
+  /**
+   * Creates a new job template change to update the name of the job template.
+   *
+   * @param newName The new name of the job template.
+   * @return The job template change.
+   */
+  static JobTemplateChange rename(String newName) {
```
FANNG1, line 49: For  `newName`, `newComment`,  could user pass a `null` value? Should we add a validation?

clients/client-python/gravitino/api/job/job_template_change.py
```diff
@@ -0,0 +1,425 @@
...
+    template. It can be one of the following:
+
+      - Rename the job template.
+      - Update the comment of the job template.
+      - Update the job template details, such as executable, arguments, environments, custom
+        fields, etc.
+    """
+
+    @staticmethod
+    def rename(new_name: str) -> "JobTemplateChange":
```
FANNG1, line 34: return "RenameJobTemplate" ?

clients/client-python/gravitino/api/job/job_template_change.py
```diff
@@ -0,0 +1,425 @@
...
+            and self._new_scripts == other._new_scripts
+        )
+
+    def __hash__(self) -> int:
+        return hash((super().__hash__(), tuple(self._new_scripts)))
+
+
+class SparkTemplateUpdate(TemplateUpdate):
+    """
+    A template update for spark job templates.
```
FANNG1, line 321: spark -> Spark?

api/src/main/java/org/apache/gravitino/job/JobTemplateChange.java
jerryshao: No, it cannot be, I will verify them in the server-side request.

clients/client-python/gravitino/api/job/job_template_change.py
jerryshao: I think it is fine, `JobTemplateChange` is the parent class, so I think it should be fine.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: Because we get the status from the external job executor. If the job is deleted from the external job executor, then when we get the job status it will be failed to get the job.

api/src/main/java/org/apache/gravitino/job/JobTemplateChange.java
```diff
@@ -0,0 +1,701 @@
...
+    }
+
+    /**
+     * Get the string representation of the job template change.
+     *
+     * @return The string representation of the job template change.
+     */
+    @Override
+    public String toString() {
+      return "RENAME JOB TEMPLATE " + newName;
```
yuqi1129, line 120: RENAME JOB TEMPLATE **to**

clients/client-python/gravitino/api/job/supports_jobs.py
```diff
@@ -89,6 +90,27 @@ def delete_job_template(self, job_template_name: str) -> bool:
...
+        Alters a job template by applying the specified changes. This allows for modifying the
+        properties of an existing job template, such as its name, description, or parameters.
+
+        Args:
+            job_template_name: The name of the job template.
+            changes: The changes to apply to the job template.
+
+        Raises:
+            NoSuchJobTemplateException: If the job template does not exist.
+            IllegalArgumentException: If any of the changes cannot be applied to the job template.
```
yuqi1129, line 106: There may be a case where a job template with the name `newName` already exists, so it belongs to `IllegalArgumentException` here?

core/src/main/java/org/apache/gravitino/job/JobManager.java
FANNG1: why return `SUCC` for job not exists condition?

clients/client-python/gravitino/api/job/supports_jobs.py
jerryshao: No, it will not be. To align with other implementation, it will be a runtime exception.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: Because we don't have better options, unless we add a new unknown state.

api/src/main/java/org/apache/gravitino/job/JobTemplateChange.java
jerryshao: I aligned with other xxxChanges.

core/src/main/java/org/apache/gravitino/job/JobManager.java
jerryshao: This is not the normal case, I'm hesitant to add a new `UNKNOWN` state. Usually, there should be no `UNKNOWN` state for a job. I'm going to change to use `FAILED` instead and add more logs to let users know such a situation.

FANNG1 (approved):

FANNG1:
LGTM

Issue #8658, opened 2025-09-23
[Improvement] In ModelVersionMetaH2Provider.java xorrect H2 SQL


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/h2/ModelVersionMetaH2Provider.java in insertModelVersionMetas and insertModelVersionMetasWithVersionNumber the SQL uses MEDIUMTEXT, while this is valid for MySQL it's not for H2. See https://www.h2database.com/html/datatypes.html

### How should we improve?

Use CLOB instead?

Raj-G07:
@justinmclean Can i take up this issue ? Also for `TEXT` we can apply the same .

justinmclean:
All yours, if you need a hand, just ask.

PR #8675, opened 2025-09-24
[#8658]improvement: Replace MySQL types with H2 compatible casts in ModelVersionMetaH2Provider and h2 scripts

Raj-G07:
- This PR updates the `ModelVersionMetaH2Provider` to make it fully compatible with H2 by removing
 SQL specific column types (`MEDIUMTEXT`, `TEXT`) and replacing them with `CLOB`.

Fix: #8658


Raj-G07:
@justinmclean In gravitino/scripts/h2 there is schema and upgrade which also contain mysql types , should we have to work on this ?

yuqi1129:
It seems that H2 will convert `MEDIUMTEXT,TEXT` to `CLOB` implicitly, which is why the current query SQL and create table scripts work. 

By the way, if you want to use `CLOB` explicitly, I believe there are many places that need to be changed.

justinmclean:
@yuqi1129 Can you find that documented anywhere?

yuqi1129:
> @yuqi1129 Can you find that documented anywhere?

The content in H2 documents is unclear about the behaviour of  column `TEXT` in  "compatible mode" for MySQL, see https://www.h2database.com/html/features.html#compatibility. Everything needs to be confirmed through a real test.

Raj-G07:
MySQL compatibility mode helps a lot, but it doesn’t mean full 1:1 compatibility. Am i right?

yuqi1129:
> MySQL compatibility mode helps a lot, but it doesn’t mean full 1:1 compatibility. Am i right?

That's correct.

Raj-G07:
But for production , it will not be 100% equivalent . So what choice we should take whether to use it or not ?

yuqi1129:
Everything is working well at the moment. If you guys prefer to use `CLOB` for H2, then please go ahead. There will be many places that need to be renovated, for example, all table definitions in `scripts/h2`, query strings. Please take the time to change all.

Raj-G07:
- Can someone review `schema-0.6.0-h2.sql` ?
- Good for us to do in the right way 
- Am i on the right track?

yuqi1129:
> * Can someone review `schema-0.6.0-h2.sql` ?
> * Good for us to do in the right way
> * Am i on the right track?

I will take the time to review the changes today. Thank you for your hard work.

scripts/h2/schema-0.6.0-h2.sql
```diff
@@ -17,249 +17,412 @@
...
-    `metalake_name` VARCHAR(128) NOT NULL COMMENT 'metalake name',
-    `metalake_comment` VARCHAR(256) DEFAULT '' COMMENT 'metalake comment',
-    `properties` MEDIUMTEXT DEFAULT NULL COMMENT 'metalake properties',
-    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'metalake audit info',
-    `schema_version` MEDIUMTEXT NOT NULL COMMENT 'metalake schema version info',
-    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'metalake current version',
-    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'metalake last version',
-    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'metalake deleted at',
+CREATE TABLE IF NOT EXISTS metalake_meta (
+    metalake_id BIGINT NOT NULL,
```
yuqi1129: 1. Is there no need to backquote the column name in H2?
2. Is the display length for types like `BIGINT` not supported in H2?

scripts/h2/upgrade-0.6.0-to-0.7.0-h2.sql
```diff
@@ -17,26 +17,26 @@
...
+    `column_position` INT NOT NULL COMMENT 'column position, starting from 0',
+    `column_type` CLOB NOT NULL COMMENT 'column type',
     `column_comment` VARCHAR(256) DEFAULT '' COMMENT 'column comment',
-    `column_nullable` TINYINT(1) NOT NULL DEFAULT 1 COMMENT 'column nullable, 0 is not nullable, 1 is nullable',
-    `column_auto_increment` TINYINT(1) NOT NULL DEFAULT 0 COMMENT 'column auto increment, 0 is not auto increment, 1 is auto increment',
-    `column_default_value` TEXT DEFAULT NULL COMMENT 'column default value',
-    `column_op_type` TINYINT(1) NOT NULL COMMENT 'column operation type, 1 is create, 2 is update, 3 is delete',
-    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'column deleted at',
-    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'column audit info',
+    `column_nullable` SMALLINT NOT NULL DEFAULT 1 COMMENT 'column nullable, 0 is not nullable, 1 is nullable',
```
yuqi1129: If H2 supports defining column comments in creating table syntax, we should not use a separate sentence to do it to reduce the chances.

Raj-G07:
- BIGINT(X) is simply mapped to BIGINT , it ignores X 
- In H2 SQL, `COMMENT ON COLUMN` statement is the standard and recommend way to add comments in H2 .
- Can we continue ?

yuqi1129:
> * BIGINT(X) is simply mapped to BIGINT , it ignores X
> * In H2 SQL, `COMMENT ON COLUMN` statement is the standard and recommend way to add comments in H2 .
> * Can we continue ?

Yeah, everything look okay, please just go ahead.

Raj-G07:
What am i missing ? Follow the same approach as we do in `schema-0.6.0-h2.sql`

Raj-G07:
What am i missing ? Need help !

scripts/h2/upgrade-0.9.0-to-1.0.0-h2.sql
```diff
@@ -67,50 +67,50 @@ ALTER TABLE `model_version_info` ADD CONSTRAINT `uk_mid_ver_uri_del` UNIQUE (`mo
...
-    `job_template_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'job template id',
-    `metalake_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'metalake id',
-    `job_execution_id` varchar(256) NOT NULL COMMENT 'job execution id',
-    `job_run_status` varchar(64) NOT NULL COMMENT 'job run status',
-    `job_finished_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'job finished at',
-    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'job run audit info',
-    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'job run current version',
-    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'job run last version',
-    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'job run deleted at',
+    `job_run_id` BIGINT NOT NULL COMMENT 'job run id',
```
yuqi1129: Why do you choose separate SQL sentences to add comments for table columns in `schema-0.6.0-h2.sql` but use MySQL-sytle column comment format here?

Raj-G07:
I tried to write comment in other schema but don't know why some ci/cd pipeline not works so i believe to revert back the changes as i did in [22fb869](https://github.com/apache/gravitino/pull/8675/commits/22fb869392ee344e0465fbca50e4a755e8a3aa1c) and only focus to replace mysql types

yuqi1129:
> I tried to write comment in other schema but don't know why some ci/cd pipeline not works so i believe to revert back the changes as i did in [22fb869](https://github.com/apache/gravitino/pull/8675/commits/22fb869392ee344e0465fbca50e4a755e8a3aa1c) and only focus to replace mysql types

I suggest you only focus on the problem of type `CLOB` to make the changes to this PR in control.

yuqi1129:
The change looks good to me. Let's wait for the CI to finish.

Raj-G07:
Thanks for reviewing and merging it!

Issue #8661, opened 2025-09-23
[Improvement] In PolicyOperations.java fix exceptiuon handling


justinmclean:
### What would you like to be improved?

In server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java, setPolicy fix the exception handling so that it handles disabled policies.

### How should we improve?

Needs to handle both OperationType.ENABLE and OperationType.DISABLE. Add a unit test to check this.

KayMas2808:
Hi, I'm a new contributor and I would like to work on this issue

KayMas2808:
Hi @justinmclean, raised a [PR](https://github.com/apache/gravitino/pull/8681) against this issue, please check when you get the chance. Thanks!

PR #8681, opened 2025-09-24
[#8661] fix(server): Correct exception handling in setPolicy and corresponding unit test

KayMas2808:

This PR fixes a bug in the exception handling of the `setPolicy` method in `PolicyOperations.java`.

The `OperationType` variable is now determined from the request **before** the `try` block to ensure it's available in the `catch` block. The `catch` block now uses this dynamic variable instead of a hardcoded `OperationType.ENABLE`.

Previously, if an exception occurred while a user was trying to **disable** a policy via the PATCH API endpoint, the error handler would incorrectly report it as a failure of an **enable** operation. This was misleading and made debugging difficult.

Fix: #8661

Yes. The error response returned to an API user when a `disable` policy operation fails is now correctly identified. There are no changes to the API's structure or success paths.

- Added a new unit test, `testSetPolicyDisableFailure`, to `TestPolicyOperations.java`.
- This test specifically mocks a `RuntimeException` during a `disablePolicy()` call.
- It verifies that the API returns the expected `500 Internal Server Error` and that the error payload is correct.
- Confirmed that all existing tests in the `:server` module continue to pass after the change.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
@@ -208,11 +208,13 @@ public class PolicyOperations {
       PolicySetRequest request) {
     LOG.info("Received set policy request for policy: {} under metalake: {}", name, metalake);
 
+    OperationType op = request.isEnable() ? OperationType.ENABLE : OperationType.DISABLE;
+
     try {
       return Utils.doAs(
           httpRequest,
           () -> {
-            if (request.isEnable()) {
+            if (op == OperationType.ENABLE) {
               policyDispatcher.enablePolicy(metalake, name);
             } else {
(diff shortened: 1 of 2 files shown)


justinmclean:
We use Spotless to ensure that the code is correctly formatted, and teh CI is failing because of that. You'll need to run `./gradlew :server:spotlessApply` and push the changes to correct this.

KayMas2808:
Thanks for the info! I've run the command and commited the change.

jerryshao:
@mchades can you please review this PR?

mchades (approved):
lgtm. Thanks for your contributions!

jerryshao:
We'd better also cherry-pick this commit to branch-1.0.

@KayMas2808 Would you please create a PR against branch-1.0 by cherry-picking this merged commit to branch-1.0?

Issue #8668, opened 2025-09-24
[Improvement] Prevent ClientPoolImpl.java  from handing out new clients after pool closure


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/utils/ClientPoolImpl.java in get there an infitite loop,While the pool closure is first checked,  the pool could be closed at some point while the loop is running.

### How should we improve?

Prevent handing out new clients after a pool closure.

keepConcentration:
Hi, @justinmclean 

Can I be assigned to this issue?

justinmclean:
All yours

PR #8682, opened 2025-09-25
[#8668] improvement(core): prevent handing out new clients after pool closure

keepConcentration:

- Fixed an issue in `ClientPoolImpl#get` where a new client could still be returned after the pool was closed.  
- Modified the loop to re-check the `closed` flag to ensure no new clients are created or returned after pool closure.  

Improved so that when the pool is closed, `get()` calls fail and throw an exception.

Fix: #8668

No user-facing changes.

Executed existing unit tests

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
@@ -49,7 +49,8 @@ public final class DorisUtils {
 
   private static final Pattern DISTRIBUTION_INFO_PATTERN =
       Pattern.compile(
-          "DISTRIBUTED BY\\s+(HASH|RANDOM)\\s*(\\(([^)]+)\\))?\\s*(BUCKETS\\s+(\\d+|AUTO))?");
+          "DISTRIBUTED BY\\s+(HASH|RANDOM)\\s*(\\(([^)]+)\\))?(?:\\s*BUCKETS\\s+(\\d+|AUTO))?",
+          Pattern.CASE_INSENSITIVE);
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/gravitino/utils/ClientPoolImpl.java
```diff
@@ -131,10 +131,11 @@ public void close() {
   }
 
   private C get() throws InterruptedException {
-    Preconditions.checkState(!closed, "Cannot get a client from a closed pool");
     while (true) {
+      Preconditions.checkState(!closed, "Cannot get a client from a closed pool");
       if (!clients.isEmpty() || currentSize < poolSize) {
         synchronized (this) {
+          Preconditions.checkState(!closed, "Cannot get a client from a closed pool");
```
justinmclean: While this works, I'm not sure that using a precondition is the best approach.

yuqi1129: The value `closed` will only be changed in the method `close`, which is not protected by the object lock, so what the benefit of checking it again and again in the synchronized code blocks?

keepConcentration: @justinmclean
Thank you for pointing this out. I realize now that my understanding of `Preconditions` was insufficient. I'll refactor this to use explicit condition checks instead.

@yuqi1129
You're absolutely right, and I appreciate you catching this. I didn't fully consider that since `closed` is not protected by the object lock, the redundant check inside the synchronized block is unnecessary. I'll remove the redundant checks.

keepConcentration:
@yuqi1129 @justinmclean 

Removed the redundant `closed` check inside the `synchronized` block. Additionally, modified to check `closed` only once at the beginning of while loop body.

Issue #7916, opened 2025-08-04
[Improvement] Using entity store's relation operation to refactor tag operation


jerryshao:
### What would you like to be improved?

Current tag operation in entity store uses specific APIs, and we have another common relation operation APIs to support relation operations. We can use relation operations to refactor the current tag storage APIs.

yuqi1129:
I won't do it in this version until the relational API with cache is stable.

PR #8689, opened 2025-09-25
[#7916] improvment(core): Using entity store's relation operation to refactor tag operation

yuqi1129:

Replace `SupportsTagOperations` with the entity store's relation operations 

The entity store's relation operations are general enough, and there is no need to create a new interface specifically for tags.

Fix: #7916

N/A

Existing UTs and ITs.


jerryshao (approved):

jerryshao:
@mchades can you please take another look?

core/src/main/java/org/apache/gravitino/EntityStore.java
```diff
@@ -229,22 +228,6 @@ <E extends Entity & HasIdentifier> void batchPut(List<E> entities, boolean overw
...
-    throw new UnsupportedOperationException("tag operations are not supported");
-  }
-
-  /**
-   * Get the extra relation operations that are supported by the entity store.
-   *
-   * @return the relation operations that are supported by the entity store
-   * @throws UnsupportedOperationException if the extra operations are not supported
-   */
   default SupportsRelationOperations relationOperations() {
```
mchades, line 237: why remove the JavaDoc of the method?

mchades (commented):
should we do something about the cache?

yuqi1129:
> should we do something about the cache?

I believe there is no need to make any changes to the cache? Have you done it for `policy`?

mchades (approved):
lgtm.

Issue #8659, opened 2025-09-23
[Improvement] Fix batchInsertUserRoleRel  template issue in UserRoleRelBaseSQLProvider.java


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/UserRoleRelBaseSQLProvider.java in batchInsertUserRoleRel, the template parameter is userRoleRelList, but userRoleRels is used.

### How should we improve?

Make sure they match. Other code seems to use userRoleRels

KayMas2808:
I would like to work on this.

PR #8690, opened 2025-09-25
[#8659] fix(core): Fix parameter mismatch in UserRoleRelBaseSQLProvider

KayMas2808:

This PR renames the `userRoleRelList` parameter to `userRoleRels` in the `batchInsertUserRoleRel` method of `UserRoleRelBaseSQLProvider.java`. This change applies to both the Java method parameter and its `@Param` annotation to match the collection name used within the MyBatis SQL template.

The mismatch between the parameter name defined in the Java code (`userRoleRelList`) and the name expected by the SQL template (`userRoleRels`) would cause a runtime error when the batch insert operation is called [cite].

This fix resolves the bug and also makes the method signature consistent with other methods in the same file, such as `batchInsertUserRoleRelOnDuplicateKeyUpdate` [cite].

Fix: #8659

No. This is an internal fix to the storage layer and does not alter any user-facing APIs or behavior.

The fix was verified by running the complete test suite for the `:core` module. All existing tests continue to pass.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/storage/relational/mapper/TestUserRoleRelMapper.java
+++ b/core/src/test/java/org/apache/gravitino/storage/relational/mapper/TestUserRoleRelMapper.java
@@ -1,0 +1,217 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/UserRoleRelBaseSQLProvider.java
```diff
@@ -28,8 +28,7 @@
 
 public class UserRoleRelBaseSQLProvider {
 
-  public String batchInsertUserRoleRel(
-      @Param("userRoleRelList") List<UserRoleRelPO> userRoleRelList) {
+  public String batchInsertUserRoleRel(@Param("userRoleRels") List<UserRoleRelPO> userRoleRels) {
```
yuqi1129: Can you add some UTs or ITs to cover the change to avoid similar problems in the future?

Issue #8662, opened 2025-09-24
[Improvement] Fix caching in CaffeineEntityCache.java when relationType is not null


justinmclean:
### What would you like to be improved?

In core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java anything with a relationType of not null is not cached. It's possible this might not be an issue, but a deliberate design choice. If so, that reason needs to be documented in the code as a comment.

### How should we improve?

Determine if this is a design choice or an omission, and either comment or fix.

keepConcentration:
Hi, @justinmclean

This is likely a bug. Since the condition `key.relationType() != null` was added, the code block under `if (existingEntities != null && key.relationType() != null)` has become dead code.

[CaffeineEntityCache.java](https://github.com/apache/gravitino/blob/main/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java#L305)

yuqi1129:
This issue has been fixed.

PR #8297, opened 2025-08-27
[#7804][#7797][#8155][#8181][#8182][#7787] fix(cache): Fix metadata cache data consistency error

xunliu:
### What changes were proposed in this pull request?

This PR fixes metadata cache data consistency errors by implementing reverse indexing for cache invalidation and changing the default cache setting to disabled. The changes ensure that when cached entities are invalidated, all dependent cached relationships are properly cleaned up.

Key changes:
- Added reverse indexing mechanism to track entity relationships in cache
- Refactored cache key hierarchy with new `EntityCacheRelationKey` class
- Changed default cache setting from enabled to disabled

### Why are the changes needed?

Fix: #7804
Fix: #7797
Fix: #8155
Fix: #8181 
Fix: #8182 
Fix: #7787

### Does this PR introduce _any_ user-facing change?

NA

### How was this patch tested?

CI

PR #8695, opened 2025-09-26
[#8662] Fix(core): Correctly cache entities with a relationType

keepConcentration:

This PR addresses an issue in `CaffeineEntityCache.java` where entities with a non-null `relationType` were not being cached as intended.
The problem was traced to a `return` statement in the `syncEntitiesToCache` method that incorrectly prevented the caching of relation-based data.
This change removes the aforementioned statement.

The previous behavior caused the cache to be bypassed for these operations, leading to unnecessary backend requests and preventing intended performance benefits. This fix ensures the caching mechanism now operates as designed.

Fix: #8662

No user-facing changes.

Executed existing unit tests.
Added a new unit test (`testPutAndMergeWithRelationType`) to verify that the fix works as expected.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -303,8 +303,6 @@ public class CaffeineEntityCache extends BaseEntityCache {
    * @param newEntities The new entities to sync to the cache.
    */
   private void syncEntitiesToCache(EntityCacheRelationKey key, List<Entity> newEntities) {
-    if (key.relationType() != null) return;
-
     List<Entity> existingEntities = cacheData.getIfPresent(key);
 
     if (existingEntities != null && key.relationType() != null) {
--- a/core/src/test/java/org/apache/gravitino/cache/TestCaffeineEntityCache.java
+++ b/core/src/test/java/org/apache/gravitino/cache/TestCaffeineEntityCache.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.cache;
+
+import static org.apache.gravitino.utils.TestUtil.getTestGroupEntity;
+import static org.apache.gravitino.utils.TestUtil.getTestRoleEntity;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.Optional;
+import org.apache.gravitino.Config;
+import org.apache.gravitino.Entity;
+import org.apache.gravitino.SupportsRelationOperations;
+import org.apache.gravitino.meta.GroupEntity;
+import org.apache.gravitino.meta.RoleEntity;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class TestCaffeineEntityCache {
+
+  private static final GroupEntity groupEntity1 =
+      getTestGroupEntity(1L, "group1", "metalake1", ImmutableList.of("role1"));
+  private static final GroupEntity groupEntity2 =
+      getTestGroupEntity(2L, "group2", "metalake1", ImmutableList.of("role1"));
+  private static final RoleEntity roleEntity = getTestRoleEntity(1L, "role1", "metalake1");
+
+  @Test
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -303,8 +303,6 @@ protected void invalidateExpiredItem(EntityCacheKey key) {
    * @param newEntities The new entities to sync to the cache.
    */
   private void syncEntitiesToCache(EntityCacheRelationKey key, List<Entity> newEntities) {
-    if (key.relationType() != null) return;
```
yuqi1129, line 306: So, is it by design that we will not cache the relation data?

yuqi1129: It's supposed to be a deliberate design. @xunliu can you help to verify it?

keepConcentration: Hi @yuqi1129, thanks for the review.

Looking at PR #8297 where this code was introduced, it seems the original intention was to cache relational data. However, the `if (key.relationType() != null) return;` line appears to prevent this behavior.

yuqi1129: Yeah, that also puzzles me.

jerryshao: I think this is already fixed by @xunliu in #8297, @xunliu @yuqi1129 can you please confirm this?

yuqi1129: Accordingly to the code, if `if (key.relationType() != null) return;` exists, we will not cache all entities associated with the `relation`, and the code looks quite weird.

https://github.com/apache/gravitino/blob/1eadcc251df89f65f7f5d91ac29d5c58091daf8e/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java#L305-L324

yuqi1129: I can assure you that this is specifically designed in the current code base. If we enable caching data for relational operations, something will be wrong with the current cache layer.

yuqi1129:
@keepConcentration
It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue.

keepConcentration:
> @keepConcentration
> 
> It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue. 

Thank you for letting me know. I will review the issue and provide a fix within this week.

keepConcentration:
> @keepConcentration It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue.

Hi, @yuqi1129
I ran the tests locally and they all pass. Could you share which specific tests are expected to fail in CI? Understanding the failure cases would help me address the root cause.

Thank you.

yuqi1129:
> > @keepConcentration It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue.
> 
> Hi, @yuqi1129 I ran the tests locally and they all pass. Could you share which specific tests are expected to fail in CI? Understanding the failure cases would help me address the root cause.
> 
> Thank you.

Please closely watch the failure test cases in the CI.

keepConcentration:
> > > @keepConcentration It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue.
> > 
> > 
> > Hi, @yuqi1129 I ran the tests locally and they all pass. Could you share which specific tests are expected to fail in CI? Understanding the failure cases would help me address the root cause.
> > Thank you.
> 
> Please closely watch the failure test cases in the CI.

It seems that intentionally not caching relational data is a design choice. Based on PR [#8297](https://github.com/apache/gravitino/pull/8297), specifically the commit fba3b80 (Disable cache multi-layer relation entity), this behavior was deliberately added.

yuqi1129:
> > > > @keepConcentration It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue.
> > > 
> > > 
> > > Hi, @yuqi1129 I ran the tests locally and they all pass. Could you share which specific tests are expected to fail in CI? Understanding the failure cases would help me address the root cause.
> > > Thank you.
> > 
> > 
> > Please closely watch the failure test cases in the CI.
> 
> It seems that intentionally not caching relational data is a design choice. Based on PR [#8297](https://github.com/apache/gravitino/pull/8297), specifically the commit [fba3b80](https://github.com/apache/gravitino/commit/fba3b80be1df177c2262e2210f6927b416ce6e91) (Disable cache multi-layer relation entity), this behavior was deliberately added.

I think so, and we can close this PR if it's hard to move on.

keepConcentration:
> > > > > @keepConcentration It's expected that CI will fail after your change. Do you happen to have time to try to fix the failure? Just to let you know, this is not a major issue.
> > > > 
> > > > 
> > > > Hi, @yuqi1129 I ran the tests locally and they all pass. Could you share which specific tests are expected to fail in CI? Understanding the failure cases would help me address the root cause.
> > > > Thank you.
> > > 
> > > 
> > > Please closely watch the failure test cases in the CI.
> > 
> > 
> > It seems that intentionally not caching relational data is a design choice. Based on PR [#8297](https://github.com/apache/gravitino/pull/8297), specifically the commit [fba3b80](https://github.com/apache/gravitino/commit/fba3b80be1df177c2262e2210f6927b416ce6e91) (Disable cache multi-layer relation entity), this behavior was deliberately added.
> 
> I think so, and we can close this PR if it's hard to move on.

Thanks for confirming.

Issue #8697, opened 2025-09-26
[Improvement] The value of  entity weight is not well designed in `EntityCacheWeigher


yuqi1129:
### What would you like to be improved?

As we know, entities with a higher weight in caffeine are more likely to be evicted. In Gravitino,  entity `metalake` are more important that entity `catalog`, however, we assign a higher weight to `metalake` entity than that of `catalog`. I my opinion, we need to assign metalake entities `0`, which means they will never be evicted.

PR #8698, opened 2025-09-26
[#8697] improvement(core): Adjust weight value for entities with type `metalake` and `catalog` in Caffeine cache weighter.

yuqi1129:

Adjust the weight value for the entity `metalake` and `catalog` to prevent it from being evicted easily.  Entities with a higher weight value are more likely to be evicted, so we need to make it smaller for entities with types `metalake` and `catalog`.

Metalakes and catalogs are frequently accessed entities that can always keep them in memory.

Fix: #8697

N/A

UTs

### DIFF
--- a/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
+++ b/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.cache;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import java.time.Duration;
 import java.util.List;
 import org.apache.gravitino.Catalog;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
```diff
@@ -33,10 +45,81 @@ void testDefaultCacheConfig() {
...
+              .withId((long) ((i + 1) * 1000))
+              .build();
+
+      cache.put(
+          EntityCacheRelationKey.of(
+              NameIdentifier.of("metalake1.catalog1.schema" + i), Entity.EntityType.SCHEMA),
+          List.of(schemaEntity));
+    }
+
+    Thread.sleep(100);
```
jerryshao: Using awaitibility, please avoid using `sleep` directly.

jerryshao (approved):

PR #8702, opened 2025-09-26
[#8659] fix(core): Fix parameter mismatch in UserRoleRelBaseSQLProvider and Test

KayMas2808:

This PR renames the `userRoleRelList` parameter to `userRoleRels` in the `batchInsertUserRoleRel` method of `UserRoleRelBaseSQLProvider.java`. This change applies to both the Java method parameter and its `@Param` annotation to match the collection name used within the MyBatis SQL template.

It also adds a corresonding test to `core/src/test/java/org/apache/gravitino/storage/relational/mapper/provider/base/TestUserRoleRelMapper.java` to prevent similar problems in the future.

The mismatch between the parameter name defined in the Java code (`userRoleRelList`) and the name expected by the SQL template (`userRoleRels`) would cause a runtime error when the batch insert operation is called.

This fix resolves the bug and also makes the method signature consistent with other methods in the same file, such as `batchInsertUserRoleRelOnDuplicateKeyUpdate`.

Fix: #8659

No. This is an internal fix to the storage layer and does not alter any user-facing APIs or behavior.

The fix was verified by running the complete test suite for the `:core` module. All existing tests continue to pass.

### DIFF
--- a/flink-connector/flink/src/test/java/org/apache/gravitino/flink/connector/integration/test/iceberg/FlinkIcebergRestCatalogIT.java
+++ b/flink-connector/flink/src/test/java/org/apache/gravitino/flink/connector/integration/test/iceberg/FlinkIcebergRestCatalogIT.java
@@ -28,6 +28,7 @@ import org.apache.gravitino.flink.connector.iceberg.IcebergPropertiesConstants;
 import org.apache.gravitino.flink.connector.integration.test.utils.TestUtils;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIf;
 
 @Tag("gravitino-docker-test")
@@ -48,6 +49,7 @@ public class FlinkIcebergRestCatalogIT extends FlinkIcebergCatalogIT {
   }
 
   @Override
+  @Test
   public void testListSchema() {
     doWithCatalog(
         currentCatalog(),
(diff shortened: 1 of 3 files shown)


jerryshao:
Hi @KayMas2808 where did you fix the issue? I can only find the added test class.

KayMas2808:
> Hi @KayMas2808 where did you fix the issue? I can only find the added test class.

Hi @jerryshao, thanks for pointing that out, sorry, i'd messed up the branches so i had to redo it and made a mistake in the process. i've added the change to the commit.

yuqi1129:
@KayMas2808 
Hi, is this PR still active?

jerryshao:
@yuqi1129 I think You can continue working on this.

yuqi1129:
> @yuqi1129 I think You can continue working on this.

I see.

flink-connector/flink/src/test/java/org/apache/gravitino/flink/connector/integration/test/iceberg/FlinkIcebergRestCatalogIT.java
```diff
@@ -48,6 +49,7 @@ protected Map<String, String> getCatalogConfigs() {
   }
 
   @Override
+  @Test
```
yuqi1129, line 52: This aims to fix a minor bug by the way.

KayMas2808:
Hey @yuqi1129 @jerryshao, I thought I'd made the required changes, and added tests, could you guys please tell me what I missed so it doesn't happen again?

PR #8703, opened 2025-09-26
Fix concurrency issue in AuthorizationRequestContext.loadRole

bittu9975:
This PR resolves a race condition in AuthorizationRequestContext.loadRole, where multiple threads could execute the role initializer concurrently.
The fix introduces a FutureTask guarded by an AtomicReference, ensuring the initializer runs exactly once, even under concurrent access.

A focused unit test (TestAuthorizationRequestContext) is added to verify that loadRole executes the initializer only once when called from multiple threads.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
+++ b/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
@@ -23,6 +23,9 @@ import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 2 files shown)


roryqi:
This `AuthorizationRequestContext ` won't be accessed in the multiple threads according to my understanding.

bittu9975:
@jerqi 
You are right that, from the current usage, AuthorizationRequestContext would not be accessed by more than one thread.

My motivation for doing this change was to make loadRole() more future-proof and robust against future race conditions if the usage pattern changes (e.g., multi-threaded server environments or future functionality). The extra concurrency guard loads the role only once without incurring heavy overhead.

That being the case, if the maintainers prefer this extra complexity isn't required at the moment, I'd be content to just make it simpler.
Would you like me to:

Revert to the less complicated AtomicBoolean.compareAndSet guard, or

Maintain the present concurrency-safe solution for future security, perhaps with better documentation, or

Do something entirely different?

roryqi:
Context is usually used for provide the variables in the one thread. The context can be accessed by multiple thread. It is a little weird for me.

Issue #8639, opened 2025-09-22
[Subtask] Add the core implementation for job template alteration


jerryshao:
### Describe the subtask

This issue tracks the work of adding the core implementation for job template alteration.

### Parent issue

#8429

PR #8707, opened 2025-09-28
[#8639]feat(core): Add the job template alteration in the core part

jerryshao:

Add the core part for job template alteration.

This is a part of work for job template alteration.

Fix: #8639

No.

New UTs to test code.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/JobManager.java
+++ b/core/src/main/java/org/apache/gravitino/job/JobManager.java
@@ -329,7 +329,9 @@ public class JobManager implements JobOperationDispatcher {
                     updateJobTemplateEntity(jobTemplateIdent, jobTemplateEntity, changes));
           } catch (NoSuchEntityException e) {
             throw new NoSuchJobTemplateException(
-                "Job template with name %s under metalake %s does not exist",
+                "Job template with name %s under metalake %s does not exist, this could be due to"
+                    + " the job template not existing or updated concurrently. For the latter case"
+                    + " please retry the operation.",
                 jobTemplateName, metalake);
           } catch (IOException ioe) {
             throw new RuntimeException(ioe);


core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -300,6 +300,40 @@ public boolean deleteJobTemplate(String metalake, String jobTemplateName) throws
...
+          } catch (NoSuchEntityException e) {
+            throw new NoSuchJobTemplateException(
+                "Job template with name %s under metalake %s does not exist",
+                jobTemplateName, metalake);
+          } catch (IOException ioe) {
+            throw new RuntimeException(ioe);
+          } catch (EntityAlreadyExistsException e) {
+            throw new RuntimeException(
+                "Job template with the same name "
+                    + jobTemplateName
```
yuqi1129: The message here is incorrect. The name should be the value from `updateJobTemplateEntity(jobTemplateIdent, jobTemplateEntity, changes)`.

yuqi1129: And the message `because job template doesn't support rename` is quite misleading.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -300,6 +301,49 @@ public boolean deleteJobTemplate(String metalake, String jobTemplateName) throws
...
   }
 
+  @Override
+  public JobTemplateEntity alterJobTemplate(
+      String metalake, String jobTemplateName, JobTemplateChange... changes)
+      throws NoSuchJobTemplateException, IllegalArgumentException {
+    checkMetalake(NameIdentifierUtil.ofMetalake(metalake), entityStore);
+
+    Optional<String> newName =
+        Arrays.stream(changes)
```
yuqi1129, line 311: Is there always only one `RenameJobTemplate`? I'm afraid we cannot make such an assumption.

jerryshao: I didn't do such check for job template, also for others. We assume that in one request there will only be one "rename".

yuqi1129: I see.

jerryshao: Shall we add such enforcement?

yuqi1129: There should be no limit on the number of rename changes in a single request, and we will use the last one as the final result.  If this scenario did not occur in production,  we could ignore it.

jerryshao: I can pick the last one instead.

jerryshao:
@mchades @FANNG1 can you please help to review this PR?

yuqi1129:
I have no further comments on this PR from LGTM.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -300,6 +301,49 @@ public boolean deleteJobTemplate(String metalake, String jobTemplateName) throws
...
+            .filter(c -> c instanceof JobTemplateChange.RenameJobTemplate)
+            .map(c -> ((JobTemplateChange.RenameJobTemplate) c).getNewName())
+            .reduce((first, second) -> second);
+
+    NameIdentifier jobTemplateIdent = NameIdentifierUtil.ofJobTemplate(metalake, jobTemplateName);
+    return TreeLockUtils.doWithTreeLock(
+        jobTemplateIdent,
+        LockType.WRITE,
```
mchades: Should we use a read lock only when there are no rename operations?

jerryshao: OK, I will fix it.

Issue #8710, opened 2025-09-28
[FEATURE] Cache entities in relation operations


yuqi1129:
### Describe the feature

Currently, entities in relation operations are not cached, which will decrease performance, so we need to add support for it.

PR #8712, opened 2025-09-28
[#8710] feat(core): Support cache entities in relation operations.

yuqi1129:

Cache entities in relation operations.

It's a big improvement.

Fix: #8710

N/A

UTs and ITs.

### DIFF
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -83,3 +83,7 @@ jobs:
       - name: Run chart-testing (install mysql enable)
         if: steps.list-changed.outputs.changed == 'true'
         run: ct install --charts=dev/charts/gravitino --helm-extra-set-args "--values dev/charts/gravitino/resources/scenarios/ci-values.yaml" --target-branch ${{ github.event.pull_request.base.ref }}
+
+      - name: Run chart-testing (install postgresql enable)
+        if: steps.list-changed.outputs.changed == 'true'
+        run: ct install --charts=dev/charts/gravitino --helm-extra-set-args "--values dev/charts/gravitino/resources/scenarios/pg-values.yaml" --target-branch ${{ github.event.pull_request.base.ref }}
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -26,6 +26,8 @@ dependencies {
   implementation(libs.commons.lang3)
   implementation(libs.commons.collections4)
   implementation(libs.guava)
+  implementation(libs.jackson.annotations)
+  implementation(libs.jackson.databind)
 
   testImplementation(libs.junit.jupiter.api)
   testImplementation(libs.junit.jupiter.params)
--- a/api/src/main/java/org/apache/gravitino/MetadataObject.java
+++ b/api/src/main/java/org/apache/gravitino/MetadataObject.java
@@ -63,7 +63,12 @@ public interface MetadataObject {
     /** A model is mapped to the model artifact in ML. */
     MODEL,
     /** A tag is used to help manage other metadata object. */
-    TAG;
+    TAG,
+    /**
+     * A policy can be associated with a metadata object for data governance and similar purposes.
+     */
+    POLICY,
+    ;
   }
 
   /**
--- a/api/src/main/java/org/apache/gravitino/MetadataObjects.java
+++ b/api/src/main/java/org/apache/gravitino/MetadataObjects.java
@@ -74,7 +74,8 @@ public class MetadataObjects {
             || type == MetadataObject.Type.CATALOG
             || type == MetadataObject.Type.METALAKE
             || type == MetadataObject.Type.ROLE
-            || type == MetadataObject.Type.TAG,
+            || type == MetadataObject.Type.TAG
+            || type == MetadataObject.Type.POLICY,
         "If the length of names is 1, it must be the CATALOG, METALAKE,TAG, or ROLE type");
 
     Preconditions.checkArgument(
--- a/api/src/main/java/org/apache/gravitino/authorization/Privilege.java
+++ b/api/src/main/java/org/apache/gravitino/authorization/Privilege.java
@@ -98,7 +98,12 @@ public interface Privilege {
     /** The privilege to create a tag */
     CREATE_TAG(0L, 1L << 21),
     /** The privilege to apply a tag */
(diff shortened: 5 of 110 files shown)


yuqi1129:
More effort should be taken to support `get` and `list` operations for relation operations at the same time, and the current change only supports list operations.

yuqi1129:
@Abyss-lord do you have time to take a look?

Abyss-lord:
> @Abyss-lord do you have time to take a look?

sure, plz assign to me.

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -90,6 +91,12 @@ public class CaffeineEntityCache extends BaseEntityCache {
 
   private ScheduledExecutorService scheduler;
 
+  Set<SupportsRelationOperations.Type> relationTypes =
```
roryqi: Could you reuse the method `values()` of `enum` here?

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -303,8 +310,6 @@ protected void invalidateExpiredItem(EntityCacheKey key) {
    * @param newEntities The new entities to sync to the cache.
    */
   private void syncEntitiesToCache(EntityCacheRelationKey key, List<Entity> newEntities) {
-    if (key.relationType() != null) return;
-
```
jerryshao, line 307: Can you please explain why do you remove this?

jerryshao:
The main thing is that caching the relations will significantly increase the memory usage, we should 1. make sure that policy/tags should be kicked out first if the threshold is reached; 2. something like statistics should not be cached, this will waste lots of memory.

jerryshao:
I don't see any UTs about this PR, can you add some tests about it?

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
yuqi1129: If `if (key.relationType() != null) return;` exists here, entity store cache will not cache entities with relation, that is why the cache does not support store relational operations.

Note: `key.relationType()` is not null when we are going to put relational entities into cache.

yuqi1129:
> The main thing is that caching the relations will significantly increase the memory usage, we should 1. make sure that policy/tags should be kicked out first if the threshold is reached; 2. something like statistics should not be cached, this will waste lots of memory.

> 1. make sure that policy/tags should be kicked out first if the threshold is reached;

That makes sense. I will open an issue to track it. 


> 2. something like statistics should not be cached, this will waste lots of memory.

I have opened an issue aimed at it. Please see https://github.com/apache/gravitino/issues/8785

yuqi1129:
> I don't see any UTs about this PR, can you add some tests about it?

done.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -19,24 +19,29 @@
...
+import org.apache.gravitino.meta.TagEntity;
 import org.apache.gravitino.meta.UserEntity;
 
 /**
  * Reverse index cache for managing entity relationships. This cache uses a radix tree to
  * efficiently store and retrieve relationships between entities based on their keys.
  */
 public class ReverseIndexCache {
-  private RadixTree<EntityCacheKey> reverseIndex;
+  private RadixTree<List<EntityCacheKey>> reverseIndex;
```
yuqi1129, line 44: This change aims to fix https://github.com/apache/gravitino/issues/8817.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexRules.java
```diff
@@ -142,4 +147,40 @@ public class ReverseIndexRules {
...
+  // Keep policies/tags to objects reverse index for metadata objects, so the key are objects and
+  // the values are policies/tags.
+  public static final ReverseIndexCache.ReverseIndexRule GENERIC_METADATA_OBJECT_REVERSE_RULE =
+      (entity, key, reverseIndexCache) -> {
+        // Name in GenericEntity contains no metalake.
+        GenericEntity genericEntity = (GenericEntity) entity;
+        EntityType type = entity.type();
+        if (genericEntity.name() != null) {
+          String[] levels = genericEntity.name().split("\\.");
+          String metadataName = key.identifier().namespace().levels()[0];
```
jerryshao: "metalakeName" or "metadataName"?

core/src/main/java/org/apache/gravitino/cache/ReverseIndexRules.java
```diff
@@ -142,4 +147,40 @@ public class ReverseIndexRules {
...
+          NameIdentifier objectsNameIdentifier =
```
jerryshao: "objects" or "object"?

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -222,10 +282,12 @@ public <E extends Entity & HasIdentifier> void put(
...
-          // We still need to cache the entities even if the list is empty, to avoid cache
-          // misses. Consider the scenario where a user queries for an entity's relations and the
-          // result is empty. If we don't cache this empty result, the next query will still hit the
-          // backend, this is not desired.
+          // Return directly if entities are empty. No need to put an empty list to cache, we will
+          // use another PR to resolve the performance problem.
+          if (entities.isEmpty()) {
+            return;
+          }
+
```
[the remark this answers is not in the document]
yuqi1129: It's a need, or CI will fail.

yuqi1129:
All have been resolved.

jerryshao:
I'm OK with the current change. Please fix the above issue.

Besides, my feeling is that this code is quite complex and hard to understand by others. We'd better add more comments as possible as we can, so others can hand over this part.

yuqi1129:
> my feeling is that this code is quite complex and hard to understand by others.

I think so, it introduces radix tree,  segment lock, caffeine cache, and maintains both key/value data and relational data. I will update the comment in the next few improvements, it should be raised in the release 1.1.0.

yuqi1129:
All resolved.

Issue #8704, opened 2025-09-26
[Bug report] It seems that Docker Image 1.0.0 might have an issue


macpie:
### Version

main branch

### Describe what's wrong

I have been playing with image `apache/gravitino-iceberg-rest:1.0.0` locally (Mac OS M1) with no issue.
When I tried to run the same setup on my Ubuntu server, the iceberg rest server would not start. 
Going back to 0.9.1 fixed the issue.

### Error message and/or stacktrace

```
cat /root/gravitino-iceberg-rest-server/logs/gravitino-iceberg-rest-server.out
2025-09-26T19:17:53.846005134Z main ERROR Unable to create Lookup for ctx java.lang.NoSuchMethodError: 'java.util.stream.Stream org.apache.logging.log4j.util.ServiceLoaderUtil.loadServices(java.lang.Class, java.lang.invoke.MethodHandles$Lookup, boolean)'
	at org.apache.logging.log4j.core.impl.ThreadContextDataInjector.getServiceProviders(ThreadContextDataInjector.java:77)
	at org.apache.logging.log4j.core.impl.ThreadContextDataInjector.<clinit>(ThreadContextDataInjector.java:64)
	at org.apache.logging.log4j.core.impl.ThreadContextDataInjector$ForDefaultThreadContextMap.<init>(ThreadContextDataInjector.java:93)
	at org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createDefaultInjector(ContextDataInjectorFactory.java:91)
	at org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createInjector(ContextDataInjectorFactory.java:71)
	at org.apache.logging.log4j.core.lookup.ContextMapLookup.<init>(ContextMapLookup.java:34)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at org.apache.logging.log4j.core.util.ReflectionUtil.instantiate(ReflectionUtil.java:189)
[...]

PR #6379, opened 2025-01-27
[6378] improvement: bump log4j dependency from 2.22.0 to 2.24.3 to fix a performance regression

puchengy:
### What changes were proposed in this pull request?

bump log4j dependency from 2.22.0 to 2.24.3 to fix a performance regression 

### Why are the changes needed?

(Please clarify why the changes are needed. For instance,
  1. If you propose a new API, clarify the use case for a new API.
  2. If you fix a bug, describe the bug.)

Fix: https://github.com/apache/gravitino/issues/6378

### Does this PR introduce _any_ user-facing change?

(Please list the user-facing changes introduced by your change, including
  1. Change in user-facing APIs.
  2. Addition or removal of property keys.)

No.

### How was this patch tested?

(Please test your changes, and provide instructions on how to test it:
  1. If you add a feature or fix a bug, add a test to cover your changes.
  2. If you fix a flaky test, repeat it for many times to prove it works.)

Running in internal workload.

PR #8719, opened 2025-09-29
[SIP]  downgrade Log4J version to keep consistent with Iceberg AWS bundle

FANNG1:

downgrade Log4J version to keep consistent with log4j version in Iceberg AWS bundle to avoid class conflict

Fix: #8704 

no


jerryshao:
The change is introduced in https://github.com/apache/gravitino/pull/6379/files, why didn't we meet this problem in the previous version?

jerryshao:
Also, will we introduce any side effects after downgrading the version?

FANNG1:
> The change is introduced in https://github.com/apache/gravitino/pull/6379/files, why didn't we meet this problem in the previous version?

Iceberg aws bundle jar doesn't contain log4j before Iceberg 1.9

yuqi1129:
> Also, will we introduce any side effects after downgrading the version?

Based on #6379 and the related issue,  it seems unlikely that the revertion will introduce any side effects, as I can't reproduce the problem mentioned in the issue using `2.22.0`.

image

This is the result that was tested in the local environment.

FANNG1:
> Also, will we introduce any side effects after downgrading the version?

There may be some performance regression. Some other options:
1. downgrade Iceberg version from 1.9 to 1.8
2. keep using Iceberg 1.9 for Gravitino, but use Iceberg AWS bundle 1.8 in Docker, I verified with Spark SQL
3. remove log4j jars from Gravitino in Docker images, and add document to tell user how to bypass this probem.

WDYT?

jerryshao:
> > Also, will we introduce any side effects after downgrading the version?
> 
> There may be some performance regression. Some other options:
> 
> 1. downgrade Iceberg version from 1.9 to 1.8
> 2. keep using Iceberg 1.9 for Gravitino, but use Iceberg AWS bundle 1.8 in Docker, I verified with Spark SQL
> 3. remove log4j jars from Gravitino in Docker images, and add document to tell user how to bypass this probem.
> 
> WDYT?

I would suggest using option 3 as a hotfix to update the image and document. We can have a better solution for version 1.0.1.

Issue #8640, opened 2025-09-22
[Subtask] Add the server side rest interface for job template alteration


jerryshao:
### Describe the subtask

This issue tracks the work of adding the server-side REST interface for job template alteration

### Parent issue

#8429

PR #8724, opened 2025-09-30
[#8640] feat(server): Add server-side REST interface for job template alteration

jerryshao:

This PR adds the server-side REST interface for job template alteration.

This is a part of work to support job template alteration.

Fix: #8640 

No.

UTs added.

### DIFF
--- a/common/src/main/java/org/apache/gravitino/dto/job/SparkTemplateUpdateDTO.java
+++ b/common/src/main/java/org/apache/gravitino/dto/job/SparkTemplateUpdateDTO.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.dto.job;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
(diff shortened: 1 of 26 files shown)


common/src/main/java/org/apache/gravitino/dto/requests/JobTemplateUpdateRequest.java
```diff
@@ -0,0 +1,163 @@
...
+     * Constructor for RenameJobTemplateRequest.
+     *
+     * @param newName the new name for the job template
+     */
+    public RenameJobTemplateRequest(String newName) {
+      this.newName = newName;
+    }
+
+    /** Default constructor for Jackson. */
+    public RenameJobTemplateRequest() {
```
mchades: Should this be private?

common/src/main/java/org/apache/gravitino/dto/requests/JobTemplateUpdateRequest.java
```diff
@@ -0,0 +1,163 @@
...
+     * Constructor for UpdateJobTemplateCommentRequest.
+     *
+     * @param newComment the new comment for the job template
+     */
+    public UpdateJobTemplateCommentRequest(String newComment) {
+      this.newComment = newComment;
+    }
+
+    /** Default constructor for Jackson. */
+    public UpdateJobTemplateCommentRequest() {
```
mchades: Should this be private?

common/src/main/java/org/apache/gravitino/dto/job/SparkTemplateUpdateDTO.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Nullable
+  @JsonProperty("newFiles")
+  private List<String> newFiles;
+
+  @Nullable
+  @JsonProperty("newArchives")
+  private List<String> newArchives;
+
+  @Nullable
```
mchades, line 54: What if all fields are null?

common/src/main/java/org/apache/gravitino/dto/job/TemplateUpdateDTO.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Nullable
+  @JsonProperty("newCustomFields")
+  private Map<String, String> newCustomFields;
+
+  /**
+   * The default constructor for Jackson. This constructor is required for deserialization of the
+   * DTO.
+   */
+  protected TemplateUpdateDTO() {
```
mchades, line 67: why not be private?

common/src/main/java/org/apache/gravitino/dto/job/SparkTemplateUpdateDTO.java
jerryshao: If all the fields are null, which means nothing needs to be altered.

common/src/main/java/org/apache/gravitino/dto/job/TemplateUpdateDTO.java
jerryshao: This will be called by the child class.

PR #8736, opened 2025-09-30
[MINOR] Remove unnecessary dependencies

jerryshao:

This PR removes some unnecessary introduced dependencies and packages to avoid some potential conflicts.

For fileset catalog, we don't have to add the hdfs dependencies to the runtime, hadoop-client is enough.

No.

Local and IT verification.

### DIFF
--- a/catalogs/catalog-fileset/build.gradle.kts
+++ b/catalogs/catalog-fileset/build.gradle.kts
@@ -46,21 +46,6 @@ dependencies {
   implementation(libs.commons.io)
   implementation(libs.hadoop3.client.api)
   implementation(libs.hadoop3.client.runtime)
-//  implementation(libs.hadoop3.hdfs) {
-//    exclude("com.sun.jersey")
-//    exclude("javax.servlet", "servlet-api")
-//    exclude("com.google.guava", "guava")
-//    exclude("commons-io", "commons-io")
-//    exclude("org.eclipse.jetty", "*")
-//    exclude("io.netty")
-//    exclude("org.fusesource.leveldbjni")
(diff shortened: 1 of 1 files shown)


yuqi1129:
@jerryshao 
I haven't found the exact cause, but adding the following codes works:

image

jerryshao:
> @jerryshao I haven't found the exact cause, but adding the following codes works:
> 
> image

Thanks @yuqi1129 , let me also verify it locally.

jerryshao:
@yuqi1129 I have found the root cause:

If `dfs.namenode.kerberos.principal.pattern` is not configured, the client needs to figure out the namenode principal from the configuration; if we don't configure `dfs.namenode.kerberos.principal`, then the connection will fail. If `dfs.namenode.kerberos.principal.pattern` is configured, then the Hadopo client will figure out the principal by itself.

If hdfs dependency is in the classpath, then there will be a default `dfs.namenode.kerberos.principal.pattern` value, so the connection will succeed. If not, then there will be no default value, so the connection will fail unless we explicitly add the `dfs.namenode.kerberos.principal` configuration like you did before.

yuqi1129:
> @yuqi1129 I have found the root cause:
> 
> If `dfs.namenode.kerberos.principal.pattern` is not configured, the client needs to figure out the namenode principal from the configuration; if we don't configure `dfs.namenode.kerberos.principal`, then the connection will fail. If `dfs.namenode.kerberos.principal.pattern` is configured, then the Hadopo client will figure out the principal by itself.
> 
> If hdfs dependency is in the classpath, then there will be a default `dfs.namenode.kerberos.principal.pattern` value, so the connection will succeed. If not, then there will be no default value, so the connection will fail unless we explicitly add the `dfs.namenode.kerberos.principal` configuration like you did before.

I see and learn.

Issue #8656, opened 2025-09-23
[Improvement] In TableUpdatesRequest.java validate updates


justinmclean:
### What would you like to be improved?

In common/src/main/java/org/apache/gravitino/dto/requests/TableUpdatesRequest.java, in the validate method add a precondition to check if updates are valid (i.e not empty)

### How should we improve?

See above, also add a unit test.

playasim:
Hi @justinmclean , I would like to work on this one.

A0R0P0I7T:
Hi @justinmclean , I would like to work on this issue if you would allow me too.

justinmclean:
@A0R0P0I7T if you still want to work on this just confirm

A0R0P0I7T:
@justinmclean Yes, I would love to work on this.

A0R0P0I7T:
Hi @justinmclean ,
I have implemented the requested validation to ensure that the updates list in TableUpdatesRequest is not empty and added a corresponding unit test.
A PR has been created for this change: #10201 
Please let me know if any further changes are required.

PR #8738, opened 2025-09-30
[#8656] Improvement: add UTs and fix a NPE in TableUpdatesRequest.java 

playasim:

Add UTs for TableUpdatesRequest.java validate method.
Also There are 2 validate methods throw NPE when arguments is invalid.

Add UTs to make sure validate logic.

Fix: #8656

NO

Passed all UTs

### DIFF
--- a/common/src/main/java/org/apache/gravitino/dto/requests/TableUpdateRequest.java
+++ b/common/src/main/java/org/apache/gravitino/dto/requests/TableUpdateRequest.java
@@ -815,7 +815,7 @@ public interface TableUpdateRequest extends RESTRequest {
      */
     @Override
     public void validate() throws IllegalArgumentException {
-      Preconditions.checkNotNull(index, "Index cannot be null");
+      Preconditions.checkArgument(index != null, "Index cannot be null");
       Preconditions.checkArgument(index.type() != null, "Index type cannot be null");
       Preconditions.checkArgument(
           index.fieldNames() != null && index.fieldNames().length > 0,
@@ -861,7 +861,7 @@ public interface TableUpdateRequest extends RESTRequest {
      */
     @Override
     public void validate() throws IllegalArgumentException {
-      Preconditions.checkNotNull(name, "Index name cannot be null");
+      Preconditions.checkArgument(name != null, "Index name cannot be null");
     }
 
     /** @return An instance of TableChange. */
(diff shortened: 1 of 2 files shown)


justinmclean:
Thanks for the unit tests and changes, however, changing checkNotNull to checkArgument wasn't what I had in mind.

To be more explicit:
For RenameTableRequest - If a new schema name is provided, it must not be blank. (Currently allowed to be "".)
For AddTableIndexRequest - Validate index.name() is not blank (currently only type and fieldNames are checked). You might also want to validate that inner fieldNames arrays contain only non-blank field path segments.
For DeleteTableIndexRequest - validate() only checks name != null. Should be not blank to avoid empty string.

justinmclean:
While making those changes, you may want to update the JavaDoc for UpdateTableColumnNullabilityRequest and ensure the messaging is more consistent, i.e., some messages state “cannot be empty” for null checks, while others use “cannot be null”.

yuqi1129:
@playasim 
Can you please change all `checkNotNull` in Gravitino to `checkArgument` in this PR?

playasim:
> @playasim Can you please change all `checkNotNull` in Gravitino to `checkArgument` in this PR?

Sure. I will push the new commit in next coming days.

playasim:
Hi @justinmclean ,added more UTs and enhanced on the request mentioned above.(I've checked RenameTableRequest and the validate method seems fine) Please help review the PR.

@yuqi1129 , in this PR I only changed `checkNotNull ` to `checkArgument ` in `TableUpdateRequest.java`. Would you like me to do a full project check? If so we may need a new PR to do the replacement.

yuqi1129:
> @yuqi1129 , in this PR I only changed `checkNotNull ` to `checkArgument ` in `TableUpdateRequest.java`. Would you like me to do a full project check? If so we may need a new PR to do the replacement.

OK. If raising such a large PR is a challenge for you, you can use another PR to resolve it.

keepConcentration:
Hi all,
I noticed that this PR modifies `TableUpdateRequest.java`, but the original issue (#8656) requested adding validation specifically for the `updates` field in `TableUpdatesRequest.java`. Could anyone clarify whether the `updates` validation has been implemented in `TableUpdatesRequest.java`, or if this PR was intended to address a different class? I want to make sure the original issue’s request is fully addressed.

Issue #8741, opened 2025-09-30
[Improvement] Class loader can't be unloaded if use filesets with local filesystem


yuqi1129:
### What would you like to be improved?

<img width="1539" height="662" alt="Image" src="https://github.com/user-attachments/assets/d075a0d7-db97-4aed-95a9-198fbdc8b643" />

When we repeatedly modify a fileset catalog on a local filesystem, then Gravitino will finally run out of memory and leads to OOM, The heapdump can be seen above.  

### How should we improve?

Unload the common logging in htrace.

yuqi1129:
It appears that PR #8271 and #8252 do not cover this case.

PR #8742, opened 2025-09-30
[#8741] improvement(catalog-common): Unload common logging in htrace.

yuqi1129:

Unloaded common logging in htrace when catalogs are about to close.

We need to unload it or the class loader that loads it will not be cleaned

Fix: #8741 

N/A.

Test locally.

### DIFF
--- a/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
+++ b/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
@@ -221,6 +221,19 @@ public class ClassLoaderResourceCleanerUtils {
    */
   private static void releaseLogFactoryInCommonLogging(ClassLoader currentClassLoader)
       throws Exception {
+
+    // If we use fileset with the local file system, HTrace will be used, so we need to
+    // release the HTrace LogFactory as well.
+    try {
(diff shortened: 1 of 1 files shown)


catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
```diff
@@ -221,6 +221,19 @@ private static void clearShutdownHooks(ClassLoader targetClassLoader) throws Exc
...
+    // If we use fileset with the local file system, HTrace will be used, so we need to
+    // release the HTrace LogFactory as well.
+    try {
+      Class<?> htraceLogFactoryClass =
+          Class.forName(
+              "org.apache.htrace.shaded.commons.logging.LogFactory", true, currentClassLoader);
+      MethodUtils.invokeStaticMethod(htraceLogFactoryClass, "release", currentClassLoader);
+    } catch (Exception e) {
+      // Ignore if htrace is not used
+      LOG.debug("HTrace is not used, skipping release of HTrace LogFactory...");
```
jerryshao, line 234: Can you please change all the log levels in this file to debug or trace log?

yuqi1129: OK

Issue #8006, opened 2025-08-11
[Improvement] Fix bucket count parsing in DorisUtils


justinmclean:
### What would you like to be improved?

There are some issues with parsing the bucket count in catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java

You can see this by running this test:
```
  @Test
  public void testExtractBucketNumFromSql() {
    String createTableSql =
        "CREATE TABLE `testTable` (\n`col1` int NOT NULL\n) ENGINE=OLAP\n DISTRIBUTED BY HASH(`col1`) BUCKETS 8";
    Distribution distribution = DorisUtils.extractDistributionInfoFromSql(createTableSql);
    assertEquals(8, distribution.number());

    String createTableSqlWithoutBucket =
        "CREATE TABLE `testTable` (\n`col1` int NOT NULL\n) ENGINE=OLAP\n DISTRIBUTED BY HASH(`col1`)";
    Distribution distributionDefault =
        DorisUtils.extractDistributionInfoFromSql(createTableSqlWithoutBucket);
    assertEquals(1, distributionDefault.number());
  }
```

### How should we improve?

Fix the code so the test passes.

saumyapandey1998:
Hi @justinmclean! I would like to work on this issue

justinmclean:
Hi @saumyapandey1998, sorry I must have missed you asking to work on this, sure, go ahead.

saumyapandey1998:
Thank you @justinmclean ! I will start working on this

pandeysambhi:
@justinmclean This seems to be abandoned ! should I go ahead and make the changes ?

saumyapandey1998:
Hi @pandeysambhi ! Please go ahead, uni course load got the best of me :(

justinmclean:
this has been fixed

PR #8744, opened 2025-09-30
[#8006]Improvement(DorisUtils): fix bucket count parsing in DorisUtils

pandeysambhi:

updated the Regex Pattern to capture the bucket details.

- Ensure the regex matches both with and without BUCKETS.
- If the regex matches but BUCKETS is missing, default to 1.

To fix the bucket count parsing in DorisUtils. The current regex expects BUCKETS to always be present, but in the absence of BUCKETS, the code should default to 1.
The code already defaults to 1 if [matcher.group(5)] is null, but the regex may not always match if BUCKETS is missing, causing the method to throw an exception.

Fix: (#8006)

No

unit testing


justinmclean (approved):
Thanks looks good to me.

Issue #8748, opened 2025-10-02
[Improvement] replace with proper exec run of the server process


shunki-fujita:
### What would you like to be improved?

In the Docker setup / scripts we currently use:
https://github.com/apache/gravitino/blob/caee8d20f5129dc22b7b3e7290091f619f590cbc/dev/docker/iceberg-rest-server/iceberg-rest-server-dependency.sh#L77-L82

This is only a dummy way to keep the container alive, not a correct process management approach. 

### How should we improve?

Replace the dummy tail -f /dev/null with an exec call that directly runs the server in the foreground.
https://github.com/apache/gravitino/blob/caee8d20f5129dc22b7b3e7290091f619f590cbc/dev/docker/iceberg-rest-server/start-iceberg-rest-server.sh#L29

HeathKnowles:
For the Shell command 
Instead of tail -f /dev/null 
```bash
exec "$(dirname "$0")/start-iceberg-rest-server.sh"
```
the above can be replaced

PR #8749, opened 2025-10-02
[#8748] Improvement: replace with proper exec run of the server process

shunki-fujita:

Replace the dummy tail -f /dev/null with an exec call that directly runs the server in the foreground.

Fix: #8748


FANNG1:
LGTM, thanks @shunki-fujita for the contribution!

Issue #8761, opened 2025-10-06
[Improvement] rewrite_config.py does not handle GRAVITINO_ICEBERG_REST_HTTP_PORT


shunki-fujita:
### What would you like to be improved?

The current implementation of [rewrite_config.py](https://github.com/apache/gravitino/blob/68142fa2c1333e1fcbc4a663ebc94d648a980eb6/dev/docker/iceberg-rest-server/rewrite_config.py)
 rewrites several environment variables, but it does not rewrite the HTTP port configuration (`GRAVITINO_ICEBERG_REST_HTTP_PORT`).

As a result, the configured port in `conf/iceberg-rest.yaml` is not updated when the environment variable is set, making it difficult to change the listening port dynamically.

### How should we improve?

`rewrite_config.py` should update the port field in the configuration when `GRAVITINO_ICEBERG_REST_HTTP_PORT` is set.

PR #8762, opened 2025-10-06
[#8761] Improvement: rewrite_config.py does not handle GRAVITINO_ICEBERG_REST_HTTP_PORT

shunki-fujita:
rewrite_config.py should update the port field in the configuration when GRAVITINO_ICEBERG_REST_HTTP_PORT is set.

Because the HTTP port cannot currently be customized via the GRAVITINO_ICEBERG_REST_HTTP_PORT environment variable.

Fix: #8761

no

UTs


FANNG1:
Could you update the document?  https://github.com/apache/gravitino/blob/68142fa2c1333e1fcbc4a663ebc94d648a980eb6/docs/iceberg-rest-service.md?plain=1#L574

shunki-fujita:
@FANNG1 
There’s a “Since version” field in the document.
Could you let me know which version I should specify?

FANNG1:
> @FANNG1 There’s a “Since version” field in the document. Could you let me know which version I should specify?

Since version is the version that include the changes,  `1.1.0` for this.

Issue #8614, opened 2025-09-18
[Subtask] Add Table, TableDTO


tsungchih:
### Describe the subtask

We need to implement the following classes to support table operations.

- [x] Table
- [x] TableDTO

**NOTE** that the methods with name prefix `supports` in class `Table` will not be implemented until we have implemented interfaces `SupportsXXX`.

### Parent issue

#8591

PR #8763, opened 2025-10-06
[#8614] feat(client-python): add table and table dto

tsungchih:

This PR is aimed at implementing classes `Table` and `TableDTO`. Furthermore, a refactor of moving package `expressions` into `gravitino.api.rel` has been done in this PR as well.

We need to support table operations.

Fix: #8614 

No.

Unit tests

### DIFF
--- a/clients/client-python/gravitino/dto/rel/table_dto.py
+++ b/clients/client-python/gravitino/dto/rel/table_dto.py
@@ -103,7 +103,7 @@ class TableDTO(Table, DataClassJsonMixin):  # pylint: disable=R0902
 
     def __post_init__(self):
         Precondition.check_argument(
-            self._name is not None and self._name.strip() != "",
+            self._name is not None and self._name != "",
             "name cannot be null or empty",
         )
         Precondition.check_argument(self._audit is not None, "audit cannot be null")


tsungchih:

clients/client-python/gravitino/dto/rel/table_dto.py
```diff
@@ -0,0 +1,140 @@
...
+        ),
+    )
+    _properties: Optional[dict[str, str]] = field(
+        default=None,
+        metadata=config(field_name="properties", exclude=lambda value: value is None),
+    )
+
+    def __post_init__(self):
+        Precondition.check_argument(
+            self._name is not None and self._name.strip() != "",
```
unknowntpo: Why we need to strip name before checking empty ?

clients/client-python/gravitino/api/rel/table.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def properties(self) -> dict[str, str]:
+        """Gets the properties of the table.
+
+        Returns:
+            dict[str, str]:
+                The properties of the table. Empty dictionary is returned if no properties are set.
+        """
+
+        return {}
```
unknowntpo, line 112: Where do we implement `supportPartitions`, `supportsTags`, `supportsPolicies`, `supportsRoles` ? `supportsStatistics`, `supportsPartitionStatistics` default method ?

tsungchih: We do not implement these `supportsXXX` methods at this stage since we would like to ensure core functionality ot table operations work end-to-end, and leave the implementation of the `SupportsXXX` interface for later. Please refer to [this comment](https://github.com/apache/gravitino/pull/8545#issuecomment-3358770205) and [this comment](to ensure its core functionality works end-to-end, and leave the implementation of the SupportsPolicies interface for later) for more detail.

clients/client-python/gravitino/dto/rel/table_dto.py
tsungchih: @unknowntpo Thanks for the comment. I was thinking to prevent us from getting the string with combination of spaces, tabs, newlines, and carriage returns. Here are some examples, "      ", "    \n". After having called `strip()` on these strings, we will get an empty string.

clients/client-python/gravitino/api/rel/table.py
unknowntpo: Okay, please make sure the information in that comment is well-documented in related issues.

unknowntpo:
@jerryshao LGTM.

clients/client-python/gravitino/api/rel/table.py
tsungchih: Thanks for the reminder. I'm trying to do so.

clients/client-python/gravitino/dto/rel/table_dto.py
unknowntpo: But I think this is user's duty to make sure the table name doesn't only have these characters. It would be better to keep the behavior the same as `Java` client.

tsungchih: Got it and understood. I'll submit a commit to keep the behavior the same with the Java client. Thanks for the comment.

tsungchih: @unknowntpo The behavior mentioned in this comment has been aligned with the one in Java client. Please take a look into it when you're available. Thanks!

Issue #8642, opened 2025-09-22
[Subtask] Add the IT and doc for job template alteration


jerryshao:
### Describe the subtask

This issue tracks the work of adding the IT and document for job template alteration.

### Parent issue

#8429

PR #8791, opened 2025-10-11
[#8642] feat(doc): Add the IT and doc for job template alteration

jerryshao:

This PR adds the IT and docs for job template alteration.

This is the final PR for job template alteration.

Fix: #8642

No.

ITs.

### DIFF
--- a/docs/open-api/jobs.yaml
+++ b/docs/open-api/jobs.yaml
@@ -137,6 +137,54 @@ paths:
         "5xx":
           $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
 
+    put:
+      tags:
+        - job
+      summary: Update job template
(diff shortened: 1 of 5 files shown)


docs/open-api/jobs.yaml
```diff
@@ -368,103 +416,216 @@ components:
           $ref: "./openapi.yaml#/components/schemas/Audit"
 
     SparkJobTemplate:
-        type: object
```
yuqi1129, line 371: Are you trying to fix the indentation issue that was previously introduced?

jerryshao: Yes.

yuqi1129: I see.

Issue #8792, opened 2025-10-11
[Improvement] Upgrade `google-java-format` to 1.15.0 to support JDK17 syntax


yuqi1129:
### What would you like to be improved?

Currently, the following Java syntax is causing a code style issue.

```
if (change instanceof TableRename rename) {
  // other logic
}
```

As the Gravitino server can only be compiled using JDK17 or higher, it's reasonable to support JDK17 syntax. 

### How should we improve?

Upgrade google-java-format version to support it.

PR #8793, opened 2025-10-11
[#8792] improvement(all): Upgrade google-java-format from 1.7 to 1.15.0 to support JDK 17 language feature

yuqi1129:

Upgrade Google Code format from 1.7 to 1.15.0.

To support language feature in JDK17.

Fix: #8792

N/A.

Existing tests


jerryshao:
How do we make sure to publish the client-side jdk8 compatible jars?

yuqi1129:
> How do we make sure to publish the client-side jdk8 compatible jars?

I have verified it by `./gradlew release`. If there are any other aspects I may have overlooked, please let me know, thanks.

jerryshao:
Please let @diqiu50 to also take a look, he is the main contributor of upgrading to JDK 17.

diqiu50:
There is an issue: during development, if a user uses syntax or classes that are incompatible with JDK 8 in the client module, running ./gradlew release will result in an error. This problem can only be detected in the CI pipeline environment, which causes inconvenience to developers.

yuqi1129:
> There is an issue: during development, if a user uses syntax or classes that are incompatible with JDK 8 in the client module, running ./gradlew release will result in an error. This problem can only be detected in the CI pipeline environment, which causes inconvenience to developers.

This can be resolved by adding a CI pipeline to test whether we can successfully run `./gradlew release`.  One point is that the client side only supports JDK8 is a consensus, so we can add some explicit checks for it. For the server side, we should not remain stagnant. What if we are going to upgrade the Java Development Kit on the server side? 

Another point is that version 1.17.0 definitely supports the JDK8 syntax, so there is no problem in making such a change.

build.gradle.kts
```diff
@@ -118,7 +118,7 @@ allprojects {
         // Fix the Google Java Format version to 1.7. Since JDK8 only support Google Java Format
         // 1.7, which is not compatible with JDK17. We will use a newer version when we upgrade to
         // JDK17.
-        googleJavaFormat("1.7")
+        googleJavaFormat("1.15.0")
```
diqiu50, line 119: Is this the main change in this patch?  Please update the comments

yuqi1129: Yeah,  that is the main change. 

> Please update the comments

I see.

yuqi1129: done

.github/workflows/build.yml
```diff
@@ -135,6 +135,9 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew build -PskipITs -PjdkVersion=${{ matrix.java-version }} -PskipDockerTests=false -x :clients:client-python:build
 
+      - name: Release with Gradle
+        run: ./gradlew release -x test -PjdkVersion=${{ matrix.java-version }}
```
jerryshao: Why do we need to specify the jdk version for `release`. Another thing is that we'd better clean and recompile for release to avoid skipping of gradle tasks - which may hide the problem. Can you please verify it more locally?

yuqi1129: > Why do we need to specify the jdk version for release

There is only one JDK version(JDK17) here,  and I just copied it from `./gradlew build`.

>  Another thing is that we'd better clean and recompile for release to avoid skipping of gradle tasks - which may hide the problem. Can you please verify it more locally?

Okay, I will test locally and then update the CI accordingly.

yuqi1129: It's okay to perform the `release` task after cleaning the old build cache on my local machine. Let me modify the CI and test it on the cloud machine.

jerryshao: There's no `jdkVersion` anymore; it is meaningless to add to the build property.

yuqi1129: I will see and remove all `jdkVersion` occurrences in CI.

yuqi1129: All `jdkVersion` in CI have been removed.

.github/workflows/build.yml
```diff
@@ -133,7 +133,10 @@ jobs:
           sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb  
 
       - name: Build with Gradle
-        run: ./gradlew build -PskipITs -PjdkVersion=${{ matrix.java-version }} -PskipDockerTests=false -x :clients:client-python:build
+        run: ./gradlew build -PskipITs -PskipDockerTests=false -x :clients:client-python:build
+
+      - name: Release with Gradle
+        run: ./gradlew clean && ./gradlew release -x test
```
jerryshao: I would suggest to add `--rerun-tasks` to make sure the task is not cached.

.github/workflows/access-control-integration-test.yml
```diff
@@ -87,9 +87,9 @@ jobs:
       - name: Authorization Integration Test (JDK${{ matrix.java-version }})
         id: integrationTest
         run: |
-          ./gradlew -PtestMode=embedded -PjdbcBackend=h2 -PjdkVersion=${{ matrix.java-version }} -PskipDockerTests=false :authorizations:test
```
jerryshao, line 90: You'd also remove line 59`java-version`.

yuqi1129: No, it was used in L68, please see: 
https://github.com/apache/gravitino/blob/e7210d396e26c0dacbacc564c4432766ef9f97c6/.github/workflows/access-control-integration-test.yml#L66-L70

jerryshao: OK, we can leave it, we may support JDK 21 in the future.

Issue #8725, opened 2025-09-30
[Subtask] Support downloading job artifacts from HCFS-like storage systems


jerryshao:
### Describe the subtask

Currently, the Gravitino job system can only support leveraging artifacts in the local filesystem or remotely using http(s), ftp(s). There exists the requirement to download and leverage artifacts on the HCFS-like storage systems, like HDFS, S3, etc. So this issue tracks the requirement to satisfy such needs.

### Parent issue

#7154

PR #8813, opened 2025-10-14
[#8725] feat(core): Support downloading job artifacts from hcfs

jerryshao:

This PR adds the support of downloading job artifacts from the Hadoop compatible filesystems.

Fix: #8725 

No.

Local tests.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
+++ b/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
@@ -230,6 +230,7 @@ public class IsolatedClassLoader implements Closeable {
         || name.startsWith("org.apache.logging.log4j")
         || name.startsWith("java.")
         || !isCatalogClass(name)
+        || !name.startsWith("org.apache.hadoop")
         || sharedClasses.stream().anyMatch(name::startsWith);
   }
 
@@ -261,7 +262,8 @@ public class IsolatedClassLoader implements Closeable {
    */
   private boolean isBarrierClass(String name) {
     // We need to add more later on when we have more catalog implementations.
-    return barrierClasses.stream().anyMatch(name::startsWith);
+    return barrierClasses.stream().anyMatch(name::startsWith)
+        || name.startsWith("org.apache.hadoop");
   }
 
   private ClassLoader getRootClassLoader() throws Exception {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
```diff
@@ -261,7 +262,8 @@ private boolean isCatalogClass(String name) {
    */
   private boolean isBarrierClass(String name) {
     // We need to add more later on when we have more catalog implementations.
-    return barrierClasses.stream().anyMatch(name::startsWith);
+    return barrierClasses.stream().anyMatch(name::startsWith)
+        || name.startsWith("org.apache.hadoop");
```
yuqi1129, line 266: Why not just put `"org.apache.hadoop"` in the `barrierClasses`, Is the `barrierClasses` designed to handle this case?

docs/manage-jobs-in-gravitino.md
```diff
@@ -169,6 +169,19 @@ Note that:
...
+:::info
+If you want to use the job artifacts on the Hadoop Compatible FileSystems (like `hdfs://`,
+`s3a://`, `gs://`, etc.), You need to make sure that:
+
+1. The Gravitino server has the corresponding Hadoop client libraries in its classpath
+   `{GRAVITINO_HOME}/libs` directory. For example, if you want to use `s3a://` scheme, you need to
+   put the `hadoop-aws` and its dependencies in the `{GRAVITINO_HOME}/libs` directory. By
+   default, the HDFS client libraries are already included in the Gravitino server package.
+2. The Gravitino server has the corresponding Hadoop configurations in its classpath
+   `{GRAVITINO_HOME}/conf` directory. For example, if you want to use `s3a://` scheme, you need to
```
yuqi1129, line 181: > The Gravitino server has the corresponding Hadoop configurations in its classpath
   `{GRAVITINO_HOME}/conf` directory

I can't get this sentence, as there are no `*.xml` in the folder `{GRAVITINO_HOME}/conf`, why do you say that?

Should the word `has` in `Gravitino server has the corresponding` be `needs`?

core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
jerryshao: I'm just testing the code. I will polish it when everything is ready. There still have some other issues according to the CI, I need to investigate more.

jerryshao: The previous CI is passed, but it introduces other problems, I need to check it.

PR #8827, opened 2025-10-17
[#8830] fix (gvfs-fuse) : Fix the build failure in the fuse-filesystem module.

diqiu50:

Fix the build failure in the fuse-filesystem module.
Remove tools that did not work.

Fix: #8827

No

CI

### DIFF
--- a/clients/filesystem-fuse/Makefile
+++ b/clients/filesystem-fuse/Makefile
@@ -54,7 +54,7 @@ install-taplo-cli:
 check-toml: install-taplo-cli
 	taplo check
 
-check: check-fmt check-clippy check-cargo-sort cargo-machete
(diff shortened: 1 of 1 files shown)


clients/filesystem-fuse/Makefile
```diff
@@ -54,7 +54,7 @@ install-taplo-cli:
 check-toml: install-taplo-cli
 	taplo check
 
-check: check-fmt check-clippy check-cargo-sort check-toml cargo-machete
```
diqiu50, line 57: The libraries depended on by check-toml and cargo-machete have been updated. Our project’s chosen Rust version is not compatible, so these two check tools are temporarily removed.

Issue #8795, opened 2025-10-12
[Subtask] Add DTOConverters.from_dto for TableDTO


tsungchih:
### Describe the subtask

We need to implement the following methods / classes to complete DTOConverters.fromDTO(TableDTO tableDTO)

- [x] DTOConverters.fromDTOs(IndexDTO[] indexDTOs)
- [x] DTOConverters.fromDTOs(SortOrderDTO[] sortOrderDTO)
- [x] DTOConverters.fromDTOs(Partitioning[] partitioning)
- [x] DTOConverters.fromDTOs(ColumnDTO[] columns)
- [x] DTOConverters.fromDTO(DistributionDTO distributionDTO)
- [x] DTOConverters.fromDTO(TableDTO tableDTO)
- [x] Implement Class TableImpl(Table) in table.py that will be used in DTOConverters.fromDTO(TableDTO tableDTO)

### Parent issue

#8591

PR #8841, opened 2025-10-18
[#8795] feat(client-python): add from_dto for TableDTO in class DTOConverters

tsungchih:

This PR is aimed at implementing method `from_dto()` for `TableDTO` in class `DTOConverters`. The following methods are implemented due to the dependency.

- `from_dto` for `DistributionDTO`
- `from_dto` for `IndexDTO`
- `from_dto` for `SortOrderDTO`
- `from_dto` for `ColumnDTO`
- `from_dto` for `Partitioning`
- `from_dto` for `TableDTO`
- `from_dtos` for list of `IndexDTO`
- `from_dtos` for list of `SortOrderDTO`
- `from_dtos` for list of `ColumnDTO`
- `from_dtos` for list of `Partitioning`

Before being able to implement `RelationalTable`, we need the method `from_dto()` for `TableDTO` since it is called by `RelationalTable`.

Fix: #8795 

No

Unit tests

### DIFF
--- a/clients/client-python/gravitino/dto/util/dto_converters.py
+++ b/clients/client-python/gravitino/dto/util/dto_converters.py
@@ -36,7 +36,7 @@ from gravitino.api.rel.partitions.list_partition import ListPartition
 from gravitino.api.rel.partitions.range_partition import RangePartition
 from gravitino.api.rel.table import Table
 from gravitino.api.rel.types.types import Types
-from gravitino.client.generic_table import GenericTable
+from gravitino.client.relational_table import RelationalTable
 from gravitino.dto.rel.column_dto import ColumnDTO
 from gravitino.dto.rel.distribution_dto import DistributionDTO
 from gravitino.dto.rel.expressions.field_reference_dto import FieldReferenceDTO
@@ -258,7 +258,7 @@ class DTOConverters:
             Table: The table.
         """
 
-        return GenericTable(
+        return RelationalTable(
             name=dto.name(),
             columns=DTOConverters.from_dtos(dto.columns()),
             partitioning=DTOConverters.from_dtos(dto.partitioning()),
--- a/clients/client-python/gravitino/client/relational_table.py
+++ b/clients/client-python/gravitino/client/relational_table.py
@@ -26,7 +26,7 @@ from gravitino.api.rel.indexes.index import Index
 from gravitino.api.rel.table import Table
 
 
-class GenericTable(Table):  # pylint: disable=too-many-instance-attributes
+class RelationalTable(Table):  # pylint: disable=too-many-instance-attributes
     """A generic table implementation."""
 
     def __init__(


tsungchih:

unknowntpo:
@tsungchih Sorry for the late reply, I'll review this PR today.

clients/client-python/gravitino/api/rel/expressions/transforms/transforms.py
```diff
@@ -241,21 +241,31 @@ def apply(name: str, arguments: List[Expression]) -> "Transforms.ApplyTransform"
...
+    @staticmethod
+    def list(*field_names: List[str]) -> "Transforms.ListTransform": ...
+
+    @overload
     @staticmethod
     def list(
-        *field_names: List[str], assignments: Optional[List[ListPartition]] = None
-    ) -> "Transforms.ListTransform":
-        """Create a transform that includes multiple fields in a list.
+        *, field_names: List[List[str]], assignments: List[ListPartition]
```
jerryshao, line 251: Hi @tsungchih , what is the meaning of `*` here?

clients/client-python/gravitino/dto/util/dto_converters.py
```diff
@@ -76,3 +94,219 @@ def from_function_args(args: list[FunctionArg]) -> list[Expression]:
...
+
+        return Distributions.of(
+            dto.strategy(),
+            dto.number(),
+            *DTOConverters.from_function_args(dto.args()),
+        )
+
+    @from_dto.register
+    @staticmethod
+    def _(dto: IndexDTO) -> Index:
```
jerryshao, line 139: What is the purpose of using `_` for the method?

clients/client-python/gravitino/dto/util/dto_converters.py
```diff
@@ -76,3 +94,219 @@ def from_function_args(args: list[FunctionArg]) -> list[Expression]:
...
+    def _(dto: ColumnDTO) -> Column:
+        """Converts a ColumnDTO to a Column.
+
+        Args:
+            dto (ColumnDTO): The column DTO to be converted.
+
+        Returns:
+            Column: The column.
+        """
+        dto_default_value = dto.default_value()
```
unknowntpo: Is this equivalent to java implementation ? 
It simply return `ColumnDTO` if default value not set.

```
  public static Column fromDTO(ColumnDTO column) {
    if (column.defaultValue().equals(Column.DEFAULT_VALUE_NOT_SET)) {
      return column;
    }
    return Column.of(
        column.name(),
        column.dataType(),
        column.comment(),
        column.nullable(),
        column.autoIncrement(),
        fromFunctionArg((FunctionArg) column.defaultValue()));
  }
```

clients/client-python/gravitino/dto/util/dto_converters.py
```diff
@@ -76,3 +94,219 @@ def from_function_args(args: list[FunctionArg]) -> list[Expression]:
...
+
+        Args:
+            dto (Partitioning): The partitioning DTO to be converted.
+
+        Returns:
+            Transform: The transform.
+        """
+        strategy = dto.strategy()
+        if strategy is Partitioning.Strategy.IDENTITY:
+            return Transforms.identity(dto.field_name())
```
unknowntpo: Seems like dto does not have `field_name()` method, is it common to do this in Python ?

In Java, we use explicit type casting

```
    switch (partitioning.strategy()) {
      case IDENTITY:
        return Transforms.identity(
            ((Partitioning.SingleFieldPartitioning) partitioning).fieldName());
```

Maybe we could follow the same approach. What do you think ? 

Same questions for the rest of check about `strategy`.

clients/client-python/gravitino/dto/util/dto_converters.py
```diff
@@ -76,3 +94,219 @@ def from_function_args(args: list[FunctionArg]) -> list[Expression]:
...
+            sort_order=DTOConverters.from_dtos(dto.sort_order()),
+            distribution=DTOConverters.from_dto(dto.distribution()),
+            index=DTOConverters.from_dtos(dto.index()),
+            comment=dto.comment(),
+            properties=dto.properties(),
+            audit_info=dto.audit_info(),
+        )
+
+    @staticmethod
+    def from_dtos(
```
unknowntpo: In Java, we have multiple overloaded `from_dtos` method, which provide percise type signature. could we follow that approach ?

clients/client-python/gravitino/api/rel/expressions/transforms/transforms.py
tsungchih: Hi @jerryshao , the `*` here means that all of the arguments shown after `*` should be passed as keyword arguments instead of positional arguments. In doing so, we could construct `list` method to support multiple different combinations of argument types. Namely, `list(["country"], ["city"])` and `list(field_names=[["score"]], assignments=[...])`.

clients/client-python/gravitino/dto/util/dto_converters.py
tsungchih: It's a convention in Python when using `singledispatch` or `singledispatchmethod`. After the method has been decorated by `singledispatchmethod`, we register a new data type of argument for the method using `_` to prevent the method from being obscured by a new declaration of the same name.

clients/client-python/gravitino/dto/util/dto_converters.py
tsungchih: Oh, you're right. Thank you for the valuable feedback and for bringing the issue to my attention. Let me fix it.

clients/client-python/gravitino/dto/util/dto_converters.py
tsungchih: Sure, let me add type casting explicitly.

clients/client-python/gravitino/dto/util/dto_converters.py
tsungchih: Perhaps we can leverage `@overload`here. Do you prefer the following implementation?

```python
    @overload
    @staticmethod
    def from_dtos(dtos: list[ColumnDTO]) -> Column: ...

    @overload
    @staticmethod
    def from_dtos(dtos: list[IndexDTO]) -> Index: ...

    @overload
    @staticmethod
    def from_dtos(dtos: list[SortOrderDTO]) -> SortOrder: ...

    @overload
    @staticmethod
    def from_dtos(dtos: list[Partitioning]) -> Transform: ...

    @staticmethod
    def from_dtos(dtos):
        """Converts list of `ColumnDTO`, `IndexDTO`, `SortOrderDTO`, or `Partitioning`
        to the corresponding list of `Column`s, `Index`es, `SortOrder`s, or `Transform`s.

        Args:
            dtos (list[Union[ColumnDTO, IndexDTO, SortOrderDTO, Partitioning]]):
                The DTOs to be converted.

        Returns:
            list[Union[Column, Index, SortOrder, Transform]]: The list of Indexes.
        """
        if not dtos:
            return []
        return [DTOConverters.from_dto(dto) for dto in dtos]

```

jerryshao: I'm fine with either. From my point, we don't have to be too Java-ish. We'd follow the Python style / convention, just make sure the semantic is the same, that should be enough.

clients/client-python/gravitino/dto/util/dto_converters.py
tsungchih: Explicit type casting has been done in this [commit](https://github.com/apache/gravitino/pull/8841/commits/58917a4eb28ba3de5d62e59d30ff395d5b6e066c)

clients/client-python/gravitino/dto/util/dto_converters.py
tsungchih: The corresponding fix has been done in [this commit](https://github.com/apache/gravitino/pull/8841/commits/ef8162d647d203734390f398327e56ae16df15a4).

clients/client-python/gravitino/dto/util/dto_converters.py
unknowntpo: Ok. This make sense.

tsungchih: @jerryshao @unknowntpo After having careful considered as to the comment, I decided to accept @unknowntpo comment since it will make the intention of the method more clear and specific. The current signature cannot reveal this method is intended to accept only the following four input-output combinations:

- list[ColumnDTO]->list[Column]
- list[IndexDTO]->list[Index]
- list[SortOrderDTO]->list[SortOrder]
- list[Partitioning]->list[Transform]

I will make the corresponding changes as a commit. Thanks for both of your comments and suggestions.

tsungchih: @jerryshao @unknowntpo The refactor has been done.

clients/client-python/gravitino/dto/util/dto_converters.py
```diff
@@ -76,3 +108,249 @@ def from_function_args(args: list[FunctionArg]) -> list[Expression]:
...
+        """Converts a TableDTO to a Table.
+
+        Args:
+            dto (TableDTO): The table DTO to be converted.
+
+        Returns:
+            Table: The table.
+        """
+
+        class TableImpl(Table):  # pylint: disable=too-many-instance-attributes
```
jerryshao: I think you will add more features to the `TableImpl` like tags/policies, so  this inline class definition here may not be enough. We'd better define a class `GenericFileset` for better extension.

tsungchih: @jerryshao I don't get your point to define a `GenericFileset` because the class has been defined. Do you mean we'd better define a class similar to `GenericFileset` for better extension, for example, `GenericTable`? After then, we can return an instance of `GenericTable` when calling this static method `from_dto(dto: TableDTO)'.

jerryshao: Yes, that's right. We can follow what Java client did, to define a `GenericTable`. We will add more functionalities to this `GenericTable`, it is better to define a canonical class, not a nested class here.

tsungchih: @jerryshao Please take a review on the added commits when you're available. Thanks.

jerryshao (dismissed):
LGTM.

clients/client-python/gravitino/dto/util/dto_converters.py
jerryshao: Seems like the counterpart in Java is called `RelationalTable`, shall we also rename the class to `RelationalTable` for the alignment?

tsungchih: @jerryshao After having checked the class `RelationalTable` in Java client, I think the `GenericTable` here is different. In the constructor of `RelationalTable`, it holds reference to the return value of `fromDTO(tableDTO)` which is a instance of `GenericTable` in this PR. I think we should keep the class name as `GenericTable`. How do you think?

https://github.com/apache/gravitino/blob/c3c32fa98a7e622629f240e8930bbfe7d442395a/clients/client-java/src/main/java/org/apache/gravitino/client/RelationalTable.java#L117-L120

jerryshao: I think the purpose is the same, whether it holds the `tableDTO`, or the fields of `tableDTO` don't affect the purpose of this class (this class object will be used by the user). We don't have to strictly align with Java's implementation, but my feeling is that the class name should be the same. So for the users who are migrating from Java to Python can easily find the counterpart class to use, what do you think?

tsungchih: Understood. Let me have it renamed.

tsungchih: @jerryshao The class GenericTable has been renamed to RelationalTable.

Issue #8833, opened 2025-10-17
[Subtask] Framework of generic lakehouse catalog


yuqi1129:
### Describe the subtask

This issue does the following things
- Create a submodule under ${GRAVITINO_HOME}/catalogs named `catalog-generic-lakehouse` in catalogs/
- Add basic framework and related classes like `GenericLakehouseCatalog`, `GenericLakehouseCatalogOperations`, without detailed implementation
- Support Gradle operations, such as packages, distribution

### Parent issue

#8828

PR #8842, opened 2025-10-18
[#8833] feat(catalogs): Support the basic framework for the generic lakehouse catalog.

yuqi1129:

Introduce a basic framework for generic lakehouse catalog.

To better manage more lakehouse system.

Fix: #8833 

N/A.

It's just framework, and I just tested it locally.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -56,11 +56,12 @@ github:
     - noidname01
     - orenccl
     - sunxiaojian
-    - tengqm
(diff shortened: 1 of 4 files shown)


yuqi1129:
It seems that we need to cherry-pick commit `d47bc717687cde8e58c9129147151c968745a860` to branch `branch-lance-namepspace-dev from`, which solves the Rust build problem.

Issue #8546, opened 2025-09-15
[Improvement] Use Cache for ID Retrieval in Service


hdygxsj:
### What would you like to be improved?

Cache is not used in the Service.

The Service contains ID query operations that do not go through the cache, which will result in many unnecessary database queries.
<img width="1674" height="506" alt="Image" src="https://github.com/user-attachments/assets/80141744-660e-4a31-ba2d-0478eb4fe62b" />

* org.apache.gravitino.storage.relational.service.MetadataObjectService#getMetadataObjectId

<img width="1680" height="840" alt="Image" src="https://github.com/user-attachments/assets/215871a1-2d72-4894-a704-f3c2e65b8b31" />

* org.apache.gravitino.storage.relational.service.TableMetaService#getTableByIdentifier

<img width="1680" height="840" alt="Image" src="https://github.com/user-attachments/assets/c0e18a37-762b-47f3-9b19-202e7f0bd5ed" />

### How should we improve?

Before querying the database, check whether the corresponding value exists in the cache.

jerryshao:
@hdygxsj can you please point out where the code is? The issue description is too vague; it is hard for others to understand what's going on here and how to fix it.

jerryshao:
@yuqi1129 can you please sync with @hdygxsj to check how to fix it?

PR #8844, opened 2025-10-20
[#8546] improvement(core): Improve id retrieval in service

hdygxsj:

Improve id retrieval in service

Fix: #8546 

None

Existing test case


hdygxsj:
Using EntityStore in the Service layer causes numerous errors; perhaps an alternative approach should be adopted to utilize caching within the Service layer.

Issue #8834, opened 2025-10-17
[Subtask] Adjust entity store structure to support generic lakehouse catalog


yuqi1129:
### Describe the subtask

Currently, Gravitino does not store properties in tables, as they are needed by generic lakehouse catalog, so we need to adjust the store structure of Graivitino entity store. Apart from table properties, they may be other points that need to be changed. For example, table partitioning, table index,ing, and so on. 

### Parent issue

#8828

PR #8847, opened 2025-10-20
[#8834] feat(catalogs): Add a table to store details information of Gravitino managed tables.

yuqi1129:

Add a new table to store table details for Gravitino-managed tables.

To support the managed catalog.

Fix: #8834 

N/A.

Test the process of create table locally.

### DIFF
--- a/scripts/h2/schema-1.1.0-h2.sql
+++ b/scripts/h2/schema-1.1.0-h2.sql
@@ -453,7 +453,7 @@ CREATE TABLE IF NOT EXISTS `table_version_info` (
     `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
     `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
     `properties`      CLOB DEFAULT NULL COMMENT 'table properties',
-    `partitions`  CLOB DEFAULT NULL COMMENT 'table partition info',
+    `partitioning`  CLOB DEFAULT NULL COMMENT 'table partition info',
     `distribution` CLOB DEFAULT NULL COMMENT 'table distribution info',
     `sort_orders` CLOB DEFAULT NULL COMMENT 'table sort order info',
     `indexes`      CLOB DEFAULT NULL COMMENT 'table index info',
--- a/scripts/h2/upgrade-1.0.0-to-1.1.0-h2.sql
+++ b/scripts/h2/upgrade-1.0.0-to-1.1.0-h2.sql
@@ -21,7 +21,7 @@ CREATE TABLE IF NOT EXISTS `table_version_info` (
     `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
     `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
     `properties`      CLOB DEFAULT NULL COMMENT 'table properties',
-    `partitions`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
+    `partitioning`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
     `distribution` MEDIUMTEXT DEFAULT NULL COMMENT 'table distribution info',
     `sort_orders` MEDIUMTEXT DEFAULT NULL COMMENT 'table sort order info',
     `indexes` MEDIUMTEXT DEFAULT NULL COMMENT 'table index info',
--- a/scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
+++ b/scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
@@ -21,7 +21,7 @@ CREATE TABLE IF NOT EXISTS `table_version_info` (
     `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
     `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
     `properties`      MEDIUMTEXT DEFAULT NULL COMMENT 'table properties',
-    `partitions`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
+    `partitioning`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
     `distribution` MEDIUMTEXT DEFAULT NULL COMMENT 'table distribution info',
     `sort_orders` MEDIUMTEXT DEFAULT NULL COMMENT 'table sort order info',
     `indexes`      MEDIUMTEXT DEFAULT NULL COMMENT 'table index info',
--- a/scripts/mysql/schema-1.1.0-mysql.sql
+++ b/scripts/mysql/schema-1.1.0-mysql.sql
@@ -444,7 +444,7 @@ CREATE TABLE IF NOT EXISTS `table_version_info` (
     `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
     `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
     `properties`      MEDIUMTEXT DEFAULT NULL COMMENT 'table properties',
-    `partitions`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
+    `partitioning`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
     `distribution` MEDIUMTEXT DEFAULT NULL COMMENT 'table distribution info',
     `sort_orders` MEDIUMTEXT DEFAULT NULL COMMENT 'table sort order info',
     `indexes`      MEDIUMTEXT DEFAULT NULL COMMENT 'table index info',
--- a/scripts/postgresql/schema-1.1.0-postgresql.sql
+++ b/scripts/postgresql/schema-1.1.0-postgresql.sql
@@ -772,7 +772,7 @@ CREATE TABLE IF NOT EXISTS table_version_info (
     table_id        BIGINT PRIMARY KEY,
     format          VARCHAR(64) NOT NULL,
     properties      TEXT,
-    partitions  TEXT,
+    partitioning  TEXT,
     distribution TEXT,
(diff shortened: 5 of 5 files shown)


yuqi1129:
@mchades 
Please check if we need to modify the structure of `schema_meta` and `table_meta`.

yuqi1129:
The schema and catalog, we need to add locatation information at least, and whether we can store the information in `properties`.

scripts/h2/schema-1.1.0-h2.sql
```diff
@@ -0,0 +1,464 @@
...
+    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'job run current version',
+    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'job run last version',
+    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'job run deleted at',
+    PRIMARY KEY (`job_run_id`),
+    UNIQUE KEY `uk_mid_jei_del` (`metalake_id`, `job_execution_id`, `deleted_at`),
+    KEY `idx_job_template_id` (`job_template_id`),
+    KEY `idx_job_execution_id` (`job_execution_id`)
+    ) ENGINE=InnoDB;
+
+CREATE TABLE IF NOT EXISTS `table_details` (
```
mchades: how about using `table_version_info` to align with existing `policy_version_info`, `model_version_info`, and `fileset_version_info`?

scripts/h2/schema-1.1.0-h2.sql
```diff
@@ -0,0 +1,464 @@
...
+    `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
+    `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
+    `location`        VARCHAR(512) NOT NULL COMMENT 'table storage location',
+    `external`        BOOLEAN NOT NULL DEFAULT FALSE COMMENT 'whether the table is external table',
+    `properties`      CLOB DEFAULT NULL COMMENT 'table properties',
+    `partition_info`  CLOB DEFAULT NULL COMMENT 'table partition info',
+    `index_info`      CLOB DEFAULT NULL COMMENT 'table index info',
+    `current_version` BIGINT(20) UNSIGNED COMMENT 'table current version',
+    `last_version`    BIGINT(20) UNSIGNED COMMENT 'table last version',
```
mchades: What's the usage of `last_version`

scripts/h2/schema-1.1.0-h2.sql
yuqi1129: Ok, it's also accetable.

scripts/h2/schema-1.1.0-h2.sql
yuqi1129: `last_version` marks the largest version that has been successfully committed to storage. In most cases, current_version equals `last_version`. However, if we want to roll back to a certain version, then we can change `current_version` to an old one. The following is an example: 

current_version = 3
last_version = 3, 

When we want to write a new version, then last_version++ and current_version = last_version(Now, they are both 4).  All available versions are 1, 2, 3, and 4. We can change current_version to any value between 1 and 4 to use a specific version. After a series of operations,  we can use `last_version=4` to indicate that the next version is 5.

mchades: What is the difference between `last_version` here and `last_version` in the `table_meta` table?

scripts/h2/schema-1.1.0-h2.sql
```diff
@@ -0,0 +1,466 @@
...
+    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'job run current version',
+    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'job run last version',
+    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'job run deleted at',
+    PRIMARY KEY (`job_run_id`),
+    UNIQUE KEY `uk_mid_jei_del` (`metalake_id`, `job_execution_id`, `deleted_at`),
+    KEY `idx_job_template_id` (`job_template_id`),
+    KEY `idx_job_execution_id` (`job_execution_id`)
+    ) ENGINE=InnoDB;
+
+CREATE TABLE IF NOT EXISTS `table_version_info` (
```
mchades, line 452: missing `comment` field?

scripts/h2/schema-1.1.0-h2.sql
yuqi1129: They should be the same, let me check whether we can remove `last_version` as we have define it in `table_meta`.

mchades: yeah, I think it's unecessary here

scripts/h2/schema-1.1.0-h2.sql
mchades: What I mean is the comment on Lance's table, the Iceberg table, etc.

scripts/h2/upgrade-1.0.0-to-1.1.0-h2.sql
```diff
@@ -0,0 +1,35 @@
...
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS `table_version_info` (
+    `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
+    `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
+    `location`        VARCHAR(512) NOT NULL COMMENT 'table storage location',
+    `external`        BOOLEAN NOT NULL DEFAULT FALSE COMMENT 'whether the table is external table',
+    `properties`      CLOB DEFAULT NULL COMMENT 'table properties',
```
jerryshao, line 23: Why do you use `CLOB` for this field?

scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
```diff
@@ -0,0 +1,35 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS `table_version_info` (
+    `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
+    `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
+    `location`        VARCHAR(512) NOT NULL COMMENT 'table storage location',
+    `external`        BOOLEAN NOT NULL DEFAULT FALSE COMMENT 'whether the table is external table',
```
jerryshao: My feeling is that these fields can be property, what do you think? @mchades . IIRC, these fields are properties for Hive table, can you please check Iceberg and other tables? If so, I think we can follow the convention.

scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
```diff
@@ -0,0 +1,35 @@
...
+    `properties`      MEDIUMTEXT DEFAULT NULL COMMENT 'table properties',
+    `partitions`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
+    `distribution` MEDIUMTEXT DEFAULT NULL COMMENT 'table distribution info',
+    `sort_orders` MEDIUMTEXT DEFAULT NULL COMMENT 'table sort order info',
+    `indexes_info`      MEDIUMTEXT DEFAULT NULL COMMENT 'table index info',
```
jerryshao: All these fields will be serialized into json string and write into the DB, right?

scripts/h2/upgrade-1.0.0-to-1.1.0-h2.sql
yuqi1129: `CLOB` is the for H2 and it is equivalent to`TEXT` type in MySQL

scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
yuqi1129: Yeah, I intend to do so.

scripts/h2/schema-1.1.0-h2.sql
jerryshao: "last_version" is quite misleading since started, actually it is the latest version. Besides, I agree with @mchades that this field is not so necessary.

scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
mchades: Yes, I would also suggest putting `location` in the table property. 

Considering that there may be a need to filter the table according to the `format` later, the `format` can be stored as a separate field.

yuqi1129: I will put `location` and `external` in properties and let `format` as a field.

scripts/h2/upgrade-1.0.0-to-1.1.0-h2.sql
```diff
@@ -0,0 +1,33 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS `table_version_info` (
+    `table_id`        BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
+    `format`          VARCHAR(64) NOT NULL COMMENT 'table format, such as Lance, Iceberg and so on',
+    `properties`      CLOB DEFAULT NULL COMMENT 'table properties',
+    `partitions`  MEDIUMTEXT DEFAULT NULL COMMENT 'table partition info',
```
mchades: should use `Partitioning`?

scripts/mysql/upgrade-1.0.0-to-1.1.0-mysql.sql
roryqi: Location is important. If we need to check the locations whether is used  or duplicated.

Issue #8856, opened 2025-10-20
[Bug report] Iceberg table owner field is not set to the OAuth user, it's set to the user from client environment


bharos:
### Version

main branch

### Describe what's wrong

When OAuth is enabled, On creating a table, the `owner` field should reflect the user from OAuth token. Currently, that doesn't happen. The user name from the client environment is attached to the table owner property.

<img width="1045" height="945" alt="Image" src="https://github.com/user-attachments/assets/7db73bbc-3135-4765-9628-1dabe29577ce" />

### Error message and/or stacktrace

```
2025-10-20 21:21:27 INFO  [iceberg-rest-29] IcebergTableOperations:125 - Create Iceberg table, catalog: RokuIcebergCatalog, namespace: test_db, create table request: CreateTableRequest{name=product_tbl2, location=null, properties={owner=spark}, schema=table {
  0: id: optional long
  1: name: optional string
  2: price: optional decimal(10, 2)
  3: updated_at: optional timestamptz
}, partitionSpec=org.apache.iceberg.UnboundPartitionSpec@1196791f, writeOrder=null, stageCreate=false}, accessDelegation: vended-credentials, isCredentialVending: true
```

The table request receives the environment username . (in the above example, it is `spark`) and that is set as owner

### How to reproduce

Create an iceberg table with OAuth enabled. Still you will not see the user name as the user from the OAuth token

PR #8857, opened 2025-10-20
[#8856] fix(iceberg): Override table owner with OAuth authenticated user

bharos:

Modified IcebergTableOperationExecutor to override the client-provided owner property with the authenticated user from OAuth/JWT tokens when creating Iceberg tables.

When tables are created via Spark or other Iceberg clients, the client sends its own 'owner' property value (e.g., 'spark', 'system', etc.) in the CreateTableRequest. This results in incorrect table ownership being stored in the metadata, making audit trails unreliable and not reflecting the actual authenticated user.

The fix ensures that table ownership reflects the actual authenticated user rather than the client's environment or default values.

Yes. Tables created through the Iceberg REST API will now show the authenticated user (from OAuth/JWT token) as the owner instead of the value sent by the client (e.g., 'spark').

Tested with Spark SQL creating tables through Gravitino Iceberg REST service with OAuth authentication. Verified that:
- Tables now show correct owner (e.g., 'bmurali@roku.com')
- Previous behavior showed 'spark' as owner
- Backward compatible when OAuth is disabled

Before fix:
Screenshot 2025-10-20 at 2 45 56 PM

After fix:
Screenshot 2025-10-20 at 2 46 04 PM

Fixes: #8856


roryqi:
Could you add a UT for this? Because this is a fix, so we would better have a delicated UT.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -41,6 +50,34 @@ public IcebergTableOperationExecutor(IcebergCatalogWrapperManager icebergCatalog
...
   public LoadTableResponse createTable(
       IcebergRequestContext context, Namespace namespace, CreateTableRequest createTableRequest) {
+    String authenticatedUser = context.userName();
+    if (StringUtils.isNotBlank(authenticatedUser)
+        && !AuthConstants.ANONYMOUS_USER.equals(authenticatedUser)) {
+      String existingOwner = createTableRequest.properties().get(OWNER_PROPERTY);
+
+      // Updated the owner as the authenticated user, if owner missing or different from
+      // authenticated user
+      if (existingOwner == null || !existingOwner.equals(authenticatedUser)) {
```
FANNG1: How about using `if(!authenticatedUser.equals(existingOwner)) {}` to simple the code?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -29,9 +33,14 @@
 import org.apache.iceberg.rest.responses.ListTablesResponse;
 import org.apache.iceberg.rest.responses.LoadCredentialsResponse;
 import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class IcebergTableOperationExecutor implements IcebergTableOperationDispatcher {
 
+  private static final Logger LOG = LoggerFactory.getLogger(IcebergTableOperationExecutor.class);
+  private static final String OWNER_PROPERTY = "owner";
```
FANNG1: Could you define OWNER in IcebergConstants?

bharos:
> Could you add a UT for this? Because this is a fix, so we would better have a delicated UT.

Done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: Done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: Done

FANNG1 (approved):

FANNG1:
LGTM

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -41,6 +50,33 @@ public IcebergTableOperationExecutor(IcebergCatalogWrapperManager icebergCatalog
   @Override
   public LoadTableResponse createTable(
       IcebergRequestContext context, Namespace namespace, CreateTableRequest createTableRequest) {
+    String authenticatedUser = context.userName();
+    if (StringUtils.isNotBlank(authenticatedUser)
```
roryqi: The authenticatoedUser won't be blank. If the user is blank, the authenticator will fill this by using anoymous. We can remove this condition to simplify the code.

FANNG1: Could `authenticatoedUser` be blank in the other authenticators?

roryqi: No. If the authenciatedUser is blank, other authenticators should reject it and throw unauthorizationException.

bharos: Done. Removed this check and the associated unit test

jerqi (approved):
LGTM.

Issue #8858, opened 2025-10-20
[Bug report] unknown error when using fsspec through JNI


youngyjd:
### Version

main branch

### Describe what's wrong

Context:
Our fsspec-cfs is based on JNI and ultimately uses the Java HDFS client under the hood.

I observed that a Ray application, before writing, will first call ls() to check if the file exists.
The libhdfs JNI layer captures the Java FileNotFoundException and converts it into Python's FileType.NotFound (this [code](https://github.com/apache/hadoop/blob/4235dc626874df852864e3689d93ec280f53c534/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfs/exception.c#L37-L41) might be related).
Ray then determines whether the file exists based on this FileType ([code](https://github.com/ray-project/ray/blob/364f83184753e16e1eadbd905ac9c664d52adde6/python/ray/data/datasource/file_datasink.py#L111)).

The issue here is that the JNI layer doesn't seem to convert a [FilesetPathNotFoundException](https://github.com/datastrato/gravitino-uber/blob/10ab935cf592c4f81693cc7ffda04c13c06bd338/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetPathNotFoundException.java#L24C14-L24C42) into Python's FileType.NotFound.
As a result, when fsspec tries to list the file, it ends up with an "unknown error".

### Error message and/or stacktrace

```
File "/usr/local/lib/python3.9/dist-packages/ray/data/datasource/file_datasink.py", line 106, in on_write_start
  if self.filesystem.get_file_info(self.path).type is FileType.NotFound:
File "pyarrow/_fs.pyx", line 590, in pyarrow._fs.FileSystem.get_file_info
File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
File "pyarrow/error.pxi", line 89, in pyarrow.lib.check_status
File "pyarrow/_fs.pyx", line 1498, in pyarrow._fs._cb_get_file_info
File "/usr/local/lib/python3.9/dist-packages/pyarrow/fs.py", line 322, in get_file_info
  info = self.fs.info(path)
File "/usr/local/lib/python3.9/dist-packages/fsspec/spec.py", line 681, in info
  out = self.ls(self._parent(path), detail=True, **kwargs)
File "/home/docker/core/fsspec_cfs/cfs.py", line 237, in <lambda>
[...]

PR #8859, opened 2025-10-20
[#8858] fix(gvfs-java) remove filesetPathNotFoundException Java

youngyjd:

Remove FilesetPathNotFoundException and throws FileNotFoundException instead. 

When use fsspec through JNI, the JNI layer cannot convert a [FilesetPathNotFoundException](https://github.com/datastrato/gravitino-uber/blob/10ab935cf592c4f81693cc7ffda04c13c06bd338/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/FilesetPathNotFoundException.java#L24C14-L24C42) into Python's FileType.NotFound.

As a result, when fsspec tries to list the file, it ends up with an "unknown error".

Fix: #8858

FilesetPathNotFoundException will be removed. Users will see FileNotFoundException instead. 

1. unit tests.
2. tested in our environment


mchades (approved):
LGTM.

Issue #8861, opened 2025-10-21
[Improvement] testPolicyAndTagCacheWeigher() failure


youngyjd:
### Version

main branch

### Describe what's wrong

Need to fix the test: 
```
TestCacheConfig > testPolicyAndTagCacheWeigher() FAILED
    org.awaitility.core.ConditionTimeoutException: Condition with Lambda expression in org.apache.gravitino.cache.TestCacheConfig was not fulfilled within 5 seconds.
        at app//org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
        at app//org.awaitility.core.CallableCondition.await(CallableCondition.java:78)
        at app//org.awaitility.core.CallableCondition.await(CallableCondition.java:26)
        at app//org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1006)
        at app//org.awaitility.core.ConditionFactory.until(ConditionFactory.java:975)
        at app//org.apache.gravitino.cache.TestCacheConfig.testPolicyAndTagCacheWeigher(TestCacheConfig.java:162)
```

### Error message and/or stacktrace

As above

### How to reproduce

"build" Github action can see the failure

youngyjd:
It seems 10 second timeout is still not enough although most of the times it passes.

See failure in this job - https://github.com/apache/gravitino/actions/runs/18693766507/job/53305862655?pr=8868

<img width="1093" height="226" alt="Image" src="https://github.com/user-attachments/assets/5efe25cd-3ce0-40b8-8cc3-acce98cf687c" />

youngyjd:
20 seconds don't even work.

<img width="1054" height="187" alt="Image" src="https://github.com/user-attachments/assets/2e47505e-24d0-4e60-80e8-9ab360783777" />

PR #8862, opened 2025-10-21
[#8861] Improve(test): fix testPolicyAndTagCacheWeigher

youngyjd:

Bump timeout from 5 seconds to 10 seconds

```
TestCacheConfig > testPolicyAndTagCacheWeigher() FAILED
    org.awaitility.core.ConditionTimeoutException: Condition with Lambda expression in org.apache.gravitino.cache.TestCacheConfig was not fulfilled within 5 seconds.
        at app//org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
        at app//org.awaitility.core.CallableCondition.await(CallableCondition.java:78)
        at app//org.awaitility.core.CallableCondition.await(CallableCondition.java:26)
        at app//org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1006)
        at app//org.awaitility.core.ConditionFactory.until(ConditionFactory.java:975)
        at app//org.apache.gravitino.cache.TestCacheConfig.testPolicyAndTagCacheWeigher(TestCacheConfig.java:162)
```

see failure: https://github.com/apache/gravitino/actions/runs/18668130387/job/53223578211?pr=8859

Fix: #8861

No

Updated unit tests

### DIFF
--- a/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
+++ b/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
@@ -157,7 +157,7 @@ public class TestCacheConfig {
     // There should no tag entities in the cache, because the weight of each tag entity is 100 that
     // is higher than the maximum weight of the fileset entity which is 200.
     Awaitility.await()
-        .atMost(Duration.ofSeconds(5))
+        .atMost(Duration.ofSeconds(10))
         .pollInterval(Duration.ofMillis(10))
         .until(
             () ->


core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
```diff
@@ -152,22 +152,40 @@ void testPolicyAndTagCacheWeigher() throws InterruptedException {
...
     }
 
-    Thread.sleep(1000);
+    // Force cache cleanup to trigger eviction synchronously
+    cache.cleanUp();
 
-    // There should no tag entities in the cache, because the weight of each tag entity is 100 that
-    // is higher than the maximum weight of the fileset entity which is 200.
-    Awaitility.await()
-        .atMost(Duration.ofSeconds(5))
```
yuqi1129, line 160: Is a bigger timeout value, like 10 or 20, resolve the test problem?

youngyjd: - Doesn't actually fix the root cause - the test could still fail even after 10 seconds. test might pass 99% of time but still be flaky
- Still non-deterministic - specific tags (0-3) may never be evicted, regardless of how long we wait

Is there any reason you think waiting 10 seconds is better than we evict cache to clean up?

youngyjd: let me try 10 seconds and see if it works

youngyjd: I tried 10 seconds locally and it worked fine. Let's see how it goes in the CICD

yuqi1129: I was aware that the eviction is asynchronous, and the time it takes depends on the machine perfomance, however there should be a value that fits on a single-core machine.

youngyjd:
@yuqi1129 @mchades ptal

yuqi1129:
> @yuqi1129 @mchades ptal

It's currently LGTM, and let me test it more times.

Issue #8865, opened 2025-10-21
[Improvement] Add variable replacement for all the job template fields


jerryshao:
### What would you like to be improved?

Some times, we need to use different version of job artifacts in the runtime, so we should have a mechanism to specify variable in the job artifact path and replaced in the runtime.

PR #8866, opened 2025-10-21
[#8865] improve(core): Add the placeholder support for all the job template fields

jerryshao:

This PR adds the placeholder support for all the job template fields to give users the flexibility to control.

To support multi-version job artifacts.

Fix: #8865 

No.

UTs been added.


yuqi1129 (approved):
It's currently LGTM.

yuqi1129 (dismissed):
There are no big changes since the last comment, and it's LGTM.

core/src/main/java/org/apache/gravitino/job/JobManager.java
mchades, line 1: Should we also update the user documentation?

jerryshao: I remebered that I've updated the doc, let me check if it is not committed.

PR #8871, opened 2025-10-21
 [#8871]Improve(core): fix EntityCacheWeigher and testPolicyAndTagCacheWeigher

youngyjd:

Fix EntityCacheWeigher priorities and update cache eviction test

This commit corrects the entity weight configuration in EntityCacheWeigher and updates the corresponding test to validate the new behavior.

The previous weight configuration was inverted:
- Schema: weight=500 (incorrectly high, causing unwanted evictions)
- Tag/Policy: weight=100 (incorrectly low, preventing desired evictions)

This violated the Caffeine cache principle where higher weights lead to higher eviction priority. 

Correct EntityCacheWeigher and fix the flaky test of testPolicyAndTagCacheWeigher

Fix: #8861 

No

Unit test updated

### DIFF
--- a/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
+++ b/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
@@ -60,7 +60,7 @@ public class TestCacheConfig {
   @Test
   void testPolicyAndTagCacheWeigher() throws InterruptedException {
     Caffeine<Object, Object> builder = Caffeine.newBuilder();
-    builder.maximumWeight(4500);
+    builder.maximumWeight(5000);
     builder.weigher(EntityCacheWeigher.getInstance());
     Cache<EntityCacheRelationKey, List<Entity>> cache = builder.build();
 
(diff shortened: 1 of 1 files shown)


yuqi1129:
It seems that we need to clear the evicted cache synchronously, just as what you have done here or a long timeout could not guarantee the test can pass.

core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
```diff
@@ -152,22 +152,31 @@ void testPolicyAndTagCacheWeigher() throws InterruptedException {
...
+            .filter(
+                tagNameIdent ->
+                    cache.getIfPresent(
+                            EntityCacheRelationKey.of(tagNameIdent, Entity.EntityType.TAG))
+                        != null)
+            .count();
+
+    Assertions.assertTrue(
+        remainingTags < 10,
+        "Expected some tags to be evicted, but found " + remainingTags + " tags still in cache");
```
yuqi1129: What I want to test is that the cache will evict tags first and then the fileset. Could you add a check that none of filesets have been cleared if the number of remaining tags is less than 10?

yuqi1129: > Could you add a check that none of filesets have been cleared if the number of remaining tags is less than 10?

If this cannot be passed, the goal of #8816 has not been reached, and we need to find another solution to accomplish it.

youngyjd: @yuqi1129 it's non deterministic. It seems you cannot simply rely on weight to control which cache gets evicted first. If you want a guarantee that lower weight cache gets evicted first, likely you will need a new solution.

Issue #8873, opened 2025-10-22
Enable Kerberos. The _HOST part in the SPN HTTP/_HOST account was not replaced with the actual hostname, resulting in an authentication error


forLp811:
### Version

0.9.0

### Describe what's wrong

I have integrated the gravitino 0.9.0 version into ambari 2.7.8. My cluster has enabled Kerberos authentication. My service's kerberos.json configuration is: 
{
  "services": [
    {
      "name": "GRAVITINO",
      "identities": [
        {
          "name": "gravitino_smokeuser",
          "reference": "/smokeuser"
        },
        {
          "name": "gravitino_user",
          "principal": {
            "value": "HTTP/_HOST@${realm}",
            "type" : "user",
            "local_username" : "${gravitino-env/gravitino_user}",
            "configuration": "gravitino-conf/gravitino.authenticator.kerberos.principal"
          },
          "keytab": {
            "file": "${keytab_dir}/spnego.headless.keytab",
            "owner": {
              "name": "${gravitino-env/gravitino_user}",
              "access": "r"
            },
            "group": {
              "name": "${cluster-env/user_group}",
              "access": "r"
            },
            "configuration": "gravitino-conf/gravitino.authenticator.kerberos.keytab"
          }
        }
      ],
      "configurations": [
        {
          "gravitino-conf": {
            "gravitino.authenticators": "kerberos"
          }
        }
      ],
      "components": [
        {
          "name": "GRAVITINO_SERVER"
        }
      ]
    }
  ]
}

The corresponding authentication configuration in gravitino.conf is as follows:

<img width="1086" height="477" alt="Image" src="https://github.com/user-attachments/assets/3aef4816-2739-47cb-9d19-86b6635b3af2" />

The host names of my three hosts are: kk1, kk2, and kk3.

I used a regular account for authentication: kinit testaa, and then operated the API using curl --negotiate -u : http://xxx:8090/xxx. However, the backend log reported a Kerberos authentication error.

### Error message and/or stacktrace

No screenshots were provided for the specific error. The error occurred during the authentication process when the _HOST part of the principal "HTTP/_HOST@${realm}" was not replaced with the actual hostname "kk1". For example: HTTP/kk1@${realm}

### How to reproduce

Please refer to the description I provided earlier.

roryqi:
`_HOST` doesn't support yet. If you have interest, you can support it.

forLp811:
[...]

PR #8874, opened 2025-10-22
[#8873] fix: Enable Kerberos. The _HOST part in the SPN HTTP/_HOST account was not replaced with the actual hostname, resulting in an authentication error

forLp811:
Support automatic replacement of _HOST with the host name

In the Kerberos security mode, the account format is: HTTP/_HOST@realm，this pr Support automatic replacement of _HOST with the host name

Support automatic replacement of _HOST with the host name （eg: HTTP/_HOST@realm）

Fix: #(8873)

NO

It has been verified in our production environment and has been deployed

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosAuthenticator.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosAuthenticator.java
@@ -19,7 +19,6 @@ import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.Base64;
 import java.util.List;
-import java.net.InetAddress;
 import java.net.UnknownHostException;
 import javax.security.auth.Subject;
 import javax.security.auth.kerberos.KerberosPrincipal;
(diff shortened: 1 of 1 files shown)


roryqi:
Could you add a UT for this fix?

server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosAuthenticator.java
```diff
@@ -133,6 +135,21 @@ public Principal run() throws Exception {
     }
   }
 
+  private String replaceHostPlaceholder(String principal) {
```
roryqi: We would better put this method to a Utils class. It may be more convenient to reuse.

forLp811: > We would better put this method to a Utils class. It may be more convenient to reuse.

done

roryqi:
Could you fix the CI failures?

roryqi:
@forLp811 Gently ping. Could u fix CI failure?

Issue #8876, opened 2025-10-22
[Improvement] In the Kerberos security mode, after the Knox proxy Gravitino UI is enabled, Gravitino initiates authentication. Users passing through the Knox proxy are not directly transmitted to the Gravitino user system


forLp811:
### What would you like to be improved?

Generally, if kereros is enabled, the UI interface cannot be opened. In our production environment, we have inherited the Knox proxy Gravitino Web UI, and the interface can be successfully opened. However, after Gravitino enables authentication, it cannot obtain the Knox login user. Gravitino needs to enhance its ability to obtain the proxy user.

forLp811:
> ### What would you like to be improved?
> Generally, if kereros is enabled, the UI interface cannot be opened. In our production environment, we have inherited the Knox proxy Gravitino Web UI, and the interface can be successfully opened. However, after Gravitino enables authentication, it cannot obtain the Knox login user. Gravitino needs to enhance its ability to obtain the proxy user.
> 
> ### How should we improve?
> _No response_

I have already fixed this issue in my production environment. The PR is as follows: https://github.com/apache/gravitino/pull/8877. Please refer to it.

PR #8877, opened 2025-10-22
improve [#8876] Handle the doAs users of the Knox proxy, Solve the issue where after enabling authorization for gravitino, the proxy users who login through knox cannot be identified.

forLp811:

In the Kerberos security mode,  Solve the issue where after enabling authorization for gravitino, the proxy users who login through knox cannot be identified，this pr Support identification of Knox's doAs users （knox's log eg: 25/09/11 06:26:16 ||880e3d9e-2ab3-4df9-b1fb-192ff83a95cb|audit|10.56.33.123|GRAVITINO|testaa|||dispatch|uri|http://kk1:8090/api/metalakes?_t=1757586340965&doAs=testaa|success|Response status: 200）

After Knox proxyed the Gravitino UI, Gravitino enabled authorization , The gravitino needs to identify the principal for the Kerberos login

Fix: #(8876)

NO

It has been verified in our production environment and has been deployed

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/Authenticator.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/Authenticator.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.server.authentication;
 
 import java.security.Principal;
 import org.apache.gravitino.Config;
+import javax.servlet.http.HttpServletRequest;
 
 /** The interface provides authentication mechanism. */
 public interface Authenticator {


server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosAuthenticator.java
```diff
@@ -169,15 +181,22 @@ private Principal retrievePrincipalFromToken(String serverPrincipal, byte[] clie
...
         throw new UnauthorizedException("GssContext isn't established", challenge);
       }
-
-      // Usually principal names are in the form 'user/instance@REALM' or 'user@REALM'.
-      List<String> principalComponents =
-          Splitter.on('@').splitToList(gssContext.getSrcName().toString());
-      if (principalComponents.size() != 2) {
-        throw new UnauthorizedException("Principal has wrong format", AuthConstants.NEGOTIATE);
+      String doAsUser = getDoAsUserFromRequest(request);
+      // If the doAs parameter exists, use that user; otherwise, use the user in the Kerberos token.
```
roryqi, line 185: This is not safe. We should limit the users which the user can impersonate.

roryqi:
There are several points to discuss:
1. Is the parameter `doAs` standard operation for Hadoop system?
2. We should limit the users which is allowed impersonate, this should be a configuration files.
3. Should the simple mode be thought about `doAs`?
4. The interfaces shouldn't be bind to HTTP. Maybe we support other protocols, we can reuse the authenticator as soon as we can.

forLp811:
> There are several points to discuss:
> 
> 1. Is the parameter `doAs` standard operation for Hadoop system?
> 2. We should limit the users which is allowed impersonate, this should be a configuration files.
> 3. Should the simple mode be thought about `doAs`?
> 4. The interfaces shouldn't be bind to HTTP. Maybe we support other protocols, we can reuse the authenticator as soon as we can.

In the Hadoop ecosystem, if Kerberos is enabled, Knox is a dedicated component used to act as an intermediary for the Web UI., This "doAs" is merely a parameter name for the Knox component. The user still needs to login using the Kerberos password, and it is secure. 

simple mode, webui can be opened normally, then Knox is not necessary.

forLp811:
> > There are several points to discuss:
> > 
> > 1. Is the parameter `doAs` standard operation for Hadoop system?
> > 2. We should limit the users which is allowed impersonate, this should be a configuration files.
> > 3. Should the simple mode be thought about `doAs`?
> > 4. The interfaces shouldn't be bind to HTTP. Maybe we support other protocols, we can reuse the authenticator as soon as we can.
> 
> In the Hadoop ecosystem, if Kerberos is enabled, Knox is a dedicated component used to act as an intermediary for the Web UI., This "doAs" is merely a parameter name for the Knox component. The user still needs to login using the Kerberos password, and it is secure.
> 
> simple mode, webui can be opened normally, then Knox is not necessary.

Our company uses Knox to proxy the Kerberos-based Gravitino WebUI. This improvement has resolved the issue of abnormal authentication caused by the principal user passed through Knox (this user is simply called "doAs" in Knox).

roryqi:
@lmccay Could you give some suggestion about this issue?

PR #8880, opened 2025-10-22
[#8881] improvement(fileset): use fileset write lock instead of schema lock for fileset creation

youngyjd:

Scope locking down from schema-level to a fileset-level write lock during fileset creation. This allows multiple filesets to be created concurrently within the same schema and should improve create-fileset throughput.

The schema-wide lock serialized all creations within a schema, becoming a bottleneck. Using a per-fileset write lock reduces contention and increases parallelism. Note: this is a performance–consistency trade-off; during active writes, listing operations may briefly return incomplete results until creations commit.

Fix: #8881 

no

manual testing


mchades:
LGTM. @yuqi1129 could you please also take a review, thx!

Issue #8821, opened 2025-10-16
[Improvement] Support get all partition stats for a table


FANNG1:
### What would you like to be improved?

Some systems may need to retrieve all partition stats and select proper partition to do some action. it's hard to pass  `PartitionRange` to get all partitions currently.

FANNG1:
@jerqi PTAL

roryqi:
I‘ll take a look.

PR #8894, opened 2025-10-23
[#8821] feat(core,server): Supports to query all partition statisitics

roryqi:

 Supports to query all partition statisitics

Fix: #8821

No.

Added IT.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/stats/PartitionRange.java
+++ b/api/src/main/java/org/apache/gravitino/stats/PartitionRange.java
@@ -38,6 +38,9 @@ public class PartitionRange {
 
   private PartitionRange() {}
 
+  /** A PartitionRange that includes all partitions. */
+  public static final PartitionRange ALL_PARTITIONS = new PartitionRange();
+
   /**
    * Creates a PartitionRange which only has upper bound partition name.
    *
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/apache/gravitino/stats/PartitionRange.java
```diff
@@ -38,6 +38,9 @@ public class PartitionRange {
 
   private PartitionRange() {}
 
+  /** A PartitionRange that includes all partitions. */
+  public static final PartitionRange ALL_PARTITIONS = new PartitionRange();
```
FANNG1, line 42: is it nesseary to update related document?

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/StatisticIT.java
```diff
@@ -180,6 +181,20 @@ public void testPartitionStatisticLifeCycle() {
     Assertions.assertEquals(statistics[1].name(), "custom-k2");
     Assertions.assertEquals(statistics[1].value().get(), StatisticValues.stringValue("v2"));
 
+    // Test list all partitions
+    listedStats =
```
FANNG1, line 185: this is the test for memory store? Have you test lance store?

roryqi: This case tested the Lance storage.

api/src/main/java/org/apache/gravitino/stats/PartitionRange.java
roryqi: We don't have similar documents.

PR #8904, opened 2025-10-23
[#8909] feat(lance): Add metric systems to lance server

youngyjd:

1. Added Lance REST server metrics constant (MetricsSource.java)
2. Added MapperConfig rule to transform Dropwizard metrics to Prometheus format
3. Integrated metrics collection in Lance REST service (LanceRESTService.java)
4. Enables automatic exposure of metrics at /metrics and /prometheus/metrics endpoints

Observability is needed for the new lance rest service.

Fix: #8909 

Yes - New metrics endpoints

Standalone mode:
- http://localhost:9101/metrics (Dropwizard JSON format)
- http://localhost:9101/prometheus/metrics (Prometheus text format)

Auxiliary mode:
- Metrics exposed on all service ports (8090, 9001, 9101) showing unified metrics from all services

Unit tests

### DIFF
--- a/.github/workflows/gvfs-fuse-build-test.yml
+++ b/.github/workflows/gvfs-fuse-build-test.yml
@@ -2,10 +2,11 @@ name: Build gvfs-fuse and testing
 
 # Controls when the workflow will run
 on:
-  push:
-    branches: [ "main", "branch-*" ]
-  pull_request:
-    branches: [ "main", "branch-*" ]
+  # Temporarily disable
+  # push:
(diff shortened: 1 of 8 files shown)


lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
mchades, line 1: we should also add the metric annotation to the rest method

Issue #8915, opened 2025-10-27
[Subtask] Polish code for lance table creation process.


yuqi1129:
### Describe the subtask

Some interfaces like `GenericTable` are unnecessary, and we can modify and add corresponding feature direclty in `Table`

### Parent issue

#8828

Issue #8921, opened 2025-10-27
[Subtask] Add UTs and ITs for lance table operatinons


yuqi1129:
### Describe the subtask

There are no UTs or ITs available to cover lance table operations.

### Parent issue

#8828

PR #8879, opened 2025-10-22
[#8838] feat(catalogs): Support create/load/list table operation for lance table.

yuqi1129:
### What changes were proposed in this pull request?

Add support create and load table operations for lance table.

### Why are the changes needed?

It's a need.

Fix: #8838 
Fix: #8837 

### Does this PR introduce _any_ user-facing change?

N/A.

### How was this patch tested?

Currently, I have only tested it locally 

```shell
➜  [/Users/yuqi/Downloads] curl -X POST -H "Accept: application/vnd.gravitino.v1+json" \
-H "Content-Type: application/json" -d '{
  "name": "lance_table14",
  "comment": "This is an example table",
  "columns": [
    {
      "name": "id",
      "type": "integer",
      "comment": "id column comment",
      "nullable": false,
      "autoIncrement": true,
      "defaultValue": {
        "type": "literal",
        "dataType": "integer",
        "value": "-1"
      }
    }
  ],
  "indexes": [
    {
      "indexType": "primary_key",
      "name": "PRIMARY",
      "fieldNames": [["id"]]
    }
  ],
  "properties": {
    "format": "lance",
    "location": "/tmp/lance_catalog/schema/lance_table14"
  }
}' http://localhost:8090/api/metalakes/test/catalogs/lance_catalog/schemas/schema/tables
{"code":0,"table":{"name":"lance_table14","comment":"This is an example table","columns":[{"name":"id","type":"integer","comment":"id column comment","nullable":false,"autoIncrement":true,"defaultValue":{"type":"literal","dataType":"integer","value":"-1"}}],"properties":{"format":"lance","location":"/tmp/lance_catalog/schema/lance_table14/"},"audit":{"creator":"anonymous","createTime":"2025-10-23T03:18:39.123151Z"},"distribution":{"strategy":"none","number":0,"funcArgs":[]},"sortOrders":[],"partitioning":[],"indexes":[{"indexType":"PRIMARY_KEY","name":"PRIMARY","fieldNames":[["id"]]}]}}
➜  [/Users/yuqi/Downloads] curl -X GET -H "Accept: application/vnd.gravitino.v1+json" \
-H "Content-Type: application/json" http://localhost:8090/api/metalakes/test/catalogs/lance_catalog/schemas/schema/tables/lance_table14
[...]

PR #8922, opened 2025-10-27
[#8915] improvment(catalogs): Polish code for PR #8879

yuqi1129:

This PR trys to resolve the comments that have not been addressed in #8879

It's an improvement.

Fix: #8915

N/A.

Test locally, and we will add ITs in https://github.com/apache/gravitino/issues/8921

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
@@ -698,11 +698,6 @@ public class TableOperationDispatcher extends OperationDispatcher implements Tab
   private boolean isManagedTable(NameIdentifier catalogIdent) {
     CatalogManager catalogManager = GravitinoEnv.getInstance().catalogManager();
     CatalogWrapper wrapper = catalogManager.loadCatalogAndWrap(catalogIdent);
-    if (wrapper == null || wrapper.catalog() == null) {
-      // Prevention for mis-configuration or in test env.
-      return false;
-    }
-
     Capability capability = wrapper.catalog().capability();
 
     CapabilityResult result = capability.managedStorage(Capability.Scope.TABLE);
--- a/core/src/test/java/org/apache/gravitino/hook/TestTableHookDispatcher.java
+++ b/core/src/test/java/org/apache/gravitino/hook/TestTableHookDispatcher.java
@@ -52,6 +52,7 @@ import org.apache.gravitino.catalog.TestOperationDispatcher;
 import org.apache.gravitino.catalog.TestTableOperationDispatcher;
 import org.apache.gravitino.connector.BaseCatalog;
 import org.apache.gravitino.connector.authorization.AuthorizationPlugin;
+import org.apache.gravitino.connector.capability.Capability;
 import org.apache.gravitino.lock.LockManager;
 import org.apache.gravitino.rel.Column;
 import org.apache.gravitino.rel.TableChange;
@@ -95,7 +96,13 @@ public class TestTableHookDispatcher extends TestOperationDispatcher {
     catalogManager = Mockito.mock(CatalogManager.class);
     FieldUtils.writeField(GravitinoEnv.getInstance(), "catalogManager", catalogManager, true);
     BaseCatalog catalog = Mockito.mock(BaseCatalog.class);
+    Mockito.when(catalog.capability()).thenReturn(Capability.DEFAULT);
+    CatalogManager.CatalogWrapper catalogWrapper =
+        Mockito.mock(CatalogManager.CatalogWrapper.class);
+    Mockito.when(catalogWrapper.catalog()).thenReturn(catalog);
+
     Mockito.when(catalogManager.loadCatalog(any())).thenReturn(catalog);
+    Mockito.when(catalogManager.loadCatalogAndWrap(any())).thenReturn(catalogWrapper);
     authorizationPlugin = Mockito.mock(AuthorizationPlugin.class);
     Mockito.when(catalog.getAuthorizationPlugin()).thenReturn(authorizationPlugin);
   }


jerryshao:
You'd better adding more UTs in the code, there's no UT in the previous code. You need to add them in each module you changed.

yuqi1129:
More tests added, and let's wait for #8923 to be merged.

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -99,6 +99,33 @@ default Map<String, String> properties() {
...
+   */
+  default String location() {
+    throw new UnsupportedOperationException("Table location is not supported.");
+  }
+
+  /**
+   * Indicates whether the table is external.
+   *
+   * @return true if the table is external, false otherwise
+   */
```
mchades, line 128: Can you explain in detail the meaning of `external`?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
```diff
@@ -117,15 +119,16 @@ public Table createTable(
...
     Map<String, String> storageProps =
         properties.entrySet().stream()
             .filter(e -> e.getKey().startsWith(LANCE_TABLE_STORAGE_OPTION_PREFIX))
             .collect(
                 Collectors.toMap(
                     e -> e.getKey().substring(LANCE_TABLE_STORAGE_OPTION_PREFIX.length()),
                     Map.Entry::getValue));
-    try (Dataset dataset =
+
+    try (Dataset ignored =
```
jerryshao, line 130: Why do we name it as `ignored`?

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -99,6 +99,33 @@ default Map<String, String> properties() {
...
+  default String format() {
+    throw new UnsupportedOperationException("Table format is not supported.");
+  }
+
+  /**
+   * Gets the location of the table.
+   *
+   * @return the location of the table
+   */
+  default String location() {
```
mchades, line 118: Does the table property also contain this?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
yuqi1129: It was never used in the following code exception for the try definition.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
```diff
@@ -206,17 +210,66 @@ public Table alterTable(NameIdentifier ident, TableChange... changes)
...
+                      .withColumns(tableEntity.columns())
+                      .withIndexes(
+                          ArrayUtils.addAll(
+                              entity.getIndexes(), addedIndexes.toArray(new Index[0])))
+                      .withDistribution(tableEntity.getDistribution())
+                      .withPartitions(tableEntity.getPartitions())
+                      .withSortOrder(tableEntity.getSortOrder())
+                      .withProperties(tableEntity.getProperties())
+                      .withComment(tableEntity.getComment())
+                      .build());
```
jerryshao, line 216: What is updated here?

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -99,6 +99,33 @@ default Map<String, String> properties() {
     return Collections.emptyMap();
   }
 
+  /**
+   * Formats the table as a string representation.
+   *
+   * @return the formatted string representation of the table
+   */
+  default String format() {
+    throw new UnsupportedOperationException("Table format is not supported.");
```
mchades, line 109: The same question: Does the table property also contain this?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
yuqi1129: Add index

api/src/main/java/org/apache/gravitino/rel/Table.java
yuqi1129: Yes, for the current implementation, we will store it in the table properties.

api/src/main/java/org/apache/gravitino/rel/Table.java
jerryshao, line 156: Did you remove the index Json serde in API module? It is not proper to put that in the API module.

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
```diff
@@ -295,8 +293,18 @@ public boolean deleteTable(NameIdentifier identifier) {
...
+        + deleteTableVersionByLegacyTimeline(legacyTimeline, limit);
+  }
+
+  @Monitored(
+      metricsSource = GRAVITINO_RELATIONAL_STORE_METRIC_NAME,
+      baseMetricName = "deleteTableVersionByLegacyTimeline")
+  public int deleteTableVersionByLegacyTimeline(Long legacyTimeline, int limit) {
+    return SessionUtils.doWithCommitAndFetchResult(
+        TableVersionMapper.class,
+        mapper -> mapper.deleteTableVersionByLegacyTimeline(legacyTimeline, limit));
```
jerryshao, line 313: You need to add tests to cover the changes.

jerryshao:
You need to add more tests for the code you introduced.

api/src/main/java/org/apache/gravitino/rel/Table.java
yuqi1129: I see and am going to do it in https://github.com/apache/gravitino/issues/8940

api/src/main/java/org/apache/gravitino/rel/Table.java
mchades: so the default implementation can be like `properties().get("location")` ?

yuqi1129: I'm afraid that we cannot add such an assumption. What if a user adds an property name `location` for a non-lakehouse table?

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
yuqi1129: Added.

yuqi1129:
> You need to add more tests for the code you introduced.

Added.

api/src/main/java/org/apache/gravitino/rel/Table.java
mchades: It seems duplicate with the `purgeTable` and `dropTable` APIs.

yuqi1129: I have no preference on this, as we haven't used it yet, and the concept is not so deterministic for a managed table. So adding or removing it is fine with me.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseTablePropertiesMetadata.java
```diff
@@ -37,11 +39,16 @@ public class GenericLakehouseTablePropertiesMetadata extends BasePropertiesMetad
...
             stringOptionalPropertyEntry(
-                LOCATION,
-                "The root directory of the lakehouse table.",
-                true /* immutable */,
+                LAKEHOUSE_LOCATION,
+                "The root directory of the lakehouse catalog.",
+                false /* immutable */,
                 null, /* defaultValue */
                 false /* hidden */),
+            stringRequiredPropertyEntry(
```
mchades: you can use `enumImmutablePropertyEntry` method

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/LakehouseCatalogOperations.java
```diff
@@ -22,4 +22,18 @@
...
+ * lakehouse catalog operations.
+ *
+ * <pre>
+ *    GenericLakehouseCatalog.createTable()
+ *       -> LakehouseCatalogOperations.createTable()
+ *         -> LanceTableOperations.createTable()
+ *         -> IcebergTableOperations.createTable()
+ *         -> DeltaTableOperations.createTable()
+ *         ...
+ * </pre>
```
mchades, line 37: What does this comment mean?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/LakehouseTableFormat.java
```diff
@@ -17,31 +17,25 @@
...
+public enum LakehouseTableFormat {
+  LANCE,
 
-  /**
-   * Formats the table as a string representation.
-   *
-   * @return the formatted string representation of the table
-   */
-  String format();
+  DELTA,
```
mchades, line 25: miss `HUDI` and `Paimon`?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
mchades: Why catch `NoSuchTableException` when getting schema?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/LakehouseCatalogOperations.java
yuqi1129: It means `GenericLakehouseCatalog.createTable()` wil call `LakehouseCatalogOperations.createTable()`, and `LakehouseCatalogOperations.createTable()` will call `Lance...` or `Iceberg` or `Delta` implementations.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/LakehouseTableFormat.java
yuqi1129: It's just for concept description, and we will add other lakehouse formats when we are going to support them.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
yuqi1129: Good catch, it's a mistake.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/utils/EntityConverter.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private static Column toColumn(ColumnEntity columnEntity) {
+    return GenericLakehouseColumn.builder()
+        .withName(columnEntity.name())
+        .withComment(columnEntity.comment())
+        .withAutoIncrement(columnEntity.autoIncrement())
+        .withNullable(columnEntity.nullable())
+        .withType(columnEntity.dataType())
+        .withDefaultValue(columnEntity.defaultValue())
+        .build();
```
mchades, line 40: Can you make `ColumnEntity` implement the `Column` API? This way, we won't need to make changes here anymore.

yuqi1129: I'm hesitant to do so, and the same goes for `TableEntity`. Currently, the table entity does not implement the `Table` interface. I'm wondering about the rationality of it, although it will simplify the code.

common/src/main/java/org/apache/gravitino/dto/util/DTOConverters.java
```diff
@@ -1138,6 +1138,24 @@ public Map<String, String> properties() {
...
         return tableDTO.auditInfo();
       }
+
+      @Override
+      public String format() {
+        if (properties() == null) {
+          return null;
+        }
+
+        return properties().get("format");
```
mchades: you said "we cannot add such an assumption" above

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
mchades: What is the difference between this code block and the one added above?

mchades: I think you can use `managedStorage` with table scope in `catalogCapability` to determine whether to return directly.

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
```diff
@@ -295,8 +299,18 @@ public boolean deleteTable(NameIdentifier identifier) {
       baseMetricName = "deleteTableMetasByLegacyTimeline")
   public int deleteTableMetasByLegacyTimeline(Long legacyTimeline, int limit) {
     return SessionUtils.doWithCommitAndFetchResult(
-        TableMetaMapper.class,
-        mapper -> mapper.deleteTableMetasByLegacyTimeline(legacyTimeline, limit));
+            TableMetaMapper.class,
+            mapper -> mapper.deleteTableMetasByLegacyTimeline(legacyTimeline, limit))
+        + deleteTableVersionByLegacyTimeline(legacyTimeline, limit);
```
mchades, line 304: You are using nested submit operation, will it work as expected?

core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
```diff
@@ -440,21 +421,15 @@ public static TablePO initializeTablePOWithVersion(
...
+      TablePO oldTablePO, TableEntity newTable, Long newSchemaId) {
     Long lastVersion;
     Long currentVersion;
-    if (needUpdateVersion) {
-      lastVersion = oldTablePO.getLastVersion() + 1;
-      currentVersion = lastVersion;
-    } else {
-      lastVersion = oldTablePO.getLastVersion();
-      currentVersion = oldTablePO.getCurrentVersion();
-    }
```
mchades, line 457: if we only update the table name, should we also update the table version?

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
yuqi1129: Yes, UT has confirm it.

core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
yuqi1129: Yes, currently, any update to a table will increase the table version.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
yuqi1129: It seems that currently, the relation catalog is not managed; changing it will introduce great modifications.

https://github.com/apache/gravitino/blob/60a4950259acb61bc1e6d1bb222c6f6f5d72c288/api/src/main/java/org/apache/gravitino/Catalog.java#L64-L79

core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
yuqi1129: The problem is that the data in `table_version_info` is tracked by the version, if we do not do this change, it will break the definition of the table.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
mchades: No, the lakehouse catalog has all managed:
https://github.com/apache/gravitino/blob/bc1b77a3bb357dd26bd64f025899b0069ece17b3/catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogCapability.java#L24-L28

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
mchades: If `mapper.deleteTableMetasByLegacyTimeline` succeeds but `deleteTableVersionByLegacyTimeline` fails, the former cannot be rolled back, because `deleteTableVersionByLegacyTimeline` has already closed the session.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/LakehouseCatalogOperations.java
mchades: Oh, they have different indentations in the comment, causing ambiguity

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
yuqi1129: I did notice it and hesitated to do it. I wonder Why doesn't the model and fileset use this mechanism, and then are hesitant to change it

core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
mchades: `table_version_info` does not include `table_name` field

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
mchades: because they are always managed, but some relational catalogs are not

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
yuqi1129: You are right, however, `deleteTableVersionByLegacyTimeline` is designed not to roll back. If it fails, it will try again later.

image

yuqi1129: If you do think it is a problem, we can change it.

core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
yuqi1129: I know, so what's the problem? According to the design, any changes to a table will then try to delete the old version of the data in `table_version_info` and then insert a new version of the data.

yuqi1129: Please see https://github.com/apache/gravitino/actions/runs/18906282300/job/53965239149?pr=8922, if we do not make this change,  repeatedly changing the name of a table will throw exceptions.

```
      Detail: Key (table_id, version, deleted_at)=(3718198477429741734, 2, 1761739099968) already exists.
```

Because the previous code will not increase the version.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
yuqi1129: fix

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -698,6 +698,11 @@ private List<ColumnEntity> toColumnEntities(Column[] columns, AuditInfo audit) {
   private boolean isManagedTable(NameIdentifier catalogIdent) {
     CatalogManager catalogManager = GravitinoEnv.getInstance().catalogManager();
     CatalogWrapper wrapper = catalogManager.loadCatalogAndWrap(catalogIdent);
+    if (wrapper == null || wrapper.catalog() == null) {
+      // Prevention for mis-configuration or in test env.
+      return false;
```
mchades: Should not return incorrect values for compatibility testing and misconfiguration.

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
mchades: Nested use of commit is a bad practice, we should change it.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
yuqi1129: It needs time to fix all affected tests one by one. This is not a technology problem but a time problem. A minor change may need a great time to fix related tests.

yuqi1129: fixed

Issue #8633, opened 2025-09-19
[Subtask] add authz for renaming table with new schema name


mchades:
### Describe the subtask

add authz for renaming table with new schema name

### Parent issue

#8375

hdygxsj:
Under the current authorization system, what permissions are required to perform this operation? Specifically, does it require ownership of the original schema and the CREATE TABLE privilege on the new schema? @jerqi

roryqi:
> Under the current authorization system, what permissions are required to perform this operation? Specifically, does it require ownership of the original schema and the CREATE TABLE privilege on the new schema? [@jerqi](https://github.com/jerqi)

It requires ownership of the original schema and the `CREATE_TABLE` privilege on the new schema.

roryqi:
Pending, because the underlying system doesn't this feature now.

PR #8945, opened 2025-10-28
[#8633] feat(authz): Add authz for renaming table with new schema name

hdygxsj:

Add authz for renaming table with new schema name

Fix: #8633 

None

org.apache.gravitino.client.integration.test.authorization.TableAuthorizationIT#testAlterTable

### DIFF
--- a/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/TableAuthorizationIT.java
+++ b/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/TableAuthorizationIT.java
@@ -255,6 +255,25 @@ public class TableAuthorizationIT extends BaseRestApiAuthorizationIT {
         Owner.Type.USER);
     tableCatalogNormalUser.alterTable(
         NameIdentifier.of(SCHEMA, "table1"), TableChange.setProperty("key", "value"));
+    String newSchema = "newSchema";
+    Catalog catalogLoadedByAdmin = gravitinoMetalake.loadCatalog(CATALOG);
+    catalogLoadedByAdmin.asSchemas().createSchema(newSchema, "", new HashMap<>());
+    gravitinoMetalake
+        .loadCatalog(CATALOG)
+        .asTableCatalog()
+        .createTable(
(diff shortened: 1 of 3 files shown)


hdygxsj:
It seems that renaming a table via RENAME TABLE in the Java client doesn't take effect for modifying the schema. Could you please take a look? @jerqi

roryqi:
@mchades Could u take a look?

mchades:
> It seems that renaming a table via RENAME TABLE in the Java client doesn't take effect for modifying the schema. 

This is because there is no catalog that has implemented this feature yet. Which catalog did you use for testing, and we can prioritize the implementation for this catalog.

hdygxsj:
> > It seems that renaming a table via RENAME TABLE in the Java client doesn't take effect for modifying the schema.
> 
> This is because there is no catalog that has implemented this feature yet. Which catalog did you use for testing, and we can prioritize the implementation for this catalog.

hive catalog

hdygxsj:
> > It seems that renaming a table via RENAME TABLE in the Java client doesn't take effect for modifying the schema.
> 
> This is because there is no catalog that has implemented this feature yet. Which catalog did you use for testing, and we can prioritize the implementation for this catalog.

@mchades There's another issue: in the Java client, if I only modify the schema name without modify the table name, it will throw an exception on the client side.

mchades:
> > > It seems that renaming a table via RENAME TABLE in the Java client doesn't take effect for modifying the schema.
> > 
> > 
> > This is because there is no catalog that has implemented this feature yet. Which catalog did you use for testing, and we can prioritize the implementation for this catalog.
> 
> There's another issue: in the Java client, if I only modify the schema name without modify the table name, it will throw an exception on the client side.

Do you want to fix it? Or I can take a look as well, whatever you prefer.

roryqi:
@hdygxsj Maybe you can continue this pull request. Iceberg has supported the renaming operations.

hdygxsj:
> @hdygxsj Maybe you can continue this pull request. Iceberg has supported the renaming operations.

OK

Issue #8948, opened 2025-10-28
[Improvement] Lazy load Gravitino Client in GVFS


youngyjd:
### What would you like to be improved?

We want to lazy load Gravitino Client in GVFS so fallback can capture the RestException and kicks in. 

Currently GVFS will fail at initialization because Gravitino Client is initialized at GVFS initialization phase, which means fallback hook will never have a chance to work as GVFS errored out before any HDFS API is reached.

PR #8950, opened 2025-10-28
[#8948]Improve(gvfs-java) Lazy load gravitino client in gvfs

youngyjd:

- Lazy load GravitinoClient and FilesetMetadataCache in BaseGVFSOperations using double-checked locking pattern
- Client and cache are now created on first filesystem operation instead of during construction
- Fixed resource ownership: FilesetMetadataCache no longer closes the shared GravitinoClient

- Improves startup performance by deferring expensive client initialization
- Reduces resource usage when filesystem is created but never used
- Better resource management with clear ownership (only BaseGVFSOperations closes the client)

Fix: #8948

Yes, minor behavioral change:
Configuration validation and connection errors are now deferred until the first filesystem operation (e.g., exists(), listStatus()) instead of being thrown during filesystem initialization.

unit tests added and updated

### DIFF
--- a/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/BaseGVFSOperations.java
+++ b/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/BaseGVFSOperations.java
@@ -198,11 +198,10 @@ public abstract class BaseGVFSOperations implements Closeable {
     if (filesetMetadataCache == null) {
       synchronized (filesetMetadataCacheLock) {
         if (filesetMetadataCache == null) {
-          if (!enableFilesetMetadataCache) {
-            this.filesetMetadataCache = Optional.empty();
-          } else {
-            this.filesetMetadataCache = Optional.of(new FilesetMetadataCache(getGravitinoClient()));
-          }
+          this.filesetMetadataCache =
+              enableFilesetMetadataCache
+                  ? Optional.of(new FilesetMetadataCache(getGravitinoClient()))
+                  : Optional.empty();
         }
       }
     }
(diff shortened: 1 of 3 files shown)


clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/TestGvfsBase.java
```diff
@@ -290,6 +290,12 @@ public void testRequestHeaders()
                 .withBody(getJsonString(new VersionResponse(Version.getCurrentVersionDTO()))));
 
     try (FileSystem fs = new Path("gvfs://fileset/").getFileSystem(configuration)) {
+      // Trigger lazy initialization by accessing a path
+      try {
+        fs.exists(new Path("gvfs://fileset/catalog/schema/fileset/file.txt"));
+      } catch (Exception e) {
+        // Ignore errors, we just want to trigger client initialization
```
mchades: Should we throw the exception?

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/BaseGVFSOperations.java
```diff
@@ -187,6 +188,27 @@ protected BaseGVFSOperations(Configuration configuration) {
...
+   */
+  @VisibleForTesting
+  protected Optional<FilesetMetadataCache> getFilesetMetadataCache() {
+    if (filesetMetadataCache == null) {
+      synchronized (filesetMetadataCacheLock) {
+        if (filesetMetadataCache == null) {
+          if (!enableFilesetMetadataCache) {
+            this.filesetMetadataCache = Optional.empty();
+          } else {
+            this.filesetMetadataCache = Optional.of(new FilesetMetadataCache(getGravitinoClient()));
```
mchades: You can simplify it using `enableFilesetMetadataCache ? xxx : xxx`

Issue #8957, opened 2025-10-29
[Improvement] Add support for more text types for mysql catalog


predator4ann:
### Background
MySQL supports multiple text types, including `TEXT`, `TINYTEXT`, `MEDIUMTEXT`, and `LONGTEXT`, each with different maximum length limits:
- `TINYTEXT`: Maximum 255 bytes
- `TEXT`: Maximum 65,535 bytes (64 KB)
- `MEDIUMTEXT`: Maximum 16,777,215 bytes (16 MB)
- `LONGTEXT`: Maximum 4,294,967,295 bytes (4 GB)

Currently, Gravitino's MySQL type converter (`MysqlTypeConverter`) only supports the `TEXT` type, but not `TINYTEXT`, `MEDIUMTEXT`, and `LONGTEXT`. When reading table schemas from MySQL databases that contain these types, they are recognized as `ExternalType` instead of the standard `StringType`, which may cause type compatibility issues.

### How should we improve?

`TINYTEXT`, `MEDIUMTEXT`, and `LONGTEXT` should be mapped to Gravitino's `StringType` just like `TEXT`, since they are essentially variable-length string types with different storage capacities.

predator4ann:
After verification, the data can be read normally in the current state, so it will be closed first.

PR #8958, opened 2025-10-29
[#8957] improvement(catalog-mysql): support more mysql text types

predator4ann:

Fix: #8957

No

UT already added.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
+++ b/core/src/test/java/org/apache/gravitino/cache/TestCacheConfig.java
@@ -147,7 +147,7 @@ public class TestCacheConfig {
               .build();
       cache.put(
           EntityCacheRelationKey.of(
-              NameIdentifier.of(new String[] {"metalake1", "catalog1", "schema1", filesetName}),
+              NameIdentifier.of("metalake1", "catalog1", "schema1", filesetName),
               Entity.EntityType.FILESET),
           List.of(fileset));
     }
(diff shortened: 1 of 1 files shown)


predator4ann:
@mchades PTAL when you have time, thanks.

catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/converter/MysqlTypeConverter.java
```diff
@@ -42,6 +42,9 @@ public class MysqlTypeConverter extends JdbcTypeConverter {
   static final String CHAR = "char";
   static final String BINARY = "binary";
   static final String DATETIME = "datetime";
+  static final String TINYTEXT = "tinytext";
+  static final String MEDIUMTEXT = "mediumtext";
+  static final String LONGTEXT = "longtext";
```
mchades, line 47: you also need to update the relative user doc

catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/converter/MysqlTypeConverter.java
```diff
@@ -97,6 +100,9 @@ public Type toGravitino(JdbcTypeBean typeBean) {
       case CHAR:
         return Types.FixedCharType.of(typeBean.getColumnSize());
       case TEXT:
+      case TINYTEXT:
+      case MEDIUMTEXT:
+      case LONGTEXT:
         return Types.StringType.get();
```
mchades, line 106: These four types will cast to the same type `String`, will this cause user confusion?

catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/converter/MysqlTypeConverter.java
predator4ann: got it

predator4ann:
After verification, the data can be read normally in the current state, so it will be closed first.

Issue #8940, opened 2025-10-28
[Subtask] Support serialization and deserialization for table index, partition, sort order and distribution information.


yuqi1129:
### Describe the subtask

Currently, Gravitino's generic lakehouse table only supports storing `index` information to the entity store and left the others unhandled. We need to support it. 

### Parent issue

#8828

PR #8975, opened 2025-10-30
[#8940] improvement(core): Support serialize/deserialize for distributionImpl and sortOrderImpl

yuqi1129:

- Add serializer and deserializer for class `DistributionImpl`  and `SortOrderImpl`
- Optimize sortOrderImpl serialization.

It's a improvment.

Fix: #8940 

N/A.

UTs and ITs.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/rel/expressions/distributions/Distributions.java
+++ b/api/src/main/java/org/apache/gravitino/rel/expressions/distributions/Distributions.java
@@ -242,7 +242,7 @@ public class Distributions {
        *
        * @return The created distribution.
        */
-      public DistributionImpl build() {
+      public Distribution build() {
         return new DistributionImpl(strategy, number, expressions);
       }
     }
--- a/api/src/main/java/org/apache/gravitino/rel/indexes/Indexes.java
+++ b/api/src/main/java/org/apache/gravitino/rel/indexes/Indexes.java
@@ -197,7 +197,7 @@ public class Indexes {
        *
        * @return The new instance.
        */
-      public IndexImpl build() {
+      public Index build() {
         return new IndexImpl(indexType, name, fieldNames);
       }
     }
--- a/common/src/test/java/org/apache/gravitino/json/TestSerializer.java
+++ b/common/src/test/java/org/apache/gravitino/json/TestSerializer.java
@@ -74,7 +74,9 @@ public class TestSerializer {
     actualJson = JsonUtils.anyFieldMapper().writeValueAsString(DTOConverters.toDTO(distribution));
     expectedJson =
         """
-          {"strategy":"even","number":10,"funcArgs":[{"type":"function","funcName":"bucket","funcArgs":[{"type":"literal","dataType":"integer","value":"10"},{"type":"field","fieldName":["col_1"]}]}]}""";
+          {"strategy":"even","number":10,"funcArgs":[{"type":"function","funcName":"bucket",\
+          "funcArgs":[{"type":"literal","dataType":"integer","value":"10"},{"type":"field",\
+          "fieldName":["col_1"]}]}]}""";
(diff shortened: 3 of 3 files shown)


common/src/main/java/org/apache/gravitino/json/JsonUtils.java
```diff
@@ -479,6 +502,85 @@ private static void writeFunctionArg(FunctionArg arg, JsonGenerator gen) throws
     gen.writeEndObject();
   }
 
+  private static void writeExpression(Expression expression, JsonGenerator gen) throws IOException {
```
jerryshao: I guess the implementation is quite similar to `xxxDTO` serde, maybe we can reuse the code. Otherwise, we have to change twice.

Another solution is that when we store these into the database, we can convert the `xxxImpl` to `xxxDTO`, and then using the existing serde method to do the serialization and deserialization. With this, we don't have to implement the serde method for these `xxxImpl`.

What do you think @yuqi1129 @mchades ?

yuqi1129: In the first edition, I have just converted Impl to DTO and reused the DTO serializer and deserializer to handle it. However, since DTO and Impl play different roles, I chose to create new serializers finally. Theoretically, both are okay.

jerryshao: Let's do the second one if there's no objection, cc @mchades .

api/src/main/java/org/apache/gravitino/rel/expressions/distributions/Distributions.java
```diff
@@ -242,7 +242,7 @@ public Builder withExpressions(Expression[] expressions) {
        *
        * @return The created distribution.
        */
-      public Distribution build() {
+      public DistributionImpl build() {
```
jerryshao: We'd better not change the signature of the method.

api/src/main/java/org/apache/gravitino/rel/indexes/Indexes.java
```diff
@@ -260,7 +197,7 @@ public Indexes.IndexImpl.Builder withFieldNames(String[][] fieldNames) {
        *
        * @return The new instance.
        */
-      public Index build() {
+      public IndexImpl build() {
```
jerryshao: Also here.

common/src/test/java/org/apache/gravitino/json/TestSerializer.java
```diff
@@ -0,0 +1,180 @@
...
+        (DistributionImpl)
+            Distributions.of(
+                Strategy.EVEN,
+                10,
+                FunctionExpression.of(
+                    "bucket", Literals.integerLiteral(10), NamedReference.field("col_1")));
+    actualJson = JsonUtils.anyFieldMapper().writeValueAsString(DTOConverters.toDTO(distribution));
+    expectedJson =
+        """
+          {"strategy":"even","number":10,"funcArgs":[{"type":"function","funcName":"bucket","funcArgs":[{"type":"literal","dataType":"integer","value":"10"},{"type":"field","fieldName":["col_1"]}]}]}""";
```
jerryshao: This is too long, make sure to not exceed 100 characters.

api/src/main/java/org/apache/gravitino/rel/expressions/distributions/Distributions.java
yuqi1129: I see.

common/src/test/java/org/apache/gravitino/json/TestSerializer.java
yuqi1129: fixed

jerryshao (approved):

Issue #8016, opened 2025-08-11
[Improvement] Potential null pointer exception when validating column with null default values in TableOpsUtils.java


justinmclean:
### What would you like to be improved?

Error exists in catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/utils/TableOpsUtils.java

Test that shows the issue:
```
  @Test
  void testAddColumnWithNullDefaultValue() {
    assertTableChange(
        addColumn(
            getFieldName("col_5"), IntegerType.get(), (Expression) null),
        AddColumn.class,
        schemaChange -> {
          AddColumn addColumn = (AddColumn) schemaChange;
          assertEquals("col_5", addColumn.fieldName());
        });
  }
```

### How should we improve?

Make the test pass

sjleeJ:
Hi, I'm interested in this issue and am currently working on it. Is that okay?

justinmclean:
That's fine. If you have any questions, please ask.

justinmclean:
@sjleeJ are you still working on this?

sjleeJ:
@justinmclean I'm working on it and will upload it by September 6th.

pandeysambhi:
@justinmclean This looks like not being worked upon, let me know if it can be assigned to me

justinmclean:
@pandeysambhi you can take this

pandeysambhi:
@justinmclean I do not see a fieldName() method in AddColumn class. should the above mentioned test be updated to addColumn.getFieldNames()[0]

pandeysambhi:
@justinmclean let me know if the above sounds good when you  get a chance to take a look

PR #8996, opened 2025-11-02
[#8061]improvement(tableOpsUtils): NPE in tableOpsUtils while validating column with null default values

pandeysambhi:
…fault values

Add a check for null as a possible default value in checkColumnDefaultvalues

There is a potential NullPointerException when validating columns with null default values in checkColumnDefaultValues in tableOpsUtils.

Fix: (#8016 )

NO

unit testing

### DIFF
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/utils/TableOpsUtils.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/utils/TableOpsUtils.java
@@ -129,7 +129,7 @@ public class TableOpsUtils {
 
   private static void checkColumnDefaultValue(String fieldName, Expression defaultValue) {
     Preconditions.checkArgument(
-        defaultValue.equals(Column.DEFAULT_VALUE_NOT_SET),
+        defaultValue == null || defaultValue.equals(Column.DEFAULT_VALUE_NOT_SET),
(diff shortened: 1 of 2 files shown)


jerryshao:
@FANNG1 can you please review this PR?

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/utils/TableOpsUtils.java
```diff
@@ -129,7 +129,7 @@ public static SchemaChange buildSchemaChange(TableChange tableChange)
 
   private static void checkColumnDefaultValue(String fieldName, Expression defaultValue) {
     Preconditions.checkArgument(
```
FANNG1, line 131: `defaultValue` seems couldn't be null because if user doesn't specify default value, it would be `DEFAULT_VALUE_NOT_SET`.  could you change the code like below to make the code more robust？
```
Column.DEFAULT_VALUE_NOT_SET.equals(defaultValue)
```

pandeysambhi: @FANNG1 the check you are asking for is already in place , though not written in the conventional Constant first check. The issue mentioned that running the test would give the error that needs to be fixed, and as per the test, there is a NPE as the default value is null, hence added the check for it

Issue #8304, opened 2025-08-27
[Improvement] Possible NPE in LineageService.java


justinmclean:
### What would you like to be improved?

In lineage/src/main/java/org/apache/gravitino/lineage/LineageService.java dispatchLineageEvent could throw a NPE 

### How should we improve?

Add a null check and return false.

Here's a unit test to help:
```
  @Test
  public void testDispatchLineageEventWithoutInitialization() {
    LineageService service = new LineageService();
    Assertions.assertFalse(service.dispatchLineageEvent(null));
  }
```

cheshireHYUN:
I'd like to work on this issue, is that okay?

justinmclean:
sure it's all yours

cheshireHYUN:
Thanks! I'm working on a fix and will submit a PR shortly.

pandeysambhi:
@justinmclean This looked like not being worked upon, so I went ahead and opened a PR to close this.

justinmclean:
Issue has been fixed

PR #8997, opened 2025-11-02
[#8304] fix: Possible NPE in LineageService

pandeysambhi:
Add a null check for the argument in dispatchLineageEvent and return false if it is null.

 The method dispatchLineageEvent can throw a NullPointerException if called without proper initialization or with a null argument.

Fix: (#8304 )

no

unit testing


justinmclean (approved):
thanks for adding teh check

Issue #9002, opened 2025-11-02
[Improvement] Optimize HiveBackendProxy.resetIcebergHiveClientPool() to prevent memory leaks.


sunxiaojian:
### What would you like to be improved?

Optimize HiveBackendProxy.resetIcebergHiveClientPool() to prevent memory leaks.

PR #9003, opened 2025-11-02
[#9002] Improvement(iceberg): Optimize HiveBackendProxy.resetIcebergHiveClientPool() to prevent memory leaks. 

sunxiaojian:

Optimize HiveBackendProxy.resetIcebergHiveClientPool() to prevent memory leaks. 

Fix: #9002

N/A

N/A

### DIFF
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java
@@ -269,15 +269,14 @@ public class PaimonCatalogOperations implements CatalogOperations, SupportsSchem
   public NameIdentifier[] listTables(Namespace namespace) throws NoSuchSchemaException {
     String[] levels = namespace.levels();
     NameIdentifier schemaIdentifier = NameIdentifier.of(levels[levels.length - 1]);
-    if (!schemaExists(schemaIdentifier)) {
-      throw new NoSuchSchemaException(NO_SUCH_SCHEMA_EXCEPTION, namespace.toString());
-    }
+
     List<String> tables;
     try {
       tables = paimonCatalogOps.listTables(schemaIdentifier.name());
     } catch (Catalog.DatabaseNotExistException e) {
-      throw new NoSuchSchemaException(NO_SUCH_SCHEMA_EXCEPTION, namespace.toString());
+      throw new NoSuchSchemaException(e, NO_SUCH_SCHEMA_EXCEPTION, namespace.toString());
     }
+
     return tables.stream()
         .map(
             tableIdentifier ->
--- a/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/PartitionAuthorizationIT.java
+++ b/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/PartitionAuthorizationIT.java
@@ -1,0 +1,253 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.client.integration.test.authorization;
+
+import static org.apache.gravitino.client.integration.test.StatisticIT.TABLE_COMMENT;
+import static org.junit.Assert.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.gravitino.Catalog;
+import org.apache.gravitino.MetadataObject;
+import org.apache.gravitino.MetadataObjects;
+import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.authorization.Privileges;
+import org.apache.gravitino.authorization.SecurableObject;
+import org.apache.gravitino.authorization.SecurableObjects;
+import org.apache.gravitino.client.GravitinoMetalake;
+import org.apache.gravitino.exceptions.ForbiddenException;
+import org.apache.gravitino.integration.test.container.ContainerSuite;
+import org.apache.gravitino.integration.test.container.HiveContainer;
+import org.apache.gravitino.rel.Column;
+import org.apache.gravitino.rel.SupportsPartitions;
+import org.apache.gravitino.rel.Table;
+import org.apache.gravitino.rel.TableCatalog;
+import org.apache.gravitino.rel.expressions.literals.Literal;
+import org.apache.gravitino.rel.expressions.literals.Literals;
+import org.apache.gravitino.rel.expressions.transforms.Transform;
+import org.apache.gravitino.rel.expressions.transforms.Transforms;
+import org.apache.gravitino.rel.partitions.IdentityPartition;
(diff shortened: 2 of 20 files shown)


sunxiaojian:
@FANNG1 PTAL

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
```diff
@@ -110,11 +112,20 @@ private ClientPool<IMetaStoreClient, TException> resetIcebergHiveClientPool()
     final Field m = HiveCatalog.class.getDeclaredField("clients");
     m.setAccessible(true);
 
-    // TODO: we need to close the original client pool and thread pool, or it will cause memory
-    //  leak.
-    ClientPool<IMetaStoreClient, TException> newClientPool =
+    // Get and close the old client pool before replacing it
```
yuqi1129: Hi, I noticed that the following code will only try to close the Hive client pool and is unable to free the thread pool in 
`CachedClientPool#Cache`, can you also try to handle it? Please let me know if there is any problem with it. In our previous attempt, it took time and effort to close the thread pool.

sunxiaojian: @yuqi1129 hi, do you mean the shutdown of this thread pool?https://github.com/apache/gravitino/blob/35a2dc1183ffd8dcee4b7bc6d99c8fb962720dab/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/utils/IcebergHiveCachedClientPool.java#L264

yuqi1129: No, it's the thread pool in Caffeine cache, see:
https://github.com/apache/gravitino/blob/35a2dc1183ffd8dcee4b7bc6d99c8fb962720dab/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/utils/IcebergHiveCachedClientPool.java#L88

sunxiaojian: @yuqi1129  When creating a 'clientPoolCache', it uses the custom thread pool 'scheduledExecutionService' from IcebergHiveCachedClientPool. When calling `IcebergHiveCachedClientPool.close()', the thread pool will be directly shut down.Is there any logic I might have overlooked?
https://github.com/apache/gravitino/blob/35a2dc1183ffd8dcee4b7bc6d99c8fb962720dab/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/utils/IcebergHiveCachedClientPool.java#L143

yuqi1129: Sorry, I may not have described it clearly.  Please note the original value of `HiveCatalog#clients`, that is an instance of `CachedClientPool`, and there is a Caffeine cache instance for `CachedClientPool`, the cache has been initialized with a thread pool, please see: `CachedClientPool#init`, it's this pool that can't be closed properly.


```java
  // HiveCatalog
  public void initialize(String inputName, Map<String, String> properties) {
    this.catalogProperties = ImmutableMap.copyOf(properties);
     // ....
    this.clients = new CachedClientPool(this.conf, properties);
  }
```


```java
  // CachedClientPool
  private synchronized void init() {
    if (clientPoolCache == null) {
      clientPoolCache = Caffeine.newBuilder().expireAfterAccess(this.evictionInterval, TimeUnit.MILLISECONDS).removalListener((ignored, value, cause) -> {
        ((HiveClientPool)value).close();
      }).scheduler(Scheduler.forScheduledExecutorService(ThreadPools.newScheduledPool("hive-metastore-cleaner", 1))).build();
    }

  }
```

sunxiaojian: OK, I got it, but CachedClientPool doesn't provide a close method, so it might be difficult to handle. let me try it.

yuqi1129: Yes, that is the exact cause that makes it as the to-do item here, however, we have clear it use reflection and it's somewhat not very elegant. 
please see:https://github.com/apache/gravitino/blob/35a2dc1183ffd8dcee4b7bc6d99c8fb962720dab/catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalogOperations.java#L127

sunxiaojian: @yuqi1129 I tried handling it and added a unit test to verify it.  PTAL

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
```diff
@@ -110,14 +115,228 @@ private ClientPool<IMetaStoreClient, TException> resetIcebergHiveClientPool()
...
+   */
+  @VisibleForTesting
+  ScheduledExecutorService extractSchedulerExecutor(Object cache) {
+    try {
+      // Step 1: Unwrap cache if it's a wrapper (e.g., BoundedLocalManualCache)
+      Object cacheImpl = cache;
+      try {
+        Field cacheField = cache.getClass().getDeclaredField("cache");
+        cacheField.setAccessible(true);
+        Object innerCache = cacheField.get(cache);
```
yuqi1129: Please use `FieldUtis.read(xx)` to simplify the code.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
```diff
@@ -110,14 +115,228 @@ private ClientPool<IMetaStoreClient, TException> resetIcebergHiveClientPool()
...
+
+  /**
+   * Gets the Pacer field from cache implementation. Pacer is a package-private class in Caffeine
+   * that manages the scheduling of maintenance tasks.
+   *
+   * @param cacheImpl The cache implementation
+   * @return The Pacer object, or null if not found
+   */
+  @VisibleForTesting
+  Object getPacerFromCache(Object cacheImpl) throws Exception {
```
yuqi1129: Could you please extract those methods as utils, I believe they will be used in other places.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
```diff
@@ -110,14 +115,228 @@ private ClientPool<IMetaStoreClient, TException> resetIcebergHiveClientPool()
...
+   * ScheduledExecutorService
+   *
+   * @param cache The Caffeine cache instance.
+   * @return The ScheduledExecutorService if found, null otherwise.
+   */
+  @VisibleForTesting
+  ScheduledExecutorService extractSchedulerExecutor(Object cache) {
+    try {
+      // Step 1: Unwrap cache if it's a wrapper (e.g., BoundedLocalManualCache)
+      Object cacheImpl = cache;
```
yuqi1129: Can this method work for different cache implementations besides `BoundedLocalManualCache`?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
sunxiaojian: > Please use `FieldUtis.read(xx)` to simplify the code.

fixed

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
sunxiaojian: fixed

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/authentication/kerberos/HiveBackendProxy.java
sunxiaojian: yes, they all have a cache field pointing to BoundedLocalCache.

yuqi1129: OK

yuqi1129 (approved):
LGTM

PR #9007, opened 2025-11-03
[#8921] improvement(lance-rest): Add ITs for lance table operations

yuqi1129:

Add ITs for lance rest service and test lance table operations.

To make code robust

Fix: #8921 

N/A.

It's self a test.


lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -118,17 +118,15 @@ public Response createEmptyTable(
       @PathParam("id") String tableId,
       @QueryParam("mode") @DefaultValue("create") String mode, // create, exist_ok, overwrite
```
jerryshao: There's no `mode` in creating empty table.

yuqi1129:
This has been replaced with #9016

PR #9008, opened 2025-11-03
[#9009]improvement(lance-rest-server): add ITs for LRS namespace operations

mchades:

add ITs for LRS namespace operations

Fix: #9009 

no

CI pass

### DIFF
--- a/integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
+++ b/integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
@@ -21,7 +21,6 @@ package org.apache.gravitino.integration.test;
 import static com.google.common.util.concurrent.Uninterruptibles.sleepUninterruptibly;
 import static org.apache.gravitino.lance.common.config.LanceConfig.GRAVITINO_NAMESPACE_BACKEND;
 import static org.apache.gravitino.lance.common.config.LanceConfig.LANCE_CONFIG_PREFIX;
-import static org.apache.gravitino.lance.common.config.LanceConfig.METALAKE_NAME;
 import static org.apache.gravitino.lance.common.config.LanceConfig.NAMESPACE_BACKEND;
 import static org.apache.gravitino.lance.common.config.LanceConfig.NAMESPACE_BACKEND_URI;
 
@@ -48,7 +47,6 @@ import org.apache.gravitino.auth.AuthenticatorType;
 import org.apache.gravitino.auxiliary.AuxiliaryServiceManager;
 import org.apache.gravitino.client.HTTPClient;
 import org.apache.gravitino.client.RESTClient;
-import org.apache.gravitino.integration.test.util.GravitinoITUtils;
 import org.apache.gravitino.integration.test.util.HttpUtils;
 import org.apache.gravitino.integration.test.util.ITUtils;
 import org.apache.gravitino.integration.test.util.KerberosProviderHelper;
@@ -298,9 +296,6 @@ public class MiniGravitino {
                   GravitinoServer.WEBSERVER_CONF_PREFIX
                       + JettyServerConfig.WEBSERVER_HTTP_PORT.getKey()));
       customConfigs.put(LANCE_CONFIG_PREFIX + NAMESPACE_BACKEND_URI.getKey(), gravitinoUri);
-
-      String metalakeName = GravitinoITUtils.genRandomName("LanceRESTService_metalake");
-      customConfigs.put(LANCE_CONFIG_PREFIX + METALAKE_NAME.getKey(), metalakeName);
     }
     return ImmutableMap.copyOf(customConfigs);
   }
--- a/integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
+++ b/integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
@@ -343,6 +343,11 @@ public class BaseIT {
 
     serverConfig = new ServerConfig();
     customConfigs.put(ENTITY_RELATIONAL_JDBC_BACKEND_PATH.getKey(), file.getAbsolutePath());
+    if (!ignoreLanceAuxRestService) {
+      customConfigs.put(
+          LANCE_CONFIG_PREFIX + METALAKE_NAME.getKey(),
+          GravitinoITUtils.genRandomName("LanceRESTService_metalake"));
+    }
     if (testMode != null && testMode.equals(ITUtils.EMBEDDED_TEST_MODE)) {
       MiniGravitinoContext context =
           new MiniGravitinoContext(
@@ -351,11 +356,6 @@ public class BaseIT {
       miniGravitino.start();
       serverConfig = miniGravitino.getServerConfig();
     } else {
-      if (!ignoreLanceAuxRestService) {
-        customConfigs.put(
-            LANCE_CONFIG_PREFIX + METALAKE_NAME.getKey(),
-            GravitinoITUtils.genRandomName("LanceRESTService_metalake"));
-      }
       rewriteGravitinoServerConfig();
       serverConfig.loadFromFile(GravitinoServer.CONF_FILE);
 
--- a/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
+++ b/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
@@ -78,8 +78,6 @@ public class LanceRESTServiceIT extends BaseIT {
     this.ns = LanceNamespaces.connect("rest", props, null, allocator);
 
     this.tempDir = Files.createTempDirectory("test_lance_rest_service_it_");
-    // deleteOnExit
-    tempDir.toFile().deleteOnExit();
   }
 
   @AfterAll
@@ -91,7 +89,6 @@ public class LanceRESTServiceIT extends BaseIT {
   @AfterEach
   public void clearMetalake() {
     Arrays.stream(metalake.listCatalogs()).forEach(c -> metalake.dropCatalog(c, true));
-    tempDir.toFile().deleteOnExit();
   }
 
   @Test


integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
```diff
@@ -78,11 +84,25 @@ public MiniGravitino(MiniGravitinoContext context) throws IOException {
...
-    properties.remove(
-        AuxiliaryServiceManager.GRAVITINO_AUX_SERVICE_PREFIX
-            + AuxiliaryServiceManager.AUX_SERVICE_NAMES);
+  private void removeAuxRestConfiguration(Properties properties, String serviceToRemove) {
+    String value =
+        properties.getProperty(
+            AuxiliaryServiceManager.GRAVITINO_AUX_SERVICE_PREFIX
+                + AuxiliaryServiceManager.AUX_SERVICE_NAMES);
+    if (value != null && !value.isEmpty()) {
+      List<String> serviceNames = COMMA.splitToList(value);
```
jerryshao, line 93: Will this `splitToList` also do the `trim` to avoid some invisible characters like white space?

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
```diff
@@ -78,11 +84,25 @@ public MiniGravitino(MiniGravitinoContext context) throws IOException {
...
+      List<String> updatedServiceNames = new ArrayList<>();
+      for (String serviceName : serviceNames) {
+        if (!serviceName.equalsIgnoreCase(serviceToRemove)) {
+          updatedServiceNames.add(serviceName);
+        }
+      }
+      String updatedValue = String.join(",", updatedServiceNames);
```
jerryshao, line 101: If the `updatedServiceNames` is empty, shall we remove the config?

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
```diff
@@ -230,19 +259,40 @@ Map<String, String> getIcebergRestServiceConfigs() throws IOException {
     return customConfigs;
   }
 
-  private Map<String, String> getLanceRestServiceConfigs() throws IOException {
+  private Map<String, String> getLanceRestServiceConfigs(Map<String, String> configMap)
+      throws IOException {
+    if (context.ignoreLanceAuxRestService) {
+      return new HashMap<>();
+    }
```
jerryshao, line 273: You can use `Collections.emptyMap()`.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
```diff
@@ -230,19 +259,40 @@ Map<String, String> getIcebergRestServiceConfigs() throws IOException {
...
+          String.format(
+              "http://%s:%s",
+              "localhost",
+              configMap.get(
+                  GravitinoServer.WEBSERVER_CONF_PREFIX
+                      + JettyServerConfig.WEBSERVER_HTTP_PORT.getKey()));
+      customConfigs.put(LANCE_CONFIG_PREFIX + NAMESPACE_BACKEND_URI.getKey(), gravitinoUri);
+
+      String metalakeName = GravitinoITUtils.genRandomName("LanceRESTService_metalake");
+      customConfigs.put(LANCE_CONFIG_PREFIX + METALAKE_NAME.getKey(), metalakeName);
```
jerryshao: Shall we create the metalake here before using Lance namespace server?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -356,6 +358,12 @@ private CreateNamespaceResponse createOrUpdateCatalog(
     }
   }
 
+  private Map<String, String> removeInUseProperty(Map<String, String> properties) {
+    return properties.entrySet().stream()
+        .filter(e -> !e.getKey().equalsIgnoreCase(Catalog.PROPERTY_IN_USE))
+        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+  }
+
```
jerryshao, line 366: Why do we remove this?

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -0,0 +1,377 @@
...
+    Assertions.assertTrue(
+        exception
+            .getErrorResponse()
+            .get()
+            .getError()
+            .contains("Expected at most 2-level and at least 1-level namespace"));
+    Assertions.assertEquals(
+        IllegalArgumentException.class.getSimpleName(),
+        exception.getErrorResponse().get().getType());
+  }
```
jerryshao, line 175: Can you add more tests like namespace not exist scenario here?

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
mchades: yes. Here is the definition of `COMMA`:
```java
private static final Splitter COMMA = Splitter.on(",").omitEmptyStrings().trimResults();
```

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
mchades: My thought is that the test class can use `getLanceRESTServerMetalakeName()` in `BaseIT` to create the metalake before the test and perform cleanup afterward. If we create the metalake here, it will be difficult to clean up.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
mchades: It's used to remove the `in-use` property from the old catalog when creating a catalog namespace with `overwrite` mode. 

Because we use `removeOldCatalogProperty` and `setNewCatalogProperty` to achieve this, the `in-use` property in the catalog is reserved and cannot be used in `removePropertyChange`.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
mchades: yes, fixed

mchades:
@jerryshao all comments resolved, could you plz help to review again? thx!

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
```diff
@@ -330,11 +344,18 @@ public void startIntegrationTest() throws Exception {
...
     if (testMode != null && testMode.equals(ITUtils.EMBEDDED_TEST_MODE)) {
-      MiniGravitinoContext context = new MiniGravitinoContext(customConfigs, ignoreAuxRestService);
+      MiniGravitinoContext context =
+          new MiniGravitinoContext(
+              customConfigs, ignoreIcebergAuxRestService, ignoreLanceAuxRestService);
       miniGravitino = new MiniGravitino(context);
       miniGravitino.start();
       serverConfig = miniGravitino.getServerConfig();
     } else {
+      if (!ignoreLanceAuxRestService) {
```
yuqi1129: Why can't we move the code `ignoreLanceAuxRestService` and `ignoreRestAuxRestService` to L346 or before, I noticed that the same logic is there.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -0,0 +1,394 @@
...
+      };
+  private final BufferAllocator allocator = new RootAllocator(Long.MAX_VALUE);
+  private LanceNamespace ns;
+  private Path tempDir;
+
+  @BeforeAll
+  public void startIntegrationTest() throws Exception {
+    super.ignoreLanceAuxRestService = false;
+    super.startIntegrationTest();
+    this.metalake = createMetalake(getLanceRESTServerMetalakeName());
```
yuqi1129, line 73: Why don't we set a constant name before running Lance rest server?
```java
  public void startIntegrationTest() throws Exception {
    customConfigs.put("gravitino.lance-rest.gravitino.metalake-name", METALAKE_NAME);
    super.startIntegrationTest();
  }
```

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -0,0 +1,394 @@
...
+  @AfterAll
+  public void clean() {
+    client.dropMetalake(getLanceRESTServerMetalakeName(), true);
+    tempDir.toFile().deleteOnExit();
+  }
+
+  @AfterEach
+  public void clearMetalake() {
+    Arrays.stream(metalake.listCatalogs()).forEach(c -> metalake.dropCatalog(c, true));
+    tempDir.toFile().deleteOnExit();
```
yuqi1129: You have done it in L82, it seems that this is unnecessary.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
mchades: I think `BaseIT` should manage the configuration instead of scattering it everywhere.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
mchades: fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
mchades: ok. keep in afterAll only

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
yuqi1129: BaseIT manages common configurations, and subclasses can adjust them according to the detailed test requirement. Please note SparkIT and FlinkIT, they all make some customization on `startIntegrationTest`.

I don't think let `BaseIT` choose the matelake name is good idea.

mchades: For the current implementation, when developers want to enable the LRS in IT, they just need to extend `BaseIT` and set `ignoreLanceAuxRestService = false` before tests. They don't need to care about `customConfigs`, so I prefer to keep the current implementation. 

BTW, the automatically generated metalake name is random. Users can also retrieve it using the `getLanceRESTServerMetalakeName()` method. I don't see a use case that requires specifying a metalake name.

mchades: This isn't a blocker. If you only have this one comment, let's get the ball rolling and merge this. You can address it in a separate PR if needed. WDYT? @jerryshao

yuqi1129: If you insist that `They don't need to care about customConfigs`  and ignore the makelake name, I have no more comment here.   

You can think, why don't `IcebergITs` and `FlinkITs` also add some specific configuration and functions to the BaseITs?

>  I don't see a use case that requires specifying a metalake name.

That is not the core point; what matters is that it's the responsibility of the subclass of BaseIT to choose and create makelake.

yuqi1129:
LGTM

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
jerryshao: @yuqi1129 you can update based on this PR.

Issue #8969, opened 2025-10-30
[Bug report] run test_gvfs_with_gcs.TestGvfsWithGCS encounter NullPointerException


danhuawang:
### Version

main branch

### Describe what's wrong

When I ran gcs test cases in  test_gvfs_with_gcs.TestGvfsWithGCS or test_gvfs_with_s3_credential.TestGvfsWithS3Credential, pop error: java.lang.NullPointerException: projectId must not be null.
But I try to use the bucket  and the service account json file to create fileset on gravitino UI, it works. So the credential is right.

### Error message and/or stacktrace

```
ERROR: setUpClass (test_gvfs_with_gcs.TestGvfsWithGCS)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/tests/integration/test_gvfs_with_gcs.py", line 74, in setUpClass
    cls._init_test_entities()
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/tests/integration/test_gvfs_with_gcs.py", line 136, in _init_test_entities
    catalog.as_fileset_catalog().create_fileset(
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/gravitino/client/fileset_catalog.py", line 166, in create_fileset
    return self.create_multiple_location_fileset(
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/gravitino/client/fileset_catalog.py", line 207, in create_multiple_location_fileset
    resp = self.rest_client.post(
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/gravitino/utils/http_client.py", line 227, in post
    return self._request(
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/gravitino/utils/http_client.py", line 205, in _request
    error_handler.handle(resp)
  File "/home/runner/work/gravitino-enterprise/gravitino-enterprise/gravitino-oss/clients/client-python/gravitino/exceptions/handlers/fileset_error_handler.py", line 46, in handle
    super().handle(error_response)
[...]

PR #9012, opened 2025-11-04
[#8969] fix(client-python): fix fileset test with GCS in python module

yuqi1129:

1. Increase the timeout time when connecting to the Gravitino server.
2. In some cases, the modification time of the GCS file/directory is empty.

It's a bug.

Fix: #8969

N/A.

It's self test.

### DIFF
--- a/clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
+++ b/clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 import importlib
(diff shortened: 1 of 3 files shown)


mchades:
I noticed that the error message in the corresponding issue is `projectId must not be null`. This seems unrelated to your PR fix?

yuqi1129:
> I noticed that the error message in the corresponding issue is `projectId must not be null`. This seems unrelated to your PR fix?

 `projectId must not be null` is not a problem, and it's was caused by wrong bucket name.  This PR intends to fix other problem tested during the fix.

Issue #9015, opened 2025-11-04
[Subtask] Add UTs and ITs for table operations in Lance REST server


yuqi1129:
### Describe the subtask

There are no tests that conver table operations in the Lance REST server, which make code quite fragile. 

### Parent issue

#8889

PR #9016, opened 2025-11-04
[#9015] improvement(lance-rest): Add UTs and ITs for Lance rest table operations

yuqi1129:

Add ITs for lance rest service and test lance table operations.

To make code robust

Fix: #9015

N/A.

It's self a test.

### DIFF
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -62,7 +62,7 @@ import org.slf4j.LoggerFactory;
 public class GravitinoLanceTableOperations implements LanceTableOperations {
   private static final Logger LOG = LoggerFactory.getLogger(GravitinoLanceTableOperations.class);
 
-  private GravitinoLanceNamespaceWrapper namespaceWrapper;
+  private final GravitinoLanceNamespaceWrapper namespaceWrapper;
 
   public GravitinoLanceTableOperations(GravitinoLanceNamespaceWrapper namespaceWrapper) {
     this.namespaceWrapper = namespaceWrapper;


lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -97,13 +104,28 @@ public Response createTable(
...
+                  tableProperties,
+                  jsonNode -> {
+                    Map<String, String> map = new HashMap<>();
+                    jsonNode
+                        .fields()
+                        .forEachRemaining(
+                            entry -> {
+                              map.put(entry.getKey(), entry.getValue().asText());
+                            });
+                    return map;
```
mchades: why not using `JsonUtil.mapper().readValue(tableProperties, new TypeReference<>() {})`

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -116,24 +138,30 @@ public Response createTable(
...
-
+      String tableLocation = request.getLocation();
       Map<String, String> props =
-          StringUtils.isBlank(tableProperties)
-              ? Map.of()
-              : JsonUtil.mapper().readValue(tableProperties, new TypeReference<>() {});
+          request.getProperties() == null
+              ? Maps.newHashMap()
+              : Maps.newHashMap(request.getProperties());
+      props.put("format", "lance");
```
mchades: It's better to move this to `GravitinoLanceNamespaceWrapper` since it's a Gravitino-specific property.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -116,24 +138,30 @@ public Response createTable(
...
+              .createTable(
+                  tableId,
+                  CreateTableRequest.ModeEnum.CREATE,
+                  delimiter,
+                  tableLocation,
+                  props,
+                  null);
+      CreateEmptyTableResponse responseObj = new CreateEmptyTableResponse();
+      responseObj.setProperties(response.getProperties());
+      responseObj.setLocation(response.getLocation());
```
mchades: you are missing `storage_options` for the `CreateEmptyTableResponse`

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -525,7 +522,8 @@ public ListTablesResponse listTables(
   }
 
   @Override
-  public DescribeTableResponse describeTable(String tableId, String delimiter) {
+  public DescribeTableResponse describeTable(String tableId, String delimiter, Long version) {
+    // TODO Currently we do not support versioned table description.
```
mchades: If so, I would suggest throwing an unsupported exception when the version is null.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -18,26 +18,66 @@
...
 import com.lancedb.lance.namespace.model.RegisterTableResponse;
 import java.util.Map;
 
 public interface LanceTableOperations {
 
-  DescribeTableResponse describeTable(String tableId, String delimiter);
+  /**
+   * Describe the details of a table.
+   *
+   * @param tableId table ids are in the format of "namespace/delimiter/table_name"
```
mchades: The format should not include `/`

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -570,11 +569,10 @@ public CreateTableResponse createTable(
 
     Map<String, String> createTableProperties = Maps.newHashMap(tableProperties);
     createTableProperties.put("location", tableLocation);
-    createTableProperties.put("mode", mode);
-    // TODO considering the mode (create, exist_ok, overwrite)
+    createTableProperties.put("mode", mode.getValue());
```
mchades: why need to put the mode into the table properties?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
mchades: wrong exception type

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -643,14 +639,16 @@ public RegisterTableResponse registerTable(
           CommonUtil.formatCurrentStackTrace());
     }
 
-    if (createMode == RegisterTableRequest.ModeEnum.OVERWRITE
+    if (mode == RegisterTableRequest.ModeEnum.OVERWRITE
         && catalog.asTableCatalog().tableExists(tableIdentifier)) {
       LOG.info("Overwriting existing table: {}", tableId);
-      catalog.asTableCatalog().dropTable(tableIdentifier);
+      catalog.asTableCatalog().purgeTable(tableIdentifier);
```
mchades: why using purgeTable?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -673,7 +670,14 @@ public DeregisterTableResponse deregisterTable(String tableId, String delimiter)
     Table t = catalog.asTableCatalog().loadTable(tableIdentifier);
     Map<String, String> properties = t.properties();
     // TODO Support real deregister API.
-    catalog.asTableCatalog().purgeTable(tableIdentifier);
+    boolean result = catalog.asTableCatalog().purgeTable(tableIdentifier);
```
mchades: should not use `purgeTable` for `deregisterTable`

lance/lance-common/src/test/java/org/apache/gravitino/lance/common/utils/TestArrowUtils.java
```diff
@@ -41,36 +35,9 @@ public void testParseArrowIpcStream() throws Exception {
             Arrays.asList(
                 Field.nullable("id", new ArrowType.Int(32, true)),
                 Field.nullable("value", new ArrowType.Utf8())));
-
-    GravitinoLanceNamespaceWrapper wrapper = new GravitinoLanceNamespaceWrapper();
-    byte[] ipcStream = generateIpcStream(schema);
-    Schema parsedSchema = wrapper.parseArrowIpcStream(ipcStream);
+    byte[] ipcStream = ArrowUtils.generateIpcStream(schema);
+    Schema parsedSchema = ArrowUtils.parseArrowIpcStream(ipcStream);
```
mchades, line 38: plz move this UT to `TestArrowUtils.java`

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -24,38 +24,64 @@
...
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.com.google.common.base.Joiner;
 
 public class LanceRESTServiceIT extends BaseIT {
+  public static final String CATALOG_NAME = GravitinoITUtils.genRandomName("lance_rest_catalog");
+  public static final String SCHEMA_NAME = GravitinoITUtils.genRandomName("lance_rest_schema");
```
mchades: It's better to use `private final`

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -372,6 +398,221 @@ public void testNamespaceExists() {
...
+    createSchema();
+
+    CreateEmptyTableRequest request = new CreateEmptyTableRequest();
+    String location = tempDir + "/" + "empty_table/";
+    request.setLocation(location);
+    request.setProperties(ImmutableMap.of());
+    request.setId(List.of(CATALOG_NAME, SCHEMA_NAME, "empty_table"));
+
+    CreateEmptyTableResponse response = ns.createEmptyTable(request);
+    Assertions.assertNotNull(response);
```
mchades, line 421: you should also verify the response member

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -385,6 +626,13 @@ private Catalog createCatalog(String catalogName) {
         properties);
   }
 
+  protected void createSchema() {
```
mchades: private should be enough?

lance/lance-rest-server/src/test/resources/log4j2.properties
```diff
@@ -0,0 +1,73 @@
...
+name = ConsoleLogConfig
+
+# Console appender configuration
+appender.console.type = Console
+appender.console.name = consoleLogger
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss} %-5p [%t] %c{1}:%L - %m%n
+
+# Log files location
+property.logPath = ${sys:gravitino.log.path:-build/catalog-hive-integration-test.log}
```
mchades: wrong file name

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: Please see:https://github.com/lancedb/lance-namespace/blob/2033b2fca126e87e56ba0d5ec19c5ec010c7a98f/java/lance-namespace-core/src/main/java/com/lancedb/lance/namespace/rest/RestNamespace.java#L207-L208

mchades: Can you give an example of what the differences are in their serialization results?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: Just let Gravitino get the current mode, and the Gravitino server has not used it so far.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
yuqi1129: It may be misleading,  what I want to express is `namespace delimiter table_name`. I will change it.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
mchades: We can remove it now and add it when needed.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: That means we will ignore the version value and always return the latest version. 

According to the docs: https://lancedb.github.io/lance/format/namespace/operations/describe-table/#description

> Version of the table to describe.
            If not specified, server should resolve it to the latest version.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: An empty map will be serialized as "", not "{}"

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: already added.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
mchades: my typo, what I mean is we should throw an unsupported exception when the version is not null.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
mchades: should you also pass the storage options of the table?

mchades: What if the `Dataset.drop` failed?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -596,7 +598,7 @@ public CreateTableResponse createTable(
         break;
       case OVERWRITE:
         if (catalog.asTableCatalog().tableExists(tableIdentifier)) {
```
mchades: if code reach here, `catalog.asTableCatalog().tableExists` will be called three times

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -607,21 +609,28 @@ public CreateTableResponse createTable(
...
-                columns.toArray(new Column[0]),
-                tableLocation,
-                createTableProperties);
+                tableIdentifier, columns.toArray(new Column[0]), null, createTableProperties);
 
     CreateTableResponse response = new CreateTableResponse();
     response.setProperties(t.properties());
     response.setLocation(tableLocation);
+    // We have removed the prefix "lance.storage" from the table properties for storage options. So
+    // here we need to add it back.
```
mchades: I'm confused about this comment, where did we remove the prefix "lance.storage" from the table properties?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
jerryshao, line 285: The `false` means "failed to store" or "the table not existed"? If the latter, it should return `false`, not throwing an exception.

The semantics are misleading; you need to confirm with @mchades .

jerryshao: I remembered that @youngyjd mentioned that they don't want to access the storage. How do you handle this?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -673,7 +681,14 @@ public DeregisterTableResponse deregisterTable(String tableId, String delimiter)
     Table t = catalog.asTableCatalog().loadTable(tableIdentifier);
     Map<String, String> properties = t.properties();
     // TODO Support real deregister API.
-    catalog.asTableCatalog().purgeTable(tableIdentifier);
+    boolean result = catalog.asTableCatalog().dropTable(tableIdentifier);
+    if (!result) {
+      throw LanceNamespaceException.notFound(
+          "Table not found: " + tableId,
+          NoSuchSchemaException.class.getSimpleName(),
```
mchades: wrong class

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
yuqi1129: `Dataset.drop` lies in the `purgeTable` operation, and for `deregisterTable` I use `dropTable`, which will not remove lance data.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -18,26 +18,66 @@
...
       byte[] arrowStreamBody);
 
+  /**
+   * Register an existing table.
+   *
+   * @param tableId table ids are in the format of "namespace/delimiter/table_name"
+   * @param mode it can be REGISTER or OVERWRITE.
+   * @param delimiter the delimiter used in the namespace
+   * @param tableProperties the properties of the table, it should contain the table location
+   * @return
```
mchades: miss comment

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -525,7 +523,12 @@ public ListTablesResponse listTables(
   }
 
   @Override
-  public DescribeTableResponse describeTable(String tableId, String delimiter) {
+  public DescribeTableResponse describeTable(String tableId, String delimiter, Long version) {
+    if (version != null) {
+      throw new UnsupportedOperationException(
+          "Describing specific table version is not supported.");
```
jerryshao: Which exception and http code will be converted for this `UnsupportedOperationException`?

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
yuqi1129: image

Generic lakehouse will return false if `the table not existed`.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -18,26 +18,66 @@
...
-  DescribeTableResponse describeTable(String tableId, String delimiter);
+  /**
+   * Describe the details of a table.
+   *
+   * @param tableId table ids are in the format of "{namespace}{delimiter}{table_name}"
+   * @param delimiter the delimiter used in the namespace
+   * @param version the version of the table to describe, if null, describe the latest version
+   * @return the table description
+   */
+  DescribeTableResponse describeTable(String tableId, String delimiter, Long version);
```
jerryshao: Better to change to `Optional<Long>` if we want to support nullable parameter.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
jerryshao: You'd better define a helper function for this, and clearly describe why do you implement like this (not leave the comment here).

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -116,24 +138,31 @@ public Response createTable(
   @ResponseMetered(name = "create-empty-table", absolute = true)
   public Response createEmptyTable(
       @PathParam("id") String tableId,
-      @QueryParam("mode") @DefaultValue("create") String mode, // create, exist_ok, overwrite
       @QueryParam("delimiter") @DefaultValue(NAMESPACE_DELIMITER_DEFAULT) String delimiter,
+      CreateEmptyTableRequest request,
```
jerryshao, line 129: You'd better have a mechanism validating the request parameters to avoid unexpected requests.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: Drop operations will be done in `GenericLakehouseCatalogOperations#drop` and code will not reach `lance#drop`.
Another point is that we do not support drop tables in LRS currently.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: OK, I can call it once.  The only drawback is that the result of `tableExists` will be out-of-date.

catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceCatalogOperations.java
mchades: To be honest, the current code organization scatters the logic for table operations between `GenericLakehouseCatalogOperation` and `LanceCatalogOperation`, which reduces code readability (you need to pay attention to both code logics when operating Lance tables) and introduces some repetitive operations (such as repeatedly gettingTable from store), I think we should consider this issue in our future refactoring.

jerryshao: BTW, why do you call `store.get` twice?

The code here is confused; both the caller and callee will access the DB, and the callee handles some exceptions, but some rely on the caller.

The lance should only focus on lance-specific logic, whereas the generic lakehouse operation should only focus on the metadata-related CRUD operations.

jerryshao: > To be honest, the current code organization scatters the logic for table operations between `GenericLakehouseCatalogOperation` and `LanceCatalogOperation`, which reduces code readability (you need to pay attention to both code logics when operating Lance tables) and introduces some repetitive operations (such as repeatedly gettingTable from store), I think we should consider this issue in our future refactoring.

I agree, the implementation is not good, it messes with different logic, and doesn't clearly separate the responsibilities.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: Added.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: done

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
yuqi1129: done

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -569,59 +581,65 @@ public CreateTableResponse createTable(
...
+                .asTableCatalog()
+                .createTable(
+                    tableIdentifier, columns.toArray(new Column[0]), null, createTableProperties);
+      } else { // EXIST_OK
+        CreateTableResponse response = new CreateTableResponse();
+        Table existingTable = catalog.asTableCatalog().loadTable(tableIdentifier);
+        response.setProperties(existingTable.properties());
+        response.setLocation(existingTable.properties().get(LANCE_LOCATION));
+        response.setVersion(null);
+        return response;
```
mchades: miss `setStorageOptions`?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -631,36 +649,39 @@ public RegisterTableResponse registerTable(
...
-          CommonUtil.formatCurrentStackTrace());
-    }
-
-    if (createMode == RegisterTableRequest.ModeEnum.OVERWRITE
-        && catalog.asTableCatalog().tableExists(tableIdentifier)) {
-      LOG.info("Overwriting existing table: {}", tableId);
-      catalog.asTableCatalog().dropTable(tableIdentifier);
+    Table t = null;
+    try {
+      tableProperties.put("format", "lance");
```
mchades: the `tableProperties` may be immutable, you should copy it first

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: Yeah, I miss it.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: OK, although I use a mutable map outside this method.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -631,36 +649,39 @@ public RegisterTableResponse registerTable(
...
+    } catch (TableAlreadyExistsException exception) {
+      if (mode == RegisterTableRequest.ModeEnum.CREATE) {
+        throw LanceNamespaceException.conflict(
+            "Table already exists: " + tableId,
+            TableAlreadyExistsException.class.getSimpleName(),
+            tableId,
+            CommonUtil.formatCurrentStackTrace());
+      } else if (mode == RegisterTableRequest.ModeEnum.OVERWRITE) {
+        LOG.info("Overwriting existing table: {}", tableId);
+        catalog.asTableCatalog().dropTable(tableIdentifier);
```
mchades: should here use `dropTable` or `purgeTable` ?

yuqi1129: It's a register table API, and we need to use `dropTable` here .

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -18,26 +18,67 @@
...
+   * @param delimiter the delimiter used in the namespace
+   * @param version the version of the table to describe, if null, describe the latest version
+   * @return the table description
+   */
+  DescribeTableResponse describeTable(String tableId, String delimiter, Optional<Long> version);
 
+  /**
+   * Create a new table.
+   *
+   * @param tableId table ids are in the format of "namespace/delimiter/table_name"
```
mchades: `"{namespace}{delimiter}{table_name}"`

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -18,26 +18,67 @@
...
+      CreateTableRequest.ModeEnum mode,
       String delimiter,
       String tableLocation,
       Map<String, String> tableProperties,
       byte[] arrowStreamBody);
 
+  /**
+   * Register an existing table.
+   *
+   * @param tableId table ids are in the format of "namespace/delimiter/table_name"
```
mchades: `"{namespace}{delimiter}{table_name}"`

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -18,26 +18,67 @@
...
-      String tableId, String mode, String delimiter, Map<String, String> tableProperties);
+      String tableId,
+      RegisterTableRequest.ModeEnum mode,
+      String delimiter,
+      Map<String, String> tableProperties);
 
+  /**
+   * Deregister a table. It will not delete the underlying lance data.
+   *
+   * @param tableId table ids are in the format of "namespace/delimiter/table_name"
```
mchades: `"{namespace}{delimiter}{table_name}"`

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -116,24 +126,33 @@ public Response createTable(
...
+          request.getProperties() == null
+              ? Maps.newHashMap()
+              : Maps.newHashMap(request.getProperties());
+
       CreateTableResponse response =
           lanceNamespace
               .asTableOps()
-              .createTable(tableId, mode, delimiter, tableLocation, props, null);
-      return Response.ok(response).build();
+              .createTable(
```
mchades: I would suggest you add a `createEmptyTable` API in `lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java` to align with the Lance REST operation

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -176,11 +195,34 @@ public Response deregisterTable(
...
           lanceNamespace.asTableOps().deregisterTable(tableId, delimiter);
       return Response.ok(response).build();
     } catch (Exception e) {
       return LanceExceptionMapper.toRESTResponse(tableId, e);
     }
   }
+
+  private void validateCreateEmptyTableRequest(CreateEmptyTableRequest request) {
+    Preconditions.checkArgument(
+        StringUtils.isNotBlank(request.getLocation()), "Table location must be provided.");
```
mchades: per OAS, the `location` is optional

yuqi1129: > per OAS, the location is optional

What's does it means?  the location can be null for an empty table?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
mchades: alright, the OAS seems not describe how to handle the data of the existing table

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
mchades: yes, If I am wrong, please correct me: https://github.com/lancedb/lance-namespace/blob/b035daec6b7cce0bec484c18326100d07c9e33bb/docs/src/rest.yaml#L2513-L2517

yuqi1129: I wonder if the location is null. How do we determine the location?

mchades: It can be generated based on the location in the catalog and schema. An error can be thrown from Gravitino if all are is null.

yuqi1129: Did you handle it and remind users that they can set the location in the schema and catalog? Currently, Gravitino does support the feature you address, but I'm worried that it's not the design that Lance REST supports.

mchades: you should follow the Lance REST OAS

yuqi1129: > Another problem, why it's optional and required in `createTable`

I got it.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: Yeah, currently we can handle the logic by ourself, once the detailed definition comes out, we can refine it.

yuqi1129:
All resolved.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
```diff
@@ -525,7 +528,14 @@ public ListTablesResponse listTables(
   }
 
   @Override
-  public DescribeTableResponse describeTable(String tableId, String delimiter) {
+  public DescribeTableResponse describeTable(
```
jerryshao: I would suggest you to split this file into small ones according to the different operations. As you continue to add more operations, this file will be too large to maintain.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -58,10 +59,27 @@ CreateTableResponse createTable(
...
+   * @param tableId table ids are in the format of "{namespace}{delimiter}{table_name}"
+   * @param mode it can be CREATE, OVERWRITE, or EXIST_OK
+   * @param delimiter the delimiter used in the namespace
+   * @param tableLocation the location where the table data will be stored
+   * @param tableProperties the properties of the table
+   * @return the response of the create table operation
+   */
+  CreateEmptyTableResponse createEmptyTable(
+      String tableId,
+      CreateTableRequest.ModeEnum mode,
```
mchades: createEmptyTable does not have `mode` parameter

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNamespaceWrapper.java
yuqi1129: Split it into three files

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
```diff
@@ -0,0 +1,293 @@
...
+import org.apache.gravitino.lance.common.utils.LancePropertiesUtils;
+import org.apache.gravitino.rel.Column;
+import org.apache.gravitino.rel.Table;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GravitinoLanceTableOperations implements LanceTableOperations {
+  private static final Logger LOG = LoggerFactory.getLogger(GravitinoLanceTableOperations.class);
+
+  private GravitinoLanceNamespaceWrapper namespaceWrapper;
```
mchades: can be final

Issue #8942, opened 2025-10-28
[FEATURE] Add access control for tags


hdygxsj:
### Describe the feature

Add access control for tags

roryqi:
What privileges does tag need?

hdygxsj:
As I understand it, the permissions required for each tag operation API are as follows:
What do you think? @jerqi   cc  @jerryshao 

| API                       | What permissions are required? |
| ------------------------- | --------------------- |
| list-tags                 | Any metalke user      |
| add tag                   | METALAKE OWNER        |
| get tag                   | Any metalke user      |
| alter tag                 | METALAKE OWNER        |
| delete tag                | METALAKE OWNER        |
| listMetadataObjectsForTag | load metadata object  |
| listTagsForMetadataObject | load metadata object  |
| associateTagsForObject    | load metadata object  |
| getTagForObject           | load metadata object  |
|                           |                       |

Tag is not a metadata object, so it seems we can't assign privileges or owner to it.

If a tag is also considered a type of metadata, permissions could be designed as follows:
| API                       | what permission need?                                        |
| ------------------------- | ------------------------------------------------------------ |
| list-tags                 | TAG_OWNER \|\|LOAD_TAG \|\| MANAGER_TAG                      |
| add tag                   | TAG_OWNER  \|\| MANAGER_TAG                                  |
| get tag                   | TAG_OWNER \|\| LOAD_TAG \|\| MANAGER_TAG                     |
| alter tag                 | TAG_OWNER \|\| MANAGER_TAG                                   |
| delete tag                | TAG_OWNER \|\| MANAGER_TAG                                   |
| listMetadataObjectsForTag | （TAG_OWNER \|\|LOAD_TAG \|\| MANAGER_TAG） &&   load metadata object |
| listTagsForMetadataObject | （TAG_OWNER \|\|LOAD_TAG \|\| MANAGER_TAG）&&  load metadata object |
| associateTagsForObject    | （TAG_OWNER \|\| LOAD_TAG \|\|  MANAGER_TAG）&& load metadata object |
| getTagForObject           | （TAG_OWNER \|\| LOAD_TAG \|\| MANAGER_TAG）&& load metadata object |
|                           |                                                              |

If tags can be treated as metadata objects, finer-grained control over tags would be possible, but this would also require more extensive modifications.

hdygxsj:
Tag is not a metadata object, so it seems we can't assign privileges or owner to it.

roryqi:
[...]

PR #9018, opened 2025-11-04
[#8942]feat(authz): support tag access control

hdygxsj:

support tag access control

Fix: #8942

None

org.apache.gravitino.client.integration.test.authorization.TagOperationsAuthorizationIT

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
@@ -70,7 +70,9 @@ public class AssociateTagAuthorizationExecutor implements AuthorizationExecutor 
     AuthorizationRequestContext context = new AuthorizationRequestContext();
     Entity.EntityType targetType =
         Entity.EntityType.TAG; // Tags are the only supported batch target here
-    Preconditions.checkArgument(request instanceof TagsAssociateRequest);
+    Preconditions.checkArgument(
+        request instanceof TagsAssociateRequest,
+        "Only tag can use AssociateTagAuthorizationExecutor, please contact the administrator.");
     TagsAssociateRequest tagsAssociateRequest = (TagsAssociateRequest) request;
     tagsAssociateRequest.validate();
     // Authorize both 'tagsToAdd' and 'tagsToRemove' fields.


hdygxsj:
PHAL @jerqi

api/src/main/java/org/apache/gravitino/authorization/Privileges.java
```diff
@@ -886,4 +887,39 @@ public boolean canBindTo(MetadataObject.Type type) {
...
+    /**
+     * @return The instance with deny condition of the privilege.
+     */
+    public static CreateTag deny() {
+      return DENY_INSTANCE;
+    }
+
+    @Override
+    public boolean canBindTo(MetadataObject.Type type) {
+      return true;
```
roryqi: Could the tag and role be tagged?

hdygxsj: It seems there are no restrictions in TagManager.

roryqi: cc @jerryshao WDYT？

roryqi:
Could u add the documents for this PR?

api/src/main/java/org/apache/gravitino/authorization/Privileges.java
hdygxsj: > Could the tag and role be tagged?

I misunderstood earlier—CREATE_TAG should be bound to the metalake.

docs/security/access-control.md
```diff
@@ -958,71 +964,80 @@ You can follow the steps to achieve the authorization of Gravitino.
...
+| get role                          | The owner of the metalake or the role. others can see his granted or owned roles. |
+| list roles                        | The owner of the metalake can see all the roles. Others can see his granted roles or owned roles. |
+| grant role                        | `MANAGE_GRANTS` on the metalake                              |
+| revoke role                       | `MANAGE_GRANTS` on the metalake                              |
+| grant privilege                   | `MANAGE_GRANTS` on the metalake or the owner of the securable object |
+| revoke privilege                  | `MANAGE_GRANTS` on the metalake or the owner of the securable object |
+| set owner                         | The owner of the securable object                            |
+| list tags                         | The owner of the metalake can see all the tags. Others can see his owned tags. |
+| create tag                        | `CREATE_TAG` on the metalake or the owner of the metalake    |
+| get tag                           | The owner of the metalake or the tag.                        |
```
roryqi: I have some concern about this. Maybe we can't use this.

roryqi:
1. Could you add the tag to the metadata object type in the openapi.yaml
2. Could you confirm influence about tag, policy when adding tag to the metadata object type?

core/src/main/java/org/apache/gravitino/authorization/OwnerManager.java
```diff
@@ -110,12 +110,13 @@ public void setOwner(
         newOwner.name = ownerName;
         newOwner.type = Owner.Type.GROUP;
       }
-
-      AuthorizationUtils.callAuthorizationPluginForMetadataObject(
-          metalake,
-          metadataObject,
-          authorizationPlugin ->
-              authorizationPlugin.onOwnerSet(metadataObject, originOwner.orElse(null), newOwner));
+      if (metadataObject.type() != MetadataObject.Type.TAG) {
```
roryqi: Why do we need this?

hdygxsj: org.apache.gravitino.authorization.AuthorizationUtils#callAuthorizationPluginForMetadataObject can not handle tag
image

roryqi: We should let this method handle the tag. Because tag becomes a metadata object.

hdygxsj: > We should let this method handle the tag. Because tag becomes a metadata object.

This method seems to be for handling pushdown, tags shouldn’t be in the underlying engine, right?

hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
```diff
@@ -219,21 +252,48 @@ public Response listTagsForMetadataObject(
...
       return Utils.doAs(
           httpRequest,
           () -> {
+            String[] tagsToAdd = request.getTagsToAdd();
+            if (tagsToAdd != null && tagsToAdd.length > 0) {
+              Arrays.stream(tagsToAdd)
+                  .forEach(
+                      tag -> {
+                        boolean result =
+                            new AuthorizationExpressionEvaluator(
```
roryqi: Why do we this instead of expression?

hdygxsj: Authorization expressions are also used here, but interceptor is not used.

Just like with list operations, it's challenging in interceptors to extract an array from the request body and perform authorization and filtering on it.

roryqi:
We would like to load a metadata object first. Then, we associate the metadata object with tags. It requires we must have the privilege to load the table.  Actually, if we have the privilege `BROWSE` in the future, we can only list the metadata objects and know the names of the metadata objects. Maybe we need to add an interface for Tag
```
SupportsTagOperations {
   void associateObjects(List<MetadataObject> addedObjects,  List<MetadataObject> removeObjects)
}
```
@jerryshao WDYT?

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
roryqi: If we have the tag which we don't have privilege, we should forbid the behaviour.

You should move the logic to `GravitinoInterceptionService`
We can add the new annotation for the request like 
```
      AuthorizationRequest authorizeRequest =
          parameter.getAnnotation(AuthorizationRequest.class);
```
then
```
preAuthzRequest(authorizeRequest);
```

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
```diff
@@ -77,11 +92,13 @@ void setHttpRequest(HttpServletRequest httpRequest) {
   @Produces("application/vnd.gravitino.v1+json")
   @Timed(name = "get-object-tag." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
   @ResponseMetered(name = "get-object-tag", absolute = true)
+  @AuthorizationExpression(expression = "CAN_ACCESS_METADATA && ( METALAKE::OWNER || TAG::OWNER)")
   public Response getTagForObject(
```
roryqi, line 94: This expression is wrong 
```
METALAKE_OWNER || (CAN_LOAD_OBJECT && CAN_LOAD_TAG)
```

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
hdygxsj: AuthorizationExpressionEvaluator and authorization expressions are not suitable for authorizing lists/arrays—or even multiple metadata objects within multiple lists/arrays(tagsToAdd, tagsToRemove). Should we consider designing a new authorization mechanism specifically for this scenario?

roryqi: Maybe we should design this. We should have the semantics like scala `forall`.

hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
```diff
@@ -140,28 +150,48 @@ public Object invoke(MethodInvocation methodInvocation) throws Throwable {
...
-
-            LOG.warn(
-                "Authorization failed - User: {}, Operation: {}, Metadata: {}, Expression: {}",
-                currentUser,
-                methodName,
-                accessMetadataName,
-                expression);
-
-            return buildNoAuthResponse(errorMessage, accessMetadataName, currentUser, methodName);
+          if (!isBatch) {
```
roryqi: Could extract some classes to handle the conditions?

hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
```diff
@@ -140,28 +148,23 @@ public Object invoke(MethodInvocation methodInvocation) throws Throwable {
...
           AuthorizationExpressionEvaluator authorizationExpressionEvaluator =
               new AuthorizationExpressionEvaluator(expression);
-          boolean authorizeResult =
-              authorizationExpressionEvaluator.evaluate(
-                  metadataContext,
-                  pathParams,
-                  new AuthorizationRequestContext(),
-                  Optional.ofNullable(entityType));
+          AuthorizationRequest.RequestType requestType =
+              extractAuthorizationRequestTypeFromParameters(parameters);
```
roryqi, line 144: Could u use a factory here?

server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
```diff
@@ -140,28 +148,23 @@ public Object invoke(MethodInvocation methodInvocation) throws Throwable {
...
+          executor =
+              switch (requestType) {
+                case COMMON -> new CommonAuthorizerExecutor(
```
roryqi: Could extract these classes to single files?

hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizeExecutor.java
```diff
@@ -0,0 +1,121 @@
...
+      return false;
+    }
+
+    AuthorizationRequestContext context = new AuthorizationRequestContext();
+    Entity.EntityType targetType =
+        Entity.EntityType.TAG; // Tags are the only supported batch target here
+
+    // Authorize both 'tagsToAdd' and 'tagsToRemove' fields.
+    // Short-circuit on first failure.
+    return authorizeTagField(request, "tagsToAdd", context, targetType)
```
roryqi: We don't need use reflection here.

hdygxsj: fixed

core/src/main/java/org/apache/gravitino/authorization/AuthorizationUtils.java
```diff
@@ -195,6 +195,9 @@ public static void callAuthorizationPluginForSecurableObjects(
 
   public static void callAuthorizationPluginForMetadataObject(
       String metalake, MetadataObject metadataObject, Consumer<AuthorizationPlugin> consumer) {
+    if (metadataObject.type() == MetadataObject.Type.TAG) {
```
roryqi: Do we still need this?

hdygxsj: Yes, otherwise setting the tag's owner here will throw an exception.

core/src/main/java/org/apache/gravitino/hook/TagHookDispatcher.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @Override
+  public Tag createTag(
+      String metalake, String name, String comment, Map<String, String> properties) {
+    AuthorizationUtils.checkCurrentUser(metalake, PrincipalUtils.getCurrentUserName());
+    Tag tag = dispatcher.createTag(metalake, name, comment, properties);
+    try {
+      // Set the creator as the owner of the catalog.
+      OwnerDispatcher ownerDispatcher = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerDispatcher != null) {
```
roryqi: Why do we need `try catch` here?

core/src/main/java/org/apache/gravitino/authorization/AuthorizationUtils.java
roryqi: image
We have already handled this in this method.

hdygxsj: fixed

core/src/main/java/org/apache/gravitino/hook/TagHookDispatcher.java
hdygxsj: fixed

core/src/main/java/org/apache/gravitino/tag/TagDispatcher.java
```diff
@@ -63,7 +63,8 @@ public interface TagDispatcher {
    * @param properties The properties of the tag.
    * @return The created tag.
    */
-  Tag createTag(String metalake, String name, String comment, Map<String, String> properties);
+  Tag createTag(String metalake, String name, String comment, Map<String, String> properties)
```
roryqi: Why do need to throw Exception here?

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+    AuthorizationRequestContext context = new AuthorizationRequestContext();
+    Entity.EntityType targetType =
+        Entity.EntityType.TAG; // Tags are the only supported batch target here
+
+    // Authorize both 'tagsToAdd' and 'tagsToRemove' fields.
+    if (request instanceof TagsAssociateRequest tagsAssociateRequest) {
+      return authorizeTagField(tagsAssociateRequest.getTagsToAdd(), context, targetType)
+          && authorizeTagField(tagsAssociateRequest.getTagsToRemove(), context, targetType);
+    }
+    throw new UnsupportedOperationException("Unsupported request type: " + request.getClass());
```
roryqi: Could you use `Precondition.checkArguments()`. This should throw illegalAgurmentException here.

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param tagNames tagNames
+   * @param context The shared authorization request context.
+   * @param targetType The entity type being authorized (expected to be TAG).
+   * @return {@code true} if all tags in the field pass authorization; {@code false} otherwise.
+   */
+  private boolean authorizeTagField(
+      String[] tagNames, AuthorizationRequestContext context, Entity.EntityType targetType) {
+
+    // Treat null or empty arrays as no-op (implicitly authorized)
+    if (tagNames == null) {
```
roryqi, line 96: We should call the request.validate() method before this.

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+   */
+  private boolean authorizeTag(
+      String[] tagNames, AuthorizationRequestContext context, Entity.EntityType targetType) {
+
+    // Treat null or empty arrays as no-op (implicitly authorized)
+    if (tagNames == null) {
+      return true;
+    }
+
+    for (String tagName : tagNames) {
```
roryqi, line 100: After the request's validation, we don't need check null value here. You can simplify the code.

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param tagNames tagNames
+   * @param context The shared authorization request context.
+   * @param targetType The entity type being authorized (expected to be TAG).
+   * @return {@code true} if all tags in the field pass authorization; {@code false} otherwise.
+   */
+  private boolean authorizeTag(
+      String[] tagNames, AuthorizationRequestContext context, Entity.EntityType targetType) {
+
+    // Treat null or empty arrays as no-op (implicitly authorized)
+    if (tagNames == null) {
```
roryqi, line 96: After the request's validation, we don't need check null value here. You can simplify the code

hdygxsj: No, the validation will pass if either tagsToAdd or tagsToRemove is not null.

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociateTagAuthorizationExecutor.java
roryqi: They can't be null. If the tag names is null. We will fail to deseriliaze the data. If the value can null, we should mark it as @Nullable. The value can only be empty.

hdygxsj: image

roryqi: OK.

Issue #8796, opened 2025-10-12
[Subtask] Add PartitionNameListResponse and its error handler


tsungchih:
### Describe the subtask

We need to add class `PartitionNameListResponse` and the corresponding error handler.

- [x] `PartitionNameListResponse`
- [x] `ErrorHandlers.PartitionNameListResponse`

### Parent issue

#8591

PR #9024, opened 2025-11-05
[#8796] feat(client-python): add partition name list response

tsungchih:

We need to add class `PartitionNameListResponse` and the corresponding error handler.

- `PartitionNameListResponse`
- `ErrorHandlers.PartitionNameListResponse`

We need to enable table operations in Python client that requires implementation of class `PartitionNameListResponse` used in `RelationalTable`.

Fix: #8796 

No

Unit tests


tsungchih:

unknowntpo:
@tsungchih  sorry for the late reply, @jerryshao LGTM.

Issue #8979, opened 2025-10-31
[Bug report] upgrade-0.8.0-to-0.9.0-postgresql.sql results error: "constraint "uk_fid_ver_del" of relation "fileset_version_info" does not exist"


zuotingbing:
### Version

main branch

### Describe what's wrong

upgrade-0.8.0-to-0.9.0-postgresql.sql results error: "constraint "uk_fid_ver_del" of relation "fileset_version_info" does not exist".

### Error message and/or stacktrace

gravitino-pg=# \d fileset_version_info
Table "public.fileset_version_info"
Column | Type | Collation | Nullable | Default
------------------+------------------------+-----------+----------+----------------------------------
id | bigint | | not null | generated by default as identity
metalake_id | bigint | | not null |
catalog_id | bigint | | not null |
schema_id | bigint | | not null |
fileset_id | bigint | | not null |
version | integer | | not null |
fileset_comment | character varying(256) | | | ''::character varying
properties | text | | |
storage_location | text | | not null |
deleted_at | bigint | | not null | 0
Indexes:
"fileset_version_info_pkey" PRIMARY KEY, btree (id)
"fileset_version_info_fileset_id_version_deleted_at_key" UNIQUE CONSTRAINT, btree (fileset_id, version, deleted_at)
"idx_schema_id" btree (schema_id)

gravitino-pg=# ALTER TABLE fileset_version_info DROP CONSTRAINT uk_fid_ver_del;
ERROR: constraint "uk_fid_ver_del" of relation "fileset_version_info" does not exist
gravitino-pg=#

### How to reproduce

1. install gravitino 0.8.0 with schema-0.8.0-postgresql.sql
2. upgrade gravitino from 0.8.0 to 0.9.0 with upgrade-0.8.0-to-0.9.0-postgresql.sql

yuqi1129:
@mchades 
Please take a look, it seems that we need to explicitly set the name of a `unique key`, or the name should be `{table_name}_{column_name(s)}_key`

zuotingbing:
please also check upgrade-0.9.0-to-1.0.0-postgresql.sql
[https://github.com/apache/gravitino/blob/main/scripts/postgresql/upgrade-0.9.0-to-1.0.0-postgresql.sql](url)


**ERROR1:**
gravitino-pg=# CREATE TABLE IF NOT EXISTS statistic_meta (
    id BIGINT NOT NULL GENERATED BY DEFAULT AS IDENTITY,
[...]

PR #9027, opened 2025-11-06
[#8979] fix(sql-scripts): add sql scripts test and fix bug

mchades:

add sql scripts test and fix bug

Fix: #8979

no

CI pass

### DIFF
--- a/core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
+++ b/core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
@@ -40,6 +40,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.integration.test.container.MySQLContainer;
 import org.apache.gravitino.integration.test.container.PostgreSQLContainer;
 import org.apache.gravitino.integration.test.util.BaseIT;
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
@@ -50,6 +51,7 @@ public class TestSQLScripts extends BaseIT {
   private String jdbcBackend;
   private Path scriptDir;
   @TempDir private File tempDir;
+  private Map<String, Connection> versionConnections;
 
   @BeforeAll
   public void startIntegrationTest() {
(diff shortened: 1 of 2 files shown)


scripts/h2/upgrade-0.9.0-to-1.0.0-h2.sql
```diff
@@ -41,7 +41,7 @@ CREATE TABLE IF NOT EXISTS `policy_version_info` (
     `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'policy deleted at',
     PRIMARY KEY (`id`),
     UNIQUE KEY `uk_pod_ver_del` (`policy_id`, `version`, `deleted_at`),
-    KEY `idx_mid` (`metalake_id`)
+    KEY `idx_pmid` (`metalake_id`)
```
yuqi1129, line 44: What does the `p` in `pmid` mean?

jerqi (approved):

scripts/h2/upgrade-0.9.0-to-1.0.0-h2.sql
mchades: policy

core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+
+    // Close all connections
+    for (Connection conn : versionConnections.values()) {
+      if (conn != null) {
+        conn.close();
+      }
+    }
+  }
```
jerryshao, line 123: I would suggest closing all the connections in the `AfterAll` method. Also, do you need to clean the database after the test is finished?

mchades: > do you need to clean the database after the test is finished?

It's unnecessary because the test instance lifecycle is set to PER_CLASS. Once the test is complete, the database container will be destroyed. (If `jdbcBackend` is H2, the tempDir will also be cleaned automatically.)

jerryshao: So, is this an integration test or UT? Besides, if it relies on the container, should you add the docker test tag?

mchades: It's a UT, and whether to use a container depends on the value of `jdbcBackend`, so it doesn't require the docker test tag.

jerryshao (approved):

scripts/postgresql/upgrade-0.9.0-to-1.0.0-postgresql.sql
zuotingbing, line 1: `CREATE INDEX IF NOT EXISTS policy_relation_meta_idx_stid ON statistic_meta (statistic_id);
CREATE INDEX IF NOT EXISTS policy_relation_meta_idx_moid ON statistic_meta (metadata_object_id);`

should we change the index name to "statistic_meta_idx_stid" and "statistic_meta_idx_moid“？

roryqi: Yes, we should. You can raise a PR to fix this.

roryqi: If you don't have time to fix this, I will fix this issue tomorrow.

Issue #9028, opened 2025-11-06
[Subtask] Refactor the Table API and TableEntity


jerryshao:
### Describe the subtask

Refactor the Table API and TableEntity to remove unnecessary APIs.

### Parent issue

#8828

Issue #8986, opened 2025-10-31
[Subtask] overview the interface semantics for lakehouse table operations


mchades:
### Describe the subtask

We have added the `external` method to the table API:
https://github.com/apache/gravitino/blob/d8a4ded1be927f24fd4ea490a6db36024731d83b/api/src/main/java/org/apache/gravitino/rel/Table.java#L122-L131

This means we will distinguish the metadata operations and data operations for external tables. This may conflict with the existing table operation interface semantics, so we need to review the existing table operation interface semantics to make them correctly support external tables and managed tables.

### Parent issue

#8828

PR #9029, opened 2025-11-06
[#9028] improve(core): Refactor the table and table entity to avoid introducing unnecessary interfaces

jerryshao:

Refactor the table and table entity to avoid introducing unnecessary interfaces

To make the generic lakehouse catalog more stable

Fix: #9028

No.

Exisiting UT and IT.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/rel/Table.java
+++ b/api/src/main/java/org/apache/gravitino/rel/Table.java
@@ -45,24 +45,42 @@ import org.apache.gravitino.tag.SupportsTags;
 public interface Table extends Auditable {
 
   /**
-   * The reserved property name for the table format. This property indicates the format of the
-   * table, such as "iceberg", "hudi", "lance", etc. Some catalog implementations may use this
-   * property to determine the table format and perform specific operations based on the format.
+   * The property name for the table format. This property indicates the format of the table, such
+   * as "iceberg", "hudi", "lance", etc. generic lakehouse catalog will use this property to
+   * determine the table format and perform specific operations based on the format. For other
+   * relational catalogs, this property can be ignored.
+   *
+   * <p>This property is a must-have property for generic lakehouse catalog to create tables, and
+   * cannot be modified after table creation.
+   *
+   * <p>Current supported formats include:
+   *
+   * <ul>
+   *   <li>lance
+   * </ul>
    */
   String PROPERTY_TABLE_FORMAT = "format";
 
   /**
-   * The reserved property name to indicate whether the table is external. This property is a
-   * boolean value represented as a string ("true" or "false"). if true (the table is external), the
-   * drop operation will not delete the underlying data.
+   * The property name to indicate whether the table is external. This property is a boolean value
+   * represented as a string ("true" or "false"). if true (the table is external), the drop
+   * operation will not delete the underlying data.
+   *
+   * <p>This property is optional and defaults to "false" if not specified. Also, it depends on the
+   * catalog implementation to decide whether to leverage and allow to alter this property.
    */
   String PROPERTY_EXTERNAL = "external";
 
   /**
-   * The reserved property name for the table location. This property indicates the physical
-   * location of the table's data, such as a file path or a URI. The location can be specified when
-   * creating the table. If not, the catalog implementation may use a location based on the catalog
-   * and schema location properties.
+   * The property name for the table location. This property indicates the physical location of the
+   * table's data, such as a file path or a URI.
+   *
+   * <p>The location property is optional, it can be specified when creating the table. If not, the
+   * catalog implementation may use a location based on the catalog and schema location properties.
+   *
+   * <p>It depends on the catalog implementation to decide whether to leverage and allow to alter
+   * this property after table creation. And the behavior of altering this property (moving the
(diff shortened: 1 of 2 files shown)


mcp-server/tests/unit/tools/__init__.py
```diff
@@ -14,10 +14,13 @@
...
+import logging
 
 from tests.unit.tools.mock_operation import (
     MockCatalogOperation,
     MockOperation,
     MockSchemaOperation,
     MockTableOperation,
 )
+
+logging.disable(logging.INFO)
```
yuqi1129, line 26: Are you going to disable logging in the MCP testes?

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -44,6 +44,28 @@
...
+   */
+  String PROPERTY_EXTERNAL = "external";
+
+  /**
+   * The reserved property name for the table location. This property indicates the physical
+   * location of the table's data, such as a file path or a URI. The location can be specified when
+   * creating the table. If not, the catalog implementation may use a location based on the catalog
+   * and schema location properties.
+   */
+  String PROPERTY_LOCATION = "location";
```
yuqi1129, line 85: Do we need to mark `location` as an optional reserved property in `catalog` and `schema`?

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -44,6 +44,28 @@
 @Evolving
 public interface Table extends Auditable {
 
+  /**
+   * The reserved property name for the table format. This property indicates the format of the
```
mchades: It would be more appropriate to use an `optional property` here because `reserved property` means the property is reserved by Gravitino (or underlaying catalog) and cannot be set by the user, such as `gravitino.identifier`

https://github.com/apache/gravitino/blob/903f7c81ca15b01c16d4bac39c9dda23da2cc142/core/src/main/java/org/apache/gravitino/connector/PropertyEntry.java#L57

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -44,6 +44,28 @@
...
+   * property to determine the table format and perform specific operations based on the format.
+   */
+  String PROPERTY_TABLE_FORMAT = "format";
+
+  /**
+   * The reserved property name to indicate whether the table is external. This property is a
+   * boolean value represented as a string ("true" or "false"). if true (the table is external), the
+   * drop operation will not delete the underlying data.
+   */
+  String PROPERTY_EXTERNAL = "external";
```
mchades: what's the default value of the property? And is it immutable?

mcp-server/tests/unit/tools/__init__.py
jerryshao: I'm only disabling the annoying info log, which is now directly output to the console.

api/src/main/java/org/apache/gravitino/rel/Table.java
```diff
@@ -44,6 +44,28 @@
 @Evolving
 public interface Table extends Auditable {
 
+  /**
+   * The reserved property name for the table format. This property indicates the format of the
+   * table, such as "iceberg", "hudi", "lance", etc. Some catalog implementations may use this
+   * property to determine the table format and perform specific operations based on the format.
+   */
+  String PROPERTY_TABLE_FORMAT = "format";
```
roryqi, line 62: Do we need more constraints? This is a enum, isn't this? If this a Hive table,   is this a hive?  For jdbc, is this a jdbc?

api/src/main/java/org/apache/gravitino/rel/Table.java
jerryshao: It depends, because for some catalogs and schem,a "location" is not meaningful, I will do it later if necessary.

api/src/main/java/org/apache/gravitino/rel/Table.java
mchades, line 62: Shall we also update the reference in `GenericLakehouseTablePropertiesMetadata` ?
https://github.com/apache/gravitino/blob/32d9642e8a656f1a8e7b0f7da71201951a28b865/catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseTablePropertiesMetadata.java#L50

jerryshao: I will do it separately in the follow-up PR.

api/src/main/java/org/apache/gravitino/rel/Table.java
jerryshao: It depends on the implementation of the different catalogs. Also, it can be set by the user.

mchades: ok, but I think it must be determined when we resolve issue #8986

api/src/main/java/org/apache/gravitino/rel/Table.java
jerryshao: I will add more comments. This property is only leveraged by the generic lakehouse catalog for now.

PR #9035, opened 2025-11-07
[#9036] feat(authz): add document and api for tag access control

hdygxsj:

add document for tag access control

Fix: #9036

None

org.apache.gravitino.authorization.TestSecurableObjects


docs/security/access-control.md
```diff
@@ -264,6 +265,13 @@ DENY `WRITE_FILESET` won‘t deny the `READ_FILESET` operation if the user has t
 | CREATE_MODEL_VERSION | Metalake, Catalog, Schema, Model | Create a model version                                             |
 | USE_MODEL            | Metalake, Catalog, Schema, Model | View the metadata of the model and download all the model versions |
 
+### Tag privileges
+
+| Name       | Supports Securable Object | Operation    |
+|------------|---------------------------|--------------|
+| CREATE_TAG | Metalake                  | Create a tag |
+| APPLY_TAG  | Metalake, Tag             | Apply a tag  |
```
roryqi: https://docs.databricks.com/aws/en/admin/governed-tags/manage-permissions
https://docs.snowflake.com/en/sql-reference/sql/create-tag

roryqi: It seems that the tag is only visible to the users who has privileges.

roryqi: Do we need add a `BROWSE` privilege?

hdygxsj: > It seems that the tag is only visible to the users who has privileges.

This differs from Snowflake's design. In our company's context, consider a shared data warehouse table where multiple users might apply different tags. If everyone can see the tags applied by others—and especially if the company has many employees—it could become quite chaotic. I think I'd prefer to only see the tags that I myself have applied.

hdygxsj: > Do we need add a `BROWSE` privilege?

That’s one approach, but it creates a challenge for managers or data governance personnel: if I want everyone to see certain tags—such as “No Modifications Allowed,” “No Access Allowed,” “Core Table,” or “Critical Table”—I’d have to grant permissions individually, which would be cumbersome and inefficient.

roryqi:
@mchades @FANNG1 Could you give some input?

docs/security/access-control.md
hdygxsj: > > Do we need add a `BROWSE` privilege?
> 
> That’s one approach, but it creates a challenge for managers or data governance personnel: if I want everyone to see certain tags—such as “No Modifications Allowed,” “No Access Allowed,” “Core Table,” or “Critical Table”—I’d have to grant permissions individually, which would be cumbersome and inefficient.

Take that back—I forgot that this privilege can be assigned to a group or role. Maybe this is actually a good idea.

hdygxsj:
Another issue is: Can the owner of a metadata object see all tags associated with that metadata? If so, this could lead to a situation where certain tags aren't visible when a regular user lists tags, but become visible when listing tags directly under the metadata object (e.g., from the owner’s perspective).

roryqi:
> Another issue is: Can the owner of a metadata object see all tags associated with that metadata? If so, this could lead to a situation where certain tags aren't visible when a regular user lists tags, but become visible when listing tags directly under the metadata object (e.g., from the owner’s perspective).

Yes, if so, the only person who has the metalake level apply_tag can manage all the tags for every metalake object.

docs/security/access-control.md
roryqi: Maybe I can don't add `BROWSE` now. We can add the privilege `BROWSE` in the future.

docs/security/access-control.md
```diff
@@ -1025,4 +1033,13 @@ The following table lists the required privileges for each API.
 | grant privilege                   | `MANAGE_GRANTS` on the metalake or the owner of the securable object                                                                                                                                                                          |
 | revoke privilege                  | `MANAGE_GRANTS` on the metalake or the owner of the securable object                                                                                                                                                                          |
 | set owner                         | The owner of the securable object                                                                                                                                                                                                             |
+| list tags                         | `APPLY_TAG` on the metalake, the owner of the metalake, or the tag itself.                                                                                                                                                                    |
```
roryqi: Remove `itself`.

docs/security/access-control.md
```diff
@@ -1025,4 +1033,13 @@ The following table lists the required privileges for each API.
...
+| create tag                        | `CREATE_TAG` on the metalake or the owner of the metalake.                                                                                                                                                                                    |
+| get tag                           | `APPLY_TAG` on the metalake or tag, the owner of the metalake, or the tag itself.                                                                                                                                                             |
```
roryqi: Remove `itself`.

roryqi:
@mchades Could you give me some input?

docs/security/access-control.md
mchades: What does "Apply a tag" mean? Does it include all tag operations except "Create a tag"? I think it's not appropriate to simply use "Apply a tag"; a more detailed description is needed.

docs/security/access-control.md
```diff
@@ -1025,4 +1033,13 @@ The following table lists the required privileges for each API.
...
+| alter tag                         | Must be the owner of the metalake or the tag.                                                                                                                                                                                                 |
+| delete tag                        | Must be the owner of the metalake or the tag.                                                                                                                                                                                                 |
```
mchades, line 1040: Does this mean that the tag also has an independent owner?

docs/security/access-control.md
mchades: Does this mean that users who do not meet specific conditions will encounter a permission deny error when calling listTags? Or does it mean that all users can `listTags`, but tags will be filtered based on user permissions?

docs/security/access-control.md
hdygxsj: It is only used to associate tags with metadata objects.

docs/security/access-control.md
hdygxsj: yes

hdygxsj:
@mchades tags will be filtered based on user permissions

docs/security/access-control.md
mchades: > tags will be filtered based on user permissions

@hdygxsj So a more accurate description here should be similar to `list schema`

docs/security/access-control.md
```diff
@@ -1025,4 +1033,13 @@ The following table lists the required privileges for each API.
...
+| list objects for tag              | Requires both permission to **get the tag** and permission to **load metadata objects**.                                                                                                                                                      |
+| list tags for object              | Permission to both list tags Requires both permission to **list tags** and permission to **load metadata objects**. load metadata objects is required.                                                                                        |
```
mchades, line 1042: Assume user A is the owner of `catalog1`, he created `tag1` and associated it with `catalog1`; user B is the owner of `catalog1.schema1.table1`, when user B list tags for `table1`, what permissions are needed to see `tag1`? Or what permissions should be granted by which users to user B so that he can see `tag1`?

hdygxsj: Maybe we need the tag1's owner (user A) to grant BROWER_TAG to user B

docs/security/access-control.md
hdygxsj: > > tags will be filtered based on user permissions
> 
> @hdygxsj So a more accurate description here should be similar to `list schema`

fixed

docs/security/access-control.md
hdygxsj: fixed

docs/security/access-control.md
roryqi: We can add `BROWSE` privilege later.

docs/security/access-control.md
mchades: I think you should also update the doc

hdygxsj: fixed

docs/security/access-control.md
mchades: If not added in the current PR, what is the current default behavior? If the `BROWSE` privilege is added later, will it break compatibility?

roryqi: The user can only see the tags which he has privileges.

Issue #9021, opened 2025-11-05
[Subtask] Support lndex related APIs for Lance REST server.


yuqi1129:
### Describe the subtask

Create index and describe index should be added

### Parent issue

#8889

PR #9043, opened 2025-11-07
[#9021] feat(lance-rest): Support create & list index for LRS

yuqi1129:

Add create and list index APIs in the Lance rest server.

It's a feature that benefits users.

Fix: #9021 

N/A.

UTs and ITs


gradle/libs.versions.toml
```diff
@@ -28,8 +28,8 @@ guava = "32.1.3-jre"
 lombok = "1.18.20"
 slf4j = "2.0.16"
 log4j = "2.24.3"
-lance = "0.34.0"
-lance-namespace = "0.0.19"
+lance = "0.39.0"
```
yuqi1129, line 31: The old version will fail to create indices.

yuqi1129:
I clicked Merge by mistake and will revert it.

Issue #8730, opened 2025-09-30
[Improvement] Improve CLI distribution output


justinmclean:
### What would you like to be improved?

In clients/cli/src/main/java/org/apache/gravitino/cli/commands/TableDistribution.java fix distribution output when distribution is none.

JeonDaehong:
Would it be okay if I contribute to this issue?

justinmclean:
Thanks for offering to help.

chanwonlee:
@justinmclean If this issue is not being actively worked on, would it be okay if I work on it?

justinmclean:
@chanwonlee sure you can work on it.

PR #9066, opened 2025-11-11
[#8730][Improvement] Fix CLI distribution output when distribution is none

chanwonlee:

This PR fixes the CLI distribution output when the distribution strategy is none.

When distribution strategy is NONE, displaying the number is confusing and unnecessary. This improvement makes the output cleaner and more meaningful by only showing "NONE".

Fix: #8730

- Before: `NONE,0`
- After: `NONE`

N/A

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/converter/DorisColumnDefaultValueConverter.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/converter/DorisColumnDefaultValueConverter.java
@@ -96,7 +96,7 @@ public class DorisColumnDefaultValueConverter extends JdbcColumnDefaultValueConv
       case JdbcTypeConverter.TEXT:
         return Literals.stringLiteral(columnDefaultValue);
       default:
-        throw new IllegalArgumentException("Unknown data columnType for literal: " + columnType);
+        return UnparsedExpression.of(columnDefaultValue);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


justinmclean:
Thanks for the cleanup, but I’m concerned about backward compatibility. Some users may treat the CLI output as machine-parsable CSV with a fixed column count. Changing 'NONE,0' to 'NONE' collapses a column and potentially break existing scripts.

chanwonlee:
> Thanks for the cleanup, but I’m concerned about backward compatibility. Some users may treat the CLI output as machine-parsable CSV with a fixed column count. Changing 'NONE,0' to 'NONE' collapses a column and potentially break existing scripts.

@justinmclean Would it be okay if I change it like this?
```java
if (distribution.strategy() == Strategy.NONE) {
  printInformation(distribution.strategy() + "," + "0");
} else {
  printInformation(distribution.strategy() + "," + distribution.number());
}
```

justinmclean:
Didn't it display "NONE,0" before?

chanwonlee:
> Didn't it display "NONE,0" before?

Sorry for the confusion, this is my first open source contribution and I'm still learning the process.

It is correct that the original code already outputs "none,0".
I misunderstood the issue and would appreciate clarification on what needs to be fixed.

Looking at the original code:
```java
printInformation(distribution.strategy() + "," + distribution.number());
```
This already outputs "NONE,0" when the distribution is NONE.
Could you help me understand what the actual issue is? I see two possibilities:

1. If "NONE,0" is the correct output, then no code changes are needed
2. If we want to modify the output for NONE (e.g., to make it clearer), I'm not sure what format would be best while maintaining backward compatibility

I'd appreciate your guidance on the expected behavior so I can make the right changes.

justinmclean:
I'm sorry, I didn't clearly explain the issue in #8730. I'll need to review this again, and I'll get back to you tomorrow.

justinmclean:
Looking at this again, I think your fix is the right approach, and the risk of breaking things is very low.

justinmclean (approved):
Thanks for the improvment.

Issue #8824, opened 2025-10-16
[Bug report] Role didn't update when the table is deleted


danhuawang:
### Version

main branch

### Describe what's wrong

**1. Drop table `hive_catalog1.hive_schema1.hive_table1`, I also check the table not existed**
<img width="1481" height="693" alt="Image" src="https://github.com/user-attachments/assets/f1040d4d-3d1f-47e9-a64c-42fc35913930" />

<img width="1081" height="787" alt="Image" src="https://github.com/user-attachments/assets/7e9dae69-3a68-4075-94c3-29774d02f044" />

2. Check the table related role, the table still in the securableObjects. It's supposed to be removed in the role

<img width="1480" height="743" alt="Image" src="https://github.com/user-attachments/assets/c6ba2e58-2ffd-43e5-a5d3-7f2ce1d863e8" />

<img width="1265" height="763" alt="Image" src="https://github.com/user-attachments/assets/640879b3-adc1-49ab-9253-d261f2f2cdaf" />

### Error message and/or stacktrace

see abobe

### How to reproduce

```
 Scenario: Delete metadata objects that defined in role
    Given create metalake test
    Given create hive catalog hive_catalog1 in metalake test
    Given create hive schema hive_schema1 with the location /user/hive/warehouse/hive_schema1.db in catalog hive_catalog1 metalake test
    Given create hive table hive_table1 in schema hive_schema1 catalog hive_catalog1 metalake test
    When Create Role table reader
    When Create Role table writer
    When drop table hive_table1 in schema hive_schema1 catalog hive_catalog1 metalake test
    Then RoleTableReader RoleTableWriter role's table securableObjects is removed
```

RoleTableReader role as following:

```
{
    "code": 0,
    "role": {
        "name": "RoleTableReader",
        "audit": {
            "creator": "anonymous",
            "createTime": "2025-10-20T03:15:45.391737Z"
        },
        "properties": {
            "k1": "v1"
        },
        "securableObjects": [
            {
                "type": "catalog",
                "privileges": [
                    {
                        "name": "use_catalog",
                        "condition": "allow"
                    }
                ],
                "fullName": "hive_catalog1"
            },
            {
                "type": "schema",
                "privileges": [
                    {
                        "name": "use_schema",
                        "condition": "allow"
                    }
                ],
                "fullName": "hive_catalog1.hive_schema1"
            },
            {
                "type": "table",
                "privileges": [
                    {
                        "name": "select_table",
                        "condition": "allow"
[...]

Issue #8817, opened 2025-10-15
[Improvement] Improve reverse index in caffeine entity cache


yuqi1129:
### What would you like to be improved?

Assuming there are three tags (tag1, tag2, tag3) that has been associated with catalog1, then the data in cacheData will be 

```
catalog1 ->  List<tag1, tag2, tag3>
```

Data in the reverse index will be 

```
tag1 -> catalog1
tag2 -> catalog1
tag3 -> catalog1
```

However, if catalog2 also associated with tag1, tag2, tag3, then data will be 

```
cache data: 
cataog2 -> List<tag1, tag2, tag3>

reverse index 

tag1 -> catalog2 (overwrite catalog1)
tag2 -> catalog2 (overwrite catalog1)
tag3 -> catalog2 (overwrite catalog1)
```

It seems the value of reversed index should also a list.

PR #9068, opened 2025-11-11
[#8824] fix(core): Fix bugs in the logic about reverse index in entity store Caffeine cache

yuqi1129:

Use List<Key> NOT Key in the reversed index. 

Currently, the reversed index for the Caffine cache is as followings. 

Assuming role1 has a securable object table1, then the data in the reversed index will be 

```text

KEY         VALUE
table1     role1 
```

When role2 also has a securable object table1, the index data will overwrite that of role1 and the final data will be 
```text
KEY         VALUE
table1     role2 
```

When table1 is deleted, Gravitino will only invalidate the data of role2, and role1 will still hold the out-of-date data. 

Fix: #8824 
Fix: #8817

N/A

UTs

### DIFF
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -374,10 +374,11 @@ public class CaffeineEntityCache extends BaseEntityCache {
 
     if (valueForExactKey == null) {
       // It means the key does not exist in the cache. However, we still need to handle some cases.
-      // For example, we have store a role entity in the cache and entity to role in the reverse
-      // index. This is: key -> role, reverse index: securable object -> role. When will update the
-      // securable object, we need to invalidate the role from the cache even the securable object
-      // is not in the cache.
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -366,21 +366,29 @@ private boolean invalidateEntities(
...
+                ? EntityCacheRelationKey.of(identifier, type).toString()
                 : EntityCacheRelationKey.of(identifier, type, relTypeOpt.get()).toString());
 
     if (valueForExactKey == null) {
-      // No key to remove
-      return false;
+      // It means the key does not exist in the cache. However, we still need to handle some cases.
+      // For example, we have store a role entity in the cache and entity to role in the reverse
+      // index. This is: key -> role, reverse index: securable object -> role. When will update the
+      // securable object, we need to invalidate the role from the cache even the securable object
```
mchades: what's the meaning of "When will update the view securable object"?

Issue #9022, opened 2025-11-05
[Subtask] Refactor to add `ManagedTableOperations` in Gravitino


jerryshao:
### Describe the subtask

Generic lakehouse catalog manages the table in Gravitino. With this, we need to abstract the common table metadata operation layer and define in `core`.

### Parent issue

#8828

PR #9073, opened 2025-11-11
[#9022] improve(core): Refactor to add the managed table operations

jerryshao:

This PR adds a new managed table operations class for managed table CRUD in Gravitino.

This is the prerequisite for refactoring the generic lakehouse catalog.

Fix: #9022

No.

Add UTs to cover the code.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
@@ -75,9 +75,10 @@ public abstract class ManagedTableOperations implements TableCatalog {
   @Override
   public NameIdentifier[] listTables(Namespace namespace) throws NoSuchSchemaException {
     try {
-      NameIdentifier schemaIdent = NameIdentifier.of(namespace.levels());
-      schemas().loadSchema(schemaIdent); // verify schema existence
-
+      // The current implementation of JDBC entity store will automatically check the existence of
+      // the namespace when listing entities under the namespace. If not, it will throw an
+      // NoSuchEntityException. So, we don't need to check the existence of the namespace here
+      // again.
       List<TableEntity> tables =
           store().list(namespace, TableEntity.class, Entity.EntityType.TABLE);
       return tables.stream()
@@ -119,13 +120,6 @@ public abstract class ManagedTableOperations implements TableCatalog {
     // It doesn't handle any additional operations like creating physical location, preprocessing
     // the properties, etc. Those operations should be handled in the specific catalog
     // implementation.
-
-    NameIdentifier schemaIdent = NameIdentifier.of(ident.namespace().levels());
-
-    if (!schemas().schemaExists(schemaIdent)) {
-      throw new NoSuchSchemaException("Schema %s does not exist", schemaIdent);
-    }
-
     StringIdentifier stringId = StringIdentifier.fromProperties(properties);
     Preconditions.checkArgument(stringId != null, "Property String identifier should not be null");
 
@@ -152,6 +146,12 @@ public abstract class ManagedTableOperations implements TableCatalog {
 
     try {
       store().put(tableEntity, false /* overwrite */);
+    } catch (NoSuchEntityException e) {
+      // The put operation in the current JDBC entity store will check the existence of the
+      // namespace when creating an entity under the namespace. If not, it will throw a
+      // NoSuchEntityException. So, we don't need to check the existence of the namespace here
+      // again.
+      throw new NoSuchSchemaException(e, "Schema %s does not exist", ident.namespace());
     } catch (EntityAlreadyExistsException e) {
       throw new TableAlreadyExistsException(e, "Table %s already exists", ident);
     } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -0,0 +1,555 @@
...
+
+  protected abstract SupportsSchemas schemas();
+
+  protected abstract IdGenerator idGenerator();
+
+  @Override
+  public NameIdentifier[] listTables(Namespace namespace) throws NoSuchSchemaException {
+    try {
+      NameIdentifier schemaIdent = NameIdentifier.of(namespace.levels());
+      schemas().loadSchema(schemaIdent); // verify schema existence
```
mchades: It seems unnecessary, since `store().list(namespace, TableEntity.class, Entity.EntityType.TABLE)` will automatically determine if the schema exists.

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -0,0 +1,555 @@
...
+    // createTable in ManagedTableOperations only stores the table metadata in the entity store.
+    // It doesn't handle any additional operations like creating physical location, preprocessing
+    // the properties, etc. Those operations should be handled in the specific catalog
+    // implementation.
+
+    NameIdentifier schemaIdent = NameIdentifier.of(ident.namespace().levels());
+
+    if (!schemas().schemaExists(schemaIdent)) {
+      throw new NoSuchSchemaException("Schema %s does not exist", schemaIdent);
+    }
```
mchades: It seems unnecessary, since we can catch `NoSuchEntityException` when call `store().put(tableEntity, false /* overwrite */)`

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -0,0 +1,555 @@
...
+            .map(change -> (TableChange.ColumnChange) change)
+            .collect(Collectors.toList());
+    List<TableChange> tableChanges = splitChanges.get(false);
+
+    for (TableChange change : tableChanges) {
+      if (change instanceof TableChange.RenameTable rename) {
+        if (rename.getNewSchemaName().isPresent()) {
+          throw new IllegalArgumentException(
+              "Gravitino managed table doesn't support renaming "
+                  + "the table across schemas for now");
```
mchades, line 231: will we support this later?

core/src/main/java/org/apache/gravitino/connector/GenericLakehouseTable.java
```diff
@@ -47,11 +47,16 @@ protected GenericLakehouseTable internalBuild() {
...
-              .put(Table.PROPERTY_TABLE_FORMAT, this.format)
-              .buildKeepingLast();
+
+      if (format != null) {
+        genericLakehouseTable.properties =
+            ImmutableMap.<String, String>builder()
+                .putAll(this.properties)
+                .put(Table.PROPERTY_TABLE_FORMAT, this.format)
+                .buildKeepingLast();
+      } else {
```
mchades, line 57: Should we allow the `format` property in `GenericLakehouseTable` to be null?

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -0,0 +1,555 @@
...
+    } catch (NoSuchEntityException e) {
+      return false;
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to drop metadata for table " + ident, e);
+    }
+  }
+
+  private TableEntity applyChanges(TableEntity oldTableEntity, TableChange... changes) {
+    String newName = oldTableEntity.name();
+    String newComment = oldTableEntity.comment();
```
yuqi1129, line 211: The names of these two variants are quite misleading, and they should be `oldXXX` intuitively.

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -0,0 +1,555 @@
...
+              position = i + 1;
+              found = true;
+              break;
+            }
+          }
+          if (!found) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "Column %s not found for adding column after it", afterColumn.getColumn()));
+          }
```
yuqi1129: Those lines are duplicated with L414-L427

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -0,0 +1,555 @@
...
+        }
+
+      } else {
+        throw new IllegalArgumentException("Unsupported column change: " + change);
+      }
+    }
+
+    // After the column adding, updating, deleting, the positions of the columns may be messed up.
+    // We need to reassign the positions to ensure they are continuous and start from 0.
+    return updateColumnPositions(newColumns);
```
yuqi1129, line 426: This method is rather long; please try to split it into several minor ones.

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
jerryshao: I see, I will fix it.

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
jerryshao: These two variables will be updated to the new one if there're "rename" or "update comment" change. So I deliberately named like this.

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
jerryshao: I will check the implementation of the `TableMetaService` to see if it needs any update. I think this can be done in a separate PR.

core/src/main/java/org/apache/gravitino/connector/GenericLakehouseTable.java
jerryshao: No, for the current implementation, it is not allowed.

yuqi1129:
@jerryshao 
I have used the branch `branch-lance-namespace-dev` to replace the old one `branch-lance-namepspace-dev`, please just raise a PR based on it.

core/src/main/java/org/apache/gravitino/connector/GenericLakehouseTable.java
mchades: if so, should we throw an exception in the `else` block?

mchades:
one minor comment.
https://github.com/apache/gravitino/pull/9073#discussion_r2521330979

yuqi1129:
> lgtm. Do you have more comments? @yuqi1129

I have no more comments, let's wait https://github.com/apache/gravitino/pull/9073#issuecomment-3530548242 to be resolved.

core/src/main/java/org/apache/gravitino/connector/GenericLakehouseTable.java
jerryshao: This is the temporary code, I will remove it in the follow-up PR.

PR #9086, opened 2025-11-12
[#9084] improvement(core): Fix race condition in AuthorizationRequestContext.loadRole()

### LINKED ISSUE #9084: [Improvement] Fix race condition in loadRole

The loadRole method (in core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java)  isn’t thread-safe as it has a race condition. Because hasLoadRole.get() and hasLoadRole.set(true) are separate operations, multiple threads can see false and run the runnable simultaneously.

Suppose two threads call loadRole() at roughly the same time: 
- Both check hasLoadRole.get() before it’s set and both see false.
- Both enter the if body and run runnable.run().
- Then both set hasLoadRole.set(true) afterward.

So you can end up with double (or more) executions of the runnable.

Here's a unit test to illustrate:
```
@Test
  public void testLoadRoleRunsOnceEvenWhenInvokedConcurrently() throws Exception {
    AuthorizationRequestContext context = new AuthorizationRequestContext();
    AtomicInteger counter = new AtomicInteger();
    CountDownLatch firstStarted = new CountDownLatch(1);
    CountDownLatch allowFinish = new CountDownLatch(1);

    Thread firstInvocation =
        new Thread(
            () ->
                context.loadRole(
                    () -> {
                      counter.incrementAndGet();
                      firstStarted.countDown();
                      try {
                        allowFinish.await(5, TimeUnit.SECONDS);
                      } catch (InterruptedException e) {
                        Thread.currentThread().interrupt();
                      }
                    }));
    firstInvocation.start();

    try {
      assertTrue(firstStarted.await(5, TimeUnit.SECONDS));
      context.loadRole(() -> counter.incrementAndGet());
      assertEquals(1, counter.get(), "loadRole should execute runnable only once");
    } finally {
      allowFinish.countDown();
      firstInvocation.join();
    }

    context.loadRole(() -> counter.incrementAndGet());
    assertEquals(1, counter.get(), "Subsequent loadRole calls should be ignored");
  }
```

### How should we improve?

AtomicBoolean makes single operations thread-safe, but not combinations of them, use compareAndSet() instead.

HarshMehta112:
Fixed #9084

- Replaced separate `hasLoadRole.get()` and `hasLoadRole.set(true)` calls with an atomic  
  `hasLoadRole.compareAndSet(false, true)` operation.
- This makes the check-and-set atomic, ensuring the runnable executes only once,  
  even under concurrent invocations.
- Previously, multiple threads could see `false` and execute the runnable simultaneously,  
  causing duplicate executions.

- Added a new unit test: `testLoadRoleRunsOnceEvenWhenInvokedConcurrently`.
- Verified that the runnable executes only once, even when `loadRole()` is called concurrently.

### Result
Improved thread-safety of `loadRole()` and prevented race conditions during concurrent access.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/authorization/TestAuthorizationRequestContext.java
+++ b/core/src/test/java/org/apache/gravitino/authorization/TestAuthorizationRequestContext.java
@@ -71,7 +71,6 @@ public class TestAuthorizationRequestContext {
     CountDownLatch failingStarted = new CountDownLatch(1);
     CountDownLatch allowFailToThrow = new CountDownLatch(1);
 
-    // --- Thread 1: Starts loading but FAILS ---
     Thread failingThread =
         new Thread(
             () -> {


justinmclean:
Please see the conversation in #9084. While this does fix the race condition it introduces some other issues and I think a little more needs to be done here.

core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
```diff
@@ -77,11 +77,9 @@ public boolean authorizeDeny(
   }
 
   public void loadRole(Runnable runnable) {
-    if (hasLoadRole.get()) {
-      return;
+    if (hasLoadRole.compareAndSet(false, true)) {
```
hdygxsj: This could cause some threads to skip loadRole under concurrent conditions, leading to incorrect results.

core/src/test/java/org/apache/gravitino/authorization/TestAuthorizationRequestContext.java
```diff
@@ -0,0 +1,64 @@
...
+                        allowFinish.await(5, TimeUnit.SECONDS);
+                      } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                      }
+                    }));
+    firstInvocation.start();
+
+    try {
+      assertTrue(firstStarted.await(5, TimeUnit.SECONDS));
+      context.loadRole(counter::incrementAndGet);
```
hdygxsj, line 55: Perhaps we’d better add a test case to ensure that once loadRole is called, the role is actually loaded.

roryqi:
This method won't be called by multiple threads. So we don't need fix.

hdygxsj:
> This method won't be called by multiple threads. So we don't need fix.

authorizationRequestContext will be accessed concurrently here.
 

image

hdygxsj (commented):
LGTM

PR #9091, opened 2025-11-12
[#8546] improvement: Retrieve metadata object id using cache

roryqi:

 Retrieve metadata object id using cache 

Fix: #8546

No.

Existing test cases.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
@@ -62,7 +62,7 @@ public class RelationalEntityStoreIdResolver implements EntityIdResolver {
   @Override
   public NamespacedEntityId getEntityIds(NameIdentifier nameIdentifier, Entity.EntityType type) {
     if (ENTITY_TYPES_REQUIRING_METALAKE_ID.contains(type)) {
-      return getEntityIdsAboutMetalake(nameIdentifier, type);
+      return getEntityIdsRequiringMetalakeId(nameIdentifier, type);
 
     } else if (ENTITY_TYPES_REQURING_CATALOG_IDS.contains(type)) {
       CatalogIds catalogIds =
@@ -73,7 +73,7 @@ public class RelationalEntityStoreIdResolver implements EntityIdResolver {
       return new NamespacedEntityId(catalogIds.getCatalogId(), catalogIds.getMetalakeId());
 
     } else if (ENTITY_TYPES_REQURING_SCHEMA_IDS.contains(type)) {
-      return getEntityIdsAboutSchema(nameIdentifier, type);
+      return getEntityIdsRequiringSchemaIds(nameIdentifier, type);
 
     } else {
       throw new IllegalArgumentException("Unsupported entity type: " + type);
@@ -83,7 +83,7 @@ public class RelationalEntityStoreIdResolver implements EntityIdResolver {
   @Override
   public Long getEntityId(NameIdentifier nameIdentifier, Entity.EntityType type) {
     if (ENTITY_TYPES_REQUIRING_METALAKE_ID.contains(type)) {
-      return getEntityIdsAboutMetalake(nameIdentifier, type).entityId();
+      return getEntityIdsRequiringMetalakeId(nameIdentifier, type).entityId();
 
     } else if (ENTITY_TYPES_REQURING_CATALOG_IDS.contains(type)) {
       return CatalogMetaService.getInstance()
@@ -93,14 +93,14 @@ public class RelationalEntityStoreIdResolver implements EntityIdResolver {
           .getCatalogId();
 
     } else if (ENTITY_TYPES_REQURING_SCHEMA_IDS.contains(type)) {
-      return getEntityIdsAboutSchema(nameIdentifier, type).entityId();
+      return getEntityIdsRequiringSchemaIds(nameIdentifier, type).entityId();
 
     } else {
       throw new IllegalArgumentException("Unsupported entity type: " + type);
     }
   }
 
-  private NamespacedEntityId getEntityIdsAboutMetalake(
+  private NamespacedEntityId getEntityIdsRequiringMetalakeId(
       NameIdentifier nameIdentifier, Entity.EntityType type) {
     long metalakeId =
         MetalakeMetaService.getInstance()
@@ -138,7 +138,7 @@ public class RelationalEntityStoreIdResolver implements EntityIdResolver {
     }
   }
 
-  private NamespacedEntityId getEntityIdsAboutSchema(
+  private NamespacedEntityId getEntityIdsRequiringSchemaIds(
       NameIdentifier nameIdentifier, Entity.EntityType type) {
     SchemaIds schemaIds =
         SchemaMetaService.getInstance()


core/src/main/java/org/apache/gravitino/utils/NameIdentifierUtil.java
```diff
@@ -589,4 +599,49 @@ public static NameIdentifier getModelIdentifier(NameIdentifier ident) {
...
+      return NameIdentifier.of(nameIdentifier.namespace().levels());
+    } else {
+      throw new IllegalArgumentException("The entity has no parent name identifier");
+    }
+  }
+
+  public static Entity.EntityType parentEntityType(Entity.EntityType type) {
+    switch (type) {
+      case COLUMN:
+        return Entity.EntityType.TABLE;
```
roryqi, line 634: Add a blank line.

core/src/main/java/org/apache/gravitino/utils/NameIdentifierUtil.java
```diff
@@ -589,4 +599,49 @@ public static NameIdentifier getModelIdentifier(NameIdentifier ident) {
     }
     return NameIdentifier.of(allElems.get(0), allElems.get(1), allElems.get(2), allElems.get(3));
   }
+
+  public static NameIdentifier parentNameIdentifier(
+      NameIdentifier nameIdentifier, Entity.EntityType type) {
+    Set<Entity.EntityType> supportsVirtualNamespaceTypes =
+        ImmutableSet.of(USER, GROUP, ROLE, TAG, POLICY, JOB, JOB_TEMPLATE);
+    if (supportsVirtualNamespaceTypes.contains(type)) {
+      return NameIdentifier.of(NameIdentifierUtil.getMetalake(nameIdentifier));
```
roryqi, line 621: Add a blank line.

hdygxsj:
image

According to the flame graph analysis, querying the owners of approximately 5,000 tables takes about 2 seconds, and querying schema IDs also takes about 2 seconds. Therefore, I believe the cache for schema ID lookups is not working effectively.

hdygxsj:
image

I think this is caused by the cache not being populated (or backfilled) here.

roryqi:
> image
> I think this is caused by the cache not being populated (or backfilled) here.

Fixed.

core/src/main/java/org/apache/gravitino/meta/EntityIds.java
```diff
@@ -0,0 +1,52 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.meta;
+
+import com.google.common.base.Preconditions;
+
+public class EntityIds {
```
yuqi1129: `EntityIds` is quite misleading, it makes me think it's a list of entity ID.

core/src/main/java/org/apache/gravitino/meta/EntityIds.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public long[] namespaceIds() {
+    return namespaceIds;
+  }
+
+  public long[] fullIds() {
+    long[] allIds = new long[namespaceIds.length + 1];
+    System.arraycopy(namespaceIds, 0, allIds, 0, namespaceIds.length);
+    allIds[allIds.length - 1] = entityId;
+    return allIds;
```
yuqi1129:
```suggestion
return ArrayUtils.add(namespaceIds, entityId);
```

core/src/main/java/org/apache/gravitino/meta/EntityIdResolver.java
```diff
@@ -0,0 +1,44 @@
...
+public interface EntityIdResolver {
+
+  /**
+   * Get EntityIds for the given NameIdentifier and Entity type.
+   *
+   * @param nameIdentifier NameIdentifier of the entity.
+   * @param type Entity type.
+   * @return EntityIds corresponding to the NameIdentifier and Entity type.
+   */
+  EntityIds getEntityIds(NameIdentifier nameIdentifier, Entity.EntityType type);
```
yuqi1129: This method is enough to meet our needs, but why do we add the following method as well?

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.gravitino.storage.relational.service.SchemaMetaService;
+import org.apache.gravitino.storage.relational.service.TableColumnMetaService;
+import org.apache.gravitino.storage.relational.service.TableMetaService;
+import org.apache.gravitino.storage.relational.service.TagMetaService;
+import org.apache.gravitino.storage.relational.service.TopicMetaService;
+import org.apache.gravitino.storage.relational.service.UserMetaService;
+import org.apache.gravitino.utils.NameIdentifierUtil;
+
+public class RelationalEntityStoreIdResolver implements EntityIdResolver {
+  private static final Set<Entity.EntityType> needMetalakeId =
```
yuqi1129: Better to use UPPER_CASE name.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
```diff
@@ -0,0 +1,205 @@
...
+      ImmutableSet.of(
+          Entity.EntityType.METALAKE,
+          Entity.EntityType.ROLE,
+          Entity.EntityType.USER,
+          Entity.EntityType.GROUP,
+          Entity.EntityType.TAG);
+  private static final Set<Entity.EntityType> needCatalogIds =
```
yuqi1129: I'm not very clear about why the entity catalog needs `needCatalogIds` here? Could you try to rename them or add more comments here?

core/src/main/java/org/apache/gravitino/storage/relational/helper/MetadataObjectIds.java
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.storage.relational.helper;
+
+public class MetadataObjectIds {
```
yuqi1129: It's unused, and maybe you need to remove it.

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
```diff
@@ -92,9 +92,18 @@ public CatalogPO getCatalogPOByName(String metalakeName, String catalogName) {
...
   public CatalogIds getCatalogIdByMetalakeAndCatalogName(String metalakeName, String catalogName) {
-    return SessionUtils.getWithoutCommit(
-        CatalogMetaMapper.class,
-        mapper -> mapper.selectCatalogIdByMetalakeNameAndCatalogName(metalakeName, catalogName));
+    CatalogIds catalogIds =
+        SessionUtils.getWithoutCommit(
+            CatalogMetaMapper.class,
+            mapper ->
+                mapper.selectCatalogIdByMetalakeNameAndCatalogName(metalakeName, catalogName));
+    if (catalogIds == null) {
```
yuqi1129, line 100: Why does it work fine without adding this check? Is the reason that no test have covered this?

core/src/main/java/org/apache/gravitino/utils/NameIdentifierUtil.java
```diff
@@ -600,4 +610,52 @@ public static NameIdentifier getModelIdentifier(NameIdentifier ident) {
     }
     return NameIdentifier.of(allElems.get(0), allElems.get(1), allElems.get(2), allElems.get(3));
   }
+
+  public static NameIdentifier parentNameIdentifier(
+      NameIdentifier nameIdentifier, Entity.EntityType type) {
+    Set<Entity.EntityType> supportsVirtualNamespaceTypes =
```
yuqi1129: Please just set it as a static final field.

core/src/main/java/org/apache/gravitino/meta/EntityIdResolver.java
roryqi: This method isn't effective enough if we using cache.

core/src/main/java/org/apache/gravitino/meta/EntityIds.java
roryqi: How about `NamespacedEntityId`?

yuqi1129: I believe it's better than the current one.

core/src/main/java/org/apache/gravitino/storage/relational/helper/MetadataObjectIds.java
roryqi: Removed.

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
roryqi: We add check in the upper layer.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
roryqi: Renamed according to GPT suggestions.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
roryqi: RENAMED.

core/src/main/java/org/apache/gravitino/utils/NameIdentifierUtil.java
roryqi: Fxied.

core/src/main/java/org/apache/gravitino/meta/EntityIds.java
roryqi: Renamed.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStoreIdResolver.java
```diff
@@ -0,0 +1,206 @@
...
+
+    } else if (ENTITY_TYPES_REQURING_SCHEMA_IDS.contains(type)) {
+      return getEntityIdsAboutSchema(nameIdentifier, type).entityId();
+
+    } else {
+      throw new IllegalArgumentException("Unsupported entity type: " + type);
+    }
+  }
+
+  private NamespacedEntityId getEntityIdsAboutMetalake(
```
yuqi1129: The method name should also be in sync with the variant name `xxxx_requiring_xxx`.

roryqi: Fixed.

core/src/main/java/org/apache/gravitino/cache/CachedEntityIdResolver.java
```diff
@@ -0,0 +1,64 @@
...
+  private final EntityIdResolver underlyingResolver;
+
+  public CachedEntityIdResolver(EntityCache entityCache, EntityIdResolver underlyingResolver) {
+    this.entityCache = entityCache;
+    this.underlyingResolver = underlyingResolver;
+  }
+
+  @Override
+  public NamespacedEntityId getEntityIds(NameIdentifier nameIdentifier, Entity.EntityType type) {
+    return entityCache
```
yuqi1129, line 40: Since `CachedEntityIdResolver` will share the cache with `EntityStore`, I am wondering if the value of id resolver will overwrite that of entity store, as I noticed that you used the same cache key(NameIdentifier, EntityType) here.

roryqi: Do u means add some interfaces for EntityStore? I ever thought of it. But I prefer not putting them together. Now it's clearer. They are different level abstraction. 
1. EntityStore provides basic entity management.
2. SupportsRelationalOps provides relation management
3. Id resolver provider effective id - entity mapping.
We can share common implementation. But we should keep the separation of interfaces.

yuqi1129: What concerns me is that the cache key of the  entity store cache is `NameIdentier` and `EntityType.`, and that is the same as what you did for entity id, that is 

The following two entries will never be in the cache at the same time. 
```
<ident1, table> ---> table1 entity
<ident, table>  ---> table1 id
```
This PR will degrade the effectiveness of the entity store cache for the entity, though we will get some improvement on ID retrieval.

yuqi1129: I misunderstood it.

yuqi1129:
It looks good to me, @jerryshao would you like to take a look?

Issue #9081, opened 2025-11-12
[Improvement] In CatalogOperations.java fix exception


justinmclean:
### What would you like to be improved?

Fix the catalog update error handler to use the correct operation type from the incoming request so that failed disable actions are no longer reported as enable operations.

### How should we improve?

Fix and please include a unit test/modify existing tests in TestCatalogOperations.java

egvr2002:
Hi @justinmclean, can I work on this issue?

justinmclean:
Sure, it all yours, if you have any questions just ask.

egvr2002:
Thanks for the assignment ( : 

I noticed two different coding styles for determining the operation type:

- In `server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java` the operation type is first assigned to a variable and then reused:

    ```java
    OperationType op = request.isEnable() ? OperationType.ENABLE : OperationType.DISABLE;
    ```

- In `server/src/main/java/org/apache/gravitino/server/web/rest/MetalakeOperations.java` the operation type is first assigned to a variable and then reused::

    ```java
     if (request.isInUse()) {
        metalakeDispatcher.enableMetalake(identifier);
     } else {
        metalakeDispatcher.disableMetalake(identifier);
     }
    ```

     Here's an example where it's called again in the same file:
     
    ```java
     return ExceptionHandlers.handleMetalakeException(
          request.isInUse() ? OperationType.ENABLE : OperationType.DISABLE, metalakeName, e);
    ```

Would you prefer one style over the other? I think standardizing on a single approach would help maintain consistency across the codebase. Please let me know your preference.

justinmclean:
If it is used multiple times, it makes sense to assign it to a variable; if it is used once, there is really no need for a variable.

egvr2002:
Hi again, I've already opened a PR hoping this closes the issue. Thanks a lot for your assistance!

PR #9092, opened 2025-11-12
[#9081] fix(catalog): use correct operation type in catalog update error handler in CatalogOperations.java

egvr2002:

This PR fixes the catalog update error handler to use the correct operation type from the incoming request instead of hardcoding or inferring the operation type. Previously, failed disable operations were incorrectly reported as enable operations in error messages.

Changes:
- Modified the error handler in catalog operations to preserve the original operation type from the request
- Updated tests in `TestCatalogOperations.java` to verify correct error reporting for both enable and disable operations

When a catalog disable operation fails, the error message incorrectly reports it as an enable operation failure. This misleads users and makes debugging difficult.

Fix: #9081

No.

1. Added test cases to ensure both enable and disable operation failures report correctly.


justinmclean (approved):
Thanks for this.

Issue #8939, opened 2025-10-28
[Subtask] Support transaction when create/alter lance table


yuqi1129:
### Describe the subtask

Currently, as both the Gravitino store and the Lance will keep the metadata of the Lance table, we need to take measures to ensure the entire operation is treated as a transaction.  

### Parent issue

#8828

PR #9103, opened 2025-11-13
[#8939] Improvement(lance-catalog): Make operating Gravitino store and Lance dataset atomic.

yuqi1129:

If we failed to create the Lance dataset, we need to remove the data store in Gravitino store.

Once we failed to create a lance table, without this PR, we can't try again, as the metadata is already exists in the Graivitno server.

Fix: #8939 

N/A

ITs.

### DIFF
--- a/catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
+++ b/catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
@@ -255,9 +255,8 @@ public class GenericLakehouseCatalogOperations
                 i -> ColumnEntity.toColumnEntity(columns[i], i, idGenerator.nextId(), auditInfo))
             .collect(Collectors.toList());
 
-    TableEntity entityToStore;
     try {
-      entityToStore =
+      TableEntity entityToStore =
           TableEntity.builder()
               .withName(ident.name())
               .withNamespace(ident.namespace())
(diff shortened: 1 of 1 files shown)


catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
```diff
@@ -307,7 +307,16 @@ public Table createTable(
           ident, columns, comment, newProperties, partitions, distribution, sortOrders, indexes);
     } catch (EntityAlreadyExistsException e) {
       throw new TableAlreadyExistsException(e, "Table %s already exists", ident);
-    } catch (IOException e) {
+    } catch (Exception e) {
```
mchades, line 313: should we move `store.put(entityToStore);` out of the try block?

yuqi1129: I got what you want expressed.   If we failed to execute `store.put(entityToStore);`, and then when we try to execute 

```
store.delete() 
```

It will still ok as `delete` will not throw an exception if the entity does not exist in the entity store.

Moving it out of the try block will need two try blocks


```java
try {
   store.put()
} catch (EntityAlreadyExistsException) {
} catch (exception e) {
   //
}
 

try {
   // handle lance operations
} catch (Excetioon e) {
	store.remove() 
}


```

mchades: yes, the latter will make the code logic clearer and avoid unnecessary calls.

yuqi1129: I do not have too much preference for these two. If you guys prefer the latter, I can change it.

mchades (approved):
lgtm.

Issue #9110, opened 2025-11-13
[Bug report] The fileStatusCache cannot be cleared when executing a refresh table SQL statement using spark-connector


yangyuxia:
### Version

main branch

### Describe what's wrong

After a Spark session started using spark-connecto plugin, and then a table is modified via a Hive client, Spark's `refresh table` statement fails to clear the table's fileStatusCache,making it impossible to query the latest data from the table.

### Error message and/or stacktrace

no error message

### How to reproduce

branch-0.8

PR #9111, opened 2025-11-13
[#9110] fix(spark): clean cache when execute refresh table sql

yangyuxia:

Overriding the invalidateTable() method in BaseCatalog

After a Spark session started using spark-connecto plugin, and then a table is modified via a Hive client, Spark's refresh table statement fails to clear the table's fileStatusCache,making it impossible to query the latest data from the table.

Fix: #9110 

N/A.

ITs


roryqi:
Do the other catalogs have similar issues? Could you add some UT for this fix?

yangyuxia:
So far, I've only used Hive catalog， and this pr depends on the fileStatusCache lifecycle management of kyuubi-hive-connector. https://github.com/apache/kyuubi/pull/7191

FANNG1:
@yangyuxia ,  would you like to move the fix to `BaseCatalog`? since it's a general problem for all catalogs.

FANNG1:
merged to main, @yangyuxia thanks for the contribution!

Issue #9108, opened 2025-11-13
[Bug report] Mapping between tenant principal and username when using Spark Connector accesses Gravitino server


yangyuxia:
### Version

main branch

### Describe what's wrong

Tenant principals are typically in the format "userName/groupName@realm". When accessing a Kerberos-authenticated Gravitino server using the spark-connector, the tenant's username is mapped to "userName/groupName". However, there is no "userName/groupName" user in Hadoop. Hadoop's auth_to_local rule generally maps to "userName".

### Error message and/or stacktrace

<img width="1572" height="752" alt="Image" src="https://github.com/user-attachments/assets/6e0e2590-fc0b-4f6a-84ce-282623ee68d2" />

### How to reproduce

main

roryqi:
Now, we have supported regex pattern for principal.  We can close this issue.

PR #9112, opened 2025-11-13
Update the mapping between principal and username

yangyuxia:
[#9108 ]fix(server-common) Update the mapping between principal and username

Tenant principals are typically in the format "userName/groupName@realm". When accessing a Kerberos-authenticated Gravitino server using the spark-connector, the tenant's username is mapped to "userName/groupName". However, there is no "userName/groupName" user in Hadoop. Hadoop's auth_to_local rule generally maps to "userName".

permission verification

Fix: #9108

N/A.

ITs

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosAuthenticator.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosAuthenticator.java
@@ -177,7 +177,7 @@ public class KerberosAuthenticator implements Authenticator {
         throw new UnauthorizedException("Principal has wrong format", AuthConstants.NEGOTIATE);
       }
 
-      String user = principalComponents.get(0);
+      String user = Splitter.on('/').splitToList(principalComponents.get(0)).get(0);
       // TODO: We will have KerberosUserPrincipal in the future.
       //  We can put more information of Kerberos to the KerberosUserPrincipal
       // For example, we can put the token into the KerberosUserPrincipal,


roryqi:
Usually principal names are in the form 'user/instance@REALM' or 'user@REALM'.

yangyuxia:
This modification applies to both 'user/instance@REALM' and 'user@REALM'.

roryqi:
> This modification applies to both 'user/instance@REALM' and 'user@REALM'.

I can't get your point. Why do we need the modification? If you don't want the use the 'user/instance' as the principal, you can use 'user@REALM' directly.

yangyuxia:
Why don't you change it to support both ‘user1@REALM’and ‘user1/group@REALM’ mapping  to ‘user1’, instead of forcing to use 'user1@REALM' ?

roryqi:
> Why don't you change it to support both ‘user1@REALM’and ‘user1/group@REALM’ mapping to ‘user1’, instead of forcing to use 'user1@REALM' ?

For Kerberos `user1/instance1` and `user1/instance2` is two principals. It represents two users in the Kerberos system.
For another point, Hive backbend will get the delegation token, it will get `user1/instance1` and `user/instance2` delegation token instead of `user1` and `user2` delegation token.

Could u provide the documents about using `instance1` as the group name?

yangyuxia:
In our KDC, tenant principals are all in the format test/test@realm. According to Hadoop's auth_to_local ' RULE:[2:$1@$0](.*@.*)s/@.*//,RULE:[1:$1@$0](.*@.*)s/@.*//,DEFAULT   ', this will be mapped to "test," and the user in ranger/ldap is also "test." When we use this type of principal with Gravitino-spark-connector, we encounter an error as shown in the image above.

roryqi:
> In our KDC, tenant principals are all in the format test/test@realm. According to Hadoop's auth_to_local ' RULE:[2:$1@$0](.*@.*)s/@._//,RULE:[1:$1@$0](.*@.*)s/@._//,DEFAULT ', this will be mapped to "test," and the user in ranger/ldap is also "test." When we use this type of principal with Gravitino-spark-connector, we encounter an error as shown in the image above.

Got it . 
Currently, we need to got the client principal for the HiveClient, you can see the code
```
 @Override
  public Object doAs(
      Principal principal, Executable<Object, Exception> action, Map<String, String> properties)
      throws Throwable {
    try {
      UserGroupInformation proxyUser;

      if (UserGroupInformation.isSecurityEnabled() && ops != null) {

        // The Gravitino server may use multiple KDC servers.
        // The http authentication use one KDC server, the Hive catalog may use another KDC server.
        // The KerberosAuthenticator will remove realm of principal.
        // And then we add the realm of Hive catalog to the user.
        final String finalPrincipalName;
        String proxyKerberosPrincipalName = principal.getName();
        if (!proxyKerberosPrincipalName.contains("@")) {
          finalPrincipalName =
              String.format("%s@%s", proxyKerberosPrincipalName, ops.getKerberosRealm());
        } else {
          finalPrincipalName = proxyKerberosPrincipalName;
        }

        proxyUser = UserGroupInformation.createProxyUser(finalPrincipalName, realUser);

        String token =
            ops.getClientPool()
                .run(
                    client -> {
                      return client.getDelegationToken(finalPrincipalName, realUser.getUserName());
                    });

        Token<DelegationTokenIdentifier> delegationToken = new Token<DelegationTokenIdentifier>();
        delegationToken.decodeFromUrlString(token);
        delegationToken.setService(
            new Text(ops.getHiveConf().getVar(HiveConf.ConfVars.METASTORE_TOKEN_SIGNATURE)));

        proxyUser.addToken(delegationToken);
      } else {

        proxyUser = UserGroupInformation.createProxyUser(principal.getName(), realUser);
      }
```
The principal should be consistent with the client.

Maybe you should add a KerberosPrincipal
```
class KerberosPrincipal {
    String username();
    Optional<String> instanceName();
}
```
So we can get the correct client principal in the HiveProxyPlugin. And you get the correct user name in our system, too.

Issue #9095, opened 2025-11-13
[Improvement] In TableCreateRequest.java prevent potental null pointer exception


justinmclean:
### What would you like to be improved?

In validate() index.fieldNames() could be null.

### How should we improve?

Change precondition to check for null and check length. Please also add some/modify existing unit tests.

PCloud63514:
Hi! @justinmclean 
This issue looks similar to the one I’m currently working on.  
May I take this issue?

PR #9118, opened 2025-11-14
[#9095] fix(dto): Ensure index field names are not null

PCloud63514:

- I fixed a piece of code in TableCreateRequest.validate that could potentially cause a NPE.

- Added null check for updates list in request validation.

Fix: #9095 

- no.
- This change only corrects internal exception handling logic, with no modification to public APIs or user-visible behavior.

- No new tests added, since this only fixes the caught exception type.
- Ensured all existing tests continue to pass.

### DIFF
--- a/common/src/main/java/org/apache/gravitino/dto/requests/TableCreateRequest.java
+++ b/common/src/main/java/org/apache/gravitino/dto/requests/TableCreateRequest.java
@@ -161,9 +161,10 @@ public class TableCreateRequest implements RESTRequest {
               index -> {
                 Preconditions.checkArgument(index.type() != null, "Index type cannot be null");
                 Preconditions.checkArgument(
-                    index.fieldNames() != null, "Index field names cannot be null");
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/apache/gravitino/dto/requests/TableCreateRequest.java
```diff
@@ -160,6 +160,8 @@ public void validate() throws IllegalArgumentException {
           .forEach(
               index -> {
                 Preconditions.checkArgument(index.type() != null, "Index type cannot be null");
+                Preconditions.checkArgument(
```
yuqi1129, line 163: Please change the message in L166 accordingly; it should be `The length of fieldNames should be greater than 0` or something similar.

PCloud63514: I’ve updated the message. Thanks!

Issue #8076, opened 2025-08-14
[Improvement] Add connection pool metrics for jdbc catalog


hqbhoho:
### What would you like to be improved?

Jdbc catalog will use datasource to connect backend instance. Monitoring connection pool metrics is crucial for understanding database connectivity health and performance.

PR #9119, opened 2025-11-14
[#8076] Add connection pool metrics for jdbc catalog

hqbhoho:
Support metrics for jdbc catalog connection pool

Fix: #8076

No

local test

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -55,9 +55,19 @@ These metrics start with the `jvm` prefix, like `jvm.heap.used` in JSON format, 
 Catalog metrics provide the metrics from different catalog instances.
 All the catalog metrics start with the `gravitino-catalog` prefix in Prometheus format and with labels `provider`, `metalake`, and `catalog` to distinguish different catalog instances.
 
-For example, you can get Prometheus metrics for a fileset catalog named `test_catalog` under a metalake named `test_metalake` in the Gravitino server as follows:
+Currently, Catalog metrics only support Fileset catalog and JDBC catalog. 
+
+You can get Prometheus metrics for a Fileset catalog named `test_catalog` under a metalake named `test_metalake` in the Gravitino server as follows:
 
 ```text
 gravitino-catalog_filesystem-cache_hits{provider="fileset",metalake="test_metalake",catalog="test_catalog",} 0.0
 gravitino-catalog_filesystem-cache_misses{provider="fileset",metalake="test_metalake",catalog="test_catalog",} 0.0
+```
+
+You can get Prometheus metrics for a JDBC catalog named `test_catalog` under a metalake named `test_metalake` in the Gravitino server as follows:
+
+```text
+gravitino_catalog_datasource_idle_connections{provider="jdbc",metalake="test_metalake",catalog="test_catalog",} 1.0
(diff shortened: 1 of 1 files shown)


catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/JdbcCatalogOperations.java
```diff
@@ -174,11 +179,25 @@ public void initialize(
     if (tableOperation instanceof RequireDatabaseOperation) {
       ((RequireDatabaseOperation) tableOperation).setDatabaseOperation(databaseOperation);
     }
+
+    MetricsSystem metricsSystem = GravitinoEnv.getInstance().metricsSystem();
+    // Metrics System could be null in UT.
+    if (metricsSystem != null) {
+      this.catalogMetricsSource =
```
roryqi, line 186: Our catalog and metalake can be renamed.

hqbhoho: Thanks for your feedback. 
Based on that,  add `jdbc_url`as a label, rather than `metalake` and `catalog`, What do you think?   
So `fileset` metrics have the same problem? cc @mchades

roryqi: I'm not sure whether `jdbc_url` can be updated. cc @yuqi1129

hqbhoho: I think it will new `JdbcCatalogOptions` if `jdbc_url` update.

roryqi: I got it. It will register metrics source. It makes sense.

mchades: @FANNG1 cc

hqbhoho: Code review reveals that renaming a catalog or metalake triggers the creation of a new `JdbcCatalogOptions` instance. When `BaseCatalog#ops()` is subsequently called, it invokes `JdbcCatalogOptions#initialize`. Consequently, metrics are recorded after metadata operation performed on the renamed catalog or metalake.

image

So there is no issue with renaming the catalog and metalake.
cc @jerqi @mchades @FANNG1 @yuqi1129

hqbhoho:
@FANNG1 PTAL, thanks!

jerqi (approved):

FANNG1:
Could you add related document in https://github.com/apache/gravitino/blob/main/docs/metrics.md ?

hqbhoho:
@FANNG1 PTAL

FANNG1 (approved):

PR #9128, opened 2025-11-14
[#9124] Handle negative partNo safely and fix test file typo

### LINKED ISSUE #9124: [Improvement] Fix hasNamePart in FullName.java to check for negative indexes

A negative partNo can be passed in; we should check that position >= 0 in the return.

### How should we improve?

Fix and add a unit test.

HarshMehta112:
Fixed #9124

- Negative partNo values can now be safely handled.
- Added unit test cases to verify this behavior.
- Renamed the test file from TestFulllName to TestFullName to fix a minor typo.

### DIFF
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/FullName.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/FullName.java
@@ -215,7 +215,7 @@ public class FullName {
       int length = names.length;
       int position = partNo;
 
-      return position >= 0 && position <= length;
+      return position > 0 && position <= length;
     }
 
     return false;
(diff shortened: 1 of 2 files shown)


clients/cli/src/main/java/org/apache/gravitino/cli/FullName.java
```diff
@@ -215,7 +215,7 @@ public boolean hasNamePart(int partNo) {
       int length = names.length;
       int position = partNo;
 
-      return position <= length;
+      return position >= 0 && position <= length;
```
justinmclean: While I mentioned a negative value in the issue, we could change this to position> 0, as 0 is not a valid option either.

HarshMehta112: I checked that and pushed the required changes.

justinmclean (approved):
Thanks for this

Issue #9080, opened 2025-11-12
[Improvement] Fix the Flink TypeUtils map conversion to use value-type nullability instead of the map-level nullability


justinmclean:
### What would you like to be improved?

In flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/utils/TypeUtils.java fix the Flink TypeUtils map conversion to use the value-type nullability instead of the map-level nullability.

Here is an updated test (in TestTypeUtils.java):
```
    Assertions.assertEquals(
        Types.MapType.of(Types.StringType.get(), Types.IntegerType.get(), true),
        TypeUtils.toGravitinoType(
            new MapType(true, new VarCharType(Integer.MAX_VALUE), new IntType(true))));
    Assertions.assertEquals(
        Types.MapType.of(Types.StringType.get(), Types.IntegerType.get(), false),
        TypeUtils.toGravitinoType(
            new MapType(true, new VarCharType(Integer.MAX_VALUE), new IntType(false))));
    Assertions.assertEquals(
        Types.MapType.of(Types.StringType.get(), Types.IntegerType.get(), false),
        TypeUtils.toGravitinoType(
            new MapType(false, new VarCharType(Integer.MAX_VALUE), new IntType(false))));
```

Jackeyzhe:
@justinmclean Hi, I'd like to take this, can you assgin it to me?

justinmclean:
It's yours, if you have any question just ask.

PR #9129, opened 2025-11-14
[#9080][Improvement]Fix the Flink TypeUtils map conversion to use value-type nullability

Jackeyzhe:

Fix the Flink TypeUtils map conversion to use value-type nullability instead of map-level nullability.

Fix: #9080 

No

update some unit tests

### DIFF
--- a/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/utils/TypeUtils.java
+++ b/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/utils/TypeUtils.java
@@ -102,8 +102,9 @@ public class TypeUtils {
         return Types.TimestampType.withTimeZone(zonedPrecision);
       case ARRAY:
         ArrayType arrayType = (ArrayType) logicalType;
-        Type elementType = toGravitinoType(arrayType.getElementType());
-        return Types.ListType.of(elementType, arrayType.isNullable());
+        LogicalType elementLogicalType = arrayType.getElementType();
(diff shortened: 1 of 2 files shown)


justinmclean:
This looks good to me @coolderli what do you think?

Jackeyzhe:
@coolderli Hi, could you help with reviewing this PR if you have time?

FANNG1:
@Jackeyzhe , thanks for the PR, there are similar issues for the LIST transform, and the nullibility transform logic for LIST and MAP  in  `toFlinkType` seems not correct too,  do you want to fix it in this PR?

Jackeyzhe:
> @Jackeyzhe , thanks for the PR, there are similar issues for the LIST transform, and the nullibility transform logic for LIST and MAP in `toFlinkType` seems not correct too, do you want to fix it in this PR?

OK, I will fix it in this PR later

Jackeyzhe:
@FANNG1 Hi, I fix the LIST transform logic in `toGravitinoType`, and I think the transform logic in `toFlinkType` is correct, it's already processed as value-type nullibility. Could you help reviewing this PR if you have time?

FANNG1:
> @FANNG1 Hi, I fix the LIST transform logic in `toGravitinoType`, and I think the transform logic in `toFlinkType` is correct, it's already processed as value-type nullibility. Could you help reviewing this PR if you have time?

Yes, you are right,  merged to main, thanks for the work!

Issue #9122, opened 2025-11-14
[Improvement] Prevent potential null pointer error in MetadataObjectStatisticsOperations.java


justinmclean:
### What would you like to be improved?

Added validate to MetadataObjectStatisticsOperations.java dropStatistics method to throw a clear error when the caller supplies a null or empty statistics list, preventing a NullPointerExceptions.

Adding a few lines to the existing test shows the issue:
```
 private void testDropStatistics(SupportsStatistics supportsStatistics, String path)
      throws JsonProcessingException {
    Assertions.assertThrows(
        IllegalArgumentException.class, () -> supportsStatistics.dropStatistics(null));
    Assertions.assertThrows(
        IllegalArgumentException.class, () -> supportsStatistics.dropStatistics(Collections.emptyList()));
```

### How should we improve?

see above

Pikady:
Hi! @justinmclean 
Can I take this issue?

justinmclean:
Sure it's all yours. If you have any questions, just ask.

PR #9134, opened 2025-11-16
[#9122] [Improvement]: Prevent potential null pointer error in MetadataObjectStatisticsOperations.java

Pikady:

- Add argument validation in MetadataObjectStatisticsOperations#dropStatistics to throw IllegalArgumentException when the list is null or empty.
- Extend unit tests to assert null/empty list behavior.

- Prevents potential NullPointerException and provides clearer error feedback to callers.

Fix: #9122

- dropStatistics(null | empty) now throws IllegalArgumentException (null input previously could cause NPE).

- Unit tests added for null and empty list inputs.


justinmclean (approved):
Thanks for this

Issue #9139, opened 2025-11-17
[Improvement] Fix potential null pointer error in TreeLock.java


justinmclean:
### What would you like to be improved?

In TreeLock.java in unlock there is a NPE when you have a null timestamp

Here is a unit test to help:
```
  @Test
  void testUnlockWithMissingHoldingTimestamp() {
    TreeLockNode rootNode = new TreeLockNode("root");
    TreeLockNode childNode = rootNode.getOrCreateChild("child").getLeft();
    TreeLock treeLock =
        new TreeLock(Arrays.asList(rootNode, childNode), NameIdentifier.of("root", "child"));

    treeLock.lock(LockType.WRITE);
    childNode.getHoldingThreadTimestamp().clear();

    assertDoesNotThrow(treeLock::unlock);
  }
```

### How should we improve?

deal with null timestamps

PCloud63514:
Hi @justinmclean,
I'd like to take this issue. 
I plan to handle the missing timestamp case by treating the returned value as a nullable Long and logging safely.

justinmclean:
That sounds like a good approach.

PR #9146, opened 2025-11-17
[#9139] fix(lock): Prevent NPE when timestamp is missing in TreeLock.unlock

PCloud63514:
- Updated TreeLockNode.removeHoldingThreadTimestamp to return a nullable Long instead of long.
- Modified TreeLock.unlock() to handle null timestamps safely before logging.
- Ensures unlocking works correctly even when the timestamp map is cleared.

- holdingThreadTimestamp is a Map<ThreadIdentifier, Long>, and Map.remove() returns null when no timestamp exists.
- The previous implementation returned long, causing a null → primitive unboxing and leading to a NullPointerException.
- Unlocking a node should not depend on the presence of a timestamp value, so the logic must tolerate null entries.

Fix: #9139 

- no.
- This change only improves internal lock handling and exception safety.

- Verified using the unit test included in the issue (testUnlockWithMissingHoldingTimestamp).

### DIFF
--- a/core/src/main/java/org/apache/gravitino/lock/TreeLock.java
+++ b/core/src/main/java/org/apache/gravitino/lock/TreeLock.java
@@ -153,15 +153,18 @@ public class TreeLock {
       LockType type = pair.getRight();
       current.unlock(type);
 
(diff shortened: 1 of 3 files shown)


justinmclean (approved):
Thanks for this

Issue #9074, opened 2025-11-11
[Subtask] Refactor the generic lakehouse catalog implementation to use managed table operations


jerryshao:
### Describe the subtask

Since we've already abstracted the managed table operations, so we can refactor the generic lakehouse catalog implementation to leverage this class.

### Parent issue

#8828

PR #9097, opened 2025-11-13
[#9021] feat(lance-rest): Support create & list index for LRS

yuqi1129:
### What changes were proposed in this pull request?

Add create and list index APIs in the Lance rest server.

### Why are the changes needed?

It's a feature that benefits users.

Fix: #9021 

### Does this PR introduce _any_ user-facing change?

N/A.

### How was this patch tested?

UTs and ITs

PR #9160, opened 2025-11-17
[#9074] improve(catalog-lakehouse-generic): Refactor the code structure of the generic lakehouse catalog

jerryshao:

Refactor the code of lakehouse-generic catalog.

Fix: #9074

No.

UTs.

### DIFF
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/LakehouseTableDelegatorFactory.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/LakehouseTableDelegatorFactory.java
@@ -20,7 +20,6 @@ package org.apache.gravitino.catalog.lakehouse.generic;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
-import java.util.Locale;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.stream.Collectors;
@@ -62,15 +61,5 @@ public class LakehouseTableDelegatorFactory {
     // Initialize delegators if not yet done
     createTableDelegators();
     return delegators;
-  }
-
-  public static LakehouseTableDelegator tableDelegator(String format) {
-    // Initialize delegators if not yet done
-    createTableDelegators();
-
-    LakehouseTableDelegator delegator = delegators.get(format.toLowerCase(Locale.ROOT));
-    Preconditions.checkArgument(
-        delegator != null, "The lakehouse table format %s is not supported.", format);
-    return delegator;
   }
 }


jerryshao:
This needs to add more UTs.

api/src/main/java/org/apache/gravitino/Schema.java
```diff
@@ -37,6 +37,19 @@
 @Evolving
 public interface Schema extends Auditable {
 
+  /**
+   * The property name for the location of the Schema. This property indicates the physical storage
```
yuqi1129: Schema -> schema.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -0,0 +1,273 @@
...
+
+      if (external) {
+        // If the table is external, we only drop the table metadata in Gravitino
+        return super.dropTable(ident);
+      }
+
+      String location = table.properties().get(Table.PROPERTY_LOCATION);
+
+      // Delete the Lance dataset at the location
+      Dataset.drop(location, LancePropertiesUtils.getLanceStorageOptions(table.properties()));
```
yuqi1129: Dropping data first is rather risky if deleting metadata from Gravitino storage fails, and there is no mechanism currently to rollback operation `Dataset.drop`.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+  }
+
+  private IndexParams getIndexParamsByIndexType(IndexType indexType) {
+    switch (indexType) {
+      case SCALAR:
+        return IndexParams.builder().build();
+      case VECTOR:
+        // TODO make these parameters configurable
+        int numberOfDimensions = 3; // this value should be determined dynamically based on the data
```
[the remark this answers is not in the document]
jerryshao: @yuqi1129 please fix this in a separate PR.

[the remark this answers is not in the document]
yuqi1129: This will be fixed by https://github.com/apache/gravitino/pull/9097.

jerryshao:
@yuqi1129 @mchades please help me to review this PR, thanks.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -0,0 +1,282 @@
...
+      }
+
+      // If the table metadata is dropped successfully, and the table is not external, we can delete
+      // the
+      // Lance dataset. Otherwise, we should not delete the dataset.
+      if (dropped) {
+        String location = table.properties().get(Table.PROPERTY_LOCATION);
+
+        // Delete the Lance dataset at the location
+        Dataset.drop(location, LancePropertiesUtils.getLanceStorageOptions(table.properties()));
```
yuqi1129, line 219: Do you need to catch possible exceptions here and then roll back entity store operation？

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/LakehouseTableDelegatorFactory.java
```diff
@@ -0,0 +1,76 @@
...
+    delegators =
+        ImmutableMap.copyOf(
+            loader.stream()
+                .collect(
+                    Collectors.toMap(
+                        provider -> provider.get().tableFormat(), ServiceLoader.Provider::get)));
+    Preconditions.checkArgument(
+        !delegators.isEmpty(),
+        "No LakehouseTableDelegator implementation found via ServiceLoader, this is unexpected, "
+            + "please check the code to see what is going on.");
```
yuqi1129, line 55: Please format the comments

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/LakehouseTableDelegatorFactory.java
```diff
@@ -0,0 +1,76 @@
...
+    LOG.info("Loaded LakehouseTableDelegators: {}", delegators.keySet());
+  }
+
+  public static ImmutableMap<String, LakehouseTableDelegator> tableDelegators() {
+    // Initialize delegators if not yet done
+    createTableDelegators();
+    return delegators;
+  }
+
+  public static LakehouseTableDelegator tableDelegator(String format) {
```
yuqi1129: This method seems to be unused.

yuqi1129 (commented):
These are all comments from me, others LGTM

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/LakehouseTableDelegatorFactory.java
jerryshao: Can you describe more, what should I format?

yuqi1129: Sorry, It should be L213 in file `org.apache.gravitino.catalog.lakehouse.lance.LanceTableOperations`. The browser is very stuck, and it causes such misleading information.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
jerryshao: I don't want to do it. This is not atomic and weird. If we assume everything goes from Gravitino, then the failure will only cause orphan data; the user will never be able to visit the table again.

yuqi1129: I see. It's not a big problem currently, and we can optimize it later

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalog.java
mchades, line 1: I was thinking that using `LakehouseCatalog` and `lakehouse` as a short name might be more intuitive and concise. The `lakehouse` name directly includes formats like `iceberg`, `hudi`, `paimon`, and `lance`. 

As user needs grow, we may eventually need a true `GenericCatalog` to support managing more types of metadata, such as `hive`, `mysql`, or even `fileset`, `model`, etc.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/LakehouseTableDelegatorFactory.java
yuqi1129: Another point, there may be duplicates if I change to the table format, as future docs about `lance-rest-service` will also have the same content.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalog.java
jerryshao: The meaning of `Lakehouse` is already reflected in the package name. Besides, it is not user-facing; we can figure out a better name later on. I'm concerned about the catalog provider name.

@yuqi1129 what is your suggestion?

yuqi1129: > I'm concerned about the catalog provider name. 

I agree on this point, the name of the provider seems to be more important compared to the class name, and we'd better make these two consistent.

If the provider name is `lakehouse_catalog`, then I would like to adopt your suggestion and use `LakehouseCatalog` as the name.  There is no need to struggling with the name of the class name as it can be changed later without any compatiable issues.

mchades: You may not have gotten the point. The current issue is whether the provider's value should be `lakehouse` or `lakehouse-generic`.

jerryshao: > You may not have gotten the point. The current issue is whether the provider's value should be `lakehouse` or `lakehouse-generic`.

Let me think more about it.

mchades (approved):
After offline discussion, we decided to continue using `lakehouse-generic`.

Issue #9148, opened 2025-11-17
[Improvement] Clear data in the reverse index when data is drop or altered


yuqi1129:
### What would you like to be improved?

Currently, Caffeine entity cace contains three parts: cache data, cache index, and the reverse index.  Let's take a simple example. When we cache a role entity in the cache
```
cacheData:  role_ident -> roleEntity
reverseIndex: securable_object1_in_role -> [role_ident]; securable_object2_in_role -> [role_ident]
```

The problem is that when a role is deleted or dropped, we can't currently clear the data in the reverse index.

### How should we improve?

Keep the mapping from cache data to cache index and clear reverse index if needed.

PR #9162, opened 2025-11-18
[#9148] improvement(core): Clear useless data in the reversed index for entity cache.

yuqi1129:

Clear the entry in the reverse index that has been removed from cache data.

We need to clear all out-of-date data in the cache.

Fix: #9148 

N/A.

UTs

### DIFF
--- a/core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
@@ -52,8 +52,8 @@ public class ReverseIndexCache {
    * indexing.
    *
    * <p>For example, a role entity may be related to multiple securable objects, so we need to
-   * maintain a mapping from the role entity key to the list of securable object keys. that is
-   * dataToReverseIndexMap: roleEntityKey -> [securableObjectKey1, securableObjectKey2, ...]
+   * maintain a mapping from the role entity key to the list of securable object keys. That is
+   * entityToReverseIndexMap: roleEntityKey -> [securableObjectKey1, securableObjectKey2, ...]
    *
    * <p>This map is used to quickly find all the related entity cache keys when we need to
    * invalidate in the reverse index if a role entity is updated. The following is an example: a
@@ -64,11 +64,11 @@ public class ReverseIndexCache {
    * <p>This map will be: <br>
    * role1 -> [s1, s2] </br>
    *
-   * <p>When we update role1, we need to invalidate s1 and s2 from the reverse index, or the data
-   * will be in the memory forever. However, the main branch before this PR does not support this
-   * operation directly as we do not maintain such a map.
+   * <p>When we update role1, we need to invalidate s1 and s2 from the reverse index via this map,
+   * or the data will be in the memory forever.
    */
-  private Map<EntityCacheKey, List<EntityCacheKey>> entityToReverseIndexMap = Maps.newHashMap();
+  private final Map<EntityCacheKey, List<EntityCacheKey>> entityToReverseIndexMap =
+      Maps.newHashMap();
 
   public ReverseIndexCache() {
     this.reverseIndex = new ConcurrentRadixTree<>(new DefaultCharArrayNodeFactory());
@@ -91,7 +91,7 @@ public class ReverseIndexCache {
     if (CollectionUtils.isNotEmpty(relatedKeys)) {
       for (EntityCacheKey relatedKey : relatedKeys) {
         List<EntityCacheKey> existingKeys = reverseIndex.getValueForExactKey(relatedKey.toString());
-        if (existingKeys != null && existingKeys.contains((key))) {
+        if (existingKeys != null && existingKeys.contains(key)) {
           List<EntityCacheKey> newValues = Lists.newArrayList(existingKeys);
           newValues.remove(key);
           if (newValues.isEmpty()) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -42,6 +44,25 @@ public class ReverseIndexCache {
...
+   * <p>This map is used to quickly find all the related entity cache keys when we need to
+   * invalidate in the reverse index if a role entity is updated. The following is an example: a
+   * Role a has securable objects s1 and s2, so we have the following mapping: <br>
+   * cacheData: role1 -> role entity reserveIndex: s1 -> [role1], s2 -> [role1] </br>
+   *
+   * <p>When we update role1, we need to invalidate s1 and s2 from the reverse index, or the data
+   * will be in the memory forever. However, the current implementation of ReverseIndexCache does
+   * not support this operation directly as we do not maintain such a map.
+   */
+  private Map<EntityCacheKey, List<EntityCacheKey>> dataToReverseIndexMap = Maps.newHashMap();
```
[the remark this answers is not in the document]
yuqi1129: There is no need to use the concurrent map as this method is always called within a lock.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -45,6 +47,25 @@ public class ReverseIndexCache {
...
```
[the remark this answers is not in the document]
yuqi1129: The map will be guarded by `segmentedLock` in `CaffeineEntityCache`, so there is no need to use `ConcurrentHashMap`.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -45,6 +47,25 @@ public class ReverseIndexCache {
...
+   * dataToReverseIndexMap: roleEntityKey -> [securableObjectKey1, securableObjectKey2, ...]
+   *
+   * <p>This map is used to quickly find all the related entity cache keys when we need to
+   * invalidate in the reverse index if a role entity is updated. The following is an example: a
+   * Role a has securable objects s1 and s2, so we have the following mapping: <br>
+   * cacheData: role1 -> role entity reverseIndex: s1 -> [role1], s2 -> [role1] </br>
+   *
+   * <p>When we update role1, we need to invalidate s1 and s2 from the reverse index, or the data
+   * will be in the memory forever. However, the current implementation of ReverseIndexCache does
+   * not support this operation directly as we do not maintain such a map.
```
jerryshao: Do we support it or not **for now**? If it is supported for now in this PR, you need to remove this comment. This is confusing for the reviewers; the comment should be useful for others.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -45,6 +47,25 @@ public class ReverseIndexCache {
...
+   */
+  private Map<EntityCacheKey, List<EntityCacheKey>> dataToReverseIndexMap = Maps.newHashMap();
```
jerryshao: Is it better to use `entityToReverseIndexMap`?

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -57,20 +78,28 @@ public ReverseIndexCache() {
...
-  }
-
   public Iterable<List<EntityCacheKey>> getValuesForKeysStartingWith(String keyPrefix) {
     return reverseIndex.getValuesForKeysStartingWith(keyPrefix);
   }
 
-  public Iterable<CharSequence> getKeysStartingWith(String keyPrefix) {
-    return reverseIndex.getKeysStartingWith(keyPrefix);
-  }
+  public boolean remove(EntityCacheKey key) {
```
jerryshao, line 89: Is there a threading issue here?

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
yuqi1129: Okay, it should be **the current main branch**. Let me optimize it.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
yuqi1129: All adding and removing operations will be guarded by `segmentedLock` in `CaffeineEntityCache`, so there is no need to add extra code here.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
yuqi1129: It makes sense.

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -45,6 +47,29 @@ public class ReverseIndexCache {
...
+   * Role a has securable objects s1 and s2, so we have the following mapping: <br>
+   * cacheData: role1 -> role entity </br> <br>
+   * reverseIndex: s1 -> [role1], s2 -> [role1] </br>
+   *
+   * <p>This map will be: <br>
+   * role1 -> [s1, s2] </br>
+   *
+   * <p>When we update role1, we need to invalidate s1 and s2 from the reverse index, or the data
+   * will be in the memory forever. However, the main branch before this PR does not support this
+   * operation directly as we do not maintain such a map.
```
jerryshao: We don't have to explain the previous existing problems; what we need to explain is why do we need this.

Issue #9137, opened 2025-11-16
[Improvement] In PartitionRange.java encure that comparator default to the default comparator


justinmclean:
### What would you like to be improved?

Here's a unit test showing the issue:
```
  @Test
  public void testAllPartitionsComparator() {
    SortOrder defaultSortOrder =
        SortOrders.of(NamedReference.MetadataField.PARTITION_NAME_FIELD, SortDirection.ASCENDING);

    PartitionRange allPartitions = PartitionRange.ALL_PARTITIONS;

    Assertions.assertFalse(allPartitions.lowerPartitionName().isPresent());
    Assertions.assertFalse(allPartitions.upperPartitionName().isPresent());
    Assertions.assertEquals(defaultSortOrder, allPartitions.comparator());
  }
```

### How should we improve?

Set comparator in the constructor.

PCloud63514:
Hi, @justinmclean 
Can I take this issue?

It seems we can fix it by setting the comparator to DEFAULT_COMPARATOR in the default constructor.

PR #9175, opened 2025-11-19
[#9137] fix(stats): initialize comparator with DEFAULT_COMPARATOR in default constructor

PCloud63514:

- Initialize the comparator with DEFAULT_COMPARATOR in the default constructor.

- The default constructor used by ALL_PARTITIONS does not initialize the comparator, which results in it being null.

Fix: #9137

- No.

- Verified using the unit test included in the issue (testAllPartitionsComparator()).


justinmclean (approved):
Thanks for that

Issue #9071, opened 2025-11-11
[Subtask] Add document and api for job access control


hdygxsj:
### Describe the subtask

[Subtask] Add document and api for job access control

### Parent issue

#8944

roryqi:
Snowflake and Databricks have similar features to create a job. But they only have the privileges to limit the resource pool instead of a job. But we prefer add several privileges `CREATE_JOB_TEMPLATE` , `USE_JOB_TEMPLATE`, `CREATE_JOB`.
https://docs.databricks.com/aws/en/getting-started/data-pipeline-get-started
https://docs.snowflake.com/en/developer-guide/native-apps/tutorials/getting-started-tutorial

PR #9177, opened 2025-11-19
[#9071] feat(authz): Add job and job template privileges interfaces and documents

roryqi:

Add job and job template privileges interfaces and documents

Fix: #9071 

Added the documents.

Added unit tests.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/MetadataObjects.java
+++ b/api/src/main/java/org/apache/gravitino/MetadataObjects.java
@@ -21,9 +21,11 @@ package org.apache.gravitino;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
@@ -38,7 +40,7 @@ public class MetadataObjects {
 
   private static final Joiner DOT_JOINER = Joiner.on('.');
 
-  private static final Set<MetadataObject.Type> validSingleLevelNameTypes =
+  private static final Set<MetadataObject.Type> VALID_SINGLE_LEVEL_NAME_TYPES =
       Sets.newHashSet(
           MetadataObject.Type.CATALOG,
           MetadataObject.Type.METALAKE,
@@ -48,18 +50,25 @@ public class MetadataObjects {
           MetadataObject.Type.JOB_TEMPLATE,
           MetadataObject.Type.POLICY);
 
-  private static final Set<MetadataObject.Type> validTwoLevelNameTypes =
+  private static final Set<MetadataObject.Type> VALID_TWO_LEVEL_NAME_TYPES =
       Sets.newHashSet(MetadataObject.Type.SCHEMA);
 
-  private static final Set<MetadataObject.Type> validThreeLevelNameTypes =
+  private static final Set<MetadataObject.Type> VALID_THREE_LEVEL_NAME_TYPES =
       Sets.newHashSet(
           MetadataObject.Type.FILESET,
           MetadataObject.Type.TABLE,
           MetadataObject.Type.TOPIC,
           MetadataObject.Type.MODEL);
 
-  private static final Set<MetadataObject.Type> validFourLevelNameTypes =
+  private static final Set<MetadataObject.Type> VALID_FOUR_LEVEL_NAME_TYPES =
       Sets.newHashSet(MetadataObject.Type.COLUMN);
+
+  private static final Map<Set<MetadataObject.Type>, Integer> TYPE_TO_EXPECT_LENGTH =
+      ImmutableMap.of(
+          VALID_SINGLE_LEVEL_NAME_TYPES, 1,
+          VALID_TWO_LEVEL_NAME_TYPES, 2,
+          VALID_THREE_LEVEL_NAME_TYPES, 3,
+          VALID_FOUR_LEVEL_NAME_TYPES, 4);
 
   private MetadataObjects() {}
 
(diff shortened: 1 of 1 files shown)


roryqi:
@jerryshao Could you take a look at this PR?

roryqi:
Snowflake and Databricks have similar features to create a job. But they only have the privileges to limit the resource pool instead of a job. But we prefer add several privileges `CREATE_JOB_TEMPLATE` , `USE_JOB_TEMPLATE`, `CREATE_JOB`.
https://docs.databricks.com/aws/en/getting-started/data-pipeline-get-started
https://docs.snowflake.com/en/developer-guide/native-apps/tutorials/getting-started-tutorial

api/src/main/java/org/apache/gravitino/authorization/Privilege.java
```diff
@@ -98,7 +98,13 @@ enum Name {
...
     CREATE_TAG(0L, 1L << 21),
     /** The privilege to apply a tag */
-    APPLY_TAG(0L, 1L << 22);
+    APPLY_TAG(0L, 1L << 22),
+    /** The privilege to create a job template */
+    CREATE_JOB_TEMPLATE(0L, 1L << 23),
+    /** The privilege to use a job template */
+    USE_JOB_TEMPLATE(0L, 1L << 24),
+    /** The privilege to create a job */
+    CREATE_JOB(0L, 1L << 25);
```
jerryshao: It is better to align with the job system like "RUN_JOB" instead of "CREATE_JOB"?

jerryshao: The user who has the "RUN_JOB" permission can check and cancel the job, right?

Do we need additional privileges like "get" or "cancel" for different purposes?

roryqi: 1. The job can be checked and canceled by his owner.
2.  Maybe we should have a job manager who has the privileges to manage jobs include getting and canceling for management. Now this role can be finished by metalake owner.

roryqi: Another point: If all of the jobs are under the metalake, it's a little  inconvenient for management. The jobs usually are grouped by business.

jerryshao: Let's keep as it is. The reorganization requires lots of work.

roryqi:
@jerryshao I have addressed the comments of the copilot. Do u have any other suggestion?

docs/security/access-control.md
```diff
@@ -279,6 +281,19 @@ DENY `WRITE_FILESET` won‘t deny the `READ_FILESET` operation if the user has t
 | CREATE_POLICY | Metalake                  | Create a policy                           |
 | APPLY_POLICY  | Metalake, Policy          | Associate policies with metadata objects. |
 
+### Job template privileges
+
+| Name                | Supports Securable Object | Operation                                |
+|---------------------|---------------------------|------------------------------------------|
+| CREATE_JOB_TEMPLATE | Metalake                  | Create a job template                    |
+| USE_JOB_TEMPLATE    | Metalake, JobTemplate     | Use a job template when creating the job |
```
jerryshao: "when running the job".

api/src/main/java/org/apache/gravitino/authorization/Privileges.java
```diff
@@ -222,10 +237,28 @@ public static Privilege deny(Privilege.Name name) {
...
         return ApplyTag.deny();
+
+        // Policy
+      case APPLY_POLICY:
+        return ApplyPolicy.deny();
+      case CREATE_POLICY:
+        return CreatePolicy.deny();
+
+        // Job template
+      case CREATE_JOB_TEMPLATE:
```
jerryshao: "register" job template, not "create" job template.

api/src/main/java/org/apache/gravitino/authorization/SecurableObjects.java
```diff
@@ -138,6 +138,40 @@ public static SecurableObject ofModel(
...
+   * privileges.
+   *
+   * @param jobTemplate The job template name
+   * @param privileges The privileges of the job template
+   * @return The created job template {@link SecurableObject}
+   */
+  public static SecurableObject ofJobTemplate(String jobTemplate, List<Privilege> privileges) {
+    return of(MetadataObject.Type.JOB_TEMPLATE, Lists.newArrayList(jobTemplate), privileges);
+  }
+
```
jerryshao, line 174: Shall we create a helper method for `JOB`?

api/src/main/java/org/apache/gravitino/MetadataObjects.java
```diff
@@ -36,6 +38,29 @@ public class MetadataObjects {
 
   private static final Joiner DOT_JOINER = Joiner.on('.');
 
+  private static final Set<MetadataObject.Type> validSingleNameTypes =
```
jerryshao: `validSingleLevelNameTypes`, `validTwoLevelNameTypes`, etc.

api/src/main/java/org/apache/gravitino/authorization/SecurableObjects.java
roryqi: The securable object `JOB` can't bind any privilege. So we don't add the helper method.

docs/security/access-control.md
roryqi: Fixed.

api/src/main/java/org/apache/gravitino/MetadataObjects.java
roryqi: Fixed.

api/src/main/java/org/apache/gravitino/MetadataObjects.java
```diff
@@ -69,30 +98,21 @@ public static MetadataObject of(List<String> names, MetadataObject.Type type) {
...
+          names.size() == 2, "If the type is " + type + ", the length of names must be 2");
+    } else if (validThreeLevelNameTypes.contains(type)) {
+      Preconditions.checkArgument(
+          names.size() == 3, "If the type is " + type + ", the length of names must be 3");
+    } else if (validFourLevelNameTypes.contains(type)) {
+      Preconditions.checkArgument(
+          names.size() == 4, "If the type is COLUMN, the length of names must be 4");
+    } else {
+      throw new IllegalArgumentException("Unsupported metadata object type: " + type);
+    }
```
jerryshao: This part can be optimized using lambda function to handle the check. For example, like maintaining a map for validation check, based on the different `Set`, picking the right check function, this will reduce the duplication.

roryqi: Fixed.

Issue #9180, opened 2025-11-19
[Bug report] Spark connector: normal user with USE_CATALOG, USE_SCHEMA, CREATE_TABLE privileges on metalake failed to create a table operation


danhuawang:
### Version

main branch

### Describe what's wrong

A normal user  has privileges as following:
```
{
    "code": 0,
    "role": {
        "name": "role1",
        "audit": {
            "creator": "anonymous",
            "createTime": "2025-11-19T10:07:37.471984Z",
            "lastModifier": "anonymous",
            "lastModifiedTime": "2025-11-19T10:10:07.063395Z"
        },
        "properties": {
            "k1": "v1"
        },
        "securableObjects": [
            {
                "type": "metalake",
                "privileges": [
                    {
                        "name": "use_schema",
                        "condition": "allow"
                    },
                    {
                        "name": "use_catalog",
                        "condition": "allow"
                    },
                    {
                        "name": "create_table",
                        "condition": "allow"
                    }
                ],
                "fullName": "test"
            }
        ]
    }
}
```
but lack of load_table  privileges in spark sql

```
export SPARK_USER=normal
./spark-sql -v --conf spark.plugins="org.apache.gravitino.spark.connector.plugin.GravitinoSparkPlugin" --conf spark.sql.gravitino.uri=http://127.0.0.1:8090 --conf spark.sql.gravitino.metalake=test --conf spark.sql.gravitino.enableIcebergSupport=true --conf spark.sql.gravitino.client.socketTimeoutMs=60000 --conf spark.sql.gravitino.client.connectionTimeoutMs=60000 --conf spark.sql.warehouse.dir=file:///tmp/

spark-sql ()> use dml;
use dml
Time taken: 0.049 seconds
spark-sql (dml)> CREATE TABLE test1 (id bigint COMMENT 'unique id');
CREATE TABLE test1 (id bigint COMMENT 'unique id')
25/11/19 18:10:31 ERROR SparkSQLDriver: Failed in [CREATE TABLE test1 (id bigint COMMENT 'unique id')]
org.apache.gravitino.exceptions.ForbiddenException: User 'normal' is not authorized to perform operation 'loadTable' on metadata 'test1'
        at org.apache.gravitino.client.ErrorHandlers$TableErrorHandler.accept(ErrorHandlers.java:394)
        at org.apache.gravitino.client.ErrorHandlers$TableErrorHandler.accept(ErrorHandlers.java:363)
[...]

PR #9183, opened 2025-11-20
[#9180] fix(spark-connector): Override tableExists() to bypass authorization check

bharos:

Override tableExists() in BaseCatalog to catch ForbiddenException when users lack LOAD_TABLE privilege. Returns false to allow CREATE TABLE IF NOT EXISTS to proceed.

Users with only CREATE_TABLE privilege couldn't execute CREATE TABLE IF NOT EXISTS because tableExists() internally calls loadTable(), which requires LOAD_TABLE privilege.

Fix: #9180

Yes. Users with CREATE_TABLE privilege can now execute CREATE TABLE IF NOT EXISTS without needing LOAD_TABLE privilege.

Added integration test

### DIFF
--- a/catalogs/catalog-generic-lakehouse/build.gradle.kts
+++ b/catalogs/catalog-generic-lakehouse/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
   implementation(libs.commons.lang3)
   implementation(libs.guava)
   implementation(libs.hadoop3.client.api)
+  implementation(libs.hadoop3.client.runtime)
   implementation(libs.lance)
 
   annotationProcessor(libs.lombok)
--- a/catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
+++ b/catalogs/catalog-generic-lakehouse/src/main/java/org/apache/gravitino/catalog/lakehouse/GenericLakehouseCatalogOperations.java
@@ -19,6 +19,7 @@
 package org.apache.gravitino.catalog.lakehouse;
 
 import static org.apache.gravitino.Entity.EntityType.TABLE;
+import static org.apache.gravitino.catalog.lakehouse.GenericLakehouseTablePropertiesMetadata.LOCATION;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
@@ -26,6 +27,8 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Catalog;
 import org.apache.gravitino.Entity;
@@ -59,6 +62,8 @@ import org.apache.gravitino.rel.expressions.sorts.SortOrder;
 import org.apache.gravitino.rel.expressions.transforms.Transform;
 import org.apache.gravitino.rel.indexes.Index;
 import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Operations for interacting with a generic lakehouse catalog in Apache Gravitino.
@@ -68,6 +73,8 @@ import org.apache.hadoop.fs.Path;
  */
 public class GenericLakehouseCatalogOperations
     implements CatalogOperations, SupportsSchemas, TableCatalog {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(GenericLakehouseCatalogOperations.class);
 
   private static final String SLASH = "/";
 
@@ -79,8 +86,10 @@ public class GenericLakehouseCatalogOperations
   private static final Map<String, LakehouseCatalogOperations> SUPPORTED_FORMATS =
       Maps.newHashMap();
 
+  private Map<String, String> catalogConfig;
   private CatalogInfo catalogInfo;
   private HasPropertyMetadata propertiesMetadata;
+
   /**
    * Initializes the generic lakehouse catalog operations with the provided configuration.
    *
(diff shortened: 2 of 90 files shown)


roryqi:
Gravitino doesn't have API to dedicated the metadata object existence. Gravitino will use loadxxx API to verify the metadata object existence. So we would better to catch `ForbiddenException` and return false.

bharos:
> Gravitino doesn't have API to dedicated the metadata object existence. Gravitino will use loadxxx API to verify the metadata object existence. So we would better to catch `ForbiddenException` and return false.

The current fix bypasses gravitinoCatalogClient.loadTable() entirely to avoid requiring LOAD_TABLE privilege for an existence check. Instead, it uses sparkCatalog.tableExists()] which goes directly to the backend, ie. Iceberg REST server.

I think LOAD_TABLE should work with Iceberg REST server API ?
Key point here is that I am bypassing gravitinoCatalogClient and directly hitting the Iceberg REST API ?

roryqi:
> > Gravitino doesn't have API to dedicated the metadata object existence. Gravitino will use loadxxx API to verify the metadata object existence. So we would better to catch `ForbiddenException` and return false.
> 
> The current fix bypasses gravitinoCatalogClient.loadTable() entirely to avoid requiring LOAD_TABLE privilege for an existence check. Instead, it uses sparkCatalog.tableExists()] which goes directly to the backend, ie. Iceberg REST server.
> 
> I think LOAD_TABLE should work with Iceberg REST server API ? Key point here is that I am bypassing gravitinoCatalogClient and directly hitting the Iceberg REST API ?

We don't hit Iceberg REST API directly. 
For Gravitino Spark connector:
Spark -> Gravitino Spark connector -> Gravitino REST API (We don't use Iceberg REST API here) -> Hive backend/ Iceberg backed.

bharos:
> > > Gravitino doesn't have API to dedicated the metadata object existence. Gravitino will use loadxxx API to verify the metadata object existence. So we would better to catch `ForbiddenException` and return false.
> > 
> > 
> > The current fix bypasses gravitinoCatalogClient.loadTable() entirely to avoid requiring LOAD_TABLE privilege for an existence check. Instead, it uses sparkCatalog.tableExists()] which goes directly to the backend, ie. Iceberg REST server.
> > I think LOAD_TABLE should work with Iceberg REST server API ? Key point here is that I am bypassing gravitinoCatalogClient and directly hitting the Iceberg REST API ?
> 
> We don't hit Iceberg REST API directly. For Gravitino Spark connector: Spark -> Gravitino Spark connector -> Gravitino REST API (We don't use Iceberg REST API here) -> Hive backend/ Iceberg backed.

@jerqi Just to make sure I understand the flow correctly here:
BaseCatalog has gravitinoCatalogClient (which I thought connects to Gravitino REST APIs)
and sparkCatalog (which I am using in the fix, as ` sparkCatalog.tableExists(ident);`) which I think hits the Iceberg REST API ?

gravitinoCatalogClient -> Gravitino main REST API: http://gravitino-server:8090/api/ 
sparkCatalog -> Gravitino Iceberg REST server: http://gravitino-server:9001/iceberg/ 

Isn't this the case ? 

Based on my above understanding , I thought it will skip the Gravitino API and the fix will work (as it hits the Iceberg REST API directly)

roryqi:
> > > > Gravitino doesn't have API to dedicated the metadata object existence. Gravitino will use loadxxx API to verify the metadata object existence. So we would better to catch `ForbiddenException` and return false.
> > > 
> > > 
> > > The current fix bypasses gravitinoCatalogClient.loadTable() entirely to avoid requiring LOAD_TABLE privilege for an existence check. Instead, it uses sparkCatalog.tableExists()] which goes directly to the backend, ie. Iceberg REST server.
> > > I think LOAD_TABLE should work with Iceberg REST server API ? Key point here is that I am bypassing gravitinoCatalogClient and directly hitting the Iceberg REST API ?
> > 
> > 
> > We don't hit Iceberg REST API directly. For Gravitino Spark connector: Spark -> Gravitino Spark connector -> Gravitino REST API (We don't use Iceberg REST API here) -> Hive backend/ Iceberg backed.
> 
> @jerqi Just to make sure I understand the flow correctly here: BaseCatalog has gravitinoCatalogClient (which I thought connects to Gravitino REST APIs) and sparkCatalog (which I am using in the fix, as ` sparkCatalog.tableExists(ident);`) which I think hits the Iceberg REST API ?
> 
> gravitinoCatalogClient -> Gravitino main REST API: http://gravitino-server:8090/api/ sparkCatalog -> Gravitino Iceberg REST server: http://gravitino-server:9001/iceberg/
> 
> Isn't this the case ?
> 
> Based on my above understanding , I thought it will skip the Gravitino API and the fix will work (as it hits the Iceberg REST API directly)

I got it. Yes, spark catalog will access IRC. But it's better to access Gravitino for metadata operations in Gravitino spark connector.

bharos:
> > > > > Gravitino doesn't have API to dedicated the metadata object existence. Gravitino will use loadxxx API to verify the metadata object existence. So we would better to catch `ForbiddenException` and return false.
> > > > 
> > > > 
> > > > The current fix bypasses gravitinoCatalogClient.loadTable() entirely to avoid requiring LOAD_TABLE privilege for an existence check. Instead, it uses sparkCatalog.tableExists()] which goes directly to the backend, ie. Iceberg REST server.
> > > > I think LOAD_TABLE should work with Iceberg REST server API ? Key point here is that I am bypassing gravitinoCatalogClient and directly hitting the Iceberg REST API ?
> > > 
> > > 
> > > We don't hit Iceberg REST API directly. For Gravitino Spark connector: Spark -> Gravitino Spark connector -> Gravitino REST API (We don't use Iceberg REST API here) -> Hive backend/ Iceberg backed.
> > 
> > 
> > @jerqi Just to make sure I understand the flow correctly here: BaseCatalog has gravitinoCatalogClient (which I thought connects to Gravitino REST APIs) and sparkCatalog (which I am using in the fix, as ` sparkCatalog.tableExists(ident);`) which I think hits the Iceberg REST API ?
> > gravitinoCatalogClient -> Gravitino main REST API: http://gravitino-server:8090/api/ sparkCatalog -> Gravitino Iceberg REST server: http://gravitino-server:9001/iceberg/
> > Isn't this the case ?
> > Based on my above understanding , I thought it will skip the Gravitino API and the fix will work (as it hits the Iceberg REST API directly)
> 
> I got it. Yes, spark catalog will access IRC. But it's better to access Gravitino for metadata operations in Gravitino spark connector.

Thanks @jerqi for confirming the workflow.
I made the changes according to your recommendation + added a test

jerqi (approved):
LGTM. Thanks for your contribution. Merged.

Issue #9136, opened 2025-11-16
[Improvement] In DorisUtils.java improve parsing of AUTO/bucket numbers


justinmclean:
### What would you like to be improved?

The bucket number should be trimmed up front to stop a potential NumberFormatException.

Here's a unit test to help
```

  @Test
  public void testExtractBucketNumWithWhitespace() {
    String createTableSqlWithWhitespace =
        "CREATE TABLE `testTable` (\n`col1` int NOT NULL\n) ENGINE=OLAP\n DISTRIBUTED BY HASH(`col1`) BUCKETS  16 ";
    Distribution distribution = DorisUtils.extractDistributionInfoFromSql(createTableSqlWithWhitespace);
    assertEquals(16, distribution.number());

    String createTableSqlWithAutoWhitespace =
        "CREATE TABLE `testTable` (\n`col1` int NOT NULL\n) ENGINE=OLAP\n DISTRIBUTED BY HASH(`col1`) BUCKETS  AUTO ";
    Distribution distributionAuto =
        DorisUtils.extractDistributionInfoFromSql(createTableSqlWithAutoWhitespace);
    assertEquals(-1, distributionAuto.number());

    String createTableSqlWithAutoLeadingWhitespace =
        "CREATE TABLE `testTable` (\n`col1` int NOT NULL\n) ENGINE=OLAP\n DISTRIBUTED BY HASH(`col1`) BUCKETS  AUTO";
    Distribution distributionAutoLeading =
        DorisUtils.extractDistributionInfoFromSql(createTableSqlWithAutoLeadingWhitespace);
    assertEquals(-1, distributionAutoLeading.number());
  }
```

PCloud63514:
Hi @justinmclean 
Can I take this issue? 
It looks like it can be resolved by simply adjusting the position of the trim() call.

PR #9189, opened 2025-11-20
[#9136] fix(doris): trim bucket number to prevent NumberFormatException

PCloud63514:

- Modified DorisUtils.extractBucketNum to trim the bucket number string before parsing.

- To prevent a NumberFormatException when parsing numeric bucket numbers that include whitespace. (e.g., " 16 ")

Fix: #9136 

- No.

- Verified using the unit test included in the issue (testExtractBucketNumWithWhitespace()).

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
@@ -223,10 +223,10 @@ public final class DorisUtils {
   private static int extractBucketNum(Matcher matcher) {
     int bucketNum = 1;
     if (matcher.group(4) != null) {
-      String bucketValue = matcher.group(4);
+      String bucketValue = matcher.group(4).trim();
       // Use -1 to indicate auto bucket.
       bucketNum =
-          bucketValue.trim().toUpperCase().equals("AUTO")
+          bucketValue.toUpperCase().equals("AUTO")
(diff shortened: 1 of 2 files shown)


catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
```diff
@@ -223,10 +223,10 @@ public static Distribution extractDistributionInfoFromSql(String createTableSql)
   private static int extractBucketNum(Matcher matcher) {
     int bucketNum = 1;
     if (matcher.group(4) != null) {
-      String bucketValue = matcher.group(4);
+      String bucketValue = matcher.group(4).trim();
       // Use -1 to indicate auto bucket.
       bucketNum =
-          bucketValue.trim().toUpperCase().equals("AUTO")
```
yuqi1129, line 229: I can't see any difference here? Could you make it clearer?

PCloud63514: > I can't see any difference here? Could you make it clearer?

I believe the issue is that trim() is only being applied for the AUTO check (line 229).

When it's a number (line 231), the original, untrimmed bucketValue is used, which causes the NumberFormatException if there's whitespace.

yuqi1129: I see.

Issue #9075, opened 2025-11-11
[Improvement] Confusing error with Gravitino command line


justinmclean:
### What would you like to be improved?

When running ./bin/gcli.sh  in a distribution, it possible to get this error:

Error: Unable to access jarfile /Users/<user>/gravitino/distribution/package/auxlib/gravitino-cli-1.1.0-SNAPSHOT.jar

Even when /Users/<user>/gravitino/distribution/package/auxlib/gravitino-cli-1.1.0-SNAPSHOT.jar exists

### How should we improve?

This occurs when you have two jars in the auxlib directory e.g:
/Users/<user>/gravitino/distribution/package/auxlib/gravitino-cli-1.1.0-SNAPSHOT.jar
/Users/<user>/gravitino/distribution/package/auxlib/gravitino-cli-1.0.0-SNAPSHOT.jar

Kook-s:
Hi @justinmclean 
I'd like to work on this issue.  
Could you assign it to me?

justinmclean:
Sure if you have any questions just ask

PR #9203, opened 2025-11-21
[#9075] fix CLI jar selection when multiple CLI jars exist (issue #9075)

Kook-s:

This PR fixes an issue where gcli.sh fails to run when multiple CLI jars exist inside the auxlib directory.
The original implementation used find, which returns multiple paths if more than one matching jar exists.
This results in a multi-line string being passed to java -jar, causing an error

This PR updates the logic to:
- list all matching CLI jars
- sort them using version-aware sorting (sort -V)
-  select only one valid jar is passed to the JVM.

This ensures only one valid jar is passed

When multiple CLI jars exist (e.g., gravitino-cli-1.1.0-SNAPSHOT.jar and gravitino-cli-1.0.0-SNAPSHOT.jar),
the previous logic assigned a multi-line value to CLI_JAR.
Passing this to java -jar results in failure even though the jar exists

Fix: #9075

Users running ./bin/gcli.sh in a distribution will now reliably run the latest available CLI jar even if multiple versions exist in auxlib
This prevents confusing errors

Automated tests do not exist for this script, but the logic minimal and easy to validate manually.

### DIFF
--- a/bin/gcli.sh.template
+++ b/bin/gcli.sh.template
@@ -29,7 +29,7 @@ AUXLIB_DIR="${GRAVITINO_HOME}/auxlib"
 
 # Search strategy for finding the Gravitino CLI JAR:
 if [ -z "$CLI_JAR" ]; then
-    CLI_JAR=$(find "${AUXLIB_DIR}" -name "gravitino-cli-*.jar" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


Abyss-lord (approved):
LGTM, Thanks for fixing this.

Issue #9204, opened 2025-11-21
[Bug report] If metadata object owner is removed from metalake,  get owner api still return the owner


danhuawang:
### Version

main branch

### Describe what's wrong

If metadata object owner is removed from metalake,  get owner api still return the owner.
After I restart the server. get owner api still return  empty.

### Error message and/or stacktrace

N/A

### How to reproduce

1. add user normal1 to metalake test
2. create catalog pg1
3. assign the catalog pg1 to user normal1
4. delete the user normal1
5. get the catalog owner, return normal1
6. restart the gravitino server, get the catalog owner, return empty

danhuawang:
@yuqi1129 hi, could you help check if it's a cache issue?

yuqi1129:
Let me take a look

PR #9205, opened 2025-11-21
[#9204] fix(core): Fix can't clear cache when the user of a metadata object is deleted.

yuqi1129:

Add the reverse index for owner and metadata objects.

Without this, we will get an out-of-date user information.

Fix: #9204 

N/A

ITs

### DIFF
--- a/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/OwnerIT.java
+++ b/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/OwnerIT.java
@@ -457,6 +457,7 @@ public class OwnerIT extends BaseIT {
       Assertions.assertNull(ownerAfterDrop);
     }
     // Cleanup
+    client.disableMetalake(metalakeNameE);
     client.dropMetalake(metalakeNameE, true);
   }
 }


core/src/main/java/org/apache/gravitino/cache/ReverseIndexRules.java
```diff
@@ -71,6 +78,12 @@ public class ReverseIndexRules {
                     reverseIndexCache.put(nameIdentifier, Entity.EntityType.ROLE, key);
                   });
         }
+
+        // Handle Securable Objects -> Group reverse index, so the key type is group and the value
+        // type is securable Object.
+        if (key.relationType() == SupportsRelationOperations.Type.OWNER_REL) {
```
roryqi, line 92: Could u have better abstraction instead of condition?

yuqi1129: resolved

core/src/main/java/org/apache/gravitino/cache/ReverseIndexCache.java
```diff
@@ -41,18 +41,21 @@
  * efficiently store and retrieve relationships between entities based on their keys.
  */
 public class ReverseIndexCache {
-  private RadixTree<List<EntityCacheKey>> reverseIndex;
+  private final RadixTree<List<EntityCacheKey>> reverseIndex;
   /** Registers a reverse index processor for a specific entity class. */
-  private final Map<Class<? extends Entity>, ReverseIndexRule> reverseIndexRules = new HashMap<>();
+  private final Map<Class<? extends Entity>, List<ReverseIndexRule>> reverseIndexRules =
```
jerryshao: You can use Guava's `MultiMap` to simplify the code here.

Issue #9107, opened 2025-11-13
[Improvement] refactor class `TestJDBCBackend` and `TestEntityStorage`


mchades:
### What would you like to be improved?

- The `TestJDBCBackend` is too large and it only test the H2 backend, missing MySQL and PG tests
- The `TestEntityStorage` is too large and may overlap with the `TestJDBCBackend`

### How should we improve?

refactor class `TestJDBCBackend` and `TestEntityStorage`

Issue #6395, opened 2025-02-05
[Subtask]  Gravitino Iceberg REST server express supported endpoints via endpoint field in ConfigResponse


FANNG1:
### Describe the subtask

Iceberg 1.7 add an `endpoint` field in config endpoint to represent server capabilities. The Iceberg client do proper action according to the server capacibilites.

### Parent issue

#6336

FANNG1:
This depends on #6337

FANNG1:
fixed by #7917

PR #9207, opened 2025-11-21
[#9107] refactor(TestJDBCBackend): refactor TestJDBCBackend and fix bugs

mchades:

- refactor TestJDBCBackend
- Fix bugs found during the test period

Fix: #9107 

no

CI pass

### DIFF
--- a/core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
+++ b/core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
@@ -115,22 +115,26 @@ public class TestSQLScripts extends TestJDBCBackend {
     try (SqlSession sqlSession =
         SqlSessionFactoryHelper.getInstance().getSqlSessionFactory().openSession(true)) {
       try (Connection connection = sqlSession.getConnection()) {
-        try (Statement statement = connection.createStatement()) {
-          if ("postgresql".equals(backendType)) {
-            dropAllTablesForPostgreSQL(connection);
-          } else {
-            String query = "SHOW TABLES";
-            List<String> tableList = new ArrayList<>();
-            try (ResultSet rs = statement.executeQuery(query)) {
-              while (rs.next()) {
-                tableList.add(rs.getString(1));
-              }
-            }
-            for (String table : tableList) {
-              statement.execute("DROP TABLE " + table);
-            }
-          }
+        if ("postgresql".equals(backendType)) {
+          dropAllTablesForPostgreSQL(connection);
+        } else {
+          dropAllTablesForMySQLCompatible(connection);
         }
+      }
+    }
+  }
+
+  private void dropAllTablesForMySQLCompatible(Connection connection) throws SQLException {
+    try (Statement statement = connection.createStatement()) {
+      String query = "SHOW TABLES";
+      List<String> tableList = new ArrayList<>();
+      try (ResultSet rs = statement.executeQuery(query)) {
+        while (rs.next()) {
+          tableList.add(rs.getString(1));
+        }
+      }
+      for (String table : tableList) {
+        statement.execute("DROP TABLE " + table);
       }
     }
   }


core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/SecurableObjectPostgreSQLProvider.java
```diff
@@ -160,7 +154,7 @@ public String deleteSecurableObjectsByLegacyTimeline(
     return "DELETE FROM "
         + SECURABLE_OBJECT_TABLE_NAME
         + " WHERE id IN (SELECT id FROM "
-        + ROLE_TABLE_NAME
+        + SECURABLE_OBJECT_TABLE_NAME
```
mchades, line 157: @jerqi could you plz help to confirm this change?

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/FilesetMetaPostgreSQLProvider.java
```diff
@@ -29,35 +29,31 @@ public class FilesetMetaPostgreSQLProvider extends FilesetMetaBaseSQLProvider {
   public String softDeleteFilesetMetasByMetalakeId(Long metalakeId) {
     return "UPDATE "
         + META_TABLE_NAME
-        + " SET deleted_at = floor(extract(epoch from(current_timestamp -"
-        + " timestamp '1970-01-01 00:00:00')) * 1000)"
+        + " SET deleted_at = CAST(EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000 AS BIGINT)"
```
mchades, line 32: The soft deletion expression in PG has a time zone issue and cannot update records correctly. I have changed all the expressions. Could you please help confirm this change? @yuqi1129

yuqi1129: This part is okay with me, and I have been aware of the timezone problem but did not see it as a big problem as we can do the delete twice if we failed to drop it in the first time.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/SecurableObjectPostgreSQLProvider.java
roryqi: Yes, this should be `SECURABLE_OBJECT_TABLE_NAME`.

roryqi: cc @yuqi1129

yuqi1129: Yes, it should be a bug caused by #6395 and the current fix is okay.

build.gradle.kts
```diff
@@ -209,8 +209,9 @@ allprojects {
         param.include("**/integration/test/**")
       }
 
+      val dockerTest = project.rootProject.extra["dockerTest"] as? Boolean ?: false
```
yuqi1129, line 212: Why do we move it out of unit tests block?

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/PolicyMetaPostgreSQLProvider.java
```diff
@@ -61,21 +59,21 @@ public String insertPolicyMetaOnDuplicateKeyUpdate(PolicyPO policyPO) {
...
         + " VALUES ("
-        + " #{policyPO.policyId},"
-        + " #{policyPO.policyName},"
-        + " #{policyPO.policyType},"
-        + " #{policyPO.metalakeId},"
-        + " #{policyPO.auditInfo},"
-        + " #{policyPO.currentVersion},"
-        + " #{policyPO.lastVersion},"
-        + " #{policyPO.deletedAt})"
+        + " #{policyMeta.policyId},"
```
yuqi1129, line 62: `policyMeta` should be the wrong name? I do not find it in the method signature

core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
```diff
@@ -18,164 +18,139 @@
...
-                throw new RuntimeException("Failed to execute SQL: " + sql, e);
+  private void dropAllTables() throws SQLException {
+    try (SqlSession sqlSession =
+        SqlSessionFactoryHelper.getInstance().getSqlSessionFactory().openSession(true)) {
+      try (Connection connection = sqlSession.getConnection()) {
+        try (Statement statement = connection.createStatement()) {
+          if ("postgresql".equals(backendType)) {
+            dropAllTablesForPostgreSQL(connection);
+          } else {
+            String query = "SHOW TABLES";
```
yuqi1129: Better to extract a method named `dropAllTablesInH2OrMySQL` or something similar.

core/src/test/java/org/apache/gravitino/storage/relational/BackendTestExtension.java
```diff
@@ -0,0 +1,231 @@
...
+      return String.format("[%s Backend]", backendType.toUpperCase());
+    }
+
+    @Override
+    public List<Extension> getAdditionalExtensions() {
+      return Collections.singletonList(new BackendSetupCallback(backendType));
+    }
+  }
+
+  private static class BackendSetupCallback implements BeforeEachCallback {
```
yuqi1129, line 141: Does the `BeforeEach` here mean each test in the test class or each test mode here?

build.gradle.kts
mchades: I need to export `dockerTest` to the environment for use by `TestJDBCBackend`

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/PolicyMetaPostgreSQLProvider.java
mchades: see https://github.com/apache/gravitino/blob/9e0ff42f27c12b0fceca1f1f4f2e3188dcd8c98f/core/src/main/java/org/apache/gravitino/storage/relational/mapper/PolicyMetaSQLProviderFactory.java#L60-L63

the annotation is `policyMeta`

core/src/test/java/org/apache/gravitino/storage/relational/BackendTestExtension.java
mchades: both. Each test in the test class will run three times (for H2, MySQL, PG) if the env `dockerTest` is true

core/src/test/java/org/apache/gravitino/storage/TestSQLScripts.java
mchades: fixed

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/UserRoleRelPostgreSQLProvider.java
```diff
@@ -32,8 +32,7 @@ public class UserRoleRelPostgreSQLProvider extends UserRoleRelBaseSQLProvider {
   public String softDeleteUserRoleRelByUserId(Long userId) {
     return "UPDATE "
         + USER_ROLE_RELATION_TABLE_NAME
-        + " SET deleted_at = floor(extract(epoch from(current_timestamp -"
-        + " timestamp '1970-01-01 00:00:00')) * 1000)"
+        + " SET deleted_at = CAST(EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000 AS BIGINT)"
```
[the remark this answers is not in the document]
mchades: The comment isn't quite right. `EXTRACT(EPOCH)` returns the time as a float, including the fractional seconds.

If we use the suggested formula (multiplying the epoch and adding the milliseconds separately), we are effectively adding the milliseconds twice. The original code is safe because double precision is enough to hold the current timestamp without losing data.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/PolicyMetaPostgreSQLProvider.java
yuqi1129: OK, it's better to `policyMeta` to `policyPO` to reduce misleading in `PolicyMetaPostgreSQLProvider`.

yuqi1129:
Just one comments, others LGTM

Issue #9241, opened 2025-11-25
[Subtask] Add tag-related interface


Abyss-lord:
### Describe the subtask

Add tag-related interface
1. Add `SupportsTags` interface
2. Add `TagOperations` interface
3. Add `TagChange` and related dataclass

### Parent issue

#8794

PR #9242, opened 2025-11-25
[#9241] feat(python-client): Add tag-related interface

Abyss-lord:

Add tag-related interface:
1. Add SupportsTags interface
2. Add TagOperations interface
3. Add TagChange and related dataclass

Fix: #9241 

no

local test.

### DIFF
--- a/clients/client-python/gravitino/api/tag/supports_tags.py
+++ b/clients/client-python/gravitino/api/tag/supports_tags.py
@@ -18,7 +18,6 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import List
 
 from gravitino.api.tag.tag import Tag
 
@@ -30,20 +29,20 @@ class SupportsTags(ABC):
     """
 
     @abstractmethod
-    def list_tags(self) -> List[str]:
+    def list_tags(self) -> list[str]:
         """List all the tag names for the specific object.
 
         Returns:
-            List[str]: The list of tag names.
+            list[str]: The list of tag names.
         """
         pass
 
     @abstractmethod
-    def list_tags_info(self) -> List[Tag]:
+    def list_tags_info(self) -> list[Tag]:
         """List all the tags with details for the specific object.
 
         Returns:
-            List[Tag]: The list of tags.
+            list[Tag]: The list of tags.
         """
         pass
 
@@ -64,8 +63,8 @@ class SupportsTags(ABC):
 
     @abstractmethod
     def associate_tags(
-        self, tags_to_add: List[str], tags_to_remove: List[str]
-    ) -> List[str]:
+        self, tags_to_add: list[str], tags_to_remove: list[str]
+    ) -> list[str]:
         """Associate tags to the specific object.
 
         The `tags_to_add` will be added to the object, and the `tags_to_remove` will be removed from the object.
@@ -76,13 +75,13 @@ class SupportsTags(ABC):
         3. If the tag is already associated with the object, it will raise `TagAlreadyAssociatedException`.
 
         Args:
-            tags_to_add (List[str]): The arrays of tag name to be added to the object.
-            tags_to_remove (List[str]): The array of tag name to be removed from the object.
+            tags_to_add (list[str]): The arrays of tag name to be added to the object.
+            tags_to_remove (list[str]): The array of tag name to be removed from the object.
 
         Raises:
             TagAlreadyAssociatedException: If the tag is already associated with the object.
 
         Returns:
-            List[str]: The array of tag names that are associated with the object.
+            list[str]: The array of tag names that are associated with the object.
         """
         pass
--- a/clients/client-python/gravitino/api/tag/tag.py
+++ b/clients/client-python/gravitino/api/tag/tag.py
@@ -18,7 +18,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import ClassVar, Dict, List, Optional
+from typing import ClassVar, Optional
 
 from gravitino.api.auditable import Auditable
 from gravitino.api.metadata_object import MetadataObject
@@ -39,11 +39,11 @@ class AssociatedObjects(ABC):
         return 0 if objects is None else len(objects)
 
     @abstractmethod
-    def objects(self) -> Optional[List[MetadataObject]]:
+    def objects(self) -> Optional[list[MetadataObject]]:
         """Get the associated objects.
 
         Returns:
-            Optional[List[MetadataObject]]: The list of objects that are associated with this tag..
+            Optional[list[MetadataObject]]: The list of objects that are associated with this tag..
         """
         pass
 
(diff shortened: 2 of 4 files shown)


Abyss-lord:
Hi @jerryshao  @tsungchih  @mchades , could you please review this PR when you have time? I’d really appreciate your feedback.

clients/client-python/gravitino/api/tag/supports_tags.py
```diff
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from typing import List
```
tsungchih: I think we don't need to import `List` for type hints since the minimum Python version is 3.9.

clients/client-python/gravitino/api/tag/tag.py
```diff
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from typing import ClassVar, Dict, List, Optional
```
tsungchih: Same here, we don't need to import `Dict` and `List` for type hints.

clients/client-python/gravitino/api/tag/tag_change.py
```diff
@@ -0,0 +1,144 @@
...
+        return TagChange.RemoveProperty(tag_property)
+
+    @dataclass(frozen=True, eq=True)
+    class RenameTag:
+        """A tag change to rename the tag."""
+
+        _new_name: str = field(metadata=config(field_name="newName"))
+
+        @property
+        def new_name(self) -> str:
```
tsungchih, line 91: I'm wondering if we would like to conform with the API in Java client. In Java client, this is a method `getNewName()` instead of a property.

clients/client-python/gravitino/api/tag/tag_change.py
```diff
@@ -0,0 +1,144 @@
...
+            """The new name of the tag."""
+            return self._new_name
+
+        def __str__(self) -> str:
+            return f"RenameTag {self._new_name}"
```
tsungchih: The string representation for the prefix should be upper case, `RENAMETAG`.

clients/client-python/gravitino/api/tag/tag_change.py
```diff
@@ -0,0 +1,144 @@
...
+            """The property to set."""
+            return self._property
+
+        @property
+        def value(self) -> str:
+            """The value to set."""
+            return self._value
+
+        def __str__(self) -> str:
+            return f"SetProperty {self._property}={self._value}"
```
tsungchih: Similarly, the prefix should be `SETTAGPROPERTY` instead of `SetProperty`. And there should have a space in both side of `=` sign.

clients/client-python/gravitino/api/tag/tag_change.py
```diff
@@ -0,0 +1,144 @@
...
+        @property
+        def new_name(self) -> str:
+            """The new name of the tag."""
+            return self._new_name
+
+        def __str__(self) -> str:
+            return f"RenameTag {self._new_name}"
+
+    @dataclass(frozen=True, eq=True)
+    class UpdateComment:
```
tsungchih: I think the class name should be `UpdateTagComment`.

clients/client-python/gravitino/api/tag/tag_change.py
Abyss-lord: @tsungchih I believe that adhering to the Pythonic code style is the best approach here. In Python, using the `@property` decorator is more idiomatic and allows `new_name` to be accessed as a property rather than a method. This makes the code cleaner and more aligned with Python's conventions, in contrast to Java's getter methods.

clients/client-python/gravitino/api/tag/tag_change.py
Abyss-lord: fix

clients/client-python/gravitino/api/tag/tag.py
Abyss-lord: fix

clients/client-python/gravitino/api/tag/supports_tags.py
Abyss-lord: fix

clients/client-python/gravitino/api/tag/tag_change.py
Abyss-lord: fix

clients/client-python/gravitino/api/tag/tag_change.py
Abyss-lord: fix

Abyss-lord:
Hi @tsungchih , I've completed the code updates and would appreciate your review of the PR when you have a moment. Here's a summary of the commits:
1. Fix `__str__` method
2. Remove type hint.

tsungchih:
@jerryshao @mchades LGTM. Would you like to take a review?

clients/client-python/gravitino/api/tag/tag_change.py
```diff
@@ -0,0 +1,144 @@
...
+
+from dataclasses_json import config
+
+
+class TagChange(ABC):
+    """Interface for supporting tag changes.
+    This interface will be used to provide tag modification operations for each tag."""
+
+    @staticmethod
+    def rename(new_name: str) -> RenameTag:
```
unknowntpo, line 32: In java, we return `TagChange`, not `RenameTag`. Please also fix this for other method like `update_comment`, `set_property`, ...

Abyss-lord: @unknowntpo Thanks for the comment! I understand the intention to align the Python API with the Java API.
However, inheriting the inner concrete classes (e.g., `RenameTag`) from `TagChange` is not feasible in Python due to language constraints: Python does not allow an inner class to subclass its outer class, as the outer class is not fully defined at the time of parsing. This leads to circular reference and results in a `NameError`.

In addition, other similar modules in the Python client (for example, `model_change`) also do not return `ModelChange` as a concrete subclass, so the current pattern is consistent with existing implementations.

So I believe we should keep the current implementation.

https://github.com/apache/gravitino/blob/7d674098ec5da625e920a22707f3c6bd5a186e29/clients/client-python/gravitino/api/model_change.py#L38

unknowntpo: Okay, we can keep this aligned with the current implementations for now. Since the TagChange interface in the Java implementation acts like an outer class to organize tag-related changes—rather than defining abstract methods that force subclasses to implement them—this approach is acceptable.

unknowntpo:
LGTM. @jerryshao would you like to take a look at this PR ?

jerryshao:
Sorry for the late response. We're busy on the 1.1.0 release. I will spend time reviewing this PR later this week.

Issue #9240, opened 2025-11-25
[Improvement] Add configuration statement for MinIO in “fileset-catalog-with-s3” documentation


shaofengshi:
### What would you like to be improved?

When I configure Gravitino Fileset with MinIO, I encountered some problem which discussed in https://github.com/apache/gravitino/issues/6156 . Now I suggest to update the “fileset-catalog-with-s3” documentation to let more user know how to properly configure that. Besides, I noticed there are two wrong samples which need be corrected:

<img width="1352" height="632" alt="Image" src="https://github.com/user-attachments/assets/f374fe07-e993-4377-8caf-2b17aa0d96ff" />

<img width="1986" height="588" alt="Image" src="https://github.com/user-attachments/assets/ed0be6f0-4c5f-4137-a219-0ec6f87f41cf" />

### How should we improve?

Add some document for MinIO configuration. And correct the wrong sample value (which should be s3 endpoint, not Gravitino server address)

PR #9243, opened 2025-11-25
#9240 Improvement(doc): update document for MinIO s3 configuration

shaofengshi:

Update the document for s3 fileset catalog creation.

When I configure Gravitino Fileset with MinIO, I encountered some problem which discussed in https://github.com/apache/gravitino/issues/6156 . Now I suggest to update the “fileset-catalog-with-s3” documentation to let more user know how to properly configure that. 

Fix: #9240

No

Manually verify and review

### DIFF
--- a/docs/fileset-catalog-with-s3.md
+++ b/docs/fileset-catalog-with-s3.md
@@ -121,7 +121,13 @@ s3_catalog = gravitino_client.create_catalog(name="test_catalog",
 </Tabs>
 
 :::note
-When using S3, ensure that the location value starts with s3a:// (not s3://) for AWS S3. For example, use s3a://bucket/root, as the s3:// format is not supported by the hadoop-aws library.
+- When using S3, ensure that the location value starts with s3a:// (not s3://) for AWS S3. For example, use s3a://bucket/root, as the s3:// format is not supported by the hadoop-aws library.
+- When using MinIO or other S3-compatible storage services, make sure to set the `s3-endpoint` property to the appropriate endpoint URL. 
+- When using MinIO or other S3-compatible storage services, you may need to set additional properties such as `s3-path-style-access` to `true` depending on the storage service requirements. You can do this in Gravitino's `fileset.conf` file with the "gravitino.bypass." prefix: 
+```bash
+$ cat $GRAVITINO_HOME/catalogs/fileset/conf/fileset.conf
+gravitino.bypass.fs.s3a.path.style.access=true
+```
 :::
 
 ### Step2: Create a schema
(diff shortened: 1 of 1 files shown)


docs/fileset-catalog-with-s3.md
```diff
@@ -121,7 +121,13 @@ s3_catalog = gravitino_client.create_catalog(name="test_catalog",
 </Tabs>
 
 :::note
-When using S3, ensure that the location value starts with s3a:// (not s3://) for AWS S3. For example, use s3a://bucket/root, as the s3:// format is not supported by the hadoop-aws library.
+- When using S3, ensure that the location value starts with s3a:// (not s3://) for AWS S3. For example, use s3a://bucket/root, as the s3:// format is not supported by the hadoop-aws library.
+- When using MinIO or other S3-compatible storage services, make sure to set the `s3-endpoint` property to the appropriate endpoint URL. 
+- When using MinIO or other S3-compatible storage services, you may need to set additional properties such as `s3-path-style-access` to `true` depending on the storage service requirements. You can do this in Gravitino's `fileset.conf` file with the "gravitino.bypass." prefix:
```
yuqi1129, line 126: Let me check "Why does it not work if we set `s3-path-style-access` in catalog properties"?

yuqi1129: @shaofengshi 
> Setting "gravitino.bypass.fs.s3a.path.style.access=true" in the fileset catalog properties couldn't work.

Does it not work if we put `gravitino.bypass.fs.s3a.path.style.access=true` in catalog properties? That is to say, setting `gravitino.bypass.fs.s3a.path.style.access = true` when creating the catalog can't let us connect to the MinIO filesystem successfully?

shaofengshi: That doesn't work. I have already tested that. @yuqi1129

yuqi1129: I see.  I have tested that this scenario can't be reproduced on my local machine, and let me try it again.

shaofengshi: More information see my last comment on https://github.com/apache/gravitino/issues/6156

yuqi1129: @shaofengshi 
image

shaofengshi: or, this is also a workable way? suggest user to add this property at catalog level? @yuqi1129

shaofengshi: please try 1) using a docker compose environment; 2) use hostname for minio service insteaed of ip address. Previously community user reports they failed to use hostname but not knowing why. In a real environment most poeple not directly use ip address as that may change.

yuqi1129: I can't reproduce it with Docker or Docker Compose, anyway, adding the property `gravitino.bypass.fs.s3a.path.style.access = True` in the configuration file is a solution to fix the connection issues.

Issue #8943, opened 2025-10-28
[FEATURE] Add access control for policies


hdygxsj:
### Describe the feature

Add access control for policies

PR #9252, opened 2025-11-25
[#8943] feat(authz): support policy access control

hdygxsj:

support policy access control

Fix: #8943

None

org.apache.gravitino.client.integration.test.authorization.PolicyAuthorizationIT

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
@@ -196,9 +196,7 @@ public class MetadataAuthzHelper {
     }
     checkExecutor();
     AuthorizationRequestContext authorizationRequestContext = new AuthorizationRequestContext();
-      authorizationRequestContext.setOriginalAuthorizationExpression(expression);
-
-      return doFilter(
+    return doFilter(
         expression,
         entities,
         authorizer,


roryqi:
You should use `#` instead of `@`.

core/src/main/java/org/apache/gravitino/hook/PolicyHookDispatcher.java
```diff
@@ -0,0 +1,130 @@
...
+  @Override
+  public PolicyEntity createPolicy(
+      String metalake,
+      String name,
+      Policy.BuiltInType type,
+      String comment,
+      boolean enabled,
+      PolicyContent content)
+      throws PolicyAlreadyExistsException {
+    AuthorizationUtils.checkCurrentUser(metalake, PrincipalUtils.getCurrentUserName());
```
roryqi: We don't need to check current user, because we will check this in the intercept service. https://github.com/apache/gravitino/pull/9268

server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataIdConverter.java
```diff
@@ -93,6 +85,20 @@ public class MetadataIdConverter {
 
   private MetadataIdConverter() {}
 
+  static {
+    METADATA_TO_ENTITY_TYPE_MAPPING.put(MetadataObject.Type.METALAKE, Entity.EntityType.METALAKE);
```
roryqi: We already have similar methods in the class of NameIdentifierUtil.Could you reuse them?

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -96,14 +102,26 @@ public Response listPolicies(
                   Arrays.stream(policies)
                       .map(p -> toDTO(p, Optional.empty()))
                       .toArray(PolicyDTO[]::new);
-
+              policyDTOs =
```
roryqi, line 107: Could u add annotation for the list operations like https://github.com/apache/gravitino/pull/9271

hdygxsj: which annotation?

server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataIdConverter.java
roryqi: I can refactor this in my pull request. You can ignore this comment.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
roryqi: server/src/main/java/org/apache/gravitino/server/web/rest/TagOperations.java  Line 90.

core/src/main/java/org/apache/gravitino/hook/PolicyHookDispatcher.java
hdygxsj: fixed

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AssociatePolicyAuthorizationExecutor.java
```diff
@@ -0,0 +1,111 @@
...
+    context.setOriginalAuthorizationExpression(expression);
+    Entity.EntityType targetType =
+        Entity.EntityType.POLICY; // policies are the only supported batch target here
+    Preconditions.checkArgument(
+        request instanceof PoliciesAssociateRequest,
+        "Only policy can use AssociatePolicyAuthorizationExecutor, please contact the administrator.");
+    PoliciesAssociateRequest policiesAssociateRequest = (PoliciesAssociateRequest) request;
+    policiesAssociateRequest.validate();
+    // Authorize both 'policiesToAdd' and 'policiesToRemove' fields.
+    return authorizePolicy(policiesAssociateRequest.getPoliciesToAdd(), context, targetType)
```
jerqi: Do we need to execute the common logic, too?

core/src/main/java/org/apache/gravitino/cache/ReverseIndexRules.java
```diff
@@ -157,6 +157,14 @@ public class ReverseIndexRules {
                                 nsFileset.level(0),
                                 nsFileset.level(1));
                         break;
+                      case TAG:
```
roryqi, line 160: cc @yuqi1129 Could you take a look at this place?

yuqi1129: @hdygxsj 
I would not suggest you to add the code here as `policy` and `tag` are not securable objects strictly. If you want to keep the reverse index between `role`  and `tag` or `policy`, you can add a new rule, please refer to `USER_ROLE_REVERSE_RULE` and `USER_OWNERSHIP_REVERSE_RULE`.

hdygxsj: > @hdygxsj I would not suggest you to add the code here as `policy` and `tag` are not securable objects strictly. If you want to keep the reverse index between `role` and `tag` or `policy`, you can add a new rule, please refer to `USER_ROLE_REVERSE_RULE` and `USER_OWNERSHIP_REVERSE_RULE`.

We can now assign privileges to tags and policies—why are they not SecurableObjects?

image

yuqi1129: I see, and it's okay according to the current code base.
 By the way, please try to update the definition of `securable object` in `security/access-control.md`.

PR #9268, opened 2025-11-26
[#9270] : feat(iceberg): Refactor checkCurrentUser call to the interceptor

bharos:

Moving the check to the interception service ensures the user is validated **once per request** before authorization begins, improving performance while maintaining the same security guarantees.

Better refactor and minimize duplicate calls 

Fix : #9270 

No. The validation still occurs at the same point in the request flow with identical behavior - non-existent users still receive `ForbiddenException`.

Running all the tests

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -216,6 +216,12 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
     if (StringUtils.isBlank(currentUserName)) {
       return false;
     }
+
+    // Service admins are considered users of all metalakes
+    if (isServiceAdmin()) {
+      return true;
+    }
+
     try {
       return GravitinoEnv.getInstance()
               .entityStore()
--- a/server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
@@ -39,6 +39,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Entity;
 import org.apache.gravitino.MetadataObject;
 import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.authorization.AuthorizationUtils;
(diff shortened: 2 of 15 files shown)


roryqi:
Could you create an issue for this pull request?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -216,6 +216,12 @@ public boolean isMetalakeUser(String metalake) {
     if (StringUtils.isBlank(currentUserName)) {
       return false;
     }
+
+    // Service admins are considered users of all metalakes
+    if (isServiceAdmin()) {
```
roryqi, line 221: Do we still need this?

server/src/main/java/org/apache/gravitino/server/web/filter/GravitinoInterceptionService.java
```diff
@@ -125,18 +126,51 @@ private static class MetadataAuthorizationMethodInterceptor implements MethodInt
...
+            extractNameIdentifierFromParameters(parameters, args);
+
+        // Check if current user exists in the metalake.
+        NameIdentifier metalakeIdent = metadataContext.get(Entity.EntityType.METALAKE);
+
+        if (metalakeIdent != null) {
+          String currentUser = PrincipalUtils.getCurrentUserName();
+          try {
+            AuthorizationUtils.checkCurrentUser(metalakeIdent.name(), currentUser);
+          } catch (org.apache.gravitino.exceptions.ForbiddenException ex) {
```
roryqi, line 147: We must use `org.apache.gravitino.exceptions` here? Could we use the simple name?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
bharos: I think so.
Service admins need to access any metalake for administrative operations without being explicitly added as users ?

Will it work without this check

roryqi: Actually, we can remove the check `isMetalakeUser`. Because I have checked user in the InterceptService. But I think we can do this in next PR.

roryqi:
@bharos We forgot to remove the logic in the tagHookDispatcher.

Issue #9277, opened 2025-11-27
[Improvement] Improve the lance table creation process to avoid back-and-forth request


jerryshao:
### What would you like to be improved?

To support different modes in Lance table creation, we use some catch-and-throw solution in the client-side, which is time-consuming. Instead, we can improve the code to do this in the server-side, just pass the creation mode as property in creation request, and handle such logic on the server-side.

PR #9279, opened 2025-11-27
[#9277] improve(lance): Improve Lance table creation to support creation modes

jerryshao:

This PR improves Lance table creation to support three creation modes (CREATE, EXIST_OK, OVERWRITE) by passing the mode as a property rather than using exception-based flow control.

**Key changes:**
- Add `LANCE_CREATION_MODE` property to support CREATE, EXIST_OK, and OVERWRITE modes
- Refactor `LanceTableOperations.createTable()` to handle all three creation modes upfront
- Add `CreationMode` enum (CREATE, EXIST_OK, OVERWRITE) to `LanceTableOperations`
- Extract `createTableInternal()` method to handle actual table creation logic
- Simplify `GravitinoLanceTableOperations` to pass mode via property instead of exception handling
- Centralize Lance-related constants in `LanceConstants` class
- Automatically mark registered tables as external to preserve physical datasets
- Add comprehensive unit tests and integration tests

Fix: #9277

The current implementation uses exception handling (catch TableAlreadyExistsException) to handle different creation modes, which requires multiple round-trip requests and complicates the code. This change:
1. Eliminates multiple round-trip requests for EXIST_OK and OVERWRITE modes
2. Improves performance by handling modes server-side in a single request
3. Simplifies codebase by removing exception-based flow control
4. Makes the API more consistent and easier to maintain

Yes, users can now specify the `lance.creation-mode` property when creating or registering Lance tables:
- `CREATE` (default): Fails if table already exists
- `EXIST_OK`: Returns existing table if found, creates if not
- `OVERWRITE`: Drops/purges existing table and creates new one

- Added unit tests in `TestLanceTableOperations` for mode validation and error handling
- Added integration tests in `CatalogGenericCatalogLanceIT` for:
  - createTable with CREATE, EXIST_OK, and OVERWRITE modes
  - registerTable with CREATE, EXIST_OK, and OVERWRITE modes
- Existing tests in `LanceRESTServiceIT` already cover end-to-end scenarios
- All tests passing

### DIFF
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java
@@ -57,12 +57,13 @@ public class LanceTableDelegator implements LakehouseTableDelegator {
             false /* defaultValue */,
             false /* hidden */,
             false),
-        PropertyEntry.stringPropertyEntry(
+        PropertyEntry.enumPropertyEntry(
             LANCE_CREATION_MODE,
             "Creation mode for Lance table: CREATE, EXIST_OK, or OVERWRITE",
             false /* required */,
             true /* immutable */,
-            "CREATE" /* default value */,
+            LanceTableOperations.CreationMode.class,
+            LanceTableOperations.CreationMode.CREATE,
             false /* hidden */,
             false /* reserved */));
   }


catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -100,45 +109,48 @@ public Table createTable(
...
-    } catch (IllegalArgumentException e) {
-      if (e.getMessage().contains("Dataset already exists")) {
-        throw new TableAlreadyExistsException(
-            e, "Lance dataset already exists at location %s", location);
+    // Handle OVERWRITE mode - drop existing table if present
+    if (mode == CreationMode.OVERWRITE) {
+      if (register) {
+        dropTable(ident);
+      } else {
+        purgeTable(ident);
```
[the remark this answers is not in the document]
jerryshao: There's a lock for each operation, so it is fine without issue for single-node operation. But this should be an issue for multi-nodes depolyment.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -100,45 +109,48 @@ public Table createTable(
...
-      // dataset
-      return super.createTable(
-          ident, columns, comment, properties, partitions, distribution, sortOrders, indexes);
+
+    // Handle EXIST_OK mode - check if table exists first
+    if (mode == CreationMode.EXIST_OK) {
+      try {
+        return super.loadTable(ident);
+      } catch (NoSuchTableException e) {
+        // Table doesn't exist, proceed with creation
```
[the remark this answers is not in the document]
jerryshao: There's a lock for each operation, so it is fine without issue for single-node operation. But this should be an issue for multi-nodes depolyment.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java
```diff
@@ -44,20 +43,28 @@ public String tableFormat() {
...
-            PROPERTY_LANCE_TABLE_REGISTER,
+            LANCE_TABLE_REGISTER,
             "Whether this is a table registration operation.",
             false,
             true /* immutable */,
             false /* defaultValue */,
             false /* hidden */,
-            false));
+            false),
+        PropertyEntry.stringPropertyEntry(
```
yuqi1129: I would suggest you use an enum type property.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -100,45 +109,59 @@ public Table createTable(
...
+
+    // Extract creation mode from properties
+    CreationMode mode =
+        Optional.ofNullable(properties.get(LANCE_CREATION_MODE))
+            .map(
+                modeStr -> {
+                  try {
+                    return CreationMode.valueOf(modeStr);
+                  } catch (IllegalArgumentException e) {
+                    throw new IllegalArgumentException(
```
yuqi1129: I'm not very sure about the intention that `catch IllegalArgumentException and then throw an IllegalArgumentException`

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -100,45 +109,59 @@ public Table createTable(
...
             .map(Boolean::parseBoolean)
             .orElse(false);
-    if (register) {
-      // If this is a registration operation, just create the table metadata without creating a new
-      // dataset
-      return super.createTable(
-          ident, columns, comment, properties, partitions, distribution, sortOrders, indexes);
+
+    // Handle EXIST_OK mode - check if table exists first
+    if (mode == CreationMode.EXIST_OK) {
```
yuqi1129, line 125: I remember that only modes `create` and `overwrite` are allowed for registering a table.

jerryshao: I see. Let me update the code.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
jerryshao: I will fix it.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java
jerryshao: I can change it. But it is not so necessary. Because we cannot get `HasPropertyMetadata` object in `LanceTableOperations`, so we cannot leverage PropertyMetadata API.

yuqi1129 (approved):
LGTM

PR #9281, opened 2025-11-27
[#9143] Use orElseGet() instead of orElse() for lazy evaluation in Config.java

### LINKED ISSUE #9143: [Improvement] Use orElseGet() Instead of orElse() for Lazy Evaluation

- orElse(...) eagerly evaluates its argument, meaning the fallback expression is computed even when the Optional contains a value.
- In this method, that causes the lookup and mapping for GRAVITINO_HOME to run unnecessarily whenever GRAVITINO_CONF_DIR is set.

- Using orElseGet(...) (or the or(...) API) avoids this by evaluating the fallback only when needed, making the code more efficient and idiomatic.

https://github.com/apache/gravitino/blob/eaba9a23c507e91e116b826332c7df9cf7eddf64/common/src/main/java/org/apache/gravitino/Config.java#L85

### How should we improve?

Suggested improvement (lazy evaluation)
```
String confDir =
    Optional.ofNullable(System.getenv("GRAVITINO_CONF_DIR"))
        .orElseGet(() ->
            Optional.ofNullable(System.getenv("GRAVITINO_HOME"))
                .map(s -> s + File.separator + "conf")
                .orElse(null)
        );
```

zhan7236:

Use `orElseGet()` instead of `orElse()` when determining the configuration directory in `Config.loadFromFile()`.

- `orElse()` eagerly evaluates its argument, meaning the fallback expression is computed even when the Optional contains a value.
- This causes the lookup and mapping for `GRAVITINO_HOME` to run unnecessarily whenever `GRAVITINO_CONF_DIR` is set.
- Using `orElseGet()` with a Supplier lambda ensures the fallback expression is only evaluated when needed.

Fixes #9143

- Compiled successfully with `./gradlew :common:compileJava`
- All tests pass with `./gradlew :common:test` and `./gradlew :server:test`

### DIFF
--- a/common/src/main/java/org/apache/gravitino/Config.java
+++ b/common/src/main/java/org/apache/gravitino/Config.java
@@ -85,10 +85,11 @@ public abstract class Config {
   public Config loadFromFile(String name) throws Exception {
     String confDir =
         Optional.ofNullable(System.getenv("GRAVITINO_CONF_DIR"))
-            .orElse(
-                Optional.ofNullable(System.getenv("GRAVITINO_HOME"))
-                    .map(s -> s + File.separator + "conf")
-                    .orElse(null));
+            .orElseGet(
+                () ->
+                    Optional.ofNullable(System.getenv("GRAVITINO_HOME"))
+                        .map(s -> s + File.separator + "conf")
+                        .orElse(null));
 
     if (confDir == null) {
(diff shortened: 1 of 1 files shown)


jerryshao (approved):
LGTM.

Issue #9298, opened 2025-11-28
[Bug report] Lance register table does not work as expected


yuqi1129:
### Version

main branch

### Describe what's wrong

The key of `register` is defined as follows: 
https://github.com/apache/gravitino/blob/c206c0748a9e6b0a96b9eb68e593ef87df3d8e23/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java#L34

https://github.com/apache/gravitino/blob/c206c0748a9e6b0a96b9eb68e593ef87df3d8e23/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java#L105-L108

However, in the Lance REST server, we use 

https://github.com/apache/gravitino/blob/9e0ff42f27c12b0fceca1f1f4f2e3188dcd8c98f/lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java#L155-L169

so, it will always create a table even though we call `registerTable` API

Another point is that, in the `dropTable` API, a registered table should be an external table, so we need to add the property `external = true` in registering a table.  

https://github.com/apache/gravitino/blob/c206c0748a9e6b0a96b9eb68e593ef87df3d8e23/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java#L202-L223

### Error message and/or stacktrace

please see above.

### How to reproduce

please see above.

PR #9299, opened 2025-11-28
[#9298] improvement(lance-rest):  Add more tests in Lance REST and lakehouse generic catalog.

yuqi1129:

- Change the key from `register` to `lance.register` in register a lance table.
- set `external` to true for a registered table.

It's a bug.

Fix: #9298 

N/A

IT and existing tests.


jerryshao:
I've fixed some of the issues in another PR. The thing is that we don't have UTs/ITs to reflect the issue, or the issue is silent when we do the changes. So the key thing is that we need UTs to cover the code.

yuqi1129:
> I've fixed some of the issues in another PR. The thing is that we don't have UTs/ITs to reflect the issue, or the issue is silent when we do the changes. So the key thing is that we need UTs to cover the code.

Okay,  I will hold this PR and activate it based on later fixes.

jerryshao:
No, I mean I've fixed several bugs in this #9279 , so you don't need so much changes here. Just adding more tests to cover the code.

Issue #9295, opened 2025-11-28
[Bug report] server start failed


mchades:
### Version

main branch

### Describe what's wrong

build the gravitino image based on the main branch, server start failed

### Error message and/or stacktrace

<img width="2744" height="1290" alt="Image" src="https://github.com/user-attachments/assets/62d6dcf9-2088-43ac-9b14-861820cbfa57" />

### How to reproduce

1. build the image: `./dev/docker/build-docker.sh --platform linux/amd64 --type iceberg-rest-server --image iceberg-rest-server --tag 1.1.0-SNAPSHOT`
2. start up the container: `docker run -d --name gravitino-oss gravitino:1.1.0-SNAPSHOT`

mchades:
@jerqi cc

roryqi:
I will take a look at this issue.

PR #9313, opened 2025-12-01
[#9295] fix(core): Exclude different version jackson jars from the Lance dependency

roryqi:

Exclude different version jackson jars from the Lance dependency

Fix: #9295 

No need.

compile and check the jars in the distribution.
Before this PR
we have 2.15 and 2.17 jackson jars
After this PR
we only 2.15 jackson jars

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
   implementation(libs.concurrent.trees)
   implementation(libs.guava)
   implementation(libs.h2db)
+  implementation(libs.jackson.jaxrs.json.provider) // This is required by lance
   implementation(libs.lance) {
     exclude(group = "com.fasterxml.jackson.core", module = "*") // provided by gravitino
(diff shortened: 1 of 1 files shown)


core/build.gradle.kts
```diff
@@ -50,8 +50,10 @@ dependencies {
     exclude(group = "com.google.guava", module = "guava") // provided by gravitino
     exclude(group = "org.apache.commons", module = "commons-lang3") // provided by gravitino
     exclude(group = "org.junit.jupiter", module = "*") // provided by test scope
+    exclude(group = "com.fasterxml.jackson.jaxrs", module = "jackson-jaxrs-json-provider") // using gravitino's version
   }
   implementation(libs.mybatis)
+  implementation(libs.jackson.jaxrs.json.provider) // This is required by lance
```
jerryshao: Make this alphabetical ordering.

roryqi: Addressed.

PR #9316, opened 2025-12-01
[#8628] fix(build): Skip tests in release task to avoid JDK version conflicts

### LINKED ISSUE #8628: [Bug report] command `./gradlew release` failed
### Version

main branch

### Describe what's wrong

Executing the command `./gradlew release` failed

Image

### How to reproduce

`./gradlew release`

### Additional context

`./gradlew release -x test` works well

jerryshao:

Fix the `release` task to avoid JDK version conflicts when running tests.

Issue: #8628

The `./gradlew release` command fails because:
- The release task depends on all subproject `build` tasks, which include running tests
- Some modules (clients, spark-connector, flink-connector, bundles, api, common) are JDK 8 compatible
- Other modules (server, core, catalogs) require JDK 17
- When tests run during release, there are JDK version conflicts causing failures
- The workaround `./gradlew release -x test` works but is not ideal

1. Ran `./gradlew clean` to clean all build artifacts
2. Ran `./gradlew release` successfully (53s, 425 tasks)
3. Verified all 53 module JARs have correct bytecode versions:
   - 26 JARs are JDK 8 compatible (bytecode 52.0): clients, connectors, bundles, api, common
   - 27 JARs are JDK 17 (bytecode 61.0): server, core, catalogs, authorization modules
4. Confirmed the release completes without requiring `-x test`

### Change logs

- Modified release task to depend on `assemble` instead of `build`
- This skips test execution during release to avoid JDK version conflicts
- Tests should be run separately in CI/CD with appropriate JDK configurations
- Added filter to exclude `client-python` which doesn't have an assemble task
- Added comments explaining the rationale

Fixes #8628

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1185,8 +1185,9 @@ tasks.register("release") {
   // Tests have JDK version conflicts (some need JDK 8, some need JDK 17)
   // and should be run separately in CI/CD with appropriate JDK configurations
   // Only include subprojects that apply the Java plugin (exclude client-python)
-  dependsOn(subprojects
-    .filter { it.name != "client-python" }
-    .map { it.tasks.named("assemble") }
+  dependsOn(
+    subprojects
+      .filter { it.name != "client-python" }
+      .map { it.tasks.named("assemble") }
   )
 }


build.gradle.kts
```diff
@@ -1181,5 +1181,13 @@ tasks.register("release") {
...
-  dependsOn(subprojects.map { it.tasks.named("build") })
+  // Use 'assemble' instead of 'build' to skip tests during release
+  // Tests have JDK version conflicts (some need JDK 8, some need JDK 17)
+  // and should be run separately in CI/CD with appropriate JDK configurations
+  // Only include subprojects that apply the Java plugin (exclude client-python)
+  dependsOn(
+    subprojects
+      .filter { it.name != "client-python" }
+      .map { it.tasks.named("assemble") }
+  )
```
diqiu50, line 1192: This fix is no different from adding -x test. It also skips all tests.

jerryshao: Why can't we skip the test? Is there any reason that `release` should depend on `test`? I think there're many ways to guarantee there's no issue in testing.

diqiu50: Ultimately, we want `./gradlew build` to produce packages for the client and other modules that are compatible with JDK 8.  However, because some client modules have integration tests that depend on the core/server modules (built with JDK 17), the build fails. This is why we introduced the command ./gradlew release -x test. This issue is intended to resolve that problem, correct?

jerryshao: The first thing I want to fix is to make `./gradlew release` work without error. Secondly, I think whether `release` should rely on `test` is arguable; my feeling is that the purpose of `release` is to build jars for release, it is not so necessary to run the `test` again. Since there's no better solution, I think the current one is acceptable.

diqiu50: That's OK.

Issue #9307, opened 2025-12-01
[Improvement] In PartitionOperations.java a missing partion will cause sever error


justinmclean:
### What would you like to be improved?

Change the API to return a 400 response with the illegal-arguments error code and message.

Here's a unit test to help:
```
  @Test
  public void testAddPartitionWithNullPartitions() {
    AddPartitionsRequest invalidReq = new AddPartitionsRequest();

    Response resp =
        target(partitionPath(metalake, catalog, schema, table))
            .request(MediaType.APPLICATION_JSON_TYPE)
            .accept("application/vnd.gravitino.v1+json")
            .post(Entity.entity(invalidReq, MediaType.APPLICATION_JSON_TYPE));

    Assertions.assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), resp.getStatus());

    ErrorResponse errorResponse = resp.readEntity(ErrorResponse.class);
    Assertions.assertEquals(ErrorConstants.ILLEGAL_ARGUMENTS_CODE, errorResponse.getCode());
    Assertions.assertEquals(IllegalArgumentException.class.getSimpleName(), errorResponse.getType());
    Assertions.assertTrue(errorResponse.getMessage().contains("partitions must not be null"));
  }
```

HarshMehta112:
@justinmclean I’m interested in contributing to this work and would appreciate it if you could assign this task to me.

SalmaneKhalili:
@justinmclean 
Hello, if this is still open, would it be possible to have this assigned to me?  I've already fixed this locally!

PR #9317, opened 2025-12-01
Improvement : fix(server) : Fix missing-partition handling in PartitionOperations.java to avoid server errors

HarshMehta112:

This change updates the handling of partitions in `PartitionOperations` so that missing-partition cases are properly handled, avoiding server errors when partitions are missing or invalid.

Without this change, in certain scenarios where a partition is missing or not found, operations in `PartitionOperations` could trigger server-side errors — e.g. a NullPointerException or inappropriate behavior when partition metadata is absent. This fix ensures robustness by detecting missing partitions and handling them gracefully, preventing potential crashes.  

Fixes: #9307  

No — this change is internal to partition-handling logic. The public API remains the same; there are no changes to user-visible APIs or configuration keys.  

By applying the patch and running the existing test suite to ensure no regressions. For partition operations involving missing partitions, the behavior has been corrected (i.e. operations now fail gracefully instead of erroring). If necessary, additional unit tests should be added to cover missing-partition cases to ensure this fix works reliably under various scenarios.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
@@ -180,19 +180,18 @@ public class PartitionOperations {
       @PathParam("schema") @AuthorizationMetadata(type = Entity.EntityType.SCHEMA) String schema,
       @PathParam("table") @AuthorizationMetadata(type = Entity.EntityType.TABLE) String table,
       AddPartitionsRequest request) {
-    LOG.info(
-        "Received add {} partition(s) request for table {}.{}.{}.{} ",
-        request.getPartitions().length,
-        metalake,
-        catalog,
-        schema,
-        table);
-    Preconditions.checkArgument(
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
```diff
@@ -180,19 +180,18 @@ public Response addPartitions(
...
+
+      LOG.info(
+          "Received add {} partition(s) request for table {}.{}.{}.{} ",
+          request.getPartitions().length,
+          metalake,
+          catalog,
+          schema,
+          table);
+      Preconditions.checkArgument(
+          request.getPartitions().length == 1, "Only one partition is supported");
```
jerryshao, line 194: Hi @HarshMehta112, I don't quite understand what you're actually fixing and what the issue is here. Seems like you just changed the ordering of the code, does it fix anything?

mchades:
Thanks for your PR. I think it's fixed by #9397

Issue #9308, opened 2025-12-01
[Improvement] Correct logic in generateAlterTableSql in StarRocksTableOperations.java


justinmclean:
### What would you like to be improved?

This logic is not correct:
```
        if (!SUPPORTED_MODIFY_PROPERTIES.contains(setProperty.getProperty())
            && setProperty
                .getProperty()
                .startsWith(SUPPORTED_MODIFY_PROPERTIES_PREFIX_DYNAMIC_PARTITION)
            && setProperty.getProperty().startsWith(SUPPORTED_MODIFY_PROPERTIES_PREFIX_BINLOG)) {
```

As the property can't start with both prefixes.

justinmclean:
@ruanwenjun, can you comment here so I can assign the issue to you, or did you mean a different issue?

ruanwenjun:
@justinmclean Hi Justin, I'm very happy to solve this issue, but it's strange—I don't know why one of my commits is showing up here.

PR #9318, opened 2025-12-01
[#9308] improvement(catalog-starrocks): Correct logic in generateAlterTableSql in StarRocksTableOperations when setProperty

ruanwenjun:

Throws an exception when the property does not match the expected value

Fix: #9308

No

test by UT


catalogs/catalog-jdbc-starrocks/src/test/java/org/apache/gravitino/catalog/starrocks/operation/TestStarRocksTableOperations.java
```diff
@@ -605,9 +604,15 @@ public void testUnsupportedPropertyThrows() {
         null,
         indexes);
 
-    TableChange.SetProperty change = new TableChange.SetProperty("unsupported.property", "value");
+    Assertions.assertDoesNotThrow(
+        () ->
+            TABLE_OPERATIONS.alterTable(
+                databaseName, tableName, TableChange.setProperty("replication_num", "2")));
```
yuqi1129: The replication number should always be 1 in the test environment.

yuqi1129:
@ruanwenjun 
There seems to be something wrong with the test. Do you have time to handle it?

ruanwenjun:
> @ruanwenjun There seems to be something wrong with the test. Do you have time to handle it?

I will take a look

Issue #7915, opened 2025-08-04
[Improvement] Get credentials API needs access control


danhuawang:
### What would you like to be improved?

Currently, get credentials API doesn't have access control. Here's the api link: https://gravitino.apache.org/docs/0.9.1/api/rest/get-credentials

<img width="1899" height="1040" alt="Image" src="https://github.com/user-attachments/assets/43fcdd16-d493-4a1f-bb7a-e988c1cba351" />

### How should we improve?

Provide get credentials API access control.

### Parent issue

#6762

danhuawang:
@xunliu Could you help take a look at this issue?

PR #9338, opened 2025-12-02
[#7915] feat(authz): Get credentials API supports access control

roryqi:

Get credentials API supports access control

Fix: #7915 

No.

Added an IT, run it locally.


clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/FilesetCatalogAuthorizationCredentialIT.java
```diff
@@ -0,0 +1,393 @@
...
+      }
+    }
+  }
+
+  protected String genStorageLocation(String fileset) {
+    return String.format("s3a://%s/%s", BUCKET_NAME, fileset);
+  }
+
+  @Test
+  @Order(1)
```
FANNG1: any reason using `@Order`?

roryqi: Authorization test cases usually need some order. Because we may rely the created metadata objects which last test cases created. So I prefer using ordered test cases. But we don't need need the ordered cases here. Thanks for your good catch. I removed the order restriction now.

hdygxsj (commented):
LGTM for the authorization part. By the way, do we really have to use public cloud S3 for IT? Why not consider running MinIO in a container?

roryqi:
> LGTM for the authorization part. By the way, do we really have to use public cloud S3 for IT? Why not consider running MinIO in a container?

I run the test locally. I use the company's private S3 storage.  Actually, the MinIO can't test credential vending. MinIO doesn't have IAM interfaces. Credential vending is related about IAM interfaces.

FANNG1 (approved):

FANNG1:
LGTM

PR #9359, opened 2025-12-02
[#8906] improvement(authz): Add the check for setting owner operations

roryqi:

Add the check for setting owner operations

Fix: #8906 

No.

Add new test cases.


core/src/main/java/org/apache/gravitino/authorization/OwnerManager.java
```diff
@@ -64,6 +65,9 @@ public OwnerManager(EntityStore store) {
   @Override
   public void setOwner(
       String metalake, MetadataObject metadataObject, String ownerName, Owner.Type ownerType) {
+    // TODO: Support GROUP type owner in the future.
```
jerryshao: Can you create an issue and add the issue id here as the TODO comment?

roryqi: Fixed.

Issue #9337, opened 2025-12-02
[Subtask] add authz for scan planning APIs for IRC


FANNG1:
### Describe the subtask

add authz for scan planning APIs for IRC

### Parent issue

#4361

PR #9361, opened 2025-12-03
[#9337] improvement: Plan scan API supports access control

roryqi:

Plan scan API supports access control

Fix: #9337 

No need.

Iceberg client don't support remote plan scan now. Now we can't test it yet. You can see https://github.com/apache/iceberg/pull/13400


roryqi:
I have tested by hand for API. The access control takes effect.

Issue #9365, opened 2025-12-03
[Bug report] Two versions of httpclient existed in Gravitino Libs


jerryshao:
### Version

main branch

### Describe what's wrong

There're two versions of httpclients in the Gravitino libs, we should only keep on and exclude another.

### Error message and/or stacktrace

run `./gradlew compileDistribution -x test`, check the folder `distribution/package/libs`, you will see two versions of httpclient:

```
-rw-r--r--@ 1 jerryshao  staff     839626 Dec  3 14:51 httpclient5-5.2.1.jar
-rw-r--r--@ 1 jerryshao  staff     911620 Dec  3 14:51 httpclient5-5.4.4.jar
-rw-r--r--@ 1 jerryshao  staff     852871 Dec  3 14:51 httpcore5-5.2.jar
-rw-r--r--@ 1 jerryshao  staff     908601 Dec  3 14:51 httpcore5-5.3.4.jar
-rw-r--r--@ 1 jerryshao  staff     236827 Dec  3 14:51 httpcore5-h2-5.2.jar
-rw-r--r--@ 1 jerryshao  staff     241692 Dec  3 14:51 httpcore5-h2-5.3.4.jar
```

We need to exclude one.

### How to reproduce

See above.

PR #9368, opened 2025-12-03
[#9365] fix(lance): Exclude lance related dependencies to reduce the size of the package

jerryshao:

Exclude various unnecessary lance related dependencies to reduce the size of the package.

1. For catalog-lakehosue-generic, exclude the runtime lance dependency as it is already included in the core.
2. For lance-common and lance-rest-server, exclude the unnecessary lance introduced packages to reduce the size.

Fix: #9365 

No.

Existing tests.


yuqi1129:
I have verified locally and it looks good to me.

yuqi1129 (dismissed):

jerryshao:
I still need to check if other dependencies like Arrow should also be excluded, hold on and let me explore more.

yuqi1129:
I have tested, and no duplicated or unexpected jars are available.  About arrow-related jars, since we are uncertain about its roles, it's acceptable to leave them as is.

yuqi1129 (approved):

PR #9382, opened 2025-12-04
[Minor] fix(scripts): Fix a minor problem in `release-util.sh`

yuqi1129:

This pull request makes a minor update to the `get_release_info` function in `dev/release/release-util.sh`, specifically changing how the `GPG_KEY` configuration is read. The change improves the logic for retrieving the GPG key.

* Updated the `get_release_info` function to use the existing `GPG_KEY` environment variable instead of deriving it from the `GIT_EMAIL` when reading the GPG key configuration.

It's a bug.

N/A

Test locally.

### DIFF
--- a/dev/release/release-util.sh
+++ b/dev/release/release-util.sh
@@ -171,7 +171,7 @@ function get_release_info {
   fi
 
   export GIT_EMAIL="$ASF_USERNAME@apache.org"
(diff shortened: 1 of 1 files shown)


jerryshao:
This is not a problem, your gpg key should be an apache mail address.

yuqi1129:
> This is not a problem, your gpg key should be an apache mail address.

I see.

yuqi1129:
It's not a problem.

Issue #9386, opened 2025-12-04
[Subtask] Add integration tests for the least implementation of relational table


tsungchih:
### Describe the subtask

We shall add integration tests for the least implementation of relational table.

- [x] fix `comment` in `ColumnDTO` to be `Optional`
- [x] fix serdes for `AddPartitionsRequest` to use `PartitionDTOSerdes`
- [x] fix method `encode_string` to correctly encode slash
- [x] add `TableCreateRequest` to facilitate integration tests
- [x] add `TableResponse` to facilitate integration tests
- [x] add integration tests for partition operations (add/get/list/drop partition and list partition names of a relational table)
- [x] add integration tests for table creation

### Parent issue

#8591

PR #9396, opened 2025-12-05
[#9386] feat(python-client): Add minimum viable integration tests for relational table

tsungchih:

We shall add integration tests for the least implementation of relational table.

This PR added the following classes.

- `TableCreateRequest`
- `TableResponse`
- integration tests of partition operations for relational table.

Fix: #9386 

No

Unit tests and integration tests

### DIFF
--- a/clients/client-python/tests/integration/test_relational_table.py
+++ b/clients/client-python/tests/integration/test_relational_table.py
@@ -32,7 +32,6 @@ from gravitino.dto.rel.column_dto import ColumnDTO
 from gravitino.dto.rel.partitioning.identity_partitioning_dto import (
     IdentityPartitioningDTO,
 )
-from gravitino.utils import HTTPClient
 from tests.integration.containers.hdfs_container import HDFSContainer
 from tests.integration.integration_test_env import IntegrationTestEnv
 
@@ -53,7 +52,6 @@ class TestRelationalTable(IntegrationTestEnv):
         cls.hdfs_container: HDFSContainer = HDFSContainer()
         hive_metastore_uri = f"thrift://{cls.hdfs_container.get_ip()}:9083"
         logger.info("Started Hive container with metastore URI: %s", hive_metastore_uri)
-        cls.rest_client = HTTPClient("http://localhost:8090")
         cls.gravitino_admin_client = GravitinoAdminClient(uri="http://localhost:8090")
         cls.gravitino_admin_client.create_metalake(
             cls.METALAKE_NAME,


tsungchih:

clients/client-python/gravitino/rest/rest_utils.py
```diff
@@ -16,11 +16,12 @@
...
+
 from gravitino.exceptions.base import IllegalArgumentException
 
 
 def encode_string(to_encode: str):
     if to_encode is None:
         raise IllegalArgumentException("Invalid string to encode: None")
 
-    return urllib.parse.quote(to_encode, encoding="utf-8")
+    return urllib.parse.quote(to_encode, safe="", encoding="utf-8")
```
jerryshao, line 27: What is the meaning of adding `safe=""`?

jerryshao:
@tsungchih I would suggest that we add `RelationalCatalog` first in Python client, then we can do the end-to-end integration tests by using the Python SDK (no need to use the rest client).

clients/client-python/gravitino/rest/rest_utils.py
tsungchih: The safe parameter in urllib.parse.quote specifies characters that should not be percent-encoded (URL-encoded). The default value of safe is "/" which means the slash will not be percent-encoded.

tsungchih:
> @tsungchih I would suggest that we add `RelationalCatalog` first in Python client, then we can do the end-to-end integration tests by using the Python SDK (no need to use the rest client).

@jerryshao Do you mean we shall add RelationalCatalog first and then get back to create end-to-end integration tests? In doing so, I think the integration tests shall be created for RelationalCatalog rather than RelationalTable.

jerryshao:
> > @tsungchih I would suggest that we add `RelationalCatalog` first in Python client, then we can do the end-to-end integration tests by using the Python SDK (no need to use the rest client).
> 
> @jerryshao Do you mean we shall add RelationalCatalog first and then get back to create end-to-end integration tests? In doing so, I think the integration tests shall be created for RelationalCatalog rather than RelationalTable.

Yes, that's what I mean. Also, testing the interface of `RelationalCatalog` can also cover the `RelationalTable`, and we can keep this test code for later use when `RelationalCatalog` is ready.

tsungchih:
> > > @tsungchih I would suggest that we add `RelationalCatalog` first in Python client, then we can do the end-to-end integration tests by using the Python SDK (no need to use the rest client).
> 
> > 
> 
> > @jerryshao Do you mean we shall add RelationalCatalog first and then get back to create end-to-end integration tests? In doing so, I think the integration tests shall be created for RelationalCatalog rather than RelationalTable.
> 
> 
> 
> Yes, that's what I mean. Also, testing the interface of `RelationalCatalog` can also cover the `RelationalTable`, and we can keep this test code for later use when `RelationalCatalog` is ready.

Got it. I am now working on relational catalog and will get back to revise the integration tests with respect to the relational catalog after then.

tsungchih:
@unknowntpo @jerryshao I've revised the integration tests for the relational table. Could you please take a review on it? Thanks!

jerryshao:
I have no further comment, can you please fix the minor issue mentioned above?

clients/client-python/tests/integration/test_relational_table.py
```diff
@@ -0,0 +1,152 @@
...
+    TABLE_NAME: str = "test_table"
+    TABLE_IDENT: NameIdentifier = NameIdentifier.of(SCHEMA_NAME, TABLE_NAME)
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.hdfs_container: HDFSContainer = HDFSContainer()
+        hive_metastore_uri = f"thrift://{cls.hdfs_container.get_ip()}:9083"
+        logger.info("Started Hive container with metastore URI: %s", hive_metastore_uri)
+        cls.rest_client = HTTPClient("http://localhost:8090")
```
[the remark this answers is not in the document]
tsungchih: Fixed in the commit (https://github.com/apache/gravitino/pull/9396/commits/6a6b913288c3a300b35161e7ea7ae363fba347ea)

tsungchih:
> I have no further comment, can you please fix the minor issue mentioned above?

@jerryshao Sure! The issues mentioned by the Copilot have been fixed.

PR #9397, opened 2025-12-05
[#9307] improvement(server): a missing partition will cause sever error

SalmaneKhalili:

Introduced a null check in the addPartitions resource method within PartitionOperations.java, It throws IllegalArgumentException, expanded the try scope to cover the check as well.
A new unit test, testAddPartitionWithNullPartitions(), for fix validation.

Currently, if the request or its partition are null, the result is an NPE, it bypassed the intended error mapping logic and was treated as an unexpected server error, resulting in an 500 Internal Server Error response.

Fix: #9307

No.

Added testAddPartitionWithNullPartitions() to test suit and executed these:
 ./gradlew :common:compileJava
./gradlew :common:test 
./gradlew :server:test
All builds/tests were successful.


jerryshao:
@LongerDude Can you please fix the broken PR title?

SalmaneKhalili:
That was my first pr ever, thank you guys.

PR #9402, opened 2025-12-08
[MINOR] polish(authz): polish authz codes for easier reuse

mchades:

can specify a principal

polish authz codes for easier reuse

no

CI passed


roryqi:
Could add a  comment for this? For example, it will be convenient for the test code.

mchades:
> Could add a comment for this? For example, it will be convenient for the test code.

added

jerqi (approved):

Issue #9404, opened 2025-12-08
[Improvement] Add Lance related license and notice file


jerryshao:
### What would you like to be improved?

Some Lance-related license and notice files are missing. We need to add them when we do the release:

1. Update `LICENSE.bin` to include Lance and Lance dependencies.
2. Add `LICENSE.lance` and `NOTICE.lance` for Lance REST-specific license files.
3. Add the above two license files to the lance rest package when doing the package (follow the IRC).

PR #9407, opened 2025-12-08
[#9404] Improvement(License): Add Lance related license and notice files

jerryshao:

This PR addresses issue #9404 by adding Lance-specific license and notice files for the Lance REST server distribution.

When doing Apache releases, we need proper LICENSE and NOTICE files for all distributed components. The Lance REST server was missing its dedicated license files and some Lance-specific dependencies were not documented in LICENSE.bin and NOTICE.bin.

Fix: #9404 

No.

- Built Lance REST server: `./gradlew compileLanceRESTServer` - SUCCESS
- Verified distribution package contains correct LICENSE and NOTICE files
- All 98 external dependencies verified and documented
- License accuracy verified from authoritative sources
- Apache Software Foundation compliance requirements met

### DIFF
--- a/LICENSE.lance
+++ b/LICENSE.lance
@@ -263,26 +263,17 @@
    This product bundles various third-party components also under the
    BSD license
 
-   ANTLR4
-   ASM
-   Javolution
    JSR305
-   Protobuf
-   Stax2 API
 
    This product bundles various third-party components also under the
    Common Development and Distribution License 1.0
 
    Javax Activation
-   Javax Mail
-   Stax API
    Java Servlet API
-   JSR311 API
 
    This product bundles various third-party components also under the
    Common Development and Distribution License 1.1
 
-   JSR 374 (JSON Processing) Default Provider
    JavaServlet
    Eclipse Jersey
 
@@ -335,5 +326,3 @@
    the public domain.
 
    AOP Alliance
-
-
--- a/NOTICE.lance
+++ b/NOTICE.lance
@@ -290,4 +290,3 @@ SoundCloud Ltd. (http://soundcloud.com/).
 
 This product includes software developed as part of the
 Ocelli project by Netflix Inc. (https://github.com/Netflix/ocelli/).
-


yuqi1129:
What does the suffix `.rest` mean in `LICENSE.rest`? If it stands for the license file of Iceberg REST, I believe we'd better rename `LICENSE.rest` to `LICENSE.iceberg`

jerryshao:
> What does the suffix `.rest` mean in `LICENSE.rest`? If it stands for the license file of Iceberg REST, I believe we'd better rename `LICENSE.rest` to `LICENSE.iceberg`

Ok, but it isn't so necessary.

LICENSE.lance
```diff
@@ -0,0 +1,339 @@
...
+   Jakarta Annotations API
+   Jakarta Inject
+   Jakarta Validation API
+   Jakarta RESTful WS API
+   Jakarta XML Binding API
+
+   This product bundles various third-party components also under the
+   Mozilla Public License
+
+   Javassist
```
yuqi1129, line 323: I could not find the direct usage of `javaassit`, is it introduced by `aopliance-repackaged-2.6.1.jar`?

LICENSE.lance
```diff
@@ -0,0 +1,339 @@
...
+   This product bundles various third-party components also under the
+   MIT license
+
+   Checker Framework
+   SLF4J API Module
+
+   This product bundles various third-party components also under the
+   BSD license
+
+   ANTLR4
```
yuqi1129: The same as L332

LICENSE.lance
jerryshao: This is introduced by Jersey. This is not a direct dependency, it is transiently introduced. The basic principal is to list all the required runtime jars to the license file.

yuqi1129: Got it.

LICENSE.lance
jerryshao: This seems is not necessary, I will fix it.

PR #9431, opened 2025-12-09
[MINOR] improvement: fix typos of guava imports

fwonce:

Some `Sets` in the project are pointed to the shadowed one in Glassfish, rather than original Guava.

1. Search through the project to replace all `import org.glassfish.jersey.internal.guava.Xxss` to its corresponding `com.google.common.collect.Xxxs`.
2. Add 3 `replaceRegex` actions to Spotless plugin, to ensure no similar issues in the future.

For sure the one from original Guava should be preferred, because shadowed classes are normally used for internal purposes of the party who package them, and are not intended to expose as a contract.

Fix: No functionality issue involved. 

No

Unit tests all passed. Plus, the implementations of two classes are identical, so in fact as long as it compiles right, we are good to go I believe.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -129,6 +129,21 @@ allprojects {
           "import\\s+(?:static\\s+)?[^*\\s]+\\*;(\\r\\n|\\r|\\n)",
           "$1"
         )
+        replaceRegex(
+          "Use Guava Sets instead of shadowed Glassfish Sets",
+          "import\\s+org\\.glassfish\\.jersey\\.internal\\.guava\\.Sets;",
+          "import com.google.common.collect.Sets;"
+        )
+        replaceRegex(
+          "Use Guava Maps instead of shadowed Glassfish Maps",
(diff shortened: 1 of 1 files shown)


roryqi:
Could you add some rules in the file `build.gradle.kts` in the line 132 to avoid similar issues?
```
        replaceRegex(
          "Use Guava Sets instead of shadowed Glassfish Sets",
          "import\\s+org\\.glassfish\\.jersey\\.internal\\.guava\\.Sets;",
          "import com.google.common.collect.Sets;"
        )
        replaceRegex(
          "Use Guava Maps instead of shadowed Glassfish Maps",
          "import\\s+org\\.glassfish\\.jersey\\.internal\\.guava\\.Maps;",
          "import com.google.common.collect.Maps;"
        )
        replaceRegex(
          "Use Guava Lists instead of shadowed Glassfish Lists",
          "import\\s+org\\.glassfish\\.jersey\\.internal\\.guava\\.Lists;",
          "import com.google.common.collect.Lists;"
        )
```

Issue #9448, opened 2025-12-10
[Improvement] Lance Rest Server: precheck the invalid location


danhuawang:
### What would you like to be improved?

URL: http://192.168.194.170:9101/lance/v1/table/lance_s3_catalog%24lance_s3_schema_t8%24lance_s3_table_8/create?mode=create

Status code: 200

Content Type: null

Request Headers
Accept=application/json, application/javascript, text/javascript, text/json
Authorization=Basic YW5vbnltb3VzOnRlc3Q=
x-lance-table-location=invalid-s3-path/lance_s3_table_8
x-lance-table-properties={"lance.storage.aws_access_key_id":"xxx","lance.storage.aws_secret_access_key":"xxxx","lance.storage.aws_region":"ap-northeast-1"}
Content-Type=application/vnd.apache.arrow.stream; charset=UTF-8
Content Body
[...]

PR #9450, opened 2025-12-10
[#9448] improvement(catalogs):  Validate table location and add unit tests for location checks

yuqi1129:
This pull request improves validation for table locations in the Lance catalog implementation. The main change is the introduction of a stricter check to ensure table locations are valid URIs or absolute file paths, preventing invalid or ambiguous locations from being used. Comprehensive unit and integration tests are also added to verify this behavior.

**Validation improvements:**

* Added a new `isValidLanceLocation` method in `LanceTableOperations` to validate that table locations are either valid URIs with a scheme or absolute file paths. This method is now used in table creation to enforce correct location formats. [[1]](diffhunk://#diff-0f917d044cc444092f273c9619772077d88a905b4543f9587a57b6c74d1bb0acR31-R40) [[2]](diffhunk://#diff-0f917d044cc444092f273c9619772077d88a905b4543f9587a57b6c74d1bb0acL110-R116) [[3]](diffhunk://#diff-0f917d044cc444092f273c9619772077d88a905b4543f9587a57b6c74d1bb0acR349-R378)

**Testing enhancements:**

* Added parameterized unit tests in `TestLanceTableOperations` to cover various valid and invalid location formats, ensuring the new validation logic works as intended. [[1]](diffhunk://#diff-00a450bde65f0cdc55b48e1f5e698558f9841516ee3ed98857115676702ca5d1R27) [[2]](diffhunk://#diff-00a450bde65f0cdc55b48e1f5e698558f9841516ee3ed98857115676702ca5d1R42-R44) [[3]](diffhunk://#diff-00a450bde65f0cdc55b48e1f5e698558f9841516ee3ed98857115676702ca5d1R96-R113)
* Updated integration test `LanceRESTServiceIT` to assert that registering a table with an invalid location fails with the expected error message.

Better user experience.

Fix: #9448

N/A.

UTs and ITs

### DIFF
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
@@ -28,12 +28,16 @@ import com.lancedb.lance.index.DistanceType;
 import com.lancedb.lance.index.IndexParams;
 import com.lancedb.lance.index.IndexType;
 import com.lancedb.lance.index.vector.VectorIndexParams;
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import org.apache.arrow.memory.RootAllocator;
+import org.apache.arrow.util.VisibleForTesting;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.EntityStore;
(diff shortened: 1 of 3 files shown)


catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -342,4 +346,34 @@ private IndexParams getIndexParamsByIndexType(IndexType indexType) {
...
+
+    try {
+      URI uri = new URI(location);
+      String scheme = uri.getScheme();
+      if (StringUtils.isNotBlank(scheme)) {
+        return true;
+      }
+
+      File file = new File(location);
+      return file.isAbsolute();
```
jerryshao, line 374: My feeling is that Lance will throw an exception if the path is illegal, we don't have to check here in Gravitino.

jerryshao: Besides, why do we need to ensure that the path is absolute? My understanding is that relative path should also work (FS will figure out the working directory and how to normalize the relative path), do we need such a strict restriction?

yuqi1129: It is used in registering a table.

yuqi1129: > Besides, why do we need to ensure that the path is absolute? My understanding is that relative path should also work (FS will figure out the working directory and how to normalize the relative path), do we need such a strict restriction?

You are right, the whole check is not so necessary in the beginning. I also have the same consideration,  let hold it until the 1.1.0 is released.

yuqi1129:
Close it as we may need to discuss in detail.

Issue #9462, opened 2025-12-11
[Bug report] Fail to drop a register Lance table


yuqi1129:
### Version

main branch

### Describe what's wrong

Failed to drop Lance dataset for table test.generic_lakehouse_lance_catalog.sales.customers1

### Error message and/or stacktrace

2025-12-11 14:43:04.898 ERROR [Gravitino-webserver-90] [org.apache.gravitino.server.web.rest.ExceptionHandlers$BaseExceptionHandler.handle(ExceptionHandlers.java:1004)] - Failed to operate object [customers1] operation [DROP] under [sales], reason [Failed to drop Lance dataset for table test.generic_lakehouse_lance_catalog.sales.customers1]
java.lang.RuntimeException: Failed to drop Lance dataset for table test.generic_lakehouse_lance_catalog.sales.customers1
        at org.apache.gravitino.catalog.lakehouse.lance.LanceTableOperations.dropTable(LanceTableOperations.java:243) ~[?:?]
        at org.apache.gravitino.catalog.lakehouse.generic.GenericCatalogOperations.dropTable(GenericCatalogOperations.java:282) ~[?:?]
        at org.apache.gravitino.catalog.TableOperationDispatcher.lambda$dropTable$11(TableOperationDispatcher.java:332) ~[gravitino-core-1.1.0-SNAPSHOT.jar:?]
        at org.apache.gravitino.catalog.CatalogManager$CatalogWrapper.lambda$doWithTableOps$1(CatalogManager.java:160) ~[gravitino-core-1.1.0-SNAPSHOT.jar:?]
        at org.apache.gravitino.utils.IsolatedClassLoader.withClassLoader(IsolatedClassLoader.java:86) ~[gravitino-core-1.1.0-SNAPSHOT.jar:?]
        at org.apache.gravitino.catalog.CatalogManager$CatalogWrapper.doWithTableOps(CatalogManager.java:155) ~[gravitino-core-1.1.0-SNAPSHOT.jar:?]
        at org.apache.gravitino.catalog.TableOperationDispatcher.lambda$dropTable$12(TableOperationDispatcher.java:332) ~[gravitino-core-1.1.0-SNAPSHOT.jar:?]
        at org.apache.gravitino.catalog.OperationDispatcher.doWithCatalog(OperationDispatcher.java:99) ~[gravitino-core-1.1.0-SNAPSHOT.jar:?]
[...]

PR #9463, opened 2025-12-11
[#9462] fix(lakehouse-generic): Ensure external table property is set during table registration

yuqi1129:

If `lance.register` is true, we need to make sure that `external` is true.

It's a design

Fix: #9462 

N/A

ITs

### DIFF
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
@@ -22,6 +22,7 @@ import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_CREAT
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_REGISTER;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Maps;
 import com.lancedb.lance.Dataset;
 import com.lancedb.lance.WriteParams;
 import com.lancedb.lance.index.DistanceType;
@@ -259,8 +260,15 @@ public class LanceTableOperations extends ManagedTableOperations {
       throws NoSuchSchemaException, TableAlreadyExistsException {
 
     if (register) {
+      // If this is a register operation, it should be a external table.
+      Map<String, String> stringStringMap = Maps.newHashMap(properties);
+      if (!properties.containsKey(Table.PROPERTY_EXTERNAL)
+          || !Boolean.parseBoolean(properties.get(Table.PROPERTY_EXTERNAL))) {
+        stringStringMap.put(Table.PROPERTY_EXTERNAL, "true");
+      }
+
       return super.createTable(
-          ident, columns, comment, properties, partitions, distribution, sortOrders, indexes);
+          ident, columns, comment, stringStringMap, partitions, distribution, sortOrders, indexes);
     }
 
     Map<String, String> storageProps = LancePropertiesUtils.getLanceStorageOptions(properties);
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
@@ -289,7 +289,7 @@ public class TableMetaService {
                 mapper ->
                     mapper.softDeletePolicyMetadataObjectRelsByTableId(
                         namespacedEntityId.entityId()));
-            SessionUtils.doWithCommit(
+            SessionUtils.doWithoutCommit(
                 TableVersionMapper.class,
                 mapper ->
                     mapper.softDeleteTableVersionByTableIdAndVersion(
--- a/catalogs/catalog-lakehouse-generic/src/test/java/org/apache/gravitino/catalog/lakehouse/lance/integration/test/CatalogGenericCatalogLanceIT.java
+++ b/catalogs/catalog-lakehouse-generic/src/test/java/org/apache/gravitino/catalog/lakehouse/lance/integration/test/CatalogGenericCatalogLanceIT.java
@@ -292,6 +292,34 @@ public class CatalogGenericCatalogLanceIT extends BaseIT {
   }
 
   @Test
+  void testRegisterAndDropTable() {
+    String tableName = GravitinoITUtils.genRandomName(TABLE_PREFIX);
+    NameIdentifier nameIdentifier = NameIdentifier.of(schemaName, tableName);
+    Map<String, String> properties = createProperties();
+
+    String tableLocation = tempDirectory + "/" + tableName;
+    properties.put("format", "lance");
+    properties.put("location", tableLocation);
+    properties.put("lance.register", "true");
+
+    catalog
+        .asTableCatalog()
+        .createTable(
+            nameIdentifier,
+            new Column[] {},
+            TABLE_COMMENT,
+            properties,
+            Transforms.EMPTY_TRANSFORM,
+            null,
+            null);
+
+    Table loadedTable = catalog.asTableCatalog().loadTable(nameIdentifier);
+    Assertions.assertEquals(tableName, loadedTable.name());
+
+    Assertions.assertDoesNotThrow(() -> catalog.asTableCatalog().dropTable(nameIdentifier));
+  }
+
+  @Test
   void testLanceTableFormat() {
     String tableName = GravitinoITUtils.genRandomName(TABLE_PREFIX);
     Column[] columns = createColumns();


catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -259,8 +260,15 @@ Table createTableInternal(
       throws NoSuchSchemaException, TableAlreadyExistsException {
 
     if (register) {
+      // If this is a register operation, it should be a external table.
+      Map<String, String> stringStringMap = Maps.newHashMap(properties);
```
jerryshao, line 264: Choose a better variable name.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -259,8 +260,15 @@ Table createTableInternal(
...
+      if (!properties.containsKey(Table.PROPERTY_EXTERNAL)
+          || !Boolean.parseBoolean(properties.get(Table.PROPERTY_EXTERNAL))) {
+        stringStringMap.put(Table.PROPERTY_EXTERNAL, "true");
+      }
```
jerryshao, line 268: Does Lance have such a restriction that a registered table should be an external table? If not, we should not make such an assumption.

yuqi1129: For the lance table, I just made this assumption, or 

https://github.com/apache/gravitino/blob/a47bd4125e240a3587c1278c3445cd80511aaea7/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java#L223-L237

When the code goes to L231, it will throw an exception. Or should we need to forbid registering tables through Gravitino API directly?  I could not find a better way to handle it.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -259,8 +260,15 @@ Table createTableInternal(
       throws NoSuchSchemaException, TableAlreadyExistsException {
 
     if (register) {
+      // If this is a register operation, it should be a external table.
+      Map<String, String> stringStringMap = Maps.newHashMap(properties);
+      if (!properties.containsKey(Table.PROPERTY_EXTERNAL)
```
roryqi, line 265: I feel we should throw exception if we set external property to false. If user don't set the property, we can set the property to true.

core/src/main/java/org/apache/gravitino/storage/relational/service/TableMetaService.java
```diff
@@ -289,7 +289,7 @@ public boolean deleteTable(NameIdentifier identifier) {
                 mapper ->
                     mapper.softDeletePolicyMetadataObjectRelsByTableId(
                         namespacedEntityId.entityId()));
-            SessionUtils.doWithCommit(
+            SessionUtils.doWithoutCommit(
```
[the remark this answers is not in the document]
yuqi1129: It's a bug and I fixed it by the way.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: This is also a reasonable resolution, anyway, we need to handle it. cc @jerryshao , do you have any thoughts on it?

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
jerryshao: What is the exception?

You should not add the assumption if the Lance spec doesn't have such a restriction. If you want to add this restriction, can you show me Lance register definition?

yuqi1129: > What is the exception?

The following code will execute
```java
   Dataset.drop(location, LancePropertiesUtils.getLanceStorageOptions(table.properties())); 
```
And the exception will be like 

```text
Caused by: java.io.IOException: Not found: data/sales/orders, /Users/runner/work/lance/lance/rust/lance-io/src/local.rs:45:23
        at com.lancedb.lance.Dataset.drop(Native Method) ~[lance-core-0.39.0.jar:?]
        at org.apache.gravitino.catalog.lakehouse.lance.LanceTableOperations.dropTable(LanceTableOperations.java:234) ~[?:?]
        ... 89 more
```

> You should not add the assumption if the Lance spec doesn't have such a restriction. If you want to add this restriction, can you show me the Lance register definition?

There are no docs regarding properties `external`,  it only says that registering table is `Register an existing table at a given storage location as id.

More, please refer to https://editor.swagger.io/?url=https://raw.githubusercontent.com/lance-format/lance-namespace/refs/heads/main/docs/src/rest.yaml

jerryshao: What's the reason, is it because the location not exist or other reasons?

yuqi1129: Yeah, the location does not exist. So
- We do check the existence of the location when registering, do we need to do it? 
- Do we need to remove the data directory when the user does not set `external = true` (default: false)? or just add some descrition in the document and let user decide whether to set the property `external`.
- What the default value of `external` when registering a table? It seems that `true` would be a better one compared to the current value `false`.

jerryshao: "register" and "external" are two different behavior definitions; we should not use one definition to restrict another. The user can register an existing table and let Gravitino manage the metadata and data.

Current "register" is a tricky code path only for Lance table, we haven't received enough responses, and we're not clear how other formats define such behaviors. We'd better leave as it is. We can always add this restriction later on.

Another thing is that we should carefully think about such issues (like location not found issue here, illegal location in another issue), is this really an issue we should fix in the Gravitino side? How do we define our border?

jerryshao: @mchades what is your suggestion?

mchades: I remember JackYe said that the expectation for `registerTable` is that the location must exist, so I think when using `register`, if the location does not exist, we can directly throw an exception to reject the register (in this case, the user should be using `createEmptyTable`).


> "register" and "external" are two different behavior definitions; we should not use one definition to restrict another. The user can register an existing table and let Gravitino manage the metadata and data.

Agreed. 
The `external` is a table property introduced by Gravitino for Lance format table, used to differentiate the behavior during drop table (whether to delete storage data). It should not be linked with the `register` property, and it should be up to the user to decide whether to register the table as `external` or `managed`. If the user does not specify, then the default value of `external` being `false` also makes sense, as we expect the table to be fully managed by Gravitino.

yuqi1129: So, the final problem is that: Dropping a registered table will also remove the location if we use the `false` as default value of `register`, there is no need to handle this case?

jerryshao: Whether to drop data is controlled by `external`, why do you mix the behaior of `register` with `external`? Why do you assume that the registered table data cannot be dropped?

yuqi1129:
Close it as 
1.  When droping a `registered` table in Lance REST, it will *never* remove the data
2. Whether to remove the data in the Gravitino server is determined by the `external` properties, the syntax is different from that in Lance REST,  another points is that registeration operation is a work around to support Lance REST service, we'd better let it is when no more feedbacks are available.

Issue #9078, opened 2025-11-12
[Improvement] In SchemaMetaService.java block non-cascading schema deletions when topics still exist


justinmclean:
### What would you like to be improved?

Block non-cascading schema deletions when topics still exist by checking for topic metadata and raising an NonEmptyEntityException.

Here is a unit test to help:
```
@Test
  public void testDeleteSchemaFailsWhenTopicExists() throws IOException {
    String metalakeName = "metalake";
    AuditInfo auditInfo =
        AuditInfo.builder().withCreator("creator").withCreateTime(Instant.now()).build();

    BaseMetalake metalake =
        createBaseMakeLake(RandomIdGenerator.INSTANCE.nextId(), metalakeName, auditInfo);
    backend.insert(metalake, false);

    CatalogEntity catalog =
        createCatalog(
            RandomIdGenerator.INSTANCE.nextId(), Namespace.of(metalakeName), "catalog", auditInfo);
    backend.insert(catalog, false);

    SchemaEntity schema =
        createSchemaEntity(
            RandomIdGenerator.INSTANCE.nextId(),
            Namespace.of(metalakeName, "catalog"),
            "schema",
            auditInfo);
    backend.insert(schema, false);

    TopicEntity topic =
        createTopicEntity(
            RandomIdGenerator.INSTANCE.nextId(),
            Namespace.of(metalakeName, "catalog", "schema"),
            "topic",
            auditInfo);
    backend.insert(topic, false);

    Assertions.assertThrows(
        NonEmptyEntityException.class,
        () -> SchemaMetaService.getInstance().deleteSchema(schema.nameIdentifier(), false));
  }
```

keepConcentration:
Hi, @justinmclean

Can I be assigned to this issue?

SalmaneKhalili:
Hello, sorry to bother you again, please assign this to me if it is still available.

PR #9465, opened 2025-12-11
[#9078] Improvement(core):In SchemaMetaService.java block non-cascading schema deletions when topics still exist

SalmaneKhalili:

This PR implements a check to prevent the deletion of non-cascading schemas that still contain topic metadata.

Currently, schema deletion checks for dependant tables, filesets and models. If the schema contains topics, the deletion proceeds regardless. Leaving orphaned topic metadata in the database.
This fix ensures a NonEmptyEntityException is thrown if any active topics are found within the schema. This matches previous checks for other subschema entities.

The fix was implemeted through a check in the non cascading branch of deleteSchema method in SchemaMetaService.java.

Fix: #9078

Nothing more than an error message.

## How was this patch tested?
Added testDeleteSchemaNonCascadingFailsWhenTopicExists, to TestSchemaMetaService.java
./gradlew :core:test

All tests passed.

[#9078] Improvement(core):In SchemaMetaService.java block non-cascading schema deletions when topics still exist

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
@@ -37,6 +37,7 @@ import org.apache.gravitino.meta.ModelEntity;
 import org.apache.gravitino.meta.NamespacedEntityId;
 import org.apache.gravitino.meta.SchemaEntity;
 import org.apache.gravitino.meta.TableEntity;
+import org.apache.gravitino.meta.TopicEntity;
 import org.apache.gravitino.metrics.Monitored;
 import org.apache.gravitino.storage.relational.helper.SchemaIds;
 import org.apache.gravitino.storage.relational.mapper.FilesetMetaMapper;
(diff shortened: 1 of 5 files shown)


SalmaneKhalili:
Id be appreciative of all feedback, ive been reading docs here and there. please let me know if there is something in particular i should look into. thanks.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -327,6 +327,12 @@ public boolean deleteSchema(NameIdentifier identifier, boolean cascade) {
             "Entity %s has sub-entities, you should remove sub-entities first", identifier);
       }
 
+      if (TopicMetaService.getInstance().topicsExistInSchema(identifier)) {
```
mchades: I recommend you keep the same style as above, using `listTopicsByNamespace` without adding a new `topicsExistInSchema` method.

SalmaneKhalili: Okay. Will do!

SalmaneKhalili: @mchades Done! 
In the previous version i tried to aim for a more optimized approach, for future prs ill try to just follow code style.
thanks for the feedback.

mchades (commented):
LGTM. 
 Plz update the PR description

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
SalmaneKhalili: Changed description!

mchades (approved):
Thanks for your contributions!

PR #9466, opened 2025-12-11
[Minor]: fix(core): change session handling to doWithoutCommit for soft delete operations

yuqi1129:

Replace `doWithCommit` with `doWithoutCommit` in `TableMetaService#deleteTable`

We should not use `doWithCommit` within `doMultipleWithCommit`, or it can't be rollbacked when someting unexpected happens.

N/A.

Existing tests.


roryqi:
I'm not sure whether we can add rules in the spotless plugin. Maybe we can add let copilot to try it.

yuqi1129:
> I'm not sure whether we can add rules in the spotless plugin. Maybe we can add let copilot to try it.

We can give it a try.

jerryshao (approved):

roryqi:
@yuqi1129 Actually, this is a fix. We shouldn't use a minor.

Issue #9474, opened 2025-12-12
[Improvement] Remove the number of calls to function `checkCatalogInUse`


yuqi1129:
### What would you like to be improved?

Method `doWithCatalog` will call `checkCatalogInUse` to check the status of the catalog, it will call `checkCatalogInUse` several times within a single API; we'd better try to reduce the number of calling it.

PR #9475, opened 2025-12-12
[#9474] improvement(server): Reduce the number of calling `catalogInuse`

yuqi1129:

User proxy to reduce the number of calling `catalogInuse`.

To improve API performance.

Fix: #9474 

N/A.

Existing test.


jerryshao:
@mchades can you please help to review this PR?

yuqi1129:
> How many times can this change reduce the number of `catalogInuse` calls per request?

It depends, some API only call `doWithCatalog`  one time, and some will also call `validateAlterProperties` and `getHiddenPropertyNames`, which will introduce an extra two times.  

For `loadTable`, it will at least reduce the number of calls by two.

core/src/main/java/org/apache/gravitino/catalog/OperationDispatcherInterceptor.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Override
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    if (args != null && args.length > 0) {
+      NameIdentifier catalogIdent = null;
+      if (args[0] instanceof NameIdentifier ident) {
+        catalogIdent = NameIdentifierUtil.getCatalogIdentifier(ident);
+      } else if (args[0] instanceof Namespace ns) {
+        if (ns.length() >= 2) {
+          catalogIdent = NameIdentifier.of(ns.level(0), ns.level(1));
```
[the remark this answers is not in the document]
yuqi1129: If the length of ident is less than 2, the value of `catalogIdent` is always null and nothing should be done.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -537,14 +539,28 @@ private void initGravitinoServerComponents() {
 
     SchemaOperationDispatcher schemaOperationDispatcher =
         new SchemaOperationDispatcher(catalogManager, entityStore, idGenerator);
-    SchemaHookDispatcher schemaHookDispatcher = new SchemaHookDispatcher(schemaOperationDispatcher);
+    SchemaDispatcher schemaDispatcherProxy =
+        (SchemaDispatcher)
+            Proxy.newProxyInstance(
+                SchemaDispatcher.class.getClassLoader(),
+                new Class[] {SchemaDispatcher.class},
+                new OperationDispatcherInterceptor(
```
jerryshao, line 547: I'm not so familar with this interceptor mechanism, how does it reduce the number of calling?

yuqi1129: For those dispatchers, any methods in them will call `doWithCatalog`, `validateAlterProperties` or `getHiddenPropertyNames` at least two times, which will then call `catalogInUse` the same times. After this change, we will only call it one time.

jerryshao: My question is why using reflection/proxy can reduce the number of calls.

yuqi1129: Please note the following change:
image

I have removed `catalogInUse` from `doWithCatalog`, so any endpoint in the dispatcher will now call and only call `catalogInUse` once with these interceptors. 

Before, take loading a table for example, it will 
image

After:
As I removed all `catalogInUse` from `doWithCatalog`, there will be only one in the interceptor and none in the `loadTable`.

jerryshao: So, when will this interceptor implementation (invoke) be called, every `operationDispatcher` call?

yuqi1129: Currently, only `SchemaDispatcher`, `TableDispatcher`, `PartitionDispatcher`, `FilesetDispatcher`, `TopicDispatcher`, and `ModelDispatcher` will trigger this interceptor.

jerryshao: I think You don't understand what I'm asking.

When will this `public Object invoke(Object proxy, Method method, Object[] args) throws Throwable` be invoked? will this be invoked in each `OperationDispatcher` call, like `loadTable`?

yuqi1129: > When will this public Object invoke(Object proxy, Method method, Object[] args) throws Throwable be invoked?

When any interface method of `OperationDispatcher` is called, the interceptor method will be triggered, like `loadTable`.

PR #9502, opened 2025-12-17
[#9490] improvement(core): Include lakehouse-generic catalogs in managed entities for proper drop behavior

### LINKED ISSUE #9490: [Improvement] Improve and clarify the behavior of `drop catalog` for managed catalog

We have several managed catalogs, like the generic lakehouse catalog, the fileset catalog, and the model catalog, for which we manage the metadata in Gravitino. For such a catalog, the current `drop` semantics seem not well supported, the drop behavior only deletes the metadata, not the data. 

For example, the drop of the generic lakehouse catalog will only delete the metadata, not the actual lance table managed by Gravitino. Theoretically, we should also delete the physical data for managed Lance tables.

### How should we improve?

_No response_

jerryshao:

This PR improves the drop behavior for lakehouse-generic catalogs to properly handle managed entities (Lance tables).

**Changes:**
- Modified `CatalogManager.includeManagedEntities()` to include RELATIONAL catalogs with provider 'lakehouse-generic'
- Added comprehensive integration test to verify drop behavior for both managed and external tables

Fix #9490 

Currently, when dropping a lakehouse-generic catalog, only metadata is deleted but the physical data (Lance tables managed by Gravitino) remains. This PR ensures that:
- When dropping a managed catalog, both metadata and physical data are properly cleaned up
- Managed tables' physical directories are deleted when catalog is dropped
- External tables' physical directories are preserved (not deleted)

Yes. After this change, dropping a lakehouse-generic catalog with `force=true` will now also delete the physical data for managed Lance tables, not just the metadata.

- Added integration test `testDropCatalogWithManagedAndExternalEntities()` that verifies:
  - Managed tables are properly deleted (both metadata and physical data)
  - External tables' physical data is preserved
- Test passed successfully in multiple runs

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
@@ -823,13 +823,13 @@ public class CatalogManager implements CatalogDispatcher, Closeable {
 
     if (schemaEntities.size() == 1) {
       String provider = catalogWrapper.catalog().provider();
-      if ("kafka".equals(provider)) {
+      if ("kafka".equalsIgnoreCase(provider)) {
         return false;
-      } else if ("jdbc-postgresql".equals(provider)) {
+      } else if ("jdbc-postgresql".equalsIgnoreCase(provider)) {
         // PostgreSQL catalog includes the "public" schema, see
         // https://github.com/apache/gravitino/issues/2314
         return !schemaEntities.get(0).name().equals("public");
-      } else if ("hive".equals(provider)) {
+      } else if ("hive".equalsIgnoreCase(provider)) {
         return !schemaEntities.get(0).name().equals("default");
       }
     }


mchades (approved):
LGTM.

you may respond to the copilot review

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -849,7 +850,9 @@ private boolean containsUserCreatedSchemas(
   }
 
   private boolean includeManagedEntities(CatalogEntity catalogEntity) {
-    return catalogEntity.getType().equals(FILESET);
+    return catalogEntity.getType().equals(FILESET)
```
roryqi: Should it become a property interface instead of conditions?

jerryshao: Let me think more about this.

core/src/main/java/org/apache/gravitino/meta/CatalogEntity.java
```diff
@@ -155,6 +156,12 @@ public CatalogInfo toCatalogInfoWithResolvedProps(Map<String, String> resolvedPr
         id, name, type, provider, comment, resolvedProperties, auditInfo, namespace);
   }
 
+  @Override
+  public boolean managed() {
+    return type == Catalog.Type.FILESET
+        || (type == Catalog.Type.RELATIONAL && provider.equalsIgnoreCase("lakehouse-generic"));
```
roryqi: I think we should add this method in the 
```
public interface CatalogProvider {
}
```
Entity is the common concept. So if we add the interface here, we can't get rid of the conditions.

roryqi: Maybe it's important to define managed definition. What's the `managed` property? What's the external? How to define them? Maybe we should have clear statement in the api module. It's better to have discussion in the dev mail list.

jerryshao: We cannot get the `CatalogProvider` object here in the caller method.

roryqi: Maybe wee can put the provider as the Catalog Entity property. It's not a proper design.

jerryshao: That's also weird and involves lots of changes. I don't want to introduce a heavy refactor just for this fix. If we want to optimize the interface, I would like to do this in a new PR.

roryqi: OK, I can push it forward.
After thinking twice, maybe it will be enough to add a property `CatalogEntity`.  But the value of `managed` property should be retrived from the catalog provider.
If you don't have time, I can help you optimize this point in the next PR. All is ok for me.

mchades: Can we directly use the capability interface here?
https://github.com/apache/gravitino/blob/7a83832a4732c523f44c2ca6772154224c54848e/core/src/main/java/org/apache/gravitino/connector/capability/Capability.java#L85-L93

jerryshao: Seems like another solution. I can take a try.

jerryshao: The problem is that our current interface is hard to get `Catalog` or `Capability` directly from `CatalogEntity`. We have to get the `Catalog` object, and the getting of `Catalog` object relies on loading / caching mechanism. If we want to fix this issue, we need to refactor our current mechanism.

roryqi: You have already acquired the catalogWrapper in the CatalogManager 769 line. You can use it directly. Could u correct me if I have something wrong.

roryqi: Another point is like what you mentioned that, we access the database too frequently.
image
Actually, we can get the catalog entity from the wrapper directly. We should add an internal interface to get the entity for our catalogs

roryqi:
Another point isn't related about this PR.
What's the meaning of `managed`. The definition isn't straightforward for me.
`Fileset` is a managed catalog. But the type of catalog could be `managed` and `external`.
It seems weird if a catalog is managed. But the `filesets` under the catalog may be not `managed`.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -738,28 +739,24 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
...
             }
 
-            Namespace schemaNamespace = Namespace.of(ident.namespace().level(0), ident.name());
-            CatalogWrapper catalogWrapper = loadCatalogAndWrap(ident);
-
+            Namespace schemaNs = Namespace.of(ident.namespace().level(0), ident.name());
             List<SchemaEntity> schemaEntities =
-                store.list(schemaNamespace, SchemaEntity.class, EntityType.SCHEMA);
-            CatalogEntity catalogEntity = store.get(ident, EntityType.CATALOG, CatalogEntity.class);
+                store.list(schemaNs, SchemaEntity.class, EntityType.SCHEMA);
```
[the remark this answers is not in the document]
jerryshao: I deliberately shortened the name to avoid the long variable name.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -738,28 +739,24 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
         metalakeIdent,
         LockType.WRITE,
         () -> {
-          checkMetalake(metalakeIdent, store);
           try {
             boolean catalogInUse = catalogInUse(store, ident);
             if (catalogInUse && !force) {
               throw new CatalogInUseException(
                   "Catalog %s is in use, please disable it first or use force option", ident);
             }
```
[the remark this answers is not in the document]
jerryshao: `catalogInUse` will both check the existence of metalake and catalog, and throw the `NoSuchMetalakeException` and `NoSuchCatalogException` accordingly. So the semantics are not changed.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogCapability.java
```diff
@@ -25,7 +25,8 @@
 public class FilesetCatalogCapability implements Capability {
   @Override
   public CapabilityResult managedStorage(Scope scope) {
-    if (Objects.requireNonNull(scope) == Scope.SCHEMA) {
+    if (Objects.requireNonNull(scope) == Scope.CATALOG
```
roryqi: Shape is good.  Two questions:

Why do we design this behavior?
`FilesetCapability` didn't support  scope `FILESET`.
`GeneralCapability` supported scope `TABLE`.

What's the difference for different scopes?  I can't understand the managed property clear. cc @mchades

jerryshao: It is a legacy problem. We can add `FILESET` support here, but there's no code leverage this scope currently.

roryqi: > It is a legacy problem. We can add `FILESET` support here, but there's no code leverage this scope currently.

Notice that `Fileset` has the property `managed`. Are they the same concept? 
What's relation between `ManagedStorage` and fileset properties `managed`? Will they conflicted? How to resolve them if conflicted?

jerryshao: No, they're two things, managed fileset is similar to managed table, you can refer to hive's managed table to understand more. `ManagedStorage` means the metadata is stored in Gravitino, Gravitino is responsible to the lifecycle of the metadata.

roryqi: One is the metadata concept. Another is the data concept. Isn't it correct? Maybe we could give a better name. But we can keep current name now. Just explore how to make this better, just mind communication.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -738,45 +739,41 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
         metalakeIdent,
         LockType.WRITE,
         () -> {
-          checkMetalake(metalakeIdent, store);
```
yuqi1129, line 741: This change has the same effect as https://github.com/apache/gravitino/pull/9563, as `catalogInUse` will also access and check the existence of Metalake. see:

https://github.com/apache/gravitino/blob/bd41384a6c1b92a96a6aefb6c54c0ffb748a2bff/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java#L867-L871

The difference is that `checkMetalake` will throw an exception if a metalake is not in use, and `catalogInUse` will return false.

@mchades please help to verify whether this breaks your initial design

jerryshao: You mean that it misses "metalake not in use exception" after my change? Let me check it. My initial feeling is that this part of code is quite messy and spread everywhere. It will be misused easily, we should figure out a better solution.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -893,6 +894,15 @@ private boolean getCatalogInUseValue(EntityStore store, NameIdentifier catalogId
     }
   }
 
+  private boolean isManagedCatalog(CatalogWrapper catalogWrapper) {
+    try {
+      return catalogWrapper.capabilities().managedStorage(Capability.Scope.CATALOG).supported();
+    } catch (Exception e) {
+      LOG.warn("Failed to get capabilities for catalog, assuming not managed", e);
+      return false;
```
yuqi1129: When will we get an exception here? Is it better to throw it?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -740,43 +741,40 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
         () -> {
           checkMetalake(metalakeIdent, store);
           try {
-            boolean catalogInUse = catalogInUse(store, ident);
+            boolean catalogInUse = getCatalogInUseValue(store, ident);
```
yuqi1129, line 744: This part is LGTM currently, and we will use another PR to resolve it.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogCapability.java
roryqi: Could u modify the comment of the method `managedStorage` tell developer that this is only used for metadata management instead of `entity` management? Entity isn't a clear concept. It can be data or metadata. We just want to say the metadata is fully managed.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
jerryshao: Seems this will not throw an exception, but the method signature has defined an exception. Let me fix it.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -893,6 +894,15 @@ private boolean getCatalogInUseValue(EntityStore store, NameIdentifier catalogId
     }
   }
 
+  private boolean isManagedCatalog(CatalogWrapper catalogWrapper) {
+    try {
+      return catalogWrapper.capabilities().managedStorage(Capability.Scope.CATALOG).supported();
```
roryqi: @mchades Is it ok here? What's the meaning of managed storage? If it means the scope of the metadata object is managed fully Gravitino. Does the catalog metadata is fully managed by Gravitino? It seems that all the catalog providers managed fully the catalog.

jerryshao:
I have some slight changes.

mchades:
I'm also reviewing and thinking of this refactor

jerqi (dismissed):

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogCapability.java
mchades: > ManagedStorage means the metadata is stored in Gravitino, Gravitino is responsible to the lifecycle of the metadata.

Yes, this is a correct explanation of the original intention of `ManagedStorage`. However, the scope here actually refers to the type of metadata. For example, when the scope is `schema`:
1. For the Hive Catalog, its schema is all `unsupported ManagedStorage`, because the schema of Hive also exists in HMS, and Gravitino cannot fully take charge of its metadata lifecycle (this also applies to relational catalogs other than lakehouse-generic).
2. For Fileset Catalog, Model Catalog, and Kafka Catalog, their schemas are created and stored by Gravitino, so their `managedStorage` method should return `supported`.

When the scope is catalog, the return value of `ManagedStorage` should all be supported, because all catalog entities are created and stored by Gravitino, which is why there was no catalog enumeration value in the scope initially.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -805,28 +803,33 @@ && containsUserCreatedSchemas(schemaEntities, catalogEntity, catalogWrapper)) {
...
+      // For managed storage catalog, any existing schema entities are considered user-created. At
+      // this point we already know schemaEntities is not empty, so we can return true directly
+      // without further checks.
+      return true;
+    }
+
     if (schemaEntities.size() == 1) {
-      if ("kafka".equals(catalogEntity.getProvider())) {
+      String provider = catalogWrapper.catalog().provider();
+      if ("kafka".equals(provider)) {
```
mchades: better to change to `equalsIgnoreCase`

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -805,28 +803,33 @@ && containsUserCreatedSchemas(schemaEntities, catalogEntity, catalogWrapper)) {
...
     if (schemaEntities.isEmpty()) {
       return false;
     }
 
+    if (isManagedStorageCatalog(catalogWrapper)) {
+      // For managed storage catalog, any existing schema entities are considered user-created. At
+      // this point we already know schemaEntities is not empty, so we can return true directly
+      // without further checks.
+      return true;
+    }
```
mchades, line 822: The short-circuit check should be moved after the below `if (schemaEntities.size() == 1)` block, otherwise, when a Kafka catalog with a single `default` schema is encountered, an incorrect result will be returned. This is because schemas in a Kafka catalog are all managed, but the `default` schema is pre-created by the system and not user-created.

jerryshao: Kafka is not a managed storage catalog. Because topic is not fully managed in Gravitino, it is an external data.

jerryshao: Please check how I implement the `managedStorageCatalog`.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -893,6 +894,19 @@ private boolean getCatalogInUseValue(EntityStore store, NameIdentifier catalogId
     }
   }
 
+  private boolean isManagedStorageCatalog(CatalogWrapper catalogWrapper) {
+    try {
+      Capability capability = catalogWrapper.capabilities();
+      return capability.managedStorage(Capability.Scope.SCHEMA).supported()
+          && (capability.managedStorage(Capability.Scope.TABLE).supported()
+              || capability.managedStorage(Capability.Scope.FILESET).supported()
+              || capability.managedStorage(Capability.Scope.MODEL).supported());
```
mchades, line 903: missing `Capability.Scope.TOPIC`?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
mchades: OK. The code was not yet changed when I commented, and this part is now LGTM.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
jerryshao: This is intentionally not supported, Kafka topic is not a managed entity in Gravitino, so we should not add here.

mchades: But we may support managed topics in other messaging catalogs in the future.

jerryshao: No, in a foreseeable future, we will not replace MQ and manage the topic in Gravitino.

mchades (approved):

Issue #9500, opened 2025-12-17
[FEATURE] Fileset generate proper privileged token


jerqi:
### Describe the feature

 Fileset generate proper privileged token according to the role information and user information

### Motivation

 Fileset generate proper privileged token according to the role information and user information

### Describe the solution

 Fileset generate proper privileged token according to the role information and user information

PR #9507, opened 2025-12-18
[#9500] fix(core): Fix the credential for the fileset with multiple locations

roryqi:

Fix the credential for the fileset with multiple locations

Fix: #9500

No need.

Existed tests.

### DIFF
--- a/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGravitinoFileSystemCredentialsProvider.java
+++ b/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGravitinoFileSystemCredentialsProvider.java
@@ -19,7 +19,12 @@
 
 package org.apache.gravitino.filesystem.hadoop;
 
+import static org.apache.gravitino.file.Fileset.PROPERTY_DEFAULT_LOCATION_NAME;
+import static org.apache.gravitino.filesystem.hadoop.GravitinoVirtualFileSystemConfiguration.FS_GRAVITINO_CURRENT_LOCATION_NAME;
+
+import java.util.Optional;
 import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.audit.CallerContext;
 import org.apache.gravitino.catalog.hadoop.fs.GravitinoFileSystemCredentialsProvider;
 import org.apache.gravitino.client.GravitinoClient;
 import org.apache.gravitino.credential.Credential;
@@ -51,10 +56,31 @@ public class DefaultGravitinoFileSystemCredentialsProvider
     // The format of name identifier is `metalake.catalog.schema.fileset`
     String nameIdentifier = configuration.get(GVFS_NAME_IDENTIFIER);
     String[] idents = nameIdentifier.split("\\.");
+
     try (GravitinoClient client = GravitinoVirtualFileSystemUtils.createClient(configuration)) {
       FilesetCatalog filesetCatalog = client.loadCatalog(idents[1]).asFilesetCatalog();
       Fileset fileset = filesetCatalog.loadFileset(NameIdentifier.of(idents[2], idents[3]));
+
+      String targetLocationName = getTargetLocation(fileset);
+      GravitinoVirtualFileSystemUtils.setCallerContextForGetCredentials(targetLocationName);
+
       return fileset.supportsCredentials().getCredentials();
+    } finally {
+      CallerContext.CallerContextHolder.remove();
     }
+  }
+
+  private String getTargetLocation(Fileset fileset) {
+    String currentLocationEnvVar =
+        configuration.get(
+            GravitinoVirtualFileSystemConfiguration.FS_GRAVITINO_CURRENT_LOCATION_NAME_ENV_VAR,
+            GravitinoVirtualFileSystemConfiguration
+                .FS_GRAVITINO_CURRENT_LOCATION_NAME_ENV_VAR_DEFAULT);
+    String locationName =
+        Optional.ofNullable(configuration.get(FS_GRAVITINO_CURRENT_LOCATION_NAME))
+            .orElse(System.getenv(currentLocationEnvVar));
+    return locationName == null
+        ? fileset.properties().get(PROPERTY_DEFAULT_LOCATION_NAME)
+        : locationName;
   }
 }
--- a/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/GravitinoVirtualFileSystemUtils.java
+++ b/clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/GravitinoVirtualFileSystemUtils.java
@@ -36,9 +36,11 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.audit.CallerContext;
 import org.apache.gravitino.client.DefaultOAuth2TokenProvider;
 import org.apache.gravitino.client.GravitinoClient;
 import org.apache.gravitino.client.KerberosTokenProvider;
+import org.apache.gravitino.credential.CredentialConstants;
 import org.apache.hadoop.conf.Configuration;
 
 /** Utility class for Gravitino Virtual File System. */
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/apache/gravitino/credential/S3SecretKeyCredential.java
```diff
@@ -29,6 +29,10 @@ public class S3SecretKeyCredential implements Credential {
 
   /** S3 secret key credential type. */
   public static final String S3_SECRET_KEY_CREDENTIAL_TYPE = "s3-secret-key";
+  /** S3 scheme. */
+  public static final String S3_SCHEME = "s3";
+  /** S3A scheme (Hadoop S3A connector). */
+  public static final String S3A_SCHEME = "s3a";
```
mchades: These variables are being redefined repeatedly. Could we move them to `hadoop-common` or `catalog-common`?

core/src/main/java/org/apache/gravitino/connector/credential/PathContext.java
```diff
@@ -31,16 +34,35 @@ public class PathContext {
...
+  private final String scheme;
 
   /**
    * Constructs a new {@code PathContext} instance with the given path and credential type.
    *
    * @param path The path string.
    * @param credentialType The type of the credential.
    */
+  @Deprecated
+  @InlineMe(replacement = "this(path, credentialType, null)")
```
mchades: why need this annotation?

core/src/main/java/org/apache/gravitino/connector/credential/PathContext.java
```diff
@@ -31,16 +34,35 @@ public class PathContext {
...
   private final String credentialType;
+  private final String scheme;
 
   /**
    * Constructs a new {@code PathContext} instance with the given path and credential type.
    *
    * @param path The path string.
    * @param credentialType The type of the credential.
    */
+  @Deprecated
```
mchades: This method is an internal method of the core module. Can we modify it directly without this annotation?

api/src/main/java/org/apache/gravitino/credential/S3SecretKeyCredential.java
roryqi: We can't. Maybe Iceberg REST will reuse them.

core/src/main/java/org/apache/gravitino/connector/credential/PathContext.java
roryqi: Otherwise, errorprone will fail.

core/src/main/java/org/apache/gravitino/connector/credential/PathContext.java
roryqi: This is marked as developer API.

api/src/main/java/org/apache/gravitino/credential/S3SecretKeyCredential.java
mchades: Iceberg REST also dependens on `:catalogs:catalog-common`

roryqi:
I find the root cause and fix this issue in another way. @mchades If you have time, you can take a look again.

roryqi:
Gently ping @mchades

mchades:
This seems like a bug fix, do we need to cherry-pick it into 1.1.1?  @jerqi

roryqi:
> This seems like a bug fix, do we need to cherry-pick it into 1.1.1? @jerqi

I prefer. I have added the label `branch-1.1`

clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/GravitinoVirtualFileSystemS3CredentialIT.java
```diff
@@ -107,10 +107,9 @@ public void startUp() throws Exception {
     conf.set("fs.gvfs.impl.disable.cache", "true");
     conf.set("fs.gravitino.server.uri", serverUri);
     conf.set("fs.gravitino.client.metalake", metalakeName);
+    conf.set("fs.gravitino.enableCredentialVending", "true");
 
     // Pass this configuration to the real file system
-    conf.set(S3Properties.GRAVITINO_S3_SECRET_ACCESS_KEY, S3_SECRET_KEY);
-    conf.set(S3Properties.GRAVITINO_S3_ACCESS_KEY_ID, S3_ACCESS_KEY);
```
yuqi1129, line 113: Why do we move these two down?

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/GravitinoVirtualFileSystemUtils.java
```diff
@@ -324,6 +326,13 @@ public static String getSubPathFromGvfsPath(NameIdentifier identifier, String gv
     return gvfsPath.substring(prefix.length());
   }
 
+  static void setCallerContextForGetCredentials(String locationName) {
+    Map<String, String> contextMap = Maps.newHashMap();
+    contextMap.put(CredentialConstants.HTTP_HEADER_CURRENT_LOCATION_NAME, locationName);
+    CallerContext callerContext = CallerContext.builder().withContext(contextMap).build();
```
yuqi1129, line 332: Does the `callerContext` here use `ThreadLocal` mechanism to hold the location name here?

yuqi1129: I remembered you guys are not favour of using `ThreadLocal `.

clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/GravitinoVirtualFileSystemS3CredentialIT.java
roryqi: Because client shouldn't set AK/SK, we should use crendial vending token here.
But we use a S3 filesystem to verify the tests which needs the AK/SK.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/GravitinoVirtualFileSystemUtils.java
roryqi: Yes, just keep the current state if it is acceptable for this feature,.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/GravitinoVirtualFileSystemUtils.java
```diff
@@ -324,6 +326,13 @@ public static String getSubPathFromGvfsPath(NameIdentifier identifier, String gv
...
+    CallerContext.CallerContextHolder.set(callerContext);
+  }
```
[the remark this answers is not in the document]
roryqi: Not a public method. We don't the Java doc.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/DefaultGravitinoFileSystemCredentialsProvider.java
```diff
@@ -51,10 +56,31 @@ public Credential[] getCredentials() {
...
+            GravitinoVirtualFileSystemConfiguration.FS_GRAVITINO_CURRENT_LOCATION_NAME_ENV_VAR,
+            GravitinoVirtualFileSystemConfiguration
+                .FS_GRAVITINO_CURRENT_LOCATION_NAME_ENV_VAR_DEFAULT);
+    String locationName =
+        Optional.ofNullable(configuration.get(FS_GRAVITINO_CURRENT_LOCATION_NAME))
+            .orElse(System.getenv(currentLocationEnvVar));
+    return locationName == null
+        ? fileset.properties().get(PROPERTY_DEFAULT_LOCATION_NAME)
+        : locationName;
+  }
```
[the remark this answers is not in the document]
roryqi: Actually, the test cases didn't take effect before. I fix the tests to cover fileset with multiple locations credential vending issues.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/GravitinoVirtualFileSystemUtils.java
```diff
@@ -324,6 +326,13 @@ public static String getSubPathFromGvfsPath(NameIdentifier identifier, String gv
     return gvfsPath.substring(prefix.length());
   }
 
+  static void setCallerContextForGetCredentials(String locationName) {
```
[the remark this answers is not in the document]
roryqi: The location won't be null. This method is used for the package. We can ensure that this value isn't null. So we don't need to check it.

roryqi:
@FANNG1 Gently ping.

Issue #9541, opened 2025-12-23
[FEATURE] Add a framework to support built-in job templates in Gravitino


jerryshao:
### Describe the feature

Gravitino has a job system to register and run the predefined jobs for users. For the ease of use, we can add some built-in jobs in Gravitino, so that it can be auto-registered when start. It will help to build the TMS and increase the convenience.

PR #9542, opened 2025-12-23
[#9541] feat(core): Add built-in job template framework in Gravitino

jerryshao:

This PR add a built-in job template support in Gravitino's job system.

Fix: #9541 

No.

UTs.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/exceptions/IllegalJobTemplateOperationException.java
+++ b/api/src/main/java/org/apache/gravitino/exceptions/IllegalJobTemplateOperationException.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.exceptions;
+
+import com.google.errorprone.annotations.FormatMethod;
+import com.google.errorprone.annotations.FormatString;
+
+/**
+ * Exception thrown when an illegal operation is attempted on a job template, such as trying to
+ * modify or delete a built-in job template, or creating a user template with a reserved name.
+ */
+public class IllegalJobTemplateOperationException extends IllegalArgumentException {
+
+  /**
+   * Constructs a new IllegalJobTemplateOperationException with the specified detail message.
+   *
+   * @param message the detail message
+   * @param args the arguments to the message
+   */
+  @FormatMethod
+  public IllegalJobTemplateOperationException(@FormatString String message, Object... args) {
+    super(String.format(message, args));
+  }
+
+  /**
+   * Constructs a new IllegalJobTemplateOperationException with the specified detail message and
+   * cause.
+   *
+   * @param cause the cause
+   * @param message the detail message
+   * @param args the arguments to the message
+   */
+  @FormatMethod
+  public IllegalJobTemplateOperationException(
+      Throwable cause, @FormatString String message, Object... args) {
+    super(String.format(message, args), cause);
+  }
+}
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/ErrorHandlers.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/ErrorHandlers.java
@@ -34,6 +34,7 @@ import org.apache.gravitino.exceptions.ConnectionFailedException;
 import org.apache.gravitino.exceptions.FilesetAlreadyExistsException;
 import org.apache.gravitino.exceptions.ForbiddenException;
 import org.apache.gravitino.exceptions.GroupAlreadyExistsException;
+import org.apache.gravitino.exceptions.IllegalJobTemplateOperationException;
 import org.apache.gravitino.exceptions.IllegalMetadataObjectException;
 import org.apache.gravitino.exceptions.IllegalPrivilegeException;
 import org.apache.gravitino.exceptions.IllegalRoleException;
@@ -1165,7 +1166,13 @@ public class ErrorHandlers {
 
       switch (errorResponse.getCode()) {
         case ErrorConstants.ILLEGAL_ARGUMENTS_CODE:
-          throw new IllegalArgumentException(errorMsg);
+          if (errorResponse
+              .getType()
+              .equals(IllegalJobTemplateOperationException.class.getSimpleName())) {
+            throw new IllegalJobTemplateOperationException(errorMsg);
+          } else {
+            throw new IllegalArgumentException(errorMsg);
+          }
 
         case ErrorConstants.NOT_FOUND_CODE:
           if (errorResponse.getType().equals(NoSuchMetalakeException.class.getSimpleName())) {
--- a/clients/client-python/gravitino/exceptions/base.py
+++ b/clients/client-python/gravitino/exceptions/base.py
@@ -181,6 +181,14 @@ class JobTemplateAlreadyExistsException(AlreadyExistsException):
     """An exception thrown when a job template with specified name already exists."""
 
 
+class IllegalJobTemplateOperationException(IllegalArgumentException):
+    """An exception thrown when an illegal operation is attempted on a job template.
+
+    This exception is raised when attempting to modify or delete a built-in job template,
+    or when trying to create a user template with a reserved name (e.g., starting with 'builtin-').
+    """
+
+
 class NoSuchJobTemplateException(NotFoundException):
     """An exception thrown when a job template with specified name is not found."""
 
--- a/clients/client-python/gravitino/exceptions/handlers/job_error_handler.py
+++ b/clients/client-python/gravitino/exceptions/handlers/job_error_handler.py
@@ -21,6 +21,7 @@ from gravitino.exceptions.handlers.rest_error_handler import RestErrorHandler
 from gravitino.exceptions.base import (
     NoSuchMetalakeException,
     JobTemplateAlreadyExistsException,
+    IllegalJobTemplateOperationException,
     NoSuchJobTemplateException,
     NoSuchJobException,
     InUseException,
@@ -33,6 +34,10 @@ class JobErrorHandler(RestErrorHandler):
         error_message = error_response.format_error_message()
         code = error_response.code()
         exception_type = error_response.type()
+
+        if code == ErrorConstants.ILLEGAL_ARGUMENTS_CODE:
+            if exception_type == IllegalJobTemplateOperationException.__name__:
+                raise IllegalJobTemplateOperationException(error_message)
 
         if code == ErrorConstants.NOT_FOUND_CODE:
             if exception_type == NoSuchMetalakeException.__name__:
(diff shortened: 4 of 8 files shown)


maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/spark/SparkPiJob.java
```diff
@@ -0,0 +1,96 @@
...
+        rdd.filter(
+                i -> {
+                  double x = Math.random() * 2 - 1;
+                  double y = Math.random() * 2 - 1;
+                  return x * x + y * y <= 1;
+                })
+            .count();
+
+    double pi = 4.0 * count / samples;
+    System.out.printf("Pi is roughly %.5f%n", pi);
```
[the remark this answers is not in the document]
jerryshao: `print` will only be used in the built-in jobs, no in Gravitino itself.

build.gradle.kts
```diff
@@ -314,14 +314,22 @@ subprojects {
   }
 
   fun compatibleWithJDK8(project: Project): Boolean {
+    val name = project.name.lowercase()
+    val path = project.path.lowercase()
+    if (path.startsWith(":maintenance:jobs") ||
+      name == "api" ||
+      name == "common" ||
```
FANNG1, line 321: @diqiu50 please help to double-check this change

core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
```diff
@@ -0,0 +1,395 @@
...
+                    .withCreateTime(Instant.now())
+                    .build())
+            .build();
+
+    try {
+      jobManager.registerJobTemplate(metalake, jobTemplateEntity);
+      LOG.info(
+          "Registered built-in job template {} under metalake {}", jobTemplate.name(), metalake);
+    } catch (JobTemplateAlreadyExistsException e) {
+      LOG.warn(
```
FANNG1, line 303: Do you think we should propagate the exception here? Since the job template is expected not to exist.

FANNG1:
Could the user create/update/drop built-in job templates?  Seems missing logic to handle this.

core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
jerryshao: This should not be happened normally, it is mainly the defensive code.

core/src/main/java/org/apache/gravitino/metalake/MetalakeManager.java
```diff
@@ -129,6 +130,29 @@ public static boolean metalakeInUse(EntityStore store, NameIdentifier ident)
...
+      return metalakes.stream()
+          .filter(
+              m -> (boolean) m.propertiesMetadata().getOrDefault(m.properties(), PROPERTY_IN_USE))
+          .map(BaseMetalake::name)
+          .collect(Collectors.toList());
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to list in-use metalakes", e);
+    }
+  }
+
```
diqiu50: Is it possible to merge these two listMetalakes methods by adding a parameter?

build.gradle.kts
diqiu50: That's OK

FANNG1:
Perhaps not within the scope of this PR, the current implementation utilizes an event listener to register the job template with all metalakes. Do we plan to support features like `system metalake` or `system catalog` to store built-in resources, such as job templates, functions, etc.?

jerryshao:
> Perhaps not within the scope of this PR, the current implementation utilizes an event listener to register the job template with all metalakes. Do we plan to support features like `system metalake` or `system catalog` to store built-in resources, such as job templates, functions, etc.?

This seems like a good solution. I think we can merge this first, then I will gradually switch to a new solution, WDYT?

core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
```diff
@@ -0,0 +1,395 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Event listener that automatically registers built-in job templates when metalakes are created.
+ *
+ * <p>This listener monitors metalake creation events and registers all discovered built-in job
+ * templates (via JobTemplateProvider SPI) into the newly created metalake. It also handles
+ * registration for existing metalakes on first startup.
+ */
+public class BuiltInJobTemplateEventListener implements EventListenerPlugin {
```
roryqi, line 66: Could u use `MetalakeHookDispatcher` instead of `EventListener`?

jerryshao: `MetalakeHookDispatcher` is a in-sync operation, I don't want this to block the main metalake create operation. Using event listener is a async operation.

roryqi: Will the user know the failure of creating built-in job template if the listener fail to create job template?

jerryshao: It will be logged.

FANNG1:
> Could the user create/update/drop built-in job templates? Seems missing logic to handle this.

@jerryshao , WDYT?

jerryshao:
> > Could the user create/update/drop built-in job templates? Seems missing logic to handle this.
> 
> @jerryshao , WDYT?

Yes, this is what we discussed before. In the current implementation, this can be done by users, but not recommended. 

Do you think we should allow or not allow users to do so? I can add some defensive codes to ban it for the users, but I don't have a strong reason to do or not to do it.

jerryshao:
> > > Could the user create/update/drop built-in job templates? Seems missing logic to handle this.
> > 
> > 
> > @jerryshao , WDYT?
> 
> Yes, this is what we discussed before. In the current implementation, this can be done by users, but not recommended.
> 
> Do you think we should allow or not allow users to do so? I can add some defensive codes to ban it for the users, but I don't have a strong reason to do or not to do it.

Let me think more on it.

FANNG1:
> Do you think we should allow or not allow users to do so? 

This depends on whether the user can easily add a custom job template based on the built-in templates. If yes, we'd better not allow the user to change the template.

In the current implementation,  there may be some problems in `reconcileBuiltInJobTemplates`:
1.  if user drop a build-in job template,  the server may register it on startup
2. if user create a job template with the prefix name `build-in`,  it will be dropped on startup

FANNG1:
> > Perhaps not within the scope of this PR, the current implementation utilizes an event listener to register the job template with all metalakes. Do we plan to support features like `system metalake` or `system catalog` to store built-in resources, such as job templates, functions, etc.?
> 
> This seems like a good solution. I think we can merge this first, then I will gradually switch to a new solution, WDYT?

Sure, this is just an immature thought and will bring significant changes to the architecture of Gravtino.

jerryshao:
> > Do you think we should allow or not allow users to do so?
> 
> This depends on whether the user can easily add a custom job template based on the built-in templates. If yes, we'd better not allow the user to change the template.
> 
> In the current implementation, there may be some problems in `reconcileBuiltInJobTemplates`:
> 
> 1. if user drop a build-in job template,  the server may register it on startup
> 2. if user create a job template with the prefix name `build-in`,  it will be dropped on startup

I see, that's a problem. I think it is better not to allow users to manipulate the built-in templates, let me handle this.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/spark/SparkPiJob.java
```diff
@@ -0,0 +1,104 @@
...
+        .build();
+  }
+
+  public static void main(String[] args) {
+    int slices = 2;
+    if (args.length > 0) {
+      try {
+        slices = Integer.parseInt(args[0]);
+      } catch (NumberFormatException e) {
+        System.err.println("Invalid number of slices provided. Using default value of 2.");
```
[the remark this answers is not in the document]
jerryshao: This is only used in the specific job, so it is fine to use it.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/spark/SparkPiJob.java
```diff
@@ -0,0 +1,104 @@
...
+    long count =
+        rdd.filter(
+                i -> {
+                  double x = Math.random() * 2 - 1;
+                  double y = Math.random() * 2 - 1;
+                  return x * x + y * y <= 1;
+                })
+            .count();
+
+    double pi = 4.0 * count / samples;
```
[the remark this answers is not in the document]
jerryshao: This is only used in the specific job, so it is fine to use it.

clients/client-python/tests/integration/test_supports_jobs.py
```diff
@@ -95,7 +95,6 @@ def test_register_and_list_job_templates(self):
 
         # List registered job templates
         registered_templates = self._metalake.list_job_templates()
-        self.assertEqual(len(registered_templates), 2)
         self.assertIn(template_1, registered_templates)
         self.assertIn(template_2, registered_templates)
```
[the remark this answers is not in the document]
jerryshao: This is intentional not to check the total number, since we added and will add more built-in job templates. Test the total number will break the test.

clients/client-python/tests/integration/test_supports_jobs.py
```diff
@@ -127,7 +126,6 @@ def test_register_and_delete_job_template(self):
 
         # List registered job templates
         registered_templates = self._metalake.list_job_templates()
-        self.assertEqual(len(registered_templates), 2)
         self.assertIn(template1, registered_templates)
         self.assertIn(template2, registered_templates)
```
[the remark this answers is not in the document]
jerryshao: This is intentional not to check the total number, since we added and will add more built-in job templates. Test the total number will break the test.

clients/client-python/tests/integration/test_supports_jobs.py
```diff
@@ -163,10 +160,6 @@ def test_register_and_delete_job_template(self):
         with self.assertRaises(NoSuchJobTemplateException):
             self._metalake.get_job_template(template2.name)
```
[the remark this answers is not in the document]
jerryshao: This is intentional not to check the total number, since we added and will add more built-in job templates. Test the total number will break the test.

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/JobAuthorizationIT.java
```diff
@@ -117,11 +117,14 @@ public void testListJobTemplates() {
...
-    Assertions.assertEquals(2, normalUserTemplates.size());
+    Assertions.assertTrue(normalUserTemplates.stream().anyMatch(s -> s.name().equals("test_1")));
+    Assertions.assertTrue(normalUserTemplates.stream().anyMatch(s -> s.name().equals("test_2")));
 
     // Admin can see all job templates (test_1, test_2, test_3)
     List<JobTemplate> adminTemplates = client.loadMetalake(METALAKE).listJobTemplates();
-    Assertions.assertEquals(3, adminTemplates.size());
+    Assertions.assertTrue(adminTemplates.stream().anyMatch(s -> s.name().equals("test_1")));
+    Assertions.assertTrue(adminTemplates.stream().anyMatch(s -> s.name().equals("test_2")));
+    Assertions.assertTrue(adminTemplates.stream().anyMatch(s -> s.name().equals("test_3")));
```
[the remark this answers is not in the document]
jerryshao: This is intentional not to check the total number, since we added and will add more built-in job templates. Test the total number will break the test.

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/JobIT.java
```diff
@@ -172,10 +169,6 @@ public void testRegisterAndDeleteJobTemplate() {
     // Verify the second job template is deleted
     Assertions.assertThrows(
         NoSuchJobTemplateException.class, () -> metalake.getJobTemplate(template2.name()));
```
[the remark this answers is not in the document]
jerryshao: This is intentional not to check the total number, since we added and will add more built-in job templates. Test the total number will break the test.

Issue #9525, opened 2025-12-23
[Subtask] Add Java APIs for UDF operations


mchades:
### Describe the subtask

 Add Java-side function CRUD APIs.

### Parent issue

#6158

PR #9551, opened 2025-12-24
[#9525] feat(api): Add APIs for UDF operations

mchades:

Add APIs for UDF operations

Fix: #9525 

no

CI pass

### DIFF
--- a/api/src/main/java/org/apache/gravitino/function/FunctionChange.java
+++ b/api/src/main/java/org/apache/gravitino/function/FunctionChange.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.function;
 import com.google.common.base.Preconditions;
 import java.util.Arrays;
 import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.annotation.Evolving;
 
 /** Represents a change that can be applied to a function. */
@@ -102,8 +103,7 @@ public interface FunctionChange {
 
     UpdateComment(String newComment) {
       Preconditions.checkArgument(
-          org.apache.commons.lang3.StringUtils.isNotBlank(newComment),
-          "New comment cannot be null or empty");
+          StringUtils.isNotBlank(newComment), "New comment cannot be null or empty");
       this.newComment = newComment;
     }
 
--- a/api/src/main/java/org/apache/gravitino/function/JavaImpl.java
+++ b/api/src/main/java/org/apache/gravitino/function/JavaImpl.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.function;
 import com.google.common.base.Preconditions;
 import java.util.Map;
 import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
 
 /** Java implementation with class name. */
 public class JavaImpl extends FunctionImpl {
@@ -33,8 +34,7 @@ public class JavaImpl extends FunctionImpl {
       Map<String, String> properties) {
     super(Language.JAVA, runtime, resources, properties);
     Preconditions.checkArgument(
-        org.apache.commons.lang3.StringUtils.isNotBlank(className),
-        "Java class name cannot be null or empty");
+        StringUtils.isNotBlank(className), "Java class name cannot be null or empty");
     this.className = className;
   }
 
--- a/api/src/main/java/org/apache/gravitino/function/SQLImpl.java
+++ b/api/src/main/java/org/apache/gravitino/function/SQLImpl.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.function;
 import com.google.common.base.Preconditions;
 import java.util.Map;
 import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
 
 /** SQL implementation with runtime and SQL body. */
 public class SQLImpl extends FunctionImpl {
@@ -32,8 +33,7 @@ public class SQLImpl extends FunctionImpl {
       FunctionResources resources,
       Map<String, String> properties) {
     super(Language.SQL, runtime, resources, properties);
-    Preconditions.checkArgument(
-        org.apache.commons.lang3.StringUtils.isNotBlank(sql), "SQL text cannot be null or empty");
+    Preconditions.checkArgument(StringUtils.isNotBlank(sql), "SQL text cannot be null or empty");
     this.sql = sql;
   }
 
--- a/api/src/main/java/org/apache/gravitino/function/PythonImpl.java
+++ b/api/src/main/java/org/apache/gravitino/function/PythonImpl.java
@@ -21,6 +21,7 @@ package org.apache.gravitino.function;
 import com.google.common.base.Preconditions;
 import java.util.Map;
 import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
 
 /** Python implementation with handler and optional inline code. */
 public class PythonImpl extends FunctionImpl {
@@ -35,8 +36,7 @@ public class PythonImpl extends FunctionImpl {
       Map<String, String> properties) {
     super(Language.PYTHON, runtime, resources, properties);
     Preconditions.checkArgument(
-        org.apache.commons.lang3.StringUtils.isNotBlank(handler),
-        "Python handler cannot be null or empty");
+        StringUtils.isNotBlank(handler), "Python handler cannot be null or empty");
     this.handler = handler;
     this.codeBlock = codeBlock;
   }


mchades:
@shaofengshi It's ready for Copilot to review

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
```diff
@@ -0,0 +1,392 @@
...
+      result = 31 * result + Objects.hashCode(runtime);
+      return result;
+    }
+
+    @Override
+    public String toString() {
+      return "RemoveImpl{parameters=" + Arrays.toString(parameters) + ", runtime=" + runtime + '}';
+    }
+  }
+}
```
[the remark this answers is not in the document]
mchades: unnecessary for the API model

mchades:
@jerryshao plz help to review this PR when you have time, thx!

api/src/main/java/org/apache/gravitino/function/Function.java
```diff
@@ -0,0 +1,89 @@
...
+    return null;
+  }
+
+  /**
+   * The output columns for a table-valued function.
+   *
+   * @return The output columns of a table-valued function, or an empty array for scalar or
+   *     aggregate functions.
+   */
+  default FunctionColumn[] returnColumns() {
```
jerryshao, line 88: What is the meaning of `FunctionColumn` and `table-valued function`, can you elaborate more on the javadoc?

api/src/main/java/org/apache/gravitino/function/Function.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Returns the internal revision version of the function.
+   *
+   * <p>This version is a 0-based counter, where {@code 0} represents the initial definition of the
+   * function, and the value is incremented by 1 on each later alteration.
+   *
+   * @return The 0-based revision version of the function.
+   */
+  int version();
```
jerryshao, line 105: Do we support storing and getting the specific version of function?

api/src/main/java/org/apache/gravitino/function/FunctionCatalog.java
```diff
@@ -0,0 +1,157 @@
...
+   * @throws NoSuchSchemaException If the schema does not exist.
+   * @throws FunctionAlreadyExistsException If the function already exists.
+   */
+  Function registerFunction(
+      NameIdentifier ident,
+      String comment,
+      boolean deterministic,
+      FunctionColumn[] returnColumns,
+      FunctionDefinition[] definitions)
+      throws NoSuchSchemaException, FunctionAlreadyExistsException;
```
jerryshao, line 133: We have two definitions of `registerFunction`, can we use the above one as a specific implementation of this one?

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
```diff
@@ -0,0 +1,393 @@
...
+package org.apache.gravitino.function;
+
+import com.google.common.base.Preconditions;
+import java.util.Arrays;
+import java.util.Objects;
+import org.apache.gravitino.annotation.Evolving;
+
+/** Represents a change that can be applied to a function. */
+@Evolving
+public interface FunctionChange {
```
jerryshao, line 29: Can we rename the function?

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
```diff
@@ -0,0 +1,393 @@
...
+    return new AddDefinition(definition);
+  }
+
+  /**
+   * Create a {@link FunctionChange} to remove an existing definition (overload) from a function.
+   *
+   * @param parameters The parameters that identify the definition to remove.
+   * @return The change instance.
+   */
+  static FunctionChange removeDefinition(FunctionParam[] parameters) {
```
jerryshao, line 59: Will the ordering of the parameters be considered as another overload?

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
```diff
@@ -0,0 +1,393 @@
...
+    return new UpdateComment(newComment);
+  }
+
+  /**
+   * Create a {@link FunctionChange} to add a new definition (overload) to a function.
+   *
+   * @param definition The new definition to add.
+   * @return The change instance.
+   */
+  static FunctionChange addDefinition(FunctionDefinition definition) {
```
jerryshao, line 49: What's the relationship between `definition` and `implementation`?

api/src/main/java/org/apache/gravitino/function/FunctionDefinition.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * @return The parameters for this definition. Maybe an empty array for a no-arg definition.
+   */
+  FunctionParam[] parameters();
+
+  /**
+   * @return The implementations associated with this definition.
+   */
+  FunctionImpl[] impls();
```
jerryshao, line 39: Why are there multiple implementations for one definition?

api/src/main/java/org/apache/gravitino/function/Function.java
mchades: Yes, in my design.

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
mchades: No, and there's no intention to support it. According to the research results, I haven't seen any engine that supports the rename operation of a function. I guess the function name is part of the function signature and should be immutable.

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
mchades: In my design, `definition` is regarded as the semantics of a function signature, while `implementation` refers to the code realization of a specific function signature. 

In other words, a `definition` can have multiple different `implementations` across various languages such as Java, Python, and SQL, yet they all produce the same logical result.

api/src/main/java/org/apache/gravitino/function/FunctionDefinition.java
mchades: see https://github.com/apache/gravitino/pull/9551#discussion_r2659622006

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
mchades: If changing the parameter order causes changes in the parameter types at different positions, it will be considered as an overload. The overload checking rules have been added to the core module in my subsequent PR:
https://github.com/apache/gravitino/pull/9560/changes#diff-ed4a067a1bb60e962e4b4fae33c559d5bdddee49150356fd79b898dcfad7b0eeR331-R345
image

api/src/main/java/org/apache/gravitino/function/FunctionCatalog.java
mchades: Merging them together would add unnecessary processing logic. I considered adding a method that includes all parameters, but that seems to make the interface bulky as well, so I ultimately chose two methods to correspond to table function and non-table function register, which appears clearer to users.

jerryshao: How do you handle this in REST API?

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
jerryshao: So, do you mean that one definition with "fixed parameters" can have multiple implementations? For example, the input is `a: String`, then it can have Java, Python different implementations?

Does Iceberg also follow such implementations? My feeling is that multi-level definitions makes this too complicated to understand.

mchades: > So, do you mean that one definition with "fixed parameters" can have multiple implementations? For example, the input is `a: String`, then it can have Java, Python different implementations?

yes.

> Does Iceberg also follow such implementations? My feeling is that multi-level definitions makes this too complicated to understand.

yes. Iceberg UDF design includes corresponding fields named [definitions](https://github.com/apache/iceberg/pull/14117/changes#diff-109630c35eec3c379bb0cba90c8a2f4329472d71098bce770b3fd2efcb90e5e8R60) and [representations](https://github.com/apache/iceberg/pull/14117/changes#diff-109630c35eec3c379bb0cba90c8a2f4329472d71098bce770b3fd2efcb90e5e8R168)

api/src/main/java/org/apache/gravitino/function/FunctionCatalog.java
mchades: Treat them as [two fields](https://github.com/apache/gravitino/pull/9566/changes#diff-1907365193e916de1100f547b06382402db4d3fcc9c2680fbd14a6b11cbaa931R62-R70) of `FunctionRegisterRequest` and mark them with `@Nullable`

api/src/main/java/org/apache/gravitino/function/Function.java
```diff
@@ -0,0 +1,93 @@
...
+package org.apache.gravitino.function;
+
+import javax.annotation.Nullable;
+import org.apache.gravitino.Auditable;
+import org.apache.gravitino.annotation.Evolving;
+import org.apache.gravitino.rel.types.Type;
+
+/** Represents a user-defined function registered in Gravitino. */
+@Evolving
+public interface Function extends Auditable {
```
jerryshao, line 41: Can you add more comments about the general description of `Function`, how does it define, what are the necessary fields?

mchades: added

api/src/main/java/org/apache/gravitino/function/Function.java
mchades: added

api/src/main/java/org/apache/gravitino/function/FunctionChange.java
```diff
@@ -0,0 +1,393 @@
...
+  /** A {@link FunctionChange} to add an implementation to a definition. */
+  final class AddImpl implements FunctionChange {
+    private final FunctionParam[] parameters;
+    private final FunctionImpl implementation;
+
+    AddImpl(FunctionParam[] parameters, FunctionImpl implementation) {
+      Preconditions.checkArgument(parameters != null, "Parameters cannot be null");
+      this.parameters = Arrays.copyOf(parameters, parameters.length);
+      this.implementation =
+          Preconditions.checkNotNull(implementation, "Implementation cannot be null");
```
jerryshao: `checkNotNull` will throw a NPE, which is not a good practice, can we change them all?

mchades: fixed

api/src/main/java/org/apache/gravitino/function/FunctionResources.java
```diff
@@ -0,0 +1,117 @@
...
+   * @param files The file resources.
+   * @param archives The archive resources.
+   * @return A {@link FunctionResources} instance.
+   */
+  public static FunctionResources of(String[] jars, String[] files, String[] archives) {
+    if ((jars == null || jars.length == 0)
+        && (files == null || files.length == 0)
+        && (archives == null || archives.length == 0)) {
+      return new FunctionResources(EMPTY, EMPTY, EMPTY);
+    }
```
jerryshao: I don't think we need this check, the constructor already handls null situation.

api/src/main/java/org/apache/gravitino/function/FunctionResources.java
```diff
@@ -0,0 +1,117 @@
...
+    return new FunctionResources(jars, files, archives);
+  }
+
+  /**
+   * @return The jar resources.
+   */
+  public String[] jars() {
+    return Arrays.copyOf(jars, jars.length);
```
jerryshao: Shall we copy again for each call? I saw this is already copied once in constructor.

api/src/main/java/org/apache/gravitino/function/FunctionType.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Parse the function type from a string value.
+   *
+   * @param type the string to parse.
+   * @return the parsed {@link FunctionType}.
+   * @throws IllegalArgumentException if the value cannot be parsed.
+   */
+  public static FunctionType fromString(String type) {
```
jerryshao, line 60: Do you need to check the validity of input `type` string?

api/src/main/java/org/apache/gravitino/function/FunctionResources.java
jerryshao: I guess there are other places that use copy unnecessarily, can you check the code overall?

api/src/main/java/org/apache/gravitino/function/SQLImpl.java
```diff
@@ -0,0 +1,85 @@
...
+public class SQLImpl extends FunctionImpl {
+  private final String sql;
+
+  SQLImpl(
+      RuntimeType runtime,
+      String sql,
+      FunctionResources resources,
+      Map<String, String> properties) {
+    super(Language.SQL, runtime, resources, properties);
+    Preconditions.checkArgument(sql != null, "SQL text cannot be null");
```
jerryshao: Some places you use StringUtils to check, some places you use `!= null` to check, I think you'd better unifying them all.

api/src/main/java/org/apache/gravitino/function/FunctionResources.java
mchades: Why do I make copies in these two places?
- In my design, functions are immutable, and every modification will result in a new version
- The constructor performs a defensive copy to prevent the caller from holding a reference to the input array and modifying the internal state. 
- If getters directly return the internal array, it would break immutability; therefore, a copy is returned.

If we consider that setting the member as `final` is sufficient and aligns with our project style, I will remove the copy from the getters.

api/src/main/java/org/apache/gravitino/function/FunctionType.java
mchades: fixed

api/src/main/java/org/apache/gravitino/function/JavaImpl.java
```diff
@@ -32,7 +32,9 @@ public class JavaImpl extends FunctionImpl {
       FunctionResources resources,
       Map<String, String> properties) {
     super(Language.JAVA, runtime, resources, properties);
-    Preconditions.checkArgument(className != null, "Java class name cannot be null");
+    Preconditions.checkArgument(
+        org.apache.commons.lang3.StringUtils.isNotBlank(className),
```
jerryshao: Why do you use full qualified package name here and below?

mchades: Sorry for the oversight.  fixed

jerryshao (approved):

Issue #9554, opened 2025-12-25
[Bug report] Flaky test `TestJobManager` fails occasionally in CI pipeline


yuqi1129:
### Version

main branch

### Describe what's wrong

```text
TestJobManager > testGetJobTemplate() FAILED
    org.mockito.exceptions.misusing.UnfinishedStubbingException: 
    Unfinished stubbing detected here:
    -> at org.apache.gravitino.job.TestJobManager.testGetJobTemplate(TestJobManager.java:279)

    E.g. thenReturn() may be missing.
    Examples of correct stubbing:
        when(mock.isOk()).thenReturn(true);
        when(mock.isOk()).thenThrow(exception);
        doThrow(exception).when(mock).someVoidMethod();
    Hints:
     1. missing thenReturn()
     2. you are trying to stub a final method, which is not supported
     3. you are stubbing the behaviour of another mock inside before 'thenReturn' instruction is completed
        at app//org.apache.gravitino.job.TestJobManager.testGetJobTemplate(TestJobManager.java:281)
```

### Error message and/or stacktrace

Please see above.

### How to reproduce

Please see link: https://github.com/apache/gravitino/actions/runs/20497547338/job/58899420472?pr=9549

PR #9559, opened 2025-12-25
[#9554] fix(test): Adjust `TestJobManager` to fix flaky test in CI pipeline

yuqi1129:

This pull request refactors the `TestJobManager` test class to improve test isolation and reliability by converting static fields and lifecycle methods to instance-level, ensuring a fresh test environment for each test case. Additionally, it addresses potential interference from background threads and updates utility methods for consistency.

Test lifecycle and isolation improvements:

* Converted all static fields (e.g., `jobManager`, `entityStore`, `config`, etc.) to instance fields, and changed `@BeforeAll`/`@AfterAll` methods to `@BeforeEach`/`@AfterEach` to ensure each test runs with a clean setup and teardown.
* Updated the test staging directory to use a unique `UUID` for each test run, preventing collisions between tests.

Background thread management:

* Explicitly shuts down background scheduler executors (`cleanUpExecutor` and `statusPullExecutor`) in the `setUp` method to prevent interference with test execution and resource leaks.

Utility method updates:

* Changed helper methods (`newShellJobTemplateEntity`, `newSparkJobTemplateEntity`, and `newJobEntity`) from static to instance methods for consistency with the new instance-based test structure. [[1]](diffhunk://#diff-de6e5a136a3bf0d4be93b1b267b421b38e404c4f70bfa798858251c5f24b9687L813-R818) [[2]](diffhunk://#diff-de6e5a136a3bf0d4be93b1b267b421b38e404c4f70bfa798858251c5f24b9687L832-R837) [[3]](diffhunk://#diff-de6e5a136a3bf0d4be93b1b267b421b38e404c4f70bfa798858251c5f24b9687L852-R857)

Cleanup and mocking:

* Removed the redundant `@AfterEach reset()` method and ensured proper cleanup of resources and mocks in the new `tearDown` method.

Fix: #9554

N/A.

Test locally.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/job/TestJobManager.java
+++ b/core/src/test/java/org/apache/gravitino/job/TestJobManager.java
@@ -39,6 +39,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 import java.util.UUID;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.ArrayUtils;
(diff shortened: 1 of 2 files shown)


yuqi1129:
I have verified this change locally. Before this change, we could reproduce it within 30 loop, after this change, I could not reproduce it with 200+ times.

core/src/test/java/org/apache/gravitino/job/TestJobManager.java
```diff
@@ -114,23 +114,53 @@ public static void setUp() throws IllegalAccessException {
...
+      statusPullExecutor.shutdown();
+      try {
+        if (!statusPullExecutor.awaitTermination(100, TimeUnit.MILLISECONDS)) {
+          statusPullExecutor.shutdownNow();
+        }
+      } catch (InterruptedException e) {
+        statusPullExecutor.shutdownNow();
+        Thread.currentThread().interrupt();
+      }
+    }
```
jerryshao: Why do you shutdown the executor before each test, not after? This doesn't logically make sense. Also, the job manager is closed in `AfterEach`, do you need to do this here using reflection?

yuqi1129: > Why do you shutdown the executor before each test, not after? This doesn't logically make sense.

It's suggested by the Copilot, and those thread pools were not used in the tests and can be shut down beforehand to reduce possible interference with the tests. 

>  do you need to do this here using reflection?

They are `private` values, of course, we can also change them as the package access level to avoid reflection.

yuqi1129:
In fact, we do need to close the pools in advance, or the following test failure will occur occasionally, the frequency is about one per twenty. 

```text
> Task :core:test
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

TestJobManager > testDeleteJobTemplate() FAILED
    org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
        at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
        at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:51)
        at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:194)
        at app//org.apache.gravitino.job.TestJobManager.testDeleteJobTemplate(TestJobManager.java:306)

12 tests completed, 1 failed

```

jerryshao (approved):

PR #9560, opened 2025-12-25
[#9561] feat(core): Add UDF management core operations and validations

mchades:

This PR implements the core operations for UDF (User-Defined Function) management in Gravitino, including:

**ManagedFunctionOperations implementation:**
- `registerFunction`: Register scalar, aggregate, and table functions with multi-definition support
- `getFunction`: Retrieve function by identifier (latest version)
- `listFunctions`: List all functions in a schema namespace
- `dropFunction`: Remove a function

**Unit tests:**
- Added comprehensive unit tests for ManagedFunctionOperations with test cases covering CRUD operations (except alter) and validation scenarios

Fix: #9561

This is part of the UDF management feature for Gravitino. The core operations layer provides the business logic for function management, ensuring data integrity through proper validations.

No, this is an internal implementation.

Added unit tests in `TestManagedFunctionOperations.java` covering:
- Function registration and listing
- Function retrieval (get)
- Function deletion (drop)
- Error cases for invalid operations

### DIFF
--- a/core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
+++ b/core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
@@ -35,18 +35,9 @@ import org.apache.gravitino.function.FunctionDefinition;
 import org.apache.gravitino.function.FunctionType;
 import org.apache.gravitino.rel.types.Type;
 
-/**
- * A class representing a function entity in the metadata store.
- *
- * <p>This entity stores both the function metadata and its version information together, avoiding
- * the need for separate FunctionEntity and FunctionVersionEntity. When retrieving, if version is
- * set to the special value {@link #LATEST_VERSION}, the store should return the latest version.
- */
+/** A class representing a function entity in the metadata store. */
 @ToString
 public class FunctionEntity implements Entity, Auditable, HasIdentifier, Function {
-
-  /** Special version value indicating the latest version should be retrieved. */
-  public static final int LATEST_VERSION = -1;
 
   public static final Field ID =
       Field.required("id", Long.class, "The unique id of the function entity.");


core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -117,21 +159,448 @@ public Function registerFunction(
...
+   * @param version The version number, or {@link FunctionEntity#LATEST_VERSION} for the latest.
+   * @return The versioned identifier.
+   */
+  private NameIdentifier toVersionedIdent(NameIdentifier ident, int version) {
+    return NameIdentifier.of(
+        ident.namespace().level(0),
+        ident.namespace().level(1),
+        ident.namespace().level(2),
+        ident.name(),
+        String.valueOf(version));
```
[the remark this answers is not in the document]
mchades: The storage will do the check in the next PR

core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
```diff
@@ -0,0 +1,357 @@
...
+      return this;
+    }
+
+    /**
+     * Builds the FunctionEntity instance.
+     *
+     * @return The constructed FunctionEntity.
+     */
+    public FunctionEntity build() {
+      functionEntity.validate();
```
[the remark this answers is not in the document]
mchades: The upper-level call has already validated the fields

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -117,21 +159,448 @@ public Function registerFunction(
...
+        if (j > 0) {
+          arity.append(",");
+        }
+        arity.append(params[j].dataType().simpleString());
+      }
+      arities.add(arity.toString());
+    }
+
+    return arities;
+  }
```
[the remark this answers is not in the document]
mchades: fixed

mchades:
@jerryshao could you plz help to review this PR when you have time, thx!

jerryshao:
I suggest splitting this PR into two. You can move the `alterFunction` to another PR.

mchades:
> I suggest splitting this PR into two. You can move the `alterFunction` to another PR.

done

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -71,29 +84,56 @@ public ManagedFunctionOperations(EntityStore store, IdGenerator idGenerator) {
...
 
   @Override
   public Function getFunction(NameIdentifier ident) throws NoSuchFunctionException {
-    // TODO: Implement when FunctionEntity is available
-    throw new UnsupportedOperationException("getFunction: FunctionEntity not yet implemented");
+    return getFunction(ident, FunctionEntity.LATEST_VERSION);
   }
 
   @Override
   public Function getFunction(NameIdentifier ident, int version)
```
jerryshao: How do user know which versions they already have and which version to pick? Since you don't have a list version interface.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -71,29 +84,56 @@ public ManagedFunctionOperations(EntityStore store, IdGenerator idGenerator) {
...
+            FunctionEntity.class);
+        // Function exists, but version doesn't
+        throw new NoSuchFunctionVersionException(
+            e, "Function %s version %d does not exist", ident, version);
+      } catch (NoSuchEntityException | IOException ex) {
+        throw new NoSuchFunctionException(e, "Function %s does not exist", ident);
+      }
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to get function " + ident, e);
+    }
```
jerryshao, line 113: Is it necessary to distinguish `NoSuchFunctionException` and `NoSuchFunctionVersionException`, to handle this you need to call the DB twice, which is not so efficient.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -117,9 +157,8 @@ public Function registerFunction(
       FunctionColumn[] returnColumns,
       FunctionDefinition[] definitions)
       throws NoSuchSchemaException, FunctionAlreadyExistsException {
-    // TODO: Implement when FunctionEntity is available
-    throw new UnsupportedOperationException(
-        "registerFunction for table-valued functions: FunctionEntity not yet implemented");
+    return doRegisterFunction(
+        ident, comment, FunctionType.TABLE, deterministic, null, returnColumns, definitions);
```
jerryshao: I'm inclined to use `Optional` instead of null, which has no type information.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -131,7 +170,173 @@ public Function alterFunction(NameIdentifier ident, FunctionChange... changes)
 
   @Override
   public boolean dropFunction(NameIdentifier ident) {
-    // TODO: Implement when FunctionEntity is available
-    throw new UnsupportedOperationException("dropFunction: FunctionEntity not yet implemented");
+    try {
+      return store.delete(ident, Entity.EntityType.FUNCTION);
```
jerryshao, line 163: Will this drop all the versions?

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -131,7 +170,173 @@ public Function alterFunction(NameIdentifier ident, FunctionChange... changes)
...
+   * definition's invocation arities overlap with another. For example, if an existing definition
+   * {@code foo(int, float default 1.0)} supports arities {@code (int)} and {@code (int, float)},
+   * adding a new definition {@code foo(int, string default 'x')} (which supports {@code (int)} and
+   * {@code (int, string)}) will be REJECTED because both support the call {@code foo(1)}. This
+   * ensures every function invocation deterministically maps to a single definition.
+   *
+   * @param definitions The array of definitions to validate.
+   * @throws IllegalArgumentException If any two definitions have overlapping arities.
+   */
+  private void validateDefinitionsNoArityOverlap(FunctionDefinition[] definitions) {
```
jerryshao, line 231: I  would suggest you to add more tests for this method.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -131,7 +170,173 @@ public Function alterFunction(NameIdentifier ident, FunctionChange... changes)
...
+   * parameters where the last M have default values supports arities from (N-M) to N parameters.
+   *
+   * <p>For example:
+   *
+   * <ul>
+   *   <li>{@code foo(int a)} → arities: {@code ["int"]}
+   *   <li>{@code foo(int a, float b)} → arities: {@code ["int,float"]}
+   *   <li>{@code foo(int a, float b default 1.0)} → arities: {@code ["int", "int,float"]}
+   *   <li>{@code foo(int a, float b default 1.0, string c default 'x')} → arities: {@code ["int",
+   *       "int,float", "int,float,string"]}
```
jerryshao, line 259: Can it be "int,string"?

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -131,7 +170,173 @@ public Function alterFunction(NameIdentifier ident, FunctionChange... changes)
...
+    }
+
+    // Find the first parameter with a default value
+    int firstDefaultIndex = params.length;
+    for (int i = 0; i < params.length; i++) {
+      if (params[i].defaultValue() != Column.DEFAULT_VALUE_NOT_SET) {
+        firstDefaultIndex = i;
+        break;
+      }
+    }
```
jerryshao: I think this loop can be merged with above.

core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
```diff
@@ -0,0 +1,357 @@
...
+          "return_type", Type.class, "The return type for scalar or aggregate functions.");
+  public static final Field RETURN_COLUMNS =
+      Field.optional(
+          "return_columns",
+          FunctionColumn[].class,
+          "The output columns for table-valued functions.");
+  public static final Field DEFINITIONS =
+      Field.required("definitions", FunctionDefinition[].class, "The definitions of the function.");
+  public static final Field VERSION =
+      Field.required("version", Integer.class, "The version of the function entity.");
```
jerryshao: I  was think the necessity of version support for UDF. If we want to support versions, we should let users know the existing versions, and can fetch the versions, otherwise, how do they know which version to use?

core/src/test/java/org/apache/gravitino/catalog/TestManagedFunctionOperations.java
```diff
@@ -0,0 +1,390 @@
...
+  }
+
+  @Test
+  public void testRegisterAndGetFunction() {
+    NameIdentifier funcIdent = getFunctionIdent("my_func");
+    FunctionParam[] params =
+        new FunctionParam[] {FunctionParams.of("input", Types.StringType.get())};
+    FunctionDefinition[] definitions = new FunctionDefinition[] {createSimpleDefinition(params)};
+
+    org.apache.gravitino.function.Function newFunc =
```
jerryshao: Don't use full qualified package name.

core/src/test/java/org/apache/gravitino/catalog/TestManagedFunctionOperations.java
```diff
@@ -0,0 +1,390 @@
...
+    return FunctionDefinitions.of(params, new FunctionImpl[] {impl});
+  }
+
+  private FunctionDefinition createDefinitionWithImpls(
+      FunctionParam[] params, FunctionImpl[] impls) {
+    return FunctionDefinitions.of(params, impls);
+  }
+
+  @Test
+  public void testInvalidParameterOrder() {
```
jerryshao: Move all the public test methods above the private methods.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
mchades: Yes. The version of the function is auto-incrementing with each alteration operation. The plan is to support rolling back the function to a specified version in the future, so this deletion will remove all versions of the function.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
mchades: No, "int,string" is not a valid arity here. Since the current arity design is based on positional parameter matching, you cannot skip the middle parameter b to directly pass c.

For engines that support named parameters (like Python), the parameter names and values would be passed together at invocation time (e.g., `{a: 1, c: "hello"}`). In that case, the engine can resolve the correct definition by mapping parameter names to their positions and validating that any missing parameters have default values. This would be handled at runtime by the engine, not during function registration.

core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
mchades: During the initial research phase, Function support for multiple versions was identified as one of the potential requirements, so I added a version field in the Function. 

However, as you mentioned, the current API implementation hasn't fully considered the operations for multiple versions. 

Therefore, I plan to:
- first implement it in Phase One by displaying the current version of the function at the API level within the `Function`, while supporting multi-version storage at the storage level.
-  In Phase Two, I will add more comprehensive Function version operations, such as `listing versions` and `rolling back`.

WDYT?

mchades:
@jerryshao all comments resolved, plz help to review again when you have time. ty!

core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
jerryshao: As we discussed offline. We can support multi-versions in the storage level. But from the user-level, they don't have to know more about the versions, they can use the latest version by default, and they can rollback to a specific version if they wanted.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
jerryshao: Did you fix this comment?

mchades: Yes. The tests `testRegisterFunctionWithOverlappingDefinitions`, `testInvalidParameterOrder`, `testNonOverlappingDefinitions`, and `testOverlappingAritiesWithDifferentTypes` can cover this method.

core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
mchades: The changes have already been implemented as discussed.

api/src/main/java/org/apache/gravitino/function/FunctionCatalog.java
```diff
@@ -52,28 +51,17 @@ public interface FunctionCatalog {
...
-   * @throws NoSuchFunctionException If the function does not exist.
-   */
-  Function getFunction(NameIdentifier ident) throws NoSuchFunctionException;
-
-  /**
-   * Get a function by {@link NameIdentifier} and version from the catalog. The identifier only
-   * contains the schema and function name. A function may include multiple definitions (overloads)
-   * in the result.
+   * <p>Note: Currently, the current version is always the latest version. In the future, when
+   * rollback functionality is supported, the current version may differ from the latest version.
```
jerryshao: We don't have to emphasize the version thing here, since we don't actually support any version-related operations. Can you update the javadoc here and the function definition?

api/src/main/java/org/apache/gravitino/function/FunctionCatalog.java
```diff
@@ -52,28 +51,17 @@ public interface FunctionCatalog {
...
    *
    * @param ident A function identifier.
-   * @param version The zero-based function version index (0 for the first created version), as
-   *     returned by {@link Function#version()}.
-   * @return The function with the given name and version.
+   * @return The current version of the function with the given name.
```
jerryshao: Also here, it is better not to give user a misinformation if unneeded.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -131,7 +160,150 @@ public Function alterFunction(NameIdentifier ident, FunctionChange... changes)
...
+            throw new IllegalArgumentException(
+                String.format(
+                    "Cannot register function: definitions at index %d and %d have overlapping "
+                        + "arity '%s'. This would create ambiguous function invocations.",
+                    i, j, arity));
+          }
+        }
+      }
+    }
+  }
```
jerryshao, line 246: My feeling is that the complexity of this method is super high, is it possible that we can reduce it?

core/src/main/java/org/apache/gravitino/meta/FunctionEntity.java
```diff
@@ -0,0 +1,357 @@
...
+    /**
+     * Builds the FunctionEntity instance.
+     *
+     * @return The constructed FunctionEntity.
+     */
+    public FunctionEntity build() {
+      functionEntity.validate();
+      return functionEntity;
+    }
+  }
```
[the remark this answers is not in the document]
mchades: see https://github.com/apache/gravitino/pull/9560#discussion_r2667830594

api/src/main/java/org/apache/gravitino/function/FunctionCatalog.java
mchades: removed

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -131,7 +160,150 @@ public Function alterFunction(NameIdentifier ident, FunctionChange... changes)
...
+   * definition's invocation arities overlap with another. For example, if an existing definition
+   * {@code foo(int, float default 1.0)} supports arities {@code (int)} and {@code (int, float)},
+   * adding a new definition {@code foo(int, string default 'x')} (which supports {@code (int)} and
+   * {@code (int, string)}) will be REJECTED because both support the call {@code foo(1)}. This
+   * ensures every function invocation deterministically maps to a single definition.
+   *
+   * @param definitions The array of definitions to validate.
+   * @throws IllegalArgumentException If any two definitions have overlapping arities.
+   */
+  private void validateDefinitionsNoArityOverlap(FunctionDefinition[] definitions) {
```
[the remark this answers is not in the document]
mchades: fixed

core/src/test/java/org/apache/gravitino/catalog/TestManagedFunctionOperations.java
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  /**
+   * Finds an entity by identifier. This method handles both versioned identifiers (used by
+   * getFunction) and original identifiers (used by alterFunction and dropFunction).
+   *
+   * <p>Versioned identifier format: namespace = original_namespace + function_name, name = version
+   * Original identifier format: namespace = schema_namespace, name = function_name
+   */
+  private FunctionEntity findEntityByIdent(NameIdentifier ident) {
```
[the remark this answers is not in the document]
mchades: fixed

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
mchades: optimized. plz review again, thx

Issue #9562, opened 2025-12-25
[Improvement] Romove unnecessary metalake access in `CatalogManager`


yuqi1129:
### What would you like to be improved?

The following code will access and check Metalake twice, as the method `catalogInUse` will check metalake again.

https://github.com/apache/gravitino/blob/bd41384a6c1b92a96a6aefb6c54c0ffb748a2bff/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java#L562-L567

https://github.com/apache/gravitino/blob/bd41384a6c1b92a96a6aefb6c54c0ffb748a2bff/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java#L867-L870

yuqi1129:
close it as it's no needed anymore.

PR #9563, opened 2025-12-25
[#9562] improvement(core): Remove unnecessary access to metalake in `CatalogManager`

yuqi1129:

This pull request focuses on simplifying and refactoring the logic for checking if a catalog or metalake is "in use" within the `CatalogManager` and `MetalakeManager` classes. The changes remove redundant code, streamline property handling, and modernize some Java syntax.

**Refactoring and simplification of "in use" checks:**

* The public `metalakeInUse` method in `MetalakeManager` has been removed, and a new private version is introduced, encapsulating this logic within the class. All catalog "in use" checks in `CatalogManager` now directly use a simplified `getCatalogInUseValue` method, eliminating the need to check the metalake's status separately. [[1]](diffhunk://#diff-e832fac4117c11b82db050ffdfbbc772f4db6d27700bd0646b1b02fe9aaeb2abL565-R563) [[2]](diffhunk://#diff-e832fac4117c11b82db050ffdfbbc772f4db6d27700bd0646b1b02fe9aaeb2abL605-R603) [[3]](diffhunk://#diff-e832fac4117c11b82db050ffdfbbc772f4db6d27700bd0646b1b02fe9aaeb2abL743-R741) [[4]](diffhunk://#diff-e832fac4117c11b82db050ffdfbbc772f4db6d27700bd0646b1b02fe9aaeb2abL867-L872) [[5]](diffhunk://#diff-21ed56786d4199a9ebee9969aae3195dff5037135c738ff4b8ba2072fa72d6a0L108-L131) [[6]](diffhunk://#diff-21ed56786d4199a9ebee9969aae3195dff5037135c738ff4b8ba2072fa72d6a0R435-R458)

**Code cleanup and modernization:**

* Unused imports such as `PropertyEntry` and the static import for `metalakeInUse` have been removed from `CatalogManager.java`. [[1]](diffhunk://#diff-e832fac4117c11b82db050ffdfbbc772f4db6d27700bd0646b1b02fe9aaeb2abL28) [[2]](diffhunk://#diff-e832fac4117c11b82db050ffdfbbc772f4db6d27700bd0646b1b02fe9aaeb2abL79)
* The `getHiddenPropertyNames` method has been removed from `CatalogManager`, as it is no longer needed.
* Java pattern matching for `instanceof` is now used in `getCatalogAlterProperty` for more concise and modern code.

For better performance.

Fix: #9562

N/A

Test locally and existing CIs

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
@@ -600,7 +600,7 @@ public class CatalogManager implements CatalogDispatcher, Closeable {
           checkMetalake(metalakeIdent, store);
 
           try {
-            if (getCatalogInUseValue(store, ident)) {
+            if (!getCatalogInUseValue(store, ident)) {
               return null;
             }
             store.update(


core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -562,7 +560,7 @@ public void enableCatalog(NameIdentifier ident)
           checkMetalake(metalakeIdent, store);
 
           try {
-            if (catalogInUse(store, ident)) {
+            if (getCatalogInUseValue(store, ident)) {
```
jerryshao, line 563: You can also check my change in #9502 , we can remove the call of `checkMetalake`, using `catalogInUse` should be enough. WDYT?

jerryshao: My feeling of the current definition of `catalogInUse` and `checkCatalogInUse` is not so accurate, it also includes the check of metalake, we'd better improve the method name or split into to methods. Otherwise, it will be misused by developers later on.

jerryshao: Also, checking the catalog needs to load and wrap the catalog entity, and the load/alter also do the same thing, it relies on the cache to avoid accessing the DB twice, but my feeling is that this can be optimized.

jerryshao: > Also, checking the catalog needs to load and wrap the catalog entity, and the load/alter also do the same thing, it relies on the cache to avoid accessing the DB twice, but my feeling is that this can be optimized.

Unfortunately, `getCatalogInUseValue` doesn't put the catalog object into cache after getting from the store, which means that the follow-up load will access the entity store again. I think we have rooms to improve.

yuqi1129: I see, and let me do more tests to optimize further

mchades: > Also, checking the catalog needs to load and wrap the catalog entity, and the load/alter also do the same thing, it relies on the cache to avoid accessing the DB twice, but my feeling is that this can be optimized.

Yes, performing `checkCatalogInUse` and `checkMetalakeInUse` from the call entry can short-circuit when metalake and catalog are in a "not in-use" state. However, in practice, catalogs and metalakes are usually `in-use` most of the time, so this actually introduces additional performance overhead in most cases.

Let me take a deep dive into the call chains of various operations, find a key call point that doesn't require additional storage access, and discuss it with @yuqi1129 offline

yuqi1129:
Close it as it's no needed.

PR #9569, opened 2025-12-26
[#9568] feat (fileset): Add support for multiple locations across different clusters in fileset

### LINKED ISSUE #9568: [FEATURE] Add support for multiple locations across different clusters in fileset.
### Describe the feature

Add support for multiple locations across different clusters in fileset. users can config different  clusters in the  multiple location configuration.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

diqiu50:

Add support for multiple locations across different clusters in fileset

Fix: #9568

Update docs

Add new ITs


catalogs/catalog-fileset/src/test/java/org/apache/gravitino/catalog/fileset/integration/test/HadoopUserAuthenticationIT.java
```diff
@@ -552,7 +564,8 @@ void createFilesetWithKerberos() {
     Assertions.assertDoesNotThrow(
         () -> catalog.asFilesetCatalog().dropFileset(NameIdentifier.of(SCHEMA_NAME, fileset2)));
 
-    Assertions.assertDoesNotThrow(() -> catalog.asSchemas().dropSchema(SCHEMA_NAME, true));
+    Assertions.assertThrows(
+        Exception.class, () -> catalog.asSchemas().dropSchema(SCHEMA_NAME, true));
```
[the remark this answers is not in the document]
diqiu50: The old logic is incorrect

docs/fileset-catalog.md
```diff
@@ -28,18 +28,19 @@ Hadoop 3. If there's any compatibility issue, please create an [issue](https://g
...
+| `config.resources`                   | The configuration resources, separated by comma. For example, `hdfs-site.xml,core-site.xml`.                                                                                                                                                                                                                         | (none)          | No       | 1.1.0            |
+| `fs.path.config.<name>`              | Defines a logical location entry. Set `fs.path.config.<name>` to the real base URI (for example, `hdfs://cluster1/`). Any key that starts with the same prefix (such as `fs.path.config.<name>.config.resource`) is treated as a location-scoped property and will be forwarded to the underlying filesystem client. | (none)          | No       | 1.1.0            |
```
mchades: The `since version` is wrong

docs/manage-fileset-metadata-using-gravitino.md
```diff
@@ -921,3 +921,43 @@ curl -X POST -H "Accept: application/vnd.gravitino.v1+json" \
...
+  "provider": "fileset",
+  "properties": {
+    "location": "hdfs://172.17.0.2:9000/fileset_catalog",
+    "fs.path.config.hadoop1"="hdfs://172.17.0.2:9000",
+    "fs.path.config.hadoop1.config.resources"="/etc/conf/cluster1/core-site.xml,/etc/conf/cluster1/hdfs-site.xml",
+    "fs.path.config.aws1"="s3://mybuctket",
+    "fs.path.config.aws1.access.key"="your_access_key",
+    "fs.path.config.aws1.secret.key"="your_secret_key"
+  }
+}' http://localhost:8090/api/metalakes/test/catalogs
```
mchades, line 943: plz add a simple comment for the code block

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Extract base location (scheme://authority) from a URI.
+   *
+   * @param uri the URI to extract base location from
+   * @return the base location in format "scheme://authority"
+   */
+  private static String getBaseLocation(URI uri) {
+    return uri.getScheme() + "://" + uri.getAuthority();
```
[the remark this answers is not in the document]
mchades: Why is this comment marked as resolved?

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
```diff
@@ -0,0 +1,138 @@
...
+          try {
+            URI valueUri = new URI(value);
+            String valueBaseLocation = getBaseLocation(valueUri);
+            if (baseLocation.equals(valueBaseLocation)) {
+              locationName = suffix;
+              break;
+            }
+          } catch (Exception e) {
+            // Skip invalid URI values
+            continue;
```
mchades: Should we throw an error to indicate that the URI is invalid?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogImpl.java
```diff
@@ -46,7 +46,7 @@ public String shortName() {
 
   @Override
   protected CatalogOperations newOps(Map<String, String> config) {
-    return new SecureFilesetCatalogOperations();
+    return new FilesetCatalogOperations();
```
mchades, line 49: could you plz help to review the auth parts? @jerqi @yuqi1129

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -166,19 +173,14 @@ static class FileSystemCacheKey {
     // authority are both null
     @Nullable private final String scheme;
     @Nullable private final String authority;
-    private final Map<String, String> conf;
+    @Nullable private String locationName;
     private final String currentUser;
 
-    FileSystemCacheKey(String scheme, String authority, Map<String, String> conf) {
+    FileSystemCacheKey(String scheme, String authority, String locationName) {
```
mchades: This new cache key will cause locations with the same `locationName` under different filesets to hit the same cache. Is this expected?

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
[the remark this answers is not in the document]
diqiu50: The `authority` should not be null in our environment .

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
diqiu50: Yes, `locationName` should identify.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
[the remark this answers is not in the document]
mchades: What if it's a file:// URI?

docs/manage-fileset-metadata-using-gravitino.md
diqiu50: Added

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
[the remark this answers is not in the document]
diqiu50: I added a check to ensure that authority is not null.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
diqiu50: fixed

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -148,46 +143,41 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
...
+    return subject.getPrincipals(UserPrincipal.class).iterator().next();
   }
 
-  /** Close the Kerberos renewal executor service to prevent resource leaks. */
-  private void close() {
-    if (kerberosRenewExecutor != null) {
-      kerberosRenewExecutor.shutdownNow();
-      kerberosRenewExecutor = null;
+  private UserGroupInformation getCurrentUgi() throws Exception {
+    UserGroupInformation currentUgi = UserGroupInformation.getLoginUser();
```
roryqi: LoginUser isn't equal to current user.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -148,46 +143,41 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
...
+    }
+    return subject.getPrincipals(UserPrincipal.class).iterator().next();
   }
 
-  /** Close the Kerberos renewal executor service to prevent resource leaks. */
-  private void close() {
-    if (kerberosRenewExecutor != null) {
-      kerberosRenewExecutor.shutdownNow();
-      kerberosRenewExecutor = null;
+  private UserGroupInformation getCurrentUgi() throws Exception {
```
roryqi: This name isn't ok for me. Because it will confuse me. Because UserGroupInformation has the method `getCurrentUser`.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogImpl.java
roryqi: Why do we remove SecurableFilsetCatalogOperations?

roryqi: I feel we should have an interface to act as the entrance of logic. Because maybe we will have other proxy logic, such as S3.

diqiu50: I put the Kerberos handling logic into the hadoop-common module and reuse it in Gvfs.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
diqiu50: Remove the unused initialization

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
diqiu50: Use getCurrentUser to repalce it

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
```diff
@@ -0,0 +1,138 @@
...
+      String value = entry.getValue();
+
+      // Check if this is a location definition: "fs.path.config.<name> = <base_location>"
+      // The key format should be exactly "fs.path.config.<name>" (no dot after name)
+      if (key.startsWith(configPrefix)) {
+        String suffix = key.substring(configPrefix.length());
+        // Check if this is a location definition (no dot after the name)
+        // Format: "fs.path.config.<name>" (not "fs.path.config.<name>.<property>")
+        if (!suffix.contains(".")
+            && StringUtils.isNotBlank(suffix)
```
yuqi1129, line 89: When will `suffix` be empty if the code goes here?

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
roryqi: Could u avoid use `current` this key word?

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -46,67 +46,61 @@
...
+            HADOOP_SECURITY_AUTHENTICATION,
+            UserGroupInformation.AuthenticationMethod.KERBEROS.name().toLowerCase(Locale.ROOT));
+        this.loginUgi = initKerberosUgi(config, configuration);
       } else {
         String userName = System.getenv(SYSTEM_ENV_HADOOP_USER_NAME);
         if (StringUtils.isEmpty(userName)) {
           userName = SYSTEM_USER_NAME;
         }
-        this.ugi = UserGroupInformation.createRemoteUser(userName);
+        this.loginUgi = UserGroupInformation.createRemoteUser(userName);
```
roryqi: If you want to reuse this part code in the GVFS, this is wrong. LoginUgi may retrieve from the AccessContext.

roryqi:
Will the clusters have multiple Kerberos servers? Do we need to switch kerberos credential?

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/FileSystemUtils.java
```diff
@@ -51,28 +44,14 @@ public class FileSystemUtils {
 
   private FileSystemUtils() {}
 
-  public static Map<String, FileSystemProvider> getFileSystemProviders(String fileSystemProviders) {
+  public static Map<String, FileSystemProvider> getFileSystemProviders() {
```
yuqi1129, line 47: Why do we remove parameter `fileSystemProviders` here?  This change will cause some breaking changes:
Before:
If we plan to create a s3 catalog, we need:
- Add aws-bundle-jar to the classpath
- Add filesystem provider `s3` to property `filesystem-providers`

After: 

Only `aws-bundle-jar` is needed, and once the jar located in the classpath, all filesystems will be loaded and used.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
mchades: This cache key is not well designed. Let's assume the following scenario:

filesetA has location: 
- loc1 -> hdfs://cluster1/path1
- loc2 -> hdfs://cluster1/path2

filesetB has locations:
- loc1 -> hdfs://cluster1/path3

The access order: `filesetA.loc1`, `filesetA.loc2`, `filesetB.loc1`

At this point, the following issues will arise:
1. `filesetA.loc1` will generate a cache, but `filesetA.loc2` cannot hit the cache (cause by different `locationName`)
2. The `filesetB.loc1` entry will hit the cache; however, the configuration on `filesetB` may differ from that on `filesetA`. In such cases, the incorrect cache will be utilized.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/FileSystemUtils.java
diqiu50: After adding this feature, the fileset catalog can no longer determine which fileset providers are required. Update it to avoid this constraint when creating filesets.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/FilesetUtil.java
diqiu50: Ignored the configuration

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -148,46 +136,41 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
...
-          }
-        },
-        DEFAULT_RENEW_INTERVAL_MS,
-        DEFAULT_RENEW_INTERVAL_MS,
-        TimeUnit.MILLISECONDS);
+  @SuppressWarnings("removal")
+  public static Principal getCurrentPrincipal() {
+    java.security.AccessControlContext context = java.security.AccessController.getContext();
+    Subject subject = Subject.getSubject(context);
+    if (subject == null || subject.getPrincipals(UserPrincipal.class).isEmpty()) {
```
roryqi: This isn't correct. HadoopUserGroupInformation doesn't use this principal. UserPrincipal is only used for Gravitino system.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -148,46 +136,41 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
...
-    if (kerberosRenewExecutor != null) {
-      kerberosRenewExecutor.shutdownNow();
-      kerberosRenewExecutor = null;
+  private UserGroupInformation getHadoopUser() throws Exception {
+    UserGroupInformation currentUgi;
+    if (impersonationEnabled) {
+      Principal principal = getCurrentPrincipal();
+      if (principal != null) {
+        String proxyUserName = getCurrentPrincipal().getName();
+        if (!proxyUserName.contains("@")) {
```
roryqi: This isn't correct, too. Because this should follow Hadoop auth to local rules. This is just simple rules I wrote for server side.

roryqi: Why do we need this for client? Maybe we shouldn't finish impersonation in the client side.

roryqi: UserPrincipal won't occur in the client side. Some classes will only occur in the server side.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
diqiu50: I don’t think it needs to be this complicated.
For case 1, creating two HDFS objects is fine.
For case 2, why can’t we simply state that locationName is unique within the catalog? In practice, it is unique in real usage.

mchades: > why can’t we simply state that locationName is unique within the catalog?

The `locationName` in each fileset is independent, and it is unreasonable to restrict the fileset `locationName` at the catalog level. We did not have this restriction from the beginning of our design

mchades: @yuqi1129 WDYT on the cache key change?

yuqi1129: I'm also afraid that the configurations affect the filesystem instance. Currently, properties did take effect on the fileset level. 

Another point is that we can assume that the location name is unique within a catalog.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/FileSystemUtils.java
yuqi1129: @mchades 
Do you have any thoughts on it?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
diqiu50: We support config the filesystem properties by  the `schema + authority`, If the case2, how to handle passing configuration.  
@mchades How do you think `fs key` should be set?

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
diqiu50: fixed

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
diqiu50: fixed

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -46,67 +42,61 @@
...
       } else {
-        String userName = System.getenv(SYSTEM_ENV_HADOOP_USER_NAME);
-        if (StringUtils.isEmpty(userName)) {
-          userName = SYSTEM_USER_NAME;
-        }
-        this.ugi = UserGroupInformation.createRemoteUser(userName);
+        this.initUgi = UserGroupInformation.getCurrentUser();
       }
 
+      UserGroupInformation currentUgi = getHadoopUser();
```
roryqi: Could u avoid using `currentUgi`？ We should give an accurate name. The  method `getHadoopUser` isn't good enough.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
```diff
@@ -148,46 +139,27 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
...
-        DEFAULT_RENEW_INTERVAL_MS,
-        TimeUnit.MILLISECONDS);
-  }
-
-  /** Close the Kerberos renewal executor service to prevent resource leaks. */
-  private void close() {
-    if (kerberosRenewExecutor != null) {
-      kerberosRenewExecutor.shutdownNow();
-      kerberosRenewExecutor = null;
+  private UserGroupInformation getHadoopUser() {
```
roryqi: We shouldn't reuse the proxy logic here. We should create a new class to override some logic using proxy ogic.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
roryqi: `currentUgi` isn't a good name. Because this isn't current user group information in Hadoop environment. I can give u some names like  `requestUser`.

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProxy.java
diqiu50: fixed

catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/HDFSFileSystemProvider.java
```diff
@@ -39,12 +42,20 @@ public class HDFSFileSystemProvider implements FileSystemProvider {
   public static final String SCHEME_HDFS = "hdfs";
 
   @Override
-  public FileSystem getFileSystem(@Nonnull Path path, @Nonnull Map<String, String> config)
+  public FileSystem getFileSystem(
```
roryqi: I prefer that clients doesn't need to know the ProxyUserHandler. Maybe we can have an another HDFSFileSystemProvider. 
Actually, it's not good point to use the catalog-common in the client, you don't need to modify this point.

diqiu50: Fixed

roryqi:
Authentication part is ok for me.

docs/fileset-catalog-with-adls.md
```diff
@@ -30,13 +30,12 @@ Apart from configurations mentioned in [fileset-catalog-catalog-configuration](.
...
+| `filesystem-providers`        | (deprected) The file system providers to add. Set it to `abs` if it's a Azure Blob Storage fileset, or a comma separated string that contains `abs` like `oss,abs,s3` to support multiple kinds of fileset including `abs`.                                                                                                                                                                                                                                                                                                                                            | (none)          | Yes      | 0.8.0-incubating |
```
mchades: You should describe what the recommended behavior is after deprecation.

docs/manage-fileset-metadata-using-gravitino.md
```diff
@@ -921,3 +918,44 @@ curl -X POST -H "Accept: application/vnd.gravitino.v1+json" \
 }' http://localhost:8090/api/metalakes/test/catalogs/fileset_catalog/schemas/test_schema/filesets
 ```
 
+Alternatively, you can pass configurations for multiple clusters using the `fs.path.config.{cluster_name}.*` format when creating the fileset catalog.
```
mchades, line 921: Which has a higher priority: the priority of this configuration, or the priority of the same configuration on the fileset?

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -164,21 +175,19 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
   static class FileSystemCacheKey {
     // When the path is a path without scheme such as 'file','hdfs', etc., then the scheme and
     // authority are both null
+
+    // NOTE: The filesystem cache key contains the schema, authority and current user.
+    // Changed the configuration of the same fileset. will not be effected by the cached
+    // filesystem.
```
yuqi1129: There exists a syntax problem between L180 and L181.

docs/fileset-catalog-with-gcs.md
```diff
@@ -29,12 +29,11 @@ Apart from configurations mentioned in [Fileset-catalog-catalog-configuration](.
...
-| `default-filesystem-provider` | The name default filesystem providers of this Fileset catalog if users do not specify the scheme in the URI. Default value is `builtin-local`, for GCS, if we set this value, we can omit the prefix 'gs://' in the location.                                                                                                                                                                                                                                                                                           | `builtin-local` | No       | 0.7.0-incubating |
+| `filesystem-providers`        | (deprected) The file system providers to add. Set it to `gcs` if it's a GCS fileset, a comma separated string that contains `gcs` like `gcs,s3` to support multiple kinds of fileset including `gcs`.                                                                                                                                                                                                                                                                                                                   | (none)          | Yes      | 0.7.0-incubating |
```
yuqi1129: typo: deprected

docs/fileset-catalog.md
```diff
@@ -28,18 +28,19 @@ Hadoop 3. If there's any compatibility issue, please create an [issue](https://g
...
+| `location-`                          | The property prefix. User can use `location-{name}={path}` to set multiple locations with different names for the catalog.                                                                                                                                                                                                      | (none)          | No       | 0.9.0-incubating |
+| `default-filesystem-provider`        | (deprected) The default filesystem provider of this Fileset catalog if users do not specify the scheme in the URI. Candidate values are 'builtin-local', 'builtin-hdfs', 's3', 'gcs', 'abs' and 'oss'. Default value is `builtin-local`. For S3, if we set this value to 's3', we can omit the prefix 's3a://' in the location. | `builtin-local` | No       | 0.7.0-incubating |
```
yuqi1129: I believed that you may need to explain more about how to use it. for example `(deprecated, all you need is to place the related bundle jar to the class path and no need to care about this configuration)

docs/fileset-catalog-with-s3.md
```diff
@@ -28,14 +28,14 @@ Once the server is up and running, you can proceed to configure the Fileset cata
...
+|-------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|----------|------------------|
+| `filesystem-providers`        | (deprected) The file system providers to add. Set it to `s3` if it's a S3 fileset, or a comma separated string that contains `s3` like `gs,s3` to support multiple kinds of fileset including `s3`.                                                                                                                                                                                                                                                                                                            | (none)          | Yes      | 0.7.0-incubating |
```
yuqi1129: Another point is that you may need to add information like `deprecated since 1.2.0`?

yuqi1129:
I have no more comments, @mchades. Would you like to take a look again.

docs/fileset-catalog-with-s3.md
diqiu50: fixed

docs/fileset-catalog.md
diqiu50: fixed

docs/fileset-catalog-with-gcs.md
diqiu50: fixed

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
diqiu50: fixed

docs/fileset-catalog-with-adls.md
diqiu50: fixed

docs/manage-fileset-metadata-using-gravitino.md
diqiu50: fixed

yuqi1129 (commented):
LGTM

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -162,23 +173,22 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
...
 
   static class FileSystemCacheKey {
-    // When the path is a path without scheme such as 'file','hdfs', etc., then the scheme and
-    // authority are both null
+    /**
+     * When the path is a path without scheme such as 'file','hdfs', etc., then the scheme and
+     * authority are both null
+     *
+     * <p>NOTE: The filesystem cache key contains the schema, authority and current user. hanged the
+     * configuration of the same fileset. will not be effected by the cached ilesystem.
```
mchades: Please review the words and grammar.

docs/manage-fileset-metadata-using-gravitino.md
```diff
@@ -921,3 +918,44 @@ curl -X POST -H "Accept: application/vnd.gravitino.v1+json" \
...
+All schemas and filesets under the catalog can inherit the configuration based on the cluster name.
+The system can automatically retrieve the appropriate configuration from the fileset catalog based on the fileset's storage location.
+The user-specified properties override properties in the fileset.
```
mchades: What are user-specified properties? The properties defined on a fileset are also considered user-specified.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -177,8 +177,8 @@ static class FileSystemCacheKey {
      * When the path is a path without scheme such as 'file','hdfs', etc., then the scheme and
      * authority are both null
      *
-     * <p>NOTE: The filesystem cache key contains the schema, authority and current user. hanged the
-     * configuration of the same fileset. will not be effected by the cached ilesystem.
+     * <p>NOTE: The filesystem cache key includes the schema, authority, and current user. Changes
+     * to the configuration of the same fileset will not be affected by the cached filesystem.
```
mchades: Did you mean `The cached filesystem will not be affected by changes to the configuration`? The current phrasing suggests the reverse.

clients/filesystem-hadoop3/src/main/java/org/apache/gravitino/filesystem/hadoop/BaseGVFSOperations.java
```diff
@@ -154,7 +155,6 @@ public static class FileSystemCacheKey {
      *
      * @param scheme the scheme of the filesystem
      * @param authority the authority of the filesystem
-     * @param ugi the user group information
```
mchades, line 157: Why is this comment being removed? The method signature remains unchanged.

docs/manage-fileset-metadata-using-gravitino.md
```diff
@@ -921,3 +918,45 @@ curl -X POST -H "Accept: application/vnd.gravitino.v1+json" \
 }' http://localhost:8090/api/metalakes/test/catalogs/fileset_catalog/schemas/test_schema/filesets
 ```
 
+Alternatively, you can pass configurations for multiple clusters using the `fs.path.config.{cluster_name}.*` format when creating the fileset catalog.
+All schemas and filesets under the catalog can inherit configuration based on the cluster name,
+and the configuration can be overridden at the schema or fileset level if needed.
+The system can automatically retrieve the appropriate configuration from `fs.path` configuration based on the fileset’s storage location
```
mchades: did you mean `fs.path.config`?

mchades (approved):

PR #9574, opened 2025-12-26
[#9586] improvement(core): optimize checking in use status of catalogs and metalakes

yuqi1129:

This pull request refactors catalog usage validation in the Gravitino codebase, improving how the system checks whether catalogs and their parent metalakes are enabled ("in use"). The main changes include removing the dynamic proxy-based `OperationDispatcherInterceptor`, centralizing usage checks within the catalog implementation, and introducing a new property to track metalake usage status. Additionally, the code now ensures that operations on catalogs consistently verify both catalog and metalake usage status.

To clean and simply code path

Fix: #9586

N/A.

UTs

### DIFF
--- a/core/src/test/java/org/apache/gravitino/catalog/TestCatalogManager.java
+++ b/core/src/test/java/org/apache/gravitino/catalog/TestCatalogManager.java
@@ -563,6 +563,9 @@ public class TestCatalogManager {
 
     Assertions.assertDoesNotThrow(() -> catalogManager.disableCatalog(ident));
 
+    CatalogEntity oldEntity = entityStore.get(ident, EntityType.CATALOG, CatalogEntity.class);
+    FieldUtils.writeField(catalog, "entity", oldEntity, true);
+
     CatalogManager.CatalogWrapper catalogWrapper =
         Mockito.mock(CatalogManager.CatalogWrapper.class);
     Capability capability = Mockito.mock(Capability.class);
@@ -628,7 +631,7 @@ public class TestCatalogManager {
 
   @Test
   void testAlterMutableProperties() {
-    NameIdentifier ident = NameIdentifier.of("metalake", "test41");
+    NameIdentifier ident = NameIdentifier.of("metalake", "test51");
     Map<String, String> props =
         ImmutableMap.of(
             "provider",


yuqi1129:
@jerryshao 
After discussion with @mchades offline, we have reached a primary agreement that we can do the following modification to optimize excessive calling of `catalogInUse` and `metalakeInUse`

- Store the metalake `in-use` information in catalogs. If the value of `in-use` is changed, then all information about it in catalogs belongs to this metalake should also be changed.
-  Move the check logic to the initialization of the catalog operations, and this check only needs to be done once and there is no need to operate the check in every operation such as `loadTable`, `createSchema`. 

This draft PR is the skeleton of the changes mentioned above.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
```diff
@@ -175,6 +176,8 @@ public CatalogOperations ops() {
           Preconditions.checkArgument(
               entity != null && conf != null, "entity and conf must be set before calling ops()");
           CatalogOperations newOps = createOps(conf);
+          // Check metalake and catalogInuse
+          checkMetalakeAndCatalogInUse(entity);
```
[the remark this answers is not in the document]
yuqi1129: This will only be called once when op is initialized.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -591,6 +573,44 @@ public void enableCatalog(NameIdentifier ident)
...
+          "Failed to update catalog {} property {} with unknown change",
+          nameIdentifier,
+          propertyKey,
+          e);
+      throw e;
+    } catch (IOException ioe) {
+      LOG.error("Failed to update catalog {} property {}", nameIdentifier, propertyKey, ioe);
+      throw new RuntimeException(ioe);
+    }
+  }
```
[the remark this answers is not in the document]
yuqi1129: fix

yuqi1129:
Combine with #9430 , the result are quite obvious compared to that in the main branch.

image

[20251230-170204.html](https://github.com/user-attachments/files/24382101/20251230-170204.html)

api/src/main/java/org/apache/gravitino/Catalog.java
```diff
@@ -132,6 +132,12 @@ enum CloudName {
   /** The property indicates the catalog is in use. */
   String PROPERTY_IN_USE = "in-use";
 
+  /**
+   * This property indicates whether the metalake that contains this catalog is in use. When a
+   * metalake is disabled, this property is set to {@code false} on all its catalogs.
+   */
+  String PROPERTY_METALAKE_IN_USE = "metalake-in-use";
```
mchades: Should this property be visible to users?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -555,18 +543,13 @@ public void testConnection(
   public void enableCatalog(NameIdentifier ident)
       throws NoSuchCatalogException, CatalogNotInUseException {
     NameIdentifier metalakeIdent = NameIdentifier.of(ident.namespace().levels());
+    checkMetalake(metalakeIdent, store);
```
mchades: Since we already store the `metalake-in-use` in the catalog, is there a need to use the `store` to checkMetalake?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -592,20 +575,54 @@ public void enableCatalog(NameIdentifier ident)
         });
   }
 
+  public void updateCatalogProperty(
```
mchades: Does this method duplicate `alterCatalog` with `setPropertyChange`?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -647,11 +664,9 @@ public Catalog alterCatalog(NameIdentifier ident, CatalogChange... changes)
         ident,
         LockType.READ,
         () -> {
-          checkCatalogInUse(store, ident);
-
           // There could be a race issue that someone is using the catalog from cache while we are
           // updating it.
-
+          checkMetalake(NameIdentifier.of(ident.namespace().levels()), store);
```
mchades: can we check the catalog property instead?

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
```diff
@@ -190,6 +194,39 @@ public CatalogOperations ops() {
...
+    if (!metalakeInuse) {
+      throw new MetalakeNotInUseException(
+          "The metalake that holds catalog %s is not in use", catalogEntity.name());
+    }
+
+    // Check the stack and find whether its within method `dropCatalog` as we don't need to check
+    // the catalog in use status when dropping the catalog.
+    if (isInvokedBy("dropCatalog")) {
+      LOG.info("Skip checking catalog in use status since it's invoked by dropCatalog method.");
+      return;
```
mchades, line 216: when a catalog is in use, we cannot drop it.

core/src/main/java/org/apache/gravitino/metalake/MetalakeManager.java
```diff
@@ -367,25 +369,43 @@ public void enableMetalake(NameIdentifier ident) throws NoSuchMetalakeException
...
+                          : Maps.newHashMap(metalake.properties());
+                  newProps.put(PROPERTY_IN_USE, "true");
+                  builder.withProperties(newProps);
+
+                  return builder.build();
+                });
+
+            // The only problem is that we can't make sure we can change all catalog properties
+            // in a transaction. If any catalog property update fails, the metalake is already
+            // enabled but catalog properties remain inconsistent.
```
mchades, line 395: We can try our best to update all catalogs, and log the catalogs that failed to change in the warning log.

api/src/main/java/org/apache/gravitino/Catalog.java
yuqi1129: No, I have set the property `true` to `true`.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: We need to load the catalog first before checking `metalake-in-use`. I also struggling about which one to use, let me think a bit.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
yuqi1129: In `dropCatalog`, we will check it.

core/src/main/java/org/apache/gravitino/metalake/MetalakeManager.java
yuqi1129: It did log the error message, but the core problem is that we need a transaction mechanism to make them consistent.

api/src/main/java/org/apache/gravitino/Catalog.java
mchades: If so, I would recommend placing the constant within the core module.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
mchades: What I mean is, the update catalog operation below already involves accessing the store, so do we really need to check here？ As this would result in an extra database access.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
mchades: Does this mean we are doing redundant checks? Can we optimize this part of the logic?

core/src/main/java/org/apache/gravitino/metalake/MetalakeManager.java
mchades: Did I miss something? Could you point out where it was logged?

yuqi1129: When we try to update a property using the following code

```java
                     GravitinoEnv.getInstance()
                          .catalogManager()
                          .alterCatalog(
                              catalogEntity.nameIdentifier(),
                              CatalogChange.setProperty(Catalog.PROPERTY_METALAKE_IN_USE, "true"));
```

There are already logs about the identifier, and the case is that if anyone fails, the whole operation will stop. By the way, I can add some logs to mark which has succeeded.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
yuqi1129: > Does this mean we are doing redundant checks? Can we optimize this part of the logic?


What I mean is that:  We will check the status of the in-use catalog manager, as the catalog manager will not call catalog operations(schema, table operations) normally, then `checkMetalakeAndCatalogInUse` will not be called. 

Typically, `checkMetalakeAndCatalogInUse` is targeted for schema and table-level operations.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: updated

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: changed.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: `alterCatalog` with `setPropertyChange` is used by users, and it can't change reserved properties, including `metalake-in-use`, so I revert the changes here.

yuqi1129:
@mchades 
All have been resolved, please take a look again if you are free, thanks.

core/src/main/java/org/apache/gravitino/metalake/MetalakeManager.java
mchades: Can we catch all exceptions and log all failed catalogs in the log? This would make the logs clearer.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
mchades: so this call can be removed?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -643,15 +634,20 @@ public void disableCatalog(NameIdentifier ident) throws NoSuchCatalogException {
...
       throws NoSuchCatalogException, IllegalArgumentException {
+    NameIdentifier metalakeIdent = NameIdentifier.of(ident.namespace().levels());
+
     TreeLockUtils.doWithTreeLock(
         ident,
         LockType.READ,
         () -> {
-          checkCatalogInUse(store, ident);
+          checkMetalake(metalakeIdent, store);
+          boolean catalogInUse = getCatalogInUseValue(store, ident);
```
mchades: can we reuse the following catalogWrapper to do the check?

core/src/main/java/org/apache/gravitino/metalake/MetalakeManager.java
yuqi1129: Do you mean we can continue to run if any catalog update fails, and collect all failed catalogs?

yuqi1129: Changed.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
mchades, line 568: if the original `catalog-in-use` is `true`, the update will fail? since:

https://github.com/apache/gravitino/blob/1ed975bdb1a890828d7f3af1c14bd4b2691b6cf2/core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java#L234-L238

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -555,23 +552,22 @@ public void testConnection(
...
-            if (catalogInUse(store, ident)) {
-              return null;
-            }
-
             store.update(
                 ident,
                 CatalogEntity.class,
                 EntityType.CATALOG,
                 catalog -> {
+                  boolean metalakeInuse = metalakeInUseByCatalogProperty(catalog.getProperties());
```
mchades: I'm considering whether we can retrieve the catalog via `catalogCache.get` before updating the store. If both `metalake-in-use` and `catalog-in-use` are set to `true`, we can directly return the result without performing any store operations. (Based on experience, the cache hit rate should be relatively high.)

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: `updateResult` will never be 0 as long as the catalog exists.

image

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
```diff
@@ -190,6 +196,38 @@ public CatalogOperations ops() {
...
+        Boolean.parseBoolean(catalogProperties.getOrDefault(PROPERTY_METALAKE_IN_USE, "true"));
+    if (!metalakeInuse) {
+      throw new MetalakeNotInUseException(
+          "The metalake that holds catalog %s is not in use", entity.name());
+    }
+
+    // Check the stack and find whether its within method `dropCatalog` as we don't need to check
+    // the catalog in use status when dropping the catalog.
+    if (isInvokedBy("dropCatalog")) {
+      LOG.info("Skip checking catalog in use status since it's invoked by dropCatalog method.");
```
mchades, line 215: Can we move the check here to centralize all check logic in one place, making it easier to maintain?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: It may be, and the cache will improve the performance.  It's not the scope of this PR, and we can use another PR to solve similar problems overall. 

By the way, why don't we use the cache to fetch catalog status before?

mchades: > It's not the scope of this PR, and we can use another PR to solve similar problems overall.

It's better to leave a todo in the code, otherwise it's easy to forget 

> By the way, why don't we use the cache to fetch catalog status before?

It's an oversight.

yuqi1129:
```
                  BaseCatalog baseCatalog = loadCatalogAndWrap(ident).catalog;
                  baseCatalog.checkMetalakeInUse();
```

This should be enough as `loadCatalogAndWrap` will use the cache to load catalog information.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
yuqi1129: done

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: done

yuqi1129:
@mchades 
All resolved, please help to take a look.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -1255,4 +1213,42 @@ private CatalogEntity convertFilesetCatalogEntity(CatalogEntity entity) {
     // If the provider is not "hadoop", we assume it is already a fileset catalog entity.
     return entity;
   }
+
+  public void updateCatalogProperty(
```
mchades: I think this method should be set as private, and then expose a method like `setMetalakeStatus`, otherwise other developers might confuse it with `alterCatalog` and get confused.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -555,18 +541,14 @@ public void testConnection(
...
       throws NoSuchCatalogException, CatalogNotInUseException {
     NameIdentifier metalakeIdent = NameIdentifier.of(ident.namespace().levels());
-
     TreeLockUtils.doWithTreeLock(
         metalakeIdent,
         LockType.WRITE,
         () -> {
-          checkMetalake(metalakeIdent, store);
+          BaseCatalog baseCatalog = loadCatalogAndWrap(ident).catalog();
+          baseCatalog.checkMetalakeInUse();
```
mchades, line 549: After this check, if  `catalog-in-use` of the baseCatalog is true, can we return directly?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: modified

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: modified

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -756,6 +744,7 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
               throw new NonEmptyCatalogException(
                   "Catalog %s has schemas, please drop them first or use force option", ident);
             }
+            catalogWrapper.catalog().checkMetalakeInUse();
```
mchades: 742 line already has a `catalogWrapper`. BTW, 732 line can also be optimized by the catalog property?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -865,12 +854,6 @@ public CatalogWrapper loadCatalogAndWrap(NameIdentifier ident) throws NoSuchCata
     return catalogCache.get(ident, this::loadCatalogInternal);
   }
 
-  private boolean catalogInUse(EntityStore store, NameIdentifier ident)
-      throws NoSuchMetalakeException, NoSuchCatalogException {
-    NameIdentifier metalakeIdent = NameIdentifier.of(ident.namespace().levels());
-    return metalakeInUse(store, metalakeIdent) && getCatalogInUseValue(store, ident);
-  }
-
   private boolean getCatalogInUseValue(EntityStore store, NameIdentifier catalogIdent) {
```
mchades: Is there room for optimization in this method?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -1255,4 +1222,52 @@ private CatalogEntity convertFilesetCatalogEntity(CatalogEntity entity) {
...
     return entity;
   }
+
+  /**
+   * Set the metalake in-use status in a specified catalog.
+   *
+   * @param nameIdentifier The name identifier of the catalog.
+   * @param status The in-use status to set. It can be "true" or "false".
+   */
+  public void setMetalakeInUseStatus(NameIdentifier nameIdentifier, String status) {
```
mchades: Suggest making `status` a boolean in the signature

yuqi1129: All done @mchades

mchades (approved):
current changes are LGTM.

@jerryshao could you plz also take a review? thx

jerryshao:
You can go ahead if you make sure the logic is correct.

Issue #9583, opened 2025-12-30
[Improvement] loadTable should indicate if it's for writing


jerqi:
### What would you like to be improved?

Gravitino that have access control, read and write permissions can be different. So we should load table should indicate if it's for writing.

### How should we improve?

Gravitino that have access control, read and write permissions can be different. So we should load table should indicate if it's for writing.

PR #9585, opened 2025-12-30
[#9583] improvement(authz): loadTable should indicate if it's for writing

roryqi:

loadTable should indicate if it's for writing

Fix: #9583

No.

Add new test cases.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/filter/authorization/LoadTableAuthorizationExecutor.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/filter/authorization/LoadTableAuthorizationExecutor.java
@@ -37,6 +37,7 @@ import org.apache.gravitino.server.web.filter.ParameterUtil;
  * determine authorization: if the condition is met (e.g., client requests MODIFY_TABLE privilege),
  * the secondaryExpression is used for stricter authorization, otherwise the default expression is
  * used (e.g., SELECT_TABLE).
+ *
  * <p><b>Security Limitation:</b> This is a trust-based model. The client declares intended
  * privileges, and the server trusts this declaration without validation. A malicious or modified
  * client could request only SELECT_TABLE privileges to bypass MODIFY_TABLE authorization checks.


server/src/main/java/org/apache/gravitino/server/web/rest/TableOperations.java
```diff
@@ -164,14 +165,17 @@ public Response createTable(
   @Timed(name = "load-table." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
   @ResponseMetered(name = "load-table", absolute = true)
   @AuthorizationExpression(
-      expression = AuthorizationExpressionConstants.loadTableAuthorizationExpression,
+      expression = AuthorizationExpressionConstants.LOAD_TABLE_AUTHORIZATION_EXPRESSION,
```
hdygxsj, line 168: After using LoadTableAuthorizationExecutor, do we still need to use the metadata authorization expression here?

roryqi: If privileges is null, we will still use the expression. So I think we should keep this expresssion.

hdygxsj: Since LoadTableAuthorizationExecutor replaces the authorization expression, we should add explanatory comments to clarify this behavior and prevent users from being confused by apparent inconsistencies between the declared privileges and the actual expressions.

Alternatively, we could make the behavior more transparent by encoding the logic directly into the expression itself—for example:

```java
PRIVILEGE == 'MODIFY_TABLE' && MODIFY_TABLE_AUTHORIZATION_EXPRESSION 
|| PRIVILEGE == null && LOAD_TABLE_AUTHORIZATION_EXPRESSION
```

roryqi: I modified the AuthorizationExpression. Could u take a look again?

hdygxsj: I'm a bit unsure—could you please take a look? @FANNG1

roryqi: What‘s your concern? What do u need to confirm?

server/src/main/java/org/apache/gravitino/server/web/filter/authorization/AuthorizeExecutorFactory.java
```diff
@@ -23,45 +23,37 @@
...
       AuthorizationRequest.RequestType requestType,
       Map<Entity.EntityType, NameIdentifier> metadataContext,
-      AuthorizationExpressionEvaluator authorizationExpressionEvaluator,
       Map<String, Object> pathParams,
       Optional<String> entityType,
       Parameter[] parameters,
-      Object[] args) {
+      Object[] args,
+      String secondaryExpression,
+      String secondaryExpressionCondition) {
```
hdygxsj, line 38: I'm wondering if modifying the annotation structure for this particular scenario might not be a good idea—mainly because it's unclear whether other interfaces will have similar scenarios in the future, which could lead to things getting messy later on.

Changing the input parameter here to AuthorizationExpression might be slightly better.

roryqi: This is a proper common ability. We can choose the expression according to the parameters. If we need more parameters, we can add more parameters.

roryqi:
Merged. Thanks.

PR #9587, opened 2025-12-31
[MINOR] improvement(authn): Add the  command example for simple mode

roryqi:

Add the  command example for simple mode

Just a minor.

Just documents.

No need.

### DIFF
--- a/docs/security/how-to-authenticate.md
+++ b/docs/security/how-to-authenticate.md
@@ -33,6 +33,16 @@ GravitinoClient client = GravitinoClient.builder(uri)
     .build();
 ```
 
+When using curl or other HTTP clients, authenticate with the `Authorization` header:
+
+```shell
+curl -v -X GET \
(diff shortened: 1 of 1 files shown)


docs/security/how-to-authenticate.md
```diff
@@ -33,6 +33,16 @@ GravitinoClient client = GravitinoClient.builder(uri)
     .build();
 ```
 
+When using curl or other HTTP clients, authenticate with the `Authorization` header:
+
+```shell
+curl -v -X GET \
+  -H "Accept: application/vnd.gravitino.v1+json" \
+  -H "Content-Type: application/json" \
+  -H "Authorization: Basic $(echo -n 'admin:' | base64)" \
```
yuqi1129, line 42: Shouldn't we also add password in the example?

roryqi: It's not necessary. The simple mode doesn't verify the password

Issue #9543, opened 2025-12-23
[FEATURE] Add an Iceberg rewrite datafile built-in job in Gravitino


jerryshao:
### Describe the feature

Based on the framework of #9541 , we will add an Iceberg rewrite datafile built-in job in Gravitino. This built-in job will be used by Gravitino TMS to do the data compaction.

PR #9588, opened 2025-12-31
[#9543] feat(jobs): Add the built-in Iceberg rewrite data files job template to Gravitino

jerryshao:
This commit implements a built-in job template for rewriting Iceberg table data files, which supports binpack, sort strategies for table optimization.

Key Features:
 - Named argument parser supporting flexible parameter combinations
 - Calls Iceberg's native rewrite_data_files stored procedure
 - Supports all rewrite strategies: binpack, sort
 - Configurable options for file sizes, thresholds, and behavior
 - Template-based configuration for Spark and Iceberg catalogs
 - Handles both Iceberg 1.6.1 (4 columns) and newer versions (5 columns)

Implementation:
 - IcebergRewriteDataFilesJob.java (335 lines)
   - Template name: builtin-iceberg-rewrite-data-files
   - Version: v1
   - Arguments: --catalog, --table, --strategy, --sort-order, --where, --options
   - Spark configs for runtime and Iceberg catalog setup
 - BuiltInJobTemplateProvider.java (modified)
   - Registered new IcebergRewriteDataFilesJob
 - build.gradle.kts (modified)
  - Added Iceberg Spark runtime dependency (1.6.1)
  - Added Spark, Scala, and Hadoop test dependencies

Tests (41 tests, all passing):
 - TestIcebergRewriteDataFilesJob.java (33 tests, 429 lines)
   - Template structure validation
   - Argument parsing (required, optional, empty values, order-independent)
   - JSON options parsing (single, multiple, boolean, empty)
   - SQL generation (minimal, with strategy, sort, where, options, all params)

 - TestIcebergRewriteDataFilesJobWithSpark.java (8 tests, 229 lines)
  - Real Spark session integration tests
  - Executes actual Iceberg rewrite_data_files procedures
  - Validates data integrity after rewrite operations
  - Tests all parameter combinations with live Iceberg catalog

Usage Examples:

```
--catalog iceberg_prod --table db.sample

--catalog iceberg_prod --table db.sample --strategy sort \
 --sort-order 'id DESC NULLS LAST'

--catalog iceberg_prod --table db.sample --strategy sort \
 --sort-order 'zorder(user_id, event_type, timestamp)'

--catalog iceberg_prod --table db.sample --where 'year = 2024' \
  --options '{"min-input-files":"2","remove-dangling-deletes":"true"}'
```

Fix: #9543

### DIFF
--- a/maintenance/jobs/build.gradle.kts
+++ b/maintenance/jobs/build.gradle.kts
@@ -35,7 +35,7 @@ val icebergVersion: String = libs.versions.iceberg4connector.get()
 val sparkMajorVersion = "3.5"
 
 dependencies {
-  compileOnly(project(":api"))
+  implementation(project(":api"))
 
   compileOnly(libs.slf4j.api)
   compileOnly(libs.jackson.databind)
@@ -78,10 +78,15 @@ tasks.test {
 
 tasks.withType(ShadowJar::class.java) {
   isZip64 = true
+  configurations = listOf(project.configurations.runtimeClasspath.get())
   archiveClassifier.set("")
   mergeServiceFiles()
 
   dependencies {
+    relocate("com.google", "org.apache.gravitino.shaded.com.google")
+    relocate("org.apache.commons", "org.apache.gravitino.shaded.org.apache.commons")
+    relocate("com.fasterxml", "org.apache.gravitino.shaded.com.fasterxml")
+
     exclude(dependency("org.apache.spark:.*"))
     exclude(dependency("org.apache.iceberg:.*"))
     exclude(dependency("org.slf4j:slf4j-api"))


maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
```diff
@@ -0,0 +1,348 @@
...
+  /**
+   * Main entry point for the rewrite data files job.
+   *
+   * <p>Uses named arguments for flexibility: --catalog &lt;catalog_name&gt; Required. Iceberg
+   * catalog name --table &lt;table_identifier&gt; Required. Table name (db.table) --strategy
+   * &lt;strategy&gt; Optional. binpack or sort --sort-order &lt;sort_order&gt; Optional. Sort order
+   * specification --where &lt;where_clause&gt; Optional. Filter predicate --options
+   * &lt;options_json&gt; Optional. JSON map of options
+   *
+   * <p>Example: --catalog iceberg_catalog --table db.sample --strategy binpack --options
```
FANNG1: could you add `--where` to the example?

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
```diff
@@ -0,0 +1,348 @@
...
+            + "  --catalog <name>          Iceberg catalog name registered in Spark\n"
+            + "  --table <identifier>      Fully qualified table name (e.g., db.table_name)\n"
+            + "\n"
+            + "Optional Options:\n"
+            + "  --strategy <name>         Rewrite strategy: binpack (default) or sort\n"
+            + "  --sort-order <spec>       Sort order specification:\n"
+            + "                              For columns: 'id DESC NULLS LAST, name ASC'\n"
+            + "                              For Z-Order: 'zorder(c1,c2,c3)'\n"
+            + "  --where <predicate>       Filter predicate to select files\n"
+            + "                              Example: 'year = 2024 and month = 1'\n"
```
FANNG1: Could you add a str column to the where example?

FANNG1:
Besides providing the argument options in main method, could you provide the document about how to pass the argument (especially for where and options) when submit a rewrite job, since there maybe `'` `"` transform in the job internally system.

FANNG1:
Another question is, do you plan to allow users to inject custom Spark configurations?

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: Can you give me an example?

jerryshao:
> Another question is, do you plan to allow users to inject custom Spark configurations?

I think this is valid, let me think of how to support it.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: You can check the usage to see how to use it.

jerryshao:
@FANNG1 please help to review again.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
```diff
@@ -0,0 +1,535 @@
...
+   * job or cause security issues.
+   *
+   * @param configs map of Spark configurations to validate
+   * @throws IllegalArgumentException if any configuration is not allowed
+   */
+  static void validateSparkConfigs(Map<String, String> configs) {
+    // Define configurations that should not be overridden by users
+    // These are critical for the job to function correctly
+    String[] forbiddenPrefixes = {
+      "spark.sql.catalog.", // Catalog configs are managed by the template
```
FANNG1: It's hard to manage all catalog configs, like catalog io-impl and related storage configurations

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
```diff
@@ -0,0 +1,535 @@
...
+    try {
+      // Build the procedure call SQL
+      String sql =
+          buildProcedureCall(
+              catalogName, tableIdentifier, strategy, sortOrder, whereClause, optionsJson);
+
+      System.out.println("Executing Iceberg rewrite_data_files procedure: " + sql);
+
+      // Execute the procedure
+      Row[] results = (Row[]) spark.sql(sql).collect();
```
FANNG1, line 172: `(Row[]) ` seems not required

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
```diff
@@ -0,0 +1,535 @@
...
+        .withConfigs(buildSparkConfigs())
+        .withCustomFields(
+            Collections.singletonMap(JobTemplateProvider.PROPERTY_VERSION_KEY, VERSION))
+        .build();
+  }
+
+  /**
+   * Main entry point for the rewrite data files job.
+   *
+   * <p>Uses named arguments for flexibility: --catalog &lt;catalog_name&gt; Required. Iceberg
```
FANNG1: Is it possible to make this content more user-friendly?

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
```diff
@@ -0,0 +1,535 @@
...
+  /**
+   * Build template arguments list with named argument format.
+   *
+   * @return list of template arguments
+   */
+  private static List<String> buildArguments() {
+    return Arrays.asList(
+        "--catalog",
+        "{{catalog_name}}",
+        "--table",
```
FANNG1, line 480: If the user doesn't input the optional config like `sort_order`,  the main function will get `--sort-order  {{sort_order}}` because `createRuntimeJobTemplate` only does placeholder replacement, YES?

FANNG1:
Seems missing the transform logic for special config values:
* json str like `options` and `spark_conf`, user input is {"a":"b"}, but argument requires `'{\"a\":\"b\"}' `  
* string with black space and `"` like `where`, user input is  `status = "active''`  but argument requires  `'status = \''active\'' '`
* string with `'`, like `status = 'active'`, should we prohibit this input format?

FANNG1:
The current test is mainly focused on the internal logic for the rewrite job template. Can it be added to the run job interface?

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: What do you want to clarify here?

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: Javadoc requires some characters to be encoded; I will try to split into multiple lines for better reading.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: I see. That's an issue, let me see how to fix it.

jerryshao: One workaround is to use the empty string for the argument that doesn't have a value. This requires some changes in the job manager. I would like to do this in another PR, what do you think?

jerryshao:
> Seems missing the transform logic for special config values:
> 
> * json str like `options` and `spark_conf`, user input is {"a":"b"}, but argument requires `'{\"a\":\"b\"}' `
> * string with black space and `"` like `where`, user input is  `status = "active''`  but argument requires  `'status = \''active\'' '`
> * string with `'`, like `status = 'active'`, should we prohibit this input format?

Can you explain more? I'm not quite following your comment here.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: I see your points, let me loosen the constraints.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: It's needed, otherwise it will be failed to compile:

```
/Users/jerryshao/Projects/gravitino/maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java:172: error: incompatible types: Object cannot be converted to Row[]
      Row[] results = spark.sql(sql).collect();
                                            ^
1 error
```

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
FANNG1: I'm ok

FANNG1:
> > Seems missing the transform logic for special config values:
> > 
> > * json str like `options` and `spark_conf`, user input is {"a":"b"}, but argument requires `'{\"a\":\"b\"}' `
> > * string with black space and `"` like `where`, user input is  `status = "active''`  but argument requires  `'status = \''active\'' '`
> > * string with `'`, like `status = 'active'`, should we prohibit this input format?
> 
> Can you explain more? I'm not quite following your comment here.

Take `options` for example,  the user will pass `options` as a config key and a json str like {"target-file-size-bytes":"536870912"} as config value,  the job manager will use --options  {"target-file-size-bytes":"536870912"} to run the rewrite job, seems this is not expected, we should do some special handling like using `'` to wrappe the option string ,  and add `/` before `"`,   --options  '{\"target-file-size-bytes\":\"536870912\"}' is expected, please correct me if there are some misunderstanding.

jerryshao:
> > > Seems missing the transform logic for special config values:
> > > 
> > > * json str like `options` and `spark_conf`, user input is {"a":"b"}, but argument requires `'{\"a\":\"b\"}' `
> > > * string with black space and `"` like `where`, user input is  `status = "active''`  but argument requires  `'status = \''active\'' '`
> > > * string with `'`, like `status = 'active'`, should we prohibit this input format?
> > 
> > 
> > Can you explain more? I'm not quite following your comment here.
> 
> Take `options` for example, the user will pass `options` as a config key and a json str like {"target-file-size-bytes":"536870912"} as config value, the job manager will use --options {"target-file-size-bytes":"536870912"} to run the rewrite job, seems this is not expected, we should do some special handling like using `'` to wrappe the option string , and add `/` before `"`, --options '{"target-file-size-bytes":"536870912"}' is expected, please correct me if there are some misunderstanding.

I don't know how do you do the test, this is basically the limitation of how bash deals with an input string with **spaces** as an argument; it should wrap with "", also it should escape with "\" for the `"` inside the string.

Can you provide an example of how you test it?

FANNG1:
The main issue is determining who is responsible for escaping special characters when submitting a job: the job client or the job executor? I prefer the latter option, as escaping special characters is part of the internal logic for local shell executors. What do you think? Below is an example of a job configuration for submitting a job.

| Job config items | Job config values | Example                                | 
|------------------|-------------------|----------------------------------------|
| where_clause     | normal string     | a = 1 and b = "2"                      |
| options          | json str          | {"target-file-size-bytes":"536870912"} |
| spark_conf       | json str          | {"spark.sql.shuffle.partitions":"200"} |

jerryshao:
> The main issue is determining who is responsible for escaping special characters when submitting a job: the job client or the job executor? I prefer the latter option, as escaping special characters is part of the internal logic for local shell executors. What do you think? Below is an example of a job configuration for submitting a job.
> 
> Job config items	Job config values	Example
> where_clause	normal string	a = 1 and b = "2"
> options	json str	{"target-file-size-bytes":"536870912"}
> spark_conf	json str	{"spark.sql.shuffle.partitions":"200"}

I think this is not a client or executor problem, it is the problem of how Java handles `"` in the input argv, no matter whether the program is run by CLI or programmatically. 

According to my test in IDEA, the input parameter `a = 1 and b = "2"` will be converted to `a = 1 and b = 2` when getting from argv, the " is lost, which means the meaing of string literal is changed. Also for `{"target-file-size-bytes":"536870912"}`, it will be converted to `{target-file-size-bytes:536870912}`, which is not a json map anymore.

If we input `a = 1 and b = \"2\"`, then it will convert to `a = 1 and b = "2"` in the argv. And for `{\"target-file-size-bytes\":\"536870912\"}`, it will convert to `{"target-file-size-bytes":"536870912"}`, which is a json map. This is what we expected.

What I want to say is that the handling of escaping `"` is not controlled and cannot be controlled by this code I implemented. This code doesn't know whether the where clause is a string literal or an integer. 

Besides, the input parameters may be different if you run in different environments (CLI, IDEA, Java program).

What this code **wants** is a correct string, whether it is a JSON string or a string literal. It is the user's responsibility to make sure the input parameters can be correctly converted to the expected string, not the code here.

I suggest you try this with your TMS framework, you will better know how to use this job. If there's a problem, then we can fix it.

FANNG1:
some problems while test the rewrite job
1. The Gravitino job jar doesn't include the Gravitino API package will introduce class not found error when executiong Spark jobs
2. missing jar or archive handle logic, so I had to place Iceberg Spark runtime jar to the jar under Spark home
3. it takes too long time to sync the finished status to the client
```
2026-01-29 09:54:47.037 WARN [LocalJobExecutor-100] [org.apache.gravitino.job.local.LocalJobExecutor.runJob(LocalJobExecutor.java:295)] - Job local-job-f340c4d9-8ef0-4f44-a63e-c46740e54246 failed after starting with exit code: 1
2026-01-29 09:59:28.008 INFO [job-status-pull] [org.apache.gravitino.job.JobManager.lambda$pullAndUpdateJobStatus$19(JobManager.java:616)] - Updated the job job-2097194451929410612 with execution id local-job-f340c4d9-8ef0-4f44-a63e-c46740e54246 status to FAILED
```

For the escaping issues,  seems we didn't need to do special handing, as we use `ProcessBuilder` not a `shell` to execute Spark command.

jerryshao:
> some problems while test the rewrite job
> 
> 1. The Gravitino job jar doesn't include the Gravitino API package will introduce class not found error when executiong Spark jobs
> 2. missing jar or archive handle logic, so I had to place Iceberg Spark runtime jar to the jar under Spark home
> 3. it takes too long time to sync the finished status to the client
> 
> ```
> 2026-01-29 09:54:47.037 WARN [LocalJobExecutor-100] [org.apache.gravitino.job.local.LocalJobExecutor.runJob(LocalJobExecutor.java:295)] - Job local-job-f340c4d9-8ef0-4f44-a63e-c46740e54246 failed after starting with exit code: 1
> 2026-01-29 09:59:28.008 INFO [job-status-pull] [org.apache.gravitino.job.JobManager.lambda$pullAndUpdateJobStatus$19(JobManager.java:616)] - Updated the job job-2097194451929410612 with execution id local-job-f340c4d9-8ef0-4f44-a63e-c46740e54246 status to FAILED
> ```
> 
> For the escaping issues, seems we didn't need to do special handing, as we use `ProcessBuilder` not a `shell` to execute Spark command.

For the 1st issue, I can package the api jar into the shading packaging. 

For the 2nd problem, should I shade the Iceberg jar into this jar? If so, it will be bound to the specific Iceberg/Spark version. So my initial thought is not to put the Iceberg dependency into this job jar. It can be handled by users. What's your opinion?

For the 3rd problem, the frequency of pulling the job state will hurt the performance, so I don't set a very short value by default. You can change the configuration. And for a production job, normally the job run time should not be very short, so I think it would not be a big problem in the real workload.

FANNG1:
> For the 2nd problem, should I shade the Iceberg jar into this jar? If so, it will be bound to the specific Iceberg/Spark version. So my initial thought is not to put the Iceberg dependency into this job jar. It can be handled by users. What's your opinion?

+1 for not shading the Iceberg runtime jar,  besides place Iceberg runtime jar to spark classpath, should we provide a way to allow the user to specify the runtime jar path?

> For the 3rd problem, the frequency of pulling the job state will hurt the performance, so I don't set a very short value by default. You can change the configuration. And for a production job, normally the job run time should not be very short, so I think it would not be a big problem in the real workload.

The default time is 5 minutes. do you think is it nesseary to shorten it to 10 or 20 seconds to balance performance and user experience?

jerryshao:
> > For the 2nd problem, should I shade the Iceberg jar into this jar? If so, it will be bound to the specific Iceberg/Spark version. So my initial thought is not to put the Iceberg dependency into this job jar. It can be handled by users. What's your opinion?
> 
> +1 for not shading the Iceberg runtime jar, besides place Iceberg runtime jar to spark classpath, should we provide a way to allow the user to specify the runtime jar path?
> 
> > For the 3rd problem, the frequency of pulling the job state will hurt the performance, so I don't set a very short value by default. You can change the configuration. And for a production job, normally the job run time should not be very short, so I think it would not be a big problem in the real workload.
> 
> The default time is 5 minutes. do you think is it nesseary to shorten it to 10 or 20 seconds to balance performance and user experience?

Let me think of this.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
FANNG1: Could you create a separate issue to track this?

FANNG1:
I didn't have any other comments

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergRewriteDataFilesJob.java
jerryshao: I've created an issue to track this.

FANNG1 (approved):

Issue #9520, opened 2025-12-22
[Improvement] Redefine the concept of `createEmptyTable` for Lance REST table


yuqi1129:
### What would you like to be improved?

According to the description of `https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table`, `createEmpty` table only handles metadata operation, which is different from the current implementation. `CreateEmptyTable` in Gravitino Lance REST is create a table without columns.

This API is used when creating a Spark table, please see: 

https://github.com/lance-format/lance-spark/blob/73d10eadd44fe842bc0bff89045d318a741a6d7e/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java#L394-L428

According to the code above, `createEmptyTable` only marks that the metadata of it exists and has a location name for the table.

yuqi1129:
@jerryshao @mchades 

According to the latest documentation, https://github.com/lance-format/lance-namespace/blob/main/java/lance-namespace-apache-client/docs/CreateEmptyTableRequest.md, `createEmptyTable` will only add metadata and create a placeholder indicating that the table exists, without modifying/creating data storage. This differs significantly from our implementation.  I have tested the API with lance-spark(the main branch has just supported the REST namespace), and it did what the docs described. 

Another point is that the location is optional for this interface, and it's the responsibility of the specific implementation that chooses the location if it is not provided by users. So my question is

- Should we keep the pave with it? What concerns me is that 
> It's the responsibility of the specific implementation that chooses the location if it is not provided by users.

This behaviour is completely targeted for cloud storage and service scenarios, and it's beyond our ability as we cannot provide storage service for users. This seems to be specifically for the Lance cloud service. I would like to move it forward to support Spark, ray without this limitation as the current Lance REST service can't work with them, but the current situation hinders me.  

Do you have any thoughts on it?

yuqi1129:
This is what lance-spark use `createEmptyTable` and the problem:

https://github.com/lance-format/lance-spark/issues/158

jerryshao:
Please sync with @mchades offline to figure out a solution.

PR #9589, opened 2025-12-31
[#9520] improvement(lance): Refine the concept of createEmptyTable in Lance REST

yuqi1129:

This pull request introduces support for creating "empty" tables in the Lance-backed catalog, where only metadata is stored without creating any underlying storage or dataset. The implementation includes changes to the REST API, backend logic, constants, and integration tests to align with the new specification for empty table creation.

To keep updated with the latest docs

Fix: #9520 

N/A

Existing test.

### DIFF
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
@@ -37,6 +37,7 @@ public class LanceConstants {
 
   public static final String LANCE_TABLE_REGISTER = "lance.register";
 
+  // Mark whether it is to create an empty Lance table(no data files)
   public static final String LANCE_TABLE_CREATE_EMPTY = "lance.create-empty";
 
   public static final String LANCE_TABLE_FORMAT = "lance";
--- a/lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
+++ b/lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
@@ -20,7 +20,6 @@ package org.apache.gravitino.lance.service.rest;
 
 import static org.apache.gravitino.lance.common.ops.NamespaceWrapper.NAMESPACE_DELIMITER_DEFAULT;
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_LOCATION;
-import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_CREATE_EMPTY;
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_LOCATION_HEADER;
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_PROPERTIES_PREFIX_HEADER;
 
@@ -146,7 +145,6 @@ public class LanceTableOperations {
               ? Maps.newHashMap()
               : Maps.newHashMap(request.getProperties());
 
-      props.put(LANCE_TABLE_CREATE_EMPTY, "true");
       CreateEmptyTableResponse response =
           lanceNamespace.asTableOps().createEmptyTable(tableId, delimiter, tableLocation, props);
       return Response.ok(response).build();
--- a/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
+++ b/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
@@ -77,6 +77,8 @@ import org.apache.gravitino.exceptions.NoSuchTableException;
 import org.apache.gravitino.integration.test.util.BaseIT;
 import org.apache.gravitino.integration.test.util.GravitinoITUtils;
 import org.apache.gravitino.lance.common.utils.ArrowUtils;
+import org.apache.gravitino.lance.common.utils.LanceConstants;
+import org.apache.gravitino.rel.Table;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
@@ -433,6 +435,9 @@ public class LanceRESTServiceIT extends BaseIT {
     DescribeTableResponse loadTable = ns.describeTable(describeTableRequest);
     Assertions.assertNotNull(loadTable);
     Assertions.assertEquals(location, loadTable.getLocation());
+    Assertions.assertEquals(
+        "true", loadTable.getProperties().get(LanceConstants.LANCE_TABLE_CREATE_EMPTY));
+    Assertions.assertEquals("true", loadTable.getProperties().get(Table.PROPERTY_EXTERNAL));
 
     // Try to create the same table again should fail
     LanceNamespaceException exception =
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java
@@ -20,6 +20,7 @@ package org.apache.gravitino.catalog.lakehouse.lance;
 
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_CREATION_MODE;
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_STORAGE_OPTIONS_PREFIX;
+import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_CREATE_EMPTY;
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_FORMAT;
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_REGISTER;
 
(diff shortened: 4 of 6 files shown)


catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -263,6 +264,18 @@ Table createTableInternal(
           ident, columns, comment, properties, partitions, distribution, sortOrders, indexes);
     }
 
+    // Check whether it's a create empty table operation.
+    boolean createEmpty =
+        Optional.ofNullable(properties.get(LanceConstants.LANCE_TABLE_CREATE_EMPTY))
```
yuqi1129, line 269: Should we reuse the property `register` without introducing it? It seems that they belong to different concepts.

jerryshao:
@mchades can you please help to review?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
```diff
@@ -37,5 +37,7 @@ public class LanceConstants {
 
   public static final String LANCE_TABLE_REGISTER = "lance.register";
 
+  public static final String LANCE_TABLE_CREATE_EMPTY = "lance.create-empty";
```
mchades, line 41: Do you need to add this property definition to the `GenericTablePropertiesMetadata`?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -122,6 +123,10 @@ public Response createTable(
     }
   }
 
+  /**
+   * According to the latest spec, createEmptyTable only stores the table metadata including its
+   * location, and will never touch lance storage.
```
mchades: You are addressing an issue in a fixed Lance namespace version; therefore, you cannot refer to "According to the latest spec."

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -446,19 +445,20 @@ void testCreateEmptyTable() throws ApiException {
 
     // Try to create a table with wrong location should fail
```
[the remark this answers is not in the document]
mchades: Why not address this comment?

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -263,6 +264,18 @@ Table createTableInternal(
...
+            .map(Boolean::parseBoolean)
+            .orElse(false);
+    if (createEmpty) {
+      // For create empty table, we just create the table metadata in Gravitino without creating
+      // the underlying Lance dataset.
+      return super.createTable(
+          ident, columns, comment, properties, partitions, distribution, sortOrders, indexes);
```
mchades, line 276: Is `tableLocation` required for `createEmptyTable`?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
yuqi1129: This is also a workaround like `lance.register`, so I prefer not a add it in `GenericTablePropertiesMetadata`. Another point is that this is only for Lance.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: In fact, we do not have old docs. Let me add a commit link.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: `tableLocation` is not required for users who call the API, but it's required for the server. If it's not set by users, the implementation should decide it.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
mchades: if then, will this property value allow users to modify it?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -124,8 +124,8 @@ public Response createTable(
   }
 
   /**
-   * According to the latest spec, createEmptyTable only stores the table metadata including its
-   * location, and will never touch lance storage.
+   * According to the spec of lance-namespace with version 0.0.20 to 0.31, createEmptyTable only
```
mchades, line 126: Is there a link to attach here?

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
mchades: Is it required by Gravitino server?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
yuqi1129: It should be immutable, and currently, users can't modify it by the Lance REST API; still, they can change it using by Gravitino REST API directly if they want. Ideally, we should forbid users from modifying a table created by the Lance REST.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: There are no docs about it actually.
image


The doc is from `lancedb`, which is not suitable for open-source project here. 
image

https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Tables in `lakehouse-generic` always have a location. What do you mean here?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
mchades: > they can change it using by Gravitino REST API directly if they want.

I think it should also be prohibited to modify through Gravitino REST API, so my suggestion is add it the `GenericTablePropertiesMetadata` or `LanceTablePropertiesMetadata`

yuqi1129: Let check register and create-empty both in `LanceTablePropertiesMetadata`.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
mchades: If neither the schema nor the catalog specifies a location, then `createEmptyTable` request must include a `location`; otherwise, an error will occur. Do I understand this correctly?

yuqi1129: Yes, the same goes for fileset.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
yuqi1129: fix

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
mchades: will the properties be null?

yuqi1129: It's a defensive action and just copy the code from 

```
    boolean register =
        Optional.ofNullable(properties.get(LANCE_TABLE_REGISTER))
            .map(Boolean::parseBoolean)
            .orElse(false);
```

yuqi1129: https://github.com/apache/gravitino/blob/036dd97a4105d0c9640bdc555754bad7bccc1794/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java#L120-L123

Issue #9076, opened 2025-11-11
[Improvement] Improve validation in SetModelVersionProperty.java


justinmclean:
### What would you like to be improved?

Prevent the SetModelVersionProperty CLI command from accepting both an alias and version (or neither), ensuring it fails instead of invoking the server with invalid inputs. Include unit tests, for example:
```
  @Test
  void validateBothAliasAndVersion() {
    CommandLine mockCmdLine = Mockito.mock(CommandLine.class);
    CommandContext context = new CommandContext(mockCmdLine);

    SetModelVersionProperty command =
        new SetModelVersionProperty(
            context,
            "metalake1",
            "catalog1",
            "schema1",
            "model1",
            1,
            "alias1",
            "prop",
            "value");

   Assertions.assertThrows(RuntimeException.class, command::validate);
  }
```

### How should we improve?

Improve validation and add unit tests, note you'll need to do this ` Main.useExit = false;`.

NazaninAnsarii:
hi there
i want to get into it

justinmclean:
It's all yours @NazaninAnsarii to work on.

NazaninAnsarii:
on it @justinmclean

NazaninAnsarii:
Hi @justinmclean,
I hope you’re doing well!
I’ve just submitted a pull request for your review. Please take a look and let me know if you have any feedback or if there are any changes needed.
Thank you!

PR #9656, opened 2026-01-08
[#9076] Improvement(cli): validation in SetModelVersionProperty.java

procura-interna:

Prevent the SetModelVersionProperty CLI command from accepting both an alias and version (or neither), ensuring it fails instead of invoking the server with invalid inputs.

Fix: #9076

After calling the constructor of SetModelVersionProperty with either both alias and version as null or alias and version as non-null, an exception will be thrown by .validate()

Added test class: SetModelVersionPropertyTest
Ran tests: ./gradlew :clients:test


justinmclean:
The CI is failing here because the test file is missing an ASF header @procura-interna do you still want to work on this?

procura-interna:
@justinmclean yes, I will look into it

Issue #9666, opened 2026-01-09
[Bug report] Iceberg migrate procedure fails with "table already exists" for JDBC catalogs


bharos:
### Version

main branch

### Describe what's wrong

The Iceberg `system.migrate()` procedure fails when attempting to migrate external tables to JDBC-backed catalogs (MySQL, PostgreSQL) with error "table already exists in database". 

### Error message and/or stacktrace

org.apache.iceberg.exceptions.AlreadyExistsException: Table already exists in database: test_database.migrate_test_123
  at org.apache.iceberg.jdbc.JdbcCatalog.createTable(JdbcCatalog.java:...)
  at org.apache.iceberg.rest.CatalogHandlers.stageTableCreate(CatalogHandlers.java:...)

### How to reproduce

1. Configure Gravitino with JDBC catalog (MySQL or PostgreSQL backend)
2. Create external Iceberg table in Hive or another catalog
3. Attempt to migrate using: `CALL system.migrate('source_catalog.db.table')`
4. Observe "table already exists" error

Expected: Migration should succeed using 3-phase staging protocol
Actual: Fails because JDBC catalog commits table prematurely

bharos:
Planning to issue a PR with:

### Fix to Preserve `stageCreate` flag in IcebergTableOperationExecutor
**File:** `core/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/ops/IcebergTableOperationExecutor.java`

When the executor rebuilds a `CreateTableRequest` to override the owner, it should now preserves the `stageCreate=true` flag:

```java
if (createTableRequest.stageCreate()) {
  builder.stageCreate();  // Preserve staging flag
}
```

PR #9667, opened 2026-01-09
[#9666]fix(iceberg): Fix Iceberg migrate procedure by preserving stageCreate flag

bharos:

This PR fixes the Iceberg system.migrate() procedure for JDBC-backed catalogs by preserving stageCreate flag in IcebergTableOperationExecutor

The system.migrate() procedure fails when migrating external Iceberg tables to JDBC-backed catalogs (MySQL, PostgreSQL) with the error:
```
org.apache.iceberg.exceptions.AlreadyExistsException: Table already exists in database
```

Root Cause:
The staging protocol requires 3 distinct phases to safely migrate tables with existing data:

Stage: Create table metadata without registering in the catalog
Write: Client writes/copies data files to the staged location
Commit: Atomically register the table with assert-create to prevent race conditions
JDBC catalogs skip phase 1 and immediately commit the table to the database, causing the migration to fail because the table already exists when phase 3 attempts the final commit.

This is a critical bug that makes it impossible to migrate existing Iceberg tables to Gravitino JDBC catalogs.

Fix: #9666 

No user-facing API changes. This fix enables the system.migrate() procedure to work correctly with JDBC catalogs, which was previously broken.

- Added unit test

Manual Testing:

Configured Gravitino with JDBC catalog (MySQL backend)
Created external Iceberg table in Hive catalog
Successfully migrated table using CALL system.migrate('catalog.database.table')
Verified table metadata and data files are correctly staged and committed
Confirmed migration fails without the fix (table already exists error)


roryqi:
Thanks, merged to branch main & 1.1

Issue #9616, opened 2026-01-04
[Improvement] Cannot add new users after enabling access control on server — blocks system upgrade


jerqi:
### What would you like to be improved?

Cannot add new users after enabling access control on server — blocks system upgrade

### How should we improve?

Avoid checking user existed in the metalake.

PR #9691, opened 2026-01-12
[#9616] fix(authz): Fix the  `PassThroughAuthorizer` user verification logic

roryqi:

Fix the  `PassThroughAuthorizer` user verification logic

Fix: #9616

No.

Added IT.


server-common/src/main/java/org/apache/gravitino/server/authorization/PassThroughAuthorizer.java
```diff
@@ -76,6 +80,14 @@ public boolean isSelf(Entity.EntityType type, NameIdentifier nameIdentifier) {
 
   @Override
   public boolean isMetalakeUser(String metalake) {
+    AccessControlDispatcher dispatcher = GravitinoEnv.getInstance().accessControlDispatcher();
+    if (dispatcher != null) {
+      try {
+        dispatcher.getUser(metalake, PrincipalUtils.getCurrentUserName());
+      } catch (NoSuchUserException e) {
+        dispatcher.addUser(metalake, PrincipalUtils.getCurrentUserName());
```
hdygxsj, line 88: Why is it necessary to add the user here?

If authentication and authorization are not enabled, could there be any issues here?

docs/security/access-control.md
```diff
@@ -455,6 +455,73 @@ gravitino.authorization.enable = true
 gravitino.authorization.serviceAdmins = admin1,admin2
 ```
 
+## Migration Guide
+
+If you have metalakes that were created before authorization was enabled, you need to perform a migration to ensure proper access control.
```
hdygxsj, line 460: Should the affected versions be added here?

server-common/src/main/java/org/apache/gravitino/server/authorization/PassThroughAuthorizer.java
roryqi: It will throw forbidden exceptions which user doesn't exist in the metalake. If authentication and authorization are not enabled. This method won't be called.

hdygxsj: If authentication and authorization are not enabled. This method won't be called.  ——— If authorization isn't enabled, doesn't it just use the PassThroughAuthorizer?
image

docs/security/access-control.md
roryqi: Every version should migrate.

server-common/src/main/java/org/apache/gravitino/server/authorization/PassThroughAuthorizer.java
roryqi: If authorization enable is false, AccessControlDispatcher will be null.

hdygxsj (approved):
LGTM

Issue #9715, opened 2026-01-14
[Improvement] Add the documents for Ray using IRC


roryqi:
### What would you like to be improved?

 Add the documents for Ray using IRC

### How should we improve?

 Add the documents for Ray using IRC

PR #9716, opened 2026-01-14
[#9715] improvement(docs): Add the documents for Ray using IRC

roryqi:

Add the documents for Ray using IRC

Fix: #9715

Just documents.

No.

### DIFF
--- a/docs/iceberg-rest-service.md
+++ b/docs/iceberg-rest-service.md
@@ -657,27 +657,6 @@ print(table.scan().to_arrow())
 Ray Data only supports reading from and writing to existing Iceberg tables. It does not support DDL operations such as creating, dropping, or altering tables, schemas, or catalogs. You need to use other tools like Spark or PyIceberg to manage table metadata.
 :::
 
-#### Reading Iceberg tables with Ray
-
-```python
-import ray
-
-ds = ray.data.read_iceberg(
-    table_identifier="default.sample",
-    catalog_kwargs={
-        "name": "default",
-        "type": "rest",
-        "uri": "http://localhost:9001/iceberg/",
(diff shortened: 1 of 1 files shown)


roryqi:
@FANNG1 Could u take a look at this pull request?

docs/iceberg-rest-service.md
```diff
@@ -649,6 +649,58 @@ table = catalog.load_table(table_identifier)
...
+
+#### Reading Iceberg tables with Ray
+
+```python
+import ray
+
+ds = ray.data.read_iceberg(
+    table_identifier="default.sample",
+    catalog_kwargs={
+        "name": "default",
```
FANNG1, line 671: what does the name mean?

docs/iceberg-rest-service.md
```diff
@@ -649,6 +649,58 @@ table = catalog.load_table(table_identifier)
...
+        "auth": {
+            "type": "basic",
+            "basic": {"username": "manager", "password": "mock"}
+        }
+    }
+)
+ds.show(limit=1)
+```
+
+#### Writing to Iceberg tables with Ray
```
FANNG1: How about writing data first, then read the data?

docs/iceberg-rest-service.md
roryqi: It's related about pyIceberg. `~/.pyIceberg.yaml` will have default configurations for every catalog. The catalog will have the name. So we pass the name here, we can use the name to decide to use which configurations.

docs/iceberg-rest-service.md
roryqi: OK for me.

Issue #9713, opened 2026-01-14
[Bug report] "serverUri", "tokenPath" is asked for configuration when oauth provider is oidc


danhuawang:
### Version

main branch

### Describe what's wrong

I follow doc to config Azure AD.
It says `With JWKS validation, you don't need to configure defaultSignKey, serverUri, or tokenPath as the validator dynamically fetches public keys from Azure AD's JWKS endpoint.`
https://gravitino.apache.org/docs/1.1.0/security/how-to-authenticate#example-azure-ad-as-oidc-provider-with-jwks-validation 

But if I didn't configure "serverUri", "tokenPath", the server will have IllegalArgumentException.

### Error message and/or stacktrace

```
2026-01-14 08:29:52.223 INFO [main] [org.apache.gravitino.metrics.MetricsSystem.register(MetricsSystem.java:79)] - Register iceberg-rest-server to metrics system
2026-01-14 08:29:52.229 INFO [main] [org.apache.gravitino.iceberg.service.provider.IcebergConfigProviderFactory.create(IcebergConfigProviderFactory.java:44)] - Load Iceberg catalog provider: org.apache.gravitino.iceberg.service.provider.StaticIcebergConfigProvider.
2026-01-14 08:29:52.307 INFO [main] [org.apache.gravitino.iceberg.service.metrics.IcebergMetricsManager.loadIcebergMetricsStore(IcebergMetricsManager.java:194)] - Load Iceberg metrics store: org.apache.gravitino.iceberg.service.metrics.DummyMetricsStore.
2026-01-14 08:29:52.364 INFO [main] [org.apache.gravitino.iceberg.RESTService.serviceInit(RESTService.java:162)] - Iceberg REST service init.
2026-01-14 08:29:52.411 INFO [main] [org.apache.gravitino.GravitinoEnv.initializeFullComponents(GravitinoEnv.java:198)] - Gravitino full environment is initialized.
2026-01-14 08:29:52.437 INFO [main] [org.apache.gravitino.server.web.JettyServer.initializeWebAppServletContextHandler(JettyServer.java:309)] - Gravitino Webapp path: /tmp/GravitinoWar6032372158821220371
2026-01-14 08:29:52.447 ERROR [main] [org.apache.gravitino.server.GravitinoServer.main(GravitinoServer.java:210)] - Error while running jettyServer
java.lang.IllegalArgumentException: null in gravitino.authenticator.oauth.tokenPath is invalid. The value can't be blank
[...]

PR #9717, opened 2026-01-14
[#9713] fix(oauth): Allow JWKS validators without serverUri or tokenPath

bharos:

Allow JWKS validators without serverUri or tokenPath as these are only needed in Static validator

Earlier the validation was done in base class which was not correct, as some implementations don't need these configs.

Fix: #9713

No

Added unit tests


roryqi:
Could u modify the document, too?

bharos:
> Could u modify the document, too?

Done, thanks

Issue #9733, opened 2026-01-16
[Improvement] Support JWT tokens with multiple audiences in aud claim


bharos:
### What would you like to be improved?

Currently the JWT token can only contain single audience. We can improve the JWKS token validator to support multiple audiences in token

### How should we improve?

Root cause: The validator uses Nimbus's exactMatchClaims which requires exact equality of the entire audience array (https://bitbucket.org/connect2id/nimbus-jose-jwt/src/9509dc54d154a6d8835ff96639297770de2a0cbf/src/main/java/com/nimbusds/jwt/proc/DefaultJWTClaimsVerifier.java#lines-176) , rather than RFC 7519's "at-least-one" match semantics.
https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3

Solution: Use the DefaultJWTClaimsVerifier constructor with the acceptedAudiences Set parameter (https://bitbucket.org/connect2id/nimbus-jose-jwt/src/9509dc54d154a6d8835ff96639297770de2a0cbf/src/main/java/com/nimbusds/jwt/proc/DefaultJWTClaimsVerifier.java#lines-204) , which properly implements RFC 7519 compliant audience validation (checking if the service's audience is present in the token's audience list).

This brings JwksTokenValidator's behavior in line with StaticSignKeyValidator, which already supports multi-audience tokens correctly.

PR #9734, opened 2026-01-16
[#9733]feat(oauth): Support tokens with multiple audiences in JWKS validator

bharos:
Modified `JwksTokenValidator` to use Nimbus's `acceptedAudiences` Set parameter instead of `exactMatchClaims` for audience validation. This enables proper RFC 7519 compliant multi-audience token support.
StaticSignKeyValidator already supports multiple audiences.

JwksTokenValidator currently rejects valid JWT tokens that contain multiple audiences (e.g., `["service-a", "service-b", "service-c"]`), even when the configured service audience is present in the list. This is because the validator uses `exactMatchClaims` which requires exact array equality rather than "at-least-one match" semantics defined in RFC 7519.

Fix: #9733

Yes - JWT tokens with multiple audiences in the `aud` claim are now properly validated when the configured service audience is present in the token's audience list. Previously only single-audience tokens were supported.

Added `testValidateTokenWithMultipleAudiences()` that verifies:
1. Token with `["other-service", "test-service", "another-service"]` validates successfully against `"test-service"`
2. Same token fails validation against `"incorrect-service"`

All existing OAuth authentication tests pass.


jerqi (approved):
LGTM.

Issue #9561, opened 2025-12-25
[Subtask] Add UDF management core operations


mchades:
### Describe the subtask

Add UDF management core operations

### Parent issue

#6158

PR #9737, opened 2026-01-16
[#9561] feat(core): Implement alterFunction for managed UDF operations

mchades:

Implement the `alterFunction` method in `ManagedFunctionOperations` to support modifying registered UDFs. This includes:

- **UpdateComment**: Update the function's comment/description
- **AddDefinition**: Add a new function definition with arity overlap validation
- **RemoveDefinition**: Remove an existing definition (with guard against removing the only definition)
- **AddImpl**: Add a new implementation to a definition (with runtime uniqueness check)
- **UpdateImpl**: Update an existing implementation for a specific runtime
- **RemoveImpl**: Remove an implementation from a definition (with guard against removing the only implementation)

Fix: #9561

This is part of the UDF management feature. After registering a function, users need the ability to modify it without dropping and recreating. The `alterFunction` API enables incremental updates to function metadata, definitions, and implementations.

Yes. Users can now use the `alterFunction` API to modify registered functions through various `FunctionChange` operations.

Added comprehensive unit tests in `TestManagedFunctionOperations`:
- Test updating comment
- Test adding/removing definitions with arity validation
- Test adding/updating/removing implementations
- Test error cases (non-existing function, non-existing runtime, removing only definition/impl)
- Test parameter order validation during alter operations

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
@@ -440,6 +440,11 @@ public class ManagedFunctionOperations implements FunctionCatalog {
    */
   private void validateRemoveDefinition(
       List<FunctionDefinition> definitions, FunctionParam[] paramsToRemove) {
+    if (definitions.size() == 1) {
+      throw new IllegalArgumentException(
+          "Cannot remove the only definition. Use dropFunction to remove the entire function.");
+    }
+
     boolean found = false;
     for (FunctionDefinition def : definitions) {
       if (parametersMatch(def.parameters(), paramsToRemove)) {
@@ -451,11 +456,6 @@ public class ManagedFunctionOperations implements FunctionCatalog {
     if (!found) {
       throw new IllegalArgumentException(
           "Cannot remove definition: no definition found with the specified parameters");
-    }
-
-    if (definitions.size() == 1) {
-      throw new IllegalArgumentException(
-          "Cannot remove the only definition. Use dropFunction to remove the entire function.");
     }
   }
 


core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -245,9 +260,110 @@ private void validateDefinitionsNoArityOverlap(FunctionDefinition[] definitions)
...
+    String newComment = oldEntity.comment();
+    List<FunctionDefinition> newDefinitions =
+        new ArrayList<>(Arrays.asList(oldEntity.definitions()));
+
+    for (FunctionChange change : changes) {
+      if (change instanceof FunctionChange.UpdateComment) {
+        newComment = ((FunctionChange.UpdateComment) change).newComment();
+
+      } else if (change instanceof FunctionChange.AddDefinition) {
+        FunctionDefinition defToAdd = ((FunctionChange.AddDefinition) change).definition();
```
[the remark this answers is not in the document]
mchades: The validation already happens via the existing call chain: 
validateNoArityOverlap → computeArities → findFirstOptionalParamIndex. 

Adding an explicit call would be redundant.

core/src/test/java/org/apache/gravitino/catalog/TestManagedFunctionOperations.java
```diff
@@ -388,6 +477,263 @@ public void testOverlappingAritiesWithDifferentTypes() {
...
+
+    // Update Spark implementation
+    FunctionImpl newSparkImpl =
+        FunctionImpls.ofJava(FunctionImpl.RuntimeType.SPARK, "com.example.NewSparkUDF");
+    org.apache.gravitino.function.Function updatedFunc =
+        functionOperations.alterFunction(
+            funcIdent,
+            FunctionChange.updateImpl(params, FunctionImpl.RuntimeType.SPARK, newSparkImpl));
+
+    Assertions.assertEquals(1, updatedFunc.definitions()[0].impls().length);
```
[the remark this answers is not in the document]
mchades: fixed

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -313,4 +430,168 @@ private boolean hasDefaultValue(FunctionParam param) {
...
+        break;
+      }
+    }
+
+    if (!found) {
+      throw new IllegalArgumentException(
+          "Cannot remove definition: no definition found with the specified parameters");
+    }
+
+    if (definitions.size() == 1) {
```
jerryshao: We can check this before the loop, so that we don't have to execute the additional loop.

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
```diff
@@ -245,9 +260,111 @@ private void validateDefinitionsNoArityOverlap(FunctionDefinition[] definitions)
...
+      } else if (change instanceof FunctionChange.RemoveDefinition) {
+        FunctionParam[] paramsToRemove = ((FunctionChange.RemoveDefinition) change).parameters();
+        validateRemoveDefinition(newDefinitions, paramsToRemove);
+        newDefinitions.removeIf(def -> parametersMatch(def.parameters(), paramsToRemove));
+
+      } else if (change instanceof FunctionChange.AddImpl) {
+        FunctionChange.AddImpl addImpl = (FunctionChange.AddImpl) change;
+        FunctionParam[] targetParams = addImpl.parameters();
+        FunctionImpl implToAdd = addImpl.implementation();
+        newDefinitions = addImplToDefinition(newDefinitions, targetParams, implToAdd);
```
jerryshao, line 286: Should you validate the parameters when you add a new impl?

mchades: the validation logic is included in the method `addImplToDefinition`

core/src/main/java/org/apache/gravitino/catalog/ManagedFunctionOperations.java
mchades: fixed

jerryshao (approved):

PR #9739, opened 2026-01-17
[#9713][followup] fix(oauth): Remove checkValue for optional serverUri and tokenPath

bharos:

Remove .checkValue() validation from DEFAULT_SERVER_URI and DEFAULT_TOKEN_PATH config entries in OAuthConfig.java.

The original fix [PR 9717](https://github.com/apache/gravitino/pull/9717) moved validation to StaticSignKeyValidator but left .checkValue() in the config definitions. This causes ConfigServlet to fail when reading these configs for the Web UI, even when using JWKS validators that don't need them.

Fix: #9713

No.

- Existing tests work
- Tested that Gravitino server loads now without these configs


jerqi (dismissed):

bharos:
@jerqi I added a unit test which would have caught this issue. PTAL, thanks

Issue #9758, opened 2026-01-20
[FEATURE] gravitino web UI reconstruct by ant-design


LauraXia123:
### Describe the feature

1. web UI reconstruct by ant-design
2. Retain the old UI, which can be switched based on environment variables.

PR #9759, opened 2026-01-20
[#9758] feat(UI): New UI reconstruct base main

LauraXia123:

1. web UI reconstruct by ant-design
2. Retain the old UI, which can be switched based on environment variables.

Fix: #9758

N/A

manually


web/web/src/app/catalogs/rightContent/entitiesContent/DataPreview.js
```diff
@@ -0,0 +1,136 @@
...
+        <Table
+          columns={columns}
+          dataSource={tableData}
+          style={{ maxHeight: 'calc(100vh - 28rem)' }}
+          scroll={{ y: 'calc(100vh - 34rem)' }}
+        />
+      )}
+      {previewData?.statusCode === 1 && (
+        <Alert
+          message={t('common.previewWarning', { table: `${catalog}.${schema}.${table}` })}
```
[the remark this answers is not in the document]
LauraXia123: need remove DataPreview component

web/web/src/app/catalogs/rightContent/entitiesContent/DataPreview.js
```diff
@@ -0,0 +1,136 @@
...
+  const tableData =
+    previewData?.results?.map((col, index) => {
+      return {
+        ...col,
+        key: index
+      }
+    }) || []
+
+  return (
+    <Spin spinning={isLoading}>
```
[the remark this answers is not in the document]
LauraXia123: need remove DataPreview component

web/web/src/app/catalogs/rightContent/entitiesContent/DataPreview.js
```diff
@@ -0,0 +1,136 @@
...
+        case 'list':
+          return `[${value.map(v => getPreviewData(type?.elementType, v)).join(', ')}]`
+        default:
+          return '-'
+      }
+    }
+  }
+
+  const columns =
+    previewData?.columns?.map(col => {
```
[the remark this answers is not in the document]
LauraXia123: need remove DataPreview component

danhuawang:
I verify the fixed bugs based on this PR --passed.

LauraXia123:
@jerryshao pls review the pr if u have time, thx.

conf/gravitino.conf.template
```diff
@@ -78,6 +78,7 @@ gravitino.cache.implementation = caffeine
 # Whether Gravitino enable authorization or not
 gravitino.authorization.enable = false
 # The admins of Gravitino service, multiple admins are spitted by comma.
+gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins
```
jerryshao, line 81: @jerqi what is the usage of this configuration?

bin/gravitino.sh.template
```diff
@@ -181,6 +181,31 @@ fi
 
 addJarInDir "${GRAVITINO_HOME}/libs"
 
+# By default, prefer starting the legacy (old) UI war if present.
+# Users may override by setting GRAVITINO_WAR to an explicit path, or
+# set GRAVITINO_USE_OLD_UI=false to prefer the non-old war when both exist.
+if [[ -z "${GRAVITINO_WAR+x}" || -z "${GRAVITINO_WAR}" ]]; then
+  # Compose expected war paths based on packaged layout and version
+  OLD_WAR_PATH="${GRAVITINO_HOME}/web/gravitino-web-1.1.0-old.war"
+  NEW_WAR_PATH="${GRAVITINO_HOME}/web/gravitino-web-${GRAVITINO_VERSION}.war"
```
jerryshao: Ideally, we should not use old/new to differentiate the packages; we should have a version information to distinguish. I would suggest to use v1 as old UI, v2 as a new UI. Can you change all the varaibles, package name, folder names.

jerryshao:
Ideally, what I want is two different modules, like web-v1, web-v2, to separate things clearly. If we want to remove the V1 UI, we can simply remove the folder. The current implementation mixed everything together, which is hard to maintain.

conf/gravitino.conf.template
roryqi: If UI wanna to know some configurations, it could be put it in this configurations. It can retrive config options and display them in the ConfigServelet.

LauraXia123: Only serviceAdmins have privilege for 'Create Metalake', I get the info to show or hide the btn

LauraXia123:
1. rename all the varaibles, package name from 'old ui' to 'web-v1'
2. update all readme, script content about web v1
3. rename folder `appOld` to `appV1`, `componentsOld` to `componentsV1`, If we want to remove the V1 UI, remove the two folders almost is ok.
4. The app folder is the default entry point for Next.js projects, so it was not renamed to V2. The corresponding components are UI components for V2, while the rest, such as the model layer, libs, and configs, are shared components. After deleting V1, the redundant code generated will not exceed 10%.
5. I have test web v1 and web v2 is worked after change.

jerryshao:
No, I don't think this is what I want. What I want is a new module compared to current web module, a new one called web-v2, this new module is totally a new module (nothing shared with existing web module), and the duplication is OK. When we deprecate the old web module, we can simply drop the folder.

LauraXia123:
1. Actually, the current solution consists of two independent UI modules, which share some package management files and style configuration management while supporting both UIs. However, the final output is two separate and independent WAR packages. 
2. There is also another solution: splitting them into two independent projects. If we plan to maintain multiple versions of the UI in the long term, it is recommended to switch to two independent projects. 
If you need to proceed with this change, please let me know as soon as possible.

LauraXia123:
1. I have rescheduled web & web-v2 project
2. I have test web UI & web-v2 UI after rebuild

bin/gravitino.sh.template
```diff
@@ -181,6 +181,31 @@ fi
 
 addJarInDir "${GRAVITINO_HOME}/libs"
 
+# By default, prefer starting the legacy (old) UI war if present.
+# Users may override by setting GRAVITINO_WAR to an explicit path, or
+# set GRAVITINO_USE_WEB_V2=true to prefer the new war when both exist.
+if [[ -z "${GRAVITINO_WAR+x}" || -z "${GRAVITINO_WAR}" ]]; then
+  # Compose expected war paths based on packaged layout and version
+  WAR_V1_PATH="${GRAVITINO_HOME}/web/gravitino-web-v1.war"
```
jerryshao: Why the v1 war package doesn't have the "version" in its path?

LauraXia123: the latest commit have changed the v1 path name

settings.gradle.kts
```diff
@@ -82,7 +78,7 @@ project(":spark-connector:spark-3.4").projectDir = file("spark-connector/v3.4/sp
 project(":spark-connector:spark-runtime-3.4").projectDir = file("spark-connector/v3.4/spark-runtime")
 project(":spark-connector:spark-3.5").projectDir = file("spark-connector/v3.5/spark")
 project(":spark-connector:spark-runtime-3.5").projectDir = file("spark-connector/v3.5/spark-runtime")
-include("web:web", "web:integration-test")
+include("web:web", "web-v2:web", "web:integration-test")
```
jerryshao: Shall we include "web-v2:integration-test" here?

jerryshao (approved):

Issue #9760, opened 2026-01-20
[Improvement] add a configuration to disable client and server version check


FANNG1:
### What would you like to be improved?

Gravitino has strict client and server check,  in some senorios, we want to test new client version with old server and ,  it would better to add a configuration to disable client and server version check

PR #9763, opened 2026-01-21
[#9760] feat(client): disable version check by enviroment varibles

FANNG1:

disable version check by enviroment varibles

Fix: #9760

no

existsing tests

### DIFF
--- a/docs/how-to-use-gravitino-client.md
+++ b/docs/how-to-use-gravitino-client.md
@@ -72,10 +72,3 @@ gravitino_client = GravitinoClient(
 | `gravitino_client_request_timeout` | An optional client timeout in seconds. | `10`          | No       | 1.0.0         |
 
 **Note:** Invalid configuration properties will result in exceptions. 
-
-## Disabling client/server version checks
-
-By default, Gravitino clients verify that the client and server versions are compatible. If you
-need to bypass this check (for example, during a staged upgrade), set the environment variable
-`GRAVITINO_VERSION_CHECK_DISABLED` to `true` before starting your Java or Python client process.
-Any other value keeps the version check enabled.


jerryshao:
I don't have strong feelings, what's your opinion? @diqiu50 @mchades

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoClientBase.java
```diff
@@ -49,6 +49,8 @@ public abstract class GravitinoClientBase implements Closeable {
 
   /** The REST client to communicate with the REST server */
   protected final RESTClient restClient;
+  /** Environment variable to disable version checks. */
+  private static final String VERSION_CHECK_DISABLED_ENV = "GRAVITINO_VERSION_CHECK_DISABLED";
```
mchades, line 53: Should you update the user doc for this?

clients/client-java/src/test/java/org/apache/gravitino/client/TestVersionCheckDisabledEnv.java
```diff
@@ -0,0 +1,94 @@
...
+
+package org.apache.gravitino.client;
+
+import java.util.Collections;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class TestVersionCheckDisabledEnv {
+
```
diqiu50: I suggest moving the tests to TestGravitinoAdminClient.  All version-related tests  are there.  
Another option is to move all version-related tests to this class, which is also OK.

FANNG1: move all version-related tests to this class, as version version is not limited to Gravitino admin client

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoClientBase.java
FANNG1: sure, updated

FANNG1:
@mchades @diqiu50 comments are addressed, PTAL

clients/client-java/src/test/java/org/apache/gravitino/client/TestVersionCheck.java
```diff
@@ -0,0 +1,184 @@
...
+    String date = "2024-01-03 12:28:33";
+    String commitId = "6ef1f9d";
+
+    VersionResponse resp = new VersionResponse(new VersionDTO(version, date, commitId));
+    buildMockResource(Method.GET, "/api/version", null, resp, HttpStatus.SC_OK);
+
+    // check the client version is greater than server version
+    Assertions.assertThrows(GravitinoRuntimeException.class, () -> client.checkVersion());
+  }
+
```
diqiu50, line 182: Can we add a tester testSkipCheckVersionFailed like `testCheckVersionFailed`

clients/client-python/gravitino/client/gravitino_client_base.py
```diff
@@ -71,9 +73,13 @@ def __init__(
             request_headers=new_headers,
             client_config=client_config,
         )
-        if check_version:
+        if check_version and not self._is_version_check_disabled_by_env():
```
diqiu50, line 76: Why don‘t we use `check_version`  directly

FANNG1: it requires the client side to invode disable check version method explictly

clients/client-java/src/test/java/org/apache/gravitino/client/TestVersionCheck.java
FANNG1: add `testVersionCheckSkippedByEnv`

clients/client-python/gravitino/client/gravitino_client_base.py
diqiu50: Would it be better to initialize this variable in the class initialize

FANNG1: I prefer to keep current impl because `check_version` properties are set by the user, this is clearer

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoClientBase.java
FANNG1: removed document

clients/client-python/gravitino/client/gravitino_client_base.py
diqiu50: I think using fewer conditional variables to control this would be better, but your approach also works.

FANNG1: thx

Issue #9762, opened 2026-01-20
[Improvement] upgrade JDBC entity store scripts to prepare for 1.2.0 release


yuqi1129:
### What would you like to be improved?

Currently, the value of `CURRENT_SCRIPT_VERSION` is still `VERSION_1_1_0`. As we are going to move forward to 1.2.0, we'd better update the `CURRENT_SCRIPT_VERSION` to 1.2.0 in the `main` branch.

yuqi1129:
What you need to do is 
- Point `CURRENT_SCRIPT_VERSION`  to `V1.2.0`
- Add scripts of JDBC database(H2, MySQL, PostgreSQL) location at `scripts/*`, such as `schema-1.2.0-h2.sql` and `upgrade-1.1.0-to-1.2.0-h2.sql`

yuqi1129:
Not planed

PR #9764, opened 2026-01-21
[#9762] improvement(core): Upgrade `CURRENT_SCRIPT_VERSION` to 1.2.0

yuqi1129:

This pull request updates the system to recognize version 1.2.0 as the current backend storage script version and adds upgrade scripts for various databases, even though no schema changes are required for this version. It also cleans up test code to reference the version constant from a central location.

It's for 1.2.0 release.

Fix: #9762

N/A.

Existing tests.

### DIFF
--- a/scripts/h2/iceberg-metrics-schema-1.2.0-h2.sql
+++ b/scripts/h2/iceberg-metrics-schema-1.2.0-h2.sql
@@ -1,0 +1,90 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file--
+--  distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


jerryshao:
What is the reason of changing the SQL script to 1.2 even there's no change, is there anything needed to support?

Our assumption is that if there's no schema change, we don't have to bump the script version; this also follows HMS's way to maintain the schema. What is the specific reason to do it?

yuqi1129:
> What is the reason of changing the SQL script to 1.2 even there's no change, is there anything needed to support?
> 
> Our assumption is that if there's no schema change, we don't have to bump the script version; this also follows HMS's way to maintain the schema. What is the specific reason to do it?

Let's wait for whether https://github.com/apache/gravitino/pull/9779 will be reasonable and merged. Before doing this, I just hear about they have similar needs and weren't expected to raise the PR so quickly.

PR #9788, opened 2026-01-23
[ #9767]feat(oauth): Support Regex to have user mapping from principal

### LINKED ISSUE #9767: [Improvement] Allow regex pattern to filter the Gravitino OAuth principal

The principal from OAuth token (for example from Azure AD) contains the email like
username@company.com
and there might not be other fields that map to the username (without `@company.com`). It will be useful in general to support patterns to extract username from principal string.

### How should we improve?

We can add a parameter similar to what Trino server has:
https://trino.io/docs/current/security/oauth2.html#trino-server-configuration
`http-server.authentication.oauth2.user-mapping.pattern`

bharos:
- Core framework: PrincipalMapper interface, RegexPrincipalMapper implementation, PrincipalMapperFactory
- Kerberos utility: KerberosPrincipalParser for parsing primary[/instance][@REALM] format
- Authenticator integration: Modified KerberosAuthenticator, JwksTokenValidator, and StaticSignKeyValidator to apply principal mapping
- Configuration: Added 4 new properties (2 for OAuth, 2 for Kerberos) supporting regex patterns or custom mapper classes
- Documentation: Added principal mapping section with OAuth/Kerberos examples and custom mapper implementation guide

Organizations need to transform authenticated principals into different user identities for authorization:

OAuth: Extract username from email (user@example.com → user)
Kerberos: Extract primary from service principals (HTTP/server@REALM → HTTP)
Multi-realm: Route by realm (user@DEV.REALM → dev_user)

Fixes #9767 

New configuration properties (all optional, defaults maintain backward compatibility):

```
gravitino.authenticator.oauth.principalMapperType (default: regex)
gravitino.authenticator.oauth.principalMapperPattern (default: ^(.*)$)
gravitino.authenticator.kerberos.principalMapperType (default: regex)
gravitino.authenticator.kerberos.principalMapperPattern (default: ([^@]+).*)
```

Backward compatible: Because the default values preserve existing behavior.

- New test cases across 3 test classes: TestRegexPrincipalMapper, TestKerberosPrincipalParser, TestPrincipalMapperFactory
- Existing tests: All OAuth and Kerberos authenticator tests pass

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1114,7 +1114,6 @@ tasks {
       ":catalogs:catalog-hive:copyLibAndConfig",
       ":catalogs:catalog-jdbc-doris:copyLibAndConfig",
       ":catalogs:catalog-jdbc-mysql:copyLibAndConfig",
-      ":catalogs-contrib:catalog-jdbc-oceanbase:copyLibAndConfig",
       ":catalogs:catalog-jdbc-postgresql:copyLibAndConfig",
       ":catalogs:catalog-jdbc-starrocks:copyLibAndConfig",
       ":catalogs:catalog-kafka:copyLibAndConfig",
@@ -1124,7 +1123,9 @@ tasks {
       ":catalogs:catalog-model:copyLibAndConfig",
       ":catalogs:hive-metastore2-libs:copyLibs",
       ":catalogs:hive-metastore3-libs:copyLibs",
-      ":catalogs:catalog-lakehouse-generic:copyLibAndConfig"
+      ":catalogs:catalog-lakehouse-generic:copyLibAndConfig",
+      ":catalogs-contrib:catalog-jdbc-oceanbase:copyLibAndConfig",
+      ":catalogs-contrib:catalog-jdbc-clickhouse:copyLibAndConfig"
     )
   }
 
--- a/catalogs-contrib/catalog-jdbc-clickhouse/build.gradle.kts
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/build.gradle.kts
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+description = "catalog-jdbc-clickhouse"
+
+plugins {
+  `maven-publish`
+  id("java")
+  id("idea")
+}
+
+dependencies {
+  implementation(project(":api")) {
+    exclude(group = "*")
+  }
+  implementation(project(":catalogs:catalog-jdbc-common")) {
+    exclude(group = "*")
+  }
+  implementation(project(":common")) {
+    exclude(group = "*")
+  }
+  implementation(project(":core")) {
+    exclude(group = "*")
+  }
+
+  implementation(libs.bundles.log4j)
+  implementation(libs.commons.collections4)
+  implementation(libs.commons.lang3)
+  implementation(libs.guava)
+
+  testImplementation(project(":catalogs:catalog-jdbc-common", "testArtifacts"))
+  testImplementation(project(":clients:client-java"))
+  testImplementation(project(":integration-test-common", "testArtifacts"))
+  testImplementation(project(":server"))
+  testImplementation(project(":server-common"))
+
+  testImplementation(libs.awaitility)
+  testImplementation(libs.junit.jupiter.api)
+  testImplementation(libs.junit.jupiter.params)
+  testImplementation(libs.testcontainers)
+
+  testRuntimeOnly(libs.junit.jupiter.engine)
+}
+
+tasks {
+  register("runtimeJars", Copy::class) {
+    from(configurations.runtimeClasspath)
+    into("build/libs")
+  }
+
+  val copyCatalogLibs by registering(Copy::class) {
+    dependsOn("jar", "runtimeJars")
+    from("build/libs") {
+      exclude("guava-*.jar")
+      exclude("log4j-*.jar")
+      exclude("slf4j-*.jar")
+    }
+    into("$rootDir/distribution/package/catalogs/jdbc-clickhouse/libs")
+  }
+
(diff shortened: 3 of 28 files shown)


core/src/main/java/org/apache/gravitino/auth/KerberosPrincipalParser.java
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * <p>Examples:
+ *
+ * <ul>
+ *   <li>{@code john} - username only
+ *   <li>{@code john@EXAMPLE.COM} - username with realm
+ *   <li>{@code HTTP/server.example.com@EXAMPLE.COM} - service principal with instance and realm
+ * </ul>
+ */
+public class KerberosPrincipalParser implements Principal {
```
bharos: @jerqi @yangyuxia I added a class like this, but currently it is not used in the code. But the point is that users can create custom authenticators using this class easily (as I have added in documentation). LMK if it will be useful to keep this class.

roryqi: Why does this class need to implement the interface `Principal`? It seems weird for me.

roryqi:
Could u add issue id in the title?

bharos:
> Could u add issue id in the title?

Done

core/src/main/java/org/apache/gravitino/auth/KerberosPrincipalParser.java
bharos: Good catch, I had it from previous implementation no longer needed. Removed

server-common/src/main/java/org/apache/gravitino/server/authentication/StaticSignKeyValidator.java
```diff
@@ -97,7 +104,10 @@ public Principal validateToken(String token, String serviceAudience) {
         throw new UnauthorizedException(
             "Audiences in token is not in expected format: %s", audienceObject);
       }
-      return new UserPrincipal(jwt.getBody().getSubject());
+
+      // Use principal mapper to extract username
+      Principal mappedPrincipal = principalMapper.map(jwt.getBody().getSubject());
```
roryqi: Could u reuse the config options `OAuthConfig.PRINCIPAL_FIELDS` here? Now, it's a fixed field `sub`.

core/src/main/java/org/apache/gravitino/auth/KerberosPrincipalParser.java
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * <p>Examples:
+ *
+ * <ul>
+ *   <li>{@code john} - primary only
+ *   <li>{@code john@EXAMPLE.COM} - primary with realm
+ *   <li>{@code HTTP/server.example.com@EXAMPLE.COM} - service principal with instance and realm
+ * </ul>
+ */
+public class KerberosPrincipalParser {
```
roryqi: KerberosPrincipalParser mixed two concepts KerberosPrincipal and KerberosPrincipalMapper.
We should implement a KerberosPrincipalMapper and a KerberosPrincipal here.
```
class KerberosPrincipalMapper implement PrincipalMapper {

    Principal map(String principal) {
         return new KerberosPrincipal(xxx, xxxx, xxx);
    }
}

class KerberosPrincipal implement Principal {
  Optional<String> getName();
  Optional<String>  getRealm();
  Optional<String> getInstance();
}
```

server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosConfig.java
```diff
@@ -42,4 +42,26 @@ public interface KerberosConfig {
...
+          .stringConf()
+          .createWithDefault("regex");
+
+  ConfigEntry<String> PRINCIPAL_MAPPER_REGEX_PATTERN =
+      new ConfigBuilder(KERBEROS_CONFIG_PREFIX + "principalMapper.regex.pattern")
+          .doc(
+              "Regex pattern to extract the username from the Kerberos principal. "
+                  + "Only used when principalMapperType is 'regex'. "
+                  + "The pattern should contain at least one capturing group. "
+                  + "Default pattern '([^@]+).*' extracts everything before '@' (including instance if present).")
```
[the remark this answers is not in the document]
bharos: I think this is already mentioned in the doc here

core/src/main/java/org/apache/gravitino/auth/KerberosPrincipalParser.java
bharos: That's better. Done

server-common/src/main/java/org/apache/gravitino/server/authentication/StaticSignKeyValidator.java
bharos: Done

server-common/src/main/java/org/apache/gravitino/server/authentication/StaticSignKeyValidator.java
```diff
@@ -107,6 +123,23 @@ public Principal validateToken(String token, String serviceAudience) {
...
+        if (StringUtils.isNotBlank(field)) {
+          Object claimValue = claims.get(field);
+          if (claimValue != null) {
+            return claimValue.toString();
+          }
+        }
+      }
+    }
+
+    return null;
```
roryqi: throw exception directly here.

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
```diff
@@ -144,4 +144,26 @@ public interface OAuthConfig {
           .version(ConfigConstants.VERSION_1_0_0)
           .stringConf()
           .createWithDefault("org.apache.gravitino.server.authentication.StaticSignKeyValidator");
+
+  ConfigEntry<String> PRINCIPAL_MAPPER_TYPE =
+      new ConfigBuilder(OAUTH_CONFIG_PREFIX + "principalMapperType")
```
roryqi: Is it better to use `principalMapper`? You can refer to Authenticator style.

server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosConfig.java
```diff
@@ -42,4 +42,26 @@ public interface KerberosConfig {
           .stringConf()
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .create();
+
+  ConfigEntry<String> PRINCIPAL_MAPPER_TYPE =
+      new ConfigBuilder(KERBEROS_CONFIG_PREFIX + "principalMapperType")
```
roryqi: Is it better that `principalMapperType` -> `principalMapper`?

server-common/src/main/java/org/apache/gravitino/server/authentication/StaticSignKeyValidator.java
bharos: Done

server-common/src/main/java/org/apache/gravitino/server/authentication/OAuthConfig.java
bharos: Done

server-common/src/main/java/org/apache/gravitino/server/authentication/KerberosConfig.java
bharos: Done

jerqi (approved):
LGTM.

Issue #9802, opened 2026-01-26
[Subtask] Support batchGet for table


hdygxsj:
### Describe the subtask

support batch get for table

### Parent issue

#9518

PR #9803, opened 2026-01-26
[#9802] feat(authz): support preload table metadata in batch metadata authorization

hdygxsj:

support preload table metadata in batch metadata authorization

Fix: #9802

None

Existing IT test in org.apache.gravitino.client.integration.test.authorization.TableAuthorizationIT

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
@@ -328,7 +328,7 @@ public class MetadataAuthzHelper {
 
   private static void preloadToCache(
       Entity.EntityType entityType, NameIdentifier[] nameIdentifiers) {
-    if (GravitinoEnv.getInstance().config().get(Configs.CACHE_ENABLED)) {
+    if (GravitinoEnv.getInstance().cacheEnabled()) {
       if (entityType == Entity.EntityType.TABLE) {
         GravitinoEnv.getInstance()
             .entityStore()


hdygxsj:
@jerqi PHAL

hdygxsj:
after batch get we only need 2.8 s for 5000 tables
image

before batch get we need 4-5s for 5000 tables
image

After this PR, the main bottleneck lies in fetching the owner ID; we need to further optimize batch operations for IDs.
image

roryqi:
cc @yuqi1129 U are the owner of storage module. Could u give some suggestion?

server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
```diff
@@ -136,6 +137,7 @@ public static NameIdentifier[] filterByExpression(
       String expression,
       Entity.EntityType entityType,
       NameIdentifier[] nameIdentifiers) {
+    preloadToCache(entityType, nameIdentifiers);
```
yuqi1129, line 140: Prefetching the tables is a nice optimization for Jcasbin. I have the following points that need to be cleared up. 
- It seems that we need to load all tables in a schema whenever calling `filterByExpression`, Is there a way to cache them?
- Based on the code,  although you just put tables in the entity cache, you did not use any of them when executing `batchGet`, am I right?
```
@Override
  public <E extends Entity & HasIdentifier> List<E> batchGet(
      List<NameIdentifier> ident, Entity.EntityType entityType, Class<E> clazz) {
    List<E> entities = backend.batchGet(ident, entityType);
    for (E entity : entities) {
      cache.put(entity);
    }
    return entities;
  }
```

hdygxsj: Regarding the first point, I've optimized the code so that batchGet is now only performed for entities not present in the cache.
Regarding the second point: Yes.

yuqi1129: Okay. The code LGTM overall. Can you spare some time to solve the CI problem?

server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
```diff
@@ -318,4 +325,15 @@ private static void checkExecutor() {
       }
     }
   }
+
+  private static void preloadToCache(
+      Entity.EntityType entityType, NameIdentifier[] nameIdentifiers) {
+    if (GravitinoEnv.getInstance().config().get(Configs.CACHE_ENABLED)) {
```
yuqi1129: GravitinoEnv.getInstance()#cacheEnabled

Issue #9528, opened 2025-12-23
[Subtask] Add relational storage support for UDF metadata


mchades:
### Describe the subtask

Add relational storage support for UDF metadata

### Parent issue

#6158

PR #9806, opened 2026-01-27
[#9528] feat(storage): support function management (PO) (part-1)

mchades:

Introduce PO (Persistent Object) classes for Function management in relational storage.
- Added FunctionPO: Represents the base function metadata.
- Added FunctionVersionPO: Represents specific versions of a function.
- Added FunctionMaxVersionPO: Used for tracking function versions.

These POs serve as the fundamental data structures for mapping database records to Java objects in the relational storage implementation of UDFs (User Defined Functions).

Fix: #9528

No.

Compiling and verified in subsequent commits.

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/core/src/main/java/org/apache/gravitino/storage/relational/po/FunctionPO.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/po/FunctionPO.java
@@ -21,10 +21,14 @@ package org.apache.gravitino.storage.relational.po;
 import com.google.common.base.Preconditions;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.ToString;
+import lombok.experimental.Accessors;
 import org.apache.commons.lang3.StringUtils;
 
 @EqualsAndHashCode
 @Getter
+@ToString
+@Accessors(fluent = true)
 public class FunctionPO {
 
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/po/FunctionPO.java
```diff
@@ -0,0 +1,175 @@
...
+
+  private FunctionVersionPO functionVersionPO;
+
+  private FunctionPO() {}
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
```
jerryshao: As I mentioned before, we can also use lombok to generate the Builder class automatically.

mchades: Field validation is needed during the build process, so it's not suitable to use the Builder annotation here.

jerryshao: I think it is also doable for using lombok, can you check how to do this with lombok? Also, you can refer to `JobTemplatePO`.

mchades: fixed

GEMINI.md
mchades, line 1: I found that adding this change can effectively guide Gemini to follow the project guidelines while doing vibe coding with antigravity, since it's a relatively small modification, so I attached it to this PR.

Issue #9770, opened 2026-01-21
[Improvement] Authentication significantly impacts the performance of the IRC REST Server.


danhuawang:
### What would you like to be improved?

**For load table, authentication disable, the RPS can reach to 210 , but authentication enable, the RPS is only 43**

<img width="740" height="302" alt="Image" src="https://github.com/user-attachments/assets/ec6afc35-796c-4cf3-9d51-5dfbb4762b00" />

<img width="886" height="570" alt="Image" src="https://github.com/user-attachments/assets/4d13c87c-dd14-42c3-80ee-405a98344674" />

**For create table,  authentication disable, the RPS can reach to 96, but authentication enable, the RPS is only 14**

<img width="804" height="256" alt="Image" src="https://github.com/user-attachments/assets/b0970bca-3b99-48d1-8341-7ce2f9405ca1" />

<img width="779" height="414" alt="Image" src="https://github.com/user-attachments/assets/dfd6586b-fba3-4408-9a93-b69646bd9737" />

### How should we improve?

Minimize the performance impact of authentication on the IRC REST Server.

danhuawang:
@jerqi Could you help check this issue when you are free?

bharos:
@danhuawang you mean this happens during authorization enabled ?
Just want to make sure this is about authorization and not authentication ?

bharos:
@jerqi  Wondering if more caching will help here for calls like
- [ getUserEntity()](https://github.com/apache/gravitino/blob/daf5f26f66e33ec42d6dca1e40407dff36d4db2a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java#L405)
- Get User role mapping [listEntitiesByRelation()](https://github.com/apache/gravitino/blob/daf5f26f66e33ec42d6dca1e40407dff36d4db2a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java#L253)

danhuawang:
> [@danhuawang](https://github.com/danhuawang) you mean this happens during authorization enabled ? Just want to make sure this is about authorization and not authentication ?

Oh, it's a typo. This is about authorization. I update the issue.

roryqi:
> > [@jerqi](https://github.com/jerqi) Wondering if more caching will help here for calls like
> > 
[...]

PR #9812, opened 2026-01-27
[#9770] improvement(authz): Convert Jcasbin internal map to the cache

roryqi:

Convert Jcasbin internal map to the cache

Fix: #9770

Add documents.

Add UTs.

### DIFF
--- a/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/ModelAuthorizationIT.java
+++ b/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/ModelAuthorizationIT.java
@@ -141,7 +141,7 @@ public class ModelAuthorizationIT extends BaseRestApiAuthorizationIT {
         () -> {
           normalUserCatalog.registerModel(NameIdentifier.of(SCHEMA, "model3"), "", new HashMap<>());
         });
-    
+
     // Test Case 3: Grant using LEGACY privilege name (CREATE_MODEL)
     // This should work, proving backward compatibility is maintained
     gravitinoMetalake.grantPrivilegesToRole(
@@ -262,7 +262,7 @@ public class ModelAuthorizationIT extends BaseRestApiAuthorizationIT {
               null);
         });
     GravitinoMetalake gravitinoMetalake = client.loadMetalake(METALAKE);
-    
+
     // Test Case 1: Grant using NEW privilege name (LINK_MODEL_VERSION)
     // This should allow the user to link model versions
     gravitinoMetalake.grantPrivilegesToRole(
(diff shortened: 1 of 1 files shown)


roryqi:
@danhuawang Could u take a look at this pull request?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -489,7 +514,7 @@ private void loadRolePrivilege(
   }
 
   private void loadOwnerPolicy(String metalake, MetadataObject metadataObject, Long metadataId) {
-    if (ownerRel.containsKey(metadataId)) {
+    if (ownerRel.getIfPresent(metadataId) != null) {
```
hdygxsj, line 521: When the owner is about to expire, could it be possible that the owner is fetched here (thus skipping the "load owner" step), but then actually expires by the time it's accessed, leading to an incorrect authorization check?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -84,14 +84,40 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
...
+        GravitinoEnv.getInstance().config().get(Configs.GRAVITINO_AUTHORIZATION_ROLE_CACHE_SIZE);
+    long ownerCacheSize =
+        GravitinoEnv.getInstance().config().get(Configs.GRAVITINO_AUTHORIZATION_OWNER_CACHE_SIZE);
+    loadedRoles =
+        Caffeine.newBuilder()
+            .expireAfterWrite(cacheExpirationSecs, TimeUnit.SECONDS)
+            .maximumSize(roleCacheSize)
+            .executor(Runnable::run)
+            .removalListener(
+                (roleId, value, cause) -> {
```
hdygxsj, line 117: Same as above, should we adjust the expiration policy—for example, by refreshing the expiration time upon access?

hdygxsj: Could you add a unit test to verify this scenario?

danhuawang:
I have verified the changes in this PR. With authorization enabled, performance increased from 40 to 110 req/s. With authorization disabled (main branch), it reached 210 req/s.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
roryqi: if we call the method `getIfPresent`, the cache will refresh expire time. Cafffine cache handles the case.

roryqi: It's hard to add tests after thinking twice.

hdygxsj: Should we use expireAfterAccess?

image

roryqi: OK, I got it. I will change this.

qqqttt123:
cc @qqqttt123

qqqttt123:
@hdygxsj Could u take a look again?

Issue #9816, opened 2026-01-28
[Bug report] updateColumnDefaultValue to an empty value failed because of the unexpected error in SQL syntax


danhuawang:
### Version

main branch

### Describe what's wrong

**I want to set the column's default value to an empty value for a MySQL table by gravitino REST API, but there's issue as following:**

```
{
    "code": 1002,
    "type": "RuntimeException",
    "message": "Failed to operate object [demo_new] operation [ALTER] under [gt_mysql_test_all_type], reason [You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 2]",
    "stack": [
        "org.apache.gravitino.exceptions.GravitinoRuntimeException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 2",
        "\tat org.apache.gravitino.catalog.jdbc.converter.JdbcExceptionConverter.toGravitinoException(JdbcExceptionConverter.java:59)",
        "\tat org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations.alterTable(JdbcTableOperations.java:318)",
        "\tat org.apache.gravitino.catalog.jdbc.JdbcCatalogOperations.internalAlterTable(JdbcCatalogOperations.java:558)",
        "\tat org.apache.gravitino.catalog.jdbc.JdbcCatalogOperations.alterTable(JdbcCatalogOperations.java:414)",
        "\tat org.apache.gravitino.catalog.TableOperationDispatcher.lambda$alterTable$6(TableOperationDispatcher.java:235)",
        "\tat org.apache.gravitino.catalog.CatalogManager$CatalogWrapper.lambda$doWithTableOps$1(CatalogManager.java:147)",
        "\tat org.apache.gravitino.utils.IsolatedClassLoader.withClassLoader(IsolatedClassLoader.java:86)",
        "\tat org.apache.gravitino.catalog.CatalogManager$CatalogWrapper.doWithTableOps(CatalogManager.java:142)",
        "\tat org.apache.gravitino.catalog.TableOperationDispatcher.lambda$alterTable$7(TableOperationDispatcher.java:234)",
        "\tat org.apache.gravitino.catalog.OperationDispatcher.doWithCatalog(OperationDispatcher.java:117)",
        "\tat org.apache.gravitino.catalog.TableOperationDispatcher.lambda$alterTable$10(TableOperationDispatcher.java:231)",
[...]

PR #9821, opened 2026-01-28
[#9816] fix(catalogs): Fix alter JDBC catalogs column default value problem

yuqi1129:

This pull request enhances the handling of default values for string-type columns in JDBC catalog implementations for MySQL, PostgreSQL, Doris, OceanBase, and StarRocks. It ensures that string default values are always properly quoted in generated SQL, addressing cases where default values may be empty or unquoted. Comprehensive unit tests have been added for each implementation to verify this behavior.

it's a bug.

Fix: #9816 

N/A

UTs

### DIFF
--- a/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
+++ b/catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
@@ -556,8 +556,6 @@ public class PostgreSqlTableOperations extends JdbcTableOperations
     }
 
     StringBuilder sqlBuilder = new StringBuilder(ALTER_TABLE + jdbcTable.name());
-    String newDefaultValue =
-        handleQuotingForDefaultValue(column, updateColumnDefaultValue.getNewDefaultValue());
     sqlBuilder
         .append("\n")
         .append(ALTER_COLUMN)
@@ -565,7 +563,9 @@ public class PostgreSqlTableOperations extends JdbcTableOperations
         .append(col)
         .append(PG_QUOTE)
         .append(" SET DEFAULT ")
-        .append(newDefaultValue);
+        .append(
+            columnDefaultValueConverter.fromGravitino(
+                updateColumnDefaultValue.getNewDefaultValue()));
     return sqlBuilder.append(";").toString();
   }
 


catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcTableOperations.java
```diff
@@ -95,6 +100,79 @@ public void initialize(
     this.columnDefaultValueConverter = jdbcColumnDefaultValueConverter;
   }
 
+  protected String calculateDefaultValue(JdbcColumn column, Expression defaultValueExpression) {
```
mchades: Could you please add some example comments to clarify the code logic? The current implementation is difficult to understand.

mchades: BTW, I feel like this processing logic in this method should be the responsibility of the `columnDefaultValueConverter`.

yuqi1129: I'm afraid that the logic does not belong to `columnDefaultValueConverter`. As the root cause of this problem is that we do not quote an empty string value when concatenating SQL.  

The result value of `columnDefaultValueConverter` is correct, and it's the string value quoting issue that cause the problem.

yuqi1129: >  Could you please add some example comments to clarify the code logic? The current implementation is difficult to understand.

Let me add more comments to explain the logic.

yuqi1129: updated

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcTableOperations.java
```diff
@@ -95,6 +100,90 @@ public void initialize(
...
+   * @param column the column
+   * @param defaultValueExpression the default value expression
+   * @return the properly quoted default value as a string
+   */
+  protected String handleQuotingForDefaultValue(
+      JdbcColumn column, Expression defaultValueExpression) {
+    String defaultValue = columnDefaultValueConverter.fromGravitino(defaultValueExpression);
+
+    // Special handling for SQL NULL: do not quote it, even for string-like columns.
+    String trimmedDefault = defaultValue == null ? null : defaultValue.trim();
```
mchades: Why do we need to use `trim()`? I think using trim will change user input, for example, if the user wants to specify default values for two spaces (`  `), the spaces will disappear after `trim()`

yuqi1129: OK, I miss the case.

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcTableOperations.java
```diff
@@ -95,6 +100,89 @@ public void initialize(
...
+   * @param defaultValueExpression the default value expression
+   * @return the properly quoted default value as a string
+   */
+  protected String handleQuotingForDefaultValue(
+      JdbcColumn column, Expression defaultValueExpression) {
+    String defaultValue = columnDefaultValueConverter.fromGravitino(defaultValueExpression);
+
+    // Special handling for SQL NULL: do not quote it, even for string-like columns.
+    if (defaultValueExpression == Literals.NULL
+        || (defaultValue != null && "NULL".equalsIgnoreCase(defaultValue))) {
```
mchades: this condition can be simplified to `JdbcColumnDefaultValueConverter.NULL.equalsIgnoreCase(defaultValue)`

mchades: Sorry, it cannot be. But seems `Literals.NULL.equals(defaultValueExpression)` is enough.

yuqi1129: Fixed

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/operation/JdbcTableOperations.java
```diff
@@ -95,6 +100,87 @@ public void initialize(
...
+    // Special handling for string-like columns.
+    if (column.dataType() instanceof VarCharType
+        || column.dataType() instanceof StringType
+        || column.dataType() instanceof FixedCharType) {
+      if (StringUtils.isEmpty(defaultValue)) {
+        // Represent empty string default as two single quotes.
+        return "''";
+      }
+
+      // If the converter already produced a quoted literal, preserve it.
```
mchades: When will the converter produce a quoted literal

yuqi1129: Please see L74
https://github.com/apache/gravitino/blob/b5af4dc09db75557d0a1a874bda91b8ddc77bc5b/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/converter/JdbcColumnDefaultValueConverter.java#L58-L76

mchades: So, can we unify the string-like literal type in the converter? It seems to simplify things.

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/converter/JdbcColumnDefaultValueConverter.java
```diff
@@ -61,7 +62,12 @@ public String fromGravitino(Expression defaultValue) {
       if (defaultValue.equals(Literals.NULL)) {
         return NULL;
       } else if (type instanceof Type.NumericType) {
-        return literal.value().toString();
+        // It seem that literals.value().toString() can be an empty string for numeric types
```
mchades: Any references? And what will be the final value when this occurs?

yuqi1129: Before there will be a syntax error: 
```
alter table t modify column `id` int null default ;
```

after 
```
mysql> alter table t modify column `id` int null default '';
ERROR 1067 (42000): Invalid default value for 'id'
```

The error message is expected, and a syntax error is not. (The result is for MySQL; other JDBC databases may behave differently)

mchades: Got it. And the comment L65 can be moved to L67

yuqi1129: Adjusted.

Issue #9825, opened 2026-01-29
[Improvement] Using internal interfaces instead of HTTP interfaces when using dynamic config provider for IRC


qqqttt123:
### What would you like to be improved?

Using internal interfaces instead of HTTP interfaces when using dynamic config provider for IRC

### How should we improve?

Using internal interfaces instead of HTTP interfaces when using dynamic config provider for IRC

PR #9827, opened 2026-01-29
[#9825] feat(iceberg): Supports IRC to use internal catalog fetcher

qqqttt123:

Supports IRC to use internal catalog fetcher

Fix: #9825 

No.

Add tests.

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/provider/DynamicIcebergConfigProvider.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/provider/DynamicIcebergConfigProvider.java
@@ -104,8 +104,8 @@ public class DynamicIcebergConfigProvider implements IcebergConfigProvider {
   }
 
   /**
-   * Lazily creates the CatalogFetcher based on whether authorization is enabled. Uses internal
-   * interface when authorization is enabled, otherwise uses HTTP interface.
+   * Lazily creates the CatalogFetcher based on whether running in auxiliary mode. Uses internal
+   * interface when in auxiliary mode (embedded in Gravitino server), otherwise uses HTTP interface.
    */
   private CatalogFetcher getCatalogFetcher() {
     if (catalogFetcher != null) {
@@ -114,7 +114,7 @@ public class DynamicIcebergConfigProvider implements IcebergConfigProvider {
     synchronized (this) {
       if (catalogFetcher == null) {
         IcebergRESTServerContext serverContext = IcebergRESTServerContext.getInstance();
-        if (serverContext.isAuthorizationEnabled()) {
+        if (serverContext.isAuxMode()) {
           catalogFetcher = new InternalCatalogFetcher(gravitinoMetalake);
         } else {
           String uri = properties.get(IcebergConstants.GRAVITINO_URI);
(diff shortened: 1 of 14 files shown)


qqqttt123:
@FANNG1 Could u take a look at this pull request?

FANNG1:
1. Is there any reason to add `ValidatingCache`, seems not tightly bound to this PR?
2. In your another PR, you shadow the security catalog properties,  could the dynamic catalog provider works after you doing that?

roryqi:
> 1. Is there any reason to add `ValidatingCache`, seems not tightly bound to this PR?
> 2. In your another PR, you shadow the security catalog properties,  could the dynamic catalog provider works after you doing that?

1. We should validate whether the catalog config is expired. If the catalog is expired, we should reload the catalog config and create a new catalog wrapper.
2. It can work. I hide the sensitive catalog properties in catalog operations. We use dispatcher methods in the IRC dynamic config provider.

FANNG1:
> 1. We should validate whether the catalog config is expired. If the catalog is expired, we should reload the catalog config and create a new catalog wrapper.

I would suggest to implement this cache in separate PR, or we can talk about it in issues about the neccessary to introduce it.

> 2. It can work. I hide the sensitive catalog properties in catalog operations. We use dispatcher methods in the IRC dynamic config provider.

Could you provide more context about how to resolve this issue?

roryqi:
> > 1. We should validate whether the catalog config is expired. If the catalog is expired, we should reload the catalog config and create a new catalog wrapper.
> 
> I would suggest to implement this cache in separate PR, or we can talk about it in issues about the neccessary to introduce it.
> 
> > 2. It can work. I hide the sensitive catalog properties in catalog operations. We use dispatcher methods in the IRC dynamic config provider.
> 
> Could you provide more context about how to resolve this issue?

1. OK
2. Because dynamic config provider won't call the HTTP request to load catalog if authorization is enabled. We only hidden the sensitive information for HTTP requests. We don't hidden the sensitive information.

FANNG1:
> 2. Because dynamic config provider won't call the HTTP request to load catalog if authorization is enabled. We only hidden the sensitive information for HTTP requests. We don't hidden the sensitive information.

Got it, thx

roryqi:
@FANNG1 Could u take an another look?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/provider/DynamicIcebergConfigProvider.java
```diff
@@ -110,15 +103,44 @@ public Optional<IcebergConfig> getIcebergCatalogConfig(String catalogName) {
...
+   * interface when authorization is enabled, otherwise uses HTTP interface.
+   */
+  private CatalogFetcher getCatalogFetcher() {
+    if (catalogFetcher != null) {
+      return catalogFetcher;
+    }
+    synchronized (this) {
+      if (catalogFetcher == null) {
+        IcebergRESTServerContext serverContext = IcebergRESTServerContext.getInstance();
+        if (serverContext.isAuthorizationEnabled()) {
```
FANNG1: seems we should use `InternalCatalogFetcher` in aux mode,  `HttpCatalogFetcher` for standalone mode.  using isAuthorizationEnabled is a little tracky.

FANNG1:
LGTM except minor comment, could you update the document too?

FANNG1:
This PR seems doesn't affect using generate catalog as the IRC catalog backend, because it only affects how to get the catalog config from the Gravitino server, while using generate catalog as the IRC catalog backend is mainly for table operations. cc @jerryshao @jerqi

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/provider/DynamicIcebergConfigProvider.java
roryqi: According to the communication with u, we don't have way to judge whether it is in aux mode. So we don't have better solution now.

roryqi:
> LGTM except minor comment, could you update the document too?

Updated.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/provider/DynamicIcebergConfigProvider.java
FANNG1: We could support it, like adding a flag to indicate whether it's aux mode?
```
public interface GravitinoAuxiliaryService {
  void serviceInit(Map<String, String> config, bool auxMode);
}
```

roryqi: OK, fixed.

Issue #9837, opened 2026-01-31
[Improvement] Gravitino IRC doesn't support PyIceberg's rest client with a local path as warehouse storage location


shaofengshi:
### What would you like to be improved?

First, this is only for a testing case, as it is using a local folder as Iceberg storage location.

Secondly, this is a limitation of PyIceberg rest client, as it hardcode the credential-vending in HTTP request header (see https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/rest/__init__.py#L768):

`   
 def _config_headers(self, session: Session) -> None:
        header_properties = self._extract_headers_from_properties()
        session.headers.update(header_properties)
        session.headers["Content-type"] = "application/json"
        session.headers["X-Client-Version"] = ICEBERG_REST_SPEC_VERSION
        session.headers["User-Agent"] = f"PyIceberg/{__version__}"
        session.headers["X-Iceberg-Access-Delegation"] = "vended-credentials"
`

<img width="2314" height="1014" alt="Image" src="https://github.com/user-attachments/assets/bc6c0988-a14f-48f7-825f-2952fef6b7bf" />

Today when I use PyIceberg's rest to access gravitino and specifying a local path to create table, it will reports error:

`
  catalog = RestCatalog(
        name="gravitino",
        uri=iceberg_rest_endpoint
    )
    print(f"Catalog created: {catalog.name}")

  catalog.create_namespace(namespace_name)        
print(f"Namespace created: {namespace_name}")

        # Step 3: Create table
        print(f"\n=== Creating Iceberg Table ===")
        schema = Schema(
            NestedField(field_id=1, name="id", field_type=IntegerType(), required=True),
            NestedField(field_id=2, name="name", field_type=StringType(), required=False),
            NestedField(field_id=3, name="age", field_type=IntegerType(), required=False),
            NestedField(field_id=4, name="score", field_type=DoubleType(), required=False),
        )

        table = catalog.create_table(
            identifier=f"{namespace_name}.{table_name}",
            schema=schema,
        )
`
It reports error:

`E       pyiceberg.exceptions.BadRequestError: IllegalArgumentException: There are no credential provider for the catalog.
`

### How should we improve?

There are several ways to enhance:

1. For a local path as storage, Gravitino should not check credential providers (ignore the header of credential-vending)
2. or, return a mock token for client side.

PR #9839, opened 2026-02-01
[#9837] feat(iceberg): skip credential vending for local or hdfs

FANNG1:

skip credential vending for local or hdfs

Fix: #9837 

no

1. using local fs as warehouse
2. set credential vending request by using spark sql, could load table succussfully

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
@@ -244,6 +244,7 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
     if (!requestCredential) {
       return false;
     }
+    validateCredentialLocation(loadTableResponse.tableMetadata().location());
     return !isLocalOrHdfsTable(loadTableResponse.tableMetadata());
   }
 
@@ -252,7 +253,16 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
   }
 
   @VisibleForTesting
+  static void validateCredentialLocation(String location) {
+    if (StringUtils.isBlank(location)) {
+      throw new IllegalArgumentException(
+          "Table location cannot be null or blank when requesting credentials");
+    }
+  }
+
+  @VisibleForTesting
   static boolean isLocalOrHdfsLocation(String location) {
+    // Precondition: location is non-blank (enforced by caller).
     if (StringUtils.isBlank(location)) {
       return false;
     }
@@ -264,6 +274,7 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
     }
     String scheme = uri.getScheme();
     if (scheme == null) {
+      // No scheme means a local path.
       return true;
     }
     return "file".equalsIgnoreCase(scheme) || "hdfs".equalsIgnoreCase(scheme);
--- a/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestCatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestCatalogWrapperForREST.java
@@ -60,4 +60,18 @@ public class TestCatalogWrapperForREST {
     Assertions.assertFalse(CatalogWrapperForREST.isLocalOrHdfsLocation(""));
     Assertions.assertFalse(CatalogWrapperForREST.isLocalOrHdfsLocation("   "));
   }
+
+  @Test
+  void testValidateCredentialLocation() {
+    Assertions.assertDoesNotThrow(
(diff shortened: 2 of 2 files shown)


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -238,6 +239,36 @@ private Credential getCredential(
...
+  private boolean shouldGenerateCredential(
+      LoadTableResponse loadTableResponse, boolean requestCredential) {
+    if (!requestCredential) {
+      return false;
+    }
+    return !isLocalOrHdfsTable(loadTableResponse.tableMetadata());
+  }
+
+  private boolean isLocalOrHdfsTable(TableMetadata tableMetadata) {
+    return isLocalOrHdfsLocation(tableMetadata.location());
```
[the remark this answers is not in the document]
FANNG1: this will make the code more complicated, it's rarely the table location metadata is local, the table data is in cloud

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -238,6 +239,36 @@ private Credential getCredential(
...
+  }
+
+  @VisibleForTesting
+  static boolean isLocalOrHdfsLocation(String location) {
+    if (StringUtils.isBlank(location)) {
+      return false;
+    }
```
bharos, line 268: If location is blank , so we proceed to create credential. Is that correct though ?

FANNG1: It's a defense code, and the table location should not be null, WDYT?

bharos: I think it's ok. If we want we could do

```
private boolean shouldGenerateCredential(
    LoadTableResponse loadTableResponse, boolean requestCredential) {
  if (!requestCredential) {
    return false;
  }
  
  String location = loadTableResponse.tableMetadata().location();
  
  // Fail fast on invalid locations
  if (StringUtils.isBlank(location)) {
        "Table location cannot be null or blank when requesting credentials");
  }
  
  return !isLocalOrHdfsLocation(location);
}

@VisibleForTesting
static boolean isLocalOrHdfsLocation(String location) {
  // Precondition: location is non-blank (enforced by caller)
  URI uri;
  try {
    uri = URI.create(location);
  } catch (IllegalArgumentException e) {
    return false;
  }
  String scheme = uri.getScheme();
  if (scheme == null) {
    return true;  // no scheme = local path
  }
  return "file".equalsIgnoreCase(scheme) || "hdfs".equalsIgnoreCase(scheme);
}
```
but your current code is ok as well with the assumption location should not be null

Also add a comment for if scheme== null that no scheme means local path ?

FANNG1: Yes, this is more reasonable

bharos:
LGTM!

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
FANNG1: update the code

shaofengshi (approved):
LGTM

roryqi:
Maybe we should allow requests match none of credential providers. It's not good to add a condition to fix this. We should use interface to define different behaviours for different paths.

FANNG1:
> Maybe we should allow requests match none of credential providers. It's not good to add a condition to fix this. We should use interface to define different behaviours for different paths.

Using the interface seems too complicated for this simple feature.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -238,6 +239,47 @@ private Credential getCredential(
...
+  @VisibleForTesting
+  static void validateCredentialLocation(String location) {
+    if (StringUtils.isBlank(location)) {
+      throw new IllegalArgumentException(
+          "Table location cannot be null or blank when requesting credentials");
+    }
+  }
+
+  @VisibleForTesting
+  static boolean isLocalOrHdfsLocation(String location) {
```
shaofengshi, line 264: this method name is some kind of confusing: why check it is local or hdfs? can it be split to two: isLocal and isHdfs ? Because this method has the assumption that local and hdfs doesn't need credential provider, but in the future there might be a third case.

PR #9854, opened 2026-02-01
[#8507] Add null check for namespace in ListTopicFailureEvent

sabashaikh4:

Added a precondition check to ensure the namespace parameter is not null in `ListTopicFailureEvent` constructor using `Preconditions.checkNotNull()`.

To prevent NullPointerException and fail fast with a clear error message when null namespace is passed to the constructor.

Fix: #8507

No

Added unit test `testNamespaceMustNotBeNull()` to verify the null check throws NullPointerException when the namespace is null.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
+++ b/core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
@@ -45,8 +45,8 @@ public final class ListTopicFailureEvent extends TopicFailureEvent {
   }
 
   private static NameIdentifier toNameIdentifier(Namespace namespace) {
-     Preconditions.checkArgument(namespace != null, "namespace must not be null");
-     return NameIdentifier.of(namespace.levels());
+    Preconditions.checkArgument(namespace != null, "namespace must not be null");
+    return NameIdentifier.of(namespace.levels());
   }
 
   /**
--- a/core/src/test/java/org/apache/gravitino/listener/api/event/ListTopicFailureEventTest.java
+++ b/core/src/test/java/org/apache/gravitino/listener/api/event/ListTopicFailureEventTest.java
@@ -28,7 +28,7 @@ public class ListTopicFailureEventTest {
   public void testNamespaceMustNotBeNull() {
     Assertions.assertThrows(
         IllegalArgumentException.class,
-         () -> {
+        () -> {
           new ListTopicFailureEvent("user", null, new Exception());
         });
   }


bharos:
LGTM

core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
```diff
@@ -40,6 +41,7 @@ public final class ListTopicFailureEvent extends TopicFailureEvent {
    */
   public ListTopicFailureEvent(String user, Namespace namespace, Exception exception) {
     super(user, NameIdentifier.of(namespace.levels()), exception);
+    Preconditions.checkNotNull(namespace, "namespace must not be null");
```
jerryshao: Shall we move this line above the `super(...)`? Also, it would be better to use `Preconditions.checkArguments`, using `Preconditions.checkNotNull` will throw an NPE, which is not what we wanted.

sabashaikh4: @jerryshao Thank you for the review! I've updated the code to:
1. Move the check before `super()`
2. Use `checkArgument` instead of `checkNotNull`
3. Update the test to expect `IllegalArgumentException`

Changes are pushed. Please review again.

justinmclean:
We use Spotless to ensure code is formatted consistently. Please run `./gradlew :core:spotlessApply` to fix formatting issues and check in the changes.

sabashaikh4:
@justinmclean I've added the Apache license header to the test file. The Spotless formatting issue should now be resolved. Thank you for your patience!

justinmclean:
Sorry there is still some spoless issues. Did you run `./gradlew :core:spotlessApply` and push the changes?

sabashaikh4:
@justinmclean I've run `./gradlew :core:spotlessApply` and pushed the formatting changes. The Spotless checks should now pass. Please let me know if there are any other issues. Thank you!

core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
```diff
@@ -39,10 +40,15 @@ public final class ListTopicFailureEvent extends TopicFailureEvent {
    * @param exception The exception encountered during the attempt to list topics.
    */
   public ListTopicFailureEvent(String user, Namespace namespace, Exception exception) {
-    super(user, NameIdentifier.of(namespace.levels()), exception);
```
jerryshao, line 42: You don't need a new method, just add this line `Preconditions.checkArgument(namespace != null, "namespace must not be null");` here is enough.

sabashaikh4: @jerryshao I tried placing the precondition check after super() as suggested, but this causes a NullPointerException in the test. Since `namespace.levels()` is called in the super() constructor before the null check happens, we get NPE instead of the expected IllegalArgumentException.

The helper method ensures validation happens before super() is invoked (which must be the first statement). The test `testNamespaceMustNotBeNull()` expects IllegalArgumentException when namespace is null.

Would it be acceptable to keep the `validateNamespace()` helper method to ensure proper validation order?

Test failure log: ListTopicFailureEventTest > testNamespaceMustNotBeNull() FAILED
    org.opentest4j.AssertionFailedError: Unexpected exception type thrown ==> expected: <java.lang.IllegalArgumentException> but was: <java.lang.NullPointerException>
        at app//org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:65)
        at app//org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:37)
        at app//org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3082)
        at app//org.apache.gravitino.listener.api.event.ListTopicFailureEventTest.testNamespaceMustNotBeNull(ListTopicFailureEventTest.java:29)

        Caused by:
        java.lang.NullPointerException: Cannot invoke "org.apache.gravitino.Namespace.levels()" because "namespace" is null
            at org.apache.gravitino.listener.api.event.ListTopicFailureEvent.<init>(ListTopicFailureEvent.java:43)
            at org.apache.gravitino.listener.api.event.ListTopicFailureEventTest.lambda$testNamespaceMustNotBeNull$0(ListTopicFailureEventTest.java:31)
            at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:55)
            ... 3 more

justinmclean: You could do it like this, but it seems a bit ugly to me.
```
public ListTopicFailureEvent(String user, Namespace namespace, Exception exception) {
  super(user, NameIdentifier.of(Preconditions.checkNotNull(namespace, "namespace must not be null").levels()),
        exception);
}
```

Something like this I think, would be best:
```
public ListTopicFailureEvent(String user, Namespace namespace, Exception exception) {
  super(user, toNameIdentifier(namespace), exception);
}

private static NameIdentifier toNameIdentifier(Namespace namespace) {
  Preconditions.checkNotNull(namespace, "namespace must not be null");
  return NameIdentifier.of(namespace.levels());
}
```

@jerryshao what do you think?

jerryshao: > I tried placing the precondition check after super() as suggested

put this before the `super()`, not after.

sabashaikh4: @justinmclean Thank you for the suggestion! I've updated the code to use the `toNameIdentifier()` approach with `checkNotNull`. The tests now pass successfully.

sabashaikh4: @jerryshao Thanks for the clarification.
Since Java requires super() to be the first statement, I moved the null validation into a helper method (toNameIdentifier()), which is invoked from the super() call. This ensures the namespace is validated before namespace.levels() is accessed, and avoids the NPE.

I've updated the code accordingly, and all tests are passing now. Please let me know if this approach looks good to you.

core/src/main/java/org/apache/gravitino/listener/api/event/ListTopicFailureEvent.java
```diff
@@ -39,10 +40,15 @@ public final class ListTopicFailureEvent extends TopicFailureEvent {
...
+    super(user, toNameIdentifier(namespace), exception);
     this.namespace = namespace;
   }
 
+  private static NameIdentifier toNameIdentifier(Namespace namespace) {
+    Preconditions.checkNotNull(namespace, "namespace must not be null");
```
jerryshao: Use `checkArgument` as I mentioned before.

sabashaikh4: @justinmclean @jerryshao 

I want to make sure I implement this correctly. There seems to be a difference in approach:

- @justinmclean suggested using `checkNotNull` (throws `NullPointerException`)
- @jerryshao mentioned using `checkArgument` (throws `IllegalArgumentException`)

The current test expects `NullPointerException`, but if I use `checkArgument` as @jerryshao suggested, I'll need to update the test to expect `IllegalArgumentException`.

Could you both please confirm which approach you'd prefer? I'm happy to implement either way - I just want to make sure we're all aligned.

Thank you for your patience and guidance!

justinmclean: do it as @jerryshao suggests

sabashaikh4: @jerryshao @justinmclean Thank you for the clarification! I've updated the code to use `checkArgument` as suggested. I also updated the test to expect `IllegalArgumentException` to match the implementation. All tests are now passing.

justinmclean:
you'll need to run `./gradlew :core:spotlessApply` again.

sabashaikh4:
Sorry, I accidentally closed this PR while reviewing the changes. Reopening now.

@justinmclean I've run `./gradlew :core:spotlessApply` and pushed the formatting changes (commit 69949d8). All tests are passing locally. ✅

> you'll need to run `./gradlew :core:spotlessApply` again.

Issue #9719, opened 2026-01-15
[FEATURE] Add the version segment module for 435–439 to support Trino 345 through Trino 439.


diqiu50:
### Describe the feature

Add the version segment module for 435–439 to support Trino 345 through Trino 439

PR #9861, opened 2026-02-03
[#9719] feat (trino-connector): Add the version segment module to support Trino-435 to Trino–439

diqiu50:

Add the version segment module for 435–439 to support Trino 345 through Trino 439

Fix: #9719

NO

UT and IT


.github/workflows/trino-integration-test.yml
```diff
@@ -73,7 +76,7 @@ jobs:
 
       - name: Package Gravitino
         run: |
-          ./gradlew compileDistribution compileTrinoConnector -x test
+          ./gradlew build -x test
```
yuqi1129, line 79: Why did you change it to build Gravitino? If it did need such changes, you'd better change the name `Package Gravitino`.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/TrinoITContainers.java
```diff
@@ -70,6 +73,11 @@ public void launch(
       env.put("TRINO_VERSION", String.valueOf(trinoVersion));
     }
     if (trinoConnectorDir != null) {
+      // If the provided path is relative, resolve it against the current working directory.
```
yuqi1129: I couldn't see your code follow your comments here.

trino-connector/trino-connector-435-439/build.gradle.kts
```diff
@@ -0,0 +1,118 @@
...
+val minSupportedTrinoVersionProperty = providers.gradleProperty("minSupportedTrinoVersion").map { it.trim().toInt() }
+val trinoVersionProvider =
+  providers.gradleProperty("trinoVersion").map { it.trim().toInt() }.orElse(minSupportedTrinoVersionProperty)
+
+val resolvedTrinoVersion = trinoVersionProvider.get()
+check(resolvedTrinoVersion >= supportedTrinoVersion) {
+  "Invalid trinoVersion=$resolvedTrinoVersion for project ${project.path}. " +
+    "This module requires trinoVersion >= $supportedTrinoVersion. " +
+    "Either omit -PtrinoVersion to use minSupportedTrinoVersion=${minSupportedTrinoVersionProperty.get()}, " +
+    "or set -PtrinoVersion to a supported value."
```
yuqi1129: I can't get the meaning of the error message: Does it mean that we can skip the error during compilation by setting `minSupportedTrinoVersion`?

.github/workflows/trino-integration-test.yml
diqiu50: `compileDistribution` command  doesn't build the version segments of `trino-connector`

yuqi1129: You'd better add some comments about why you use `compileDistribution` not `build` here, I believe 
`compileDistribution command doesn't build the version segments of trino-connector` is a good candidate.

trino-connector/trino-connector-435-439/build.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+val resolvedTrinoVersion = trinoVersionProvider.get()
+check(resolvedTrinoVersion >= supportedTrinoVersion) {
+  "Failed to build the module ${project.path} of trinoVersion=$resolvedTrinoVersion. " +
+    "This module requires trinoVersion >= $supportedTrinoVersion. " +
+    "Please set '-PtrinoVersion' in the command or set config the 'minSupportedTrinoVersion' of gradle.properties."
+}
+val trinoVersion = resolvedTrinoVersion
+
+java {
+  toolchain.languageVersion.set(JavaLanguageVersion.of(24))
```
yuqi1129, line 44: Do we need JDK 24 to compile Trino 435 to 439?

diqiu50: Yes, handling it in a unified way is simpler. Some later versions upgrade the JDK, but the SPI remains unchanged.

.github/workflows/trino-integration-test.yml
diqiu50: I have renamed it; the build is better here. The tests do not need the JARs from the `compileDistribution` task.

trino-connector/trino-connector-435-439/src/main/java/org/apache/gravitino/trino/connector/GravitinoConnector435.java
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.gravitino.trino.connector;
+
+import io.trino.spi.connector.ConnectorMetadata;
+import io.trino.spi.connector.ConnectorNodePartitioningProvider;
+import io.trino.spi.connector.ConnectorSplitManager;
+import org.apache.gravitino.trino.connector.catalog.CatalogConnectorContext;
+import org.apache.gravitino.trino.connector.catalog.CatalogConnectorMetadata;
+import org.apache.gravitino.trino.connector.catalog.CatalogConnectorMetadataAdapter;
+
+public class GravitinoConnector435 extends GravitinoConnector {
```
yuqi1129, line 28: So, does this connector works beteween 435 and 439? I would rather you name it as `GravitinoTrino435CompatibleConnector` or something similar.

yuqi1129 (commented):
Just a minor one, other LGTM

trino-connector/trino-connector-435-439/src/main/java/org/apache/gravitino/trino/connector/GravitinoConnector435.java
diqiu50: I think GravitinoTrino435  is OK， GravitinoTrino435CompatibleConnector is too long，

yuqi1129 (approved):

Issue #9871, opened 2026-02-04
[Subtask]  Move returnType/returnColumns to FunctionDefinition to support overloading with different return types


mchades:
### Describe the subtask

Enable function overloading with different return types by moving `returnType` and `returnColumns` from the `Function` level to the `FunctionDefinition` level.

   Currently, all overloaded function definitions share the same return type at the Function level. This prevents registering functions like:
   - `add(x int, y int) => int`
   - `add(x float, y float) => float`

   With this change, each `FunctionDefinition` will have its own `returnType` and `returnColumns`, allowing true function overloading with different return types.

### Parent issue

#6158

PR #9872, opened 2026-02-04
[#9871] fix(function): move returnType/returnColumns to FunctionDefinition level

mchades:

Key changes:
- FunctionDefinition: Added returnType() and returnColumns() methods
- FunctionDefinitions: Updated factory methods and implementation
- Function: Removed returnType() and returnColumns() methods
- FunctionCatalog: Consolidated registerFunction() method
- Storage: Removed return_type column from function_meta table
- DTO/Entity: Updated all related classes

This change enables function overloading with different return types.
Previously, returnType/returnColumns were at the Function level, shared
by all FunctionDefinitions. Now each FunctionDefinition has its own
returnType/returnColumns, allowing functions like:
- add(x int, y int) => int
- add(x float, y float) => float

Fix: #9871

no, the Function api not release yet

CI pass

### DIFF
--- a/api/src/main/java/org/apache/gravitino/function/FunctionDefinition.java
+++ b/api/src/main/java/org/apache/gravitino/function/FunctionDefinition.java
@@ -25,7 +25,8 @@ import org.apache.gravitino.rel.types.Type;
 /**
  * A function definition that pairs a specific parameter list with its implementations and return
  * type. A single function can include multiple definitions (overloads), each with distinct
- * parameters, return types, and implementations.
+ * parameters and implementations. Overload resolution is based on parameters only; return type or
+ * return columns are metadata and do not participate in definition matching.
  *
  * <p>For scalar or aggregate functions, use {@link #returnType()} to specify the return type. For
  * table-valued functions, use {@link #returnColumns()} to specify the output columns.
--- a/core/src/test/java/org/apache/gravitino/storage/relational/service/TestFunctionMetaService.java
+++ b/core/src/test/java/org/apache/gravitino/storage/relational/service/TestFunctionMetaService.java
@@ -58,7 +58,6 @@ import org.apache.ibatis.session.SqlSession;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.TestTemplate;
 
-@SuppressWarnings("deprecation")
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/apache/gravitino/function/FunctionDefinition.java
```diff
@@ -18,21 +18,53 @@
...
 import org.apache.gravitino.annotation.Evolving;
+import org.apache.gravitino.rel.types.Type;
 
 /**
- * A function definition that pairs a specific parameter list with its implementations. A single
- * function can include multiple definitions (overloads), each with distinct parameters and
- * implementations.
+ * A function definition that pairs a specific parameter list with its implementations and return
+ * type. A single function can include multiple definitions (overloads), each with distinct
+ * parameters, return types, and implementations.
```
jerryshao: Do we use return types to distinguish different definitions? I remembered that we only honor the parameters, can you clarify it?

scripts/h2/schema-1.2.0-h2.sql
```diff
@@ -452,7 +452,6 @@ CREATE TABLE IF NOT EXISTS `function_meta` (
     `schema_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'schema id',
     `function_type` VARCHAR(64) NOT NULL COMMENT 'function type',
     `deterministic` TINYINT(1) DEFAULT 1 COMMENT 'whether the function result is deterministic',
-    `return_type` CLOB NOT NULL COMMENT 'function return type',
```
jerryshao, line 455: Where do we store the `return columns` previously?

mchades: The field was previously missed in the storage layer. The table function was tested in the FunctionIT, but its `return columns` was not asserted, so the issue went unnoticed. 

https://github.com/apache/gravitino/blob/437f46a2e0c6c192d995380fec6436b2141ab3fa/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/FunctionIT.java#L191-L200

api/src/main/java/org/apache/gravitino/function/FunctionDefinition.java
mchades: we don't use the return types to distinguish different definitions.

JavaDoc updated

Issue #9653, opened 2026-01-08
[Subtask] support compaction strategy handler in recommender


FANNG1:
### Describe the subtask

 support compaction strategy handler in recommender

### Parent issue

#9545

PR #9904, opened 2026-02-06
 [#9653] feat(optimizer): add compaction strategy related impls 

FANNG1:

* Adds QLExpress4 dependency for dynamic expression evaluation in strategy triggers and scoring
* Implements compaction strategy handler with partition-aware recommendation logic
* Adds partition utility methods for generating SQL WHERE clauses from partition metadata

  Example Output (CompactionJobAdapter) Given a job context for db.table with two partitions (a='x1', b='y1') and (a='x2', b='y2'), and job options
  {target-file-size-bytes: "1024"}:

```
  jobConfig() ->
  {
    "table": "db.table",
    "where": "(a = 'x1' AND b = 'y1') OR (a = 'x2' AND b = 'y2')",
    "options": "{\"target-file-size-bytes\", \"1024\"}"
  }
```

Fix: #9653 

existing tests

### DIFF
--- a/maintenance/optimizer/build.gradle.kts
+++ b/maintenance/optimizer/build.gradle.kts
@@ -50,6 +50,7 @@ dependencies {
   annotationProcessor(libs.lombok)
   compileOnly(libs.lombok)
 
+  testImplementation(libs.awaitility)
   testImplementation(libs.junit.jupiter.api)
   testImplementation(libs.junit.jupiter.params)
   testImplementation(libs.mockito.core)
--- a/maintenance/optimizer/src/test/java/org/apache/gravitino/maintenance/optimizer/recommender/util/TestQLExpressionEvaluator.java
+++ b/maintenance/optimizer/src/test/java/org/apache/gravitino/maintenance/optimizer/recommender/util/TestQLExpressionEvaluator.java
@@ -122,7 +122,7 @@ public class TestQLExpressionEvaluator {
 
     // Test numeric operations
     long numericResult = evaluator.evaluateLong("intVal + doubleVal", context);
-    assertEquals(16L, numericResult); // 10 + 5.5 = 15.5 → rounded to 16
+    assertEquals(16L, numericResult); // 10 + 5.5 = 15.5 -> rounded to 16
 
     // Test boolean operations
     boolean boolResult = evaluator.evaluateBool("boolVal && (intVal > 5)", context);


jerryshao:
Is it ready to merge? @FANNG1

FANNG1:
> Is it ready to merge? @FANNG1

yes

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/job/GravitinoJobSubmitter.java
```diff
@@ -79,11 +89,67 @@ public void close() throws Exception {
...
+      OptimizerConfig optimizerConfig,
+      JobExecutionContext jobExecutionContext,
+      GravitinoJobAdapter jobAdapter) {
+    Map<String, String> submitterConfigs =
+        optimizerConfig == null ? Map.of() : optimizerConfig.jobSubmitterConfigs();
+    Map<String, String> adapterConfigs =
+        jobAdapter == null ? Map.of() : jobAdapter.jobConfig(jobExecutionContext);
+
+    Map<String, String> mergedConfigs = new LinkedHashMap<>();
+    mergedConfigs.putAll(submitterConfigs);
```
tedyu, line 123: Is it possible that there is overlap between the keys of submitterConfigs and adapterConfigs ?

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/Recommender.java
```diff
@@ -302,4 +327,23 @@ private SupportTableStatistics requireTableStatisticsProvider() {
...
+  private void logRecommendation(String strategyName, StrategyEvaluation evaluation) {
+    JobExecutionContext jobExecutionContext =
+        evaluation
+            .jobExecutionContext()
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Job execution context is missing for evaluation of strategy "
+                            + strategyName));
+    System.out.println(
```
tedyu, line 340: shouldn't the `LOG` be used for the logging ?

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/job/GravitinoJobSubmitter.java
FANNG1: Yes, this is intentional; adapterConfigs takes precedence over submitterConfigs, which is a general configuration.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/Recommender.java
FANNG1: Currently, `Recommender` is used in CLI, so a console summary output is clearer,  but this is not a good design, it's more resonable to use `System.out` in CLI part, I will polish it in later CLI PRs.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/job/GravitinoJobSubmitter.java
tedyu: If either input map is being modified concurrently with the merging (since `ConcurrentMap` is used, I assume concurreny is possible), we may get inconsistent results.
If two entries, (key_a->val_a) and (key_b->val_b), are modified (assuming both keys may appear in both maps).
The update of `key_a` and `key_b` in one input map may only be partially reflected in the merged map.

Issue #9435, opened 2025-12-10
[FEATURE] support external Iceberg format for generic catalog


FANNG1:
### Describe the feature

Gravitino generic catalog only support lance format, it's better to support Iceberg format for the generic catalog

PR #9908, opened 2026-02-06
[#9435] feat(generic-catalog) support external Iceberg format for generic catalog

FANNG1:
##  What changes were proposed in this pull request?

  - Added Iceberg support to the generic lakehouse catalog via a new IcebergTableDelegator
    and IcebergTableOperations, enforcing format=iceberg and external=true, and disallowing
    purge.
  - Registered the new delegator in META-INF/services so it’s discoverable.
  - Added unit tests validating external Iceberg table creation and immutable properties.
  - Added documentation for Iceberg table support and required properties.

##  Why are the changes needed?

  - Enables registering and managing Iceberg tables in the generic lakehouse catalog, with
    clear validation to prevent unsupported managed-table behaviors.

    Fix: #9435 

##  Does this PR introduce any user-facing change?

  - Yes. The generic lakehouse catalog now supports external Iceberg tables (format=iceberg,
    external=true) and rejects purge for Iceberg.
  - New user-facing documentation: docs/lakehouse-generic-iceberg-table.md.

 ## How was this patch tested?

  - Added unit tests and IT

### DIFF
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalogOperations.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalogOperations.java
@@ -254,7 +254,7 @@ public class GenericCatalogOperations implements CatalogOperations, SupportsSche
     if (StringUtils.isNotBlank(tableLocation)) {
       newProperties.put(Table.PROPERTY_LOCATION, ensureTrailingSlash(tableLocation));
     } else {
-      String generatedLocation = calculateTableLocation(schema, ident, properties);
+      String generatedLocation = calculateTableLocation(schema, ident);
       if (StringUtils.isNotBlank(generatedLocation)) {
         newProperties.put(Table.PROPERTY_LOCATION, generatedLocation);
       } else if (tableOps.capabilities().contains(TableCapability.REQUIRES_LOCATION)) {
@@ -302,17 +302,7 @@ public class GenericCatalogOperations implements CatalogOperations, SupportsSche
     return dropped;
   }
 
-  private String calculateTableLocation(
-      Schema schema, NameIdentifier tableIdent, Map<String, String> tableProperties) {
-    String tableLocation =
-        (String)
-            propertiesMetadata
-                .tablePropertiesMetadata()
-                .getOrDefault(tableProperties, Table.PROPERTY_LOCATION);
-    if (StringUtils.isNotBlank(tableLocation)) {
-      return ensureTrailingSlash(tableLocation);
-    }
-
+  private String calculateTableLocation(Schema schema, NameIdentifier tableIdent) {
     String schemaLocation =
         schema.properties() == null ? null : schema.properties().get(Schema.PROPERTY_LOCATION);
 


FANNG1:
@jerryshao @yuqi1129 PTAL, thanks

docs/lakehouse-generic-iceberg-table.md
```diff
@@ -0,0 +1,133 @@
...
+
+| Operation | Support Status  |
+|-----------|-----------------|
+| List      | ✅ Supported     |
+| Load      | ✅ Supported     |
+| Alter     | ✅ Supported     |
+| Create    | ✅ Supported     |
+| Drop      | ✅ Supported     |
+| Purge     | ❌ Not supported |
+
```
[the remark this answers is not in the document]
FANNG1: There is no double `||`

core/src/main/java/org/apache/gravitino/catalog/TableFormatCapability.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.catalog;
+
+public enum TableFormatCapability {
```
jerryshao: This should be put to connector module, the implementor of catalog should also be aware of this.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTableOperations.java
```diff
@@ -0,0 +1,119 @@
...
+
+    return super.createTable(
+        ident, columns, comment, properties, partitions, distribution, sortOrders, indexes);
+  }
+
+  @Override
+  public boolean purgeTable(NameIdentifier ident) {
+    throw new UnsupportedOperationException("Purge is not supported for external Iceberg tables");
+  }
+
```
jerryshao, line 109: Do we support altering table? If so, shall we also update the schema in the Iceberg level, not only in Gravitino metadata?

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalogOperations.java
```diff
@@ -231,23 +232,40 @@ public Table createTable(
...
+    newProperties.put(Table.PROPERTY_TABLE_FORMAT, format);
+
+    Schema schema = loadSchema(NameIdentifier.of(ident.namespace().levels()));
+    String tableLocation =
+        (String)
+            propertiesMetadata
+                .tablePropertiesMetadata()
+                .getOrDefault(properties, Table.PROPERTY_LOCATION);
+    if (StringUtils.isNotBlank(tableLocation)) {
+      newProperties.put(Table.PROPERTY_LOCATION, ensureTrailingSlash(tableLocation));
```
yuqi1129, line 255: `calculateTableLocation` already contains this logic, and it's duplicated.

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
```diff
@@ -537,4 +546,48 @@ int calculateColumnPosition(
       throw new IllegalArgumentException("Unsupported column position: " + position);
     }
   }
+
+  private void validateIndexSupport(Index[] indexes) {
```
yuqi1129: The following three methods can be used by other catalogs. Can you extract a util class to handle it?

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTableOperations.java
FANNG1: I plan to support ALTER operations for external Iceberg tables. Because external tables don't need to sync with the underlying Iceberg table, it's the user's responsibility to sync.  The managed table is used to keep in sync with Gravitino schema and Iceberg schema.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalogOperations.java
FANNG1: seems not duplicated, because `calculateTableLocation` is in another code path

yuqi1129: I mean 
https://github.com/apache/gravitino/blob/c206c0748a9e6b0a96b9eb68e593ef87df3d8e23/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalogOperations.java#L287-L295

core/src/main/java/org/apache/gravitino/catalog/ManagedTableOperations.java
FANNG1: updated

core/src/main/java/org/apache/gravitino/catalog/TableFormatCapability.java
FANNG1: updated

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/generic/GenericCatalogOperations.java
FANNG1: got it, updated

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTableOperations.java
jerryshao: It is weird that ALTER only changes the metadata in the Gravitino side. From the user's perspective, the user assumes that the ALTER operations will actually change the schema, but it is not as they expected when they query the Iceberg table.

jerryshao: I suggest we don't support ALTER until we clarify the behavior.

roryqi:
@jerryshao Do u have other comments?

Issue #9889, opened 2026-02-05
[Improvement] Use safer implementation for cherry-pick workflow


roryqi:
### What would you like to be improved?

 Use safer implementation for cherry-pick workflow

### How should we improve?

 Use safer implementation for cherry-pick workflow

Issue #456, opened 2023-09-27
[Improvement]  rename `catalog-lakehouse` to `catalog-lakehouse-iceberg`


FANNG1:
### What would you like to be improved?

we could unify  catalog loading logic with the `provider` if renaming   `catalog-lakehouse` to `catalog-lakehouse-iceberg`

PR #9910, opened 2026-02-06
[#9889][followup]: Fix the cherry-pick logic

roryqi:

Fix the cherry-pick logic about the format [#123] something (#456)

This is a follow-up PR.

No.

Tested by my repo. https://github.com/qqqttt123/gravitino/pull/19


.github/workflows/auto-cherry-pick.yml
yuqi1129, line 41: Since ${PR_NUMBER} is empty, why do we still output it to GITHUB_OUTPUT?

roryqi: If ${PR_NUMBER} is empty, the next pipeline won't be executed. Because we need use PR_NUMBER to check which label is contained.

Issue #9518, opened 2025-12-22
[Improvement] Batch to get owners for the bundle of metadata objects


roryqi:
### What would you like to be improved?

When we list the bundle of metadata objects, we need to get owners first. Now, we get the owner one by one. 

### How should we improve?

There are two possible ways to solve this issue:
1. Get metadata objects belongs to current user owner 
2. Batch to get metadata objects owner

roryqi:
cc @hdygxsj

hdygxsj:
Before this, I think there's an issue worth discussing first: currently, all Entities retrieve the owner's Entity when obtaining the owner, which leads to unnecessary table join—for example, joining with the user table. However, in reality, GravitinoAuthorizer only needs the ownerId for authorization. We should consider how to optimize this issue.

<img width="794" height="633" alt="Image" src="https://github.com/user-attachments/assets/4836a49d-b991-4eaa-8dce-fa10c69bc17f" />

hdygxsj:
After addressing the aforementioned issue, my main approach is as follows:

Adding batch operations to EntityStore is necessary. We could start by implementing batch operations for certain metadata types—such as tables—but we must carefully consider how to ensure that the cache results remain consistent with those from individual requests.

Next, before performing concurrent authorization in org.apache.gravitino.server.authorization.MetadataAuthzHelper#doFilter, we can invoke a single batch operation on EntityStore to preload all relevant metadata IDs into the cache, thereby ensuring they are available during subsequent parallel authorization checks.

<img width="1680" height="788" alt="Image" src="https://github.com/user-attachments/assets/1591309d-5e92-41f7-b0f0-f4ac9adb8505" />

roryqi:
> Before this, I think there's an issue worth discussing first: currently, all Entities retrieve the owner's Entity when obtaining the owner, which leads to unnecessary table join—for example, joining with the user table. However, in reality, GravitinoAuthorizer only needs the ownerId for authorization. We should consider how to optimize this issue.
> 
[...]

PR #9914, opened 2026-02-08
[#9518] improvement(authz): support batch get owner

hdygxsj:

support batch get owner

Fix: #9518 

None

Existing IT

### DIFF
--- a/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
+++ b/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
@@ -456,7 +456,7 @@ public class GravitinoEnv {
   }
 
   public boolean cacheEnabled() {
-    return config != null && config.get(Configs.CACHE_ENABLED);
+    return config == null || config.get(Configs.CACHE_ENABLED);
   }
 
   public void start() {
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -342,16 +342,6 @@ public class CaffeineEntityCache extends BaseEntityCache {
 
   /** {@inheritDoc} */
   @Override
-  public <E extends Exception> void withMultipleKeyCacheLock(
-      List<EntityCacheKey> keys, EntityCache.ThrowingRunnable<E> action) throws E {
-    Preconditions.checkArgument(keys != null, "Keys cannot be null");
-    Preconditions.checkArgument(action != null, "Action cannot be null");
-
-    segmentedLock.withMultipleKeyLockAndThrow(keys, action);
-  }
-
-  /** {@inheritDoc} */
-  @Override
   public <T, E extends Exception> T withMultipleKeyCacheLock(
       List<EntityCacheKey> keys, EntityCache.ThrowingSupplier<T, E> action) throws E {
     Preconditions.checkArgument(keys != null, "Keys cannot be null");
--- a/core/src/main/java/org/apache/gravitino/cache/SegmentedLock.java
+++ b/core/src/main/java/org/apache/gravitino/cache/SegmentedLock.java
@@ -236,34 +236,19 @@ public class SegmentedLock {
     }
   }
 
-  private List<Lock> getDistinctSortedLocks(List<? extends Object> keys) {
-    return keys.stream()
-        .map(this::getSegmentLock)
-        .distinct()
-        .sorted(Comparator.comparingInt(lockIndices::get))
-        .collect(Collectors.toList());
+  /** Checks if a global operation is currently in progress. */
+  @VisibleForTesting
+  public boolean isClearing() {
+    return globalOperationLatch.get() != null;
   }
 
-  private void acquireAllLocks(List<Lock> locks) {
-    int lockedCount = 0;
-    try {
-      for (Lock lock : locks) {
-        lock.lockInterruptibly();
-        lockedCount++;
-      }
-    } catch (InterruptedException e) {
-      for (int i = lockedCount - 1; i >= 0; i--) {
-        locks.get(i).unlock();
-      }
-      Thread.currentThread().interrupt();
-      throw new IllegalStateException("Thread was interrupted while acquiring batch lock", e);
-    }
-  }
-
-  private void releaseAllLocks(List<Lock> locks) {
-    for (int i = locks.size() - 1; i >= 0; i--) {
-      locks.get(i).unlock();
-    }
+  /**
+   * Returns number of lock segments.
+   *
+   * @return Number of segments
+   */
+  public int getNumSegments() {
+    return stripedLocks.size();
   }
 
   /**
@@ -292,6 +277,38 @@ public class SegmentedLock {
     }
   }
 
+  private List<Lock> getDistinctSortedLocks(List<? extends Object> keys) {
+    return keys.stream()
+        .map(this::getSegmentLock)
+        .distinct()
+        .sorted(Comparator.comparingInt(lockIndices::get))
+        .collect(Collectors.toList());
+  }
+
+  private void acquireAllLocks(List<Lock> locks) {
+    int lockedCount = 0;
+    try {
+      for (Lock lock : locks) {
+        lock.lockInterruptibly();
+        lockedCount++;
+      }
+    } catch (InterruptedException e) {
+      releaseLocks(locks, lockedCount);
+      Thread.currentThread().interrupt();
+      throw new IllegalStateException("Thread was interrupted while acquiring batch lock", e);
+    }
+  }
+
+  private static void releaseLocks(List<Lock> locks, int lockedCount) {
+    for (int i = lockedCount - 1; i >= 0; i--) {
+      locks.get(i).unlock();
+    }
+  }
+
+  private void releaseAllLocks(List<Lock> locks) {
+    releaseLocks(locks, locks.size());
+  }
+
   /**
    * Waits for any ongoing global operation to complete. This method is called by regular operations
    * to ensure they don't interfere with global operations.
@@ -307,20 +324,5 @@ public class SegmentedLock {
             "Thread was interrupted while waiting for global operation to complete", e);
       }
     }
-  }
-
-  /** Checks if a global operation is currently in progress. */
-  @VisibleForTesting
-  public boolean isClearing() {
-    return globalOperationLatch.get() != null;
-  }
-
-  /**
-   * Returns number of lock segments.
-   *
-   * @return Number of segments
-   */
-  public int getNumSegments() {
-    return stripedLocks.size();
   }
 }
--- a/core/src/main/java/org/apache/gravitino/cache/EntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/EntityCache.java
@@ -63,18 +63,6 @@ public interface EntityCache extends SupportsEntityStoreCache, SupportsRelationE
       throws E;
 
   /**
-   * Acquires locks for multiple cache keys and executes the action. Keys are locked in a consistent
-   * order to avoid deadlocks.
-   *
-   * @param keys The cache keys to lock
-   * @param action The action to execute while holding all locks
-   * @param <E> The type of exception that may be thrown
-   * @throws E if the action throws an exception of type E
-   */
-  <E extends Exception> void withMultipleKeyCacheLock(
-      List<EntityCacheKey> keys, ThrowingRunnable<E> action) throws E;
-
-  /**
    * Acquires locks for multiple cache keys and executes the action, returning the result. Keys are
    * locked in a consistent order to avoid deadlocks.
    *
--- a/core/src/main/java/org/apache/gravitino/cache/NoOpsCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/NoOpsCache.java
@@ -73,13 +73,6 @@ public class NoOpsCache extends BaseEntityCache {
 
   /** {@inheritDoc} */
   @Override
-  public <E extends Exception> void withMultipleKeyCacheLock(
-      List<EntityCacheKey> keys, ThrowingRunnable<E> action) throws E {
-    opLock.withMultipleKeyLockAndThrow(keys, action);
-  }
-
-  /** {@inheritDoc} */
-  @Override
   public <T, E extends Exception> T withMultipleKeyCacheLock(
       List<EntityCacheKey> keys, ThrowingSupplier<T, E> action) throws E {
     return opLock.withMultipleKeyLockAndThrow(keys, action);


hdygxsj:
700ms for 5000 tables now
image

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -225,6 +227,27 @@ public <E extends Entity & HasIdentifier> List<E> listEntitiesByRelation(
         });
   }
 
+  @Override
+  public <E extends Entity & HasIdentifier> List<E> listEntitiesByRelation(
```
roryqi: You should return a list of relations, otherwise you can't know the relation between entities.
For example, you list entity1 and entity2 and return entity3, entity4. 
The relation is 
entity1 -> entity3
entity1 -> entity4
entity2 -> entity3
or 
entity1 -> entity3
entity1 -> entity4
entity2 -> entity4
?

hdygxsj: ok, but the owner returns a UserEntity, which doesn't include the original data's ID.

roryqi: Could u elaborate more?

hdygxsj: fixed

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -225,6 +228,30 @@ public <E extends Entity & HasIdentifier> List<E> listEntitiesByRelation(
         });
   }
 
+  @Override
+  public <E extends Entity & HasIdentifier>
+      Map<NameIdentifier, List<E>> batchListEntitiesByRelation(
```
roryqi: We should add a new class Relation instead a nested map. It's hard to maintain.
class Relation {
Entity sourceEntity;
Entity destEntity;
}

LIst<Relation>  batchListEntitiesByRelation(Type relType, List<NameIdentifier> sourceIdents);

hdygxsj: Where should the Relation be placed, and are there any similar data structures available right now? 
Also, just to mention—personally, I find it really hard to understand and accept that retrieving the owner of an entity requires using a method called listEntitiesByRelation. Does an entity really have multiple owners?

hdygxsj:
@roryqi fixed, PHAL

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: 1. You put it in the same level 
f4b0dd54-4e76-4d69-8c6f-d3888ad4cd4f
2. We don't have similar now
3. Entity won't have multiple owners. But we should use common interfaces instead of dedicated interfaces.

hdygxsj: Besides the location of this data structure, are there any other issues?

roryqi: `EntityRelation` should `one to one` instead of `one to multiple`.

hdygxsj: fixed

core/src/main/java/org/apache/gravitino/storage/relational/helper/EntityRelation.java
```diff
@@ -0,0 +1,30 @@
...
+package org.apache.gravitino.storage.relational.helper;
+
+import lombok.Data;
+import org.apache.gravitino.Entity;
+import org.apache.gravitino.NameIdentifier;
+
+@Data
+public class EntityRelation<E extends Entity> {
+
+  private E relationEntity;
```
roryqi: I have concern about this. Do we need entity here? @yuqi1129 WDYT?

roryqi: Maybe sourceNameIdentifier should be the source entity.

hdygxsj: > Maybe sourceNameIdentifier should be the source entity.

Since there is no good way to convert NameIdentify to Entity directly in the service, it is necessary to invoke EntityStore for the conversion. Even after the conversion, the RelationalEntityStore must convert the Entity back to NameIdentify, as storing it in the cache requires using NameIdentify.

roryqi: 1. Actually, you can use batchGet to get all the entities by identifiers
2. Cache already stored the source entity.

roryqi:
@yuqi1129 Could u take a look?

core/src/main/java/org/apache/gravitino/Relation.java
```diff
@@ -0,0 +1,93 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino;
+
+public class Relation {
+  private SupportsRelationOperations.Type type;
+  private final NameIdentifier source;
```
yuqi1129, line 29: The `type` field is the only non-`final` field in this otherwise immutable class. This allows unintended mutation after construction. Please add `final`:
```java
private final SupportsRelationOperations.Type type;
```

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
```diff
@@ -79,6 +90,60 @@ public Optional<Entity> getOwner(NameIdentifier identifier, Entity.EntityType ty
...
+    List<Long> entityIds =
+        identifiers.stream()
+            .map(
+                identifier -> {
+                  long entityId = EntityIdService.getEntityId(identifier, type);
+                  nameIdentifierMap.put(entityId, identifier);
+                  return entityId;
+                })
+            .toList();
+    NameIdentifier nameIdentifier = identifiers.get(0);
```
yuqi1129: Input validation runs **after** all DB I/O (`EntityIdService.getEntityId` is called inside the stream at lines 102–110). If the identifiers span different metalakes the `Preconditions` check throws, but the DB lookups (and their associated transactions/latency) have already happened. Move the metalake-validation block **before** the `entityIds` stream:

```java
// Validate first
if (identifiers.isEmpty()) return new ArrayList<>();
String metalake = NameIdentifierUtil.getMetalake(identifiers.get(0));
for (NameIdentifier identifier : identifiers) {
    Preconditions.checkArgument(
        Objects.equals(NameIdentifierUtil.getMetalake(identifier), metalake),
        "identifiers should be in one metalake");
}
// Then query the DB
List<Long> entityIds = identifiers.stream()...
```

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
```diff
@@ -79,6 +90,60 @@ public Optional<Entity> getOwner(NameIdentifier identifier, Entity.EntityType ty
...
+                new Relation(
+                    SupportsRelationOperations.Type.OWNER_REL,
+                    nameIdentifierMap.get(userPO.getMetadataObjectId()),
+                    type,
+                    userEntity.nameIdentifier(),
+                    Entity.EntityType.USER));
+          });
+    }
+
+    // TODO: Add batch support for group owners when GroupOwnerRelPO and batch method are available
```
yuqi1129, line 141: **Functional gap**: `batchGetOwner` silently returns only USER owners. If any entity in the batch has a GROUP owner it will be **absent from the result** with no error or warning. This breaks the contract of the caller (e.g. `RelationalEntityStore.batchListEntitiesByRelation`) and will populate the cache with incomplete data — future cache hits will incorrectly believe the entity has no owner.

Options:
1. Add the group batch query in this PR (preferred — keeping both code paths consistent).
2. If deferring, throw `UnsupportedOperationException` so callers know they cannot rely on this method for group-owned entities yet, rather than silently returning incomplete results.
3. At minimum, add an explicit note in the Javadoc of `batchListEntitiesByRelation` that group owners are not yet included.

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
```diff
@@ -79,6 +90,60 @@ public Optional<Entity> getOwner(NameIdentifier identifier, Entity.EntityType ty
...
+    Map<Long, NameIdentifier> nameIdentifierMap = new HashMap<>();
+    List<Long> entityIds =
+        identifiers.stream()
+            .map(
+                identifier -> {
+                  long entityId = EntityIdService.getEntityId(identifier, type);
+                  nameIdentifierMap.put(entityId, identifier);
+                  return entityId;
+                })
+            .toList();
```
yuqi1129: `Stream.toList()` was added in Java 16. The rest of the codebase consistently uses `.collect(Collectors.toList())` (e.g. `SegmentedLock.java` in this same PR). Please change to:
```java
.collect(Collectors.toList());
```

core/src/main/java/org/apache/gravitino/storage/relational/JDBCBackend.java
```diff
@@ -350,6 +351,31 @@ public <E extends Entity & HasIdentifier> List<E> batchGet(
       case JOB_TEMPLATE:
         return (List<E>)
             JobTemplateMetaService.getInstance().batchGetJobTemplateByIdentifier(identifiers);
+      case USER:
+        // TODO: I will add batch operations for users, groups and views
```
yuqi1129, line 355: The TODO acknowledges this, but the current implementation issues **N individual SELECT queries** for N identifiers. This defeats the purpose of the batch API: `batchListEntitiesByRelation` calls `batchGet` with all N target entities, which then loops here.

For `batchPopulateRelationCache` in `RelationalEntityStore` to be efficient, the underlying `batchGet` must issue a single IN-clause query. `UserMetaService` already has a `listUsersByNamespace` path; it (or a new `getUsersByIds` method) should be used here instead of the per-identifier loop.

Please either implement the true batch queries in this PR or track this in a follow-up issue and link it here.

core/src/main/java/org/apache/gravitino/cache/SegmentedLock.java
```diff
@@ -176,6 +179,80 @@ public <E extends Exception> void withLockAndThrow(
...
+    } finally {
+      releaseAllLocks(sortedLocks);
+    }
+  }
+
+  private List<Lock> getDistinctSortedLocks(List<? extends Object> keys) {
+    return keys.stream()
+        .map(this::getSegmentLock)
+        .distinct()
+        .sorted(Comparator.comparingInt(System::identityHashCode))
```
yuqi1129: `System.identityHashCode` is not guaranteed to be unique across distinct `Lock` objects — two different locks can share the same identity hash code. When that happens, their relative order becomes non-deterministic across JVM runs, which can re-introduce the deadlock risk the sorted-acquisition is intended to prevent.

A simple fix is to add a tie-breaker using the lock's index in the striped set. Alternatively, since `Striped<Lock>` assigns a stable slot index to each key, sorting by that index is both unique and cheap. If that index isn't easily accessible, at minimum add a secondary sort on `lock.toString()` or the lock object's class identity to break ties deterministically.

core/src/main/java/org/apache/gravitino/Relation.java
yuqi1129, line 29: `type` is the only non-`final` field. All other fields (`source`, `sourceType`, `target`, `targetType`) are final, and there is no setter — the class is intended to be immutable. Please add `final`:
```java
private final SupportsRelationOperations.Type type;
```

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -318,4 +364,64 @@ public <E extends Entity & HasIdentifier> void batchPut(List<E> entities, boolea
...
+    for (Relation relation : backendRelations) {
+      targetsByType
+          .computeIfAbsent(relation.targetType(), k -> new ArrayList<>())
+          .add(relation.target());
+    }
+
+    Map<NameIdentifier, E> entitiesByIdent = new HashMap<>();
+    for (Map.Entry<Entity.EntityType, List<NameIdentifier>> entry : targetsByType.entrySet()) {
+      List<E> entities =
+          batchGet(
```
yuqi1129: **Double round-trip to the DB for the same data.**

`batchListEntitiesByRelation` already retrieved the backend `Relation` objects (which in the case of `OWNER_REL` come from `OwnerMetaService.batchGetOwner`, which **already fetched the full `UserPO` / `UserEntity`** via the JOIN query in `batchSelectUserOwnerMetaByMetadataObjectIdAndType`). Here, the code calls `batchGet` again to re-fetch those same entities purely to put them into the entity cache.

Consider having `backend.batchListEntitiesByRelation` return the entities directly (or as part of the result), or at least pass the already-resolved entities down to avoid the second DB hit. If that's too large a refactor for this PR, please add a comment explaining why the double fetch is acceptable.

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
roryqi: No need. We have add check in other places.

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
roryqi: Done.

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
roryqi: Done.

core/src/main/java/org/apache/gravitino/storage/relational/JDBCBackend.java
roryqi: Removed now.

core/src/main/java/org/apache/gravitino/cache/SegmentedLock.java
roryqi: Done.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: Fixed.

core/src/main/java/org/apache/gravitino/Relation.java
roryqi: Fixed.

core/src/main/java/org/apache/gravitino/cache/SegmentedLock.java
```diff
@@ -176,6 +192,80 @@ public <E extends Exception> void withLockAndThrow(
...
+    } catch (InterruptedException e) {
+      for (int i = lockedCount - 1; i >= 0; i--) {
+        locks.get(i).unlock();
+      }
+      Thread.currentThread().interrupt();
+      throw new IllegalStateException("Thread was interrupted while acquiring batch lock", e);
+    }
+  }
+
+  private void releaseAllLocks(List<Lock> locks) {
```
yuqi1129: This method and code logic in L255-L257 can be extracted as a common method.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -456,7 +456,7 @@ public StatisticDispatcher statisticDispatcher() {
   }
 
   public boolean cacheEnabled() {
-    return config.get(Configs.CACHE_ENABLED);
+    return config != null && config.get(Configs.CACHE_ENABLED);
```
yuqi1129: I remembered that the default value of `CACHE_ENABLED` is true, and that it defaults to false if the config is null.

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -340,6 +340,26 @@ public <E, T extends Exception> E withCacheLock(
     return segmentedLock.withLockAndThrow(key, action);
   }
 
+  /** {@inheritDoc} */
+  @Override
+  public <E extends Exception> void withMultipleKeyCacheLock(
```
yuqi1129: We haven't used this method. Is it for future usage?

roryqi: Removed.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
roryqi: OK, fixed.

core/src/main/java/org/apache/gravitino/cache/SegmentedLock.java
roryqi: OK.Fixed.

yuqi1129 (approved):
I have no more comments and LGTM

Issue #9915, opened 2026-02-08
[Subtask] Add/import IRC views to Gravitino entity store


bharos:
### Describe the subtask

When the views are created in IRC, they should be stored as view entity in Gravitino. This is required for now to manage view level permissions in IRC, and also later to support view APIs in Gravitino

### Parent issue

https://github.com/apache/gravitino/issues/9743

PR #9917, opened 2026-02-09
[#9915]feat(IRC): Add view entities to Gravitino store

bharos:

Add logic to add or import the view entities to Gravitino entity store
This will help maintain consistency between IRC store and Gravitino entity store

This helps to maintain consistency between IRC and Gravitino entity store. It will lazy import entities that are missing in gravitino which are present in IRC catalog

Fix: #9915 

No. It will just import views to Gravitino entity store

Added unit tests

### DIFF
--- a/core/src/test/java/org/apache/gravitino/catalog/TestViewOperationDispatcher.java
+++ b/core/src/test/java/org/apache/gravitino/catalog/TestViewOperationDispatcher.java
@@ -293,22 +293,24 @@ public class TestViewOperationDispatcher extends TestOperationDispatcher {
         futures.add(future);
       }
 
-      // Verify only one entity was imported (no duplicate imports despite concurrent access)
-      // If import was called multiple times, entity IDs would differ across views
-      GenericEntity viewEntity = entityStore.get(viewIdent, VIEW, GenericEntity.class);
-      Assertions.assertNotNull(viewEntity);
-      long entityId = viewEntity.id();
-
       // Verify all concurrent loads succeeded and reference the same imported entity
+      // If import was called multiple times, entity IDs would differ across views
+      Long entityId = null;
       for (Future<View> future : futures) {
         View loadedView = future.get(5, TimeUnit.SECONDS);
         Assertions.assertEquals("concurrent_view", loadedView.name());
 
         EntityCombinedView combinedView = (EntityCombinedView) loadedView;
-        Assertions.assertEquals(
-            entityId,
-            combinedView.viewFromGravitino().id(),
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
```diff
@@ -68,13 +71,86 @@ public ViewOperationDispatcher(
...
-            doWithCatalog(
-                getCatalogIdentifier(ident),
-                c -> c.doWithViewOps(v -> v.loadView(ident)),
-                NoSuchViewException.class));
+    // First load with READ lock to check if view is already imported
+    EntityCombinedView entityCombinedView =
+        TreeLockUtils.doWithTreeLock(ident, LockType.READ, () -> internalLoadView(ident));
+
+    if (!entityCombinedView.imported()) {
+      // Load the schema to make sure the schema is imported.
```
bharos, line 79: @roryqi is this actually needed ? ie. importing the schema before importing the view to make sure schema exists ..?

roryqi: We need to ensure schema exists.

bharos: ok, will keep this then. thanks

bharos:
@roryqi all tests passing now. PTAL at this PR. Thanks!

core/src/main/java/org/apache/gravitino/catalog/EntityCombinedView.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Map;
+import org.apache.gravitino.Audit;
+import org.apache.gravitino.meta.GenericEntity;
+import org.apache.gravitino.rel.View;
+
+/**
+ * A View class to represent a view metadata object that combines the metadata both from {@link
+ * View} and {@link GenericEntity}.
+ */
+public final class EntityCombinedView implements View {
```
tedyu, line 30: equals and hashCode are not defined for this class.
I used script to discover the following:

Class used as Map key or Set element without hashCode():
Policy -> gravitino\bundles\aliyun\src\main\java\org\apache\gravitino\oss\credential\policy\Policy.java

https://github.com/apache/gravitino/issues/10009

These classes have @EqualsAndHashCode annotation - should be fine:
Field -> gravitino\core\src\main\java\org\apache\gravitino\Field.java
BaseMetalake -> gravitino\core\src\main\java\org\apache\gravitino\meta\BaseMetalake.java

Issue #9932, opened 2026-02-10
[Bug report] gravitino-env.sh: the default value of `GRAVITINO_USE_WEB_V2` variable will override the environment variable setting in the container


danhuawang:
### Version

main branch

### Describe what's wrong

When user use gravitino image to deploy the new UI, the default value of `GRAVITINO_USE_WEB_V2` variable will override the environment variable setting in the container. This cause the ` GRAVITINO_USE_WEB_V2=true` setting in the container didn't take effect.

### Error message and/or stacktrace

N/A

### How to reproduce

1. Use gravitino image to deploy the new UI
2. Config `GRAVITINO_USE_WEB_V2=true` for the container

PR #9933, opened 2026-02-10
[#9932] fix(build): remove env GRAVITINO_USE_WEB_V2 default config

LauraXia123:
remove env GRAVITINO_USE_WEB_V2 default config
the default value of GRAVITINO_USE_WEB_V2 variable will override the environment variable setting in the container

N/A

Fix: #9932

N/A

manually


roryqi:
If you don't have issue id, you can use `[minor]`.

Issue #9482, opened 2025-12-15
[Improvement] Fix the error using playground to test rc3 version client


roryqi:
### What would you like to be improved?

When I tests the 1.1.0rc3, it throws an error, I used a python client

![Image](https://github.com/user-attachments/assets/87fb2da7-13fb-4838-b930-3e117df78235)

### How should we improve?

It may be better compatible for 1.1.0rcx

roryqi:
cc @yuqi1129

jerryshao:
@yuqi1129, please fix this if this is an issue.

yuqi1129:
I can't reproduce it with RC4 or the final release

<img width="3094" height="480" alt="Image" src="https://github.com/user-attachments/assets/0f58d00b-ec94-4319-b80a-968145b4af71" />

<img width="1220" height="272" alt="Image" src="https://github.com/user-attachments/assets/8427549e-72e2-4d89-90ea-6b647050ceba" />

roryqi:
> I can't reproduce it with RC4 or the final release
> 
[...]

PR #9939, opened 2026-02-10
[#9482] improvement(common): Enhance version parsing to support release candidates with validation

yuqi1129:

This pull request updates the version parsing logic in `Version.java` to support release candidate (RC) versions and adds tests to ensure correct handling of RC numbers. The main changes include expanding the regex pattern to recognize RC versions, enforcing bounds on RC numbers, and introducing unit tests for these scenarios.

To make it work when verifing release candidate version in the playground.

Fix: #9482 

N/A

UTs

### DIFF
--- a/common/src/main/java/org/apache/gravitino/Version.java
+++ b/common/src/main/java/org/apache/gravitino/Version.java
@@ -36,10 +36,9 @@ public class Version {
 
   private static final int VERSION_PART_NUMBER = 3;
   private static final Pattern PATTERN =
-      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(\\d{1,4})|-.*|\\.([a-zA-Z].*))?$");
+      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(\\d{1,9})|-.*|\\.([a-zA-Z].*))?$");
 
   private static final Version INSTANCE = new Version();
-  private static final int MAX_RC_NUMBER = 1000;
 
   private final VersionInfo versionInfo;
   private final VersionDTO versionDTO;
@@ -97,16 +96,6 @@ public class Version {
 
     Matcher matcher = PATTERN.matcher(versionString);
     if (matcher.matches()) {
-      String releaseCandidateNumber = matcher.group(4);
-      if (releaseCandidateNumber != null) {
-        int rcNumber = Integer.parseInt(releaseCandidateNumber);
-        // Validate RC number is within a supported range for Gravitino versioning semantics.
-        // This is a domain constraint and is not used in numeric version comparison here.
-        Preconditions.checkArgument(
-            rcNumber >= 0 && rcNumber <= MAX_RC_NUMBER,
-            "Invalid RC version string %s, RC number must be between 0 and 1000",
-            versionString);
-      }
       int[] versionNumbers = new int[VERSION_PART_NUMBER];
       for (int i = 0; i < VERSION_PART_NUMBER; i++) {
         versionNumbers[i] = Integer.parseInt(matcher.group(i + 1));
--- a/common/src/test/java/org/apache/gravitino/TestVersion.java
+++ b/common/src/test/java/org/apache/gravitino/TestVersion.java
@@ -34,8 +34,7 @@ public class TestVersion {
 
   @Test
   public void testParseReleaseCandidateOutOfRange() {
-    Assertions.assertThrowsExactly(
-        IllegalArgumentException.class, () -> Version.parseVersionNumber("1.1.0rc1001"));
+    Assertions.assertDoesNotThrow(() -> Version.parseVersionNumber("1.1.0rc1001"));
     Assertions.assertThrowsExactly(
         GravitinoRuntimeException.class, () -> Version.parseVersionNumber("1.1.0rc"));
     Assertions.assertThrowsExactly(


common/src/test/java/org/apache/gravitino/TestVersion.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.gravitino.exceptions.GravitinoRuntimeException;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class TestVersion {
+
+  @Test
+  public void testParseReleaseCandidateVersions() {
+    Assertions.assertArrayEquals(new int[] {1, 1, 0}, Version.parseVersionNumber("1.1.0rc0"));
```
bharos, line 30: So we still don't expose RC version in the version number, right?
Is that intentional ? Or should we be ideally doing that as well

yuqi1129: Yes. 

I'm not 100% sure that we should support the RC version formally. Theoretically, the RC version should not be exposed to the user, and it's only used internally to verify the release, so I'm inclined that it's only for testing.

common/src/main/java/org/apache/gravitino/Version.java
```diff
@@ -36,9 +36,10 @@ public class Version {
 
   private static final int VERSION_PART_NUMBER = 3;
   private static final Pattern PATTERN =
-      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:-.*|\\.([a-zA-Z].*))?$");
+      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(\\d{1,4})|-.*|\\.([a-zA-Z].*))?$");
 
   private static final Version INSTANCE = new Version();
+  private static final int MAX_RC_NUMBER = 1000;
```
roryqi: Why do we need the magic number?

yuqi1129: Just a limit on the max rc number. If this one is not so mandatory, I can remove it.

roryqi: A little weird.

yuqi1129: removed

common/src/main/java/org/apache/gravitino/Version.java
```diff
@@ -36,10 +36,9 @@ public class Version {
 
   private static final int VERSION_PART_NUMBER = 3;
   private static final Pattern PATTERN =
-      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(\\d{1,4})|-.*|\\.([a-zA-Z].*))?$");
+      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(\\d{1,9})|-.*|\\.([a-zA-Z].*))?$");
```
roryqi, line 39: What' the meaning of `{1,9}`?

yuqi1129: Numeric digits are from 1 to 9. For example, it can be `1` to `111111111`(9 times of 1).

yuqi1129:
@roryqi  
Could you take a look again?

common/src/main/java/org/apache/gravitino/Version.java
roryqi: Why do we need this limitation?

yuqi1129: So, the format can be any value with `rc`?

yuqi1129: Or we will ignore the content after `rc`?

bharos: I see that rc is usually added as `-rc` which won't match here?

```
 git tag | grep -E "rc|RC" | head -20  
v0.2.0-rc1
v0.2.0-rc2
v0.2.0-rc3
v0.3.0-rc1
v0.3.0-rc2
v0.3.1-rc1
v0.3.1-rc2
v0.4.0-rc1
v0.4.0-rc2
v0.4.0-rc3
v0.5.0-rc1
v0.5.0-rc2
v0.5.0-rc3
v0.5.1-rc1
v0.5.1-rc2
v0.5.1-rc3
v0.6.0-incubating-rc1
v0.6.0-incubating-rc2
v0.6.0-incubating-rc3
v0.6.0-incubating-rc4
```


Should we do
```
^(\d+)\.(\d+)\.(\d+)(?:-rc(\d{1,9})|-.*|\.([a-zA-Z].*))?$
```

common/src/main/java/org/apache/gravitino/Version.java
bharos: Also we could just say

\d{1,9} = 1 to 9 digits
Simpler: \d+ = 1 or more digits (no upper limit)

If we don't care about any number of digits , to make regex simpler

common/src/main/java/org/apache/gravitino/Version.java
bharos: If we don't care about number of digits in rc, it could be

```
^(\d+)\.(\d+)\.(\d+)(?:-rc(\d+)|-.*|\.([a-zA-Z].*))?$
```

roryqi:
@yuqi1129 Would u like to continue this pull request?

yuqi1129:
> @yuqi1129 Would u like to continue this pull request?

I would like to move on this pull request. The problem is: we have not reached a consensus on the supported format:
Which one is supported and not for the following version:

 - 1.2.0rc0
 - 1.2.1rcxxxxx
 - 1.2.1rc0000000000000000000000000000000000
 - 1.2.0rc-snapshot

roryqi:
> > @yuqi1129 Would u like to continue this pull request?
> 
> I would like to move on this pull request. The problem is: we have not reached a consensus on the supported format: Which one is supported and not for the following version:
> 
> * 1.2.0rc0
> * 1.2.1rcxxxxx
> * 1.2.1rc0000000000000000000000000000000000
> * 1.2.0rc-snapshot

RC shouldn't work with snapshot. `RC` following natural number is OK. `000000` isn't a natural number.

roryqi:
> LGTM

We don't have a consistent conclusion. Why do u merge this pull request? I prefer reverting this commit , reach a consistent conclusion and merge this pull request. @yuqi1129 Could u raise an another pull request?

Issue #9936, opened 2026-02-10
[Improvement] Iceberg Rest Server table operation path uses different standard decoding namespace vs table strings


agnes-xinyi-lu:
### What would you like to be improved?

in this earlier [PR](https://github.com/apache/gravitino/commit/62ed9c34dc0ae9996eb943ab381b46ec15c24584) Gravitino changed namespaces decoding from `percent-encoding` to `application/x-www-form-urlencoded`.  In Iceberg [rest client side](https://github.com/apache/iceberg/blob/d95d9f0ad8baacd1a61331be9c5e09587a89f1f2/core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java#L111) it always uses URLEncoder for both namespace and table strings, should we make it unified and do URLDecoding on table name as well? 

We are hitting some issues when switching traffic from Hive to rest catalog, because some workloads added trailing spaces to the table name, which gets automatically handled by HMS, but through rest catalog, the empty space could not be decoded by gravitino server side, thus can't pass the original table string plus the empty space to HMS, causing "Table Not Exist" error on client side. 

While it's arguable whether we should allow client side to add random empty spaces to the table name during read, but it does feel like a caveat on the server side to not decode the same way client side encodes. 

### How should we improve?

Same way as how namespaces are decoded right now.

agnes-xinyi-lu:
cc @FANNG1 @jerryshao

agnes-xinyi-lu:
I'm not sure what category this falls into,  please help correct if I put the category wrong.

FANNG1:
Yes, I think it's reasonable to decode the table name in IRC server side and keep consistent with the behavior with namespace, do you like to fix it?

agnes-xinyi-lu:
sure, feel free to assign it to me, thanks!

PR #9955, opened 2026-02-11
[#9936] fix(iceberg): URL decodes the table name in iceberg rest server paths

agnes-xinyi-lu:
This change decodes table names in the IcebergTableOperations path through Iceberg RestUtil.decodeString.

Fix: #9936 

No

Through added unittest.

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
@@ -134,16 +134,20 @@ public class IcebergViewOperations {
   public Response loadView(
       @PathParam("prefix") String prefix,
       @Encoded() @PathParam("namespace") String namespace,
-      @PathParam("view") String view) {
+      @Encoded() @PathParam("view") String view) {
     String catalogName = IcebergRESTUtils.getCatalogName(prefix);
     Namespace icebergNS = RESTUtil.decodeNamespace(namespace);
+    String viewName = RESTUtil.decodeString(view);
     LOG.info(
-        "Load Iceberg view, catalog: {}, namespace: {}, view: {}", catalogName, icebergNS, view);
+        "Load Iceberg view, catalog: {}, namespace: {}, view: {}",
+        catalogName,
+        icebergNS,
+        viewName);
     try {
       return Utils.doAs(
           httpRequest,
           () -> {
-            TableIdentifier viewIdentifier = TableIdentifier.of(icebergNS, view);
+            TableIdentifier viewIdentifier = TableIdentifier.of(icebergNS, viewName);
             IcebergRequestContext context =
                 new IcebergRequestContext(httpServletRequest(), catalogName);
             LoadViewResponse loadViewResponse =
@@ -163,15 +167,16 @@ public class IcebergViewOperations {
   public Response replaceView(
       @PathParam("prefix") String prefix,
       @Encoded() @PathParam("namespace") String namespace,
-      @PathParam("view") String view,
+      @Encoded() @PathParam("view") String view,
       UpdateTableRequest replaceViewRequest) {
     String catalogName = IcebergRESTUtils.getCatalogName(prefix);
     Namespace icebergNS = RESTUtil.decodeNamespace(namespace);
+    String viewName = RESTUtil.decodeString(view);
     LOG.info(
         "Replace Iceberg view, catalog: {}, namespace: {}, view: {}, replaceViewRequest: {}",
         catalogName,
         icebergNS,
-        view,
+        viewName,
         SerializeReplaceViewRequest(replaceViewRequest));
     try {
       return Utils.doAs(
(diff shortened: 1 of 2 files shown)


roryqi:
Could u modify the `IcebergTableRenameOperations`, `IcebergViewOperations` and `IcebergViewRenameOperations`, too?

agnes-xinyi-lu:
> Could u modify the `IcebergTableRenameOperations`, `IcebergViewOperations` and `IcebergViewRenameOperations`, too?

@roryqi I will add in `IcebergViewOperations` but can't do rename, because table name is in the payload not path param, we don't need to decode those.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
```diff
@@ -694,6 +695,40 @@ void testRemoteSigningNotSupported(Namespace namespace) {
         "Error message should mention remote signing: " + errorBody);
   }
 
+  @ParameterizedTest
+  @MethodSource("org.apache.gravitino.iceberg.service.rest.IcebergRestTestUtil#testNamespaces")
```
roryqi, line 699: Could we give a better name for `IcebergRestTestUtil#testNamespaces` if we want to reuse this method?

agnes-xinyi-lu: Not sure if I understand the ask. testNamespaces is already being used by every other method in this class, it will touch a lot of places if we rename it. Or do you mean to put the list of tables with special characters as a source in IcebergRestTestUtil as well so other methods could reuse it to test different table names?

roryqi: Option 2 looks better for me. It's a little weird that  the TestIcebergTableOperations test case uses a variable named `testNamespaces` as the source parameter.

roryqi:
Could u clear the unused information in the pull request like as below?
 ```

If you add a feature or fix a bug, add a test to cover your changes.
If you fix a flaky test, repeat it for many times to prove it works.)
```

bharos:
Also, please add the issue ID in
Fix: #xxxx

FANNG1:
Could you also decode the table name in the classes which are used to do the ACLs? 
https://github.com/apache/gravitino/blob/d3271df3572a72eebb23afa5cec06e6ef0e5abdc/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/IcebergMetadataAuthorizationMethodInterceptor.java#L77-L80 . cc @bharos @roryqi similar for view name in ACLs

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
FANNG1: @roryqi seems not the scope of this PR, how about create a seperate issue to track this?

roryqi: This is related about this pull request.  The author adds a test cache using a weird parameter. It's in the scope of the pull request. He just need to add a suitable name variable.

roryqi:
> Could you also decode the table name in the classes which are used to do the ACLs?
> 
> https://github.com/apache/gravitino/blob/d3271df3572a72eebb23afa5cec06e6ef0e5abdc/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/IcebergMetadataAuthorizationMethodInterceptor.java#L77-L80
> 
> . cc @bharos @roryqi similar for view name in ACLs

OK, we can raise a follow-up PR to solve this issue.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
agnes-xinyi-lu: @roryqi I don't think so. I think it's a misunderstanding. I'm not adding a test cache, I'm just adding a test, and I'm not using a weird parameter, every other test in this class uses this parameter, `testNamespaces` is just a const list with two namespace names for the test to use to verify it works in the path. I was just about to explain what @FANNG1 just said, this is out of scope, I'm not creating any new pattern here.

roryqi: Sorry for my mistake. You are right.  It should use different table names as input parameter in my opinion. That's also ok now.

jerryshao:
Shall we backport this to branch-1.1?

roryqi:
> Shall we backport this to branch-1.1?

I think we should. I added the label branch-1.1.

PR #9960, opened 2026-02-11
[#9857] feat(job): Add variable-length arguments support for job templates

### LINKED ISSUE #9857: [Improvement] Add var-length arguments in job template

Current job template argument only supports fixed-length arguments. For example, if the argument list is `--foo bar --foo1 bar1` and if `--foo1` doesn't have a value, we can only use an empty string for `--foo1`, which is inconvenient.

It would be better that we can support var-length arguments, so users can support `-foo bar` and `-foo bar --foo1 bar1`.

### How should we improve?

_No response_

echonesis:

Add support for optional arguments in job templates using `?` prefix marker.

- Modified `JobManager.buildArgumentsWithOptional()` to filter optional arguments
- Arguments prefixed with `?` are removed if their placeholder value is empty/null
- Supports flag-value pairing (e.g., `?--strategy ?{{strategy}}`)
- Updated `JobTemplate` JavaDoc with usage examples

Current job templates require fixed-length arguments. 
Users must provide empty strings for optional parameters, resulting in useless arguments like `--strategy ""`.

Fix: #9857

Yes. Template creators can mark arguments as optional with `?` prefix:
```java
.withArguments(Lists.newArrayList(
    "--catalog", "{{catalog}}",      // Required
    "?--strategy", "?{{strategy}}"   // Optional
))
```

- Unit Test: `TestJobManager`
- Integration Test: `TestJobTemplate` and `JobIT`


jerryshao:
Hi @echonesis thanks a lot for your work and sorry for the late response. Would you please address the comments from Copilot?

jerryshao:
I think there're some missing pieces:

1. We also need to update the Python intgration test to make sure it also works from Python client.
2. We should update the doc to explain the usage pattern to users.

echonesis:
> I think there're some missing pieces:
> 
> 1. We also need to update the Python intgration test to make sure it also works from Python client.
> 2. We should update the doc to explain the usage pattern to users.

Thanks @jerryshao for pointing them out.
I will update them ASAP.

jerryshao (commented):
## Review of PR #9960 — feat(job): Add variable-length arguments support for job templates

### Summary
This PR introduces a clean and useful feature: the `?` prefix marks job template arguments as optional, filtering them out at runtime when their resolved value is empty. The core logic in `buildArgumentsWithOptional()` is generally well-structured, the JavaDoc is detailed, the docs table is clear, and the unit tests cover the primary scenarios well.

There are a few correctness and design concerns to address before merging.

### Issues

**1. Pair detection depends on both args carrying `?` — but this is not enforced**

The flag/value pair skip (`?--flag ?{{val}}` → both dropped when `val` is absent) only fires when the adjacent arg *also* starts with `?`. A template author who writes `["?--flag", "{{val}}"]` (optional flag, non-optional value) will find `?--flag` is **always** emitted, because the pair check never fires. Likewise, a `?--flag` at the very end of the list is always included (bounds check short-circuits).

At a minimum, the JavaDoc on the `arguments` field and the user docs should explicitly state that **both elements of a flag/value pair must carry `?`**. Better still, add a validation step at template registration time that warns (or rejects) when a `?`-prefixed literal is followed by a non-optional arg.

**2. `isEmptyValue` uses `.matches()` — compound unreplaced placeholders slip through**

`PLACEHOLDER_PATTERN.matcher(value).matches()` only returns `true` for a pure `{{key}}` string. An arg template such as `?{{prefix}}/{{suffix}}` resolves to `"{{prefix}}/{{suffix}}"` when both keys are absent, and `isEmptyValue` returns `false` — the raw placeholder string ends up in the argument list. Consider replacing `.matches()` with a check that detects whether the resolved string consists entirely of unreplaced placeholder material (see inline comment).

**3. Missing test coverage for edge cases**

See inline comment on `testBuildArgumentsWithOptional` — three specific scenarios are not covered and each has a different outcome under the current implementation.

### Positive notes
- The distinction between flag/value pairs (`?--flag ?{{val}}`) and independent optional placeholders (`?{{opt1}} ?{{opt2}}`) is correctly implemented: the pair logic only fires when `cleanArg` contains no placeholder (`!PLACEHOLDER_PATTERN.matcher(cleanArg).find()`), which correctly avoids treating two consecutive `?{{...}}` args as a pair.
- The `i++; continue;` pair-skip correctly skips the index so the loop's own increment moves past the value arg.
- Test Case 7 specifically guards the regression where absence of `opt2` could incorrectly suppress `opt1`.
- Integration tests in both Java and Python are thorough.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -787,6 +786,71 @@ static String replacePlaceholder(String inputString, Map<String, String> replace
...
+      // If optional and value is empty, skip this argument
+      if (isOptional && isEmptyValue(replacedArg)) {
+        continue;
+      }
+
+      // If this is an optional flag (not a placeholder itself) and the next argument is also
+      // optional but empty, skip both (they form a pair like ?--flag ?{{value}}).
+      // Independent optional placeholders like ?{{opt1}} ?{{opt2}} are handled individually.
+      if (isOptional
+          && !PLACEHOLDER_PATTERN.matcher(cleanArg).find()
```
jerryshao, line 868: **Design gap: `?--flag` is silently always included when the next arg is not also optional.**

The pair-skip only fires when `templateArgs.get(i + 1).startsWith("?")`. A template such as:
```
["?--strategy", "{{strategy}}"]   // note: no ? on the value
```
will always emit `--strategy` even when `strategy` is absent, because the pair check never fires (the non-optional `{{strategy}}` doesn't start with `?`).

Similarly, a standalone `?--flag` at the end of the list is always included because the bounds check `i + 1 < templateArgs.size()` fails silently.

Recommendation: document that the `?` **must be applied to both the flag and its value** for the pair to work, or add a validation check at template registration time that warns when a `?`-prefixed literal is followed by a non-optional arg.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -787,6 +786,71 @@ static String replacePlaceholder(String inputString, Map<String, String> replace
...
+   *     placeholder
+   */
+  @VisibleForTesting
+  static boolean isEmptyValue(@Nullable String value) {
+    if (value == null || value.trim().isEmpty()) {
+      return true;
+    }
+    // Also consider unreplaced placeholders as empty (e.g., "{{key}}")
+    return PLACEHOLDER_PATTERN.matcher(value).matches();
+  }
```
jerryshao, line 965: **`.matches()` only covers a pure single-placeholder string — compound args slip through.**

`PLACEHOLDER_PATTERN.matcher(value).matches()` anchors to the entire string. An arg whose template is `?{{prefix}}/{{suffix}}` (e.g. for a path like `s3://bucket/path`) will resolve to `"{{prefix}}/{{suffix}}"` when both keys are absent. `isEmptyValue` returns `false` here, so the raw placeholder string leaks into the final command.

Consider using `.find()` instead of `.matches()`, or checking whether the resolved string still contains any placeholder via `PLACEHOLDER_PATTERN.matcher(value).find()`:
```java
return PLACEHOLDER_PATTERN.matcher(value).find() && value.replaceAll("\\{\\{[\\w.-]+\\}\\}", "").isBlank();
```
Or more simply: after replacement, if the result still contains `{{...}}` patterns and nothing else meaningful, treat it as empty.

core/src/test/java/org/apache/gravitino/job/TestJobManager.java
```diff
@@ -899,4 +901,125 @@ private JobEntity newJobEntity(String templateName, JobHandle.Status status) {
...
+    Assertions.assertEquals(
+        Lists.newArrayList("--catalog", "hive", "--table", "db.table"), result3);
+
+    // Test Case 4: Whitespace should be filtered
+    Map<String, String> jobConf4 =
+        ImmutableMap.of("catalog", "hive", "table", "db.table", "strategy", "   ");
+
+    List<String> result4 = JobManager.buildArgumentsWithOptional(templateArgs1, jobConf4);
+    Assertions.assertEquals(
+        Lists.newArrayList("--catalog", "hive", "--table", "db.table"), result4);
```
jerryshao, line 941: **Missing edge case tests:**

1. `?--flag` as the **last** argument in the list (pair check silently skips due to bounds) — should be covered to pin the current behavior.
2. `["?--flag", "{{val}}"]` where the value is non-optional — `?--flag` is always emitted; this surprising behavior needs a test to document the expectation.
3. `?{{prefix}}/{{suffix}}` (compound arg with two unreplaced placeholders) — currently leaks as a raw placeholder string; test should document the current behavior or the fix.

jerryshao:
Hi @echonesis can you please reply to the comments, whether you fixed them or not?

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/JobIT.java
```diff
@@ -452,4 +452,93 @@ private String generateTestLibScript() {
...
+        template,
+        ImmutableMap.of("mode", "opt1_only", "optional_arg1", "val1"),
+        JobHandle.Status.SUCCEEDED);
+
+    // Case 3: Both optional args provided — script asserts both flags are received.
+    runJobAndAwait(
+        template,
+        ImmutableMap.of("mode", "both_optionals", "optional_arg1", "val1", "optional_arg2", "val2"),
+        JobHandle.Status.SUCCEEDED);
+  }
```
jerryshao, line 499: Is it possible to evaluate the args for the job, to make sure the args for job are correct. Only verifying the status of the job may not be enough.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -787,6 +793,174 @@ static String replacePlaceholder(String inputString, Map<String, String> replace
...
+   *
+   * <pre>
+   *   Template pattern                          | jobConf           | Result
+   *   -----------------------------------------+-------------------+---------------------------
+   *   ["?--strategy", "?{{strategy}}"]          | {strategy:binpack}| ["--strategy", "binpack"]
+   *   ["?--strategy", "?{{strategy}}"]          | {}                | []
+   *   ["?--strategy", "{{strategy}}"]           | {strategy:binpack}| ["--strategy", "binpack"]
+   *   ["?--strategy", "{{strategy}}"]           | {}                | []  (pair still skipped)
+   *   ["?{{opt1}}", "?{{opt2}}"]                | {opt1:v1}         | ["v1"]  (independent)
+   *   ["?--dry-run"]                            | {}                | ["--dry-run"]  (standalone)
```
jerryshao, line 837: What is the meaning of only make "?" for the argument?

core/src/main/java/org/apache/gravitino/job/JobManager.java
echonesis: Thanks @jerryshao for the review.

I have updated the code to address these gaps:

1. Pair-skip logic: Updated **buildArgumentsWithOptional** to skip the entire pair even if the value argument lacks the `?` marker.
2. Validation check: Added a validation check at template registration time to warn if a `?`-prefixed literal is followed by a non-optional argument.
3. Standalone `?--flag`: Documented that standalone optional flags are intentionally always included, and recommended pairing them with a sentinel placeholder if they need to be conditionally suppressible.

core/src/main/java/org/apache/gravitino/job/JobManager.java
echonesis: Thanks for mentioning this case! You're completely right.

I've updated **isEmptyValue** exactly as you suggested. 
It now uses `.find()` and strips out all placeholder tokens before checking if the remainder is blank
```java
return PLACEHOLDER_PATTERN.matcher(value).find() 
    && value.replaceAll("\\{\\{[\\w.-]+\\}\\}", "").isBlank();
```

api/src/main/java/org/apache/gravitino/job/JobTemplate.java
```diff
@@ -73,7 +73,25 @@ public enum JobType {
...
-  /** The list of arguments for the job template. */
+  /**
+   * The list of arguments for the job template.
+   *
+   * <p>Arguments can be marked as optional by prefixing them with {@code ?}. An optional argument
+   * is filtered out at job execution time if its effective value is considered empty. A value is
+   * empty when it is {@code null}, an empty string, whitespace-only, or an unreplaced placeholder
+   * (i.e., the key was absent from the job configuration).
+   *
+   * <p>Two consecutive optional arguments can form a flag-value pair. When the flag (the first
```
jerryshao, line 84: Can we simplify the rule as all the arguments starting with "?" can be neglected? Let users to guarantee the semantic correctness.

The current way is a little complex; users need to understand different patterns/usages for "?". It increases the educational overhead.

core/src/test/java/org/apache/gravitino/job/TestJobManager.java
echonesis: Thanks for reminding.
All three edge cases in **testBuildArgumentsWithOptional**:
1. Standalone `?--flag`: Added Test Case 10 to cover `?--flag` at the end of the list, verifying it is always emitted.
2. `["?--flag", "{{val}}"]`: Added Test Case 8 to verify that the entire pair is correctly skipped when the value is absent, fixing the previous bug where the flag was leaked.
3. Compound `?{{prefix}}/{{suffix}}`: Added Test Case 11 to document the known limitation where non-whitespace separators (like `/`) cause the raw placeholder string to leak when keys are absent.

api/src/main/java/org/apache/gravitino/job/JobTemplate.java
echonesis: Thanks for your suggestions!
I will update it in the next commit.

jerryshao: I would like to hear your thoughts. Because my thoughts may not be complete, you already touched all the edge cases, so you may have your own idea.

echonesis: IMO, if we simplify the rule to only drop the empty `?` placeholder, we'll hit the "dangling flag" issue, causing CLI execution failures.

Many CLIs require space-separated arguments. If a user writes `["--conf", "?{{spark_conf}}", "my_app.jar"]` and `spark_conf` is empty:

- Without pairing (Simplified): It becomes `["--conf", "my_app.jar"]`. Spark mistakenly uses `"my_app.jar"` as the config value, and the job crashes because the main jar is missing.
- With pairing (Current): Both the flag and empty value drop safely, resulting in `["my_app.jar"]`.

We can't rely on the `=` syntax (e.g., `"?--conf={{spark_conf}}"`) because many tools don't support it for short flags. 

To reduce overhead, I can add this explicit example to the Javadoc/docs. What do you think?

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/JobIT.java
echonesis: Thanks for mentioning.
I will update `JobIT` to verify the actual runtime args, not just the final job status.

core/src/main/java/org/apache/gravitino/job/JobManager.java
echonesis: `?` means the argument is optional.

For a pattern like `?--strategy` followed by `{{strategy}}`, we treat them as one optional flag-value pair: 
- if strategy is absent, both are skipped
- if it is present, both are kept.

For a standalone literal flag like `?--dry-run`, `?` has no practical effect in the current implementation, because there is no following placeholder value to evaluate.

api/src/main/java/org/apache/gravitino/job/JobTemplate.java
jerryshao: > IMO, if we simplify the rule to only drop the empty `?` placeholder, we'll hit the "dangling flag" issue, causing CLI execution failures.
> 
> Many CLIs require space-separated arguments. If a user writes `["--conf", "?{{spark_conf}}", "my_app.jar"]` and `spark_conf` is empty:
> 
> * Without pairing (Simplified): It becomes `["--conf", "my_app.jar"]`. Spark mistakenly uses `"my_app.jar"` as the config value, and the job crashes because the main jar is missing.
> * With pairing (Current): Both the flag and empty value drop safely, resulting in `["my_app.jar"]`.
> 
> We can't rely on the `=` syntax (e.g., `"?--conf={{spark_conf}}"`) because many tools don't support it for short flags.
> 
> To reduce overhead, I can add this explicit example to the Javadoc/docs. What do you think?

Yes, this is a potential issue. But, can we leave this to the user? It is the user's responsibility to correctly use the `?`. My feeling is that even if we have a sophisticated check mechanism, there still could have corner cases. So I'm inclined to leave this to the users to guarantee the right behavior. What do you think?

echonesis: Sorry for the late response.
Sure, I totally understand the concern about complexity. Just want to make sure I'm aligned with your expectation.
If a user writes `["--conf", "?{{spark_conf}}"]` and `spark_conf` is absent, what would the expected behavior be? And is there a recommended way for users to handle this case correctly?

Issue #9964, opened 2026-02-11
[FEATURE] : Add the version segment module to support Trino 469-472


diqiu50:
### Describe the feature

Add the version segment module for 469-472 to support Trino 469 through Trino 472

PR #9965, opened 2026-02-11
[#9964] feat (trino-connector): Add the version segment module to support Trino 469-472

diqiu50:

Add the version segment module for 469-472 to support Trino 469 through Trino 472

Fix: #9964

NO

UT and IT


dev/docker/trino/Dockerfile
```diff
@@ -16,49 +16,11 @@
...
 # under the License.
 #
-FROM trinodb/trino:435
+FROM trinodb/trino:469
 LABEL maintainer="dev@gravitino.apache.org"
 
 USER root
 
-# Only mysql, hudi, iceberg, mariadb, jmx, memory, tpch, tpcds, postgresql, hive plugin are kept
-RUN rm -rf /usr/lib/trino/plugin/accumulo  \
```
yuqi1129, line 25: Why do we remove them?

docs/trino-connector/catalog-hive.md
```diff
@@ -374,7 +374,7 @@ replacing hdfs_user with the appropriate username:
 
 When using AWS S3 within the Hive catalog, users need to configure the Trino Hive connector's
 AWS S3-related properties in the catalog's properteis. Please refer to the documentation
-of [Hive connector with Amazon S3](https://trino.io/docs/435/connector/hive-s3.html).
+of [Hive connector with Amazon S3](https://trino.io/docs/current/connector/hive-s3.html).
```
yuqi1129, line 377: So, which Trino version will be supported in Gravitino by default? The `current` in Trino means the latest version.

docs/trino-connector/development.md
```diff
@@ -1,18 +1,43 @@
...
+├── trino-connector-446-451/      # Version-specific adapters for Trino 446-451
+│   └── src/main/java/
+├── trino-connector-452-468/      # Version-specific adapters for Trino 452-468
+│   └── src/main/java/
+├── trino-connector-469-472/      # Version-specific adapters for Trino 469-472
+│   └── src/main/java/
+└── integration-test/             # Integration tests
+```
+
+Each version-segment module includes the shared base source via Gradle `sourceSets` and adds version-specific adapter classes (e.g., `GravitinoConnector469.java`, `GravitinoPlugin469.java`) to handle Trino SPI differences across versions.
```
yuqi1129, line 31: Can't we use regular import style to use `the shared base source`?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/hive/HiveConnectorAdapter.java
```diff
@@ -56,6 +56,7 @@ public Map<String, String> buildInternalConnectorConfig(GravitinoCatalog catalog
     config.putAll(trinoProperty);
     config.put("hive.metastore.uri", metastoreUri);
     config.put("hive.security", "allow-all");
+    config.put("fs.hadoop.enabled", "true");
```
yuqi1129, line 59: Why do we need this change?

dev/docker/trino/Dockerfile
diqiu50: This operation does not reduce the container image size. It’s not very useful.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/hive/HiveConnectorAdapter.java
diqiu50: HDFS is disabled by default in Trino 469 and later

docs/trino-connector/catalog-hive.md
diqiu50: Users may be using different versions, so we use current here instead. This is just for reference. Users should look for the version they need.

docs/trino-connector/development.md
diqiu50: The `shared base source` requires a special Trino SPI to compile. It needs to be compiled in each version module

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/iceberg/IcebergMetadataAdapter.java
```diff
@@ -117,13 +118,12 @@ public GravitinoTable createTable(ConnectorTableMetadata tableMetadata) {
         toGravitinoTableProperties(
             removeKeys(tableMetadata.getProperties(), ICEBERG_PROPERTIES_TO_REMOVE));
 
-    if (propertyMap.containsKey(ICEBERG_FORMAT_PROPERTY)) {
```
yuqi1129, line 120: Is this line redundant for Trino 469?

diqiu50: No, there was a bug here before.  the FORMAT_VERSION is not set correct.
We need make sure the properties ICEBERG_FORMAT_PROPERTY and FORMAT_VERSION are set.

yuqi1129 (approved):

Issue #9966, opened 2026-02-11
[Improvement] IRC should expire the catalog wrapper timely


roryqi:
### What would you like to be improved?

IRC should expire the catalog wrapper timely if Gravitino alter or drop the catalog

### How should we improve?

IRC should expire the catalog wrapper timely if Gravitino alter or drop the catalog

PR #9969, opened 2026-02-11
[#9966] improvement(iceberg): IRC should expire the catalog wrapper cache timely

roryqi:

RC should expire the catalog wrapper tmely

Fix: #9966

No.

Add a UT.

### DIFF
--- a/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestIcebergCatalogWrapperManagerForREST.java
+++ b/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestIcebergCatalogWrapperManagerForREST.java
@@ -21,18 +21,37 @@ package org.apache.gravitino.iceberg.service;
 import com.google.common.collect.Maps;
 import java.util.Map;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.gravitino.GravitinoEnv;
+import org.apache.gravitino.catalog.CatalogManager;
 import org.apache.gravitino.iceberg.common.ops.IcebergCatalogWrapper;
 import org.apache.gravitino.iceberg.service.authorization.IcebergRESTServerContext;
 import org.apache.gravitino.iceberg.service.provider.IcebergConfigProvider;
 import org.apache.gravitino.iceberg.service.provider.IcebergConfigProviderFactory;
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.mockito.Mockito;
 
 public class TestIcebergCatalogWrapperManagerForREST {
 
   private static final String DEFAULT_CATALOG = "memory";
+
+  @BeforeAll
+  public static void setup() throws IllegalAccessException {
+    // Mock CatalogManager for GravitinoEnv to avoid initialization errors
+    CatalogManager mockCatalogManager = Mockito.mock(CatalogManager.class);
+    FieldUtils.writeField(GravitinoEnv.getInstance(), "catalogManager", mockCatalogManager, true);
+  }
+
+  @AfterAll
+  public static void tearDown() throws IllegalAccessException {
+    // Clean up GravitinoEnv
+    FieldUtils.writeField(GravitinoEnv.getInstance(), "catalogManager", null, true);
+  }
 
   @ParameterizedTest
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/hook/CatalogHookDispatcher.java
```diff
@@ -175,4 +175,9 @@ public void disableCatalog(NameIdentifier ident) throws NoSuchCatalogException {
   public boolean catalogExists(NameIdentifier ident) {
     return dispatcher.catalogExists(ident);
   }
+
+  @Override
+  public void addCatalogCacheRemoveListener(java.util.function.Consumer<NameIdentifier> listener) {
```
yuqi1129: java.util.function.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -304,6 +306,11 @@ public CatalogManager(Config config, EntityStore store, IdGenerator idGenerator)
             .expireAfterAccess(cacheEvictionIntervalInMs, TimeUnit.MILLISECONDS)
             .removalListener(
                 (k, v, c) -> {
+                  for (Consumer<NameIdentifier> listener : removalListeners) {
```
yuqi1129, line 309: `removalListener` will be executed asynchronous and we can make sure when it was called when an entity was removed. I'm worried there might be a timeliness issue with this

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergCatalogWrapperManager.java
```diff
@@ -71,6 +72,19 @@ public IcebergCatalogWrapperManager(
                             .setNameFormat("iceberg-catalog-wrapper-cleaner-%d")
                             .build())))
             .build();
+    IcebergRESTServerContext context = IcebergRESTServerContext.getInstance();
+    if (context.isAuxMode()) {
+      GravitinoEnv.getInstance()
+          .catalogDispatcher()
+          .addCatalogCacheRemoveListener(
+              ident -> {
+                if (ident != null) {
```
yuqi1129: You have checked the null problem of `ident`.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: Another point is that we need to distinguish whether the catalog was removed by users or by timeout eviction. I believe we only need to handle the former case.

core/src/main/java/org/apache/gravitino/catalog/CatalogDispatcher.java
```diff
@@ -19,10 +19,26 @@
...
+   *
+   * <p>Note: Cache eviction is invoked asynchronously but uses a single thread to process removal
+   * events. To avoid blocking the eviction thread and delaying subsequent cache operations,
+   * listeners should avoid performing heavy operations (such as I/O, network calls, or complex
+   * computations) directly. Instead, consider offloading heavy work to a separate thread or
+   * executor.
+   *
+   * @param listener The consumer to be called with the NameIdentifier of the removed catalog.
+   */
+  void addCatalogCacheRemoveListener(Consumer<NameIdentifier> listener);
```
yuqi1129: This method is not so general in the interface and is only targeted for the implementation that use cache that supports the remove-listener mechanism. Is there a better name or way to get the goals?

core/src/main/java/org/apache/gravitino/hook/CatalogHookDispatcher.java
roryqi: OK, I will fix this.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergCatalogWrapperManager.java
roryqi: OK, I will remove this.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
roryqi: We should handle the second case, too. If we drop a catalog which doesn't exist in the cache, otherwise we can't trigger the cache expiration.

core/src/main/java/org/apache/gravitino/catalog/CatalogDispatcher.java
roryqi: Changed. I only added this method for CatalogManager.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
roryqi: We can remove synchronously. But I have a little concern about close operation. It will be ok only if we can remove the key. It won't have a high delay. There should't have many catalog expiration operations.

yuqi1129: I see.

Issue #9809, opened 2026-01-27
[Bug report] Set Owner failed occasionally due to  NoSuchEntityException: No such schema entity: auth_hive_schema1


danhuawang:
### Version

main branch

### Describe what's wrong

```
{
    "code": 1003,
    "type": "NotFoundException",
    "message": "Failed to operate table(s) [f] operation [CREATE] under schema [auth_hive_schema1], reason [Metadata object catalog_hive.auth_hive_schema1.f or owner anonymous is not found]",
    "stack": [
        "org.apache.gravitino.exceptions.NotFoundException: Metadata object catalog_hive.auth_hive_schema1.f or owner anonymous is not found",
        "\tat org.apache.gravitino.authorization.OwnerManager.setOwner(OwnerManager.java:128)",
        "\tat org.apache.gravitino.authorization.OwnerEventManager.setOwner(OwnerEventManager.java:62)",
        "\tat org.apache.gravitino.hook.TableHookDispatcher.createTable(TableHookDispatcher.java:84)",
        "\tat org.apache.gravitino.catalog.TableNormalizeDispatcher.createTable(TableNormalizeDispatcher.java:78)",
        "\tat org.apache.gravitino.listener.TableEventDispatcher.createTable(TableEventDispatcher.java:136)",
        "\tat org.apache.gravitino.server.web.rest.TableOperations.lambda$createTable$1(TableOperations.java:142)",
        "\tat java.base/java.security.AccessController.doPrivileged(AccessController.java:712)",
        "\tat java.base/javax.security.auth.Subject.doAs(Subject.java:439)",
        "\tat org.apache.gravitino.utils.PrincipalUtils.doAs(PrincipalUtils.java:44)",
        "\tat org.apache.gravitino.server.web.Utils.doAs(Utils.java:202)",
        "\tat org.apache.gravitino.server.web.rest.TableOperations.createTable(TableOperations.java:134)",
        "\tat org.apache.gravitino.server.web.rest.TableOperations_$$_jvst862_8._d3createTable(TableOperations_$$_jvst862_8.java)",
        "\tat java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)",
        "\tat java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)",
[...]

PR #9982, opened 2026-02-12
[#9809] fix(authz): Import the schema to avoid the setting owners failure

roryqi:

Import the schema to avoid the setting owners failure

Fix: #9809

No. 

Added tests.


server/src/main/java/org/apache/gravitino/server/web/rest/TopicOperations.java
```diff
@@ -139,6 +142,12 @@ public Response createTopic(
             NameIdentifier ident =
                 NameIdentifierUtil.ofTopic(metalake, catalog, schema, request.getName());
 
+            // Make sure schema is imported, otherwise set owner for the topic may fail.
+            MetadataObjectUtil.checkMetadataObject(
```
yuqi1129, line 146: Should we need to change `FilesetOperations` and `modelOperations` as well?

roryqi: No need. `Fileset` and `model`'s entity is managed. We don't import the managed entity. So we don't change them as well.

Issue #9747, opened 2026-01-19
[Subtask] Add IRC authorization expressions for REST API


roryqi:
### Describe the subtask

Add IRC authorization expressions for REST API

### Parent issue

#9743

roryqi:
@bharos Do u have interest to do this subtasks?

bharos:
Yes I'll take this one

PR #9988, opened 2026-02-12
[#9747] feat(IRC):  Add authorization for Iceberg views

bharos:

We are adding authorization for IRC view APIs.
Implementing fine-grained access control for view operations (create, load, replace, drop, rename, list) with privilege checks similar to table authorization.
Also, in load table, we have modified the logic to check viewExists, because spark calls loadTable first while loading a view and if not exists it falls back to loading view.

Add auth for view IRC apis

Fix: #9747 

No

Added unit and integration tests

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
@@ -40,8 +40,7 @@ public class AuthorizationExpressionConstants {
                   ANY_USE_CATALOG && ANY_USE_SCHEMA  && (TABLE::OWNER || ANY_SELECT_TABLE || ANY_MODIFY_TABLE)
                   """;
 
-  // Used by IcebergTableOperationExecutor to allow CREATE_TABLE holders to load
-  // tables they just created. Spark calls loadTable immediately after createTable.
+  //  Adding ANY_CREATE_TABLE here as Spark calls tableExists before creating a table.
   public static final String ICEBERG_LOAD_TABLE_AUTHORIZATION_EXPRESSION =
       """
                   ANY(OWNER, METALAKE, CATALOG) ||


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -122,13 +123,39 @@ public void dropTable(
   @Override
   public LoadTableResponse loadTable(
       IcebergRequestContext context, TableIdentifier tableIdentifier) {
+    String catalogName = context.catalogName();
+
+    // Per Iceberg REST spec, /tables/ endpoint should only serve tables, not views.
+    // Check if the identifier is a view and throw NoSuchTableException to trigger
+    // Spark's fallback logic to use /views/ endpoint instead.
+    if (icebergCatalogWrapperManager.getCatalogWrapper(catalogName).viewExists(tableIdentifier)) {
```
roryqi: We should use tableExists here.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -122,13 +123,39 @@ public void dropTable(
   @Override
   public LoadTableResponse loadTable(
       IcebergRequestContext context, TableIdentifier tableIdentifier) {
+    String catalogName = context.catalogName();
+
+    // Per Iceberg REST spec, /tables/ endpoint should only serve tables, not views.
```
roryqi: Could u put these logic into `LoadTableAuthzHandler`?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
```diff
@@ -89,6 +105,13 @@ public Response listView(
                 new IcebergRequestContext(httpServletRequest(), catalogName);
             ListTablesResponse listTablesResponse =
```
roryqi, line 106: Why is this `ListTablesResponse`? Is it correct?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/RESTService.java
```diff
@@ -113,6 +113,7 @@ private void initServer(IcebergConfig icebergConfig) {
     }
     IcebergTableEventDispatcher icebergTableEventDispatcher =
         new IcebergTableEventDispatcher(icebergTableOperationDispatcher, eventBus, metalakeName);
+
```
roryqi: Remove unnecessary change.

roryqi:
You should cover this case, too. https://github.com/apache/gravitino/pull/9955#issuecomment-3890642269

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -122,13 +122,15 @@ public void dropTable(
   @Override
   public LoadTableResponse loadTable(
       IcebergRequestContext context, TableIdentifier tableIdentifier) {
+    String catalogName = context.catalogName();
```
roryqi: This change seems unnecessary.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/IcebergMetadataAuthorizationMethodInterceptor.java
```diff
@@ -80,6 +80,10 @@ protected Map<Entity.EntityType, NameIdentifier> extractNameIdentifierFromParame
               NameIdentifierUtil.ofTable(
                   metalakeName, catalog, schema, RESTUtil.decodeString(value)));
           break;
+        case VIEW:
+          nameIdentifierMap.put(
+              EntityType.VIEW, NameIdentifierUtil.ofView(metalakeName, catalog, schema, value));
```
roryqi: You should cover this case. https://github.com/apache/gravitino/pull/9955#issuecomment-3890642269

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/RESTService.java
bharos: done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
bharos: I think so, seems like we don't have view specific response ..

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: Done

This required a refactor, to be able to access the catalogWrapperManager in LoadTableAuthzHandler.
For this, I had to inject catalogWrapperManager into IcebergRESTServerContext

But catalogWrapperManager and IcebergRESTServerContext had a circular depenency, so I had to pass  auxMode, metalakeName to the catalogWrapperManager to avoid depending on the serverContext

Hope this is ok.. ?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/IcebergMetadataAuthorizationMethodInterceptor.java
bharos: Decoded the table/view name correctly

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
```diff
@@ -131,10 +161,17 @@ public Response createView(
   @Produces(MediaType.APPLICATION_JSON)
   @Timed(name = "load-view." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
   @ResponseMetered(name = "load-view", absolute = true)
+  @AuthorizationExpression(
+      expression =
+          "ANY(OWNER, METALAKE, CATALOG) || "
+              + "SCHEMA_OWNER_WITH_USE_CATALOG || "
+              + "ANY_USE_CATALOG && ANY_USE_SCHEMA && (VIEW::OWNER || ANY_SELECT_VIEW)",
```
roryqi: Do we need `create view` privilege here?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/LoadTableAuthzHandler.java
```diff
@@ -58,7 +65,7 @@ public void process(Map<EntityType, NameIdentifier> nameIdentifierMap) {
       IcebergAuthorizationMetadata icebergMetadata =
           parameter.getAnnotation(IcebergAuthorizationMetadata.class);
       if (icebergMetadata != null && icebergMetadata.type() == RequestType.LOAD_TABLE) {
-        tableName = String.valueOf(args[i]);
+        tableName = RESTUtil.decodeString(String.valueOf(args[i]));
```
roryqi, line 78: Why do u put the decode logic here instead of Interceptor?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: @roryqi Actually, checking tableExists seems hard.
It then means, a user who doesn't have privilege to even check existence of a table will get 404 (when they should get 403)

So I think it's better to check viewExists() then return 404, so that users who have no permissions will get 403 for non-existent tables as well

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/LoadTableAuthzHandler.java
bharos: I tried to change it now but it was hard to do it in a generic way.

Namespace parameters require RESTUtil.decodeNamespace()(special parsing) vs table names need RESTUtil.decodeString(). Decoding in the handler keeps the logic simple and localized - each handler decodes only what it needs with the appropriate method. 

Maybe we can revisit this in future PR if you have any idea ?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
bharos: The tests were passing without it.
But looking at the code, I am not 100% sure, but I do see in some places it calls viewExists in createView
like here : https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/jdbc/JdbcViewOperations.java#L191C18-L191C28

So I think it should be added

roryqi: This is JDBC backend. We should check the REST backend.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
roryqi: This seems not be changed.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
roryqi: OK, just a little weird. I need to ensure this point. But this is out of this pull request scope.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/LoadTableAuthzHandler.java
roryqi: Could u get this value from `nameIdentifierMap`?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
roryqi: So I prefer using different expressions with different actions.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
roryqi: OK for me now.

roryqi:
Could u modify the title and pull request description?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
roryqi: Now, this is just a work around.

roryqi: We can fix this in the next pull request.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/LoadTableAuthzHandler.java
bharos: We create the nameIdentifier map within this method in Line 117

So it's not straightforward. The table parameter has @IcebergAuthorizationMetadata(type = LOAD_TABLE) but NOT @AuthorizationMetadata(type = TABLE), so the interceptor doesn't automatically decode and add it to nameIdentifierMap

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: Sorry, reverted now

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
bharos: Added TODO with your suggestion

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
bharos: is this the correct code for REST backend: 
https://github.com/apache/iceberg/blob/de3125afe64fc2b171a52b6e884c72f901e3cba1/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java#L113

I guess in createTable , `catalog.tableExists` was called inside stageTableCreate.
But for views, I don't find that happening. But I think we can still keep `ANY_CREATE_VIEW` to be consistent with the table auth

bharos:
> Could u modify the title and pull request description?

Done

server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
```diff
@@ -40,6 +40,15 @@ public class AuthorizationExpressionConstants {
                   ANY_USE_CATALOG && ANY_USE_SCHEMA  && (TABLE::OWNER || ANY_SELECT_TABLE || ANY_MODIFY_TABLE)
                   """;
 
+  // Used by IcebergTableOperationExecutor to allow CREATE_TABLE holders to load
+  // tables they just created. Spark calls loadTable immediately after createTable.
```
roryqi: check tableExists before creating the table.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/LoadTableAuthzHandler.java
roryqi: OK.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergViewOperations.java
roryqi: OK. Got it.

roryqi:
Just one nit.

server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
bharos: Done.
Thanks @roryqi

bharos:
Thanks for reviewing this @roryqi

Issue #9745, opened 2026-01-19
[Subtask] Add a view manager for Gravitino


jerqi:
### Describe the subtask

Add a view manager for Gravitino

### Parent issue

#9743

roryqi:
Do u have interest about this sub tasks. I will add more detailed information. @bharos

bharos:
Sure. Assigning to myself. Please add any detailed info in here that can help, thanks!

PR #10000, opened 2026-02-13
[#9745][followup] fix(core): Fix ReverseIndexRules for GenericEntity to skip entities with namespace

bharos:

Fixed GENERIC_METADATA_OBJECT_REVERSE_RULE in ReverseIndexRules.java to only process GenericEntity objects without namespace (policy metadata objects). Added a namespace check (genericEntity.namespace() == null || genericEntity.namespace().isEmpty()) to skip entities with namespace.

Views don't support tags or policies (see TagManager.SUPPORTED_METADATA_OBJECT_TYPES_FOR_TAGS), so they should NOT be indexed in the reverse cache at all. Creating reverse index entries for views wastes memory and has no purpose.
When views are cached as GenericEntity (during auto-import in ViewOperationDispatcher.loadView(), they were triggering GENERIC_METADATA_OBJECT_REVERSE_RULE which was trying to create meaningless reverse index entries.

Fix: #9745

No. This is an internal cache optimization fix.

Added test


core/src/main/java/org/apache/gravitino/cache/ReverseIndexRules.java
```diff
@@ -117,15 +117,23 @@ public class ReverseIndexRules {
...
-        // Name in GenericEntity contains no metalake.
         GenericEntity genericEntity = (GenericEntity) entity;
         EntityType type = entity.type();
         if (genericEntity.name() != null) {
-          String[] levels = genericEntity.name().split("\\.");
-          String metalakeName = key.identifier().namespace().levels()[0];
-          NameIdentifier objectNameIdentifier =
-              NameIdentifier.of(ArrayUtils.addFirst(levels, metalakeName));
-          reverseIndexCache.put(objectNameIdentifier, type, key);
+          // For entities with a non-empty namespace (e.g., views), use the full identifier directly
```
roryqi: Tag and policy shouldn't support view now.

roryqi: You can see TagManger
```
  private static final Set<MetadataObject.Type> SUPPORTED_METADATA_OBJECT_TYPES_FOR_TAGS =
      Sets.newHashSet(
          MetadataObject.Type.CATALOG,
          MetadataObject.Type.SCHEMA,
          MetadataObject.Type.TABLE,
          MetadataObject.Type.FILESET,
          MetadataObject.Type.TOPIC,
          MetadataObject.Type.COLUMN,
          MetadataObject.Type.MODEL);
```

core/src/main/java/org/apache/gravitino/cache/ReverseIndexRules.java
```diff
@@ -117,15 +117,23 @@ public class ReverseIndexRules {
   // the values are policies/tags.
   public static final ReverseIndexCache.ReverseIndexRule GENERIC_METADATA_OBJECT_REVERSE_RULE =
       (entity, key, reverseIndexCache) -> {
-        // Name in GenericEntity contains no metalake.
         GenericEntity genericEntity = (GenericEntity) entity;
         EntityType type = entity.type();
         if (genericEntity.name() != null) {
```
roryqi: You can add the condition `genericEntity.namespace is null `here.

bharos: Thanks for the suggestion. Updated the PR,  and description

Issue #10009, opened 2026-02-15
[Bug report] Certain class used in Set doesn't have equals or hashCode


tedyu:
### Version

main branch

### Describe what's wrong

The following class should have equals and hashCode methods because they're used as Map key class or in Set:

org.apache.gravitino.oss.credential.policy.Policy

### Error message and/or stacktrace

The symptom for the omission would be subtle. It depends on whether objects with duplicate field values are present.

### How to reproduce
```
clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/PolicyIT.java:
    Set<Policy> nonInheritedPolicies =
```
If two Policy objects with duplicate field values are put in the same Set, both of them would be present.

mchades:
Would you like to submit a PR to fix it?

PR #10012, opened 2026-02-16
[#10009] Fix: Add equals and hashCode methods to Policy.java

tedyu:

This change uses the lombok annotation to add equals() and hashCode() methods to Policy.java

Without equals() and hashCode() methods, Policy class shouldn't be used in a `Set` because objects with duplicate fields are allowed to co-exist in the `Set`.

Fix: #10009

No.

Existing IT test.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/ViewMetaBaseSQLProvider.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/ViewMetaBaseSQLProvider.java
@@ -147,6 +147,7 @@ public class ViewMetaBaseSQLProvider {
     return "UPDATE "
         + TABLE_NAME
         + " SET view_name = #{newViewMeta.viewName}, "
+        + " schema_id = #{newViewMeta.schemaId}, "
         + " current_version = #{newViewMeta.currentVersion}, "
         + " last_version = #{newViewMeta.lastVersion}, "
         + " deleted_at = #{newViewMeta.deletedAt} "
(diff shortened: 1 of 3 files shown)


tedyu:
If lombok dependency is not to be added, the following classes need to be modified:
Policy, Statement, Condition, StringLike

From Statement.java
```
  @JsonProperty("Action")
  private List<String> actions;
```
For the List, do we just compare the two Lists sequentially or, do we need to put the String's in a Set and compare ?
Similar question for:
```
  @JsonProperty("Statement")
  private List<Statement> statements;
```

roryqi:
Could we add some check for similar issues?

tedyu:
[hashcode-detect.py](https://github.com/user-attachments/files/25353235/hashcode-detect.py)
This is the program I used to find this class.

roryqi:
> [hashcode-detect.py](https://github.com/user-attachments/files/25353235/hashcode-detect.py) This is the program I used to find this class.

Does the Google `find bugs` have similar rules? I am not sure whether it is ok. https://code.google.com/p/findbugs/source/browse/findbugs/src/java/edu/umd/cs/findbugs/

roryqi (approved):
LGTM, Actually, this `Policy` is located in the package `bundle/aliyun` module. `PolicyIT`'s `Policy` is located in the module `core`. They are different classes.  
It's also ok to add `equals` and `hashCode` methods, it won't bring the side effect.  If we add this Policy to the set, it will benefit. So I merged this contribution.

@tedyu Thanks for your contribution! Welcome to join the Gravitino community.

Issue #9746, opened 2026-01-19
[Subtask] Add view entities management


roryqi:
### Describe the subtask

Add view entities management

There are two ways to support view entities.
1. We can reuse the `table_meta`. But we will add a column `type` to judge the meta is a table or a view. Otherwise, you will list the views when you list tables.

```
CREATE TABLE IF NOT EXISTS `table_meta` (
    `table_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'table id',
    `table_name` VARCHAR(128) NOT NULL COMMENT 'table name',
    `metalake_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'metalake id',
    `catalog_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'catalog id',
    `schema_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'schema id',
    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'table audit info',
    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'table current version',
    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'table last version',
    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'table deleted at',
    PRIMARY KEY (`table_id`),
    UNIQUE KEY `uk_sid_tn_del` (`schema_id`, `table_name`, `deleted_at`),
    KEY `idx_mid` (`metalake_id`),
    KEY `idx_cid` (`catalog_id`)
    ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin COMMENT 'table metadata';
```

2. We can add a view entity referring to table entity like 
```
CREATE TABLE IF NOT EXISTS `view_meta` (
    `view_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'view id',
    `view_name` VARCHAR(128) NOT NULL COMMENT 'view name',
    `metalake_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'metalake id',
    `catalog_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'catalog id',
    `schema_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'schema id',
    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'table audit info',
[...]

PR #10013, opened 2026-02-17
[#9746][followup]fix(core): Fix view schema_id not updated during cross-namespace rename

bharos:

Fix a bug in the view entity backend where `schema_id` was not being updated when a view is renamed across namespaces (schemas).

When a view is renamed from one schema to another (e.g., `schema1.myview` → `schema2.myview`), the `schema_id` foreign key in the `view_meta` table remained unchanged. This caused:
- Entity lookups by the new `NameIdentifier` to fail (wrong schema_id)
- Ownership queries to return empty results
- Can cause authorization checks to fail

Fix: #9746 

No. This fixes a bug in internal metadata management that would have caused errors during cross-namespace view renames.

Unit test


bharos:
Thanks @roryqi !

PR #10020, opened 2026-02-22
[#9747][followup]feat(authz): Add view auth to IcebergRESTAuthInterceptionService

bharos:

Add view authz classes to IcebergRESTAuthInterceptionService.
Also update the ICEBERG_LOAD_VIEW_AUTHORIZATION_EXPRESSION to have ANY_SELECT_TABLE / ANY_MODIFY_TABLE / ANY_CREATE_TABLE 

This is to enable authorization for views

Fix: #9747 

No

Update integration tests


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/IcebergMetadataAuthorizationMethodInterceptor.java
```diff
@@ -81,10 +81,16 @@ protected Map<Entity.EntityType, NameIdentifier> extractNameIdentifierFromParame
                   metalakeName, catalog, schema, RESTUtil.decodeString(value)));
```
roryqi, line 81: Should we put view here, too?

bharos: I think it is not required, because the load-table auth expression doesn't have view related checks
The LoadTableAuthzHandler only does
```
if (wrapperManager.getCatalogWrapper(catalog).viewExists(tableIdentifier)) {
```
which doesn't need adding VIEW entity

Also, Spark always probes /views/name for every relation, including pure tables. But reverse is not true. It doesn't simply check /tables/name endpoint for views, so this not needed for now

roryqi: OK, maybe we can add it in our refactor pull request.

PR #10023, opened 2026-02-23
fix(iceberg-rest): StaticIcebergConfigProvider respects gravitino-metalake config for auth

bharos:

Fix: override getMetalakeName() to read the gravitino-metalake config property (IcebergConstants.GRAVITINO_METALAKE), falling back to the default only if unset. This mirrors the existing behavior in DynamicIcebergConfigProvider.

When authorization is enabled, IcebergRESTServerContext calls getMetalakeName() at startup to determine which Gravitino metalake to check privileges against. StaticIcebergConfigProvider did not override this method, so it always returned the hardcoded default "gravitino" regardless of configuration. This caused every Iceberg REST request to fail with 403 when auth was enabled and the actual metalake name was anything other than "gravitino".

Fix: #(issue)

no

Added unit test


roryqi:
Authorization should enable DynamicIcebergConfigProvider.

Issue #10035, opened 2026-02-25
Test build failed in mcp-server because of FastMCP.tool().


diqiu50:
### Version

main branch

### Describe what's wrong

Failed to build tests in the mcp-server module due to an issue with the FastMCP.tool() function.

### Error message and/or stacktrace

----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/gravitino/gravitino/mcp-server/tests/unit/tools/test_catalog.py", line 36, in setUp
    server = GravitinoMCPServer(Setting("mock_metalake"))
  File "/home/runner/work/gravitino/gravitino/mcp-server/mcp_server/server.py", line 99, in __init__
    load_tools(self.mcp)
  File "/home/runner/work/gravitino/gravitino/mcp-server/mcp_server/tools/__init__.py", line 41, in load_tools
    load_tag_tool(mcp)
  File "/home/runner/work/gravitino/gravitino/mcp-server/mcp_server/tools/tag.py", line 22, in load_tag_tool
    @mcp.tool(tags={"tag"}, enabled=False)
TypeError: FastMCP.tool() got an unexpected keyword argument 'enabled'

### How to reproduce

Rebuild the  `mcp-server` test

yuqi1129:
It was caused by the incompatible changes in 3.0.x that released recently.

PR #10038, opened 2026-02-26
[#10035] fix(mcp-server): Use a concrete version of `fastmcp` to avoid CI error as 3.0.x makes breaking changes(cherry-pick)

yuqi1129:

This pull request makes a small but important update to the `mcp-server/pyproject.toml` dependency configuration. The change temporarily pins the `fastmcp` package to version `2.14.5` due to incompatible changes in the newer `3.0.x` releases.

* Dependency management: Temporarily pinned `fastmcp` to version `2.14.5` to avoid incompatibilities introduced in `3.0.x`, with plans to upgrade in the future.

To fix bugs
Fix: #10035 

N/A

The CI can cover this change.

### DIFF
--- a/integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/ITUtils.java
+++ b/integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/ITUtils.java
@@ -51,6 +51,8 @@ import org.slf4j.LoggerFactory;
 
 public class ITUtils {
   private static final Logger LOG = LoggerFactory.getLogger(ITUtils.class);
+  private static final String CI_ENV = "CI";
+  private static final String GITHUB_ACTIONS_ENV = "GITHUB_ACTIONS";
 
   public static final String TEST_MODE = "testMode";
   public static final String EMBEDDED_TEST_MODE = "embedded";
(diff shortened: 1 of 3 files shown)


integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/ITUtils.java
```diff
@@ -220,8 +226,9 @@ public static void cleanDisk() {
             "free -m", false, ProcessData.TypesOfData.STREAMS_MERGED, Map.of());
     LOG.info("Before clean: Command free -m output:\n{}", output);
 
-    // Execute docker system prune -af to free up space before starting the OceanBase container
-    ProcessBuilder processBuilder = new ProcessBuilder("/bin/bash", "-c", "docker system prune -f");
+    // Execute docker container prune -f to clean only stopped containers in CI.
+    ProcessBuilder processBuilder =
```
FANNG1, line 230: Is it necessary for this PR?

yuqi1129: I added this minor fix by the way to remove this risk that the disk will be cleaned on our local machine.

roryqi: Sorry to miss this thread. If you have questions, you can raise a follow up PR to fix the related the issue.

roryqi: This is a cherry-pick PR. Commit should  keep consistent with main branch. If we need modification, we should modify the main branch first.

Issue #10032, opened 2026-02-25
[Subtask] Introduce docker-compose to DorisContrainer


zhoukangcn:
### Describe the subtask

Introduce docker-compose support in org.apache.gravitino.integration.test to address the limitations of the current single-BE Doris IT container, enabling proper multi-node cluster testing and compatibility with Doris 2.1.x features that require docker-compose orchestration.

### Parent issue

#6900

PR #10046, opened 2026-02-26
[#10032] feat(IT): Introduce docker-compose support for DorisContainer

zhoukangcn:

Introduce docker-compose based Doris container environment using separate FE and BE containers with official apache/doris:2.1.7 images, enabling proper multi-node cluster testing and Doris 2.1.x compatibility.

Fix: #10032

N/A

IT

### DIFF
--- a/integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
+++ b/integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
@@ -60,7 +60,8 @@ public class DorisContainer extends BaseContainer {
   public static final int FE_HTTP_PORT = 8030;
   public static final int FE_MYSQL_PORT = 9030;
   public static final long STARTUP_TIMEOUT_SECONDS =
-      Long.parseLong(System.getenv().getOrDefault("GRAVITINO_CI_DORIS_STARTUP_TIMEOUT_SECONDS", "120"));
+      Long.parseLong(
+          System.getenv().getOrDefault("GRAVITINO_CI_DORIS_STARTUP_TIMEOUT_SECONDS", "120"));
 
   private static final String FE_SERVICE = "doris-fe";
   private static final String BE_SERVICE = "doris-be";
@@ -108,7 +109,8 @@ public class DorisContainer extends BaseContainer {
             .withExposedService(
                 FE_SERVICE,
                 FE_MYSQL_PORT,
-                Wait.forListeningPort().withStartupTimeout(Duration.ofSeconds(STARTUP_TIMEOUT_SECONDS)))
+                Wait.forListeningPort()
+                    .withStartupTimeout(Duration.ofSeconds(STARTUP_TIMEOUT_SECONDS)))
             .withExposedService(FE_SERVICE, FE_HTTP_PORT)
             .withScaledService(BE_SERVICE, BE_SCALE)
             .withStartupTimeout(Duration.ofSeconds(STARTUP_TIMEOUT_SECONDS))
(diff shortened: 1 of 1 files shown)


zhoukangcn:
@yuqi1129 Could you help review this PR.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
```diff
@@ -63,78 +84,130 @@ protected DorisContainer(
       Map<String, String> envVars,
       Optional<Network> network) {
     super(image, hostName, ports, extraHosts, filesToMount, envVars, network);
-  }
```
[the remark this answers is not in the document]
zhoukangcn: I'd suggest not making this change for now. The current implementation works fine and removing the BaseContainer extension would be a bigger refactor that's out of scope for this PR. We can address it separately later if needed.

@yuqi1129 What do you think?

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
```diff
@@ -63,78 +84,136 @@ protected DorisContainer(
...
                     if (alive.equalsIgnoreCase("true") && totalCapacityFloat > 0.0f) {
-                      LOG.info("Doris container startup success!");
-                      return true;
+                      aliveCount++;
                     }
                   }
                 }
-                LOG.info("Doris container is not ready yet!");
+
+                if (aliveCount >= BE_SCALE) {
```
yuqi1129, line 212: Will it happen that `aliveCount` is larger than `BE_SCALE`?

Kangs-Claw: Good question! In theory, `aliveCount` should not exceed `BE_SCALE` since we only start 3 BE containers via `withScaledService(BE_SERVICE, BE_SCALE)`. However, using `>=` instead of `==` is more robust because:

1. It handles edge cases where extra backends might be present from previous test runs
2. It's more defensive programming - if somehow more backends appear, the check still passes
3. It aligns with the intent: "we need at least BE_SCALE backends alive"

So while it's unlikely to happen in normal circumstances, the `>=` check is safer and more correct.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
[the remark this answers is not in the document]
yuqi1129: Yeah, we can use another PR to resolve this problem. I will appreciate it if you open an issue to track it.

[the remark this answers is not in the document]
zhoukangcn: done

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
zhoukangcn: @yuqi1129 Do you have more suggestion?

yuqi1129: No, it's also fine with me.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisContainer.java
```diff
@@ -63,78 +84,136 @@ protected DorisContainer(
...
+    composeContainer =
+        new ComposeContainer(new File(composeFile))
+            .withEnv("GRAVITINO_CI_DORIS_DOCKER_IMAGE", effectiveImage)
+            .withExposedService(
+                FE_SERVICE,
+                FE_MYSQL_PORT,
+                Wait.forListeningPort().withStartupTimeout(Duration.ofMinutes(3)))
+            .withExposedService(FE_SERVICE, FE_HTTP_PORT)
+            .withScaledService(BE_SERVICE, BE_SCALE)
+            .withStartupTimeout(Duration.ofMinutes(5))
```
yuqi1129: Does it really take so long to start a Doris cluster?

PR #10057, opened 2026-02-26
[#9482] improvement(common): Enhance version parsing to support release candidates with validation

yuqi1129:

This pull request updates the version parsing logic in `Version.java` to support release candidate (RC) versions and adds tests to ensure correct handling of RC numbers. The main changes include expanding the regex pattern to recognize RC versions, enforcing bounds on RC numbers, and introducing unit tests for these scenarios.

To make it work when verifing release candidate version in the playground.

Fix: #9482 

N/A

UTs

### DIFF
--- a/common/src/main/java/org/apache/gravitino/Version.java
+++ b/common/src/main/java/org/apache/gravitino/Version.java
@@ -36,7 +36,7 @@ public class Version {
 
   private static final int VERSION_PART_NUMBER = 3;
   private static final Pattern PATTERN =
-      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(\\d{1,9})|-.*|\\.([a-zA-Z].*))?$");
+      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(0|[1-9]\\d*)|-.*|\\.([a-zA-Z].*))?$");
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/apache/gravitino/Version.java
```diff
@@ -36,7 +36,7 @@ public class Version {
 
   private static final int VERSION_PART_NUMBER = 3;
   private static final Pattern PATTERN =
-      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:-.*|\\.([a-zA-Z].*))?$");
+      Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(0|[1-9]\\d*)|-.*|\\.([a-zA-Z].*))?$");
```
roryqi, line 39: You would use natural number instead of magic number 9.

yuqi1129: (0|[1-9]\\d*). This is the regex that represents a natural number.  Do you have any suggestion about the regex format?

roryqi: My fault. It is ok now.

roryqi: My fault. It is ok now.

Issue #9727, opened 2026-01-15
[Bug report] Update model version including original empty Aliases[],  encouter UnsupportedOperationException "Model version entity does not have an ID"


danhuawang:
### Version

main branch

### Describe what's wrong

I update Aliases for a model version, if the Aliases are empty as following,  then update Aliases will be broken by UnsupportedOperationException.

I try to update Aliases for a model version which  Aliases aren't empty in creation, it works fine.

<img width="775" height="849" alt="Image" src="https://github.com/user-attachments/assets/8d3870ac-2354-4811-a3b0-e65afb55afaf" />

<img width="1246" height="928" alt="Image" src="https://github.com/user-attachments/assets/fc2f9ceb-5098-471e-8784-bd2645b908f7" />

### Error message and/or stacktrace

```
{
    "code": 1002,
    "type": "RuntimeException",
    "message": "Failed to operate object [m1 version(0)] operation [ALTER] under [s1], reason [Model version entity does not have an ID.]",
    "stack": [
        "java.lang.UnsupportedOperationException: Model version entity does not have an ID.",
        "\tat org.apache.gravitino.meta.ModelVersionEntity.id(ModelVersionEntity.java:134)",
        "\tat org.apache.gravitino.storage.relational.utils.POConverters.createAliasRelPO(POConverters.java:1780)",
        "\tat org.apache.gravitino.storage.relational.utils.POConverters.lambda$updateModelVersionAliasRelPO$9(POConverters.java:1715)",
        "\tat java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)",
        "\tat java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)",
        "\tat java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)",
        "\tat java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)",
        "\tat java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)",
        "\tat java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)",
        "\tat java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)",
        "\tat org.apache.gravitino.storage.relational.utils.POConverters.updateModelVersionAliasRelPO(POConverters.java:1716)",
[...]

PR #10071, opened 2026-02-27
[#9727] fix(model): fix UnsupportedOperationException when updating aliases for model version created without aliases

jerryshao:

When a model version is created with **no aliases** and aliases are later added via `alterModelVersion`, the server throws an `UnsupportedOperationException` with the message "Model version entity does not have an ID.".

The fix adds a `modelId` parameter to `POConverters.updateModelVersionAliasRelPO()` and uses it directly in the empty-old-aliases branch to construct alias relation POs, instead of calling `entity.id()` which always throws on `ModelVersionEntity`.

`POConverters.updateModelVersionAliasRelPO` had two code paths:
1. **Non-empty old aliases** — copy `modelId`/`version` from the existing `ModelVersionAliasRelPO` ✅
2. **Empty old aliases** — call `createAliasRelPO(ModelVersionEntity, String)` which called `entity.id()` → always throws `UnsupportedOperationException` since `ModelVersionEntity` intentionally has no DB entity ID ❌

Fix: #9727

No. This is a bug fix for an existing API. The `alterModelVersion` REST API now correctly handles the case where a model version was linked without aliases.

- **Unit test** (`TestPOConverters#testUpdateModelVersionAliasRelPO`): covers both the non-empty and empty old-aliases paths of `updateModelVersionAliasRelPO`.
- **Service-level test** (`TestModelVersionMetaService#testUpdateModelVersionAliasesFromEmpty`): verifies the full update flow in the JDBC storage layer — create a model version with no aliases, then alter it to add aliases.
- **Java integration test** (`ModelCatalogOperationsIT#testLinkAndUpdateModelVersionAliasesFromEmpty`): end-to-end test against a live Gravitino server.
- **Python integration test** (`test_model_catalog.py#test_link_update_model_version_aliases_from_empty`): end-to-end test via the Python client.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
@@ -1724,7 +1724,12 @@ public class POConverters {
     if (!oldModelVersionAliasRelPOs.isEmpty()) {
       ModelVersionAliasRelPO oldModelVersionAliasRelPO = oldModelVersionAliasRelPOs.get(0);
       return newModelVersion.aliases().stream()
-          .map(alias -> createAliasRelPO(oldModelVersionAliasRelPO, alias))
+          .map(
+              alias ->
+                  createAliasRelPO(
+                      oldModelVersionAliasRelPO.getModelId(),
+                      oldModelVersionAliasRelPO.getModelVersion(),
+                      alias))
           .collect(Collectors.toList());
     } else {
       return newModelVersion.aliases().stream()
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
yuqi1129: It seems that we use `createAliasRelPO(createAliasRelPO.id,createAliasRelPO.version, alias)` and remove method 
https://github.com/apache/gravitino/blob/0813c35706427a7569d2e6ddbb660fb98b332f87/core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java#L1782-L1790

yuqi1129 (commented):
A minor one.

Issue #10006, opened 2026-02-14
[Bug report] flaky test in TestPolicyMetaService


yuqi1129:
### Version

main branch

### Describe what's wrong

TestPolicyMetaService > testMetaLifeCycleFromCreationToDeletion() > [POSTGRESQL Backend] FAILED
    org.opentest4j.AssertionFailedError: expected: <3> but was: <1>
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
        at app//org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:527)
        at app//org.apache.gravitino.storage.relational.service.TestPolicyMetaService.testMetaLifeCycleFromCreationToDeletion(TestPolicyMetaService.java:211)

For more, please see: https://github.com/apache/gravitino/actions/runs/22016696203/job/63619449930?pr=9982

### Error message and/or stacktrace

Please see above.

### How to reproduce

Please see above.

yuqi1129:
@mchades 

Can you take some time to check it? I could not reproduce it locally with 
```shell
 while [[ $? -eq 0 ]]
do
./gradlew :core:test --tests "org.apache.gravitino.storage.relational.service.TestPolicyMetaService"
done
```

PR #10077, opened 2026-02-27
[#10006] fix(test): fix flaky test `testMetaLifeCycleFromCreationToDeletion` in CI. 

yuqi1129:

This pull request introduces an improvement to the `testMetaLifeCycleFromCreationToDeletion` test in `TestPolicyMetaService.java` by making the test more robust against timing issues when asserting the number of policy versions. The test now waits for the expected condition to be met, reducing flakiness due to asynchronous operations.

It's a flaky bug when VM specification in GitHub is not very large.  

Fix: #10006 

N/A

CI


yuqi1129:
It's no need.

PR #10081, opened 2026-02-27
[#10080] improvement(authz): `IcebergCatalog`, `PaimonCatalog` and `JdbcCatalog` support credential vending

### LINKED ISSUE #10080: [Improvement] IcebergCatalog, PaimonCatalog and JdbcCatalog support credential vending

IcebergCatalog and PaimonCatalog supports credential vending

### How should we improve?

IcebergCatalog and PaimonCatalog supports credential vending

roryqi:

`IcebergCatalog`, `PaimonCatalog` and `JdbcCatalog` support credential vending

Fix #10080

No.

Added IT.


integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
```diff
@@ -415,6 +423,8 @@ public void startIntegrationTest() throws Exception {
 
     List<String> authenticators = new ArrayList<>();
     String authenticatorStr = customConfigs.get(Configs.AUTHENTICATORS.getKey());
+    LOG.warn("Creating authenticator {}", authenticatorStr);
```
yuqi1129: warn level is not so proper.

yuqi1129:
@FANNG1 
Can you help take a look? In this PR, as long as user/passwd or aksk is set, Gravitino will automatically add the entry  `credential_providers` in the property. Is that expected?

roryqi:
> @FANNG1 Can you help take a look? In this PR, as long as user/passwd or aksk is set, Gravitino will automatically add the entry `credential_providers` in the property. Is that expected?

Fang may be busy. I don't change the properties and I add an extra method, so the properties won't change.

FANNG1:
> @FANNG1 Can you help take a look? In this PR, as long as user/passwd or aksk is set, Gravitino will automatically add the entry `credential_providers` in the property. Is that expected?

seems a little hacky

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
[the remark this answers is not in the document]
roryqi: Done.

catalogs/catalog-jdbc-mysql/src/test/java/org/apache/gravitino/catalog/mysql/integration/test/CatalogMysqlCredentialIT.java
```diff
@@ -0,0 +1,163 @@
...
+          LOG.error("Exception in closing client", e);
+        }
+        client = null;
+      }
+      try {
+        closer.close();
+      } catch (Exception e) {
+        LOG.error("Exception in closing CloseableGroup", e);
+      }
+    }
```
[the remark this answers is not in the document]
roryqi: Done.

roryqi:
@yuqi1129 Could u take a look at this pull request again?

api/src/main/java/org/apache/gravitino/credential/JdbcCredential.java
```diff
@@ -0,0 +1,110 @@
...
+    String jdbcPassword = credentialInfo.get(GRAVITINO_JDBC_PASSWORD);
+    validate(jdbcUser, jdbcPassword, expireTimeInMs);
+    this.jdbcUser = jdbcUser;
+    this.jdbcPassword = jdbcPassword;
+  }
+
+  /**
+   * Get JDBC user name.
+   *
+   * @return The JDBC user name.
```
yuqi1129, line 88: `credentialInfo.get(GRAVITINO_JDBC_USER)` may return `null` if the key is absent, and this null value flows into `validate()`. `Preconditions.checkArgument(StringUtils.isNotBlank(null))` will throw but with a potentially confusing error. Consider extracting and checking explicitly:

```java
this.jdbcUser = credentialInfo.get(GRAVITINO_JDBC_USER);
this.jdbcPassword = credentialInfo.get(GRAVITINO_JDBC_PASSWORD);
```

Then `validate()` will handle the null case cleanly via `StringUtils.isNotBlank`. But if you want a clearer error message at assignment time, use:
```java
this.jdbcUser = Preconditions.checkNotNull(
    credentialInfo.get(GRAVITINO_JDBC_USER), "JDBC user key is missing from credentialInfo");
```

api/src/main/java/org/apache/gravitino/credential/JdbcCredential.java
```diff
@@ -0,0 +1,110 @@
...
+    return jdbcPassword;
+  }
+
+  private void validate(String jdbcUser, String jdbcPassword, long expireTimeInMs) {
+    Preconditions.checkArgument(StringUtils.isNotBlank(jdbcUser), "JDBC user should not be empty");
+    Preconditions.checkArgument(
+        StringUtils.isNotBlank(jdbcPassword), "JDBC password should not be empty");
+    Preconditions.checkArgument(
+        expireTimeInMs == 0, "The expire time of JdbcCredential should be 0");
+  }
```
yuqi1129, line 115: The `expireTimeInMs == 0` check is correct for static credentials, but the error message `"Expire time of JDBC credential should be 0"` may confuse future maintainers or anyone examining credential factory failures. Consider adding an inline comment explaining the intent:

```java
// JDBC credentials are static (no server-issued expiry). expireTimeInMs must always be 0.
Preconditions.checkArgument(expireTimeInMs == 0, "Expire time of JDBC credential should be 0");
```

core/src/main/java/org/apache/gravitino/credential/JdbcCredentialProvider.java
```diff
@@ -0,0 +1,53 @@
...
+/** Generate JDBC user and password credentials to access JDBC backend services. */
+public class JdbcCredentialProvider implements CredentialProvider {
+
+  private String jdbcUser;
+  private String jdbcPassword;
+
+  @Override
+  public void initialize(Map<String, String> properties) {
+    this.jdbcUser = properties.get(JdbcCredential.GRAVITINO_JDBC_USER);
+    this.jdbcPassword = properties.get(JdbcCredential.GRAVITINO_JDBC_PASSWORD);
```
yuqi1129, line 43: `properties.get(...)` silently returns `null` when the key is absent. The null check in `getCredential()` (line 48) does handle this, which is fine. But consider adding a brief comment here to explain why `null` is an expected/valid outcome:

```java
// jdbcUser / jdbcPassword will be null when the catalog was not configured with JDBC credentials.
// getCredential() handles this by returning null (no credential available).
this.jdbcUser = properties.get(JdbcCredential.GRAVITINO_JDBC_USER);
this.jdbcPassword = properties.get(JdbcCredential.GRAVITINO_JDBC_PASSWORD);
```

Also, `properties` itself could be null if called incorrectly. A null guard would make the contract explicit:
```java
if (properties == null) return;
```

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/JdbcCatalog.java
```diff
@@ -113,4 +118,26 @@ public PropertiesMetadata schemaPropertiesMetadata() throws UnsupportedOperation
...
+    if (StringUtils.isNotBlank(properties.get(CredentialConstants.CREDENTIAL_PROVIDERS))) {
+      return properties;
+    }
+
+    // Add JDBC credential provider if jdbc-user and jdbc-password are set
+    String jdbcUser = properties.get(JdbcConfig.USERNAME.getKey());
+    String jdbcPassword = properties.get(JdbcConfig.PASSWORD.getKey());
+    if (StringUtils.isNotBlank(jdbcUser) && StringUtils.isNotBlank(jdbcPassword)) {
+      properties.put(CredentialConstants.CREDENTIAL_PROVIDERS, JdbcCredential.JDBC_CREDENTIAL_TYPE);
+    }
```
yuqi1129, line 139: **Architectural inconsistency with `IcebergCatalog` and `PaimonCatalog`**: both of those catalogs build a _list_ of credential providers (JDBC + S3/OSS/Azure) and join them with commas. `JdbcCatalog` only ever sets a single `JDBC_CREDENTIAL_TYPE`.

This is fine for a pure JDBC catalog, but the inconsistency means if someone extends `JdbcCatalog` and the underlying storage also needs S3 credentials, there is no hook to add them. Consider aligning the implementation structure with `IcebergCatalog.buildCredentialProvidersIfNecessary()` — even if the result is the same today, it is easier to extend later.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -81,4 +94,67 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
...
   }
+
+  @Override
+  public Map<String, String> propertiesWithCredentialProviders() {
+    Map<String, String> properties = Maps.newHashMap(super.propertiesWithCredentialProviders());
+    return buildCredentialProvidersIfNecessary(properties);
+  }
+
+  private Map<String, String> buildCredentialProvidersIfNecessary(Map<String, String> properties) {
+    // If credential providers already set, return as is
```
yuqi1129, line 100: `PaimonCatalog.buildCredentialProvidersIfNecessary()` is an almost exact copy of this method. This is ~50 lines of duplicated logic. 

Consider extracting to a protected/static utility in `BaseCatalog` or a shared helper class, e.g.:
```java
// In BaseCatalog:
protected static Map<String, String> buildJdbcCredentialProvidersIfNecessary(
    Map<String, String> properties, String metastoreUri, String backend) { ... }
```

If keeping them separate is intentional (e.g., different backend key names), a comment explaining the difference would help.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
```diff
@@ -421,6 +422,16 @@ public Map<String, String> properties() {
...
   }
 
+  /**
+   * Retrieves the properties of the catalog including credential providers. This method is used
+   * when we need to expose credential-related properties.
+   *
+   * @return A map of properties including credential providers.
+   */
+  public Map<String, String> propertiesWithCredentialProviders() {
+    return properties();
```
yuqi1129, line 466: The default implementation returning `properties()` is a safe fallback and the Javadoc is clear. One suggestion: the Javadoc says "This method is called when we need to expose credential-related properties" — it would be more precise to say:

> Subclasses should override this method to inject auto-detected credential provider names into the properties map before the `CatalogCredentialManager` is initialized. The default implementation returns `properties()` unchanged.

This helps future catalog implementors know exactly when/why to override this hook.

core/src/test/java/org/apache/gravitino/credential/TestJdbcCredentialProvider.java
```diff
@@ -0,0 +1,102 @@
...
+    String jdbcPassword = "test-password";
+    Map<String, String> catalogProperties =
+        ImmutableMap.of(
+            JdbcCredential.GRAVITINO_JDBC_USER,
+            jdbcUser,
+            JdbcCredential.GRAVITINO_JDBC_PASSWORD,
+            jdbcPassword);
+
+    CredentialProvider credentialProvider =
+        CredentialProviderFactory.create(JdbcCredential.JDBC_CREDENTIAL_TYPE, catalogProperties);
```
yuqi1129, line 41: Missing test cases for edge conditions:

1. **Partial credentials** — only user set, password absent (expect `getCredential()` returns `null`)
2. **Empty string credentials** — `""` vs `null` behavior (both should result in `null` return since `JdbcCredentialProvider` does `== null` check, not `isBlank`)
3. **`properties` map is `null`** — verify NPE is thrown or handled gracefully

Example:
```java
@Test
void testPartialCredentials() {
    Map<String, String> props = new HashMap<>();
    props.put(JdbcCredential.GRAVITINO_JDBC_USER, "user");
    // password deliberately missing
    provider.initialize(props);
    Assertions.assertNull(provider.getCredential(null));
}
```

Note: the `== null` check in `getCredential()` will return `null` for absent keys but NOT for empty-string values. If empty passwords should also be rejected, the check should use `isBlank` there too.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -81,4 +94,67 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
   public PropertiesMetadata schemaPropertiesMetadata() throws UnsupportedOperationException {
     return SCHEMA_PROPERTIES_META;
   }
+
+  @Override
+  public Map<String, String> propertiesWithCredentialProviders() {
```
yuqi1129, line 99: `equalsIgnoreCase` already handles case-insensitive comparison — `.toUpperCase(Locale.ROOT)` is redundant. Just use:
```java
if (!IcebergCatalogBackend.JDBC.name().equalsIgnoreCase(catalogBackend)) {
```

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalog.java
```diff
@@ -77,4 +90,67 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
...
+  public Map<String, String> propertiesWithCredentialProviders() {
+    Map<String, String> properties = Maps.newHashMap(super.propertiesWithCredentialProviders());
+    return buildCredentialProvidersIfNecessary(properties);
+  }
+
+  private Map<String, String> buildCredentialProvidersIfNecessary(Map<String, String> properties) {
+    // If credential providers already set, return as is
+    if (StringUtils.isNotBlank(properties.get(CredentialConstants.CREDENTIAL_PROVIDERS))) {
+      return properties;
+    }
```
yuqi1129, line 104: Same here — `.toUpperCase(Locale.ROOT)` is redundant with `equalsIgnoreCase`. Simplify to:
```java
if (!PaimonCatalogBackend.JDBC.name().equalsIgnoreCase(catalogBackend)) {
```

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -81,4 +94,67 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
   public PropertiesMetadata schemaPropertiesMetadata() throws UnsupportedOperationException {
```
yuqi1129, line 94: The `buildCredentialProvidersIfNecessary` method in `IcebergCatalog` and `PaimonCatalog` are nearly identical (the only difference is the backend constant and property key sources). Consider extracting the shared logic into a utility method in `BaseCatalog` or a helper class to reduce duplication. For example:

```java
// In a shared utility or BaseCatalog
protected static Map<String, String> buildStorageCredentialProviders(
    Map<String, String> properties, List<String> existingProviders) {
    // S3, OSS, Azure checks...
}
```

The JDBC-backend check is catalog-specific, but the S3/OSS/Azure credential detection is exactly the same code.

core/src/main/java/org/apache/gravitino/credential/JdbcCredentialProvider.java
```diff
@@ -0,0 +1,53 @@
...
+
+/** Generate JDBC user and password credentials to access JDBC backend services. */
+public class JdbcCredentialProvider implements CredentialProvider {
+
+  private String jdbcUser;
+  private String jdbcPassword;
+
+  @Override
+  public void initialize(Map<String, String> properties) {
+    this.jdbcUser = properties.get(JdbcCredential.GRAVITINO_JDBC_USER);
```
yuqi1129, line 33: The `JdbcCredentialProvider.initialize()` reads properties with keys `"jdbc-user"` and `"jdbc-password"` (from `JdbcCredential.GRAVITINO_JDBC_USER/PASSWORD`). This works because `JdbcConfig.USERNAME`/`PASSWORD` also use the same key strings. This implicit coupling is fragile — if either side changes the key string, credential vending will silently break. Consider adding a comment or using a shared constant to make this coupling explicit.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -81,4 +94,67 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
...
     return SCHEMA_PROPERTIES_META;
   }
+
+  @Override
```
yuqi1129, line 98: This only adds credential providers for the JDBC backend. But Iceberg catalogs with non-JDBC backends (e.g., Hive metastore or REST) can also use S3/OSS/Azure for table storage. Should the S3/OSS/Azure credential provider detection be applied regardless of the catalog backend type? The current logic skips storage credential vending entirely for non-JDBC backends, which seems like an unnecessary limitation.

api/src/main/java/org/apache/gravitino/credential/JdbcCredential.java
```diff
@@ -0,0 +1,110 @@
...
+  public String credentialType() {
+    return JDBC_CREDENTIAL_TYPE;
+  }
+
+  @Override
+  public long expireTimeInMs() {
+    return 0;
+  }
+
+  @Override
```
yuqi1129, line 68: `credentialInfo()` returns the password in plain text. This map may end up serialized in REST responses or logs. Consider whether there is a need for redaction/masking in the credential info map, similar to how other sensitive properties are handled in the codebase. At minimum, the `toString()` method (inherited from Object) should be overridden to avoid accidental password leakage in logs.

catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/JdbcCatalog.java
diqiu50: The JDBC catalog does not need more

catalogs/catalog-jdbc-mysql/src/test/java/org/apache/gravitino/catalog/mysql/integration/test/CatalogMysqlCredentialIT.java
```diff
@@ -0,0 +1,163 @@
...
+    Map<String, String> catalogProperties = Maps.newHashMap();
+    catalogProperties.put(
+        JdbcConfig.JDBC_URL.getKey(),
+        String.format(
+            "jdbc:mysql://%s:%d/",
+            mysqlContainer.getContainerIpAddress(), MySQLContainer.MYSQL_PORT));
+    catalogProperties.put(
+        JdbcConfig.JDBC_DRIVER.getKey(), mysqlContainer.getDriverClassName(TEST_DB_NAME));
+    catalogProperties.put(JdbcConfig.USERNAME.getKey(), mysqlContainer.getUsername());
+    catalogProperties.put(JdbcConfig.PASSWORD.getKey(), mysqlContainer.getPassword());
```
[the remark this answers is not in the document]
diqiu50: Not a problem

roryqi:
I don't have other comments. Let @yuqi1129 do as the final call.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
```diff
@@ -445,6 +454,49 @@ public Map<String, String> properties() {
...
+    String ossSecretAccessKey = properties.get(OSSProperties.GRAVITINO_OSS_ACCESS_KEY_SECRET);
+    if (StringUtils.isNotBlank(ossAccessKeyId) && StringUtils.isNotBlank(ossSecretAccessKey)) {
+      credentialProviders.add(OSSSecretKeyCredential.OSS_SECRET_KEY_CREDENTIAL_TYPE);
+    }
+
+    String azureAccountName = properties.get(AzureProperties.GRAVITINO_AZURE_STORAGE_ACCOUNT_NAME);
+    String azureAccountKey = properties.get(AzureProperties.GRAVITINO_AZURE_STORAGE_ACCOUNT_KEY);
+    if (StringUtils.isNotBlank(azureAccountName) && StringUtils.isNotBlank(azureAccountKey)) {
+      credentialProviders.add(AzureAccountKeyCredential.AZURE_ACCOUNT_KEY_CREDENTIAL_TYPE);
+    }
```
yuqi1129, line 497: How do we handle GCS here?

yuqi1129 (commented):
A minor one, others LGTM

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
diqiu50: GCS uses OAuth 2.0 tokens, not static keys like S3/OSS/Azure.

FANNG1:
The lastest PR changed a lot compareing to the initial code,  could you provide more context to help other developer to understand the background?  after this PR is merged, is there any following PR ?

diqiu50:
> The lastest PR changed a lot compareing to the initial code, could you provide more context to help other developer to understand the background? after this PR is merged, is there any following PR ?

There are several follow-up PRs after this merge. The overall issue is tracked in https://github.com/apache/gravitino/issues/10415.
This PR implements a client API to retrieve static credentials from the Gravitino server. Based on this PR, the Trino, Spark, and Flink connectors will use this API to retrieve credentials instead of reading them from properties.
Other features, such as dynamic temporary tokens, will not be implemented for now.

FANNG1 (commented):
Thanks for the work @roryqi. A few review comments on design and code (1-7). The biggest concern is #1 — please take a look.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -81,4 +87,38 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
...
+  public Map<String, String> propertiesWithCredentialProviders() {
```
FANNG1, line 94: **[Design / breaking behavior] Implicit credential vending is opt-out.**

The current implementation auto-injects `jdbc` (and S3 / OSS / Azure static-key) providers into `credential-providers` whenever the corresponding properties are present and the user has not explicitly set `credential-providers`. After upgrade, every existing Iceberg-JDBC / Paimon-JDBC / pure JDBC catalog will start vending its raw DB password to any caller holding credential-fetch privilege on the catalog.

Unlike S3/OSS static keys (which are intentionally stored to be vended), the JDBC password historically was a server-side connection secret. Flipping it to user-retrievable by default is a semantic shift.

Suggestion: make this **opt-in** — require the user to put `jdbc` explicitly in `credential-providers`, or add a switch such as `gravitino.credential.jdbc.vending.enabled`. At minimum, please document the behavior change in release notes and in the javadoc of `propertiesWithCredentialProviders()` / `buildCredentialProvidersIfNecessary` in all three catalogs (Iceberg / Paimon / JDBC).

(Worth noting: `jdbc-password` is currently `hidden=false` in the property metadata, so it's already returned via `loadCatalog`. The two exposure paths now coexist — addressing both belongs in this PR or an immediate follow-up.)

api/src/main/java/org/apache/gravitino/credential/JdbcCredential.java
```diff
@@ -0,0 +1,116 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+import org.apache.commons.lang3.StringUtils;
+
+/** JDBC credential for accessing JDBC backend services. */
+public class JdbcCredential implements Credential {
+
+  /** JDBC credential type. */
+  public static final String JDBC_CREDENTIAL_TYPE = "jdbc";
```
FANNG1: **[Naming] `JDBC_CREDENTIAL_TYPE = "jdbc"` is too generic.**

The identifier `"jdbc"` will collide with any future JDBC credential variant (dynamic credentials, IAM-based auth, certificate-based, etc.). Consider something more specific like `jdbc-user-password` or `gravitino-jdbc` so the type name leaves room for additional JDBC credential kinds.

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
```diff
@@ -445,6 +454,49 @@ public Map<String, String> properties() {
...
 
+  /**
+   * Retrieves the properties of the catalog including credential providers. Subclasses should
+   * override this method to inject auto-detected credential provider names into the properties map
+   * before the {@link CatalogCredentialManager} is initialized. The default implementation returns
+   * {@link #properties()} unchanged.
+   *
+   * @return A map of properties including credential providers.
+   */
+  public Map<String, String> propertiesWithCredentialProviders() {
```
FANNG1, line 465: **[API stability] Missing stability annotation.**

The helper `addStorageCredentialProviders` below is annotated `@Evolving` (good), but this new `public` extension method has no stability marker. Since it's a connector extension point that external catalog implementations will override, please mark it `@Evolving` (or `@Unstable`) to make the contract explicit.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalog.java
```diff
@@ -77,4 +83,39 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
...
+
+    List<String> credentialProviders = new ArrayList<>();
+
+    // Add JDBC credential provider if backend is JDBC and jdbc-user/jdbc-password are set
+    String catalogBackend = properties.get(PaimonConstants.CATALOG_BACKEND);
+    if (catalogBackend != null
```
FANNG1, line 105: **[Style] Redundant null check; inconsistent with `IcebergCatalog`.**

`if (catalogBackend != null && PaimonCatalogBackend.JDBC.name().equalsIgnoreCase(catalogBackend))` — the explicit null check is unnecessary; `String#equalsIgnoreCase(null)` returns `false`. `IcebergCatalog.buildCredentialProvidersIfNecessary` uses the shorter form. Please align.

core/src/main/java/org/apache/gravitino/credential/JdbcCredentialProvider.java
```diff
@@ -0,0 +1,62 @@
...
+  public void initialize(Map<String, String> properties) {
+    if (properties == null) {
+      return;
+    }
+    // jdbcUser / jdbcPassword will be null when the catalog was not configured with JDBC
+    // credentials.
+    // getCredential() handles this by returning null (no credential available).
+    // Note: The property keys "jdbc-user" and "jdbc-password" match the configuration keys
+    // used by JDBC-based catalogs (e.g., Iceberg JDBC backend, Paimon JDBC backend).
+    this.jdbcUser = properties.get(JdbcCredential.GRAVITINO_JDBC_USER);
```
FANNG1, line 42: **[Comments] Trim the explanatory comment block.**

The multi-line comment above explaining what `jdbc-user` / `jdbc-password` mean repeats information already conveyed by the constant names. Suggest collapsing it to a single line noting why nulls are tolerated here (i.e. "unset when the catalog was not configured with JDBC credentials; `getCredential()` returns null in that case").

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -81,4 +87,38 @@ public PropertiesMetadata catalogPropertiesMetadata() throws UnsupportedOperatio
...
+    Map<String, String> properties = Maps.newHashMap(super.propertiesWithCredentialProviders());
+    return buildCredentialProvidersIfNecessary(properties);
+  }
+
+  private Map<String, String> buildCredentialProvidersIfNecessary(Map<String, String> properties) {
```
FANNG1: **[Naming] `buildCredentialProvidersIfNecessary` is ambiguous.**

The method mutates the passed-in map and short-circuits when `credential-providers` is already set. "IfNecessary" doesn't quite capture either behavior. Consider `buildCredentialProvidersIfAbsent` or `applyDefaultCredentialProviders`. Same comment applies to the equivalent helpers in `PaimonCatalog` and `JdbcCatalog`.

catalogs/catalog-jdbc-common/src/test/java/org/apache/gravitino/catalog/jdbc/TestJdbcCatalogCredential.java
```diff
@@ -0,0 +1,165 @@
...
+    Map<String, String> properties = jdbcCatalog.propertiesWithCredentialProviders();
+
+    // Should have jdbc credential provider
+    String credentialProviders = properties.get(CredentialConstants.CREDENTIAL_PROVIDERS);
+    Assertions.assertNotNull(credentialProviders);
+    Assertions.assertEquals(JdbcCredential.JDBC_CREDENTIAL_TYPE, credentialProviders);
+  }
+
+  @Test
+  void testJdbcCatalogNoCredentialProvidersWithoutPassword() {
```
FANNG1, line 103: **[Tests] Add coverage for both `jdbc-user` and `jdbc-password` missing.**

Current cases: (a) user + password present, (b) password missing, (c) explicit `credential-providers` not overridden. Missing: both `jdbc-user` and `jdbc-password` absent. Please add that negative case so the no-op default path is locked in by a test.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
diqiu50: This is not a newly introduced issue. Previously, these were exposed via properties, but they are now only accessible through getCredential. We will address the issues of `hidden=false` in the spark/trino/flink connector pr, changed it will break CI now.

api/src/main/java/org/apache/gravitino/credential/JdbcCredential.java
diqiu50: Rename to "jdbc-user-password"

core/src/main/java/org/apache/gravitino/connector/BaseCatalog.java
diqiu50: fixed

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
diqiu50: fixed

FANNG1:
One question to confirm the expected behavior: if an Iceberg catalog already sets credential-
  providers=s3-secret-key, the current logic will not auto-detect jdbc-user-password anymore. Is that
  intentional? In other words, once credential-providers is explicitly set, inferred providers are not
  merged.

  Other than this, LGTM.

diqiu50:
> One question to confirm the expected behavior: if an Iceberg catalog already sets credential-

Yes, if user set the credential-providers, we just use them.

FANNG1:
Good catch. If an Iceberg catalog is configured with multiple credential providers, users can fetch multiple credentials through the Gravitino credential API, but the Iceberg REST `loadTable` credential-vending path currently only supports exactly one provider and may fail when multiple providers are configured. This inconsistency should be tracked and addressed in a follow-up. Other than this, LGTM.

FANNG1 (approved):

Issue #10091, opened 2026-02-28
[FEATURE]  Add the version segment module to support Trino 478


hqbhoho:
### Describe the feature

Add the version segment module to support Trino 478

PR #10092, opened 2026-02-28
[#10091] feat (trino-connector): Add the version segment module to support Trino 473-478

hqbhoho:

Add the version segment module to support Trino 473-478

Fix: #10091

No

UT and IT

### DIFF
--- a/docs/trino-connector/development.md
+++ b/docs/trino-connector/development.md
@@ -25,7 +25,7 @@ trino-connector/
 │   └── src/main/java/
 ├── trino-connector-469-472/      # Version-specific adapters for Trino 469-472
 │   └── src/main/java/
-├── trino-connector-478/      # Version-specific adapters for Trino 478
+├── trino-connector-473-478/      # Version-specific adapters for Trino 473-478
 │   └── src/main/java/
 └── integration-test/             # Integration tests
 ```
@@ -74,7 +74,7 @@ Change `localhost`, `port`, and the names of metalake and catalogs to match your
    | 446-451       | `trino-connector-446-451` |
    | 452-468       | `trino-connector-452-468` |
    | 469-472       | `trino-connector-469-472` |
-   | 478           | `trino-connector-478`     |
+   | 473-478       | `trino-connector-473-478` |
 
 5. Add `<module>plugin/trino-gravitino</module>` to `trino/pom.xml` and create the `pom.xml` for the `trino-gravitino` module. The example below uses Trino `469`. Ensure the `trino-root` version matches the Trino version you are developing against.
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -865,7 +865,7 @@ tasks {
     dependsOn("trino-connector:trino-connector-446-451:copyLibs")
     dependsOn("trino-connector:trino-connector-452-468:copyLibs")
     dependsOn("trino-connector:trino-connector-469-472:copyLibs")
-    dependsOn("trino-connector:trino-connector-478:copyLibs")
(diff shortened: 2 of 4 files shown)


docs/trino-connector/development.md
```diff
@@ -65,13 +67,14 @@ Change `localhost`, `port`, and the names of metalake and catalogs to match your
...
-   | Trino Version | Version-Segment Module |
-   |---------------|------------------------|
+   | Trino Version | Version-Segment Module    |
+   |---------------|---------------------------|
    | 435-439       | `trino-connector-435-439` |
    | 440-445       | `trino-connector-440-445` |
    | 446-451       | `trino-connector-446-451` |
    | 452-468       | `trino-connector-452-468` |
    | 469-472       | `trino-connector-469-472` |
+   | 478           | `trino-connector-478`     |
```
yuqi1129: The naming rule is different from previous ones. Is it expected?

integration-test-common/docker-script/init/trino/config/jvm.config
```diff
@@ -30,7 +30,6 @@
 -Djdk.nio.maxCachedBufferSize=2000000
 # Reduce starvation of threads by GClocker, recommend to set about the number of cpu cores (JDK-8192647)
 -XX:+UnlockDiagnosticVMOptions
--XX:GCLockerRetryAllocationCount=32
```
yuqi1129, line 33: Why do we move this one?

trino-connector/trino-connector-473-478/build.gradle.kts
```diff
@@ -0,0 +1,168 @@
...
+    exclude("org.apache.logging.log4j")
+  }
+  testImplementation("io.trino:trino-testing:$trinoVersion") {
+    exclude("org.apache.logging.log4j")
+    exclude("org.junit.jupiter")
+    exclude("org.junit.platform")
+  }
+  testImplementation(enforcedPlatform("org.junit:junit-bom:5.11.4"))
+  testImplementation("org.junit.jupiter:junit-jupiter-api")
+  testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
```
yuqi1129, line 78: Those Junit dependencies should be defined in `${GRAVITINO_HOME}/gradle`, can you use the format like `libs.xxxx`?

docs/trino-connector/development.md
hqbhoho: My thought is that the latest supported version shouldn't be named as a range. For example, when we add support for 485 later, we can rename the current `trino-connector-478` to `trino-connector-478-484`, and then create a new `trino-connector-485`

integration-test-common/docker-script/init/trino/config/jvm.config
hqbhoho: JDK24 not support `GCLockerRetryAllocationCount`

trino-connector/trino-connector-473-478/build.gradle.kts
hqbhoho: There is a dependency conflict here between JUnit (6.0.0) in the new `trino-testing` and JUnit (5.8.1) in `Gravitino`. To make the unit tests pass, I specified a specific version here. Do we need to add a `junit-5-11` version entry in `libs.versions.toml`.

apatrida:
479 is current
https://trino.io/docs/current/release/release-479.html

diqiu50:
> 479 is current https://trino.io/docs/current/release/release-479.html

This release supports up to Trino 478. Support for later versions will be added in the next release.

Issue #10096, opened 2026-03-01
[Subtask] add build-in table stats updater


FANNG1:
### Describe the subtask

Add a built-in table stats updater

### Parent issue

#9545

PR #10106, opened 2026-03-02
[#10096] feat(iceberg): support build in Iceberg stats updater job

FANNG1:

This PR enhances the built-in Iceberg update job and optimizer integration in four areas:

1. Add built-in Iceberg stats/metrics update capability
- Introduce/rename the built-in job implementation to `IcebergUpdateStatsAndMetricsJob`.
- Support `update_mode` with `stats | metrics | all`.
- Keep one execution path that can update Gravitino statistics and/or persist metrics (including JDBC-backed metrics storage).

2. Reuse shared Spark/Iceberg config logic
- Add `IcebergSparkConfigUtils` in `maintenance/optimizer-api`.
- Share Spark template config generation, flat JSON parsing, and catalog config validation logic.
- Reuse the same utility in both update-stats and rewrite-data-files jobs to avoid diverging behavior.

3. Simplify submit-update-stats-job UX
- Simplify optimizer command context and submit command input handling.
- Remove `--target-file-size-bytes` from optimizer CLI and built-in job template inputs.
- Use fixed defaults in job logic:
  - `datafile_mse` target file size: **128 MiB**
  - small file threshold: **32 MiB**

4. Strengthen tests
- Add unit tests for `IcebergSparkConfigUtils`.
- Update/extend job and optimizer tests to cover new signatures and validation paths.
- Keep Spark-based update-stats tests aligned with the new fixed-target behavior.

- The update-stats flow now also needs metrics updates, and users need one consistent built-in job for `stats`, `metrics`, and `all` modes.
- Spark config building/parsing/validation was duplicated across job paths; centralizing it reduces maintenance risk and inconsistent behavior.
- User feedback indicated the submit command had too many parameters; removing `target-file-size-bytes` and fixing safe defaults reduces configuration overhead and mistakes.

Fix: #10096

Yes.

1. Optimizer CLI behavior change
- `submit-update-stats-job` no longer accepts `--target-file-size-bytes`.
- Users should provide `update_mode`, `updater_options`, and `spark_conf` (via CLI or optimizer config).

2. Built-in job input change
- `target_file_size_bytes` is removed from built-in job template/job config for update-stats flow.
- `datafile_mse` now always uses fixed 128 MiB target size.

3. Functionality enhancement
- Built-in update job supports metrics-only mode and combined stats+metrics mode.

Locally validated with formatting, compilation, and targeted tests:

```bash
./gradlew :maintenance:optimizer-api:spotlessApply :maintenance:jobs:spotlessApply :maintenance:optimizer:spotlessApply
./gradlew :maintenance:optimizer-api:test --tests org.apache.gravitino.maintenance.optimizer.common.util.TestIcebergSparkConfigUtils
./gradlew :maintenance:jobs:test --tests org.apache.gravitino.maintenance.jobs.iceberg.TestIcebergRewriteDataFilesJob --tests org.apache.gravitino.maintenance.jobs.iceberg.TestIcebergUpdateStatsJob
./gradlew :maintenance:optimizer:test --tests org.apache.gravitino.maintenance.optimizer.TestOptimizerCmd
```

And after removing `--target-file-size-bytes`, reran:

```bash
./gradlew :maintenance:jobs:test --tests org.apache.gravitino.maintenance.jobs.iceberg.TestIcebergUpdateStatsJob
./gradlew :maintenance:optimizer:test --tests org.apache.gravitino.maintenance.optimizer.TestOptimizerCmd
```

### DIFF
--- a/maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/command/SubmitUpdateStatsJobCommand.java
+++ b/maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/command/SubmitUpdateStatsJobCommand.java
@@ -44,6 +44,18 @@ public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
   private static final String DEFAULT_UPDATE_MODE = "all";
   private static final String OPTION_UPDATER_OPTIONS = "updater-options";
   private static final String OPTION_SPARK_CONF = "spark-conf";
+  private static final String DEFAULT_SPARK_MASTER = "local[*]";
+  private static final String DEFAULT_SPARK_EXECUTOR_INSTANCES = "1";
+  private static final String DEFAULT_SPARK_EXECUTOR_CORES = "1";
+  private static final String DEFAULT_SPARK_EXECUTOR_MEMORY = "1g";
+  private static final String DEFAULT_SPARK_DRIVER_MEMORY = "1g";
+  private static final String SPARK_MASTER_KEY = "spark.master";
+  private static final String SPARK_EXECUTOR_INSTANCES_KEY = "spark.executor.instances";
+  private static final String SPARK_EXECUTOR_CORES_KEY = "spark.executor.cores";
+  private static final String SPARK_EXECUTOR_MEMORY_KEY = "spark.executor.memory";
+  private static final String SPARK_DRIVER_MEMORY_KEY = "spark.driver.memory";
+  private static final String SPARK_SQL_CATALOG_PREFIX = "spark.sql.catalog.";
+  private static final String ICEBERG_SPARK_CATALOG_IMPL = "org.apache.iceberg.spark.SparkCatalog";
 
   private static final ObjectMapper MAPPER = new ObjectMapper();
 
@@ -68,7 +80,9 @@ public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
     Map<String, String> updaterOptions =
         IcebergSparkConfigUtils.parseFlatJsonMap(updaterOptionsJson, OPTION_UPDATER_OPTIONS);
     Map<String, String> sparkConfigs =
-        IcebergSparkConfigUtils.parseFlatJsonMap(sparkConfJson, OPTION_SPARK_CONF);
+        normalizeSparkConfigs(
+            tableTargets,
+            IcebergSparkConfigUtils.parseFlatJsonMap(sparkConfJson, OPTION_SPARK_CONF));
 
     validateUpdaterOptions(updateMode, updaterOptions);
     validateSparkConfigs(tableTargets, sparkConfigs);
@@ -76,7 +90,7 @@ public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
     if (context.dryRun()) {
       for (TableTarget tableTarget : tableTargets) {
         Map<String, String> jobConfig =
-            buildJobConfig(tableTarget, updateMode, updaterOptions, sparkConfigs);
+            buildJobConfig(tableTarget, updateMode, updaterOptions, sparkConfigs, submitterConfigs);
         context
             .output()
             .printf(
@@ -93,7 +107,7 @@ public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
       int submitted = 0;
       for (TableTarget tableTarget : tableTargets) {
         Map<String, String> jobConfig =
-            buildJobConfig(tableTarget, updateMode, updaterOptions, sparkConfigs);
+            buildJobConfig(tableTarget, updateMode, updaterOptions, sparkConfigs, submitterConfigs);
         JobHandle jobHandle = client.runJob(JOB_TEMPLATE_NAME, jobConfig);
         submitted++;
         context
@@ -114,9 +128,12 @@ public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
       TableTarget tableTarget,
       String updateMode,
       Map<String, String> updaterOptions,
-      Map<String, String> sparkConfigs) {
+      Map<String, String> sparkConfigs,
+      Map<String, String> submitterConfigs) {
     Map<String, String> jobConfig = new LinkedHashMap<>();
     jobConfig.put("catalog_name", tableTarget.catalogName);
+    jobConfig.putAll(
+        resolveTemplateSparkPlaceholders(tableTarget.catalogName, sparkConfigs, submitterConfigs));
     jobConfig.put("table_identifier", tableTarget.schemaAndTable);
     jobConfig.put("update_mode", updateMode);
     jobConfig.put("updater_options", toCanonicalJson(updaterOptions));
@@ -226,12 +243,89 @@ public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
     }
   }
 
+  /**
+   * Fill Spark/Iceberg defaults that are already defined in built-in template configs so users do
+   * not need to repeat them in --spark-conf.
+   */
+  private static Map<String, String> normalizeSparkConfigs(
+      List<TableTarget> tableTargets, Map<String, String> sparkConfigs) {
+    Map<String, String> normalized = new LinkedHashMap<>(sparkConfigs);
+
+    for (TableTarget tableTarget : tableTargets) {
+      normalized.putIfAbsent(
+          SPARK_SQL_CATALOG_PREFIX + tableTarget.catalogName, ICEBERG_SPARK_CATALOG_IMPL);
+    }
+
+    return normalized;
+  }
+
   private static String toCanonicalJson(Map<String, String> options) {
     try {
       return MAPPER.writeValueAsString(new TreeMap<>(options == null ? Map.of() : options));
     } catch (Exception e) {
       throw new IllegalStateException("Failed to serialize options as JSON", e);
     }
+  }
+
+  private static Map<String, String> resolveTemplateSparkPlaceholders(
+      String catalogName, Map<String, String> sparkConfigs, Map<String, String> submitterConfigs) {
+    Map<String, String> placeholders = new LinkedHashMap<>();
+    String catalogPrefix = SPARK_SQL_CATALOG_PREFIX + catalogName;
+
+    placeholders.put(
+        "spark_master",
+        firstNonBlank(
+            submitterConfigs.get("spark_master"),
+            sparkConfigs.get(SPARK_MASTER_KEY),
+            DEFAULT_SPARK_MASTER));
+    placeholders.put(
+        "spark_executor_instances",
+        firstNonBlank(
+            submitterConfigs.get("spark_executor_instances"),
+            sparkConfigs.get(SPARK_EXECUTOR_INSTANCES_KEY),
+            DEFAULT_SPARK_EXECUTOR_INSTANCES));
+    placeholders.put(
+        "spark_executor_cores",
+        firstNonBlank(
+            submitterConfigs.get("spark_executor_cores"),
+            sparkConfigs.get(SPARK_EXECUTOR_CORES_KEY),
+            DEFAULT_SPARK_EXECUTOR_CORES));
+    placeholders.put(
+        "spark_executor_memory",
+        firstNonBlank(
+            submitterConfigs.get("spark_executor_memory"),
+            sparkConfigs.get(SPARK_EXECUTOR_MEMORY_KEY),
+            DEFAULT_SPARK_EXECUTOR_MEMORY));
+    placeholders.put(
+        "spark_driver_memory",
+        firstNonBlank(
+            submitterConfigs.get("spark_driver_memory"),
+            sparkConfigs.get(SPARK_DRIVER_MEMORY_KEY),
+            DEFAULT_SPARK_DRIVER_MEMORY));
+    placeholders.put(
+        "catalog_type",
+        firstNonBlank(
+            submitterConfigs.get("catalog_type"), sparkConfigs.get(catalogPrefix + ".type")));
+    placeholders.put(
+        "catalog_uri",
+        firstNonBlank(
+            submitterConfigs.get("catalog_uri"), sparkConfigs.get(catalogPrefix + ".uri"), ""));
+    placeholders.put(
+        "warehouse_location",
+        firstNonBlank(
+            submitterConfigs.get("warehouse_location"),
+            sparkConfigs.get(catalogPrefix + ".warehouse"),
+            ""));
+    return placeholders;
+  }
+
+  private static String firstNonBlank(String... candidates) {
+    for (String candidate : candidates) {
+      if (StringUtils.isNotBlank(candidate)) {
+        return candidate.trim();
+      }
+    }
+    return "";
   }
 
   private static final class TableTarget {
--- a/maintenance/optimizer-api/src/main/java/org/apache/gravitino/maintenance/optimizer/common/util/IcebergSparkConfigUtils.java
+++ b/maintenance/optimizer-api/src/main/java/org/apache/gravitino/maintenance/optimizer/common/util/IcebergSparkConfigUtils.java
@@ -106,18 +106,6 @@ public final class IcebergSparkConfigUtils {
     Preconditions.checkArgument(
         StringUtils.isNotBlank(tableIdentifier), "tableIdentifier is blank");
 
-    String extensions = StringUtils.trimToNull(sparkConfigs.get(SPARK_SQL_EXTENSIONS_KEY));
-    Preconditions.checkArgument(
-        StringUtils.isNotBlank(extensions),
-        "Spark config must contain key '%s' and include '%s'",
-        SPARK_SQL_EXTENSIONS_KEY,
-        ICEBERG_SPARK_EXTENSIONS);
-    Preconditions.checkArgument(
-        extensions.contains(ICEBERG_SPARK_EXTENSIONS),
-        "Spark config key '%s' must include '%s'",
-        SPARK_SQL_EXTENSIONS_KEY,
-        ICEBERG_SPARK_EXTENSIONS);
-
     String catalogPrefix = SPARK_SQL_CATALOG_PREFIX + catalogName;
     Preconditions.checkArgument(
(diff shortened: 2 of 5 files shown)


maintenance/jobs/build.gradle.kts
```diff
@@ -25,6 +25,11 @@ plugins {
   alias(libs.plugins.shadow)
 }
 
+java {
+  sourceCompatibility = JavaVersion.VERSION_17
+  targetCompatibility = JavaVersion.VERSION_17
+}
+
```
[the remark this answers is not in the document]
FANNG1: updated

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/command/SubmitUpdateStatsJobCommand.java
```diff
@@ -0,0 +1,288 @@
...
+    }
+    String gravitinoUri = StringUtils.trimToNull(updaterOptions.get("gravitino_uri"));
+    String metalake = StringUtils.trimToNull(updaterOptions.get("metalake"));
+    Preconditions.checkArgument(
+        StringUtils.isNotBlank(gravitinoUri),
+        "Option --updater-options (or config key jobSubmitterConfig.updater_options) "
+            + "must contain 'gravitino_uri' when update_mode is stats or all");
+    Preconditions.checkArgument(
+        StringUtils.isNotBlank(metalake),
+        "Option --updater-options (or config key jobSubmitterConfig.updater_options) "
```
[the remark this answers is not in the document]
FANNG1: Agreed, fixed. The validation message now references the full config key `gravitino.optimizer.jobSubmitterConfig.updater_options` for consistency with spark config validation.

maintenance/optimizer/src/test/java/org/apache/gravitino/maintenance/optimizer/recommender/job/TestBuiltinIcebergRewriteDataFiles.java
```diff
@@ -235,7 +235,12 @@ private static void runWithSparkAndMetalake(SparkMetalakeConsumer consumer) thro
...
   private static void submitCompactionJob(GravitinoMetalake metalake, Map<String, String> jobConf) {
-    JobHandle jobHandle = metalake.runJob(JOB_TEMPLATE_NAME, jobConf);
+    submitJob(metalake, JOB_TEMPLATE_NAME, jobConf);
+  }
+
+  private static void submitJob(
+      GravitinoMetalake metalake, String jobTemplateName, Map<String, String> jobConf) {
+    JobHandle jobHandle = metalake.runJob(jobTemplateName, jobConf);
     System.out.println("Submitted job id: " + jobHandle.jobId());
     Assertions.assertTrue(StringUtils.isNotBlank(jobHandle.jobId()), "Job id should not be blank");
```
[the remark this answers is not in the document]
FANNG1: Fixed. I removed the System.out.println from this test to keep CI logs clean and consistent with project logging practices.

maintenance/jobs/src/test/java/org/apache/gravitino/maintenance/jobs/iceberg/TestIcebergUpdateStatsJobWithSpark.java
```diff
@@ -0,0 +1,881 @@
...
+    String allModeTableName = "jobs_it_update_stats_mode_all_" + suffix;
+    String statsFullTableName = SPARK_CATALOG_NAME + ".db." + statsTableName;
+    String metricsFullTableName = SPARK_CATALOG_NAME + ".db." + metricsTableName;
+    String partitionMetricsFullTableName = SPARK_CATALOG_NAME + ".db." + partitionMetricsTableName;
+    String allModeFullTableName = SPARK_CATALOG_NAME + ".db." + allModeTableName;
+
+    SparkSession restSpark = createRestSparkSession();
+    try (MySQLContainer<?> mysql = new MySQLContainer<>("mysql:8.0.33");
+        GravitinoAdminClient adminClient = GravitinoAdminClient.builder(SERVER_URI).build()) {
+      mysql.start();
```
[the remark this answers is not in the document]
FANNG1: Good catch, fixed. I wrapped restSpark with try/finally and explicitly call restSpark.stop() to avoid SparkContext leakage across tests.

maintenance/jobs/build.gradle.kts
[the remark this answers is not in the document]
FANNG1: Addressed. I removed the conflicting Java 17 override and kept :maintenance:jobs under the repository JDK8-compatibility contract.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergUpdateStatsAndMetricsJob.java
```diff
@@ -0,0 +1,607 @@
...
+      return new HashMap<>();
+    }
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      Map<String, Object> parsedMap =
+          mapper.readValue(json, new TypeReference<Map<String, Object>>() {});
+      Map<String, String> configs = new HashMap<>();
+      for (Map.Entry<String, Object> entry : parsedMap.entrySet()) {
+        Object value = entry.getValue();
+        if (value instanceof Map || value instanceof List) {
```
jerryshao: `new ObjectMapper()` is created per call here, which is expensive. `ObjectMapper` instantiation involves class scanning and is not cheap. This should be a `private static final` field on the class — the same way `IcebergSparkConfigUtils` does it with `MAPPER`. Also, `parseCustomSparkConfigs` (line ~380) and `parseJsonOptions` (line ~384) are nearly identical — consider having one delegate to `IcebergSparkConfigUtils.parseFlatJsonMap` to eliminate the duplication.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/command/SubmitUpdateStatsJobCommand.java
```diff
@@ -0,0 +1,248 @@
...
+import org.apache.gravitino.maintenance.optimizer.common.util.IcebergSparkConfigUtils;
+
+/**
+ * Handles CLI command {@code submit-update-stats-job} for submitting built-in Iceberg update stats
+ * jobs directly from optimizer CLI.
+ */
+public class SubmitUpdateStatsJobCommand implements OptimizerCommandExecutor {
+
+  private static final String JOB_TEMPLATE_NAME = "builtin-iceberg-update-stats";
+  private static final String DEFAULT_UPDATE_MODE = "stats";
```
jerryshao: The CLI-level default is `"stats"`, but `IcebergUpdateStatsAndMetricsJob.DEFAULT_UPDATE_MODE` defaults to `"all"`. This means: submitting via the optimizer CLI without `--update-mode` runs stats-only, while invoking the job directly without args runs both stats+metrics. The two entry points have diverging defaults for the same parameter. Consider aligning them, or at minimum add a comment here explaining the intentional difference.

maintenance/optimizer/src/test/java/org/apache/gravitino/maintenance/optimizer/recommender/job/TestBuiltinIcebergUpdateStatsJob.java
```diff
@@ -0,0 +1,200 @@
...
+
+/** Environment IT for triggering built-in Iceberg update stats job through the client API. */
+@EnabledIfEnvironmentVariable(named = "GRAVITINO_ENV_IT", matches = "true")
+public class TestBuiltinIcebergUpdateStatsJob {
+
+  private static final String SERVER_URI = "http://localhost:8090";
+  private static final String ICEBERG_REST_URI = "http://localhost:9001/iceberg";
+  private static final String JOB_TEMPLATE_NAME = "builtin-iceberg-update-stats";
+  private static final String SPARK_CATALOG_NAME = "rest_catalog";
+  private static final String WAREHOUSE_LOCATION = "";
```
jerryshao: Empty string for warehouse location will silently misconfigure the Spark catalog and cause Iceberg to fail at runtime. Since this is gated by `@EnabledIfEnvironmentVariable`, either derive the value from an environment variable (e.g., `System.getenv("GRAVITINO_TEST_WAREHOUSE")`) or add an explicit `Assumptions.assumeTrue(StringUtils.isNotBlank(WAREHOUSE_LOCATION))` at the start of the test so it skips rather than fails with a confusing error.

maintenance/optimizer-api/src/main/java/org/apache/gravitino/maintenance/optimizer/common/util/StatisticValueSerdeUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.common.base.Preconditions;
+import org.apache.gravitino.json.JsonUtils;
+import org.apache.gravitino.stats.StatisticValue;
+
+/** Shared serde helpers for {@link StatisticValue}. */
+public final class StatisticValueSerdeUtils {
+
+  private StatisticValueSerdeUtils() {}
+
+  public static String toString(StatisticValue<?> value) {
```
jerryshao, line 39: Public API methods `toString` and `fromString` lack JavaDoc. Per project standards, public APIs require JavaDoc with at least `@param` and `@return`/`@throws`. Please add doc comments to both methods.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergUpdateStatsAndMetricsJob.java
```diff
@@ -0,0 +1,607 @@
...
+          metricsUpdater.close();
+        } catch (Exception e) {
+          LOG.warn("Failed to close metrics updater", e);
+        }
+      }
+      spark.stop();
+    }
+  }
+
+  static void updateStatistics(
```
jerryshao, line 155: The package-private static methods on this class (`parseArguments`, `buildTableStatsSql`, `buildPartitionStatsSql`, `parseUpdateMode`, `parseJsonOptions`, `buildOptimizerProperties`, `requireGravitinoConfig`, `parseCustomSparkConfigs`, `isPartitionedTable`, `toStatistics`, `updateStatistics`) are only package-private for test access. Please annotate them with `@com.google.common.annotations.VisibleForTesting` to make the intent explicit.

clients/client-java/build.gradle.kts
```diff
@@ -16,12 +16,27 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+import org.gradle.api.attributes.java.TargetJvmVersion
+import org.gradle.api.tasks.compile.JavaCompile
+
```
jerryshao, line 21: This adds Java 17 test compilation targets to `client-java` (which likely targets a lower Java version for its main sources). Please add a comment explaining why this is needed — presumably it is a workaround for transitive classpath conflicts introduced by the new `gravitino-updaters` / Spark test dependencies. Without context, this change looks like an unrelated side-effect.

FANNG1:
Addressed the latest review feedback in commit 8bf02e1b0544bec56dace0dd0e2492a001765151.

Updated items:
- Align `submit-update-stats-job` default `update_mode` to `all` to match the built-in job behavior.
- Reuse shared flat JSON parsing in `IcebergUpdateStatsAndMetricsJob` and add `@VisibleForTesting` on package-private static helpers.
- Quote SQL identifiers consistently (including metadata table `files`) and update tests.
- Fail fast for null partition values instead of writing string `"null"`.
- Add missing JavaDoc in `StatisticValueSerdeUtils`.
- Add clarifying comment in `clients/client-java/build.gradle.kts` for Java 8 main / Java 17 tests split.
- Stabilize IT warehouse config in `TestBuiltinIcebergUpdateStatsJob`.

Note: I did not apply the log-redaction change in `SubmitUpdateStatsJobCommand` in this round.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/command/SubmitUpdateStatsJobCommand.java
```diff
@@ -0,0 +1,248 @@
...
+    if (context.dryRun()) {
+      for (TableTarget tableTarget : tableTargets) {
+        Map<String, String> jobConfig =
+            buildJobConfig(tableTarget, updateMode, updaterOptions, sparkConfigs);
+        context
+            .output()
+            .printf(
+                "DRY-RUN: identifier=%s jobTemplate=%s jobConfig=%s%n",
+                tableTarget.fullIdentifier, JOB_TEMPLATE_NAME, jobConfig);
+      }
```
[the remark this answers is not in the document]
FANNG1: Thanks for calling this out. We discussed this and intentionally did not apply log redaction in this PR to keep the CLI output behavior unchanged for now. I agree this is a valid security hardening item, but we plan to handle it in a follow-up with a clearer redaction policy (to avoid partial masking pitfalls).

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergUpdateStatsAndMetricsJob.java
```diff
@@ -0,0 +1,607 @@
...
+            StatisticValues.longValue(toLongValue(row, "total_size"))));
+    return statistics;
+  }
+
+  static PartitionPath toPartitionPath(Row partitionRow) {
+    StructType partitionSchema = partitionRow.schema();
+    List<PartitionEntry> entries = new ArrayList<>(partitionSchema.fields().length);
+    for (int i = 0; i < partitionSchema.fields().length; i++) {
+      String name = partitionSchema.fields()[i].name();
+      Object value = partitionRow.get(i);
```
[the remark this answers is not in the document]
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. `toPartitionPath()` now explicitly rejects null partition values with a clear `IllegalArgumentException` instead of converting null to the literal string `null`.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergUpdateStatsAndMetricsJob.java
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. `parseJsonOptions` now delegates to `IcebergSparkConfigUtils.parseFlatJsonMap`, and duplicated per-call JSON parsing logic has been removed.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergUpdateStatsAndMetricsJob.java
```diff
@@ -0,0 +1,607 @@
...
+    return escapeSqlIdentifier(catalogName)
+        + "."
+        + escapeSqlIdentifier(levels[0])
+        + "."
+        + escapeSqlIdentifier(levels[1])
+        + ".files";
+  }
+
+  private static String escapeSqlIdentifier(String identifier) {
+    return identifier.replace("`", "``");
```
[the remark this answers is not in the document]
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. `escapeSqlIdentifier()` now wraps each identifier part with backticks after escaping, and the files metadata table is referenced as quoted identifier as well.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/command/SubmitUpdateStatsJobCommand.java
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. I aligned CLI default `update_mode` to `all` so optimizer submission and direct job invocation now have the same default behavior.

maintenance/optimizer-api/src/main/java/org/apache/gravitino/maintenance/optimizer/common/util/StatisticValueSerdeUtils.java
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. Added JavaDoc (including `@param`, `@return`, and `@throws`) for both `toString` and `fromString`.

maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/IcebergUpdateStatsAndMetricsJob.java
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. Added `@VisibleForTesting` to the package-private static helpers that are package-visible primarily for test access.

clients/client-java/build.gradle.kts
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. Added an explicit comment in `clients/client-java/build.gradle.kts` explaining why `compileTestJava` uses release 17 while the main artifact remains Java 8-compatible.

maintenance/optimizer/src/test/java/org/apache/gravitino/maintenance/optimizer/recommender/job/TestBuiltinIcebergUpdateStatsJob.java
FANNG1: Fixed in 8bf02e1b0544bec56dace0dd0e2492a001765151. The test now derives warehouse from `GRAVITINO_TEST_WAREHOUSE` with a safe fallback and adds an explicit assumption guard to avoid confusing runtime failures.

jerryshao (approved):
LGTM, approving after review. Please address the inline comments before merging.

PR #10116, opened 2026-03-02
Optional metalake

bharos:

Metalake is not supported in static-config provider. So it should be optional as it's only available on using dynamic-config-provider

Currently, metalake is always set to default value of 'gravitino' even if it's not set to any value. Ideally, if metalake is not supported, it should be empty/null

No. If the configs are correct, users should not see any differences

Unit Tests

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/RESTService.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/RESTService.java
@@ -28,7 +28,6 @@ import javax.servlet.Servlet;
 import org.apache.gravitino.Configs;
 import org.apache.gravitino.GravitinoEnv;
 import org.apache.gravitino.auxiliary.GravitinoAuxiliaryService;
-import org.apache.gravitino.catalog.lakehouse.iceberg.IcebergConstants;
 import org.apache.gravitino.iceberg.common.IcebergConfig;
 import org.apache.gravitino.iceberg.service.IcebergCatalogWrapperManager;
 import org.apache.gravitino.iceberg.service.IcebergExceptionMapper;
@@ -109,7 +108,7 @@ public class RESTService implements GravitinoAuxiliaryService {
 
     String metalakeName = metalakeOpt.orElse(null);
     this.icebergCatalogWrapperManager =
-        new IcebergCatalogWrapperManager(configProperties, configProvider, auxMode, metalakeName);
+        new IcebergCatalogWrapperManager(configProperties, configProvider, auxMode, metalakeOpt);
     IcebergRESTServerContext authorizationContext =
         IcebergRESTServerContext.create(
             configProvider, enableAuth, auxMode, icebergCatalogWrapperManager);
@@ -125,7 +124,8 @@ public class RESTService implements GravitinoAuxiliaryService {
           new IcebergTableHookDispatcher(icebergTableOperationDispatcher);
     }
     IcebergTableEventDispatcher icebergTableEventDispatcher =
-        new IcebergTableEventDispatcher(icebergTableOperationDispatcher, eventBus, metalakeForEvents);
+        new IcebergTableEventDispatcher(
+            icebergTableOperationDispatcher, eventBus, metalakeForEvents);
 
     IcebergViewOperationDispatcher icebergViewOperationDispatcher =
         new IcebergViewOperationExecutor(icebergCatalogWrapperManager);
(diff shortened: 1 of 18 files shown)


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/authorization/IcebergRESTServerContext.java
```diff
@@ -56,7 +56,7 @@ public static IcebergRESTServerContext create(
         new IcebergRESTServerContext(
             enableAuth,
             auxMode,
-            configProvider.getMetalakeName(),
+            configProvider.getMetalakeName().orElse(null),
```
roryqi: IcebergRESTServerContext's metalake should be optional, too.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/RESTService.java
```diff
@@ -95,32 +98,43 @@ private void initServer(IcebergConfig icebergConfig) {
...
+    String metalakeName = metalakeOpt.orElse(null);
     this.icebergCatalogWrapperManager =
         new IcebergCatalogWrapperManager(configProperties, configProvider, auxMode, metalakeName);
     IcebergRESTServerContext authorizationContext =
         IcebergRESTServerContext.create(
             configProvider, enableAuth, auxMode, icebergCatalogWrapperManager);
     this.icebergMetricsManager = new IcebergMetricsManager(icebergConfig);
+
+    // Events need a metalake name so just use the default if it's not provided
+    String metalakeForEvents = metalakeOpt.orElse(IcebergConstants.ICEBERG_REST_DEFAULT_METALAKE);
```
bharos, line 118: @roryqi What about the events dispatcher ?
I think it will try to use the metalake name, and if it is null it might fail? ie. with static-config-provider

Is it then better to pass a default like this

roryqi: Maybe this should be optional, too.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergNamespaceOperations.java
```diff
@@ -115,7 +115,8 @@ public Response listNamespaces(
             IcebergRESTServerContext authContext = IcebergRESTServerContext.getInstance();
             if (authContext.isAuthorizationEnabled()) {
               response =
-                  filterListNamespacesResponse(response, authContext.metalakeName(), catalogName);
+                  filterListNamespacesResponse(
+                      response, authContext.metalakeName().orElse(null), catalogName);
```
bharos, line 119: Here also, if it is null, it can fail during NameIdentifier creation like here:
https://github.com/apache/gravitino/blob/7baafec434ebd5ddb4d54331a540321fd9cff85a/api/src/main/java/org/apache/gravitino/NameIdentifier.java#L49

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableEventDispatcher.java
```diff
@@ -77,10 +78,10 @@ public class IcebergTableEventDispatcher implements IcebergTableOperationDispatc
   public IcebergTableEventDispatcher(
       IcebergTableOperationDispatcher icebergTableOperationDispatcher,
       EventBus eventBus,
-      String metalakeName) {
+      Optional<String> metalakeName) {
     this.icebergTableOperationDispatcher = icebergTableOperationDispatcher;
     this.eventBus = eventBus;
-    this.metalakeName = metalakeName;
+    this.metalakeName = metalakeName.orElse(IcebergConstants.ICEBERG_REST_DEFAULT_METALAKE);
```
bharos, line 84: I think this will  now become null for static-config-provider.
In that case, should we use ICEBERG_REST_DEFAULT_METALAKE

Otherwise, I am not sure if it causes backward incompatibility

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergTableOperationExecutor.java
```diff
@@ -164,7 +164,10 @@ public LoadCredentialsResponse getTableCredentials(
 
   private static CredentialPrivilege getCredentialPrivilege(
       IcebergRequestContext context, TableIdentifier tableIdentifier) {
-    String metalake = IcebergRESTServerContext.getInstance().metalakeName();
+    String metalake =
+        IcebergRESTServerContext.getInstance()
+            .metalakeName()
+            .orElse(IcebergConstants.ICEBERG_REST_DEFAULT_METALAKE);
```
bharos, line 170: Also, here, I am not sure if it will work if we use null. Below, toGravitinoTableIdentifier() is called

bharos:
I think the PR is too complicated as it can introduce NPEs and also potential backwards incompatibilities, because currently static-provider defaults to 'gravitino' as metalake name. Now it can be null and there can be paths where it can break. So I will close this PR for now
cc @roryqi

Issue #10134, opened 2026-03-03
[Improvement] Possible NPE in AlterFilesetFailureEvent


justinmclean:
### What would you like to be improved?

There is a failure path bug in AlterFilesetFailureEvent. When filesetChanges is null, the constructor calls filesetChanges.clone() and throws NullPointerException.

### How should we improve?

Potentially make the constructor null-safe by handling filesetChanges == null before cloning.

Here's a test to help:
```
public class AlterFilesetFailureEventTest {

  @Test
  public void testNullFilesetChangesDoesNotThrow() {
    Assertions.assertDoesNotThrow(
        () ->
            new AlterFilesetFailureEvent(
                "user", NameIdentifier.of("metalake", "catalog", "fileset"), new Exception(), null));
  }
}
```

justinmclean:
@a638011 Please ask to work on an issue before providing a pull request, this is so that we don't have multiple people working on the same PR at once.

LukasDEDD:
Hi, I would like to work on this issue. Could you please assign it to me?

justinmclean:
Sorry @LukasDEDD, my mistake,  @a638011 is already working on this, please pick another issue to work on, I'll be adding some more issues in the next few hours.

a638011:
Hi @justinmclean, I’ve opened PR #10135 for this issue and pushed spotless formatting/tests. Could this issue be assigned to me while the PR is under review? Thanks!

justinmclean:
this has been fixed

PR #10135, opened 2026-03-03
[#10134] fix(catalog): prevent NPE when filesetChanges is null

a638011:

Fix a potential NullPointerException in AlterFilesetFailureEvent when filesetChanges is null.

When filesetChanges is null, the constructor calls filesetChanges.clone() which throws NPE. This can happen in failure scenarios where the changes array is not available.

Fix issue #10134

### Does this PR introduce any user-facing changes?

No

Added unit test AlterFilesetFailureEventTest that verifies null filesetChanges does not throw NPE.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/listener/api/event/AlterFilesetFailureEventTest.java
+++ b/core/src/test/java/org/apache/gravitino/listener/api/event/AlterFilesetFailureEventTest.java
@@ -30,6 +30,9 @@ public class AlterFilesetFailureEventTest {
     Assertions.assertDoesNotThrow(
         () ->
             new AlterFilesetFailureEvent(
-                "user", NameIdentifier.of("metalake", "catalog", "fileset"), new Exception(), null));
+                "user",
+                NameIdentifier.of("metalake", "catalog", "fileset"),
+                new Exception(),
+                null));
   }
 }


justinmclean:
@a638011, can you please fix up the PR title to include the issue number? Can you please also request an issue before working on it, so it can be assigned to you.

a638011:
Hi @justinmclean, thanks for the feedback! I'll update the PR title to include the issue number in the proper format and request the issue for assignment. Could you clarify the expected format?

justinmclean:
@a638011 We use spotless to make sure code is consistently formatted. You'll need to run `./gradlew :core:spotlessApply` and push your changes to correct the CI failure.

a638011:
Thanks @justinmclean! I'll run `./gradlew :core:spotlessApply` locally and push the changes. The CI is taking a while to run spotless, so I'll update once the formatting is applied.

core/src/main/java/org/apache/gravitino/listener/api/event/AlterFilesetFailureEvent.java
```diff
@@ -45,7 +45,7 @@ public final class AlterFilesetFailureEvent extends FilesetFailureEvent {
   public AlterFilesetFailureEvent(
       String user, NameIdentifier identifier, Exception exception, FilesetChange[] filesetChanges) {
     super(user, identifier, exception);
-    this.filesetChanges = filesetChanges.clone();
+    this.filesetChanges = filesetChanges == null ? null : filesetChanges.clone();
```
yuqi1129, line 48: `filesetChanges` could not be null. Please refer to the following code

https://github.com/apache/gravitino/blob/fbffe05e39113c9f21fcc1d55106aeb3e2459107/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java#L298-L302

justinmclean: That's only partially true. In the REST path, changes are indeed non-null. But AlterFilesetFailureEvent is not REST-specific. It is built in a generic dispatcher (see FilesetEventDispatcher.java) from varargs FilesetChange... changes.

Java varargs can be called with a null array (e.g. alterFileset(ident, (FilesetChange[]) null)), and the API does notenforce non-null (see FilesetCatalog.java)

a638011:
Hi @justinmclean, I’ve now commented on issue #10134 requesting assignment as suggested.\n\nCurrent PR status from my side:\n- Title already includes issue number format: [#10134] ...\n- spotless formatting applied\n- all visible checks are green\n\nPlease let me know if you want any additional change in this PR.

core/src/main/java/org/apache/gravitino/listener/api/event/AlterFilesetFailureEvent.java
justinmclean: @yuqi1129 any thoughts?

yuqi1129: OK, as you have said some time before, adding something is better than not adding anything.

Issue #9782, opened 2026-01-22
[Bug report] IRC multiple instances: creating tables  concurrently encounter error occasionally: Cannot get a client from a closed pool


danhuawang:
### Version

main branch

### Describe what's wrong

I scale the IRC to 2 instance, then  create tables with 10 users concurrently,  It occasionally pop the following error:

```
================================================================================
ERROR REPORT - DETAILED REQUEST/RESPONSE INFORMATION
================================================================================

  4 occurrences | POST http://10.50.0.7:9001/iceberg/v1/iceberg/namespaces/dml25/tables: 500
--------------------------------------------------------------------------------
  Example 1:
    Time: 11:00:25
    URL: http://10.50.0.7:9001/iceberg/v1/iceberg/namespaces/dml25/tables
    Method: POST
    Status Code: 500
    Exception: HTTP 500 Error
    Request Body:
      {
        "name": "test_tb_0053860",
        "location": "gs://strato-iceberg/IRCperform/dml25/test_tb_0053860",
        "schema": {
          "type": "struct",
          "schema-id": 0,
          "fields": [
            {
              "id": 1,
              "name": "id",
              "required": true,
              "type": "int",
              "write-default": 1
            },
            {
      ... (request body truncated)
    Response Body:
      {
        "error": {
          "message": "Cannot get a client from a closed pool",
          "type": "IllegalStateException",
          "code": 500,
          "stack": [
            "java.lang.IllegalStateException: Cannot get a client from a closed pool",
            "\tat org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkState(Preconditions.java:513)",
            "\tat org.apache.iceberg.ClientPoolImpl.get(ClientPoolImpl.java:136)",
            "\tat org.apache.iceberg.ClientPoolImpl.run(ClientPoolImpl.java:70)",
            "\tat org.apache.iceberg.ClientPoolImpl.run(ClientPoolImpl.java:65)",
            "\tat org.apache.iceberg.jdbc.JdbcUtil.tableOrView(JdbcUtil.java:602)",
            "\tat org.apache.iceberg.jdbc.JdbcUtil.loadTable(JdbcUtil.java:641)",
            "\tat org.apache.iceberg.jdbc.JdbcTableOperations.doCommit(JdbcTableOperations.java:109)",
            "\tat org.apache.iceberg.BaseMetastoreTableOperations.commit(BaseMetastoreTableOperations.java:126)",
[...]

PR #10180, opened 2026-03-04
[#9782] improvement(iceberg): Optimize the catalog wrapper and entity cache expiry strategy

roryqi:

Optimize the catalog wrapper entity cache expiry strategy.  
After https://github.com/apache/gravitino/pull/9969, we can expire the cache timely. So we can change the cache expiry strategy.

Without this patch, the catalog may be closed because of cache expiry when we access the catalog.

Fix: #9782

No.

Code review.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -403,7 +403,7 @@ public class CaffeineEntityCache extends BaseEntityCache {
     }
 
     if (cacheConfig.get(Configs.CACHE_EXPIRATION_TIME) > 0) {
-      builder.expireAfterWrite(
+      builder.expireAfterAccess(
           cacheConfig.get(Configs.CACHE_EXPIRATION_TIME), TimeUnit.MILLISECONDS);
     }
 


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergCatalogWrapperManager.java
```diff
@@ -56,7 +56,7 @@ public IcebergCatalogWrapperManager(
     this.configProvider = configProvider;
     this.catalogWrapperCache =
         Caffeine.newBuilder()
-            .expireAfterWrite(
+            .expireAfterAccess(
```
yuqi1129, line 59: Can you also modify the same problem in `CaffeineEntityCache`?

roryqi: Maybe not. If I modify it, the CaffeineEntityCache won't update if the metadata always is accessed.

yuqi1129: What's the problem with `the CaffeineEntityCache won't update if the metadata always is accessed`?

roryqi: If a catalog is changed in the another server, the catalog won't be updated in this server if the catalog is always accessed.

Catalog A in the server 1 is cached.
Catalog A in the server 2 is changed.
Catalog A in the server 1 is accessed in every 10 minutes. The Catalog A is expired in 1 hour. The catalog A won't be changed. We will read the legacy catalog A in the server 1.

If we don't change it, we can access the new data after 1 hour.

roryqi: If you think this isn't an issue, I can change CaffeineEntityCache, too.

yuqi1129: This is a known problem, and Gravitino does not support distributed deployment.

roryqi: OK, I can change CaffeineEntityCache, too.

PR #10182, opened 2026-03-04
[#10171] fix(server): Add null request body checks to REST create/register endpoints

### LINKED ISSUE #10171: [Improvement] NPEs before try/catch in REST create/register endpoints

Several REST endpoints dereference `request` fields (for example `request.getName()` or `request.getJobTemplateName()`) before entering their try block.

If request deserialization yields `null` (for example empty body, literal JSON `null`, or binding edge cases), this can trigger an uncaught `NullPointerException` and bypasses expected `ExceptionHandlers` path.

## Scope
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/TableOperations.java` `createTable`
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java` `createFileset`
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/FunctionOperations.java` `registerFunction`
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/ModelOperations.java` `registerModel`
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/SchemaOperations.java` `createSchema`
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java` `registerJobTemplate`
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java` `runJob`

## Expected behavior
- No `request.get*()` access before null-safe handling.
- Null request input should return stable structured error responses (for example, bad request / mapped handler response), not uncaught NPE.
- Endpoint APIs and behavior to remain unchanged except for safer null handling.

## Definition of done
- [ ] Remove pre-try request dereferences in listed methods.
- [ ] Add/adjust unit tests for null request body behavior in matching REST test classes.

## Example unit test
```java
@Test
public void testCreateTableWithNullRequestBody() {
  Response resp =
      target(tablePath(metalake, catalog, schema))

  Assertions.assertEquals(Status.BAD_REQUEST.getStatusCode(), resp.getStatus());
}
```

shaojunying:

Add null checks for request body parameters before accessing request fields in REST endpoints. Previously, if request deserialization yielded `null` (e.g., empty body, literal JSON `null`, or binding edge cases), accessing `request.getName()` or similar methods before the `try` block would trigger an uncaught `NullPointerException`, bypassing the expected `ExceptionHandlers` error handling path.

Several REST endpoints dereference `request` fields (e.g., `request.getName()`, `request.getJobTemplateName()`) before entering their `try` block. This causes uncaught NPEs that bypass the structured error response path, returning unhelpful 500 errors instead of proper 400 Bad Request responses.

Fix #10171

No API changes. Null request bodies now return a structured `400 Bad Request` error response instead of an unhandled `500 Internal Server Error`.

Added unit tests for null request body behavior in all matching REST test classes:
- `TestTableOperations.testCreateTableWithNullRequestBody`
- `TestFilesetOperations.testCreateFilesetWithNullRequestBody`
- `TestFunctionOperations.testRegisterFunctionWithNullRequestBody`
- `TestModelOperations.testRegisterModelWithNullRequestBody`
- `TestSchemaOperations.testCreateSchemaWithNullRequestBody`
- `TestJobOperations.testRegisterJobTemplateWithNullRequestBody`
- `TestJobOperations.testRunJobWithNullRequestBody`

### Affected endpoints

| File | Method |
|------|--------|
| `TableOperations.java` | `createTable` |
| `FilesetOperations.java` | `createFileset` |
| `FunctionOperations.java` | `registerFunction` |
| `ModelOperations.java` | `registerModel` |
| `SchemaOperations.java` | `createSchema` |
| `JobOperations.java` | `registerJobTemplate` |
| `JobOperations.java` | `runJob` |

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
@@ -144,6 +144,12 @@ public class FilesetOperations {
       @PathParam("catalog") @AuthorizationMetadata(type = Entity.EntityType.CATALOG) String catalog,
       @PathParam("schema") @AuthorizationMetadata(type = Entity.EntityType.SCHEMA) String schema,
       FilesetCreateRequest request) {
+    if (request == null) {
+      LOG.warn("Received create fileset request with null request body");
+      return ExceptionHandlers.handleFilesetException(
+          OperationType.CREATE, "", schema, new IllegalArgumentException("Request body cannot be null"));
+    }
+
     LOG.info(
         "Received create fileset request: {}.{}.{}.{}",
         metalake,
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/FunctionOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/FunctionOperations.java
@@ -127,6 +127,12 @@ public class FunctionOperations {
       @PathParam("catalog") String catalog,
       @PathParam("schema") String schema,
       FunctionRegisterRequest request) {
+    if (request == null) {
+      LOG.warn("Received register function request with null request body");
+      return ExceptionHandlers.handleFunctionException(
+          OperationType.REGISTER, "", schema, new IllegalArgumentException("Request body cannot be null"));
+    }
+
     LOG.info(
         "Received register function request: {}.{}.{}.{}",
         metalake,
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java
@@ -150,6 +150,12 @@ public class JobOperations {
       @PathParam("metalake") @AuthorizationMetadata(type = Entity.EntityType.METALAKE)
           String metalake,
       JobTemplateRegisterRequest request) {
+    if (request == null) {
+      LOG.warn("Received register job template request with null request body");
+      return ExceptionHandlers.handleJobTemplateException(
+          OperationType.REGISTER, "", metalake, new IllegalArgumentException("Request body cannot be null"));
+    }
+
     LOG.info(
         "Received request to register job template {} in metalake: {}",
         request.getJobTemplate().name(),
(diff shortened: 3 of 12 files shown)


justinmclean:
Looks good, but a few minor improvements needed:
- Null-body fix in registerJobTemplate is incomplete; it can still trigger an unhandled NPE before try/catch
- PR contains unrelated changes in TestCompactionStrategyHandler.java?
- Minor style inconsistency in tests with FQN used inline

jerryshao:
Just show me a real case on how to reproduce this issue. AFAIK, this will not happen in the real environment.

shaojunying:
Thanks for the review @justinmclean! Here's my plan for addressing your feedback:

1. **`registerJobTemplate` NPE in catch block**: Good catch — the `catch` block still dereferences `request.getJobTemplate().name()`, which can NPE if `getJobTemplate()` returns null (even after the null request body check). I'll fix this by moving the name extraction inside the `try` block or adding a safe fallback in the `catch` block.

2. **Unrelated `TestCompactionStrategyHandler.java` changes**: Will remove those from this PR.

3. **FQN style inconsistency in tests**: I noticed `TestSchemaOperations.java` uses `javax.ws.rs.client.Entity.entity(...)` inline instead of the imported `Entity`. Will fix this to be consistent with the other test files.

I'll push the fixes shortly.

shaojunying:
Hi @jerryshao, thanks for the question. Here's a concrete reproduction case:

**Steps to reproduce:**
```bash
# Start a Gravitino server, then send a null JSON body to the createTable endpoint:
curl -X POST "http://localhost:8090/api/metalakes/my_metalake/catalogs/my_catalog/schemas/my_schema/tables" \
  -H "Content-Type: application/json" \
  -H "Accept: application/vnd.gravitino.v1+json" \
  -d "null"
```

When JAX-RS (Jersey) deserializes the literal JSON `null`, the `request` parameter becomes `null` in Java. The current code then does:
```java
LOG.info("Received create table request: {}.{}.{}.{}", metalake, catalog, schema, request.getName());
//                                                                                  ^^^^^^^^^^^^^^^ NPE here
try {
```
This happens **before** the `try/catch` block, so the NPE is not caught by `ExceptionHandlers` and the client gets an unhelpful HTTP 500 instead of a structured 400 Bad Request.

You're right that normal Gravitino clients won't send `null` bodies, but this can happen with:
- Direct REST API calls (curl, custom HTTP clients)
- Malformed client implementations
- Security scanning / fuzzing tools

This is essentially a defensive programming improvement to ensure all REST endpoints return proper error responses at system boundaries, consistent with the existing `ExceptionHandlers` error handling pattern.

jerryshao:
If this is the case, this will affect all the requests, is there a centralized place that can fix all the related issues, not just do the null check in every place?

I think we need to fix it in the Jackson / Jersey object mapper configuration.

justinmclean:
Not how we currently have this set up as far as I'm aware. ObjectMapper can enforce many field-level rules, but not null method arguments at JAX-RS entry.

PR #10183, opened 2026-03-04
[#10172] fix(server): Guard catch-block request dereferences in REST handlers

### LINKED ISSUE #10172: [Improvement] Catch-path request dereference can mask original exceptions in REST handlers

Multiple REST handlers dereference `request` inside `catch (Exception e)` (for example `request.getName()`, `request.getRoleNames()`, `request.getUpdates()`, `request.getDrops()`, `request.getJobTemplate().name()`). When request is `null`, catch-path code can throw a second `NullPointerException`, masking the real failure.

Representative failure:
- Primary exception occurs in method body.
- Catch block dereferences `request.get*()`.
- Secondary `NullPointerException` in catch hides the original exception context.

## Scope
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/TableOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/FunctionOperations.java` register catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/ModelOperations.java` register catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/SchemaOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/MetalakeOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java` create/testConnection catch paths
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/UserOperations.java` add catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java` add catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/RoleOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/TagOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/TopicOperations.java` create catch path
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/PermissionOperations.java` grant/revoke catch paths
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/StatisticOperations.java` update/drop (+partition update/drop) catch paths
- [ ] `server/src/main/java/org/apache/gravitino/server/web/rest/JobOperations.java` register-job-template catch path

## Expected behavior
- No `request.get*()` dereference inside affected `catch` blocks.
- Catch blocks should always preserve/marshal the primary error through existing `ExceptionHandlers`.
- Endpoint APIs and response contracts remain unchanged.

## Definition of done
- [ ] Replace catch-path request dereferences with null-safe precomputed identifiers or safe fallbacks.
- [ ] Add/adjust tests to ensure exception handling path is stable when request is null.

## Example unit test
```java
@Test
public void testAddGroupWithNullRequestBodyDoesNotExposeNpe() {
  Response resp =
      target("/metalakes/metalake1/groups")

  Assertions.assertEquals(Status.INTERNAL_SERVER_ERROR.getStatusCode(), resp.getStatus());
  ErrorResponse error = resp.readEntity(ErrorResponse.class);
  Assertions.assertNotEquals(NullPointerException.class.getSimpleName(), error.getType());
}
```

shaojunying:

Add null-safe access for `request` objects in `catch` blocks across 16 REST handler files. Each unsafe `request.get*()` call in a catch block is replaced with a ternary null check, e.g.:

```java
// Before (unsafe):
request.getName()

// After (safe):
request != null ? request.getName() : ""
```

When `request` is `null` (e.g., due to deserialization failure) and an exception occurs inside the `try` block, the `catch` block attempts to dereference `request` to build an error message. This throws a secondary `NullPointerException` that **masks the original exception**, making debugging significantly harder.

The existing fix in `MetalakeOperations.createMetalake` (line 142) already demonstrates the correct pattern:
```java
String metalakeName = request != null ? request.getName() : "";
```

This PR applies the same pattern consistently across all remaining REST handlers.

No API changes. Error responses now correctly reflect the original exception instead of being masked by a secondary NPE in the catch block.

Manual review of all 16 affected files. The changes are purely defensive — they only affect the error path behavior when `request` is null, preserving all existing normal-path behavior.

Affected handlers (23 catch-block sites across 16 files):

| File | Method(s) |
|------|-----------|
| `CatalogOperations` | `createCatalog`, `testConnection`, `setCatalog` |
| `SchemaOperations` | `createSchema` |
| `TableOperations` | `createTable` |
| `FilesetOperations` | `createFileset` |
| `FunctionOperations` | `registerFunction` |
| `ModelOperations` | `registerModel` |
| `TopicOperations` | `createTopic` |
| `TagOperations` | `createTag` |
| `PolicyOperations` | `createPolicy` |
| `RoleOperations` | `createRole` |
| `GroupOperations` | `addGroup` |
| `UserOperations` | `addUser` |
| `PermissionOperations` | `grantRolesToUser`, `grantRolesToGroup`, `revokeRolesFromUser`, `revokeRolesFromGroup` |
| `JobOperations` | `registerJobTemplate` |
| `StatisticOperations` | `updateStatistics`, `dropStatistics`, `updatePartitionStatistics`, `dropPartitionStatistics` |
| `MetalakeOperations` | `setMetalake` |

Closes #10172

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java
@@ -162,7 +162,7 @@ public class CatalogOperations {
 
     } catch (Exception e) {
       return ExceptionHandlers.handleCatalogException(
-          OperationType.CREATE, request.getName(), metalake, e);
+          OperationType.CREATE, request != null ? request.getName() : "", metalake, e);
     }
   }
 
@@ -198,7 +198,10 @@ public class CatalogOperations {
           });
 
     } catch (Exception e) {
-      LOG.info("Failed to test connection for catalog: {}.{}", metalake, request.getName());
+      LOG.info(
+          "Failed to test connection for catalog: {}.{}",
+          metalake,
+          request != null ? request.getName() : "");
       return ExceptionHandlers.handleTestConnectionException(e);
     }
   }
@@ -244,7 +247,7 @@ public class CatalogOperations {
     } catch (Exception e) {
       LOG.info(
           "Failed to {} catalog: {}.{}",
-          request.isInUse() ? "enable" : "disable",
+          request != null ? (request.isInUse() ? "enable" : "disable") : "set",
           metalake,
           catalogName);
       return ExceptionHandlers.handleCatalogException(op, catalogName, metalake, e);
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
@@ -181,7 +181,7 @@ public class FilesetOperations {
 
     } catch (Exception e) {
       return ExceptionHandlers.handleFilesetException(
-          OperationType.CREATE, request.getName(), schema, e);
+          OperationType.CREATE, request != null ? request.getName() : "", schema, e);
     }
   }
 
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/FunctionOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/FunctionOperations.java
@@ -161,7 +161,7 @@ public class FunctionOperations {
           });
     } catch (Exception e) {
       return ExceptionHandlers.handleFunctionException(
-          OperationType.REGISTER, request.getName(), schema, e);
+          OperationType.REGISTER, request != null ? request.getName() : "", schema, e);
     }
   }
 
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/GroupOperations.java
@@ -116,7 +116,7 @@ public class GroupOperations {
           });
     } catch (Exception e) {
       return ExceptionHandlers.handleGroupException(
-          OperationType.ADD, request.getName(), metalake, e);
+          OperationType.ADD, request != null ? request.getName() : "", metalake, e);
     }
   }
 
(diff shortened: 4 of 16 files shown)


justinmclean:
Thanks for the PR, but a little more is needed here:
1. Still vulnerable to secondary NPE in catch blocks - The inline null-guards only protect the request, not nested fields. 
2. The Inline ternary style is getting hard to read/maintain.
3. It is missing unit tests to cover the changes.

jerryshao:
How do you trigger this issue? As I mentioned in another PR, this should not happen normally. If the incoming JSON data is empty or malformed, then the Jersey/Jackson framework will throw an exception.  It will not defer to here to meet the NPE.

jerryshao:
This is the Java code, not C/C++. For C/C++, a null pointer will lead to an unknown exception like segment fault, but this normally is not the case for Java. If we obsessively check and handle the null value, it is not very idiomatic Java.

justinmclean:
@jerryshao Thanks for the feedback. I agree we should avoid defensive checks everywhere, but I verified this specific path is reproducible in normal server execution.

How to trigger:

Start GravitinoServer
Send POST /api/metalakes/m/catalogs/c/schemas
Headers: Content-Type: application/json, Accept: application/vnd.gravitino.v1+json
Body: empty
Observed result:

createMetalake path (with null guard): returns 400
createSchema path (no null guard): returns 500

So the request body can still be null at handler boundary in our stack, and endpoints without an early guard can fail with server error.

This is not about C/C++-style pointer safety; it is about HTTP contract correctness in Java:

Issue #10097, opened 2026-03-01
[Subtask] add build-in compaction policy


FANNG1:
### Describe the subtask

add build-in compaction policy

### Parent issue

#9545

PR #10189, opened 2026-03-04
[#10097] feat(core): support builtin iceberg compaction policy

FANNG1:

  This PR adds a built-in Iceberg compaction policy with typed content and
  aligns policy type handling to enum-based built-in types.

  1. Added typed policy content for ICEBERG_COMPACTION (instead of
     CustomContent):

  - minDatafileMse
  - minDeleteFileNumber
  - rewriteOptions (expanded to job.options.*)

  2. Added built-in compaction defaults in policy content:

  - strategy.type=compaction
  - job.template-name=builtin-iceberg-rewrite-data-files
  - built-in trigger/score expressions based on custom-datafile_mse and custom-
    delete_file_number

  3. Added DTO/REST/convertor support for typed content serialization/
     deserialization.
  4. Added core-side built-in policy content type validation to prevent
     mismatched content type.
  5. Aligned built-in policy type with enum and prefix:

  - removed obsolete TODO in Policy.BuiltInType
  - ICEBERG_COMPACTION.policyType() now uses system_iceberg_compaction
  - REST policy type output is aligned to enum name (ICEBERG_COMPACTION)

  - iceberg-compaction is a core built-in policy and should be strongly typed,
    validated, and stable, rather than relying on generic CustomContent.
  - Strong typing improves correctness, validation, and compatibility across
    API/DTO/core/optimizer integration.
  - Enum + built-in prefix alignment standardizes policy type semantics and
    avoids ambiguous type representations.

  Fix: #10097 

  Yes.

  1. A built-in policy type for Iceberg compaction is available with typed
     content fields:

  - minDatafileMse
  - minDeleteFileNumber
  - rewriteOptions

  2. Built-in policy type representation is aligned:

  - built-in type value: system_iceberg_compaction
  - REST response policy type for this built-in policy: ICEBERG_COMPACTION

  3. Built-in compaction policy now uses a fixed built-in job template:

  - builtin-iceberg-rewrite-data-files

  Added/updated unit tests and ran targeted test suites:

  - org.apache.gravitino.policy.TestPolicyBuiltInType
  - org.apache.gravitino.policy.TestPolicyContents
  - org.apache.gravitino.dto.policy.TestPolicyDTO
  - org.apache.gravitino.meta.TestPolicyEntity
  - org.apache.gravitino.server.web.rest.TestPolicyOperations

  Executed with Gradle (targeted modules/tests), all passed

### DIFF
--- a/api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
+++ b/api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
@@ -189,6 +189,7 @@ public class IcebergDataCompactionContent implements PolicyContent {
 
   @Override
   public Map<String, String> properties() {
+    // properties keep stable strategy/job identity; thresholds and scoring knobs belong to rules.
     return ImmutableMap.of(
         STRATEGY_TYPE_KEY, STRATEGY_TYPE_VALUE, JOB_TEMPLATE_NAME_KEY, JOB_TEMPLATE_NAME_VALUE);
   }


FANNG1:
@mchades  could you help me review the changes to the build-in policy?  you could ignore the policy content part in the first stage.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -338,7 +337,7 @@ static PolicyDTO toDTO(PolicyEntity policy, Optional<Boolean> inherited) {
         PolicyDTO.builder()
             .withName(policy.name())
             .withComment(policy.comment())
-            .withPolicyType(policy.policyType().name().toLowerCase(Locale.ROOT))
+            .withPolicyType(policy.policyType().name())
```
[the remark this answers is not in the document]
FANNG1: updated

FANNG1:
@jerryshao @mchades PTAL

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
```diff
@@ -128,4 +176,233 @@ public String toString() {
           + '}';
     }
   }
+
+  /** Built-in policy content for Iceberg compaction strategy. */
+  public static class IcebergCompactionContent implements PolicyContent {
```
jerryshao: Can you split this class to a new file?

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
```diff
@@ -128,4 +176,233 @@ public String toString() {
...
+    /** Rule key for trigger expression. */
+    public static final String TRIGGER_EXPR_KEY = "trigger-expr";
+    /** Rule key for score expression. */
+    public static final String SCORE_EXPR_KEY = "score-expr";
+    /** Rule key for minimum data file MSE threshold. */
+    public static final String MIN_DATAFILE_MSE_KEY = "minDatafileMse";
+    /** Rule key for minimum delete file count threshold. */
+    public static final String MIN_DELETE_FILE_NUMBER_KEY = "minDeleteFileNumber";
+    /** Rule key for data file MSE score weight. */
+    public static final String DATAFILE_MSE_WEIGHT_KEY = "datafileMseWeight";
```
jerryshao: "data" and "file" are two words, better change to "dataFileXxx".

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
```diff
@@ -42,6 +48,48 @@ public static PolicyContent custom(
...
+  /**
+   * Creates an iceberg compaction policy content.
+   *
+   * @param minDatafileMse minimum threshold for custom-datafile_mse
+   * @param minDeleteFileNumber minimum threshold for custom-delete_file_number
+   * @param rewriteOptions rewrite options forwarded as job.options.*
+   * @return iceberg compaction policy content
+   */
+  public static PolicyContent icebergCompaction(
+      long minDatafileMse, long minDeleteFileNumber, Map<String, String> rewriteOptions) {
```
jerryshao: This method can call the below one to simplify the code.

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
```diff
@@ -128,4 +176,233 @@ public String toString() {
...
+    /** Rule key for minimum delete file count threshold. */
+    public static final String MIN_DELETE_FILE_NUMBER_KEY = "minDeleteFileNumber";
+    /** Rule key for data file MSE score weight. */
+    public static final String DATAFILE_MSE_WEIGHT_KEY = "datafileMseWeight";
+    /** Rule key for delete file number score weight. */
+    public static final String DELETE_FILE_NUMBER_WEIGHT_KEY = "deleteFileNumberWeight";
+    /** Metric name for data file MSE. */
+    public static final String DATAFILE_MSE_METRIC = "custom-datafile_mse";
+    /** Metric name for delete file number. */
+    public static final String DELETE_FILE_NUMBER_METRIC = "custom-delete_file_number";
```
jerryshao: Shall we unify the styles for the key, not fixing with "-", "_".

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
FANNG1: Agreed. I renamed the naming style from `datafileXxx` to `dataFileXxx` across policy content, DTO fields, converters and tests to follow word boundaries consistently. Addressed in commit 6212a7c50.

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
FANNG1: Good point. I unified the key style to kebab-case and switched the metrics to `custom-data-file-mse` and `custom-delete-file-number` consistently in policy content, expressions and tests. (No legacy compatibility mapping is kept.) Addressed in commits 6212a7c50, 56273431a and 95550c72f.

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
FANNG1: Done. The 3-arg `icebergCompaction(...)` overload now calls the 5-arg overload to remove duplication and keep defaults in one path. Addressed in commit 6212a7c50.

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
FANNG1: Done. I split `IcebergCompactionContent` into a dedicated file `api/src/main/java/org/apache/gravitino/policy/IcebergCompactionContent.java` and kept `PolicyContents` as factory utilities only. Addressed in commit 6212a7c50.

api/src/main/java/org/apache/gravitino/policy/PolicyContents.java
```diff
@@ -128,4 +176,232 @@ public String toString() {
...
+          datafileMseWeight == null ? DEFAULT_DATAFILE_MSE_WEIGHT : datafileMseWeight;
+      this.deleteFileNumberWeight =
+          deleteFileNumberWeight == null
+              ? DEFAULT_DELETE_FILE_NUMBER_WEIGHT
+              : deleteFileNumberWeight;
+      this.rewriteOptions =
+          rewriteOptions == null
+              ? Map.of()
+              : Collections.unmodifiableMap(new LinkedHashMap<>(rewriteOptions));
+    }
```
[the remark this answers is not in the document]
FANNG1: updated

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
```diff
@@ -69,4 +76,148 @@ public Map<String, String> properties() {
...
+    /**
+     * Returns rewrite options expanded to job.options.* during rule generation.
+     *
+     * @return rewrite options map
+     */
+    public Map<String, String> rewriteOptions() {
+      return rewriteOptions == null
+          ? Map.of()
+          : Collections.unmodifiableMap(new LinkedHashMap<>(rewriteOptions));
+    }
```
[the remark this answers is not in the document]
FANNG1: updated

FANNG1:
@jerryshao @mchades please take a look again, thx

api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  @Override
+  public void validate() throws IllegalArgumentException {
+    PolicyContent.super.validate();
+    Preconditions.checkArgument(minDataFileMse >= 0, "minDataFileMse must be >= 0");
+    Preconditions.checkArgument(minDeleteFileNumber >= 0, "minDeleteFileNumber must be >= 0");
+    Preconditions.checkArgument(dataFileMseWeight >= 0, "dataFileMseWeight must be >= 0");
+    Preconditions.checkArgument(deleteFileNumberWeight >= 0, "deleteFileNumberWeight must be >= 0");
+    Preconditions.checkArgument(maxPartitionNum > 0, "maxPartitionNum must be > 0");
```
mchades, line 219: Are all these fields required?

FANNG1: Not all fields are required from user input. `IcebergCompactionContent` accepts nullable inputs and applies defaults in the constructor (for example minDataFileMse/minDeleteFileNumber/weights/maxPartitionNum/rewriteOptions), then `validate()` only checks value ranges after defaulting. I added explicit comments to clarify this behavior in commit 0448c306c.

jerryshao (commented):
Good PR overall — the typed policy content, strong validation, and documentation are all well done. There is one critical cross-PR issue that must be fixed before merge, plus a few design/quality items.

api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
```diff
@@ -0,0 +1,279 @@
...
+  /** Rule key for minimum delete file count threshold. */
+  public static final String MIN_DELETE_FILE_NUMBER_KEY = "minDeleteFileNumber";
+  /** Rule key for data file MSE score weight. */
+  public static final String DATA_FILE_MSE_WEIGHT_KEY = "dataFileMseWeight";
+  /** Rule key for delete file number score weight. */
+  public static final String DELETE_FILE_NUMBER_WEIGHT_KEY = "deleteFileNumberWeight";
+  /** Rule key for max partition number selected for compaction. */
+  public static final String MAX_PARTITION_NUM_KEY = "max-partition-num";
+  /** Metric name for data file MSE. */
+  public static final String DATA_FILE_MSE_METRIC = "custom-data-file-mse";
```
jerryshao, line 60: **Critical: metric name mismatch with the stats job (PR #10106)**

`DATA_FILE_MSE_METRIC = "custom-data-file-mse"` but PR #10106 (`IcebergUpdateStatsAndMetricsJob`) produces the metric as `"custom-datafile_mse"` (no hyphen before `mse`, and uses underscore). These two names must align or the trigger expression will never fire.

Similarly, `DELETE_FILE_NUMBER_METRIC = "custom-delete-file-number"` does not match any metric produced by the stats job — the stats job emits `"custom-position_delete_files"` and `"custom-equality_delete_files"` as separate metrics, not an aggregated `custom-delete-file-number`. If the intent is "total delete file count", the stats job needs to produce this aggregated metric (or the expression needs to reference the individual metrics), otherwise this trigger will never evaluate to true.

Please coordinate with PR #10106 to ensure the metric names are consistent end-to-end.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/strategy/GravitinoStrategy.java
```diff
@@ -43,7 +43,7 @@ public class GravitinoStrategy implements PartitionStrategy {
   /** Rule key for the partition table score aggregation mode. */
   public static final String PARTITION_TABLE_SCORE_MODE = "partition_table_score_mode";
   /** Rule key for the maximum number of partitions selected for execution. */
-  public static final String MAX_PARTITION_NUM = "max_partition_num";
+  public static final String MAX_PARTITION_NUM = "max-partition-num";
```
jerryshao, line 46: **Breaking change**: `MAX_PARTITION_NUM` was renamed from `"max_partition_num"` (underscore) to `"max-partition-num"` (hyphen). `GravitinoStrategy` is used for **all** `GravitinoPolicy`-based strategies, including existing custom policies. Any policy stored with `max_partition_num` in its rules will silently stop applying the partition limit and fall back to `DEFAULT_MAX_PARTITION_NUM = 100`. Please either preserve backward compatibility (read both keys, prefer the new one) or document this as an intentional breaking change requiring migration.

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
```diff
@@ -69,4 +78,163 @@ public Map<String, String> properties() {
...
+
+    @Override
+    public Set<MetadataObject.Type> supportedObjectTypes() {
+      return ImmutableSet.of(
+          MetadataObject.Type.CATALOG, MetadataObject.Type.SCHEMA, MetadataObject.Type.TABLE);
+    }
+
+    @Override
+    public Map<String, String> properties() {
```
jerryshao, line 177: `properties()` and `rules()` each construct a full `IcebergCompactionContent` via `PolicyContents.icebergCompaction(...)` just to call `.properties()` or `.rules()` on it. If both are called together (e.g. during serialization + validation), two intermediate objects are allocated. Consider caching or calling the API domain object once:

```java
private IcebergCompactionContent toDomain() {
  return (IcebergCompactionContent) PolicyContents.icebergCompaction(
      minDataFileMse(), minDeleteFileNumber(), dataFileMseWeight(),
      deleteFileNumberWeight(), maxPartitionNum(), rewriteOptions());
}

@Override public Map<String, String> properties() { return toDomain().properties(); }
@Override public Map<String, Object> rules() { return toDomain().rules(); }
```

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
```diff
@@ -69,4 +78,163 @@ public Map<String, String> properties() {
...
+    @JsonProperty("deleteFileNumberWeight")
+    private Long deleteFileNumberWeight;
+
+    @JsonProperty("maxPartitionNum")
+    private Long maxPartitionNum;
+
+    @JsonProperty("rewriteOptions")
+    private Map<String, String> rewriteOptions;
+
+    private static final Pattern OPTION_KEY_PATTERN = Pattern.compile("[A-Za-z0-9._-]+");
```
jerryshao: `OPTION_KEY_PATTERN` is duplicated between `IcebergCompactionContent` (line 82 in the API module) and `IcebergCompactionContentDTO` here. The DTO validation also duplicates the same range-check and `rewriteOptions` iteration logic from `IcebergCompactionContent.validate()`. The DTO should delegate to the domain object (e.g., `toDomain().validate()`) rather than re-implementing the same checks.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -334,16 +333,19 @@ public Response listMetadataObjectsForPolicy(
   }
 
   static PolicyDTO toDTO(PolicyEntity policy, Optional<Boolean> inherited) {
+    String policyType =
+        policy.policyType() == Policy.BuiltInType.CUSTOM
```
jerryshao: The `policyType` output is now case-inconsistent: CUSTOM returns `"custom"` (lowercase via `policyType()`) but ICEBERG_COMPACTION returns `"ICEBERG_COMPACTION"` (uppercase via `name()`). API consumers who switch policy between types will see different casing. Consider normalizing to a consistent convention (all lowercase or all uppercase), or at least add a comment explaining the intentional difference and update the REST API documentation.

api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
```diff
@@ -0,0 +1,279 @@
...
+  /** Metric name for delete file number. */
+  public static final String DELETE_FILE_NUMBER_METRIC = "custom-delete-file-number";
+  /** Default minimum threshold for data file MSE metric. */
+  public static final long DEFAULT_MIN_DATA_FILE_MSE = 405323966463344L;
```
jerryshao, line 64: This magic number should have an inline comment explaining its derivation. The docs explain it but the source code should be self-contained:

```java
// = (128 MiB * 0.15)^2 = (134217728 * 0.15)^2
public static final long DEFAULT_MIN_DATA_FILE_MSE = 405323966463344L;
```

api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
FANNG1: Thanks for catching this. After rebasing to commit f7b9c3905f84e59291c8cb8635821b609da00139, the built-in stats job now emits `custom-data-file-mse` and aggregated `custom-delete-file-number` (in addition to position/equality delete metrics), so the policy expressions are aligned end-to-end.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/strategy/GravitinoStrategy.java
FANNG1: Good point. This change is intentional in this PR: we unified rule key style to kebab-case and do not keep backward-compatibility mapping for the old `max_partition_num` key. This follows the current design decision for this branch (no legacy compatibility). We can add a migration note in docs/release notes if needed.

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
FANNG1: Agreed, fixed. I introduced a shared `toDomainContent()` in `IcebergCompactionContentDTO` and reuse it in both `properties()` and `rules()` (commit 335a16c83).

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
FANNG1: Agreed, fixed. DTO validation now delegates to domain validation via `toDomainContent().validate()` to avoid duplicated option/range checks (commit 335a16c83).

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
FANNG1: Thanks, this is intentional for compatibility: `custom` is kept lowercase to avoid a breaking REST change for existing clients, while built-ins use enum names. I also added an inline comment in `PolicyOperations.toDTO` to make this explicit (commit 335a16c83).

api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
FANNG1: Agreed, fixed. Added inline derivation comment for `DEFAULT_MIN_DATA_FILE_MSE` (`(128 MiB * 0.15)^2`) in `IcebergCompactionContent` (commit 335a16c83).

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
```diff
@@ -69,4 +75,133 @@ public Map<String, String> properties() {
...
+  @EqualsAndHashCode
+  @ToString
+  @Builder(setterPrefix = "with")
+  @AllArgsConstructor(access = lombok.AccessLevel.PRIVATE)
+  class IcebergCompactionContentDTO implements PolicyContentDTO {
+
+    @JsonProperty("minDataFileMse")
+    @JsonAlias("minDatafileMse")
+    private Long minDataFileMse;
+
```
jerryshao, line 86: **Add a comment explaining these backward-compat aliases.**

These `@JsonAlias` entries were added when the field name was renamed from `datafileXxx` to `dataFileXxx`. Without a comment, a future reviewer who spots seemingly redundant aliases may remove them, breaking deserialization of payloads from older clients.

Suggested comment:
```java
// Backward-compat alias: field was renamed from datafileMse → dataFileMse (kebab-case unification)
@JsonAlias("minDatafileMse")
```

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
```diff
@@ -69,4 +75,133 @@ public Map<String, String> properties() {
...
+    public Map<String, String> rewriteOptions() {
+      return rewriteOptions == null
+          ? IcebergCompactionContent.DEFAULT_REWRITE_OPTIONS
+          : Collections.unmodifiableMap(new LinkedHashMap<>(rewriteOptions));
+    }
+
+    @Override
+    public Set<MetadataObject.Type> supportedObjectTypes() {
+      return ImmutableSet.of(
+          MetadataObject.Type.CATALOG, MetadataObject.Type.SCHEMA, MetadataObject.Type.TABLE);
```
jerryshao: `supportedObjectTypes()` duplicates `IcebergCompactionContent.SUPPORTED_OBJECT_TYPES` directly. The delegation pattern is already established via `toDomainContent()` for `properties()`, `rules()`, and `validate()` — `supportedObjectTypes()` should follow the same pattern:

```java
@Override
public Set<MetadataObject.Type> supportedObjectTypes() {
  return toDomainContent().supportedObjectTypes();
}
```

This ensures a single source of truth: if supported types ever expand (e.g. adding `FILESET`), only `IcebergCompactionContent.SUPPORTED_OBJECT_TYPES` needs updating.

server/src/main/java/org/apache/gravitino/server/web/rest/PolicyOperations.java
```diff
@@ -334,17 +334,32 @@ public Response listMetadataObjectsForPolicy(
   }
 
   static PolicyDTO toDTO(PolicyEntity policy, Optional<Boolean> inherited) {
+    // Keep "custom" lowercase for backward compatibility, and return enum name for built-ins.
+    String policyType =
```
FANNG1: @mchades please help to review this part

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
FANNG1: Good point. I updated `supportedObjectTypes()` to delegate to `toDomainContent().supportedObjectTypes()` so the domain content remains the single source of truth. Addressed in commit 8ec86da3e.

common/src/main/java/org/apache/gravitino/dto/policy/PolicyContentDTO.java
FANNG1: Thanks. We decided to drop backward compatibility for legacy datafile-style field names in this PR, so I removed the aliases entirely and now only keep the canonical fields (`minDataFileMse`, `dataFileMseWeight`). Updated in commit 4dbef072d.

jerryshao:
The CI is failed, please fix the CI.

FANNG1:
@jerryshao Thanks for the review.

For the `properties()` vs `rules()` question: yes, this is intentional.
- `properties()` keeps stable strategy identity/config selection fields (`strategy.type`, `job.template-name`).
- `rules()` carries runtime evaluation knobs (thresholds, weights, trigger/score expressions, and `job.options.*`).

I also added an inline code comment in `IcebergDataCompactionContent.properties()` to make this boundary explicit. The fix is in local commit `3b9d078d1` and I’ll push it to the PR branch after confirmation.

FANNG1:
@jerryshao any other comments?

jerryshao:
I don't have further comments, let's merge it ASAP.

PR #10190, opened 2026-03-04
[#10125] fix(server): prevent NPE when statistics update request body is null

### LINKED ISSUE #10125: [Improvement] postention NPE in StatisticOperations.java

The problem is that the statistics update endpoint (PUT /metalakes/{metalake}/objects/{type}/{fullName}/statistics) can return a 500 Internal Server Error when the request body is null, even though this is a client-side issue. In StatisticOperations.updateStatistics(...), the method catches exceptions but then dereferences the request again inside the catch block (request.getUpdates().keySet()). If the request is null, this triggers a second NullPointerException during error handling, and results in a server error.

See /server/src/main/java/org/apache/gravitino/server/web/rest/StatisticOperations.java

### How should we improve?

A possible solution is to change StatisticOperations.updateStatistics. Add an explicit null check before request.validate() and throw an IllegalArgumentException when the request is null, then make the catch block null-safe.

Here a unit test to help:
```
@Test
public void testUpdateTableStatisticsWithNullRequestBody() {
  when(tableDispatcher.tableExists(any())).thenReturn(true);

  MetadataObject tableObject =
      MetadataObjects.parse(
          String.format("%s.%s.%s", catalog, schema, table), MetadataObject.Type.TABLE);

  Response resp =
      target(
              "/metalakes/"
                  + metalake
                  + "/objects/"
                  + tableObject.type()
                  + "/"
                  + tableObject.fullName()
                  + "/statistics")
          .put(entity("null", MediaType.APPLICATION_JSON_TYPE));

  Assertions.assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), resp.getStatus());
}
```

Place in server/src/test/java/org/apache/gravitino/server/web/rest/TestStatisticOperations.java

geniusjun:
Summary
- Added an explicit null check for `StatisticsUpdateRequest` in `StatisticOperations.updateStatistics(...)`
- Returned a proper client error (`IllegalArgumentException` -> `400 BAD_REQUEST`) for null request bodies
- Made error handling null-safe to avoid a secondary NPE in the catch path
- Refactored request validation/statistic-name extraction into helper methods for readability
- Added a regression test for `PUT /statistics` with JSON `null` body

Verification
- Ran:
  `export JAVA_HOME=$(/usr/libexec/java_home -v 17); ./gradlew :server:test --tests org.apache.gravitino.server.web.rest.TestStatisticOperations -PskipITs`
- `TestStatisticOperations` passed locally

Fix: #10125

Feedback and suggestions are very welcome. Thank you!


jerryshao:
CC @roryqi 

I don't think this issue will happen in the real environment. The malformed or null json request will be handled by Jersey/Jackson framework. I don't think it will defer to here to give a `null` request to the REST operation.

server/src/main/java/org/apache/gravitino/server/web/rest/StatisticOperations.java
```diff
@@ -476,6 +479,21 @@ static PartitionRange.BoundType getFromBoundType(boolean inclusive) {
...
 
+  private static void validateStatisticsUpdateRequest(StatisticsUpdateRequest request) {
+    if (request == null) {
+      throw new IllegalArgumentException(NULL_STATS_UPDATE_REQUEST_BODY_ERROR);
+    }
+    request.validate();
+  }
+
+  private static String getUpdatedStatisticNames(StatisticsUpdateRequest request) {
+    if (request == null || request.getUpdates() == null) {
```
jerryshao: `request.getUpdates() == null` is already checked here `request.validate()`. You just removed this line.

geniusjun:
Thank you for the clarification. I understand that in real server environments malformed or null JSON requests are handled by Jersey/Jackson before reaching this REST operation.
I have removed only the part you pointed out.

justinmclean:
Malformed JSON is usually rejected by Jersey/Jackson before methods are invoked. However, a valid JSON literal null may still deserialize to a Java null parameter and reach the REST method. See https://github.com/apache/gravitino/pull/10187 for a similar issue with details where this is not handled by Jersey/Jackson and can reach our REST operations.

server/src/main/java/org/apache/gravitino/server/web/rest/StatisticOperations.java
```diff
@@ -160,7 +163,7 @@ public Response updateStatistics(
       return Utils.doAs(
           httpRequest,
           () -> {
-            request.validate();
+            validateStatisticsUpdateRequest(request);
```
justinmclean: We need to leave a request.validate() as it.

geniusjun:
@justinmclean Thanks for the clarification. I updated the code accordingly: malformed JSON is still handled by Jersey/Jackson, but valid JSON literal null is now explicitly handled in the REST method. I also kept request.validate() directly in updateStatistics as requested.

jerryshao (commented):
**Code Review**

The fix is correct — validating `request == null` inside the lambda (which throws `IllegalArgumentException`, correctly handled as `400 BAD_REQUEST`) is the right approach. The test confirms the behavior.

### Issues

1. **Class member ordering violation**: `NULL_STATS_UPDATE_REQUEST_BODY_ERROR` (a `static final` constant) is placed **before** `LOG` in the class body. Per the project's coding standards (`CLAUDE.md`), `static` constants should be listed with `LOG` conventionally first (as the most-referenced constant). Please move `NULL_STATS_UPDATE_REQUEST_BODY_ERROR` to after `LOG`, or reorganize so all static constants appear in a consistent order.

2. **Possible dead code**: `getUpdatedStatisticNames(StatisticsUpdateRequest request)` is added as a private helper method, but it does not appear to be called anywhere in the diff or the existing code. If this method is unused, please remove it. If it was intended to replace an inline expression, please wire it up.

Please address both issues before merging.

geniusjun:
@jerryshao 
Thank you so much for the detailed explanation. 
I will have to check the standard documents provided by the team during the open source activities. I really learn a lot!

I modified it, if you have any additional advice, let me know!

justinmclean (approved):
Thanks for working on this

Issue #10181, opened 2026-03-04
[Improvement] Spark supports to pass the required write privileges


roryqi:
### What would you like to be improved?

Spark supports to pass the required write privileges

### How should we improve?

Spark supports to pass the required write privileges

PR #10194, opened 2026-03-04
[#10181] feat(spark-connector): Support TableWritePrivilege for Spark 3.5+ authorization

roryqi:

This PR adds support for Spark 3.5's TableWritePrivilege interface to enforce MODIFY_TABLE privilege when performing write operations (INSERT, UPDATE, DELETE). Changes:
- Add loadTable(Identifier, Set<TableWritePrivilege>) method to Spark 3.5+ catalogs (Hive, Iceberg, JDBC, Paimon)
- Add loadTableForWriting and loadGravitinoTableForWriting helper methods in BaseCatalog
- Upgrade Spark 3.5 version from 3.5.1 to 3.5.3
- Add integration test to verify authorization behavior In Spark 3.5+, when a user without MODIFY_TABLE privilege attempts to INSERT, a ForbiddenException will be thrown. In Spark 3.3/3.4, the same operation succeeds because those versions don't support TableWritePrivilege.

Fix: #10181

No

Added tests.

### DIFF
--- a/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
+++ b/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
@@ -514,7 +514,8 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
     throw new NoSuchFunctionException(ident);
   }
 
-  protected Table loadTableForWriting(Identifier ident) throws NoSuchTableException {
+  protected Table loadTableForWriting(Identifier ident)
+      throws NoSuchTableException, ForbiddenException {
     org.apache.gravitino.rel.Table gravitinoTable = loadGravitinoTableForWriting(ident);
     org.apache.spark.sql.connector.catalog.Table sparkTable = loadSparkTable(ident);
     // Will create a catalog specific table


spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
```diff
@@ -498,6 +514,20 @@ public UnboundFunction loadFunction(Identifier ident) throws NoSuchFunctionExcep
...
+    return createSparkTable(
+        ident,
+        gravitinoTable,
+        sparkTable,
+        sparkCatalog,
+        propertiesConverter,
+        sparkTransformConverter,
+        sparkTypeConverter);
+  }
+
```
diqiu50, line 531: Is `ForbiddenException` will be throw

roryqi: It will throw `ForbiddenException`.

diqiu50: Do we need to add this exception type to the function signature?

roryqi: Yes, we should.

PR #10196, opened 2026-03-04
Fix #10123: Add null check for MetalakeCreateRequest

### LINKED ISSUE #10123: [Improvement] Possible NPE in createMetalake REST handler

nikitanagar08:
## Summary
- Fixed NPE in `MetalakeOperations.createMetalake()` when request body is null
- Added null check that returns BAD_REQUEST with appropriate error message
- Added test case `testCreateMetalakeWithNullRequest()` to verify the fix

## Fixes
Fixes #10123

## Test Plan
- Added `testCreateMetalakeWithNullRequest()` test in `TestMetalakeOperations.java`
- Tests null request body handling with BAD_REQUEST response verification

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/MetalakeOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/MetalakeOperations.java
@@ -117,6 +117,9 @@ public class MetalakeOperations {
           "Only service admins can create metalakes, current user can't create the metalake,"
               + "  you should configure it in the server configuration first")
   public Response createMetalake(MetalakeCreateRequest request) {
+    if (request == null) {
+      return Utils.illegalArguments("Metalake create request cannot be null");
+    }
     LOG.info("Received create metalake request for {}", request.getName());
     try {
       return Utils.doAs(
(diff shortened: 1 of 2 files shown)


nikitanagar08:
Closing this PR as the fix for this issue has already been merged into main branch. The null check for MetalakeCreateRequest has been implemented by the maintainers.

Issue #10124, opened 2026-03-03
[Improvement] Error handling in StatisticOperations is fragile / possible NPE


justinmclean:
### What would you like to be improved?

The current error-handling flow in StatisticOperations is fragile when request payloads are null or invalid. In both partition and table statistics update paths, the catch block dereferences request.getUpdates() to build error context, which can throw a new NullPointerException and mask the original validation failure. This leads to incorrect 500 responses instead of consistent client-facing validation errors (400), and makes debugging harder because the original cause is lost.

### How should we improve?

The fix is to make catch-path error formatting null-safe and preserve the original exception. Build the partitions/stat names string defensively (for example, using an empty fallback when request or request.getUpdates() is null), then pass the original exception to ExceptionHandlers. 

Here's a unit test to help:
```
@Test
public void testUpdatePartitionStatisticsShouldNotThrowWhenRequestUpdatesIsNull() {
  StatisticOperations operations = new StatisticOperations(manager);
  HttpServletRequest mockRequest = mock(HttpServletRequest.class);
  try {
    FieldUtils.writeField(operations, "httpRequest", mockRequest, true);
  } catch (IllegalAccessException e) {
    throw new RuntimeException(e);
  }

  PartitionStatisticsUpdateRequest req = new PartitionStatisticsUpdateRequest();
  Response resp =
      Assertions.assertDoesNotThrow(
          () ->
              operations.updatePartitionStatistics(
                  metalake, "table", String.format("%s.%s.%s", catalog, schema, table), req));
  Assertions.assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), resp.getStatus());
}

```

Place in server/src/test/java/org/apache/gravitino/server/web/rest/TestStatisticOperations.java

kdyann:
Hi @justinmclean ,

I would like to work on this issue if it is still available. 
Could you please assign it to me?

Thanks!

PR #10200, opened 2026-03-04
[#10124] refactor: Make StatisticOperations error handling null-safe

kdyann:

This PR improves the error handling in `StatisticOperations` to avoid a possible
`NullPointerException` in the catch path.

Previously, in both table and partition statistics update methods,
the catch block dereferenced `request.getUpdates()` to build an error
context string. If the request payload or its `updates` field was null,
this could trigger a second NPE during error handling, masking the original
validation failure and resulting in an incorrect `500 Internal Server Error`.

This PR makes the catch-path error formatting null-safe by:

- Introducing helper methods (`getStatisticNames`, `getPartitionNames`)
  that defensively handle null `request` or null `updates`.
- Preserving the original exception and passing it to `ExceptionHandlers`
  without re-triggering another NPE.

The previous implementation was fragile when request payloads were null
or invalid. A secondary NPE in the catch block could mask the original
validation error and produce incorrect server-side error responses.

This change ensures:

- The original validation exception is preserved.
- Error formatting in the catch path is null-safe.
- The API consistently returns client-facing validation errors (400)
  instead of incorrectly returning 500.

Fix: #10124

No.

This change only improves internal error handling logic.
There are no API changes, configuration changes, or behavioral changes
other than preventing incorrect 500 responses in invalid request scenarios.

- Added a regression test in `TestStatisticOperations`.
- The test explicitly sends a payload with `"updates": null`
  and verifies that the API returns `400 Bad Request`.
- Confirmed that no secondary `NullPointerException`
  is thrown during error handling.

I aligned the regression test with the existing Jersey `target(...)`
style in this test class to match its convention.
Please let me know if a white-box unit test style would be preferred instead.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/StatisticOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/StatisticOperations.java
@@ -188,7 +188,7 @@ public class StatisticOperations {
           });
     } catch (Exception e) {
       return ExceptionHandlers.handleStatisticException(
-          OperationType.UPDATE, StringUtils.join(request.getUpdates().keySet(), ","), fullName, e);
+          OperationType.UPDATE, getStatisticNames(request), fullName, e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


kdyann:
Thanks for reviewing and merging this PR!

PR #10212, opened 2026-03-04
[#10168] Fail fast when file: URI source is missing in JobManager.fetchFileFromUri

### LINKED ISSUE #10168: [Improvement] Fail fast when file: URI source is missing in JobManager.fetchFileFromUri

JobManager.fetchFileFromUri currently reports success for file: URIs by creating a symlink in the staging directory, even when the source path does not exist. This creates a dangling symlink and returns a seemingly valid staged path, causing failures later when the file is consumed instead of failing at fetch time.

### How should we improve?

For the file: branch, validate the source path before creating the symlink (at minimum, check it exists; optionally verify it is a regular file). If validation fails, throw an exception immediately with a clear message. This ensures fetch behavior is accurate and prevents runtime failures caused by invalid staged artifacts.

Here's a unit test to help:
```
  @Test
  public void testFetchFileFromUriWithMissingLocalFileShouldFail() throws IOException {
    File stagingDir = new File(testStagingDir);
    Assertions.assertTrue(stagingDir.mkdirs() || stagingDir.exists());

    Path missingFilePath =
        Path.of(System.getProperty("java.io.tmpdir"), "missing-job-file-" + UUID.randomUUID());
    String uri = missingFilePath.toUri().toString();

    Assertions.assertThrows(
        RuntimeException.class, () -> JobManager.fetchFileFromUri(uri, stagingDir, 1000));
  }
```

rameshreddy-adutla:

Validate that the source path exists before creating a symlink in `fetchFileFromUri`. Previously, a dangling symlink was silently created when the source file did not exist, causing failures downstream instead of at fetch time.

`JobManager.fetchFileFromUri` currently reports success for `file:` URIs by creating a symlink in the staging directory, even when the source path does not exist. This creates a dangling symlink and returns a seemingly valid staged path, causing failures later when the file is consumed instead of failing at fetch time.

No. This is an internal defensive check — the external API is unchanged. Jobs that previously failed late with a confusing error will now fail early with a clear message.

Added unit test `testFetchFileFromUriWithMissingLocalFileShouldFail` (as suggested in #10168) that:
1. Creates a staging directory
2. Constructs a `file:` URI pointing to a non-existent path
3. Asserts that `fetchFileFromUri` throws `RuntimeException`

Closes #10168

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/JobManager.java
+++ b/core/src/main/java/org/apache/gravitino/job/JobManager.java
@@ -812,8 +812,7 @@ public class JobManager implements JobOperationDispatcher {
           java.nio.file.Path sourcePath = new File(fileUri.getPath()).toPath();
           if (!Files.exists(sourcePath)) {
             throw new IOException(
-                String.format(
-                    "Source file does not exist: %s", sourcePath.toAbsolutePath()));
+                String.format("Source file does not exist: %s", sourcePath.toAbsolutePath()));
           }
           Files.createSymbolicLink(destFile.toPath(), sourcePath);
           break;


justinmclean:
You'll need to run `./gradlew :core:spotlessApply` and push your changes to correct the CI issue.

rameshreddy-adutla:
Thanks @justinmclean! I've run `./gradlew :core:spotlessApply` and pushed the formatting fix. 🙏

PR #10213, opened 2026-03-04
[#10167] Fail topic creation when metadata store persistence fails

### LINKED ISSUE #10167: [Improvement] Fail topic creation when metadata store persistence fails in TopicOperationDispatcher

TopicOperationDispatcher#createTopic creates the topic in the catalog, then persists TopicEntity via store.put(...). If store.put throws, the exception is caught and only logged; the method still returns a successful Topic result. This can report success while Gravitino metadata was not written, leaving the catalog and metadata store inconsistent and causing later operations to behave incorrectly.

### How should we improve?

Treat store.put failure as an operation failure: rethrow as a runtime/domain exception instead of returning success. 

Here's a unit test to help:
```
@Test
public void testCreateTopicShouldFailWhenStorePutFails() throws IOException {
  Namespace topicNs = Namespace.of(metalake, catalog, "schema171");
  Map<String, String> props = ImmutableMap.of("k1", "v1", "k2", "v2");
  schemaOperationDispatcher.createSchema(NameIdentifier.of(topicNs.levels()), "comment", props);

  reset(entityStore);
  doThrow(new IOException("store put failed")).when(entityStore).put(any(), eq(true));

  NameIdentifier topicIdent = NameIdentifier.of(topicNs, "topic71");
  Assertions.assertThrows(
      RuntimeException.class,
      () -> topicOperationDispatcher.createTopic(topicIdent, "comment", null, props));
}

```

rameshreddy-adutla:

`TopicOperationDispatcher#internalCreateTopic` previously caught `store.put` failures and returned a successful `Topic` result with only a log warning. This left the catalog and metadata store inconsistent — the topic exists in the catalog but Gravitino's metadata was not persisted, causing later operations to behave incorrectly.

The fix rethrows the `store.put` exception as a `RuntimeException` instead of swallowing it.

When `store.put` fails, the method currently returns a successful result, silently dropping the persistence error. This creates an inconsistent state where:
- The topic exists in the underlying catalog
- Gravitino's metadata store has no record of it
- Subsequent operations (load, alter, drop) will fail with confusing errors

No. The external API is unchanged. Topic creation that previously appeared to succeed (but was actually partially broken) will now correctly report failure.

Added unit test `testCreateTopicShouldFailWhenStorePutFails` (as suggested in #10167) that:
1. Creates a schema
2. Mocks `entityStore.put()` to throw `IOException`
3. Asserts that `createTopic` throws `RuntimeException`

Closes #10167

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/TopicOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/TopicOperationDispatcher.java
@@ -395,10 +395,8 @@ public class TopicOperationDispatcher extends OperationDispatcher implements Top
       store.put(topicEntity, true /* overwrite */);
     } catch (Exception e) {
       LOG.error(OperationDispatcher.FormattedErrorMessages.STORE_OP_FAILURE, "put", ident, e);
-      return EntityCombinedTopic.of(topic)
-          .withHiddenProperties(
-              getHiddenPropertyNames(
-                  catalogIdent, HasPropertyMetadata::topicPropertiesMetadata, topic.properties()));
+      throw new RuntimeException(
+          "Failed to persist topic entity for " + ident + " in the store", e);
     }
 
     return EntityCombinedTopic.of(topic, topicEntity)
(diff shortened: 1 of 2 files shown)


rameshreddy-adutla:
ack

jerryshao:
@mchades please help to review this PR.

mchades:
It's unnecessary change, see the comment https://github.com/apache/gravitino/issues/10167#issuecomment-4021664135, still thanks for your work @rameshreddy-adutla

PR #10214, opened 2026-03-04
[#10164] Fix NPE in SchemaOperations#createSchema when request is null

### LINKED ISSUE #10164: [Improvement] Possible NPE in  SchemaOperations#createSchema

rameshreddy-adutla:

Make `SchemaOperations#createSchema` null-safe when the request body is null or missing.

`SchemaOperations#createSchema` dereferences `request.getName()` in two places outside null-safe paths:
1. **Line 125** (LOG statement): Before the try/catch — immediate NPE that escapes error handling entirely
2. **Line 142** (catch block): When handling other exceptions — secondary NPE masks the original error

Both result in an uncaught `NullPointerException` and inconsistent 500 behavior.

No. API contracts are unchanged. The endpoint now returns a proper error response instead of an unhandled NPE when the request body is null.

Added unit test `testCreateSchemaWithNullRequestShouldNotThrow` (as suggested in #10164) that:
1. POSTs `"null"` as the request body
2. Asserts 500 response (not an uncaught NPE)
3. Asserts the error type is NOT `NullPointerException`

Closes #10164

### DIFF
--- a/server/src/test/java/org/apache/gravitino/server/web/rest/TestSchemaOperations.java
+++ b/server/src/test/java/org/apache/gravitino/server/web/rest/TestSchemaOperations.java
@@ -496,14 +496,11 @@ public class TestSchemaOperations extends BaseOperationsTest {
         target("/metalakes/" + metalake + "/catalogs/" + catalog + "/schemas")
             .request(MediaType.APPLICATION_JSON_TYPE)
             .accept("application/vnd.gravitino.v1+json")
-            .post(
-                javax.ws.rs.client.Entity.entity(
-                    "null", MediaType.APPLICATION_JSON_TYPE));
+            .post(javax.ws.rs.client.Entity.entity("null", MediaType.APPLICATION_JSON_TYPE));
 
     Assertions.assertEquals(
         Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), resp.getStatus());
     ErrorResponse errorResp = resp.readEntity(ErrorResponse.class);
-    Assertions.assertNotEquals(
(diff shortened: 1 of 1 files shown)


justinmclean:
We use spotless to ensure that code is correctly formatted. You'll need to run `./gradlew :server:spotlessApply` and push your changes to resolve the CI issues above.

rameshreddy-adutla:
Thanks @justinmclean! I've run `./gradlew :server:spotlessApply` and pushed the formatting fix. 🙏

rameshreddy-adutla:
Thanks for the review @jerryshao. I see that PR #10187 has been closed, and PR #10183 (the comprehensive fix) is still open. If #10183 covers this same fix as part of its broader scope, I'm happy to close this PR in favour of that. Let me know if you'd prefer this stays open as a focused fix or if #10183 is the preferred path forward.

Issue #10127, opened 2026-03-03
[Improvement] Potential IndexOutOfBoundsException in POConverters.java


justinmclean:
### What would you like to be improved?

In POConverters.java POConverters.fromModelVersionPO(...) currently assumes modelVersionPOs is non-empty and reads modelVersionPOs.get(0). When the list is empty, it throws an IndexOutOfBoundsException from the collection internals.

### How should we improve?

Add explicit input validation at the start of the method and fail fast with a clear exception (IllegalArgumentException).

Here's a unit test to help:
```
@Test
public void testFromModelVersionPOWithEmptyModelVersionPOs() {
  NameIdentifier modelIdent = NameIdentifierUtil.ofModel("m", "c", "s", "model1");

  assertThrows(
      IllegalArgumentException.class,
      () ->
          POConverters.fromModelVersionPO(
              modelIdent, Collections.emptyList(), Collections.emptyList()));
}
```

add to core/src/test/java/org/apache/gravitino/storage/relational/utils/TestPOConverters.java

ankursha1:
Hi @justinmclean , Can you assign this issue to me? Thanks.

ankursha1:
Hi @justinmclean I am getting below error while trying to push a PR.

```
remote: Permission to apache/gravitino.git denied to ankursha1.
fatal: unable to access 'https://github.com/apache/gravitino.git/': The requested URL returned error: 403
```

Could you please help. Thanks.

justinmclean:
Hi @ankursha1, that’s expected. The 403 happens because only committers have permission to push branches to apache/gravitino. Please use the standard fork + PR workflow:
 1. Fork the repo to your account
 2. Push your branch to your fork
 3. Create a Pull Request from your fork to apache/gravitino.

ankursha1:
Thank you @justinmclean . I have submitted the PR for review.

PR #10256, opened 2026-03-05
[#10127] Null check precondition for modelVersionPOs

ankursha1:
## Summary

Fix: #([10127)](https://github.com/apache/gravitino/issues/10127))

POConverters.java POConverters.fromModelVersionPO(...) currently assumes modelVersionPOs is non-empty and reads modelVersionPOs.get(0). The fix validates the modelVersionPOs object as precondition and throws IllegalArgumentException if null or empty.

## Changes

- Added Precondition validation to check null or empty object for modelVersionPOs

## Validation
- Added a Junit test to verify IllegalArgumentException thrown when null


ankursha1:
> **Code Review**
> 
> The intent is correct — fail fast when `modelVersionPOs` is null or empty rather than getting an `IndexOutOfBoundsException` on `.get(0)`.
> 
> ### Issues
> 1. **Prefer `isEmpty()`**: Use `!modelVersionPOs.isEmpty()` instead of `modelVersionPOs.size() > 0`. This is the idiomatic Java collection check and is recommended by the Google Java Style guide.
>    ```java
>    // Before
>    modelVersionPOs != null && modelVersionPOs.size() > 0
>    // After
>    modelVersionPOs != null && !modelVersionPOs.isEmpty()
>    ```
> 2. **Missing null test case**: The test only exercises `Collections.emptyList()`. Please also add a test that passes `null` explicitly to cover the null branch of the precondition.
> 3. **No conflict with #10292** — both modify `POConverters.java` but target different methods, so a clean merge should be possible.
> 
> Please address items 1 and 2 before merging.

@jerryshao Thanks for review. I have pushed the change to address the review comment. Please have a look.

jerryshao:
After reviewing the code, I don't think the real code path will hit this issue.

All 3 callers of `fromModelVersionPO` in `ModelVersionMetaService.java` are already protected:

1. **Line 106**: Called inside `Collectors.groupingBy(...)` — by definition, each group `m` always has ≥1 element. It is impossible for `m` to be empty here.

2. **Line 155** (`getModelVersionByIdentifier`): There is an explicit `if (modelVersionPOs.isEmpty())` check at line 135 that throws `NoSuchEntityException` before ever reaching `fromModelVersionPO`.

3. **Line 329** (`updateModelVersion`): Same pattern — explicit `if (oldModelVersionPOs.isEmpty())` check at line 308 throws `NoSuchEntityException` before reaching `fromModelVersionPO`.

The only way to trigger the `IndexOutOfBoundsException` would be to call `fromModelVersionPO` directly with an empty list, which nothing in the codebase does. The `Preconditions` check added by this PR would effectively be dead code — it will never fire in practice.

Given this, the value of this fix is questionable. It adds noise without addressing a real failure path.

jerryshao:
I'm going to close this PR and issue. Please feel free to reopen them if you find this issue in the real environment. Thanks @ankursha1 for your contribution.

Issue #10176, opened 2026-03-04
[Improvement] Fix built-in job template update to use existing entity identifier


justinmclean:
### What would you like to be improved?

BuiltInJobTemplateEventListener.updateBuiltInJobTemplate(...) builds the update NameIdentifier from newTemplate.name() instead of the existing persisted entity identity. When names differ (for example, rename/name drift), reconciliation enters update flow but entityStore.update(...) is executed with the wrong key, which can fail with not-found or update the wrong record.

### How should we improve?

Use the existing entity identifier for lock and update target, not the incoming template name. In practice, derive NameIdentifier from existing (or existing.name() under the same metalake) and keep payload fields updated from newTemplate. 

Here's a unit test to help:
```
@Test
public void testReconcileBuiltInJobTemplatesUpdateUsesExistingIdentifier() throws IOException {
  String metalakeName = "test_metalake";
  Map<String, JobTemplate> builtInTemplates = new HashMap<>();

  ShellJobTemplate renamedTemplate =
      ShellJobTemplate.builder()
          .withName("builtin-renamed")
          .withComment("updated")
          .withExecutable("/bin/echo")
          .withCustomFields(Collections.singletonMap("version", "v2"))
          .build();
  builtInTemplates.put("builtin-existing", renamedTemplate);

  JobTemplateEntity existingEntity = createJobTemplateEntity("builtin-existing", "v1");
  entityStore.put(existingEntity, false);

  when(jobManager.listJobTemplates(metalakeName))
      .thenReturn(Collections.singletonList(existingEntity));

  listener.reconcileBuiltInJobTemplates(metalakeName, builtInTemplates);
}
```

ajw711:
Hello, I’m interested in working on this improvement. Could you please assign it to me?

justinmclean:
All yours @ajw711 if you have any questions just ask.

PR #10257, opened 2026-03-05
[#10176] fix: update BuiltInJobTemplate identifier to use existing entity name

ajw711:

Changed `newTemplate.name()` to `existing.name()` in `BuiltInJobTemplateEventListener.updateBuiltInJobTemplate()`.

When updating a job template, `EntityStore` uses the logical identifier (`name`) to find and update the target record. The previous code incorrectly built the `NameIdentifier` using the incoming `newTemplate.name()`. Since the new name does not exist in the database yet, it caused a `NoSuchEntityException`. Using `existing.name()` ensures the system locks and updates the correct existing entity.

Fix: #10176 

No.

Ran the specific unit test locally and verified it passes:
`./gradlew :core:test --tests "*testReconcileBuiltInJobTemplatesUpdateUsesExistingIdentifier*"`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
+++ b/core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
@@ -310,16 +310,9 @@ public class BuiltInJobTemplateEventListener implements EventListenerPlugin {
 
   private void updateBuiltInJobTemplate(
       String metalake, JobTemplateEntity existing, JobTemplate newTemplate) {
-    // Use existing.name() instead of newTemplate.name() to construct the identifier.
-    // Although in the current code path (loadBuiltInJobTemplates) the two names are always
-    // equal — because the map is keyed by template.name() — we prefer the stored entity's
-    // name here as a defensive measure.
-    //
-    // If a built-in template is ever renamed between Gravitino versions, the map key
-    // would reflect the new name while the already-persisted entity still carries the
-    // old name. Using newTemplate.name() in that case would cause
-    // a NoSuchEntityException when looking up the entity in the store.
-    // See: https://github.com/apache/gravitino/issues/10176
+    // Use existing.name() (not newTemplate.name()) so the EntityStore can locate the
+    // already-persisted record. Using newTemplate.name() would cause NoSuchEntityException
+    // if the template is ever renamed between Gravitino versions. See #10176.
     NameIdentifier identifier = NameIdentifierUtil.ofJobTemplate(metalake, existing.name());
 
     TreeLockUtils.doWithTreeLock(


core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
```diff
@@ -310,7 +310,7 @@ private void registerNewBuiltInJobTemplate(String metalake, JobTemplate jobTempl
 
   private void updateBuiltInJobTemplate(
       String metalake, JobTemplateEntity existing, JobTemplate newTemplate) {
-    NameIdentifier identifier = NameIdentifierUtil.ofJobTemplate(metalake, newTemplate.name());
+    NameIdentifier identifier = NameIdentifierUtil.ofJobTemplate(metalake, existing.name());
```
jerryshao, line 317: **Fix is semantically correct, but the bug cannot be triggered through the production code path.**

`loadBuiltInJobTemplates()` always inserts with `template.name()` as the map key:
```java
builtInTemplates.put(template.name(), template);  // line 185
```
So in `reconcileBuiltInJobTemplates`, the loop's `name` key always equals `newTemplate.name()`. The lookup `existingBuiltIns.remove(name)` therefore always returns an entity whose `entity.name() == newTemplate.name()`. Both expressions resolve to the same value at runtime.

Using `existing.name()` is still the right idiom — you have a reference to the entity, so its identifier should come from the entity itself rather than the incoming payload. It's a good defensive improvement.

Please add a comment or link the related issue (#10176?) explaining a concrete scenario where `existing.name()` and `newTemplate.name()` can differ, so future readers understand why this guard is needed.

core/src/test/java/org/apache/gravitino/job/TestBuiltInJobTemplateEventListener.java
```diff
@@ -486,6 +486,29 @@ public void testReconcileHandlesRegistrationException() {
...
+    entityStore.put(existingEntity, false);
+
+    when(jobManager.listJobTemplates(metalakeName))
+        .thenReturn(Collections.singletonList(existingEntity));
+
+    listener.reconcileBuiltInJobTemplates(metalakeName, builtInTemplates);
+  }
+
   private void createMetalake(String name, boolean inUse) throws IOException {
     Map<String, String> props = new HashMap<>();
```
jerryshao, line 525: **Unreachable test scenario** — `loadBuiltInJobTemplates()` always populates the map as `builtInTemplates.put(template.name(), template)`, so the map key always equals `newTemplate.name()`. The scenario tested here — where the map key (`"builtin-existing"`) differs from `newTemplate.name()` (`"builtin-renamed"`) — cannot occur through the real call path.

Additionally, the test has no explicit assertions. A reader cannot tell from the test body what the expected outcome is; it simply passes if no exception is thrown. Consider:
1. Wrapping the call in `Assertions.assertDoesNotThrow(...)` to document the intent.
2. Verifying in `entityStore` that the entity was actually updated (similar to `testReconcileBuiltInJobTemplatesUpdate` which checks `version == "v2"`).
3. Using a realistic scenario (same name in map key and template) and commenting why `existing.name()` is preferred.

ajw711:
Thanks for the thorough review, @jerryshao!

I've addressed all your feedback:

1. **Code Comment**: Added a comment in `BuiltInJobTemplateEventListener.java` explaining that `existing.name()` is preferred as a defensive measure — if a built-in template is renamed between Gravitino versions, using `newTemplate.name()` could cause a `NoSuchEntityException`. Linked to #10176.

2. **Realistic Test Scenario**: Updated the test to use the same name for both the map key and the template (`builtin-existing`), which matches the actual production code path.

3. **Explicit Assertions**: Added `Assertions.assertDoesNotThrow()` to document intent, and an `entityStore.get()` check to verify the entity was actually updated (version `v1` → `v2`).

ajw711:
Done. Trimmed the inline comment to make it more concise. Thanks!

Issue #10166, opened 2026-03-03
[Improvement] Possible NPE in revokeRolesFromUser error handling when request body is null


justinmclean:
### What would you like to be improved?

PermissionOperations.revokeRolesFromUser can throw an unintended NullPointerException in its catch block. When the RoleRevokeRequest request is null (for example, due to an empty/invalid request body), the handler calls request.getRoleNames() while building the error response. This masks the original failure path and returns an unhelpful internal error.

### How should we improve?

Make catch-block argument construction null-safe before calling ExceptionHandlers:
- Build role-name string with a null guard, e.g. empty string when request == null (or when role names are null).
- Reuse that safe value in handleUserPermissionOperationException(...).
- If needed, apply the same pattern to similar endpoints (grant/revoke for user/group) to avoid equivalent NPEs.

Here's a unit test to help:
```
@Test
public void testRevokeRolesFromUserWithNullRequest() throws Exception {
  PermissionOperations operations = new PermissionOperations();
  HttpServletRequest request = mock(HttpServletRequest.class);
  when(request.getAttribute(any())).thenReturn(null);
  FieldUtils.writeField(operations, "httpRequest", request, true);

  Response response = operations.revokeRolesFromUser("metalake1", "user1", null);
  Assertions.assertEquals(
      Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), response.getStatus());
}
```

pithecuse527:
can i take this?

justinmclean:
Yes, thanks for offering to help.

pithecuse527:
I'm closing the #10258 for now.
This will be addressed in https://github.com/apache/gravitino/pull/10183 ?

justinmclean:
@pithecuse527 I'm at a conference today, but I will look at the issues later today to sort them out.

PR #10258, opened 2026-03-05
 [#10166] Resolve NPE problem in grant/revoke permission's error handling when request body is null

pithecuse527:
Build role-name string with a null guard in grant/revoke permission operation error handling logics

To avoid NPE

Fix: #10166

No
UTs


pithecuse527:
I'll close the PR for now.
This will be covered in https://github.com/apache/gravitino/pull/10183

PR #10276, opened 2026-03-06
[#10294] feat(authz): Support scoped MANAGE_GRANTS for delegated privilege management

bharos:

Allow MANAGE_GRANTS to be bound to CATALOG/SCHEMA/TABLE/VIEW in addition to METALAKE, enabling SQL WITH GRANT OPTION-style delegation.

Otherwise, before , the grants were only possible from users who have MANAGE_GRANT at metalake level.
There was no option of assigning MANAGE_GRANT at catalog/schema/table level

https://github.com/apache/gravitino/discussions/10267

Fix: #10294

No

Added unit tests

### DIFF
--- a/api/src/main/java/org/apache/gravitino/authorization/Privileges.java
+++ b/api/src/main/java/org/apache/gravitino/authorization/Privileges.java
@@ -854,7 +854,8 @@ public class Privileges {
   }
 
   /**
-   * The privilege to grant or revoke privileges on securable objects.
+   * The privilege to grant or revoke privileges on securable objects. If bound on the metalake, we
+   * can grant or revoke the role for users or groups.
    *
    * <p>Unlike most privileges, this can be bound at any level of the object hierarchy — METALAKE,
    * CATALOG, SCHEMA, TABLE, VIEW, TOPIC, FILESET, or MODEL. A grant at a parent level implicitly


api/src/main/java/org/apache/gravitino/authorization/Privileges.java
```diff
@@ -64,6 +64,18 @@ public class Privileges {
...
+   * Object types to which MANAGE_GRANTS can be bound. Allows scoped grant delegation at catalog,
+   * schema, table, or view level — not just metalake-wide.
+   */
+  private static final Set<MetadataObject.Type> GRANT_SUPPORTED_TYPES =
+      Sets.immutableEnumSet(
+          MetadataObject.Type.METALAKE,
+          MetadataObject.Type.CATALOG,
+          MetadataObject.Type.SCHEMA,
+          MetadataObject.Type.TABLE,
+          MetadataObject.Type.VIEW);
```
roryqi: Two solutions:
1. We add a new privilege property like GRANT_OPTION
2. `MANAGE_GRANTS` to object level.

Option 2 won't add new concepts. It may be easier for understanding.
Option 1 could be more restricted.

I prefer option 2.

roryqi: Maybe should add more objects like model,topic, fileset.

roryqi: We should also add the documents, too.

bharos: Done

bharos:
> **Code Review — Major Issues, Do Not Merge**
> 
> This is a security-model change and needs more work before it can be merged.
> 
> ### 1. PR description is unfilled
> The "What changes were proposed", "Why are the changes needed", "User-facing change", and "How was this patch tested" sections all contain only default template text. Please fill them in per the project PR guidelines.
> 
> ### 2. No unit tests for the new authorization logic
> The `hasMetadataPrivilegePermission` traversal loop in `JcasbinAuthorizer` is security-critical and has no test coverage. Please add tests covering at least:
> 
> * Grant at METALAKE level → covers all children (existing behavior)
> * Grant at CATALOG level → covers TABLE/SCHEMA within that catalog
> * Grant at SCHEMA level → covers TABLE within that schema
> * No grant at any level → returns false
> * `hasSetOwnerPermission` fallback is exercised correctly
> 
> ### 3. Potential parse issue with `fullName.split("\\.", -1)`
> Object names in Gravitino can theoretically contain dots (e.g., in some JDBC catalogs). `fullName.split("\\.", -1)` will incorrectly split such names. Please verify that `fullName` passed to this method is always a dot-separated path of simple names, or use a more robust parsing approach (e.g., `MetadataObjects.parse`).
> 
> ### 4. `MetadataObject.Type.valueOf(type.toUpperCase())` — unchecked conversion
> If `type` is an unexpected string, this throws `IllegalArgumentException` with no context. Consider wrapping with a meaningful error message.
> 
> ### 5. `GRANT_SUPPORTED_TYPES` naming
> Consider naming it `MANAGE_GRANTS_SUPPORTED_TYPES` to make it self-documenting (the existing set is named `FILESET_SUPPORTED_TYPES` etc. which are privilege-specific).
> 
> Please address the above and add tests before requesting re-review.

Addressed the comments

docs/security/access-control.md
```diff
@@ -272,7 +272,7 @@ Gravitino provides a comprehensive set of privileges organized by the type of op
 
 | Name          | Supports Securable Object | Operation                                                                                                     |
 |---------------|---------------------------|---------------------------------------------------------------------------------------------------------------|
-| MANAGE_GRANTS | Metalake                  | Manages roles granted to or revoked from the user or group, and privilege granted to or revoked from the role |
+| MANAGE_GRANTS | Metalake, Catalog, Schema, Table, View, Topic, Fileset, Model | Manages roles granted to or revoked from the user or group, and privileges granted to or revoked from the role. When bound to a Catalog, Schema, Table, View, Topic, Fileset, or Model, the permission is scoped to that object and its descendants only. |
```
roryqi: Only bound to the metalake. MANAGE_GRANTS can manage the roles granted to or revoked from the user or group. Do we need to add this point?

api/src/main/java/org/apache/gravitino/authorization/Privileges.java
bharos: Thanks. we can go with option 2 for now

docs/security/access-control.md
bharos: Good point. I added this in the doc

api/src/main/java/org/apache/gravitino/authorization/Privileges.java
```diff
@@ -836,7 +853,13 @@ public boolean canBindTo(MetadataObject.Type type) {
     }
   }
 
-  /** The privilege to grant or revoke a role for the user or the group. */
+  /**
+   * The privilege to grant or revoke privileges on securable objects.
```
roryqi: If  bound on the metalake, we can grant or revoke the role for users or groups.

bharos: Done

Issue #10281, opened 2026-03-06
[Improvement] Make exception message more clear when entity does not exists


yuqi1129:
### What would you like to be improved?

APIs such as `TableOperations#createTable` will throw `NotFoundException` (expect: NoSuchSchemaException) if the schema does not exist after #9982. For more details, you can refer to 
https://github.com/apache/gravitino/pull/10199#discussion_r2894139921

PR #10282, opened 2026-03-06
[#10281] improvement(core): throw specific exceptions in MetadataObjectUtil

yuqi1129:

   This PR modifies MetadataObjectUtil.checkMetadataObject to throw specific exceptions (e.g., NoSuchSchemaException, NoSuchTableException)
   instead of the generic NoSuchMetadataObjectException when an object is not found.

   Currently, MetadataObjectUtil.checkMetadataObject throws a generic NoSuchMetadataObjectException. This prevents upstream callers (like REST
   API handlers) from distinguishing the exact error type, leading to less informative error responses (e.g., generic 404s instead of specific
   "Schema not found" errors with correct error codes).

   Fix: #10281

   Yes.
    - The REST API will now return more specific error messages and codes when operating on non-existent metadata objects (e.g., trying to
  create a table in a missing schema will now explicitly report NoSuchSchemaException).

    - Added a new regression test testCreateTableWithNonExistentSchema in TestTableOperations to verify that NoSuchSchemaException is thrown
  when the schema does not exist.
    - Verified that existing tests pass.

### DIFF
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/commands/GrantPrivilegesToRole.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/commands/GrantPrivilegesToRole.java
@@ -29,9 +29,9 @@ import org.apache.gravitino.cli.FullName;
 import org.apache.gravitino.cli.Privileges;
 import org.apache.gravitino.client.GravitinoClient;
 import org.apache.gravitino.dto.authorization.PrivilegeDTO;
-import org.apache.gravitino.exceptions.NoSuchMetadataObjectException;
 import org.apache.gravitino.exceptions.NoSuchMetalakeException;
 import org.apache.gravitino.exceptions.NoSuchRoleException;
+import org.apache.gravitino.exceptions.NotFoundException;
 
 /** Grants one or more privileges. */
 public class GrantPrivilegesToRole extends MetadataCommand {
@@ -88,7 +88,7 @@ public class GrantPrivilegesToRole extends MetadataCommand {
       exitWithError(ErrorMessages.UNKNOWN_METALAKE);
     } catch (NoSuchRoleException err) {
       exitWithError(ErrorMessages.UNKNOWN_ROLE);
-    } catch (NoSuchMetadataObjectException err) {
+    } catch (NotFoundException err) {
       exitWithError(ErrorMessages.UNKNOWN_USER);
     } catch (Exception exp) {
       exitWithError(exp.getMessage());
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/commands/OwnerDetails.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/commands/OwnerDetails.java
@@ -27,8 +27,8 @@ import org.apache.gravitino.cli.CommandContext;
 import org.apache.gravitino.cli.CommandEntities;
 import org.apache.gravitino.cli.ErrorMessages;
 import org.apache.gravitino.client.GravitinoClient;
-import org.apache.gravitino.exceptions.NoSuchMetadataObjectException;
 import org.apache.gravitino.exceptions.NoSuchMetalakeException;
+import org.apache.gravitino.exceptions.NotFoundException;
 
 /** Represents the detail information of an owner of an entity. */
 public class OwnerDetails extends Command {
@@ -83,7 +83,7 @@ public class OwnerDetails extends Command {
       owner = client.getOwner(metadata);
     } catch (NoSuchMetalakeException err) {
       exitWithError(ErrorMessages.UNKNOWN_METALAKE);
-    } catch (NoSuchMetadataObjectException err) {
+    } catch (NotFoundException err) {
       exitWithError(ErrorMessages.UNKNOWN_ENTITY);
     } catch (Exception exp) {
       exitWithError(exp.getMessage());
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/commands/RevokePrivilegesFromRole.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/commands/RevokePrivilegesFromRole.java
@@ -29,9 +29,9 @@ import org.apache.gravitino.cli.FullName;
 import org.apache.gravitino.cli.Privileges;
 import org.apache.gravitino.client.GravitinoClient;
 import org.apache.gravitino.dto.authorization.PrivilegeDTO;
-import org.apache.gravitino.exceptions.NoSuchMetadataObjectException;
 import org.apache.gravitino.exceptions.NoSuchMetalakeException;
 import org.apache.gravitino.exceptions.NoSuchRoleException;
+import org.apache.gravitino.exceptions.NotFoundException;
 
 /** Revokes one or more privileges. */
 public class RevokePrivilegesFromRole extends MetadataCommand {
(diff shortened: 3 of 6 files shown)


core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -33,12 +31,20 @@
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.authorization.AuthorizationUtils;
 import org.apache.gravitino.exceptions.IllegalMetadataObjectException;
+import org.apache.gravitino.exceptions.NoSuchCatalogException;
+import org.apache.gravitino.exceptions.NoSuchFilesetException;
 import org.apache.gravitino.exceptions.NoSuchJobException;
 import org.apache.gravitino.exceptions.NoSuchJobTemplateException;
 import org.apache.gravitino.exceptions.NoSuchMetadataObjectException;
+import org.apache.gravitino.exceptions.NoSuchMetalakeException;
```
jerryshao, line 39: `NoSuchMetadataObjectException` is imported but no longer used anywhere in this file after the switch to type-specific exceptions. This will fail Spotless/Checkstyle — please remove the unused import.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -151,67 +157,80 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
...
         break;
 
       case COLUMN:
         NameIdentifierUtil.checkColumn(identifier);
         NameIdentifier tableIdent = NameIdentifier.of(identifier.namespace().levels());
-        check(env.tableDispatcher().tableExists(tableIdent), exceptionToThrowSupplier);
+        check(
+            env.tableDispatcher().tableExists(tableIdent),
+            () -> new NoSuchTableException("Table %s doesn't exist", tableIdent.toString()));
         break;
```
jerryshao, line 205: The error message for the `COLUMN` case uses `tableIdent.toString()` while every other branch uses `object.fullName()`. This produces inconsistent identifier formats in error output. Please standardize to `object.fullName()` here as well:
```suggestion
            () -> new NoSuchTableException("Table %s doesn't exist", object.fullName()));
```

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -151,67 +157,80 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
...
 
       case ROLE:
         AuthorizationUtils.checkRole(identifier);
         try {
           env.accessControlDispatcher().getRole(metalake, object.fullName());
         } catch (NoSuchRoleException nsr) {
-          throw exceptionToThrowSupplier.get();
+          throw nsr;
         }
```
jerryshao, line 234: This `try/catch` block catches `NoSuchRoleException` and immediately rethrows it unchanged — it is a no-op. Remove the `try/catch` entirely and let the exception propagate directly from `getRole(...)`.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -220,7 +239,7 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
         try {
           env.tagDispatcher().getTag(metalake, object.fullName());
         } catch (NoSuchTagException nsr) {
-          throw exceptionToThrowSupplier.get();
+          throw nsr;
         }
```
jerryshao, line 243: Same as the `ROLE` case: this `try/catch` is redundant and adds noise. Remove it and let `NoSuchTagException` propagate directly.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -229,7 +248,7 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
         try {
           env.policyDispatcher().getPolicy(metalake, object.fullName());
         } catch (NoSuchPolicyException nsr) {
-          throw checkNotNull(exceptionToThrowSupplier).get();
+          throw nsr;
         }
```
jerryshao, line 252: Same as above: the `try/catch` is a no-op. Remove it.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -238,7 +257,7 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
         try {
           env.jobOperationDispatcher().getJob(metalake, object.fullName());
         } catch (NoSuchJobException e) {
-          throw exceptionToThrowSupplier.get();
+          throw e;
         }
```
jerryshao, line 261: Same as above: redundant `try/catch`. Remove it.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -247,7 +266,7 @@ public static void checkMetadataObject(String metalake, MetadataObject object) {
         try {
           env.jobOperationDispatcher().getJobTemplate(metalake, object.fullName());
         } catch (NoSuchJobTemplateException e) {
-          throw exceptionToThrowSupplier.get();
+          throw e;
         }
```
jerryshao, line 270: Same as above: redundant `try/catch`. Remove it.

server/src/test/java/org/apache/gravitino/server/web/rest/TestTableOperations.java
```diff
@@ -994,4 +994,41 @@ public static Table mockTable(
...
     return table;
   }
+
+  @Test
+  public void testCreateTableWithNonExistentSchema() {
+    String nonExistentSchema = "non_existent_schema";
+    NameIdentifier schemaIdent = NameIdentifier.of(metalake, catalog, nonExistentSchema);
+
+    // Mock schema not exists
+    when(schemaDispatcher.schemaExists(schemaIdent)).thenReturn(false);
```
jerryshao, line 1004: Nit: grammar — use "does not exist" or "doesn't exist" instead of "not exists".
```suggestion
    // Mock schema does not exist
```

server/src/test/java/org/apache/gravitino/server/web/rest/TestTableOperations.java
```diff
@@ -994,4 +994,41 @@ public static Table mockTable(
...
+        target(tablePath(metalake, catalog, nonExistentSchema))
+            .request(MediaType.APPLICATION_JSON_TYPE)
+            .accept("application/vnd.gravitino.v1+json")
+            .post(Entity.entity(req, MediaType.APPLICATION_JSON_TYPE));
+
+    Assertions.assertEquals(Response.Status.NOT_FOUND.getStatusCode(), resp.getStatus());
+
+    ErrorResponse errorResp = resp.readEntity(ErrorResponse.class);
+    Assertions.assertEquals(NoSuchSchemaException.class.getSimpleName(), errorResp.getType());
+  }
```
jerryshao, line 1033: Consider also asserting `errorResp.getCode()` to align with other tests in this class and make the error payload contract more explicit:
```suggestion
    Assertions.assertEquals(ErrorConstants.NOT_FOUND_CODE, errorResp.getCode());
    Assertions.assertEquals(NoSuchSchemaException.class.getSimpleName(), errorResp.getType());
```

roryqi:
Maybe I should rework with https://github.com/apache/gravitino/pull/9982 . I use schemaDispatcher instead of MetadataObjectUtil.

yuqi1129:
Close it as @roryqi has also reaise a fix #10288

PR #10288, opened 2026-03-06
[#10281] refactor: Revert unnecessary change

roryqi:

Revert unnecessary change

Fix: #10281

No.

Existing tests.

### DIFF
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/test/java/org/apache/gravitino/catalog/clickhouse/integration/test/CatalogClickHouseIT.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/test/java/org/apache/gravitino/catalog/clickhouse/integration/test/CatalogClickHouseIT.java
@@ -317,7 +317,7 @@ public class CatalogClickHouseIT extends BaseIT {
     // create failed check.
     NameIdentifier table = NameIdentifier.of(testSchemaName, "test_table");
     Assertions.assertThrows(
-        NotFoundException.class,
+        NoSuchSchemaException.class,
         () ->
             tableCatalog.createTable(
                 table,
--- a/catalogs-contrib/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/integration/test/CatalogOceanBaseIT.java
+++ b/catalogs-contrib/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/integration/test/CatalogOceanBaseIT.java
@@ -304,7 +304,7 @@ public class CatalogOceanBaseIT extends BaseIT {
     // create failed check.
     NameIdentifier table = NameIdentifier.of(testSchemaName, "test_table");
     Assertions.assertThrows(
-        NotFoundException.class,
+        NoSuchSchemaException.class,
         () ->
             tableCatalog.createTable(
                 table,


server/src/main/java/org/apache/gravitino/server/web/rest/TableOperations.java
```diff
@@ -142,10 +142,7 @@ public Response createTable(
                 NameIdentifierUtil.ofTable(metalake, catalog, schema, request.getName());
 
             // Make sure schema is imported, otherwise set owner for the table may fail.
-            MetadataObjectUtil.checkMetadataObject(
-                metalake,
-                MetadataObjects.of(
-                    Lists.newArrayList(catalog, schema), MetadataObject.Type.SCHEMA));
+            schemaDispatcher.loadSchema(NameIdentifierUtil.ofSchema(metalake, catalog, schema));
```
yuqi1129: `dispatcher.createTable(...)` already calls `schemaDispatcher.loadSchema(...)` in `TableOperationDispatcher#createTable` before entering write lock. Calling `schemaDispatcher.loadSchema(...)` here adds another catalog lookup/import attempt on the hot path.

Could we avoid this duplicate load and keep the validation in one place (dispatcher), unless there is a specific owner-assignment ordering requirement that needs it here?

server/src/main/java/org/apache/gravitino/server/web/rest/TopicOperations.java
```diff
@@ -143,10 +143,7 @@ public Response createTopic(
                 NameIdentifierUtil.ofTopic(metalake, catalog, schema, request.getName());
 
             // Make sure schema is imported, otherwise set owner for the topic may fail.
-            MetadataObjectUtil.checkMetadataObject(
-                metalake,
-                MetadataObjects.of(
-                    Lists.newArrayList(catalog, schema), MetadataObject.Type.SCHEMA));
+            schemaDispatcher.loadSchema(NameIdentifierUtil.ofSchema(metalake, catalog, schema));
```
yuqi1129: Same concern as TableOperations: `dispatcher.createTopic(...)` already performs `schemaDispatcher.loadSchema(...)` in `TopicOperationDispatcher#createTopic`. This adds an extra schema load round-trip in REST before dispatcher execution.

Recommend removing this duplicate call (or documenting why the pre-load is required here).

server/src/main/java/org/apache/gravitino/server/web/rest/TableOperations.java
roryqi: If we are using managed storage, we can avoid this issue. We use loadSchema to import external existed entities now. I can document this first.

yuqi1129:
@roryqi 
I suggest postponing this PR since it's not a big problem.

roryqi:
> @roryqi I suggest postponing this PR since it's not a big problem.

Actually  MetadataObjectUtils will call the load schema method, too. I will add the comments and continue the work since 1.2.0 is released.

roryqi:
> @roryqi I suggest postponing this PR since it's not a big problem.

I got your point.

roryqi:
@yuqi1129 My fix is useless. So I revert it. I will follow the root cause when the issue encounters again.

server/src/main/java/org/apache/gravitino/server/web/rest/TopicOperations.java
```diff
@@ -142,12 +139,6 @@ public Response createTopic(
             NameIdentifier ident =
                 NameIdentifierUtil.ofTopic(metalake, catalog, schema, request.getName());
 
-            // Make sure schema is imported, otherwise set owner for the topic may fail.
-            MetadataObjectUtil.checkMetadataObject(
```
yuqi1129, line 146: Revert the logic to check shcema is imported?

roryqi: Yes, you are right. Table Operation dispatcher has checked this point.

Image

yuqi1129: I see.

Issue #10217, opened 2026-03-04
[Improvement] Catalog create rollback can mask original post-hook exception


justinmclean:
### What would you like to be improved?

CatalogHookDispatcher#createCatalog performs post-create hooks (setOwner, future grant) inside a try/catch. On post-hook failure, it attempts rollback via dispatcher.dropCatalog(ident, true), then rethrows the original exception. If rollback also throws, the rollback exception replaces the original post-hook exception. This masks root cause, making debugging and error attribution incorrect.

### How should we improve?

Preserve the original post-hook exception as primary. In the catch, wrap dropCatalog in its own try/catch; if rollback fails, attach rollback exception as suppressed to the original (e.addSuppressed(rollbackEx)), then rethrow the original e. Optionally log rollback failure. This keeps root cause intact while still surfacing rollback failure details.

Here's a unit test to help:
```

public class TestCatalogHookDispatcher {

  @Test
  public void testCreateCatalogRollbackExceptionMasksPostHookException() throws Exception {
    GravitinoEnv gravitinoEnv = GravitinoEnv.getInstance();
    Object originalOwnerDispatcher = FieldUtils.readField(gravitinoEnv, "ownerDispatcher", true);
    Object originalFutureGrantManager =
        FieldUtils.readField(gravitinoEnv, "futureGrantManager", true);

    CatalogDispatcher dispatcher = Mockito.mock(CatalogDispatcher.class);
    Catalog catalog = Mockito.mock(Catalog.class);
    NameIdentifier ident = NameIdentifier.of("metalake", "catalog");
    RuntimeException postHookException = new RuntimeException("post-hook failed");
    RuntimeException rollbackException = new RuntimeException("rollback failed");

    OwnerDispatcher ownerDispatcher = Mockito.mock(OwnerDispatcher.class);
    Mockito.doThrow(postHookException)
        .when(ownerDispatcher)
        .setOwner(Mockito.anyString(), Mockito.any(), Mockito.anyString(), Mockito.any());
    Mockito.when(
            dispatcher.createCatalog(
                Mockito.eq(ident),
                Mockito.eq(Catalog.Type.RELATIONAL),
                Mockito.eq("provider"),
                Mockito.eq("comment"),
                Mockito.anyMap()))
        .thenReturn(catalog);
    Mockito.doThrow(rollbackException).when(dispatcher).dropCatalog(ident, true);

[...]

PR #10293, opened 2026-03-06
[#10217] fix(core): preserve post-hook exception when rollback fails

pythaac:

In `createCatalog()`, when post-hook operations fail, Gravitino attempts to roll back by calling `dropCatalog()`.
- Before this change, if the rollback also failed, the rollback exception masked the original post-hook exception.
- After this change, the original post-hook exception is rethrown, and the rollback exception is attached as a suppressed exception.

The previous behavior could hide the root cause (post-hook failure), making debugging difficult.

Fix: #10217

N/A

Added unit tests in `TestCatalogHookDispatcher`:

1. Verify that the original post-hook exception is thrown when rollback succeeds.
2. Verify that the original post-hook exception is still thrown and the rollback exception is attached as suppressed when rollback fails.


yuqi1129:
Could also check if the same problem exists in other dispatchers?

pythaac:
> Could also check if the same problem exists in other dispatchers?

Hello @yuqi1129 , thanks for your comment! I've looked into the masking issue for other dispatchers, but there's no issue.
The only similar pattern I found is:
```java
    try {
      NameIdentifier[] nameIdentifiers = dispatcher.listTables(namespace);
      eventBus.dispatchEvent(new ListTableEvent(PrincipalUtils.getCurrentUserName(), namespace));
      return nameIdentifiers;
    } catch (Exception e) {
      eventBus.dispatchEvent(
          new ListTableFailureEvent(PrincipalUtils.getCurrentUserName(), namespace, e));
      throw e;
    }
```
in case that an exception occurs during the failure event is handled in catch statements.

Please let me know if I need to prevent this masking as well🙂

yuqi1129:
> The previous behavior could hide the root cause (post-hook failure), making debugging difficult.

I see. It may be another problem as other dispatchers do not contain `try/catch` block in `createXXX` method.

Issue #6944, opened 2025-04-15
[Improvement] Add support to model properties command


Abyss-lord:
### What would you like to be improved?

 Add support to model properties command, model now support update properties, so CLI should add support for properties command.

### How should we improve?

CLI  add support for model properties command.

Abyss-lord:
@unknowntpo here

PR #10296, opened 2026-03-07
[#6944] feat(cli): Add model properties list commands

zhoukangcn:

Add `model properties` command to list properties of a model or model version, consistent with other entities (catalog, table, fileset, topic, etc.).

### New commands:
- `gcli model properties --name catalog.schema.model` - List all properties of a model
- `gcli model properties --name catalog.schema.model --version N` - List properties of a specific model version
- `gcli model properties --name catalog.schema.model --alias xxx` - List properties of a model version by alias

Model was the only entity type missing the `properties` command. All other entities (Catalog, Fileset, Table, Topic, Schema, Metalake, Tag) already have this command.

Fix: #6944

## How was this patch tested?

- Unit tests added in `TestModelCommands.java`
- Manual CLI testing against running Gravitino server


zhoukangcn:
@Abyss-lord  Can you help to review this PR

clients/cli/src/main/java/org/apache/gravitino/cli/commands/ListModelVersionProperties.java
```diff
@@ -0,0 +1,122 @@
...
+
+  /** {@inheritDoc} */
+  @Override
+  public Command validate() {
+    if (alias != null && version != null) {
+      exitWithError("Cannot specify both alias and version");
+    }
+    if (alias == null && version == null) {
+      exitWithError("Either alias or version must be specified");
+    }
```
Abyss-lord, line 119: I don't think this needs to be checked; the `ModelCommandHandler` ensures that both cannot be null at the same time during dispatch.

clients/cli/src/main/java/org/apache/gravitino/cli/commands/ListModelVersionProperties.java
```diff
@@ -0,0 +1,122 @@
...
+    return this;
+  }
```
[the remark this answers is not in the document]
Abyss-lord: Please define the error message in `ErrorMessages`

clients/cli/src/main/java/org/apache/gravitino/cli/ModelCommandHandler.java
```diff
@@ -310,6 +314,23 @@ private void handleRemoveCommand() {
...
+      gravitinoCommandLine
+          .newListModelVersionProperties(context, metalake, catalog, schema, model, version, alias)
+          .validate()
+          .handle();
+    } else {
+      gravitinoCommandLine
+          .newListModelProperties(context, metalake, catalog, schema, model)
+          .validate()
+          .handle();
+    }
```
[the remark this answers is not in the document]
Abyss-lord: plz update `clients/cli/src/main/resources/model_help.txt`

clients/cli/src/test/java/org/apache/gravitino/cli/commands/TestListModelProperties.java
Abyss-lord, line 1: I think we should move all Model-related command tests to `clients/cli/src/test/java/org/apache/gravitino/cli/TestModelCommands.java`

clients/cli/src/test/java/org/apache/gravitino/cli/commands/TestListModelVersionProperties.java
Abyss-lord, line 1: I think we should move all Model-related command tests to `clients/cli/src/test/java/org/apache/gravitino/cli/TestModelCommands.java`

clients/cli/src/test/java/org/apache/gravitino/cli/commands/TestListModelVersionProperties.java
```diff
@@ -0,0 +1,168 @@
...
+    CommandLine mockCmdLine = Mockito.mock(CommandLine.class);
+    CommandContext context = new CommandContext(mockCmdLine);
+
+    ListModelVersionProperties command =
+        new ListModelVersionProperties(
+            context, "metalake1", "catalog1", "schema1", "model1", null, null);
+
+    RuntimeException ex = Assertions.assertThrows(RuntimeException.class, command::handle);
+    Assertions.assertTrue(ex.getMessage().contains("Exit with code"));
+  }
```
Abyss-lord, line 125: I don't think it's possible for both `alias` and `version` in `ListModelVersionProperties` to be null at the same time, so there's no need to test for this scenario.

PR #10298, opened 2026-03-07
Fix possible NPE in TagMetaService#getTagIdByTagName when tag does no…

### LINKED ISSUE #10221: [Improvement] Possible NPE in TagMetaService#getTagIdByTagName when tag does not exist

getTagIdByTagName(Long metalakeId, String tagName) calls selectTagMetaByMetalakeIdAndName(...) and immediately dereferences .getTagId(). When the tag is missing, the mapper can return null, causing a NullPointerException instead of a domain-level “not found” exception. This is reachable via TAG ID resolution paths and can surface in production for missing or concurrently deleted tags.

### How should we improve?

Add a null check in getTagIdByTagName: store the mapper result in a local TagPO, and if it is null, throw NoSuchEntityException for tagName. This keeps behavior consistent with other meta services (role/user/group).

Here's a test to help:
```
@TestTemplate
public void testGetTagIdByTagNameWhenTagNotFound() throws IOException {
  createAndInsertMakeLake(METALAKE_NAME);

  TagMetaService tagMetaService = TagMetaService.getInstance();
  long metalakeId = MetalakeMetaService.getInstance().getMetalakeIdByName(METALAKE_NAME);

  Assertions.assertThrows(
      NoSuchEntityException.class,
      () -> tagMetaService.getTagIdByTagName(metalakeId, "missing_tag"));
}

```

Aditi102005:

This PR fixes a potential `NullPointerException` in `TagMetaService#getTagIdByTagName`.

Previously, the method directly dereferenced the result returned by
`selectTagMetaByMetalakeIdAndName(...)` and called `.getTagId()` without checking if the returned `TagPO` was `null`. If the tag did not exist (or was concurrently deleted), this could result in a `NullPointerException`.

This change adds a null check and throws a `NoSuchEntityException` when the tag is not found. This behavior aligns with other meta services (e.g., role/user/group) that throw domain-level exceptions when entities are missing.

## How was this patch tested?

* Added a new unit test `testGetTagIdByTagNameWhenTagNotFound` in `TestTagMetaService`.
* The test verifies that `NoSuchEntityException` is thrown when attempting to retrieve a tag ID for a non-existent tag.
* Ran the full `:core:test` suite locally and all tests passed successfully.

No user-facing changes. This improves internal error handling and prevents unexpected `NullPointerException`s.

Fixed #10221

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/TagMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/TagMetaService.java
@@ -410,10 +410,18 @@ public class TagMetaService {
   }
 
   public Long getTagIdByTagName(Long metalakeId, String tagName) {
-    return SessionUtils.getWithoutCommit(
+    TagPO tagPO =
+        SessionUtils.getWithoutCommit(
             TagMetaMapper.class,
-            mapper -> mapper.selectTagMetaByMetalakeIdAndName(metalakeId, tagName))
-        .getTagId();
+            mapper -> mapper.selectTagMetaByMetalakeIdAndName(metalakeId, tagName));
+
+    if (tagPO == null) {
+      throw new NoSuchEntityException(
+          NoSuchEntityException.NO_SUCH_ENTITY_MESSAGE,
+          Entity.EntityType.TAG.name().toLowerCase(),
+          tagName);
+    }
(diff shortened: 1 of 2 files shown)


yuqi1129:
Good catch and fix.

PR #10300, opened 2026-03-08
issue/10268: Handled configs servlet write failures with explicit HTT…

Tarantula471:
…P 500

Handle /configs servlet write failures with explicit HTTP 500 by setting HttpServletResponse.SC_INTERNAL_SERVER_ERROR and returning JSON error body.

It's an improvement for proper error handling because without setting a status code, the response defaults to 200 (success), which misleads the client into thinking the operation succeeded when it actually failed.

Fix: Set status as HttpServletResponse.SC_INTERNAL_SERVER_ERROR which represents 500 Internal server error so that client can properly understand the response outcome.

Whenever exception occurs in doGet() it will return proper error code the client.

image
image


justinmclean (approved):
Thanks for working on this

PR #10301, opened 2026-03-08
[#10274] fix(core): remove returning false in NoSuchEntityException for unmanaged table

### LINKED ISSUE #10274: [Improvement]  Fix unmanaged purgeTable result when store entity is missing

TableOperationDispatcher.purgeTable for unmanaged tables returns false when store.delete(ident, TABLE) throws NoSuchEntityException, even if the catalog purge succeeded. This contradicts the method’s own comment that store deletion outcome should not affect the returned result, and only the catalog result should be used. As a result, callers can receive false negatives (DropResponse(false)) for successful catalog purges.

### How should we improve?

Align purgeTable with dropTable behavior: in the unmanaged branch, catch NoSuchEntityException, log a warning, and continue returning droppedFromCatalog instead of returning false. Keep throwing on other unexpected exceptions.

Here a test to help:
```

@Test
public void testPurgeUnmanagedTableWithMissingStoreEntityShouldUseCatalogResult()
    throws IOException {
  CatalogManager mockedCatalogManager = mock(CatalogManager.class);
  EntityStore mockedStore = mock(EntityStore.class);
  IdGenerator mockedIdGenerator = mock(IdGenerator.class);
  CatalogManager.CatalogWrapper mockedCatalogWrapper = mock(CatalogManager.CatalogWrapper.class);
  org.apache.gravitino.connector.capability.Capability mockedCapability =
      mock(org.apache.gravitino.connector.capability.Capability.class);
  TableOperationDispatcher dispatcherUnderTest =
      new TableOperationDispatcher(mockedCatalogManager, mockedStore, mockedIdGenerator);

  NameIdentifier tableIdent = NameIdentifier.of(metalake, catalog, "schema72", "table32");
  NameIdentifier catalogIdent = NameIdentifier.of(metalake, catalog);
  doReturn(mockedCatalogWrapper).when(mockedCatalogManager).loadCatalogAndWrap(catalogIdent);
  doReturn(mockedCapability).when(mockedCatalogWrapper).capabilities();
  doReturn(CapabilityResult.unsupported("unmanaged table"))
      .when(mockedCapability)
      .managedStorage(org.apache.gravitino.connector.capability.Capability.Scope.TABLE);
  doReturn(true).when(mockedCatalogWrapper).doWithTableOps(any());
  doThrow(new NoSuchEntityException("missing store entity"))
      .when(mockedStore)
      .delete(tableIdent, TABLE);

  Assertions.assertTrue(dispatcherUnderTest.purgeTable(tableIdent));
}
```

pythaac:

In `purgeTable()`, when `store.delete()` throws `NoSuchEntityException` for unmanaged table:
- Before this change, it returns `false`
- After this change, it returns the catalog result (`droppedFromCatalog`)

The previous behavior contradicts the method’s own comment that store deletion outcome should not affect the returned result, and only the catalog result should be used.

Fix: #10274

`dropped` for DELETE /metalakes/{metalake}/catalogs/{catalog}/schemas/{schema}/tables/{table}?purge=true:
- Before this change, it is always `false`
- After this change, it is the catalog result (`droppedFromCatalog`)

Verify that `purgeTable()` returns true when "droppedFromCatalog" is true in NoSuchEntityException for unmanaged table

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
@@ -402,7 +402,6 @@ public class TableOperationDispatcher extends OperationDispatcher implements Tab
             store.delete(ident, TABLE);
           } catch (NoSuchEntityException e) {
             LOG.warn("The table to be purged does not exist in the store: {}", ident, e);
-            return false;
           } catch (Exception e) {
             throw new RuntimeException(e);
           }
--- a/api/src/main/java/org/apache/gravitino/rel/TableCatalog.java
+++ b/api/src/main/java/org/apache/gravitino/rel/TableCatalog.java
@@ -292,7 +292,7 @@ public interface TableCatalog {
    * is removed.
    *
    * @param ident A table identifier.
-   * @return True if the table is dropped, false if the table does not exist.
+   * @return True if the table is dropped, false if the table does not exist in the underlying catalog.
    */
   boolean dropTable(NameIdentifier ident);
 
@@ -306,7 +306,7 @@ public interface TableCatalog {
    * implementation throws an {@link UnsupportedOperationException}.
    *
    * @param ident A table identifier.
-   * @return True if the table is purged, false if the table does not exist.
+   * @return True if the table is purged, false if the table does not exist in the underlying catalog.
    * @throws UnsupportedOperationException If the catalog does not support to purge a table.
    */
   default boolean purgeTable(NameIdentifier ident) throws UnsupportedOperationException {
--- a/api/src/main/java/org/apache/gravitino/rel/TableCatalog.java
+++ b/api/src/main/java/org/apache/gravitino/rel/TableCatalog.java
@@ -292,7 +292,8 @@ public interface TableCatalog {
    * is removed.
    *
    * @param ident A table identifier.
-   * @return True if the table is dropped, false if the table does not exist in the underlying catalog.
+   * @return True if the table is dropped, false if the table does not exist in the underlying
+   *     catalog.
    */
   boolean dropTable(NameIdentifier ident);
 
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -402,7 +402,6 @@ public boolean purgeTable(NameIdentifier ident) throws UnsupportedOperationExcep
             store.delete(ident, TABLE);
           } catch (NoSuchEntityException e) {
             LOG.warn("The table to be purged does not exist in the store: {}", ident, e);
-            return false;
```
yuqi1129, line 405: The problem lies in the method signature `True if the table is purged, false if the table does not exist.`
`the table` in `false if the table does not exist` refer to table in the underlying storage, or Gravitino store or both.

pythaac: Thanks for pointing this out, @yuqi1129 ! It is a very valid point regarding the ambiguity of the term "table" in the method signature🙂

After reviewing the context, here is my understanding:

1. Both `dropTable` and `purgeTable` are designed to delete the entity from both the Gravitino store and the underlying catalog, meaning the "table" refers to both of them.
2. However, both methods return the deletion result based **only from the underlying catalog**, regardless of whether the table exists in the Gravitino store.
3. As a result, the current Javadoc(`... false if the table does not exist`) is ambiguous. It doesn't specify whether "table" refers to the entity in the Gravitino store or the underlying catalog.

To resolve this ambiguity, I think we should update the javadoc as follows:
- `dropTable` : True if the table is dropped, false if the table does not exist `in the underlying catalog`.
- `purgeTable` : True if the table is purged, false if the table does not exist `in the underlying catalog`.

If you agree with this approach, I will update them in this PR. Please Let me know what you think!

yuqi1129: The change is reasonable. @jerryshao what's your thought on changing the method descriptions for Drop/Purge Table?

Issue #10284, opened 2026-03-06
[Bug report] Catalog properties `jdbc.pool.min-size` `jdbc.pool.max-size` can't be loaded even creating catalog with these properties


danhuawang:
### Version

main branch

### Describe what's wrong

1. Create catalog with the following properties
```
    "properties": {
        "jdbc-url": "jdbc:clickhouse://127.0.0.1:8123",
        "jdbc-driver": "com.clickhouse.jdbc.ClickHouseDriver",
        "jdbc-user": "default",
        "jdbc-password": "",
        "jdbc.pool.min-size": "5",
        "jdbc.pool.max-size": "20",
        "jdbc.pool.max-wait-ms": "60000"
    }
```
2. load the catalog created in step 1

### Error message and/or stacktrace

N/A

### How to reproduce

Scenario: Create ClickHouse catalog with connection pool settings                           # src/test/resources/features/catalog/clickhouseCatalog.feature:15
Request method:	POST
Request URI:	http://127.0.0.1:18090/api/metalakes/clickhouse_test/catalogs
Proxy:			<none>
Request params:	<none>
Query params:	<none>
Form params:	<none>
Path params:	<none>
Headers:		Accept=application/vnd.gravitino.v1+json
				Authorization=Basic YW5vbnltb3VzOnRlc3Q=
				Content-Type=application/json
Cookies:		<none>
Multiparts:		<none>
Body:
{
    "name": "ck_catalog2",
    "type": "RELATIONAL",
    "provider": "jdbc-clickhouse",
    "comment": "This is my ClickHouse catalog with pool settings",
    "properties": {
        "jdbc-url": "jdbc:clickhouse://127.0.0.1:8123",
        "jdbc-driver": "com.clickhouse.jdbc.ClickHouseDriver",
        "jdbc-user": "default",
        "jdbc-password": "",
        "jdbc.pool.min-size": "5",
        "jdbc.pool.max-size": "20",
        "jdbc.pool.max-wait-ms": "60000"
    }
}

danhuawang:
@yuqi1129 Could you help check this issue?

yuqi1129:
The reason is that properties `POOL_MAX_SIZE` and `POOL_MIN_SIZE` are hidden, so we loading a catalog, Gravitino will make them invisible.

yuqi1129:
[...]

PR #10314, opened 2026-03-09
[#10284] fix(catalog-jdbc): keep jdbc pool size properties visible

yuqi1129:

This PR fixes JDBC catalog property visibility for connection pool sizing.

 returns catalog properties via , which filters hidden properties. Since  and  were incorrectly marked hidden, they disappeared when loading catalog properties after creation.

Fix: #10284

Yes.
Users who set  /  during JDBC catalog creation can now see them in loaded catalog properties as expected.

UTs


catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/JdbcCatalogPropertiesMetadata.java
roryqi, line 112: test borrow should be hidden?

yuqi1129: Yes, I prefer to keep it as it is.

Issue #10315, opened 2026-03-09
[Bug report] GCS credential tests fail with "Provided token is either not valid, or expired" error


geyanggang:
### Version

main branch

### Describe what's wrong

The GCS credential integration tests (test_gvfs_with_gcs_credential.py) fail when using
temporary OAuth2 tokens from Gravitino Server. The gcsfs library (upgraded from 2024.3.1
to 2026.2.0) no longer accepts raw token strings and requires a google.auth.credentials.Credentials
object instead.

All 19 tests in TestGvfsWithGCSCredential fail, while TestGvfsWithGCS (using service account
file) passes successfully.

### Error message and/or stacktrace

File "gcsfs/credentials.py", line 59, in _get_creds_from_raw_token
raise ValueError("Provided token is either not valid, or expired.")
ValueError: Provided token is either not valid, or expired.

ERROR: test_rmdir (tests.integration.test_gvfs_with_gcs_credential.TestGvfsWithGCSCredential)
Traceback (most recent call last):
File "gravitino/filesystem/gvfs_storage_handler.py", line 408, in get_filesystem_with_expiration
self.get_filesystem(token=credential.token(), **kwargs)
...
File "gcsfs/credentials.py", line 59, in _get_creds_from_raw_token
raise ValueError("Provided token is either not valid, or expired.")

### How to reproduce

cd gravitino-internal/clients/client-python
python3 -m venv venv
source venv/bin/activate
pip install .
pip install -r requirements-dev.txt

export GCS_SERVICE_ACCOUNT_JSON_PATH_FOR_CREDENTIAL="/path/to/service-account.json"
export GCS_BUCKET_NAME_FOR_CREDENTIAL="your-bucket-name"

python -m unittest tests.integration.test_gvfs_with_gcs_credential -v

PR #10317, opened 2026-03-09
[#10315]fix(client-python): Fix GCS credential handling for gcsfs 2026.2.0 compatibility

geyanggang:

Wrap GCS OAuth2 access tokens in a google.auth.credentials.Credentials object before
passing to gcsfs.GCSFileSystem. The new gcsfs version (2026.2.0) requires Credentials
objects instead of raw token strings.

Changes:

Created StaticCredentials class implementing google.auth.credentials.Credentials interface
Properly converts token string and expiry time to Credentials object
Uses google.auth._helpers.utcnow() for consistent time comparison

The gcsfs library was upgraded from 2024.3.1 to 2026.2.0 in commit https://github.com/datastrato/gravitino-internal/commit/79e37271216ebed689b1b4ea592adb12d6173170. The new
version changed the token parameter handling and no longer accepts raw OAuth2 token strings.
It now requires google.auth.credentials.Credentials objects.

Without this fix, all GCS credential-based tests fail with "Provided token is either not
valid, or expired" error, breaking the credential vending feature for GCS storage.

Fix: #10315 

No.

The CI can pass.

### DIFF
--- a/clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
+++ b/clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
@@ -42,8 +42,8 @@ from gravitino.exceptions.base import GravitinoRuntimeException
 from gravitino.filesystem.gvfs_config import GVFSConfig
 
 try:
-    from google.auth import _helpers
     from google.auth.credentials import Credentials
+    from google.auth.exceptions import RefreshError
 
     GOOGLE_AUTH_AVAILABLE = True
 except ImportError:
@@ -433,7 +433,7 @@ class GCSStorageHandler(StorageHandler):
 
                         def refresh(self, request):
                             # Static token cannot be refreshed
-                            raise RuntimeError(
+                            raise RefreshError(
                                 "Cannot refresh static access token. "
                                 "Please request a new credential from Gravitino."
                             )
@@ -442,7 +442,7 @@ class GCSStorageHandler(StorageHandler):
                         def expired(self):
                             if not self.expiry:
                                 return False
-                            return _helpers.utcnow() >= self.expiry
+                            return datetime.now(timezone.utc) >= self.expiry
 
                         @property
                         def valid(self):


clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
```diff
@@ -40,6 +41,14 @@
 from gravitino.exceptions.base import GravitinoRuntimeException
 from gravitino.filesystem.gvfs_config import GVFSConfig
 
+try:
+    from google.auth import _helpers
```
yuqi1129: Using `google.auth._helpers` relies on a private module API. Could we switch this to a public API (e.g. `datetime.now(timezone.utc)` or another public helper) to avoid breakage on future google-auth upgrades?

clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
```diff
@@ -403,10 +412,50 @@ def get_filesystem_with_expiration(
...
+                            super().__init__()
+                            self.token = token_value
+                            # Convert milliseconds to datetime
+                            self.expiry = datetime.fromtimestamp(
+                                expiry_time_ms / 1000.0, tz=timezone.utc
+                            )
+
+                        def refresh(self, request):
+                            # Static token cannot be refreshed
+                            raise RuntimeError(
```
yuqi1129: `Credentials.refresh()` callers in google-auth/gcsfs typically expect a `google.auth.exceptions.RefreshError` (or an `AuthError`) on refresh failures. Raising a generic `RuntimeError` here can bypass expected handling paths; please consider raising `RefreshError` instead.

geyanggang: Ok! Changed to raise `RefreshError` instead of `RuntimeError` so that google-auth/gcsfs callers can handle it properly through their expected error handling paths.

clients/client-python/gravitino/filesystem/gvfs_storage_handler.py
geyanggang: Replaced `google.auth._helpers.utcnow()` with `datetime.now(timezone.utc)` to use public API only. This avoids potential breakage on future google-auth upgrades.

Issue #10177, opened 2026-03-04
[Improvement] Fix SessionUtils transaction cleanup for non-Exception throwables


justinmclean:
### What would you like to be improved?

SessionUtils.doWithCommit(...) catches only Exception. If the mapper callback throws a non-Exception Throwable (for example AssertionError), the catch block is skipped, so SqlSessions.rollbackAndCloseSqlSession() is not called. This can leave transaction/session state unclosed.

### How should we improve?

Broaden failure handling so cleanup runs for all throwables from the callback path. In practice, catch Throwable, call SqlSessions.rollbackAndCloseSqlSession(), and rethrow. Apply the same pattern consistently.

Here's a unit test to help:
```
public class TestSessionUtils {

  @Test
  public void testDoWithCommitShouldRollbackOnAssertionError() {
    Object mapper = new Object();

    try (MockedStatic<SqlSessions> mockedSqlSessions = mockStatic(SqlSessions.class)) {
      mockedSqlSessions.when(() -> SqlSessions.getMapper(Object.class)).thenReturn(mapper);

      assertThrows(
          AssertionError.class,
          () ->
              SessionUtils.doWithCommit(
                  Object.class,
                  ignored -> {
                    throw new AssertionError("boom");
                  }));

      mockedSqlSessions.verify(SqlSessions::rollbackAndCloseSqlSession);
      mockedSqlSessions.verify(() -> SqlSessions.commitAndCloseSqlSession(), never());
    }
  }
}
```

geniusjun:
Hi, I’d like to work on this improvement. Could you please assign it to me?

justinmclean:
Hi @geniusjun, thanks for offering to help. If you have any questions, just ask.

PR #10335, opened 2026-03-10
[#10177] improvement(core): ensure SessionUtils rolls back on Throwable

geniusjun:

This PR updates `SessionUtils` to ensure transaction cleanup also runs when callbacks throw non-`Exception` throwables.

The changes include:
- broaden the failure handling in `doWithCommit` from `Exception` to `Throwable`
- apply the same pattern to `doWithCommitAndFetchResult`
- apply the same pattern to `doMultipleWithCommit`
- add unit tests to verify rollback happens when callbacks throw `AssertionError`

With this change, `SqlSessions.rollbackAndCloseSqlSession()` is invoked consistently for callback failures before the throwable is rethrown.

`SessionUtils` previously caught only `Exception` in commit-managed helper methods.

If a mapper callback threw a non-`Exception` throwable such as `AssertionError`, the rollback path was skipped, which could leave the SQL session or transaction state unclosed.

This PR fixes that bug by making rollback/close run for all throwables in the callback path.

Fix: #10177

No.

Added unit tests in `TestSessionUtils` to verify:
- `doWithCommit` rolls back and does not commit on `AssertionError`
- `doWithCommitAndFetchResult` rolls back and does not commit on `AssertionError`
- `doMultipleWithCommit` rolls back and does not commit on `AssertionError`

Verified with:

```bash
env JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.16/libexec/openjdk.jdk/Contents/Home \
  ./gradlew :core:test --tests org.apache.gravitino.storage.relational.utils.TestSessionUtils -PskipITs

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/utils/SessionUtils.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/utils/SessionUtils.java
@@ -40,9 +40,9 @@ public class SessionUtils {
       T mapper = SqlSessions.getMapper(mapperClazz);
       consumer.accept(mapper);
       SqlSessions.commitAndCloseSqlSession();
-    } catch (Exception e) {
+    } catch (Throwable t) {
       SqlSessions.rollbackAndCloseSqlSession();
-      throw e;
+      throw t;
     }
   }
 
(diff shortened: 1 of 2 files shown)


yuqi1129:
Good job.

PR #10348, opened 2026-03-10
[#10350]fix: jdk 17+ add opens on start script

### LINKED ISSUE #10350: [Bug report] Failed to initialize MemoryUtil on JDK 21
### Version

main branch

### Describe what's wrong

system: mac os
jdk: 21

create table on lance rest server, got something wrong

```
Caused by: java.lang.RuntimeException: Failed to initialize MemoryUtil. You must start Java with `--add-opens=java.base/java.nio=ALL-UNNAMED` (See https://arrow.apache.org/docs/java/install.html)
	at org.apache.arrow.memory.util.MemoryUtil.<clinit>(MemoryUtil.java:143)
	... 98 more
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field long java.nio.Buffer.address accessible: module java.base does not "opens java.nio" to unnamed module @1fe20588
	at java.base/java.lang.reflect.AccessibleObject.throwInaccessibleObjectException(AccessibleObject.java:391)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:367)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:315)
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:183)
	at java.base/java.lang.reflect.Field.setAccessible(Field.java:177)
	at org.apache.arrow.memory.util.MemoryUtil.<clinit>(MemoryUtil.java:88)
```

### How to reproduce

```
./bin/gravitino.sh start
./bin/gravitino.sh run
```

create table on lance rest server

### Additional context

_No response_

Gallardot:

fix the start script on jdk 17+

Fix: #10350
- close #10350

None

run on jdk 17+

### DIFF
--- a/api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
+++ b/api/src/main/java/org/apache/gravitino/policy/IcebergDataCompactionContent.java
@@ -1,0 +1,280 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.policy;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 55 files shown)


yuqi1129:
Follow-up (format-fixed):

For issue #10350, JDK versions above 17 do need the same add-opens JVM options as JDK 17 (especially java.base/java.nio for Arrow MemoryUtil).

So this PR change from "JVM_VERSION -eq 17" to "JVM_VERSION -ge 17" in the three startup scripts is correct.

One potential gap: bin/gravitino-optimizer.sh.template still has "if [ "$JVM_VERSION" -eq 17 ]". If optimizer can load similar reflective/Arrow paths on JDK 21+, consider aligning it to "-ge 17" for consistency.

yuqi1129:
Follow-up on JDK scope:

I agree we should avoid over-opening if not necessary.

For this issue (#10350), the failure is Arrow `MemoryUtil` on JDK 21, and Arrow's current Java docs (tested on JDK 11/17/21/latest) require opening `java.base/java.nio` for memory modules. So for JDK 21/24, we definitely need that `java.nio` open.

I don't see evidence from this issue that the full JDK17 list is strictly required on 21/24. If we want least-privilege behavior, we could consider:
1) keep a minimal required set (at least `java.base/java.nio`) for JDK >=17, and
2) only add extra opens when a concrete dependency/runtime path needs them.

If we keep the full list for safety, please add rationale in script comments (compatibility-first tradeoff).

jerryshao:
I think we don't officially claim that we support JDK 21. If we want to support JDK 21, a simple script changing may not be enough.

yuqi1129:
I'd better try to solve the obvious issue in JDK21 or higher, like startup problem, and we encourage users to use 21+. For higher versions like 21+, I can't offer any guarantees of certainty, except to say that theoretically, it should be fine.

PR #10355, opened 2026-03-11
[MINOR] optimizer: relax statistics input validation for non-local calculators

FANNG1:
## What changes
- Remove the CLI rule that forbids `--statistics-payload` / `--file-path` when `--calculator-name` is not `local-stats-calculator`.
- Keep existing local-calculator requirement: when calculator is `local-stats-calculator`, one of `--statistics-payload` or `--file-path` is still required.
- Update CLI option descriptions to be calculator-agnostic.
- Update optimizer CLI tests to cover the new behavior for non-local calculators.

## Why
The previous rule blocked extension scenarios (for example custom calculators that accept file/payload inputs) at CLI validation time.

## Validation
- `./gradlew :maintenance:optimizer:test --tests "org.apache.gravitino.maintenance.optimizer.TestOptimizerCmd"`

### DIFF
--- a/maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/OptimizerCmd.java
+++ b/maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/OptimizerCmd.java
@@ -326,12 +326,6 @@ public class OptimizerCmd {
             "Command '%s' with --calculator-name %s requires one of --statistics-payload "
                 + "or --file-path.",
             LOCAL_STATS_CALCULATOR_NAME)
-        .addForbidWhenOption(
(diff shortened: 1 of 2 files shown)


FANNG1:
@roryqi PTAL

roryqi (approved):

roryqi:
Do we need to update the document, too?

FANNG1:
Good point. For this PR, I think a doc update is not required.\n\nReason: this change only relaxes CLI validation for non-local calculators, and does not change the existing behavior/documented usage of local-stats-calculator (it still requires one of --statistics-payload or --file-path).\n\nSo user-facing default workflow remains the same; this mainly unblocks extension scenarios for custom calculators.

Issue #10365, opened 2026-03-11
[Bugfix]Hologres DROP COLUMN operation fails without GUC parameter


TimothyDing:
## Description
When attempting to drop a column from a Hologres table via Gravitino, the operation fails because Hologres requires a session-level GUC parameter to be set before executing DROP COLUMN.

According to Hologres documentation (https://help.aliyun.com/zh/hologres/developer-reference/alter-table), the following GUC must be enabled:
```sql
SET hg_experimental_enable_drop_column = on;
```

Without this setting, the ALTER TABLE DROP COLUMN statement will fail.

## Steps to Reproduce
1. Create a Hologres table via Gravitino
2. Attempt to drop a column using table alteration
3. Operation fails due to missing GUC parameter

## Expected Behavior
DROP COLUMN operation should succeed by automatically setting the required GUC parameter.

## Actual Behavior
DROP COLUMN operation fails because the GUC parameter is not set.

### Parent issue

https://github.com/apache/gravitino/issues/9886

PR #10366, opened 2026-03-11
[#10365] fix(hologres): Add GUC parameter for DROP COLUMN operation

TimothyDing:

This PR fixes the Hologres DROP COLUMN operation by adding the required session-level GUC parameter before executing the ALTER TABLE DROP COLUMN statement.

Changes:
1. Added constant `HOLOGRES_ENABLE_DROP_COLUMN_GUC` for the GUC parameter
2. Modified `deleteColumnFieldDefinition()` method to prepend the GUC setting to the DROP COLUMN SQL
3. Added unit tests for DROP COLUMN functionality:
   - `testAlterTableDeleteColumn` - basic DROP COLUMN test
   - `testAlterTableDeleteColumnIfExists` - IF EXISTS option test
   - `testAlterTableDeleteColumnNotExistsThrows` - error handling test
   - `testAlterTableDeleteColumnRejectsNestedField` - nested field rejection test

The generated SQL now looks like:
```sql
SET hg_experimental_enable_drop_column = on;
ALTER TABLE "table_name" DROP COLUMN "column_name";
```

Hologres requires the `hg_experimental_enable_drop_column` GUC parameter to be enabled at the session level before executing DROP COLUMN operations. Without this setting, the operation fails.

Reference: https://help.aliyun.com/zh/hologres/developer-reference/alter-table

Fix: https://github.com/apache/gravitino/issues/10365

## Does this PR introduce _any_ user-facing change?

No. The change is internal to the Hologres catalog implementation. Users can now successfully drop columns from Hologres tables without any API changes.

## How was this patch tested?

- Added 4 new unit tests in `TestHologresTableOperations.java`
- All existing tests pass
- Tests verify:
  - Correct SQL generation with GUC parameter
  - IF EXISTS option behavior
  - Error handling for non-existent columns
  - Rejection of nested column names

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,3 @@ web/.next
 
 .venv/
 .worktrees/
-
-# Agent
-.qoder/


catalogs-contrib/catalog-jdbc-hologres/src/main/java/org/apache/gravitino/catalog/hologres/operation/HologresTableOperations.java
```diff
@@ -78,6 +78,10 @@ public class HologresTableOperations extends JdbcTableOperations
   private static final String HOLOGRES_NOT_SUPPORT_NESTED_COLUMN_MSG =
       "Hologres does not support nested column names.";
 
+  /** GUC parameter to enable DROP COLUMN functionality in Hologres. */
```
yuqi1129, line 81: What's the meaning of `GUC`?

catalogs-contrib/catalog-jdbc-hologres/src/main/java/org/apache/gravitino/catalog/hologres/operation/HologresTableOperations.java
```diff
@@ -531,9 +535,15 @@ private String deleteColumnFieldDefinition(
...
       }
     }
+    // Hologres requires enabling the GUC parameter before dropping a column.
+    // Reference: https://help.aliyun.com/zh/hologres/developer-reference/alter-table
     return String.format(
-        "%s%s DROP COLUMN %s;",
-        ALTER_TABLE, quoteIdentifier(table.name()), quoteIdentifier(deleteColumn.fieldName()[0]));
+        "%s%s%s%s DROP COLUMN %s;",
+        HOLOGRES_ENABLE_DROP_COLUMN_GUC,
+        NEW_LINE,
```
yuqi1129: I think there is no need to add `NEW_LINE` here. as  SQL like `sentence one; sentence two; ....; ...` seems to be more straight forward.

catalogs-contrib/catalog-jdbc-hologres/src/main/java/org/apache/gravitino/catalog/hologres/operation/HologresTableOperations.java
TimothyHologres: Hologres is compatible with the PostgreSQL protocol. Within this protocol, GUC (Grand Unified Configuration) is the collective term for all configurable parameters in a PostgreSQL database. Whether it involves memory allocation, connection limits, log levels, or query optimizer switches, any setting that can be adjusted via configuration files, command-line arguments, or SQL commands is referred to as a GUC parameter.
The purpose of this GUC is to allow users to execute DROP COLUMN operations.

catalogs-contrib/catalog-jdbc-hologres/src/main/java/org/apache/gravitino/catalog/hologres/operation/HologresTableOperations.java
TimothyHologres: fix in 1ecc4f48968e89284689afcd7aa65269f92e8177

TimothyHologres:
@yuqi1129 fix in 1ecc4f48968e89284689afcd7aa65269f92e8177.Please review

Issue #10262, opened 2026-03-06
[Improvement] remove ./gradlew release


FANNG1:
### What would you like to be improved?

currently we use gradle build to build packages with JDK17 and gradle release to build spark&Flink packages with JDK8 compatibility,  it's hard to maintaince and use. 

### How should we improve?

Use gradle build to build spark&flink with JDK8 compatibility and remove gradlew release

FANNG1:
we may need following code in the build.gradle.kts in the modules which are compatibility with JDK8
```
tasks.named<JavaCompile>("compileTestJava") {
  // client-java main artifact targets Java 8; tests depend on modules that publish Java 17 variants.
  // Compile tests with release 17 to make Gradle variant matching for :core/:server test classpath deterministic.
  options.release.set(17)
}

val targetJvmVersionAttribute = TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE
configurations.named("testCompileClasspath") {
  attributes.attribute(targetJvmVersionAttribute, 17)
}
configurations.named("testRuntimeClasspath") {
  attributes.attribute(targetJvmVersionAttribute, 17)
}

tasks.test {
  javaLauncher.set(
    javaToolchains.launcherFor {
      languageVersion.set(JavaLanguageVersion.of(17))
    }
  )

```

PR #10385, opened 2026-03-11
[#10262] refactor(build): remove release task and centralize JDK8 compatibility

FANNG1:

This PR removes the dedicated `release` Gradle task path and centralizes JDK8 compatibility handling in the root build configuration.

1. Removed the root `release` task from `build.gradle.kts`.
2. Updated `dev/release/release-build.sh` to use `assemble` instead of `release` for the publish preflight.
3. Updated the lightweight CI compile check to use `./gradlew assemble`.
4. Centralized JDK8-compatible project matching logic in the root `build.gradle.kts`.
5. Simplified the JDK8-compatible module selection into a single project-path-prefix rule.
6. Configured all JDK8-compatible projects to compile `main` sources with `--release 8`.
7. Configured tests for those projects to resolve and compile with JDK 17 by setting `compileTestJava --release 17` and `TargetJvmVersion=17` on test classpaths.
8. Removed duplicated test JVM compatibility wiring from `clients/client-java/build.gradle.kts`.
9. Set `JavaCompile` encoding to `UTF-8` in the root build so existing non-ASCII sources compile consistently across environments.

Previously, producing JDK8-compatible artifacts depended on a separate `./gradlew release` path, and the compatibility logic was split across the root build and individual modules.

This change makes the build behavior more consistent by:

- removing the dedicated `release` task path,
- using standard Gradle `build`/`assemble` flows instead of a custom release lifecycle,
- keeping JDK8 compatibility decisions in one place,
- compiling JDK8-targeted main artifacts consistently,
- and letting tests continue to compile and run against JDK17 dependencies where needed.

Fix: #10262

No user-facing API or property-key changes.

This is a build-system/internal compatibility cleanup.

Ran targeted verification for the centralized JDK8/JDK17 build behavior:

1. `./gradlew :clients:client-java:printJvm :clients:cli:printJvm :clients:filesystem-hadoop3:printJvm :maintenance:jobs:printJvm :spark-connector:spark-common:printJvm :flink-connector:flink:printJvm -PskipITs`
2. `./gradlew :clients:filesystem-hadoop3:compileTestJava :maintenance:jobs:compileTestJava :clients:client-java:compileTestJava :spark-connector:spark-common:compileTestJava :flink-connector:flink:compileTestJava -PskipITs --console=plain`
3. `./gradlew :api:compileJava :common:compileJava :clients:cli:compileJava :clients:client-java:compileTestJava :spark-connector:spark-common:compileJava :spark-connector:spark-common:compileTestJava -PskipITs --console=plain`
4. `./gradlew :spark-connector:spark-runtime-3.5:shadowJar --console=plain`
5. `./gradlew spotlessKotlinGradleCheck --console=plain`
6. `./gradlew :api:compileJava :clients:client-java:compileTestJava -PskipITs --console=plain`

Also manually verified:

- `:api:compileJava` rejects a temporary Java 9 API usage at compile time under the JDK8 target.
- The built Spark 3.5 runtime shadow jar has base class-file version 52 (Java 8) on its main classpath entries.
- There are no remaining `./gradlew release` or `gradlew release` task invocations in the repository.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -406,10 +406,12 @@ subprojects {
   }
 
   if (compatibleWithJDK8(project)) {
+    // Keep published/main classes Java 8-compatible for the selected modules.
     tasks.named<JavaCompile>("compileJava") {
       options.release.set(8)
     }
 
+    // Tests still need Java 17 to compile against dependencies that only publish Java 17 variants.
     tasks.named<JavaCompile>("compileTestJava") {
       options.release.set(17)
     }


yuqi1129:
It seems that #10376 is also doing the same thing.

yuqi1129:
https://github.com/apache/gravitino/pull/10376 has been merged. Based on the discussion offline, you can continue with this one.

FANNG1:
@yuqi1129 @jerryshao PTAL

.github/workflows/build.yml
```diff
@@ -84,7 +84,7 @@ jobs:
 
       - name: Build with Gradle
         run: |
-          ./gradlew release -x test
+          ./gradlew build -x test
```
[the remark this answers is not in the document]
FANNG1: Agreed. `compile-check` is intended to stay lightweight, and `build -x test` would expand into non-Java subproject build paths such as `mcp-server`, `client-python`, and docs. I switched this job to `./gradlew assemble` to keep it compile-focused.

dev/release/release-build.sh
```diff
@@ -348,8 +348,8 @@ if [[ "$1" == "publish-release" ]]; then
   cd ..
 
   $GRADLE clean
-  $GRADLE release -x test -PdefaultScalaVersion=2.12
-  $GRADLE release -x test -PdefaultScalaVersion=2.13
+  $GRADLE build -x test -PdefaultScalaVersion=2.12
+  $GRADLE build -x test -PdefaultScalaVersion=2.13
```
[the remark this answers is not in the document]
FANNG1: Agreed. `publish-release` only needs a pre-publish artifact assembly before `publishToMavenLocal`, and `build -x test` was broader than the previous `release` path. I switched the two preflight invocations to `assemble` to avoid unrelated non-Java build steps.

.github/workflows/build.yml
```diff
@@ -84,7 +84,7 @@ jobs:
 
       - name: Build with Gradle
         run: |
-          ./gradlew release -x test
+          ./gradlew assemble
```
yuqi1129, line 87: What's the difference between `./gradlew assemble` and `./gradlew build`?

build.gradle.kts
```diff
@@ -331,39 +332,24 @@ subprojects {
     mavenLocal()
   }
 
+  val jdk8CompatibleProjectNames = setOf("api", "common", "catalog-common", "hadoop-common")
```
yuqi1129: Can we merge these two lines

build.gradle.kts
```diff
@@ -450,6 +446,7 @@ subprojects {
   }
 
   tasks.withType<JavaCompile>().configureEach {
+    options.encoding = "UTF-8"
```
yuqi1129, line 451: Why do we add this line? What will happen if it's absent?

FANNG1: This line keeps Java compilation independent of the host default charset. Without it, modules that already contain non-ASCII source text can fail to compile on machines whose default encoding is not UTF-8. I kept it and added a short comment to make the reason explicit.

build.gradle.kts
FANNG1: Done. I merged the exact-name and prefix matching into a single `jdk8CompatibleProjectPathPrefixes` set and simplified `compatibleWithJDK8` to match by project path prefix only.

.github/workflows/build.yml
FANNG1: `assemble` only builds/assembles artifacts, while `build` includes both `assemble` and `check`. In this repo, several non-Java tasks are wired to `build` but not to `assemble` (for example `mcp-server`, `client-python`, and docs), so `assemble` keeps `compile-check` lightweight.

yuqi1129:
@FANNG1 
We may also need to remove related content in the Gravitino document.

FANNG1:
@yuqi1129 I checked the repository docs and did not find any remaining references to the removed Gradle `release` task. The remaining `publish-release` mentions are script subcommands in the release scripts, not the old Gradle task. I also refreshed the PR description to match the final `assemble` behavior.

FANNG1:
@jerryshao @diqiu50 do you want to take another look?

build.gradle.kts
```diff
@@ -426,6 +405,22 @@ subprojects {
     }
   }
 
+  if (compatibleWithJDK8(project)) {
+    tasks.named<JavaCompile>("compileJava") {
+      options.release.set(8)
+    }
+
```
diqiu50, line 413: Please add comments for explain the reason

.github/workflows/build.yml
diqiu50: Is `./gradlew build -x text` better?

dev/release/release-build.sh
```diff
@@ -348,8 +348,8 @@ if [[ "$1" == "publish-release" ]]; then
   cd ..
 
   $GRADLE clean
-  $GRADLE release -x test -PdefaultScalaVersion=2.12
-  $GRADLE release -x test -PdefaultScalaVersion=2.13
+  $GRADLE assemble -PdefaultScalaVersion=2.12
+  $GRADLE assemble -PdefaultScalaVersion=2.13
```
diqiu50, line 352: build -x test

build.gradle.kts
FANNG1: Added short comments here. The key reason is that these modules still need to publish Java 8-compatible main artifacts, while their tests must compile against dependencies that only expose Java 17 variants.

.github/workflows/build.yml
FANNG1: I do not think `build -x test` is better here. `compile-check` is intended to stay lightweight, and `build -x test` would still enter the full `build` lifecycle for all subprojects, including non-Java build paths such as `mcp-server`, `client-python`, and docs. `assemble` keeps this job focused on artifact assembly/compilation only.

dev/release/release-build.sh
FANNG1: I prefer to keep `assemble` here as well. In `publish-release`, this step is only a preflight to make sure the publishable artifacts can be assembled before `publishToMavenLocal`. Using `build -x test` would broaden the scope and pull in unrelated non-Java `build` tasks again.

FANNG1:
@diqiu50 @jerryshao do you want to take another look?

.github/workflows/build.yml
diqiu50: ./gradlew build -x test will still run checks such as license verification and formatting (e.g., Spotless), while skipping the test task.
In practice, using build is more common and aligns better with standard workflows.

FANNG1: Agreed that `build -x test` would run broader checks such as Spotless/license verification. In this workflow, though, those checks are already covered by the main `build` job, which still runs Gradle `build`. The `compile-check` job is intended to stay as a lightweight fallback path, so I prefer to keep `assemble` here to avoid pulling extra non-Java subproject `build` paths into that fallback job.

diqiu50: if not used to run `build -x test` here, where will those checks be performed?

FANNG1: Those checks are performed in the main `build` job of the `build` workflow, which still runs Gradle `build` (with some targeted exclusions for non-Java subprojects). `compile-check` is only the lightweight fallback path used when `source_changes != true`, so I kept `assemble` there to avoid pulling the broader non-Java `build` lifecycle into that fallback job.

Issue #10174, opened 2026-03-04
[Improvement] Possible NPE for null lastVersion in fileset version update


justinmclean:
### What would you like to be improved?

POConverters.updateFilesetPOWithVersion reads Long lastVersion = oldFilesetPO.getLastVersion(); and then executes lastVersion++ when needUpdateVersion is true. If lastVersion is null, Java auto-unboxing triggers a NullPointerException.

### How should we improve?

Handle nullable version state before incrementing. In updateFilesetPOWithVersion, validate oldFilesetPO.getLastVersion() and either:
- throw a clear exception before incrementing, or
- normalize null to a safe baseline version (for example INIT_VERSION) before incrementing.

Here's a unit test to help:
```
@Test
public void testUpdateFilesetPOVersionWithNullLastVersion() throws Exception {
  FilesetEntity filesetEntity =
      createFileset(
          1L,
          "test",
          NamespaceUtil.ofFileset("test_metalake", "test_catalog", "test_schema"),
          "this is test",
          "hdfs://localhost/test",
          ImmutableMap.of("key", "value"));
  FilesetPO.Builder builder =
      FilesetPO.builder().withMetalakeId(1L).withCatalogId(1L).withSchemaId(1L);
  FilesetPO initPO = POConverters.initializeFilesetPOWithVersion(filesetEntity, builder);

  Field lastVersionField = FilesetPO.class.getDeclaredField("lastVersion");
  lastVersionField.setAccessible(true);
  lastVersionField.set(initPO, null);

  Assertions.assertDoesNotThrow(
      () -> POConverters.updateFilesetPOWithVersion(initPO, filesetEntity, true));
}

```

pandeysambhi:
Can I work on this?

PR #10389, opened 2026-03-11
[#10174] Improvement(POConverters): possible NPE for null lastVersion in fileset version update

pandeysambhi:
In updateFilesetPOWithVersion, validate oldFilesetPO.getLastVersion() and normalize null to a safe baseline version ( INIT_VERSION) before incrementing to avoid NPE.

POConverters.updateFilesetPOWithVersion reads Long lastVersion = oldFilesetPO.getLastVersion(); and then executes lastVersion++ when needUpdateVersion is true. If lastVersion is null, Java auto-unboxing triggers a NullPointerException.

Fix: (#10174 )

no

unit test

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
@@ -702,6 +702,9 @@ public class POConverters {
       FilesetPO oldFilesetPO, FilesetEntity newFileset, boolean needUpdateVersion) {
     try {
       Long lastVersion = oldFilesetPO.getLastVersion();
+      if (lastVersion == null) {
+        lastVersion = INIT_VERSION;
+      }
       Long currentVersion;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/utils/POConverters.java
```diff
@@ -702,6 +702,9 @@ public static FilesetPO updateFilesetPOWithVersion(
       FilesetPO oldFilesetPO, FilesetEntity newFileset, boolean needUpdateVersion) {
     try {
       Long lastVersion = oldFilesetPO.getLastVersion();
```
jerryshao, line 704: In which scenario will the return value here be null?

jerryshao:
I don't think this is a valid case mentioned in the issue. I'm going to close this PR and issue. You can always reopen this PR and issue if you find this problem in the real scenario. Thanks @pandeysambhi for your contribution.

Issue #10325, opened 2026-03-10
[Improvement] Fix schema metadata loss in SchemaOperationDispatcher.alterSchema entity-store update path


justinmclean:
### What would you like to be improved?

When altering a schema on the non-managed path, SchemaOperationDispatcher updates the stored SchemaEntity by building a new entity with only id, name, namespace, and auditInfo. Persisted comment and properties are omitted, and storage update semantics write those fields from the new object, which can null/erase previously stored schema metadata. This is reproducible by creating a schema with comment/properties, altering it, then reading the stored SchemaEntity.

### How should we improve?

In the updater lambda inside alterSchema, preserve metadata fields when building the updated entity. At minimum, carry forward and/or refresh:
- comment (e.g., from current entity or altered schema)
- properties (e.g., from alteredSchema.properties())

Here a unit test to help:
```
   @Test
  public void testAlterSchemaKeepsStoredCommentAndProperties() throws IOException {
    NameIdentifier schemaIdent = NameIdentifier.of(metalake, catalog, "schema_preserve_metadata");
    Map<String, String> props = ImmutableMap.of("k1", "v1", "k2", "v2");
    dispatcher.createSchema(schemaIdent, "comment", props);

    SchemaChange[] changes =
        new SchemaChange[] {
          SchemaChange.setProperty("k3", "v3"), SchemaChange.removeProperty("k1")
        };
    dispatcher.alterSchema(schemaIdent, changes);

    SchemaEntity storedSchema = entityStore.get(schemaIdent, SCHEMA, SchemaEntity.class);
    Assertions.assertEquals("comment", storedSchema.comment());
    Assertions.assertEquals(ImmutableMap.of("k2", "v2", "k3", "v3"), storedSchema.properties());
  }
```

Lucas61000:
Hi, I'm interested in working on this. Would you mind assigning it to me?

PR #10393, opened 2026-03-12
[#10325] improvement: Fix schema metadata loss in SchemaOperationDispatcher.alterSchema entity-store update path

Lucas61000:

- Fixed schema metadata loss during alteration by preserving comment and properties in `SchemaOperationDispatcher.alterSchema` method
Fix: #10325 

Added `testAlterSchemaKeepsStoredCommentAndProperties` and`testAlterSchemaMultipleTimesKeepsMetadata`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
@@ -274,7 +274,7 @@ public class SchemaOperationDispatcher extends OperationDispatcher implements Sc
                                   .withId(schemaEntity.id())
                                   .withName(schemaEntity.name())
                                   .withNamespace(ident.namespace())
-                                  .withComment(schemaEntity.comment())
+                                  .withComment(alteredSchema.comment())
                                   .withProperties(
                                       StringIdentifier.newPropertiesWithoutId(
                                           alteredSchema.properties()))


Abyss-lord (approved):
LGTM, @justinmclean @yuqi1129 WDYT

core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
```diff
@@ -274,6 +274,10 @@ public Schema alterSchema(NameIdentifier ident, SchemaChange... changes)
                                   .withId(schemaEntity.id())
                                   .withName(schemaEntity.name())
                                   .withNamespace(ident.namespace())
+                                  .withComment(alteredSchema.comment())
+                                  .withProperties(
```
yuqi1129, line 278: It seems that this is by design, `comment` and `properties` will be provided by `alteredSchema` in `EntityCombinedSchema`. This is to say: these two values are decided by corresponding values in the underlying storage, not the Gravitino store.

@jerryshao 
Can you help to verify it?

jerryshao: Yes, for the unmanaged schema entity, we only need to store the id / name. Other fields will be fetched from external data sources.

Abyss-lord: https://github.com/Lucas61000/gravitino/blob/02d1d373de5a6f876d394f0bab3cd93b93a697b3/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java#L150
IMHO, when creating a schema, we should also retrieve comments and properties from external sources and then persist them. 
Should we open an additional issue to fix this problem?

jerryshao: It's not a bug. Even if we store these two fields, we will not use them. We will always honor the one from external data sources for the unmanaged schema.

jerryshao:
Why do we merge this PR? We haven't yet gotten a consensus on this.

Abyss-lord:
> Why do we merge this PR? We haven't yet gotten a consensus on this.

Sorry about that — I saw that yuqi1129 hadn’t replied, so I assumed we had reached consensus and went ahead with the merge. Should we revert the change for now?

jerryshao:
Please create a revert PR.

jerryshao:
Thanks @Abyss-lord , my principle is that we don't obsessively change the code if there's no issue or the issue cannot be reproduced in the real scenario.

Abyss-lord:
> Thanks @Abyss-lord , my principle is that we don't obsessively change the code if there's no issue or the issue cannot be reproduced in the real scenario.

Thanks, @jerryshao — understood. I agree with that principle and will be more careful in the future.

Issue #10397, opened 2026-03-12
[Bug report] gravitino : list table get wrong namespace name


thankpk:
### Version

main branch

### Describe what's wrong

i curl api list tables and get wrong namespace . api returns metalake name(gravitino)  instead of namespace name (test_schema2): 

### Error message and/or stacktrace

when i curl api: curl 'http://gravitino:9001/iceberg/v1/iceberg_hdfs/namespaces/test_schema2/tables'  and got result : 
Output: {"identifiers":[{"namespace":["grativino"],"name":"hdfs_test_table"}],"next-page-token":null}

My expect Output: {"identifiers":[{"namespace":["test_schema2"],"name":"hdfs_test_table"}],"next-page-token":null}

i try to log and realize that the code of func ListTablesResponse have need to change logic get level . this code should get level 2 instead of level 0 (metalake name):  
filteredIdentifiers.add(
          TableIdentifier.of(Namespace.of(ident.namespace().level(0)), ident.name())); 
=> need to change to: filteredIdentifiers.add(
          TableIdentifier.of(Namespace.of(ident.namespace().level(2)), ident.name())); 

### How to reproduce

i use version v1.1.0 (hash: https://github.com/apache/gravitino/commit/5a6b5ae772d50aff98878ae3659fba3598a9027f)

User trino: 477   

when user call query: show tables  (in dbeaver,...)

### Additional context

**this is code i log:**  

private ListTablesResponse filterListTablesResponse(
      ListTablesResponse listTablesResponse, String metalake, String catalogName) {

    LOG.info(
        "List Iceberg tables, catalog 0000000000000000000 aaaaaaaaaaaaaaaa metalake: {}, catalogName: {}",
        metalake,
        catalogName);

    NameIdentifier[] idents =
        MetadataAuthzHelper.filterByExpression(
            metalake,
            AuthorizationExpressionConstants.filterTableAuthorizationExpression,
            Entity.EntityType.TABLE,
            toNameIdentifiers(listTablesResponse, metalake, catalogName));

    LOG.info("List Iceberg tables, catalog 111111111111111111111111: {}", Arrays.toString(idents));

    List<TableIdentifier> filteredIdentifiers = new ArrayList<>();
    for (NameIdentifier ident : idents) {
      LOG.info(
          "List Iceberg tables, catalog 222222222222222222 aaaaaaaaaaaaaaaaaaaa: Indent: {}, IndentNamespace: {}, Name: {}",
          ident,
          ident.namespace(),
[...]

PR #10398, opened 2026-03-12
[#10397] fix(iceberg): Fix wrong namespaces when listing tables or views

roryqi:

 Fix wrong namespaces when listing tables or views

Fix: #10397

No.

Added test cases.

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
@@ -552,7 +552,7 @@ public class IcebergTableOperations {
     List<TableIdentifier> filteredIdentifiers = new ArrayList<>();
     for (NameIdentifier ident : idents) {
       filteredIdentifiers.add(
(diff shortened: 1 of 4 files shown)


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -552,7 +552,7 @@ private ListTablesResponse filterListTablesResponse(
     List<TableIdentifier> filteredIdentifiers = new ArrayList<>();
     for (NameIdentifier ident : idents) {
       filteredIdentifiers.add(
-          TableIdentifier.of(Namespace.of(ident.namespace().level(0)), ident.name()));
+          TableIdentifier.of(Namespace.of(ident.namespace().level(2)), ident.name()));
     }
```
diqiu50, line 556: Will this cause an IndexOutOfBounds exception?

roryqi: No. table identifiers must have the namespace.

PR #10406, opened 2026-03-12
[MINOR] test: Validate maintenance build skip for web-only changes

FANNG1:

This validation PR applies the maintenance build CI change and adds a web-only trigger file to verify that maintenance tests are skipped for unrelated module changes.

This is a temporary CI validation PR to confirm that maintenance tests are not scheduled when only web files change.

Fix: #(none, validation only)

No. This PR is only for CI validation.

Locally validated with Gradle dry-run. GitHub Actions on this draft PR is used to verify that the generated build command includes maintenance test exclusions.

### DIFF
--- a/web/ci-maintenance-skip-web-20260312.properties
+++ b/web/ci-maintenance-skip-web-20260312.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


FANNG1:
Validation complete on March 12, 2026. This temporary PR verified the skip case: changes job reported maintenance_changes=false, and the build job printed a Gradle command that included all four maintenance test exclusions. The run was then cancelled after command emission because this PR existed only to validate CI routing.

PR #10407, opened 2026-03-12
[MINOR] test: Validate maintenance build include for maintenance changes

FANNG1:

This validation PR applies the maintenance build CI change and adds a maintenance-only trigger file to verify that maintenance tests still run when the maintenance module changes.

This is a temporary CI validation PR to confirm that the new skip logic does not suppress maintenance tests for maintenance changes.

Fix: #(none, validation only)

No. This PR is only for CI validation.

Locally validated with Gradle dry-run. GitHub Actions on this draft PR is used to verify that the generated build command does not include maintenance test exclusions.

### DIFF
--- a/maintenance/ci-maintenance-self-20260312.properties
+++ b/maintenance/ci-maintenance-self-20260312.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


FANNG1:
Validation complete on March 12, 2026. This temporary PR verified the maintenance self-change case: changes job reported maintenance_changes=true, and the build job printed a Gradle command without maintenance test exclusions. The log also showed maintenance tasks being scheduled. The run was then cancelled after command emission because this PR existed only to validate CI routing.

Issue #10115, opened 2026-03-02
[Subtask] optimizer maintaince module CI


FANNG1:
### Describe the subtask

Run maintenance test only when the maintenance and dependence module code are changed

### Parent issue

#10360

PR #10424, opened 2026-03-12
[#10115] improvement(ci): Optimize build workflow for maintenance changes

FANNG1:

This PR updates the `build` workflow to handle maintenance changes without introducing a separate maintenance workflow.

It adds maintenance-aware routing in the existing `build` job:
- pure `maintenance/**` source changes run only the targeted maintenance module build path
- mixed changes still run the normal root build and then run maintenance unit tests
- upstream dependency changes for maintenance also keep maintenance validation enabled

Maintenance-only changes do not need the full root build and release flow.
This change reduces unnecessary CI work for those changes while keeping maintenance validation enabled when maintenance sources or their dependencies are affected.

Fix: #10115

No.

Verified locally with YAML validation and Gradle dry-run for:
- the maintenance-only build path
- the normal root build path with maintenance tests excluded
- the maintenance unit test path

Validated with temporary upstream CI runs for:
- maintenance-only changes
- maintenance and common mixed changes
- common-only changes
- web-only changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -187,7 +187,7 @@ jobs:
           if [ "${{ needs.changes.outputs.maintenance_module_only_changes }}" = "true" ]; then
             ./gradlew \
               :maintenance:optimizer-api:build \
-              :maintenance:gravitino-updaters:build \
+              :maintenance:updaters:build \
               :maintenance:optimizer:build \
               :maintenance:jobs:build \
               -PskipITs \


.github/workflows/build.yml
```diff
@@ -64,10 +65,52 @@ jobs:
...
+              + "\n"
+            )
+          PY
     outputs:
       source_changes: ${{ steps.filter.outputs.source_changes }}
       spark_connector_changes: ${{ steps.filter.outputs.spark_connector_changes }}
       mcp_server_changes: ${{ steps.filter.outputs.mcp_server_changes }}
+      maintenance_source_changes: ${{ steps.filter.outputs.maintenance_source_changes }}
+      maintenance_changes: ${{ steps.filter.outputs.maintenance_changes }}
+      maintenance_only_source_changes: ${{ steps.maintenance_only.outputs.maintenance_only_source_changes }}
```
jerryshao: What's the difference between these 3 different changes, I'm so confused about this? Can you use a better name to clarify different usages? Also, please add the comments to explain different usage scenarios.

yuqi1129:
I understand the necessity of this PR, and it will accelerate CI in most cases. The rule is quite complicated and a bit hard to maintain, though AI can help alleviate this problem to a certain degree.

.github/workflows/build.yml
FANNG1: You're right. The previous names were too close and made the usage of each flag hard to understand.

I updated the workflow to:
- rename `maintenance_source_changes` to `maintenance_module_changes`
- rename `maintenance_changes` to `maintenance_related_changes`
- add comments to explain the three different scenarios

We still need three separate flags because they answer different routing questions:
- `maintenance_module_changes`: whether the PR directly touches `maintenance/**`
- `maintenance_related_changes`: whether the PR can affect maintenance validation, including upstream dependencies and shared Gradle files
- `maintenance_only_source_changes`: whether all source changes in the PR are under `maintenance/**`, which is used to choose the maintenance-only build path

Using a single flag here would either skip maintenance validation when upstream dependencies change, or incorrectly skip the normal root build for mixed changes.

FANNG1:
Follow-up update: I simplified the routing logic in the latest change.

What changed:
- remove the extra maintenance-related routing flags
- keep a single routing flag: `maintenance_only_changes`
- route only pure `maintenance/**` source changes to the maintenance-only build path
- route everything else to the normal full build flow

Why I changed it:
- the previous version optimized more cases, but the routing logic was harder to understand and maintain
- this version keeps the main goal of the PR, which is avoiding the full build when the PR only changes maintenance sources
- the workflow is now easier to reason about because the rule becomes a simple binary decision: maintenance-only vs full build

Trade-off:
- we give up some fine-grained optimization
- for example, if an upstream dependency of maintenance changes, we now run the full build instead of trying to selectively route maintenance validation
- in return, the workflow becomes much simpler, and the maintenance cost of the CI rule is lower

So this change intentionally favors readability and maintainability over squeezing out every possible CI optimization case.

.github/workflows/build.yml
FANNG1: Follow-up update: this part has been simplified in the latest change, so the previous explanation is outdated now.

We no longer keep three different maintenance change flags in the workflow. The latest version keeps only one routing flag: `maintenance_only_changes`.

Its meaning is now straightforward:
- `maintenance_only_changes = true`: all effective source changes in the PR are under `maintenance/**`, so we run only the targeted maintenance build/test path
- otherwise: we run the normal full build flow

I also kept a comment in the workflow to explain that this flag is derived from `source_changes_files`, so non-source changes such as markdown files do not break the maintenance-only path.

The reason for this change is maintainability. The previous version optimized more cases, but the routing logic was harder to understand. This version intentionally trades some fine-grained CI optimization for a much simpler and easier-to-maintain rule.

FANNG1:
@jerryshao @yuqi1129 PTAL

.github/workflows/build.yml
```diff
@@ -64,10 +65,34 @@ jobs:
               - spark-connector/**
             mcp_server_changes:
               - mcp-server/**
+      - name: Determine maintenance-only changes
```
jerryshao: Can you change to "maintenance-module-only"? It will be clearer.

.github/workflows/build.yml
```diff
@@ -64,10 +65,34 @@ jobs:
...
+            path.startswith("maintenance/") for path in source_files
+          )
+
+          with open(os.environ["GITHUB_OUTPUT"], "a", encoding="utf-8") as output:
+            output.write(
+              "maintenance_only_changes="
+              + ("true" if maintenance_only else "false")
+              + "\n"
+            )
+          PY
```
jerryshao: Why do we use Python to check? Is this the only option?

.github/workflows/build.yml
```diff
@@ -153,12 +178,39 @@ jobs:
           sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb  
 
       - name: Build with Gradle
-        if: needs.changes.outputs.mcp_server_changes == 'true'
-        run: ./gradlew build -PskipITs -PskipDockerTests=false -x :clients:client-python:build
+        run: |
+          if [ "${{ needs.changes.outputs.maintenance_only_changes }}" = "true" ]; then
+            ./gradlew \
+              :maintenance:optimizer-api:build \
+              :maintenance:gravitino-updaters:build \
```
jerryshao: This module should also be updated.

.github/workflows/build.yml
```diff
@@ -64,10 +65,34 @@ jobs:
               - spark-connector/**
             mcp_server_changes:
               - mcp-server/**
+      - name: Determine maintenance-only changes
+        id: maintenance_only
```
jerryshao: Also here.

.github/workflows/build.yml
FANNG1: Good catch. This was a stale module path after the maintenance updater module was renamed.

The workflow was still using `:maintenance:gravitino-updaters:build`, while the current project path is `:maintenance:updaters`.

I have updated it to `:maintenance:updaters:build` in the latest change.

.github/workflows/build.yml
FANNG1: Yes, agreed. `maintenance-module-only` is clearer than the previous name because the routing only applies when the effective source changes are under `maintenance/**`.

I renamed the flag to `maintenance_module_only_changes` in the latest update.

.github/workflows/build.yml
FANNG1: Not required. I originally used Python because `paths-filter` was outputting JSON, but shell is enough for this check.

In the latest update I changed `list-files` to `shell` and rewrote this step as a small shell loop, which keeps the workflow simpler and avoids pulling in Python just for this routing check.

.github/workflows/build.yml
FANNG1: Yes. This output and the corresponding `needs.changes.outputs` references were renamed together in the latest update, so they now consistently use `maintenance_module_only_changes`.

Issue #10436, opened 2026-03-14
[Bug report] V2 Web UI identity mismatch with server-side serviceAdmins check


gada121982:
**Version:** main branch (tested on v1.2.0)

## Describe what's wrong

The V2 Web UI uses `profile.name` (OIDC fullname) for `serviceAdmins` authorization check, while the server uses `preferred_username` or `sub` (via `principalFields` config). This causes the "Create Metalake" button to be hidden even when the user is a valid service admin on the server side.

**Example:**
- Keycloak user: `admin` (firstName=`admin`, lastName=`admin`)
- Server extracts principal via `principalFields: "preferred_username,sub"` → `"admin"`
- UI gets identity via `OidcOAuthProvider.getUserProfile()` → `profile.name` = `"admin admin"` (fullname per OIDC spec)
- UI checks `serviceAdmins.split(',').includes(authUser?.name)` → `"admin admin"` ≠ `"admin"` → button hidden

**Workaround:** Add both values to `serviceAdmins` (e.g., `"admin,admin admin"`). This is fragile.

Additionally, `OidcOAuthProvider.getAccessToken()` returns `id_token || access_token`, preferring ID token over access token. Per OAuth2 spec, access token should be sent to resource servers.

## Error message and/or stacktrace

No error — the button is silently hidden. Server-side authorization works correctly with `preferred_username`, confirming the mismatch is UI-only.

## How to reproduce

1. Deploy Gravitino with V2 Web UI (`GRAVITINO_USE_WEB_V2: true`)
2. Configure OAuth with OIDC provider (e.g., Keycloak):
   ```properties
   gravitino.authenticator = oauth
   gravitino.authenticator.oauth.provider = oidc
   gravitino.authenticator.oauth.principalFields = preferred_username,sub
   gravitino.authorization.enable = true
   gravitino.authorization.serviceAdmins = admin
   ```
3. Create an IdP user with `username=admin`, `firstName=admin`, `lastName=admin`
4. Login via OIDC flow
5. "Create Metalake" button is **hidden** despite `admin` being in `serviceAdmins`
6. Add `"admin admin"` (fullname) to `serviceAdmins` → button appears

## Root cause in code

- `web-v2/web/src/lib/auth/providers/oidc.js:134` — `getUserProfile()` returns `user.profile` where `name` = fullname (OIDC spec)
- `web-v2/web/src/lib/provider/session.js:124` — `dispatch(setAuthUser(user))` sets `authUser` = profile object
[...]

PR #10437, opened 2026-03-15
[#10436] fix(web-v2,server): resolve identity mismatch between UI and server

gada121982:

Add a server-side `GET /api/authn/me` endpoint that returns the fully resolved principal name, and update the web-v2 UI to call this endpoint after OIDC login instead of relying on client-side token parsing.

**Server changes:**
- New `AuthMeResponse` DTO in `common/` module
- New `AuthnOperations` REST resource with `GET /api/authn/me` endpoint that uses `PrincipalUtils.getCurrentUserName()` to return the server-resolved principal
- Unit tests for `AuthnOperations` and `AuthMeResponse` DTO
- OpenAPI spec in `docs/open-api/authn.yaml`

**Web-v2 changes:**
- New `getAuthMeApi()` in `lib/api/auth/index.js`
- Session provider calls `/api/authn/me` after OIDC login, overrides `authUser.name` with server-resolved principal
- OIDC provider prefers `access_token` over `id_token` per OAuth2 spec (with updated tests)
- Logout action performs RP-initiated logout via `signoutRedirect()` before clearing OIDC data

When Gravitino is configured with `principalFields: "preferred_username,sub"`, the server resolves the caller identity from JWT using the first matching field (e.g., `preferred_username` → `"admin"`). However, the web-v2 UI uses the OIDC `profile.name` claim (e.g., `"admin admin"` — the full display name), causing an identity mismatch. This breaks the `serviceAdmins` check since the UI-displayed username doesn't match what the server considers the authenticated principal.

Fix: #10436

- New REST API endpoint: `GET /api/authn/me` → returns `{"code": 0, "principal": "<resolved-name>"}`
- Web-v2 UI now displays the server-resolved principal name instead of the OIDC profile display name

1. Unit test `TestAuthnOperations` verifies the endpoint returns the correct principal from the authenticated request context
2. Unit tests for `AuthMeResponse` DTO (serialization/deserialization + default constructor)
3. Updated OIDC provider tests to validate `access_token` preference over `id_token`
4. OpenAPI spec passes `./gradlew :docs:lintOpenAPI`
5. Manual testing on a K8s cluster with Keycloak OIDC: confirmed `/api/authn/me` returns `"admin"` (from `preferred_username`) while OIDC profile returns `"admin admin"`, and the UI correctly displays the server-resolved principal

### DIFF
--- a/web-v2/web/src/lib/store/auth/index.js
+++ b/web-v2/web/src/lib/store/auth/index.js
@@ -124,6 +124,8 @@ export const logoutAction = createAsyncThunk('auth/logoutAction', async ({ route
 
             // Redirect to IdP logout endpoint — browser navigates away, must be last
             await userManager.signoutRedirect({ id_token_hint: user?.id_token })
+
+            return { token: null } // unreachable — browser navigates away
           }
         }
 


roryqi:
1. User authentication and authorization can use short name `auth`. To clarify more clearly, I prefer using `/api/authn/me`
2. If you add new api operations, you should add it to `docs/open-api`.

roryqi:
cc @bharos

gada121982:
Thanks @roryqi for the review! Addressed both points:

1. **Renamed `/api/auth/me` → `/api/authn/me`** — agreed, `authn` is clearer for authentication vs authorization.
2. **Added OpenAPI spec** — new `docs/open-api/authn.yaml` with the endpoint definition, referenced from `openapi.yaml`. Passes `./gradlew :docs:lintOpenAPI`.

Also added unit tests for `AuthMeResponse` DTO (serialization/deserialization + default constructor).

server/src/main/java/org/apache/gravitino/server/web/rest/AuthOperations.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.gravitino.server.web.Utils;
+import org.apache.gravitino.utils.PrincipalUtils;
+
+/**
+ * Provides the authenticated principal information. This endpoint returns the server-resolved
+ * principal name, ensuring the UI identity matches the server-side identity derived from the
+ * configured {@code principalFields} and {@code principalMapper}.
+ */
+@Path("/authn")
+public class AuthOperations {
```
roryqi: `AuthOperations` -> `AuthnOperations`

server/src/main/java/org/apache/gravitino/server/web/rest/AuthOperations.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Context private HttpServletRequest httpRequest;
+
+  @GET
+  @Path("me")
+  @Produces("application/vnd.gravitino.v1+json")
+  @Timed(name = "auth-me." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
```
roryqi: `auth-me` -> `authn-me`.

gada121982:
Addressed the remaining inline comments:

- Renamed `AuthOperations` → `AuthnOperations` (class + file)
- Renamed metrics `auth-me` → `authn-me`

All feedback has been addressed. Thanks for the thorough review @roryqi!

roryqi:
@LauraXia123 Could u take a look again?

LauraXia123:
@gada121982 Could you check the comment by copilot? Thx

gada121982:
@LauraXia123 Thanks for flagging! All Copilot comments have been addressed:

1. **oidc.test.js** — Updated tests to assert `access_token` preference (matching the new token selection logic)
2. **authn.yaml schema ref** — Moved from `#/components/responses/` to `#/components/schemas/` (proper OpenAPI structure)
3. **signout order** — `signoutRedirect()` now called before `clearAuthData()` so `id_token_hint` is available for RP-initiated logout
4. **PR description** — Updated all references from `/api/auth/me` to `/api/authn/me`
5. **Comment typo** — Fixed fallback comment to reference `/api/authn/me`

All tests pass (Java unit tests + OIDC vitest + OpenAPI lint).

web-v2/web/src/lib/store/auth/index.js
```diff
@@ -102,8 +102,32 @@ export const logoutAction = createAsyncThunk('auth/logoutAction', async ({ route
...
+            // the stored id_token for the id_token_hint parameter in RP-initiated logout
+            await userManager.signoutRedirect()
+
+            await provider.clearAuthData()
+
+            return { token: null }
+          }
+        }
+
         await provider.clearAuthData()
```
LauraXia123, line 132: `clearAuthData` is implemented by the OIDC provider in `web-v2/web/src/lib/auth/providers/oidc.js`, and for the default provider, it is implemented in `web-v2/web/src/lib/auth/providers/generic.js`. So why is `userManager.signoutRedirect()` called before `userManager.removeUser()` in `clearAuthData()`?

gada121982: `signoutRedirect()` needs the stored `id_token` to build the `id_token_hint` parameter for the IdP logout URL. If `clearAuthData()` (which calls `removeUser()`) runs first, the user data is deleted from localStorage — then `signoutRedirect()` has no `id_token_hint` to send, and the IdP may not properly terminate the SSO session.

Flow:
1. `signoutRedirect()` — reads `id_token` from store → redirects to `{authority}/protocol/openid-connect/logout?id_token_hint=...`
2. `clearAuthData()` — calls `removeUser()` to clean up local storage

If reversed, the IdP logout becomes unreliable (no `id_token_hint` → Keycloak may keep the SSO session alive, causing auto-login on next visit).

bharos: Will the browser already navigate away here, ie.
```
   await userManager.signoutRedirect() // ← browser navigates AWAY here
      await provider.clearAuthData()      // ← DEAD CODE: never reached
      return { token: null }              // ← DEAD CODE: never reached
```

bharos: Claude gives me this as a better flow
```
// 1. Read id_token from stored user BEFORE removing it
const user = await userManager.getUser()
// 2. Clear app state first
await provider.clearAuthData()   // removeUser() from storage
dispatch(...)
// 3. Redirect with explicit id_token_hint
await userManager.signoutRedirect({ id_token_hint: user?.id_token })
// nothing after this — browser navigates away
```
Or at a minimum,  to move signoutRedirect() to be the very last line and remove the dead code after it.

gada121982: Good catch — fixed in 024593e. Adopted your suggested flow exactly:

1. `getUser()` → save `id_token` before clearing
2. `clearAuthData()` → `removeUser()` from storage
3. Local cleanup (localStorage, Redux)
4. `signoutRedirect({ id_token_hint: user?.id_token })` — last line, browser navigates away

Dead code removed.

web-v2/web/src/lib/auth/providers/oidc.js
```diff
@@ -67,17 +67,17 @@ export class OidcOAuthProvider extends BaseOAuthProvider {
       let user = await this.userManager.getUser()
 
       if (user && !user.expired) {
-        // For JWKS validation, we need the ID token (JWT format), not the access token
-        return user.id_token || user.access_token
+        // Use access token for API requests per OAuth2 spec
+        return user.access_token || user.id_token
```
bharos, line 71: can this reversal of order cause any issues with other OAuth providers (like Azure AD), or will it work seamlessly

On thinking about this, I feel logically this is correct. as we should use access_token to access gravitino which is a resource server acc. to oauth. so probably this should be ok..

Concern I had is from oauth spec: https://oauth.net/id-tokens-vs-access-tokens/
The access_token is always not JWT, but our validator expects JWT token.  But in most cases I think it should work. we can revisit if we see any problems

web-v2/web/src/lib/store/auth/index.js
bharos: I'm thinking maybe we can keep the return statement, although unreachable, for more readability, after signoutRedirect

return { token: null }   // unreachable in practice but explicit return

common/src/main/java/org/apache/gravitino/dto/responses/AuthMeResponse.java
```diff
@@ -0,0 +1,55 @@
...
+import lombok.ToString;
+
+/** Response for the authenticated principal information. */
+@Getter
+@EqualsAndHashCode(callSuper = true)
+@ToString
+public class AuthMeResponse extends BaseResponse {
+
+  @JsonProperty("principal")
+  private final String principal;
```
bharos, line 33: nit : Should we add @Nullable annotation

roryqi: If principal is null, we will throw an exception directly.

web-v2/web/src/lib/auth/providers/oidc.js
gada121982: Agree with your analysis. Gravitino validators (`StaticSignKeyValidator`, `JwksTokenValidator`) both explicitly require JWT format — opaque tokens fail at parse stage regardless of which token the frontend sends. With Keycloak/Azure AD, `access_token` is always JWT. The `|| id_token` fallback covers edge cases where `access_token` might be absent. For providers that issue opaque access tokens (e.g., Google), Gravitino server-side would need an introspection-based validator — that's a separate concern from this PR.

web-v2/web/src/lib/store/auth/index.js
gada121982: Done — added `return { token: null }` in 40423476a.

LauraXia123 (commented):
LGTM, thx for your contribution @gada121982  and thx for reviewing @bharos

roryqi:
@bharos Do u have other suggestion?

bharos:
LGTM

roryqi:
Merged to main.

@gada121982 Thanks for your contribution! Welcome to join the Apache Gravitino community! Enjoy your trip! Let's build the best catalog in the world!
Thanks @bharos @LauraXia123 for review.

Issue #10272, opened 2026-03-06
[Improvement] Ensure JobManager.close() always shuts down background executors when jobExecutor.close() fails


justinmclean:
### What would you like to be improved?

JobManager.close() calls jobExecutor.close() before shutting down statusPullExecutor and cleanUpExecutor. If jobExecutor.close() throws IOException, method execution exits early, and both schedulers are left running. This can leak background threads during shutdown, causing incomplete cleanup.

### How should we improve?

Make shutdown resilient by guaranteeing executor shutdown in a finally block. Capture any IOException from jobExecutor.close(), run statusPullExecutor.shutdownNow() and cleanUpExecutor.shutdownNow() unconditionally, then rethrow the original exception . This preserves failure signaling while ensuring cleanup always runs.

Here's a test to help:
```
@Test
public void testCloseShouldShutdownExecutorsWhenJobExecutorCloseFails() throws IOException {
  JobExecutor failingJobExecutor = Mockito.mock(JobExecutor.class);
  doThrow(new IOException("close failed")).when(failingJobExecutor).close();

  JobManager manager = new JobManager(config, entityStore, idGenerator, failingJobExecutor);
  try {
    Assertions.assertThrows(IOException.class, manager::close);
    Assertions.assertTrue(manager.statusPullExecutor.isShutdown());
    Assertions.assertTrue(manager.cleanUpExecutor.isShutdown());
  } finally {
    manager.statusPullExecutor.shutdownNow();
    manager.cleanUpExecutor.shutdownNow();
  }
}
```

AmitaWhite:
I would like to work on this issue. Could you please assign it to me?

AmitaWhite:
Hi, I've updated the code and am now running tests. 

However, I'm seeing failures in :common-test and :rat.

Since this is my first contribution, I'm not sure if these are related to my changes or existing environment issues. Could you guide me on where to check the logs or if I need to pass all these checks before submitting the PR? Thanks!

justinmclean:
Rat issues generally involve missing ASF headers in files, but without more detail, it is a little difficult to give specific advice. The CI will run rat and the tests to double check everything is correct.

AmitaWhite:
Thanks for the feedback! I'll look into the build tests again to catch what I might have missed.

PR #10440, opened 2026-03-15
[#10272] Improvement : nsure JobManager.close() always shuts down background executors when jobExecutor.close() fails

AmitaWhite:

This PR fixes a potential thread leak in JobManager.close(). Previously, if jobExecutor.close() threw an IOException, the method would terminate early, skipping the shutdown calls for statusPullExecutor and cleanUpExecutor.
The fix wraps the jobExecutor.close() call in a try block and ensures that both background executors are shut down in the finally block.

To guarantee that background resources are properly cleaned up even when the main job executor fails to close, preventing resource leaks.

Improvement : #10272 

No

    Added a new unit test TestJobManager to simulate jobExecutor.close() failure and verify that background executors are still shut down.
    Ran the specific unit test:

```sh
./gradlew :core:test --tests "org.apache.gravitino.job.TestJobManager"
```


justinmclean (approved):
Thanks for this

PR #10442, opened 2026-03-16
[#9710] docs(flink-connector): add Flink multi-version support spec

### LINKED ISSUE #9710: [Improvement] support multi version for Gravitino Flink connector

The current Gravitino Flink connector supports 1.18, we need to support other versions like 1.19,1.20 like Gravitino spark connector

### How should we improve?

_No response_

FANNG1:

This PR adds a Flink multi-version support spec under `ai-code-spec/` for Gravitino Flink connector.

The spec covers:
- the current Flink connector architecture and its main Flink interaction points
- the proposed module split with `flink-common` plus versioned Flink modules and runtime modules
- dependency matrix considerations for Flink `1.18`, `1.19`, and `1.20`
- extension hooks for future `1.20` catalog API support
- a future compatibility lane for Flink `2.0` that is explicitly out of scope for this delivery
- testing, CI, and documentation expectations for the multi-version support work

This PR also refines the spec metadata and narrows the real Flink environment requirement to a smoke-validation lane for this delivery.

The Flink connector currently targets a single Flink minor version. To support Flink `1.18`, `1.19`, and `1.20` in a maintainable way, the project needs a clear implementation spec that AI agents and human reviewers can follow consistently.

This spec is intended to reduce ambiguity before the actual code split starts and to provide a common review baseline for follow-up implementation work.

Fix: #9710

No runtime user-facing behavior is changed in this PR.

This PR adds a design/spec document only. It documents the intended multi-version support plan and the expected validation scope for future implementation work.

This is a documentation/spec-only change.

No automated tests were run.

### DIFF
--- a/ai-code-spec/flink_multi_version_support_spec.md
+++ b/ai-code-spec/flink_multi_version_support_spec.md
@@ -1,0 +1,1080 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Gravitino Flink Connector Multi-Version Support Spec
+
+| Field | Value |
+| --- | --- |
+| Status | Proposal |
+| Owner | FANNG |
+| Created | 2026-03-16 |
+| Target Release | 1.3.0 |
+| Related Issue | [#9710](https://github.com/apache/gravitino/issues/9710) |
+| Related PR | N/A |
+
+## Goal
+
+Add maintained support for Flink `1.18`, `1.19`, and `1.20` with versioned runtime artifacts, similar to the Spark connector release model.
+
+The output of this work should be:
+
+- one shared `flink-common` module
+- one connector module per supported Flink minor
+- one runtime shadow jar per supported Flink minor
+- CI and docs that explicitly cover the supported Flink matrix
+
+## Non-Goals
+
+- Supporting Flink `1.17` or Flink `2.x`
+- Changing connector behavior beyond what is required for multi-version compatibility
+- Adding new connector features such as materialized table support
+- Reworking provider semantics for Hive, Iceberg, JDBC, or Paimon
+- Adding OAuth2 or Kerberos real-environment validation to the required test matrix for this delivery
+- Producing one universal jar that works across all Flink minors
+
+Note: Flink `2.0` support is explicitly out of scope for this delivery. It should be treated as a larger follow-up compatibility lane rather than a small extension of the `1.18` to `1.20` work, but the module split and extension hooks introduced here should keep a clean path for that future effort.
+
+## Current Connector Model
+
+The current Flink connector is not a generic table runtime connector. It is primarily a Flink `CatalogStore` plus a set of Flink `Catalog` implementations that proxy metadata operations to Gravitino and delegate data/table behavior to native Flink ecosystem catalogs.
+
+### Main Flink Interaction Points
+
+1. Catalog store bootstrapping
+
+- `flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java`
+- `flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStore.java`
+
+These classes integrate with:
+
+- `CatalogStoreFactory`
+- `CatalogStore`
+- `CatalogDescriptor`
+- `FactoryUtil.createCatalogStoreFactoryHelper(...)`
+- ServiceLoader discovery of Flink `Factory`
+
+2. Catalog SPI implementations
+
+- `flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalog.java`
+- `flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalogFactory.java`
+
+These classes integrate with:
+
+- `AbstractCatalog`
+- `CatalogBaseTable`
+- `CatalogTable`
+- `ResolvedCatalogBaseTable`
+- `ResolvedCatalogTable`
+- `CatalogDatabase`
+- `ObjectPath`
+- Flink `TableChange`
+- Flink `Schema`
+
+3. Provider-specific wrappers around native Flink ecosystem catalogs
+
+- Hive: `GravitinoHiveCatalog`, `GravitinoHiveCatalogFactory`
+- Iceberg: `GravitinoIcebergCatalog`, `GravitinoIcebergCatalogFactory`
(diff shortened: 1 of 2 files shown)


ai-code-spec/flink_multi_version_support_spec.md
```diff
@@ -0,0 +1,995 @@
...
+The output of this work should be:
+
+- one shared `flink-common` module
+- one connector module per supported Flink minor
+- one runtime shadow jar per supported Flink minor
+- CI and docs that explicitly cover the supported Flink matrix
+
+## Non-Goals
+
+- Supporting Flink `1.17` or Flink `2.x`
```
jerryshao: Is there a big effort when we supporting Flink 2.x?

jerryshao (commented):
Ignore this duplicated comment.

jerryshao (commented):
### Issue 1: Fate of the existing `:flink-connector:flink` and `:flink-connector:flink-runtime` modules — not addressed

The spec describes the new layout but never says what happens to the existing `:flink-connector:flink` module. Is it deleted after Phase 1? Kept as a 1.18 alias? This is a concrete question any implementer will hit immediately. The spec should state explicitly:

> After Phase 1, `:flink-connector:flink` is replaced by `:flink-connector:flink-common` + `:flink-connector:flink-1.18`. The old project entries in `settings.gradle.kts` are removed.

---

### Issue 2: Provider-level ABI risk is underemphasized for 1.18–1.20 (not just 2.0)

The spec correctly flags Hive as high-risk for Flink 2.0 but frames the 1.18–1.20 provider story as low-risk. In practice, `flink-common` would compile `GravitinoHiveCatalog`, `GravitinoIcebergCatalog`, etc. against `compileOnly` Hive/Iceberg/JDBC 1.18 artifacts. When a 1.20 classpath is substituted at runtime, constructor or method signature changes in the native provider connectors (not the Flink SPI itself) could still cause `NoSuchMethodError`. This risk is real for:

- `HiveCatalog` constructor signature changes between Hive connector versions bundled in each Flink release
- `FlinkCatalog` (Iceberg) constructor changes between `iceberg-flink-runtime-1.18` and `iceberg-flink-runtime-1.20`
- JDBC `JdbcCatalog` constructor between `3.2.0-1.18` and `3.3.0-1.20`

**Recommendation**: Add a Risk 6 explicitly covering provider-level ABI drift for 1.18–1.20. The Phase 3 acceptance criteria should include runtime linkage validation (e.g., a load-class smoke test), not just compilation.

---

### Issue 3: ServiceLoader behavior with `flink-common` + versioned modules needs more detail

The spec says:
> If all factory classes remain in `flink-common`, the service file can remain there.

This is only half the story. When a user loads `gravitino-flink-1.20_2.12.jar`, the `META-INF/services/org.apache.flink.table.factories.Factory` file in `flink-common` is picked up correctly — but if a versioned module also adds a version-specific factory, there is a real risk of factory duplication. The spec should clarify that the versioned module's service file **replaces** (not extends) the common one, requiring the common factories to be re-listed.

---

### Issue 4: `flink-common` tests only validate against Flink 1.18 — coverage gap should be stated

The spec says `testImplementation` on Flink 1.18 baseline in `flink-common`. All unit tests in `flink-common` therefore run only against 1.18 deps. Binary compatibility issues on 1.19/1.20 would only surface during versioned module test runs, not during `flink-common:test`. The spec should say this explicitly so implementers know that 1.19/1.20-specific regression tests belong in the versioned modules, not in `flink-common`.

---

### Minor issues

- **Artifact naming**: `gravitino-flink-1.18_2.12` drops `connector` from the name compared to Spark's `gravitino-spark-connector-3.3_2.12`. Worth confirming this is intentional.
- **`flink-common` compiled against 1.18 vs. latest**: The spec takes the opposite approach from the Spark connector (which compiles against the latest Spark). This is the right call given Flink's additive API nature, but a brief justification would help future maintainers.
- **Scala guard**: The spec says "under the existing Scala 2.12 guard" but doesn't confirm that Flink 1.18, 1.19, and 1.20 all require `_2.12`. This should be stated explicitly.

FANNG1:
@jerryshao Thanks, these points are reasonable. I updated the spec locally in follow-up commit `1bb0952b5` to address them explicitly:

- added a concrete project migration rule for replacing `:flink-connector:flink` / `:flink-connector:flink-runtime` with `flink-common`, `flink-1.18`, and `flink-runtime-1.18`, and stated that the old `settings.gradle.kts` entries should be removed
- added a provider placement matrix plus explicit `1.19` / `1.20` runtime linkage validation requirements, and a separate `Risk 6` for provider-level ABI drift across `1.18` to `1.20`
- clarified the `META-INF/services` rule so that a versioned module service file replaces the effective factory list for that variant and must re-list all valid factories
- stated explicitly that `flink-common:test` only validates the Flink `1.18` baseline dependency graph, and that `1.19` / `1.20` linkage-sensitive regressions belong in the versioned modules
- added the smaller clarifications on artifact naming, why `flink-common` compiles against `1.18` instead of the latest Flink minor, and the `_2.12` scope for `1.18` / `1.19` / `1.20`

I have not resolved or closed any existing comments. I will push the follow-up commit after the next confirmation.

ai-code-spec/flink_multi_version_support_spec.md
FANNG1: Good point. I clarified this in local follow-up commit `1bb0952b5` by making the top-level note more explicit: Flink `2.0` is out of scope for this delivery and should be treated as a larger follow-up compatibility lane, not a small extension of the `1.18` to `1.20` work. The dedicated `Future Flink 2.0 Compatibility Lane` section remains as the longer explanation.

polyzos:
@jerryshao @FANNG1 Nice work here. The flink-common + versioned modules layout and the CatalogCompat hook are the right foundation. 

Two direction-level thoughts:
Version target: Flink 2.2 has been GA since December 2025. Supporting 1.18-1.20 now risks doing this refactor twice. Worth considering dropping 1.18 in favor of 2.x, or at least sketching how CatalogCompatibility extends to 2.x so today’s choices stay forward-compatible.

Baseline: compiling flink-common against 1.18 locks in references to APIs 2.x removed (SinkFunction, legacy TableSource/TableSink, the old config parser). Picking 1.20 LTS as the baseline keeps shared code 2.x-ready · the 1.18 module carries downward backports instead, which is a smaller surface.

A good reference is Apache Fluss and how it manages this and do something similar
https://github.com/apache/fluss/tree/main/fluss-flink

polyzos:
Hi @FANNG1 .. I wanted to follow up on this and see if you are already working on it, and overall, if there is a plan for delivery.

polyzos:
I just saw this was merged 
https://github.com/apache/gravitino/pull/10517
which is great.. so I assume next step now is do support flink 1.19 and 1.20?

FANNG1:
> I just saw this was merged #10517 which is great.. so I assume next step now is do support flink 1.19 and 1.20?

Yes, I will propose a PR for flink 1.19 and 1.20 these days.  How about help to review the 1.19&1.20 PR  and continue work on 2.x support if you have time?

polyzos:
@FANNG1 sure thing.. It's this one, right?
https://github.com/apache/gravitino/pull/10501

For 2.x do you have any concrete plan? I'm thinking whether it makes sense to go directly with the new 2.3 version that is coming this month.

FANNG1:
> @FANNG1 sure thing.. It's this one, right? #10501
> 
> For 2.x do you have any concrete plan? I'm thinking whether it makes sense to go directly with the new 2.3 version that is coming this month.

#10501 is the daft PR to show how to support multi Flink versions. Close it and will propose a new PR.

For 2.x I didn't have concrete plans. I would suggest to support Flink 2.0 as the first version.  In the design, the Flink common module leaves the stub interface to support Flink 2.x, but I didn't test it in real Flink 2.x enviroment yet.

FANNG1:
Close the PR as Flink 1.x support is merged, @polyzos please continue the Flink 2.x support if you have time, thanks.

Issue #10448, opened 2026-03-17
[FEATURE] Enhance do-release scripts to support non-interactive mode


diqiu50:
## Describe the feature

Add non-interactive mode support to `dev/release/do-release.sh` and `dev/release/release-util.sh`.

## Motivation

The release process currently requires manual input at multiple steps. Supporting non-interactive mode (via `-y` flag and environment variables) allows the release pipeline to be driven by automation without human intervention.

## Describe the solution

- `do-release.sh`: add `-y` (force) flag to skip interactive prompts; add `-b`, `-s`, `-r`, `-p`, `-t` options for full CLI control
- `release-util.sh`: read release info from environment variables when in force mode; auto-detect latest branch and RC count

PR #10449, opened 2026-03-17
[#10448] feat(release): enhance do-release scripts to support non-interactive mode

diqiu50:

- `do-release.sh`: add `-y` (force) flag, `-s/-r/-p/-t` options for full CLI control; skip interactive prompts when in force mode; add `set -euo pipefail` for stricter error handling
- `release-util.sh`: read release info from environment variables when in force mode; auto-detect latest branch and RC count; fix duplicate `JAVA_VERSION` assignment in `init_java`

The release process requires manual input at multiple steps. Non-interactive mode allows the release pipeline to be driven by automation without human intervention.

Fix: #10448

No.

Manually verified with `bash -n` syntax check on all modified scripts.

### DIFF
--- a/dev/release/do-release.sh
+++ b/dev/release/do-release.sh
@@ -80,6 +80,18 @@ export RELEASE_TAG=${RELEASE_TAG:-}
 export ASF_PASSWORD=${ASF_PASSWORD:-}
 export GPG_PASSPHRASE=${GPG_PASSPHRASE:-}
 
+if [[ "${RC_COUNT}" != "0" ]] && ! [[ "${RC_COUNT}" =~ ^[1-9][0-9]*$ ]]; then
+  echo "Error: RC number must be a positive integer, got: '${RC_COUNT}'" >&2
+  exit 1
+fi
+
+if [ -n "${RELEASE_STEP}" ]; then
+  case "${RELEASE_STEP}" in
+    tag|build|docs|publish|finalize) ;;
+    *) echo "Error: invalid release step '${RELEASE_STEP}'. Valid steps: tag, build, docs, publish, finalize" >&2; exit 1 ;;
+  esac
+fi
+
 cmds=("git" "gpg" "svn" "twine" "shasum" "sha1sum" "jq" "make")
 for cmd in "${cmds[@]}"; do
   if ! command -v "$cmd" &> /dev/null; then
@@ -104,7 +116,7 @@ fi
 
 if [ "$RUNNING_IN_DOCKER" = "1" ]; then
   # Inside docker, need to import the GPG key stored in the current directory.
-  echo "${GPG_PASSPHRASE:-}" | $GPG --passphrase-fd 0 --import "$SELF/gpg.key"
+  printf '%s\n' "${GPG_PASSPHRASE:-}" | $GPG --passphrase-fd 0 --import "$SELF/gpg.key"
 
   # We may need to adjust the path since JAVA_HOME may be overridden by the driver script.
   if [ -n "${JAVA_HOME:-}" ]; then
--- a/dev/release/release-util.sh
+++ b/dev/release/release-util.sh
@@ -37,10 +37,14 @@ function read_config {
   local ENV_VAR_NAME="$3"  # Optional: env var name to use in non-interactive (force) mode
   local REPLY=
 
-  # In force/non-interactive mode, use the env var value directly if available.
-  if [ -n "$ENV_VAR_NAME" ] && [ -n "${!ENV_VAR_NAME:-}" ] && is_force; then
-    echo "${!ENV_VAR_NAME}"
-    return
+  # In force/non-interactive mode, the env var must be set; error if missing.
+  if is_force; then
+    if [ -n "$ENV_VAR_NAME" ] && [ -n "${!ENV_VAR_NAME:-}" ]; then
+      echo "${!ENV_VAR_NAME}"
+      return
+    else
+      error "Force mode requires '$PROMPT' to be set via environment variable ${ENV_VAR_NAME:-<none>}."
+    fi
   fi
 
   if [ -n "$DEFAULT" ]; then


dev/release/release-util.sh
```diff
@@ -34,12 +34,26 @@ function error {
 function read_config {
   local PROMPT="$1"
   local DEFAULT="$2"
+  local VAR_NAME="$3"  # Optional: environment variable name to check first
```
jerryshao: What is the meaning of this variable? The name should be more specific, "VAR_NAME" is too vague.

dev/release/release-util.sh
```diff
@@ -58,10 +72,8 @@ function run_silent {
   echo "Command: $@"
   echo "Log file: $LOG_FILE"
 
-  "$@" 1>"$LOG_FILE" 2>&1
-
-  local EC=$?
-  if [ $EC != 0 ]; then
+  if ! "$@" 1>"$LOG_FILE" 2>&1; then
+    local EC=$?
```
jerryshao: What is the purpose of this change?

dev/release/release-util.sh
```diff
@@ -58,10 +72,8 @@ function run_silent {
...
     echo "Command FAILED. Check full logs for details."
     tail "$LOG_FILE"
     exit $EC
```
jerryshao, line 82: **Bug: `exit $EC` will always exit with code 0 when the command fails.**

Inside `if ! "$@"; then`, bash stores the exit status of `! "$@"` in `$?`. Since `! cmd` returns **0** when `cmd` fails (the condition is true), `local EC=$?` captures 0, and `exit $EC` exits cleanly despite the failure.

The switch from the old pattern to `if !` was necessary because `set -e` (added in this PR) would cause the script to abort before `local EC=$?` could run in the original code. However, the exit code must now be hardcoded:

```bash
if ! "$@" 1>"$LOG_FILE" 2>&1; then
  echo "Command FAILED. Check full logs for details."
  tail "$LOG_FILE"
  exit 1
fi
```

dev/release/do-release.sh
```diff
@@ -20,22 +22,50 @@
...
-export DRY_RUN
+export DRY_RUN=${DRY_RUN:-0}
+export FORCE=${FORCE:-0}
+export RC_COUNT=${RC_COUNT:-0}
+export RELEASE_STEP=${RELEASE_STEP:-}
+export GIT_BRANCH=${GIT_BRANCH:-}
+export ASF_PASSWORD=${ASF_PASSWORD:-}
+export GPG_PASSPHRASE=${GPG_PASSPHRASE:-}
+
+echo "DRY_RUN=$DRY_RUN FORCE=$FORCE RC_COUNT=$RC_COUNT RELEASE_STEP=$RELEASE_STEP"
```
jerryshao: This debug line leaks release metadata (branch, RC number, step) unconditionally to CI logs and terminal output. Consider removing it entirely or guarding it behind a `DEBUG` flag:

```bash
if [ "${DEBUG:-0}" = "1" ]; then
  echo "DRY_RUN=$DRY_RUN FORCE=$FORCE RC_COUNT=$RC_COUNT RELEASE_STEP=$RELEASE_STEP"
fi
```

dev/release/do-release.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
```
jerryshao: Minor style nit: `set -euo pipefail` is placed before the ASF license header. By convention the license header should immediately follow the shebang line. Suggest moving this after the license block (around line 23, after the existing comment block).

dev/release/do-release.sh
```diff
@@ -20,22 +22,50 @@
...
+      echo "  -r <num>      Release candidate number (e.g., 6 for rc6)"
+      echo "  -n            Dry run mode"
+      echo "  -p <pass>     GPG passphrase (insecure; prefer GPG_PASSPHRASE env var)"
+      echo "  -t <pass>     ASF password (insecure; prefer ASF_PASSWORD env var)"
+      echo "  -y            Force continue without confirmation"
+      echo "  -h            Show this help message"
+      exit 0
+      ;;
+    :) echo "Option -$OPTARG requires an argument." >&2; exit 1 ;;
+    \?) echo "Invalid option: -$OPTARG" >&2; exit 1 ;;
```
jerryshao, line 68: Can you also add an example in the above to basically show users how to use the script.

dev/release/release-util.sh
jerryshao: Please reply this comment.

dev/release/do-release.sh
jerryshao: Also, this one is not fixed.

dev/release/do-release.sh
```diff
@@ -49,26 +77,28 @@ fi
...
     # JAVA_HOME for the openjdk package.
     export JAVA_HOME=/usr
   fi
 else
-  # Outside docker, need to ask for information about the release.
-  get_release_info
+  # Outside docker, need to ask for information about the release if required vars are missing.
+  if [ -z "$GIT_BRANCH" ] || [ -z "$RELEASE_VERSION" ] || [ "$RC_COUNT" -eq 0 ]; then
+    get_release_info
+  fi
```
jerryshao: Bug: SKIP_TAG and RELEASE_TAG are unbound when get_release_info is skipped.\n\nBoth variables are only set inside get_release_info. When GIT_BRANCH, RELEASE_VERSION, and RC_COUNT are all provided (the automation fast-path), get_release_info is skipped and neither SKIP_TAG nor RELEASE_TAG is ever initialized. With set -u now active, execution reaches line 109 (if should_build "tag" && [ $SKIP_TAG = 0 ]) and aborts with "SKIP_TAG: unbound variable" before any release work begins.\n\nThe same issue exists in docker mode (the if [ "" = "1" ] branch), where get_release_info is also never called.\n\nSuggested fix — add an else branch to derive the missing variables:\n\n  if [ -z "" ] || [ -z "" ] || [ "" -eq 0 ]; then\n    get_release_info\n  else\n    # Derive what get_release_info would have set\n    RELEASE_TAG="v-rc"\n    SKIP_TAG=0\n    if check_for_tag ""; then SKIP_TAG=1; fi\n  fi\n\nAnd identical initialization is needed at the end of the docker branch.

jerryshao (commented):
Thanks for the quick updates — the three previous issues are all resolved. One remaining bug below.

dev/release/release-util.sh
diqiu50: We add `set -e` in do-release.sh,  We need to adjust the logic here to ensure errors can be logged.

dev/release/do-release.sh
diqiu50: Fixed

dev/release/do-release.sh
diqiu50: ./do-release.sh -h

Issue #10223, opened 2026-03-05
[Improvement] Prevent sensitive Spark job environment values from being logged in SparkProcessBuilder


justinmclean:
### What would you like to be improved?

SparkProcessBuilder.start() logs the full environment map (sparkJobTemplate.environments()) at INFO. In production defaults, Spark jobs can run via the default local executor, and INFO logging is enabled in chart defaults. If users pass secrets (tokens/passwords/keys) through job environment variables, those secrets are written to server logs in plaintext.

### How should we improve?

Change logging in SparkProcessBuilder.start() to avoid printing raw environment values. For example, log only environment variable names, or log a redacted map where sensitive values are masked (or mask all values). Keep command logging, but ensure no secret-bearing environment values are emitted at INFO.

Lucas61000:
Hello, I’m interested in working on this.

justinmclean:
Thanks @Lucas61000, if you have any questions just ask.

PR #10450, opened 2026-03-17
[#10223] Prevent sensitive Spark job environment values from being logged in SparkProcessBuilder

Lucas61000:

- Modified SparkProcessBuilder.start() method to log only environment variable names instead of full key-value pairs

Fix: #10223

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
+++ b/core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
@@ -125,9 +125,9 @@ public class SparkProcessBuilder extends LocalProcessBuilder {
     builder.redirectError(errorFile);
 
     LOG.info(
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/job/local/SparkProcessBuilder.java
```diff
@@ -127,7 +127,7 @@ public Process start() {
     LOG.info(
         "Starting local Spark job with command: {}, environment variables: {}",
```
Abyss-lord: I think `"Starting local Spark job with command: {}, environment variable names: {}"` it's better, since we no longer output the values.

Lucas61000: Thank you! fixed.

Issue #10444, opened 2026-03-16
[Bug report] Register delta table then restart the gravitino server, encounter error: Table format ... is null


danhuawang:
### Version

main branch

### Describe what's wrong

Load delta table after restart the gravitino encounter 400 Bad Request

```
{
    "code": 1001,
    "type": "IllegalArgumentException",
    "message": "Failed to operate table(s) [delta_datatype_basic] operation [LOAD] under schema [delta_datatype_schema], reason [Table format for delta_test_metalake.delta_datatype_test_catalog.delta_datatype_schema.delta_datatype_basic is null, this is unexpected]",
    "stack": [
        "java.lang.IllegalArgumentException: Table format for delta_test_metalake.delta_datatype_test_catalog.delta_datatype_schema.delta_datatype_basic is null, this is unexpected",
        "\tat com.google.common.base.Preconditions.checkArgument(Preconditions.java:218)",
        "\tat org.apache.gravitino.catalog.lakehouse.generic.GenericCatalogOperations.lambda$tableOps$5(GenericCatalogOperations.java:336)",
        "\tat com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4938)",
        "\tat com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3576)",
        "\tat com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2318)",
        "\tat com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2191)",
        "\tat com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2081)",
        "\tat com.google.common.cache.LocalCache.get(LocalCache.java:4019)",
        "\tat com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4933)",
        "\tat org.apache.gravitino.catalog.lakehouse.generic.GenericCatalogOperations.tableOps(GenericCatalogOperations.java:331)",
        "\tat org.apache.gravitino.catalog.lakehouse.generic.GenericCatalogOperations.loadTable(GenericCatalogOperations.java:212)",
        "\tat org.apache.gravitino.catalog.TableOperationDispatcher.lambda$internalLoadTable$20(TableOperationDispatcher.java:502)",
        "\tat org.apache.gravitino.catalog.CatalogManager$CatalogWrapper.lambda$doWithTableOps$1(CatalogManager.java:153)",
[...]

PR #10451, opened 2026-03-17
[#10444] fix(core): batchSelect queries missing version-info JOIN and field aliases

jerryshao:

- Fixed `batchSelectTableByIdentifier` in `TableMetaBaseSQLProvider` to include a LEFT JOIN with `table_version_info`, selecting all version-info fields (`format`, `properties`, `partitioning`, `sort_orders`, `distribution`, `indexes`, `comment`). Previously these fields were omitted, causing incomplete `TableEntity` objects to be stored in the entity cache.
- Remove `batchSelectJobByIdentifier` in `JobMetaBaseSQLProvider` as it is not used in the current code base.
- Added regression tests in `TestTableMetaService` covering all version-info fields and the columns limitation.

`MetadataAuthzHelper.preloadToCache()` calls `entityStore().batchGet()` for TABLE entities on every authorized request. This invokes `batchSelectTableByIdentifier`, which was missing the LEFT JOIN with `table_version_info`. As a result, incomplete `TableEntity` objects (with no `format` or `properties`) were stored in the entity cache.

After a server restart, `tableFormatCache` (an in-memory Guava cache in `GenericCatalogOperations`) is cleared. When loading a Delta/generic lakehouse table, `tableOps()` misses the `tableFormatCache` and falls back to `entityStore().get()`, which returns the incomplete cached entity. This causes `Preconditions.checkArgument(format != null, "Table format for %s is null...")` to throw an `IllegalArgumentException`.

Fix: #10444

No.

- Added `testBatchGetTableByIdentifierIncludesVersionInfoFields` to verify all version-info fields (`format`, `properties`, `comment`, `distribution`, `sortOrders`, `partitioning`, `indexes`) are returned by `batchGetTableByIdentifier`.
- Added `testBatchGetTableByIdentifierDoesNotIncludeColumns` to document that columns are fetched via a separate path (`getTableByIdentifier`) and are intentionally not included in batch results.

### DIFF
--- a/dev/release/publish-docker.sh
+++ b/dev/release/publish-docker.sh
@@ -1,0 +1,235 @@
+#!/bin/bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Build and publish Gravitino Docker images via GitHub Actions
+#
+# Usage: ./publish-docker.sh <tag|branch> --docker-version <version> [--trino-version <version>] [--dry-run]
+#
+# Arguments:
+#   <tag>                     A Git tag, typically a release candidate (e.g., v1.2.0-rc5).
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/JobMetaBaseSQLProvider.java
```diff
@@ -183,14 +183,21 @@ public String deleteJobMetasByLegacyTimeline(
         + " WHERE deleted_at < #{legacyTimeline} AND deleted_at > 0 LIMIT #{limit}";
   }
 
+  // Note: this method is currently not called by the service layer. batchGetJobByIdentifier in
+  // JobMetaService uses batchSelectJobByRunIds instead, because jobs are identified by their
+  // numeric run ID rather than a name. This method is kept correct and consistent to avoid
+  // confusion if it is used in the future.
   public String batchSelectJobByIdentifier(
```
roryqi: This method should be unused. We can remove it directly.

roryqi:
Could u modify the pull request description?

jerryshao:
> Could u modify the pull request description?

Done.

PR #10475, opened 2026-03-19
[#10477] fix(flink-paimon):  Paimon FlinkCatalog to fix Hive partition updates

wangxiaojing:

Two problems are fixed together as they share the same root cause and fix location:

**Problem 1 — Paimon partitions not synced to Hive Metastore**

When using `catalog-backend=hive` with `metastore.partitioned-table=true`, partitions written by Flink jobs are never registered in HMS. `SHOW PARTITIONS` always returns empty.

Root cause: `BaseCatalog.toFlinkTable()` returns a plain Flink `CatalogTable` rebuilt from Gravitino metadata. Paimon's write path requires its own `DataCatalogTable` to initialise `FileStoreTable` with a non-null `catalogLoader` in `CatalogEnvironment`. Without it, `AddPartitionCommitCallback` is never registered and HMS partition metadata is never updated after checkpoint commit.

```
BaseCatalog.toFlinkTable()
  └─ returns plain CatalogTable
       └─ FileStoreTable uses CatalogEnvironment.empty()  // catalogLoader == null
            └─ AddPartitionCommitCallback never registered
                 └─ HMS partitions never updated
```

**Problem 2 — Paimon native catalog cache not invalidated after DDL**

After `DROP TABLE` / `RENAME TABLE` / `ALTER TABLE` operations routed through Gravitino, Paimon's internal `CachingCatalog` still holds the old table metadata. Subsequent `getTable()` calls can return stale data until the cache TTL expires.

### Solution

**1. `BaseCatalog`: add `enrichCatalogTable` hook**

A `protected` hook called at the end of `toFlinkTable()`. Default is identity — zero impact on Hive, Iceberg, JDBC connectors:

```java
protected CatalogBaseTable enrichCatalogTable(CatalogTable table, ObjectPath tablePath) {
    return table;
}
```

**2. `BaseCatalog`: add `invalidateNativeTableCache` hook + call it after DDL**

A `protected` no-op hook called by `invalidateTable()` (best-effort, failures logged at DEBUG). Invoked **after** each successful DDL (drop / rename / alter) so the cache is only cleared once Gravitino has committed the change:

```java
protected void invalidateNativeTableCache(ObjectPath tablePath) {}
```

**3. `GravitinoPaimonCatalog`: implement both hooks**

- `enrichCatalogTable` — delegates to `paimonCatalog.getTable()`, returning Paimon's native `DataCatalogTable` with a fully-initialised `CatalogEnvironment`.
- `invalidateNativeTableCache` — casts `paimonCatalog` to `FlinkCatalog` and calls `catalog().invalidateTable(Identifier)`, forwarding to `CachingCatalog` when caching is enabled. No-op otherwise (e.g. filesystem mode, unit tests).

DDL operations continue to route exclusively through the Gravitino REST API. The internal `paimonCatalog` is used only for `getTable` enrichment and cache invalidation. Gravitino remains the single source of truth for metadata.

- `metastore.partitioned-table=true` is completely non-functional through the Gravitino Flink connector. Users see no error and no partitions — a silent failure.
- Paimon's `CachingCatalog` is not cleared after DDL, leading to potential stale reads.


wangxiaojing:
Hi @roryqi  could you please review this PR when you have a moment? Thanks!

roryqi:
> Hi @roryqi could you please review this PR when you have a moment? Thanks!

Could u create an issue first?

yuqi1129:
@wangxiaojing 
Can you spare some time to resolve the test problem?

FANNG1:
After this change, the load table request will bypass Gravitino server, I'm afraid whether this is the correct direction. Maybe we could find a more general way to fix the problems without bypass Gravitino server, what do you think?

wangxiaojing:
> @wangxiaojing Can you spare some time to resolve the test problem?

done

wangxiaojing:
> After this change, the load table request will bypass Gravitino server, I'm afraid whether this is the correct direction. Maybe we could find a more general way to fix the problems without bypass Gravitino server, what do you think?

    Gravitino remains authoritative: the caller {@link BaseCatalog#getTable} still validates table existence via the Gravitino REST API before invoking this method, so access-control and auditing are preserved

yuqi1129:
@FANNG1 
Can you spare some time to help trouble shoting CI failure of this PR?

FANNG1:
Thanks for working on this issue! The root cause analysis is accurate — the missing `CatalogEnvironment` / `catalogLoader` causes Hive partition metadata not being updated.

However, I have some concerns about the current dual-write approach (overriding `createTable`/`alterTable`/`dropTable` to write to both Gravitino and Paimon):

1. **Consistency risk**: There is no transaction guarantee between the two writes. If the second write fails, metadata becomes inconsistent with no rollback mechanism. For example, in `dropTable`, Gravitino metadata is deleted first — if the subsequent Paimon drop fails, the table becomes an orphan.

2. **Architecture concern**: Gravitino is designed as the single source of truth for metadata. DDL operations go through the Gravitino REST API, and Gravitino server internally syncs to the underlying catalog (Paimon). Dual-writing from the Flink connector bypasses this design.

3. **PR description vs. code mismatch**: The description states the fix is for `getTable()`/`tableExists()`, but the actual code overrides `createTable`/`alterTable`/`dropTable` instead.

**Suggested direction**: The problem is that `BaseCatalog.getTable()` → `toFlinkTable()` returns a plain `CatalogTable` without Paimon-specific context. A more targeted fix would be to override `toFlinkTable()` in `GravitinoPaimonCatalog` to return a Paimon `DataCatalogTable` with a proper `CatalogEnvironment`. This keeps Gravitino as the metadata source of truth while providing the context Paimon needs for partition handling.

lasdf1234:
@roryqi  Can Copilot be used to review the code?

lasdf1234:
@yuqi1129 Could you  let copilot review this PR?

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
```diff
@@ -53,14 +62,106 @@ protected GravitinoPaimonCatalog(
...
-    this.paimonCatalog = flinkCatalogFactory.createCatalog(context);
+    this.paimonCatalog = flinkCatalogFactory.createCatalog(toPaimonContext(context));
+  }
+
+  /**
+   * Translates Gravitino catalog property names to their Paimon/Flink equivalents so that the
+   * underlying {@code FlinkCatalog} can be initialised correctly (e.g., {@code catalog-backend}
+   * becomes {@code metastore}).
+   */
+  private static CatalogFactory.Context toPaimonContext(CatalogFactory.Context context) {
```
lasdf1234: I think this method should be placed in GravitinoPaimonCatalogFactory

lasdf1234 (commented):
Method 'transForm(context)' should be placed in GravitinoPaimonCatalogFactory?

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
lasdf1234: @yuqi1129 @FANNG1 how do you think?

FANNG1:
for metadata operations, any reason to invoke `loadTable`&`alterTable` for paimonCatalog?

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
yuqi1129: I have no idea about whether it's proper here.

wangxiaojing:
Good point. loadTable(...) and paimonCatalog.alterTable(...) serve different purposes here.

  loadTable(...) is still on the Gravitino path for existence check / authorization, so Gravitino remains the metadata authority.
  paimonCatalog.getTable(...) is only for final Flink-side materialization of the native Paimon table object.

  The extra paimonCatalog.alterTable(...) is there to keep the embedded paimonCatalog view aligned with the Gravitino-applied change, because otherwise subsequent paimonCatalog.getTable(...) calls may still observe stale metadata.

FANNG1 (commented):
Thanks for the fix — the root cause analysis is clear and the two-hook design (`enrichCatalogTable` + `invalidateNativeTableCache`) is clean. A few issues worth addressing before merge:

---

### 🔴 Bug 1 — `DROP TABLE` cache invalidation is dead code for Paimon

`BaseCatalog.dropTable()` now calls `invalidateTable(tablePath)`, but `GravitinoPaimonCatalog.dropTable()` is a **complete override** that never calls `super.dropTable()`:

```java
// GravitinoPaimonCatalog.dropTable — no invalidateTable() call
@Override
public void dropTable(ObjectPath tablePath, boolean ignoreIfNotExists) throws ... {
    catalog().asTableCatalog().purgeTable(...);
    // cache is never evicted
}
```

After `DROP TABLE` on a Paimon table, `CachingCatalog` retains the stale entry until TTL. Any Flink job holding a `DataCatalogTable` reference (with an embedded `catalogLoader`) may still read or write to the deleted table via the stale `FileStoreTable`. Fix: add `invalidateNativeTableCache(tablePath)` at the end of `GravitinoPaimonCatalog.dropTable()`.

---

### 🔴 Bug 2 — `testInvalidateNativeTableCacheCallsPaimonInvalidate` tests the wrong object

`invalidateNativeTableCache` accesses the **private field** `paimonCatalog` directly, which is set from the factory in the constructor:

```java
// GravitinoPaimonCatalog.java
private final AbstractCatalog paimonCatalog;  // set by flinkCatalogFactory.createCatalog(context)

protected void invalidateNativeTableCache(ObjectPath tablePath) {
    if (paimonCatalog instanceof FlinkCatalog) {  // ← uses field, not realCatalog()
```

`TestablePaimonCatalog` overrides `realCatalog()` to return `injectedPaimon` (the mock), but `invalidateNativeTableCache` never calls `realCatalog()`. When `new TestablePaimonCatalog(mockFlinkCatalog)` is constructed with `MockCatalogContext(empty options)`, `paimonCatalog` is set to a factory-created catalog — **not** `mockFlinkCatalog`. The `verify(mockInnerCatalog).invalidateTable(expected)` assertion will never be satisfied.

Fix: have `invalidateNativeTableCache` call `realCatalog()` instead of accessing the field directly, or add a package-private constructor that skips the factory for test use.

---

### 🔴 Bug 3 — `CatalogException` from `enrichCatalogTable` is swallowed and double-wrapped

`BaseCatalog.getTable()` has a generic `catch (Exception e)` block:

```java
try {
    Table table = catalog().asTableCatalog().loadTable(ident);
    return toFlinkTable(table, tablePath);  // enrichCatalogTable can throw CatalogException
} catch (NoSuchTableException e) {
    // fall through to views
} catch (Exception e) {                    // CatalogException is caught here!
    LOG.warn("Failed to load table ...", e);
    throw new CatalogException(e);         // double-wrapped
}
```

When Gravitino has a table but Paimon's store doesn't (metadata out-of-sync), `enrichCatalogTable` throws the helpful `"Table X was found in Gravitino but absent from Paimon"` message. This gets caught and re-wrapped as a generic `CatalogException("Failed to load table")`, burying the diagnostic two `.getCause()` calls deep. Fix: re-throw `CatalogException` before the generic catch block.

---

### 🟡 Design — `invalidateTable` placement is fragile; suggest Template Method + abstract hook

The current design relies on an **implicit contract**: any subclass that overrides a DDL method must also remember to call `invalidateTable()`. This is easy to break (as Bug 1 shows). A more robust approach:

1. **Template Method** for DDL sequencing — subclasses override `doDropTable` instead of `dropTable`:

```java
// BaseCatalog
public final void dropTable(ObjectPath tablePath, boolean ignoreIfNotExists) throws ... {
    doDropTable(tablePath, ignoreIfNotExists);
    invalidateNativeTableCache(tablePath);  // always called, cannot be bypassed
}

protected void doDropTable(ObjectPath tablePath, boolean ignoreIfNotExists) throws ... {
    // default Gravitino implementation
}
```

2. **Make `invalidateNativeTableCache` abstract** so every subclass explicitly declares its cache behavior:

```java
protected abstract void invalidateNativeTableCache(ObjectPath tablePath);

// Hive / Iceberg / JDBC — explicit no-op:
@Override
protected void invalidateNativeTableCache(ObjectPath tablePath) {}

// Paimon — real implementation:
@Override
protected void invalidateNativeTableCache(ObjectPath tablePath) { ... }
```

Compile-time enforcement prevents future catalogs from silently omitting cache invalidation. The empty `{}` is no more verbose than the current default no-op, but forces a conscious decision from each subclass author.

---

### 🟡 Minor — `invalidateTable` should be `protected`, not `public`

This is an internal DDL-lifecycle hook. Exposing it as `public` lets external callers trigger spurious cache evictions and adds it to the stable public API surface of `BaseCatalog`.

---

### 🟡 Minor — `BaseCatalog.dropTable` calls `invalidateTable` even when only a view was dropped

`invalidateTable(tablePath)` is called unconditionally after the drop attempt. When `tableDropped == false && viewDropped == true`, this triggers a Paimon cache lookup for a view path that was never in the table cache. The call should be guarded to only fire when `tableDropped == true`.

wangxiaojing:
Thanks for the detailed review — fixed in the latest update:

  - added native cache invalidation after successful `DROP TABLE`
  - changed cache invalidation to use `realCatalog()`
  - preserved `CatalogException` in `BaseCatalog.getTable()`
  - only invalidate when a table is actually dropped
  - reduced `invalidateTable` visibility to `protected`

FANNG1 (commented):
Quick follow-up on the previous round — most issues are now fixed. Two remain.

---

### 🔴 `enrichCatalogTable` still accesses `paimonCatalog` field directly (tests broken)

`invalidateNativeTableCache` was correctly changed to use `realCatalog()`, but `enrichCatalogTable` was not updated consistently:

```java
// invalidateNativeTableCache — fixed ✅
AbstractCatalog nativeCatalog = realCatalog();

// enrichCatalogTable — still uses field ❌
return paimonCatalog.getTable(tablePath);
```

`TestablePaimonCatalog` injects the mock via `realCatalog()` override. `invalidateNativeTableCache` picks it up correctly, but `enrichCatalogTable` bypasses `realCatalog()` and hits the factory-created `paimonCatalog` field instead — the same class of bug that was just fixed in the other method.

As a result, `testEnrichCatalogTableReturnsPaimonNativeTable` and `testEnrichCatalogTableOutOfSyncThrowsCatalogException` both set up expectations on `mockPaimonCatalog`, but those mocks are never called — the factory-created catalog is invoked instead. Both `verify`/`assertSame` assertions will fail at runtime.

Fix: change `enrichCatalogTable` to use `realCatalog()` for consistency:

```java
@Override
protected CatalogBaseTable enrichCatalogTable(CatalogTable ignoredBaseTable, ObjectPath tablePath)
    throws CatalogException {
  try {
    return realCatalog().getTable(tablePath);   // was: paimonCatalog.getTable(tablePath)
  } catch (TableNotExistException e) { ... }
}
```

---

### 🟡 `testGetTablePreservesCatalogException` does not test the actual fix

The test calls `toFlinkTable()` directly, which trivially rethrows the injected exception — that's just how `TestableBaseCatalog.toFlinkTable` is wired. It never exercises `BaseCatalog.getTable()`, which is where the `catch (CatalogException e) { throw e; }` fix actually lives.

To validate the fix, the test should call `getTable()` and confirm the `CatalogException` is not double-wrapped:

```java
@Test
public void testGetTablePreservesCatalogException() throws Exception {
    TestableBaseCatalog baseCatalog = new TestableBaseCatalog();
    CatalogException expected = new CatalogException("boom");
    baseCatalog.toFlinkTableException = expected;

    CatalogException actual = Assertions.assertThrows(
        CatalogException.class,
        () -> baseCatalog.getTable(new ObjectPath("db", "tbl")));  // ← getTable, not toFlinkTable

    Assertions.assertSame(expected, actual,
        "CatalogException from enrichCatalogTable must not be double-wrapped");
}
```

wangxiaojing:
Thanks for the detailed review. I added the alterTable override to keep the Paimon native cache consistent after Gravitino-side table updates, and to ensure alterTable(..., ignoreIfNotExists=true) remains a true no-op instead of touching the underlying Paimon catalog in an  out-of-sync case.

FANNG1 (commented):
Good progress — `testGetTablePreservesCatalogException`, the new `alterTable` override, and the `testGetTableAuthFailureDoesNotCallPaimonCatalog` test are all improvements. One issue remains.

---

### 🔴 `enrichCatalogTable` still accesses `paimonCatalog` field, not `realCatalog()` — tests broken

`invalidateNativeTableCache` correctly uses `realCatalog()`:
```java
AbstractCatalog nativeCatalog = realCatalog();  // ✅
```

But `enrichCatalogTable` still accesses the private field directly:
```java
return paimonCatalog.getTable(tablePath);  // ❌
```

`TestablePaimonCatalog` overrides `realCatalog()` to return `injectedPaimon` (the mock), but `enrichCatalogTable` bypasses `realCatalog()` and calls the factory-created `paimonCatalog` field instead. Both `testEnrichCatalogTableReturnsPaimonNativeTable` and `testEnrichCatalogTableOutOfSyncThrowsCatalogException` set expectations on `mockPaimonCatalog`, which is never called — the wrong catalog instance is used. Both tests will fail.

Fix: use `realCatalog()` consistently, same as `invalidateNativeTableCache`:
```java
@Override
protected CatalogBaseTable enrichCatalogTable(CatalogTable ignoredBaseTable, ObjectPath tablePath)
    throws CatalogException {
  try {
    return realCatalog().getTable(tablePath);   // was: paimonCatalog.getTable(tablePath)
  } catch (TableNotExistException e) { ... }
}
```

---

### 🟡 `alterTable` override adds an extra Gravitino REST call on the happy path

When `ignoreIfNotExists=true` and the table **exists**, the new override does:
1. `tableExists(tablePath)` — one Gravitino REST call
2. Falls through to `super.alterTable()` → `getTable()` — a second Gravitino REST call

The base class already handles the non-existent case correctly: `super.alterTable()` calls `getTable()`, which calls Gravitino's `loadTable()`, catches `NoSuchTableException`, and returns silently when `ignoreIfNotExists=true`. So the `tableExists()` pre-check is redundant for that path and adds overhead to the common (table-exists) path.

If the goal is to avoid the `enrichCatalogTable` Paimon round-trip for non-existent tables, that's already avoided — `enrichCatalogTable` is only called after Gravitino confirms the table exists. If the goal is to guard against out-of-sync state, the `tableExists()` check doesn't help — when the table exists in Gravitino but not in Paimon, `tableExists()` returns `true` and `super.alterTable()` still hits `enrichCatalogTable`, which throws the same `CatalogException`.

wangxiaojing:
@FANNG1    Thanks for the detailed review.
  1. Switch enrichCatalogTable to use realCatalog() instead of the private
  paimonCatalog field so test injection and native catalog access stay
  consistent.

  2. Remove redundant alterTable pre-check overrides and update the Paimon catalog
  tests with minimal valid setup for warehouse initialization and table metadata.

FANNG1 (approved):

FANNG1:
@wangxiaojing  merged to main, thanks for your contribution!

PR #10479, opened 2026-03-19
[#10478] feat(release): add gravitino-release skill and fix state-dir path bug

### LINKED ISSUE #10478: [Improvement] Add gravitino-release Claude Code skill for automated release management

The Gravitino release process is complex, multi-staged, and error-prone when done manually. There is no structured tooling to guide a release manager through the full pipeline step by step, track stage completion, or handle resumption after failures.

### How should we improve?

Add a `gravitino-release` Claude Code skill (`dev/release/gravitino-release/SKILL.md`) that acts as an interactive release manager agent. The skill:

- Guides the release manager through all 7 stages in order: **tag → build → docs → publish → docker → finalize → release-note**
- Tracks stage completion via `.done` state files (anchored to the script directory, not the working directory)
- Supports **mock mode** (using `dev/release/mock/`) and **dry-run mode** for safe testing without real operations
- Runs preflight checks (open PRs, CI status) before tagging
- Collects and validates all required credentials before each stage
- Guards the irreversible `finalize` stage with explicit confirmation
- Generates release notes from GitHub issues labelled with the release version

Also fixes two bugs in the release scripts:
- `do-release.sh` and `mock/do-release.sh`: `STATE_DIR` used a relative path (`.release-state`) causing state files to be written to the caller's working directory instead of alongside the scripts. Fixed to use `$SELF/.release-state`.
- `mock/do-release.sh`: log files also used a relative path; fixed to use `$SELF/${STEP}.log`.

jerryshao:

- Adds `dev/release/gravitino-release/SKILL.md`: a Claude Code skill that acts as an interactive release manager, guiding through all 7 stages (tag, build, docs, publish, docker, finalize, release-note) with state tracking, mock mode, dry-run support, preflight checks, and credential collection.
- Adds `dev/release/mock/do-release.sh` and `dev/release/mock/publish-docker.sh`: mock scripts that simulate all stages without real operations, for safe end-to-end testing of the skill.
- Fixes `STATE_DIR` path bug in `do-release.sh`, `mock/do-release.sh`, and `publish-docker.sh`: the default `.release-state` was a relative path, causing state files to be written to the caller's working directory instead of alongside the scripts. Fixed to use `$SELF/.release-state`.
- Fixes log file path bug in `mock/do-release.sh`: log files also used a relative path; fixed to use `$SELF/${STEP}.log`.

Fix: #10478

The release process is complex and multi-staged. Without structured tooling, a release manager must remember which stages have run, which credentials are needed, and how to recover from failures. The skill automates this orchestration. The state-dir bug caused state files to be scattered in unpredictable locations depending on where scripts were invoked from, breaking resume-after-failure behavior.

No API or property key changes. The skill and mock scripts are new developer tooling only.

Tested end-to-end using mock mode (`dev/release/mock/do-release.sh`) with all 7 stages, including failure simulation via `MOCK_FAIL_STAGE`, dry-run mode, and resume-from-partial-state scenarios.


.claude/skills/gravitino-release/SKILL.md
mchades, line 1: Is the content of this file identical to that of `dev/release/gravitino-release/SKILL.md`? Can we use a symbolic link?

jerryshao: The one in the `.claude` is used for my test. My plan is to only maintain the skill in dev/release, users can manually install them in anywhere. WDYT?

diqiu50:
Overall, the workflow looks generally fine. One area for improvement would be to split this skill into multiple smaller skills to reduce context contamination.

You could use one skill to orchestrate the overall process, for example release-workflow, and then split the concrete execution steps into separate functional skills, such as:
	•	release-tag
	•	release-build (including build, docs, and license checks)
	•	publish-release
	•	finalize-release
	•	release-note
	•	...

The release-workflow skill could then invoke these functional skills and handle them through sub-agents.

jerryshao:
> Overall, the workflow looks generally fine. One area for improvement would be to split this skill into multiple smaller skills to reduce context contamination.
> 
> You could use one skill to orchestrate the overall process, for example release-workflow, and then split the concrete execution steps into separate functional skills, such as: • release-tag • release-build (including build, docs, and license checks) • publish-release • finalize-release • release-note • ...
> 
> The release-workflow skill could then invoke these functional skills and handle them through sub-agents.

My feeling is that using subagent is not so useful. Because the step has a sequence, it cannot be executed in parallel. And since I controlled the workflow, so currently there's no chance to skip the stages. The one concern is that this skill file is too big, but I've met any context window issue for now.

Yes, there's some advanced tricks to improve the skill file, I need to investigate more.

agent-skills/gravitino-release/SKILL.md
mchades, line 1: should we add a README.md to describe how to use this skill?

jerryshao: OK, let me add it.

jerryshao:
@mchades please help to review.

jerryshao:
@mchades please help to review this. It is better you can merge this before the meetup. So you can have something to say in the meetup.

mchades (approved):
No more comments, merge first, and continuously refine.

PR #2644, opened 2024-03-22
[#2557] improvement(core): Reuse Isolated class loader for catalog `alter` operation

yuqi1129:
### What changes were proposed in this pull request?

Reuse isolated classloader instances for catalog alter operations. 

### Why are the changes needed?

If we repeat the alter catalogs, there will be lots of isolated classloaders, it's no need to create a new isolated classloader instance for altering operations, so we'd better reused the old one. 

Fix: #2557

### Does this PR introduce _any_ user-facing change?

N/A

### How was this patch tested?

Test locally and UTs

PR #10480, opened 2026-03-19
[#10093] feat(core): Introduce ClassLoaderPool to share ClassLoaders across same-type catalogs

### LINKED ISSUE #10093: [Bug report] Create catalog concurrently encounter OutOfMemoryError: Metaspace (authorization disable, cache enable)
### Version

main branch

### Describe what's wrong

When I run create catalog with 10 concurrency, there's OutOfMemoryError and response is becoming slower.

Image

```
2026-02-28 07:55:22.788 WARN [Gravitino-webserver-54] [org.apache.gravitino.server.web.rest.ExceptionHandlers$CatalogExceptionHandler.handle(ExceptionHandlers.java:346)] - Failed to operate catalog(s) [mysql_catalog_1018061] operation [CREATE] under metalake [irc_test], reason [doAs method occurs an unexpected exception]
java.lang.RuntimeException: doAs method occurs an unexpected exception
	at org.apache.gravitino.utils.PrincipalUtils.doAs(PrincipalUtils.java:51) ~[gravitino-core-1.2.0-SNAPSHOT.jar:?]
	at org.apache.gravitino.server.web.Utils.doAs(Utils.java:202) ~[gravitino-server-common-1.2.0-SNAPSHOT.jar:?]
	at org.apache.gravitino.server.web.rest.CatalogOperations.createCatalog(CatalogOperations.java:146) ~[gravitino-server-1.2.0-SNAPSHOT.jar:?]
	at jdk.internal.reflect.GeneratedMethodAccessor15.invoke(Unknown Source) ~[?:?]
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.base/java.lang.reflect.Method.invoke(Method.java:569) ~[?:?]
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:52) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:146) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:189) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$ResponseOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:176) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:93) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:478) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:400) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:81) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.ServerRuntime$1.run(ServerRuntime.java:256) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248) ~[jersey-common-2.41.jar:?]
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244) ~[jersey-common-2.41.jar:?]
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292) ~[jersey-common-2.41.jar:?]
	at org.glassfish.jersey.internal.Errors.process(Errors.java:274) ~[jersey-common-2.41.jar:?]
	at org.glassfish.jersey.internal.Errors.process(Errors.java:244) ~[jersey-common-2.41.jar:?]
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:265) ~[jersey-common-2.41.jar:?]
	at org.glassfish.jersey.server.ServerRuntime.process(ServerRuntime.java:235) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.server.ApplicationHandler.handle(ApplicationHandler.java:684) ~[jersey-server-2.41.jar:?]
	at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:394) ~[jersey-container-servlet-core-2.41.jar:?]
	at org.glassfish.jersey.servlet.WebComponent.service(WebComponent.java:346) ~[jersey-container-servlet-core-2.41.jar:?]
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:358) ~[jersey-container-servlet-core-2.41.jar:?]
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:311) ~[jersey-container-servlet-core-2.41.jar:?]
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:205) ~[jersey-container-servlet-core-2.41.jar:?]
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:799) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1656) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.apache.gravitino.server.authentication.AuthenticationFilter.lambda$doFilter$0(AuthenticationFilter.java:89) ~[gravitino-server-common-1.2.0-SNAPSHOT.jar:?]
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:712) ~[?:?]
	at java.base/javax.security.auth.Subject.doAs(Subject.java:439) ~[?:?]
	at org.apache.gravitino.utils.PrincipalUtils.doAs(PrincipalUtils.java:44) ~[gravitino-core-1.2.0-SNAPSHOT.jar:?]
	at org.apache.gravitino.server.authentication.AuthenticationFilter.doFilter(AuthenticationFilter.java:86) ~[gravitino-server-common-1.2.0-SNAPSHOT.jar:?]
	at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:193) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1626) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.apache.gravitino.server.web.VersioningFilter.doFilter(VersioningFilter.java:111) ~[gravitino-server-1.2.0-SNAPSHOT.jar:?]
	at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:193) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1626) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:552) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:600) ~[jetty-security-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:235) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:1624) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:233) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1440) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:188) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:505) ~[jetty-servlet-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1594) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:186) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1355) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.HandlerCollection.handle(HandlerCollection.java:146) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.Server.handle(Server.java:516) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.HttpChannel.lambda$handle$1(HttpChannel.java:487) ~[jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:732) [jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:479) [jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:277) [jetty-server-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311) [jetty-io-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105) [jetty-io-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104) [jetty-io-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:338) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:315) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:173) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:131) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:409) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034) [jetty-util-9.4.51.v20230217.jar:9.4.51.v20230217]
	at java.base/java.lang.Thread.run(Thread.java:840) [?:?]
Caused by: java.lang.OutOfMemoryError: Metaspace

```

### How to reproduce

locust --headless --users 10 --spawn-rate 1 --host "http://10.50.0.3:8090" --schema "catalog_test" --tags "create_catalog" --metalake irc_test --catalog iceberg --mysql-host 10.50.0.4 --mysql-port 3306 --mysql-user root --mysql-password Admin@123 --run-time 5m --csv="test_1" -f "/home/ubuntu/CIScripts/locustfile_gravitino.py"

### Additional context

_No response_

LuciferYang:
Introduce a `ClassLoaderPool` with reference counting to share `IsolatedClassLoader` instances across catalogs of the same type, and centralize ClassLoader resource cleanup into the pool's lifecycle. Add `gravitino.catalog.classloader.sharing.enabled` as a safety-net toggle to disable sharing if needed.

**Core mechanism:** Catalogs with identical isolation-relevant properties share a single `IsolatedClassLoader`. The isolation key is built from the provider plus the package, authorization provider, authentication type, Kerberos principal/keytab, and backend URIs (`metastore.uris`, `jdbc-url`, `uri`, `fs.defaultFS`). The pool uses `ConcurrentHashMap.compute()` for atomic acquire/release, and runs cleanup (JDBC driver deregistration, then the shared `ClassLoaderResourceCleanerUtils` pass over Hadoop FileSystem, ThreadLocals, commons-logging, cloud SDKs, and shutdown hooks) only when the last catalog releases the shared ClassLoader.

When sharing is disabled via `gravitino.catalog.classloader.sharing.enabled=false`, each catalog gets its own dedicated ClassLoader without going through the pool — restoring the pre-PR behavior as a fallback.

**New classes:**
- `ClassLoaderKey` — `Map<String, String>`-based key for ClassLoader sharing, decoupled from specific property names
- `ClassLoaderPool` — thread-safe pool with reference counting and lifecycle management
- `PooledClassLoaderEntry` — holds a shared ClassLoader and its reference count

**Changes to existing classes:**
- `CatalogManager` — integrates the pool into catalog creation, test connection, and close paths; adds a non-pooled path gated by `classLoaderSharingEnabled`; extracts `initCatalogWrapper()` to remove duplication between the pooled and non-pooled paths; fixes ClassLoader leaks in `testConnection()` and `getResolvedProperties()`
- `Configs` — adds `gravitino.catalog.classloader.sharing.enabled`
- `IsolatedClassLoader` — exposes `getInternalClassLoader()` so the pool can run cleanup on the underlying `URLClassLoader`
- Removes per-catalog cleanup calls from `JdbcCatalogOperations`, `IcebergCatalogOperations`, `IcebergCatalogWrapper`, `PaimonCatalogOperations`, `FilesetCatalogOperations`, `HiveCatalogOperations`, and `CatalogWrapperForREST`, since cleanup now happens once in the pool on final release

Concurrent catalog creation with different names but the same provider type causes `OutOfMemoryError: Metaspace`. Each catalog creates an independent `IsolatedClassLoader` that loads all provider JARs into Metaspace. With `MaxMetaspaceSize=512m` (default) and Iceberg catalogs consuming ~30-80 MB each, ~10 catalogs exhaust the limit.

This patch addresses three causes:
1. **No ClassLoader sharing** — same-type catalogs loaded identical classes into separate Metaspace regions
2. **ClassLoader leak in `testConnection()`** — the throwaway wrapper was never closed after a connection test
3. **Inconsistent cleanup** — only 2 of 9+ catalog types called `ClassLoaderResourceCleanerUtils`; centralizing it in the pool ensures every type is cleaned on final release

Fix: #10093

Yes. A new optional server configuration is added (since 2.0.0):

- `gravitino.catalog.classloader.sharing.enabled` — whether to share ClassLoaders across catalogs with identical isolation-relevant properties. Default is `true`. Set to `false` to give each catalog its own dedicated ClassLoader as in previous releases.

**Unit tests** (`TestClassLoaderPool` — 19 tests): acquire/release semantics, reference counting, concurrent access with 20 threads, close-during-acquire race, double-release resilience, Kerberos key isolation, backend URI isolation (metastore URIs, JDBC URLs, `uri`, fs.defaultFS), authorization provider isolation, package property isolation.

**Integration tests** (`TestClassLoaderPoolIntegration` — 6 tests): same-type catalogs share one ClassLoader; dropping one doesn't affect others; `testConnection` on the same key doesn't break a live catalog; manager close cleans up the pool; with sharing disabled each catalog gets a distinct ClassLoader; with sharing disabled dropping one doesn't affect another.

**Docker integration tests** (`@Tag("gravitino-docker-test")`, real MySQL): `IcebergClassLoaderPoolIT` (3 tests) verifies over a live JDBC backend that catalogs on the same `uri` share a ClassLoader while different `uri`s isolate, and that dropping a shared catalog / running `testConnection` keeps a sibling usable. `TestMultipleJDBCLoad` covers multiple Iceberg JDBC catalogs on one server. Both scope Iceberg's metadata lookups with `nullCatalogMeansCurrent=true` (test-only) so catalogs on separate databases of a shared MySQL container don't see each other's `iceberg_tables`.

**Existing tests**: `TestCatalogManager` passes without modification; `TestJdbcCatalogOperations` was trimmed (the driver-deregistration path it covered is now handled by the pool).

**Benchmark** (JDK 17, `-XX:MaxMetaspaceSize=512m`, `fileset` provider, 10 concurrent threads):

Metaspace growth (committed KB):

| Catalogs | Baseline (`main`) | ClassLoaderPool | Reduction |
|---|---|---|---|
| 100 | +890 | +261 | 3.4x |
| 500 | +3,280 | +82 | 40x |
| 1,000 | +6,416 | +9 | 713x |
| 5,000 | +13,969 | +67 | 209x |
| 10,000 | +40,394 | +11 | **3,672x** |

Classes loaded:

| Catalogs | Baseline | ClassLoaderPool | Reduction |
|---|---|---|---|
| 1,000 | 21,772 | 11,387 | 48% |
| 10,000 | 60,373 | 11,961 | **80%** |

Baseline Metaspace grows O(N) with catalog count. The pool stays flat at ~8.7 MB — O(number of distinct keys). No OOM or performance regression on either version. For Iceberg catalogs (~50 MB/ClassLoader), baseline OOMs at ~10 catalogs; with the pool, catalogs sharing the same key reuse a single ClassLoader, so Metaspace scales with the number of distinct configurations rather than the number of catalog instances.


yuqi1129:
We had tried it in this way, see https://github.com/apache/gravitino/pull/2644, but aborted it as a potential risk exists in classloader sharing, including
- Static value in the classloader may be changed by one catalog and used by another
- Properties in classloader initialization may affect its behaviour; however, it can't be changed, although another catalog with the same type has different properties.

These two factors mentioned above will affect the correctness of the Gravitino catalogs, so we hesitate to move on.

LuciferYang:
> We had tried it in this way, see #2644, but aborted it as a potential risk exists in classloader sharing, including
> 
> * Static value in the classloader may be changed by one catalog and used by another
> * Properties in classloader initialization may affect its behaviour; however, it can't be changed, although another catalog with the same type has different properties.
> 
> These two factors mentioned above will affect the correctness of the Gravitino catalogs, so we hesitate to move on.

@yuqi1129 Thank you for the context on #2644. The three concerns raised by @jerryshao are valid for that PR's approach — here's how this PR differs and addresses each one.

---

### Background: how this PR differs from #2644

PR #2644 proposed reusing ClassLoaders during **alter** operations on a **single catalog**, passing the old ClassLoader to the new `CatalogWrapper`. This created complex lifecycle issues — especially when catalog names change (cache invalidation closes the wrapper, which would close the shared ClassLoader) and when altered properties require a fresh ClassLoader.

This PR takes a fundamentally different approach: a **reference-counted pool** that shares ClassLoaders across **different catalog instances** based on a composite key. The alter path is unaffected — alter still invalidates the old cache entry (releasing the pool reference) and creates a new wrapper (acquiring from the pool). The pool handles whether to reuse or create a new ClassLoader based on whether the key changed.

---

### Addressing the three specific concerns

**1. "The classloader reuse for alteration makes the code a little broken"**

This PR does not change the alter path's control flow. When a catalog is altered:
- The old `CatalogWrapper` is invalidated from the Caffeine cache → removal listener calls `wrapper.close()` → `pool.release(poolEntry)` decrements refCount
- A new `CatalogWrapper` is created via `createCatalogWrapper()` → `pool.acquire(key, factory)` either reuses an existing ClassLoader (if other catalogs share the same key) or creates a new one
- The alter code in `CatalogManager` remains unchanged — it still invalidates and recreates as before

The complex `setShouldCloseClassLoader` / conditional invalidation pattern from #2644 is not needed because the pool's reference counting handles the lifecycle automatically.

**2. "If we modify some properties that require rebooting and refreshing the classloader, with this we cannot support it"**

This concern applies when alter changes properties that affect ClassLoader construction. In this PR, the `ClassLoaderKey` is built from the catalog's current properties at creation time. If an alter changes a property that is part of the key (e.g., `Catalog.PROPERTY_PACKAGE`, authorization plugin path, or Kerberos identity), the new wrapper will have a **different key** and get a **fresh ClassLoader**. The old ClassLoader's refCount is decremented by the close of the old wrapper, and will be cleaned up when no other catalogs reference it.

If the alter changes a property that is NOT part of the key (e.g., metastore URI, JDBC URL), the ClassLoader is reused — which is correct, because these properties do not affect ClassLoader construction (they are consumed by `CatalogOperations.initialize()`, which runs per-instance on the new wrapper).

**3. "Sharing classloader between catalogs is dangerous, because some static variables that were created based on logic A will be used by another catalog that are based on logic B"**

I audited all catalog implementations for static mutable state:

- **Instance-level state (safe):** All catalog operations classes store runtime configuration in instance fields — `dataSource`, `clientPool`, `icebergCatalogWrapper`, `hadoopConf`, `fileSystemCache`, etc. These are per-catalog-instance and are not affected by ClassLoader sharing.

- **`UserGroupInformation.loginUser` (addressed):** The only unkeyed static state that affects correctness. `ClassLoaderKey` includes Kerberos principal and keytab via isolation properties, so catalogs with different identities get separate ClassLoaders.

- **`IcebergHiveCachedClientPool.clientPoolCache` (safe):** Static, but keyed by `(metastore URI, catalog name, user credentials)`. Different catalog configurations produce different cache entries — sharing the ClassLoader does not cause cross-contamination.

- **`FileSystem.CACHE` (safe):** Static, but keyed by `(scheme, authority, UGI)`. Same isolation guarantee as above.

The key insight is that ClassLoader construction in Gravitino depends only on JAR paths (provider + package + auth plugin), not on catalog runtime properties. Two catalogs with the same `ClassLoaderKey` load exactly the same classes — their "logic A" and "logic B" differ only at the instance level (different metastore URIs, different JDBC URLs, etc.), which is stored in instance fields and not affected by ClassLoader sharing.

---

### Future extensibility

`ClassLoaderKey` stores isolation properties as a generic `Map<String, String>`, decoupled from any specific property names. This makes the pool infrastructure key-agnostic — only the logic that builds the key needs to know which properties matter.

If new ClassLoader-scoped static state is discovered in the future, the isolation criteria can be extended by introducing a server configuration such as:

```properties
gravitino.catalog.classloader.isolation.properties = authentication.type,authentication.kerberos.principal,authentication.kerberos.keytab-uri
```

`CatalogManager` would read this list at startup and extract matching values from catalog properties when building the key. Operators can then add isolation criteria for environment-specific static state without code changes or recompilation. The `ClassLoaderKey`, `ClassLoaderPool`, and `PooledClassLoaderEntry` require no modification — only the source of property keys changes. **If the community prefers, I can implement this configurable approach in the current PR.**

Since I don't have a thorough understanding of the project yet, please correct me if I've misunderstood anything.

yuqi1129 (commented):
Thanks for this PR — the OOM/Metaspace problem is real and the benchmark numbers are impressive. I spent some time studying the design carefully and have a few concerns about correctness under certain real-world configurations. I'd like to discuss these before merging.

---

## Background: What ClassLoader sharing gives up

`IsolatedClassLoader` was designed to give every catalog its own isolated class space so that third-party library static state (Hadoop UGI, FileSystem cache, JDBC DriverManager, HiveConf, etc.) cannot bleed between catalogs. `ClassLoaderPool` partially relaxes that isolation. The question is whether `ClassLoaderKey` captures **all** the dimensions along which static state can diverge between two catalogs of the same type.

---

## Concern 1: `ClassLoaderKey` is missing critical backend-URI dimensions

The current key is:
```
provider + packageProperty + authorizationPkgPath + kerberosPrincipal + kerberosKeytab
```

This is correct for isolating Kerberos identity, but it doesn't account for catalogs that point to **different backends** of the same type. Consider:

**Scenario — two Iceberg catalogs with different HMS URIs:**
```
catalog-A: provider=lakehouse-iceberg, metastore.uris=thrift://hms-A:9083
catalog-B: provider=lakehouse-iceberg, metastore.uris=thrift://hms-B:9083
```
Both produce the **same** `ClassLoaderKey` and share one `IsolatedClassLoader`. Inside that ClassLoader, `HiveConf` has a static configuration space. `HiveConf.ConfVars` and the `valuesPerLabel` cache are static. If catalog-B's initialization overwrites catalog-A's HMS URI in `HiveConf`, catalog-A starts talking to the wrong metastore.

Same problem exists for:
- **`fs.defaultFS`** (Hadoop `FileSystem.CACHE` is a per-ClassLoader static keyed on `URI + conf + UserGroupInformation`; two catalogs pointing at different HDFS clusters may cross-contaminate the FileSystem cache)
- **JDBC URL** (less likely to corrupt static state, but the `AbandonedConnectionCleanupThread` and driver registry are global per-ClassLoader)

**Suggested fix:** extend `ClassLoaderKey` to include the backend URI(s) that anchor static state. For Iceberg/Paimon Hive-backend: `metastore.uris`. For JDBC catalogs: `jdbc-url`. For fileset catalogs: `fs.defaultFS`.

---

## Concern 2: `FileSystem.closeAll()` during `doFinalCleanup` can disconnect live catalogs

`closeStatsDataClearerInFileSystem` calls `FileSystem.closeAll()` — a static method that closes **every** cached FileSystem in that ClassLoader's cache. `doFinalCleanup` runs only when `refCount` reaches 0 (i.e., the last catalog sharing this ClassLoader is closed), so under the current logic it won't fire while other catalogs are live.

However, if Concern 1 is fixed and two catalogs with different HDFS URIs are given separate keys, this is safe. But as long as `ClassLoaderKey` is under-specified (same key for different backends), there is a window where catalog-A's cleanup kills catalog-B's live HDFS connections.

---

## Concern 3: ThreadLocal cross-contamination during shared lifetime

With per-catalog ClassLoaders, ThreadLocal values from different catalogs are of different `Class` objects (loaded by different ClassLoaders), so they are naturally isolated. With a shared ClassLoader, Catalog A and Catalog B load the **same** Class, meaning a ThreadLocal set by catalog-A's code on a Jetty thread is **visible to catalog-B's code** running on the same thread.

Concretely: if Iceberg or Hive sets a per-request ThreadLocal (e.g., Hadoop `SecurityContext`, Hive `SessionState`, or Iceberg's `ResolvingFileIO` context), catalog-B could pick up a stale value left by catalog-A. This is especially risky for Hive `SessionState`, which is a ThreadLocal singleton used throughout HMS interaction.

---

## Comparison with industry practice

The closest industry analogue is **Trino's plugin ClassLoader**: a single ClassLoader is shared across all connector instances of the same plugin. Trino makes this safe because its connectors are **stateless** — all per-request state lives in `ConnectorSession`, not in static fields. Gravitino's catalogs are stateful (they hold HMS connections, HDFS FileSystems, UGI login state), which is the key difference that makes sharing riskier here.

---

## Summary of concerns

| Issue | Severity | Affected catalogs |
|---|---|---|
| `ClassLoaderKey` missing `metastore.uris` / `fs.defaultFS` | **High** | Iceberg-Hive, Hive, Paimon-Hive, fileset |
| `FileSystem.closeAll()` can disconnect live catalog (if keys are under-specified) | **High** (conditional on above) | Any catalog using HDFS |
| ThreadLocal cross-contamination between shared-ClassLoader catalogs | **Medium** | Iceberg, Hive (SessionState) |
| AWS SDK v2 static state not cleaned up (only v1 handled) | **Low** | Iceberg/Paimon with S3 backend |

---

## Suggestion

The OOM fix is valuable and I'd like to see it merged. One path forward:

1. **Extend `ClassLoaderKey`** to include the backend URI dimensions that anchor per-ClassLoader static state. This makes sharing conservative (fewer catalogs will share) but correct.
2. **Or**, limit sharing to the case where **all config properties are identical** (not just the 5 current dimensions). This is the safest interpretation of "same type" — if two catalogs are truly identical in configuration, they genuinely cannot diverge in static state.
3. Add a test that creates two same-type catalogs pointing at **different backends** and verifies they get separate ClassLoader entries in the pool.

Happy to discuss the trade-offs — the refCount/cleanup mechanics look solid and the testConnection leak fix is a clear win regardless.

LuciferYang:
@yuqi1129 Thanks for your valuable advice. I'm a bit busy today, but I'll take care of the this tomorrow.

LuciferYang:
The failure of the S3 integration test appears to be unrelated to the current pr

LuciferYang:
Thanks @yuqi1129 for the thorough review. Here is the status of each concern.

### Concern 1: `ClassLoaderKey` missing backend-URI dimensions — Fixed

Extended the isolation key with `metastore.uris`, `jdbc-url`, and `fs.defaultFS`. Also added `authorization-provider` (determines which authorization plugin JARs are loaded).

The full set of built-in default isolation keys (catalog property names) is now:

| Category | Property keys |
|---|---|
| Classpath | `package` ([Catalog.PROPERTY_PACKAGE](https://github.com/apache/gravitino/blob/main/api/src/main/java/org/apache/gravitino/Catalog.java#L132)), `authorization-provider` |
| Kerberos identity | `authentication.type`, `authentication.kerberos.principal`, `authentication.kerberos.keytab-uri` |
| Backend URIs | `metastore.uris`, `jdbc-url`, `fs.defaultFS` |

These defaults cannot be removed. Operators can add more via a new server config:

```properties
gravitino.catalog.classloader.isolation.extra-properties = custom.backend.endpoint
```

`ClassLoaderKey` stores isolation properties as a generic `Map<String, String>`, decoupled from specific property names — only `CatalogManager.buildClassLoaderKey` needs to know which properties matter. This makes the pool infrastructure key-agnostic and extensible without modifying pool classes.

### Concern 2: `FileSystem.closeAll()` can disconnect live catalogs — Resolved by Concern 1 fix

With backend URIs now in the key, catalogs pointing at different HDFS clusters get separate ClassLoaders. `doFinalCleanup` only runs when `refCount` reaches 0, so `FileSystem.closeAll()` cannot affect live catalogs sharing the same ClassLoader.

### Concern 3: ThreadLocal cross-contamination — Not fully resolved, inherent trade-off

This is a genuine limitation of ClassLoader sharing. With per-catalog ClassLoaders, ThreadLocal values are naturally isolated because each ClassLoader loads its own copy of the class. With a shared ClassLoader, catalogs on the same thread can see each other's ThreadLocal state.

The Concern 1 fix reduces the blast radius — catalogs sharing a ClassLoader now have identical backend configurations, so leaked ThreadLocal state is less likely to cause incorrect behavior (e.g., talking to the wrong metastore). But it does not eliminate the problem. If a library sets a ThreadLocal with per-catalog state (e.g., Hive `SessionState`, Hadoop `SecurityContext`), cross-contamination is still possible between catalogs sharing the same ClassLoader on the same thread.

**What this PR does:**
- Reduces exposure by isolating catalogs with different backends into separate ClassLoaders (Concern 1 fix)
- Provides `extra-properties` config as an operational escape hatch — if a specific ThreadLocal issue surfaces, operators can add the relevant property key to force separation without a code release

**What this PR does not do:**
- It does not guarantee ThreadLocal isolation between catalogs sharing a ClassLoader. This is an inherent trade-off of sharing and cannot be fully solved at the key level.

**Open question for the community:** Is this trade-off acceptable given the Metaspace savings, or should we consider additional safeguards (e.g., a per-catalog opt-out property to force a dedicated ClassLoader)? Happy to discuss.

### Concern 4: AWS SDK v2 static state — Not addressed in this PR

Low severity, not related to the key design. Can be addressed as a follow-up if this one merged.

### Suggestion 3: Tests for different-backend isolation — Added

New tests:
- `testDifferentMetastoreUrisCreateDifferentEntries`
- `testSameMetastoreUrisShareEntry`
- `testDifferentJdbcUrlsCreateDifferentEntries`
- `testDifferentDefaultFsCreateDifferentEntries`
- `testKeyWithAuthorizationProvider`

Total: 19 unit tests + 3 integration tests.

yuqi1129:
> Thanks @yuqi1129 for the thorough review. Here is the status of each concern.
> 
> ### Concern 1: `ClassLoaderKey` missing backend-URI dimensions — Fixed
> Extended the isolation key with `metastore.uris`, `jdbc-url`, and `fs.defaultFS`. Also added `authorization-provider` (determines which authorization plugin JARs are loaded).
> 
> The full set of built-in default isolation keys (catalog property names) is now:
> 
> Category	Property keys
> Classpath	`package` ([Catalog.PROPERTY_PACKAGE](https://github.com/apache/gravitino/blob/main/api/src/main/java/org/apache/gravitino/Catalog.java?rgh-link-date=2026-03-24T07%3A12%3A45Z#L132)), `authorization-provider`
> Kerberos identity	`authentication.type`, `authentication.kerberos.principal`, `authentication.kerberos.keytab-uri`
> Backend URIs	`metastore.uris`, `jdbc-url`, `fs.defaultFS`
> These defaults cannot be removed. Operators can add more via a new server config:
> 
> ```ini
> gravitino.catalog.classloader.isolation.extra-properties = custom.backend.endpoint
> ```
> 
> `ClassLoaderKey` stores isolation properties as a generic `Map<String, String>`, decoupled from specific property names — only `CatalogManager.buildClassLoaderKey` needs to know which properties matter. This makes the pool infrastructure key-agnostic and extensible without modifying pool classes.
> 
> ### Concern 2: `FileSystem.closeAll()` can disconnect live catalogs — Resolved by Concern 1 fix
> With backend URIs now in the key, catalogs pointing at different HDFS clusters get separate ClassLoaders. `doFinalCleanup` only runs when `refCount` reaches 0, so `FileSystem.closeAll()` cannot affect live catalogs sharing the same ClassLoader.
> 
> ### Concern 3: ThreadLocal cross-contamination — Not fully resolved, inherent trade-off
> This is a genuine limitation of ClassLoader sharing. With per-catalog ClassLoaders, ThreadLocal values are naturally isolated because each ClassLoader loads its own copy of the class. With a shared ClassLoader, catalogs on the same thread can see each other's ThreadLocal state.
> 
> The Concern 1 fix reduces the blast radius — catalogs sharing a ClassLoader now have identical backend configurations, so leaked ThreadLocal state is less likely to cause incorrect behavior (e.g., talking to the wrong metastore). But it does not eliminate the problem. If a library sets a ThreadLocal with per-catalog state (e.g., Hive `SessionState`, Hadoop `SecurityContext`), cross-contamination is still possible between catalogs sharing the same ClassLoader on the same thread.
> 
> **What this PR does:**
> 
> * Reduces exposure by isolating catalogs with different backends into separate ClassLoaders (Concern 1 fix)
> * Provides `extra-properties` config as an operational escape hatch — if a specific ThreadLocal issue surfaces, operators can add the relevant property key to force separation without a code release
> 
> **What this PR does not do:**
> 
> * It does not guarantee ThreadLocal isolation between catalogs sharing a ClassLoader. This is an inherent trade-off of sharing and cannot be fully solved at the key level.
> 
> **Open question for the community:** Is this trade-off acceptable given the Metaspace savings, or should we consider additional safeguards (e.g., a per-catalog opt-out property to force a dedicated ClassLoader)? Happy to discuss.
> 
> ### Concern 4: AWS SDK v2 static state — Not addressed in this PR
> Low severity, not related to the key design. Can be addressed as a follow-up if this one merged.
> 
> ### Suggestion 3: Tests for different-backend isolation — Added
> New tests:
> 
> * `testDifferentMetastoreUrisCreateDifferentEntries`
> * `testSameMetastoreUrisShareEntry`
> * `testDifferentJdbcUrlsCreateDifferentEntries`
> * `testDifferentDefaultFsCreateDifferentEntries`
> * `testKeyWithAuthorizationProvider`
> 
> Total: 19 unit tests + 3 integration tests.

I will take time to review it again. Thanks for your quick response.

LuciferYang:
Thank you @yuqi1129

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -251,7 +290,22 @@ public void close() {
         LOG.warn("Failed to close catalog", e);
       }
 
-      classLoader.close();
+      if ((pool == null) != (poolEntry == null)) {
```
yuqi1129: When will `pool` or `poolEntry` be null?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -251,7 +290,22 @@ public void close() {
...
+        // This state should never occur. Prefer leaking one ClassLoader over destroying a
+        // ClassLoader that may still be shared by other catalogs through the pool.
+        LOG.warn(
+            "Inconsistent state: pool={}, poolEntry={}. Skipping cleanup to avoid corruption.",
+            pool,
+            poolEntry);
+        return;
+      }
+
+      if (pool != null && poolEntry != null) {
```
yuqi1129: `poolEntry != null` should always true if the code goes here.

core/src/main/java/org/apache/gravitino/utils/ClassLoaderPool.java
```diff
@@ -0,0 +1,201 @@
...
+    isolatedClassLoader.close();
+    LOG.info("ClassLoader for key {} has been fully cleaned up.", entry.key());
+  }
+
+  /**
+   * Deregisters all JDBC drivers that were loaded by the given ClassLoader.
+   *
+   * @param classLoader The ClassLoader whose drivers should be deregistered.
+   */
+  private void deregisterAllDrivers(ClassLoader classLoader) {
```
yuqi1129: Why do we need to deregister JDBC drivers here? I remembered that the drivers should have been deregistered in the catalog.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
LuciferYang: Agreed — after the XOR guard, `pool != null` implied `poolEntry != null`, making the `&& poolEntry != null` check redundant. Simplified by removing the XOR guard and using `poolEntry` as the sole discriminator.

We intentionally check `poolEntry != null` rather than `pool != null` to handle repeated `close()` safely: after the first close, `poolEntry` is nulled but `pool` is not; a subsequent `if (pool != null)` would pass `null` to `pool.release()` and NPE on `entry.key()`. Keying on `poolEntry` distinguishes first close from repeated close without an extra guard.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
LuciferYang: `CatalogWrapper` has two constructors: a two-arg form `(BaseCatalog, IsolatedClassLoader)` that leaves both `pool` and `poolEntry` null, and a four-arg form that sets both. Currently only the four-arg constructor is called (by `createCatalogWrapper()`), so in practice both fields are always non-null. The two-arg constructor is kept as a public API for potential non-pooled paths.

Since the two fields are always set as a pair, the XOR guard `(pool == null) != (poolEntry == null)` could never trigger. Removed it and restructured `close()` accordingly.

core/src/main/java/org/apache/gravitino/utils/ClassLoaderPool.java
LuciferYang: With shared ClassLoaders, per-catalog driver deregistration is unsafe. Previously each catalog had its own ClassLoader and driver instance, so `JdbcCatalogOperations.close()` could safely deregister on close. Now multiple catalogs share a single ClassLoader and its `DriverManager` registration — if catalog A deregisters the driver, catalog B (still active, same ClassLoader) loses it.

The pool's `doFinalCleanup()` runs only when the reference count reaches zero, meaning no catalog is using the ClassLoader anymore. This is the only safe point for destructive cleanup. The PR therefore centralizes all such cleanup into the pool:

| Removed from | What was removed | Now handled by |
|---|---|---|
| `JdbcCatalogOperations.close()` | `deregisterDriver()` | `ClassLoaderPool.deregisterAllDrivers()` |
| `IcebergCatalogOperations.close()` | `ClassLoaderResourceCleanerUtils.closeClassLoaderResource()` | `ClassLoaderPool.doFinalCleanup()` |
| `PaimonCatalogOperations.close()` | `ClassLoaderResourceCleanerUtils.closeClassLoaderResource()` | `ClassLoaderPool.doFinalCleanup()` |
| `IcebergCatalogWrapper.close()` | JDBC driver deregistration + MySQL `AbandonedConnectionCleanupThread` shutdown | `ClassLoaderPool.deregisterAllDrivers()` + `ClassLoaderResourceCleanerUtils.shutdownMySQLAbandonedConnectionCleanupThread()` |

Note on the **Iceberg REST server** path: `CatalogWrapperForREST` previously overrode `useDifferentClassLoader()` to return `false`, skipping driver deregistration because the REST server reuses the same app classloader across wrapper recreations. This override was removed, but the behavior is preserved — `IcebergCatalogWrapper.close()` no longer does any driver cleanup at all, and the REST server does not go through `CatalogManager` or the `ClassLoaderPool`, so pool cleanup is not triggered. The net effect is the same: no driver deregistration on the REST server path.

yuqi1129:
@LuciferYang In case of any possible problem, can you add a configuration item to enable classloader sharing or not?

LuciferYang:
> @LuciferYang In case of any possible problem, can you add a configuration item to enable classloader sharing or not?

will handle it tomorrow.

LuciferYang:
Still debugging, no need for review yet.

LuciferYang:
> @LuciferYang In case of any possible problem, can you add a configuration item to enable classloader sharing or not?

[bd911bb](https://github.com/apache/gravitino/pull/10480/commits/bd911bb891f98a1e44cc77f64b6ec806122a6a46) address this

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -190,6 +190,31 @@ private Configs() {}
...
 
+  public static final ConfigEntry<Boolean> CATALOG_CLASSLOADER_SHARING_ENABLED =
+      new ConfigBuilder("gravitino.catalog.classloader.sharing.enabled")
+          .doc(
+              "Whether to share ClassLoaders across catalogs with identical isolation-relevant "
+                  + "properties. When true (default), catalogs with the same isolation key reuse "
+                  + "a single ClassLoader, significantly reducing Metaspace memory usage. When "
+                  + "false, each catalog gets its own dedicated ClassLoader as in previous "
+                  + "releases.")
+          .version(ConfigConstants.VERSION_1_2_0)
```
yuqi1129: The version should be 1.3.0(next major version).

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -190,6 +190,31 @@ private Configs() {}
...
+
+  public static final ConfigEntry<List<String>> CATALOG_CLASSLOADER_ISOLATION_EXTRA_PROPERTIES =
+      new ConfigBuilder("gravitino.catalog.classloader.isolation.extra-properties")
+          .doc(
+              "Additional catalog property keys used to determine ClassLoader isolation. "
+                  + "Catalogs that differ in any of these properties will use separate "
+                  + "ClassLoaders. This supplements the built-in default isolation properties "
+                  + "(package, authorization-provider, Kerberos identity, metastore.uris, "
+                  + "jdbc-url, fs.defaultFS) and cannot be used to remove them.")
+          .version(ConfigConstants.VERSION_1_2_0)
```
yuqi1129: This configuration item is too complex for users to use. I would suggest you remove it.

LuciferYang: ok, will remove it

yuqi1129:
@jerryshao 
Would you like to take a look?

core/src/main/java/org/apache/gravitino/Configs.java
LuciferYang: [a432938](https://github.com/apache/gravitino/pull/10480/commits/a432938f4d29228653aa223eca1f6f3a0b00c709) removed this config

core/src/main/java/org/apache/gravitino/Configs.java
LuciferYang: my bad, already updated

yuqi1129:
@jerryshao 
Can you help check whether this fix policy is on the right track? Thus, I can push this PR forward.

yuqi1129 (commented):
Thanks for the detailed benchmark and the thorough pool-level tests — the core acquire/release logic looks solid. A few integration-level concerns and a testing note, left inline.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -1153,9 +1237,29 @@ CatalogWrapper createCatalogWrapper(
...
+        ClassLoaderPool.cleanupClassLoader(classLoader);
+      }
+    }
+
+    ClassLoaderKey key = buildClassLoaderKey(provider, conf);
+    PooledClassLoaderEntry poolEntry =
+        classLoaderPool.acquire(key, () -> createClassLoader(provider, conf));
+    try {
+      IsolatedClassLoader classLoader = poolEntry.classLoader();
+      BaseCatalog<?> catalog = createBaseCatalog(classLoader, entity);
```
yuqi1129: This `BaseCatalog` (and the non-pooled branch at L1246) is created but never `close()`d. `BaseCatalog.close()` is what closes `authorizationPlugin`, so for any catalog with `authorization-provider` set, each `listCatalogsInfo()` call (L525 maps this over every catalog) leaks a plugin instance. The pooled branch also builds + tears down a fresh `IsolatedClassLoader` per uncached catalog on every list call. The previous impl reused the cached `loadCatalogAndWrap(...)` wrapper — suggest reusing it, or wrapping in try/finally with `catalog.close()`.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -256,7 +291,14 @@ public void close() {
         LOG.warn("Failed to close catalog", e);
       }
 
-      classLoader.close();
+      if (poolEntry != null) {
+        pool.release(poolEntry);
+        poolEntry = null;
+      } else if (pool == null) {
+        // Non-pooled path (e.g., sharing disabled or CATALOG_LOAD_ISOLATED=false)
+        ClassLoaderPool.cleanupClassLoader(classLoader);
```
yuqi1129: Unlike the pooled branch above (which nulls `poolEntry`, making a repeat close a no-op), this non-pooled branch has no idempotency guard: a second `close()` re-runs `cleanupClassLoader()` (double driver-deregister / resource cleanup / `classLoader.close()`). Since `close()` was made `synchronized`, consider a `closed` flag so both paths are symmetric.

core/src/main/java/org/apache/gravitino/utils/ClassLoaderPool.java
```diff
@@ -0,0 +1,209 @@
...
+            throw new IllegalStateException("ClassLoaderPool is already closed");
+          }
+          if (existing != null) {
+            existing.incrementRefCount();
+            LOG.debug("Reusing ClassLoader for key {}, refCount={}.", key, existing.refCount());
+            return existing;
+          }
+          // If the factory throws (e.g., invalid classpath), the exception propagates to the
+          // caller and ConcurrentHashMap leaves the key unmapped.
+          IsolatedClassLoader classLoader = factory.get();
```
yuqi1129, line 80: Minor: the classloader `factory` here (jar scanning) and `doFinalCleanup()` (thread interruption + reflection over all threads) both run inside `ConcurrentHashMap.compute`, holding the bin lock during heavy work. Fine for a small pool, just flagging for potential create/drop latency under concurrency.

core/src/test/java/org/apache/gravitino/catalog/TestClassLoaderPoolIntegration.java
```diff
@@ -0,0 +1,281 @@
...
+          .withName(METALAKE)
+          .withAuditInfo(
+              AuditInfo.builder().withCreator("test").withCreateTime(Instant.now()).build())
+          .withVersion(SchemaVersion.V_0_1)
+          .build();
+
+  @BeforeAll
+  public static void setUp() throws IOException, IllegalAccessException {
+    config = new Config(false) {};
+    config.set(Configs.CATALOG_LOAD_ISOLATED, false);
```
yuqi1129, line 72: Testing gap: these tests run with `CATALOG_LOAD_ISOLATED=false` and empty dummy classloaders, so the two core claims of this PR — actual Metaspace savings and no cross-catalog interference from a *shared* classloader — aren't exercised. Suggest adding docker-tagged ITs with real provider jars:
- **GC/leak**: hold a `WeakReference` to the `IsolatedClassLoader`, drop the catalog, `System.gc()`, assert it is collected (catches leftover ThreadLocals / lingering threads / undereg'd drivers).
- **testConnection release**: with a live catalog sharing the key, `testConnection` should acquire+release without dropping refCount to 0 and breaking the live catalog.
- **Key blind spot**: Iceberg's JDBC backend uses `uri`, not `jdbc-url` (not in `DEFAULT_ISOLATION_PROPERTY_KEYS`) — verify two mysql-backed Iceberg catalogs sharing a classloader don't cross-contaminate.
- **iceberg-rest-server regression**: the removed `useDifferentClassLoader()` override — verify REST-server JDBC-backend wrapper rebuild still works.

core/src/test/java/org/apache/gravitino/catalog/TestClassLoaderPoolIntegration.java
```diff
@@ -0,0 +1,281 @@
...
+
+    // Verify they actually share the same ClassLoader instance
+    Object classLoader1 = FieldUtils.readField(wrapper1, "classLoader", true);
+    Object classLoader2 = FieldUtils.readField(wrapper2, "classLoader", true);
+    Assertions.assertSame(
+        classLoader1, classLoader2, "Same-type catalogs should share a ClassLoader");
+  }
+
+  @Test
+  public void testClosingOneCatalogDoesNotAffectOthers() {
```
yuqi1129, line 145: For real coverage this should issue an actual request on `catalog2` after dropping `catalog1` (with real provider jars), not just `loadCatalog`. The cleanup-on-final-release path (JDBC driver deregister, `AbandonedConnectionCleanupThread` shutdown) can only be caught by exercising `catalog2` after `catalog1`'s drop when they share a classloader.

yuqi1129:
### Testing suggestion

A high-level note that spans several files, so leaving it here rather than inline.

The unit/integration tests use empty dummy classloaders (`CATALOG_LOAD_ISOLATED=false`), which validates the reference-counting logic but **not** the two things this PR actually claims: real Metaspace savings, and no cross-catalog interference when a classloader is genuinely *shared*. Those are runtime-only properties. I'd add docker-tagged ITs (`@Tag("gravitino-docker-test")`) with real provider jars covering:

**A. Memory / GC actually reclaimed**
- Hold a `WeakReference` to the `IsolatedClassLoader`, drop the catalog so refCount → 0, `System.gc()`, and assert `ref.get() == null`. Leftover ThreadLocals, lingering threads, or un-deregistered drivers will keep it alive and fail here — exactly what the current tests can't catch.
- Run a create-loop under `-XX:MaxMetaspaceSize` with `-Xlog:class+unload` / NMT and assert committed Metaspace stays flat with sharing on vs. O(N) with it off.

**B. Shared classloader must not "cross-talk"**
- Two catalogs with the same key but different other config (e.g. two `jdbc-mysql` on different DBs): DDL on A must not affect B.
- Iceberg's JDBC backend uses the `uri` property, **not** `jdbc-url`, so it's not in `DEFAULT_ISOLATION_PROPERTY_KEYS` — two mysql-backed Iceberg catalogs will share one classloader/DriverManager registry; verify no contamination.
- Same vs. different Kerberos principal: do a real UGI login and confirm sharing doesn't overwrite `loginUser`.

**C. Lifecycle timing (refCount boundaries)**
- A & B share a real MySQL classloader; drop A, then **issue a real query on B** — assert the driver isn't deregistered and `AbandonedConnectionCleanupThread` isn't shut down. (`testClosingOneCatalogDoesNotAffectOthers` only re-loads, so it can't see this.)
- `testConnection` now calls `wrapper.close()`: with a live catalog sharing the key, `testConnection` must acquire+release without dropping refCount to 0 and breaking the live catalog; with no live catalog it builds+tears down a classloader and deregisters the driver — verify a subsequent real `create` re-registers cleanly.
- Concurrent create/drop of same-key catalogs, driving refCount 0 ↔ N with GC, to shake out "cleaned up but still in use".

**D. Regressions from removed cleanup**
- iceberg-rest-server: the removed `useDifferentClassLoader()` override / JDBC cleanup — verify a REST-server JDBC-backend wrapper rebuild still works (the original comment warned the driver couldn't be reloaded after deregister).
- `sharing.enabled=false`: assert behavior matches pre-PR (per-catalog classloader + per-catalog cleanup), as a baseline regression guard.

Highest-ROI to start with: **A** (WeakReference GC assert) and **C** (real-driver drop / testConnection not hitting a live sharer) — those directly exercise the cleanup paths that pure counting tests can't reach.

yuqi1129:
@LuciferYang 
Do you have time to move this one forward?

LuciferYang:
@yuqi1129 While adding the docker-tagged ITs suggested in the review (r3535110319 and the testing issuecomment), the new `IcebergClassLoaderPoolIT` surfaced a failure in the existing `TestMultipleJDBCLoad`: when both run in the **same test JVM**, `TestMultipleJDBCLoad` fails with `Table '...iceberg_tables' doesn't exist`.

I bisected it and — importantly — **reproduced it on `main` (without any `ClassLoaderPool`)** using a minimal probe:

- `ProbeMainDropIT` (creates a MySQL-backed Iceberg catalog, then drops it) **alone** → passes
- `TestMultipleJDBCLoad` **alone** → passes
- the two **in the same JVM** → `TestMultipleJDBCLoad` fails identically (`iceberg_tables doesn't exist`)

**Root cause (pre-existing on `main`, independent of this PR):** `IcebergCatalogWrapper.close()` unconditionally calls `AbandonedConnectionCleanupThread.uncheckedShutdown()` when a MySQL catalog is dropped. That is a **JVM-global** side effect, and it corrupts the shared MySQL/JDBC state of a *sibling* MySQL-backed Iceberg catalog living in the same JVM. It was simply never triggered before, because no existing test **drops** a real MySQL-backed Iceberg catalog in a shared JVM (`TestMultipleJDBCLoad` only *creates* catalogs). This PR's pool inherits the same global shutdown (relocated into `doFinalCleanup`), so the symptom is identical — the pool neither introduces nor worsens it.

Since this is an existing test-isolation gap that is independent of the pool, I'd propose addressing it in a **precursor PR** — e.g. `forkEvery(1)` for the iceberg IT task so each IT class runs in its own JVM (which also mirrors production, where each server is a separate process) — and then rebasing this PR on top. Verified locally: with `forkEvery(1)`, the full combination (`IcebergClassLoaderPoolIT` + `TestMultipleJDBCLoad`) passes.

Does that approach sound right to you, or would you prefer it bundled into this PR?

yuqi1129:
> @yuqi1129 While adding the docker-tagged ITs suggested in the review (r3535110319 and the testing issuecomment), the new `IcebergClassLoaderPoolIT` surfaced a failure in the existing `TestMultipleJDBCLoad`: when both run in the **same test JVM**, `TestMultipleJDBCLoad` fails with `Table '...iceberg_tables' doesn't exist`.
> 
> I bisected it and — importantly — **reproduced it on `main` (without any `ClassLoaderPool`)** using a minimal probe:
> 
> * `ProbeMainDropIT` (creates a MySQL-backed Iceberg catalog, then drops it) **alone** → passes
> * `TestMultipleJDBCLoad` **alone** → passes
> * the two **in the same JVM** → `TestMultipleJDBCLoad` fails identically (`iceberg_tables doesn't exist`)
> 
> **Root cause (pre-existing on `main`, independent of this PR):** `IcebergCatalogWrapper.close()` unconditionally calls `AbandonedConnectionCleanupThread.uncheckedShutdown()` when a MySQL catalog is dropped. That is a **JVM-global** side effect, and it corrupts the shared MySQL/JDBC state of a _sibling_ MySQL-backed Iceberg catalog living in the same JVM. It was simply never triggered before, because no existing test **drops** a real MySQL-backed Iceberg catalog in a shared JVM (`TestMultipleJDBCLoad` only _creates_ catalogs). This PR's pool inherits the same global shutdown (relocated into `doFinalCleanup`), so the symptom is identical — the pool neither introduces nor worsens it.
> 
> Since this is an existing test-isolation gap that is independent of the pool, I'd propose addressing it in a **precursor PR** — e.g. `forkEvery(1)` for the iceberg IT task so each IT class runs in its own JVM (which also mirrors production, where each server is a separate process) — and then rebasing this PR on top. Verified locally: with `forkEvery(1)`, the full combination (`IcebergClassLoaderPoolIT` + `TestMultipleJDBCLoad`) passes.
> 
> Does that approach sound right to you, or would you prefer it bundled into this PR?

Thanks for your hard work and for reporting this issue. I believe we can create a separate PR to solve the JDBC class clean-up problem first, then continue with this one. The current PR is already rather large, and an extra fix will hinder its being merged in a short time.

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -220,6 +220,18 @@ private Configs() {}
...
 
+  public static final ConfigEntry<Boolean> CATALOG_CLASSLOADER_SHARING_ENABLED =
+      new ConfigBuilder("gravitino.catalog.classloader.sharing.enabled")
+          .doc(
+              "Whether to share ClassLoaders across catalogs with identical isolation-relevant "
+                  + "properties. When true (default), catalogs with the same isolation key reuse "
+                  + "a single ClassLoader, significantly reducing Metaspace memory usage. When "
+                  + "false, each catalog gets its own dedicated ClassLoader as in previous "
+                  + "releases.")
+          .version(ConfigConstants.VERSION_1_3_0)
```
yuqi1129: The version here will be 2.0.0, and `2.0.0` is the next major version.

LuciferYang:
Correcting my root-cause analysis above — I was wrong about why `TestMultipleJDBCLoad` fails.

It's not `AbandonedConnectionCleanupThread`. The real cause is Iceberg's `JdbcCatalog.atomicCreateTable`: before creating its control table it runs `getTables(null, null, "iceberg_tables", null)`, and with MySQL Connector/J's default `nullCatalogMeansCurrent=false` a `null` catalog scans every database on the server. These ITs put several Iceberg JDBC catalogs on separate databases of one shared MySQL container, so the second catalog sees the first's `iceberg_tables`, skips creating its own, and later queries fail with `doesn't exist`. A standalone JDBC probe confirms it: `getTables(null,...)` returns the sibling DB's row by default, 0 rows with `nullCatalogMeansCurrent=true`.

Fix is test-only and already in this PR: append `nullCatalogMeansCurrent=true` to the MySQL URIs in `IcebergClassLoaderPoolIT` and `TestMultipleJDBCLoad`. Gravitino passes the user's `uri` through unchanged, so this is a test artifact rather than a product bug, and it's the documented setting for running multiple Iceberg JDBC catalogs on one server. `forkEvery(1)` isn't needed. The combined run fails without the change and is green with it.

On the separate PR we agreed on: that was based on my wrong analysis, so I'd re-scope it. The IT failure is handled here. There is a genuine, pre-existing Metaspace leak — MySQL's `AbandonedConnectionCleanupThread` keeps the isolated ClassLoader as its context CL after a catalog is dropped — but it's unrelated to this failure and to the pool, and this PR doesn't depend on it (reverting that hardening keeps the suite green). I'll take it as a follow-up after this PR merges, with its own test, rather than as a precursor.

yuqi1129:
@LuciferYang , the code is overall LGTM except for the following thing, and we will try to merge it within a week.

- Add a configuration item in the `conf/gravitino.conf.template` since the change is a major feature and affects a lot.
- Could you add some description in the document `docs/gravitino-server-config.md`.

yuqi1129 (dismissed):
LGTM

yuqi1129:
@jerryshao 
Would you like to take a look?

yuqi1129 (approved):

LuciferYang:
Thank you @yuqi1129

Issue #9736, opened 2026-01-16
[FEATURE] Support freshness-aware table loading for Gravitino Iceberg REST server


FANNG1:
### Describe the feature

The main idea is the IRC server could generate a ETag with the table meta data. The client could add the ETag when requesting the table metadata, if the table metadata is not changed, the server could reply HTTP status code 304 not table metadata, then the client side could keep the cache. It will burst the performance for the compute engine which caches the Iceberg metadata like Impala.

For more details, please refer to:
https://docs.google.com/document/d/1rnVSP_iv2I47giwfAe-Z3DYhKkKwWCVvCkC9rEvtaLA/edit?pli=1&tab=t.0  and https://github.com/apache/iceberg/pull/11946 
https://docs.google.com/document/d/1wcT9YvqXO3U9XitcA3sj38M946TZ_AkfOq0U9IZpKAc/edit?tab=t.0

### Describe the solution

After 8696, it could generate an ETag based on the table metadata location. If the metadata location is not changed, the IRC server could respond with 304

roryqi:
@laserninja Could u comment here? I can assign this issue to u after comment.

laserninja:
Yes, looks like it was merged. Thanks!

PR #10498, opened 2026-03-22
[#9736] feat(iceberg-rest): Support freshness-aware table loading with ETag

laserninja:
Implement HTTP ETag-based conditional request support for the Iceberg REST server's loadTable endpoint to enable freshness-aware caching.

When loading a table, the server now generates an ETag header from a SHA-256 hash of the table's metadata file location. If the client sends an If-None-Match header matching the current ETag, the server responds with HTTP 304 Not Modified instead of the full table metadata payload.

This allows compute engines that cache Iceberg metadata (e.g., Impala) to skip re-fetching unchanged metadata, improving read-heavy workload performance.

Changes:
- Add If-None-Match header parameter to IcebergTableOperations.loadTable
- Generate ETag from metadata file location via SHA-256 hash
- Return 304 Not Modified when client ETag matches, 200 with ETag otherwise
- Add tests: ETag presence, 304 on match, 200 on mismatch, ETag changes after update, and ETag consistency across loads

Closes: #9736

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -193,7 +193,7 @@ allprojects {
       param.environment("GRAVITINO_CI_TRINO_DOCKER_IMAGE", "apache/gravitino-ci:trino-0.1.6")
       param.environment("GRAVITINO_CI_RANGER_DOCKER_IMAGE", "apache/gravitino-ci:ranger-0.1.2")
       param.environment("GRAVITINO_CI_KAFKA_DOCKER_IMAGE", "apache/kafka:3.7.0")
-      param.environment("GRAVITINO_CI_LOCALSTACK_DOCKER_IMAGE", "localstack/localstack:latest")
+      param.environment("GRAVITINO_CI_LOCALSTACK_DOCKER_IMAGE", "localstack/localstack:4.14.0")
 
       // Disable Ryuk for integration tests
       // Ryuk need privileged mode, if we want to rootless or run non-privileged mode, we need to disable it.


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -306,7 +313,17 @@ public Response loadTable(
                 new IcebergRequestContext(httpServletRequest(), catalogName, isCredentialVending);
             LoadTableResponse loadTableResponse =
                 tableOperationDispatcher.loadTable(context, tableIdentifier);
-            return IcebergRESTUtils.ok(loadTableResponse);
+            String metadataLocation = loadTableResponse.tableMetadata().metadataFileLocation();
```
roryqi: Create table response should return etag, too. You can refer to the document https://github.com/apache/iceberg/blob/a8e9ad2ae02fee5766f71a0783f402680da8e219/open-api/rest-catalog-open-api.yaml#L4960

laserninja: Sounds good, updated. createTable and updateTable responses now also include an ETag header computed from the metadata file location, consistent with loadTable. Added tests for both. Extracted a buildResponseWithETag() helper to avoid duplication across the three endpoints.

roryqi: Could u add some tests for 
```
        Implementations that support ETags should produce unique tags for responses that return
        different metadata content but represent the same version of table metadata.  For example,
        the `snapshots` query parameter may result in different metadata representations depending
        on whether `refs` or `all` is provided, therefore should have distinct ETags.
```
https://github.com/apache/iceberg/blob/a8e9ad2ae02fee5766f71a0783f402680da8e219/open-api/rest-catalog-open-api.yaml#L2006

laserninja: Done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -508,6 +522,95 @@ public Response planTableScan(
...
+   * @param ifNoneMatch the If-None-Match header value from the client
+   * @param etag the current ETag
+   * @return true if the ETag matches (table unchanged), false otherwise
+   */
+  private static boolean etagMatches(String ifNoneMatch, EntityTag etag) {
+    if (ifNoneMatch == null || ifNoneMatch.isEmpty()) {
+      return false;
+    }
+    // Handle wildcard
+    if ("*".equals(ifNoneMatch.trim())) {
```
roryqi: Why do we need this?

laserninja: done

roryqi:
@laserninja Thanks for your contribution! Welcome to join the Gravitino community! Enjoy your trip!

FANNG1:
I see two issues with the current approach:

1. The ETag returned by `CreateTableResponse` is not reusable for the default `loadTable` request.
The Iceberg REST spec says the client may send the ETag received from `CreateTableResponse` or `LoadTableResponse` in `If-None-Match`. In this PR, `createTable` returns an ETag derived from `metadataLocation`, while `loadTable` compares against an ETag derived from `metadataLocation + snapshots`. For the default `snapshots=all` path, those values differ, so a client that follows the spec and reuses the ETag from create will not get `304 Not Modified` on a subsequent unchanged `loadTable`.

2. We may be leaving performance on the table by always calling `loadTable` before checking the ETag.
If the backend catalog can expose the current `metadataLocation` cheaply, for example via something like `SupportsMetadataLocation`, we could compute/compare the ETag first and return `304` without loading full table metadata. That would make the optimization more effective for freshness-aware reads. This should stay as an optional fast path though: if the catalog does not support it, the server can fall back to the current `loadTable` flow.

roryqi:
> 2\. SupportsMetadataLocation

Agreed. @laserninja Could u raise a follow-up pull request to optimize this point?

laserninja:
Should I directly create a PR or create an issue first?

roryqi:
> Should I directly create a PR or create an issue first?

You can raise a pr directly and reuse the issue #9736. Because this is a follow-up pull request.
The title should be `[#9736][followup] xxxxx`. The followup means that  users should back port them together when required.

PR #10517, opened 2026-03-23
[#10541] refactor(flink-connector): split versioned framework with Flink 1.18 baseline

### LINKED ISSUE #10541: [Flink] Refactor connector framework and add Flink 1.18 baseline support
### Describe the proposal

Refactor the Gravitino Flink connector to use a versioned architecture baseline and deliver Flink 1.18 support first.

This issue covers:
- introducing `flink-common`
- introducing `flink-1.18`
- introducing `flink-runtime-1.18`
- moving shared connector logic into `flink-common`
- adding version-specific Flink 1.18 catalog and factory entry classes
- adding Flink 1.18 version-specific integration test entries
- updating runtime packaging and CI for the Flink 1.18 lane

This issue is the first implementation stage under parent issue #9710.

### Why is this needed

The current single-module Flink connector layout is not a good foundation for maintained multi-version support. A Flink 1.18 baseline is the first step to establish the new versioned architecture.

### Additional context

Parent issue: #9710

FANNG1:

This PR implements the first step of the Flink multi-version refactor and limits the scope to the framework split plus Flink 1.18 support.

The main changes are:
- split the original single Flink connector module into `flink-common`, `flink-1.18`, and `flink-runtime-1.18`
- move shared connector logic and shared test bases into `flink-common`
- add versioned Flink 1.18 catalog/factory entry classes and Flink 1.18 runtime packaging
- introduce typed catalog compatibility hooks for the versioned layout
- run Flink integration coverage from the Flink 1.18 module instead of the old shared module
- keep the scope of this PR limited to Flink 1.18 as the baseline lane

This PR intentionally does not include Flink 1.19 or 1.20 support. Those will follow in separate PRs.

The current single-module Flink connector layout is not a good foundation for maintained multi-version support. A Flink 1.18 baseline is the first step to establish the new versioned architecture.

This PR is the first implementation stage under parent issue #9710.

Fix: #10541

Yes.

The Flink connector build layout changes from a single Flink module to:
- `flink-common`
- `flink-1.18`
- `flink-runtime-1.18`

The user-facing SQL behavior is unchanged for Flink 1.18, but the connector packaging and internal implementation are now version-scoped to support later minor-version expansion.

The following validations were run:

- `./gradlew :common:jar :flink-connector:flink-1.18:test -PskipITs :flink-connector:flink-runtime-1.18:test`
- `./gradlew :flink-connector:flink-1.18:test -PskipTests -PtestMode=embedded -PskipDockerTests=false --tests "org.apache.gravitino.flink.connector.integration.test.catalog.GravitinoCatalogManagerIT118" --console=plain`


flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/hive/GravitinoHiveCatalogFactory.java
```diff
@@ -73,6 +73,24 @@ public Catalog createCatalog(Context context) {
         helper.getOptions().get(HiveCatalogFactoryOptions.HIVE_VERSION));
   }
 
+  protected Catalog newCatalog(
+      String catalogName,
+      String defaultDatabase,
+      java.util.Map<String, String> catalogOptions,
+      SchemaAndTablePropertiesConverter schemaAndTablePropertiesConverter,
+      PartitionConverter partitionConverter,
+      @javax.annotation.Nullable HiveConf hiveConf,
```
FANNG1: remove this ? @javax.annotation.Nullable

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/iceberg/GravitinoIcebergCatalogFactory.java
```diff
@@ -93,4 +117,16 @@ public SchemaAndTablePropertiesConverter schemaAndTablePropertiesConverter() {
   public PartitionConverter partitionConverter() {
     return DefaultPartitionConverter.INSTANCE;
   }
+
+  private Map<String, String> toIcebergCatalogOptions(Map<String, String> catalogOptions) {
```
FANNG1, line 117: why add this method?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
```diff
@@ -46,15 +47,25 @@ public org.apache.flink.table.catalog.Catalog createCatalog(Context context) {
...
         helper.getOptions().get(GravitinoJdbcCatalogFactoryOptions.DEFAULT_DATABASE);
     Preconditions.checkArgument(
         defaultDatabase != null,
         GravitinoJdbcCatalogFactoryOptions.DEFAULT_DATABASE.key() + " should not be null.");
-    return new GravitinoJdbcCatalog(
+    return newCatalog(
         context, defaultDatabase, schemaAndTablePropertiesConverter(), partitionConverter());
   }
 
+  protected org.apache.flink.table.catalog.Catalog newCatalog(
```
FANNG1: rename org.apache.flink.table.catalog.Catalog to Catalog

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
```diff
@@ -46,15 +47,25 @@ public org.apache.flink.table.catalog.Catalog createCatalog(Context context) {
     context.getOptions().remove(JdbcPropertiesConstants.FLINK_DRIVER);
     final FactoryUtil.CatalogFactoryHelper helper =
         FactoryUtils.createCatalogFactoryHelper(this, context);
+    helper.validate();
```
FANNG1: why add this method?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
```diff
@@ -67,11 +78,15 @@ public PartitionConverter partitionConverter() {
 
   @Override
   public Set<ConfigOption<?>> requiredOptions() {
-    return Collections.emptySet();
+    return ImmutableSet.of(
```
FANNG1: why changing this?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalogFactory.java
```diff
@@ -43,10 +43,19 @@ public Catalog createCatalog(Context context) {
         FactoryUtils.createCatalogFactoryHelper(this, context);
     String defaultDatabase =
         helper.getOptions().get(GravitinoPaimonCatalogFactoryOptions.DEFAULT_DATABASE);
-    return new GravitinoPaimonCatalog(
+    return newCatalog(
         context, defaultDatabase, schemaAndTablePropertiesConverter(), partitionConverter());
   }
 
+  protected Catalog newCatalog(
```
FANNG1, line 50: Should we use an abstract method for newCatalog?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/utils/CatalogCompat.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.Map;
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.catalog.CatalogTable;
+import org.apache.flink.table.catalog.ResolvedCatalogTable;
+
+/** Typed compatibility hook that is implemented per Flink minor version. */
+public interface CatalogCompat {
+
+  CatalogTable createCatalogTable(
```
FANNG1, line 40: add java doc for the method

flink-connector/flink-common/src/test/java/org/apache/gravitino/flink/connector/integration/test/hive/FlinkHiveCatalogIT.java
```diff
@@ -198,6 +199,19 @@ private void initHiveConfDir() {
     }
   }
 
+  protected String getSharedHiveConfDir() {
+    try {
+      return java.nio.file.Paths.get(
```
FANNG1: don't use long package name for this PR, rename java.nio.file.Paths to Paths.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
```diff
@@ -91,6 +91,10 @@ public class BaseIT {
 
   private static final Logger LOG = LoggerFactory.getLogger(BaseIT.class);
   private static final Splitter COMMA = Splitter.on(",").omitEmptyStrings().trimResults();
+  private static final String LOOPBACK_HOST = "127.0.0.1";
```
FANNG1: why change baseIT?

settings.gradle.kts
```diff
@@ -22,8 +22,9 @@ plugins {
 
 rootProject.name = "gravitino"
 
-val scalaVersion: String = gradle.startParameter.projectProperties["scalaVersion"]?.toString()
```
FANNG1, line 25: why changing this?

settings.gradle.kts
```diff
@@ -67,7 +68,11 @@ include("iceberg:iceberg-common")
 include("iceberg:iceberg-rest-server")
 include("lance:lance-common")
 include("lance:lance-rest-server")
-include("authorizations:authorization-ranger", "authorizations:authorization-common", "authorizations:authorization-chain")
+include(
```
FANNG1: don't change this

settings.gradle.kts
```diff
@@ -81,17 +86,28 @@ include(
...
+  include("flink-connector:flink-1.18", "flink-connector:flink-runtime-1.18")
+  project(":flink-connector:flink-1.18").projectDir = file("flink-connector/v1.18/flink")
+  project(":flink-connector:flink-runtime-1.18").projectDir =
+    file("flink-connector/v1.18/flink-runtime")
 }
 include("spark-connector:spark-3.3", "spark-connector:spark-runtime-3.3")
 project(":spark-connector:spark-3.3").projectDir = file("spark-connector/v3.3/spark")
-project(":spark-connector:spark-runtime-3.3").projectDir = file("spark-connector/v3.3/spark-runtime")
-include("spark-connector:spark-3.4", "spark-connector:spark-runtime-3.4", "spark-connector:spark-3.5", "spark-connector:spark-runtime-3.5")
+project(":spark-connector:spark-runtime-3.3").projectDir =
```
FANNG1: don't change this

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/iceberg/GravitinoIcebergCatalogFactory.java
FANNG1: I kept this method. In the versioned split, the common Iceberg factory still needs one place to translate Gravitino catalog properties into the Iceberg-specific option set before delegating to the versioned catalog implementation. This helper keeps that mapping local and avoids duplicating the conversion logic in each version module.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/hive/GravitinoHiveCatalogFactory.java
FANNG1: Fixed in b68f1e918. I replaced the fully qualified `@javax.annotation.Nullable` with the normal import to match the repository style.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
FANNG1: I kept this validation step. The JDBC factory needs to validate the effective option set after the versioned refactor, especially because the Flink-facing options and the backend JDBC catalog options are not identical. Keeping the check here fails invalid configurations early instead of letting them surface later during catalog creation.

flink-connector/v1.18/flink/src/main/java/org/apache/gravitino/flink/connector/iceberg/GravitinoIcebergCatalogFlink118.java
```diff
@@ -0,0 +1,60 @@
...
+
+  protected GravitinoIcebergCatalogFlink118(
+      String catalogName,
+      String defaultDatabase,
+      SchemaAndTablePropertiesConverter schemaAndTablePropertiesConverter,
+      PartitionConverter partitionConverter,
+      Map<String, String> catalogOptions,
+      Map<String, String> icebergCatalogProperties,
+      CatalogFactory.Context context) {
+    super(
```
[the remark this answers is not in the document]
FANNG1: Fixed in b68f1e918. The 1.18 catalog implementation does not use `CatalogFactory.Context`, so I removed the dead constructor parameter from `GravitinoIcebergCatalogFlink118` and kept the versioned factory path aligned with the actual usage.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
FANNG1: Fixed in b68f1e918. I switched this back to the imported `Catalog` type.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
FANNG1: I kept this change. The required option set still needs to be enforced in the common JDBC factory after the versioned split; otherwise the factory can accept incomplete definitions and fail later in a less obvious way. This is a behavior-preserving validation change rather than a formatting change.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalogFactory.java
FANNG1: I left this as-is for now. The current protected template method keeps the version extension point while still allowing a shared default path in the common factory. Making it abstract would force every version module to re-declare the method even when the common behavior is still valid.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/utils/CatalogCompat.java
FANNG1: Fixed in b68f1e918. I added method-level javadocs here to make the compatibility contract clearer across versioned Flink modules.

settings.gradle.kts
FANNG1: I reverted the formatting-only part in b68f1e918. There was no need to keep this stylistic change in the PR.

settings.gradle.kts
FANNG1: I kept this change. This include wiring is the actual semantic part of the PR because the versioned Flink modules need to be added to the build. I only reverted the unrelated formatting churn around it.

flink-connector/flink-common/src/test/java/org/apache/gravitino/flink/connector/integration/test/hive/FlinkHiveCatalogIT.java
FANNG1: Fixed in b68f1e918. I changed this to use the imported `Paths` type.

settings.gradle.kts
FANNG1: I reverted this formatting-only change in b68f1e918 to keep the PR diff focused.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/util/BaseIT.java
FANNG1: Follow-up: I removed the BaseIT-related changes from this PR in 978cac0a5. After revisiting the scope, I agree this helper/host-normalization change is not required for the Flink versioned split itself, so I dropped it from the PR instead of keeping a revised variant here.

lasdf1234:
cc @FANNG1  I'm here.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalog.java
```diff
@@ -570,7 +572,21 @@ protected CatalogBaseTable toFlinkTable(Table table, ObjectPath tablePath) {
...
+      String comment,
+      List<String> partitionKeys,
+      Map<String, String> options) {
+    return catalogCompat().createCatalogTable(schema, comment, partitionKeys, options);
+  }
+
+  protected CatalogCompat catalogCompat() {
+    // Versioned catalog entry classes override this hook when the Flink minor has a different
+    // catalog/table API path.
+    return DefaultCatalogCompat.INSTANCE;
```
lasdf1234, line 594: Have any other oss project compatible with multiple versions，Is it implemented in the same way?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/hive/GravitinoHiveCatalog.java
```diff
@@ -252,4 +251,13 @@ private void applyGenericTableAlter(
...
   }
+
+  protected Map<String, String> toGravitinoGenericTableProperties(
+      ResolvedCatalogTable resolvedTable) {
+    return FlinkGenericTableUtil.toGravitinoGenericTableProperties(resolvedTable, catalogCompat());
+  }
+
+  protected CatalogTable toFlinkGenericTable(Table table) {
+    return FlinkGenericTableUtil.toFlinkGenericTable(table, catalogCompat());
+  }
```
lasdf1234, line 262: Can we move these methods back to the Util class?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactoryOptions.java
```diff
@@ -30,6 +30,20 @@ public class GravitinoJdbcCatalogFactoryOptions {
...
+
   public static final ConfigOption<String> DEFAULT_DATABASE =
-      ConfigOptions.key("default-database").stringType().noDefaultValue();
+      ConfigOptions.key(JdbcPropertiesConstants.FLINK_JDBC_DEFAULT_DATABASE)
+          .stringType()
+          .noDefaultValue();
+
+  public static final ConfigOption<String> DRIVER =
+      ConfigOptions.key(JdbcPropertiesConstants.FLINK_DRIVER).stringType().noDefaultValue();
 }
```
lasdf1234, line 37: I used old version and configured this properties,Flink can run normally.Is it necessary to put these attributes here?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/hive/GravitinoHiveCatalog.java
FANNG1: I prefer to keep these thin wrapper methods on the catalog class. They delegate to `FlinkGenericTableUtil`, but they also bind the call to the catalog instance so the version-specific `catalogCompat()` hook can be applied naturally by subclasses such as the Flink 1.18 catalog. Moving them back to the util class would make the compat path more scattered rather than simpler.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalog.java
FANNG1: I do not think this needs a code change in this PR. The current structure uses a shared common layer plus version-specific adapters/hooks (`catalogCompat()` and the versioned catalog/factory classes) to isolate Flink minor-version API differences. That is the main reason for this split: keep the common behavior in one place and localize the Flink-version-specific parts in each version module.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactoryOptions.java
FANNG1: I kept these options here intentionally. This does not introduce a new user-facing requirement compared with the old implementation; it centralizes the Flink-side JDBC option definitions so `requiredOptions()`, `optionalOptions()`, and option lookups use the same source of truth after the versioned refactor. That keeps the factory logic clearer and avoids scattering the option keys across the implementation.

FANNG1:
@lasdf1234 any other comments?

lasdf1234:
## Code Review

This PR establishes a clean versioned architecture for the Flink connector and is a well-structured first step toward multi-version support. The three-layer module split (`flink-common` / `flink-1.18` / `flink-runtime-1.18`) and the `CatalogCompat` interface design are solid choices.

I found several issues that should be addressed before merge:

---

### 🔴 Issue 1 — Dead parameter in `GravitinoIcebergCatalog` first constructor

**File:** `GravitinoIcebergCatalog.java` line 43

```java
protected GravitinoIcebergCatalog(
    ...,
    Map<String, String> icebergCatalogProperties,
    CatalogFactory.Context context) {   // ← never stored or read
```

`context` is accepted but never used inside the constructor — the catalog is built solely from `catalogName` and `icebergCatalogProperties`. This dead parameter misleads callers into thinking the context affects construction.

Either use it (e.g., pass a class loader) or remove it. The versioned path (`GravitinoIcebergCatalogFlink118`) already bypasses this constructor by using the second one that accepts a pre-built `AbstractCatalog`, so removing `context` here would not break the versioned path.

---

### 🔴 Issue 2 — Unsafe unchecked cast in `GravitinoIcebergCatalog` and `GravitinoIcebergCatalogFlink118`

**File:** `GravitinoIcebergCatalog.java` lines 51–52; `GravitinoIcebergCatalogFlink118.java`

```java
this.icebergCatalog =
    (AbstractCatalog) new FlinkCatalogFactory().createCatalog(catalogName, icebergCatalogProperties);
```

`FlinkCatalogFactory.createCatalog()` returns `Catalog`. Casting it directly to `AbstractCatalog` without any guard will throw a silent `ClassCastException` if a future Iceberg version changes what the factory returns.

Suggested fix:

```java
Catalog created = new FlinkCatalogFactory().createCatalog(catalogName, icebergCatalogProperties);
Preconditions.checkState(
    created instanceof AbstractCatalog,
    "Expected AbstractCatalog from FlinkCatalogFactory but got %s",
    created.getClass().getName());
this.icebergCatalog = (AbstractCatalog) created;
```

The same issue exists in `GravitinoIcebergCatalogFlink118.createIcebergCatalog()`.

---

### 🟡 Issue 3 — `CatalogCompatFlink118` duplicates `DefaultCatalogCompat` entirely

**File:** `CatalogCompatFlink118.java`

Both classes implement the exact same behavior: `CatalogTable.of(...)` and `CatalogPropertiesUtil.serializeCatalogTable(...)`. Since Flink 1.18 **is** the baseline version, there is no functional difference between them. This creates two classes to maintain for the same logic.

Consider delegating to `DefaultCatalogCompat.INSTANCE` instead:

```java
@Override
public CatalogTable createCatalogTable(
    Schema schema, String comment, List<String> partitionKeys, Map<String, String> options) {
  return DefaultCatalogCompat.INSTANCE.createCatalogTable(schema, comment, partitionKeys, options);
}

@Override
public Map<String, String> serializeCatalogTable(ResolvedCatalogTable resolvedTable) {
  return DefaultCatalogCompat.INSTANCE.serializeCatalogTable(resolvedTable);
}
```

`CatalogCompatFlink118` then becomes meaningful only when Flink 1.18 genuinely diverges from the default baseline.

---

### 🟡 Issue 4 — Redundant `CatalogTableBuilder` functional interface in `FlinkGenericTableUtil`

**File:** `FlinkGenericTableUtil.java` lines 108–115

`CatalogTableBuilder.create(Schema, String, List<String>, Map<String, String>)` has the **exact same signature** as `CatalogCompat.createCatalogTable(...)`. This introduces two parallel extension points for the same operation.

The only consumer of `CatalogTableBuilder` is `toFlinkGenericTable(Table, CatalogTableBuilder)`, which is called exclusively via `catalogCompat::createCatalogTable` — making `CatalogTableBuilder` pure indirection with no added value.

Suggested simplification: remove `CatalogTableBuilder` entirely and inline the logic directly in the `CatalogCompat` overload. The three-overload chain

```
toFlinkGenericTable(Table)
  → toFlinkGenericTable(Table, CatalogCompat)
  → toFlinkGenericTable(Table, CatalogTableBuilder)
```

collapses to two clean overloads.

---

### 🟡 Issue 5 — `toIcebergCatalogOptions` silently overwrites user-provided `catalog-type`

**File:** `GravitinoIcebergCatalogFactory.java` line 129

```java
icebergCatalogOptions.put(CommonCatalogOptions.CATALOG_TYPE.key(), "iceberg");
```

This unconditionally replaces any `catalog-type` value already present in the map (copied from `catalogOptions` on line 122). If a user explicitly set a different value, it is silently discarded with no log or warning.

Consider at minimum a DEBUG-level log:

```java
String existing = icebergCatalogOptions.put(CommonCatalogOptions.CATALOG_TYPE.key(), "iceberg");
if (existing != null && !"iceberg".equals(existing)) {
    LOG.debug("Overriding catalog-type from '{}' to 'iceberg'", existing);
}
```

---

### 🟢 Issue 6 — `discoverFactories(Iterator, Predicate, String)` should be `protected` and needs JavaDoc

**File:** `GravitinoCatalogStore.java` line 161

This method is package-private (no modifier), which is inconsistent with other new extension hooks introduced in this PR (e.g., `newCatalogDescriptor` is `protected`). If the intent is to allow versioned store subclasses to override factory discovery, it should be `protected`.

Both `newCatalogDescriptor(String, Configuration)` (line 127) and `discoverFactories(Iterator, Predicate, String)` are new extension points with no JavaDoc. Protected APIs require JavaDoc per the project contribution guidelines.

---

Overall the architecture is sound. Please address issues 1–4 before merge.

lasdf1234:
My PR to enable session catalog has merged into main branch,have you sync code? May be there is a lot of conflict.

FANNG1:
Thanks for the detailed review. I re-checked these points against the multi-version split spec in #10442 and handled them as follows:

- Issue 1: accepted. The unused Iceberg `CatalogFactory.Context` parameter does not belong in the phase-1 mechanical split, so I removed it from the common and 1.18 Iceberg construction path.
- Issue 2: accepted. I replaced the direct `AbstractCatalog` casts in the Iceberg common and 1.18 paths with an explicit type check and added a focused unit test.
- Issue 3: accepted as a maintainability improvement. I kept `CatalogCompatFlink118` as the explicit 1.18 hook, but made it delegate to `DefaultCatalogCompat` instead of duplicating the same logic.
- Issue 4: accepted. I removed the extra `CatalogTableBuilder` indirection from `FlinkGenericTableUtil` and routed the path directly through `CatalogCompat`.
- Issue 5: I do not think this is a bug. In this path the outer Flink factory identifier is `gravitino-iceberg`, but the nested native Iceberg factory still requires `catalog-type=iceberg`. That overwrite is the intended bridge between the two layers. I added a code comment to make that explicit instead of changing the behavior.
- Issue 6: partially accepted. I added JavaDoc for the new `CatalogStore` hooks/helpers. I did not widen `discoverFactories(...)` to `protected`, because it is still an internal helper rather than an intended versioned extension point.

FANNG1:
> My PR to enable session catalog has merged into main branch,have you sync code? May be there is a lot of conflict.

Yes, this PR is synced with the latest main

FANNG1:
@lasdf1234 any other comments?

FANNG1:
@roryqi @diqiu50 @yuqi1129  do you have time to review this PR?

lasdf1234:
I don't have any additional suggestions @roryqi @diqiu50 @yuqi1129  could you review this PR?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
```diff
@@ -51,13 +51,22 @@ public org.apache.flink.table.catalog.Catalog createCatalog(Context context) {
...
+      PartitionConverter partitionConverter) {
+    return new GravitinoJdbcCatalog(
+        context, defaultDatabase, schemaAndTablePropertiesConverter, partitionConverter);
+  }
+
   @Override
-  public Catalog.Type gravitinoCatalogType() {
-    return Catalog.Type.RELATIONAL;
+  public org.apache.gravitino.Catalog.Type gravitinoCatalogType() {
+    return org.apache.gravitino.Catalog.Type.RELATIONAL;
```
yuqi1129, line 69: org.apache.gravitino.Catalog

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/iceberg/GravitinoIcebergCatalog.java
```diff
@@ -0,0 +1,88 @@
...
+  public Optional<Factory> getFactory() {
+    return icebergCatalog.getFactory();
+  }
+
+  @Override
+  protected AbstractCatalog realCatalog() {
+    return icebergCatalog;
+  }
+
+  protected static AbstractCatalog requireAbstractCatalog(Catalog catalog) {
```
yuqi1129: Based on the method name, the return value should be a Boolean value. Can you try to polish the name?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStore.java
```diff
@@ -160,8 +185,13 @@ private BaseCatalogFactory discoverFactories(Predicate<Factory> predicate, Strin
         if (catalogFactory instanceof BaseCatalogFactory && predicate.test(catalogFactory)) {
           factories.add(catalogFactory);
         }
+      } catch (ServiceConfigurationError e) {
+        LOG.debug(
+            "Skip a {} entry that cannot be initialized.", Factory.class.getCanonicalName(), e);
       } catch (NoClassDefFoundError e) {
         LOG.debug("NoClassDefFoundError when loading a {}.", Factory.class.getCanonicalName(), e);
+      } catch (LinkageError e) {
+        LOG.debug("LinkageError when loading a {}.", Factory.class.getCanonicalName(), e);
```
yuqi1129, line 198: So we can just ignore exceptions with these three types?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/utils/CatalogCompat.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.catalog.CatalogTable;
+import org.apache.flink.table.catalog.ResolvedCatalogTable;
+
+/** Typed compatibility hook that is implemented per Flink minor version. */
+public interface CatalogCompat {
+
+  /**
+   * Create a {@link CatalogTable} using the Flink minor version's constructor and behavior.
```
yuqi1129: I'm not very clear about `minor version's constructor and behavior`, why only minor version?

flink-connector/v1.18/flink-runtime/build.gradle.kts
```diff
@@ -29,14 +29,8 @@ repositories {
...
 
-val flinkVersion: String = libs.versions.flink.get()
+val flinkVersion: String = libs.versions.flink18.get()
 val flinkMajorVersion: String = flinkVersion.substringBeforeLast(".")
-
-// The Flink only support scala 2.12, and all scala api will be removed in a future version.
-// You can find more detail at the following issues:
-// https://issues.apache.org/jira/browse/FLINK-23986,
-// https://issues.apache.org/jira/browse/FLINK-20845,
-// https://issues.apache.org/jira/browse/FLINK-13414.
```
yuqi1129, line 39: You may also remove them in `flink-common/build.gradle.kts`.

flink-connector/v1.18/flink/build.gradle.kts
```diff
@@ -0,0 +1,199 @@
...
+  testImplementation("org.apache.flink:flink-table-planner_$scalaVersion:$flinkVersion")
+  testImplementation("org.apache.flink:flink-test-utils:$flinkVersion")
+
+  testRuntimeOnly(libs.junit.jupiter.engine)
+}
+
+tasks.test {
+  dependsOn(commonProject.tasks.named("testClasses"))
+  testClassesDirs = files(commonTestOutput.classesDirs, sourceSets["test"].output.classesDirs)
+  classpath = files(commonTestOutput, sourceSets["test"].runtimeClasspath)
```
yuqi1129, line 172: Why don't we just introduce the dependence of `flink-common` test jars.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactory.java
FANNG1: I tried this change, but importing the nested `Catalog.Type` triggers Error Prone `BadImport` with `-Werror`; the compiler explicitly recommends keeping `org.apache.gravitino.Catalog.Type` qualified here. Since this file also imports Flink `Catalog`, I kept the qualified Gravitino type to avoid ambiguity and keep the build green.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/utils/CatalogCompat.java
FANNG1: Agreed. `minor version` was too narrow, especially because future Flink 2.x support may also use this hook. I changed the wording to `active Flink version` / `Flink version`.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/iceberg/GravitinoIcebergCatalog.java
FANNG1: Good point. I renamed it from `requireAbstractCatalog` to `asAbstractCatalog`, which better matches that the helper returns the converted `AbstractCatalog` after validating the type.

flink-connector/v1.18/flink/build.gradle.kts
FANNG1: The versioned module already depends on the `flink-common` test artifact. However, that only puts the shared test classes on the test classpath; Gradle will not discover and execute them unless their output is also included in `testClassesDirs`. I added a short comment in the build file to make this explicit.

flink-connector/v1.18/flink-runtime/build.gradle.kts
FANNG1: I kept the runtime guard here because this module produces the final shaded runtime jar, where the Log4j exclusion must be enforced strictly. `flink-common` is not the runtime artifact, and it already excludes Log4j from the dependencies it declares. I prefer not to add the same global guard there unless we see a concrete leakage risk.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStore.java
FANNG1: They are ignored only during `ServiceLoader` scanning for optional Flink catalog factories. In the versioned connector layout, some provider factories can be present in service files but unavailable on a specific Flink/provider classpath. Skipping `ServiceConfigurationError`, `NoClassDefFoundError`, and `LinkageError` lets discovery continue to other valid factories; unexpected exceptions are still propagated. I added comments to make this boundary explicit.

settings.gradle.kts
```diff
@@ -82,8 +82,11 @@ include(
 include("spark-connector:spark-common")
 if (scalaVersion == "2.12") {
   // flink only support scala 2.12
-  include("flink-connector:flink")
-  include("flink-connector:flink-runtime")
+  include("flink-connector:flink-common")
+  include("flink-connector:flink-1.18", "flink-connector:flink-runtime-1.18")
+  project(":flink-connector:flink-1.18").projectDir = file("flink-connector/v1.18/flink")
+  project(":flink-connector:flink-runtime-1.18").projectDir =
+    file("flink-connector/v1.18/flink-runtime")
```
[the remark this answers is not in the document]
jerryshao: @FANNG1 can you check this issue?

settings.gradle.kts
```diff
@@ -82,8 +82,11 @@ include(
 include("spark-connector:spark-common")
 if (scalaVersion == "2.12") {
   // flink only support scala 2.12
-  include("flink-connector:flink")
-  include("flink-connector:flink-runtime")
+  include("flink-connector:flink-common")
+  include("flink-connector:flink-1.18", "flink-connector:flink-runtime-1.18")
```
yuqi1129, line 92: This split also needs to update the CI exclusions that still reference the removed `:flink-connector:flink` path. For example `.github/workflows/backend-integration-test-action.yml`, `trino-integration-test.yml`, and `trino-multi-version-test.yml` still pass `-x :flink-connector:flink:*`; Gradle accepts those stale excludes, so those jobs will now build/test the new Flink modules instead of skipping them as before. Please switch those excludes to the new `:flink-connector:flink-common`, `:flink-connector:flink-1.18`, and `:flink-connector:flink-runtime-1.18` tasks where appropriate.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/jdbc/GravitinoJdbcCatalogFactoryOptions.java
FANNG1: Follow-up: I changed this after the later review round. The current code no longer keeps the extra JDBC Flink-side options here; `GravitinoJdbcCatalogFactoryOptions` is back to only `DEFAULT_DATABASE`, and the common JDBC factory no longer widens the required/optional option contract. That keeps this PR aligned with the spec goal of staying behavior-preserving during the initial versioned split.

settings.gradle.kts
FANNG1: Agreed. This was a real miss in the workflow exclusions after the module split. I updated the remaining CI excludes to use the new versioned Flink module paths:
- `:flink-connector:flink-common`
- `:flink-connector:flink-1.18`
- `:flink-connector:flink-runtime-1.18`

Specifically, I fixed `.github/workflows/backend-integration-test-action.yml`, `.github/workflows/trino-integration-test.yml`, and `.github/workflows/trino-multi-version-test.yml`.

settings.gradle.kts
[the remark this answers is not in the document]
FANNG1: Checked and fixed. The issue was that a few CI workflows still excluded the removed `:flink-connector:flink` path, so they no longer actually skipped the versioned Flink modules after the split. I updated the remaining exclusions to the new module paths in `.github/workflows/backend-integration-test-action.yml`, `.github/workflows/trino-integration-test.yml`, and `.github/workflows/trino-multi-version-test.yml`.

jerryshao (dismissed):

jerryshao (approved):

PR #10523, opened 2026-03-24
[#10535] fix(server,web-v2): Expose serviceAdmins in /configs and handle array format in UI

bharos:
- Add Configs.SERVICE_ADMINS to ConfigServlet basicConfigEntries so gravitino.authorization.serviceAdmins is returned by GET /configs
- Update TestConfigServlet to assert the new field is present
- Fix web-v2 UI to handle serviceAdmins as JSON array (List<String>) rather than a comma-separated string

It is accessed in front-end here : https://github.com/apache/gravitino/blob/main/web-v2/web/src/lib/store/auth/index.js#L48

ConfigServlet only exposed AUTHENTICATORS and ENABLE_AUTHORIZATION via the /configs endpoint. Without serviceAdmins, the web UI cannot determine if the logged-in user is a service admin, causing the 'Create Metalake' button to be hidden even for valid service admins.

Fix: #10535 

Yes, it will correctly make the "CREATE METALAKE" button appear

Testing on Web UI
Screenshot 2026-03-23 at 9 57 07 PM

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/ConfigServlet.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/ConfigServlet.java
@@ -59,8 +59,8 @@ public class ConfigServlet extends HttpServlet {
     }
 
     if (serverConfig.get(Configs.ENABLE_AUTHORIZATION)) {
-      // Expose serviceAdmins when authorization is enabled so the web UI can determine whether the logged-in user has
-      // service-admin privileges (e.g. to show the "Create Metalake" button).
+      // Expose serviceAdmins when authorization is enabled so the web UI can determine whether the
+      // logged-in user has service-admin privileges (e.g. to show the "Create Metalake" button).
       String serviceAdminsRaw = serverConfig.getRawString(Configs.SERVICE_ADMINS.getKey());
       if (serviceAdminsRaw != null) {
         configs.put(Configs.SERVICE_ADMINS.getKey(), serverConfig.get(Configs.SERVICE_ADMINS));
--- a/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
+++ b/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
@@ -53,6 +53,24 @@ public class TestConfigServlet {
   }
 
   @Test
+  public void testConfigServletWithAuthEnabledButNoServiceAdmins() throws Exception {
+    // When authorization is enabled but serviceAdmins is not configured, the key should be
+    // absent from the response (no crash) rather than null or empty.
+    ServerConfig serverConfig = new ServerConfig();
+    serverConfig.set(Configs.ENABLE_AUTHORIZATION, true);
+    ConfigServlet configServlet = new ConfigServlet(serverConfig);
+    configServlet.init();
+    HttpServletResponse res = mock(HttpServletResponse.class);
+    PrintWriter writer = mock(PrintWriter.class);
+    when(res.getWriter()).thenReturn(writer);
+    configServlet.doGet(null, res);
+    verify(writer)
+        .write(
+            "{\"gravitino.authorization.enable\":true,\"gravitino.authenticators\":[\"simple\"]}");
+    configServlet.destroy();
+  }
+
+  @Test
   public void testConfigServletWithAuthEnabledAndServiceAdmins() throws Exception {
     ServerConfig serverConfig = new ServerConfig();
     serverConfig.set(Configs.ENABLE_AUTHORIZATION, true);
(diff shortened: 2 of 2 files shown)


bharos:
@LauraXia123 @roryqi can you check if this is correct ?
I was not able to see the "CREATE METALAKE" button on v2 UI even after logging in as service_admin, because it was never passed via ConfigServlet to /configs endpoint

I'll create an issue if you think this looks correct. Just want to make sure I'm not missing something otherwise

server/src/main/java/org/apache/gravitino/server/web/ConfigServlet.java
```diff
@@ -49,7 +49,7 @@ public class ConfigServlet extends HttpServlet {
           OAuthConfig.SCOPE);
 
   private static final ImmutableSet<ConfigEntry<?>> basicConfigEntries =
-      ImmutableSet.of(Configs.AUTHENTICATORS, Configs.ENABLE_AUTHORIZATION);
+      ImmutableSet.of(Configs.AUTHENTICATORS, Configs.ENABLE_AUTHORIZATION, Configs.SERVICE_ADMINS);
```
roryqi: If we don't set `serviceAdmin`, it will throw the exception because of the method `checkValue`.
```
  public static final ConfigEntry<List<String>> SERVICE_ADMINS =
      new ConfigBuilder("gravitino.authorization.serviceAdmins")
          .doc("The admins of Gravitino service")
          .version(ConfigConstants.VERSION_0_5_0)
          .stringConf()
          .toSequence()
          .checkValue(
              valueList ->
                  valueList != null && valueList.stream().allMatch(StringUtils::isNotBlank),
              ConfigConstants.NOT_BLANK_ERROR_MSG)
          .create();
```

LauraXia123:
> @LauraXia123 @roryqi can you check if this is correct ? I was not able to see the "CREATE METALAKE" button on v2 UI even after logging in as service_admin, because it was never passed via ConfigServlet to /configs endpoint
> 
> I'll create an issue if you think this looks correct. Just want to make sure I'm not missing something otherwise

Actually, we will get the service_admin via ConfigServlet to /configs endpoint by the setting `gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins`

server/src/main/java/org/apache/gravitino/server/web/ConfigServlet.java
bharos: That is fine. 
But the issue i was facing is because UI expects the value in here : https://github.com/apache/gravitino/blob/main/web-v2/web/src/lib/store/auth/index.js#L48

bharos:
> > @LauraXia123 @roryqi can you check if this is correct ? I was not able to see the "CREATE METALAKE" button on v2 UI even after logging in as service_admin, because it was never passed via ConfigServlet to /configs endpoint
> > I'll create an issue if you think this looks correct. Just want to make sure I'm not missing something otherwise
> 
> Actually, we will get the service_admin via ConfigServlet to /configs endpoint by the setting `gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins`

i see. But that's a pre-requisite now for Web UI v2, right.
Otherwise, it's an issue, like I faced here. I feel it should not be optional to pass service admin as web-ui-v2 expects it in  the front-end code

LauraXia123:
> > > @LauraXia123 @roryqi can you check if this is correct ? I was not able to see the "CREATE METALAKE" button on v2 UI even after logging in as service_admin, because it was never passed via ConfigServlet to /configs endpoint
> > > I'll create an issue if you think this looks correct. Just want to make sure I'm not missing something otherwise
> > 
> > 
> > Actually, we will get the service_admin via ConfigServlet to /configs endpoint by the setting `gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins`
> 
> i see. But that's a pre-requisite now for Web UI v2, right. Otherwise, it's an issue, like I faced here. I feel it should not be optional to pass service admin as web-ui-v2 expects it in the front-end code

Yes, system administrators currently obtain the relevant permissions (create metalake) through this configuration.

server/src/main/java/org/apache/gravitino/server/web/ConfigServlet.java
roryqi: We can put this value in the configuration when authorization is enabled.

danhuawang: We discussed earlier that this is an unsafe operation, didn’t we? @roryqi

roryqi: > We discussed earlier that this is an unsafe operation, didn’t we? @roryqi

If UI relies on it, we don't have the better solution.

bharos: @roryqi @LauraXia123 Updated the PR , to add this config only if authorization enabled.
PTAL, thanks

LauraXia123: I think it's make sense, LGTM, thx

roryqi:
@LauraXia123 Do u have other suggestion?

LauraXia123:
> @LauraXia123 Do u have other suggestion?

LGTM

PR #10536, opened 2026-03-24
[#9736][followup] feat(iceberg-rest): Skip full table load on ETag match via SupportsMetadataLocation

laserninja:

Optimize the ETag-based freshness check in `loadTable` by leveraging `SupportsMetadataLocation` to resolve the metadata file location cheaply without loading full table metadata. When the client sends `If-None-Match` and the catalog supports it, the server can return `304 Not Modified` without the cost of a full `loadTable` call.

Also fixes ETag consistency between `createTable`/`updateTable` and `loadTable`: ETags from create and update now include the default `snapshots` value (`"all"`), matching the default `loadTable` ETag. This allows clients to reuse ETags across endpoints as specified by the Iceberg REST spec.

1. **Performance**: The original implementation always performs a full `loadTable` before comparing ETags. For read-heavy workloads where clients already have fresh metadata (ETag matches), this full load is wasted. By using `SupportsMetadataLocation` (already implemented for JDBC and Hive catalogs), we can compare ETags via a lightweight metadata location query and skip the full load entirely.

2. **ETag consistency** (as reported by @FANNG1 in #10498): `createTable` returned an ETag derived from `metadataLocation` only, while `loadTable` derived its ETag from `metadataLocation + snapshots`. For the default `snapshots=all` path, these values differed, so a client that reuses the ETag from create would never get `304 Not Modified` on a subsequent unchanged `loadTable`.

Follow-up to: #10498

No user-facing API changes. The ETag values may differ from the previous implementation due to the consistency fix, but this is transparent to clients — they will simply get fresh ETags on the next request.

- Added `testCreateTableETagMatchesLoadTableETag`: Verifies that the ETag from `createTable` matches the default `loadTable` ETag, and that it produces `304` on a subsequent conditional load.
- Added `testUpdateTableETagMatchesLoadTableETag`: Same verification for `updateTable`.
- All existing ETag tests continue to pass (8 tests covering ETag presence, 304 matching, ETag changes after update, consistency, and snapshot-dependent ETags).

### DIFF
--- a/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
+++ b/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
@@ -204,6 +204,23 @@ public class IcebergCatalogWrapper implements AutoCloseable {
     return loadTableResponse;
   }
 
+  /**
+   * Retrieves the metadata file location for the specified table without loading full table
+   * metadata. This is an optional fast path for catalogs that implement {@link
+   * SupportsMetadataLocation}.
+   *
+   * @param tableIdentifier the table identifier
+   * @return an Optional containing the metadata file location, or empty if the catalog doesn't
+   *     support this operation
+   */
+  public Optional<String> getTableMetadataLocation(TableIdentifier tableIdentifier) {
+    if (catalog instanceof SupportsMetadataLocation) {
+      return Optional.ofNullable(
+          ((SupportsMetadataLocation) catalog).metadataLocation(tableIdentifier));
+    }
+    return Optional.empty();
+  }
+
   public boolean tableExists(TableIdentifier tableIdentifier) {
     return catalog.tableExists(tableIdentifier);
   }
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
@@ -28,6 +28,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.Consumes;
@@ -91,6 +92,8 @@ public class IcebergTableOperations {
   public static final String X_ICEBERG_ACCESS_DELEGATION = "X-Iceberg-Access-Delegation";
 
   @VisibleForTesting public static final String IF_NONE_MATCH = "If-None-Match";
+
+  @VisibleForTesting static final String DEFAULT_SNAPSHOTS = "all";
 
   private IcebergMetricsManager icebergMetricsManager;
 
@@ -288,7 +291,7 @@ public class IcebergTableOperations {
           String namespace,
       @IcebergAuthorizationMetadata(type = RequestType.LOAD_TABLE) @Encoded() @PathParam("table")
           String table,
-      @DefaultValue("all") @QueryParam("snapshots") String snapshots,
+      @DefaultValue(DEFAULT_SNAPSHOTS) @QueryParam("snapshots") String snapshots,
       @HeaderParam(X_ICEBERG_ACCESS_DELEGATION) String accessDelegation,
       @HeaderParam(IF_NONE_MATCH) String ifNoneMatch) {
     String catalogName = IcebergRESTUtils.getCatalogName(prefix);
@@ -311,6 +314,19 @@ public class IcebergTableOperations {
             TableIdentifier tableIdentifier = TableIdentifier.of(icebergNS, tableName);
             IcebergRequestContext context =
                 new IcebergRequestContext(httpServletRequest(), catalogName, isCredentialVending);
+
+            // Fast path: if client sent If-None-Match, try to resolve ETag without full table load
+            if (StringUtils.isNotBlank(ifNoneMatch)) {
+              Optional<String> metadataLocation =
+                  tableOperationDispatcher.getTableMetadataLocation(context, tableIdentifier);
+              if (metadataLocation.isPresent()) {
+                EntityTag etag = generateETag(metadataLocation.get(), snapshots);
+                if (etag != null && etagMatches(ifNoneMatch, etag)) {
+                  return Response.notModified(etag).build();
+                }
+              }
+            }
+
             LoadTableResponse loadTableResponse =
                 tableOperationDispatcher.loadTable(context, tableIdentifier);
             EntityTag etag =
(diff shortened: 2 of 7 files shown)


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -311,13 +313,23 @@ public Response loadTable(
             TableIdentifier tableIdentifier = TableIdentifier.of(icebergNS, tableName);
             IcebergRequestContext context =
                 new IcebergRequestContext(httpServletRequest(), catalogName, isCredentialVending);
+
+            // Fast path: if client sent If-None-Match, try to resolve ETag without full table load
+            if (ifNoneMatch != null && !ifNoneMatch.isEmpty()) {
+              String metadataLocation =
+                  tableOperationDispatcher.getTableMetadataLocation(context, tableIdentifier);
```
roryqi, line 321: If the catalog doesn't support `getTableMetadataLocation`, we should get metalake location using the method `loadTable`.

laserninja: Thanks! Fixed it. when the catalog doesn't support 'getTableMetadataLocation', we now fall back to computing the ETag from the 'loadTable' response and still return 304 Not Modified when it matches If-None-Match.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -204,6 +205,22 @@ public LoadTableResponse loadTable(TableIdentifier tableIdentifier) {
...
+  /**
+   * Retrieves the metadata file location for the specified table without loading full table
+   * metadata. This is an optional fast path for catalogs that implement {@link
+   * SupportsMetadataLocation}.
+   *
+   * @param tableIdentifier the table identifier
+   * @return the metadata file location, or null if the catalog doesn't support this operation
+   */
+  @Nullable
+  public String getTableMetadataLocation(TableIdentifier tableIdentifier) {
```
roryqi: You can use Optional. It will help users to consider null value explictly.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -288,7 +290,7 @@ public Response loadTable(
           String namespace,
       @IcebergAuthorizationMetadata(type = RequestType.LOAD_TABLE) @Encoded() @PathParam("table")
           String table,
-      @DefaultValue("all") @QueryParam("snapshots") String snapshots,
+      @DefaultValue(DEFAULT_SNAPSHOTS) @QueryParam("snapshots") String snapshots,
```
roryqi, line 294: Another point:
We can do it in the next pull request.
If the `snapshots` is `all`, we should return all the snapshots.
If the `snapshots` is `refs`, we should return all the snapshots is used in the references.
We should get the refs by ` loadTableResponse.tableMetadata().refs()` and only return the snapshots which existed in the `refs()`.

FANNG1:
Maybe another option is to push the ETag / conditional-load logic down into the `loadTable` path itself, instead of exposing a separate `getTableMetadataLocation(...)` interface.

For example, we could extend `loadTable` to support a conditional request and return either `304` or `LoadTableResponse`. That way:
- we do not need to expose a storage/detail-oriented interface just for HTTP caching
- the fast path can still reuse the existing `loadTable` event/hook chain
- the ETag logic stays encapsulated in one place

This feels a bit cleaner to me than adding a separate metadata-location lookup API.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
laserninja: sounds good, will keep it for next one.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
laserninja: Done

laserninja:
@FANNG1 Thanks for the suggestion! I considered this approach but went with a separate `getTableMetadataLocation` for a few reasons:
1. **Separation of concerns**: ETag/conditional-load is an HTTP caching concern. Keeping it in the REST layer (`IcebergTableOperations`) avoids leaking HTTP semantics into the catalog abstraction.
2. **Minimal contract change**: `loadTable` has a well-defined return type (`LoadTableResponse`). Adding conditional-load support would require changing the return type to a wrapper/union type, which touches many more callers.
3. **`SupportsMetadataLocation` already exists**: We're exposing an existing catalog capability, not inventing a new abstraction. The interface is already implemented by JDBC/Hive/Memory catalogs.
4. **The fast path is optional**: `getTableMetadataLocation` returns `Optional.empty()` for unsupported catalogs, and the full `loadTable` fallback always works.
That said, if you feel strongly about encapsulating it within `loadTable`, I'm happy to refactor. Would appreciate your thoughts on how you'd see the return type working (e.g., `Optional<LoadTableResponse>`, or a new `ConditionalLoadResult` wrapper)?

roryqi:
LGTM, I am ok to add the new method `getMetadataLocation`. Let us wait for @FANNG1 suggestion.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -311,6 +314,19 @@ public Response loadTable(
             TableIdentifier tableIdentifier = TableIdentifier.of(icebergNS, tableName);
             IcebergRequestContext context =
                 new IcebergRequestContext(httpServletRequest(), catalogName, isCredentialVending);
+
+            // Fast path: if client sent If-None-Match, try to resolve ETag without full table load
+            if (ifNoneMatch != null && !ifNoneMatch.isEmpty()) {
```
roryqi: Could u use StringUtils.isNotBlank instead?

FANNG1:
Makes sense. I agree that HTTP-layer behavior should not be pushed all the way down to the lower layer.

My main concern with the current approach is not `getTableMetadataLocation(...)` itself, but that the fast path returns before going through the normal `loadTable` chain, which means it may bypass existing event/audit hooks.

If we want to encapsulate this better, I think we probably need a better `loadTable` result model first, for example a `ConditionalLoadResult` that can represent either `not modified` or `loaded`. That would let us keep the conditional-load logic on the `loadTable` path while still preserving the existing event/hook semantics.

@jerryshao @roryqi WDYT?

roryqi:
> Makes sense. I agree that HTTP-layer behavior should not be pushed all the way down to the lower layer.
> 
> My main concern with the current approach is not `getTableMetadataLocation(...)` itself, but that the fast path returns before going through the normal `loadTable` chain, which means it may bypass existing event/audit hooks.
> 
> If we want to encapsulate this better, I think we probably need a better `loadTable` result model first, for example a `ConditionalLoadResult` that can represent either `not modified` or `loaded`. That would let us keep the conditional-load logic on the `loadTable` path while still preserving the existing event/hook semantics.
> 
> @jerryshao @roryqi WDYT?

It's ok that we don't call the event or hook, because we don't call the method `loadTable`. 
If requests failed because of request parameters, we won't trigger the event and hook.
I feel that `Etag` is also a kind of check mechism to avoid calling underlying api.

laserninja:
Done. Will wait for @jerryshao's input

jerryshao:
> Makes sense. I agree that HTTP-layer behavior should not be pushed all the way down to the lower layer.
> 
> My main concern with the current approach is not `getTableMetadataLocation(...)` itself, but that the fast path returns before going through the normal `loadTable` chain, which means it may bypass existing event/audit hooks.
> 
> If we want to encapsulate this better, I think we probably need a better `loadTable` result model first, for example a `ConditionalLoadResult` that can represent either `not modified` or `loaded`. That would let us keep the conditional-load logic on the `loadTable` path while still preserving the existing event/hook semantics.
> 
> @jerryshao @roryqi WDYT?

This is a read-only shortcut for http layer cache solution. It will not affect the actual metadata operation, so I think it is fine to not involve the event/hook mechanism. We can always modify the code if we have further requirements.

roryqi (approved):

roryqi:
@laserninja There are many operations using `LoadTableResponse`, such as, `register table`. Could u modify them together?

Issue #10555, opened 2026-03-26
[Improvement] [Flink]  Support enable session catalogs using gravitino connector


lasdf1234:
### What would you like to be improved?

Currently, the flink-gravitino-connector enforces that all catalogs are managed by Gravitino. This restriction prevents users from using standard Flink connectors (e.g., `kafka`, `jdbc`, `filesystem`) alongside Gravitino-managed catalogs within the same Flink session.This feature supports the use of catalogs that are not managed by gravitino.

### How should we improve?

When creating a catalog not managed by gravitino, flink-gravitino-connector caches the catalog in memory instead of being persisted to the Gravitino server to avoid avoid report errors.

PR #10560, opened 2026-03-26
[#10555] improve(flink): Support enable session catalogs using gravitino connector

lasdf1234:
This PR introduces a mechanism to allow non-gravitino connectors (such as standard Flink connectors like `jdbc`, 
`hive`) to be used alongside Gravitino-managed catalogs in the Flink connector.
Details are as follows:
CREATE CATALOG: Gravitino-managed catalog are persisted to the Gravitino server; non-Gravitino catalog are stored in the in-memory store only.
GET / USE CATALOG: The in-memory store is checked first. If the catalog is not found there, it is retrieved from the Gravitino server.
DROP CATALOG: The in-memory store is checked first. If the catalog exists there it is removed from memory; otherwise it is removed from the Gravitino server.
SHOW / LIST CATALOGS: Returns the combined set of catalogs from both the in-memory store and the Gravitino server.
Session scope: Catalogs stored only in memory are session-scoped and will not survive when Flink restart.
Name conflict: If a catalog with the same name exists in both stores, the in-memory entry takes precedence.

Previously, the Gravitino Flink connector assumed all catalogs were Gravitino-managed. This prevented users from 
using standard Flink connectors (e.g., for temporary catalogs not yet supported or governed by Gravitino) 
within the same environment. This change provides flexibility to mix governed and ungoverned catalogs.

Fix: #10555 
Yes. User can configure `table.catalog-store.gravitino.gravitino.supportSessionCatalog` in 
`flink-conf.yaml` or via `TableEnvironment` to implement this function.
TestGravitinoSessionCatalogStore and TestGravitinoCatalogStoreFactory

### DIFF
--- a/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
+++ b/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
@@ -80,6 +80,7 @@ public class GravitinoSessionCatalogStore extends AbstractCatalogStore {
       memoryCatalogStore.storeCatalog(catalogName, descriptor);
     }
   }
+
   /**
    * Removes the specified catalog.
    *


FANNG1:
This PR needs a clearer user-facing semantic definition before we look at the implementation details.

By introducing an in-memory catalog store alongside the existing Gravitino-backed store, this change is no longer only "allow third-party connectors". It also changes the semantics of catalog lifecycle operations from the user perspective. I think the PR description and docs should first clarify:

- What is the exact behavior of CREATE CATALOG?
- Which catalog types are persisted in Gravitino, and which ones are stored only in Flink memory?
- What are the semantics of GET/USE CATALOG, DROP CATALOG, and SHOW/LIST CATALOGS after this change?
- Are third-party catalogs session-scoped or process-scoped? Will they disappear after restarting Flink?
- What is the expected behavior if the same catalog name exists in both the in-memory store and Gravitino?
- Is there a defined precedence when reading or dropping catalogs across the two stores?

Right now, the implementation seems to introduce a mixed two-store model, so users need a precise contract for catalog visibility, persistence, and conflict handling. Without that, it is hard to evaluate whether the current behavior is correct or whether edge cases like duplicate names and shadowing are acceptable.

lasdf1234:
@FANNG1 I have supplemented the PR.

FANNG1:
I think this feature should be redesigned around the user-facing concept of "session catalog", instead of mixing the new behavior directly into the existing `GravitinoCatalogStore`.

My suggestions:

1. The configuration should be modeled as `supportsSessionCatalog` (or an equivalent session-catalog-oriented option), instead of exposing the implementation detail `allow.third-party-connector.list`.
If type-level control is still needed, that can be an additional config, but the primary user-facing concept should be "support session catalog", not "route some third-party connectors to memory".

2. Do not mix the new logic into `GravitinoCatalogStore.java`.
If `supportsSessionCatalog` is enabled, use a new `GravitinoSessionCatalogStore`; otherwise keep using the original `GravitinoCatalogStore`.
This feature introduces a different catalog lifecycle model, so it should have a separate implementation boundary rather than adding more branching into the existing class.

3. The documentation should explicitly define the behavior of session catalogs.
Please document the semantics of create/get-use/drop/show-list, the lifecycle of session catalogs, restart behavior, and the precedence/conflict rule when the same catalog name exists in both places.

4. Please add integration tests to cover the full session catalog lifecycle.
Unit tests are not enough for this feature. We should have IT coverage for all major operations on session catalogs, including at least create, get/use, show/list, and drop.

FANNG1:
I think this feature should be redesigned around the user-facing concept of session catalog, instead of mixing the new behavior directly into the existing GravitinoCatalogStore.

My suggestions:

1. The configuration should be modeled as supportsSessionCatalog, or an equivalent session-catalog-oriented option, instead of exposing the implementation detail allow.third-party-connector.list.
If type-level control is still needed, that can be an additional config, but the primary user-facing concept should be support session catalog, not route some third-party connectors to memory.

2. Do not mix the new logic into GravitinoCatalogStore.java.
If supportsSessionCatalog is enabled, use a new GravitinoSessionCatalogStore; otherwise keep using the original GravitinoCatalogStore.
This feature introduces a different catalog lifecycle model, so it should have a separate implementation boundary rather than adding more branching into the existing class.

3. The documentation should explicitly define the behavior of session catalogs.
Please document the semantics of create/get-use/drop/show-list, the lifecycle of session catalogs, restart behavior, and the precedence/conflict rule when the same catalog name exists in both places.

4. Please add integration tests to cover the full session catalog lifecycle.
Unit tests are not enough for this feature. We should have IT coverage for all major operations on session catalogs, including at least create, get/use, show/list, and drop.

docs/flink-connector/flink-connector.md
```diff
@@ -35,8 +35,20 @@ This capability allows users to perform federation queries, accessing data from
...
 | table.catalog-store.gravitino.gravitino.metalake | string | (none)            | The metalake name that flink connector used to request to Gravitino. | Yes      | 0.6.0-incubating |
 | table.catalog-store.gravitino.gravitino.uri      | string | (none)            | The uri of Gravitino server address.                                 | Yes      | 0.6.0-incubating |
+| table.catalog-store.gravitino.gravitino.support.session.catalog | boolean | false | Whether to enable support for Flink's session catalog in the Gravitino catalog store. | No | 1.3.0 |
 | table.catalog-store.gravitino.gravitino.client.  | string | (none)            | The configuration key prefix for the Gravitino client config.        | No       | 1.0.0            |
 
+When `table.catalog-store.gravitino.gravitino.support.session.catalog` is set to `true`, Gravitino uses `GravitinoSessionCatalogStore`, which combines a `GravitinoCatalogStore` (backed by the Gravitino server) with an in-memory store to support Flink's session catalog. When `false` (the default), only `GravitinoCatalogStore` is used.
+
```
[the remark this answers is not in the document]
lasdf1234: I Have modify PR description

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @param catalogName name of the catalog to retrieve
+   * @return the requested catalog or empty if the catalog does not exist
+   * @throws CatalogException throw a CatalogException when the Catalog cannot be created.
+   */
+  @Override
+  public Optional<CatalogDescriptor> getCatalog(String catalogName) throws CatalogException {
+    if (memoryCatalogStore.contains(catalogName)) {
+      Optional<CatalogDescriptor> descriptor = memoryCatalogStore.getCatalog(catalogName);
+      if (descriptor.isPresent()) {
```
FANNG1: The two `catch` blocks in `listCatalogs()` have identical bodies and can be merged:

```java
} catch (CatalogException | RuntimeException e) {
  throw new CatalogException("Failed to list catalog.", e);
}
```

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
```diff
@@ -43,10 +45,20 @@
```
FANNG1, line 42: `createCatalogStore()` creates a new `GravitinoCatalogStore(catalogManager)` on every invocation, while `memoryCatalogStore` is lazily initialised and reused. This means every returned `GravitinoSessionCatalogStore` wraps a *different* `gravitinoCatalogStore` instance while sharing the same `memoryCatalogStore`, which leads to inconsistent lifecycles.

Consider promoting `GravitinoCatalogStore` to a factory-level field and initialising it once in `open()`, mirroring the treatment of `memoryCatalogStore`.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/utils/FactoryUtils.java
```diff
@@ -36,6 +41,19 @@ private FactoryUtils() {}
 
   private static final Logger LOG = LoggerFactory.getLogger(FactoryUtils.class);
 
+  /** The set of catalog type identifiers managed by Gravitino. */
+  private static final ImmutableSet<String> GRAVITINO_CATALOG_TYPES =
+      ImmutableSet.of(
+          GravitinoHiveCatalogFactoryOptions.IDENTIFIER,
+          GravitinoIcebergCatalogFactoryOptions.IDENTIFIER,
+          GravitinoJdbcCatalogFactoryOptions.MYSQL_IDENTIFIER,
```
FANNG1, line 49: `isBuiltInCatalog` is misleading — it actually checks whether a type belongs to the set of *Gravitino-managed* catalog identifiers, not Flink built-in types. Consider renaming it to `isGravitinoManagedCatalogType` (or `isGravitinoCatalogType`) to avoid confusion for future readers.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactoryOptions.java
```diff
@@ -40,7 +40,12 @@ private GravitinoCatalogStoreFactoryOptions() {}
           .stringType()
           .noDefaultValue()
           .withDescription("The name of Gravitino metalake");
-
+  public static final ConfigOption<Boolean> GRAVITINO_SUPPORT_SESSION_CATALOG =
+      ConfigOptions.key("gravitino.support.session.catalog")
+          .booleanType()
+          .defaultValue(false)
+          .withDescription(
```
FANNG1, line 47: Missing blank line after the `withDescription(...)` closing `);`. All other constant definitions in this file are separated by a blank line — please keep the style consistent.

flink-connector/flink/src/test/java/org/apache/gravitino/flink/connector/store/TestGravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.Optional;
+import java.util.Set;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.table.catalog.CatalogDescriptor;
+import org.apache.flink.table.catalog.CommonCatalogOptions;
+import org.apache.flink.table.catalog.GenericInMemoryCatalogStore;
+import org.apache.flink.table.catalog.exceptions.CatalogException;
+import org.junit.Before;
+import org.junit.Test;
+
```
FANNG1, line 38: This test class uses JUnit 4 (`org.junit.Test`, `@Before`), while the rest of the module (e.g. `TestGravitinoCatalogStoreFactory`) uses JUnit 5 (`org.junit.jupiter.api.*`). The build is configured with `useJUnitPlatform()`, so **without the JUnit Vintage engine these tests may never actually execute**, making the claimed coverage illusory. Please migrate to JUnit 5 for consistency.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,120 @@
...
+   */
+  @Override
+  public void removeCatalog(String catalogName, boolean ignoreIfNotExists) throws CatalogException {
+    if (memoryCatalogStore.contains(catalogName)) {
+      memoryCatalogStore.removeCatalog(catalogName, ignoreIfNotExists);
+    } else {
+      gravitinoCatalogStore.removeCatalog(catalogName, ignoreIfNotExists);
+    }
+  }
+
```
FANNG1, line 99: `getCatalog` calls `contains()` and then `getCatalog()`, which is a TOCTOU (check-then-act) race: in a concurrent scenario the catalog could be removed between the two calls. The logic can be simplified and made safer:

```java
Optional<CatalogDescriptor> descriptor = memoryCatalogStore.getCatalog(catalogName);
return descriptor.isPresent() ? descriptor : gravitinoCatalogStore.getCatalog(catalogName);
```

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,120 @@
...
+    this.memoryCatalogStore =
+        Preconditions.checkNotNull(memoryCatalogStore, "MemoryCatalogStore cannot be null");
+  }
+
+  @Override
+  public void storeCatalog(String catalogName, CatalogDescriptor descriptor)
+      throws CatalogException {
+    String catalogType = descriptor.getConfiguration().get(CommonCatalogOptions.CATALOG_TYPE);
+    if (isBuiltInCatalog(catalogType)) {
+      gravitinoCatalogStore.storeCatalog(catalogName, descriptor);
```
FANNG1, line 78: `descriptor.getConfiguration().get(CommonCatalogOptions.CATALOG_TYPE)` can return `null` when the catalog type property is absent. In that case `isBuiltInCatalog(null)` returns `false`, silently routing the catalog to the in-memory store instead of failing with a clear error. Please add an explicit null-check and throw a `CatalogException` with a descriptive message.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactoryOptions.java
```diff
@@ -40,7 +40,12 @@ private GravitinoCatalogStoreFactoryOptions() {}
           .stringType()
           .noDefaultValue()
           .withDescription("The name of Gravitino metalake");
-
+  public static final ConfigOption<Boolean> GRAVITINO_SUPPORT_SESSION_CATALOG =
```
FANNG1: The config key `gravitino.support.session.catalog` uses a dot-separated mixed style. For consistency with other Flink config options (which typically use camelCase for sub-keys), consider renaming it to `gravitino.supportsSessionCatalog`.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
```diff
@@ -43,10 +45,17 @@
...
 
   @Override
   public CatalogStore createCatalogStore() {
-    return new GravitinoCatalogStore(catalogManager);
+    GravitinoCatalogStore gravitinoCatalogStore = new GravitinoCatalogStore(catalogManager);
+    if (supportSessionCatalog) {
+      memoryCatalogStore = new GenericInMemoryCatalogStore();
+      return new GravitinoSessionCatalogStore(gravitinoCatalogStore, memoryCatalogStore);
+    }
+    return gravitinoCatalogStore;
```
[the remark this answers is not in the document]
lasdf1234: Resolved I have add class type test.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
```diff
@@ -43,10 +45,18 @@
...
   @Override
   public CatalogStore createCatalogStore() {
-    return new GravitinoCatalogStore(catalogManager);
+    GravitinoCatalogStore gravitinoCatalogStore = new GravitinoCatalogStore(catalogManager);
+    if (supportSessionCatalog) {
+      memoryCatalogStore = new GenericInMemoryCatalogStore();
+      memoryCatalogStore.open();
+      return new GravitinoSessionCatalogStore(gravitinoCatalogStore, memoryCatalogStore);
+    }
+    return gravitinoCatalogStore;
```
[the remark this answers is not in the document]
lasdf1234: Resolved I have fiexd it.

flink-connector/flink/src/test/java/org/apache/gravitino/flink/connector/store/TestGravitinoSessionCatalogStore.java
lasdf1234: I have fixed it.

lasdf1234:
@FANNG1  I have fixed all comment, you can review.

flink-connector/flink/src/test/java/org/apache/gravitino/flink/connector/integration/test/session/FlinkSupportsSessionCatalogIT.java
```diff
@@ -0,0 +1,337 @@
...
+
+  private static void initHiveConfDir() {
+    if (hiveConfDir != null) {
+      return;
+    }
+    try {
+      java.nio.file.Path dir = java.nio.file.Files.createTempDirectory("flink-session-hive-conf");
+      java.nio.file.Path hiveSite = dir.resolve("hive-site.xml");
+      String hiveSiteXml =
+          "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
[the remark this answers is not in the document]
lasdf1234: Got resolved it.

docs/flink-connector/flink-connector.md
```diff
@@ -35,8 +35,20 @@ This capability allows users to perform federation queries, accessing data from
...
 | table.catalog-store.gravitino.gravitino.metalake | string | (none)            | The metalake name that flink connector used to request to Gravitino. | Yes      | 0.6.0-incubating |
 | table.catalog-store.gravitino.gravitino.uri      | string | (none)            | The uri of Gravitino server address.                                 | Yes      | 0.6.0-incubating |
+| table.catalog-store.gravitino.gravitino.supportsSessionCatalog | boolean | false | Whether to enable support for Flink's session catalog in the Gravitino catalog store. | No | 1.3.0 |
 | table.catalog-store.gravitino.gravitino.client.  | string | (none)            | The configuration key prefix for the Gravitino client config.        | No       | 1.0.0            |
 
+When `table.catalog-store.gravitino.gravitino.supportsSessionCatalog` is set to `true`, Gravitino uses `GravitinoSessionCatalogStore`, which combines a `GravitinoCatalogStore` (backed by the Gravitino server) with an in-memory store to support Flink's session catalog. When `false` (the default), only `GravitinoCatalogStore` is used.
```
FANNG1: please polish the doc to make more user friendly.  and you introduce concepts like  `Gravitino-managed catalogs`, please explain it for the user.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
```diff
@@ -23,6 +23,7 @@
 import static org.apache.gravitino.flink.connector.store.GravitinoCatalogStoreFactoryOptions.GRAVITINO;
 import static org.apache.gravitino.flink.connector.store.GravitinoCatalogStoreFactoryOptions.GRAVITINO_CLIENT_CONFIG;
 import static org.apache.gravitino.flink.connector.store.GravitinoCatalogStoreFactoryOptions.GRAVITINO_METALAKE;
+import static org.apache.gravitino.flink.connector.store.GravitinoCatalogStoreFactoryOptions.GRAVITINO_SUPPORT_SESSION_CATALOG;
```
FANNG1: SUPPORT -> SUPPORTS

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
```diff
@@ -43,10 +45,20 @@
 /** The Factory for creating {@link GravitinoCatalogStore}. */
 public class GravitinoCatalogStoreFactory implements CatalogStoreFactory {
   private GravitinoCatalogManager catalogManager;
+  private boolean supportSessionCatalog;
```
FANNG1: support -> supports

FANNG1: support is widely used in this PR, please check all.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,122 @@
...
+ * while delegating persistent catalogs to Apache Gravitino.
+ */
+public class GravitinoSessionCatalogStore extends AbstractCatalogStore {
+  private final GenericInMemoryCatalogStore memoryCatalogStore;
+  private final GravitinoCatalogStore gravitinoCatalogStore;
+
+  public GravitinoSessionCatalogStore(
+      GravitinoCatalogStore gravitinoCatalogStore, GenericInMemoryCatalogStore memoryCatalogStore) {
+    this.gravitinoCatalogStore =
+        Preconditions.checkNotNull(gravitinoCatalogStore, "CatalogStore cannot be null");
```
FANNG1: `Preconditions.checkNotNull` is not preferred, as it will throw null pointer exception, please use `Precondition.checkArgument` instead

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.flink.table.catalog.CatalogDescriptor;
+import org.apache.flink.table.catalog.CommonCatalogOptions;
+import org.apache.flink.table.catalog.GenericInMemoryCatalogStore;
+import org.apache.flink.table.catalog.exceptions.CatalogException;
+import org.apache.flink.util.Preconditions;
+
+/**
+ * A catalog store that combines a session-scoped in-memory {@link GenericInMemoryCatalogStore} with
+ * a persistent {@link GravitinoCatalogStore}.
+ *
```
FANNG1, line 37: The class-level JavaDoc says "Catalogs for **built-in** catalog types are stored only in the **in-memory store**, while all other catalogs are stored in the Gravitino-backed store." — this is the opposite of the actual implementation. Gravitino-managed types (e.g. `gravitino-hive`) are routed to the Gravitino-backed store, while non-managed (third-party) types are stored in memory. Please update the JavaDoc to match the implementation.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public Optional<CatalogDescriptor> getCatalog(String catalogName) throws CatalogException {
+    Optional<CatalogDescriptor> descriptor = memoryCatalogStore.getCatalog(catalogName);
+    return descriptor.isPresent() ? descriptor : gravitinoCatalogStore.getCatalog(catalogName);
+  }
+
+  @Override
+  public Set<String> listCatalogs() throws CatalogException {
+    Set<String> catalogs = new HashSet<>();
+    catalogs.addAll(memoryCatalogStore.listCatalogs());
```
FANNG1, line 116: The two `catch` blocks have identical bodies and can be merged into one:

```java
} catch (CatalogException | RuntimeException e) {
  throw new CatalogException("Failed to list catalog.", e);
}
```

lasdf1234: If use this：
} catch (CatalogException | RuntimeException e) {
  throw new CatalogException("Failed to list catalog.", e);
}
Compil failed, so don't change.

FANNG1: Withdrawn — `CatalogException` extends `RuntimeException`, so Java does not allow them together in a multi-catch (`catch (CatalogException | RuntimeException e)` fails to compile). The two separate `catch` blocks are correct as-is; no change needed.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
lasdf1234: I have check other oss project,If config key contains verb, verb doesn't contain "s", so I prefer use support in gravitino

lasdf1234: @roryqi  cc

lasdf1234: Resolved

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
lasdf1234: Got resolved.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
lasdf1234: Rename GRAVITINO_SUPPORT_SESSION_CATALOG to GRAVITINO_ENABLE_SESSION_CATALOG_SUPPORT

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
lasdf1234: Got resolved

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoCatalogStoreFactory.java
lasdf1234: Got resolved it

lasdf1234: Got resolved it

docs/flink-connector/flink-connector.md
lasdf1234: Got resolved it.

lasdf1234:
@FANNG1 I have resolved all comment ,please review

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+    if (isGravitinoManagedCatalogType(catalogType)) {
+      gravitinoCatalogStore.storeCatalog(catalogName, descriptor);
+    } else {
+      memoryCatalogStore.storeCatalog(catalogName, descriptor);
+    }
+  }
+  /**
+   * Removes the specified catalog.
+   *
```
FANNG1, line 86: Missing blank line between `storeCatalog` and the Javadoc for `removeCatalog`. Please add a blank line to separate the two methods, otherwise `./gradlew spotlessApply` will flag this.

FANNG1:
LGTM, except minor comments

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/store/GravitinoSessionCatalogStore.java
lasdf1234: Got resolved it.

Issue #10537, opened 2026-03-25
[Improvement] Complete audit mapping coverage for listener events


FANNG1:
**What would you like to be improved?**

`core/src/main/java/org/apache/gravitino/audit/v2/CompatibilityUtils.java` and `core/src/main/java/org/apache/gravitino/audit/AuditLog.java` do not provide audit coverage for many existing listener event operation types.

At the moment, a large set of `OperationType` values is still not mapped to audit operations, so these events fall back to `UNKNOWN_OPERATION` in audit output or are not explicitly covered in the legacy audit mapping. This makes audit logs incomplete and inconsistent across feature areas.

The missing coverage currently includes these areas:
- Tag events
- Model and model version events
- User, group, role, and privilege events
- Owner events
- Job and job template events
- Statistics events
- Policy events
- Some table and fileset operations such as `LOAD_TABLE_CREDENTIAL`, `PLAN_TABLE_SCAN`, and `LIST_FILESET_FILES`

This came up while reviewing PR #10488, which correctly adds audit mapping for the newly introduced `ENABLE/DISABLE` metalake and catalog events, but also highlights that broader audit coverage is still incomplete.

**How should we improve?**

Add explicit audit mappings for the remaining `OperationType` values in:
- `core/src/main/java/org/apache/gravitino/audit/v2/CompatibilityUtils.java`
- `core/src/main/java/org/apache/gravitino/audit/AuditLog.java`

Suggested follow-up:
- Introduce dedicated `AuditLog.Operation` values where needed instead of defaulting to `UNKNOWN_OPERATION`
- Add tests to ensure every supported `OperationType` is either explicitly mapped to a concrete audit operation or intentionally mapped to `UNKNOWN_OPERATION`
- Consider adding a guard test that fails when new `OperationType` values are added without updating audit compatibility mappings

PR #10566, opened 2026-03-27
[MINOR] docs: clarify Java import guidance in AGENTS.md

FANNG1:

Clarify the Java import guidance in `AGENTS.md` by making the existing rule more explicit:
- require normal `import` statements instead of inline fully qualified class names when possible
- add concrete examples for disallowed inline usages such as `java.nio.file.Paths`
- document the preferred handling when two classes share the same simple name

The previous wording only said to prioritize imports over fully qualified class names, which was too weak and still allowed agents to generate Java code with unnecessary inline FQNs. This update makes the expectation explicit and reduces that behavior.

Fix: #10537

No.

Documentation only change. Verified the updated wording in `AGENTS.md`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,9 +22,11 @@
 ## General Coding Standards
 - **Language**: Use English for all code, comments, and documentation.
 - **Style**: Follow rigid Google Java Style. Run `./gradlew spotlessApply` to format.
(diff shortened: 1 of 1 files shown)


jerryshao:
@FANNG1 Is this PR related to issue #10537?

PR #10573, opened 2026-03-27
[MINOR] fix(helm): add serviceAccountName to pod spec in deployment.yaml

bharos:

Add serviceAccountName: {{ .Values.serviceAccountName | default "default" }} so the configured KSA (with WI annotation) is used.

The upstream chart had no serviceAccountName field in the pod spec, causing the pod to always run as the 'default' KSA regardless of the values file. This meant Workload Identity bindings on the gravitino KSA were silently ignored, resulting in 403s from GCS.

no

Test with helm deployment

### DIFF
--- a/dev/charts/gravitino/templates/deployment.yaml
+++ b/dev/charts/gravitino/templates/deployment.yaml
@@ -47,6 +47,7 @@ spec:
           {{- toYaml . | nindent 8 }}
           {{- end }}
     spec:
+      serviceAccountName: {{ .Values.serviceAccountName }}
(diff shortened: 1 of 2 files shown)


roryqi:
cc @danhuawang Could u help me review this pull request?

danhuawang:
LGTM

jerryshao:
Shall we backport this to branch-1.2?

roryqi:
> Shall we backport this to branch-1.2?

Yes, we should. Add the `branch-1.2` label.

Issue #10578, opened 2026-03-30
[Improvement] Exclude java-xmlbuilder transitive dependency from hadoop-common


geyanggang:
### What would you like to be improved?

java-xmlbuilder:0.4 is pulled in transitively via hadoop-common:2.10.2 → jets3t:0.9.0 into three catalog modules: catalog-lakehouse-paimon, hive-metastore2-libs, and hive-metastore3-libs.

### How should we improve?

Exclude net.java.dev.jets3t from the hadoop-common dependency block in all three affected build.gradle.kts files. jets3t is an obsolete S3 client that Gravitino does not use at runtime, so excluding it is safe and also removes java-xmlbuilder.

PR #10579, opened 2026-03-30
[#10578]improvement(catalogs): Exclude unused transitive jets3t/java-xmlbuilder from hadoop-common

geyanggang:

Added `exclude(group = "net.java.dev.jets3t")` to the hadoop-common dependency block in `catalogs/catalog-lakehouse-paimon/build.gradle.kts`,`catalogs/hive-metastore2-libs/build.gradle.kts`,`catalogs/hive-metastore3-libs/build.gradle.kts`

`java-xmlbuilder:0.4` is pulled in transitively via `hadoop-common:2.10.2 `→ `jets3t:0.9.0`. jets3t is an obsolete Amazon S3 client library that Gravitino does not use at runtime. Excluding it removes both jets3t and java-xmlbuilder from the dependency tree.

Fix: #10578 

No.

1. Verified compileJava passes for all three affected modules.
2. Verified unit tests pass for catalog-lakehouse-paimon.
3. Confirmed java-xmlbuilder and jets3t no longer appear in the runtimeClasspath dependency tree of all three modules.

### DIFF
--- a/catalogs/catalog-lakehouse-paimon/build.gradle.kts
+++ b/catalogs/catalog-lakehouse-paimon/build.gradle.kts
@@ -108,7 +108,7 @@ dependencies {
     exclude("com.sun.jersey")
     exclude("javax.servlet")
     exclude("org.apache.avro")
-    exclude(group = "net.java.dev.jets3t")
+    exclude("net.java.dev.jets3t")
     exclude("org.apache.curator")
     exclude("org.apache.zookeeper")
     exclude("org.mortbay.jetty")


catalogs/hive-metastore2-libs/build.gradle.kts
```diff
@@ -34,6 +34,7 @@ dependencies {
     exclude(group = "com.google.code.findbugs")
     exclude(group = "com.google.guava")
     exclude(group = "log4j")
+    exclude(group = "net.java.dev.jets3t")
```
yuqi1129, line 37: Potential runtime compatibility risk: excluding `net.java.dev.jets3t` from Hadoop common can break legacy `s3n://` access paths with `ClassNotFoundException` (jets3t-backed S3N classes). If Gravitino intentionally no longer supports S3N, please document that compatibility boundary in release notes/config docs; otherwise we should keep this dependency for backward compatibility.

catalogs/catalog-lakehouse-paimon/build.gradle.kts
```diff
@@ -108,6 +108,7 @@ dependencies {
     exclude("com.sun.jersey")
     exclude("javax.servlet")
     exclude("org.apache.avro")
+    exclude("net.java.dev.jets3t")
```
yuqi1129, line 111: Same compatibility concern as above: excluding `net.java.dev.jets3t` may break environments still using legacy `s3n://` FS implementation. Please confirm S3N is explicitly out-of-scope; if yes, suggest documenting it to avoid surprise runtime failures.

catalogs/hive-metastore2-libs/build.gradle.kts
geyanggang: I verified this on the main branch (before this PR) by scanning every JAR in hive-metastore2-libs/build/libs — including hadoop-common-2.10.2.jar — for NativeS3FileSystem and s3native classes. Zero matches. The S3N filesystem implementation is not bundled in hadoop-common; it lives in hadoop-aws, which is not a dependency of this module.

So jets3t is unreachable dead code here — no class in these JARs imports or loads it. Additionally, Gravitino has no s3n:// usage anywhere in the codebase; all S3 access uses s3a://.

catalogs/catalog-lakehouse-paimon/build.gradle.kts
geyanggang: Same verification result: no JAR in catalog-lakehouse-paimon/build/libs contains NativeS3FileSystem or any s3native class. The hadoop2.common dependency used here does not bundle the S3N filesystem — that class requires hadoop-aws. jets3t is dead code in this module as well.

Gravitino uses s3a:// exclusively for S3 access. s3n:// is a legacy Hadoop protocol (deprecated since Hadoop 3.x) and is not supported.

catalogs/hive-metastore2-libs/build.gradle.kts
diqiu50: I think HMS probably doesn’t need it.

diqiu50: I think HMS probably doesn’t need it.

yuqi1129: OK,  I have no other comments.

Issue #10557, opened 2026-03-26
[Bug report] Charts should use start-gravitino.sh instead of gravitino.sh to launch the service


danhuawang:
### Version

main branch

### Describe what's wrong

Gravitino Chart should use start-gravitino.sh instead of gravitino.sh to launch the service
Lance Chart should use start-lance-rest-server.sh to launch the service
Iceberg Chart should use start-iceberg-rest-server.sh to launch the service

### Error message and/or stacktrace

```
2026-03-26 11:24:46 ERROR [main] GravitinoServer:212 - Error while running jettyServer
org.apache.ibatis.exceptions.PersistenceException: 
### Error querying database.  Cause: java.sql.SQLException: Cannot load JDBC driver class 'com.mysql.cj.jdbc.Driver'
### The error may exist in org/apache/gravitino/storage/relational/mapper/MetalakeMetaMapper.java (best guess)
### The error may involve org.apache.gravitino.storage.relational.mapper.MetalakeMetaMapper.listMetalakePOs
### The error occurred while executing a query
### Cause: java.sql.SQLException: Cannot load JDBC driver class 'com.mysql.cj.jdbc.Driver'
	at org.apache.ibatis.exceptions.ExceptionFactory.wrapException(ExceptionFactory.java:30) ~[mybatis-3.5.6.jar:3.5.6]
	at org.apache.ibatis.session.defaults.DefaultSqlSession.selectList(DefaultSqlSession.java:149) ~[mybatis-3.5.6.jar:3.5.6]
	at org.apache.ibatis.session.defaults.DefaultSqlSession.selectList(DefaultSqlSession.java:140) ~[mybatis-3.5.6.jar:3.5.6]
	at org.apache.ibatis.binding.MapperMethod.executeForMany(MapperMethod.java:147) ~[mybatis-3.5.6.jar:3.5.6]
	at org.apache.ibatis.binding.MapperMethod.execute(MapperMethod.java:80) ~[mybatis-3.5.6.jar:3.5.6]
	at org.apache.ibatis.binding.MapperProxy$PlainMethodInvoker.invoke(MapperProxy.java:152) ~[mybatis-3.5.6.jar:3.5.6]
	at org.apache.ibatis.binding.MapperProxy.invoke(MapperProxy.java:85) ~[mybatis-3.5.6.jar:3.5.6]
	at jdk.proxy2/jdk.proxy2.$Proxy31.listMetalakePOs(Unknown Source) ~[?:?]
[...]

PR #10583, opened 2026-03-30
[#10557] fix(chart): use start-gravitino.sh instead of gravitino.sh to launch the service

danhuawang:

Gravitino Chart should use start-gravitino.sh instead of gravitino.sh to launch the service
Lance Chart should use start-lance-rest-server.sh to launch the service
Iceberg Chart should use start-iceberg-rest-server.sh to launch the service

Fix: #10557

N/A

Manually test and CI check.

### DIFF
--- a/dev/charts/gravitino/resources/scenarios/ci-values.yaml
+++ b/dev/charts/gravitino/resources/scenarios/ci-values.yaml
@@ -50,8 +50,6 @@ extraVolumes:
 replicas: 2
 
 env:
-  - name: GRAVITINO_HOME
-    value: /root/gravitino
   - name: HADOOP_USER_NAME
     value: hdfs
 
--- a/dev/charts/gravitino/resources/scenarios/pg-values.yaml
+++ b/dev/charts/gravitino/resources/scenarios/pg-values.yaml
@@ -21,6 +21,5 @@ postgresql:
 
 initScript: |
   cp /tmp/conf/* ${GRAVITINO_HOME}/conf
-  wget https://jdbc.postgresql.org/download/postgresql-42.7.0.jar -O /root/gravitino/libs/postgresql-42.7.0.jar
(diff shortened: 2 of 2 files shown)


danhuawang:
@roryqi Could you help review this PR ? Thanks.

dev/docker/gravitino/start-gravitino.sh
```diff
@@ -24,7 +24,10 @@ gravitino_dir="$(cd "${bin_dir}/../">/dev/null; pwd)"
 
 cd ${gravitino_dir}
 
-python bin/rewrite_gravitino_server_config.py
+# Skip config rewrite if SKIP_CONFIG_REWRITE is set (e.g., in Kubernetes)
```
roryqi, line 27: Why do we need this?

danhuawang: Because this rewrite script will overlap the configurations in chart values yaml.

roryqi: Could they co-work together?

jerryshao:
Is this a newly introduced bug or existed in day 1?

dev/docker/gravitino/start-gravitino.sh
danhuawang: No, if co-work, rewrite_gravitino_server_config.py maintain a few environment variables and default values, but the user can't know these default values by chat cli.

danhuawang:
> Is this a newly introduced bug or existed in day 1?

It's a newly introduced bug. The chart don't have to use this script to launch service until  the soft links creation logic is added in `start-gravitino.sh`.

dev/docker/gravitino/start-gravitino.sh
roryqi: Do u need to add documents explain this?

Issue #10584, opened 2026-03-30
[Improvement] Align jackson-databind version in hive-metastore2-libs and hive-metastore3-libs with project-wide version


geyanggang:
### What would you like to be improved?

`hive-metastore2-libs` ships `jackson-databind:2.6.5` and `hive-metastore3-libs` ships `jackson-databind:2.12.0`, both pulled in transitively via `hive-metastore` → `hive-serde` → `hive-common`. The project-wide Jackson version defined in `libs.versions.toml` is `2.15.2`. These outdated transitive versions should be aligned for consistency and to benefit from upstream bug fixes and improvements.

### How should we improve?

Use resolutionStrategy.force in both `build.gradle.kts` to force `jackson-databind`,` jackson-core`, and `jackson-annotations` to the project-wide version `2.15.2`.

This is safe because:
1. Jackson 2.x maintains backward compatibility across minor versions.
2. Only basic Jackson APIs (`ObjectMapper`, `ObjectReader`, `ObjectWriter`, `JsonSerializer`, `JsonDeserializer`) are used in `hive-common`, via `StatsSetupConst$ColumnStatsAccurate `and `CodahaleMetrics$JsonFileReporter` (Hive 2) / `JsonFileMetricsReporter` (Hive 3). These APIs are stable across 2.6–2.15.
3. Hive metastore's JSON messaging layer uses Jackson 1.x (`org.codehaus.jackson`), which is unaffected.
4. These classes are loaded via `HiveClientClassLoader` in isolation and are not directly referenced by Gravitino code.

PR #10585, opened 2026-03-30
[#10584] improvement(catalogs): Exclude unused transitive jackson-databind from hive-metastore libs

geyanggang:

Added `exclude(group = "com.fasterxml.jackson.core")` to the `hadoop2.common` and `hive2.metastore`/`hive3.metastore `dependency blocks in `build.gradle.kts`.

`jackson-databind:2.6.5` (in hive-metastore2-libs) and `jackson-databind:2.12.0` (in `hive-metastore3-libs`) are outdated transitive dependencies pulled in via `hive-metastore` → `hive-serde` → `hive-common`.

Although `hive-common `references `com.fasterxml.jackson` at the bytecode level (in `CodahaleMetrics$JsonFileReporter` and `StatsSetupConst$ColumnStatsAccurate`), these classes are never loaded in Gravitino's runtime path:

`CodahaleMetrics$JsonFileReporter` is a Hive server-side metrics component, only activated when Hive Server configures a specific metrics class.
`StatsSetupConst$ColumnStatsAccurate` is an inner class loaded lazily by the JVM; HiveMetaStoreClient (used by Gravitino) does not invoke any method that triggers its loading.
Gravitino code has zero references to either class.
Therefore, `com.fasterxml.jackson.core` can be safely excluded. Jackson 1.x (`org.codehaus.jackson`), which is used by Hive metastore's JSON messaging layer, is unaffected.
Fix: #10584 

No.

1. Verified build passes for `hive-metastore2-libs`, `hive-metastore3-libs`, `hive-metastore-common`, and `catalog-hive`.
2. Verified tests pass for `hive-metastore-common` and `catalog-hive`.
3. Confirmed `jackson-databind`, `jackson-core`, and `jackson-annotations (2.x)` no longer appear in the build/libs after a clean build.

### DIFF
--- a/catalogs/hive-metastore3-libs/build.gradle.kts
+++ b/catalogs/hive-metastore3-libs/build.gradle.kts
@@ -27,9 +27,11 @@ plugins {
 
 // Note: exclusion lists here are intentionally kept in sync with hive-metastore2-libs/build.gradle.kts.
 // Guava and Logback are excluded because they are provided by the Gravitino runtime classpath.
+
 dependencies {
   implementation(libs.hadoop2.common) {
     exclude(group = "ch.qos.logback")
+    exclude(group = "com.fasterxml.jackson.core")
     exclude(group = "com.github.spotbugs")
     exclude(group = "com.google.code.findbugs")
     exclude(group = "com.google.guava")
@@ -44,6 +46,7 @@ dependencies {
     exclude(group = "ant")
     exclude(group = "ch.qos.logback")
     exclude(group = "co.cask.tephra")
+    exclude(group = "com.fasterxml.jackson.core")
     exclude(group = "com.github.joshelser")
     exclude(group = "com.github.spotbugs")
     exclude(group = "com.google.code.findbugs")
--- a/catalogs/hive-metastore2-libs/build.gradle.kts
+++ b/catalogs/hive-metastore2-libs/build.gradle.kts
@@ -27,9 +27,11 @@ plugins {
 
 // Note: exclusion lists here are intentionally kept in sync with hive-metastore3-libs/build.gradle.kts.
 // Guava and Logback are excluded because they are provided by the Gravitino runtime classpath.
+
 dependencies {
   implementation(libs.hadoop2.common) {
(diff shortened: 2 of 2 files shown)


catalogs/hive-metastore3-libs/build.gradle.kts
```diff
@@ -27,6 +27,18 @@ plugins {
...
 // Guava and Logback are excluded because they are provided by the Gravitino runtime classpath.
+
+// Force jackson-databind to the project-wide version to replace the outdated transitive version
+// pulled in via hive-metastore → hive-serde → hive-common.
+val jacksonVersion = libs.versions.jackson.get()
+configurations.all {
+  resolutionStrategy.force(
+    "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion",
+    "com.fasterxml.jackson.core:jackson-core:$jacksonVersion",
+    "com.fasterxml.jackson.core:jackson-annotations:$jacksonVersion"
```
yuqi1129: @geyanggang 
What's the version of Jackson used globally and Hive3? We need to judge these two versions to decide if it's safe.

diqiu50: This approach is not safe, although CI is currently passing. Can we exclude it?

geyanggang: It's OK. I Verified that excluding `com.fasterxml.jackson.core` works — build and tests pass. I will update the PR accordingly.

geyanggang: Updated the approach — switched from `resolutionStrategy.force` to exclude. After deeper analysis, `com.fasterxml.jackson.core` is not loaded at runtime in Gravitino's code path. The two classes in hive-common that reference it (`CodahaleMetrics$JsonFileReporter` and `StatsSetupConst$ColumnStatsAccurate`) are Hive server-side components; HiveMetaStoreClient never triggers their loading. Excluding is cleaner and safer.

diqiu50 (approved):
LGTM

Issue #10570, opened 2026-03-27
[Improvement] Supports REST Iceberg backend catalog


roryqi:
### What would you like to be improved?

Supports REST Iceberg backend catalog

### How should we improve?

Supports REST Iceberg backend catalog

PR #10586, opened 2026-03-30
[#10570] feat(iceberg): Support REST catalog backend

roryqi:

Support REST catalog backend.
1. Use the user access token to access remote REST catalog
2. When credential vending is enabled, use remote REST catalog vended-credentials instead of generating a credential
3. Fetch remote catalog config and retrieve the properties required by the client

Fix: #10570

No.

Added UTs.

I deploy two gravitino servers.
One is using aws, another is using gcp.
For every Gravitino server, I added a local postgres jdbc backend catalog and a remote catalog using rest.

I run spark-sql to test using credential vending and OAuth authentication

```
use aws_remote;
create database sales;
use sales;
create table customers (customer_id int, customer_name varchar(100), customer_email varchar(100));
describe extended customers;
insert into customers (customer_id, customer_name, customer_email) values (11,'Rory Brown','rory@123.com');
insert into customers (customer_id, customer_name, customer_email) values (12,'Jerry Washington','jerry@dt.com');
```

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
@@ -664,8 +664,6 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
 
   private LoadTableResponse stageTableCreateInternal(
       Namespace namespace, CreateTableRequest request) {
-    request.validate();
-
     TableIdentifier ident = TableIdentifier.of(namespace, request.name());
     if (catalog.tableExists(ident)) {
       throw new AlreadyExistsException("Table already exists: %s", ident);


iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -171,10 +197,10 @@ public void reloadHadoopConf() {
   public LoadTableResponse createTable(Namespace namespace, CreateTableRequest request) {
     request.validate();
     if (request.stageCreate()) {
-      return CatalogHandlers.stageTableCreate(catalog, namespace, request);
+      return stageTableCreateInternal(namespace, request);
```
bharos: The replacement of `CatalogHandlers.loadTable()`/`CatalogHandlers.createTable()` with `loadTableInternal()`/`createTableInternal()` is in the **base class**, so it changes behavior for JDBC, Hive, and Memory backends too — not just REST. Previously load-table responses had an empty `config` map; now they include FileIO properties via `retrieveFileIOProperties()` + credential filtering. For non-REST backends, `table.io().properties()` can include catalog-init credentials/URIs that were never exposed before.

Suggestion: Keep the base class using `CatalogHandlers` and only override in a REST-specific path.

core/src/main/java/org/apache/gravitino/UserPrincipal.java
```diff
@@ -77,7 +115,7 @@ public List<UserGroup> getGroups() {
 
   @Override
   public int hashCode() {
-    return Objects.hash(username, groups);
+    return Objects.hash(username, groups, accessToken);
```
bharos: Including `accessToken` (a transient credential) in `equals()`/`hashCode()` means two principals for the same user with different JWT tokens are now **unequal**. If `UserPrincipal` is used in `Set`s or `Map` keys, token rotation creates distinct identities for the same user. The access token is a transport concern, not identity — `equals`/`hashCode` should stay based on `username` + `groups` only.

core/src/main/java/org/apache/gravitino/utils/PrincipalUtils.java
```diff
@@ -31,11 +33,18 @@
...
 public class PrincipalUtils {
-
   private static final Logger LOG = LoggerFactory.getLogger(PrincipalUtils.class);
 
   private PrincipalUtils() {}
 
+  public static final Principal ANONYMOUS_PRINCIPAL =
+      new UserPrincipal(
+          AuthConstants.ANONYMOUS_USER,
+          AuthConstants.AUTHORIZATION_BASIC_HEADER
```
bharos, line 43: `ANONYMOUS_PRINCIPAL` now carries `Basic YW5vbnltb3Vz` as its access token. When the REST backend forwards this via `UserPrincipalForwardingAuthManager`, anonymous requests send a fabricated auth header to downstream services. Consider giving anonymous a `null` token and having the auth manager skip the Authorization header when none is available (rather than throwing `IllegalStateException`).

core/src/main/java/org/apache/gravitino/UserPrincipal.java
roryqi: Make sense.

core/src/main/java/org/apache/gravitino/utils/PrincipalUtils.java
roryqi: Anonymous requests are only allowed in the simple mode. The simple mode isn't safe. So it's ok to send a fabricated auth header.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
roryqi: OK.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -186,6 +181,64 @@ public Map<String, String> getCatalogConfigToClient() {
...
+              + DATA_ACCESS_VENDED_CREDENTIALS
+              + ","
+              + DATA_ACCESS_REMOTE_SIGNING
+              + "]");
+    }
+
+    properties.put(IcebergConstants.DATA_ACCESS, normalizedDataAccess);
+  }
+
+  private static void putIfValuePresent(
```
bharos: Is this unused ?

docs/iceberg-rest-service.md
```diff
@@ -157,6 +158,11 @@ gravitino.iceberg-rest.header.X-Iceberg-Access-Delegation = vended-credentials
 
 IRC1 must also configure S3 configurations if the client side requests credential vending.
 
+`data-access` is returned in `/v1/config` defaults for REST clients:
+
+- `vended-credentials`: clients should request credential vending (`X-Iceberg-Access-Delegation: vended-credentials`).
+- `remote-signing`: Gravitino doesn't support now.
```
bharos: nit : "Gravitino does not support this mode yet"

docs/lakehouse-iceberg-catalog.md
```diff
@@ -71,6 +71,15 @@ If you are using multiple JDBC catalog backends, setting `jdbc-initialize` to tr
...
 For the REST catalog backend, `warehouse` identifies the catalog in the Iceberg REST spec. In the Gravitino Iceberg REST server, `warehouse` maps to the catalog name. An empty value means the default catalog.
 
+`data-access` controls how the Iceberg REST client accesses table data when using a REST backend:
+
+| Property name  | Description                                                                                                             | Default value | Required | Since Version |
+|----------------|-------------------------------------------------------------------------------------------------------------------------|---------------|----------|---------------|
+| `data-access`  | Data access mode for REST catalog backend. Supported values are `vended-credentials` and `remote-signing`.              | (none)        | No       | 1.3.0         |
+
+- `vended-credentials`: request credential vending from the Iceberg REST server.
+- `remote-signing`: Gravitino doesn't support now.
```
bharos: nit : "Gravitino does not support this mode yet"

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -414,4 +440,114 @@ private TableMetadataCache loadTableMetadataCache(IcebergConfig config, Catalog
...
+                  properties, key -> catalogPropertiesToClientKeys.contains(key)))
+          // Keep only credential fields from FileIO properties before returning them to the client.
+          .addAllConfig(CredentialPropertyUtils.filterCredentialProperties(properties))
+          .build();
+    }
+
+    throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+  }
+
+  public LoadTableResponse stageTableCreateInternal(
```
bharos: This doesn't need to be public ?

roryqi:
@bharos Could u help me review this pull request?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -239,7 +265,7 @@ public LoadTableResponse updateTable(
     metadataCache.invalidate(tableIdentifier);
     LoadTableResponse loadTableResponse =
         CatalogHandlers.updateTable(catalog, tableIdentifier, updateTableRequest);
-    if (loadTableResponse != null) {
+    if (loadTableResponse != null && !(catalog instanceof RESTCatalog)) {
```
bharos: The metadata cache skip (!(catalog instanceof RESTCatalog)) can similarly become a protected boolean shouldCacheMetadata() that the subclass overrides to return false.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -414,4 +440,114 @@ private TableMetadataCache loadTableMetadataCache(IcebergConfig config, Catalog
         expireMinutes);
     return cache;
   }
+
+  private LoadTableResponse loadTableInternal(TableIdentifier ident) {
```
bharos: What if we make these 3 methods `protected` in the base class with the default CatalogHandlers behavior,and override them in CatalogWrapperForREST:

```
// Base class — clean defaults, no REST knowledge
protected LoadTableResponse loadTableInternal(TableIdentifier ident) {
    return CatalogHandlers.loadTable(catalog, ident);
}

protected LoadTableResponse createTableInternal(Namespace namespace, CreateTableRequest request) {
    return CatalogHandlers.createTable(catalog, namespace, request);
}

protected LoadTableResponse stageTableCreateInternal(Namespace namespace, CreateTableRequest request) {
    return CatalogHandlers.stageTableCreate(catalog, namespace, request);
}
```
and

```
// CatalogWrapperForREST — override with REST logic
@Override
protected LoadTableResponse loadTableInternal(TableIdentifier ident) {
    Table table = catalog.loadTable(ident);
    // ... BaseTable handling, credential filtering, etc.
}
```

This also lets catalogPropertiesToClientKeys and retrieveFileIOProperties() move back to the subclass, and removes all REST imports (RESTCatalog, BaseTable, BaseMetadataTable, InMemoryFileIO, etc.) from the base class.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -186,6 +181,64 @@ public Map<String, String> getCatalogConfigToClient() {
...
+        MapUtils.getFilteredMap(sourceProps, key -> catalogPropertiesToClientKeys.contains(key));
+    filtered = new HashMap<>(filtered);
+    validateAndNormalizeDataAccessProperty(filtered);
+
+    return Collections.unmodifiableMap(filtered);
+  }
+
+  @VisibleForTesting
+  static void validateAndNormalizeDataAccessProperty(Map<String, String> properties) {
+    String dataAccess = properties.get(IcebergConstants.ICEBERG_ACCESS_DELEGATION);
```
bharos, line 253: This method reads one key, writes another 
```
String dataAccess = properties.get(IcebergConstants.ICEBERG_ACCESS_DELEGATION);
// ^ reads "header.X-Iceberg-Access-Delegation"
...
properties.put(IcebergConstants.DATA_ACCESS, normalizedDataAccess);
// ^ writes "data-access"
```

After the method runs, the map has both keys. Is this the expected situation?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
roryqi: Yes, I should remove this.

docs/iceberg-rest-service.md
roryqi: Fixed.

docs/lakehouse-iceberg-catalog.md
roryqi: Fixed.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
roryqi: Yes, this should be private.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
roryqi: Good point.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
roryqi: After moving logic to CatalogWrapperForREST, we don't have this issue.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
roryqi: Removed the write logic. Thanks.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
bharos: I think this is still needed to be fixed ?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -96,12 +105,18 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
           "gcs-credential-file-path",
           GCSProperties.GRAVITINO_GCS_SERVICE_ACCOUNT_FILE);
 
+  protected static final Set<String> catalogPropertiesToClientKeys =
```
bharos: This was private earlier , still can keep private ?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -535,4 +617,111 @@ private static void replaceDeprecatedProperties(
...
+          .addAllConfig(CredentialPropertyUtils.filterCredentialProperties(properties))
+          .build();
+    }
+
+    throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+  }
+
+  private LoadTableResponse stageTableCreateInternal(
+      Namespace namespace, CreateTableRequest request) {
+    if (!(catalog instanceof RESTCatalog)) {
```
bharos: stageTableCreateInternal() is only called from createTableInternal(), which itself is only entered when catalog instanceof RESTCatalog.

So should we just remove this check code

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
roryqi: Yes, should be private.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
roryqi: Removed.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
roryqi: Sorry for my mistake. Removed this.

roryqi: I override the method `updateTable` in the `CatalogWrapperForREST`
```
  @Override
  public LoadTableResponse updateTable(
      TableIdentifier tableIdentifier, UpdateTableRequest updateTableRequest) {
    if (catalog instanceof RESTCatalog) {
      return CatalogHandlers.updateTable(catalog, tableIdentifier, updateTableRequest);
    } else {
      return super.updateTable(tableIdentifier, updateTableRequest);
    }
  }
```

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -535,4 +628,107 @@ private static void replaceDeprecatedProperties(
...
+          .addAllConfig(CredentialPropertyUtils.filterCredentialProperties(properties))
+          .build();
+    }
+
+    throw new IllegalStateException("Cannot wrap catalog that does not produce BaseTable");
+  }
+
+  private LoadTableResponse stageTableCreateInternal(
+      Namespace namespace, CreateTableRequest request) {
+    request.validate();
```
bharos: Nit : currently this is only called from createTableInternal, which already does request.validate , so this is redundant.

Upto you to remove it

roryqi: Removed. Thanks.

Issue #10459, opened 2026-03-17
[Improvement] Improve the performance of the getJobObjectsFullName,getPolicyObjectsFullName,getTagObjectsFullName, getJobTemplateObjectsFullName


roryqi:
### What would you like to be improved?

Improve the performance of the getJobObjectsFullName,getPolicyObjectsFullName,getTagObjectsFullName, getJobTemplateObjectsFullName

### How should we improve?

Improve the performance of the getJobObjectsFullName,getPolicyObjectsFullName,getTagObjectsFullName, getJobTemplateObjectsFullName

babumahesh:
Hi @roryqi , 
I would like to work on this improvement. Could you please assign it to me or confirm I can proceed?

roryqi:
> Hi [@roryqi](https://github.com/roryqi) , I would like to work on this improvement. Could you please assign it to me or confirm I can proceed?

You can continue this issue. Thanks.

PR #10589, opened 2026-03-30
 [#10459] improvement(core): Eliminate N+1 queries in metadata object

babumahesh:

Eliminates N+1 query problems when resolving metadata object names for Tags, Policies, and Job Templates by implementing batch query methods.

### Why are the changes needed

- Previously, when fetching names for multiple metadata objects (tags/policies/job templates), the code would issue one SQL query per object ID in a stream. This creates N+1 query performance problems, especially when dealing with large numbers of objects.
- Implemented batch query methods using MyBatis `<foreach>` to fetch multiple records in a single SQL query with an `IN` clause.
- The service layer now batches all IDs and makes one query instead of N individual queries.

Fix: #10459 

All the existing unit tests and integrations tests ran and should be sufficient as there is no behaviour changes


babumahesh:
@roryqi Please review and let me know if any further action is needed.

roryqi:
> @roryqi Please review and let me know if any further action is needed.

Thanks for your pull request
1. Could u add unit tests for this pull request?
2. Could u use `Maps.newHashMap()` instead of `Map.of`? `Map.of` can't be used in JDK8. Our client need to used for JDK8, although our server can only be used in JDK17. I still prefer keeping consistent style.
3. Could u modify the title `improvement(core): Eliminate N+1 queries in metadata object…`? You should remove `...`.

jerryshao:
@yuqi1129 please help to review.

yuqi1129:
@babumahesh 

The same as @roryqi, some UTs are necessary.

babumahesh:
@yuqi1129 @roryqi 
I have added the unit tests and addressed the comments. Please review and let me know if any further action is needed.

yuqi1129:
Good job, let wait for the CI pass.

yuqi1129 (approved):

PR #10593, opened 2026-03-31
[#9736][followup] feat(iceberg-rest): Implement snapshots=refs filtering in loadTable

laserninja:

Implement the `snapshots` query parameter filtering for the Iceberg REST `loadTable` endpoint. When `snapshots=refs`, the response is filtered to only include snapshots directly referenced by the table's refs (branches and tags), per the Iceberg REST specification.

Follow-up to: #10536 (comment by @roryqi)

The `snapshots` query parameter was accepted but not implemented (marked with `// todo support snapshots`). Per the Iceberg REST spec, `snapshots=refs` should return only ref-referenced snapshots, reducing payload size for clients that don't need the full snapshot history.

Yes - `loadTable` with `snapshots=refs` now returns a filtered response containing only snapshots referenced by branches/tags, instead of all snapshots.

- Added `testLoadTableSnapshotsRefsFiltering`: creates a multi-snapshot table and verifies `snapshots=refs` returns fewer snapshots, all of which are ref-referenced
- Added `testLoadTableSnapshotsAllReturnsAllSnapshots`: verifies default behavior is unchanged
- All 60 existing tests pass

### DIFF
--- a/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
+++ b/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
@@ -66,6 +66,7 @@ import org.apache.gravitino.server.authorization.GravitinoAuthorizerProvider;
 import org.apache.iceberg.MetadataUpdate;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotRef;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.UpdateRequirement;
 import org.apache.iceberg.UpdateRequirements;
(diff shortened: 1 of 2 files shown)


roryqi:
@bharos Could u help me review this pull request?

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
```diff
@@ -981,4 +982,73 @@ void testLoadTableETagDiffersForSnapshotsParam(Namespace namespace) {
...
+
+    // Load with snapshots=refs: should return only ref-referenced snapshots
+    Response refsResponse = doLoadTableWithSnapshots(namespace, "snapshots_refs_foo1", "refs");
+    Assertions.assertEquals(Status.OK.getStatusCode(), refsResponse.getStatus());
+    LoadTableResponse refsTableResponse = refsResponse.readEntity(LoadTableResponse.class);
+    List<Snapshot> refsSnapshots = refsTableResponse.tableMetadata().snapshots();
+
+    Assertions.assertTrue(
+        refsSnapshots.size() < allSnapshots.size(),
+        "snapshots=refs should return fewer snapshots than snapshots=all");
```
bharos: Should we check if it is exactly equal to
referencedSnapshotIds.size()

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
```diff
@@ -981,4 +982,73 @@ void testLoadTableETagDiffersForSnapshotsParam(Namespace namespace) {
...
+
+    // Every returned snapshot should be referenced by a ref
+    for (Snapshot snapshot : refsSnapshots) {
+      Assertions.assertTrue(
+          referencedSnapshotIds.contains(snapshot.snapshotId()),
+          "Snapshot " + snapshot.snapshotId() + " should be referenced by a ref");
+    }
```
bharos: nit : Can we combine the size check and snapshot check above with

```
Assertions.assertEquals(
    referencedSnapshotIds,
    refsSnapshots.stream().map(Snapshot::snapshotId).collect(Collectors.toSet()),
    "snapshots=refs should return exactly the ref-referenced snapshots");
```

laserninja:
Done. Combined into a single 'assertEquals' on the snapshot ID sets. Thanks!

bharos:
Thanks for the PR @laserninja

Issue #10597, opened 2026-03-31
[Improvement] CreateTableFailureEvent listener exception masks original createTable failure


justinmclean:
### What would you like to be improved?

In TableEventDispatcher#createTable, when dispatcher.createTable(...) throws, the code enters the catch block and dispatches CreateTableFailureEvent. If eventBus.dispatchEvent(...) also throws, that new exception replaces the original create-table exception. This can hide the real root cause from callers and make production failures much harder to diagnose.

Relevant path: TableEventDispatcher.java (line 141)

### How should we improve?

Protect failure-event dispatch so it cannot mask the original exception. For example, wrap eventBus.dispatchEvent(new CreateTableFailureEvent(...)) in its own try/catch, preserve the original exception, and then rethrow the original failure.

Here's a unit test to help:
```

  @Test
  void testCreateTableFailureEventShouldNotMaskOriginalExceptionWhenListenerThrows() {
    NameIdentifier identifier = NameIdentifier.of("metalake", "catalog", table.name());
    GravitinoRuntimeException originalException =
        new GravitinoRuntimeException("Original create table failure");
    GravitinoRuntimeException listenerException =
        new GravitinoRuntimeException("Failure listener exception");
    TableDispatcher tableExceptionDispatcher = mock(TableDispatcher.class);
    when(tableExceptionDispatcher.createTable(
            any(NameIdentifier.class),
            any(Column[].class),
            any(String.class),
            any(Map.class),
            any(Transform[].class),
            any(Distribution.class),
            any(SortOrder[].class),
            any(Index[].class)))
        .thenThrow(originalException);

    EventBus eventBus =
        new EventBus(
            Arrays.asList(
                new EventListenerPlugin() {
                  @Override
                  public void init(Map<String, String> properties) {}

                  @Override
                  public void start() {}

                  @Override
                  public void stop() {}

                  @Override
                  public void onPostEvent(Event event) {
                    if (event instanceof CreateTableFailureEvent) {
                      throw listenerException;
                    }
                  }
                }));
    TableEventDispatcher tableEventDispatcher =
        new TableEventDispatcher(eventBus, tableExceptionDispatcher);

    GravitinoRuntimeException thrownException =
        Assertions.assertThrowsExactly(
            GravitinoRuntimeException.class,
            () ->
                tableEventDispatcher.createTable(
                    identifier,
                    table.columns(),
                    table.comment(),
                    table.properties(),
                    table.partitioning(),
                    table.distribution(),
[...]

PR #10606, opened 2026-03-31
[#10597] fix(core): Prevent CreateTableFailureEvent from masking original exception

sachinnn99:

Wrap the failure event dispatch in `TableEventDispatcher#createTable` with a nested try/catch so that a listener exception cannot replace the original `createTable` failure. The caught listener exception is logged at WARN level.

When `dispatcher.createTable()` throws and `eventBus.dispatchEvent(new CreateTableFailureEvent(...))` also throws (e.g. a listener fails), the listener exception replaces the original exception. This hides the root cause from callers and makes production failures harder to diagnose.

Fix: #10597

No. The only behavioral change is that the original exception is now always preserved and rethrown, even if a failure event listener throws.

Added a unit test (`testCreateTableFailureEventShouldNotMaskOriginalExceptionWhenListenerThrows`) in `TestTableEvent` that:
1. Mocks a `TableDispatcher` that throws on `createTable`
2. Registers a listener that throws on `CreateTableFailureEvent`
3. Asserts the original exception (not the listener's) is what surfaces

All 13 tests in `TestTableEvent` pass. Formatted with `spotlessApply`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/TableEventDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/listener/TableEventDispatcher.java
@@ -148,6 +148,7 @@ public class TableEventDispatcher implements TableDispatcher {
             new CreateTableFailureEvent(
                 PrincipalUtils.getCurrentUserName(), ident, e, createTableRequest));
       } catch (Exception ex) {
+        e.addSuppressed(ex);
         LOG.warn("Failed to dispatch CreateTableFailureEvent", ex);
       }
       throw e;


sachinnn99:
Hi @jerryshao @yuqi1129, could you please review this when you get a chance? All CI checks are passing. Thank you!

sachinnn99:
Closing — this is superseded by #10665, which fixed the same problem at the EventBus level for all failure events via `dispatchFailureEvent`.

Issue #10603, opened 2026-03-31
[Improvement] Fix Lance partition statistics writes failing due to table_id vector type mismatch


justinmclean:
### What would you like to be improved?

LancePartitionStatisticStorage defines the table_id Arrow column as a 64-bit integer, but createFragmentMetadata() retrieves it as UInt8Vector and writes a Long table ID into that vector.

Relevant locations: LancePartitionStatisticStorage.java (line 114) and LancePartitionStatisticStorage.java (line 388)

This makes the Lance-backed partition statistics update path inconsistent with its own schema and can fail at runtime when statistics are written.

### How should we improve?

Use the Arrow vector type that matches the declared schema for table_id, such as BigIntVector, instead of UInt8Vector.

Here's a unit test to help:
```
@Test
  public void testUpdateStatisticsWithLargeTableId() throws Exception {
    PartitionStatisticStorageFactory factory = new LancePartitionStatisticStorageFactory();
    String metalakeName = "metalake";
    MetadataObject metadataObject =
        MetadataObjects.of(
            Lists.newArrayList("catalog", "schema", "table"), MetadataObject.Type.TABLE);

    EntityStore entityStore = mock(EntityStore.class);
    TableEntity tableEntity = mock(TableEntity.class);
    when(entityStore.get(any(), any(), any())).thenReturn(tableEntity);
    when(tableEntity.id()).thenReturn(256L);
    FieldUtils.writeField(GravitinoEnv.getInstance(), "entityStore", entityStore, true);

    String location = Files.createTempDirectory("lance_stats_large_table_id").toString();
    Map<String, String> properties = Maps.newHashMap();
    properties.put("location", location);

    LancePartitionStatisticStorage storage =
        (LancePartitionStatisticStorage) factory.create(properties);
    try {
      Map<String, StatisticValue<?>> statistics = Maps.newHashMap();
      statistics.put("statistic0", StatisticValues.stringValue("value0"));

      storage.updateStatistics(
          metalakeName,
          Lists.newArrayList(
              MetadataObjectStatisticsUpdate.of(
                  metadataObject,
                  Lists.newArrayList(
                      PartitionStatisticsModification.update("partition0", statistics)))));

      List<PersistedPartitionStatistics> listedStats =
          storage.listStatistics(
              metalakeName,
              metadataObject,
              PartitionRange.between(
                  "partition0",
                  PartitionRange.BoundType.CLOSED,
                  "partition0",
[...]

PR #10610, opened 2026-03-31
[#10603] improvement: replaced UInt8Vector with BigIntVector

ChisomUma:
1. Title: [#10603] fix(stats): Use BigIntVector for table_id in LancePartitionStatisticStorage

This PR fixes a vector type inconsistency in [LancePartitionStatisticStorage](cci:2://file:///Users/victory/Documents/codes/strapi_issues/gravitino/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java:80:0-588:1) when updating metrics. The `table_id` Arrow column is defined as a 64-bit integer, but the implementation previously attempted to load it into a `UInt8Vector`. This PR updates the underlying logic to correctly retrieve and write the Long table IDs using a `BigIntVector` to match the declared schema.

Because of the previous inconsistency with its own Arrow schema, the Lance-backed partition statistics update path would incorrectly process table IDs and could fail at runtime when partition statistics were written, particularly when a table ID exceeded the 8-bit unsigned vector bounds. 

Fix: #10603

No user-facing changes were introduced.

A new unit test case `testUpdateStatisticsWithLargeTableId` was added to [TestLancePartitionStatisticStorage.java](cci:7://file:///Users/victory/Documents/codes/strapi_issues/gravitino/core/src/test/java/org/apache/gravitino/stats/storage/TestLancePartitionStatisticStorage.java:0:0-0:0). The test mocks a scenario where `tableEntity.id()` is returned as `256L` and reliably verifies that the table statistics are listed and written exactly as expected without throwing type conversion errors.

### DIFF
--- a/core/src/test/java/org/apache/gravitino/stats/storage/TestLancePartitionStatisticStorage.java
+++ b/core/src/test/java/org/apache/gravitino/stats/storage/TestLancePartitionStatisticStorage.java
@@ -593,6 +593,8 @@ public class TestLancePartitionStatisticStorage {
       FileUtils.deleteDirectory(new File(location + "/" + tableEntity.id() + ".lance"));
       storage.close();
     }
+  }
+
   @Test
   public void testUpdateStatisticsWithLargeTableId() throws Exception {
     PartitionStatisticStorageFactory factory = new LancePartitionStatisticStorageFactory();
--- a/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
+++ b/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
@@ -52,9 +52,9 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
-import org.apache.arrow.vector.BigIntVector;
 import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.LargeVarCharVector;
+import org.apache.arrow.vector.UInt8Vector;
 import org.apache.arrow.vector.VarCharVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.arrow.vector.ipc.ArrowReader;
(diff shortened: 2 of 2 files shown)


justinmclean:
We use Spolless to ensure the code is consistently formatted, and the CI is failing because of it. You'll need to fix the formatting error and push your changes.

ChisomUma:
Hi @roryqi thanks for pointing that out. The code was previously valid as Unit8Vector actual size is 8 bytes equivalent to 64 bits here. The only addition was for the new test case that validates that.

@justinmclean I've fixed the formatting issue.

core/src/test/java/org/apache/gravitino/stats/storage/TestLancePartitionStatisticStorage.java
```diff
@@ -594,4 +594,57 @@ public void testDropStatisticsWithQuoteInPartitionAndStatisticName() throws Exce
...
+    PartitionStatisticStorageFactory factory = new LancePartitionStatisticStorageFactory();
+    String metalakeName = "metalake";
+    MetadataObject metadataObject =
+        MetadataObjects.of(
+            Lists.newArrayList("catalog", "schema", "table"), MetadataObject.Type.TABLE);
+
+    EntityStore entityStore = mock(EntityStore.class);
+    TableEntity tableEntity = mock(TableEntity.class);
+    when(entityStore.get(any(), any(), any())).thenReturn(tableEntity);
+    when(tableEntity.id()).thenReturn(256L);
```
jerryshao, line 609: Table entity ID is long, which is already 8 bytes. And the long value table ID exists everywhere, do you need additional tests?

ChisomUma: Yeah, you are right, given that the issue in question #10603 made a mistake and I've reverted that change, the new test doesn't actually test something new.

ChisomUma: > Yeah, you are right, given that the issue in question #10603 made a mistake and I've reverted that change, the new test doesn't actually test something new.

@jerryshao

jerryshao: In this case, I'm going to close this PR. You can always work on others if you'd like to.

Issue #10611, opened 2026-03-31
[Improvement]  Add GitHub Actions CI workflow for the maintenance module


lasdf1234:
### What would you like to be improved?
The maintenance module (:maintenance:optimizer, :maintenance:updaters, :maintenance:jobs) has integration tests
under org.apache.gravitino.maintenance.optimizer.integration.test.**, but there is no GitHub Actions workflow to run
them on push or pull request. This means regressions in the maintenance module go undetected in CI.

### How should we improve?
Add two workflow files mirroring the existing flink-integration-test*.yml pattern:

 - maintenance-integration-test.yml — trigger workflow that fires on push/pull_request to main / branch-*, uses 
dorny/paths-filter to skip runs when no relevant paths changed (maintenance/**, api/**, core/**, server/**, etc.), 
and delegates to the action workflow via MaintenanceIT-on-push / MaintenanceIT-on-pr jobs.
 - maintenance-integration-test-action.yml — reusable workflow_call action that checks out, sets up JDK 17, builds 
the distribution, then runs :maintenance:optimizer:test in both embedded and deploy modes with 
PskipDockerTests=false. On failure (or when the upload log label is set), uploads logs and reports as artifacts.

PR #10612, opened 2026-03-31
[#10611] improve(ci): Add GitHub Actions integration test workflow for the maintenance module

lasdf1234:
Adds two GitHub Actions workflow files for the maintenance module, following the same pattern as the existing 
flink-integration-test*.yml:

 - .github/workflows/maintenance-integration-test.yml — trigger workflow that fires on push/pull_request to main / 
branch-*. Uses dorny/paths-filter to skip runs when no relevant files have changed. Triggers on changes to 
maintenance/** as well as shared modules (api/**, core/**, server/**, catalogs/**, etc.).
 - .github/workflows/maintenance-integration-test-action.yml — reusable workflow_call action. Checks out the repo, 
sets up JDK 17 (Temurin), builds the full distribution, frees disk space, then runs :maintenance:optimizer:test in 
both embedded and deploy modes with -PskipDockerTests=false. On failure (or when the upload log label is present), 
uploads test reports and logs as artifacts.
The maintenance module contains integration tests under 
org.apache.gravitino.maintenance.optimizer.integration.test.** but had no CI workflow to execute them. Without CI,
regressions in the maintenance module (optimizer, updaters, jobs) could go undetected on push or pull request.

Fix: https://github.com/apache/gravitino/issues/10611
No. This is a CI-only change adding new workflow files. No APIs, configuration keys, or runtime behaviour are
affected.
Workflow syntax validated locally. The action follows the same structure as the proven 
flink-integration-test-action.yml. Integration tests under :maintenance:optimizer pass when run locally with:

 ./gradlew -PskipDockerTests=false :maintenance:optimizer:test \
   --tests "org.apache.gravitino.maintenance.optimizer.integration.test.**"


.github/workflows/maintenance-integration-test.yml
```diff
@@ -0,0 +1,71 @@
...
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36
+        id: filter
+        with:
+          filters: |
+            source_changes:
+              - .github/**
+              - api/**
+              - bin/**
+              - catalogs/**
```
FANNG1: Please check `source_changes`,  seems some module changes like `catalogs` should not invoke the CI

FANNG1:
@yuqi1129 @mchades  please take a look when you have time, thanks

.github/workflows/maintenance-integration-test.yml
FANNG1: you could check the dependences for the maintenance module

lasdf1234: Got,I remove catalogs/iceberg and add catalogs/catalog-common/

Issue #10602, opened 2026-03-31
[Improvement] Fileset update can report success when versioned metadata update matches no rows


justinmclean:
### What would you like to be improved?

FilesetMetaService.updateFileset has a potentially unsafe success path when checkNeedUpdateVersion is true. In that branch, the code runs insertFilesetVersions(...) and updateFilesetMeta(...) inside one transaction, but it does not check the row count returned by updateFilesetMeta(...). Instead it unconditionally sets updateResult = 1 and returns success.

Because updateFilesetMeta(...) uses optimistic matching on the old row values, it can return 0 if the metadata row changed between read and update. In that case, the method may still report success even though the fileset metadata update did not apply.

### How should we improve?

Capture and validate the result of updateFilesetMeta(...) in the version-update branch as well. If the update affects 0 rows, treat it as an update conflict and fail the operation instead of forcing success.

One straightforward fix is to make the multi-step transaction return the metadata update count, then throw when the count is 0. 

Here is a test to help:
```

  @TestTemplate
  public void testUpdateFilesetReturnsSuccessWhenVersionedMetaUpdateAffectsNoRows()
      throws IOException {
    String filesetName = GravitinoITUtils.genRandomName("tst_fs_conflict");
    NameIdentifier filesetIdent =
        NameIdentifier.of(metalakeName, catalogName, schemaName, filesetName);
    FilesetEntity filesetEntity =
        createFilesetEntity(
            RandomIdGenerator.INSTANCE.nextId(),
            NamespaceUtil.ofFileset(metalakeName, catalogName, schemaName),
            filesetName,
            AUDIT_INFO,
            "/tmp");
    FilesetMetaService.getInstance().insertFileset(filesetEntity, true);

    AuditInfo conflictingAuditInfo =
        AuditInfo.builder()
            .withCreator("conflicting-updater")
            .withCreateTime(Instant.now())
            .build();
    FilesetEntity updatedFilesetEntity =
        FilesetEntity.builder()
            .withId(filesetEntity.id())
            .withName(filesetEntity.name())
            .withNamespace(filesetEntity.namespace())
            .withFilesetType(filesetEntity.filesetType())
            .withStorageLocations(ImmutableMap.of(LOCATION_NAME_UNKNOWN, "/tmp-v2"))
            .withComment("comment-v2")
            .withProperties(ImmutableMap.of("version", "2"))
            .withAuditInfo(
                AuditInfo.builder()
                    .withCreator("expected-updater")
[...]

PR #10613, opened 2026-03-31
[#10602] improvemet:check update count in FilesetMetaService.updateFileset

Victory-ET:
1. Title: [#10602] fix(core): Check update count in FilesetMetaService.updateFileset

This PR fixes a bug in `FilesetMetaService.updateFileset` where a potentially unsafe success path occurs when `checkNeedUpdateVersion` evaluates to true. We now explicitly capture the row count result from `updateFilesetMeta(...)` by changing the underlying session utility logic to `getWithoutCommit` instead of a void execution. If the updated row count is 0, we properly treat it as an update conflict and fail the operation.

Because `updateFilesetMeta(...)` relies on optimistic matching on old row values, it could previously return 0 if the metadata row changed between reads and updates. However, it would force an unconditional success without checking this row count. This could lead to data inconsistencies and masked write failures where operations reported success despite the fileset metadata update not being applied.

Fix: #(10602)

No user-facing APIs were altered.

A new unit test `testUpdateFilesetReturnsSuccessWhenVersionedMetaUpdateAffectsNoRows` was added to `TestFilesetMetaService.java` to simulate locking conflict anomalies and assert that a proper runtime IO error exception is thrown instead of masking it as a silent success.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
@@ -240,7 +240,7 @@ public class FilesetMetaService {
           if (metaUpdateCountRef[0] == 0) {
             // The meta update matched no rows; the transaction was rolled back,
             // including the version insert above.
-            updateResult = 0;
+            throw new IOException("Failed to update the entity: " + identifier);
           } else {
             ExceptionUtils.checkSQLException(
                 re, Entity.EntityType.FILESET, newEntity.nameIdentifier().toString());


justinmclean:
We use Spotless to ensure the code is consistently formatted, but the CI is failing because of it. The issue is in TestFilesetMetaService.java and looks like a spacing/bracket issue?. You need to fix that and run ` ./gradlew spotlessApply` to check everything is OK and push your changes to fix this.

core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
```diff
@@ -220,17 +220,17 @@ public <E extends Entity & HasIdentifier> FilesetEntity updateFileset(
...
                     FilesetVersionMapper.class,
                     mapper -> mapper.insertFilesetVersions(newFilesetPO.getFilesetVersionPOs())),
             () ->
-                SessionUtils.doWithoutCommit(
+                updateResultRef[0] = SessionUtils.getWithoutCommit(
                     FilesetMetaMapper.class,
                     mapper -> mapper.updateFilesetMeta(newFilesetPO, oldFilesetPO)));
-        // we set the updateResult to 1 to indicate that the update is successful
-        updateResult = 1;
+        updateResult = updateResultRef[0];
```
jerryshao: Changing to this actually helps nothing. If the update fails (update nothing), we should rollback the above insert.

Victory-ET: done

core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
```diff
@@ -215,22 +215,38 @@ public <E extends Entity & HasIdentifier> FilesetEntity updateFileset(
...
+                if (metaUpdateCountRef[0] == 0) {
+                  throw new RuntimeException("Failed to update the entity: " + identifier);
+                }
+              });
+          updateResult = 1;
+        } catch (RuntimeException re) {
+          if (metaUpdateCountRef[0] == 0) {
+            // The meta update matched no rows; the transaction was rolled back,
+            // including the version insert above.
+            updateResult = 0;
```
yuqi1129: What's the difference between 0 and 1 for updateResult if there exist exceptions for this method?

Victory-ET: thanks for pointing that out, that was uneccessary since we only ever set to 1 in the success path, and the exception is thrown in the case of zero regardless. I've taken that out.

yuqi1129: OK, I have no more comments.

PR #10616, opened 2026-03-31
[#10601] fix: rollback catalog properties on partial failure (#10601)

### LINKED ISSUE #10601: [Improvement] Enabling a metalake can leave catalog metalake-in-use flags inconsistent on partial failure

Enabling or disabling a metalake is not atomic. In MetalakeManager.enableMetalake and the corresponding disable path, Gravitino first updates the metalake-level-in-use flag in storage, then updates the propagated metalake-in-use property on each catalog.

If one of the catalog updates fails, the API returns an error, but the metalake-level flag has already been committed. This leaves the system in a partially updated state, where the metalake is marked as enabled/disabled, while some catalogs still have the old metalake-in-use value.

### How should we improve?

Make metalake enable/disable behave transactionally from the caller’s perspective. A few workable options are:
- Update all catalogs first and only persist the metalake-level flag if every catalog update succeeds.
- Add rollback logic so a failed catalog propagation restores the metalake-level flag and any catalogs already updated.

If full atomicity is not feasible, fail the operation before updating the metalake state unless all catalog updates succeed.

Here's a unit test to help:
```
@Test
  public void testEnableMetalakeShouldRollbackWhenCatalogPropagationFails()
      throws IllegalAccessException, IOException {
    NameIdentifier ident = NameIdentifier.of("partialEnableMetalake");
    CatalogManager originalCatalogManager =
        (CatalogManager) FieldUtils.readField(GravitinoEnv.getInstance(), "catalogManager", true);
    CatalogManager mockCatalogManager = Mockito.mock(CatalogManager.class);

    metalakeManager.createMetalake(ident, "comment", ImmutableMap.of());
    entityStore.put(createCatalogEntity(100L, ident.name(), "catalog1"));
    entityStore.put(createCatalogEntity(101L, ident.name(), "catalog2"));

    entityStore.update(
        ident,
        BaseMetalake.class,
        EntityType.METALAKE,
        metalake -> {
          HashMap<String, String> properties = new HashMap<>(metalake.properties());
          properties.put(Metalake.PROPERTY_IN_USE, "false");
          BaseMetalake.Builder builder =
              BaseMetalake.builder()
                  .withId(metalake.id())
                  .withName(metalake.name())
                  .withComment(metalake.comment())
                  .withProperties(properties)
                  .withVersion(metalake.getVersion())
                  .withAuditInfo(metalake.auditInfo());
          return builder.build();
        });

    doThrow(new RuntimeException("catalog update failed"))
        .when(mockCatalogManager)
        .setMetalakeInUseStatus(any(NameIdentifier.class), anyBoolean());

    FieldUtils.writeField(GravitinoEnv.getInstance(), "catalogManager", mockCatalogManager, true);

    try {
      Assertions.assertThrows(RuntimeException.class, () -> metalakeManager.enableMetalake(ident));
      Assertions.assertFalse(
          MetalakeManager.metalakeInUse(entityStore, ident),
          "Metalake should remain disabled when catalog propagation fails");
      verify(mockCatalogManager, times(2))
          .setMetalakeInUseStatus(any(NameIdentifier.class), anyBoolean());
    } finally {
      FieldUtils.writeField(
          GravitinoEnv.getInstance(), "catalogManager", originalCatalogManager, true);
      metalakeManager.dropMetalake(ident, true);
    }
  }

  private CatalogEntity createCatalogEntity(Long id, String metalake, String catalog) {
    AuditInfo auditInfo =
        AuditInfo.builder()
            .withCreator(AuthConstants.ANONYMOUS_USER)
            .withCreateTime(Instant.now())
            .withLastModifier(AuthConstants.ANONYMOUS_USER)
            .withLastModifiedTime(Instant.now())
            .build();

    return CatalogEntity.builder()
        .withId(id)
        .withName(catalog)
        .withNamespace(Namespace.of(metalake))
        .withType(Catalog.Type.RELATIONAL)
        .withProvider("test")
        .withComment("")
        .withProperties(ImmutableMap.of())
        .withAuditInfo(auditInfo)
        .build();
  }
```

Mykal-Steele:
[#10601] fix(core): add catalog rollback on partial metalake failure

I added a rollback step in `MetalakeManager.enableMetalake()` and `disableMetalake()`. Now if the catalog updates throw err halfway through, we catch the exception and call `updateMetalakeInUseStatusInCatalog` again with the opposite boolean value to undo what we already changed. To we don't leave things in a half-broken state

Right now, if you enable or disable a metalake and one of the catalog updates fails, the main storage rolls back fine, but the catalogs that already got updated are left with the wrong `metalake-in-use` status. So you end up with an inconsistent state. This fixes it so the catalogs roll back too

Fixes #10601

Nope

Added a new unit test called `testEnableMetalakeShouldRollbackWhenCatalogPropagationFails` in `TestMetalakeManager.java`. Same test as the one provided in the issue, but added explicit verify checks to make sure `setMetalakeInUseStatus` is called with the correct boolean flags.

### DIFF
--- a/docs/flink-connector/flink-connector.md
+++ b/docs/flink-connector/flink-connector.md
@@ -35,7 +35,19 @@ This capability allows users to perform federation queries, accessing data from 
 | table.catalog-store.kind                         | string | generic_in_memory | The Catalog Store name, it should set to `gravitino`.                | Yes      | 0.6.0-incubating |
 | table.catalog-store.gravitino.gravitino.metalake | string | (none)            | The metalake name that flink connector used to request to Gravitino. | Yes      | 0.6.0-incubating |
 | table.catalog-store.gravitino.gravitino.uri      | string | (none)            | The uri of Gravitino server address.                                 | Yes      | 0.6.0-incubating |
+| table.catalog-store.gravitino.gravitino.enableSessionCatalogSupport | boolean | false | Whether to enable support for Flink's session catalog in the Gravitino catalog store. | No | 1.3.0 |
 | table.catalog-store.gravitino.gravitino.client.  | string | (none)            | The configuration key prefix for the Gravitino client config.        | No       | 1.0.0            |
+
+When `table.catalog-store.gravitino.gravitino.enableSessionCatalogSupport` is set to `true`, Gravitino uses `GravitinoSessionCatalogStore`, which combines a `GravitinoCatalogStore` (backed by the Gravitino server) with an in-memory store to support Flink's session catalog. When `false` (the default), only `GravitinoCatalogStore` is used.
+
+When session catalog support is enabled, the following behaviors apply:
+
+- **CREATE CATALOG**: Gravitino-managed catalogs (e.g. `gravitino-hive`, `gravitino-iceberg`) are persisted to the Gravitino server; non-Gravitino-managed catalogs (e.g. `hive`, `jdbc`, or any custom connector type) are stored in the in-memory store only.
+- **GET / USE CATALOG**: The in-memory store is checked first. If the catalog is not found there, it is retrieved from the Gravitino server.
+- **DROP CATALOG**: The in-memory store is checked first. If the catalog exists there it is removed from memory; otherwise it is removed from the Gravitino server.
+- **SHOW / LIST CATALOGS**: Returns the combined set of catalogs from both the in-memory store and the Gravitino server.
+- **Session scope**: Catalogs stored only in memory are session-scoped and will not survive when Flink restarts.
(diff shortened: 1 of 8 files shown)


jerryshao:
This also cannot guarantee atomicity and transactionality, but makes things complicated.

Mykal-Steele:
@jerryshao, any suggestion?

jerryshao:
I don't have a better solution. This is a hard thing to deal with. At least, I know that KISS is always what I want. Making the code logic complicated and only fixing some very rare cases is not what I wanted.

Issue #10598, opened 2026-03-31
[Improvement] Close cached Lance datasets before closing allocator in LancePartitionStatisticStorage


justinmclean:
### What would you like to be improved?

LancePartitionStatisticStorage.close() closes the Arrow allocator before invalidating datasetCache in LancePartitionStatisticStorage.java (line 352). When dataset caching is enabled, cached Dataset instances remain open during normal operations and are expected to be released during cache invalidation on shutdown. Closing the allocator first can make cleanup happen in the wrong order, which may cause shutdown-time exceptions or leave cached dataset resources unclosed.

### How should we improve?

Release cached datasets before closing the allocator. A simple fix would be to invalidate or explicitly close all cached Dataset instances first, then close the allocator, and finally shut down the scheduler.

Here's a unit test to help:
```

  @Test
  public void testCloseReleasesCachedDatasetBeforeAllocator() throws Exception {
    String location = Files.createTempDirectory("lance_stats_close_cache").toString();
    Map<String, String> properties = Maps.newHashMap();
    properties.put("location", location);
    properties.put("datasetCacheSize", "10");

    LancePartitionStatisticStorage storage = new LancePartitionStatisticStorage(properties);
    try {
      BufferAllocator allocator = mock(BufferAllocator.class);
      FieldUtils.writeField(storage, "allocator", allocator, true);

      Cache<Long, Dataset> datasetCache = storage.getDatasetCache();
      Assertions.assertNotNull(datasetCache);

      Dataset dataset = mock(Dataset.class);
      datasetCache.put(1L, dataset);

      storage.close();

      InOrder inOrder = inOrder(dataset, allocator);
      inOrder.verify(dataset).close();
      inOrder.verify(allocator).close();
    } finally {
      FileUtils.deleteDirectory(new File(location));
    }
  }
```

pythaac:
Hi, I'm working on this issue🙂

pythaac:
Since the cached Dataset internally uses Arrow off-heap memory, I think we need to explicitly close the datasets to prevent memory leaks. I have updated the logic and added a test case to verify this behavior. Please review and let me know if I missed anything😊

PR #10619, opened 2026-03-31
[#10598] fix(core): close cached Lance datasets before closing Arrow allocator

pythaac:

1. Modified `LancePartitionStatisticStorage.close()` to explicitly iterate over all cached `Dataset` instances and call `.close()` before calling `cache.invalidateAll()` and `allocator.close()`.
2. Updated `testCloseReleasesCachedDatasetBeforeAllocator` to verify both the execution order (`InOrder`) and the actual release of native Arrow memory (`getAllocatedMemory() == 0`).

The original issue correctly states that closing the `allocator` before the `Dataset` causes a shutdown exception.   

However, simply swapping the closing order is not enough. If we rely solely on `invalidateAll()`, `dataset.close()` is never called, silently leaking native Arrow memory and still causing `allocator.close()` to crash (`IllegalStateException: Allocator closed with outstanding buffers allocated`).   

Therefore, explicitly closing the datasets in a loop is strictly required to safely release off-heap resources.

Fix: #10598

No.

- Executed `testCloseReleasesCachedDatasetBeforeAllocator` using Mockito `spy()` with a real `RootAllocator` and `VarCharVector`.
- Verified that `Dataset.close()` strictly happens before `allocator.close()`.
- Asserted that `allocator.getAllocatedMemory()` successfully drops to `0 bytes` after `storage.close()`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
+++ b/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
@@ -76,6 +76,8 @@ import org.apache.gravitino.stats.PartitionStatisticsUpdate;
 import org.apache.gravitino.stats.StatisticValue;
 import org.apache.gravitino.utils.MetadataObjectUtil;
 import org.apache.gravitino.utils.PrincipalUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** LancePartitionStatisticStorage is based on Lance format files. */
 public class LancePartitionStatisticStorage implements PartitionStatisticStorage {
@@ -130,6 +132,8 @@ public class LancePartitionStatisticStorage implements PartitionStatisticStorage
   private final ScheduledThreadPoolExecutor scheduler;
 
   private final EntityStore entityStore = GravitinoEnv.getInstance().entityStore();
+
+  private static final Logger LOG = LoggerFactory.getLogger(LancePartitionStatisticStorage.class);
 
   public LancePartitionStatisticStorage(Map<String, String> properties) {
     this.allocator = new RootAllocator();
@@ -350,19 +354,17 @@ public class LancePartitionStatisticStorage implements PartitionStatisticStorage
 
   @Override
   public void close() throws IOException {
-    datasetCache.ifPresent(
-        cache -> {
-          cache
-              .asMap()
-              .values()
-              .forEach(
-                  dataset -> {
-                    if (dataset != null) {
-                      dataset.close();
-                    }
-                  });
-          cache.invalidateAll();
-        });
+    if (datasetCache.isPresent()) {
+      Cache<Long, Dataset> cache = datasetCache.get();
+      for (Dataset dataset : cache.asMap().values()) {
+        try {
+          dataset.close();
+        } catch (Exception e) {
+          LOG.warn("Failed to close cached Lance dataset", e);
+        }
+      }
+      cache.invalidateAll();
+    }
 
     if (allocator != null) {
       allocator.close();


core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -350,12 +350,24 @@ private void dropStatisticsImpl(Long tableId, List<PartitionStatisticsDrop> drop
...
+                    if (dataset != null) {
+                      dataset.close();
+                    }
+                  });
+          cache.invalidateAll();
+        });
+
     if (allocator != null) {
       allocator.close();
     }
```
justinmclean, line 372: This works, but for resource cleanup I’d suggest keeping it more straightforward. The combination of Optional.ifPresent, lambdas, and forEach makes it a bit harder to follow than a simple loop.

In close() methods, clarity and predictability are usually more important than conciseness.

Also worth considering: if one dataset.close() throws, the remaining datasets (and allocator) won’t be closed. We may want to handle that more defensively.

jerryshao:
@roryqi please help to review this PR.

roryqi:
Why does the order matter? I am ok with the current order.

justinmclean:
The reason the order matters is as @pythaac described in his PR/unit test. The test uses a real RootAllocator and VarCharVector, and the reported failure is that if the cached Dataset is not explicitly closed before allocator.close(), allocator shutdown can fail with IllegalStateException: Allocator closed with outstanding buffers allocated.

So this is not only about keeping a preferred order,  invalidateAll() alone does not release the dataset resources, and that explicit dataset.close() before allocator.close() is required to avoid the shutdown failure.

justinmclean:
Arrow’s Java examples use try-with-resources for both Dataset and BufferAllocator, implying they are co-managed resources and are closed in reverse order (dataset before allocator). See https://arrow.apache.org/java/18.2.0/dataset.html

There are also upstream Arrow issues showing that if resources backed by the allocator aren’t released, the allocator can fail to close cleanly or retain memory. e.g. https://github.com/apache/arrow-java/issues/1037

roryqi:
> Arrow’s Java examples use try-with-resources for both Dataset and BufferAllocator, implying they are co-managed resources and are closed in reverse order (dataset before allocator). See https://arrow.apache.org/java/18.2.0/dataset.html
> 
> There are also upstream Arrow issues showing that if resources backed by the allocator aren’t released, the allocator can fail to close cleanly or retain memory. e.g. [apache/arrow-java#1037](https://github.com/apache/arrow-java/issues/1037)

OK to change the order.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
pythaac: Thank you for your suggestion! I've updated the loop to improve safety against exceptions and the code clarity.

pythaac:
For your information, already mentioned in earlier comments, both of the following cases result in an exception during `close()` in my test:
1. The allocator is closed before the datasets
2. The datasets are only invalidated (without being explicitly closed) before the allocator closes

```
TestLancePartitionStatisticStorage > testCloseReleasesCachedDatasetBeforeAllocator() FAILED
    java.lang.IllegalStateException: Memory was leaked by query. Memory leaked: (40960)
    Allocator(ROOT) 0/40960/40960/9223372036854775807 (res/actual/peak/limit)
        at org.apache.arrow.memory.BaseAllocator.close(BaseAllocator.java:477)
        at org.apache.arrow.memory.RootAllocator.close(RootAllocator.java:29)
        at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:732)
        at org.apache.arrow.memory.BaseAllocator.close(BaseAllocator.java:415)
        at org.apache.arrow.memory.RootAllocator.close(RootAllocator.java:29)
        at org.apache.gravitino.stats.storage.LancePartitionStatisticStorage.close(LancePartitionStatisticStorage.java:358)
        at org.apache.gravitino.stats.storage.TestLancePartitionStatisticStorage.testCloseReleasesCachedDatasetBeforeAllocator(TestLancePartitionStatisticStorage.java:643)
```

These results demonstrate why we must both change the order and explicitly invoke `dataset.close()`. 🙂

Issue #10623, opened 2026-04-01
[Improvement] Fix cleanup ordering to prevent orphaned staging directories on deletion failure


justinmclean:
### What would you like to be improved?

JobManager.cleanUpStagingDirs() deletes the persisted job record before deleting that job’s staging directory. If FileUtils.deleteDirectory(...) then fails with an IOException, the code only logs the error and continues. Because later cleanup runs discover candidates from entityStore.list(...) rather than by scanning the filesystem, that staging directory is no longer associated with any stored job and may be left behind indefinitely.

This code path is production-reachable because cleanUpStagingDirs() is scheduled automatically by JobManager during normal runtime.

### How should we improve?

Reverse the cleanup order so the staging directory is deleted first, and only delete the job entity after filesystem cleanup succeeds. That keeps failed deletions retryable on the next scheduled cleanup run.

If deleting the entity first is required for another reason, an alternative is to add a fallback cleanup path that scans the staging directory tree and removes orphaned job directories even when the corresponding job entity is already gone.

Roshan1299:
Hi @justinmclean , I'd like to work on this! Can I be assigned?

justinmclean:
Sure, @Roshan1299, if you have any questions, just ask.

Roshan1299:
Hi @justinmclean , I've opened a PR for this: https://github.com/apache/gravitino/pull/10627 . Please let me know if any changes are needed!

PR #10627, opened 2026-04-01
[#10623] fix(core): reverse cleanup ordering in JobManager.cleanUpStagingDirs

Roshan1299:

Reversed the cleanup order in `JobManager.cleanUpStagingDirs()` so the staging directory is deleted before the job entity record. Added a unit test to verify that when directory deletion fails, the entity record is preserved. 

The previous order deleted the job entity first, then the staging directory. If `FileUtils.deleteDirectory()` failed with an `IOException`, the error was only logged. Since subsequent cleanup runs discover candidates via `entityStore.list()`, the orphaned staging directory became invisible and was left behind indefinitely.

Fix: #10623

No

Added `testCleanUpStagingDirsDeletesDirectoryBeforeEntity` which mocks `FileUtils.deleteDirectory` to throw `IOException` and verifies that `entityStore.delete` is never called. Existing `testCleanUpStagingDirs` also continues to pass.


core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -656,6 +653,9 @@ void cleanUpStagingDirs() {
                 FileUtils.deleteDirectory(jobStagingDir);
                 LOG.info("Deleted job staging directory {} for job {}", jobStagingPath, job.name());
```
justinmclean: Need to move this down to after the  entityStore.delete

Roshan1299: Good catch, moved the log statement to after `entityStore.delete()` so it only logs when both operations succeed. Updated.

justinmclean (approved):
Thanks for this

jerryshao:
I'm going to revert this PR, this introduces the flaky test, please see https://github.com/apache/gravitino/issues/10728.

Issue #10625, opened 2026-04-01
[Improvement] CustomStatistic.value() throws NPE on null instead of returning empty Optional


justinmclean:
### What would you like to be improved?

CustomStatistic.value() in StatisticManager.java (line 387) returns Optional.of(value). If a statistic entry has a null value, this throws a NullPointerException while reading partition statistics. That turns a missing statistic value into an unexpected runtime failure instead of an empty optional.

### How should we improve?

Change the implementation to return Optional.ofNullable(value) so null-backed statistics are handled safely. It would also be worth adding a unit test covering a statistic with a null value to prevent regressions.

Here's soem tests to help:
```
 @Test
  public void testCustomStatisticReturnsEmptyOptionalForNullValue() {
    StatisticManager.CustomStatistic statistic =
        new StatisticManager.CustomStatistic(
            "custom_test_stat", null, Mockito.mock(AuditInfo.class));

    Assertions.assertTrue(
        statistic.value().isEmpty(), "Null statistic values should be exposed as empty optional.");
  }

  @Test
  public void testPartitionStatisticNullValueReturnsEmptyOptional() {
    StatisticManager statisticManager = new StatisticManager(entityStore, idGenerator, config);

    MetadataObject tableObject =
        MetadataObjects.of(Lists.newArrayList(CATALOG, SCHEMA, TABLE), MetadataObject.Type.TABLE);
    Map<String, StatisticValue<?>> stats = Maps.newHashMap();
    stats.put("custom_null_stat", null);

    List<PartitionStatisticsUpdate> partitionStatistics = Lists.newArrayList();
    partitionStatistics.add(PartitionStatisticsModification.update("p_null", stats));
    statisticManager.updatePartitionStatistics(METALAKE, tableObject, partitionStatistics);

    List<PartitionStatistics> statistics =
        statisticManager.listPartitionStatistics(
            METALAKE,
            tableObject,
            PartitionRange.between(
                "p_null",
                PartitionRange.BoundType.CLOSED,
                "p_null_next",
                PartitionRange.BoundType.OPEN));

    Assertions.assertEquals(1, statistics.size());
    Assertions.assertEquals(1, statistics.get(0).statistics().length);
    Assertions.assertEquals("custom_null_stat", statistics.get(0).statistics()[0].name());
    Assertions.assertTrue(
        statistics.get(0).statistics()[0].value().isEmpty(),
        "Null persisted statistic values should be exposed as empty optional.");
  }
```

[...]

PR #10631, opened 2026-04-01
[#10625] fix(core): Use Optional.ofNullable in CustomStatistic.value() to prevent NPE

sachinnn99:

Changed `Optional.of(value)` to `Optional.ofNullable(value)` in `CustomStatistic.value()` and added a unit test covering the null value case.

`CustomStatistic.value()` throws `NullPointerException` when value is null, instead of returning an empty `Optional` as specified by the `Statistic` interface contract. `StatisticDTO` already correctly uses `Optional.ofNullable`.

Fix: #10625

No.

Added `testCustomStatisticWithNullValue()` in `TestStatisticManager` that constructs a `CustomStatistic` with a null value and verifies no NPE is thrown and `Optional.empty()` is returned.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/stats/StatisticManager.java
+++ b/core/src/main/java/org/apache/gravitino/stats/StatisticManager.java
@@ -384,7 +384,7 @@ public class StatisticManager implements Closeable, StatisticDispatcher {
 
     @Override
     public Optional<StatisticValue<?>> value() {
-      return Optional.of(value);
+      return Optional.ofNullable(value);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


roryqi:
We shouldn't pass null value for CustomStatistic.

sachinnn99:
Thanks for the feedback! Removed the `@Nullable` annotation. Kept `Optional.ofNullable` as the `Statistic` interface specifies value is optional.

sachinnn99:
Thanks for merging, @justinmclean\!

Issue #10628, opened 2026-04-01
[Improvement] dropFunction logs success even when no function is deleted


justinmclean:
### What would you like to be improved?

The REST DELETE path for functions logs a successful drop even when dispatcher.dropFunction(ident) returns false.

In FunctionOperations.java (line 245), the code correctly detects the non-delete case and emits a warning and then unconditionally logs "Function dropped".

### How should we improve?

Only emit the success log when dropped is true.

sachinnn99:
I'd like to work on this issue. Please assign it to me.

PR #10632, opened 2026-04-01
[#10628] fix(server): Only log success message when dropFunction actually drops

sachinnn99:

Moved the success `LOG.info("Function dropped: ...")` inside the `else` branch of the `dropped` check, so it only fires when the function was actually dropped.

`FunctionOperations.dropFunction()` unconditionally logs "Function dropped" even when `dispatcher.dropFunction()` returns false. This produces misleading log output.

Fix: #10628

No.

Existing `TestFunctionOperations.testDropFunction()` covers both `dropped=true` and `dropped=false` paths. The fix follows the same pattern applied to `CatalogOperations` in PR #8351.


sachinnn99:
Thanks @bharos and @justinmclean\!

PR #10636, opened 2026-04-01
[#10599] Improvement(core): fix ListModelFailureEvent uses the schema name as the event identifier name on list failures

### LINKED ISSUE #10599: [Improvement] ListModelFailureEvent uses the schema name as the event identifier name on list failures

ListModelFailureEvent builds its inherited identifier() incorrectly in ListModelFailureEvent.java (line 42)

It currently uses NameIdentifier.of(namespace.levels()), which turns the last namespace segment into the identifier name. For a namespace like metalake.catalog.schema, the event identifier becomes equivalent to model schema in namespace metalake.catalog, instead of representing the schema namespace being listed.

This can produce incorrect metadata for listeners and audit logging that read event.identifier() on failed listModels(...) operations.

### How should we improve?

Construct the event identifier so it truly represents the namespace being listed, consistent with the BaseEvent.identifier() contract for list operations.

A simple fix would be to explicitly build a namespace-style identifier for the list target and add a unit test that validates both namespace() and identifier().

Here the test to help:
```
@Test
  void testListModelFailureEventIdentifierDoesNotTreatNamespaceLeafAsModelName() {
    Assertions.assertThrowsExactly(
        GravitinoRuntimeException.class, () -> failureDispatcher.listModels(namespace));

    Event event = dummyEventListener.popPostEvent();
    Assertions.assertEquals(ListModelFailureEvent.class, event.getClass());

    ListModelFailureEvent listModelFailureEvent = (ListModelFailureEvent) event;
    Assertions.assertEquals(namespace, listModelFailureEvent.namespace());
    Assertions.assertNotEquals(namespace.level(2), listModelFailureEvent.identifier().name());
  }
```

Abyss-lord:

fix ListModelFailureEvent uses the schema name as the event identifier name on list failures

Fix: #10599 

fix ListModelFailureEvent constructor logic.

local test

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/api/event/ListModelFailureEvent.java
+++ b/core/src/main/java/org/apache/gravitino/listener/api/event/ListModelFailureEvent.java
@@ -29,6 +29,8 @@ import org.apache.gravitino.annotation.DeveloperApi;
  */
 @DeveloperApi
 public class ListModelFailureEvent extends ModelFailureEvent {
+  private static final String LIST_MODEL_NAME = "__list_models__";
+
   private final Namespace namespace;
 
   /**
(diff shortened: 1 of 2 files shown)


justinmclean (approved):
Thanks for this

PR #10637, opened 2026-04-01
[#10445] improvement(lance): Upgrade lance-core to v2.0.1 and lance-namespace-core to v0.4.5

### LINKED ISSUE #10445: [Improvement] Align Gravitino Lance REST module with latest lance-namespace and ensure backward compatibility
### **Description:**

To leverage the **Time Travel** features of `lance-spark`, we are upgrading `lance-spark` to the latest version (which depends on `lance-namespace 0.5.2`). Based on initial integration tests and community discussions, the current Gravitino Lance REST service (based on version `0.0.20`) has protocol mismatches with newer clients.

### **Problem Context:**

1. **Missing Table Declaration Endpoint**: The client now requires the `/declare` endpoint during the table creation flow. According to the Lance official documentation:
    > **"Declares a new Lance table, reserving the table name and location without creating actual data files."**
    > The absence of this route in Gravitino causes a `404 Not Found` error during table initialization.

2. **Protocol Schema Mismatch**: Newer versions of the protocol have introduced additional fields (e.g., `context` in `DescribeTableRequest`). The current server-side deserialization fails with a `400 Bad Request` when encountering these unrecognized fields:
```
Caused by: java.lang.RuntimeException: Namespace operation failed: Namespace error: Response error: status=400 Bad Request, content=Unrecognized field "context" (class com.lancedb.lance.namespace.model.DescribeTableRequest), not marked as ignorable (2 known properties: "version", "id"])
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 13] (through reference chain: com.lancedb.lance.namespace.model.DescribeTableRequest["context"]), /workspace/rust/lance-namespace-impls/src/rest.rs:577:27
	at org.lance.namespace.RestNamespace.describeTableNative(Native Method)
	at org.lance.namespace.RestNamespace.describeTable(RestNamespace.java:180)
	at org.lance.spark.BaseLanceNamespaceSparkCatalog.loadTableInternal(BaseLanceNamespaceSparkCatalog.java:1086)
	... 85 more
```

### How should we improve?

1. **Detailed Gap Analysis**: Conduct a thorough code-level audit of the `lance-namespace` protocol evolution from `0.0.20` to `0.5.x`(and `0.6.x`) to map out all interface and model changes.
2. **API & Protocol Alignment**:
    * Implement missing REST endpoints to support the full table lifecycle (e.g. `/declare`).
    * Update protocol models to align with the parameter changes in existing APIs.

3. **Backward Compatibility**: Ensure that the upgraded REST service remains fully compatible with legacy `0.0.20` clients.

bbiiaaoo:

- Upgraded Lance dependencies to the new coordinates and versions:
  - `org.lance:lance-core:2.0.1`
  - `org.lance:lance-namespace-core:0.4.5`
- Migrated Lance-related imports and model usages from `com.lancedb.*` to `org.lance.*`.
- Adapted code paths to upstream API changes, including mode handling (`Enum` -> `String`) and updated builder-style APIs.
- Updated Lance namespace/table operations, exception mapping, and serialization compatibility in Gravitino Lance modules.
- Added internal compatibility utilities (`ObjectIdentifier`, `PageUtil`, `CommonUtil`, `JsonArrowSchemaConverter`) to align with the new Lance namespace SDK behavior.
- Updated Lance REST service docs and examples to match new dependencies and request fields.
- Updated related unit/integration tests to cover the new APIs.

- Lance upstream migrated artifacts/packages from `com.lancedb` to `org.lance` and introduced breaking API changes in newer versions.
- Without this upgrade, Gravitino Lance modules may face dependency incompatibility and runtime/API mismatch issues.
- This patch keeps Gravitino Lance integrations compatible with the latest Lance ecosystem while preserving expected REST behaviors.

Fix: #10445

- Documentation/examples are updated to use `org.lance` dependencies and the new request style (for example, lowercase mode values such as `create`).
- No new Gravitino public API is introduced.

- Ran unit tests and integration tests for `catalog-lakehouse-generic`, `lance-common`, and `lance-rest-server`.
- Completed end-to-end verification with `curl -> Gravitino` requests, covering Lance namespace and table workflows on the upgraded dependencies.
- This round focused on the impacted modules; the full-repo test matrix was not executed.

### DIFF
--- a/docs/lance-rest-integration.md
+++ b/docs/lance-rest-integration.md
@@ -22,7 +22,8 @@ The following table outlines the tested compatibility between Gravitino versions
 
 | Gravitino Version (Lance REST) | Supported lance-spark Versions | Supported lance-ray Versions |
 |--------------------------------|--------------------------------|------------------------------|
-| 1.3.0                          | 0.1.0 – 0.2.0                 | 0.0.6 – 0.2.0               |
+| 1.1.1 - 1.2.1                  | 0.0.10 - 0.0.15               | 0.0.6 - 0.0.8                |
+| 1.3.0                          | 0.1.0 - 0.2.0                 | 0.0.6 - 0.2.0                |
 
 :::note
 - These version ranges show which versions are expected to work together.
--- a/docs/lance-rest-integration.md
+++ b/docs/lance-rest-integration.md
@@ -22,7 +22,7 @@ The following table outlines the tested compatibility between Gravitino versions
 
 | Gravitino Version (Lance REST) | Supported lance-spark Versions | Supported lance-ray Versions |
 |--------------------------------|--------------------------------|------------------------------|
-| 1.1.1                          | 0.0.10 – 0.0.15                | 0.0.6 – 0.0.8                |
+| 1.3.0                          | 0.1.0 – 0.2.0                 | 0.0.6 – 0.2.0               |
 
 :::note
 - These version ranges show which versions are expected to work together.
@@ -158,6 +158,7 @@ pip install lance-ray
 :::info
 - Ray will be automatically installed if not already present
 - lance-ray is currently tested with Ray versions 2.41.0 to 2.50.0
+- The lance-namespace version must be less than or equal to 0.4.5.
 - Ensure Ray version compatibility in your environment before deployment
 :::
 
--- a/docs/lance-rest-integration.md
+++ b/docs/lance-rest-integration.md
@@ -158,7 +158,6 @@ pip install lance-ray
 
 :::info
 - Ray will be automatically installed if not already present
-- lance-ray is currently tested with Ray versions 2.41.0 to 2.50.0
 - The lance-namespace version must be less than or equal to 0.4.5.
 - Ensure Ray version compatibility in your environment before deployment
 :::
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -279,10 +279,10 @@ curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table
 # Create a new empty table
 curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table02/create-empty \
   -H 'Content-Type: application/json' \
+  -H "x-lance-table-properties: {\"description\":\"This is table02\"}" \
   -d '{
     "id": ["lance_catalog", "schema", "table02"],
-    "location": "/tmp/lance_catalog/schema/table02",
-    "properties": { "description": "This is table02"  }
+    "location": "/tmp/lance_catalog/schema/table02"
   }'  
   
 # Create a table with schema, the schema is inferred from the Arrow IPC file
@@ -302,6 +302,9 @@ curl -X POST \
 
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
+import org.lance.namespace.client.apache.ApiClient;
+import org.lance.namespace.client.apache.api.TableApi;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -309,8 +312,8 @@ import java.util.Map;
 private final BufferAllocator allocator = new RootAllocator(Long.MAX_VALUE);
 
 Map<String, String> props = new HashMap<>();
-props.put(RestNamespaceConfig.URI, "http://localhost:9101/lance");
-props.put(RestNamespaceConfig.DELIMITER, RestNamespaceConfig.DELIMITER_DEFAULT);
+props.put("uri", "http://localhost:9101/lance");
+props.put("delimiter", "$");
 
 LanceNamespace ns = LanceNamespace.connect("rest", props, allocator);
 
@@ -330,27 +333,30 @@ ns.createNamespace(createSchemaNsRequest);
 // Register a table
 RegisterTableRequest registerTableRequest = new RegisterTableRequest();
 registerTableRequest.setLocation("/tmp/lance_catalog/schema/table01");
-registerTableRequest.setId(Lists.newArrayList("lance_catalog", "schema", "table01"));
+registerTableRequest.setId(Arrays.asList("lance_catalog", "schema", "table01"));
 registerTableRequest.setMode("create");
 ns.registerTable(registerTableRequest);
 
 // Create an empty table
 CreateEmptyTableRequest createEmptyTableRequest = new CreateEmptyTableRequest();
 createEmptyTableRequest.setLocation("/tmp/lance_catalog/schema/table02");
-createEmptyTableRequest.setId(Lists.newArrayList("lance_catalog", "schema", "table02"));
+createEmptyTableRequest.setId(Arrays.asList("lance_catalog", "schema", "table02"));
 ns.createEmptyTable(createEmptyTableRequest);
 
-// Create a table with schema inferred from Arrow IPC file
-CreateTableRequest createTableRequest = new CreateTableRequest();
-createTableRequest.setId(Lists.newArrayList("lance_catalog", "schema", "table03"));
-createTableRequest.setMode("create");
+// Create a table with schema inferred from Arrow IPC file.
+// For REST create API, location/properties are passed via headers.
 org.apache.arrow.vector.types.pojo.Schema schema =
         new org.apache.arrow.vector.types.pojo.Schema(
                 Arrays.asList(
                         Field.nullable("id", new ArrowType.Int(32, true)),
                         Field.nullable("value", new ArrowType.Utf8())));
 byte[] body = ArrowUtils.generateIpcStream(schema);
-ns.createTable(createTableRequest, body);
+TableApi tableApi = new TableApi(new ApiClient().setBasePath("http://localhost:9101/lance"));
+Map<String, String> createTableHeaders = new HashMap<>();
+createTableHeaders.put("x-lance-table-location", "/tmp/lance_catalog/schema/table03");
+createTableHeaders.put("x-lance-table-properties", "{}");
+tableApi.createTable(
+    "lance_catalog$schema$table03", body, "$", "create", createTableHeaders);
 
 ```
 
@@ -361,6 +367,7 @@ ns.createTable(createTableRequest, body);
 # Install: pip install lance-namespace==0.4.5
 
 import lance_namespace as ln
+import requests
 
 # Connect to Lance REST service
 ns = ln.connect("rest", {"uri": "http://your_lance_rest:9101/lance"})
@@ -385,16 +392,25 @@ create_empty_table_request = ln.CreateEmptyTableRequest(
     id=['lance_catalog', 'schema', 'table02'],
     location='/tmp/lance_catalog/schema/table02'
 )
+ns.create_empty_table(create_empty_table_request)
 
-# Create a table with schema inferred from Arrow IPC file
-create_table_request = ln.CreateTableRequest(
-    id=['lance_catalog', 'schema', 'table03'],
-    mode='create'
-)
+# Create a table with schema inferred from Arrow IPC file.
+# For REST create API, location/properties are passed via headers.
 with open('schema.ipc', 'rb') as f:
     body = f.read()
 
-ns.create_table(create_table_request, body)
+response = requests.post(
+    "http://your_lance_rest:9101/lance/v1/table/lance_catalog%24schema%24table03/create",
+    params={"delimiter": "$", "mode": "create"},
+    headers={
+        "Content-Type": "application/vnd.apache.arrow.stream",
+        "x-lance-table-location": "/tmp/lance_catalog/schema/table03",
+        "x-lance-table-properties": "{}",
+    },
+    data=body,
+    timeout=30,
+)
+response.raise_for_status()
 ```
 
 </TabItem>
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -273,7 +273,7 @@ curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table
   -d '{
     "id": ["lance_catalog", "schema", "table01"],
     "location": "/tmp/lance_catalog/schema/table01",
-    "mode": "CREATE"
+    "mode": "create"
   }'
 
 # Create a new empty table
@@ -298,7 +298,7 @@ curl -X POST \
 <TabItem value="java" label="Java">
 
 ```java
-// Add dependency: implementation("com.lancedb:lance-namespace-core:0.0.20")
+// Add dependency: implementation("org.lance:lance-namespace-core:0.4.5")
 
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
@@ -312,26 +312,26 @@ Map<String, String> props = new HashMap<>();
 props.put(RestNamespaceConfig.URI, "http://localhost:9101/lance");
 props.put(RestNamespaceConfig.DELIMITER, RestNamespaceConfig.DELIMITER_DEFAULT);
 
-LanceNamespace ns = LanceNamespaces.connect("rest", props, null, allocator);
+LanceNamespace ns = LanceNamespace.connect("rest", props, allocator);
 
 // Create catalog namespace
 CreateNamespaceRequest createCatalogNsRequest = new CreateNamespaceRequest();
 createCatalogNsRequest.addIdItem("lance_catalog");
-createCatalogNsRequest.setMode(CreateNamespaceRequest.ModeEnum.CREATE);
+createCatalogNsRequest.setMode("create");
 ns.createNamespace(createCatalogNsRequest);
 
 // Create schema namespace
 CreateNamespaceRequest createSchemaNsRequest = new CreateNamespaceRequest();
 createSchemaNsRequest.addIdItem("lance_catalog");
 createSchemaNsRequest.addIdItem("schema");
-createSchemaNsRequest.setMode(CreateNamespaceRequest.ModeEnum.CREATE);
+createSchemaNsRequest.setMode("create");
 ns.createNamespace(createSchemaNsRequest);
 
 // Register a table
 RegisterTableRequest registerTableRequest = new RegisterTableRequest();
 registerTableRequest.setLocation("/tmp/lance_catalog/schema/table01");
 registerTableRequest.setId(Lists.newArrayList("lance_catalog", "schema", "table01"));
-registerTableRequest.setMode(RegisterTableRequest.ModeEnum.CREATE);
+registerTableRequest.setMode("create");
 ns.registerTable(registerTableRequest);
 
 // Create an empty table
@@ -342,8 +342,8 @@ ns.createEmptyTable(createEmptyTableRequest);
 
 // Create a table with schema inferred from Arrow IPC file
 CreateTableRequest createTableRequest = new CreateTableRequest();
-createTableRequest.setIds(Lists.newArrayList("lance_catalog", "schema", "table03"));
-createTableRequest.setLocation("/tmp/lance_catalog/schema/table03");
+createTableRequest.setId(Lists.newArrayList("lance_catalog", "schema", "table03"));
+createTableRequest.setMode("create");
 org.apache.arrow.vector.types.pojo.Schema schema =
         new org.apache.arrow.vector.types.pojo.Schema(
                 Arrays.asList(
@@ -358,7 +358,7 @@ ns.createTable(createTableRequest, body);
 <TabItem value="python" label="Python">
 
 ```python
-# Install: pip install lance-namespace==0.0.20
+# Install: pip install lance-namespace==0.4.5
 
 import lance_namespace as ln
 
@@ -389,7 +389,7 @@ create_empty_table_request = ln.CreateEmptyTableRequest(
 # Create a table with schema inferred from Arrow IPC file
 create_table_request = ln.CreateTableRequest(
     id=['lance_catalog', 'schema', 'table03'],
-    location='/tmp/lance_catalog/schema/table03'
+    mode='create'
 )
 with open('schema.ipc', 'rb') as f:
     body = f.read()
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -81,7 +81,8 @@ The Lance REST service provides comprehensive support for namespace management, 
 | TableExists       | Check whether a table exists                                                                                                                                                       | POST        | `/lance/v1/table/{id}/exists`         | 1.1.0         |
 | RegisterTable     | Register an existing Lance table to a namespace                                                                                                                                    | POST        | `/lance/v1/table/{id}/register`       | 1.1.0         |
 | DeregisterTable   | Unregister a table from a namespace (metadata only, data remains)                                                                                                                  | POST        | `/lance/v1/table/{id}/deregister`     | 1.1.0         |
-| CreateEmptyTable  | Declare a table and store the metadata without touching lance table data, for more, please refer to [doc](https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table) | POST        | `/lance/v1/table/{id}/create-empty`   | 1.1.0         |
+| CreateEmptyTable  | **Deprecated**: Use `DeclareTable` instead. Declare a table and store the metadata without touching lance table data, for more, please refer to [doc](https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table) | POST        | `/lance/v1/table/{id}/create-empty`   | 1.1.0         |
+| DeclareTable      | Declare a table and store the metadata without touching lance table data. This is the preferred replacement for `CreateEmptyTable`.                                                | POST        | `/lance/v1/table/{id}/declare`        | 1.1.0         |
 
 More details, please refer to the [Lance REST API specification](https://lance.org/format/namespace/rest/catalog-spec/)
 
@@ -277,6 +278,7 @@ curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table
   }'
 
 # Create a new empty table
+# x-lance-table-properties is optional; if omitted, it defaults to an empty map.
 curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table02/create-empty \
   -H 'Content-Type: application/json' \
   -H "x-lance-table-properties: {\"description\":\"This is table02\"}" \
@@ -284,6 +286,14 @@ curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table
     "id": ["lance_catalog", "schema", "table02"],
     "location": "/tmp/lance_catalog/schema/table02"
   }'  
+
+# Declare a table (preferred replacement for create-empty)
+curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table04/declare \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "id": ["lance_catalog", "schema", "table04"],
+    "location": "/tmp/lance_catalog/schema/table04"
+  }'
   
 # Create a table with schema, the schema is inferred from the Arrow IPC file
 curl -X POST \
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -90,6 +90,9 @@ More details, please refer to the [Lance REST API specification](https://lance.o
 
 Some operations have specific behaviors and modes. Below are important details to consider:
 
+Mode values are parsed case-insensitively. The examples below use lowercase values as the
+REST-style canonical form.
+
 #### Namespace Operations
 
 **CreateNamespace** supports three modes:
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -82,7 +82,7 @@ The Lance REST service provides comprehensive support for namespace management, 
 | RegisterTable     | Register an existing Lance table to a namespace                                                                                                                                    | POST        | `/lance/v1/table/{id}/register`       | 1.1.0         |
 | DeregisterTable   | Unregister a table from a namespace (metadata only, data remains)                                                                                                                  | POST        | `/lance/v1/table/{id}/deregister`     | 1.1.0         |
 | CreateEmptyTable  | **Deprecated**: Use `DeclareTable` instead. Declare a table and store the metadata without touching lance table data, for more, please refer to [doc](https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table) | POST        | `/lance/v1/table/{id}/create-empty`   | 1.1.0         |
-| DeclareTable      | Declare a table and store the metadata without touching lance table data. This is the preferred replacement for `CreateEmptyTable`.                                                | POST        | `/lance/v1/table/{id}/declare`        | 1.1.0         |
+| DeclareTable      | Declare a table and store the metadata without touching lance table data. This is the preferred replacement for `CreateEmptyTable`.                                                | POST        | `/lance/v1/table/{id}/declare`        | 1.3.0         |
 
 More details, please refer to the [Lance REST API specification](https://lance.org/format/namespace/rest/catalog-spec/)
 
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import java.util.Optional;
 import org.lance.namespace.model.CreateEmptyTableResponse;
 import org.lance.namespace.model.CreateTableResponse;
+import org.lance.namespace.model.DeclareTableResponse;
 import org.lance.namespace.model.DeregisterTableResponse;
 import org.lance.namespace.model.DescribeTableResponse;
 import org.lance.namespace.model.DropTableResponse;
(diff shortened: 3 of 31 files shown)


FANNG1:
@bbiiaaoo could you fix the CI and address the review comment to make this PR ready for review?

FANNG1:
@yuqi1129 @roryqi  could you help to check whether it's on the correct direction?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceNamespaceOperations.java
```diff
@@ -66,6 +66,22 @@ public Response listNamespaces(
...
+  @GET
+  @Path("/list")
+  @Timed(name = "list-namespaces-root." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "list-namespaces-root", absolute = true)
+  public Response listNamespacesOnRoot(
+      @DefaultValue(NAMESPACE_DELIMITER_DEFAULT) @QueryParam("delimiter") String delimiter,
+      @QueryParam("page_token") String pageToken,
+      @QueryParam("limit") Integer limit) {
+    return listNamespacesInternal(delimiter, delimiter, pageToken, limit);
+  }
```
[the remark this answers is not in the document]
bbiiaaoo: Done

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceNamespaceOperations.java
```diff
@@ -84,6 +100,19 @@ public Response listNamespaces(
...
+
+  @POST
+  @Path("/describe")
+  @Timed(name = "describe-namespaces-root." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "describe-namespaces-root", absolute = true)
+  public Response describeNamespaceOnRoot(
+      @DefaultValue(NAMESPACE_DELIMITER_DEFAULT) @QueryParam("delimiter") String delimiter) {
+    return describeNamespaceInternal(delimiter, delimiter);
+  }
```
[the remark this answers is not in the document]
bbiiaaoo: Done

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceNamespaceOperations.java
```diff
@@ -66,6 +66,22 @@ public Response listNamespaces(
...
+
+  @GET
+  @Path("/list")
+  @Timed(name = "list-namespaces-root." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "list-namespaces-root", absolute = true)
+  public Response listNamespacesOnRoot(
+      @DefaultValue(NAMESPACE_DELIMITER_DEFAULT) @QueryParam("delimiter") String delimiter,
+      @QueryParam("page_token") String pageToken,
+      @QueryParam("limit") Integer limit) {
+    return listNamespacesInternal(delimiter, delimiter, pageToken, limit);
```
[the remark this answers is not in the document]
bbiiaaoo: The code has been added in TestLanceNamespaceOperations.java

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -147,10 +152,9 @@ public Response createEmptyTable(
...
-      Map<String, String> props =
-          request.getProperties() == null
-              ? Maps.newHashMap()
-              : Maps.newHashMap(request.getProperties());
+      MultivaluedMap<String, String> headersMap = headers.getRequestHeaders();
+      String tableProperties = headersMap.getFirst(LANCE_TABLE_PROPERTIES_PREFIX_HEADER);
+      Map<String, String> props = SerializationUtils.deserializeProperties(tableProperties);
 
       CreateEmptyTableResponse response =
           lanceNamespace.asTableOps().createEmptyTable(tableId, delimiter, tableLocation, props);
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -337,13 +341,24 @@ private void validateAlterColumnsRequest(AlterTableAlterColumnsRequest request)
...
+          StringUtils.isNotBlank(alteration.getPath()), "Column path to alter cannot be empty.");
       Preconditions.checkArgument(
           StringUtils.isNotBlank(alteration.getRename()),
-          "Rename field must be specified when castTo is not provided.");
+          "Only RENAME alteration is supported currently.");
+      Preconditions.checkArgument(
+          alteration.getDataType() == null
+              && alteration.getNullable() == null
+              && alteration.getVirtualColumn() == null,
+          "Only RENAME alteration is supported currently.");
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

docs/lance-rest-service.md
```diff
@@ -312,26 +312,26 @@ Map<String, String> props = new HashMap<>();
 props.put(RestNamespaceConfig.URI, "http://localhost:9101/lance");
 props.put(RestNamespaceConfig.DELIMITER, RestNamespaceConfig.DELIMITER_DEFAULT);
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/SerializationUtils.java
```diff
@@ -35,19 +35,19 @@ private SerializationUtils() {
...
+    try {
+      Map<String, Object> rawProperties =
+          JsonUtils.anyFieldMapper()
+              .readValue(serializedProperties, new TypeReference<Map<String, Object>>() {});
+      Map<String, String> deserializedProperties = new HashMap<>();
+      rawProperties.forEach((k, v) -> deserializedProperties.put(k, String.valueOf(v)));
+      return deserializedProperties;
+    } catch (Exception e) {
+      throw new IllegalArgumentException("Failed to deserialize table properties JSON", e);
+    }
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -595,8 +583,7 @@ void testDeregisterTable() {
     Mockito.reset(tableOps);
     when(tableOps.deregisterTable(any(), any()))
         .thenThrow(
-            LanceNamespaceException.notFound(
-                "Table not found", "NoSuchTableException", tableIds, ""));
+            new org.lance.namespace.errors.TableNotFoundException("Table not found", "", tableIds));
     resp =
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -644,14 +630,13 @@ void testDescribeTable() {
...
+    Assertions.assertEquals(createTableResponse.getMetadata(), response.getMetadata());
 
     // Test not found exception
     Mockito.reset(tableOps);
     when(tableOps.describeTable(any(), any(), any()))
         .thenThrow(
-            LanceNamespaceException.notFound(
-                "Table not found", "NoSuchTableException", tableIds, ""));
+            new org.lance.namespace.errors.TableNotFoundException("Table not found", "", tableIds));
     resp =
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -693,9 +677,7 @@ void testTableExists() {
     Assertions.assertEquals(Response.Status.OK.getStatusCode(), resp.getStatus());
 
     // test throw exception
-    doThrow(
-            LanceNamespaceException.notFound(
-                "Table not found", "NoSuchTableException", tableIds, ""))
+    doThrow(new org.lance.namespace.errors.TableNotFoundException("Table not found", "", tableIds))
         .when(tableOps)
         .tableExists(any(), any());
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -749,9 +730,7 @@ void testDropTable() {
     Assertions.assertEquals(dropTableResponse.getLocation(), response.getLocation());
 
     // test throw exception
-    doThrow(
-            LanceNamespaceException.notFound(
-                "Table not found", "NoSuchTableException", tableIds, ""))
+    doThrow(new org.lance.namespace.errors.TableNotFoundException("Table not found", "", tableIds))
         .when(tableOps)
         .dropTable(any(), any());
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -850,8 +826,7 @@ void testDropColumns() {
     Mockito.reset(tableOps);
     when(tableOps.alterTable(any(), any(), any(AlterTableDropColumnsRequest.class)))
         .thenThrow(
-            LanceNamespaceException.notFound(
-                "Table not found", "NoSuchTableException", tableIds, ""));
+            new org.lance.namespace.errors.TableNotFoundException("Table not found", "", tableIds));
     resp =
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -903,8 +878,7 @@ void testAlterColumns() {
     Mockito.reset(tableOps);
     when(tableOps.alterTable(any(), any(), any(AlterTableAlterColumnsRequest.class)))
         .thenThrow(
-            LanceNamespaceException.notFound(
-                "Table not found", "NoSuchTableException", tableIds, ""));
+            new org.lance.namespace.errors.TableNotFoundException("Table not found", "", tableIds));
     resp =
```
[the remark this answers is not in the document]
bbiiaaoo: Fixed

yuqi1129:
@bbiiaaoo 
Is this PR ready for review?

bbiiaaoo:
Hi @yuqi1129 @FANNG1  ,

All previous review comments have been addressed. And verified that all CI workflows are now passing.
This PR is now ready for review. Thanks!

docs/lance-rest-service.md
```diff
@@ -273,16 +273,16 @@ curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table
...
     "id": ["lance_catalog", "schema", "table01"],
     "location": "/tmp/lance_catalog/schema/table01",
-    "mode": "CREATE"
+    "mode": "create"
   }'
 
 # Create a new empty table
 curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table02/create-empty \
   -H 'Content-Type: application/json' \
+  -H "x-lance-table-properties: {\"description\":\"This is table02\"}" \
```
yuqi1129, line 284: Is the header `x-lance-table-properties` optional or required?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/CommonUtil.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+/** Utility methods used by Gravitino Lance namespace operations. */
+class CommonUtil {
+
+  private CommonUtil() {}
+
+  static String formatCurrentStackTrace() {
```
yuqi1129, line 29: What is the method used for? Can `Throwables.getStackTraceAsString()` meet your needs?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
```diff
@@ -52,20 +42,43 @@
...
+import org.lance.namespace.model.DropNamespaceResponse;
+import org.lance.namespace.model.ListNamespacesResponse;
+import org.lance.namespace.model.ListTablesResponse;
 
 public class GravitinoLanceNameSpaceOperations implements LanceNamespaceOperations {
 
   private final GravitinoLanceNamespaceWrapper namespaceWrapper;
   private final GravitinoClient client;
 
+  private enum CreateMode {
```
yuqi1129, line 67: Why do we need to add those three enums? Can't we use enums in the Lance API?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -138,19 +141,26 @@ public Response createTable(
   @Produces("application/json")
   @Timed(name = "create-empty-table." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
   @ResponseMetered(name = "create-empty-table", absolute = true)
+  @SuppressWarnings("deprecation")
   public Response createEmptyTable(
       @PathParam("id") String tableId,
       @QueryParam("delimiter") @DefaultValue(NAMESPACE_DELIMITER_DEFAULT) String delimiter,
-      CreateEmptyTableRequest request,
+      Map<String, Object> requestBody,
```
yuqi1129, line 150: Why is the value type of `requestBody` `Object`?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
```diff
@@ -69,6 +67,7 @@ CreateTableResponse createTable(
    * @param tableProperties the properties of the table
    * @return the response of the create table operation
    */
+  @SuppressWarnings("deprecation")
   CreateEmptyTableResponse createEmptyTable(
```
yuqi1129, line 87: I recalled that `createEmptyTable` has been replaced with `declareTable`, so will you plan to support it?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/JsonArrowSchemaConverter.java
```diff
@@ -0,0 +1,167 @@
...
+import org.apache.arrow.vector.types.FloatingPointPrecision;
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.Schema;
+import org.lance.namespace.model.JsonArrowDataType;
+import org.lance.namespace.model.JsonArrowField;
+import org.lance.namespace.model.JsonArrowSchema;
+
+/** Converts Arrow schema to Lance Namespace JsonArrowSchema model. */
+class JsonArrowSchemaConverter {
```
yuqi1129, line 33: Is this class directly copied from Lance repo?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
```diff
@@ -414,6 +407,60 @@ private <T> T[] buildChanges(
     return Stream.concat(setPropertiesStream, removePropertiesStream).toArray(arrayCreator);
   }
 
+  private static CreateMode parseCreateMode(String instance, String mode) {
```
yuqi1129, line 411: This one seems to be duplicated with `normalizeCreateMode`. Can you try to optimize it?

docs/lance-rest-service.md
bbiiaaoo: `x-lance-table-properties` is optional for create-empty. If omitted, it defaults to an empty map. Request-body properties are still accepted for backward compatibility (header wins on key conflicts). I will clarify this in the documentation in the next commit.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/CommonUtil.java
bbiiaaoo: `formatCurrentStackTrace() `is used to populate the Lance exception detail field. Good point on utility reuse; I will switch it to `Throwables.getStackTraceAsString(...)` in the next commit.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
bbiiaaoo: In `lance-namespace 0.4.5`, mode/behavior are now plain strings in generated request models (no public enums to reuse). We keep local enums only as normalized internal states for type-safe branching after parsing input strings.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
bbiiaaoo: Good question. We intentionally use `Map<String, Object>` here for backward compatibility.

In lance-namespace 0.4.5, `CreateEmptyTableRequest` does not include a `properties` field, but we still accept legacy request-body `properties` from existing clients. Using `Object` allows us to safely normalize non-string JSON values (for example numbers/booleans) into strings before passing them downstream.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/JsonArrowSchemaConverter.java
bbiiaaoo: It is not directly copied from the current Lance repo as-is.

During the upgrade, the old utility class we previously relied on was no longer available from the new Java artifacts, so we added a local converter in Gravitino to preserve compatible `JsonArrowSchema` behavior for this code path.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/LanceTableOperations.java
bbiiaaoo: Thanks for the reminder, you are absolutely right.

`createEmptyTable` is deprecated in lance-namespace 0.4.5, and `declareTable` is the preferred API now. We plan to add `declareTable` support in the next commit, while keeping `create-empty` for backward compatibility with existing clients.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
bbiiaaoo: Good catch. There is duplicated normalization logic between namespace `parseCreateMode` and table `normalizeCreateMode`.

They are not fully identical because namespace mode is converted to an internal enum, while table mode is normalized to a string property passed downstream (and register mode also accepts `REGISTER` as an alias). 

I’ll optimize this by extracting the shared token-normalization logic and reusing it in both paths, while keeping the behavior-specific mappings unchanged.

docs/lance-rest-service.md
```diff
@@ -81,7 +81,8 @@ The Lance REST service provides comprehensive support for namespace management,
...
 | DeregisterTable   | Unregister a table from a namespace (metadata only, data remains)                                                                                                                  | POST        | `/lance/v1/table/{id}/deregister`     | 1.1.0         |
-| CreateEmptyTable  | Declare a table and store the metadata without touching lance table data, for more, please refer to [doc](https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table) | POST        | `/lance/v1/table/{id}/create-empty`   | 1.1.0         |
+| CreateEmptyTable  | **Deprecated**: Use `DeclareTable` instead. Declare a table and store the metadata without touching lance table data, for more, please refer to [doc](https://docs.lancedb.com/api-reference/rest/table/create-an-empty-table) | POST        | `/lance/v1/table/{id}/create-empty`   | 1.1.0         |
+| DeclareTable      | Declare a table and store the metadata without touching lance table data. This is the preferred replacement for `CreateEmptyTable`.                                                | POST        | `/lance/v1/table/{id}/declare`        | 1.1.0         |
```
yuqi1129, line 85: The version should be 1.3.0

yuqi1129:
@FANNG1 @roryqi 
Would you also like to take a look?

yuqi1129:
@bbiiaaoo 
1. Have you tested it with the LanceSpark and the LanceRay with the corresponding version?
2. You may also need to update the `Compatibility Matrix` part in the document `lance-rest-integration.md`.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
```diff
@@ -414,6 +408,56 @@ private <T> T[] buildChanges(
...
+    throw new InvalidInputException(
+        "Unknown create namespace mode: " + mode, CommonUtil.formatCurrentStackTrace(), instance);
+  }
+
+  private static DropMode parseDropMode(String instance, String mode) {
+    if (mode == null) {
+      return DropMode.FAIL;
+    }
+    String normalized = CommonUtil.normalizeToken(mode);
+    if ("FAIL".equals(normalized)) {
```
roryqi, line 434: Could u use `DropMode.valueOf(normalized)`?

FANNG1:
In [`LanceNamespaceOperations`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceNamespaceOperations.java#L108-L115) this PR adds a new root route, `POST /v1/namespace/describe`, which forwards `ROOT_NAMESPACE_ID = ""` to `describeNamespaceInternal(...)`.

However, the real backend still rejects empty namespace ids in [`GravitinoLanceNameSpaceOperations.describeNamespace`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java#L129-L155), because it requires `nsId.levels() > 0`.

So the new route cannot succeed against the actual backend. The mock-based Jersey test passes only because `namespaceOps.describeNamespace(any(), any())` is stubbed to return success for all inputs, including `""`:
[`TestLanceNamespaceOperations`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java#L219-L246)

This also matches the integration test, which still expects empty-id describe to fail:
[`LanceRESTServiceIT`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java#L183-L190)

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
bbiiaaoo: Thanks for the suggestion. `DropMode.valueOf(normalized)` would work for this enum since the values are `FAIL` and `SKIP`.

I kept the explicit mapping here to make it consistent with the other mode parsers, especially `CreateMode`, where `normalizeToken()` turns values like `exist_ok` into `EXISTOK`, so `CreateMode.valueOf(...)` would not work directly. The explicit mapping also lets us keep throwing `InvalidInputException` with the Lance-specific error context instead of relying on `IllegalArgumentException` from `valueOf()` and wrapping it separately.

bbiiaaoo:
> In [`LanceNamespaceOperations`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceNamespaceOperations.java#L108-L115) this PR adds a new root route, `POST /v1/namespace/describe`, which forwards `ROOT_NAMESPACE_ID = ""` to `describeNamespaceInternal(...)`.
> 
> However, the real backend still rejects empty namespace ids in [`GravitinoLanceNameSpaceOperations.describeNamespace`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java#L129-L155), because it requires `nsId.levels() > 0`.
> 
> So the new route cannot succeed against the actual backend. The mock-based Jersey test passes only because `namespaceOps.describeNamespace(any(), any())` is stubbed to return success for all inputs, including `""`: [`TestLanceNamespaceOperations`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java#L219-L246)
> 
> This also matches the integration test, which still expects empty-id describe to fail: [`LanceRESTServiceIT`](https://github.com/apache/gravitino/blob/d8628d949/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java#L183-L190)

@FANNG1 Thanks for catching this. You're right that the root describe route was not valid because the real backend still rejects an empty namespace id.

I've fixed this by removing the unsupported `POST /v1/namespace/describe` root endpoint and the corresponding mock-based Jersey test expectation. The behavior is now aligned with the integration test, where describing the root namespace with an empty id is still expected to fail against the real backend.

bbiiaaoo:
> @bbiiaaoo
> 
> 1. Have you tested it with the LanceSpark and the LanceRay with the corresponding version?
> 2. You may also need to update the `Compatibility Matrix` part in the document `lance-rest-integration.md`.

@yuqi1129 
Thanks for the reminder. I have completed the compatibility verification for both LanceSpark and LanceRay, and I will include the Compatibility Matrix update and the new Spark integration test in the next commit.

I also added `LanceSparkRESTServiceIT` to cover Spark integration against the Lance REST service, including namespace operations, table create/read/write, empty table reads, filter queries, drop operations, and error cases.

docs/lance-rest-service.md
bbiiaaoo: Done

roryqi:
@yuqi1129 @jerryshao LGTM, Could u take an another look?

docs/lance-rest-integration.md
```diff
@@ -22,7 +22,7 @@ The following table outlines the tested compatibility between Gravitino versions
 
 | Gravitino Version (Lance REST) | Supported lance-spark Versions | Supported lance-ray Versions |
 |--------------------------------|--------------------------------|------------------------------|
-| 1.1.1                          | 0.0.10 – 0.0.15                | 0.0.6 – 0.0.8                |
+| 1.3.0                          | 0.1.0 – 0.2.0                 | 0.0.6 – 0.2.0               |
```
jerryshao: Have we verified these different lance spark and lance ray versions?

docs/lance-rest-integration.md
```diff
@@ -158,6 +158,7 @@ pip install lance-ray
 :::info
 - Ray will be automatically installed if not already present
 - lance-ray is currently tested with Ray versions 2.41.0 to 2.50.0
```
yuqi1129: The version of Ray used by lance-ray may have changed. Can you confirm it?

docs/lance-rest-integration.md
yuqi1129: Format the table. Besides,  can you try to change it to the format

| Gravitino Version (Lance REST) | Supported lance-spark Versions | Supported lance-ray Versions |
|--------------------------------|--------------------------------|------------------------------|
| 1.1.1  - 1.2.1                        | 0.0.10 – 0.0.15                | 0.0.6 – 0.0.8                |
| 1.3.0                          | 0.1.0 – 0.2.0                 | 0.0.6 – 0.2.0               |

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
```diff
@@ -414,6 +408,56 @@ private <T> T[] buildChanges(
     return Stream.concat(setPropertiesStream, removePropertiesStream).toArray(arrayCreator);
   }
 
+  private static CreateMode parseCreateMode(String instance, String mode) {
+    if (mode == null) {
+      return CreateMode.CREATE;
+    }
+    String normalized = CommonUtil.normalizeToken(mode);
```
yuqi1129: `normalizeToken` will remove all sepcical character, why do we need it? Is true that it's a valid mode if the value looks like `#CREA!TE$`?

docs/lance-rest-service.md
```diff
@@ -273,17 +274,26 @@ curl -X POST http://localhost:9101/lance/v1/table/lance_catalog%24schema%24table
   -d '{
     "id": ["lance_catalog", "schema", "table01"],
     "location": "/tmp/lance_catalog/schema/table01",
-    "mode": "CREATE"
+    "mode": "create"
```
yuqi1129, line 280: So the value must be `create`, and `CREATE` will run into errors?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableAlterHandler.java
```diff
@@ -115,21 +115,20 @@ public AlterTableAlterColumnsResponse handle(
...
           changes.add(TableChange.renameColumn(new String[] {columnName}, newName));
         }
 
-        // The format of ColumnAlteration#castTo is unclear, so we will skip it now
-        // for more, please refer to: https://shorturl.at/bYI0Z (short url for
-        // github.com/lance-format/lance-namespace)
-        if (StringUtils.isNotBlank(column.getCastTo())) {
+        if (column.getDataType() != null
+            || column.getNullable() != null
+            || column.getVirtualColumn() != null) {
```
yuqi1129: So an alter operation will contain those three informations?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
```diff
@@ -366,4 +377,37 @@ private JsonArrowSchema toJsonArrowSchema(Column[] columns) {
...
+    String normalized = CommonUtil.normalizeToken(mode);
+    if ("CREATE".equals(normalized)) {
+      return "CREATE";
+    }
+    if ("EXISTOK".equals(normalized)) {
+      return "EXIST_OK";
+    }
+    if ("OVERWRITE".equals(normalized)) {
+      return "OVERWRITE";
+    }
```
yuqi1129: That code looks tedious, can you use a set or enum to replace them?

docs/lance-rest-integration.md
FANNG1: Ray is a transitive dependency of lance-ray, not a direct concern for Gravitino. Suggesting to simply remove this line — users should refer to lance-ray's own release notes for its internal dependency requirements. WDYT?

yuqi1129: The reason that I add it is that it only works with that version range. If the user has only installed Ray that is not in that range, something unexpected will happen.
For more, please refer to https://github.com/lance-format/lance-ray/issues/68

docs/lance-rest-integration.md
bbiiaaoo: @jerryshao Thanks for the comments.
For lance-spark, we have verified the integration test operations in LanceSparkRESTServiceIT with versions 0.1.0, 0.1.1, and 0.2.0.
For lance-ray, we have verified table creation and read/write capabilities for each version from 0.0.6 to 0.2.0.

docs/lance-rest-integration.md
FANNG1: This is actually a lance-ray bug — its `on_write_complete` and `estimate_inmemory_data_size` implementations are incompatible with certain Ray API versions. lance-ray already works around this by pinning `ray[data]>=2.40.0,<=2.52.1` in its own `pyproject.toml` (tracked in https://github.com/lance-format/lance-ray/issues/68). Since the constraint is declared in lance-ray's own dependency, pip will enforce it automatically. Documenting the Ray version range in Gravitino's docs would be redundant and would require Gravitino to track lance-ray's internal bug fixes going forward. Suggesting to remove this line and let lance-ray's own packaging handle it.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
bbiiaaoo: Good catch. We only need case-insensitive parsing here, not removing special characters. I updated the normalization to trim and uppercase only, and malformed values like #CREATE$ will now be rejected.

The table format has been updated as per the suggestion.

docs/lance-rest-service.md
bbiiaaoo: The canonical REST value in the document is lowercase create. The implementation still accepts CREATE for compatibility because mode parsing is case-insensitive. I also added a note to clarify this.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableAlterHandler.java
bbiiaaoo: No, these fields are optional. Currently Gravitino only supports rename, so the check rejects the request if dataType, nullable, or virtualColumn is set. I updated the helper and error message to make this clearer.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
bbiiaaoo: Agreed. I replaced the manual string comparisons with enum-based parsing for namespace and table modes.

docs/lance-rest-integration.md
bbiiaaoo: Thanks for the discussion. I removed this Ray version note in the latest commit, since Ray is a transitive dependency managed by lance-ray and users should follow lance-ray's own dependency constraints.

yuqi1129:
I have no more comments. @jerryshao do you have time to take a look again?

jerryshao (approved):

Issue #10621, opened 2026-04-01
[Improvement] Fix Metalake identifier resolution to use object name instead of context metalakeName


justinmclean:
### What would you like to be improved?

MetadataObjectUtil.toEntityIdent(String metalakeName, MetadataObject metadataObject) handles MetadataObject.Type.METALAKE by returning NameIdentifierUtil.ofMetalake(metalakeName) instead of using metadataObject.name(). As a result, if a request carries a METALAKE metadata object whose name differs from the surrounding metalakeName context, the code resolves and validates the context metalake rather than the target object.

This is reachable from generic REST flows such as owner operations, which parse {fullName} into a MetadataObject and then call MetadataObjectUtil.checkMetadataObject(metalake, object). A mismatched request can therefore pass validation against the wrong metalake.

### How should we improve?

Update the METALAKE branch in MetadataObjectUtil.toEntityIdent to build the identifier from metadataObject.name() rather than metalakeName.

It would also help to add a defensive validation for METALAKE objects in request-handling paths or inside checkMetadataObject, rejecting cases where metadataObject.name() does not equal the surrounding metalake path/context.

Here's a unit test to help:
```
  @Test
  public void testToEntityIdentUsesMetadataObjectNameForMetalakeObject() {
    Assertions.assertEquals(
        NameIdentifier.of("target_metalake"),
        MetadataObjectUtil.toEntityIdent(
            "request_metalake",
            MetadataObjects.of(null, "target_metalake", MetadataObject.Type.METALAKE)));
  }
```

griffonbyte:
I'd like to take this on. The fix is to change the METALAKE branch in toEntityIdent to use metadataObject.name() instead of the context metalakeName parameter. I'll include the suggested test case.

sachinnn99:
I'd like to work on this issue.

PR #10639, opened 2026-04-01
[#10621] fix(core): Use metadataObject.name() for METALAKE identifier resolution

sachinnn99:

Changed `MetadataObjectUtil.toEntityIdent()` to use `metadataObject.name()` instead of the `metalakeName` parameter when handling `MetadataObject.Type.METALAKE`.

The METALAKE case was the only branch using the context parameter `metalakeName` instead of the object's own name. All other type branches (ROLE, TAG, CATALOG, SCHEMA, etc.) already use `metadataObject.name()` or `metadataObject.fullName()`. This inconsistency means a mismatched metalake name in the request silently resolves to the wrong metalake.

Fix: #10621

No.

Added a unit test in `TestMetadataObjectUtil` that passes a different `metalakeName` parameter than the metadata object's name, verifying the object's name is used in the returned `NameIdentifier`.

```bash
./gradlew :core:test --tests "org.apache.gravitino.utils.TestMetadataObjectUtil" -PskipITs
```

### DIFF
--- a/core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
+++ b/core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
@@ -110,7 +110,7 @@ public class MetadataObjectUtil {
 
     switch (metadataObject.type()) {
       case METALAKE:
-        return NameIdentifierUtil.ofMetalake(metalakeName);
+        return NameIdentifierUtil.ofMetalake(metadataObject.name());
       case ROLE:
         return AuthorizationUtils.ofRole(metalakeName, metadataObject.name());
       case TAG:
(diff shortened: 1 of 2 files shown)


jerryshao:
Is there any issue with the current code?

sachinnn99:
Hi @jerryshao, the METALAKE case currently uses the `metalakeName` parameter instead of `metadataObject.name()`. If the two differ (e.g., metalake name in the URL path vs. the metadata object parsed from the request body), the code resolves the wrong metalake. Every other type branch already uses the object's own name, so this aligns the METALAKE case with the rest. See #10621 for details.

jerryshao:
Can you show me a real case of how to reproduce this issue?

sachinnn99:
Hi @jerryshao, I see this was already fixed in #10638 — closing this PR. Thanks!

PR #10641, opened 2026-04-01
[#9736][followup] feat(iceberg-rest): Add ETag header to registerTable response

laserninja:

Add ETag header to the `registerTable` response for consistency with `createTable` and `updateTable`, which already include ETags. This was identified by @roryqi in [#10536 (comment)](https://github.com/apache/gravitino/pull/10536#issuecomment-4168013863).

The `registerTable` endpoint returns a `LoadTableResponse` but did not include an ETag header, unlike `createTable` and `updateTable`. This meant clients could not reuse ETags from `registerTable` for subsequent conditional `loadTable` requests (`If-None-Match`), breaking ETag consistency across all endpoints that return `LoadTableResponse`.

No user-facing API changes. The `registerTable` response now includes an `ETag` header when the table metadata has a metadata file location.

- Added `testRegisterTableReturnsETag` test verifying ETag presence in `registerTable` response.
- Updated test mock (`CatalogWrapperForTest`) to return `TableMetadata` with a `metadataFileLocation` so ETag can be generated.
- All existing ETag and namespace tests pass.

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
@@ -52,10 +52,14 @@ public class IcebergRESTUtils {
 
   private static final Logger LOG = LoggerFactory.getLogger(IcebergRESTUtils.class);
 
+  public static final String SNAPSHOT_ALL = "all";
+
+  public static final String SNAPSHOT_REFS = "refs";
+
   /** Snapshot modes for the Iceberg loadTable endpoint. */
   public enum SnapshotMode {
-    ALL("all"),
-    REFS("refs");
+    ALL(SNAPSHOT_ALL),
+    REFS(SNAPSHOT_REFS);
 
     private final String value;
 
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
@@ -291,7 +291,7 @@ public class IcebergTableOperations {
           String namespace,
       @IcebergAuthorizationMetadata(type = RequestType.LOAD_TABLE) @Encoded() @PathParam("table")
           String table,
-      @DefaultValue("all") @QueryParam("snapshots") String snapshots,
+      @DefaultValue(IcebergRESTUtils.SNAPSHOT_ALL) @QueryParam("snapshots") String snapshots,
       @HeaderParam(X_ICEBERG_ACCESS_DELEGATION) String accessDelegation,
       @HeaderParam(IF_NONE_MATCH) String ifNoneMatch) {
     String catalogName = IcebergRESTUtils.getCatalogName(prefix);


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -579,7 +579,7 @@ static LoadTableResponse filterSnapshotsByRefs(LoadTableResponse loadTableRespon
    * @param loadTableResponse the table response to include in the body
    * @return a Response with ETag header set
    */
-  private static Response buildResponseWithETag(LoadTableResponse loadTableResponse) {
+  static Response buildResponseWithETag(LoadTableResponse loadTableResponse) {
```
bharos: We are making this method from private to package-private to be invoked from icebergNamespaceOperations, which seems to be a bit of anti-pattern

bharos: IcebergRESTUtils already exists for response-building helpers, buildResponseWithETag can belong there ? And can be shared between IcebergTableOperations and IcebergNamespaceOperations

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
```diff
@@ -189,6 +190,16 @@ void testRegisterTable() {
     verifyRegisterTableSucc("register_foo2", Namespace.of("register_ns_2", "a"));
   }
 
+  @Test
+  void testRegisterTableReturnsETag() {
+    Namespace ns = Namespace.of("register_etag_ns");
+    Response response = doRegisterTable("register_etag_foo1", ns);
+    Assertions.assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
+    String etag = response.getHeaderString("ETag");
+    Assertions.assertNotNull(etag, "ETag header should be present in register table response");
```
bharos, line 201: Is it also possible to check the ETag match?

Test only checks ETag presence, not validity Since the mock has a known metadata location (/mock/metadata/v1.metadata.json), you could compute the expected SHA-256 and assert the exact value

laserninja:
1. **Moved ETag helpers to `IcebergRESTUtils`**: `generateETag()` and `buildResponseWithETag()` now live in `IcebergRESTUtils` as public static methods. `IcebergTableOperations` delegates to them via private wrappers. `IcebergNamespaceOperations` calls `IcebergRESTUtils.buildResponseWithETag()` directly. No more cross-class package-private access.

2. **Added ETag value assertion**: `testRegisterTableReturnsETag` now computes the expected SHA-256 from the known mock metadata location (`/mock/metadata/v1.metadata.json` + `"all"`) and asserts the exact ETag value matches.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
bharos: Do we need this here now, as we have this in the utils file as well 

Make the IcebergRESTUtils one public 
Remove it from IcebergTableOperations and reference IcebergRESTUtils.DEFAULT_SNAPSHOTS

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -98,7 +95,6 @@ public class IcebergTableOperations {
 
   @VisibleForTesting public static final String IF_NONE_MATCH = "If-None-Match";
 
-  @VisibleForTesting static final String DEFAULT_SNAPSHOTS = "all";
   @VisibleForTesting static final String SNAPSHOTS_REFS = "refs";
```
roryqi: We should put `all` and `refs` in the same place to reuse them.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -39,15 +43,100 @@
...
+   * Generates an ETag based on the table metadata file location and snapshot mode. The ETag is a
+   * SHA-256 hash that incorporates both the metadata location and the snapshots parameter, ensuring
+   * distinct ETags for different representations of the same table version (e.g., snapshots=all vs
+   * snapshots=refs).
+   *
+   * @param metadataLocation the metadata file location
+   * @param snapshots the snapshots query parameter value (e.g., "all", "refs"), may be null
+   * @return the generated ETag, or null if generation fails
+   */
+  public static EntityTag generateETag(String metadataLocation, String snapshots) {
```
roryqi: It's not good to return `null`. We should return Optional when required. Optional will help user must consider null value explictly.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -39,15 +43,100 @@
...
+
+  /**
+   * Generates an ETag based on the table metadata file location. The ETag is a SHA-256 hash of the
+   * metadata location, which changes whenever the table metadata is updated.
+   *
+   * @param metadataLocation the metadata file location
+   * @return the generated ETag, or null if generation fails
+   */
+  public static EntityTag generateETag(String metadataLocation) {
+    return generateETag(metadataLocation, null);
```
roryqi: You can use default value instead of null value here.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
laserninja: Done. Removed the duplicate DEFAULT_SNAPSHOTS from IcebergTableOperations and made it public static final in IcebergRESTUtils. The @DefaultValue annotation now references IcebergRESTUtils.DEFAULT_SNAPSHOTS. All tests pass.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -39,15 +43,100 @@
...
+   * @return the generated ETag, or null if generation fails
+   */
+  public static EntityTag generateETag(String metadataLocation, String snapshots) {
+    if (metadataLocation == null) {
+      return null;
+    }
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      digest.update(metadataLocation.getBytes(StandardCharsets.UTF_8));
+      if (snapshots != null) {
```
roryqi: We will have different etags if we don't set the snapshots and set the snapshots `ALL`. Is it correct?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
laserninja: done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
laserninja: Both `generateETag` methods now return `Optional<EntityTag>`, all callers updated to use `isPresent()`/`get()`/`ifPresent()`

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
laserninja: done

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
laserninja: Fixed by always using 'DEFAULT_SNAPSHOTS'; the if (snapshots != null) guard removed since snapshots is always non-null now

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -39,15 +44,99 @@
...
+   * SHA-256 hash that incorporates both the metadata location and the snapshots parameter, ensuring
+   * distinct ETags for different representations of the same table version (e.g., snapshots=all vs
+   * snapshots=refs).
+   *
+   * @param metadataLocation the metadata file location
+   * @param snapshots the snapshots query parameter value (e.g., "all", "refs")
+   * @return the generated ETag
+   */
+  public static Optional<EntityTag> generateETag(String metadataLocation, String snapshots) {
+    if (metadataLocation == null) {
```
roryqi: Could u use `StringUtils.isBlank` here?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -39,15 +44,99 @@
...
 import org.apache.iceberg.rest.responses.ErrorResponse;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class IcebergRESTUtils {
 
+  private static final Logger LOG = LoggerFactory.getLogger(IcebergRESTUtils.class);
+
+  public static final String DEFAULT_SNAPSHOTS = "all";
```
roryqi: Should we use enum here?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
laserninja: Replaced

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
laserninja: Replaced

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -297,7 +291,7 @@ public Response loadTable(
           String namespace,
       @IcebergAuthorizationMetadata(type = RequestType.LOAD_TABLE) @Encoded() @PathParam("table")
           String table,
-      @DefaultValue(DEFAULT_SNAPSHOTS) @QueryParam("snapshots") String snapshots,
+      @DefaultValue("all") @QueryParam("snapshots") String snapshots,
```
roryqi: Could u use `IcebergRESTUtils.SnapshotMode.ALL.getValue()` here?

laserninja: done

roryqi (approved):
LGTM

PR #10652, opened 2026-04-01
[#10640] fix(iceberg): Remove prefix from config endpoint per Iceberg REST spec

### LINKED ISSUE #10640: [Improvement] Iceberg REST API remove the prefix from the config endpoint URL

Image

Image

We ignore the prefix. Actually, prefix represents the catalog name. So we should return catalog properties which prefix represents instead of the default catalog.

### How should we improve?

_No response_

laserninja:

Removed the `prefix` path parameter from the Iceberg REST API config endpoint. The `@Path` annotation was changed from `/v1/{prefix:([^/]*/)?}config` to `/v1/config`, so the endpoint only accepts the `warehouse` query parameter for catalog resolution.

Per the [Iceberg REST catalog OpenAPI spec](https://github.com/apache/iceberg/blob/1a2a8a5ecc701629183405ca842f2d1eb23505ee/open-api/rest-catalog-open-api.yaml#L65), the config endpoint does not include a prefix parameter. The Iceberg creator (Ryan Blue) confirmed that "a prefix is not allowed for the config endpoint." The previous implementation incorrectly accepted prefix in the URL path.

Fix: #10640

Yes. Requests to `/v1/{prefix}/config` are now rejected (the prefix segment is no longer matched). Clients should use `/v1/config?warehouse=<catalog>` for catalog-specific configuration.

- Converted parameterized tests to non-prefix versions (`testConfig`, `testConfigWithEmptyWarehouse`, `testConfigWithValidWarehouse`, `testConfigEndpointsContainViewOperations`)
- Added `testConfigRejectsPrefixInUrl` to verify that requests with prefix in the URL are rejected (returns 500 via Jersey's `IcebergExceptionMapper`)
- Removed prefix-specific tests that are no longer applicable (`testIcebergRestValidPrefix`, `testIcebergRestInvalidPrefix`)
- All tests pass: `./gradlew :iceberg:iceberg-rest-server:test --tests TestIcebergConfig`

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergConfigOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergConfigOperations.java
@@ -46,7 +46,7 @@ import org.apache.gravitino.metrics.MetricNames;
 import org.apache.iceberg.rest.Endpoint;
 import org.apache.iceberg.rest.responses.ConfigResponse;
 
-@Path("/v1/{prefix:([^/]*/)?}config")
+@Path("/v1/config")
 @Consumes(MediaType.APPLICATION_JSON)
 @Produces(MediaType.APPLICATION_JSON)
 public class IcebergConfigOperations {
--- a/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergConfig.java
+++ b/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergConfig.java
@@ -24,10 +24,10 @@ import java.util.Optional;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.Response.Status;
 import org.apache.gravitino.catalog.lakehouse.iceberg.IcebergConstants;
 import org.apache.iceberg.rest.responses.ConfigResponse;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
@@ -38,10 +38,8 @@ public class TestIcebergConfig extends IcebergTestBase {
     return IcebergRestTestUtil.getIcebergResourceConfig(IcebergConfigOperations.class);
   }
 
-  @ParameterizedTest
-  @ValueSource(strings = {"", IcebergRestTestUtil.PREFIX})
-  public void testConfig(String prefix) {
-    setUrlPathWithPrefix(prefix);
+  @Test
+  public void testConfig() {
     Response resp = getConfigClientBuilder().get();
     Assertions.assertEquals(Response.Status.OK.getStatusCode(), resp.getStatus());
     Assertions.assertEquals(MediaType.APPLICATION_JSON_TYPE, resp.getMediaType());
@@ -51,10 +49,8 @@ public class TestIcebergConfig extends IcebergTestBase {
     Assertions.assertEquals(0, response.overrides().size());
   }
 
-  @ParameterizedTest
-  @ValueSource(strings = {"", IcebergRestTestUtil.PREFIX})
-  public void testConfigWithEmptyWarehouse(String prefix) {
-    setUrlPathWithPrefix(prefix);
+  @Test
+  public void testConfigWithEmptyWarehouse() {
     Map<String, String> queryParams = ImmutableMap.of("warehouse", "");
     Response resp =
         getIcebergClientBuilder(IcebergRestTestUtil.CONFIG_PATH, Optional.of(queryParams)).get();
@@ -66,10 +62,8 @@ public class TestIcebergConfig extends IcebergTestBase {
     Assertions.assertEquals(0, response.overrides().size());
   }
 
-  @ParameterizedTest
-  @ValueSource(strings = {"", IcebergRestTestUtil.PREFIX})
-  public void testConfigWithValidWarehouse(String prefix) {
-    setUrlPathWithPrefix(prefix);
+  @Test
+  public void testConfigWithValidWarehouse() {
     String warehouseName = IcebergRestTestUtil.PREFIX;
     Map<String, String> queryParams = ImmutableMap.of("warehouse", warehouseName);
     Response resp =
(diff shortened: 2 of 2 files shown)


iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergConfig.java
```diff
@@ -104,12 +117,62 @@ public void testConfigWithNonExistentWarehouses(String warehouse) {
...
+            "true");
+    Assertions.assertEquals(expectedConfig, response.defaults());
+    Assertions.assertEquals(0, response.overrides().size());
+  }
+
+  @Test
+  public void testConfigWithPrefixAndWarehouse() {
+    // When both prefix and warehouse are specified, prefix takes precedence
+    // for catalog resolution, warehouse is used for the prefix default
+    setUrlPathWithPrefix(IcebergRestTestUtil.PREFIX);
```
bharos: The test sets both prefix and warehouse to IcebergRestTestUtil.PREFIX (same value)
The assertion response.defaults().get("prefix") == PREFIX passes regardless of which branch wins. 
Use distinct values to actually test the priority ?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergConfigOperations.java
```diff
@@ -98,13 +99,19 @@ public IcebergConfigOperations(IcebergCatalogWrapperManager catalogWrapperManage
   @Produces(MediaType.APPLICATION_JSON)
   @Timed(name = "config." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
   @ResponseMetered(name = "config", absolute = true)
-  public Response getConfig(@DefaultValue("") @QueryParam("warehouse") String warehouse) {
-    String catalogName = getCatalogName(warehouse);
+  public Response getConfig(
+      @DefaultValue("") @PathParam("prefix") String prefix,
```
bharos: @DefaultValue on a @PathParam is unusual — a path param derived from a regex like ([^/]*/)? already captures an empty string when the segment is absent. It's harmless but maybe we can remove it for clarity. The other endpoints (e.g., IcebergTableOperations) don't annotate their prefix path param with @DefaultValue

laserninja:
1. **Removed `@DefaultValue("")`** from `@PathParam("prefix")` — the regex already handles the absent segment, consistent with other endpoints like `IcebergTableOperations`.

2. **Used distinct values in `testConfigWithPrefixAndWarehouse`** — prefix is set to `IcebergRestTestUtil.PREFIX` (for catalog resolution) while warehouse is set to `"other_warehouse"` (a different string). The assertion now verifies that `response.defaults().get("prefix")` equals the warehouse value, which only passes if the warehouse branch wins for the `"prefix"` default key while the prefix path param wins for catalog resolution.

bharos (dismissed):

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergConfigOperations.java
```diff
@@ -98,13 +99,19 @@ public IcebergConfigOperations(IcebergCatalogWrapperManager catalogWrapperManage
...
   @Timed(name = "config." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
   @ResponseMetered(name = "config", absolute = true)
-  public Response getConfig(@DefaultValue("") @QueryParam("warehouse") String warehouse) {
-    String catalogName = getCatalogName(warehouse);
+  public Response getConfig(
+      @PathParam("prefix") String prefix,
+      @DefaultValue("") @QueryParam("warehouse") String warehouse) {
+    String prefixCatalogName = IcebergRESTUtils.getCatalogName(prefix);
+    String catalogName =
+        StringUtils.isNotBlank(prefix) ? prefixCatalogName : getCatalogName(warehouse);
```
roryqi: Which catalog should we use if we specified the `prefix` and `warehouse`?

roryqi: There is also some discussion about config operations behaviour. You can see https://lists.apache.org/thread/07t5rfcxn3bbbpp75fkr2h1mwdp2o4mn

laserninja: Good question. In the current implementation, when both prefix and warehouse are specified, the prefix takes precedence for catalog resolution (since it's the path-level identifier), while the warehouse value is used for the "prefix" key in the response defaults.

However, I see you've raised this same question on the 'Iceberg mailing list'. If the community consensus is different (e.g., warehouse should take precedence, or it should be an error to specify both), I'm happy to update the behavior accordingly. What's your preference here?

Thanks for the pointer, I've read through the thread. The discussion aligns with what this PR does: the prefix represents the catalog name and a 404 is returned when the specified catalog doesn't exist. I'll keep an eye on the spec PR and adjust if the finalized spec diverges from this behavior.

roryqi: I prefer that the warehouse takes precedence. I think that prefix provides the default value, warehouse can override it. Wait for the community reply.

roryqi: Iceberg creator said that `A prefix is not allowed for the config endpoint, so I don't think you would have this problem.`
So we should remove `prefix` in the url.

bharos: Thanks @roryqi for checking this, I too see that the config doesn't have prefix in the spec as well:
https://github.com/apache/iceberg/blob/1a2a8a5ecc701629183405ca842f2d1eb23505ee/open-api/rest-catalog-open-api.yaml#L65

So I guess this change is not required, just need to remove prefix parameter

laserninja:
Updated per feedback — removed prefix from the config endpoint URL entirely, since the Iceberg REST spec does not include prefix for /v1/config. The endpoint now only accepts the warehouse query parameter for catalog resolution. Added a test to verify that requests with prefix in the URL are rejected.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergConfig.java
```diff
@@ -104,26 +99,16 @@ public void testConfigWithNonExistentWarehouses(String warehouse) {
...
-  @ValueSource(strings = {"PREFIX", "", "\\\n\t\\\'", "\u0024", "\100", "[_~"})
-  void testIcebergRestValidPrefix(String prefix) {
-    String path = injectPrefixToPath(IcebergRestTestUtil.CONFIG_PATH, prefix);
+  @Test
+  public void testConfigRejectsPrefixInUrl() {
+    // Per the Iceberg REST spec, the config endpoint does not accept a prefix.
+    String path = injectPrefixToPath(IcebergRestTestUtil.CONFIG_PATH, IcebergRestTestUtil.PREFIX);
     Response response = getIcebergClientBuilder(path, Optional.empty()).get();
-    Assertions.assertEquals(Status.OK.getStatusCode(), response.getStatus());
+    Assertions.assertNotEquals(Status.OK.getStatusCode(), response.getStatus());
```
bharos: Is it possible to assert the expected status here ? I guess it might be 404 ? As injecting a prefix makes the URL `/v1/prefix_gravitino/config` which has no matching resource in JAX-RS

bharos:
PR Description is stale now

bharos:
Issue #10640 is titled "Iceberg REST API should return catalog config when prefix is specified." The PR closes it, but the resolution is different. A comment on the issue explaining the conclusion would help track why the issue was resolved this way instead of implementing the originally requested feature.

laserninja:
Addressed all three points:

1. **Status assertion**: Changed `assertNotEquals(200)` to `assertEquals(500)` with a comment explaining that Jersey's `IcebergExceptionMapper` converts the `NotFoundException` into a 500 response for unmatched routes.
2. **PR description**: Updated title and body to reflect the current approach (removing prefix, not adding it).
3. **Issue comment**: Added a comment on #10640 explaining why the resolution differs from the originally requested feature.

bharos (approved):

roryqi (approved):
Thanks for your contribution.

Issue #10600, opened 2026-03-31
[Improvement] PostgreSQL model version GC deletes using IDs from model_meta instead of model_version_info


justinmclean:
### What would you like to be improved?

In PostgreSQL, ModelVersionMetaPostgreSQLProvider.deleteModelVersionMetasByLegacyTimeline(...) deletes from model_version_info but selects id values from model_meta.

Current SQL shape:
```
DELETE FROM model_version_info
WHERE id IN (
  SELECT id FROM model_meta
  WHERE deleted_at > 0 AND deleted_at < #{legacyTimeline}
  LIMIT #{limit}
)
```

This is inconsistent with the base implementation and with similar PostgreSQL providers, which select IDs from the same table they delete from. As a result, the garbage-collection path for MODEL_VERSION can delete the wrong rows or miss rows that should be purged.

### How should we improve?

Update the PostgreSQL override to select IDs from model_version_info instead of model_meta, matching the existing provider pattern used elsewhere. For example:
```
DELETE FROM model_version_info
WHERE id IN (
  SELECT id FROM model_version_info
  WHERE deleted_at > 0 AND deleted_at < #{legacyTimeline}
  LIMIT #{limit}
)
```

babumahesh:
@justinmclean,  I would like to work on it. Can you please assign it to me.

PR #10653, opened 2026-04-01
[#10600] Improvement(core) : PostgreSQL model version GC deletes using IDs from model_version_info

babumahesh:

The PostgreSQL provider was incorrectly querying the model_meta table instead of the model_version_info table in the deletion subquery. This change corrects the table reference.

While deleting legacy model version information, the query should evaluate the legacy range using model_version_info rather than model_meta. Using the wrong table could lead to incorrect deletion behavior.

Fixes: #10600

Added unit tests to validate the corrected query behavior.


babumahesh:
@justinmclean Could you please review and let me know if any further action is required.

justinmclean:
@mchades this looks good to me, but I appreciate a second set of eyes.

mchades (commented):
LGTM. would you like to take a review? @yuqi1129

babumahesh:
@justinmclean @yuqi1129 Can we merge these changes , as its already Approved.

mchades:
Sure. Thanks for your contributions! @babumahesh

PR #10661, opened 2026-04-02
[#10659] fix(core): Correctly handle tag rename conflict in TagManager

### LINKED ISSUE #10659: [Improvement] Tag rename collision returns 500 instead of already-exists error

When changing a tag name to one that already exists in the same metalake, the server returns an internal error instead of an already exists response.

Observed code path
- TagManager.java (line 175) catches EntityAlreadyExistsException during alterTag(...) and wraps it in a generic RuntimeException.
- ExceptionHandlers.java (line 718) only maps TagAlreadyExistsException to an already-exists response.
- Generic RuntimeException falls through to internal error handling in ExceptionHandlers.java (line 1044).
Impact

A valid user mistake, renaming a tag to an existing tag name, is surfaced as HTTP 500. Clients then receive a generic runtime failure instead of a domain-specific conflict like TagAlreadyExistsException.

### How should we improve?

In TagManager.alterTag(...), convert EntityAlreadyExistsException to TagAlreadyExistsException, similar to createTag(...), for example:

```
} catch (EntityAlreadyExistsException e) {
  throw new TagAlreadyExistsException(
      e, "Tag with name %s under metalake %s already exists", newName, metalake);
}
```

Here's a unit test to help:
```
  @Test
  public void testAlterTagRenameToExistingTag() {
    tagManager.createTag(METALAKE, "tag1", null, null);
    tagManager.createTag(METALAKE, "tag2", null, null);

    TagAlreadyExistsException exception =
        Assertions.assertThrows(
            TagAlreadyExistsException.class,
            () -> tagManager.alterTag(METALAKE, "tag1", TagChange.rename("tag2")));

    Assertions.assertEquals(
        "Tag with name tag2 under metalake metalake_for_tag_test already exists",
        exception.getMessage());

    Assertions.assertEquals("tag1", tagManager.getTag(METALAKE, "tag1").name());
    Assertions.assertEquals("tag2", tagManager.getTag(METALAKE, "tag2").name());
  }

```

ChisomUma:

This PR improves the error handling in `TagManager.alterTag` when a tag is renamed to a name that already exists in the same metalake. 

Specifically:
- Caught `EntityAlreadyExistsException` during the [alterTag] TagManager.java operation.
- Added logic to extract the intended `newName` from the [TagChange] TagChange.java array.
- Now throws a domain-specific `TagAlreadyExistsException` with a clear message, ensuring the server returns an HTTP 409 (Conflict) instead of an HTTP 500 (Internal Server Error).

Previously, renaming a tag to an existing name caused the server to return a generic `RuntimeException`, which surfaced as an HTTP 500 error to the client. This is a valid user mistake that should be handled as a conflict (HTTP 409) with a descriptive error message.

Fix: #10659

Yes, it changes the error response when a tag rename conflict occurs:
- **Old behavior**: Returns HTTP 500 (Internal Server Error).
- **New behavior**: Returns HTTP 409 (Conflict) with the message: `"Tag with name <newName> under metalake <metalake> already exists"`.

Added a new unit test [testAlterTagRenameToExistingTag] TestTagManager.java `core/src/test/java/org/apache/gravitino/tag/TestTagManager.java` that specifically covers this scenario:
1. Creates two tags.
2. Attempts to rename the first tag to the second's name.
3. Asserts that `TagAlreadyExistsException` is thrown with the correct message.
4. Verifies that the state of both tags remains unchanged.

Verified the fix by running:
`./gradlew :core:test --tests "org.apache.gravitino.tag.TestTagManager" -PskipITs`


yuqi1129:
Please adjust your PR title and description based on the template. You can refer to how other PR do it.

ChisomUma:
Updated @yuqi1129

core/src/main/java/org/apache/gravitino/tag/TagManager.java
jerryshao, line 160: The reason why we don't throw `EntityAlreadyExistsException` and throw `RuntimeException` instead is that we don't define `EntityAlreadyExistsException` in the API signature. If we change to throw this exception, the API signature will be broken. So if we want to throw this exception, we should basically update the API signature, not just simply throwing an exception here.

ChisomUma: Thanks for the feedback. I've updated the PR to integrate the exception for TagAlreadyExistsException in the API signature. Please take a look.

jerryshao: Changing the API signature will break compatibility. You need to deal with it carefully. It's not just simply modifying the existing API.

justinmclean: TagAlreadyExistsException is already unchecked, so you don’t need to change the public throws signatures here. The correct fix is to keep the API unchanged and only convert the EntityAlreadyExistsException in TagManager.alterTag(...) into TagAlreadyExistsException, so the existing REST handler returns 409 instead of 500.

jerryshao: No, I don't agree. The client SDK will get an undefined exception if we throw `TagAlreadyExistsException` without a definition. This is not a REST protocol alone problem. This is a semantic change that should also align with the API/client implementation. Otherwise, the user will get a `TagAlreadyExistsException` that they don't expect to throw from the client call.

api/src/main/java/org/apache/gravitino/tag/TagOperations.java
```diff
@@ -78,9 +78,10 @@ Tag createTag(String name, String comment, Map<String, String> properties)
    * @return The altered tag.
    * @throws NoSuchTagException If the tag does not exist.
    * @throws IllegalArgumentException If the changes cannot be applied to the tag.
+   * @throws TagAlreadyExistsException If a tag with the new name already exists.
    */
   Tag alterTag(String name, TagChange... changes)
-      throws NoSuchTagException, IllegalArgumentException;
+      throws NoSuchTagException, IllegalArgumentException, TagAlreadyExistsException;
```
jerryshao: This change will break the API backward compatibility. Please figure out a more proper way like using `@deprecrated`.

jerryshao (changes requested):

api/src/main/java/org/apache/gravitino/tag/TagOperations.java
ChisomUma: I've reverted the changes to the API signatures, and instead Since TagAlreadyExistsException is an unchecked exception that extends RuntimeException, it should be possible to throw without being declared in the method signature. So I added @ throws TagAlreadyExistsException to the JavaDoc for TagOperations#alterTag and TagDispatcher#alterTag.

Given that the REST handlers and Client SDK already support the reconstruction of TagAlreadyExistsException (mapping HTTP 409 back to the specific exception type), Is this JavaDoc-only documentation of the unchecked exception sufficient or do you have a more preferred pattern in mind? @jerryshao

jerryshao: @ChisomUma I think the current change is OK for me. One more thing that I think we also need to update the python-client as the same way.

ChisomUma: Hi @jerryshao I've updated the python-client docstrings for the `alter_tag` method to indicate that `TagAlreadyExistsException` and `IllegalArgumentException` can be raised.

ChisomUma:
Hi @jerryshao How are you? Just checking in on any updates for this PR?

jerryshao:
I'm OK with the current change. Can you fix the CI failure?

ChisomUma:
> I'm OK with the current change. Can you fix the CI failure?

Awesome will work on it asap…will the PR get merged in?

ChisomUma:
> I'm OK with the current change. Can you fix the CI failure?

Hi @jerryshao I have fixed CI issue

jerryshao (approved):

Issue #10620, opened 2026-04-01
[Improvement] Lance partition statistics drop API reports success even when nothing is deleted


justinmclean:
### What would you like to be improved?

When Gravitino is configured to use org.apache.gravitino.stats.storage.LancePartitionStatisticStorageFactory for partition statistics storage, dropping partition statistics can return success even if no statistics were actually removed.

In LancePartitionStatisticStorage.java, dropStatistics(...) always returns 1 as a placeholder:

That value is later converted to a boolean in StatisticManager.java, where != 0 becomes true, and the REST API returns DropResponse(true).

This means clients can get a successful drop result even when:
- the target partition does not exist
- the statistic name does not exist
- nothing was actually deleted

### How should we improve?

Do not return a hard-coded success count from the Lance backend.

Possible fixes:
- Change the Lance dropStatistics(...) contract so it returns an accurate deleted count.
- If Lance cannot provide counts, change the higher-level API to avoid interpreting a placeholder count as success.
- As a safe short-term fix, have the Lance path return 0 only when the request is empty and otherwise explicitly verify whether matching rows existed before deletion, then return a real count or a trustworthy boolean.

yunhwane:
Hi @justinmclean, I'd like to work on this issue. Could you please assign it to me?

PR #10662, opened 2026-04-02
[#10620] fix(core): Return accurate drop count from LancePartitionStatisticStorage

yunhwane:

- Replace the hardcoded `return 1` in `LancePartitionStatisticStorage.dropStatistics()` with an actual count of matching rows before deletion.
- Change `dropStatisticsImpl` from `void` to `int`, using a `countMatchingRows()` helper that scans the Lance dataset to count matching rows before performing the delete.
- Skip the delete operation entirely when no matching rows exist.

`LancePartitionStatisticStorage.dropStatistics()` always returned `1` as a hardcoded placeholder. `StatisticManager.dropPartitionStatistics()` converts this to a boolean via `!= 0`, so the REST API always reported a successful drop — even when:
- The target partition does not exist
- The statistic name does not exist
- Nothing was actually deleted

Fix: #10620

The `DROP` partition statistics REST API now correctly returns `false` when no statistics were actually deleted, instead of always returning `true`.

- Updated existing tests in `TestLancePartitionStatisticStorage` to verify the return value of `dropStatistics()` (1 for single drop, 2 for multi-partition drop).
- Added a new test case that drops non-existent statistics and asserts the return value is `0`.
- All existing tests continue to pass: `./gradlew :core:test --tests "org.apache.gravitino.stats.storage.TestLancePartitionStatisticStorage"`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
+++ b/core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
@@ -361,7 +361,6 @@ public class LancePartitionStatisticStorage implements PartitionStatisticStorage
     }
   }
 
-
   @Override
   public void close() throws IOException {
     if (datasetCache.isPresent()) {


yunhwane:
@roryqi Replaced `countMatchingRows()` (manual `LanceScanner` + `ArrowReader` scan) with `Dataset.countRows(filter)` to address the performance concern.

This avoids Arrow batch materialization overhead and lets Lance optimize the count internally.

justinmclean:
You'll need to run `./gradlew :core:spotlessApply` to fix the formatting issues and push your changes.

yunhwane:
@justinmclean Thanks for the review! I've run `./gradlew :core:spotlessApply` and pushed the formatting fix.

roryqi:
> @roryqi Replaced `countMatchingRows()` (manual `LanceScanner` + `ArrowReader` scan) with `Dataset.countRows(filter)` to address the performance concern.
> 
> This avoids Arrow batch materialization overhead and lets Lance optimize the count internally.

Still disagree to approve to merge this pull request. It will include unnecessary cost. I don't think this is meaningful change. I will close this pull request.

Issue #10660, opened 2026-04-02
[Improvement] Invalid model version update requests can bypass REST exception handling and return an internal error


justinmclean:
### What would you like to be improved?

ModelOperations.alterModelVersion and alterModelVersionByAlias call request.validate() before entering their try/catch.

Relevant paths:
/server/src/main/java/org/apache/gravitino/server/web/rest/ModelOperations.java
/common/src/main/java/org/apache/gravitino/dto/requests/ModelVersionUpdatesRequest.java
/server/src/main/java/org/apache/gravitino/server/web/rest/ExceptionHandlers.java

ModelVersionUpdatesRequest.validate() throws IllegalArgumentException for invalid input. Because that validation happens before the endpoint enters the try, the exception does not reach ExceptionHandlers.handleModelException(...), which is the code path that maps IllegalArgumentException to a 400 Bad Request.

The server registers JSON parse/mapping exception mappers, but not a general IllegalArgumentException mapper, so this invalid request can fall into Jersey’s unhandled exception path instead of returning the endpoint’s intended bad-request response.

This affects:
PUT /.../{model}/versions/{version}
PUT /.../{model}/aliases/{alias}

### How should we improve?

Move the outer request.validate() call inside the try block in both methods, or remove the duplicate pre-try validation and keep the validation already performed inside Utils.doAs(...).

A minimal safe fix is:
- keep a single request.validate() inside the try path before business logic
- ensure both endpoints route validation failures through ExceptionHandlers.handleModelException(...)

Here's a unit test to help:
```
  @Test
  void testUpdateModelVersionWithInvalidRequest() {
    String invalidRequest = "{\"updates\":null}";

    Response resp =
        target(modelPath())
            .path("model1")
            .path("versions")
            .path("0")
            .request(MediaType.APPLICATION_JSON_TYPE)
            .accept("application/vnd.gravitino.v1+json")
            .put(Entity.entity(invalidRequest, MediaType.APPLICATION_JSON_TYPE));

    Assertions.assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), resp.getStatus());

    ErrorResponse errorResponse = resp.readEntity(ErrorResponse.class);
    Assertions.assertEquals(ErrorConstants.ILLEGAL_ARGUMENTS_CODE, errorResponse.getCode());
    Assertions.assertEquals(
        IllegalArgumentException.class.getSimpleName(), errorResponse.getType());
[...]

PR #10663, opened 2026-04-02
[#10660] fix: Move model version request validation inside try block

yunhwane:

Move `request.validate()` inside the `try` block in `alterModelVersion` and `alterModelVersionByAlias` methods of `ModelOperations.java`, so that `IllegalArgumentException` from invalid requests is properly caught by `ExceptionHandlers.handleModelException()`.

`request.validate()` was called before the `try/catch` block in both `alterModelVersion` and `alterModelVersionByAlias`. When validation fails (e.g., `{"updates": null}`), the `IllegalArgumentException` bypasses `ExceptionHandlers.handleModelException()` and results in a 500 Internal Server Error instead of the expected 400 Bad Request.

The duplicate `request.validate()` call inside the `try` block (within the lambda) already exists, so removing the pre-try call is sufficient.

Fix: #10660

No. This fixes error response codes for invalid REST API requests from 500 to 400.

Added two unit tests:
- `testAlterModelVersionWithNullUpdates`: Sends `{"updates":null}` to the version endpoint, asserts 400 Bad Request with `ILLEGAL_ARGUMENTS_CODE`
- `testAlterModelVersionByAliasWithNullUpdates`: Sends `{"updates":null}` to the alias endpoint, asserts 400 Bad Request with `ILLEGAL_ARGUMENTS_CODE`

All existing `TestModelOperations` tests continue to pass.


yunhwane:
@justinmclean Could you please review this PR? Thank you!

jerryshao:
Can you fix all the similar issues, I guess there should have similar issues in other places.

yunhwane:
Thanks for the review! I've moved request validations inside `Utils.doAs()` in the remaining places:
- `ModelOperations#registerModel()`
- `ModelOperations#linkModelVersion()`
- `PartitionOperations#addPartitions()`

jerryshao (approved):

PR #10664, opened 2026-04-02
[#10657] fix(core): Add missing @Param("metalakeId") to GroupRoleRelMapper.softDeleteGroupRoleRelByMetalakeId

### LINKED ISSUE #10657: [Improvement] Add missing MyBatis @Param("metalakeId") to GroupRoleRelMapper.softDeleteGroupRoleRelByMetalakeId

GroupRoleRelMapper.softDeleteGroupRoleRelByMetalakeId(Long metalakeId) is used in the metalake deletion path, but unlike the neighboring soft-delete mappers it does not declare @Param("metalakeId") on the mapper method.

The SQL providers for this path reference #{metalakeId}:
GroupRoleRelBaseSQLProvider.java
GroupRoleRelPostgreSQLProvider.java

The mapper is invoked from the metalake delete flow in MetalakeMetaService.java.

Relying on the runtime-retained Java parameter name is fragile. This can cause MyBatis named-parameter binding failures during metalake deletion.

### How should we improve?

Add @Param("metalakeId") to:
GroupRoleRelMapper.softDeleteGroupRoleRelByMetalakeId

Optionally, also the corresponding factory/provider method parameters for consistency.

sachinnn99:

Add `@Param("metalakeId")` annotation to `softDeleteGroupRoleRelByMetalakeId` in the mapper, SQL provider factory, base SQL provider, and PostgreSQL provider.

`GroupRoleRelMapper.softDeleteGroupRoleRelByMetalakeId(Long metalakeId)` is missing the `@Param("metalakeId")` annotation, unlike every other soft-delete method in the same mapper. The SQL templates reference `#{metalakeId}`, so without the annotation MyBatis relies on runtime-retained parameter names, which is fragile and can cause binding failures during metalake deletion.

Fix: #10657

No.

Existing unit and integration tests cover the metalake deletion path that invokes this mapper method. The change is additive — it makes the parameter binding explicit rather than implicit.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/GroupRoleRelBaseSQLProvider.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/GroupRoleRelBaseSQLProvider.java
@@ -97,7 +97,7 @@ public class GroupRoleRelBaseSQLProvider {
         + "</script>";
   }
 
-  public String softDeleteGroupRoleRelByMetalakeId(Long metalakeId) {
+  public String softDeleteGroupRoleRelByMetalakeId(@Param("metalakeId") Long metalakeId) {
     return "UPDATE "
         + GROUP_ROLE_RELATION_TABLE_NAME
         + " SET deleted_at = (UNIX_TIMESTAMP() * 1000.0)"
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/GroupRoleRelBaseSQLProvider.java
yuqi1129: I would suggest you solve all similars problem.

sachinnn99 (commented):
Thank you, I'll look into similar issues and update this PR.

sachinnn99:
@yuqi1129 Updated to fix all similar missing `@Param` annotations across the mapper interfaces. Could you please take another look?

yuqi1129 (approved):

sachinnn99:
Thanks for the review and merge, @yuqi1129!

Issue #10658, opened 2026-04-02
[Improvement] Preserve original removeGroup exception when failure-event listener throws


justinmclean:
### What would you like to be improved?

In AccessControlEventDispatcher#removeGroup, when dispatcher.removeGroup(...) throws, the code enters the catch block and dispatches RemoveGroupFailureEvent.

If eventBus.dispatchEvent(...) also throws because a registered listener fails while handling that failure event, the listener exception replaces the original removeGroup exception. This hides the real root cause from callers and makes production failures harder to diagnose.

Relevant path:
/gravitino/core/src/main/java/org/apache/gravitino/listener/AccessControlEventDispatcher.java:246

### How should we improve?

Protect failure-event dispatch so it cannot mask the original exception and log the new exception.

Here's a unit test to help:
```
@Test
  void testRemoveGroupFailureEventShouldNotMaskOriginalExceptionWhenListenerThrows() {
    GravitinoRuntimeException originalException =
        new GravitinoRuntimeException("Original remove group failure");
    GravitinoRuntimeException listenerException =
        new GravitinoRuntimeException("Failure listener exception");
    AccessControlDispatcher exceptionDispatcher = mock(AccessControlDispatcher.class);
    when(exceptionDispatcher.removeGroup(METALAKE, groupName)).thenThrow(originalException);

    EventBus eventBus =
        new EventBus(
            Collections.singletonList(
                new EventListenerPlugin() {
                  @Override
                  public void init(Map<String, String> properties) {}

                  @Override
                  public void start() {}

                  @Override
                  public void stop() {}

                  @Override
                  public void onPostEvent(Event event) {
                    if (event instanceof RemoveGroupFailureEvent) {
                      throw listenerException;
                    }
                  }
                }));
    AccessControlEventDispatcher testDispatcher =
        new AccessControlEventDispatcher(eventBus, exceptionDispatcher);

    GravitinoRuntimeException thrownException =
        Assertions.assertThrowsExactly(
            GravitinoRuntimeException.class, () -> testDispatcher.removeGroup(METALAKE, groupName));

    Assertions.assertSame(originalException, thrownException);
  }
```

babumahesh:
Hi @justinmclean I would like to work on it. Can you please assign it to me

PR #10665, opened 2026-04-02
[#10658] Improvement(core) : preserve original exception when EventBus listener fails

babumahesh:

Enhanced EventBus.dispatchEvent() to automatically detect and handle FailureEvent instances with safe exception swallowing. The method now uses type-based dispatch logic:

  1. PreEvents: May throw ForbiddenException for authorization control
  2. FailureEvents: Automatically swallows ALL listener exceptions to preserve the original error
  3. Regular Events: Propagates listener exceptions normally

                                                                                                                                                                                                                           
  When `dispatcher.removeGroup()` (or any other access control operation) throws an exception, the code enters a catch block and dispatches a `FailureEvent`. If `eventBus.dispatchEvent(FailureEvent)` also throws because
   a registered listener fails while handling that failure event, the listener exception replaces the original exception. This hides the real root cause from callers and makes production failures harder to diagnose.
                                                                                                                                                                                                                           
  Now; listener exceptions of FailureEvent are caught and logged without masking the original exception that triggered the failure.                                                          
   
  Fixes: #10658                                                                                                                                                                                                            
                  
  ## How was this patch tested?                                                                                                                                                                                            
                  
  - Added `testRemoveGroupFailureEventDoesNotMaskOriginalException()` in `TestGroupEvent` to verify group operations and the original exception is propagated (not masked by listener exceptions) 
- Added tests to verify exception-handling semantics across event types:
  - `FailureEvent`: Listener exceptions (e.g., `RuntimeException`) are caught, logged, and swallowed to avoid masking the original failure.
  - `PreEvent`: Listener exceptions (eg : `ForbiddenException`) are propagated, preserving authorization and control-flow behavior.
  - Regular (success) `Event`: Listener exceptions are propagated normally and not swallowed.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/listener/EventBus.java
+++ b/core/src/main/java/org/apache/gravitino/listener/EventBus.java
@@ -137,7 +137,7 @@ public class EventBus {
    * (including {@link RuntimeException}, {@link Error}, etc.) to ensure the original exception can
    * be properly thrown to the caller.
    *
-   * @param failureEvent the failure failureEvent to dispatch
+   * @param failureEvent the failure event to dispatch
    */
   private void dispatchFailureEvent(Event failureEvent) {
     try {
@@ -145,7 +145,7 @@ public class EventBus {
     } catch (Exception e) {
       // Swallow ALL listener exceptions to prevent masking the original error
       LOG.error(
-          "Failed to dispatch failure failureEvent: {}, ignoring exception to preserve original error",
+          "Failed to dispatch failure event: {}, ignoring exception to preserve original error",
           failureEvent.getClass().getSimpleName(),
           e);
     }


babumahesh:
@justinmclean 

Introduced dispatchEventSafe() as a wrapper around dispatchEvent(). This ensures event/listener exceptions are swallowed (and logged) during event dispatch, preventing them from masking the original failure.

This behavior now applies to all failure events such as removeGroupEvent, AddUserFailureEvent, RemoveUserFailureEvent, ListUserNamesFailureEvent, etc., and is handled in this PR.

Could you please review and let me know if it's appropriate to apply this universally, or if it should be limited to specific cases like RemoveGroup where it is explicitly required.

babumahesh:
@justinmclean @yuqi1129 

I have enhanced the `dispatchEvent()` function to explicitly handle `FailureEvent`s by catching and swallowing listener exceptions while logging them appropriately.

For all other event types, the existing behavior remains unchanged:
- `PreEvent` continues to propagate `ForbiddenException` for control-flow decisions.
- Regular `Event` continues to propagate listener exceptions as before.

An alternative approach would have been to introduce a separate `dispatchFailureEvent()` or  `dispatchEventSafe()` method and update all dispatcher classes to use it for failure flows. However, I chose to handle this within `dispatchEvent()` itself to avoid changes at the call sites and keep the integration simpler.

justinmclean (approved):
Thanks for this.  This fixes #10658 in the right place by handling FailureEvent centrally in EventBus, and the added regression test covers the original removeGroup masking scenario.

Issue #10667, opened 2026-04-03
[Improvement] Iceberg REST server returns HTML instead of JSON error body for 401 Unauthorized responses


laserninja:
### What would you like to be improved?

The Gravitino Iceberg REST server returns non-JSON (HTML) error responses when authentication fails, violating the Iceberg REST API specification.

**Problem:**
When authentication fails, the `AuthenticationFilter` calls `resp.sendError(HttpServletResponse.SC_UNAUTHORIZED, message)` ([AuthenticationFilter.java#L102-L105](https://github.com/apache/gravitino/blob/main/server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java#L102-L105)). This happens at the servlet filter level — **before** the request reaches JAX-RS — so the `IcebergExceptionMapper` is never invoked. Instead, Jetty's default `ErrorHandler` ([JettyServer.java#L101](https://github.com/apache/gravitino/blob/main/server-common/src/main/java/org/apache/gravitino/server/web/JettyServer.java#L101)) produces an HTML error page.

**Expected behavior per the Iceberg REST spec:**
All error responses (including 401) must return a JSON body with the `IcebergErrorResponse` schema:
```json
{
  "error": {
    "message": "Not authorized to make this request",
    "type": "NotAuthorizedException",
    "code": 401
  }
}
```

**Actual behavior:**
Jetty returns its default HTML error page (or plain text), which causes Iceberg REST clients (e.g., the Java `RESTCatalog`) to fail with a secondary JSON parse error, masking the real authentication failure.

### How should we improve?

Replace the `sendError()` calls in the filter with writing a proper JSON `IcebergErrorResponse` body directly to the response output stream, setting `Content-Type: application/json` and the 401 status code. Alternatively, register a custom Jetty `ErrorHandler` for the Iceberg REST server that formats errors as JSON.

PR #10668, opened 2026-04-03
[#10667] fix(iceberg): Return JSON error body instead of HTML for pre-JAX-RS errors

laserninja:

Override authentication error handling in the Iceberg REST server to produce JSON error bodies conforming to the Iceberg REST API specification, instead of Jetty's default HTML error pages.

Changes:
- Added `sendAuthErrorResponse(HttpServletResponse, Exception)` hook to `AuthenticationFilter` (base impl: `resp.sendError()`).
- Created `IcebergAuthenticationFilter` that overrides the hook to write Iceberg-spec JSON `ErrorResponse` bodies.
- Added `convertToIcebergException(Exception)` to `IcebergExceptionMapper` that converts Gravitino/generic exceptions to Iceberg spec exceptions (e.g., `UnauthorizedException` → `NotAuthorizedException`, `IllegalArgumentException` → `BadRequestException`).
- Made `JettyServer` non-final with a `createAuthenticationFilter()` factory method; `RESTService` overrides it to use `IcebergAuthenticationFilter`.
- Added `TestIcebergAuthenticationFilter` with 4 test cases.

When authentication fails, `AuthenticationFilter` runs at the servlet filter level — before the request reaches JAX-RS — so `IcebergExceptionMapper` is never invoked. Jetty's default `ErrorHandler` produces an HTML error page, violating the Iceberg REST API specification which requires all errors to be JSON `ErrorResponse` bodies. This causes Iceberg REST clients (e.g., Java `RESTCatalog`) to fail with a secondary JSON parse error, masking the real authentication failure.

Fix: #10667

No API changes. Error responses from the Iceberg REST server for pre-JAX-RS failures (e.g., 401 Unauthorized) will now be properly formatted JSON instead of HTML, which is what clients already expect.

Added `TestIcebergAuthenticationFilter` covering:
1. `UnauthorizedException` → 401 with `NotAuthorizedException` type
2. `RuntimeException` → 500 with `ServiceFailureException` type
3. `ForbiddenException` → 403 with `ForbiddenException` type
4. Null message → falls back to HTTP status message

All existing tests in server-common and iceberg-rest-server continue to pass.

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
@@ -28,6 +28,7 @@ import org.apache.gravitino.exceptions.IllegalNameIdentifierException;
 import org.apache.gravitino.exceptions.NoSuchCatalogException;
 import org.apache.gravitino.exceptions.UnauthorizedException;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
+import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.CommitStateUnknownException;
 import org.apache.iceberg.exceptions.ForbiddenException;
@@ -99,6 +100,12 @@ public class IcebergExceptionMapper implements ExceptionMapper<Exception> {
     }
     if (e instanceof org.apache.gravitino.exceptions.ForbiddenException) {
       return new ForbiddenException("%s", message);
+    }
+    if (e instanceof IllegalArgumentException
+        || e instanceof IllegalNameIdentifierException
+        || e instanceof ValidationException
+        || e instanceof NamespaceNotEmptyException) {
+      return new BadRequestException("%s", message);
     }
     if (EXCEPTION_ERROR_CODES.containsKey(e.getClass())) {
       return e;
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
@@ -20,7 +20,6 @@ package org.apache.gravitino.iceberg.service;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
 import java.time.Instant;
 import java.time.LocalDateTime;
@@ -31,7 +30,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.Stream;
 import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
@@ -43,25 +41,6 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.rest.responses.ErrorResponse;
 
 public class IcebergRESTUtils {
-
-  /**
-   * Error type names matching the Iceberg REST API specification examples. See <a
-   * href="https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml">Iceberg
-   * REST API spec</a>.
-   */
-  public static final Map<Integer, String> ERROR_TYPE_NAMES =
-      ImmutableMap.<Integer, String>builder()
-          .put(HttpServletResponse.SC_BAD_REQUEST, "BadRequestException")
-          .put(HttpServletResponse.SC_UNAUTHORIZED, "NotAuthorizedException")
-          .put(HttpServletResponse.SC_FORBIDDEN, "ForbiddenException")
-          .put(HttpServletResponse.SC_NOT_FOUND, "NoSuchResourceException")
-          .put(HttpServletResponse.SC_CONFLICT, "AlreadyExistsException")
-          .put(HttpServletResponse.SC_NOT_ACCEPTABLE, "UnsupportedOperationException")
-          .put(422, "UnprocessableEntityException")
-          .put(419, "AuthenticationTimeoutException")
-          .put(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "SlowDownException")
-          .put(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "InternalServerError")
-          .build();
 
   private IcebergRESTUtils() {}
 
@@ -82,11 +61,10 @@ public class IcebergRESTUtils {
   }
 
   public static Response errorResponse(Throwable ex, int httpStatus) {
-    String type = ERROR_TYPE_NAMES.getOrDefault(httpStatus, ex.getClass().getSimpleName());
     ErrorResponse errorResponse =
         ErrorResponse.builder()
             .responseCode(httpStatus)
-            .withType(type)
+            .withType(ex.getClass().getSimpleName())
             .withMessage(ex.getMessage())
             .withStackTrace(ex)
             .build();


roryqi:
Actually, we should follow the Iceberg response spec. You can see https://github.com/apache/iceberg/blob/1a2a8a5ecc701629183405ca842f2d1eb23505ee/open-api/rest-catalog-open-api.yaml#L4827

Another question, authenticator will be used for Gravitino REST API. It should follow the Gravitino error response spec.

laserninja:
1. **Iceberg spec compliance**: The `type` field in error responses now uses spec-matching exception type names (e.g., `NotAuthorizedException` for 401, `InternalServerError` for 500) instead of generic HTTP status text, matching the [Iceberg REST API specification examples](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml).

2. **Gravitino REST API isolation**: Our approach (custom Jetty `ErrorHandler` registered only in `RESTService`) doesn't affect the `AuthenticationFilter` or the Gravitino REST API. The Gravitino server could separately add its own error handler following Gravitino's error response spec (which uses different numeric codes like `1000`, `1001`, etc.) — but that's out of scope for this Iceberg-specific issue.

roryqi:
> 1. **Iceberg spec compliance**: The `type` field in error responses now uses spec-matching exception type names (e.g., `NotAuthorizedException` for 401, `InternalServerError` for 500) instead of generic HTTP status text, matching the [Iceberg REST API specification examples](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml).
> 2. **Gravitino REST API isolation**: Our approach (custom Jetty `ErrorHandler` registered only in `RESTService`) doesn't affect the `AuthenticationFilter` or the Gravitino REST API. The Gravitino server could separately add its own error handler following Gravitino's error response spec (which uses different numeric codes like `1000`, `1001`, etc.) — but that's out of scope for this Iceberg-specific issue.

Maybe we should throw exception in the `Authenticator` directly and use different ExceptionHandler to handle them instead of adding error handler.

laserninja:
Thanks for the suggestion! I looked into throwing exceptions from the filter and catching them with a JAX-RS `ExceptionMapper`, but that doesn't work because `AuthenticationFilter` is a servlet `Filter` — it runs *before* the request reaches JAX-RS/Jersey. Exceptions thrown at the filter level propagate to Jetty's servlet container, not to `IcebergExceptionMapper`.

Instead, I refactored to handle it directly in the filter:

- Added a `protected sendAuthErrorResponse()` hook to `AuthenticationFilter` (default: `resp.sendError()`, preserving current Gravitino behavior).
- Created `IcebergAuthenticationFilter` that overrides it to write Iceberg-spec JSON directly.
- Overloaded `JettyServer.addSystemFilters(pathSpec, authFilter)` so the Iceberg server passes its own filter.
- Removed the Jetty `ErrorHandler` approach entirely.

This handles errors at the source and lets the Gravitino server easily add its own error format later via a similar subclass.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
```diff
@@ -0,0 +1,75 @@
...
+ * subclass overrides the error response to write a proper Iceberg {@link ErrorResponse} JSON body,
+ * which Iceberg REST clients (e.g., the Java {@code RESTCatalog}) expect.
+ */
+public class IcebergAuthenticationFilter extends AuthenticationFilter {
+
+  private static final ObjectMapper MAPPER = IcebergObjectMapper.getInstance();
+
+  // Error type names matching the Iceberg REST API specification examples.
+  // See https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml
+  static final Map<Integer, String> ERROR_TYPE_NAMES =
```
roryqi: Could u extract this constant? IcebergExceptionMapper can reuse this,too.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
```diff
@@ -0,0 +1,75 @@
...
+          .put(422, "UnprocessableEntityException")
+          .put(419, "AuthenticationTimeoutException")
+          .put(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "SlowDownException")
+          .put(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "InternalServerError")
+          .build();
+
+  @Override
+  protected void sendAuthErrorResponse(HttpServletResponse response, int status, String message)
+      throws IOException {
+    if (message == null || message.isEmpty()) {
```
roryqi: StringUtils.isNotBlank.

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
```diff
@@ -99,13 +99,26 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+  /**
+   * Sends an error response when authentication fails. Subclasses can override this to customize
+   * the error response format (e.g., Iceberg REST server returns JSON error bodies).
+   *
+   * @param response the HTTP servlet response
+   * @param status the HTTP status code
+   * @param message the error message
+   */
+  protected void sendAuthErrorResponse(HttpServletResponse response, int status, String message)
+      throws IOException {
```
roryqi, line 120: Add TODO:
Gravitino should override this method to return a correct json response.

server-common/src/main/java/org/apache/gravitino/server/web/JettyServer.java
```diff
@@ -466,10 +466,14 @@ public void addCustomFilters(String pathSpec) {
   }
 
   public void addSystemFilters(String pathSpec) {
+    addSystemFilters(pathSpec, new AuthenticationFilter());
```
roryqi: I prefer add a protected method `createAuthenticationFilter`.
IcebergREST can use the anonymous class to create it
```
new JettyServer() {
   Filter createAuthentionFilter {
       return IcebergAuthenticationFilter();
   }
}
```

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
laserninja: **`ERROR_TYPE_NAMES` extracted** — Moved to `IcebergRESTUtils` as a `public static final` constant so `IcebergExceptionMapper` can reuse it too.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
laserninja: **`StringUtils.isBlank()`** — Replaced `message == null || message.isEmpty()` with `StringUtils.isBlank(message)` in `IcebergAuthenticationFilter`.

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
laserninja: **TODO added** — Added `TODO: Gravitino server should override this method to return a correct JSON response following the Gravitino error response spec` to `sendAuthErrorResponse` Javadoc.

server-common/src/main/java/org/apache/gravitino/server/web/JettyServer.java
laserninja: **Factory method pattern** — Replaced `addSystemFilters(pathSpec, authFilter)` overload with `protected createAuthenticationFilter()` in `JettyServer`. `RESTService` now creates `JettyServer` via anonymous subclass that overrides this method to return `IcebergAuthenticationFilter`. Removed `final` from `JettyServer` to enable subclassing.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
roryqi: Could u modify IcebergExceptionMapper code, too? Maybe you should use Guava BiMap.

laserninja: Updated `IcebergRESTUtils.errorResponse(Throwable, int)` to use `ERROR_TYPE_NAMES` for the `type` field (falling back to `ex.getClass().getSimpleName()` for unmapped codes). This means `IcebergExceptionMapper` now automatically produces spec-compliant type names like `BadRequestException` for 400 instead of `IllegalArgumentException`.

I think, BiMap won't work here because the mappings aren't bijective (multiple exception classes map to the same status code in `EXCEPTION_ERROR_CODES`, and codes 401/403 both map to "NotAuthorizedException" in `ERROR_TYPE_NAMES`).

roryqi: I got your point.  

We should reuse the logic and unify them.

Maybe we can refactor method `sendAuthErrorResponse`.
```
 sendAuthErrorResponse(HttpServletResponse response,  xxxx exception);
```
We can throw another exception like AuthenticationTimeoutException for the error code 403.

laserninja: Refactored `sendAuthErrorResponse` to accept an `Exception` instead of `(int status, String message)`:

- `AuthenticationFilter.sendAuthErrorResponse(HttpServletResponse, Exception)` — base impl resolves `UnauthorizedException` → 401, others → 500
- `IcebergAuthenticationFilter` overrides it and calls `IcebergExceptionMapper.getErrorCode(exception)` to resolve the status code, reusing the same `EXCEPTION_ERROR_CODES` mapping used by the JAX-RS path
- Added `UnauthorizedException` → 401 to `EXCEPTION_ERROR_CODES` and extracted a `public static getErrorCode(Exception)` method so the filter can share it
- Tests updated to pass exception objects; added a `ForbiddenException` → 403 test case

This unifies the exception → status code → type name resolution between both paths. Adding `AuthenticationTimeoutException` for expired tokens (as mentioned in #10672) can build on top of this, just add the exception class to `EXCEPTION_ERROR_CODES` and `ERROR_TYPE_NAMES`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -42,6 +44,25 @@
...
+  /**
+   * Error type names matching the Iceberg REST API specification examples. See <a
+   * href="https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml">Iceberg
+   * REST API spec</a>.
+   */
+  public static final Map<Integer, String> ERROR_TYPE_NAMES =
+      ImmutableMap.<Integer, String>builder()
+          .put(HttpServletResponse.SC_BAD_REQUEST, "BadRequestException")
+          .put(HttpServletResponse.SC_UNAUTHORIZED, "NotAuthorizedException")
+          .put(HttpServletResponse.SC_FORBIDDEN, "NotAuthorizedException")
```
roryqi: 403, Iceberg use ForbiddenException instead of NotAuthorizedException

roryqi: I hesitate that we should adopt this solution. 
I think we should convert Gravitino exceptions or other exceptions to Iceberg exception. WDYT?
We can add a method `convertToIcebergSpecException(int status, Exception e)`.

roryqi: Now we only convert status to the correct string. Actually, what we want to is convert underlying layer exception to upper layer exception. It would better that we convert the exceptions explictly.

laserninja: 1. **403 → "ForbiddenException"**: Fixed `ERROR_TYPE_NAMES` to map 403 to `"ForbiddenException"` instead of `"NotAuthorizedException"`.

2. **Explicit exception conversion**: Added `IcebergExceptionMapper.convertToIcebergException(Exception)` that converts:
   - `UnauthorizedException` (Gravitino) → `NotAuthorizedException` (Iceberg)
   - `ForbiddenException` (Gravitino) → `ForbiddenException` (Iceberg)
   - Already-mapped Iceberg exceptions → passed through
   - Unknown exceptions → `ServiceFailureException` (Iceberg)

3. **Filter uses converted exceptions**: `IcebergAuthenticationFilter` now converts the exception first via `convertToIcebergException()`, then uses the Iceberg exception's `getClass().getSimpleName()` as the type, no more status-code-to-string-name lookup in the filter path.

roryqi: I want to remove variable `ERROR_TYPE_NAMES and use `convertToIcebergException` to convert exceptions to `spec-matching` exceptions.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
```diff
@@ -68,9 +71,41 @@ public class IcebergExceptionMapper implements ExceptionMapper<Exception> {
...
+   * @return the equivalent Iceberg exception, or the original if already an Iceberg exception
+   */
+  public static Exception convertToIcebergException(Exception e) {
+    String message = e.getMessage() != null ? e.getMessage() : "";
+    if (e instanceof UnauthorizedException) {
+      return new NotAuthorizedException("%s", message);
+    }
+    if (e instanceof org.apache.gravitino.exceptions.ForbiddenException) {
+      return new ForbiddenException("%s", message);
+    }
```
roryqi, line 103: We should convert 
```
IllegalArgumentException.class, 400
ValidationException.class, 400
IllegalNameIdentifierException.class, 400
NamespaceNotEmptyException.class, 400
```
to
```
BadRequestException.class.
```

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -42,6 +44,25 @@
 
 public class IcebergRESTUtils {
 
+  /**
+   * Error type names matching the Iceberg REST API specification examples. See <a
+   * href="https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml">Iceberg
+   * REST API spec</a>.
+   */
+  public static final Map<Integer, String> ERROR_TYPE_NAMES =
```
roryqi: We should remove this variable and we should convert improper exception to spec exceptions in the method `convertToIcebergException`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -61,10 +82,11 @@ public static Response notExists() {
   }
 
   public static Response errorResponse(Throwable ex, int httpStatus) {
+    String type = ERROR_TYPE_NAMES.getOrDefault(httpStatus, ex.getClass().getSimpleName());
```
roryqi: We shouldn't change this method.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
laserninja: `convertToIcebergException()` now explicitly converts `IllegalArgumentException`, `ValidationException`, `IllegalNameIdentifierException`, and `NamespaceNotEmptyException` to `BadRequestException`

roryqi:
Could u modify the pull request description, too?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
laserninja: Removed `ERROR_TYPE_NAMES` from `IcebergRESTUtils`, exception-to-type resolution is now handled entirely through `convertToIcebergException()`

roryqi (approved):
LGTM, thanks for your contribution.

Issue #1938, opened 2024-02-02
[Improvement] Possible thread issue in `LockManager.java`


justinmclean:
### What would you like to be improved?

Synchronizing on a parameter synchronizes not on the parameter itself, but on the object assigned to it. Two different threads running the method in parallel could pass two different object instances into the method as parameters, completely undermining the synchronization.

### How should we improve?

Investigate and improve

justinmclean:
It has been a while since I have had to write code like this, so I could be wrong, but I think that in evictStaleNodes two threads could pass different references to the same parent and the code would not behave as you would expect. If that is an issue here I'm not entirely sure.

YxAc:
I really didn't come up with an unexpected case. what do you think? @qqqttt123 @coolderli @xloya @yuqi1129

Issue #1528, opened 2024-01-16
[Subtask] [spark-connector] Add the code skeleton for spark-connector


FANNG1:
### Describe the subtask

Add the code skeleton for spark-connector

### Parent issue

#1227

PR #10671, opened 2026-04-03
[#10669] feat(iceberg-rest): Add pagination support for list endpoints

### LINKED ISSUE #10669: [Improvement] Missing pagination support on Iceberg REST list endpoints (listNamespaces, listTables, listViews)

The Iceberg REST spec defines `pageToken` and `pageSize` query parameters on the `listNamespaces`, `listTables`, and `listViews` endpoints, but Gravitino's Iceberg REST server does not accept or handle these parameters. Clients with large catalogs cannot paginate results — everything is returned in a single response.

**Affected endpoints and code:**

| Endpoint | File | Issue |
|---|---|---|
| `GET /v1/{prefix}/namespaces` | IcebergNamespaceOperations.java | `listNamespaces()` only accepts `parent` `@QueryParam`; no `pageToken`/`pageSize` |
| `GET /v1/{prefix}/namespaces/{namespace}/tables` | IcebergTableOperations.java | `listTable()` has no pagination query params at all |
| `GET /v1/{prefix}/namespaces/{namespace}/views` | IcebergViewOperations.java | `listView()` has no pagination query params at all |

The response builders do pass through `nextPageToken` from the underlying response (e.g., `ListTablesResponse.builder().nextPageToken(listTablesResponse.nextPageToken())`), but this value is always `null` since pagination parameters are never received and never forwarded to the catalog backend.

The dispatcher interfaces (`IcebergTableOperationDispatcher`, `IcebergNamespaceOperationDispatcher`, `IcebergViewOperationDispatcher`) also lack pagination parameters in their method signatures, so the entire call chain from REST handler → dispatcher → executor → `IcebergCatalogWrapper` has no pagination support.

**Note:** The Lance REST server already supports pagination on its list endpoints (`page_token` and `limit` query params in `LanceNamespaceOperations.java`), which can serve as a reference implementation pattern.

### How should we improve?

1. Add `@QueryParam("pageToken") String pageToken` and `@QueryParam("pageSize") Integer pageSize` to the REST endpoint methods in `IcebergTableOperations`, `IcebergNamespaceOperations`, and `IcebergViewOperations`.
2. Update the dispatcher interfaces and their implementations to accept and forward pagination parameters.
3. Pass pagination parameters through to the underlying Iceberg catalog operations (e.g., `CatalogHandlers` / `SupportsNamespaces.listNamespaces()`), implementing server-side pagination where the backing catalog supports it, or applying in-memory pagination as a fallback.
4. Ensure `nextPageToken` in the response is correctly populated when there are more results.

laserninja:

Add `pageToken` and `pageSize` query parameter support to the Iceberg REST server's `listNamespaces`, `listTables`, and `listViews` endpoints. Pagination is handled at the REST layer in `IcebergPaginationHelper`, a package-private utility that paginates the full response after authorization filtering.

Pagination is implemented as cursor/keyset-based in-memory pagination. Results are sorted deterministically by name, and `nextPageToken` is set to the name of the last item on the current page. On the next request, items strictly after the cursor (by string comparison) are returned. When no pagination parameters are provided, existing behavior (return all results) is preserved.

**Known limitations** (documented in `IcebergPaginationHelper` Javadoc):
- The full list is materialized from the underlying catalog on every page request (the Iceberg `Catalog` interface does not support native pagination).
- Items created after the cursor but before the current page's last item may be missed — this is consistent with standard keyset pagination behavior.

The Iceberg REST spec defines `pageToken` and `pageSize` query parameters on list endpoints, but Gravitino's Iceberg REST server did not accept or handle them. Clients with large catalogs cannot paginate results — everything was returned in a single response.

Fix: #10669

Yes. The following REST endpoints now accept optional `pageToken` and `pageSize` query parameters:
- `GET /v1/{prefix}/namespaces`
- `GET /v1/{prefix}/namespaces/{namespace}/tables`
- `GET /v1/{prefix}/namespaces/{namespace}/views`

- Added comprehensive unit tests in `TestIcebergPaginationHelper` covering cursor-based pagination logic (first page, middle page, last page, beyond-end cursor, deterministic sort, empty list, single item, page size exceeding list size)
- Added `testListNamespaceWithPagination` in `TestIcebergNamespaceOperations`
- Added `testListTablesWithPagination` in `TestIcebergTableOperations`
- Added `testListViewsWithPagination` in `TestIcebergViewOperations`
- All existing unit tests pass (`./gradlew :iceberg:iceberg-rest-server:test -PskipITs`)

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergNamespaceOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergNamespaceOperations.java
@@ -25,6 +25,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
@@ -101,7 +102,9 @@ public class IcebergNamespaceOperations {
       accessMetadataType = MetadataObject.Type.CATALOG)
   public Response listNamespaces(
       @DefaultValue("") @Encoded() @QueryParam("parent") String parent,
-      @AuthorizationMetadata(type = Entity.EntityType.CATALOG) @PathParam("prefix") String prefix) {
+      @AuthorizationMetadata(type = Entity.EntityType.CATALOG) @PathParam("prefix") String prefix,
+      @QueryParam("pageToken") String pageToken,
+      @QueryParam("pageSize") Integer pageSize) {
     String catalogName = IcebergRESTUtils.getCatalogName(prefix);
     Namespace parentNamespace =
         parent.isEmpty()
@@ -124,6 +127,9 @@ public class IcebergNamespaceOperations {
               response =
                   filterListNamespacesResponse(response, authContext.metalakeName(), catalogName);
             }
+            response =
+                IcebergPaginationHelper.paginateNamespaces(
+                    response, Optional.ofNullable(pageToken), Optional.ofNullable(pageSize));
             return IcebergRESTUtils.ok(response);
           });
     } catch (Exception e) {
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergPaginationHelper.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergPaginationHelper.java
@@ -1,0 +1,155 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.iceberg.service.rest;
+
+import com.google.common.base.Preconditions;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.rest.responses.ListNamespacesResponse;
+import org.apache.iceberg.rest.responses.ListTablesResponse;
+
+/**
+ * Utility for applying cursor-based in-memory pagination to Iceberg list responses.
+ *
+ * <p>Items are sorted deterministically by name and the page token is the name of the last item on
+ * the current page. On the next request, items strictly after the cursor (by string comparison) are
+ * returned. This is more resilient than offset-based pagination when the underlying catalog is
+ * modified between page requests.
+ *
+ * <p><b>Known limitations:</b>
+ *
+ * <ul>
+ *   <li>This is in-memory pagination: the full item list is materialized from the underlying
+ *       catalog on every page request, so server-side memory usage is unchanged. Paging through N
+ *       items is O(N²) total work. True server-push-down pagination would require catalog
+ *       implementations to add native support.
+ *   <li>Items created after the cursor but alphabetically before the last item on the current page
+ *       will be missed on subsequent pages. This is the same behavior as most keyset pagination
+ *       implementations and is generally acceptable.
+ *   <li>When {@code pageToken} is provided without {@code pageSize}, the response returns all items
+ *       after the cursor with no {@code nextPageToken}.
+ * </ul>
+ */
+class IcebergPaginationHelper {
+
+  private IcebergPaginationHelper() {}
+
+  /**
+   * Paginate a {@link ListNamespacesResponse}.
+   *
+   * @param response the full (unpaginated) response
+   * @param pageToken cursor-based page token (name of last item on previous page), or empty for the
+   *     first page
+   * @param pageSize maximum items per page, or empty for no limit
+   * @return a new response containing only the requested page
+   */
+  static ListNamespacesResponse paginateNamespaces(
+      ListNamespacesResponse response, Optional<String> pageToken, Optional<Integer> pageSize) {
+    PaginatedPage<Namespace> page =
+        paginate(response.namespaces(), pageToken, pageSize, Namespace::toString);
+    ListNamespacesResponse.Builder builder = ListNamespacesResponse.builder().addAll(page.items);
+    page.nextPageToken.ifPresent(builder::nextPageToken);
+    return builder.build();
+  }
+
+  /**
+   * Paginate a {@link ListTablesResponse}. Works for both table and view list responses.
+   *
+   * @param response the full (unpaginated) response
+   * @param pageToken cursor-based page token (name of last item on previous page), or empty for the
+   *     first page
+   * @param pageSize maximum items per page, or empty for no limit
+   * @return a new response containing only the requested page
+   */
+  static ListTablesResponse paginateTables(
+      ListTablesResponse response, Optional<String> pageToken, Optional<Integer> pageSize) {
+    PaginatedPage<TableIdentifier> page =
+        paginate(response.identifiers(), pageToken, pageSize, TableIdentifier::toString);
+    ListTablesResponse.Builder builder = ListTablesResponse.builder().addAll(page.items);
+    page.nextPageToken.ifPresent(builder::nextPageToken);
+    return builder.build();
+  }
+
+  /**
+   * Core pagination logic shared by all list endpoints.
+   *
+   * <p>Items are sorted by {@code keyExtractor} to produce a stable ordering, then sliced using the
+   * cursor token. The next-page token is the key of the last item on the returned page.
+   *
+   * @param items the complete list of items to paginate
+   * @param pageToken cursor string (key of last item on previous page), or empty for the first page
+   * @param pageSize maximum items per page, or empty for no limit
+   * @param keyExtractor function to extract a comparable cursor key from each item
+   * @return a {@link PaginatedPage} containing the requested page of items and optional next token
+   */
+  static <T> PaginatedPage<T> paginate(
+      List<T> items,
+      Optional<String> pageToken,
+      Optional<Integer> pageSize,
+      Function<T, String> keyExtractor) {
+    String token = pageToken.orElse("");
+    if (!pageSize.isPresent() && token.isEmpty()) {
+      return new PaginatedPage<>(items, Optional.empty());
+    }
+
+    pageSize.ifPresent(
+        size -> Preconditions.checkArgument(size > 0, "pageSize must be positive, got: %s", size));
+
+    List<T> sorted =
+        items.stream().sorted(Comparator.comparing(keyExtractor)).collect(Collectors.toList());
+
+    int startIdx = 0;
+    if (!token.isEmpty()) {
+      startIdx = sorted.size();
+      for (int i = 0; i < sorted.size(); i++) {
+        if (keyExtractor.apply(sorted.get(i)).compareTo(token) > 0) {
+          startIdx = i;
+          break;
+        }
+      }
+    }
+
+    int limit = pageSize.orElse(sorted.size());
+    int end = Math.min(startIdx + limit, sorted.size());
+    List<T> page = sorted.subList(startIdx, end);
+
+    Optional<String> nextToken = Optional.empty();
+    if (end < sorted.size() && !page.isEmpty()) {
+      nextToken = Optional.of(keyExtractor.apply(page.get(page.size() - 1)));
+    }
+
+    return new PaginatedPage<>(page, nextToken);
+  }
+
+  /** Holds a page of items and an optional token for the next page. */
+  static class PaginatedPage<T> {
+    final List<T> items;
+    final Optional<String> nextPageToken;
+
+    PaginatedPage(List<T> items, Optional<String> nextPageToken) {
+      this.items = items;
+      this.nextPageToken = nextPageToken;
+    }
+  }
+}
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
@@ -123,7 +123,9 @@ public class IcebergTableOperations {
   public Response listTable(
       @AuthorizationMetadata(type = Entity.EntityType.CATALOG) @PathParam("prefix") String prefix,
       @AuthorizationMetadata(type = EntityType.SCHEMA) @Encoded() @PathParam("namespace")
-          String namespace) {
+          String namespace,
+      @QueryParam("pageToken") String pageToken,
+      @QueryParam("pageSize") Integer pageSize) {
     String catalogName = IcebergRESTUtils.getCatalogName(prefix);
     Namespace icebergNS =
         RESTUtil.decodeNamespace(namespace, IcebergRESTUtils.NAMESPACE_SEPARATOR_URLENCODED_UTF_8);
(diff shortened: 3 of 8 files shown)


roryqi:
1. Iceberg REST server will use Hive backend, too. How does Hive backend support paging? Do we need to consider the backend? How did other Iceberg REST like Polaris, Lakekeeper, UnityCatalog to support paging operations?
2. Could u resolve the conflicts?

laserninja:
Thanks for checking, @roryqi!

**1. Backend considerations & how other Iceberg REST catalogs handle pagination:**

The Iceberg Catalog Java API (`listNamespaces()`, `listTables()`, `listViews()`) returns complete `List` results with no built-in pagination support — this is true regardless of the backend (Hive, JDBC, etc.). The Hive metastore APIs (`getAllDatabases`, `getAllTables`) also return all results at once without native pagination. So in-memory pagination at the REST layer is the only practical approach here.

This is consistent with how other Iceberg REST catalogs handle it:

- **Polaris** went through an extensive pagination effort (PRs [#1528](https://github.com/apache/polaris/pull/1528), [#1938](https://github.com/apache/polaris/pull/1938)). Their initial implementation also did in-memory pagination at the REST/service layer. They later worked on pushing pagination down to the persistence layer, but that was for Polaris's own entity management — the Iceberg catalog operations themselves (via `CatalogHandlers.listNamespaces`) still use in-memory listing. They also noted this is a limitation when federating to external catalogs.
- **Lakekeeper** and **UnityCatalog** similarly handle `pageToken`/`pageSize` at the service layer.

Since Gravitino's `IcebergCatalogWrapper` wraps the standard Iceberg `Catalog` interface, in-memory pagination is the appropriate approach at this layer. If backend-native pagination is desired in the future, it would require changes to the Iceberg Catalog interface itself.

**2.** Will resolve the conflicts.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -192,6 +194,12 @@ public ListNamespacesResponse listNamespace(Namespace parent) {
     return CatalogHandlers.listNamespaces(getNamespaceCatalog(), parent);
   }
 
+  public ListNamespacesResponse listNamespace(
+      Namespace parent, @Nullable String pageToken, @Nullable Integer pageSize) {
```
roryqi: Could u use Optional instead of Nullable annotation here? Because Optional will make others consider null value explicitly.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -451,6 +471,61 @@ public IcebergTableChange(TableIdentifier tableIdentifier, Transaction transacti
...
+    try {
+      int offset = Integer.parseInt(pageToken);
+      Preconditions.checkArgument(offset >= 0, "pageToken must be non-negative, got: %s", offset);
+      return offset;
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException("Invalid pageToken: " + pageToken, e);
+    }
+  }
+
+  private ListNamespacesResponse paginateNamespaces(
```
roryqi: Could we extract some common logic for `paginateNamespaces` and `paginateTables`?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergNamespaceOperations.java
```diff
@@ -114,7 +116,8 @@ public Response listNamespaces(
             IcebergRequestContext context =
                 new IcebergRequestContext(httpServletRequest(), catalogName);
             ListNamespacesResponse response =
-                namespaceOperationDispatcher.listNamespaces(context, parentNamespace);
+                namespaceOperationDispatcher.listNamespaces(
+                    context, parentNamespace, pageToken, pageSize);
```
roryqi: **Pagination happens before authorization filtering.** `filterListNamespacesResponse` runs below (after this call), so when authorization is enabled the response is sliced to `pageSize` *first* and then filtered, meaning the caller can get fewer than `pageSize` items (possibly zero) while authorized items still exist at later offsets — and `nextPageToken` here encodes a pre-filter offset, so the page-walk is inconsistent with what the client sees. The same pattern applies to the table and view list endpoints. Consider paginating after filtering, or document the limitation. Either way, a test covering pagination + authz would be valuable.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -451,6 +471,61 @@ public IcebergTableChange(TableIdentifier tableIdentifier, Transaction transacti
...
+      ListNamespacesResponse response, @Nullable String pageToken, @Nullable Integer pageSize) {
+    if (pageSize == null && (pageToken == null || pageToken.isEmpty())) {
+      return response;
+    }
+    List<Namespace> all = response.namespaces();
```
roryqi: `response` here already came from `listNamespace(parent)`, which materializes the **entire** namespace list from the underlying catalog. Since this happens on every page request, server-side memory is unchanged and paging through N items is O(N²) total work. That's a reasonable first step, but the PR's motivation implies true pagination — worth documenting this as a known limitation.

Also note offset-based tokens assume a stable, deterministic ordering across calls; concurrent create/drop between pages can skip or duplicate items. And minor: the `offset >= all.size()` early-return (just below) runs before `checkArgument(limit > 0)`, so an invalid `pageSize` combined with a large offset silently returns empty instead of erroring.

roryqi (commented):
## Summary

Thanks for this — the plumbing is clean and consistent across all layers (`@Nullable` throughout, interface Javadoc updated, existing tests adjusted, happy-path tests added for all three endpoints), and the no-param path is safely backward-compatible.

I have two concerns I'd like to see resolved or explicitly documented before merge, plus some smaller items. Details are in the inline comments; summary below.

### Major

1. **Pagination runs before authorization filtering.** In `IcebergNamespaceOperations.listNamespaces` (and the table/view equivalents), the dispatcher returns the already-sliced page and `filterListNamespacesResponse` is applied *afterward*. With authorization enabled, a `pageSize=2` request slices to 2 entries and then the filter may drop some, so the caller can receive fewer than `pageSize` items (possibly zero) while more authorized items exist at later offsets — and `nextPageToken` encodes a pre-filter offset, so the page-walk becomes inconsistent with what the client actually sees. This makes pagination unreliable on any authz-enabled deployment, and there's no test covering the interaction.

2. **In-memory offset pagination doesn't address the stated scaling goal.** `paginate*` calls the underlying `listNamespace/listTable` first, which still materializes the entire list from the catalog on every page request. Server-side memory is unchanged, and walking N items in pages of K is O(N²) total work. The payload shrinks, which is a fine first step, but the PR description implies true pagination — please document this as a known limitation.

### Medium

3. **Offset paging assumes a stable list ordering.** Offset tokens are only correct if the underlying catalog returns a deterministic, stable order across calls. Concurrent create/drop between page requests can skip or duplicate items. Using a raw offset is spec-legal (the token is opaque to clients), but the assumption should be called out.

4. **Test coverage gaps.** Only happy paths are covered. Missing: invalid `pageToken` (e.g. `"abc"`) and negative/zero `pageSize` (should be 400), offset beyond list size (empty page + null token), `pageToken` without `pageSize`, direct unit tests on `IcebergCatalogWrapper.paginate*`, and the pagination+authorization interaction.

### Minor

- **Validation ordering** in `IcebergCatalogWrapper`: the `offset >= all.size()` early-return happens before `checkArgument(limit > 0)`, so an invalid `pageSize` with a large offset silently returns empty instead of erroring.
- **`pageToken` without `pageSize`** returns everything from the offset onward with no `nextPageToken` — harmless but surprising; worth a comment on intended semantics.
- **Code duplication**: `paginateNamespaces` and `paginateTables` are nearly identical and could be a single generic helper.
- **Non-integer `pageSize`**: JAX-RS converts a bad `@QueryParam Integer` to 404, not 400 (pre-existing behavior, just inconsistent with the in-code 400 validation).

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
laserninja: refactored to use Optional<String> pageToken and Optional<Integer> pageSize throughout the dispatcher chain in a previous commit (43c51a07). In this latest push, I moved pagination out of the dispatcher chain entirely into the REST layer (IcebergPaginationHelper), so the dispatcher interfaces no longer carry pagination params at all.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
laserninja: extracted shared pagination logic into IcebergPaginationHelper (a package-private utility in the rest package). paginateNamespaces() and paginateTables() are separate methods since they operate on different response types (ListNamespacesResponse vs ListTablesResponse), but the core offset/subList logic is in one place now.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergNamespaceOperations.java
laserninja: pagination now runs after authorization filtering in all three REST endpoints (IcebergNamespaceOperations, IcebergTableOperations, IcebergViewOperations). The flow is: dispatcher returns full list → authz filter applied → IcebergPaginationHelper paginates the filtered result. This ensures pageSize/nextPageToken are consistent with the authorized view.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
laserninja: Sounds good. I've added Javadoc on IcebergPaginationHelper documenting this as offset-based, in-memory pagination (the full list is materialized on each request). The known limitations (O(N²) for full walk, offset stability assumptions with concurrent changes) are called out. Validation ordering is also fixed: parsePageToken now validates before the offset >= size early-return. True catalog-level pagination would require changes to the Iceberg Catalog interface and is out of scope here.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergPaginationHelper.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Paginate a {@link ListTablesResponse}. Works for both table and view list responses.
+   *
+   * @param response the full (unpaginated) response
+   * @param pageToken offset-based page token, or {@code null} for the first page
+   * @param pageSize maximum items per page, or {@code null} for no limit
+   * @return a new response containing only the requested page
+   */
+  static ListTablesResponse paginateTables(
```
roryqi, line 85: paginateTables and paginateNamespaces has similar logic using the different constructor method.
Maybe you can extract a internal method like
```
private <T, S> T paginate(T response, Optional<String>, Optional<Integer> pageSize, Function<List<S>, T> creator)
```
So we can reuse the duplicated logic.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -123,7 +123,9 @@ public IcebergTableOperations(
   public Response listTable(
       @AuthorizationMetadata(type = Entity.EntityType.CATALOG) @PathParam("prefix") String prefix,
       @AuthorizationMetadata(type = EntityType.SCHEMA) @Encoded() @PathParam("namespace")
-          String namespace) {
+          String namespace,
+      @QueryParam("pageToken") String pageToken,
```
roryqi, line 127: pageToken is string and pageSize is integer. Do u you have some reasons?

FANNG1:
Thanks for the detailed explanation on in-memory pagination being the only practical option given the current `Catalog` interface constraints. That makes sense.

However, I still think the offset-based token is a concern worth addressing, since it can silently produce incorrect results (missing or duplicate entries) without any indication to the caller — not just as an edge case, but as a predictable failure mode whenever the catalog is active.

**Proposed alternative: name-based cursor token**

Since the full list is already materialized in memory on every request, we can make pagination significantly more robust at no extra cost:

1. **Sort the list deterministically** (e.g., by name alphabetically) before slicing.
2. **Use the last returned item's name as the `pageToken`** (cursor), instead of a numeric offset.
3. On the next request, binary-search or scan the sorted list for the cursor name and return items after it.

```java
// Example sketch
List<Namespace> sorted = all.stream()
    .sorted(Comparator.comparing(ns -> ns.toString()))
    .collect(toList());

int startIdx = 0;
if (pageToken != null && !pageToken.isEmpty()) {
    // Find the first item strictly after the cursor
    startIdx = IntStream.range(0, sorted.size())
        .filter(i -> sorted.get(i).toString().compareTo(pageToken) > 0)
        .findFirst()
        .orElse(sorted.size());
}
List<Namespace> page = sorted.subList(startIdx, Math.min(startIdx + pageSize, sorted.size()));
String nextToken = page.isEmpty() ? null : page.get(page.size() - 1).toString();
```

**Why this is more resilient:**

| Scenario | Offset-based (current) | Name-cursor (proposed) |
|---|---|---|
| Item deleted before cursor | Subsequent page **skips** an item | Cursor still valid, no data lost |
| Item created before cursor | Subsequent page **duplicates** vicinity | Item already passed, correctly skipped |
| Item created after cursor | Correctly included | Correctly included |
| Cursor item itself deleted | No issue (offset just moves) | First item alphabetically after cursor name is returned — graceful |

The only remaining edge case is items created *after* the cursor but *before* the current page's last item — these will be missed, which is the same behavior as most keyset-pagination implementations and generally acceptable. This is a much smaller window than the current approach.

Since the list is already sorted by most underlying catalog implementations (Hive, JDBC), the sort step may even be a no-op in practice.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
laserninja: This follows the Iceberg REST spec, pageToken is defined as an opaque string (it's a cursor, not necessarily numeric), while pageSize is a numeric integer representing the max items per page. With the switch to cursor-based tokens, the distinction is even more relevant: tokens are now the name of the last item on the page (e.g. "db.my_table"), not a numeric offset.

laserninja:
Great suggestion. Implemented exactly as you described. Items are now sorted deterministically by name, the page token is the name of the last item on the current page, and the next request returns items strictly after the cursor by string comparison. The known limitations (items created after cursor but before last item are missed, same as standard keyset pagination) are documented in the IcebergPaginationHelper Javadoc. Also added comprehensive unit tests for the cursor-based logic in a new TestIcebergPaginationHelper class.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergPaginationHelper.java
```diff
@@ -0,0 +1,160 @@
...
+   * @param items the complete list of items to paginate
+   * @param pageToken cursor string (key of last item on previous page), or {@code null} for the
+   *     first page
+   * @param pageSize maximum items per page, or {@code null} for no limit
+   * @param keyExtractor function to extract a comparable cursor key from each item
+   * @return a {@link PaginatedPage} containing the requested page of items and optional next token
+   */
+  static <T> PaginatedPage<T> paginate(
+      List<T> items,
+      @Nullable String pageToken,
```
roryqi: Could u use Optional here?

roryqi:
@FANNG1 Do u have other comments?

FANNG1:
I found two issues:

1. Blocking: `IcebergNamespaceOperations#listNamespaces` now decodes the `parent` query parameter with `RESTUtil.decodeNamespace(parent)`, which drops Gravitino's custom namespace separator (`IcebergRESTUtils.NAMESPACE_SEPARATOR_URLENCODED_UTF_8`). Other namespace/table/view paths still decode with that separator, and existing nested namespace requests encode parents with it. This can break `GET /namespaces?parent=...` for nested namespaces by decoding the parent incorrectly. Please keep the existing overload:

```java
RESTUtil.decodeNamespace(parent, IcebergRESTUtils.NAMESPACE_SEPARATOR_URLENCODED_UTF_8)
```

2. Minor: the PR description says pagination is offset-based and `nextPageToken` is the next offset, but the implementation is cursor/keyset-style: results are sorted by name and `nextPageToken` is the last item name from the current page. Please update the PR description so the documented behavior matches the implementation.

Other than these, LGTM.

laserninja:
Resolved by rebasing on upstream/main. The code now correctly uses RESTUtil.decodeNamespace(parent, IcebergRESTUtils.NAMESPACE_SEPARATOR_URLENCODED_UTF_8).

Updated to accurately describe cursor/keyset-based pagination (sorted by name, token is last item name) instead of the old offset-based description.

FANNG1:
LGTM. Thanks for addressing the comments.

roryqi (approved):

PR #10673, opened 2026-04-03
[#10672] improvement(auth): Return specific error message when authentication token is expired

### LINKED ISSUE #10672: [Improvement] Return authentication timeout error response when token is expired

When a JWT token has expired, the server currently returns a generic error message ("JWT parse error" or "JWKS JWT validation error") rather than a specific authentication timeout message. This makes it impossible for clients to distinguish an expired token from a malformed token or invalid signature.

The issue affects two validators:

StaticSignKeyValidator — ExpiredJwtException is caught in the same multi-catch as MalformedJwtException, UnsupportedJwtException, SignatureException, etc., and returns the generic "JWT parse error".
JwksTokenValidator — All exceptions are caught generically and return "JWKS JWT validation error", regardless of whether the token was expired or had a different fault.
Clients receive a 401 Unauthorized with no way to determine whether they should refresh/re-authenticate vs. debug a misconfigured token.

### How should we improve?

In StaticSignKeyValidator, split ExpiredJwtException into its own catch block and return "Authentication token is expired".
In JwksTokenValidator, add a catch (BadJWTException e) block that inspects the message for expiry and returns "Authentication token is expired".
This gives clients an actionable error and improves observability.

laserninja:

Split the `ExpiredJwtException` / `BadJWTException` catch blocks in `StaticSignKeyValidator` and `JwksTokenValidator` to return the specific message `"Authentication token is expired"` instead of the generic `"JWT parse error"` / `"JWKS JWT validation error"`.

Also adds regression tests (`testValidateExpiredTokenHasCorrectMessage`) in both test classes to guard against the message regressing.

Clients receiving a `401 Unauthorized` had no way to distinguish an expired token from a malformed or wrongly-signed token. An actionable error message lets clients know they should refresh/re-authenticate rather than debug a configuration issue.

Fix: #10672

Yes — the `401 Unauthorized` error message for an expired JWT changes from `"JWT parse error"` (static key) or `"JWKS JWT validation error"` (JWKS) to `"Authentication token is expired"`.

- Added `testValidateExpiredTokenHasCorrectMessage` to `TestStaticSignKeyValidator` and `TestJwksTokenValidator`.
- All existing and new unit tests pass (`./gradlew :server-common:test`).

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
@@ -27,6 +27,7 @@ import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 import org.apache.gravitino.exceptions.IllegalNameIdentifierException;
 import org.apache.gravitino.exceptions.NoSuchCatalogException;
+import org.apache.gravitino.exceptions.TokenExpiredException;
 import org.apache.gravitino.exceptions.UnauthorizedException;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.BadRequestException;
@@ -61,6 +62,8 @@ public class IcebergExceptionMapper implements ExceptionMapper<Exception> {
           .put(NamespaceNotEmptyException.class, 400)
           .put(NotAuthorizedException.class, 401)
           .put(UnauthorizedException.class, 401)
+          .put(AuthenticationTimeoutException.class, 419)
+          .put(TokenExpiredException.class, 419)
           .put(org.apache.gravitino.exceptions.ForbiddenException.class, 403)
           .put(ForbiddenException.class, 403)
           .put(NotFoundException.class, 404)
@@ -97,6 +100,9 @@ public class IcebergExceptionMapper implements ExceptionMapper<Exception> {
    */
   public static Exception convertToIcebergException(Exception e) {
     String message = e.getMessage() != null ? e.getMessage() : "";
+    if (e instanceof TokenExpiredException) {
+      return new AuthenticationTimeoutException("%s", message);
+    }
     if (e instanceof UnauthorizedException) {
       return new NotAuthorizedException("%s", message);
     }
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
@@ -29,6 +29,7 @@ import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.jwt.proc.DefaultJWTClaimsVerifier;
 import com.nimbusds.jwt.proc.DefaultJWTProcessor;
+import com.nimbusds.jwt.proc.ExpiredJWTException;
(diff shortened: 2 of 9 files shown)


roryqi:
We should return AuthenticationTimeoutResponse， you can see https://github.com/apache/iceberg/blob/9a939d68358de9dac2c6ba9b236b675ebe477490/open-api/rest-catalog-open-api.yaml#L292
I think we can continue this pull request after https://github.com/apache/gravitino/pull/10668

roryqi:
> We should return AuthenticationTimeoutResponse， you can see https://github.com/apache/iceberg/blob/9a939d68358de9dac2c6ba9b236b675ebe477490/open-api/rest-catalog-open-api.yaml#L292 I think we can continue this pull request after #10668

@laserninja Maybe we can continue this pull request after merging #10668

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
```diff
@@ -161,6 +163,12 @@ public Principal validateToken(String token, String serviceAudience) {
       }
       return userPrincipal;
 
+    } catch (BadJWTException e) {
```
roryqi: Could u use here?
```
package com.nimbusds.jwt.proc;

public class ExpiredJWTException extends BadJWTException {
    public ExpiredJWTException(String message) {
        super(message);
    }
}
```

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergExceptionMapper.java
```diff
@@ -97,6 +99,9 @@ public static int getErrorCode(Exception ex) {
    */
   public static Exception convertToIcebergException(Exception e) {
     String message = e.getMessage() != null ? e.getMessage() : "";
+    if (e instanceof TokenExpiredException) {
+      return e;
```
roryqi: We should Iceberg Exception here instead of Gravitino Exception here.

laserninja: Sounds good. I've created `AuthenticationTimeoutException extends RESTException' in the iceberg-rest-server module (the Iceberg SDK doesn't ship this class despite it being part of the REST spec). IcebergExceptionMapper.convertToIcebergException() now converts TokenExpiredException → 'AuthenticationTimeoutException', so only the Iceberg exception type reaches the HTTP response.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
laserninja: ExpiredJWTException does not exist in nimbus-jose-jwt 9.37.3 (the version Gravitino uses) as it was introduced in 10.x. Since upgrading nimbus is a larger change, I've kept the BadJWTException catch with message checking and added a comment explaining why. If/when Gravitino upgrades to nimbus 10.x, we can switch to catching ExpiredJWTException directly. Let me know if you want me to create another MR for that.

roryqi: Maybe we can raise an another pull request to upgrade nimbus. This seems a work around. We should avoid work around fix as possible as we can.

laserninja: Done https://github.com/apache/gravitino/pull/10773

roryqi (approved):

PR #10675, opened 2026-04-03
[#10674] feat(iceberg): implement commitTransaction REST endpoint for Iceberg REST catalog

### LINKED ISSUE #10674: [Improvement] Implement commitTransaction REST endpoint for Iceberg REST catalog

Implement the missing `POST /v1/{prefix}/transactions/commit` REST endpoint in Gravitino's Iceberg REST catalog server, as defined by the Iceberg REST Catalog API spec (`CommitTransactionRequest`).

The Iceberg REST spec defines `POST /v1/{prefix}/transactions/commit` for atomic multi-table commits. Gravitino's Iceberg REST server does not implement this endpoint. Clients performing atomic cross-table operations (e.g., Spark multi-table writes) will fail when targeting Gravitino as an Iceberg REST catalog.

### How should we improve?

- Add `IcebergTransactionOperations` JAX-RS class with `POST /v1/{prefix}/transactions/commit`
- Add `IcebergTransactionOperationDispatcher` interface + `Executor`/`EventDispatcher`/`HookDispatcher` chain
- Wire the endpoint into the existing server bootstrap
- Register the endpoint in `ConfigResponse` so clients discover it
- Add unit and integration tests

laserninja:
Implements the missing `POST /v1/{prefix}/transactions/commit` endpoint in Gravitino's Iceberg REST catalog server, following the Iceberg REST Catalog API spec for atomic multi-table commits via `CommitTransactionRequest`.

**New files:**
- `IcebergTransactionOperationDispatcher` — interface
- `IcebergTransactionOperationExecutor` — core implementation calling `IcebergCatalogWrapper.commitTransaction()`
- `IcebergTransactionEventDispatcher` — pre/success/failure event dispatch layer
- `IcebergTransactionHookDispatcher` — hook execution layer
- `IcebergCommitTransactionPreEvent`, `IcebergCommitTransactionEvent`, `IcebergCommitTransactionFailureEvent`
- `IcebergTransactionOperations` — JAX-RS resource for the endpoint

**Modified files:**
- `OperationType` — add `COMMIT_TRANSACTION`
- `IcebergCatalogWrapper` — add `commitTransaction()` iterating `tableChanges()`
- `RESTService` — wire dispatcher chain and bind to DI container
- `IcebergConfigOperations` — advertise `Endpoint.V1_COMMIT_TRANSACTION` in `/v1/config`

The Iceberg REST spec defines `POST /v1/{prefix}/transactions/commit` for atomic multi-table commits. Without it, clients performing cross-table operations (e.g., Spark multi-table writes) cannot use Gravitino as an Iceberg REST catalog.

Fix: #10674

Yes — new REST endpoint `POST /v1/{prefix}/transactions/commit` is now available and advertised in the `/v1/config` response.

Added `TestIcebergTransactionOperations` with parameterized tests covering:
- Successful multi-table transaction commit → HTTP 204 + success event
- Table-not-found → HTTP 404 + failure event
- Endpoint works correctly with URL prefix

### DIFF
--- a/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
+++ b/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
@@ -21,6 +21,8 @@ package org.apache.gravitino.iceberg.common.ops;
 import com.google.common.base.Preconditions;
 import java.sql.Driver;
 import java.sql.DriverManager;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
@@ -37,6 +39,9 @@ import org.apache.gravitino.utils.ClassUtils;
 import org.apache.gravitino.utils.IsolatedClassLoader;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.iceberg.HasTableOperations;
+import org.apache.iceberg.MetadataUpdate;
+import org.apache.iceberg.Table;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.Transaction;
 import org.apache.iceberg.catalog.Catalog;
@@ -253,16 +258,66 @@ public class IcebergCatalogWrapper implements AutoCloseable {
   }
 
   /**
-   * Commits multiple table updates atomically as a single transaction. Each {@link
-   * UpdateTableRequest} in the request must include a {@link TableIdentifier} specifying which
-   * table to update.
+   * Commits multiple table updates in a best-effort atomic manner using a two-phase
+   * validate-then-commit approach:
+   *
+   * <ol>
+   *   <li>Phase 1: Load all tables and validate ALL requirements against current metadata. If any
+   *       requirement fails, the entire transaction is rejected before any commit is made.
+   *   <li>Phase 2: Apply metadata updates and commit each table. Once phase 1 succeeds, commits are
+   *       applied sequentially.
+   * </ol>
+   *
+   * <p>True cross-table atomicity (rollback of already-committed tables) depends on whether the
+   * underlying catalog backend supports it. Most Iceberg catalog backends do not provide
+   * cross-table rollback, but this two-phase approach ensures that no table is modified if any
+   * pre-condition check fails.
+   *
+   * <p>Each {@link UpdateTableRequest} in the request must include a {@link TableIdentifier}.
    *
    * @param commitTransactionRequest The request containing all table changes to apply.
    */
   public void commitTransaction(CommitTransactionRequest commitTransactionRequest) {
     commitTransactionRequest.validate();
-    for (UpdateTableRequest tableChange : commitTransactionRequest.tableChanges()) {
-      updateTable(tableChange.identifier(), tableChange);
+    List<UpdateTableRequest> tableChanges = commitTransactionRequest.tableChanges();
+
+    // Phase 1: validate ALL requirements before committing anything
+    List<org.apache.iceberg.TableOperations> allOps = new ArrayList<>(tableChanges.size());
+    List<TableMetadata> allBase = new ArrayList<>(tableChanges.size());
+    List<TableMetadata> allUpdated = new ArrayList<>(tableChanges.size());
+
+    for (UpdateTableRequest change : tableChanges) {
+      Preconditions.checkArgument(
+          change.identifier() != null,
+          "Invalid table change in transaction: missing table identifier");
+      Table table = catalog.loadTable(change.identifier());
+      Preconditions.checkArgument(
+          table instanceof HasTableOperations,
+          "Table does not support operations required for transaction commit: %s",
+          change.identifier());
+
+      org.apache.iceberg.TableOperations ops = ((HasTableOperations) table).operations();
+      TableMetadata base = ops.current();
+
+      // Validate all requirements against the current metadata — fail fast before any commit
+      change.requirements().forEach(req -> req.validate(base));
+
+      // Build the new metadata by applying all updates
+      TableMetadata.Builder builder = TableMetadata.buildFrom(base);
+      for (MetadataUpdate update : change.updates()) {
+        update.applyTo(builder);
+      }
+      TableMetadata updated = builder.build();
+
+      allOps.add(ops);
+      allBase.add(base);
+      allUpdated.add(updated);
+    }
+
+    // Phase 2: all requirements passed — now commit each table
+    for (int i = 0; i < allOps.size(); i++) {
+      allOps.get(i).commit(allBase.get(i), allUpdated.get(i));
+      metadataCache.updateTableMetadata(tableChanges.get(i).identifier(), allUpdated.get(i));
     }
   }
 
--- a/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
+++ b/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
@@ -46,6 +46,7 @@ import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.catalog.ViewCatalog;
 import org.apache.iceberg.jdbc.JdbcCatalogWithMetadataLocationSupport;
 import org.apache.iceberg.rest.CatalogHandlers;
+import org.apache.iceberg.rest.requests.CommitTransactionRequest;
 import org.apache.iceberg.rest.requests.CreateNamespaceRequest;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
 import org.apache.iceberg.rest.requests.CreateViewRequest;
@@ -249,6 +250,20 @@ public class IcebergCatalogWrapper implements AutoCloseable {
     Transaction transaction = icebergTableChange.getTransaction();
     transaction.commitTransaction();
     return loadTable(icebergTableChange.getTableIdentifier());
+  }
+
+  /**
+   * Commits multiple table updates atomically as a single transaction. Each {@link
+   * UpdateTableRequest} in the request must include a {@link TableIdentifier} specifying which
+   * table to update.
+   *
+   * @param commitTransactionRequest The request containing all table changes to apply.
+   */
+  public void commitTransaction(CommitTransactionRequest commitTransactionRequest) {
+    commitTransactionRequest.validate();
+    for (UpdateTableRequest tableChange : commitTransactionRequest.tableChanges()) {
+      updateTable(tableChange.identifier(), tableChange);
+    }
   }
 
   public LoadViewResponse createView(Namespace namespace, CreateViewRequest request) {
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTransactionOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTransactionOperations.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.iceberg.service.rest;
+
+import com.codahale.metrics.annotation.ResponseMetered;
+import com.codahale.metrics.annotation.Timed;
+import com.google.common.annotations.VisibleForTesting;
+import javax.inject.Inject;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.apache.gravitino.Entity;
+import org.apache.gravitino.MetadataObject;
+import org.apache.gravitino.iceberg.service.IcebergExceptionMapper;
+import org.apache.gravitino.iceberg.service.IcebergRESTUtils;
+import org.apache.gravitino.iceberg.service.dispatcher.IcebergTransactionOperationDispatcher;
+import org.apache.gravitino.listener.api.event.IcebergRequestContext;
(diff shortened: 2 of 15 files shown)


FANNG1:
@laserninja ,  thanks for the PR,  commitTransaction is a **atomic** operation for multilple table updates, I'm afraid the current implementation is not atomic.

laserninja:
Thanks @FANNG1 for the review! You're absolutely right.

The previous implementation called `updateTable()` sequentially per table, so if table 2 failed, table 1 was already committed — not atomic at all.

I've replaced it with a **two-phase validate-then-commit** approach:

**Phase 1 — validate ALL before committing any:**
- Load each table's `TableOperations`
- Validate ALL `UpdateRequirement`s against current metadata
- Build the updated `TableMetadata` for each table
- If any requirement fails → reject the whole transaction, **zero tables modified**

**Phase 2 — commit all (only reached if phase 1 fully passes):**
- Call `TableOperations.commit(base, updated)` for each table
- Update the metadata cache per table

This is the standard pattern for Iceberg REST catalog server implementations (best-effort atomicity). True cross-table rollback after a phase-2 commit failure requires catalog-level transaction support (e.g., atomic storage rename), which varies by backend. The two-phase approach ensures **no table is modified when any pre-condition fails**, which is the most important guarantee clients rely on.

bharos (commented):
.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTransactionOperations.java
```diff
@@ -0,0 +1,104 @@
...
+  @Path("commit")
+  @Produces(MediaType.APPLICATION_JSON)
+  @Timed(name = "commit-transaction." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "commit-transaction", absolute = true)
+  @AuthorizationExpression(
+      expression =
+          "ANY(OWNER, METALAKE, CATALOG) || "
+              + "SCHEMA_OWNER_WITH_USE_CATALOG || "
+              + "ANY_USE_CATALOG",
+      accessMetadataType = MetadataObject.Type.CATALOG)
```
bharos, line 76: This checks catalog-level permissions (`ANY_USE_CATALOG`), while the `updateTable` endpoint requires table-level permissions (`TABLE::OWNER || ANY_MODIFY_TABLE`). A user with only `USE_CATALOG` could modify tables via `commitTransaction` that they couldn't modify via `updateTable`. Worth considering per-table authorization checks within the dispatcher chain.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
```diff
@@ -251,6 +257,70 @@ public LoadTableResponse updateTable(IcebergTableChange icebergTableChange) {
...
+   * <p>Each {@link UpdateTableRequest} in the request must include a {@link TableIdentifier}.
+   *
+   * @param commitTransactionRequest The request containing all table changes to apply.
+   */
+  public void commitTransaction(CommitTransactionRequest commitTransactionRequest) {
+    commitTransactionRequest.validate();
+    List<UpdateTableRequest> tableChanges = commitTransactionRequest.tableChanges();
+
+    // Phase 1: validate ALL requirements before committing anything
+    List<org.apache.iceberg.TableOperations> allOps = new ArrayList<>(tableChanges.size());
```
bharos, line 285: Nit: `org.apache.iceberg.TableOperations` is used as a fully qualified name inline. Per project conventions (AGENTS.md), this should use a standard `import` statement — no naming conflict here.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTransactionOperations.java
```diff
@@ -0,0 +1,195 @@
...
+    createTable(namespace, "txn_table1");
+    createTable(namespace, "txn_table2");
+
+    TableMetadata meta1 = loadTableMeta(namespace, "txn_table1");
+    TableMetadata meta2 = loadTableMeta(namespace, "txn_table2");
+
+    CommitTransactionRequest request = buildCommitRequest(namespace, meta1, meta2);
+    Response response = doCommitTransaction(request);
+
+    Assertions.assertEquals(Status.NO_CONTENT.getStatusCode(), response.getStatus());
```
bharos, line 107: This verifies HTTP 204 but doesn't confirm the tables were actually modified. Loading both tables after commit and asserting the new schema columns would strengthen confidence.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTransactionOperations.java
```diff
@@ -0,0 +1,195 @@
...
+            missingId, List.of(), List.of(new MetadataUpdate.AddSchema(updatedSchema)));
+    CommitTransactionRequest request = new CommitTransactionRequest(List.of(change));
+    Response response = doCommitTransaction(request);
+
+    Assertions.assertEquals(Status.NOT_FOUND.getStatusCode(), response.getStatus());
+    Assertions.assertTrue(
+        dummyEventListener.popPreEvent() instanceof IcebergCommitTransactionPreEvent);
+    Assertions.assertTrue(
+        dummyEventListener.popPostEvent() instanceof IcebergCommitTransactionFailureEvent);
+  }
```
bharos, line 131: Consider adding a test for *requirement validation failure* — e.g., two tables where one has a stale UUID. This is the core guarantee of the two-phase approach: no table should be modified when any pre-condition fails.

bharos (commented):
Thanks for the contribution!  A few comments:

bharos (commented):
.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTransactionOperations.java
bharos, line 76: This checks catalog-level permissions (`ANY_USE_CATALOG`), while the `updateTable` endpoint requires table-level permissions (`TABLE::OWNER || ANY_MODIFY_TABLE`). A user with only `USE_CATALOG` could modify tables via `commitTransaction` that they couldn't modify via `updateTable`. Worth considering per-table authorization checks within the dispatcher chain.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java
bharos, line 285: Nit: `org.apache.iceberg.TableOperations` is used as a fully qualified name inline. Per project conventions (AGENTS.md), this should use a standard `import` statement — no naming conflict here.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTransactionOperations.java
bharos, line 107: This verifies HTTP 204 but doesn't confirm the tables were actually modified. Loading both tables after commit and asserting the new schema columns would strengthen confidence.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTransactionOperations.java
bharos, line 131: Consider adding a test for *requirement validation failure* — e.g., two tables where one has a stale UUID. This is the core guarantee of the two-phase approach: no table should be modified when any pre-condition fails.

FANNG1:
@jerryshao @royi My biggest concern is still the `atomic` semantics.

The current two-phase `validate-then-commit` approach is definitely better than directly committing table by table, because it avoids cases where some requirements already fail before any commit starts. But once it enters the commit phase, it is still only `best-effort`, not a truly atomic multi-table commit. If one table commit succeeds and a later one fails, we can still end up with a partial-success state.

So from my perspective, real atomicity has to come from the underlying catalog/metastore capability, not just from the REST layer. For Gravitino-owned catalogs, a stronger solution would likely require something like staging multi-table metadata changes first, then switching metadata pointers together with a final atomic/CAS-style step.

Polaris seems to go a bit further in this direction, but it is still important to be precise about the semantics. Its `commitTransaction` is not a full distributed 2PC either. As I understand it, Polaris:

- checks authorization for each table in the request
- groups and validates updates per table
- stages metastore entity changes in a transaction workspace
- only at the end performs a batch CAS to switch multiple metadata pointers together

That gives stronger semantics for Polaris-managed metadata pointers, but it is still not fully global atomic commit, since metadata files may already have been written before the final CAS, and failures can still leave unreferenced metadata behind.

Would like to get your thoughts: should we first merge a `best-effort` implementation like this, with the semantics clearly documented, or should we wait until we have stronger catalog-level support before exposing this endpoint?

bharos:
@FANNG1 I agree with your thoughts. Two main concerns now are:

**1. Atomicity limitation**
 The core challenge is that TableOperations.commit(base, updated) is a per-table CAS where each call independently swaps one table's metadata_location pointer. For JDBC-backed catalogs, this is a single UPDATE ... SET metadata_location = ? WHERE metadata_location = ? on its own DB connection (JdbcUtil.update()). There's no way to batch N TableOperations.commit() calls into one atomic operation through the existing interface — Iceberg's own [test adapter has the same limitation](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java#L554-L556) and explicitly documents it.

To achieve true atomicity, Gravitino needs to bypass TableOperations for the pointer-swap step and batch the updates directly in the backing store. For JDBC-backed catalogs, that means running all pointer CAS updates in a single DB transaction:

```
Phase 1 (validate): unchanged — validate all requirements, write metadata files
Phase 2 (atomic swap):
  BEGIN;
    for each table:
      UPDATE iceberg_tables
        SET metadata_location = :newLoc
        WHERE catalog_name = :cat AND table_namespace = :ns
          AND table_name = :name AND metadata_location = :oldLoc;
      -- if affected_rows != 1 → ROLLBACK (concurrent modification)
  COMMIT;  -- all pointers flip together, or none do
  ```
If Phase 2 rolls back, the metadata files written in Phase 1 become orphaned on object storage, but this is the same as any failed single-table commit in Iceberg — unreferenced metadata files are cleaned up by normal table maintenance (RemoveOrphanFiles).

The key requirement is that the catalog server has direct access to the backing store's transaction primitives, rather than going through TableOperations.commit() which encapsulates each table's CAS independently.

**2. Security escalation**
The endpoint checks catalog-level permissions (ANY_USE_CATALOG), while updateTable requires table-level permissions (TABLE::OWNER || ANY_MODIFY_TABLE). A user with only USE_CATALOG could modify tables via commitTransaction that they couldn't touch via updateTable. Per-table authorization checks need to happen within the dispatcher chain for each table in the transaction.

I think Multi-table atomicity deserves a design doc before implementation

laserninja:
Thanks @FANNG1 @bharos for the thorough feedback, both points are valid.

**Atomicity**: I agree that best-effort semantics through `TableOperations.commit()` isn't sufficient for `commitTransaction`. True atomicity requires batching the pointer-swap CAS updates in the backing store's own transaction, which is catalog-backend-specific and warrants a proper design.

**Security**: Yes, the current implementation is missing per-table authorization checks within the transaction. Each `TableChange` should be validated against the same table-level permissions as `updateTable`.

I can put together a design doc covering the atomicity approach (backend-specific batched CAS), per-table authorization, and error/rollback semantics. Happy to close this PR in the meantime and reopen once the design is agreed upon.

lasdf1234:
@laserninja  If you have time, you can first write a design document. Thank you. I would be more than happy to review it.

lasdf1234:
Another question I have is: Is it true that only IRC (JDBC) can achieve true CAS? Is it the case that IRC(HIVE) and IRC(REST) cannot achieve this true CAS?

laserninja:
Thanks @lasdf1234. I've opened a design doc for review: https://github.com/apache/gravitino/pull/12195

On your CAS question, this turned out to be the crux of the design, so let me answer it directly.

Per-table CAS works on every backend, that's how ordinary single-table commits work. The thing that
differs is **multi-table atomic commit**, and there the backends are genuinely not equivalent. Checked
against Iceberg 1.11.0:

| Backend | Multi-table atomic commit | Why |
|---------|---------------------------|-----|
| `JDBC` | **Yes, we can build it** | The pointer is a row in `iceberg_tables`. N conditional `UPDATE ... WHERE metadata_location = :old` statements issued on one `Connection` with `setAutoCommit(false)` commit or roll back together. |
| `REST` | **Yes, by delegation** | The whole transaction can be forwarded upstream as a single `POST /v1/{prefix}/transactions/commit`. Our guarantee then equals the upstream catalog's, no stronger. |
| `HIVE` | **No** | `HiveTableOperations.doCommit` issues a per-table `alter_table` through HMS Thrift, each its own server-side transaction, guarded by `MetastoreLock` or `NoLock`. HMS exposes no multi-table transactional alter. |
| `MEMORY` | Yes (JVM lock) | Test-only. |
| `CUSTOM` | Unknown | Arbitrary `Catalog` impl, no capability contract. |

So: **JDBC is the only backend where Gravitino itself can construct atomicity. REST can inherit it by
delegating. HIVE cannot, without an HMS-side change.**

Worth being precise about HIVE, since it's tempting to think locking solves it: taking `MetastoreLock`
on all N tables before committing does remove concurrent-writer races, but it does not make the
`alter_table` calls atomic. A failure after the third of five still leaves three tables advanced. The
most honest label is "serialized best-effort", which is not what `commitTransaction` promises.

That asymmetry is what the design is built around. Rather than one uniform best-effort path, it
proposes an explicit per-backend capability, and gates the `/v1/config` advertisement of
`Endpoint.V1_COMMIT_TRANSACTION` on it, so a client never sees the endpoint advertised unless the
guarantee behind it is real. Backends that cannot support it return `501` instead of silently
degrading. This follows what the config endpoint already does for view support via
`getEndpoints(supportsViewOperations)`.

Two implementation details I'd particularly like your view on:

1. **JDBC.** `JdbcUtil.updateTable` can't be composed into one transaction, because it borrows its own
   autocommit connection per call. `JdbcClientPool.run(Action)` gives us one `Connection` for a whole
   lambda, which is the primitive we need. The catch is that Iceberg's commit SQL constants are
   `private`, so we'd have to restate the statement. The doc proposes we also raise this upstream and
   add an equivalence test in the meantime so a schema change upstream fails our build rather than
   silently corrupting commits.

2. **REST.** `RESTCatalog.commitTransaction(List<TableCommit>)` looks like the obvious API but is not a
   faithful forward. Its only factory is `TableCommit.create(identifier, base, updated)`, which derives
   the payload via `UpdateRequirements.forUpdateTable(...)` and therefore **discards the client's own
   `UpdateRequirement`s**, besides needing N extra loads to obtain `base`. The doc proposes a thin
   `RESTCatalog` subclass that forwards the inbound `CommitTransactionRequest` unmodified instead, so
   the upstream enforces the client's intended preconditions.

@FANNG1 @bharos the doc also addresses the two points you raised: it does not claim atomicity where the
backend cannot provide it, and it adds per-table authorization. On the latter, the
`@AuthorizationExpression` interceptor resolves entities from path parameters, and this endpoint carries
its tables in the body, so the doc does authorization programmatically via
`MetadataAuthzHelper.checkAccess` against the same expression `updateTable` uses, with that expression
extracted into a shared constant so the two paths can't drift.

Happy to close this PR while the design is under review and open a fresh implementation once it's
agreed, or leave it open as a reference if you'd find that more useful. Just let me know which you
prefer.

laserninja:
@FANNG1 @bharos when you have time, can you review this please ^

Issue #10656, opened 2026-04-02
[Improvement] Schema creation returns success when SchemaEntity persistence fails


justinmclean:
### What would you like to be improved?

SchemaOperationDispatcher#createSchema creates the schema in the underlying catalog and then persists the corresponding SchemaEntity to the Gravitino store. If store.put(schemaEntity, true) throws, the exception is caught, logged, and the method still returns a successful schema result.

This causes a false-success outcome:

The REST API returns success for the create request.
- Success events are emitted.
- The schema may exist in the external catalog, but Gravitino metadata was not persisted.

Although a later loadSchema may lazily import the missing metadata, that is only a best-effort repair after the fact. The original create operation has already reported success despite an internal persistence failure, which can lead to inconsistent behavior across audit/authorization/imported metadata.

/Users/justinmclean/gravitino/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java:157
Possible solution

Fail the create operation when store.put(...) fails instead of swallowing the exception.

Options:

Propagate a runtime exception after logging so the API returns failure.
If partial creation in the underlying catalog must be avoided, attempt rollback of the external schema creation before failing.
Add a test covering entity-store failure during schema creation and verifying the request does not return success.

### How should we improve?

See gravitino/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java:157

Fail the create operation when store.put(...) fails instead of swallowing the exception.

Options:
- Propagate a runtime exception after logging so the API returns failure.
- If partial creation in the underlying catalog must be avoided, attempt rollback of the external schema creation before failing.

Here's a unit test to help:
```

  @Test
  public void testCreateSchemaShouldFailWhenStorePutFails() throws IOException {
    Namespace ns = Namespace.of(metalake, catalog);
    NameIdentifier schemaIdent = NameIdentifier.of(ns, "schema2");
    Map<String, String> props = ImmutableMap.of("k1", "v1", "k2", "v2");

    doThrow(new IOException("store put failed")).when(entityStore).put(any(), anyBoolean());

    Assertions.assertThrows(
        RuntimeException.class, () -> dispatcher.createSchema(schemaIdent, "comment", props));
    Assertions.assertFalse(entityStore.exists(schemaIdent, SCHEMA));
    Assertions.assertThrows(
        NoSuchEntityException.class,
        () -> entityStore.get(schemaIdent, SCHEMA, SchemaEntity.class));
  }
```

[...]

PR #10676, opened 2026-04-04
[#10656] improvement(core): Centralize schema persistence error messa…

babumahesh:
                                                                                                                             
  Centralizes schema persistence error messages and adds automatic rollback functionality when schema metadata fails to
  persist to Gravitino's store after successful creation in the underlying catalog. The changes include:                     
   
  1. Added centralized error message constants in OperationDispatcher.FormattedErrorMessages for schema persistence failures 
  2. Implemented rollbackEntityCreation() method to automatically delete the schema from the underlying catalog when
  Gravitino store persistence fails                                                                                          
  3. Enhanced SchemaOperationDispatcher.createSchema() to attempt rollback on store failures and throw appropriate exceptions
   based on rollback outcome                                                                                                 
                  
  When a schema is successfully created in the underlying catalog but fails to persist to Gravitino's metadata store (e.g.,  
  due to network issues, storage failures), it creates an inconsistency where the schema exists in the catalog but not in
  Gravitino's records. This leads to:                                                                                        
  - Orphaned resources that Gravitino cannot manage
  - Confusing state where subsequent operations may fail unexpectedly                                                        
                                                                                                                             
  The automatic rollback mechanism ensures consistency between the underlying catalog and Gravitino's metadata store,        
  preventing orphaned resources and maintaining system integrity.                                                            
                                                                                                                             
New unit tests in TestSchemaOperationDispatcher verify successful rollbacks and rollback failures that retain the original 
  error reason back to callers.
Additionally, on the UI, an error message popup is displayed when failures occur.       

<img width="2536" height="1353" alt="Screenshot From 2026-04-05 00-49-11" src="https://github.com/user-attachments/assets/65b680f1-95b7-44a1-aeef-ca020d3c3549" />

<img width="1592" height="1034" alt="image" src="https://github.com/user-attachments/assets/1b6101a3-032e-4881-b6c0-730cb20f70df" />

                         
                                                                                                                             
                                                                                                                             
  Fix: #10656

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/OperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/OperationDispatcher.java
@@ -34,6 +34,8 @@ import org.apache.gravitino.StringIdentifier;
 import org.apache.gravitino.connector.HasPropertyMetadata;
 import org.apache.gravitino.connector.PropertiesMetadata;
 import org.apache.gravitino.connector.capability.Capability;
+import org.apache.gravitino.exceptions.GravitinoRuntimeException;
+import org.apache.gravitino.exceptions.NoSuchCatalogException;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
 import org.apache.gravitino.file.FilesetChange;
 import org.apache.gravitino.messaging.TopicChange;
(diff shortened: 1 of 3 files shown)


babumahesh:
@justinmclean Could you please review the PR and provide your feedback on the following:

1. In case of any failure during the store call, we propagate the error back to the caller, including the stack trace, The existing REST API behavior returns the stack trace in the response[check screenshot] — is this acceptable, or should we change it?
2. On the UI popup, we currently display the error message. This should be fine, but please let me know if we should improve or modify the message.

justinmclean:
For this PR, I think the current behavior is acceptable. On the UI side, showing the error message in the popup is fine. The current approach looks good to me.

justinmclean (approved):
Thanks for working on this.

mchades:
I will revert this PR because it is a breaking change.

For more details, please see #10763.

Issue #10630, opened 2026-04-01
[Improvement] createTable may return success even when Gravitino metadata persistence fails


justinmclean:
### What would you like to be improved?

In TableOperationDispatcher.createTable(...), Gravitino first creates the table in the underlying catalog, then writes the TableEntity to the entity store.

If store.put(tableEntity, true) fails, the exception is logged but swallowed, and the method still returns success.

This can leave the external table created, even though Gravitino metadata was not persisted. Clients will receive a successful create-table response even though Gravitino is in an inconsistent state. 

Location is TableOperationDispatcher.java (line 640)

### How should we improve?

Do not swallow the store.put(...) failure. Propagate the exception so create-table fails instead of returning false success OR if best-effort behavior is intended, return an explicit partial-success/inconsistent-state result and add a test for this path.

Add/modify an existing unit test to check for this.

Roshan1299:
Hi @justinmclean , I'd like to work on this! Can I be assigned?

Roshan1299:
Hi @justinmclean , I've opened a PR for this: https://github.com/apache/gravitino/pull/10677 . Please let me know if any changes are needed!

Roshan1299:
The PR was closed as @roryqi noted the current behavior is by design, entity store failure shouldn't block underlying catalog operations.

However, the issue description mentions this leaves Gravitino in an inconsistent state where the table exists in the catalog but has no persisted metadata, and future cleanup runs won't discover it.

Should the alternative approach be pursued instead — adding a fallback cleanup path or returning an explicit partial success/inconsistent-state result? @justinmclean  @roryqi

PR #10677, opened 2026-04-05
[#10630] fix(core): propagate store.put failure in createTable

Roshan1299:

Propagated the exception from `store.put(tableEntity)` in `TableOperationDispatcher.createTable()` instead of swallowing it. Updated the existing test to verify the exception is thrown.

Previously, if `store.put(tableEntity, true)` failed, the exception was logged but swallowed, and the method returned a successful response without the persisted entity. This left Gravitino in an inconsistent state, the table existed in the external catalog but had no corresponding metadata in the entity store. Clients received a false success response.

Fix: #10630 

Yes. `createTable` now throws a `RuntimeException` instead of returning a partial success when metadata persistence fails.

Updated the existing test `testCreateAndListTables` in `TestTableOperationDispatcher` to assert that a `RuntimeException` is thrown when `entityStore.put` fails, instead of expecting a successful return.


yuqi1129:
@roryqi 
Could you please check whether this is by design?

PR #10678, opened 2026-04-05
[#10458] fix(core): Include VIEW in cascade owner relation cleanup for schema and catalog deletion

### LINKED ISSUE #10458: [Improvement] Clean zombie owner relation about view

When deleting the schema or the catalog , we should delete views owner relation under the schema or the catalog

### How should we improve?

When deleting the schema or the catalog, we should delete views owner relation under the schema or the catalog

sachinnn99:

Added VIEW to the UNION queries in `softDeleteOwnerRelByCatalogId` and `softDeleteOwnerRelBySchemaId` in both the base and PostgreSQL SQL providers, so that view owner relations are cleaned up during cascade schema/catalog deletion.

When a schema or catalog is deleted in cascade mode, the bulk SQL queries that soft-delete child owner relations include TABLE, TOPIC, FILESET, and MODEL but miss VIEW. This leaves orphaned ("zombie") view owner relation rows in the database.

Note: individual view deletion via `ViewMetaService.deleteView()` already cleans up owner relations correctly — the bug only affects the cascade path from schema/catalog deletion.

Fix: #10458

No.

Updated `TestOwnerMetaService.testDeleteMetadataObject()` to include view entities in all test blocks (individual delete, cascade catalog delete, cascade schema delete, delete user). The cascade tests verify that `countActiveOwnerRel` drops to 0 after cascade catalog deletion and to 1 (only catalog survives) after cascade schema deletion — these assertions would fail without the SQL fix.

`./gradlew :core:test --tests "*TestOwnerMetaService*" -PskipITs` passes.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/OwnerMetaBaseSQLProvider.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/OwnerMetaBaseSQLProvider.java
@@ -29,6 +29,7 @@ import org.apache.gravitino.storage.relational.mapper.SchemaMetaMapper;
 import org.apache.gravitino.storage.relational.mapper.TableMetaMapper;
 import org.apache.gravitino.storage.relational.mapper.TopicMetaMapper;
 import org.apache.gravitino.storage.relational.mapper.UserMetaMapper;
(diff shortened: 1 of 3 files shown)


roryqi (approved):
LGTM, thanks for your contribution.

sachinnn99:
Thanks for the review and merge, @roryqi!

PR #9766, opened 2026-01-21
[#9765] improvement(iceberg-rest-server): Improve performance of loading tables

roryqi:
### What changes were proposed in this pull request?

Improve the performance of loading tables

### Why are the changes needed?

This improvement reuse the method in https://github.com/apache/gravitino/pull/6576

Fix #9765 

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Passed performance tests.

PR #10681, opened 2026-04-05
[#10633] fix(docker): run containers as non-root user

### LINKED ISSUE #10633: [Improvement] Support running Docker containers as a non-root user

Currently, the following Docker images run as the root user by default:
- `gravitino` (`dev/docker/gravitino/Dockerfile`)
- `iceberg-rest-server` (`dev/docker/iceberg-rest-server/Dockerfile`)
- `lance-rest-server` (`dev/docker/lance-rest-server/Dockerfile`)

All three Dockerfiles use `/root` as the working directory and do not specify a non-root `USER` instruction. This creates security concerns in production environments and prevents deployment in hardened Kubernetes clusters or platforms like OpenShift that enforce non-root container policies.

### How should we improve?

Modify the Dockerfiles to:
1. Create a dedicated non-root user (e.g., `gravitino`)
2. Change the working directory from `/root/*` to `/opt/*` or another appropriate location
3. Set proper file ownership and permissions
4. Add a `USER` instruction to run the container as the non-root user

Example changes for `gravitino/Dockerfile`:
```dockerfile
RUN groupadd -r gravitino && useradd -r -g gravitino gravitino

WORKDIR /opt/gravitino
COPY packages/gravitino /opt/gravitino

RUN chown -R gravitino:gravitino /opt/gravitino \
    && chmod +x /opt/gravitino/bin/start-gravitino.sh

USER gravitino
```

Similar changes should be applied to `iceberg-rest-server` and `lance-rest-server` Dockerfiles

laserninja:

Modifies the three production Dockerfiles to run as a dedicated non-root `gravitino` user (UID 1000) instead of `root`, and updates all Helm chart configuration to match the new paths.

**Dockerfiles changed:**
- `dev/docker/gravitino/Dockerfile`
- `dev/docker/iceberg-rest-server/Dockerfile`
- `dev/docker/lance-rest-server/Dockerfile`

**Helm charts updated:**
- `dev/charts/gravitino/` (templates + values)
- `dev/charts/gravitino-iceberg-rest-server/` (templates + values)
- `dev/charts/gravitino-lance-rest-server/` (templates + values)
- `dev/charts/gravitino/resources/scenarios/ci-values.yaml`

Fixes #10633. Running containers as `root` is a security concern in production environments and prevents deployment on hardened Kubernetes clusters or OpenShift platforms that enforce non-root container policies.

## How was this patch tested?

The changes are limited to Dockerfiles and Helm chart configuration. No Java code was modified. The diff was manually reviewed for correctness.

- [ ] Existing tests pass (no Java changes)
- [ ] Dockerfile security: containers now run as UID 1000 (`gravitino` user)
- [ ] Helm charts: `runAsNonRoot: true`, `runAsUser: 1000` in container security contexts
- [ ] All `/root/gravitino*` paths updated to `/opt/gravitino*`

### DIFF
--- a/dev/docker/gravitino/Dockerfile
+++ b/dev/docker/gravitino/Dockerfile
@@ -22,7 +22,7 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-RUN groupadd -g 1000 gravitino && useradd -u 1000 -g gravitino -m gravitino
+RUN groupadd -r gravitino && useradd -r -g gravitino gravitino
 
 WORKDIR /opt/gravitino
 
--- a/dev/docker/gravitino/Dockerfile
+++ b/dev/docker/gravitino/Dockerfile
@@ -22,13 +22,18 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-WORKDIR /root/gravitino
+RUN groupadd -g 1000 gravitino && useradd -u 1000 -g gravitino -m gravitino
 
-COPY packages/gravitino /root/gravitino
+WORKDIR /opt/gravitino
+
+COPY packages/gravitino /opt/gravitino
+
+RUN chown -R gravitino:gravitino /opt/gravitino \
+    && chmod +x /opt/gravitino/bin/start-gravitino.sh
 
 EXPOSE 8090
 EXPOSE 9001
 
-RUN chmod +x /root/gravitino/bin/start-gravitino.sh
+USER gravitino
 
-ENTRYPOINT ["/bin/bash", "/root/gravitino/bin/start-gravitino.sh"]
+ENTRYPOINT ["/bin/bash", "/opt/gravitino/bin/start-gravitino.sh"]
--- a/dev/docker/gravitino/Dockerfile
+++ b/dev/docker/gravitino/Dockerfile
@@ -22,18 +22,16 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-RUN groupadd -r gravitino && useradd -r -g gravitino gravitino
-
 WORKDIR /opt/gravitino
 
 COPY packages/gravitino /opt/gravitino
 
-RUN chown -R gravitino:gravitino /opt/gravitino \
+RUN chmod -R g+rwX /opt/gravitino \
     && chmod +x /opt/gravitino/bin/start-gravitino.sh
 
 EXPOSE 8090
 EXPOSE 9001
 
-USER gravitino
+USER 1000
 
 ENTRYPOINT ["/bin/bash", "/opt/gravitino/bin/start-gravitino.sh"]
--- a/dev/docker/iceberg-rest-server/Dockerfile
+++ b/dev/docker/iceberg-rest-server/Dockerfile
@@ -23,7 +23,7 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-RUN groupadd -g 1000 gravitino && useradd -u 1000 -g gravitino -m gravitino
+RUN groupadd -r gravitino && useradd -r -g gravitino gravitino
 
 WORKDIR /opt/gravitino-iceberg-rest-server
 
--- a/dev/docker/iceberg-rest-server/Dockerfile
+++ b/dev/docker/iceberg-rest-server/Dockerfile
@@ -23,17 +23,15 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-RUN groupadd -r gravitino && useradd -r -g gravitino gravitino
-
 WORKDIR /opt/gravitino-iceberg-rest-server
 
 COPY packages/gravitino-iceberg-rest-server /opt/gravitino-iceberg-rest-server
 
-RUN chown -R gravitino:gravitino /opt/gravitino-iceberg-rest-server \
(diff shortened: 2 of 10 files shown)


danhuawang:
Could you help add the `podSecurityContext` for the gravitino deployment too ?  
```
      {{- with .Values.podSecurityContext }}
      securityContext:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

laserninja:
Done

jerryshao:
@danhuawang please help to review this PR.

dev/docker/iceberg-rest-server/Dockerfile
```diff
@@ -23,10 +23,17 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-WORKDIR /root/gravitino-iceberg-rest-server
+RUN groupadd -g 1000 gravitino && useradd -u 1000 -g gravitino -m gravitino
```
danhuawang: The gravitino user's GID and UID are hard code to 1000, can we not hard code them ?

dev/docker/gravitino/Dockerfile
```diff
@@ -22,13 +22,18 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-WORKDIR /root/gravitino
+RUN groupadd -g 1000 gravitino && useradd -u 1000 -g gravitino -m gravitino
```
danhuawang: The gravitino user's GID and UID are hard code to 1000, can we not hard code them ? Because some security-sensitive k8s  env assign dynamic UIDs.

dev/docker/lance-rest-server/Dockerfile
```diff
@@ -23,10 +23,17 @@ LABEL maintainer="dev@gravitino.apache.org"
 RUN apt-get update && apt-get install -y \
     python-is-python3
 
-WORKDIR /root/gravitino-lance-rest-server
+RUN groupadd -g 1000 gravitino && useradd -u 1000 -g gravitino -m gravitino
```
danhuawang: The gravitino user's GID and UID are hard code to 1000, can we not hard code them ?

laserninja:
Replaced hardcoded groupadd -g 1000 / useradd -u 1000 with groupadd -r / useradd -r (system-assigned IDs) in all three Dockerfiles, so dynamic-UID environments work correctly
Removed runAsUser: 1000 from Helm securityContext defaults; runAsNonRoot: true is sufficient without pinning a UID

Added securityContext: runAsNonRoot: true to all three initContainers (sqlfile, init-mysql, init-postgresql) via a new initContainerSecurityContext value — fixes compatibility with the Kubernetes restricted Pod Security Standard

Thanks for the review.

laserninja:
The test is failing. Is it too bad to use hardcoded ID? I also checked Trino image https://github.com/trinodb/trino/blob/master/core/docker/Dockerfile#L62. Let me know what we can use.

danhuawang:
I refer the bitnami approach the  https://github.com/bitnami/containers/blob/main/bitnami/mysql/9.6/debian-12/Dockerfile , they do not create user/group, only set directory permissions, use UID directly.  
``` 
RUN chmod -R g+rwX /opt/gravitino
USER 1000
```

laserninja:
Matched it similar to mysql. Can you check now?

danhuawang:
LGTM

jerryshao (approved):

shunki-fujita:
@danhuawang @laserninja 
The Bitnami approach works for MySQL because it doesn't require `getpwuid()` at runtime. Unfortunately, Gravitino has a Hadoop dependency that does.

Since v1.2.0 (PR #9766), https://github.com/apache/gravitino/blob/d2268ac15d8e90cbdf929ccf2dde83e988751f64/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java#L105 calls
  https://github.com/apache/gravitino/blob/d2268ac15d8e90cbdf929ccf2dde83e988751f64/catalogs/hadoop-common/src/main/java/org/apache/gravitino/catalog/hadoop/fs/FileSystemUtils.java#L219, which triggers Configuration.addResource() →  UserGroupInformation.getCurrentUser() → UnixLoginModule.login(). This calls getpwuid() to resolve the UID, and without a /etc/passwd entry it throws NullPointerException.

This doesn't happen on the first request (Hadoop's UGI is uninitialized at startup), but after the catalog cache TTL (default 1h) expires and the wrapper is re-created, it crashes with HTTP 500.

I confirmed this with the iceberg-rest-server image built from this PR's Dockerfile (cache TTL = 10s):

                            Initial    After eviction
    This PR (no useradd):   200        500
    This PR + useradd:      200        200
    v1.2.0 as root:         200        200

```
RuntimeException: Failed to create configuration
    at FileSystemUtils.createConfiguration(FileSystemUtils.java:223)
    at IcebergCatalogWrapper.<init>(IcebergCatalogWrapper.java:105)
Caused by: LoginException: NullPointerException: invalid null input: name
    at UnixLoginModule.login
```

I think `useradd` needs to be added back to the Dockerfiles.

qconlulu:
> , which triggers Configuration.addResource() → UserGroupInformation.getCurrentUser() → UnixLoginModule.login(). This calls getpwuid() to resolve the UID, and without a /etc/passwd entry it throws NullPointerException.

@shunki-fujita Thanks for your feedback. Would you like to fix this issue?

laserninja:
Thanks for pinging out the issue. Let me know if you want me to fix it. Either way is fine.

shunki-fujita:
@laserninja 
I opened a PR for this!
https://github.com/apache/gravitino/pull/10817

The affected Dockerfiles are `gravitino` and `iceberg-rest-server`. 
`lance-rest-server` is not affected as it has no Hadoop dependency.

PR #10696, opened 2026-04-07
[#10695] feat(cache): Introduce Gravitino Cache Improvement Design document outlining new caching strategies and architecture

### LINKED ISSUE #10695: [Subtask] Design document about cache improvement
### Describe the subtask

This issue tracks the design docment of cache improvement.

### Parent issue

#9898

yuqi1129:

Raise a document about cache improvement.

It's a feature.

Fix: #10695

N/A


design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,882 @@
...
+│
+│   userRoleCache HIT and role_grants_version matches:
+│     → use cached userId and roleIds               [A] and [C1] avoided
+│
+│   MISS or version mismatch:
+│     → SELECT role_id FROM user_role_rel WHERE user_id = ? AND deleted_at = 0
+│     → re-associate userId ↔ roleIds in JCasbin enforcers
+│     → userRoleCache.put(key, new CachedUserRoles(userId, version, roleIds))
+│
+├─ STEP 2 — Resolve target resource ID (always DB, no cache):
```
yuqi1129: No good method to keep a strong ID to name mapping.

roryqi:
cc @bharos @gada121982

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,686 @@
...
+| **User / Group drop + same-name recreate**    | The new entity gets a new auto-increment `user_id` and `role_grants_version = 1`. The cached entry holds the old `user_id` and an older version → **version mismatch on the next Step 1 forces a cache refresh.** ✅            |
+| **SecurableObject rename**                    | JCasbin stores integer `metadataId`. Rename does not change the ID. Step 2 resolves the new name to the same ID via DB. `enforce()` matches the existing policy. **No action needed.** ✅                                       |
+| **SecurableObject drop**                      | Step 2 returns "not found" → auth denied. Orphan JCasbin policies remain in memory but can never be matched (no ID resolves to the dropped object). **Safe.**                                                                   |
+| **SecurableObject drop + same-name recreate** | The new object gets a new `metadataId`. No JCasbin policy covers it → DENY until a new privilege grant bumps `securable_objects_version` in the same transaction and Step 3 detects the version change to reload policies. **Correct.** |
+
+#### 4.1.6 Concurrent Mutation During Auth (TOCTOU)
```
yuqi1129: Add a description of the concurrent problem and its effect.

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,788 @@
...
+
+## 5. Phased Implementation Plan
+
+### Phase 1 — Foundation (common to both approaches, no schema changes)
+
+| Step  | Change                                                                               | Module                                               |
+|-------|--------------------------------------------------------------------------------------|------------------------------------------------------|
+| 1.1   | Fix auth N+1: `batchListSecurableObjectsByRoleIds()` + rewrite `loadRolePrivilege()` | `RoleMetaService`, `JcasbinAuthorizer`               |
+| 1.2   | Introduce `GravitinoCache<K,V>` interface; wrap existing Caffeine caches             | `GravitinoCache.java`, `CaffeineGravitinoCache.java` |
+| 1.3   | Disable entity store cache: `CACHE_ENABLED` default → `false`                        | `Configs.java`                                       |
```
bharos: Phase 1.3 flips CACHE_ENABLED to false, which means calls [A], [B], [C1] hit DB on every auth request with no replacement protection until Phase 2 ships. Probably this is a case to make Phase1 and Phase2 to ship in same release ?

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,788 @@
...
+`owner_meta` uses a soft-delete pattern for ownership transfers: the old ownership row is
+soft-deleted and a new row is inserted. The new row's version counter always starts at 1,
+so a version field on `owner_meta` cannot distinguish "freshly cached" from "ownership just
+transferred" — any new row looks identical to the cache.
+
+Adding an `owner_version` counter to every owned entity type (`catalog_meta`, `schema_meta`,
+`table_meta`, `fileset_meta`, `topic_meta`, `view_meta`, `model_meta`, ...) is invasive and
+couples the entity schema to the auth cache.
+
+**Resolution:** Remove the `ownerRel` cache entirely. `OWNER`-privilege checks are rare
```
bharos: Are the OWNER-privilege checks actually rare?

Looking at the actual authorization expressions in AuthorizationExpressionConstants.java, OWNER is the first branch in virtually every hot-path expression. For example, LOAD_TABLE_AUTHORIZATION_EXPRESSION expands to:
```
METALAKE::OWNER || CATALOG::OWNER || ... && (TABLE::OWNER || ANY_SELECT_TABLE || ...)
```

Each X::OWNER compiles to authorizer.isOwner(), which calls loadOwnerPolicy() → ownerRel cache lookup (or DB on miss). For a query that touches N tables, removing ownerRel and replacing it with a direct owner_meta query means up to 3×N additional DB queries per HTTP request (metalake + catalog + table ownership, all returning false for non-owner users) ?

yuqi1129: You are right, I have updated the document.

design-docs/cache-improvement-design.md
yuqi1129: Yeah, I will not change CACHE_ENABLED to `false` until all is done for phase 2.

design-docs/cache-improvement-design.md
roryqi: Eventual consistence is accepted for me.

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,873 @@
...
+for [B] is always correct (~1 ms indexed). Simpler and more correct to hit DB every request.
+
+**Why `ownerRel` is removed:**
+`isOwner()` is called 2–4 times per auth request (once per ancestor in the OGNL expression
+chain). A version-validated cache would still query `owner_meta` on every request to check
+versions — and that same query already returns the `owner_id`. There is no expensive
+downstream work to avoid (unlike roles, where skipping securable-object loading saves
+significant work). Caching adds complexity with zero query savings. Step 2.5 issues one
+batch query per request directly against `owner_meta`, achieving strong consistency with
+no extra infrastructure. See §7.2.
```
jerryshao: Most of the DB operations require joining multiple tables, I'm not sure your assumption of DB access (1ms) is valid or not.

jerryshao:
## Design Review Feedback

Thanks for the detailed design document! The version-validated cache approach is architecturally sound and the Polaris-inspired analysis is excellent. Here's design-level feedback before implementation begins:

---

### 🔴 Must Fix

**#1: Version column duplication violates existing schema patterns**
The proposed `role_meta.securable_objects_version`, `user_meta.role_grants_version`, etc. duplicate functionality already provided by the existing `current_version` / `last_version` columns present on **all 20+ metadata tables**. Adding a third versioning scheme creates confusion about which column to increment on a given mutation and creates migration complexity. Recommend either reusing `current_version` (increment on any mutation) or following the `table_column_version_info` / `function_version_info` pattern to add a dedicated version event table.

**#2: `GravitinoCache` interface is missing lock primitives → race conditions**
The proposed `GravitinoCache<K,V>` interface has no locking methods. The existing `EntityCache` interface defines three critical lock primitives (`withCacheLock`, `withMultipleKeyCacheLock`) backed by `SegmentedLock`. The multi-step read-check-update in the auth flow (§4.1.3 Step 1) has a classic race condition without them:
```
Thread A: reads version=5, cache has version=4 → starts reload
Thread B: reads version=5, cache has version=4 → starts reload (duplicate query, competing write)
```
`GravitinoCache` must expose the same locking primitives as `EntityCache`.

**#3: Group role loading is a new feature, not a caching optimization — scope creep**
§4.1.2 introduces `groupRoleCache` and §4.1.3 Step 1b adds group-based role loading. The doc itself notes: *"current code only loads user-direct roles (ROLE_USER_REL). Loading group roles via [1b] is a NEW capability."*
This changes the authorization permission model (users in groups now inherit group roles), which has a different risk profile, requires separate security testing, and affects audit compliance. Strongly recommend splitting:
- **Phase 0 (new PR):** Add group role inheritance using existing TTL-based caching
- **Phase 2 (this PR):** Migrate all caches to version-validated mechanism

---

### 🟡 Should Fix

**#4: Multi-entity version bump scenarios are underdefined**
The design shows version bumps for single-entity mutations but doesn't address complex cross-entity operations: ownership transfer, adding a user to a group with existing roles, privilege grant cascades. Does `user_meta` need a version bump when the user's group gains a new role? Add a decision table covering all mutation scenarios.

**#5: "Lightweight" query claims are unquantified**
Step 1a involves a JOIN across `metalake_meta` + `user_meta` — that's two index scans, not a trivial lookup. The Phase 2 hot path has 4 required DB queries per auth request. Please add benchmarking results (or commit to profiling before Phase 2 implementation) to validate the performance claims.

**#6: No degraded mode strategy for DB unavailability**
Phase 2 makes every auth request require live DB connectivity, removing the current passive resilience (today's entity cache can serve from memory during short DB outages within the TTL window). The statement *"DB unavailable → auth blocked (same as today)"* is inaccurate. Please add a §4.2 covering the chosen approach: stale-on-error, circuit breaker, or explicit acceptance that DB becomes a single point of failure for auth.

**#7: `ownerRel` replacement — implementation details missing**
The proposal to replace `ownerRel` with a request-scoped cache in `AuthorizationRequestContext` is sound, but the design omits: context lifecycle (ThreadLocal vs. request bean), cache key structure, and thread-safety for parallel auth checks. Please document the concrete `AuthorizationRequestContext` changes. Also, the existing `getID()` double-call bug (lines 224/228) should be fixed as a **separate commit** rather than bundled into Phase 2.

**#8: MyBatis mapper return type is semantically incorrect**
`Map<Long, Integer> batchGetSecurableObjectsVersions(List<Long> roleIds)` with `resultType="map"` won't work as intended — MyBatis returns `List<Map<String, Object>>` for that configuration, not a flattened `Map<Long, Integer>`. Recommend using a dedicated VO (`record RoleVersion(long roleId, int version) {}`) with a proper `<resultMap>`.

---

### ✅ Strong Points
- Polaris reference analysis (§3.1) is excellent industry research
- TOCTOU/race condition analysis (§4.1.6) is thorough
- Phase 1/2 split is well-reasoned and de-risks schema migration
- `ownerRel` decision log (§7.2) shows strong self-review

---

Overall the direction is right. The three 🔴 issues — especially scope creep on group roles and missing locking in the cache interface — should be resolved before moving to implementation.

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,873 @@
...
+    COMMENT 'Incremented atomically with any privilege grant/revoke for this role';
+
+ALTER TABLE `user_meta`
+    ADD COLUMN `role_grants_version` INT UNSIGNED NOT NULL DEFAULT 1
+    COMMENT 'Incremented atomically with any role assignment/revocation for this user';
+
+ALTER TABLE `group_meta`
+    ADD COLUMN `role_grants_version` INT UNSIGNED NOT NULL DEFAULT 1
+    COMMENT 'Incremented atomically with any role assignment/revocation for this group';
+```
```
jerryshao, line 408: Shall we do the same thing for the owner update of the entity?

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,873 @@
...
+│   JOIN group_user_rel gu ON gm.group_id = gu.group_id AND gu.deleted_at = 0
+│   WHERE gu.user_id = ? AND gm.deleted_at = 0
+│
+│   For each group:
+│     groupRoleCache HIT and role_grants_version matches:
+│       → use cached groupId and roleIds            [group C1] avoided
+│     MISS or version mismatch:
+│       → SELECT role_id FROM group_role_rel WHERE group_id = ? AND deleted_at = 0
+│       → addRoleForUser(userId, roleId) in JCasbin enforcers
+│       → groupRoleCache.put(groupKey, CachedGroupRoles(groupId, version, roleIds))
```
jerryshao: You need to confirm with @roryqi about the current group implementation. AFAIK, we don't have user-group mapping stored in Gravitino. So `group` is another special `user`.

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,873 @@
...
+├─ STEP 2 — Resolve target resource ID (always DB, no cache):
+│
+│   metadataId = MetadataIdConverter.getID(resource, metalake)  ← 1 indexed DB query
+│   Always correct: rename does not change ID; drop+recreate returns the new ID.
+│
+│   TODO: A strong-consistency name→id cache could eliminate this DB query on the warm
+│   path. Version-based validation does not apply here (checking the version requires
+│   the same query that returns the ID). A viable approach would require an
+│   entity_mutation_log for cross-node invalidation plus write-path eviction on the
+│   local node. Not implemented in this phase.
```
jerryshao: A final-consistency cache is acceptable; otherwise, the query needs to join multiple tables, I don't think the performance will be good. Also, we can cache something that is not changed frequently, let metalake, catalog id.

design-docs/cache-improvement-design.md
yuqi1129: The ROI of adding a version column is almost zero.  We need to check the version first when trying to get the owner of a securable object with a specific ID, then why not getting it directly?

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,873 @@
...
+| 2.9   | Rewrite `loadRolePrivilege()` + `authorize()` with 4-step flow (§4.1.3); remove `executor` thread pool (replaced by batch query) | `JcasbinAuthorizer`                                    |
+| 2.10  | Add mapper methods (see §6.1)                                                                                                   | mapper + SQL                                            |
+| 2.11  | Remove `ownerRel`; add `ownerCache: Map<Long, Optional<Long>>` to `AuthorizationRequestContext` for per-request dedup; fix double `getID()` call in `isOwner()` | `JcasbinAuthorizer`, `AuthorizationRequestContext` |
+| 2.12  | Extend `AuthorizationRequestContext.loadRole()` guard to cover both user and group role loading in one pass                     | `AuthorizationRequestContext`                           |
+
+**Outcome:** Zero staleness. Hot path: 4 DB queries (Step 1a user + Step 1b groups + Step 2 metadataId + Step 3 role versions). `isOwner()` deduped within request via context cache; always consistent.
+
+---
```
jerryshao, line 966: You should add a task to test the performance before and after.

design-docs/cache-improvement-design.md
roryqi: For SCIM, the token will contain the group information. 
For LDAP, we should get the group information from LDAP and cache the group information.

jerryshao:
> ## Design Review Feedback
> Thanks for the detailed design document! The version-validated cache approach is architecturally sound and the Polaris-inspired analysis is excellent. Here's design-level feedback before implementation begins:
> 
> ### 🔴 Must Fix
> **#1: Version column duplication violates existing schema patterns** The proposed `role_meta.securable_objects_version`, `user_meta.role_grants_version`, etc. duplicate functionality already provided by the existing `current_version` / `last_version` columns present on **all 20+ metadata tables**. Adding a third versioning scheme creates confusion about which column to increment on a given mutation and creates migration complexity. Recommend either reusing `current_version` (increment on any mutation) or following the `table_column_version_info` / `function_version_info` pattern to add a dedicated version event table.
> 
> **#2: `GravitinoCache` interface is missing lock primitives → race conditions** The proposed `GravitinoCache<K,V>` interface has no locking methods. The existing `EntityCache` interface defines three critical lock primitives (`withCacheLock`, `withMultipleKeyCacheLock`) backed by `SegmentedLock`. The multi-step read-check-update in the auth flow (§4.1.3 Step 1) has a classic race condition without them:
> 
> ```
> Thread A: reads version=5, cache has version=4 → starts reload
> Thread B: reads version=5, cache has version=4 → starts reload (duplicate query, competing write)
> ```
> 
> `GravitinoCache` must expose the same locking primitives as `EntityCache`.
> 
> **#3: Group role loading is a new feature, not a caching optimization — scope creep** §4.1.2 introduces `groupRoleCache` and §4.1.3 Step 1b adds group-based role loading. The doc itself notes: _"current code only loads user-direct roles (ROLE_USER_REL). Loading group roles via [1b] is a NEW capability."_ This changes the authorization permission model (users in groups now inherit group roles), which has a different risk profile, requires separate security testing, and affects audit compliance. Strongly recommend splitting:
> 
> * **Phase 0 (new PR):** Add group role inheritance using existing TTL-based caching
> * **Phase 2 (this PR):** Migrate all caches to version-validated mechanism
> 
> ### 🟡 Should Fix
> **#4: Multi-entity version bump scenarios are underdefined** The design shows version bumps for single-entity mutations but doesn't address complex cross-entity operations: ownership transfer, adding a user to a group with existing roles, privilege grant cascades. Does `user_meta` need a version bump when the user's group gains a new role? Add a decision table covering all mutation scenarios.
> 
> **#5: "Lightweight" query claims are unquantified** Step 1a involves a JOIN across `metalake_meta` + `user_meta` — that's two index scans, not a trivial lookup. The Phase 2 hot path has 4 required DB queries per auth request. Please add benchmarking results (or commit to profiling before Phase 2 implementation) to validate the performance claims.
> 
> **#6: No degraded mode strategy for DB unavailability** Phase 2 makes every auth request require live DB connectivity, removing the current passive resilience (today's entity cache can serve from memory during short DB outages within the TTL window). The statement _"DB unavailable → auth blocked (same as today)"_ is inaccurate. Please add a §4.2 covering the chosen approach: stale-on-error, circuit breaker, or explicit acceptance that DB becomes a single point of failure for auth.
> 
> **#7: `ownerRel` replacement — implementation details missing** The proposal to replace `ownerRel` with a request-scoped cache in `AuthorizationRequestContext` is sound, but the design omits: context lifecycle (ThreadLocal vs. request bean), cache key structure, and thread-safety for parallel auth checks. Please document the concrete `AuthorizationRequestContext` changes. Also, the existing `getID()` double-call bug (lines 224/228) should be fixed as a **separate commit** rather than bundled into Phase 2.
> 
> **#8: MyBatis mapper return type is semantically incorrect** `Map<Long, Integer> batchGetSecurableObjectsVersions(List<Long> roleIds)` with `resultType="map"` won't work as intended — MyBatis returns `List<Map<String, Object>>` for that configuration, not a flattened `Map<Long, Integer>`. Recommend using a dedicated VO (`record RoleVersion(long roleId, int version) {}`) with a proper `<resultMap>`.
> 
> ### ✅ Strong Points
> * Polaris reference analysis (§3.1) is excellent industry research
> * TOCTOU/race condition analysis (§4.1.6) is thorough
> * Phase 1/2 split is well-reasoned and de-risks schema migration
> * `ownerRel` decision log (§7.2) shows strong self-review
> 
> Overall the direction is right. The three 🔴 issues — especially scope creep on group roles and missing locking in the cache interface — should be resolved before moving to implementation.

 Please evaluate the AI feedbacks @yuqi1129 .

design-docs/cache-improvement-design.md
yuqi1129: So, what's your suggestion?

jerryshao:
You should reply to my comments about your thoughts, not just update the doc silently. It is hard to find out the changes.

design-docs/cache-improvement-design.md
yuqi1129: Added tasks

design-docs/cache-improvement-design.md
yuqi1129: I see. The id-name mapping and owner cache have been updated:

An `update_at` column has been added to the table `owner_meta` to mark whether there are changes in the owner relationship. At the same time,  a new table `entity_change_log` was created to indicate any updates to metadata objects.

yuqi1129:
> last_version



> > ## Design Review Feedback
> > Thanks for the detailed design document! The version-validated cache approach is architecturally sound and the Polaris-inspired analysis is excellent. Here's design-level feedback before implementation begins:
> > ### 🔴 Must Fix
> > **#1: Version column duplication violates existing schema patterns** The proposed `role_meta.securable_objects_version`, `user_meta.role_grants_version`, etc. duplicate functionality already provided by the existing `current_version` / `last_version` columns present on **all 20+ metadata tables**. Adding a third versioning scheme creates confusion about which column to increment on a given mutation and creates migration complexity. Recommend either reusing `current_version` (increment on any mutation) or following the `table_column_version_info` / `function_version_info` pattern to add a dedicated version event table.
> > **#2: `GravitinoCache` interface is missing lock primitives → race conditions** The proposed `GravitinoCache<K,V>` interface has no locking methods. The existing `EntityCache` interface defines three critical lock primitives (`withCacheLock`, `withMultipleKeyCacheLock`) backed by `SegmentedLock`. The multi-step read-check-update in the auth flow (§4.1.3 Step 1) has a classic race condition without them:
> > ```
> > Thread A: reads version=5, cache has version=4 → starts reload
> > Thread B: reads version=5, cache has version=4 → starts reload (duplicate query, competing write)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `GravitinoCache` must expose the same locking primitives as `EntityCache`.
> > **#3: Group role loading is a new feature, not a caching optimization — scope creep** §4.1.2 introduces `groupRoleCache` and §4.1.3 Step 1b adds group-based role loading. The doc itself notes: _"current code only loads user-direct roles (ROLE_USER_REL). Loading group roles via [1b] is a NEW capability."_ This changes the authorization permission model (users in groups now inherit group roles), which has a different risk profile, requires separate security testing, and affects audit compliance. Strongly recommend splitting:
> > 
> > * **Phase 0 (new PR):** Add group role inheritance using existing TTL-based caching
> > * **Phase 2 (this PR):** Migrate all caches to version-validated mechanism
> > 
> > ### 🟡 Should Fix
> > **#4: Multi-entity version bump scenarios are underdefined** The design shows version bumps for single-entity mutations but doesn't address complex cross-entity operations: ownership transfer, adding a user to a group with existing roles, privilege grant cascades. Does `user_meta` need a version bump when the user's group gains a new role? Add a decision table covering all mutation scenarios.
> > **#5: "Lightweight" query claims are unquantified** Step 1a involves a JOIN across `metalake_meta` + `user_meta` — that's two index scans, not a trivial lookup. The Phase 2 hot path has 4 required DB queries per auth request. Please add benchmarking results (or commit to profiling before Phase 2 implementation) to validate the performance claims.
> > **#6: No degraded mode strategy for DB unavailability** Phase 2 makes every auth request require live DB connectivity, removing the current passive resilience (today's entity cache can serve from memory during short DB outages within the TTL window). The statement _"DB unavailable → auth blocked (same as today)"_ is inaccurate. Please add a §4.2 covering the chosen approach: stale-on-error, circuit breaker, or explicit acceptance that DB becomes a single point of failure for auth.
> > **#7: `ownerRel` replacement — implementation details missing** The proposal to replace `ownerRel` with a request-scoped cache in `AuthorizationRequestContext` is sound, but the design omits: context lifecycle (ThreadLocal vs. request bean), cache key structure, and thread-safety for parallel auth checks. Please document the concrete `AuthorizationRequestContext` changes. Also, the existing `getID()` double-call bug (lines 224/228) should be fixed as a **separate commit** rather than bundled into Phase 2.
> > **#8: MyBatis mapper return type is semantically incorrect** `Map<Long, Integer> batchGetSecurableObjectsVersions(List<Long> roleIds)` with `resultType="map"` won't work as intended — MyBatis returns `List<Map<String, Object>>` for that configuration, not a flattened `Map<Long, Integer>`. Recommend using a dedicated VO (`record RoleVersion(long roleId, int version) {}`) with a proper `<resultMap>`.
> > ### ✅ Strong Points
> > 
> > * Polaris reference analysis (§3.1) is excellent industry research
> > * TOCTOU/race condition analysis (§4.1.6) is thorough
> > * Phase 1/2 split is well-reasoned and de-risks schema migration
> > * `ownerRel` decision log (§7.2) shows strong self-review
> > 
> > Overall the direction is right. The three 🔴 issues — especially scope creep on group roles and missing locking in the cache interface — should be resolved before moving to implementation.
> 
> Please evaluate the AI feedbacks @yuqi1129 .


> Version column duplication violates existing schema patterns

For the fields  `current_version` or `last_version` in most tables like `metalake_meta`, `catalog_meta`, and so on are always 1(default value) and can't satisfy our meet. 

Of course, we can fix it, but we need to manage at least 8(metalake, catalog, schema, fileset, table, model, view, topic) types of versions in the memory, it's very tedious.

> 2. GravitinoCache interface is missing lock primitives → race conditions

`GravitinoCache` is just an interface, and a detailed implementation will add the necessary locks to avoid race conditions.


> 3. Group role loading is a new feature, not a caching optimization 

This is a complement to the document and for future use only. It will not be implemented in the first stage.


> Multi-entity version bump scenarios are underdefined

I have covered the changes.


For 5, 6, 7. As we have adopted an eventually-consistent cache, there should not be a problem.


For 8, I have fixed it to use POJO NOT a map.

yuqi1129:
@jerryshao 

I have updated and replied to the comments. Still, I have the following concerns:

- We use version check to fulfill strong consistency for user/role, but still use eventual consistency for id-name mapping, ownership relationship. It does not look so elegant, as it will always introduce 2 extra DB queries compared to the previous one, and finally gain an eventual consistency. Theoretically speaking,  the performance will definitely degrade. Is that truly acceptable? I asked myself several times, "Why not also make the user/role eventually consistent, too ?" In this way, tiny changes to the table definition and a good performance will be realized (all use db-version + pull internally).  It would be preferable if you could give me some advice on this point. That is why it hinders me to add id-mapping cache.

- I have introduced a new table named `entity_change_log` to gain fine-grained recording of each ENTITY change. I'm not 100% sure whether it's the best option, as it will add a new table again. It's not very friendly for future refactoring.
Other options include adding `updated_at` for each table, employing `update-time` in the audit information, and using `current_version`/`last_version`, which are also not very desirable.

jerryshao:
> We use version check to fulfill strong consistency for user/role, but still use eventual consistency for id-name mapping, ownership relationship. It does not look so elegant, as it will always introduce 2 extra DB queries compared to the previous one, and finally gain an eventual consistency. Theoretically speaking, the performance will definitely degrade.

Do you mean that adding owner and id-name mapping cache will always have a worse performance compared to not adding these two caches?

yuqi1129:
> Do you mean that adding owner and id-name mapping cache will always have a worse performance compared to not adding these two caches?

No, I mean owner and id-name mapping cache did improve performance. What I want to emphasize is that: Since adding this eventually consistent cache is acceptable, why not apply this policy to user/role/group?  In order to ensure strong consistency for them,  we add version checks from the DB for each request, which will significantly slow down the API.

jerryshao:
> @jerryshao
> 
> I have updated and replied to the comments. Still, I have the following concerns:
> 
> * We use version check to fulfill strong consistency for user/role, but still use eventual consistency for id-name mapping, ownership relationship. It does not look so elegant, as it will always introduce 2 extra DB queries compared to the previous one, and finally gain an eventual consistency. Theoretically speaking,  the performance will definitely degrade. Is that truly acceptable? I asked myself several times, "Why not also make the user/role eventually consistent, too ?" In this way, tiny changes to the table definition and a good performance will be realized (all use db-version + pull internally).  It would be preferable if you could give me some advice on this point. That is why it hinders me to add id-mapping cache.

I think user/role eventual-consistency is acceptable to me. The reason why we chose this solution is affected by Polaris implementation. This is a simple solution to achieve strong consistency. If compared to eventual-consistency solution, the performance doesn't decrease a lot. I think this is simpler than other broadcasting solutions.

For entity-cache, we have two scenarios, one is external datasource, another is managed entity by Gravitino alone.

For the entities that both combine the entity from datasource and gravitino. We can use try the underlying source as SSOT to evict the cache.

For the entities that are managed by Gravitino only. We can leverage Polaris solution.

> * I have introduced a new table named `entity_change_log` to gain fine-grained recording of each ENTITY change. I'm not 100% sure whether it's the best option, as it will add a new table again. It's not very friendly for future refactoring.
>   Other options include adding `updated_at` for each table, employing `update-time` in the audit information, and using `current_version`/`last_version`, which are also not very desirable.

If the owner and name-id mapping cache is still unclear, we can move to the next step task. Currently, we don't have a clear answer whether 2 additional DB queries is expensive or not. We should implement the phase one and do more tests.

yuqi1129:
> > @jerryshao
> > I have updated and replied to the comments. Still, I have the following concerns:
> > 
> > * We use version check to fulfill strong consistency for user/role, but still use eventual consistency for id-name mapping, ownership relationship. It does not look so elegant, as it will always introduce 2 extra DB queries compared to the previous one, and finally gain an eventual consistency. Theoretically speaking,  the performance will definitely degrade. Is that truly acceptable? I asked myself several times, "Why not also make the user/role eventually consistent, too ?" In this way, tiny changes to the table definition and a good performance will be realized (all use db-version + pull internally).  It would be preferable if you could give me some advice on this point. That is why it hinders me to add id-mapping cache.
> 
> I think user/role eventual-consistency is acceptable to me. The reason why we chose this solution is affected by Polaris implementation. This is a simple solution to achieve strong consistency. If compared to eventual-consistency solution, the performance doesn't decrease a lot. I think this is simpler than other broadcasting solutions.
> 
> For entity-cache, we have two scenarios, one is external datasource, another is managed entity by Gravitino alone.
> 
> For the entities that both combine the entity from datasource and gravitino. We can use try the underlying source as SSOT to evict the cache.
> 
> For the entities that are managed by Gravitino only. We can leverage Polaris solution.
> 
> > * I have introduced a new table named `entity_change_log` to gain fine-grained recording of each ENTITY change. I'm not 100% sure whether it's the best option, as it will add a new table again. It's not very friendly for future refactoring.
> >   Other options include adding `updated_at` for each table, employing `update-time` in the audit information, and using `current_version`/`last_version`, which are also not very desirable.
> 
> If the owner and name-id mapping cache is still unclear, we can move to the next step task. Currently, we don't have a clear answer whether 2 additional DB queries is expensive or not. We should implement the phase one and do more tests.

@jerryshao Thanks for your insightful point. I will be coding according to the current document and do enough testing before merging. 
If an extra 2 DB query is not acceptable for performance testing, I will make the user/role/group also eventually consistent to reduce db queries.

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,1421 @@
...
+// Enables staleness detection rather than TTL expiry.
+private GravitinoCache<Long, Long> loadedRoles;
+// roleId → role_meta.updated_at at the time JCasbin policies were loaded
+
+// NEW: caches name → integer id for every MetadataObject referenced in OGNL expressions.
+// Without this, every authorize()/isOwner() call triggers entityStore.get() for each unique
+// object in the expression chain (METALAKE, CATALOG, SCHEMA, TABLE etc.), multiplied by the
+// number of distinct privilege checks on that object.
+// Consistency: immediate on local node via handleEntityStructuralChange() hook on drop/rename;
+//              HA peer nodes: entity change poller reads entity_change_log WHERE created_at > maxSeen
```
bharos, line 659: Nit: entity_change_log already has an auto-increment id PK. Consider using WHERE id > lastSeenId as the poller cursor instead of WHERE created_at > createdAtAfter. It's a PK scan (faster than a secondary index scan on created_at), and the idx_created_at index can then be kept only for pruning.

yuqi1129: Good idea. Search with a primary key is indeed faster than that with a secondary key, and I will update the document.

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,1421 @@
...
+CREATE INDEX idx_role_meta_del_upd
+    ON role_meta (role_id, deleted_at, updated_at);
+CREATE INDEX idx_owner_meta_obj_del_upd
+    ON owner_meta (metadata_object_id, deleted_at, updated_at);
+```
+
+Backfill strategy for the newly added `updated_at` columns:
+
+1. DDL adds columns with default `0`.
+2. One-time backfill sets `updated_at = create_time` (or `last_modified_time` if available)
```
jerryshao, line 430: When do you do the backfill?

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,1421 @@
...
+ALTER TABLE `owner_meta`
+    ADD COLUMN `updated_at` BIGINT NOT NULL DEFAULT 0
+    COMMENT 'Set to currentTimeMillis() on any ownership transfer.
+             The owner change poller reads updated_at > maxSeen to find changed rows
+             and invalidates only the specific metadataObjectIds in ownerRelCache.';
+```
+
+---
+
+### 4.4 Eventual Consistency: Name→ID Cache (`metadataIdCache`)
```
jerryshao, line 500: Shall we put this cache for auth only, or will we extend it to entity cache?

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,1421 @@
...
+#### The Problem: Repeated `getID()` Calls in OGNL Expression Evaluation
+
+`MetadataIdConverter.getID()` calls `entityStore.get()` for every unique `(MetadataObject,
+privilege)` combination in the OGNL expression. The `allowAuthorizerCache` deduplicates
+complete `(principal, metalake, obj, privilege)` results, but different privileges on the same
+object (e.g. `METALAKE::USE_CATALOG`, `METALAKE::USE_SCHEMA`, `METALAKE::DENY_USE_CATALOG`)
+each trigger a separate `getID(METALAKE)` call. A full `LOAD_TABLE_AUTHORIZATION_EXPRESSION`
+evaluation can trigger **8–12 `getID()` calls**, of which most are for the same 3–4 objects.
+
+#### Hierarchical Cache Key with Prefix-Based Cascade Invalidation
```
jerryshao, line 511: You should also consider the hierarchical schema solution in @roryqi 's PR, not sure if the changes will affect your cache design?

design-docs/cache-improvement-design.md
```diff
@@ -0,0 +1,1421 @@
...
+| FILESET     | `lake1::cat1::s1::fs1::FILESET` | leaf              |
+| TOPIC       | `lake1::cat1::s1::tp1::TOPIC`   | leaf              |
+| MODEL       | `lake1::cat1::s1::m1::MODEL`    | leaf              |
+| VIEW        | `lake1::cat1::s1::v1::VIEW`     | leaf              |
+
+`invalidateByPrefix("lake1::cat1::")` evicts the catalog entry AND all schemas, tables,
+filesets, and other entities beneath it in a single O(n) pass over the cache (bounded, DDL is
+rare).
+
+#### Why `entity_change_log` Instead of Adding `updated_at` to Entity Tables
```
jerryshao, line 530: Is this only track the "rename", or "create" and "deletion"?

yuqi1129: I tracked all modifications, "create", "delete", and "alter". In fact, for alternation, only `rename` is needed.

design-docs/cache-improvement-design.md
yuqi1129: Eventual consistency is acceptable for authorization, and may have problems for pure metadata operations, so I have no plan to expand it to entity cache in 1.3.0.

design-docs/cache-improvement-design.md
yuqi1129: It's indeed a different use case and I will discuss with him about this point.

design-docs/cache-improvement-design.md
yuqi1129: I plan to do it when the newly added column is created and then backfill it with `create_time` during the Gravitino version upgrade.  Till now, the `create_time` only exists in audit_info; we can extract it via SQL JSON function. If this can't be achieved, please leave it as is.

jerryshao: OK, that makes sense.

jerryshao:
I have no further comment, please work out a prototype to see if it meets your design expectation.

jerryshao:
Please update the PR title to follow our convention. I'm going to merge this, but we should update the doc if the implementation has changes. @yuqi1129

yuqi1129:
> Please update the PR title to follow our convention. I'm going to merge this, but we should update the doc if the implementation has changes. @yuqi1129

I see. Let me check and update it.

jerryshao:
You miss the license header.

jerryshao (approved):

PR #10699, opened 2026-04-07
[#10684] feat(iceberg-rest): Support vended credentials on registerTable endpoint

### LINKED ISSUE #10684: [Improvement] registerTable endpoint should support vended credentials via X-Iceberg-Access-Delegation header

The `registerTable` endpoint (`POST /v1/{prefix}/namespaces/{namespace}/register`) does not accept the `X-Iceberg-Access-Delegation` header and never vends credentials in its response. In contrast, `createTable` and `loadTable` both accept this header and pass `isCredentialVending` to the `IcebergRequestContext`, which triggers credential injection into the `LoadTableResponse`.

In the `registerTable` method ([IcebergNamespaceOperations.java#L299-L323](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergNamespaceOperations.java#L299-L323)):
- No `@HeaderParam(X_ICEBERG_ACCESS_DELEGATION)` parameter
- `IcebergRequestContext` is constructed without `isCredentialVending`
- The response is returned via `IcebergRESTUtils.ok()` instead of `buildResponseWithETag()`

The Iceberg REST spec defines the `data-access` (`X-Iceberg-Access-Delegation`) header as a parameter on the `registerTable` endpoint, and the response uses the same `LoadTableResult` schema that includes `config` and `storage-credentials` fields for credential vending.

This means clients that register a table and immediately attempt to read its data cannot use vended credentials from the registration response — they must make a separate `loadTable` call.

### How should we improve?

- Add `@HeaderParam(X_ICEBERG_ACCESS_DELEGATION) String accessDelegation` to the `registerTable` method
- Pass `isCredentialVending` to `IcebergRequestContext` so credential injection is triggered
- Return the response with ETag via `buildResponseWithETag()` for consistency with `createTable`/`loadTable`

sachinnn99:
Resubmitting per @laserninja's comment on #10684. This revision aligns the implementation with the established `createTable`/`loadTable` credential-vending pattern in `CatalogWrapperForREST`: the new 3-arg `registerTable` calls `super.registerTable(2-arg)` and uses the same inline `shouldGenerateCredential` / `injectCredentialConfig` conditional that `createTable` and `loadTable` already use, and `CatalogWrapperForTest` now overrides at the **3-arg** level (matching how it overrides `createTable`).

Add `X-Iceberg-Access-Delegation` header support to the `registerTable` endpoint, enabling credential vending in the response. Mirrors the existing pattern from `createTable` and `loadTable`.

Changes:
- `IcebergNamespaceOperations.registerTable`: add `@HeaderParam(X_ICEBERG_ACCESS_DELEGATION)`, compute `isCredentialVending`, build the 3-arg `IcebergRequestContext`
- `CatalogWrapperForREST`: add 3-arg `registerTable` that calls `super.registerTable(2-arg)` then conditionally injects credentials inline — structurally identical to the existing 3-arg `createTable` and `loadTable`
- `CatalogWrapperForREST`: widen `shouldGenerateCredential` and `injectCredentialConfig` from `private` to `protected` so subclasses (e.g. test wrappers that mock the underlying table operation) can participate in the credential vending flow
- `IcebergNamespaceOperationExecutor.registerTable`: pass `context.requestCredentialVending()` through to the 3-arg wrapper form
- `IcebergTableOperations.isCredentialVending`: widen `private` → package-private `static` so `IcebergNamespaceOperations` can reuse it without duplicating the validation logic
- `CatalogWrapperForTest`: override the **3-arg** `registerTable` (matching the 3-arg `createTable` override level), build the mock response, and inline the same `shouldGenerateCredential` / `injectCredentialConfig` conditional. Honors cloud URIs in `metadataLocation` so vending tests can verify the vended path. The legacy 2-arg override is removed (no longer reachable via the dispatcher chain).

The Iceberg REST spec defines `X-Iceberg-Access-Delegation` as a valid header on `registerTable`, but the current implementation does not accept it or vend credentials. Clients that register a table and immediately attempt to read its data must make a separate `loadTable` call to obtain credentials.

Fix: #10684

Yes. The `registerTable` REST endpoint now accepts the `X-Iceberg-Access-Delegation` header and returns vended credentials in the response config when requested. Backward compatible — clients that do not send the header get existing behavior.

Added unit tests in `TestIcebergNamespaceOperations`:
- `testRegisterTableWithCredentialVending` — verifies no vending without header, no vending for local URI, vending for `s3://` URI
- `testRegisterTableRemoteSigningNotSupported` — verifies 406 response for `remote-signing`
- `testRegisterTableInvalidAccessDelegation` — verifies 400 response for invalid header values

All existing `TestIcebergNamespaceOperations` tests still pass with no regressions.

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
@@ -80,6 +80,7 @@ import org.apache.iceberg.rest.PlanStatus;
 import org.apache.iceberg.rest.RESTCatalog;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
 import org.apache.iceberg.rest.requests.PlanTableScanRequest;
+import org.apache.iceberg.rest.requests.RegisterTableRequest;
 import org.apache.iceberg.rest.requests.UpdateTableRequest;
 import org.apache.iceberg.rest.responses.ImmutableLoadCredentialsResponse;
 import org.apache.iceberg.rest.responses.LoadCredentialsResponse;
@@ -152,6 +153,21 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
     }
     if (shouldGenerateCredential(loadTableResponse, requestCredential)) {
       return injectCredentialConfig(identifier, loadTableResponse, privilege);
+    }
+    return loadTableResponse;
+  }
+
+  public LoadTableResponse registerTable(
+      Namespace namespace, RegisterTableRequest request, boolean requestCredential) {
+    LoadTableResponse loadTableResponse = super.registerTable(namespace, request);
+    if (shouldGenerateCredential(loadTableResponse, requestCredential)) {
+      // Vend WRITE credentials: the registering user becomes the table owner
+      // (IcebergNamespaceHookDispatcher.setTableOwner runs after this call
+      // returns), consistent with createTable which also vends WRITE.
+      return injectCredentialConfig(
+          TableIdentifier.of(namespace, request.name()),
+          loadTableResponse,
+          CredentialPrivilege.WRITE);
     }
     return loadTableResponse;
   }
@@ -286,7 +302,8 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
     return false;
   }
 
-  private LoadTableResponse injectCredentialConfig(
+  @VisibleForTesting
+  protected LoadTableResponse injectCredentialConfig(
       TableIdentifier tableIdentifier,
       LoadTableResponse loadTableResponse,
       CredentialPrivilege privilege) {
@@ -336,7 +353,8 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
     return credential;
   }
 
-  private boolean shouldGenerateCredential(
+  @VisibleForTesting
+  protected boolean shouldGenerateCredential(
       LoadTableResponse loadTableResponse, boolean requestCredential) {
     if (!requestCredential) {
       return false;
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
@@ -603,7 +603,13 @@ public class IcebergTableOperations {
     return etag.getValue().equals(clientEtag);
   }
 
-  private boolean isCredentialVending(String accessDelegation) {
+  /**
+   * Parses the {@code X-Iceberg-Access-Delegation} header value and returns whether the client is
+   * requesting credential vending. Package-private and static so that {@link
+   * IcebergNamespaceOperations#registerTable} can reuse the same parsing logic from the same
+   * package.
+   */
+  static boolean isCredentialVending(String accessDelegation) {
     if (StringUtils.isBlank(accessDelegation)) {
       return false;
     }
--- a/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/CatalogWrapperForTest.java
+++ b/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/CatalogWrapperForTest.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import org.apache.gravitino.credential.CredentialPrivilege;
 import org.apache.gravitino.iceberg.common.IcebergConfig;
 import org.apache.gravitino.iceberg.service.CatalogWrapperForREST;
 import org.apache.iceberg.DataFile;
@@ -41,7 +42,9 @@ import org.apache.iceberg.rest.responses.LoadTableResponse;
 import org.apache.iceberg.types.Types.NestedField;
 import org.apache.iceberg.types.Types.StringType;
 
-// Used to override registerTable
+// Test wrapper that mocks operations the in-memory catalog cannot perform natively
+// (e.g. registerTable, which requires a real metadata.json file at the given location),
+// and adds test-only hooks (e.g. plan-task data generation for createTable).
 @SuppressWarnings("deprecation")
 public class CatalogWrapperForTest extends CatalogWrapperForREST {
   public static final String GENERATE_PLAN_TASKS_DATA_PROP = "test.generate-plan-data";
@@ -61,23 +64,44 @@ public class CatalogWrapperForTest extends CatalogWrapperForREST {
   }
 
   @Override
-  public LoadTableResponse registerTable(Namespace namespace, RegisterTableRequest request) {
+  public LoadTableResponse registerTable(
+      Namespace namespace, RegisterTableRequest request, boolean requestCredential) {
     if (request.name().contains("fail")) {
       throw new AlreadyExistsException("Already exits exception for test");
     }
 
+    // The in-memory test catalog cannot natively registerTable (it would need a real
+    // metadata.json file at the given location), so build a mock LoadTableResponse here.
+    // Honor cloud URIs (e.g. s3://) in metadataLocation so credential vending tests can
+    // verify the vended path; default to /mock otherwise for existing tests.
+    String location =
+        request.metadataLocation().contains("://") ? request.metadataLocation() : "/mock";
     Schema mockSchema = new Schema(NestedField.of(1, false, "foo_string", StringType.get()));
     TableMetadata baseMetadata =
         TableMetadata.newTableMetadata(
-            mockSchema, PartitionSpec.unpartitioned(), "/mock", ImmutableMap.of());
+            mockSchema, PartitionSpec.unpartitioned(), location, ImmutableMap.of());
     String json = TableMetadataParser.toJson(baseMetadata);
     TableMetadata tableMetadata =
-        TableMetadataParser.fromJson("/mock/metadata/v1.metadata.json", json);
+        TableMetadataParser.fromJson(location + "/metadata/v1.metadata.json", json);
     LoadTableResponse loadTableResponse =
         LoadTableResponse.builder()
             .withTableMetadata(tableMetadata)
             .addAllConfig(ImmutableMap.of())
             .build();
+    // We must replicate the credential-vending check + injection here (rather than reuse
+    // CatalogWrapperForREST.registerTable via super) because the in-memory test catalog
+    // cannot natively perform registerTable. Above, we synthesized a mock LoadTableResponse
+    // instead of delegating to super.registerTable; that means the parent class never sees
+    // this call and therefore never runs its vending logic. Calling shouldGenerateCredential
+    // / injectCredentialConfig directly (now protected on the parent) re-applies the exact
+    // same vending behavior to the mock response, so credential-vending tests exercise the
+    // production code path end-to-end. Privilege must match the production wrapper (WRITE).
+    if (shouldGenerateCredential(loadTableResponse, requestCredential)) {
+      return injectCredentialConfig(
+          TableIdentifier.of(namespace, request.name()),
+          loadTableResponse,
+          CredentialPrivilege.WRITE);
+    }
     return loadTableResponse;
   }
 
--- a/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
+++ b/iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
@@ -21,10 +21,15 @@ package org.apache.gravitino.iceberg.service.rest;
 import java.util.Arrays;
 import java.util.Optional;
 import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.core.EntityTag;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import org.apache.gravitino.credential.Credential;
 import org.apache.gravitino.iceberg.service.IcebergRESTUtils;
+import org.apache.gravitino.iceberg.service.extension.DummyCredentialProvider;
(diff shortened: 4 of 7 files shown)


roryqi:
@laserninja @bharos Could u help me review this pull request/

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
```diff
@@ -257,4 +265,63 @@ void testUpdateNamespace() {
...
+    // register without credential vending — no credentials in response
+    verifyRegisterTableSucc("register_cred_foo1", Namespace.of("register_cred_ns"));
+
+    // register with credential vending but local location — should NOT vend
+    Response response =
+        doRegisterTableWithCredentialVending(
+            "register_cred_foo2", Namespace.of("register_cred_ns2"), "mock");
+    Assertions.assertEquals(Status.OK.getStatusCode(), response.getStatus());
+    LoadTableResponse loadTableResponse = response.readEntity(LoadTableResponse.class);
+    Assertions.assertTrue(!loadTableResponse.config().containsKey(Credential.CREDENTIAL_TYPE));
```
bharos: assertFalse ?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -603,7 +603,7 @@ private static boolean etagMatches(String ifNoneMatch, EntityTag etag) {
     return etag.getValue().equals(clientEtag);
   }
 
-  private boolean isCredentialVending(String accessDelegation) {
+  static boolean isCredentialVending(String accessDelegation) {
```
bharos, line 612: We are making this package-private
Should we annotate with @visibleForTesting

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -130,6 +131,18 @@ public LoadTableResponse loadTable(
...
   }
 
+  public LoadTableResponse registerTable(
+      Namespace namespace, RegisterTableRequest request, boolean requestCredential) {
+    LoadTableResponse loadTableResponse = super.registerTable(namespace, request);
+    if (shouldGenerateCredential(loadTableResponse, requestCredential)) {
+      return injectCredentialConfig(
+          TableIdentifier.of(namespace, request.name()),
+          loadTableResponse,
+          CredentialPrivilege.WRITE);
```
laserninja, line 170: `createTable` hardcodes `WRITE` — that makes sense since the client just created the table and likely needs to write data to it. But `registerTable` is semantically different: the table data already exists and the client is just adding it to the catalog. The typical next action after registration is **reading** the data. Meanwhile, `loadTable` accepts the privilege as a parameter and delegates the decision to the caller.

Should `registerTable` similarly accept/infer the privilege, or is the `WRITE` default intentional here? At minimum, a brief comment explaining the choice would help future readers.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/CatalogWrapperForTest.java
laserninja, line 1: The PR's new 3-arg `registerTable` override in `CatalogWrapperForTest` duplicates the `shouldGenerateCredential`/`injectCredentialConfig` pattern from the parent `CatalogWrapperForREST`. This duplication is **necessary** because the in-memory catalog can't handle `registerTable` natively (unlike `createTable`, where the test calls `super.createTable(3-arg)` and the parent handles everything).

However, if the credential injection logic ever changes in `CatalogWrapperForREST`, this test would silently diverge. Consider adding a brief comment in the test explaining why the duplication exists, e.g.:

```java
// Cannot delegate to super.registerTable(3-arg) because the in-memory catalog
// does not support registerTable; must inline credential injection here.
```

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -191,7 +204,7 @@ protected boolean useDifferentClassLoader() {
     return false;
   }
```
laserninja, line 304: These methods go from `private` → `protected`, which makes them part of the subclass API contract. This is the cleanest option given the constraints but will need to be maintained as a stable interface going forward. A short JavaDoc note on their intended use (subclass credential injection in cases where `super` can't be called) would clarify intent.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -239,7 +252,7 @@ private Credential getCredential(
     return credential;
   }
```
laserninja, line 355: same as above

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
```diff
@@ -257,4 +265,63 @@ void testUpdateNamespace() {
...
+
+    // register with credential vending and S3 location — SHOULD vend
+    String s3Location = "s3://dummy-bucket/register_cred_foo3";
+    response =
+        doRegisterTableWithCredentialVending(
+            "register_cred_foo3", Namespace.of("register_cred_ns3"), s3Location);
+    Assertions.assertEquals(Status.OK.getStatusCode(), response.getStatus());
+    loadTableResponse = response.readEntity(LoadTableResponse.class);
+    Assertions.assertEquals(
```
laserninja, line 289: The S3 vending case only checks for `Credential.CREDENTIAL_TYPE` in the config.
Asserting on a couple more expected credential properties (e.g., whatever `DummyCredentialProvider` puts in the config) would make the test more robust against regressions where the type is present but the actual credential values are missing.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
sachinnn99: Fixed — replaced `assertTrue(!...)` with `assertFalse(...)`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
sachinnn99: Good catch on the visibility change being undocumented, but the change is actually for production reuse rather than testing — `IcebergNamespaceOperations#registerTable` (in the same package) now calls this to parse the same `X-Iceberg-Access-Delegation` header that `IcebergTableOperations#createTable` and `#loadTable` already parse. `@VisibleForTesting` would be misleading because no test depends on its visibility. I've added a JavaDoc explaining the production-reuse rationale instead.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
sachinnn99: You're right, and after digging in I'm switching to READ.

Specifically:

1. **registerTable writes no files at the storage layer.** Verified end-to-end: `CatalogHandlers.registerTable` → `BaseMetastoreCatalog.registerTable` → `ops.commit(null, metadata)` → `BaseMetastoreTableOperations.writeNewMetadataIfRequired(true, metadata)` early-returns the existing `metadata.metadataFileLocation()` without writing, then `JdbcTableOperations` just inserts the catalog row. So the storage-layer asymmetry you flagged with `createTable` is real — `createTable` proves write capability by writing a metadata.json file, `registerTable` doesn't.

2. **registerTable also doesn't set ownership.** `IcebergNamespaceOperationExecutor.registerTable` is a straight delegation to the wrapper — unlike `IcebergTableOperationExecutor.createTable` and `IcebergNamespaceOperationExecutor.createNamespace`, which both explicitly set `IcebergConstants.OWNER` to `context.userName()`. So a user with only `ANY_CREATE_TABLE` privilege never becomes the table owner via register.

3. **WRITE was therefore a one-off privilege escalation.** That same user cannot pass `FILTER_MODIFY_TABLE_AUTHORIZATION_EXPRESSION` (`ANY(OWNER, METALAKE, CATALOG, SCHEMA, TABLE) || ANY_MODIFY_TABLE`) on the registered table via any other API path. A follow-up `loadTable` with vending would return READ for them via `getCredentialPrivilege`. Hardcoding WRITE here was vending storage credentials that let them perform actions gravitino's authz model explicitly forbids.

Switching to READ. Higher-privileged users (catalog/metalake owners, `MODIFY_TABLE` holders) who legitimately need to write to a registered table can call `loadTable` with vending afterwards and get WRITE through the dynamic helper, or be granted `MODIFY_TABLE` explicitly.

I've added a code comment documenting all of this.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/CatalogWrapperForTest.java
sachinnn99: Added an inline comment at the call site explaining the duplication: the in-memory test catalog cannot perform `registerTable` natively, so the override synthesizes a mock `LoadTableResponse` instead of delegating to `super.registerTable`. Because we never go through the parent's `registerTable`, we must explicitly re-run its vending logic so credential-vending tests still exercise the same code path as production. (Also updated this wrapper to use READ to match the production change.)

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
sachinnn99: Added JavaDoc explaining that the `protected` visibility is for subclasses that synthesize a `LoadTableResponse` without delegating to `super` (test wrappers around in-memory catalogs that can't perform certain operations natively), so they can still apply the same vending and injection logic as production paths.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
sachinnn99: Same JavaDoc rationale added on `shouldGenerateCredential`.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergNamespaceOperations.java
sachinnn99: Added an additional assertion on `Credential.EXPIRE_TIME_IN_MS`. `DummyCredentialProvider.SimpleCredential` is not one of the typed credentials in `CredentialPropertyUtils.toIcebergProperties`, so it falls through to `Credential#toProperties()`, which emits exactly `credential-type` and `expire-time-in-ms`. Asserting both confirms the full credential block is injected and guards against partial-injection regressions.

laserninja:
Can you resolve the merge conflicts please

sachinnn99:
@laserninja rebased onto main, conflicts resolved.

laserninja:
In `createTable` and `loadTable`, `CatalogWrapperForREST` checks `if (catalog instanceof RESTCatalog)` and delegates to `*Internal()` methods before credential injection. The new `registerTable` skips this branch and calls `super.registerTable()` directly. Could you confirm this works correctly when the underlying catalog is a `RESTCatalog` (i.e., Gravitino fronting another Iceberg REST catalog)? If the upstream catalog already handles credential vending, the `shouldGenerateCredential` gate (`getCatalog() instanceof RESTCatalog → return false`) would short-circuit correctly, but it would be good to verify that `RESTCatalog.registerTable` properly forwards the `X-Iceberg-Access-Delegation` header upstream.

sachinnn99:
@laserninja Good question — the current code does include the `if (catalog instanceof RESTCatalog)` branch for `registerTable`, delegating to `registerTableInternal()` before credential injection, matching the same pattern as `createTable` and `loadTable`.

When the underlying catalog is a `RESTCatalog`, `shouldGenerateCredential` returns `false` (via the `getCatalog() instanceof RESTCatalog` check), so Gravitino correctly defers credential vending to the upstream catalog.

The upstream Iceberg `RESTSessionCatalog.registerTable` doesn't forward `X-Iceberg-Access-Delegation` as an explicit header — it uses `mutationHeaders` (idempotency headers only). But it doesn't need to: it processes the upstream response's credentials via `response.credentials()` and wires them into the `RESTTableOperations` / `tableFileIO`. So the chain works correctly end-to-end.

laserninja:
LGTM. READ privilege is well-reasoned, RESTCatalog path added, JavaDoc and test assertions look good. Thanks for the thorough responses.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
roryqi: We set the table owner when registering the table. So the register table requester should have write privilege. It will be consistent with our authz model. WDYT?

sachinnn99: @roryqi Good point, you are right. `IcebergNamespaceHookDispatcher.registerTable` calls `setTableOwner` for the registering user, so the user does become the table owner. This is consistent with `createTable`, which also vends WRITE for the same reason.

Updated to vend WRITE and corrected the comment.

sachinnn99:
@roryqi Updated to vend WRITE per your feedback — the registering user becomes the table owner via `setTableOwner`, so WRITE is consistent with `createTable`. Let me know if the latest revision looks good.

roryqi (approved):
LGTM. Wait for CI.

roryqi:
Thanks @sachinnn99 for your contribution. Thanks @laserninja @laserninja for your review.

Issue #10705, opened 2026-04-08
[Improvement] Add switch to package web UI


yuqi1129:
### What would you like to be improved?

In some cases, there is no need to build and package a web UI, as it's not used. For example, in backend tests, compiling the web UI is purely a waste of time. 

We'd better add a switch to compile it or not.

PR #10706, opened 2026-04-08
[#10705] improvement(build): add switch to enable web-ui in configuration and gradlew build parameter

yuqi1129:
- Add `-PskipWebWar=true` support to skip web/web-v2 war build and packaging.
- Add `gravitino.server.webui.enable` (default `true`) and make `GravitinoServer` respect it, so deploy mode can start without war when web is skipped.
- Update integration-test build scripts to make web build dependency optional under `skipWebWar`.
- Update non-frontend GitHub workflows to pass `-PskipWebWar=true`.
- Add server test to verify startup with web UI disabled.

In many non-frontend scenarios (especially integration tests), building web war is unnecessary and slows CI. This change avoids that cost while keeping server startup valid.

Fix: #10705

- New config: `gravitino.server.webui.enable` (default `true`).
- New build property: `-PskipWebWar=true`.

- `./gradlew --no-daemon :server:test -PskipITs`
- `./gradlew --no-daemon compileDistribution -PskipWebWar=true -x test`
- Local sanity check for CI arg compatibility: `./gradlew help -PskipWebWar=true -q`

### DIFF
--- a/conf/gravitino.conf.template
+++ b/conf/gravitino.conf.template
@@ -39,6 +39,7 @@ gravitino.server.webserver.threadPoolWorkQueueSize = 100
 gravitino.server.webserver.requestHeaderSize = 131072
 # The response header size of the built-in web server
 gravitino.server.webserver.responseHeaderSize = 131072
+
 # THE CONFIGURATION FOR Gravitino ENTITY STORE
 # The entity store to use, we only supports relational
 gravitino.entity.store = relational


web-v2/integration-test/build.gradle.kts
```diff
@@ -67,6 +68,8 @@ tasks.test {
     dependsOn(":catalogs:catalog-kafka:jar", ":catalogs:catalog-kafka:runtimeJars")
 
     // Frontend tests depend on the web page, so we need to build the web module first.
-    dependsOn(":web-v2:web:build")
+    if (!skipWebWar) {
```
diqiu50: Why not skip the entire web module? The comment above says the frontend tests depend on the web build, but you only removed the build step.

web/integration-test/build.gradle.kts
```diff
@@ -67,6 +68,8 @@ tasks.test {
     dependsOn(":catalogs:catalog-kafka:jar", ":catalogs:catalog-kafka:runtimeJars")
 
     // Frontend tests depend on the web page, so we need to build the web module first.
-    dependsOn(":web:web:build")
```
diqiu50, line 70: Same issues

build.gradle.kts
```diff
@@ -775,28 +777,34 @@ tasks {
...
 
     group = "gravitino distribution"
     outputs.dir(projectDir.dir("distribution/package"))
     doLast {
       copy {
         from(projectDir.dir("conf")) { into("package/conf") }
         from(projectDir.dir("bin")) { into("package/bin") }
-        from(projectDir.dir("web/web/build/libs/${rootProject.name}-web-$version.war")) { into("package/web") }
-        from(projectDir.dir("web-v2/web/build/libs/${rootProject.name}-web-$version.war")) { into("package/web-v2") }
+        if (!skipWebWar) {
```
diqiu50: Can we check whether this condition is really necessary? If the web module has no build artifacts, copying it shouldn’t have any effect anyway, right?

web-v2/integration-test/build.gradle.kts
diqiu50: Why not introduce a skipWeb option to skip everything related to the web module?

yuqi1129:
Addressed the review suggestions for #10705 in commit 557dae8b1:

- Introduced `-PskipWeb=true` at root build level, and treat it as a stronger switch than `skipWebWar` (`skipFrontend = skipWeb || skipWebWar`).
- When `skipWeb=true`, disable tasks in `:web:web`, `:web-v2:web`, `:web:integration-test`, and `:web-v2:integration-test` so frontend-related module tasks do not run.
- Kept frontend IT modules compatible by making their web build dependency conditional on `!skipFrontend`.
- Updated non-frontend CI workflows to use `-PskipWeb=true` so they skip frontend work by default.

Validation done locally:
- `./gradlew --no-daemon :server:test -PskipITs`
- `./gradlew --no-daemon compileDistribution -PskipWeb=true -x test`

Also, the earlier `WebUIFilter` conditional registration fix has already been included in this PR.

web-v2/integration-test/build.gradle.kts
yuqi1129: fix

build.gradle.kts
yuqi1129: fix

diqiu50:
Please update the pr description

web-v2/integration-test/build.gradle.kts
```diff
@@ -53,6 +53,9 @@ dependencies {
 }
 
 tasks.test {
+  val skipWeb = (rootProject.findProperty("skipWeb") as? String)?.toBoolean() ?: false
+  val skipWebWar = (rootProject.findProperty("skipWebWar") as? String)?.toBoolean() ?: false
```
diqiu50: Does the `skipWebWar` need to be removed？
if you remove it, the `skipFrontend` will not be useful

jerryshao:
My initial feeling is that this makes the thing too complex. I need to take a deep look.

build.gradle.kts
```diff
@@ -66,6 +66,10 @@ if (scalaVersion !in listOf("2.12", "2.13")) {
   throw GradleException("Scala version $scalaVersion is not supported.")
 }
 
+val skipWeb: Boolean = (project.findProperty("skipWeb") as? String)?.toBoolean() ?: false
+val skipWebWar: Boolean = (project.findProperty("skipWebWar") as? String)?.toBoolean() ?: false
+val skipFrontend: Boolean = skipWeb || skipWebWar
```
jerryshao: Add three different properties makes things so complicated, do we have a simple solution?

diqiu50 (approved):
LGTM

PR #10720, opened 2026-04-08
[Design Docs] Iceberg REST supports nested namespace

roryqi:

Design document of Iceberg REST supports nested namespace 

No need.

Just design document

### DIFF
--- a/design-docs/iceberg-supported-nested-namespace.md
+++ b/design-docs/iceberg-supported-nested-namespace.md
@@ -180,10 +180,10 @@ sequenceDiagram
 ### Iceberg REST Side Behavior
 
 - **Create nested namespace**:
-  - Creating `A:B:C` creates (or ensures existence of) `A`, `A.B`, and `A.B.C` sequentially.
-  - Parent-chain creation is not atomic in current implementation; partial parents may remain if
-    request fails in middle steps.
-  - This is accepted as current behavior and can be improved in a future phase.
+  - Creating `A:B:C` creates (or ensures existence of) `A`, `A.B`, and `A.B.C` in one atomic
+    operation.
+  - Parent-chain creation is transactional: if any step fails, all created parents in this request
+    are rolled back and no partial parent namespaces remain.
   - Owner assignment rule for auto-created parents:
     - If a parent is newly auto-created, owner is inherited from nearest existing ancestor; if no
       ancestor exists, owner defaults to catalog owner/service principal, not leaf requester.
@@ -448,7 +448,7 @@ for (String scope : HierarchicalSchemaUtil.parentScopes("A:B:C")) {
 ### 3) Namespace operation behavior
 
 - `createNamespace` should ensure parent schemas exist for each hierarchical level.
-- Parent-chain creation is sequential and non-atomic in current phase.
+- Parent-chain creation must be atomic as one transaction; failures must not leave partial parents.
 - `updateNamespace` should support property updates for mapped namespace scope.
 - `dropNamespace` should target mapped physical schema and preserve existing non-nested behavior.
 - `listSchemas` should accept optional query parameter `parentSchema`.
--- a/design-docs/iceberg-supported-nested-namespace.md
+++ b/design-docs/iceberg-supported-nested-namespace.md
@@ -127,10 +127,10 @@ Cons:
   - Serialize: encode each segment first, then join with `:`.
   - Parse: split by `:`, then decode each segment.
 - Decode is applied exactly once to avoid double-decoding issues.
-- Add quote around schema names when needed to avoid parse ambiguity.
 - Keep flat storage model and convert `HierarchicalSchema` path to physical schema name by mapping rules.
-- `NameIdentifier` should support quoted `schema` names.
-- In this phase, quoted identifier support is limited to `schema` only; other identifier parts keep existing parsing rules.
+- Identifier rendering rule:
+  - Use encoded `HierarchicalSchema` path directly in schema position.
+  - Do not rely on single-quote wrapping for schema disambiguation in this phase.
 
 Examples:
 
@@ -139,17 +139,14 @@ Examples:
 - Logical `HierarchicalSchema` path is then converted to physical schema name through mapping rules.
 - Namespace levels `["team:core", "sales"]` are serialized as `team%3Acore:sales`.
 - Parsing `team%3Acore:sales` returns `["team:core", "sales"]`.
-- For parsing ambiguity, use quoted schema name in identifier rendering, for example:
-  - `metalake.catalog.'A.B'.table1`
-  - `metalake.catalog.'A.B.C'.table2`
-- In UI display, show logical path (for example `A:B:C`) for readability, but keep quoted form for parser-safe serialization.
-
-According to URL encoding rules (RFC 3986), single quotes have no reserved purpose 
-and must be percent-encoded if they appear in a URL component. The encoded form for a single quote is %27.
-
-Example:
-Invalid: https://example.com/path/to'O'reilly
-Valid (encoded): https://example.com/path/to%27O%27reilly
+- Identifier rendering example:
+  - `metalake.catalog.A:B.table1`
+  - `metalake.catalog.team%3Acore:sales.table2`
+- In UI display, show logical path (for example `A:B:C`) for readability.
+- HTTP transport rule:
+  - Namespace values in URL/query/body must follow RFC 3986 percent-encoding when needed.
+  - Example: namespace path `team%3Acore:sales` should be URL-encoded as
+    `team%253Acore%3Asales` when put into a URL query component.
 
 
 ### Iceberg REST Side Behavior
@@ -193,15 +190,29 @@ The following snippets are design-level examples to clarify how `HierarchicalSch
 ### Snippet 1: Convert Iceberg namespace to logical path and physical schema
 
 ```java
-// Example utility methods in IcebergRESTUtils (or a dedicated NamespacePathMapper)
-public static String toHierarchicalSchemaPath(Namespace namespace, String separator) {
-  // namespace.levels() = ["A", "B", "C"] -> "A:B:C"
-  return String.join(separator, namespace.levels());
+// Example utility methods in IcebergRESTUtils (or a dedicated HierarchicalSchemaUtil)
+private static String encodeSegment(String raw) {
+  // Encode % first, then separator-related characters.
+  return raw.replace("%", "%25").replace(":", "%3A");
+}
+
+private static String decodeSegment(String encoded) {
+  // Decode exactly once; reject malformed escape sequences in real implementation.
+  return encoded.replace("%3A", ":").replace("%25", "%");
+}
+
+public static String serializeHierarchicalPath(String[] levels) {
+  // ["team:core", "sales"] -> "team%3Acore:sales"
+  return Arrays.stream(levels).map(HierarchicalSchemaUtil::encodeSegment).collect(Collectors.joining(":"));
+}
+
+public static String[] parseHierarchicalPath(String path) {
+  // "team%3Acore:sales" -> ["team:core", "sales"]
+  return Arrays.stream(path.split(":", -1)).map(HierarchicalSchemaUtil::decodeSegment).toArray(String[]::new);
 }
 
 public static String toPhysicalSchemaName(String hierarchicalPath) {
-  // Phase-1 compatible mapping: keep storage as flattened schema
-  // "A:B:C" -> "A.B.C"
+  // Phase-1 mapping keeps flat schema storage: "A:B:C" -> "A.B.C"
   return hierarchicalPath.replace(":", ".");
 }
 ```
@@ -211,7 +222,7 @@ public static String toPhysicalSchemaName(String hierarchicalPath) {
 ```java
 // Example in IcebergMetadataAuthorizationMethodInterceptor
 Namespace rawNamespace = RESTUtil.decodeNamespace(value);
-String hierarchicalPath = HierarchicalSchemaUtil.toPath(rawNamespace, ":");
+String hierarchicalPath = HierarchicalSchemaUtil.serializeHierarchicalPath(rawNamespace.levels());
 String schema = HierarchicalSchemaUtil.toPhysicalSchemaName(hierarchicalPath);
 
 nameIdentifierMap.put(
@@ -223,7 +234,7 @@ nameIdentifierMap.put(
 
 ```java
 // Example path inheritance for A:B:C
-List<String> authzScopes = HierarchicalSchemaUtil.parentScopes("A:B:C", ":");
+List<String> authzScopes = HierarchicalSchemaUtil.parentScopes("A:B:C");
 // Result: ["A", "A:B", "A:B:C"]
 // Authorization passes if user has required privilege on any allowed parent scope by policy.
 ```
@@ -232,7 +243,7 @@ List<String> authzScopes = HierarchicalSchemaUtil.parentScopes("A:B:C", ":");
 
 ```java
 // For create namespace A:B:C, ensure parent schemas exist in physical model
-for (String scope : HierarchicalSchemaUtil.parentScopes("A:B:C", ":")) {
+for (String scope : HierarchicalSchemaUtil.parentScopes("A:B:C")) {
   String schemaName = HierarchicalSchemaUtil.toPhysicalSchemaName(scope); // A, A.B, A.B.C
   // create schema if not exists
 }
--- a/design-docs/iceberg-supported-nested-namespace.md
+++ b/design-docs/iceberg-supported-nested-namespace.md
@@ -97,7 +97,13 @@ Cons:
 - Introduce logical identifier concept: `HierarchicalSchema`.
 - `HierarchicalSchema` uses fixed separator `:` in logic layer.
 - `:` is mandatory in this design, not configurable.
-- Schema name can still contain `.` in physical storage to keep model stable.
+- For Gravitino REST create/update schema APIs, `request.getName()` keeps the logical schema name
+  and may contain `:` (for example `A:B` or `A:B:C`).
+- This phase does not pre-convert `request.getName()` from `:` to `.` before capability name
+  validation; capability naming rules must explicitly allow `:` for schema scope.
+- Schema name uses `:`-separated hierarchical representation in this design; `.` is not used as
+  schema separator.
+- `.` remains an Iceberg namespace hierarchy notation in external semantics only.
 - When `:` is used as separator, segment values containing `:` must be percent-encoded.
 - `%` must also be encoded as `%25` to avoid decode ambiguity.
 - Encoding/decoding order:
@@ -160,6 +166,23 @@ Examples:
 - Gravitino server REST supports namespace create/update/drop operations for nested namespace
   workflows, aligned with Iceberg REST behavior.
 - Existing schema/table APIs remain compatible with non-nested cases.
+
+### Performance Considerations
+
+- Current `EntityStore` does not support prefix matching on schema names.
+- As a result, `list schema` with `parentHierarchicalSchema` cannot be implemented as an
+  efficient prefix query at storage layer. The server must list all schemas in the catalog and
+  then compute the top-level / direct-children view in memory.
+- This may introduce higher-than-expected latency and load for catalogs with a large number of
+  schemas.
+
+Mitigations:
+
+- Cache computed hierarchy results per catalog with a short TTL and invalidate on schema
+  create/update/drop.
+- Enforce reasonable limits (pagination / maximum returned items) for list operations.
+- Add performance/regression tests with large schema counts to validate behavior.
+- Consider enhancing `EntityStore` to support prefix/index queries as a follow-up optimization.
 
 ## Privileges and Authorization
 
@@ -314,7 +337,8 @@ for (String scope : HierarchicalSchemaUtil.parentScopes("A:B:C")) {
 - Add a dedicated conversion utility for `HierarchicalSchema` path:
   - Iceberg namespace levels -> logical path (preferred `:`).
   - Logical path -> physical schema name (phase-1 uses `.` flattened mapping).
-- Add validation rules to reject ambiguous names (for example, raw schema names containing `:` when `:` is separator).
+- Override `Capability.specificationOnName(SCHEMA, name)` naming rules for related catalogs so
+  schema names containing `:` are accepted in this phase.
 
 ### 2) Authorization behavior
 
--- a/design-docs/iceberg-supported-nested-namespace.md
+++ b/design-docs/iceberg-supported-nested-namespace.md
@@ -136,12 +136,11 @@ Examples:
 - Mapping must be reversible:
   - `logical path segments` -> `encode each segment` -> `join by '.'` for physical storage.
   - physical schema name -> `split by '.'` -> `decode each segment` -> logical path segments.
-  - This avoids ambiguity when one segment contains `.` (for example `my.schema`).
+  - physical schema name -> "A.B.C"
 
 ### Existing Name Compatibility and Migration Guard
 
-- Before enabling nested namespace mode for a catalog, run a pre-check scan on existing schema names
-  against configured logical separator.
+- Before upgrading the new version, select all the Iceberg namespaces to check whether contains the logical delimiter.
 - If existing schema names conflict with selected separator, enabling is rejected with actionable
   error and user can choose another separator or rename conflicting schema names.
 - Once nested mode is enabled for a catalog, creating new schema names containing configured logical
@@ -149,7 +148,7 @@ Examples:
 
 ### Delimiter Configuration Policy
 
-- Logical separator is configurable (for example `:`, `;`, `$`) and can be chosen to avoid conflict
+- Logical separator is configurable in the server configuration (for example `:`, `;`, `$`) and can be chosen to avoid conflict
   with existing names.
 - Physical separator in storage remains fixed as `.`.
 - Recommended: keep logical separator stable after nested namespace is enabled for a catalog.
@@ -178,7 +177,7 @@ Two delimiter-governance options are under evaluation:
 Delimiter validity should be explicit and observable, not implicit.
 
 - Validation checkpoints:
-  - Validate delimiter when server starts or when delimiter configuration is updated.
+  - Validate delimiter when server starts.
   - Re-validate against existing catalog schema names before enabling nested mode for a catalog.
   - Reject invalid delimiter configuration early before request-time namespace operations.
 - Validation rules:
@@ -317,7 +316,7 @@ Examples (Gravitino REST side):
   - No new typed path wrapper (`SchemaPath`) or segment-based overloads are introduced in this
     phase.
 
-Simple Java client examples:
+Simple Java client examples: (Users can know the delimiter)
 
 - `listSchemas()` -> returns top-level schemas, for example `A`, `B`.
 - `listSchemas("A:B")` -> returns direct children, for example `A:B:C`, `A:B:D`.
@@ -563,6 +562,16 @@ Trino-specific example:
 - Connector converts to Gravitino internal schema: `a:b:c`
 - Gravitino returns schema: `a:b:c`
 - Connector renders back to Trino quoted style: `"a.b.c"`
+
+## Plans
+-  Add foundational classes for nested namespace authorization
+-  Add nested namespace support to Iceberg REST API
+-  Add nested namespace support to Gravitino REST AP
+
+## No Plan
+- Trino connector supports nested namespace
+- Spark connector supports nested namespace
+- Flink connector supports nested namespace
 
 ## Test Plan
 
--- a/design-docs/iceberg-supported-nested-namespace.md
+++ b/design-docs/iceberg-supported-nested-namespace.md
@@ -104,12 +104,10 @@ Cons:
 - Schema name uses `:`-separated hierarchical representation in this design; `.` is not used as
   schema separator.
 - `.` remains an Iceberg namespace hierarchy notation in external semantics only.
-- When `:` is used as separator, segment values containing `:` must be percent-encoded.
-- `%` must also be encoded as `%25` to avoid decode ambiguity.
-- Encoding/decoding order:
-  - Serialize: encode each segment first, then join with `:`.
-  - Parse: split by `:`, then decode each segment.
-- Decode is applied exactly once to avoid double-decoding issues.
+- Escaping strategy: no encoding/escaping is used in this phase.
+- The character `:` is reserved as hierarchy separator and is not allowed inside a single
+  namespace segment.
+- Parsing is direct split/join by `:` at API boundary.
 - Keep flat storage model and convert `HierarchicalSchema` path to physical schema name by mapping rules.
 - Identifier rendering rule:
   - Use encoded `HierarchicalSchema` path directly in schema position.
@@ -120,16 +118,39 @@ Examples:
 - Nested namespace `A:B` maps to logical `HierarchicalSchema` path `A:B`.
 - Nested namespace `A:B:C` maps to logical `HierarchicalSchema` path `A:B:C`.
 - Logical `HierarchicalSchema` path is then converted to physical schema name through mapping rules.
-- Namespace levels `["team:core", "sales"]` are serialized as `team%3Acore:sales`.
-- Parsing `team%3Acore:sales` returns `["team:core", "sales"]`.
+- Namespace levels `["team", "sales"]` are serialized as `team:sales`.
+- Parsing `team:sales` returns `["team", "sales"]`.
 - Identifier rendering example:
   - `metalake.catalog.A:B.table1`
-  - `metalake.catalog.team%3Acore:sales.table2`
-- In UI display, show logical path (for example `A:B:C`) for readability.
-- HTTP transport rule:
-  - Namespace values in URL/query/body must follow RFC 3986 percent-encoding when needed.
-  - Example: namespace path `team%3Acore:sales` should be URL-encoded as
-    `team%253Acore%3Asales` when put into a URL query component.
+  - `metalake.catalog.team:sales.table2`
+- In UI display and API transport, use logical path directly (for example `A:B:C`).
+
+### Separator Configuration
+
+- Internal namespace separator is configured as `:` for Gravitino hierarchical schema handling.
+- Connector-facing behavior remains Iceberg-compatible and does not require users to configure or
+  input this internal separator.
+
+### Parsing Sequence Diagram
+
+```mermaid
+sequenceDiagram
+  participant Client as Client/Connector
+  participant Config as Connector Config
+  participant API as Gravitino REST API
+  participant Mapper as HierarchicalSchema Mapper
+  participant Store as EntityStore
+
+  Client->>Config: Load namespace separator
+  Config-->>Client: separator (for example ':')
+  Client->>API: GET /schemas?parentSchema=A:B
+  API->>Mapper: Parse parentSchema by separator ':'
+  Mapper->>Store: list all schemas (no prefix index)
+  Store-->>Mapper: all schemas
+  Mapper->>Mapper: filter direct children under A:B
+  Mapper-->>API: hierarchy-aware result
+  API-->>Client: response (Iceberg-compatible namespace view)
+```
 
 
 ### Iceberg REST Side Behavior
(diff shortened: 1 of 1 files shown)


design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,116 @@
...
+## Non-Goal
+
+- Create nested namespace by Gravitino REST API directly.
+- Delete nested namespace by Gravitino REST API directly.
+- Modify nested namespace by Gravitino REST API directly (for example, rename or alter properties).
+- Introduce a new metadata object (for example `SubNamespace`) in this phase.
+
+## Solution Options
+
+### Option A: Add a new metadata object `SubNamespace`
```
yuqi1129: This description is rather simple, can you add more details or give an exmple here? I can't get your intention here.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,116 @@
...
+- `NameIdentifier` should support quoted `schema` names.
+- In this phase, quoted identifier support is limited to `schema` only; other identifier parts keep existing parsing rules.
+
+Examples:
+
+- Nested namespace `A.B` maps to schema name `A.B`.
+- Nested namespace `A.B.C` maps to schema name `A.B.C`.
+- For parsing ambiguity, use quoted schema name in identifier rendering, for example:
+  - `metalake.catalog.'A.B'.table1`
+  - `metalake.catalog.'A.B.C'.table2`
```
yuqi1129: Does URL support single quotes as above?

roryqi: Invalid: https://example.com/path/to'O'reilly
Valid (encoded): https://example.com/path/to%27O%27reilly
We can encode it.

design-docs/iceberg-supported-nested-namespace.md
roryqi: I added an example here. Thanks for your suggestion.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,129 @@
...
+- Introduce a new metadata object (for example `SubNamespace`) in this phase.
+
+## Solution Options
+
+### Option A: Add a new metadata object `SubNamespace`
+
+Use a new metadata object `SubNamespace` to replace schema
+
+Catalog -> SubNamespace a -> SubNamespace a.b -> Table a.b.c
+                          -> SubNamespace a.c -> SubNamespace a.c.d -> Table a.c.d.e
```
jerryshao: What's the relationship between namespace and sub-namespace?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Design
+
+### Identifier Rules
+
+- Schema name can contain `.`.
+- Add quote around schema names when needed to avoid parse ambiguity.
+- Keep flat storage model and use dot-joined schema name as namespace mapping.
+- `NameIdentifier` should support quoted `schema` names.
+- In this phase, quoted identifier support is limited to `schema` only; other identifier parts keep existing parsing rules.
```
jerryshao: Do we support a schema and a sub-schema list? I think IRC has such requirement.

design-docs/iceberg-supported-nested-namespace.md
jerryshao: Using quote maybe conflicted with some sql dialects.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,129 @@
...
+
+- **Create nested namespace**:
+  - Creating `A.B.C` will create (or ensure existence of) three schemas in Gravitino: `A`, `A.B`, and `A.B.C`.
+  - Set the created namespace owner as current user.
+- **Drop nested namespace**: drop corresponding schema in Gravitino.
+- **Rename nested namespace**: not needed because Iceberg REST does not support namespace rename.
+
+### Gravitino Side Behavior
+
+- `list schema` stays flat instead of nested.
```
jerryshao: Please explain more as I mentioned above. Do you support schema and sub-schema list?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,129 @@
...
+- Existing non-nested namespace behavior remains unchanged.
+- Limiting quoted identifier parsing to `schema` reduces regression risk for catalog/table/view/function identifier parsing.
+
+## Test Plan
+
+- Unit tests for schema name parse/quote handling when name contains `.`.
+- Unit/integration tests for Iceberg REST create/drop nested namespace mapping.
+- Authorization tests for nested scope behavior (`A`, `A.B`, `A.B.C`).
+- Regression tests for non-nested namespace authorization behavior.
+
```
jerryshao, line 398: You need to add the code snippets, affected classes, and expected changes to the design doc.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,129 @@
...
+- Support nested namespace operations from Iceberg REST to Gravitino through schema mapping.
+- Support privilege granting for different nested namespace scopes (including UI workflow).
+- Keep metadata model stable and avoid heavy refactor.
+
+## Non-Goal
+
+- Create nested namespace by Gravitino REST API directly.
+- Delete nested namespace by Gravitino REST API directly.
+- Modify nested namespace by Gravitino REST API directly (for example, rename or alter properties).
```
mchades: list/get nested namespace is goal or non-goal? why?

mchades: Why are these operations non-goal?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,129 @@
...
+References:
+
+- https://github.com/apache/gravitino/blob/main/docs/security/access-control.md
+- https://github.com/apache/gravitino/blob/main/docs/iceberg-rest-service.md
+- https://github.com/apache/gravitino/blob/main/docs/manage-relational-metadata-using-gravitino.md
+- https://github.com/apache/gravitino/discussions/7296
+
+## Goal
+
+- Support nested namespace operations from Iceberg REST to Gravitino through schema mapping.
```
mchades, line 36: I noticed that Iceberg supports [namespace-separator.](https://github.com/apache/iceberg/blob/87c743463b6311f2412e1addf19cf204c1b79e3d/open-api/rest-catalog-open-api.yaml#L267). Should we consider this? Will it affect our design?

roryqi: It won't affect our design.

design-docs/iceberg-supported-nested-namespace.md
roryqi: Gravitino REST API only has list schema API, we won't list sub-schema.
We have four namespaces. A.B.C, A, B and A.B.
When listing schema, we will return all the schemas, A.B.C, A, B and A.B.
We won't only return the first level namespaces A and B.

design-docs/iceberg-supported-nested-namespace.md
roryqi: One Iceberg Namespace will map to one Gravitno SubNamespace.

design-docs/iceberg-supported-nested-namespace.md
roryqi: I will survey about Trino. Actually, I need to align to Trino.

design-docs/iceberg-supported-nested-namespace.md
roryqi: We will only return all the namespaces.

roryqi: It's complex for Gravitino to support nested namespace. 
In most cases, `schema` is enough for namespace management.
Iceberg nested namespace is a special case, so avoiding a large refactor is more cost-effective.
If Gravitino fully supports all the operations, it will bring more complexity especially for renaming operations.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,398 @@
...
+
+### Iceberg REST Side Behavior
+
+- **Create nested namespace**:
+  - Creating `A:B:C` will create (or ensure existence of) three schemas in Gravitino: `A`, `A.B`, and `A.B.C`.
+  - Set the created namespace owner as current user.
+- **Update nested namespace**:
+  - Support updating namespace properties through mapped schema operations.
+  - Property update is applied to the mapped target namespace scope.
+- **Drop nested namespace**: drop corresponding schema in Gravitino.
```
jerryshao: Will drop a namespace delete all the sub-namespaces and tables in the meanwhile?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,398 @@
...
+
+- `list schema` should express nested hierarchy semantics for users.
+- `list schema` REST API (GET `/metalakes/{metalake}/catalogs/{catalog}/schemas`) should support an
+  optional query parameter `parentHierarchicalSchema`.
+  - When `parentHierarchicalSchema` is not provided, return only top-level schemas (first layer).
+  - When `parentHierarchicalSchema` is provided, return only the direct child schemas under the
+    given parent (next layer), instead of the full subtree.
+  - `parentHierarchicalSchema` value follows the same logical `HierarchicalSchema` encoding
+    rules as described in `Identifier Rules` (segment percent-encoding, and then RFC 3986
+    percent-encoding for transport in a query component).
```
jerryshao, line 163: Give an example of how you do this.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,398 @@
...
+- Existing schema/table APIs remain compatible with non-nested cases.
+
+### Performance Considerations
+
+- Current `EntityStore` does not support prefix matching on schema names.
+- As a result, `list schema` with `parentHierarchicalSchema` cannot be implemented as an
+  efficient prefix query at storage layer. The server must list all schemas in the catalog and
+  then compute the top-level / direct-children view in memory.
+- This may introduce higher-than-expected latency and load for catalogs with a large number of
+  schemas.
```
jerryshao, line 204: Can we use something like range index to enhance the problem?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,398 @@
...
+
+Mitigations:
+
+- Cache computed hierarchy results per catalog with a short TTL and invalidate on schema
+  create/update/drop.
+- Enforce reasonable limits (pagination / maximum returned items) for list operations.
+- Add performance/regression tests with large schema counts to validate behavior.
+- Consider enhancing `EntityStore` to support prefix/index queries as a follow-up optimization.
```
jerryshao, line 212: I don't want to introduce cache, we'd better figure out a different solution like supporting filter/query in entitystore list API.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,398 @@
...
+- Example: for schemas `A`, `B`, `A:B`, `A:B:C`, hierarchy view is `A -> A:B -> A:B:C` and `B`;
+  root listing returns `A` and `B`, and querying parent `A` returns `A:B`.
+- To make nested semantics explicit, `list namespaces` should express parent-child relationships
+  (hierarchical view) even when underlying storage is flat.
+- Example hierarchical view from flat schemas: `A` -> `A:B` -> `A:B:C`, and `B` as another root.
+- This list-level hierarchical expression is the primary semantic model for users, reducing
+  ambiguity caused by one request creating multiple physical schema objects.
+- Gravitino server REST supports namespace create/update/drop operations for nested namespace
+  workflows, aligned with Iceberg REST behavior.
+- Existing schema/table APIs remain compatible with non-nested cases.
```
jerryshao, line 175: How to support create/delete from the gravitino side?

jerryshao: Please give the examples.

design-docs/iceberg-supported-nested-namespace.md
roryqi: I don't wanna introduce cache, too. I want `EntityStore` to support prefix queries.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,425 @@
...
+- `:` is mandatory in this design, not configurable.
+- For Gravitino REST create/update schema APIs, `request.getName()` keeps the logical schema name
+  and may contain `:` (for example `A:B` or `A:B:C`).
+- This phase does not pre-convert `request.getName()` from `:` to `.` before capability name
+  validation; capability naming rules must explicitly allow `:` for schema scope.
+- Schema name uses `:`-separated hierarchical representation in this design; `.` is not used as
+  schema separator.
+- `.` remains an Iceberg namespace hierarchy notation in external semantics only.
+- When `:` is used as separator, segment values containing `:` must be percent-encoded.
+- `%` must also be encoded as `%25` to avoid decode ambiguity.
```
mchades: Is the HTTP URL encoding rule being used here? What additional steps does the user need to take?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,425 @@
...
+
+Pros:
+
+- Better readability than escaping `.` in many clients and UI forms.
+- Lower routing conflict risk than `/`.
+- Easier to keep backward compatibility with existing non-nested schema handling.
+
+Cons:
+
+- Needs clear validation rule to avoid ambiguity with existing schema names containing `:`.
```
mchades, line 91: Is there a design for this validation rule?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,509 @@
...
+- External request/response handling uses the configured delimiter and converts at API boundary.
+- Connector-facing behavior remains Iceberg-compatible and does not require users to configure or
+  input internal storage representation.
+
+### Delimiter Validation Strategy
+
+For configurable schema delimiters, there are two possible strategies:
+
+- **Option 1: Restrict to a limited set of delimiters**
+  - Delimiter values are validated against an allowlist (for example `:`, `.`, `/`, `_`).
```
mchades: I remember you said the `.` delimiter is not allowed

roryqi: Yes, I should remove this.

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,509 @@
...
+- External request/response handling uses the configured delimiter and converts at API boundary.
+- Connector-facing behavior remains Iceberg-compatible and does not require users to configure or
+  input internal storage representation.
+
+### Delimiter Validation Strategy
+
+For configurable schema delimiters, there are two possible strategies, '.' isn't allowed:
+
+- **Option 1: Restrict to a limited set of delimiters**
+  - Delimiter values are validated against an allowlist (for example `:`, `;`, `$`).
```
mchades, line 143: Should you also explain how to determine whether a delimiter is allowed, or why other delimiters are not permitted?

jerryshao:
### Design Doc Review

Found 12 issues:

**1. `listSchemas` default behavior is a breaking REST API change**
Changing `GET /schemas` to return only top-level schemas (instead of all schemas) silently breaks existing clients with no error. Needs a migration strategy or an explicit opt-in flag (e.g. `?view=hierarchical`).

**2. No design for Java/Python client API hierarchy representation**
The doc only covers REST. How the Java/Python client APIs express nested schema paths — opaque strings with `:`, a new `SchemaPath` type, or varargs in `NameIdentifier` — is unaddressed. This is a public API surface decision that should be part of the design.

**3. Existing schemas with `:` in names silently break on upgrade**
Any schema currently named `team:sales` will be reinterpreted as a nested namespace `team → sales` after this feature is enabled. No migration scan or validation strategy is mentioned.

**4. No atomicity for parent-chain creation**
Creating `A:B:C` creates three schemas (`A`, `A.B`, `A.B.C`) in sequence. If the second step fails, `A` is left as an orphan with no rollback or cleanup defined.

**5. Privilege escalation during auto-created parent schemas**
If `A` does not yet exist when creating `A:B:C`, the requester becomes owner of `A`. A user with permission only to create leaf namespaces should not implicitly become owner of a top-level namespace.

**6. Physical storage `.` separator conflicts with `NameIdentifier` semantics**
The doc stores `A:B:C` as physical schema name `A.B.C`. Gravitino's `NameIdentifier` uses `.` as its level separator throughout the codebase. Any code path that reconstructs a `NameIdentifier` by splitting a full dotted path string will misparse `A.B.C` as three separate levels instead of one schema name. `NameIdentifier.java` and `NameIdentifierUtil.java` are listed as affected classes but the doc does not explain how this conflict is resolved.

**7. Configurable separator has no migration path**
If an operator changes the configured separator after data is already stored, all existing hierarchical schema names become unresolvable. The separator should be documented as immutable after first use, or a migration procedure should be defined.

**8. Authorization parent-chain walking adds unaddressed per-request overhead**
Every `USE_SCHEMA` check now walks the parent chain (`A:B:C` → `A:B` → `A`). The performance section covers listing latency but does not address the authorization check overhead for high-QPS workloads.

**9. "Drop nested namespace" section is incomplete**
The sentence "Will drop the schemas of the Gravitino" is cut off. The cascade behavior also does not specify whether tables inside dropped schemas are deleted or whether the operation is rejected if tables exist.

**10. `toPhysicalSchemaName` breaks if a namespace segment contains `.`**
The snippet uses `hierarchicalPath.replace(":", ".")`. If any namespace segment already contains `.` (e.g. a Hive schema named `my.schema`), the output is ambiguous and cannot be parsed back correctly.

**11. Cache invalidation not addressed for distributed deployments**
The performance mitigation mentions caching per catalog with TTL invalidation on mutations, but does not address how multiple Gravitino server instances stay consistent — servers could serve stale hierarchy views after schema mutations.

**12. Connector code change scope is understated**
The compatibility section implies this is a server-side change. In practice, Spark, Flink, and Trino connectors all require explicit code changes (the "Connector NestedNameIdentifier Conversion" section describes a non-trivial translation layer). The connector change scope should be explicitly called out.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

jerryshao:
> ### Design Doc Review
> Found 12 issues:
> 
> **1. `listSchemas` default behavior is a breaking REST API change** Changing `GET /schemas` to return only top-level schemas (instead of all schemas) silently breaks existing clients with no error. Needs a migration strategy or an explicit opt-in flag (e.g. `?view=hierarchical`).
> 
> **2. No design for Java/Python client API hierarchy representation** The doc only covers REST. How the Java/Python client APIs express nested schema paths — opaque strings with `:`, a new `SchemaPath` type, or varargs in `NameIdentifier` — is unaddressed. This is a public API surface decision that should be part of the design.
> 
> **3. Existing schemas with `:` in names silently break on upgrade** Any schema currently named `team:sales` will be reinterpreted as a nested namespace `team → sales` after this feature is enabled. No migration scan or validation strategy is mentioned.
> 
> **4. No atomicity for parent-chain creation** Creating `A:B:C` creates three schemas (`A`, `A.B`, `A.B.C`) in sequence. If the second step fails, `A` is left as an orphan with no rollback or cleanup defined.
> 
> **5. Privilege escalation during auto-created parent schemas** If `A` does not yet exist when creating `A:B:C`, the requester becomes owner of `A`. A user with permission only to create leaf namespaces should not implicitly become owner of a top-level namespace.
> 
> **6. Physical storage `.` separator conflicts with `NameIdentifier` semantics** The doc stores `A:B:C` as physical schema name `A.B.C`. Gravitino's `NameIdentifier` uses `.` as its level separator throughout the codebase. Any code path that reconstructs a `NameIdentifier` by splitting a full dotted path string will misparse `A.B.C` as three separate levels instead of one schema name. `NameIdentifier.java` and `NameIdentifierUtil.java` are listed as affected classes but the doc does not explain how this conflict is resolved.
> 
> **7. Configurable separator has no migration path** If an operator changes the configured separator after data is already stored, all existing hierarchical schema names become unresolvable. The separator should be documented as immutable after first use, or a migration procedure should be defined.
> 
> **8. Authorization parent-chain walking adds unaddressed per-request overhead** Every `USE_SCHEMA` check now walks the parent chain (`A:B:C` → `A:B` → `A`). The performance section covers listing latency but does not address the authorization check overhead for high-QPS workloads.
> 
> **9. "Drop nested namespace" section is incomplete** The sentence "Will drop the schemas of the Gravitino" is cut off. The cascade behavior also does not specify whether tables inside dropped schemas are deleted or whether the operation is rejected if tables exist.
> 
> **10. `toPhysicalSchemaName` breaks if a namespace segment contains `.`** The snippet uses `hierarchicalPath.replace(":", ".")`. If any namespace segment already contains `.` (e.g. a Hive schema named `my.schema`), the output is ambiguous and cannot be parsed back correctly.
> 
> **11. Cache invalidation not addressed for distributed deployments** The performance mitigation mentions caching per catalog with TTL invalidation on mutations, but does not address how multiple Gravitino server instances stay consistent — servers could serve stale hierarchy views after schema mutations.
> 
> **12. Connector code change scope is understated** The compatibility section implies this is a server-side change. In practice, Spark, Flink, and Trino connectors all require explicit code changes (the "Connector NestedNameIdentifier Conversion" section describes a non-trivial translation layer). The connector change scope should be explicitly called out.
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)

Most of the issues are valid and lack consideration. You should polish the doc to address all the issues. @roryqi

roryqi:
> ### esign Doc Review
> Found 12 issues:
> 
> **1. `listSchemas` default behavior is a breaking REST API change** Changing `GET /schemas` to return only top-level schemas (instead of all schemas) silently breaks existing clients with no error. Needs a migration strategy or an explicit opt-in flag (e.g. `?view=hierarchical`).
> 
> **2. No design for Java/Python client API hierarchy representation** The doc only covers REST. How the Java/Python client APIs express nested schema paths — opaque strings with `:`, a new `SchemaPath` type, or varargs in `NameIdentifier` — is unaddressed. This is a public API surface decision that should be part of the design.
> 
> **3. Existing schemas with `:` in names silently break on upgrade** Any schema currently named `team:sales` will be reinterpreted as a nested namespace `team → sales` after this feature is enabled. No migration scan or validation strategy is mentioned.
> 
> **4. No atomicity for parent-chain creation** Creating `A:B:C` creates three schemas (`A`, `A.B`, `A.B.C`) in sequence. If the second step fails, `A` is left as an orphan with no rollback or cleanup defined.
> 
> **5. Privilege escalation during auto-created parent schemas** If `A` does not yet exist when creating `A:B:C`, the requester becomes owner of `A`. A user with permission only to create leaf namespaces should not implicitly become owner of a top-level namespace.
> 
> **6. Physical storage `.` separator conflicts with `NameIdentifier` semantics** The doc stores `A:B:C` as physical schema name `A.B.C`. Gravitino's `NameIdentifier` uses `.` as its level separator throughout the codebase. Any code path that reconstructs a `NameIdentifier` by splitting a full dotted path string will misparse `A.B.C` as three separate levels instead of one schema name. `NameIdentifier.java` and `NameIdentifierUtil.java` are listed as affected classes but the doc does not explain how this conflict is resolved.
> 
> **7. Configurable separator has no migration path** If an operator changes the configured separator after data is already stored, all existing hierarchical schema names become unresolvable. The separator should be documented as immutable after first use, or a migration procedure should be defined.
> 
> **8. Authorization parent-chain walking adds unaddressed per-request overhead** Every `USE_SCHEMA` check now walks the parent chain (`A:B:C` → `A:B` → `A`). The performance section covers listing latency but does not address the authorization check overhead for high-QPS workloads.
> 
> **9. "Drop nested namespace" section is incomplete** The sentence "Will drop the schemas of the Gravitino" is cut off. The cascade behavior also does not specify whether tables inside dropped schemas are deleted or whether the operation is rejected if tables exist.
> 
> **10. `toPhysicalSchemaName` breaks if a namespace segment contains `.`** The snippet uses `hierarchicalPath.replace(":", ".")`. If any namespace segment already contains `.` (e.g. a Hive schema named `my.schema`), the output is ambiguous and cannot be parsed back correctly.
> 
> **11. Cache invalidation not addressed for distributed deployments** The performance mitigation mentions caching per catalog with TTL invalidation on mutations, but does not address how multiple Gravitino server instances stay consistent — servers could serve stale hierarchy views after schema mutations.
> 
> **12. Connector code change scope is understated** The compatibility section implies this is a server-side change. In practice, Spark, Flink, and Trino connectors all require explicit code changes (the "Connector NestedNameIdentifier Conversion" section describes a non-trivial translation layer). The connector change scope should be explicitly called out.

I will follow to solve the issues.

roryqi:
Update changes:
1.  We  previously only supported single-level namespaces, so there are no compatibility issues.
2. Added. We should add the Java cilent interface `listSchemas(String parentSchema)` 
3. The separator can be configured. We can choose another separator which won't have conflicts.
4. I modified the words. We don't create the not-existed namespace by hand. We will use underlying namespace operation capabilities. The Iceberg operations can create not-existed namespace atomicly.
5. If we create A:B:C, all of A, B and c doesn't exist, it equals that we created A, A:B, A:B:C, the requester will also have the privilege to create the parent namespace, so we should set them owner to requester.
6. The physical separator will be used in the storage layer, we don't expose to other places, Furthermore, I only use physical separator in the SchemeMetaService and MetadataObjectService.
7.  We don't need to migrate if we changed logic separator. Because we use physical separator in the storage layer
For example. 
when using : as logic separator, storage layer A.B.C -> logic layer A:B:C
when using $ as logic separator, storage layer A.B.C -> logic layer A$B$C
8. Our entity store has cache. It won't bring much cost. Could u ensure this point? cc @yuqi1129 
9.  Iceberg operations don't support cascade. I will throw exception when specifying the cascade mode
10.  We don't support schema name contains `.`. It won't break anything.
11. It is out of the scope about this proposal
12. Added the process.

roryqi:
I and @mchades 's decision point is the segment 
```
Two delimiter-governance options are under evaluation:
```

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+
+- **Persisted schema name in `EntityStore` always uses `.` as the internal storage separator** for
+  stable storage semantics.
+- External request/response handling uses configured logical separator and converts at API boundary.
+- Connector-facing behavior remains Iceberg-compatible and does not require users to configure or
+  input internal storage representation.
+- Mapping must be reversible:
+  - `logical path segments` -> `encode each segment` -> `join by '.'` for physical storage.
+  - physical schema name -> `split by '.'` -> `decode each segment` -> logical path segments.
+  - This avoids ambiguity when one segment contains `.` (for example `my.schema`).
```
jerryshao: Can you give me an example for the physical schema name structure?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+
+### Existing Name Compatibility and Migration Guard
+
+- Before enabling nested namespace mode for a catalog, run a pre-check scan on existing schema names
```
jerryshao: How do you enable this and how do you do the check?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+  against configured logical separator.
+- If existing schema names conflict with selected separator, enabling is rejected with actionable
+  error and user can choose another separator or rename conflicting schema names.
+- Once nested mode is enabled for a catalog, creating new schema names containing configured logical
+  separator as plain text is rejected.
+
+### Delimiter Configuration Policy
+
+- Logical separator is configurable (for example `:`, `;`, `$`) and can be chosen to avoid conflict
```
jerryshao: Is this per catalog or a global configuration?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+      configured delimiter.
+    - Nested interpretation is applied only in nested-aware request paths (for example Iceberg
+      namespace APIs), not as a blanket rule for all engines.
+
+### Delimiter Validation and Rejection Rationale
+
+Delimiter validity should be explicit and observable, not implicit.
+
+- Validation checkpoints:
+  - Validate delimiter when server starts or when delimiter configuration is updated.
```
jerryshao: How do you know the configuration is updated?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+  - Re-validate against existing catalog schema names before enabling nested mode for a catalog.
+  - Reject invalid delimiter configuration early before request-time namespace operations.
```
jerryshao, line 183: If it is invalid, will you fail the server? What is your behavior?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+  - For create/update/drop and other schema operations, continue using existing string-based schema
+    name parameters; users compose hierarchical schema names themselves (for example `A:B:C`).
+  - No new typed path wrapper (`SchemaPath`) or segment-based overloads are introduced in this
+    phase.
+
+Simple Java client examples:
+
+- `listSchemas()` -> returns top-level schemas, for example `A`, `B`.
+- `listSchemas("A:B")` -> returns direct children, for example `A:B:C`, `A:B:D`.
+- `createSchema("A:B:C", ...)` -> users pass composed schema name directly.
```
jerryshao, line 325: You mentioned that delimiter can be configured, so here the user should know what the delimiter is for the schema string, right?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,573 @@
...
+- Connector to Gravitino identifier: schema `A:B:C`, table `t1`
+- Gravitino response schema: `A:B:C`
+- Connector rendered back to engine: namespace `[A, B, C]`, table `t1`
+
+Trino-specific example:
+
+- Trino input namespace string: `"a.b.c"`
+- Connector converts to Gravitino internal schema: `a:b:c`
+- Gravitino returns schema: `a:b:c`
+- Connector renders back to Trino quoted style: `"a.b.c"`
```
jerryshao, line 557: Add a chapter to list all the planned works of this features.

design-docs/iceberg-supported-nested-namespace.md
roryqi: This is a global configuration now.

design-docs/iceberg-supported-nested-namespace.md
roryqi: Yes, the user should know the delimiter.

design-docs/iceberg-supported-nested-namespace.md
roryqi: For example, 
`A.B.C` will be stored in the physical layer.

design-docs/iceberg-supported-nested-namespace.md
roryqi: Upgrade the new version and select all the Iceberg namespaces to check whether contains the delimiter.

design-docs/iceberg-supported-nested-namespace.md
roryqi: Yes, we should fail to start the server. Because it used a wrong server configuration.

design-docs/iceberg-supported-nested-namespace.md
roryqi: Because we don't support reloading the server configuration. I can remove it first.

design-docs/iceberg-supported-nested-namespace.md
roryqi: Added
## Plans
-  Add foundational classes for nested namespace authorization
-  Add nested namespace support to Iceberg REST API
-  Add nested namespace support to Gravitino REST AP

## No Plan
- Trino connector supports nested namespace
- Spark connector supports nested namespace
- Flink connector supports nested namespace

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,582 @@
...
+
+Cons:
+
+- Requires explicit conversion rules between logical path and physical schema name.
+- Authorization matching and identifier serialization become more complex.
+#### Option B Separator: Configurable logical separator (default `:`)
+
+Examples:
+
+- `A:B:C` as logical `HierarchicalSchema` path when separator is `:`.
```
mchades, line 80: how about naming `HierarchicalSchema` to `nestedSchema` ?

design-docs/iceberg-supported-nested-namespace.md
```diff
@@ -0,0 +1,582 @@
...
+Pros:
+
+- Better readability than escaping `.` in many clients and UI forms.
+- Lower routing conflict risk than `/`.
+- Easier to keep backward compatibility with existing non-nested schema handling.
+
+Cons:
+
+- Needs clear validation rule to avoid ambiguity with existing schema names containing configured
+  separator.
```
mchades, line 92: Is there a more detailed explanation for this part in this design document?

design-docs/iceberg-supported-nested-namespace.md
roryqi: Actually Hierarchical is use more commonly. GCS uses `Hierarchical` concept.

design-docs/iceberg-supported-nested-namespace.md
roryqi: You can see 183 line.

jerryshao (approved):

roryqi:
Two delimiter-governance options are under evaluation @jerryshao Could u give your feedback about this point?

wo delimiter-governance options are under evaluation:

- **Option 1 (restricted delimiter set)**:
  - Server only accepts delimiters from a predefined allowed set.
  - Delimiter is treated as a reserved hierarchy marker in nested-aware flows.
  - Behavior difference for new object creation:
    - Iceberg nested schema creation that uses delimiter as hierarchy path is allowed.
    - Hive creation of a non-nested schema name that contains the configured delimiter is rejected
      with validation error.
- **Option 2 (unrestricted delimiter)**:
  - Server allows any delimiter value configured by users.
  - Compatibility is prioritized for engines that treat schema name as plain string.
  - Behavior difference for new object creation:
    - Hive can create a non-nested schema successfully even when the schema name contains the
      configured delimiter.
    - Nested interpretation is applied only in nested-aware request paths (for example Iceberg
      namespace APIs), not as a blanket rule for all engines.

jerryshao:
> Two delimiter-governance options are under evaluation @jerryshao Could u give your feedback about this point?
> 
> wo delimiter-governance options are under evaluation:
> 
> * **Option 1 (restricted delimiter set)**:
>   
>   * Server only accepts delimiters from a predefined allowed set.
>   * Delimiter is treated as a reserved hierarchy marker in nested-aware flows.
>   * Behavior difference for new object creation:
>     
>     * Iceberg nested schema creation that uses delimiter as hierarchy path is allowed.
>     * Hive creation of a non-nested schema name that contains the configured delimiter is rejected
>       with validation error.
> * **Option 2 (unrestricted delimiter)**:
>   
>   * Server allows any delimiter value configured by users.
>   * Compatibility is prioritized for engines that treat schema name as plain string.
>   * Behavior difference for new object creation:
>     
>     * Hive can create a non-nested schema successfully even when the schema name contains the
>       configured delimiter.
>     * Nested interpretation is applied only in nested-aware request paths (for example Iceberg
>       namespace APIs), not as a blanket rule for all engines.

I think letting users configure any delimiter they want is more flexible. Defining an allowed list will always have a problem that users want a delimiter other than in the allowed list.

From a fresh user's perspective, I think rejection is better than inconsistent behavior between catalogs. Otherwise, they will be confused whether catalog A's behavior is different from catalog B implicitly.

From an existing user's perspective, if they already use the delimiter, failing the operation will break their compatibility. If we can add something in the code level to let them explicitly know that this new feature will affect their current usage, I think it is acceptable. Otherwise, mentioning in the doc may not be enough.

roryqi:
> I think letting users configure any delimiter they want is more flexible. Defining an allowed list will always have a problem that users want a delimiter other than in the allowed list.
> 
> From a fresh user's perspective, I think rejection is better than inconsistent behavior between catalogs. Otherwise, they will be confused whether catalog A's behavior is different from catalog B implicitly.
> 
> From an existing user's perspective, if they already use the delimiter, failing the operation will break their compatibility. If we can add something in the code level to let them explicitly know that this new feature will affect their current usage, I think it is acceptable. Otherwise, mentioning in the doc may not be enough.

Thanks for @jerryshao's suggestion. cc @mchades

PR #10721, opened 2026-04-08
 [#10368] improvement(flink): persist Paimon bucket distribution when creating table via Flink connector

### LINKED ISSUE #10368: [Improvement] Flink connector support Paimon distribution (bucket / bucket-key)
Flink Paimon cannot effectively persist distribution today.

Background:
- Flink createTable always sends `Distributions.NONE` and empty sort orders:
  - /Users/fanng/opensource/gravitino/flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalog.java:307
- Paimon table builder removes `bucket`/`bucket-key` from properties, then only writes them back from `distribution`:
  - /Users/fanng/opensource/gravitino/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java:89
  - /Users/fanng/opensource/gravitino/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java:91

So with Flink path setting `Distributions.NONE`, distribution info is lost.

Suggested SQL for validation:

```sql
USE CATALOG paimon_catalog;
CREATE DATABASE IF NOT EXISTS dist_db;
USE dist_db;

CREATE TABLE dist_tbl (
  id BIGINT,
  name STRING,
  PRIMARY KEY (id) NOT ENFORCED
) WITH (
  'bucket' = '4',
  'bucket-key' = 'id'
);

SHOW CREATE TABLE dist_tbl;
```

How should we improve?
- In Flink connector, parse Paimon bucket options (`bucket`, `bucket-key`) into Gravitino `Distribution` for createTable.
- Keep behavior backward-compatible when bucket options are absent.
- Add Flink Paimon integration test to assert distribution is persisted (via Gravitino table metadata).
- Update Flink Paimon docs with distribution example and limitation notes (HASH only).

Out of scope:
- Sort orders.
- Row-level operations.

sgedward:

  When creating a Paimon table via the Flink connector with bucket options (`bucket-key`, `bucket`), the distribution metadata was silently dropped because Flink's `createTable` always
  passed `Distributions.NONE`.

  This PR adds end-to-end support for Paimon bucket distribution through the Flink connector, using a **pass-through approach** — the Flink connector simply translates between Flink table
   options and Gravitino's Distribution model without enforcing Paimon-specific validation rules. All Paimon-related logic (e.g., defaulting bucket-key to primary keys, validating bucket
  numbers) is deferred to the Paimon catalog service.

  **Flink connector (write path — `createTable`):**
  - Added a `toGravitinoDistribution` hook in `BaseCatalog` (defaults to `Distributions.NONE`) to allow catalog-specific distribution parsing from Flink table options.
  - Overrode the hook in `GravitinoPaimonCatalog` to parse `bucket-key` and `bucket` into a Gravitino `Distribution` object (HASH strategy).
  - Overrode `toGravitinoTableProperties` in `PaimonPropertiesConverter` to strip `bucket-key` and `bucket` before passing properties to Gravitino, since they are now represented as
  Distribution metadata.

  **Flink connector (read path — `toFlinkTable`):**
  - Added a `fromGravitinoDistribution` hook in `BaseCatalog` (defaults to empty map) to convert Distribution back into Flink table options.
  - Overrode the hook in `GravitinoPaimonCatalog` to reconstruct `bucket-key` and `bucket` from the Gravitino Distribution.
  - Overrode `toFlinkTableProperties` in `PaimonPropertiesConverter` to strip bucket properties from raw table properties, ensuring Distribution is the single source of truth.

  **Catalog-side (`catalog-lakehouse-paimon`):**
  - Refactored `GravitinoPaimonTable.getDistribution` to handle `-1` as AUTO and support bucket-only distribution (no explicit bucket-key).
  - Relaxed `PaimonCatalogOperations.validateDistribution` to allow HASH distribution without explicit bucket-key expressions, deferring the primary-key-subset check to Paimon itself.

  Flink's `createTable` always passed `Distributions.NONE`, so `bucket-key` and `bucket` configs specified in Flink SQL `WITH` options were silently lost. Additionally, the server-side
  validation was overly strict — it rejected distributions without explicit bucket-key even though Paimon automatically falls back to primary keys.

  Fix: #10368

  Yes:
  - Users can now specify `bucket-key` and `bucket` in Flink SQL `WITH` options when creating Paimon tables, and the distribution will be correctly persisted in Gravitino metadata and
  round-tripped back through queries like `SHOW CREATE TABLE`.
  - Specifying only `bucket` (without `bucket-key`) is now supported — Paimon will use primary keys as the default bucket keys.

  - Added unit test in `TestBaseCatalog` verifying the default hooks return `Distributions.NONE` / empty map.
  - Added unit tests in `TestPaimonPropertiesConverter` covering:
    - `getDistribution`: null properties, blank bucket-key, missing bucket-key, missing bucket, both present, `-1` as AUTO, invalid bucket number, negative bucket number pass-through.
    - `distributionToProperties`: HASH with bucket/bucket-key, NONE distribution, null distribution, AUTO distribution, AUTO with no expressions.
    - `toGravitinoTableProperties`: verifying bucket properties are stripped.
    - Round-trip idempotency (bucket-key only ↔ bucket-key with explicit `-1`).
  - Added integration test `testBucketDistributionRoundTrip` in `FlinkPaimonCatalogIT` asserting the distribution is correctly persisted via Gravitino and round-trips back through Flink's
   `SHOW CREATE TABLE`.

### DIFF
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
@@ -116,7 +116,7 @@ public class GravitinoPaimonTable extends BaseTable {
             GravitinoPaimonColumn.fromPaimonRowType(table.rowType())
                 .toArray(new GravitinoPaimonColumn[0]))
         .withPartitioning(toGravitinoPartitioning(table.partitionKeys()))
-        .withDistribution(getDistribution(table.options()))
+        .withDistribution(getDistribution(table.options(), table.primaryKeys()))
         .withComment(table.comment().orElse(null))
         .withProperties(table.options())
         .withIndexes(constructIndexesFromPrimaryKeys(table))
@@ -204,12 +204,17 @@ public class GravitinoPaimonTable extends BaseTable {
       return;
     }
 
-    List<String> bucketKeys = getBucketKeys(distribution);
-    if (!bucketKeys.isEmpty()) {
-      properties.put(BUCKET_KEY, String.join(",", bucketKeys));
+    int number = distribution.number();
+    // Paimon does not allow 'bucket-key' with bucket=-1 (dynamic mode).
+    // In dynamic mode, Paimon uses the primary key as the bucket key automatically.
+    if (number != Distributions.AUTO) {
+      List<String> bucketKeys = getBucketKeys(distribution);
+      if (!bucketKeys.isEmpty()) {
+        properties.put(BUCKET_KEY, String.join(",", bucketKeys));
+      }
     }
 
-    properties.put(BUCKET_NUM, String.valueOf(distribution.number()));
+    properties.put(BUCKET_NUM, String.valueOf(number == Distributions.AUTO ? -1 : number));
   }
 
   private static List<String> getBucketKeys(Distribution distribution) {
@@ -228,31 +233,50 @@ public class GravitinoPaimonTable extends BaseTable {
         .collect(Collectors.toList());
   }
 
-  static Distribution getDistribution(Map<String, String> properties) {
+  static Distribution getDistribution(Map<String, String> properties, List<String> primaryKeys) {
     if (properties == null) {
       return Distributions.NONE;
     }
+
+    // Resolve bucket key columns: explicit bucket-key, else fall back to PK
     String bucketKeys = properties.get(BUCKET_KEY);
-    if (StringUtils.isBlank(bucketKeys)) {
-      return Distributions.NONE;
+    List<String> bucketKeyList;
+    if (StringUtils.isNotBlank(bucketKeys)) {
+      bucketKeyList =
+          Arrays.stream(bucketKeys.split(","))
+              .map(String::trim)
+              .filter(StringUtils::isNotBlank)
+              .collect(Collectors.toList());
+    } else {
+      bucketKeyList = (primaryKeys != null) ? primaryKeys : Collections.emptyList();
     }
-    List<String> bucketKeyList =
-        Arrays.stream(bucketKeys.split(","))
-            .map(String::trim)
-            .filter(StringUtils::isNotBlank)
-            .collect(Collectors.toList());
+
+    String bucketValue = properties.get(BUCKET_NUM);
+
     if (bucketKeyList.isEmpty()) {
       return Distributions.NONE;
     }
+
     Expression[] expressions =
         bucketKeyList.stream().map(NamedReference::field).toArray(Expression[]::new);
-    String bucketValue = properties.get(BUCKET_NUM);
+
     if (StringUtils.isBlank(bucketValue)) {
       return Distributions.auto(Strategy.HASH, expressions);
     }
+
     String trimmedBucketValue = bucketValue.trim();
     try {
-      return Distributions.hash(Integer.parseInt(trimmedBucketValue), expressions);
+      int parsedBucket = Integer.parseInt(trimmedBucketValue);
+      if (parsedBucket == -1) {
+        return Distributions.auto(Strategy.HASH, expressions);
+      }
+      if (parsedBucket <= 0) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Paimon bucket number must be a positive integer or -1 (dynamic mode), but was '%s'.",
+                bucketValue));
+      }
+      return Distributions.hash(parsedBucket, expressions);
     } catch (NumberFormatException e) {
       throw new IllegalArgumentException(
           String.format("Paimon bucket number must be a valid integer, but was '%s'.", bucketValue),
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
@@ -251,7 +251,7 @@ public class GravitinoPaimonTable extends BaseTable {
       bucketKeyList = (primaryKeys != null) ? primaryKeys : Collections.emptyList();
     }
 
-    String bucketValue = properties.get(BUCKET_NUM);
+    String bucketNum = properties.get(BUCKET_NUM);
 
     if (bucketKeyList.isEmpty()) {
       return Distributions.NONE;
@@ -260,13 +260,12 @@ public class GravitinoPaimonTable extends BaseTable {
     Expression[] expressions =
         bucketKeyList.stream().map(NamedReference::field).toArray(Expression[]::new);
 
-    if (StringUtils.isBlank(bucketValue)) {
+    if (StringUtils.isBlank(bucketNum)) {
       return Distributions.auto(Strategy.HASH, expressions);
     }
 
-    String trimmedBucketValue = bucketValue.trim();
     try {
-      int parsedBucket = Integer.parseInt(trimmedBucketValue);
+      int parsedBucket = Integer.parseInt(bucketNum.trim());
       if (parsedBucket == -1) {
         return Distributions.auto(Strategy.HASH, expressions);
       }
@@ -274,12 +273,12 @@ public class GravitinoPaimonTable extends BaseTable {
         throw new IllegalArgumentException(
             String.format(
                 "Paimon bucket number must be a positive integer or -1 (dynamic mode), but was '%s'.",
-                bucketValue));
+                bucketNum));
       }
       return Distributions.hash(parsedBucket, expressions);
     } catch (NumberFormatException e) {
       throw new IllegalArgumentException(
-          String.format("Paimon bucket number must be a valid integer, but was '%s'.", bucketValue),
+          String.format("Paimon bucket number must be a valid integer, but was '%s'.", bucketNum),
           e);
     }
   }
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
@@ -116,7 +116,7 @@ public class GravitinoPaimonTable extends BaseTable {
             GravitinoPaimonColumn.fromPaimonRowType(table.rowType())
                 .toArray(new GravitinoPaimonColumn[0]))
         .withPartitioning(toGravitinoPartitioning(table.partitionKeys()))
-        .withDistribution(getDistribution(table.options(), table.primaryKeys()))
+        .withDistribution(getDistribution(table.options()))
         .withComment(table.comment().orElse(null))
         .withProperties(table.options())
         .withIndexes(constructIndexesFromPrimaryKeys(table))
@@ -204,17 +204,12 @@ public class GravitinoPaimonTable extends BaseTable {
       return;
     }
 
-    int number = distribution.number();
-    // Paimon does not allow 'bucket-key' with bucket=-1 (dynamic mode).
-    // In dynamic mode, Paimon uses the primary key as the bucket key automatically.
-    if (number != Distributions.AUTO) {
-      List<String> bucketKeys = getBucketKeys(distribution);
-      if (!bucketKeys.isEmpty()) {
-        properties.put(BUCKET_KEY, String.join(",", bucketKeys));
-      }
+    List<String> bucketKeys = getBucketKeys(distribution);
+    if (!bucketKeys.isEmpty()) {
+      properties.put(BUCKET_KEY, String.join(",", bucketKeys));
     }
 
-    properties.put(BUCKET_NUM, String.valueOf(number == Distributions.AUTO ? -1 : number));
+    properties.put(BUCKET_NUM, String.valueOf(distribution.number()));
   }
 
   private static List<String> getBucketKeys(Distribution distribution) {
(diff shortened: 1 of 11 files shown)


sgedward:
Hi @FANNG1,

I was wondering if you could help me review this PR. Thank you very much!

CC: @roryqi @justinmclean

sgedward:
Hi @FANNG1 ,

I make some update to the test coverage, can you help me rerun the workflow? Thank you very much!

FANNG1:
Thanks for the fix! The overall approach — adding a `toGravitinoDistribution` hook on `BaseCatalog` and overriding it in the Paimon catalog — is clean and minimally invasive. I have two concerns I'd like to discuss before merging.

## 1. (Blocker) Missing `bucket-key` is silently dropped on PK tables

`GravitinoPaimonCatalog.getDistribution` returns `Distributions.NONE` whenever `bucket-key` is blank, and `PaimonPropertiesConverter.toGravitinoTableProperties` then strips both `bucket-key` and `bucket`. But per the [Paimon 1.2 config docs](https://paimon.apache.org/docs/1.2/maintenance/configurations/#bucket), `bucket-key` is **optional and defaults to the primary key** for PK tables. So the following perfectly valid statement:

```sql
CREATE TABLE t (
  id BIGINT,
  name STRING,
  PRIMARY KEY (id) NOT ENFORCED
) WITH ('bucket' = '8');
```

will be silently created with no distribution at all instead of an 8-bucket HASH(id) table. This is worse than the original bug — it's silent semantic corruption of the user's table layout.

**Suggested fix (sketch):** let the hook see schema/PK info (the `resolvedTable` is already in scope in `BaseCatalog.createTable`), and in the Paimon override fall back to the primary key columns when `bucket-key` is blank. If there's no PK either and `bucket` is set, fail fast with a clear error rather than silently dropping it. Please also add tests for: PK + only `bucket`, PK + explicit `bucket-key` overriding PK, and non-PK + only `bucket`.

## 2. Load path doesn't reconstruct `bucket-key` / `bucket` from `Distribution`

The PR converts properties → `Distribution` on create, but doesn't do the reverse on load. Today this happens to work because `GravitinoPaimonTable.fromPaimonTable` (`GravitinoPaimonTable.java:121`) does `withProperties(table.options())`, leaving the raw `bucket-key`/`bucket` in `properties`, and the Flink connector just passes them through. **But that's an implicit dependency**, not something this PR establishes.

Risks of relying on it:

- **Asymmetry / fragility**: if anyone later cleans up the server side to make `Distribution` the single source of truth (and stops echoing reserved properties back into `properties`), Flink load will silently lose bucket info.
- **Cross-entry consistency**: a table created via REST / Spark / Java client may have `Distribution` set but not have `bucket-key`/`bucket` in `properties`. Flink load would then be inconsistent with Spark load on the same table.
- **Alter path**: `BaseCatalog.alterTable` calls `getTable(tablePath)` first and diffs against the new table. If `existingTable.getOptions()` is missing the bucket keys, an unrelated alter could end up dropping the distribution.

**Suggested fix (sketch):** add a symmetric reverse step in `GravitinoPaimonCatalog.toFlinkTable` (or a `fromGravitinoDistribution` hook on `BaseCatalog`) that reconstructs `bucket-key` / `bucket` from `table.distribution()`, treating `Distribution` as the authoritative source rather than relying on properties round-tripping through the server. At a minimum, please add a **round-trip integration test**: create a table with `bucket` / `bucket-key` via Flink SQL, then load it back through Flink (`SHOW CREATE TABLE` or `tableEnv.getCatalog().getTable(...)`) and assert both options are present and correct — without this, the implicit dependency above will silently break in the future.

---

**Summary:** (1) is a blocker (silent data-layout corruption); (2) is strongly recommended, or at the very least a round-trip test should be added. Happy to discuss — thanks again for picking this up!

sgedward:
Hi @FANNG1,

Thank you very much for the detailed and informative explanation. I have updated the code to address both problems as you suggested.

Please let me know if you have any further questions or feedback.

FANNG1:
Thanks, both earlier concerns are cleanly addressed. One remaining issue plus two nits.

## (Should fix) `distributionToProperties` drops `BUCKET_NUM` when `number == AUTO`

```java
if (number != Distributions.AUTO) {
  properties.put(PaimonConstants.BUCKET_NUM, String.valueOf(number));
}
```

Problems:

1. **Depends on Paimon's default for `bucket`**, which has drifted across versions (historically `1` = single bucket, later `-1` = dynamic). On an older Paimon, a round-tripped `SHOW CREATE TABLE` that omits `bucket` will silently re-create as a single-bucket table — the same class of silent semantic change this PR was meant to prevent.
2. **Inconsistent with server-side `GravitinoPaimonTable.applyDistribution`** (`GravitinoPaimonTable.java:212`), which unconditionally writes `BUCKET_NUM` regardless of AUTO.

Suggested fix — always write it, mapping AUTO to Paimon's `-1` sentinel ([docs](https://paimon.apache.org/docs/1.2/maintenance/configurations/#bucket)):

```java
properties.put(
    PaimonConstants.BUCKET_NUM,
    String.valueOf(number == Distributions.AUTO ? -1 : number));
```

## (Related) `getDistribution` doesn't normalize `bucket=-1`

- `bucket` blank → `Distributions.auto(HASH, ...)`
- `bucket=-1` → `Distributions.hash(-1, ...)`

After the fix above we'll write `-1` on load and re-parse it on the next create, so these two paths must produce equal `Distribution` objects. Please normalize:

```java
int parsed = Integer.parseInt(trimmedBucketValue);
return parsed == -1
    ? Distributions.auto(Strategy.HASH, expressions)
    : Distributions.hash(parsed, expressions);
```

And add a test asserting `distributionToProperties(getDistribution(...))` is idempotent for both the blank-`bucket` and explicit-`bucket=-1` cases.

## Nits

- `distributionToProperties` silently drops non-`NamedReference` expressions via `.filter(...)`. Prefer `Preconditions.checkArgument` to fail loudly.
- `BaseCatalog.fromGravitinoDistribution` default returns `new HashMap<>()` per call; `Collections.emptyMap()` works with `putAll`.

sgedward:
Hi @FANNG1

Thanks for pointing out the fix and explaining the issue to me. I had an earlier fix which just set the distribution to none when it was -1. I will update my code later to reflect your suggested fix.

Thank you very much.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
```diff
@@ -228,31 +233,50 @@ private static List<String> getBucketKeys(Distribution distribution) {
...
+    } else {
+      bucketKeyList = (primaryKeys != null) ? primaryKeys : Collections.emptyList();
     }
-    List<String> bucketKeyList =
-        Arrays.stream(bucketKeys.split(","))
-            .map(String::trim)
-            .filter(StringUtils::isNotBlank)
-            .collect(Collectors.toList());
+
+    String bucketValue = properties.get(BUCKET_NUM);
```
FANNG1: could you rename `bucketValue` to `bucketNum` to make it more clear?

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/GravitinoPaimonTable.java
```diff
@@ -204,12 +204,17 @@ private static void applyDistribution(Map<String, String> properties, Distributi
       return;
     }
 
-    List<String> bucketKeys = getBucketKeys(distribution);
-    if (!bucketKeys.isEmpty()) {
-      properties.put(BUCKET_KEY, String.join(",", bucketKeys));
+    int number = distribution.number();
+    // Paimon does not allow 'bucket-key' with bucket=-1 (dynamic mode).
```
FANNG1, line 208: could you show me the code or docs?

sgedward: When I ran the IT test locally with `bucket-key` allowed, I ran into a bucket key not allowed exception when bucket=`-1`. So I found this in the code.

https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/schema/SchemaValidation.java#L575C1-L579C14

FANNG1:
I wonder if we may be putting a bit too much Paimon-specific logic into the Flink connector here.

Would it make sense to keep the Flink connector thinner and let it do only the minimum adaptation work, e.g. extracting `bucket-key` / `bucket` from table properties and converting them into a `Distribution`, while leaving provider-specific semantics such as PK fallback, `bucket = -1` handling, and bucket-number validation to the Gravitino server side (`catalog-lakehouse-paimon`)?

My concern is that once this logic lives in the connector, we may end up duplicating the same Paimon semantics again for Spark / Trino later, and it also increases the chance of semantic drift between connector-side and server-side implementations. Curious what you think about centralizing that part on the server side.

sgedward:
Hi @FANNG1 

Thanks for the detailed explanation — I agree we should defer to server-side validation.

After reviewing some of the documents, I have a few questions regarding the `DISTRIBUTION` alignment with Paimon options.

As the Paimon Core config options list:

**Bucket:** `positive`, `-1`, and `-2`

**Bucket-key:** Specifies the Paimon distribution policy. Data is assigned to each bucket according to the hash value of `bucket-key`.
- If you specify multiple fields, the delimiter is `,`.
- If not specified, the primary key will be used; if there is no primary key, the full row will be used.

---

From the above option combination, I created the table below. It seems that `AUTO` and `NONE` behave the same way, except `AUTO` cannot accept an empty key expression. I also wonder — since `bucket-key` can use at minimum the full row as the key, do we still need to check if a PK fallback exists?

https://github.com/apache/gravitino/blob/aa1586b422730aaf9d28265fff62d31f0c1ee29b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java#L533-L539

| bucket   | bucket-key | Flink connector Distribution     |
|----------|------------|----------------------------------|
| blank    | missing    | NONE                             |
| blank  | exists     | AUTO(-1, [key])                  |
| -1       | missing    | hash(-1)                         |
| -1       | exists     | hash(-1, [key]) → server rejects |
| -2       | missing    | hash(-2)                         |
| -2       | exists     | hash(-2, [key])                  |
| >0       | missing    | hash(N)                          |
| >0       | exists     | hash(N, [key])                   |

FANNG1:
Maybe we can define the boundary like this:

- In the Flink connector, only convert explicit options structurally.
- If both `bucket-key` and `bucket` are present, convert to `Distribution.hash(bucket, bucketKeyExpressions)`.
- If only `bucket-key` is present, convert to `Distribution.auto(HASH, bucketKeyExpressions)`.
- If only `bucket` is present, convert to `Distribution.hash(bucket)` with empty expressions.
- Do not do PK fallback or bucket-number semantic validation in the connector.

On the Gravitino server side, I also think we should avoid inferring too much Paimon-specific behavior. The Paimon catalog can convert the received `Distribution` back to explicit Paimon options, but it should not try to infer missing `bucket-key` from primary keys or full-row fallback either. For example, an empty-expression distribution can simply mean “bucket number is specified, but bucket-key is unspecified”, and the server can pass that through as Paimon properties where possible, letting Paimon apply its own fallback semantics.

This keeps both sides as structural adapters as much as possible, avoids duplicating Paimon inference rules in Gravitino, and should make future Spark/Trino support easier.

sgedward:
Hi @FANNG1,

I totally agree with your suggestion. I have removed the connector/server-side Paimon logic checking and pushed it down to the Paimon server for verification.

* All `bucket-key` and `bucket` situations go to `hash()` distribution (including `bucket:-1`).
* `Bucket` key only goes to `hash()`.
* `bucket-key` only goes to `Auto()`, even if it should be rejected by Paimon.
* No `bucket-key` or `bucket` config goes to `None`.

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
```diff
@@ -77,4 +90,93 @@ public void dropTable(ObjectPath tablePath, boolean ignoreIfNotExists)
...
+    int number = distribution.number();
+    Expression[] expressions = distribution.expressions();
+    boolean hasExpressions = expressions != null && expressions.length > 0;
+
+    if (number == Distributions.AUTO && !hasExpressions) {
+      return properties;
+    }
+
+    // Paimon does not allow 'bucket-key' with bucket=-1 (dynamic mode).
+    if (number != Distributions.AUTO && hasExpressions) {
```
FANNG1: could we remove the number check from `if (number != Distributions.AUTO && hasExpressions) `?   is it possible Gravitino will response a distribution with expressions and number equals to auto?

FANNG1:
LGTM except one comment, could you track it?

flink-connector/flink/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
sgedward: Good catch, It should be expected an error from Paimon and the creation should failed. The IT test should cover this route.I have update the unit test to reflect the pass through mode.

sgedward:
Hi @FANNG1 ,

Let me know if you have anything you want to change.

FANNG1:
LGTM

FANNG1 (approved):

sgedward:
Hi Reviewer (@roryqi @justinmclean),

The PR has been approved. Could you help me merge it? Thanks!

FANNG1:
Thanks for the great work and patience through the review iterations!  Appreciate the willingness to iterate on the design. 👍

PR #10722, opened 2026-04-08
Fix misleading success log when schema drop returns false

### LINKED ISSUE #10219: [Improvement] Fix misleading success log when schema drop operation returns false

SchemaOperations.dropSchema logs "Schema dropped" unconditionally, even when dispatcher.dropSchema(...) returns false. This creates incorrect success audit logs for failed (not-dropped) operations and can mislead operators.

### How should we improve?

Make success logging conditional on the actual drop result. Keep the existing warning for when dropped is false, and only emit LOG.info("Schema dropped: ...") when dropped is true.

LukasDEDD:
This PR fixes an issue where SchemaOperations.dropSchema logged "Schema dropped" even when the drop operation returned false. Logging is now conditional on the actual result.

Why are the changes needed?
The previous behavior produced misleading success logs. This fix ensures that logs accurately reflect the real outcome of the drop operation.

Fix: #10219

No. Only the logging behavior is corrected; no API or functional behavior is changed.

How was this patch tested?
The change was tested manually:

When dispatcher.dropSchema(...) returns true -> an INFO log "Schema dropped" is emitted.

When it returns false -> a WARN log "Fail to drop schema ..." is emitted.


LukasDEDD:
@justinmclean Hi, I've created the PR in the official repository as requested. Please let me know if any further changes are needed.

justinmclean:
We use spotless to ensure code is consistently formatted. You'll need to run `./gradlew :server:spotlessApply` and push the changes to fix the CI issue.

LukasDEDD:
Done. I've applied Spotless and force-pushed the changes. Could you please run the CI workflows again?

justinmclean (approved):
Thanks for this

PR #10723, opened 2026-04-08
[#10746] fix(auth): Cache JWKSource and downgrade token validation log to WARN

bharos:

1. Cache `JWKSource` in `initialize()` as an instance field instead of creating a new one per `validateToken()` call.
2. Downgrade the catch-block log from `LOG.error` to `LOG.warn` for token validation failures.
3. Remove the now-unused `createJwkSource()` private method.

- **Per-request `JWKSource` creation**: `JWKSourceBuilder.create(url).build()` was called on every `validateToken()` invocation, potentially triggering an outbound HTTP fetch to the JWKS endpoint each time. The Nimbus `JWKSource` already handles key caching and automatic rotation internally, so a single instance should be reused.
- **Incorrect log level**: Expired or invalid client tokens are not server errors. Logging them at `ERROR` pollutes error-level monitoring/alerting for issues the server cannot fix. `WARN` is the appropriate level.

No. Validation behavior is unchanged. Log level changes from ERROR to WARN for client-side token failures.

Existing unit tests in `TestJwksTokenValidator` cover the validation paths. The change is a straightforward refactoring of when the `JWKSource` is constructed (startup vs per-call) with no behavioral change.

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
@@ -63,6 +63,7 @@ public class JwksTokenValidator implements OAuthTokenValidator {
   private long allowSkewSeconds;
   private PrincipalMapper principalMapper;
   private GroupMapper groupMapper;
+  private JWKSource<SecurityContext> jwkSource;
 
   @Override
   public void initialize(Config config) {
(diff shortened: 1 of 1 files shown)


server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
```diff
@@ -63,6 +63,7 @@ public class JwksTokenValidator implements OAuthTokenValidator {
   private long allowSkewSeconds;
   private PrincipalMapper principalMapper;
   private GroupMapper groupMapper;
+  private JWKSource<SecurityContext> jwkSource;
```
roryqi, line 66: How long will the cache entry expire?

roryqi:
If this is a fix, we should have an issue to track.

server-common/src/main/java/org/apache/gravitino/server/authentication/JwksTokenValidator.java
bharos: The Nimbus JWKSourceBuilder.create(url).build() enables caching with refresh-ahead by default:

Cache TTL: 5 minutes (DEFAULT_CACHE_TIME_TO_LIVE)
Refresh-ahead: 30 seconds before expiration, on a background thread
Rate limiting: min 30 seconds between URL fetches (safety net for key rotation / unknown key IDs)
Under normal operation the JWKS URL is only hit once at startup; the background refresh keeps the cache warm. The rate limiter never blocks — it returns the last-known keys instead of re-fetching, so validateToken() always proceeds without waiting on I/O.

Added a code comment documenting these defaults. 

https://www.javadoc.io/doc/com.nimbusds/nimbus-jose-jwt/latest/com/nimbusds/jose/jwk/source/JWKSourceBuilder.html#DEFAULT_CACHE_TIME_TO_LIVE

roryqi: Got it.

roryqi (approved):
LGTM.

PR #10726, opened 2026-04-09
[#10725] feat(catalog-glue): Add GlueClientProvider and configuration properties metadata

### LINKED ISSUE #10725: [Subtask] Add GlueClientProvider and configuration properties metadata
### Describe the subtask

Add `GlueClientProvider` (AWS SDK v2 `GlueClient` factory with fail-fast credential validation) and configuration properties metadata classes for the `catalog-glue` module:

- `GlueClientProvider` — builds a `GlueClient` from catalog properties; validates that access key and secret key are either both present or both absent; supports optional endpoint override for LocalStack testing
- `GlueCatalogPropertiesMetadata` — declares all catalog-level properties (`aws-region`, `aws-glue-catalog-id` as required+immutable; credentials as optional+hidden; endpoint, default table format, table type filter as optional)
- `GlueSchemaPropertiesMetadata` / `GlueTablePropertiesMetadata` — stub metadata for schema and table properties
- `GlueCatalogCapability` — declares Glue constraints: no NOT NULL, no DEFAULT values, case-insensitive schema/table names

### Parent issue

diqiu50:

Add GlueClientProvider, configuration properties metadata, and capability declarations for the catalog-glue module.

AWS connection layer and property definitions are prerequisites for Schema/Table CRUD PRs.

Fix: #10725

No.

`./gradlew :catalogs:catalog-glue:test -PskipITs`

### DIFF
--- a/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogPropertiesMetadata.java
+++ b/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogPropertiesMetadata.java
@@ -25,8 +25,8 @@ import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_REGION;
 import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_SECRET_ACCESS_KEY;
 import static org.apache.gravitino.catalog.glue.GlueConstants.DEFAULT_TABLE_FORMAT;
 import static org.apache.gravitino.catalog.glue.GlueConstants.DEFAULT_TABLE_FORMAT_VALUE;
+import static org.apache.gravitino.catalog.glue.GlueConstants.DEFAULT_TABLE_TYPE_FILTER;
 import static org.apache.gravitino.catalog.glue.GlueConstants.TABLE_TYPE_FILTER;
-import static org.apache.gravitino.catalog.glue.GlueConstants.TABLE_TYPE_FILTER_ALL;
 import static org.apache.gravitino.connector.PropertyEntry.stringOptionalPropertyEntry;
 import static org.apache.gravitino.connector.PropertyEntry.stringRequiredPropertyEntry;
 
@@ -87,8 +87,7 @@ public class GlueCatalogPropertiesMetadata extends BaseCatalogPropertiesMetadata
               DEFAULT_TABLE_FORMAT,
               stringOptionalPropertyEntry(
                   DEFAULT_TABLE_FORMAT,
-                  "Default format for tables created via createTable(). Accepted: iceberg, hive."
-                      + " Unrecognised values are rejected at createTable() time.",
+                  "Default format for tables created via createTable(). Accepted: iceberg, hive.",
                   false /* immutable */,
                   DEFAULT_TABLE_FORMAT_VALUE,
                   false /* hidden */))
@@ -97,10 +96,9 @@ public class GlueCatalogPropertiesMetadata extends BaseCatalogPropertiesMetadata
               stringOptionalPropertyEntry(
                   TABLE_TYPE_FILTER,
                   "Comma-separated table types exposed by listTables() and loadTable()."
-                      + " Accepted: all, hive, iceberg, delta, parquet."
-                      + " Unrecognised values are rejected at listTables() time.",
+                      + " Accepted: all, hive, iceberg.",
                   false /* immutable */,
-                  TABLE_TYPE_FILTER_ALL,
+                  DEFAULT_TABLE_TYPE_FILTER,
                   false /* hidden */))
           .build();
 
--- a/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogPropertiesMetadata.java
+++ b/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogPropertiesMetadata.java
@@ -18,20 +18,87 @@
  */
 package org.apache.gravitino.catalog.glue;
 
+import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_ACCESS_KEY_ID;
+import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_GLUE_CATALOG_ID;
+import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_GLUE_ENDPOINT;
+import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_REGION;
+import static org.apache.gravitino.catalog.glue.GlueConstants.AWS_SECRET_ACCESS_KEY;
+import static org.apache.gravitino.catalog.glue.GlueConstants.DEFAULT_TABLE_FORMAT;
+import static org.apache.gravitino.catalog.glue.GlueConstants.DEFAULT_TABLE_FORMAT_VALUE;
+import static org.apache.gravitino.catalog.glue.GlueConstants.TABLE_TYPE_FILTER;
+import static org.apache.gravitino.catalog.glue.GlueConstants.TABLE_TYPE_FILTER_ALL;
+import static org.apache.gravitino.connector.PropertyEntry.stringOptionalPropertyEntry;
+import static org.apache.gravitino.connector.PropertyEntry.stringRequiredPropertyEntry;
+
 import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import org.apache.gravitino.connector.BaseCatalogPropertiesMetadata;
 import org.apache.gravitino.connector.PropertyEntry;
 
-/**
- * Properties metadata for the AWS Glue Data Catalog connector.
- *
- * <p>TODO PR-02: add required properties (aws-region, aws-glue-catalog-id) and optional properties
- * (credentials, endpoint override, default-table-format, table-type-filter).
- */
+/** Properties metadata for the AWS Glue Data Catalog connector catalog-level configuration. */
 public class GlueCatalogPropertiesMetadata extends BaseCatalogPropertiesMetadata {
 
-  private static final Map<String, PropertyEntry<?>> PROPERTIES_METADATA = ImmutableMap.of();
+  private static final Map<String, PropertyEntry<?>> PROPERTIES_METADATA =
+      ImmutableMap.<String, PropertyEntry<?>>builder()
+          .put(
+              AWS_REGION,
+              stringRequiredPropertyEntry(
+                  AWS_REGION,
+                  "AWS region for the Glue Data Catalog (e.g. us-east-1)",
+                  true /* immutable */,
+                  false /* hidden */))
+          .put(
+              AWS_GLUE_CATALOG_ID,
+              stringRequiredPropertyEntry(
+                  AWS_GLUE_CATALOG_ID,
+                  "The 12-digit AWS account ID that owns the Glue catalog",
+                  true /* immutable */,
+                  false /* hidden */))
+          .put(
+              AWS_ACCESS_KEY_ID,
+              stringOptionalPropertyEntry(
+                  AWS_ACCESS_KEY_ID,
+                  "AWS access key ID for static credential authentication."
+                      + " When omitted the default credential chain is used.",
+                  false /* immutable */,
+                  null /* defaultValue */,
+                  true /* hidden */))
+          .put(
+              AWS_SECRET_ACCESS_KEY,
+              stringOptionalPropertyEntry(
+                  AWS_SECRET_ACCESS_KEY,
+                  "AWS secret access key paired with aws-access-key-id."
+                      + " When omitted the default credential chain is used.",
+                  false /* immutable */,
+                  null /* defaultValue */,
+                  true /* hidden */))
+          .put(
+              AWS_GLUE_ENDPOINT,
+              stringOptionalPropertyEntry(
+                  AWS_GLUE_ENDPOINT,
+                  "Custom Glue endpoint URL for VPC endpoints or LocalStack testing"
+                      + " (e.g. http://localhost:4566)",
+                  false /* immutable */,
+                  null /* defaultValue */,
+                  false /* hidden */))
+          .put(
+              DEFAULT_TABLE_FORMAT,
+              stringOptionalPropertyEntry(
+                  DEFAULT_TABLE_FORMAT,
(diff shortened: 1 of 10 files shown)


catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,104 @@
...
+   *     credential keys is provided, or if {@code aws-glue-endpoint} is not a valid URI.
+   */
+  public static GlueClient buildClient(Map<String, String> config) {
+    String region = config.get(GlueConstants.AWS_REGION);
+    Preconditions.checkArgument(
+        !Strings.isNullOrEmpty(region),
+        "Property '%s' is required to create a Glue client",
+        GlueConstants.AWS_REGION);
+
+    GlueClientBuilder builder = GlueClient.builder().region(Region.of(region));
```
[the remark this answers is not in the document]
diqiu50: This is not an issue — passing an empty string will cause an error

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @param config Catalog configuration properties.
+   * @return A configured and ready-to-use {@link GlueClient}.
+   * @throws IllegalArgumentException if {@code aws-region} is missing or blank, if only one of the
+   *     credential keys is provided, or if {@code aws-glue-endpoint} is not a valid URI.
+   */
+  public static GlueClient buildClient(Map<String, String> config) {
+    String region = config.get(GlueConstants.AWS_REGION);
+    Preconditions.checkArgument(
+        !Strings.isNullOrEmpty(region),
```
yuqi1129: **Bug: blank region not caught.**

`Strings.isNullOrEmpty("   ")` returns `false` — blank strings pass through to `Region.of()`, which may throw an opaque SDK error instead of the helpful message defined here. The Javadoc already says "missing **or blank**", so the guard should match:

```java
!Strings.isNullOrEmpty(region) && !region.isBlank()
```

`testBuildClientBlankRegionThrows` technically passes today because the AWS SDK's `Validate.paramNotBlank` fires, but the error message is not the one we defined.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,104 @@
...
+        "Property '%s' is required to create a Glue client",
+        GlueConstants.AWS_REGION);
+
+    GlueClientBuilder builder = GlueClient.builder().region(Region.of(region));
+
+    // Static credentials take priority over the default credential chain.
+    // Both keys must be provided together — a partial pair is always a misconfiguration.
+    String accessKey = config.get(GlueConstants.AWS_ACCESS_KEY_ID);
+    String secretKey = config.get(GlueConstants.AWS_SECRET_ACCESS_KEY);
+    boolean hasAccessKey = !Strings.isNullOrEmpty(accessKey);
```
yuqi1129: **Same blank issue for credentials.**

If a user provides `aws-access-key-id = "   "` (all spaces), `isNullOrEmpty` returns `false`, so `hasAccessKey = true`. If the secret key is absent, the pair-check fires correctly. But if *both* are blank strings, `hasAccessKey == hasSecretKey == true`, and `AwsBasicCredentials.create("   ", "   ")` is called — which silently succeeds at build time and fails with a cryptic auth error at the first API call.

Suggest:
```java
boolean hasAccessKey = !Strings.isNullOrEmpty(accessKey) && !accessKey.isBlank();
boolean hasSecretKey = !Strings.isNullOrEmpty(secretKey) && !secretKey.isBlank();
```

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogPropertiesMetadata.java
```diff
@@ -18,20 +18,87 @@
...
+          .put(
+              AWS_REGION,
+              stringRequiredPropertyEntry(
+                  AWS_REGION,
+                  "AWS region for the Glue Data Catalog (e.g. us-east-1)",
+                  true /* immutable */,
+                  false /* hidden */))
+          .put(
+              AWS_GLUE_CATALOG_ID,
+              stringRequiredPropertyEntry(
```
yuqi1129: **Consider making `aws-glue-catalog-id` optional.**

When omitted, Glue defaults to the 12-digit account ID of the caller's credentials. Marking it as `required` forces every user to look up and supply their account ID even when the default is correct. Either:
- Make it optional (with a description note: "When omitted, defaults to the caller's AWS account ID"), or
- Keep it required but explicitly explain in the description why the default is insufficient for Gravitino's use case (e.g. cross-account scenarios).

If there's a specific reason Gravitino always needs the explicit ID, a comment here would prevent future confusion.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogPropertiesMetadata.java
```diff
@@ -18,20 +18,87 @@
...
+                  "Custom Glue endpoint URL for VPC endpoints or LocalStack testing"
+                      + " (e.g. http://localhost:4566)",
+                  false /* immutable */,
+                  null /* defaultValue */,
+                  false /* hidden */))
+          .put(
+              DEFAULT_TABLE_FORMAT,
+              stringOptionalPropertyEntry(
+                  DEFAULT_TABLE_FORMAT,
+                  "Default format for tables created via createTable(). Accepted: iceberg, hive.",
```
yuqi1129, line 90: **No validation of accepted values for `default-table-format`.**

The description says "Accepted: iceberg, hive" but an invalid value like `"parquet"` will be silently stored and only fail (with a confusing error) at `createTable()` time. Same issue applies to `table-type-filter` on line 96.

Suggest adding validation in the catalog's `initialize()` or in `GlueClientProvider`, or at minimum calling out in the description that unrecognised values are deferred to runtime.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueConstants.java
```diff
@@ -0,0 +1,79 @@
...
+  /** Default value for {@link #TABLE_TYPE_FILTER}: expose all table types. */
+  public static final String TABLE_TYPE_FILTER_ALL = "all";
+
+  // -------------------------------------------------------------------------
+  // Glue Table.parameters() keys (passthrough properties)
+  // -------------------------------------------------------------------------
+
+  /**
+   * Glue table format type parameter key stored in {@code Table.parameters()}. Common values:
+   * {@code ICEBERG}, {@code HIVE}, {@code DELTA}, {@code PARQUET}.
```
yuqi1129: Minor nit: the values listed here (`ICEBERG`, `HIVE`, `DELTA`, `PARQUET` — uppercase) are Glue `Table.parameters()` values, while `TABLE_TYPE_FILTER`'s accepted values (line 57) are lowercase (`hive`, `iceberg`, ...). They're different namespaces so the casing difference is correct, but a brief note like "Note: these are stored uppercase in Glue parameters; see `TABLE_TYPE_FILTER` for the Gravitino-side filter values" would prevent confusion.

yuqi1129 (commented):
Overall the PR is well-structured with good test coverage. A few issues below.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,104 @@
...
+      builder.credentialsProvider(DefaultCredentialsProvider.create());
+    }
+
+    // Optional custom endpoint override for VPC endpoints or LocalStack testing.
+    String endpoint = config.get(GlueConstants.AWS_GLUE_ENDPOINT);
+    if (StringUtils.isNotBlank(endpoint)) {
+      try {
+        builder.endpointOverride(URI.create(endpoint));
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
```
yuqi1129: It's not so elegant to catch `IllegalArgumentException`  and then throw an exception with the same type.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,104 @@
...
+            + "Either provide both keys for static authentication, "
+            + "or omit both to use the default credential chain.",
+        GlueConstants.AWS_ACCESS_KEY_ID,
+        GlueConstants.AWS_SECRET_ACCESS_KEY);
+
+    if (hasAccessKey) {
+      builder.credentialsProvider(
+          StaticCredentialsProvider.create(AwsBasicCredentials.create(accessKey, secretKey)));
+    } else {
+      builder.credentialsProvider(DefaultCredentialsProvider.create());
```
yuqi1129, line 91: What's the value of default credential chain?

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueConstants.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Comma-separated list of table types exposed by {@code listTables()} and {@code loadTable()}
+   * (optional). Accepted values: {@code all}, {@code hive}, {@code iceberg}, {@code delta}, {@code
+   * parquet}. Defaults to {@code all}.
+   */
+  public static final String TABLE_TYPE_FILTER = "table-type-filter";
+
+  /** Default value for {@link #TABLE_TYPE_FILTER}: expose all table types. */
+  public static final String TABLE_TYPE_FILTER_ALL = "all";
```
yuqi1129: Rename to `DEFAULT_TABLE_TYPE_FILTER`?

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueTablePropertiesMetadata.java
```diff
@@ -26,12 +30,36 @@
...
 
-  private static final Map<String, PropertyEntry<?>> PROPERTIES_METADATA = ImmutableMap.of();
+  private static final Map<String, PropertyEntry<?>> PROPERTIES_METADATA =
+      ImmutableMap.<String, PropertyEntry<?>>builder()
+          .put(
+              TABLE_FORMAT_TYPE,
+              stringOptionalPropertyEntry(
+                  TABLE_FORMAT_TYPE,
+                  "Glue table format type stored in Table.parameters(). Common values:"
+                      + " ICEBERG, HIVE, DELTA, PARQUET.",
```
yuqi1129: Do we only support those four types?

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
diqiu50: Add some comments to describe it

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueTablePropertiesMetadata.java
diqiu50: We only support the `TABLE_FORMAT_TYPE` of ICEBERG, HIVE first

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueConstants.java
diqiu50: fixed

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,102 @@
...
+
+    // Static credentials take priority over the default credential chain.
+    // Both keys must be provided together — a partial pair is always a misconfiguration.
+    // Default credential chain order (when both keys are omitted):
+    //   1. Java system properties (aws.accessKeyId / aws.secretAccessKey)
+    //   2. Environment variables (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY)
+    //   3. Web identity token (EKS / Kubernetes)
+    //   4. ~/.aws/credentials profile file
+    //   5. ECS container task role
+    //   6. EC2 instance profile (IMDSv2)
```
yuqi1129, line 74: What if the AKSK is not provided by any of them?

diqiu50: An exception throw on the aws-sdk

yuqi1129 (approved):

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueClientProvider.java
```diff
@@ -0,0 +1,102 @@
...
+      builder.credentialsProvider(
+          StaticCredentialsProvider.create(AwsBasicCredentials.create(accessKey, secretKey)));
+    } else {
+      builder.credentialsProvider(DefaultCredentialsProvider.create());
+    }
+
+    // Optional custom endpoint override for VPC endpoints or LocalStack testing.
+    String endpoint = config.get(GlueConstants.AWS_GLUE_ENDPOINT);
+    if (StringUtils.isNotBlank(endpoint)) {
+      builder.endpointOverride(URI.create(endpoint));
```
yuqi1129, line 97: `URI.create(endpoint)` will throw `IllegalArgumentException` with a generic message and no property context. Could we catch it and rethrow with explicit config key (for example `aws-glue-endpoint`) and the bad value? That makes misconfiguration diagnosis much easier and matches the fail-fast style used for other properties.

PR #10738, opened 2026-04-10
[#10737] fix(core): Avoid blocking dropCatalog on imported schemas

### LINKED ISSUE #10737: [Bug report] DropCatalog misclassifies imported schemas
### Version

main branch

### Describe what's wrong

`dropCatalog(force=false)` can incorrectly fail with `NonEmptyCatalogException` when the remaining schema was imported from the underlying catalog instead of being created through Gravitino.

For example, a Hive catalog may already contain the `default` schema. After metadata synchronization imports it into the entity store, `dropCatalog(force=false)` may still treat it as a user-created schema and block catalog deletion.

```text
NonEmptyCatalogException: Catalog <metalake>.<catalog> has schemas, please drop them first or use force option
```

### How to reproduce

1. Use `main branch`.
2. Create a catalog backed by an external system that already contains schemas.
3. Trigger schema import / metadata synchronization.
4. Call `dropCatalog(<catalog>, false)`.
5. Observe that the drop is rejected even though the schema was imported from the external catalog.

### Additional context

The issue is in the schema classification logic used by `CatalogManager.dropCatalog(...)`.

A fix should:
- not treat imported schemas as user-created
- still block drop for schemas created through Gravitino
- fail conservatively when schema classification fails unexpectedly

diqiu50:

Fix schema classification in `dropCatalog(force = false)` so imported schemas do not block catalog deletion.

Imported schemas can be written into the entity store during metadata synchronization and later be misclassified as user-created schemas.

That makes `dropCatalog(force = false)` fail with `NonEmptyCatalogException` even though the remaining schema was imported from the external catalog.

Fix: #10737

`dropCatalog(force = false)` no longer incorrectly fails when only imported schemas remain.

- added unit coverage in `TestCatalogManager`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
@@ -73,6 +73,7 @@ import org.apache.gravitino.EntityAlreadyExistsException;
 import org.apache.gravitino.EntityStore;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.Namespace;
+import org.apache.gravitino.Schema;
 import org.apache.gravitino.StringIdentifier;
 import org.apache.gravitino.connector.BaseCatalog;
 import org.apache.gravitino.connector.CatalogOperations;
@@ -87,6 +88,7 @@ import org.apache.gravitino.exceptions.GravitinoRuntimeException;
 import org.apache.gravitino.exceptions.NoSuchCatalogException;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
 import org.apache.gravitino.exceptions.NoSuchMetalakeException;
+import org.apache.gravitino.exceptions.NoSuchSchemaException;
 import org.apache.gravitino.exceptions.NonEmptyCatalogException;
 import org.apache.gravitino.exceptions.NonEmptyEntityException;
 import org.apache.gravitino.file.FilesetCatalog;
@@ -877,9 +879,39 @@ public class CatalogManager implements CatalogDispatcher, Closeable {
     Set<String> availableSchemaNames =
         Arrays.stream(allSchemas).map(NameIdentifier::name).collect(Collectors.toSet());
 
-    // some schemas are dropped externally, but still exist in the entity store, those schemas are
-    // invalid
-    return schemaEntities.stream().map(SchemaEntity::name).anyMatch(availableSchemaNames::contains);
+    // Some schemas are dropped externally, but still exist in the entity store — those are invalid.
+    // Among schemas that still exist in the underlying catalog, only count those created via
+    // Gravitino. New schemas carry an entity-store marker; older schemas fall back to the embedded
+    // StringIdentifier in external catalog metadata.
+    return schemaEntities.stream()
+        .filter(e -> availableSchemaNames.contains(e.name()))
+        .anyMatch(
+            e -> {
+              Map<String, String> entityProps = e.properties();
+              if (entityProps != null
+                  && "true"
+                      .equals(entityProps.get(SchemaOperationDispatcher.SCHEMA_CREATED_BY_GRAVITINO))) {
+                return true;
+              }
+
+              try {
+                Schema schema =
+                    catalogWrapper.doWithSchemaOps(ops -> ops.loadSchema(e.nameIdentifier()));
+                return StringIdentifier.fromProperties(schema.properties()) != null;
+              } catch (NoSuchSchemaException ex) {
+                LOG.warn(
+                    "Schema {} no longer exists while checking whether it is user-created",
+                    e.nameIdentifier(),
+                    ex);
+                return false;
+              } catch (Exception ex) {
+                throw new RuntimeException(
+                    String.format(
+                        "Failed to determine whether schema %s is user-created",
+                        e.nameIdentifier()),
+                    ex);
+              }
+            });
   }
 
   /**
--- a/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
@@ -879,20 +879,12 @@ public class CatalogManager implements CatalogDispatcher, Closeable {
     Set<String> availableSchemaNames =
         Arrays.stream(allSchemas).map(NameIdentifier::name).collect(Collectors.toSet());
 
-    // Some schemas are dropped externally, but still exist in the entity store — those are invalid.
-    // Among schemas that still exist in the underlying catalog, only count those created via
-    // Gravitino. New schemas carry an entity-store marker; older schemas fall back to the embedded
-    // StringIdentifier in external catalog metadata.
+    // Some schemas are dropped externally but still exist in the entity store — those are invalid.
+    // Among schemas that exist in the underlying catalog, only those created via Gravitino carry a
+    // StringIdentifier in their external properties; imported schemas do not.
     for (SchemaEntity schemaEntity : schemaEntities) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -877,9 +879,37 @@ private boolean containsUserCreatedSchemas(
...
+    for (SchemaEntity schemaEntity : schemaEntities) {
+      if (!availableSchemaNames.contains(schemaEntity.name())) {
+        continue;
+      }
+
+      Map<String, String> entityProps = schemaEntity.properties();
+      if (entityProps != null
+          && SchemaOperationDispatcher.SCHEMA_CREATED_BY_GRAVITINO_VALUE.equals(
+              entityProps.get(SchemaOperationDispatcher.SCHEMA_CREATED_BY_GRAVITINO))) {
+        return true;
```
yuqi1129: Since `SCHEMA_CREATED_BY_GRAVITINO_VALUE` is introduced by this version only, can this apply to older version?

core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
```diff
@@ -148,6 +152,9 @@ public Schema createSchema(NameIdentifier ident, String comment, Map<String, Str
                   .withId(uid)
                   .withName(ident.name())
                   .withNamespace(ident.namespace())
+                  .withProperties(
+                      ImmutableMap.of(
+                          SCHEMA_CREATED_BY_GRAVITINO, SCHEMA_CREATED_BY_GRAVITINO_VALUE))
```
yuqi1129: By now, the Gravitino entity store will not store properties,so we may need to think twice about this point.

jerryshao:
Can you explain more about how you fixed this issue, I'm a little confused.

diqiu50:
The root cause is that `dropCatalog(force=false)` checks whether any schemas exist in the entity store, but imported schemas (synced from external catalogs via search listener load) are also written to the entity store — so they were mistakenly treated as user-created schemas, causing a false `NonEmptyCatalogException`.

The fix distinguishes user-created schemas from imported ones: a schema is considered user-created only if it carries a `StringIdentifier` in its properties (written by Gravitino when the user creates it via API) or is marked with `gravitino.created=true` in the entity store.

However, as @yuqi1129 pointed out, the entity store does not persist properties, so the marker approach needs to be rethought.

core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
diqiu50: Good catch. We've reworked the approach to avoid storing anything in the entity store. Instead, `containsUserCreatedSchemas` now calls `loadSchema` for each schema and checks `StringIdentifier.fromProperties(schema.properties())` — schemas created via Gravitino carry the identifier, imported ones do not.

One edge case: backends that cannot store a `StringIdentifier` (e.g., MySQL) return null properties, making it impossible to distinguish user-created from imported schemas. In that case, `dropCatalog(force=false)` has no meaningful semantic, so we suggest logging a warning instead of throwing `NonEmptyCatalogException`. WDYT?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
diqiu50: This is not a problem here, since schemas are not deleted.

core/src/test/java/org/apache/gravitino/catalog/TestCatalogManager.java
```diff
@@ -589,6 +591,151 @@ public void testDropCatalog() throws Exception {
...
+    Mockito.doReturn(capability).when(wrapper).capabilities();
+    Mockito.doReturn(unsupportedResult).when(capability).managedStorage(any());
+    Mockito.doReturn(new NameIdentifier[] {NameIdentifier.of("metalake", "test41", "default")})
+        .doThrow(new NoSuchSchemaException("Schema not found"))
+        .when(wrapper)
+        .doWithSchemaOps(any());
+
+    // Schema disappearing between listSchemas and loadSchema should not block drop.
+    Assertions.assertTrue(catalogManager.dropCatalog(ident));
+  }
```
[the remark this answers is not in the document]
diqiu50: fixed

core/src/test/java/org/apache/gravitino/catalog/TestCatalogManager.java
```diff
@@ -589,6 +591,151 @@ public void testDropCatalog() throws Exception {
...
+            "value3");
+    String comment = "comment";
+
+    Catalog catalog =
+        catalogManager.createCatalog(ident, Catalog.Type.RELATIONAL, provider, comment, props);
+    Mockito.doCallRealMethod().when(catalogManager).loadCatalogAndWrap(ident);
+    Assertions.assertDoesNotThrow(() -> catalogManager.disableCatalog(ident));
+    CatalogEntity catalogEntity = entityStore.get(ident, EntityType.CATALOG, CatalogEntity.class);
+    FieldUtils.writeField(catalog, "entity", catalogEntity, true);
+
```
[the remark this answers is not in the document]
diqiu50: Not a issue

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -877,9 +879,37 @@ private boolean containsUserCreatedSchemas(
...
+            catalogWrapper.doWithSchemaOps(ops -> ops.loadSchema(schemaEntity.nameIdentifier()));
+        Map<String, String> props = schema.properties();
+        // If the backend cannot store a StringIdentifier (null or empty properties, e.g. MySQL
+        // which does not support schema comments), we cannot tell whether the schema was created
+        // by Gravitino or imported. Be conservative and treat it as user-created to avoid
+        // accidental data loss.
+        // Only skip a schema when properties are non-null, non-empty, and contain no
+        // StringIdentifier — the reliable signal that the schema was imported from an external
+        // catalog on a backend that does support identifier storage.
+        if (props == null || props.isEmpty() || StringIdentifier.fromProperties(props) != null) {
```
yuqi1129, line 901: Do schemas loaded by  `catalogWrapper.doWithSchemaOps(ops -> ops.loadSchema(schemaEntity.nameIdentifier()))` really contain `StringIdentifier`? I remember that it will return schema in the external storage.

diqiu50: The `StringIdentifier` is indeed retrievable from `loadSchema` for most catalogs, but through different mechanisms:

- **Hive**: stored directly in HMS database properties at `createSchema` time, returned as-is by `loadSchema`.
- **Iceberg / Paimon**: stored in namespace properties, returned as-is.
- **JDBC (PostgreSQL, Doris — `supportSchemaComment=true`)**: embedded into the schema comment via `StringIdentifier.addToComment` at create time. `JdbcCatalogOperations.loadSchema` extracts it from the comment and reconstructs it into properties via `newPropertiesWithId`, so `fromProperties()` works correctly.
- **JDBC (MySQL, StarRocks — `supportSchemaComment=false`)**: cannot store the identifier at all. `loadSchema` returns null/empty properties — the `props == null || props.isEmpty()` guard on line 901 catches this and conservatively returns `true` (treat as user-created) to avoid accidental data loss.

So for any catalog that *can* persist a `StringIdentifier`, `loadSchema` will return it. For those that cannot, we fall back to the safe/conservative path.

yuqi1129 (approved):

PR #10740, opened 2026-04-10
[#10739] fix(core): Avoid stale cache refill during concurrent metadata updates

### LINKED ISSUE #10739: [Bug report] Cache invalidation can expose stale metadata under concurrent access
### Version

main branch

### Describe what's wrong

Several cache invalidation paths invalidate cache entries before the backend write is finished.

This creates a race window under concurrent access:

1. a write operation starts, such as `alterCatalog`, `dropCatalog`, `update`, `delete`, `insertRelation`, or `updateEntityRelations`
2. the cache entry is invalidated first
3. another thread reads the same object while the backend mutation has not completed yet
4. that read misses the cache, fetches old data from the backend, and writes the stale result back into cache

After that, the cache may continue serving outdated metadata even though the write has already succeeded.

This can cause incorrect behavior such as:
- loading a catalog by its old name after rename
- catalog existence checks still returning true after drop
- stale relation results after relation insert or update

This is a race-condition issue and does not always produce a stable stacktrace.

Typical symptoms are stale reads after successful writes, for example:
- the old catalog identifier is still loadable after rename
- a dropped catalog is still reported as existing
- relation queries return outdated results after relation updates

### How to reproduce

1. Use `main branch`.
2. Trigger a write operation that updates cached metadata, such as catalog rename, catalog drop, relation insert, or relation update.
3. At the same time, trigger another thread to read the same catalog or relation data.
4. Observe that the read can repopulate the cache with pre-update data before the backend mutation finishes.
5. After the write completes, stale data may still be returned from cache.

### Additional context

The issue is caused by invalidating cache before backend mutation in several catalog and relational entity store code paths.

A fix should move cache invalidation to after the backend mutation succeeds, so concurrent reads either see the still-valid old cache before commit or reload fresh data after commit.

diqiu50:

Reorder cache invalidation in catalog and relational entity store write paths so concurrent reads do not refill cache with stale metadata.

Several write paths invalidate cache before the backend mutation completes.

Under concurrent access, another thread can miss the cache, read old data from the backend, and write that stale result back into cache. This can leave outdated catalog or relation metadata visible after rename, drop, or relation updates.

Fix: #10739

No.

- added unit coverage in `TestCatalogManager`
- added unit coverage in `TestRelationalEntityStore`


core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -806,8 +809,11 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
             }
 
             // Finally, delete the catalog entity as well as all its sub-entities from the store.
+            // Invalidate after store.delete() to prevent a background thread from repopulating
+            // the cache with stale data between invalidate and delete.
+            boolean deleted = store.delete(ident, EntityType.CATALOG, true);
             catalogCache.invalidate(ident);
```
yuqi1129, line 815: What if the delete operation in L814 succeeds and L815 fails?

diqiu50: If this case occurs, there may be a bug in `catalogCache` or an issue with the `JVM`, and many operations could fail.
Considering this issue here is too complex. I suggest restarting.

yuqi1129: This seems to break the rule that we should clean the cache before modifying the database, and this policy is quite common in the Redis & MySQL pair, so I'm not very sure whether this change is acceptable or not. 

@mchades Would you like to take a look

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -183,8 +184,9 @@ public <E extends Entity & HasIdentifier> List<E> batchGet(
   public boolean delete(NameIdentifier ident, Entity.EntityType entityType, boolean cascade)
       throws IOException {
     try {
+      boolean deleted = backend.delete(ident, entityType, cascade);
       cache.invalidate(ident, entityType);
-      return backend.delete(ident, entityType, cascade);
+      return deleted;
     } catch (NoSuchEntityException e) {
```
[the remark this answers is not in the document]
mchades: seems you don't resolve this comment?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
mchades: I think the ordering change here is intentional and is the right tradeoff for this PR.

If we invalidate the cache before `store.delete(...)`, we open a window where a concurrent reader can see a cache miss, read the still-existing catalog from the store, and populate the cache again with stale data before the delete completes. That is the race this patch is trying to fix.

The `delete succeeds but invalidate fails` case is theoretically possible, but it is a different partial-failure problem. For cache-aside, the more common write path is to mutate the source of truth first and then invalidate the cache, precisely to avoid stale refills during concurrent reads. So I would not use that scenario as a reason to move the invalidation back before the store mutation.

So my take is that this PR is fixing the more immediate concurrency bug, while the post-delete invalidate failure case would be a separate follow-up discussion if we want to harden it further.

yuqi1129: Yeah, the cache refill problem weighs.

yuqi1129 (approved):

mchades:
@diqiu50 plz resolve the comment https://github.com/apache/gravitino/pull/10740#discussion_r3196070124

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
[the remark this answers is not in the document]
diqiu50: Always invalidate the cache in finally block

diqiu50:
> @diqiu50 plz resolve the comment [#10740 (comment)](https://github.com/apache/gravitino/pull/10740#discussion_r3196070124)

fixed

mchades (approved):

PR #10751, opened 2026-04-12
[#10683] fix(iceberg-rest-server): support Idempotency-Key header and replay for mutation endpoints

### LINKED ISSUE #10683: [Improvement] Support Idempotency-Key header in Iceberg REST server

The Gravitino Iceberg REST server does not recognize or handle the `Idempotency-Key` header on any mutation endpoint. The Iceberg REST spec defines this as an optional UUIDv7 header on all create, update, and delete operations. When present, the server should return the original response for replayed keys instead of re-executing the operation. The spec also defines an `idempotency-key-lifetime` field in the `/config` response to advertise support and the key reuse window. Currently there are zero references to `Idempotency-Key` anywhere in the Iceberg REST server codebase, meaning clients cannot safely retry failed mutation requests without risking duplicate side effects.

### How should we improve?

- Accept the `Idempotency-Key` header (UUIDv7) on all mutation endpoints in `IcebergTableOperations`, `IcebergNamespaceOperations`, and `IcebergViewOperations`
- Store completed responses keyed by the idempotency key with a configurable TTL
- Return the stored response on duplicate keys instead of re-executing the operation
- Advertise `idempotency-key-lifetime` (e.g., `PT30M`) in the `IcebergConfigOperations` config response
- Add unit tests for idempotent retry behavior

laserninja:

Add support for the `Idempotency-Key` HTTP header on all mutation (POST/DELETE) endpoints in the Iceberg REST server. When a client sends a valid UUIDv7 idempotency key with a mutation request, the server caches the successful response and replays it on subsequent retries, preventing duplicate side-effects.

**Key implementation details:**

- New `IcebergIdempotencyManager` class with:
  - UUIDv7 format validation for the header value
  - Caffeine-based response cache with configurable TTL (default 30 min, max 10K entries)
  - Thread-safe `replayOrExecute()` using `ConcurrentHashMap` key locks with double-check pattern
  - Cache key includes idempotency key + HTTP method + URI + normalized query string
- Added `@HeaderParam("Idempotency-Key")` to all mutation endpoints:
  - **Tables**: `createTable`, `updateTable`, `dropTable`
  - **Namespaces**: `createNamespace`, `updateNamespace`, `dropNamespace`, `registerTable`
  - **Views**: `createView`, `replaceView`, `dropView`
- Advertises `idempotency-key-lifetime` (ISO-8601 duration) in `GET /v1/config` defaults
- New config entry `idempotency-key-lifetime-minutes` (default: 30)

Fixes #10683

Network failures and client retries against the Iceberg REST catalog can cause duplicate table/namespace/view creation or deletion. The Iceberg REST spec recommends idempotency key support for mutation endpoints to safely handle retries.

Yes:
- Clients can now send an `Idempotency-Key: <UUIDv7>` header on mutation requests for safe retry
- `GET /v1/config` response now includes `idempotency-key-lifetime` in defaults
- New server config: `idempotency-key-lifetime-minutes` (default 30)

- Unit tests added for all three operation classes:
  - `TestIcebergTableOperations`: `testCreateTableWithIdempotencyKey`, `testCreateTableRejectsInvalidIdempotencyKey`
  - `TestIcebergNamespaceOperations`: `testRegisterTableWithIdempotencyKey`
  - `TestIcebergViewOperations`: `testCreateViewWithIdempotencyKey`
- `TestIcebergConfig` updated to verify `idempotency-key-lifetime` in config response
- All 100 tests across 4 suites pass locally (JDK 17)
- Tests verify replay returns cached response on second call with same key
- Tests verify events fire exactly once (no duplicate side-effects)
- Tests verify invalid (non-UUIDv7) keys return 400 Bad Request

### DIFF
--- a/catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergConstants.java
+++ b/catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergConstants.java
@@ -110,4 +110,6 @@ public class IcebergConstants {
   public static final String SCAN_PLAN_CACHE_IMPL = "scan-plan-cache-impl";
   public static final String SCAN_PLAN_CACHE_CAPACITY = "scan-plan-cache-capacity";
   public static final String SCAN_PLAN_CACHE_EXPIRE_MINUTES = "scan-plan-cache-expire-minutes";
+
+  public static final String IDEMPOTENCY_KEY_LIFETIME_MINUTES = "idempotency-key-lifetime-minutes";
 }
--- a/catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergConstants.java
+++ b/catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergConstants.java
@@ -128,4 +128,6 @@ public class IcebergConstants {
   public static final String SCAN_PLAN_CACHE_IMPL = "scan-plan-cache-impl";
   public static final String SCAN_PLAN_CACHE_CAPACITY = "scan-plan-cache-capacity";
   public static final String SCAN_PLAN_CACHE_EXPIRE_MINUTES = "scan-plan-cache-expire-minutes";
+
+  public static final String IDEMPOTENCY_KEY_LIFETIME_MINUTES = "idempotency-key-lifetime-minutes";
 }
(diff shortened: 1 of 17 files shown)


lasdf1234:
@laserninja Could you follow up this PR？

laserninja:
Resolved the conflicts. Let me know if you need some other changes too.

lasdf1234:
Thank you very much for your contribution. However, after reading the PR, I have a question. How can we ensure consistency of multi-node caching? How can we ensure that the returned value is consistent when a single node fails and the cache is lost between two calls?

Should we introduce a database to store this information?

You can refer to the PR of "polaris".https://github.com/apache/polaris/pull/3205

@roryqi @FANNG1 How do you think?

roryqi:
Maybe you could write a design document to cover the multiple node cases.

laserninja:
I can create that as a starting point if you want. New PR or this works?

lasdf1234:
> I can create that as a starting point if you want. New PR or this works?

You should create an Epic issue. The first task is to design the document, and the remaining tasks are to implement the code.

laserninja:
Done

PR #10757, opened 2026-04-13
[#10750] feat(trino-connector): Support CTAS (CREATE TABLE AS SELECT)

### LINKED ISSUE #10750: [FEATURE] Support CTAS (CREATE TABLE AS SELECT) in Gravitino Iceberg connector
### Describe the feature

CREATE TABLE AS SELECT (CTAS) is currently unsupported in the Gravitino Iceberg connector when queried via Trino. Attempting CTAS returns:

```
This connector does not support creating tables with data
```

Native Apache Iceberg catalogs accessed directly via Trino support CTAS. The limitation is specific to the Gravitino connector wrapper.

### Motivation

CTAS is a standard and commonly-used SQL pattern for creating Iceberg tables from existing data sources — particularly during lakehouse migrations, ETL pipeline development, and exploratory data work. 

Without CTAS, users must fall back to a two-step CREATE TABLE + INSERT INTO pattern. This is not only more verbose, but introduces a failure mode: if the CREATE succeeds but the INSERT fails (e.g. due to a timeout or transient error), the table is registered in the Iceberg REST catalog pointing to metadata files that were never written. Subsequent attempts to DROP or reload the table then fail because the connector tries to load the missing metadata before executing the drop. Recovery requires bypassing Gravitino entirely and calling the IRC REST API directly.

CTAS support would eliminate this failure mode entirely by making table creation and data load atomic.

### Describe the solution

Implement CTAS support in the Gravitino Iceberg connector so that `CREATE TABLE ... AS SELECT` queries issued via Trino (and Spark) are passed through correctly to the underlying Iceberg REST catalog. The behavior should be consistent with what a user would expect from a native Iceberg catalog — the table is created and populated atomically, and if the operation fails no partial registration is left behind.

### Additional context

Discovered while working through end-to-end Iceberg setup with Gravitino 1.2.0, Trino, and standard Apache Spark 3.5.5. Related to #10698 (multi-level namespace) and the Iceberg connector JAR compatibility investigation. The orphaned-registration failure mode caused by the missing CTAS support is documented in the companion bug filed in that thread.

laserninja:

Support CTAS (CREATE TABLE AS SELECT) in the Gravitino Trino connector by
implementing `beginCreateTable()`, `finishCreateTable()`, and the associated
page sink provider and handle wrapper.

The key design choice is that `beginCreateTable()` first creates the table in
the Gravitino catalog, then delegates data writing to the internal connector's
`beginInsert()` path. This avoids double table creation in the original
connector.

**Files changed:**
- `GravitinoMetadata.java` — `beginCreateTable()` creates table via Gravitino, then calls `internalMetadata.beginInsert()`; `getNewTableLayout()` delegates to internal
- `GravitinoOutputTableHandle.java` — **new** handle wrapper that wraps `ConnectorInsertTableHandle` (implements `ConnectorOutputTableHandle`)
- `GravitinoMetadata{435,440,446,452,469,478}.java` — version-specific `finishCreateTable()` delegating to `finishInsert()`
- `GravitinoPageSinkProvider.java` — `createPageSink(OutputTableHandle)` unwraps to insert handle and uses insert-path sink
- `JsonCodec.java` — registered `ConnectorOutputTableHandle` in `AbstractTypedJacksonModule` for polymorphic serialization

CTAS is a standard SQL pattern commonly used during lakehouse migrations, ETL
development, and exploratory data work. Without it, users must use a two-step
`CREATE TABLE` + `INSERT INTO` pattern, which introduces a failure mode where
partial table registrations can be left behind if the insert fails.

Fix: #10750

Yes. Users can now execute `CREATE TABLE ... AS SELECT` queries through the
Gravitino Trino connector. Previously this returned:
`This connector does not support creating tables with data.`

- Added `testCreateTableAsSelect` test in `TestGravitinoConnector`
- All existing tests pass across all 6 Trino SPI version modules (435–478)
- Ran `./gradlew spotlessApply` for formatting compliance


diqiu50 (commented):
Please add integration tests and documentation, especially for distributed environments. The number of workers for the underlying test containers can be configured via environment variables.

You can replace the data preparation in the TPC-DS integration tests with CTAS.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoOutputTableHandle.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * The GravitinoOutputTableHandle is used for handling CTAS (CREATE TABLE AS SELECT) operations.
+ *
+ * <p>Internally wraps a {@link ConnectorInsertTableHandle} because the Gravitino connector creates
+ * the table first via catalogConnectorMetadata, then delegates data writing to the internal
+ * connector's insert path, avoiding double table creation.
+ *
+ * <p>Also stores the {@link SchemaTableName} so that {@code rollbackCreateTable} can drop the table
+ * from the Gravitino catalog if the CTAS operation fails.
+ */
```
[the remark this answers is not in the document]
laserninja: Fixed.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoMetadata.java
```diff
@@ -220,6 +221,46 @@ public void createTable(
...
+
+    // Get the table handle from the internal connector for the newly created table
+    SchemaTableName tableName = tableMetadata.getTable();
+    ConnectorTableHandle internalTableHandle =
+        internalMetadata.getTableHandle(session, tableName, Optional.empty(), Optional.empty());
+
+    // Delegate to the internal connector's insert path to write data,
+    // avoiding double table creation in the original connector
+    List<ColumnHandle> columns =
+        new ArrayList<>(internalMetadata.getColumnHandles(session, internalTableHandle).values());
```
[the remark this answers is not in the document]
laserninja: Fixed. Changed from iterating getColumnHandles().values() (unordered Map) to iterating tableMetadata.getColumns() and looking up each column handle by name, ensuring deterministic column ordering that matches the table definition.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoMetadata.java
```diff
@@ -220,6 +221,46 @@ public void createTable(
...
+    ConnectorInsertTableHandle insertTableHandle =
+        internalMetadata.beginInsert(session, internalTableHandle, columns, retryMode);
+    return new GravitinoOutputTableHandle(insertTableHandle, tableName);
```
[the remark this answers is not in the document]
laserninja: Added a try/catch block around the getTableHandle/beginInsert sequence, if either fails after the table has been created in Gravitino, we now call catalogConnectorMetadata.dropTable(tableName) to clean up the orphaned table before re-throwing the exception. Also added a null check for internalTableHandle to handle the case where getTableHandle returns null.

laserninja:
Added:
Integration test 00013_ctas.sql/00013_ctas.txt with 5 CTAS scenarios (basic, with alias, filtered, aggregated, cross-schema)
Updated catalog_iceberg_prepare.sql to use CTAS for all 8 TPC-H data preparation tables
Added CREATE TABLE AS SELECT to the SQL support documentation in 'sql-support.md'

diqiu50:
Can you fix the CI failure

laserninja:
Pushed 426161814: reverted tpch/catalog_iceberg_prepare.sql to the original explicit CREATE TABLE + INSERT INTO ... SELECT * pattern.

Root cause: switching that script to CREATE TABLE ... AS SELECT * FROM tpch.tiny.* failed because the TPC-H connector exposes columns as VARCHAR(N), and IcebergDataTypeTransformer rejects length-bounded VARCHAR (Iceberg does not support the datatype VARCHAR with length). The previous form sidesteps this by declaring unbounded varchar columns up front and letting INSERT cast.

CTAS coverage stays in 00013_ctas.sql, so the TPC-H prep doesn't need to also demonstrate CTAS.

Also, his surfaces a separate limitation: CTAS into Iceberg from any source with bounded VARCHAR currently fails. Happy to file a follow-up issue to relax IcebergDataTypeTransformer to accept VARCHAR(N) and map it to unbounded, but it feels out of scope for this PR.

diqiu50:
@laserninja Please address the conflict

diqiu50:
> Added: Integration test 00013_ctas.sql/00013_ctas.txt with 5 CTAS scenarios (basic, with alias, filtered, aggregated, cross-schema) Updated catalog_iceberg_prepare.sql to use CTAS for all 8 TPC-H data preparation tables Added CREATE TABLE AS SELECT to the SQL support documentation in 'sql-support.md'

There some issues:
I don’t see your changes to `catalog_iceberg_prepare.sql`.
You only tested the Iceberg catalog.
 Most importantly, we need to verify that it works on a multi-worker cluster.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoMetadata.java
```diff
@@ -223,6 +224,72 @@ public void createTable(
...
+      }
+
+      // Delegate to the internal connector's insert path to write data,
+      // avoiding double table creation in the original connector
+      ConnectorInsertTableHandle insertTableHandle =
+          internalMetadata.beginInsert(session, internalTableHandle, columns, retryMode);
+      return new GravitinoOutputTableHandle(insertTableHandle, tableName);
+    } catch (Exception e) {
+      // Clean up the table created in the Gravitino catalog on failure
+      catalogConnectorMetadata.dropTable(tableName);
```
diqiu50: Exception may not  be throw when call dropTable Failed

laserninja:
The existing test infrastructure already supports multi-worker verification via the --trino_worker_num=N flag in TrinoQueryTestTool. The CTAS integration tests (for Iceberg, MySQL, and PostgreSQL) will run on a multi-worker cluster when that flag is set, no additional code changes are needed.

I've verified locally with --trino_worker_num=2 and the CTAS flow works correctly across workers since the table creation happens on the coordinator via 'beginCreateTable'/finishCreateTable, and the data write uses the existing insert path which already supports distributed execution.

laserninja:
Added CTAS integration tests for jdbc-mysql and jdbc-postgresql catalogs in this latest commit. These cover basic CTAS, CTAS with column subset + transformation, and CTAS with empty source.

Re: [catalog_iceberg_prepare.sql](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) changes:

Those were reverted in a previous commit because VARCHAR(N) caused issues with Iceberg's type system (it only supports unbounded VARCHAR). The prepare script didn't need changes since the CTAS test creates its own schema and source table.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoMetadata.java
laserninja: its is now wrapped in try/catch with LOG.warn so cleanup failures don't mask the original exception.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoMetadata.java
```diff
@@ -223,6 +224,76 @@ public void createTable(
     catalogConnectorMetadata.createTable(table, saveMode == SaveMode.IGNORE);
   }
 
+  @Override
+  public ConnectorOutputTableHandle beginCreateTable(
+      ConnectorSession session,
+      ConnectorTableMetadata tableMetadata,
+      Optional<ConnectorTableLayout> layout,
+      RetryMode retryMode,
+      boolean noExistingData) {
```
diqiu50: The last parameter in the Trino SPI is `replace`, not `noExistingData`—please handle it accordingly.

diqiu50: @laserninja Can you fixed this issue

laserninja: Renamed the last parameter from noExistingData to replace to match the Trino SPI, and added handling for CREATE OR REPLACE TABLE AS SELECT: when replace=true and the table already exists, we drop it before recreating. Otherwise existing behavior is preserved.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/GravitinoMetadata.java
```diff
@@ -230,12 +230,19 @@ public ConnectorOutputTableHandle beginCreateTable(
...
-      boolean noExistingData) {
-    // First, create the table in the Gravitino catalog
+      boolean replace) {
+    SchemaTableName tableName = tableMetadata.getTable();
+
+    // CREATE OR REPLACE TABLE AS SELECT: drop the existing table first if present.
+    if (replace
+        && catalogConnectorMetadata.tableExists(
+            tableName.getSchemaName(), tableName.getTableName())) {
+      catalogConnectorMetadata.dropTable(tableName);
```
diqiu50: This feature is not covered by any tests.

laserninja: Thanks for catching this. Added an explicit `CREATE OR REPLACE TABLE AS SELECT` scenario to the Iceberg CTAS integration test in 662241502 (`trino-connector/integration-test/src/test/resources/trino-ci-testset/testsets/lakehouse-iceberg/00013_ctas.sql`, Test 6). It creates a table, runs `CREATE OR REPLACE TABLE ... AS SELECT` against the existing table with a different column set and row count, and asserts the post-replace `SELECT` reflects the new schema/data — exercising the `replace && tableExists` drop+recreate branch in `GravitinoMetadata#beginCreateTable`.

diqiu50: Why do we remove the feature?

diqiu50: If there have problems, you can add some comment and update the document

laserninja: Thanks for the feedback. The CREATE OR REPLACE TABLE AS SELECT path was removed rather than kept with a workaround because of a fundamental incompatibility with the Iceberg internal connector's table-identity tracking.

Root cause: The internal Iceberg connector caches the table's UUID at query-plan time. The CTAS transaction flow is:
beginCreateTable → beginInsert → commitTransaction.
When replace=true, we dropped and recreated the table inside beginCreateTable. But beginInsert then calls beginTransaction → refresh(), which compares the cached UUID against the new table and throws IllegalStateException: Table UUID does not match: current=<old> != refreshed=<new>. There is no public API in the internal connector to reset this cache.

Throwing NOT_SUPPORTED is the standard Trino pattern for operations the connector genuinely cannot execute, it's what the default ConnectorMetadata base class does for any unimplemented method. The error message surfaces cleanly to the user: "This connector does not support replacing a table".

I'm happy to add an inline comment in the code explaining the UUID cache incompatibility so future maintainers understand why replace=true is rejected rather than implemented. I can also add a note to the docs if there's a connector-limitations section to update, please point me to the right file. Would that address the concern?

diqiu50: That’s fine. 
Please add comments in the code. 
For the documentation, we can include some notes in the CTAS section.

laserninja: added an inline comment in GravitinoMetadata#beginCreateTable explaining the UUID cache root cause, and updated catalog-iceberg.md, catalog-mysql.md, catalog-postgresql.md, and sql-support.md to reflect that CTAS is now supported while CREATE OR REPLACE TABLE AS SELECT is not.

docs/trino-connector/catalog-postgresql.md
```diff
@@ -27,8 +27,11 @@ Otherwise, schema names, table names or column names containing uppercase letter
 ## Create table
 
 At present, the Apache Gravitino Trino connector only supports basic PostgreSQL table creation statements, which involve fields, null allowances, and comments. However, it does not support advanced features like primary keys, indexes, default values, and auto-increment.
+The Gravitino Trino connector supports `CREATE TABLE AS SELECT`.
 
-The Gravitino Trino connector does not support `CREATE TABLE AS SELECT`.
+:::note
+`CREATE OR REPLACE TABLE AS SELECT` is not supported. Use `DROP TABLE` followed by `CREATE TABLE AS SELECT` as an alternative.
+:::
```
diqiu50, line 34: It looks like the Hive catalog is missing. We should either include it consistently in the `sql-support` document or avoid omitting it.

diqiu50 (commented):
LGTM

docs/trino-connector/catalog-postgresql.md
laserninja: I've updated catalog-hive.md to reflect that CTAS is now supported (with the same note about CREATE OR REPLACE TABLE AS SELECT not being supported), and added a Hive CTAS integration test (00018_ctas.sql/txt) with the same three scenarios used for MySQL and PostgreSQL.

diqiu50 (approved):
LGTM

laserninja:
thanks for the detailed review!

PR #10779, opened 2026-04-14
[#10762] fix(common): Lazy-initialize Version to fix TestVersion without jar task

### LINKED ISSUE #10762: [Bug report] TestVersion fails when run without prior jar task due to missing gravitino-build-info.properties
### Version

main branch

### Describe what's wrong

`TestVersion` fails when tests are run directly (e.g. `./gradlew :common:test`) without first running the `jar` task. The `gravitino-build-info.properties` file is only generated inside `common:jar { doFirst }`, not during `processResources`. Since `test` does not depend on `jar`, the file is absent from the test classpath, causing `Version`'s static initializer to throw. All four `TestVersion` test methods fail because loading the `Version` class triggers the static `INSTANCE` initialization, even though the tests only call `parseVersionNumber()` which never uses `INSTANCE`.

```
java.lang.ExceptionInInitializerError
    Caused by: org.apache.gravitino.exceptions.GravitinoRuntimeException:
    Failed to get Gravitino version: Build info properties file is missing.
        Caused by: java.io.IOException: Resource gravitino-build-info.properties not found
            at org.apache.gravitino.Version.<init>(Version.java:51)
            at org.apache.gravitino.Version.<clinit>(Version.java:41)
```

### How to reproduce

1. Run `./gradlew clean :common:test -PskipITs` (without a prior `./gradlew build` or `jar`)
2. Observe all four `TestVersion` test methods failing

### Additional context

Two possible fixes:
- Move `writeProjectPropertiesFile()` from `jar { doFirst }` to a dedicated task that `processResources` depends on, so the file is always available for tests.
- Lazy-initialize `Version.INSTANCE` so it is only created when `getCurrentVersion()` / `getCurrentVersionDTO()` are called, not when `parseVersionNumber()` is called.

sachinnn99:

Replace eager static initialization of `Version.INSTANCE` with the initialization-on-demand holder idiom. This ensures `parseVersionNumber()` can be called without triggering the properties file load.

`TestVersion` fails when tests are run directly (`./gradlew clean :common:test -PskipITs`) without first running the `jar` task. The `gravitino-build-info.properties` file is only generated inside `jar { doFirst }`, not during `processResources`. Since `test` does not depend on `jar`, the file is absent, causing `Version`'s static initializer to throw — even though all `TestVersion` methods only call `parseVersionNumber()`, which never uses `INSTANCE`.

The holder pattern is already established in this codebase by `ObjectMapperProvider` and `GravitinoEnv`.

Fix: #10762

No.

`./gradlew clean :common:test -PskipITs --tests "org.apache.gravitino.TestVersion"` — all 4 tests pass without a prior `jar` task.

### DIFF
--- a/common/src/main/java/org/apache/gravitino/Version.java
+++ b/common/src/main/java/org/apache/gravitino/Version.java
@@ -38,7 +38,9 @@ public class Version {
   private static final Pattern PATTERN =
       Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(?:rc(0|[1-9]\\d*)|-.*|\\.([a-zA-Z].*))?$");
 
-  private static final Version INSTANCE = new Version();
+  private static class InstanceHolder {
+    private static final Version INSTANCE = new Version();
+  }
 
   private final VersionInfo versionInfo;
   private final VersionDTO versionDTO;
@@ -74,14 +76,14 @@ public class Version {
    * @return the current versionInfo
    */
   public static VersionInfo getCurrentVersion() {
-    return INSTANCE.versionInfo;
+    return InstanceHolder.INSTANCE.versionInfo;
   }
 
   /**
    * @return the current version DTO
    */
   public static VersionDTO getCurrentVersionDTO() {
-    return INSTANCE.versionDTO;
+    return InstanceHolder.INSTANCE.versionDTO;
   }
 
   /**


jerryshao (approved):

sachinnn99:
Thanks for the review and merge, @jerryshao!

PR #10785, opened 2026-04-14
[#10765] feat(iceberg): Support s3-proxy-endpoint property for Iceberg REST server

### LINKED ISSUE #10765: [FEATURE] Iceberg Rest: Support of S3 Proxy Host and Port
### Describe the feature

Add support of S3:
- s3.proxy.host
- s3.proxy.port
to the Iceberg Catalog definition, so that a client can also fetch S3 proxy settings along side with already existing properties like endpoint, pathStyleAccess and region.

### Motivation

Allow Catalog Clients like Flink and Spark to use proxy settings to be able to work S3 behind the HTTP PROXY.

### Describe the solution

Add two properties the Iceberg REST Catalog definition so that they can be passed down to S3 HTTP Client.

### Additional context

_No response_

sachinnn99:

Add first-class support for configuring an S3 proxy endpoint in the Iceberg REST catalog. The new Gravitino property `s3-proxy-endpoint` maps to Iceberg's `http-client.proxy-endpoint` (`HttpClientProperties.PROXY_ENDPOINT`), allowing clients like Spark and Flink behind an HTTP proxy to fetch proxy settings from the catalog configuration.

Users behind an HTTP proxy cannot configure S3 proxy settings through the Iceberg REST catalog. While the `gravitino.bypass.` mechanism can pass arbitrary properties, a first-class managed property provides a consistent experience alongside existing S3 properties like `s3-endpoint`, `s3-region`, and `s3-path-style-access`.

Fixes #10765

Yes. Adds a new optional catalog property:

| Gravitino property | Iceberg property | Description |
|---|---|---|
| `s3-proxy-endpoint` | `http-client.proxy-endpoint` | The proxy endpoint for S3 requests, e.g. `http://proxy-host:8080` |

- Added unit test in `TestIcebergPropertiesUtils` for the property mapping
- Added proxy endpoint to `IcebergRestTestUtil` test catalog config
- Updated `TestIcebergConfig.testConfigWithValidWarehouse` to verify proxy endpoint in REST config response
- Updated `TestCatalogWrapperForREST` to verify proxy endpoint passes through client config filtering
- All existing tests pass: `./gradlew :catalogs:catalog-common:test :iceberg:iceberg-rest-server:test :catalogs:catalog-lakehouse-iceberg:test -PskipITs`


novakov-alexey:
Hi @sachinnn99 ,

This looks very promising. Thanks.

I would also advise to check whether it changes something in your PR, when a proxy endpoint uses `https` scheme instead of `http`. 

I know at least at least one point. In the runtime, Gravitino JVM process may require CA Certs to be imported.
Another thing could be some SSL dependencies.

sachinnn99:
@novakov-alexey Thanks for the feedback. I checked this — the `s3-proxy-endpoint` value is passed through as a string to Iceberg's `http-client.proxy-endpoint`, which is consumed by the AWS SDK HTTP client. Gravitino never opens a connection to this URL itself.

When the scheme is `https`, the AWS SDK handles TLS using the JVM's default truststore — the same way `s3-endpoint` works today (which also supports `https` without any special SSL handling in Gravitino). So no code changes are needed for HTTPS support; it works out of the box.

The CA cert / truststore configuration you mentioned is an operational concern at deployment time (ensuring the JVM trusts the proxy's certificate), not something this property-mapping layer needs to address.

novakov-alexey:
@sachinnn99 Perfect, appreciate your investigation.

roryqi:
Have u tested it using the real environment instead of unit tests or integration tests?

Issue #10457, opened 2026-03-17
[Improvement] Shrink the package size of Hive Metastore 2 and Hive Metastore 3 catalogs


jerryshao:
### What would you like to be improved?

The distribution packages for the Hive Metastore 2 and Hive Metastore 3 catalogs are currently too large. This increases download time, storage requirements, and deployment overhead for users.

### How should we improve?

Investigate and reduce the package size by:
- Auditing bundled dependencies to identify large or redundant JARs
- Excluding transitive dependencies that are already provided by the Gravitino server runtime
- Shading or relocating only the necessary classes instead of bundling full dependency JARs
- Removing any test or unused dependencies included in the final package

PR #10789, opened 2026-04-15
[#10788] improvement(deps): upgrade Hadoop from 2.10.2 to 3.3.6 and remove hadoop2 dependency line

### LINKED ISSUE #10788: [Improvement] Upgrade Hadoop from 2.10.2 to 3.3.6 and unify on a single Hadoop 3.x dependency line

Gravitino currently maintains two parallel Hadoop dependency lines:` hadoop2 (2.10.2)` and `hadoop3 (3.3.1)`. This dual-version setup increases maintenance burden,  causes classpath conflicts, and brings in outdated transitive dependencies from the Hadoop 2.x line.

### How should we improve?

Remove the hadoop2 dependency line entirely and unify all modules on `hadoop3 (3.3.6)`. Bump hadoop3 from `3.3.1` to `3.3.6` for the latest patch-level fixes and compatibility improvements.

geyanggang:

- Remove `hadoop2` version (2.10.2) and its 6 library definitions from libs.versions.toml
- Bump `hadoop3` from 3.3.1 to 3.3.6 (including `hadoop3-abs`, `hadoop3-aliyun`, `hadoop-minikdc`)
- Replace all `libs.hadoop2.*` references with `libs.hadoop3.*` in 8 build files
- Add Hadoop 3.x runtime dependencies stripped by `exclude("*")`: `hadoop-shaded-guava`, `hadoop-shaded-protobuf`, `commons-configuration2`, `re2j`,`hadoop-hdfs-client`
- Remove outdated hadoop2/hadoop3 conflict comment in `catalog-hive`

- The `hadoop2` (2.10.2) dependency line brings in outdated transitive dependencies (Jackson 2.6.x, Netty 3.x, Jetty 9.3.x) that are no longer maintained
- `hive-metastore3-libs` was incorrectly configured with `hadoop2` despite Hive 3.1.3 natively depending on Hadoop 3.1.0
- Unifying on a single Hadoop version eliminates classpath conflicts and simplifies dependency management

Fix: #10788 

No.

- Full compilation (`./gradlew build -x test`) — passed
- Verified `hadoop-common-3.3.6` resolves in all affected modules, no Hadoop 2.x JARs in distribution
- All unit tests passed across 12 key modules including `catalog-hive`, `hive-metastore-common`, `catalog-lakehouse-hudi`, `catalog-fileset`, `filesystem-hadoop3`, `hadoop-common`, `authorization-ranger`, `spark-connector`, `flink-connector`, `core`
- Integration tests pending CI validation


trino-connector/integration-test/build.gradle.kts
```diff
@@ -41,13 +41,19 @@ dependencies {
   testImplementation(libs.bundles.jetty)
   testImplementation(libs.bundles.log4j)
   testImplementation(libs.commons.cli)
-  testImplementation(libs.hadoop2.common) {
+  testImplementation(libs.hadoop3.common) {
     exclude("*")
```
diqiu50, line 45: I think the trino integration should not depend on hadoop

trino-connector/integration-test/build.gradle.kts
```diff
@@ -41,13 +41,19 @@ dependencies {
...
-  testImplementation(libs.hadoop2.mapreduce.client.core) {
+  testImplementation(libs.hadoop3.hdfs)
+  testImplementation(libs.hadoop3.hdfs.client)
+  testImplementation(libs.hadoop3.mapreduce.client.core) {
     exclude("*")
   }
+  // Hadoop 3.x runtime requirements (stripped by exclude("*") above)
+  testImplementation(libs.hadoop3.shaded.guava)
+  testImplementation(libs.hadoop3.shaded.protobuf)
+  testImplementation("org.apache.commons:commons-configuration2:2.8.0")
```
diqiu50: We should not use the version directly here

flink-connector/flink/build.gradle.kts
```diff
@@ -141,9 +141,15 @@ dependencies {
...
-  testImplementation(libs.hadoop2.mapreduce.client.core) {
+  testImplementation(libs.hadoop3.hdfs.client)
+  testImplementation(libs.hadoop3.mapreduce.client.core) {
     exclude("*")
   }
+  // Hadoop 3.x runtime requirements (stripped by exclude("*") above)
+  testImplementation(libs.hadoop3.shaded.guava)
+  testImplementation(libs.hadoop3.shaded.protobuf)
+  testImplementation("org.apache.commons:commons-configuration2:2.8.0")
+  testImplementation(libs.re2j)
```
diqiu50, line 152: Same problem.
I found you add more dependency for using haddoop3, Is it correct?

trino-connector/integration-test/build.gradle.kts
geyanggang: Replacing hadoop2 with hadoop3. The trino integration test previously depended on hadoop2.common, now it uses hadoop3.common with exclude("*") to keep the same minimal-dependency approach.

trino-connector/integration-test/build.gradle.kts
geyanggang: OK, I'll move `commons-configuration2` to `libs.versions.toml` as a managed dependency instead of hardcoding the version string.

flink-connector/flink/build.gradle.kts
geyanggang: Yes, this is expected. Hadoop 2.x bundled everything inside `hadoop-common`. Hadoop 3.x extracted several internal dependencies into separate artifacts:

- `hadoop-shaded-guava` — Guava relocated to avoid conflicts
- `hadoop-shaded-protobuf` — Protobuf relocated
- `commons-configuration2` — replaced `commons-configuration` 1.x
- `re2j` — new dependency used by `Globber.doGlob()` since 3.3.6

Since we use `exclude("*")` to strip all transitives from `hadoop-common`, these must be declared explicitly. Without them, you get `NoClassDefFoundError` at runtime.

trino-connector/integration-test/build.gradle.kts
geyanggang: Fixed.

gradle/libs.versions.toml
```diff
@@ -197,22 +197,22 @@ hive2-common = { group = "org.apache.hive", name = "hive-common", version.ref =
...
 hadoop3-client-api = { group = "org.apache.hadoop", name = "hadoop-client-api", version.ref = "hadoop3"}
 hadoop3-client-runtime = { group = "org.apache.hadoop", name = "hadoop-client-runtime", version.ref = "hadoop3"}
+hadoop3-mapreduce-client-core = { group = "org.apache.hadoop", name = "hadoop-mapreduce-client-core", version.ref = "hadoop3"}
 hadoop3-minicluster = { group = "org.apache.hadoop", name = "hadoop-minicluster", version.ref = "hadoop-minikdc"}
 hadoop3-gcs = { group = "com.google.cloud.bigdataoss", name = "gcs-connector", version.ref = "hadoop3-gcs"}
 hadoop3-oss = { group = "org.apache.hadoop", name = "hadoop-aliyun", version.ref = "hadoop3-aliyun"}
 hadoop3-abs = { group = "org.apache.hadoop", name = "hadoop-azure", version.ref = "hadoop3-abs"}
+hadoop3-shaded-guava = { group = "org.apache.hadoop.thirdparty", name = "hadoop-shaded-guava", version = "1.1.1" }
+hadoop3-shaded-protobuf = { group = "org.apache.hadoop.thirdparty", name = "hadoop-shaded-protobuf_3_7", version = "1.1.1" }
+re2j = { group = "com.google.re2j", name = "re2j", version = "1.7" }
```
diqiu50: Don't hard code the version here

trino-connector/integration-test/build.gradle.kts
diqiu50: I think you don’t need to exclude * for libs.hadoop3.common. The following dependency was added because of it.

geyanggang: This was the same approach used with hadoop2.common before.

gradle/libs.versions.toml
geyanggang: Fixed.

trino-connector/integration-test/build.gradle.kts
geyanggang: The exclude("*") is intentional and consistent with other modules (catalog-hive, flink-connector, etc.). We exclude all transitive deps from hadoop-common, then explicitly add back only what's needed (hadoop3-shaded-guava, hadoop3-shaded-protobuf, commons-configuration2, re2j).

geyanggang: The exclude("*") is intentional — without it, hadoop-common pulls in many transitive dependencies that cause conflicts. We exclude all and explicitly add back only what's needed. This is consistent with other modules (catalog-hive, flink-connector, etc.).

diqiu50: Could you give it a try? I think excluding everything and then including several dependencies back is not a good approach. 
@yuqi1129  What would you think

geyanggang: OK,Gave it a try on trino test — removing exclude("*") from hadoop3.common works fine, no transitive dependency conflicts found. Also removed the 4 explicit deps (hadoop3-shaded-guava, hadoop3-shaded-protobuf, commons-configuration2, re2j) since they're now pulled in transitively. This is testImplementation only so no production impact. Updated.

geyanggang: I tried removing exclude("*") and it compiled fine locally with no dependency conflicts. However, CI failed with a javaCompiler error in trino-connector-435-439:compileJava, which may be related to the dependency change affecting Gradle's task graph. Reverted to the original exclude("*") approach to be safe. Will revisit this separately if needed.

diqiu50: This was a separate issue related to the CI toolchain configuration. It has been fixed, and the latest changes should resolve it. Please update and try again.

authorizations/authorization-chain/build.gradle.kts
```diff
@@ -104,6 +104,7 @@ dependencies {
     exclude("javax.servlet", "servlet-api")
     exclude("io.netty")
   }
+  testImplementation(libs.hadoop3.hdfs.client)
```
jerryshao, line 107: Why don't we need this before?

catalogs/catalog-hive/build.gradle.kts
```diff
@@ -49,12 +49,17 @@ dependencies {
...
   }
-  implementation(libs.hadoop2.common) {
+  implementation(libs.hadoop3.common) {
     exclude("*")
   }
+  // Hadoop 3.x runtime requirements (stripped by exclude("*") above)
+  implementation(libs.hadoop3.shaded.guava)
+  implementation(libs.hadoop3.shaded.protobuf)
+  implementation(libs.commons.configuration2)
+  implementation(libs.re2j)
```
jerryshao, line 62: Excluding everything and adding back dependencies one by one is very tricky. Simply, because of bumping the major version will introduce lots of dependency changes, the previous way that works with Hadoop 2 may not apply to Hadoop 3, unless you check all the dependencies of Hadoop 3 and make sure that you pull in the ones you wanted exactly.

clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/GravitinoVirtualFileSystemABSCredentialIT.java
```diff
@@ -111,7 +111,7 @@ public void startUp() throws Exception {
     conf.set(AzureProperties.GRAVITINO_AZURE_STORAGE_ACCOUNT_KEY, ABS_ACCOUNT_KEY);
     conf.set("fs.abfss.impl", "org.apache.hadoop.fs.azurebfs.SecureAzureBlobFileSystem");
 
-    conf.set("fs.gravitino.client.useCloudStoreCredential", "true");
+    conf.set("fs.gravitino.enableCredentialVending", "true");
```
jerryshao, line 114: @yuqi1129 @mchades can you confirm this change? Is it right?

trino-connector/integration-test/build.gradle.kts
geyanggang: OK,I'll try it again.

clients/filesystem-hadoop3/src/test/java/org/apache/gravitino/filesystem/hadoop/integration/test/GravitinoVirtualFileSystemABSCredentialIT.java
yuqi1129: This is an out-of-date configuration item.

authorizations/authorization-chain/build.gradle.kts
geyanggang: In **Hadoop 2**, `hadoop-hdfs `depends on `hadoop-hdfs-client` with **compile** scope, so it was pulled in transitively. In **Hadoop 3**, this was changed to **provided** scope, meaning it's no longer transitive and must be declared explicitly. That's why we need to add it now.

catalogs/catalog-hive/build.gradle.kts
geyanggang: For implementation scope (like catalog-hive), I think exclude("*") + explicit deps is the safer approach. Hadoop 3.3.6 pulls in many transitive deps (Guava, Protobuf, Jetty, etc.) that are likely to conflict with Hive 2.3.9's own dependencies. Removing exclude("*") would require extensive per-dependency exclusions to resolve conflicts, which is harder to maintain and more fragile. The current approach gives us precise control over the classpath.

geyanggang: I tested removing exclude("*") — it compiles fine, but the runtime classpath balloons from ~20 to 150+ jars. More importantly, hadoop-common:3.3.6 transitively pulls in several vulnerable dependencies: log4j:1.2.17, jackson-mapper-asl:1.9.13, protobuf-java:2.5.0, jettison:1.1, etc.

trino-connector/integration-test/build.gradle.kts
geyanggang: Done

diqiu50: Should any packages be excluded?

geyanggang: I tried selective excludes but `hadoop-common`'s CVE deps (Jersey 1.x, jettison, reload4j, etc.) also come in through deep transitive chains (hadoop-auth → zookeeper → ...), which Gradle's per-dependency exclude can't fully block. Reverted to `exclude("*")` + explicit add-back for a clean classpath. This is consistent with all other modules (`catalog-hive`, `hudi,` `flink`, `hive-metastore-common`).

catalogs/catalog-hive/build.gradle.kts
geyanggang: I analyzed both `hadoop-common:2.10.2` (41 compile deps) and `3.3.6` (44 compile deps) source pom.xml. Hadoop 3 added 10 new deps and removed 7 old ones. Of the 10 new: 4 explicitly added back (`hadoop-shaded-guava`, `hadoop-shaded-protobuf`, `commons-configuration2`, `re2j`), 1 auto-resolved transitively, 5 safely excluded. The excluded categories are identical to Hadoop 2.
    I also tried selective excludes instead of `exclude("*")`, but CVE deps (Jersey 1.x, jettison, reload4j, etc.) leak in through deep transitive chains (`hadoop-auth` → `zookeeper` → ...) that Gradle's per-dependency exclude can't fully block. `exclude("*")` + explicit add-back is the only reliable approach for a clean classpath.

geyanggang: I also put a detailed side-by-side comparison of Hadoop 2 vs 3 dependencies here:
[hadoop-exclude-dependency-analysis.md](https://docs.google.com/document/d/13C0IG8DZEN-PxXsb2z2llKIwOwaDLppBTNTK-VeK3iQ/edit?tab=t.0) @jerryshao @diqiu50

jerryshao, line 62: For re2, why do we need this?

geyanggang: `re2j` is a new compile dependency in Hadoop 3 (not present in Hadoop 2). It's used by `GlobPattern` and `GlobFilter` in `hadoop-common`, which are on the `FileSystem.globStatus()` code path. Without it, we get `NoClassDefFoundError: com/google/re2j/PatternSyntaxException` at runtime.

trino-connector/integration-test/build.gradle.kts
```diff
@@ -41,11 +41,20 @@ dependencies {
...
+    exclude("*")
+  }
+  // Hadoop 3.x runtime requirements (stripped by exclude("*") above)
+  testImplementation(libs.hadoop3.shaded.guava)
+  testImplementation(libs.hadoop3.shaded.protobuf)
+  testImplementation(libs.commons.configuration2)
+  testImplementation(libs.re2j)
+  testImplementation(libs.hadoop3.mapreduce.client.core) {
     exclude("*")
   }
```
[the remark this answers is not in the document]
geyanggang: Done.

catalogs/catalog-lakehouse-paimon/build.gradle.kts
```diff
@@ -103,7 +103,7 @@ dependencies {
   }
   implementation(libs.commons.lang3)
   implementation(libs.guava)
-  implementation(libs.hadoop2.common) {
+  implementation(libs.hadoop3.common) {
```
jerryshao, line 106: Do we need to exclude and adjust the dependencies here and below? Is changing Hadoop version alone enough?

catalogs/catalog-lakehouse-paimon/build.gradle.kts
```diff
@@ -113,16 +113,16 @@ dependencies {
     exclude("org.apache.zookeeper")
     exclude("org.mortbay.jetty")
   }
-  implementation(libs.hadoop2.hdfs) {
+  implementation(libs.hadoop3.hdfs) {
     exclude("*")
   }
-  implementation(libs.hadoop2.hdfs.client) {
+  implementation(libs.hadoop3.hdfs.client) {
```
jerryshao: Also here.

catalogs/catalog-lakehouse-paimon/build.gradle.kts
geyanggang: Fixed — Hadoop 3 changed several transitive deps (Mortbay Jetty → Eclipse Jetty, `com.sun.jersey:jersey-json` → `com.github.pjfanning:jersey-json`, added `reload4j`), so the old exclude lists weren't covering them. Updated excludes in `hadoop3.common`, `hadoop3.mapreduce.client.core`, and cleaned up `hadoop3.hdfs.client` (its Hadoop 2 era excludes are no longer needed in Hadoop 3).

catalogs/catalog-lakehouse-paimon/build.gradle.kts
geyanggang: The same issue in `hive-metastore2-libs`, `hive-metastore3-libs`. Fixed.

jerryshao:
I don't have further comments. @yuqi1129 @diqiu50 can you please also take a review?

catalogs/hive-metastore2-libs/build.gradle.kts
```diff
@@ -29,27 +29,35 @@ plugins {
...
     exclude(group = "net.java.dev.jets3t")
     exclude(group = "org.apache.avro")
     exclude(group = "org.apache.logging.log4j")
+    exclude(group = "org.eclipse.jetty")
     exclude(group = "org.eclipse.jetty.aggregate", module = "jetty-all")
     exclude(group = "org.eclipse.jetty.orbit", module = "javax.servlet")
     exclude(group = "org.slf4j")
   }
-  implementation(libs.hadoop2.mapreduce.client.core) {
+  implementation(libs.hadoop3.mapreduce.client.core) {
```
yuqi1129, line 50: Why does Hive2 need this and Hive3 don't?

geyanggang: Hive2 metastore has a compile-scope dependency on mapreduce-client-core (uses JobConf, etc.), while Hive3 standalone-metastore only needs it as provided or not at all. That's why hive-metastore2-libs has had this dependency since the initial shim (74e9881), and hive-metastore3-libs never did. This line just bumps it from hadoop2 → hadoop3 — no functional change.

geyanggang: This difference predates this PR — hadoop-mapreduce-client-core was added to hive-metastore2-libs in a previous PR (#10457) to shrink package size, but was never added to hive-metastore3-libs. Hive 2 metastore requires some MapReduce client classes that Hive 3 metastore doesn't depend on directly.

yuqi1129:
I have no more comments, @diqiu50 would you like to take a look?

diqiu50 (commented):
LGTM

jerryshao (approved):

PR #10800, opened 2026-04-16
[#10685] feat(iceberg-rest): Support vended credentials on planTableScan endpoint

### LINKED ISSUE #10685: [Improvement] planTableScan endpoint should support X-Iceberg-Access-Delegation and return storage credentials

The `planTableScan` endpoint (`POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/scan`) does not accept the `X-Iceberg-Access-Delegation` header and does not return storage credentials in its response. The Iceberg REST spec defines the `data-access` header as a parameter on the `planTableScan` endpoint, and the `CompletedPlanningResult` schema includes a `storage-credentials` field for clients to access the data files returned in the scan result.

In the current `planTableScan` method ([IcebergTableOperations.java#L513-L545](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java#L513-L545)):
- No `@HeaderParam(X_ICEBERG_ACCESS_DELEGATION)` parameter
- `IcebergRequestContext` is constructed without `isCredentialVending`
- The `CompletedPlanningResult` response does not include `storage-credentials`

In contrast, `createTable` and `loadTable` accept this header and inject credentials into the response. Without this, clients performing server-side scan planning must make a separate `GET .../credentials` call to obtain storage access credentials before reading the data files returned in the scan plan.

### How should we improve?

- Add `@HeaderParam(X_ICEBERG_ACCESS_DELEGATION) String accessDelegation` to `planTableScan`
- Pass `isCredentialVending` to `IcebergRequestContext` so credential injection can be triggered
- When credential vending is enabled, include `storage-credentials` in the `CompletedPlanningResult` response so clients can immediately access the returned data files without a separate credentials call

sachinnn99:

Add `X-Iceberg-Access-Delegation` header support to the `planTableScan` endpoint and return storage credentials in the response when `vended-credentials` is requested. Mirrors the existing credential-vending flow from `createTable`/`loadTable`/`registerTable` (#10684).

Changes:
- `IcebergTableOperations.planTableScan`: read `@HeaderParam(X_ICEBERG_ACCESS_DELEGATION)`, compute `isCredentialVending`, build `IcebergRequestContext` with credential flag
- `IcebergTableOperationExecutor.planTableScan`: resolve `CredentialPrivilege` and pass credential-vending flag through the dispatcher to `CatalogWrapperForREST`
- `CatalogWrapperForREST.planTableScan`: add 4-arg overload that accepts `requestCredentialVending` and `CredentialPrivilege`; inject credentials via `PlanTableScanResponse.Builder.withCredentials()` *after* cache lookup so cached plans remain credential-free
- `CatalogWrapperForREST.injectScanCredentials`: reuses the already-loaded `Table` object from scan planning (no redundant `loadTable`), catches only `ServiceUnavailableException`
- `CatalogWrapperForREST.getCredentialFromTable`: extracts credentials from a `Table` object directly, following the same location/property logic as `getCredential(TableMetadata, ...)`
- `FederatedCatalogWrapper.planTableScan`: overrides to fetch credentials from the remote REST catalog instead of the local credential manager
- `TestCatalogWrapperForREST.testPlanTableScanCacheDoesNotLeakCredentials`: verifies cached plans don't leak credentials to non-vended requests and that vended requests on cache hits get fresh credentials
- `TestIcebergTableOperations`: three new tests — credential vending (no header / local / S3 with prefix verification), remote-signing rejection (406), invalid header (400)

The Iceberg REST spec defines `X-Iceberg-Access-Delegation` as a valid header on `planTableScan` and `CompletedPlanningResult` includes a `storage-credentials` field. Currently, clients performing server-side scan planning must make a separate `GET .../credentials` call to obtain storage access credentials before reading the data files returned in the scan plan.

Fixes #10685

Yes. The `planTableScan` REST endpoint now accepts the `X-Iceberg-Access-Delegation` header and returns vended credentials in the response `storage-credentials` field when requested. Backward compatible -- clients that do not send the header get existing behavior.

Added unit tests:
- `TestCatalogWrapperForREST.testPlanTableScanCacheDoesNotLeakCredentials` -- vended request returns credentials, non-vended cache hit returns none, vended cache hit returns fresh credentials
- `TestIcebergTableOperations.testPlanTableScanWithCredentialVending` -- no vending without header, no vending for `file://` location, vending present for `s3://` location with correct credential type and prefix
- `TestIcebergTableOperations.testPlanTableScanRemoteSigningNotSupported` -- 406 response
- `TestIcebergTableOperations.testPlanTableScanInvalidAccessDelegation` -- 400 response

All existing tests pass (no regressions).

### DIFF
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
@@ -201,6 +201,49 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
     }
   }
 
+  /**
+   * Vend credentials for a table if eligible. Unlike {@link #getTableCredentials}, this method
+   * respects the same vending gate ({@link #shouldGenerateCredential}) used by
+   * createTable/loadTable, so local/HDFS tables correctly return empty.
+   *
+   * <p>Returns credentials separately (not embedded in a response config map) because {@link
+   * PlanTableScanResponse} in Iceberg 1.10.1 has no config/credentials field.
+   *
+   * @param identifier the table identifier
+   * @param requestCredential whether the client requested credential vending
+   * @param privilege the credential privilege level (READ for scans)
+   * @return credentials response, empty if the table is not eligible for vending
+   */
+  public LoadCredentialsResponse getCredentialsIfEligible(
+      TableIdentifier identifier, boolean requestCredential, CredentialPrivilege privilege) {
+    try {
+      LoadTableResponse loadTableResponse = super.loadTable(identifier);
+      if (!shouldGenerateCredential(loadTableResponse, requestCredential)) {
+        return ImmutableLoadCredentialsResponse.builder().build();
+      }
+      Credential credential = getCredential(loadTableResponse, privilege);
+      org.apache.iceberg.rest.credentials.Credential icebergCredential =
+          new org.apache.iceberg.rest.credentials.Credential() {
+            @Override
+            public String prefix() {
+              return "";
+            }
+
+            @Override
+            public Map<String, String> config() {
+              return CredentialPropertyUtils.toIcebergProperties(credential);
+            }
+
+            @Override
+            public void validate() {}
+          };
+      return ImmutableLoadCredentialsResponse.builder().addCredentials(icebergCredential).build();
+    } catch (ServiceUnavailableException e) {
+      LOG.warn("Failed to generate scan credentials for table: {}", identifier, e);
+      return ImmutableLoadCredentialsResponse.builder().build();
+    }
+  }
+
   @Override
   public void close() throws Exception {
     try {
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
@@ -39,6 +39,7 @@ import org.apache.gravitino.credential.CatalogCredentialManager;
 import org.apache.gravitino.credential.Credential;
 import org.apache.gravitino.credential.CredentialConstants;
 import org.apache.gravitino.credential.CredentialPrivilege;
+import org.apache.gravitino.credential.CredentialPropertyUtils;
 import org.apache.gravitino.credential.PathBasedCredentialContext;
 import org.apache.gravitino.iceberg.common.IcebergConfig;
 import org.apache.gravitino.iceberg.common.ops.IcebergCatalogWrapper;
@@ -48,6 +49,7 @@ import org.apache.gravitino.storage.GCSProperties;
 import org.apache.gravitino.utils.ClassUtils;
 import org.apache.gravitino.utils.MapUtils;
 import org.apache.gravitino.utils.PrincipalUtils;
+import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.IncrementalAppendScan;
 import org.apache.iceberg.Scan;
@@ -373,7 +375,19 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
   }
 
   /**
-   * Plan table scan and return scan tasks.
+   * Plan table scan without credential vending.
+   *
+   * @param tableIdentifier The table identifier.
+   * @param scanRequest The scan request parameters.
+   * @return PlanTableScanResponse with status=COMPLETED and file scan tasks.
+   */
+  public PlanTableScanResponse planTableScan(
+      TableIdentifier tableIdentifier, PlanTableScanRequest scanRequest) {
+    return planTableScan(tableIdentifier, scanRequest, false, CredentialPrivilege.READ);
+  }
+
+  /**
+   * Plan table scan and optionally inject vended storage credentials.
    *
    * <p>This method performs server-side scan planning to optimize query performance by reducing
    * client-side metadata loading and enabling parallel task execution.
@@ -384,18 +398,27 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
    * different from asynchronous mode (SUBMITTED status) where a plan ID is returned for later
    * retrieval.
    *
+   * <p>When {@code requestCredentialVending} is true and the table is eligible (non-local,
+   * non-HDFS), storage credentials are injected directly into the response using the table already
+   * loaded for scan planning -- avoiding a redundant {@code loadTable} call.
+   *
    * <p>Referenced from Iceberg PR #13400 for scan planning implementation.
    *
    * @param tableIdentifier The table identifier.
    * @param scanRequest The scan request parameters including filters, projections, snapshot-id,
    *     etc.
+   * @param requestCredentialVending whether the client requested credential vending
+   * @param privilege the credential privilege level for vending
    * @return PlanTableScanResponse with status=COMPLETED and file scan tasks.
    * @throws IllegalArgumentException if scan request validation fails
    * @throws org.apache.gravitino.exceptions.NoSuchTableException if table doesn't exist
    * @throws RuntimeException for other scan planning failures
    */
   public PlanTableScanResponse planTableScan(
-      TableIdentifier tableIdentifier, PlanTableScanRequest scanRequest) {
+      TableIdentifier tableIdentifier,
+      PlanTableScanRequest scanRequest,
+      boolean requestCredentialVending,
+      CredentialPrivilege privilege) {
 
     LOG.debug(
         "Planning scan for table: {}, snapshotId: {}, startSnapshotId: {}, endSnapshotId: {}, select: {}, caseSensitive: {}",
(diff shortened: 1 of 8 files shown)


roryqi:
@sachinnn99 Could u resolve the conflicts? @laserninja Could u help me review this pull request?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -509,27 +517,50 @@ public Response planTableScan(
...
 
             PlanTableScanResponse scanResponse =
                 tableOperationDispatcher.planTableScan(context, tableIdentifier, scanRequest);
 
+            if (isCredentialVending) {
+              try {
+                LoadCredentialsResponse credentialsResponse =
+                    icebergCatalogWrapperManager
+                        .getCatalogWrapper(catalogName)
+                        .getCredentialsIfEligible(tableIdentifier, true, CredentialPrivilege.READ);
```
laserninja: **[HIGH] Bypasses the dispatcher/event-listener layer**

All other credential operations go through `tableOperationDispatcher` (e.g. `getTableCredentials`), which triggers event listeners and authorization hooks. Here the endpoint accesses `IcebergCatalogWrapperManager` directly, so:
- Credential vending on `planTableScan` won't emit events
- Authorization hooks won't fire for the credential fetch

The existing `loadTable`/`createTable` endpoints delegate credential vending to `CatalogWrapperForREST` via the dispatcher. The same pattern should be followed here — e.g. add a `planTableScan` overload in `CatalogWrapperForREST` that accepts `requestCredential`, or route through a dispatcher method.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -509,27 +517,50 @@ public Response planTableScan(
...
+                if (!credentialsResponse.credentials().isEmpty()) {
+                  return buildScanResponseWithCredentials(scanResponse, credentialsResponse);
+                }
+              } catch (Exception e) {
```
laserninja: **[HIGH] Overly broad `catch (Exception e)` silently swallows errors**

This catches *all* exceptions including `IllegalArgumentException`, `NullPointerException`, authorization failures, etc. A client requesting `vended-credentials` that fails due to a misconfiguration would silently get no credentials with only a WARN log — very hard to debug.

The existing `getTableCredentials` method only catches `ServiceUnavailableException`. This should follow the same pattern:
```java
} catch (ServiceUnavailableException e) {
```
Other exceptions (auth failures, illegal state, NPE) should propagate so they surface as proper error responses.

laserninja (commented):
Thanks for the PR! Posted some Inline comments

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -201,6 +201,49 @@ public void validate() {}
...
+   *
+   * @param identifier the table identifier
+   * @param requestCredential whether the client requested credential vending
+   * @param privilege the credential privilege level (READ for scans)
+   * @return credentials response, empty if the table is not eligible for vending
+   */
+  public LoadCredentialsResponse getCredentialsIfEligible(
+      TableIdentifier identifier, boolean requestCredential, CredentialPrivilege privilege) {
+    try {
+      LoadTableResponse loadTableResponse = super.loadTable(identifier);
```
laserninja: **[HIGH] Redundant `loadTable` call**

`planTableScan` already loads and scans the table via `tableOperationDispatcher.planTableScan(...)`. Then `getCredentialsIfEligible` calls `super.loadTable(identifier)` *again* — a second full table load just to check the location and generate credentials. This is wasteful for every credential-vended scan.

Consider either:
1. Passing the table metadata/location obtained during `planTableScan` into the credential vending method, or
2. Caching or reusing the already-loaded table within the same request

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergTableOperations.java
```diff
@@ -623,6 +654,19 @@ private boolean isCredentialVending(String accessDelegation) {
     }
   }
 
+  private Response buildScanResponseWithCredentials(
```
laserninja: **[MEDIUM] JSON-level merge is fragile — add a TODO with issue reference**

Manually constructing JSON via `ObjectNode`/`ArrayNode` is fine as a temporary workaround for Iceberg 1.10.1's lack of `PlanTableScanResponse.Builder.withCredentials()`, but:

1. Please add a `// TODO(#XXXX): Replace with PlanTableScanResponse.Builder.withCredentials() after Iceberg 1.11.0 upgrade` comment so this is tracked for cleanup.
2. Verify the JSON structure matches the Iceberg REST spec's `CompletedPlanningResult` schema — specifically the `storage-credentials` array format.
3. This also bypasses the standard `IcebergRESTUtils.ok()` serialization path. Any future changes to response serialization (e.g. custom serializers, response wrapping) won't apply here.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergTableOperations.java
```diff
@@ -1046,4 +1046,109 @@ void testLoadTableSnapshotsAllReturnsAllSnapshots(Namespace namespace) {
...
+    Response s3ScanResponse =
+        doPlanTableScanWithCredentialVending(namespace, s3TableName, emptyRequest);
+    Assertions.assertEquals(Status.OK.getStatusCode(), s3ScanResponse.getStatus());
+    try {
+      JsonNode s3ScanJson = JsonUtil.mapper().readTree(s3ScanResponse.readEntity(String.class));
+      Assertions.assertTrue(
+          s3ScanJson.has("storage-credentials"), "S3 table should have storage-credentials");
+      JsonNode credentials = s3ScanJson.get("storage-credentials");
+      Assertions.assertTrue(credentials.isArray() && credentials.size() > 0);
+      JsonNode firstCred = credentials.get(0);
```
laserninja, line 1173: **[MEDIUM] Tests don't verify `prefix` or refresh properties**

The tests verify `config` contains the credential type, but don't verify:
1. The `prefix` field is set correctly — currently `getCredentialsIfEligible` returns `""` (empty string), but the existing `getTableCredentials`/`injectCredentialConfig` methods use the table's metadata location as the prefix. This discrepancy should be tested.
2. Refresh properties (e.g. `client.refresh-credentials-endpoint`) are present — `IcebergRESTUtils.toRESTCredential()` adds these but this code path uses raw `CredentialPropertyUtils.toIcebergProperties()` which omits them.
3. The overall credential structure matches what `loadTable` with `vended-credentials` returns for the same table, ensuring consistency.

Consider adding:
```java
Assertions.assertFalse(firstCred.get("prefix").asText().isEmpty(),
    "Credential prefix should be the table location");
```

sachinnn99:
@laserninja Addressed everything in the latest push:

- Dispatcher bypass: moved credential vending into the executor + wrapper, matching the `loadTable` pattern. Privilege is now determined via `getCredentialPrivilege()` instead of hardcoded READ.
- Broad catch: eliminated entirely from the REST layer. Only `ServiceUnavailableException` is caught inside the wrapper now.
- Redundant loadTable: `planTableScan` reuses the `Table` already loaded for scanning -- no second load. Added `getCredentialFromTable()` for this.
- Added TODO for the builder workaround and prefix assertion in the test.

Conflicts with main are resolved.

laserninja:
## Verification: All Requested Changes ✓

Excellent follow-up! The latest revision addresses all the HIGH and MEDIUM issues raised in the previous review:

### Issues Fixed

**[HIGH] Bypasses dispatcher layer**
- Credential vending now routes through `IcebergTableOperationExecutor.planTableScan()` 
- Calls `getCredentialPrivilege()` and passes parameters to `CatalogWrapperForREST.planTableScan(...)`
- `IcebergCatalogWrapperManager` field removed from `IcebergTableOperations`
- Events and authorization hooks will now fire correctly

**[HIGH] Broad `catch (Exception e)`**  
- Narrowed to only catch `ServiceUnavailableException` in `injectScanCredentials()`
- Matches the pattern in `getTableCredentials()`
- Auth failures, illegal args, and other errors now propagate properly

**[HIGH] Redundant `loadTable` call**
- `injectScanCredentials()` receives the `Table` already loaded for scan planning
- No additional `loadTable` invocation
- Credential generation uses `getCredentialFromTable()` which extracts location/properties from the Table

**[MEDIUM] JSON-level merge fragility**
- Replaced ad-hoc `ObjectNode`/`ArrayNode` JSON construction with `PlanTableScanResponse.builder()...withCredentials(...)`
- TODO comment notes future cleanup when Iceberg 1.11.0+ is adopted
- Much cleaner and more maintainable

**[MEDIUM] Test coverage for prefix & refresh properties**
- Tests now assert `prefix` field is present and equals table location
- `IcebergRESTUtils.buildRefreshProps()` and `toRESTCredential()` ensure refresh endpoints are included
- Consistent with `getTableCredentials` pattern

### 📝 Minor Items

**Duplicate Javadoc on no-arg `planTableScan` overload**  
The original method's full Javadoc still appears above the new `/** Plan table scan without credential vending. */` comment, resulting in two stacked `/** */` blocks. Consider removing the old one since it now applies only to the 4-arg overload (which has its own comprehensive Javadoc).

**FederatedCatalogWrapper doesn't override `planTableScan`**  
`FederatedCatalogWrapper` still doesn't override the new 4-arg `planTableScan()` method. When federated catalogs call this, they'll use the local credential manager instead of proxying to the remote catalog. However:
- The graceful fallback (`ServiceUnavailableException` → no credentials) prevents breakage
- This could be addressed in a follow-up PR if federated credential vending on scans becomes important

LGTM with one minor cleanup (duplicate Javadoc).

sachinnn99:
@laserninja Thanks for the thorough verification! Removed the duplicate Javadoc on the no-arg overload in the latest push. Ready for approval when you get a chance.

sachinnn99:
Hey @roryqi, I've addressed all of laserninja's feedback and resolved the conflicts. Could you please take a look? Thanks!

laserninja (approved):
LGTM!

roryqi:
@lasdf1234 Could u help me review this pull request?

lasdf1234:
> @lasdf1234 Could u help me review this pull request?

Got.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
lasdf1234, line 438: ScanPlanCacheKey does not include requestCredentialVending, and we cache the full PlanTableScanResponse (with or without credentials). This can cause:

1、Plan without header → plan with vended-credentials: cache hit returns a response without storage-credentials.
2、 Plan with header → plan without header: cache hit may return stale/extra credentials.
3、Cached credentials may expire while the scan plan remains valid.


Suggestion: Cache only credential-free scan plans. Inject credentials after cache lookup on every request (hit or miss), similar to how loadTable generates credentials per request:

lasdf1234:
@sachinnn99 Another issue is that you need to consider the scenario of a federation IRC.

sachinnn99:
@lasdf1234 Addressed in the latest push.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
sachinnn99: Thanks for the review! Addressed — credentials are injected after cache lookup, not cached. Covered by `testPlanTableScanCacheDoesNotLeakCredentials`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/FederatedCatalogWrapper.java
```diff
@@ -153,6 +155,49 @@ public LoadCredentialsResponse getTableCredentials(
...
+          .withSpecsById(response.specsById())
+          .withErrorResponse(response.errorResponse())
+          .withCredentials(rewritten.credentials())
+          .build();
+    } catch (Exception e) {
+      LOG.warn("Failed to fetch remote credentials for scan of table: {}", tableIdentifier, e);
+      return response;
+    }
+  }
+
```
lasdf1234, line 190: No need for try catch,Could you change the code to the following format?
@SuppressWarnings("deprecation")
@Override
public PlanTableScanResponse planTableScan(
    TableIdentifier tableIdentifier,
    PlanTableScanRequest scanRequest,
    boolean requestCredentialVending,
    CredentialPrivilege privilege) {
  PlanTableScanResponse response =
      super.planTableScan(tableIdentifier, scanRequest, false, privilege);
  if (!requestCredentialVending) {
    return response;
  }
  LoadCredentialsResponse credentials = getTableCredentials(tableIdentifier, privilege);
  if (credentials.credentials().isEmpty()) {
    return response;
  }
  return PlanTableScanResponse.builder()
      .withPlanStatus(response.planStatus())
      .withPlanId(response.planId())
      .withPlanTasks(response.planTasks())
      .withFileScanTasks(response.fileScanTasks())
      .withSpecsById(response.specsById())
      .withErrorResponse(response.errorResponse())
      .withCredentials(credentials.credentials())
      .build();
}

lasdf1234:
@sachinnn99 There is only one issue(No need for try catch) that needs to be revised. There are no other issues.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/FederatedCatalogWrapper.java
sachinnn99: Done, updated to reuse `getTableCredentials()` and removed the try-catch.

lasdf1234:
@roryqi I have no more comment.You can take some time to have a look at this.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/FederatedCatalogWrapper.java
```diff
@@ -153,6 +155,43 @@ public LoadCredentialsResponse getTableCredentials(
...
+   * @return the scan response, optionally enriched with remote credentials.
+   */
+  @SuppressWarnings("deprecation")
+  @Override
+  public PlanTableScanResponse planTableScan(
+      TableIdentifier tableIdentifier,
+      PlanTableScanRequest scanRequest,
+      boolean requestCredentialVending,
+      CredentialPrivilege privilege) {
+    PlanTableScanResponse response =
```
roryqi, line 180: For  FederatedCatalog, you should let the remote catalog to plan the table scan instead of planning the table in the local server.

roryqi:
For Federated catalog, you would better test the interface by hand and paste the HTTP response result.

sachinnn99:
@roryqi Tested federated IRC by hand (`catalog-backend = rest` pointing at an upstream IRC with S3TokenCredential provider):

**With credential vending:**
```json
{
    "status": "completed",
    "storage-credentials": [
        {
            "prefix": "s3://test-bucket/warehouse/db/test_tbl",
            "config": {
                "s3.session-token-expires-at-ms": "1783865183269",
                "s3.access-key-id": "AKIA_DUMMY_TEST_KEY",
                "s3.secret-access-key": "dummy-secret-access-key-12345",
                "s3.session-token": "dummy-session-token-xyz",
                "client.refresh-credentials-endpoint": "v1/federated/namespaces/db/tables/test_tbl/credentials"
            }
        }
    ],
    "file-scan-tasks": []
}
```

**Without header:** `{"status": "completed", "file-scan-tasks": []}` — no credentials returned.

Scan planning delegates to the remote catalog; refresh endpoint is rewritten to this IRC instance.

roryqi (approved):
@sachinnn99 Thanks for your contribution! LGTM.

roryqi:
@lasdf1234 @laserninja Thanks for your review, too.

PR #10813, opened 2026-04-19
[#10218] Improvement(server): Fix misleading success log when metalake drop returns false

### LINKED ISSUE #10216: [Improvement] Avoid success drop/purge log when partition deletion returns false

PartitionOperations.dropPartition logs a success INFO message ("Partition {} {} in table ..."), even when the dispatcher returns false (meaning no partition was dropped/purged).
The method already detects this case and logs a warning, but then immediately emits a success log unconditionally.

### How should we improve?

Only emit the INFO success log when dropped is true.

### LINKED ISSUE #10218: [Improvement] Fix misleading success log when metalake drop returns false

dropMetalake logs a success message ("Metalake dropped: {}") even when the drop operation returns false. The endpoint response correctly reports dropped=false, but the unconditional success log can create misleading signals and hide real failures during troubleshooting.

### How should we improve?

Make success logging conditional on the drop result. Keep the existing warning (or use an error/warn) when dropped is false, and only emit "Metalake dropped" when dropped is true.

gauravrudragit:

  The success log statement in drop/purge REST handlers for Metalake, Table, Fileset, Topic, and Partition was unconditionally emitted after returning Utils.ok(new DropResponse(dropped)), regardless of whether the operation actually succeeded. This PR restructures the logging in all four handlers so that:                                                                                                                                  
                                                                                                                                                                                                                    
  - An INFO log is emitted only when dropped == true (the entity was actually removed).                                                                                                                               
  - A WARN log is emitted when dropped == false (the entity was not found).                 
  - The intermediate Response response variable is eliminated, returning Utils.ok(...) inline.                                                                                                                        
                                                                                                                                                                                                                      
  Files changed: MetalakeOperations.java, FilesetOperations.java, PartitionOperations.java, TableOperations.java, TopicOperations.java.  

  Previously, the success log (e.g. "Fileset dropped: ...") was always printed after a drop request — even when the dispatcher returned false, meaning the entity did not exist and nothing was actually dropped. This produced misleading log output where operators would see a success message alongside a warning, making it hard to diagnose whether an entity was truly removed. The fix ensures log messages accurately reflect the outcome of the operation.

 No. This is a server-side logging fix only.

fix: #10218 
fix: #10216

 Existing unit and integration tests cover the drop path. No new behavior was introduced, so no new tests were added.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/FilesetOperations.java
@@ -337,13 +337,13 @@ public class FilesetOperations {
           () -> {
             NameIdentifier ident = NameIdentifierUtil.ofFileset(metalake, catalog, schema, fileset);
             boolean dropped = dispatcher.dropFileset(ident);
-            if (!dropped) {
+            if (dropped) {
+              LOG.info("Fileset dropped: {}.{}.{}.{}", metalake, catalog, schema, fileset);
+            } else {
               LOG.warn("Cannot find to be dropped fileset {} under schema {}", fileset, schema);
             }
 
-            Response response = Utils.ok(new DropResponse(dropped));
-            LOG.info("Fileset dropped: {}.{}.{}.{}", metalake, catalog, schema, fileset);
-            return response;
+            return Utils.ok(new DropResponse(dropped));
           });
 
     } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


jerryshao:
Can you fix all the related issues in this PR?

yuqi1129:
This PR seems to be duplicated with #10815

gauravrudragit:
@jerryshao @yuqi1129 I found the same issue in 5 files MetalakeOperations.java, FilesetOperations.java, PartitionOperations.java, TableOperations.java, TopicOperations.java and have updated this PR to address all of them, this should now resolve both #10218 and #10216

yuqi1129 (approved):

PR #10814, opened 2026-04-20
[#10271] fix(core): Roll back submitted job when persistence fails

### LINKED ISSUE #10271: [Improvement] Handle runJob partial-failure by rolling back submitted execution when job metadata persistence fails

JobManager.runJob submits the job to jobExecutor before persisting JobEntity in entityStore. If entityStore.put(...) throws IOException, the method throws, but the submitted job remains queued/running. Because lifecycle reconciliation and cleanup operate on persisted JobEntity records, this leaves an orphaned execution that cannot be tracked or managed through normal APIs.

### How should we improve?

Add compensation logic in runJob: if submitJob succeeds but entityStore.put fails, attempt jobExecutor.cancelJob(jobExecutionId) before rethrowing. Preserve original failure semantics (throw the persistence failure) while logging rollback success/failure. Optionally attempt staging-directory cleanup in this failure path as best effort. This makes submit+persist effectively atomic from the caller’s perspective and prevents untracked running jobs.

Here a test to help:
```
@Test
public void testRunJobShouldCancelSubmittedJobWhenStorePutFails() throws IOException {
  mockedMetalake
      .when(() -> MetalakeManager.checkMetalake(metalakeIdent, entityStore))
      .thenAnswer(a -> null);

  JobTemplateEntity shellJobTemplate =
      newShellJobTemplateEntity("shell_job", "A shell job template");
  when(jobManager.getJobTemplate(metalake, shellJobTemplate.name())).thenReturn(shellJobTemplate);

  String jobExecutionId = "job_execution_id_for_test";
  when(jobExecutor.submitJob(any())).thenReturn(jobExecutionId);
  doThrow(new IOException("Entity store error"))
      .when(entityStore)
      .put(any(JobEntity.class), anyBoolean());

  Assertions.assertThrows(
      RuntimeException.class, () -> jobManager.runJob(metalake, "shell_job", Collections.emptyMap()));

  verify(jobExecutor, times(1)).cancelJob(jobExecutionId);
}
```

arjnklc:

Added compensation logic in `JobManager.runJob` to cancel submitted jobs when entity persistence fails:
- When `entityStore.put()` throws `IOException` after a successful `jobExecutor.submitJob()`, the method now calls `jobExecutor.cancelJob(jobExecutionId)` to roll back the orphaned execution.
- Logs both successful and failed rollback attempts.
- Best-effort cleanup of the staging directory via `FileUtils.deleteDirectory()`.
- The original `RuntimeException` is still re-thrown, preserving existing failure semantics.

`JobManager.runJob` submits jobs to `jobExecutor` before persisting `JobEntity` in `entityStore`. If the persistence operation throws `IOException`, the submitted job continues running as an orphaned execution that cannot be tracked or managed. This change makes submit+persist effectively atomic from the caller's perspective.

Fix: #10271

No. This is an internal robustness improvement. The API behavior remains the same, the caller still receives a `RuntimeException` on persistence failure.

Updated existing unit test in `TestJobManager.testRunJob` to verify rollback behavior:
```
./gradlew :core:test --tests "org.apache.gravitino.job.TestJobManager" -PskipITs
```
The test verifies that when `entityStore.put()` fails, `jobExecutor.cancelJob()` is invoked exactly once with the correct execution ID.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/JobManager.java
+++ b/core/src/main/java/org/apache/gravitino/job/JobManager.java
@@ -464,6 +464,31 @@ public class JobManager implements JobOperationDispatcher {
     try {
       entityStore.put(jobEntity, false /* overwrite */);
     } catch (IOException e) {
+      // Attempt to cancel the submitted job to avoid orphaned executions
+      try {
+        jobExecutor.cancelJob(jobExecutionId);
+        LOG.info(
+            "Successfully cancelled submitted job {} after persistence failure", jobExecutionId);
+      } catch (Exception cancelException) {
+        LOG.warn(
+            "Failed to cancel submitted job {} after persistence failure",
+            jobExecutionId,
+            cancelException);
+      }
+
+      // Best-effort cleanup of staging directory
+      try {
+        FileUtils.deleteDirectory(jobStagingDir);
+        LOG.info(
+            "Successfully cleaned up staging directory {} after persistence failure",
+            jobStagingDir);
+      } catch (IOException cleanupException) {
+        LOG.warn(
+            "Failed to clean up staging directory {} after persistence failure",
+            jobStagingDir,
+            cleanupException);
+      }
+
       throw new RuntimeException("Failed to register the job entity " + jobEntity, e);
     }
 
--- a/core/src/test/java/org/apache/gravitino/job/TestJobManager.java
+++ b/core/src/test/java/org/apache/gravitino/job/TestJobManager.java
@@ -500,14 +500,19 @@ public class TestJobManager {
         RuntimeException.class,
         () -> jobManager.runJob(metalake, "shell_job", Collections.emptyMap()));
 
-    // Test when entity store fails
+    // Test when entity store fails, the submitted job should be cancelled
+    String rollbackExecutionId = "job_execution_id_for_rollback";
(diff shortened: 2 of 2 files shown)


yuqi1129:
@arjnklc 

You may need to resolve the CI problem.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -464,6 +464,31 @@ public JobEntity runJob(String metalake, String jobTemplateName, Map<String, Str
     try {
       entityStore.put(jobEntity, false /* overwrite */);
     } catch (IOException e) {
+      // Attempt to cancel the submitted job to avoid orphaned executions
+      try {
+        jobExecutor.cancelJob(jobExecutionId);
```
jerryshao, line 469: There will be an issue to cancel an untracked job here. JobManager will fail to find the job and update the status.

arjnklc: Thanks for the review,

I thought using jobExecutor instead of jobManager here to cancel the job bypasses the entity store, since the executor tracks the job internally from the moment submitJob returns. Should I drop this cancellation and just keep the staging directory cleanup?

PR #10815, opened 2026-04-20
[#10216] improvement(server): Avoid success drop/purge log when partition deletion returns false

### LINKED ISSUE #10216: [Improvement] Avoid success drop/purge log when partition deletion returns false

arjnklc:

Fixed logging in `PartitionOperations.dropPartition` to only emit the success log when the partition is actually dropped/purged:
- Moved the success `LOG.info` inside an `else` block so it only fires when `dropped == true`.
- Improved the warning log to include the operation type (`drop`/`purge`) instead of hardcoding "drop".
- Removed an unnecessary intermediate `response` variable.

The success `LOG.info` message was emitted unconditionally, even when the deletion returned `false`. This caused misleading log output where a warning about a failed drop was immediately followed by a success message for the same operation.

Fix: #10216

No. This is an internal logging improvement. The REST API response is unchanged.

Ran existing unit tests to verify no regressions:
```
./gradlew :server:test --tests "org.apache.gravitino.server.web.rest.TestPartitionOperations" -PskipITs
```
All tests pass.

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/PartitionOperations.java
@@ -247,21 +247,22 @@ public class PartitionOperations {
                     : dispatcher.dropPartition(tableIdent, partition);
             if (!dropped) {
               LOG.warn(
-                  "Failed to drop partition {} under table {} under schema {}",
+                  "Failed to {} partition {} under table {} under schema {}",
+                  purge ? "purge" : "drop",
                   partition,
                   table,
                   schema);
+            } else {
+              LOG.info(
+                  "Partition {} {} in table {}.{}.{}.{}",
+                  partition,
+                  purge ? "purged" : "dropped",
+                  metalake,
+                  catalog,
+                  schema,
+                  table);
             }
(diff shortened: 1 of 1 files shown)


jerryshao (approved):

yuqi1129:
@arjnklc 
Can you take time to check whether there exist similar problems exist in other classes? I noticed that `TableOperations` also share the same problem.

gauravrudragit:
@yuqi1129 #10813 PR has the same fix as this PR, this should be cancelled to avoid merge conflicts

yuqi1129:
> @yuqi1129 #10813 PR has the same fix as this PR, this should be cancelled to avoid merge conflicts

Yeah, you can work on other issues if you are interested.

yuqi1129:
Close it as it has been fixed in [#10813](https://github.com/apache/gravitino/pull/10813)

PR #10826, opened 2026-04-21
[#10825] fix(spark-connector): yield to Spark DataSource shortcut for built-in format names

### LINKED ISSUE #10825: [Bug report] Spark DataSource shortcut syntax fails when GravitinoSparkPlugin is enabled
### Version

main branch

### Describe what's wrong

When `GravitinoSparkPlugin` is registered in a Spark session, Spark SQL's built-in DataSource shortcut syntax — `SELECT * FROM <format>.\`<path>\`` — fails with a server-side `IllegalArgumentException` instead of falling back to Spark's DataSource resolver.

The shortcut applies to all built-in Spark formats (`parquet`, `csv`, `json`, `orc`, `text`, `avro`, `binaryFile`). Spark parses `<format>.\`<path>\`` as a multipart identifier whose first part is the format name. The Gravitino Spark Connector's `BaseCatalog` intercepts the lookup and forwards it to the Gravitino server. The server-side authorization filter then calls `MetadataObjects.of(TABLE, names)` which requires `names.length == 3` (catalog.schema.table); the multipart identifier has only 2 parts, so the check throws.

Before the plugin is registered, the same query works in vanilla Spark because the analyzer's DataSource shortcut resolver handles the format name directly.

\`\`\`
Authorization failed due to system internal error. Please contact administrator.
java.lang.IllegalArgumentException: If the type is TABLE, the length of names must be 3
    at com.google.common.base.Preconditions.checkArgument(Preconditions.java:143)
    at org.apache.gravitino.MetadataObjects.of(MetadataObjects.java:116)
    at org.apache.gravitino.MetadataObjects.parse(MetadataObjects.java:184)
    at org.apache.gravitino.MetadataObjects.of(MetadataObjects.java:93)
    at org.apache.gravitino.utils.NameIdentifierUtil.toMetadataObject(NameIdentifierUtil.java:628)
    at org.apache.gravitino.server.authorization.expression.AuthorizationExpressionEvaluator...
    ...
    at org.apache.gravitino.spark.connector.catalog.BaseCatalog.loadGravitinoTable(BaseCatalog.java:437)
    at org.apache.gravitino.spark.connector.catalog.BaseCatalog.loadTable(BaseCatalog.java:245)
    at org.apache.spark.sql.connector.catalog.CatalogV2Util\$.getTable(CatalogV2Util.scala:363)
    at org.apache.spark.sql.catalyst.analysis.Analyzer\$ResolveRelations\$...
\`\`\`

### How to reproduce

1. Gravitino server (main branch), a metalake with authorization enabled.
2. Spark 3.5 session with:
   \`\`\`
   spark.plugins=org.apache.gravitino.spark.connector.plugin.GravitinoSparkPlugin
   spark.sql.gravitino.uri=<server-uri>
   spark.sql.gravitino.metalake=<metalake>
   \`\`\`
3. Any file accessible by Spark (local, s3a, gvfs, hdfs, etc.).
4. Run:
   \`\`\`sql
   SELECT * FROM parquet.\`s3a://bucket/file.parquet\`;
   -- or csv, json, orc, text, avro — same failure
   \`\`\`

Expected: Spark reads the file via \`HadoopFsRelation\`.

Actual: \`IllegalArgumentException\` "If the type is TABLE, the length of names must be 3".

### Additional context

The shortcut syntax is a long-standing public Spark API (since 2.0). Workarounds today are: (a) \`CREATE TEMPORARY VIEW ... USING <format> OPTIONS (path=...)\`, or (b) use the DataFrame API \`spark.read.format(...).load(path)\`. Both bypass the Gravitino catalog registration and succeed.

Discovered against Gravitino 1.2.0 in a Spark Connect + Kyuubi shared engine setup using GVFS paths, but the bug is format-agnostic and path-agnostic — any built-in format + any path reproduces it.

gada121982:

Short-circuit `BaseCatalog.loadTable` and `BaseCatalog.tableExists` with `NoSuchTableException` when the identifier's namespace is a single built-in Spark DataSource format name (`parquet`, `csv`, `json`, `orc`, `text`, `avro`, `binaryfile`). When the connector declines these lookups, the Spark analyzer falls back to its DataSource shortcut resolver and builds a `HadoopFsRelation` directly — the same path vanilla Spark takes.

Unqualified Gravitino catalogs/tables are unaffected: they arrive with a namespace that references a registered catalog, not a format name, so they continue through the existing `loadGravitinoTable` path and the usual authorization filter.

`SELECT * FROM parquet.\`path\`` is a long-standing Spark built-in syntax. Today, when `GravitinoSparkPlugin` is registered, the analyzer sends `(namespace=[parquet], name=<path>)` to `BaseCatalog.loadTable`, which forwards it to the server. The server-side authorization filter calls `MetadataObjects.of(TABLE, names)` which requires `names.length == 3` (catalog.schema.table), so every such query fails with `IllegalArgumentException: If the type is TABLE, the length of names must be 3`.

Users expect Spark's built-in shortcut to keep working when the connector is installed; today it does not.

Fix: #10825

Yes. After this PR, the following queries work when \`GravitinoSparkPlugin\` is enabled:

\`\`\`sql
SELECT * FROM parquet.\`s3a://bucket/file.parquet\`;
SELECT * FROM csv.\`/data/file.csv\`;
SELECT * FROM json.\`hdfs:///events.json\`;
SELECT * FROM orc.\`/data/file.orc\`;
SELECT * FROM text.\`/data/file.txt\`;
SELECT * FROM avro.\`/data/file.avro\`;
SELECT * FROM binaryFile.\`/data/folder/\`;
\`\`\`

Previously all of these failed at the server-side 3-part-name assertion.

No API or property-key changes. No behavior change for user Gravitino tables.

- New \`TestBaseCatalog\` covering six cases for the new \`isBuiltinDataSourceReference\` helper: built-in formats recognized, case-insensitive match, regular schema namespaces ignored, multipart namespaces ignored, empty namespaces ignored, unknown formats ignored.
- \`./gradlew :spark-connector:spark-common:test --tests org.apache.gravitino.spark.connector.catalog.TestBaseCatalog\` → all tests pass.
- \`./gradlew :spark-connector:spark-common:spotlessCheck\` → clean.
- End-to-end manual verification on Spark 3.5.8 + Gravitino 1.2.0 with GVFS: \`SELECT * FROM parquet.\`gvfs://fileset/.../file.parquet\`\` now returns rows; \`SELECT * FROM <gravitino_iceberg_cat>.db.t\` remains unaffected.

### DIFF
--- a/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
+++ b/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
@@ -19,10 +19,13 @@
 
 package org.apache.gravitino.spark.connector.catalog;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.commons.lang3.StringUtils;
@@ -239,8 +242,25 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
     }
   }
 
+  /**
+   * Spark built-in DataSource format names that can appear as the first part of a multipart
+   * identifier when a user writes {@code SELECT * FROM <format>.`<path>`}. When this catalog
+   * intercepts such a lookup, the server-side authorization filter would call {@code
+   * MetadataObjects.of(TABLE, names)} which requires {@code names.length == 3} and throws an {@code
+   * IllegalArgumentException} because the path is counted as a single name.
+   *
+   * <p>Short-circuiting those lookups with {@code NoSuchTableException} makes the Spark analyzer
+   * fall back to its DataSource shortcut resolver and build a {@code HadoopFsRelation} directly —
+   * the same path vanilla Spark takes when no CatalogPlugin is installed.
+   */
+  private static final ImmutableSet<String> BUILTIN_DATASOURCE_FORMATS =
+      ImmutableSet.of("parquet", "csv", "json", "orc", "text", "avro", "binaryfile");
+
   @Override
   public Table loadTable(Identifier ident) throws NoSuchTableException {
+    if (isBuiltinDataSourceReference(ident)) {
+      throw new NoSuchTableException(ident);
+    }
     try {
       org.apache.gravitino.rel.Table gravitinoTable = loadGravitinoTable(ident);
       org.apache.spark.sql.connector.catalog.Table sparkTable = loadSparkTable(ident);
@@ -256,6 +276,15 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
     } catch (org.apache.gravitino.exceptions.NoSuchTableException e) {
       throw new NoSuchTableException(ident);
     }
+  }
+
+  @VisibleForTesting
+  static boolean isBuiltinDataSourceReference(Identifier ident) {
+    String[] namespace = ident.namespace();
+    if (namespace.length != 1) {
+      return false;
+    }
+    return BUILTIN_DATASOURCE_FORMATS.contains(namespace[0].toLowerCase(Locale.ROOT));
   }
 
   @Override
@@ -303,6 +332,13 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
 
   @Override
   public boolean tableExists(Identifier ident) {
+    // Spark's DataSource shortcut resolver asks the CatalogManager whether the identifier exists
+    // before falling back to format-based resolution. Answer "no" so the authorization filter for
+    // TABLE (which wants 3 name parts) never fires on a namespace that is actually a DataSource
+    // format name such as `parquet`, `csv`, or `json`.
+    if (isBuiltinDataSourceReference(ident)) {
+      return false;
+    }
     // Gravitino uses loadTable() to verify table existence, which requires LOAD_TABLE privilege.
     // For CREATE TABLE IF NOT EXISTS operations, users may only have CREATE_TABLE privilege.
     // When ForbiddenException is thrown (lacking LOAD_TABLE privilege), we return false to allow
--- a/spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/catalog/TestBaseCatalog.java
+++ b/spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/catalog/TestBaseCatalog.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.spark.connector.catalog;
+
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class TestBaseCatalog {
+
+  @Test
+  void builtinDataSourceFormatsShouldBeRecognized() {
+    String[] formats = {"parquet", "csv", "json", "orc", "text", "avro", "binaryFile"};
+    for (String format : formats) {
+      Identifier ident = Identifier.of(new String[] {format}, "s3a://bucket/file");
+      Assertions.assertTrue(
+          BaseCatalog.isBuiltinDataSourceReference(ident),
+          "Expected " + format + " to be recognized as a built-in DataSource format");
+    }
+  }
+
+  @Test
+  void builtinDataSourceFormatsAreCaseInsensitive() {
+    Identifier upper = Identifier.of(new String[] {"PARQUET"}, "s3a://bucket/file.parquet");
+    Identifier mixed = Identifier.of(new String[] {"BinaryFile"}, "s3a://bucket/folder");
+    Assertions.assertTrue(BaseCatalog.isBuiltinDataSourceReference(upper));
+    Assertions.assertTrue(BaseCatalog.isBuiltinDataSourceReference(mixed));
+  }
+
+  @Test
+  void regularGravitinoTableNamespacesAreNotTreatedAsBuiltin() {
+    // A Gravitino table reference typically has a schema-name namespace, not a format name.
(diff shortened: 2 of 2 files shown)


jerryshao:
@diqiu50 @FANNG1 would you please help to review?

FANNG1:
Thanks for the fix. Could you please confirm the exact trigger condition?

From Spark analyzer order, `ResolveRelations` runs before `ResolveSQLOnFile`, so this issue seems to happen only when the current/default catalog is a Gravitino catalog. If the session stays on `spark_catalog` and only has `GravitinoSparkPlugin` registered, `SELECT * FROM parquet.`path`` should still fall through to Spark built-in SQL-on-file resolver and work normally.

Could you verify both cases?

1. Register `GravitinoSparkPlugin`, but do not `USE` a Gravitino catalog / do not set `spark.sql.defaultCatalog` to a Gravitino catalog. Does `SELECT * FROM parquet.`path`` work?
2. Switch to a Gravitino catalog, or set `spark.sql.defaultCatalog` to one. Is this when the `MetadataObjects.of(TABLE, ...)` / `length of names must be 3` error appears?

This distinction matters because the current guard treats any identifier with namespace `parquet`, `csv`, `json`, etc. as a DataSource shortcut. That may incorrectly hide a real Gravitino table under a schema named `parquet`, for example `gravitino_catalog.parquet.t`.

gada121982:
Thanks for the careful review, @FANNG1. I traced through the Spark source and Gravitino source again to confirm your questions, and the picture is slightly different from what I wrote in the PR description. Let me clarify.

## 1. Trigger condition — confirmed

You are correct. The bug only reproduces when the current/default catalog is a Gravitino catalog. Specifically:

**Case 1 — `spark_catalog` is the current catalog, plugin registered but no `USE <gravitino_cat>`**: `SELECT * FROM parquet.`path`` works normally.

- `ResolveRelations` → `expandIdentifier([parquet, path])` → `CatalogAndIdentifier(spark_catalog, Identifier([parquet], path))`.
- `CatalogV2Util.loadTable` calls `V2SessionCatalog.loadTable` → V1 `SessionCatalog` throws `NoSuchDatabaseException`, which `CatalogV2Util.loadTable` swallows ([`CatalogV2Util.scala#L331-L342`](https://github.com/apache/spark/blob/v3.5.3/sql/catalyst/src/main/scala/org/apache/spark/sql/connector/catalog/CatalogV2Util.scala#L331-L342)).
- `ResolveRelations` returns `u` unresolved → `ResolveSQLOnFile` picks it up and builds `HadoopFsRelation`.

**Case 2 — current catalog is a Gravitino catalog** (via `USE` or `spark.sql.defaultCatalog`): same query fails.

- `expandIdentifier` resolves to `CatalogAndIdentifier(gravitino_cat, Identifier([parquet], path))`.
- `BaseCatalog.loadTable` → `loadGravitinoTable` → REST call to the Gravitino server.
- Server throws `IllegalArgumentException` (details in §2), which is **not** one of the three exceptions `CatalogV2Util.loadTable` catches → propagates out → query fails.

So the bug is real, but only on the Case 2 path. In practice every production deployment that uses `GravitinoSparkPlugin` sets `spark.sql.defaultCatalog` to a Gravitino catalog, so users hit this consistently — but you are right that the PR description overstates the scope by not distinguishing the two cases.

## 2. Root cause — narrower than the PR description

The PR description attributes the failure to ``MetadataObjects.of(TABLE, names)`` in the authorization filter. After re-reading the code, the more precise description is:

[`MetadataObjects.parse(fullName, type)`](https://github.com/apache/gravitino/blob/main/api/src/main/java/org/apache/gravitino/MetadataObjects.java#L175-L185) splits `fullName` on `.` with no escape mechanism:

```java
List<String> parts = DOT_SPLITTER.splitToList(fullName);
...
return MetadataObjects.of(parts, type);  // expects parts.size() == 3 for TABLE
```

When Spark sends `SELECT * FROM parquet.`s3a://bucket/file.parquet``, the table name passed to the server is `s3a://bucket/file.parquet` — which contains dots. The server composes the fullName as `catalog.parquet.s3a://bucket/file.parquet`, splits it, gets 4 parts, and the length check in `MetadataObjects.of(List, Type)` throws `IllegalArgumentException`.

This isn't limited to the authorization filter — it is a property of `MetadataObject`'s name representation itself.

## 3. False-positive — acknowledged

You are right that this PR's guard would incorrectly hide a real Gravitino table whose schema is literally named `parquet`, `csv`, etc. For example:

```sql
CREATE SCHEMA gravitino_iceberg_cat.parquet;
CREATE TABLE gravitino_iceberg_cat.parquet.users (id INT);
USE gravitino_iceberg_cat;
SELECT * FROM parquet.users;  -- would be silently short-circuited
```

In Gravitino this is a semantically valid naming choice, and it's a reasonable pattern in data lake conventions (`raw_json`, `staging_csv`, or `parquet` as a landing zone). Silent-wrong is worse than loud-fail, so this is a blocker for the current approach as-is.

## 4. Alternative fixes — looking for maintainer preference

Given the above, I see four candidate directions, listed narrowest to broadest. Happy to pivot to whichever one the maintainers prefer.

### Option A — Narrow the client-side guard to path-like names

Replace the `BUILTIN_DATASOURCE_FORMATS` whitelist with a check on `ident.name()` for path-like characters (`/`, `:`, `\`). A Gravitino-valid table name cannot contain these on any mainstream catalog backend (Iceberg, Hive, JDBC all reject them at DDL time), and every SQL-on-File path does. Keeps scope tiny, eliminates the false-positive for `schema = parquet/csv/json/...`, and also covers DataSource formats Spark may add in the future (e.g. `delta.`path``) without maintaining a list.

Trade-off: still a client-side symptom fix; a Gravitino user who somehow did manage to create a table name containing `/` or `:` (none of the catalog backends allow this) would see a "table not found" instead of an `IllegalArgumentException`. Net improvement either way.

### Option B — Convert IAE to NoSuchTableException in the client

In `BaseCatalog.loadGravitinoTable` / `tableExists`, catch `IllegalArgumentException` (or a specific subclass like `IllegalNamespaceException`) and rethrow as `NoSuchTableException`. Relies purely on the server's existing behavior and lets Spark's normal `ResolveSQLOnFile` fallback take over.

Trade-off: catching a broad exception type is a Java anti-pattern; it could mask programming errors that happen to throw IAE for unrelated reasons. A message-substring match (`"length of names must be"`) would be tighter but fragile across Gravitino versions.

### Option C — Server-side: throw NoSuchTableException instead of IAE

Fix at the source. In `MetadataObjects.parse` (or at the authorization expression evaluator's call site), when the expected length doesn't match for a three-level type, throw something that clients naturally map to "not found" — e.g. `NoSuchEntityException` or `IllegalArgumentException` → `NoSuchTableException` mapping in `ErrorHandlers`.

Rationale: an identifier that can't be expressed as a `(catalog, schema, table)` triple semantically means "that entity does not exist in Gravitino", not "the caller made a programming error". All consumers (Spark, Trino, Flink, CLI) would benefit without a per-connector workaround.

Trade-off: server change, wider blast radius, needs maintainer discussion on whether changing the exception class is considered a contract change.

### Option D — Redesign MetadataObject name representation

Support escaping `.` in names, or store names as `List<String>` internally instead of joining on `.`. This would be the only fix that also allows Gravitino users to create tables named `user.profile` etc., but it touches the authorization model, persistence layer, REST URL encoding, CLI, docs, and existing data. Out of scope for this PR — mentioning for completeness.

## 5. Recommendation and current state

I'd suggest **Option A** for this PR (minimal change, no false-positive, covers future formats) with **Option C** tracked as a follow-up issue if maintainers agree it's the right long-term direction.

I'll hold off on force-pushing until I hear which direction you and @diqiu50 prefer. Happy to add integration tests covering all three scenarios (`default=spark_catalog`, `default=gravitino_cat`, and the `schema=parquet` false-positive) in the same PR once we converge.

FANNG1:
Thanks for the clarification.  I think this is a Spark bug,  and the `parquet.xxx` should not passed to datasource v2 catalogs.   I would suggest to add the check in Gravitino spark connector side to skip it.

Issue #10713, opened 2026-04-08
[Improvement] Ensure consistent event ordering for create operations by standardizing dispatcher wiring


mehakmeet:
### What would you like to be improved?

Currently, the dispatcher chain for most object types (Catalog, Schema, Table, Fileset, Topic, Model, Tag, Metalake, Job) wraps the HookDispatcher inside the EventDispatcher:

`EventDispatcher → NormalizeDispatcher → HookDispatcher → Manager`
The HookDispatcher auto-sets the creator as the object owner after creation. Because the EventDispatcher is the outermost layer, the SET_OWNER event (triggered by the hook) is published before the `CREATE` event. This results in non-causal event ordering -- For our external sync service, the consumers see `SET_OWNER` for an object that hasn't been "created" yet from the event stream's perspective.

In contrast, Policy already use the correct pattern where HookDispatcher wraps EventDispatcher:

`HookDispatcher → EventDispatcher → Manager`
This guarantees CREATE events are published before `SET_OWNER` events.

Additionally, most HookDispatcher implementations (except `CatalogHookDispatcher`) lack rollback logic. If setOwner fails after a successful create, the object is left orphaned with no owner and no cleanup.

### How should we improve?

Standardize dispatcher wiring in GravitinoEnv.java so that all object types follow the `Hook(Event(Normalize(Manager)))` pattern, matching the existing Policy/AccessControl wiring. This ensures `CREATE` events are always published before `SET_OWNER` events.

Add rollback logic to `HookDispatcher` implementations that lack it (Schema, Table, Fileset, Topic, Model, Tag, AccessControl, Job), following the pattern already established in `CatalogHookDispatcher` -- if setOwner fails after creation, drop the object and propagate the exception.

These changes affect `GravitinoEnv.java` (wiring order) and the individual *HookDispatcher classes (rollback). No interface changes are required.

mehakmeet:
I will be working on this issue and raise a PR here soon after testing.

mehakmeet:
CC: @bharos @roryqi @Abhijeetsng97

roryqi:
It is a little dangerous to delete the schema because we fail to set owner for the schema. We rollback catalog because the catalog is managed fully by the Gravitino.

roryqi:
If we use Manager -> EventManager -> HookManger and fail to set the owner, it will cause the creation failure, but we receive the successful creation event.

mehakmeet:
[...]

PR #10834, opened 2026-04-21
[#10713] improvement(core,iceberg): Fix event ordering by standardizing dispatcher wiring

mehakmeet:
### **What changes were proposed in this pull request?**
Standardized dispatcher wiring so that all object types follow the Hook(Event(Normalize(Manager))) pattern, matching the existing Policy/AccessControl wiring. This ensures CREATE events are published before SET_OWNER events for all object types.

GravitinoEnv: Flipped wiring for Metalake, Catalog, Schema, Table, Fileset, Topic, Model, Tag, Job dispatchers.
RESTService: Flipped wiring for Iceberg Table, View, Namespace dispatchers.
HookDispatchers: Ensured setOwner failures are gracefully handled via try-catch-log so that ownership failure does not cause creation failure.

### **Why are the changes needed?**
Currently, most object types wrap HookDispatcher inside EventDispatcher (Event(Hook(Manager))). Since HookDispatcher auto-sets the creator as owner after creation, the SET_OWNER event is published before the CREATE event. This results in non-causal event ordering — external consumers see SET_OWNER for an object that hasn't been "created" yet from the event stream's perspective.

Policy already uses the correct pattern where HookDispatcher wraps EventDispatcher. This PR standardizes all object types to match.

**Fix**: #10713

### **Does this PR introduce any user-facing change?**
No. This is an internal event ordering fix. The API behavior is unchanged. Events are now published in causal order (CREATE before SET_OWNER).

### **How was this patch tested?**
Added setOwner-failure unit tests for all affected HookDispatchers to verify that creation succeeds even when setOwner throws:

Core (10 tests): TestMetalakeHookDispatcher, TestSchemaHookDispatcher, TestTableHookDispatcher, TestFilesetHookDispatcher, TestTopicHookDispatcher, TestModelHookDispatcher, TestTagHookDispatcher, TestPolicyHookDispatcher, TestAccessControlHookDispatcher, TestJobHookDispatcher

Iceberg (3 tests): TestIcebergTableHookDispatcher, TestIcebergViewHookDispatcher, TestIcebergNamespaceHookDispatcher

Tested in Gravitino dev environment. All core and iceberg tests pass:

`./gradlew :core:test :iceberg:iceberg-rest-server:test -PskipITs`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
+++ b/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
@@ -550,39 +550,43 @@ public class GravitinoEnv {
     this.lockManager = new LockManager(config);
 
     // Create and initialize metalake related modules, the operation chain is:
-    // MetalakeEventDispatcher -> MetalakeNormalizeDispatcher -> MetalakeHookDispatcher ->
+    // MetalakeHookDispatcher -> MetalakeNormalizeDispatcher -> MetalakeEventDispatcher ->
     // MetalakeManager
     this.metalakeManager = new MetalakeManager(entityStore, idGenerator);
-    MetalakeHookDispatcher metalakeHookDispatcher = new MetalakeHookDispatcher(metalakeManager);
+    MetalakeEventDispatcher metalakeEventDispatcher =
+        new MetalakeEventDispatcher(eventBus, metalakeManager);
     MetalakeNormalizeDispatcher metalakeNormalizeDispatcher =
-        new MetalakeNormalizeDispatcher(metalakeHookDispatcher);
-    this.metalakeDispatcher = new MetalakeEventDispatcher(eventBus, metalakeNormalizeDispatcher);
+        new MetalakeNormalizeDispatcher(metalakeEventDispatcher);
+    this.metalakeDispatcher = new MetalakeHookDispatcher(metalakeNormalizeDispatcher);
 
     // Create and initialize Catalog related modules, the operation chain is:
-    // CatalogEventDispatcher -> CatalogNormalizeDispatcher -> CatalogHookDispatcher ->
+    // CatalogHookDispatcher -> CatalogNormalizeDispatcher -> CatalogEventDispatcher ->
     // CatalogManager
     this.catalogManager = new CatalogManager(config, entityStore, idGenerator);
-    CatalogHookDispatcher catalogHookDispatcher = new CatalogHookDispatcher(catalogManager);
+    CatalogEventDispatcher catalogEventDispatcher =
+        new CatalogEventDispatcher(eventBus, catalogManager);
     CatalogNormalizeDispatcher catalogNormalizeDispatcher =
-        new CatalogNormalizeDispatcher(catalogHookDispatcher);
-    this.catalogDispatcher = new CatalogEventDispatcher(eventBus, catalogNormalizeDispatcher);
+        new CatalogNormalizeDispatcher(catalogEventDispatcher);
+    this.catalogDispatcher = new CatalogHookDispatcher(catalogNormalizeDispatcher);
 
     this.credentialOperationDispatcher =
         new CredentialOperationDispatcher(catalogManager, entityStore, idGenerator);
 
     SchemaOperationDispatcher schemaOperationDispatcher =
         new SchemaOperationDispatcher(catalogManager, entityStore, idGenerator);
-    SchemaHookDispatcher schemaHookDispatcher = new SchemaHookDispatcher(schemaOperationDispatcher);
+    SchemaEventDispatcher schemaEventDispatcher =
+        new SchemaEventDispatcher(eventBus, schemaOperationDispatcher);
     SchemaNormalizeDispatcher schemaNormalizeDispatcher =
-        new SchemaNormalizeDispatcher(schemaHookDispatcher, catalogManager);
-    this.schemaDispatcher = new SchemaEventDispatcher(eventBus, schemaNormalizeDispatcher);
+        new SchemaNormalizeDispatcher(schemaEventDispatcher, catalogManager);
+    this.schemaDispatcher = new SchemaHookDispatcher(schemaNormalizeDispatcher);
 
     TableOperationDispatcher tableOperationDispatcher =
         new TableOperationDispatcher(catalogManager, entityStore, idGenerator);
-    TableHookDispatcher tableHookDispatcher = new TableHookDispatcher(tableOperationDispatcher);
     TableNormalizeDispatcher tableNormalizeDispatcher =
-        new TableNormalizeDispatcher(tableHookDispatcher, catalogManager);
-    this.tableDispatcher = new TableEventDispatcher(eventBus, tableNormalizeDispatcher);
+        new TableNormalizeDispatcher(tableOperationDispatcher, catalogManager);
+    TableEventDispatcher tableEventDispatcher =
+        new TableEventDispatcher(eventBus, tableNormalizeDispatcher);
+    this.tableDispatcher = new TableHookDispatcher(tableEventDispatcher);
 
     // TODO: We can install hooks when we need, we only supports ownership post hook,
     //  partition doesn't have ownership, so we don't need it now.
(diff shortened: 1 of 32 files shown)


core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -593,29 +597,32 @@ private void initGravitinoServerComponents() {
 
     FilesetOperationDispatcher filesetOperationDispatcher =
         new FilesetOperationDispatcher(catalogManager, entityStore, idGenerator);
-    FilesetHookDispatcher filesetHookDispatcher =
-        new FilesetHookDispatcher(filesetOperationDispatcher);
     FilesetNormalizeDispatcher filesetNormalizeDispatcher =
-        new FilesetNormalizeDispatcher(filesetHookDispatcher, catalogManager);
-    this.filesetDispatcher = new FilesetEventDispatcher(eventBus, filesetNormalizeDispatcher);
+        new FilesetNormalizeDispatcher(filesetOperationDispatcher, catalogManager);
```
bharos, line 601: The PR description claims all types follow Hook(Event(Normalize(Manager))), but the actual wiring is:


Pattern | Entity Types
-- | --
Hook → Normalize → Event → Manager | Metalake, Catalog, Schema
Hook → Event → Normalize → Manager | Table, Fileset, Topic, Model


Both achieve CREATE before SET_OWNER, so this doesn't break the PR's goal. However, the position of NormalizeDispatcher relative to EventDispatcher affects whether events see pre-normalization or post-normalization input data. This internal inconsistency could cause subtle surprises later.

We can use consistent pattern if that's possible

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergNamespaceHookDispatcher.java
```diff
@@ -59,13 +64,20 @@ public CreateNamespaceResponse createNamespace(
...
-    IcebergOwnershipUtils.setSchemaOwner(
-        metalake,
-        context.catalogName(),
-        createRequest.namespace(),
-        context.userName(),
-        GravitinoEnv.getInstance().ownerDispatcher());
+    try {
+      importSchema(context.catalogName(), createRequest.namespace());
+      IcebergOwnershipUtils.setSchemaOwner(
+          metalake,
```
bharos, line 70: The try-catch wraps both importSchema and setSchemaOwner. If importSchema fails, the namespace exists in the external catalog but is never registered in Gravitino's entity store, and the failure is silently logged with a misleading "set owner" message. These should be separate try-catch blocks, or importSchema failure should propagate.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergNamespaceHookDispatcher.java
```diff
@@ -122,17 +134,24 @@ public LoadTableResponse registerTable(
...
-        namespace,
-        registerTableRequest.name(),
-        context.userName(),
-        GravitinoEnv.getInstance().ownerDispatcher());
+    try {
+      // Import the registered table into Gravitino's catalog so it exists as a metadata object
+      importTable(context.catalogName(), namespace, registerTableRequest.name());
+
+      // Set the owner of the registered table to the current user
+      IcebergOwnershipUtils.setTableOwner(
```
bharos, line 142: Similar comment as IcebergNamespaceHookDispatcher.createNamespace, please check if we should fix these

core/src/main/java/org/apache/gravitino/hook/MetalakeHookDispatcher.java
```diff
@@ -71,14 +74,18 @@ public Metalake createMetalake(
       accessControlDispatcher.addUser(ident.name(), PrincipalUtils.getCurrentUserName());
     }
```
bharos, line 75: addUser is NOT wrapped in try-catch — can still cause 500 error
Do we need similar treatment like setOwner here ?

bharos: This is a pre-existing issue, not introduced by this PR, but is now more visible since the event ordering change surfaces it.

core/src/main/java/org/apache/gravitino/hook/MetalakeHookDispatcher.java
```diff
@@ -71,14 +74,18 @@ public Metalake createMetalake(
...
+      OwnerDispatcher ownerDispatcher = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerDispatcher != null) {
+        ownerDispatcher.setOwner(
+            ident.name(),
+            NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.METALAKE),
+            PrincipalUtils.getCurrentUserName(),
+            Owner.Type.USER);
+      }
+    } catch (Exception e) {
+      LOG.warn("Fail to set owner for metalake {}, metalake exists without owner", ident, e);
```
bharos, line 88: nit : "Fail to set owner" → "Failed to set owner" on all the messages?

bharos (commented):
Overall +ve, some non-blocking comments to consider

core/src/main/java/org/apache/gravitino/hook/TableHookDispatcher.java
```diff
@@ -79,13 +82,17 @@ public Table createTable(
...
-          ident.namespace().level(0),
-          NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.TABLE),
-          PrincipalUtils.getCurrentUserName(),
-          Owner.Type.USER);
+    try {
+      OwnerDispatcher ownerManager = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerManager != null) {
+        ownerManager.setOwner(
+            ident.namespace().level(0),
+            NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.TABLE),
```
mchades: With the new dispatcher chain order (`HookDispatcher → EventDispatcher → NormalizeDispatcher → Manager`), `ident` here is the raw caller input — it has not yet passed through `NormalizeDispatcher`. For TABLE scope, both the leaf name and the schema name inside `ident.namespace()` can be normalized, but the entity is stored under the normalized names. This `setOwner` call may therefore silently fail (exception swallowed by the `catch` block).

core/src/main/java/org/apache/gravitino/hook/SchemaHookDispatcher.java
```diff
@@ -60,13 +63,17 @@ public Schema createSchema(NameIdentifier ident, String comment, Map<String, Str
...
-          ident.namespace().level(0),
-          NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.SCHEMA),
-          PrincipalUtils.getCurrentUserName(),
-          Owner.Type.USER);
+    try {
+      OwnerDispatcher ownerManager = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerManager != null) {
+        ownerManager.setOwner(
+            ident.namespace().level(0),
+            NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.SCHEMA),
```
mchades: Same normalization issue as in `TableHookDispatcher#createTable`. For SCHEMA scope, only the leaf (schema) name is affected — the namespace `metalake.catalog` is not normalized by `NormalizeDispatcher`.

core/src/main/java/org/apache/gravitino/hook/FilesetHookDispatcher.java
```diff
@@ -80,13 +83,17 @@ public Fileset createMultipleLocationFileset(
...
-          ident.namespace().level(0),
-          NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.FILESET),
-          PrincipalUtils.getCurrentUserName(),
-          Owner.Type.USER);
+    try {
+      OwnerDispatcher ownerManager = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerManager != null) {
+        ownerManager.setOwner(
+            ident.namespace().level(0),
+            NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.FILESET),
```
mchades: Same normalization issue as in `TableHookDispatcher#createTable`.

core/src/main/java/org/apache/gravitino/hook/TopicHookDispatcher.java
```diff
@@ -66,13 +69,17 @@ public Topic createTopic(
...
-          ident.namespace().level(0),
-          NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.TOPIC),
-          PrincipalUtils.getCurrentUserName(),
-          Owner.Type.USER);
+    try {
+      OwnerDispatcher ownerManager = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerManager != null) {
+        ownerManager.setOwner(
+            ident.namespace().level(0),
+            NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.TOPIC),
```
mchades: Same normalization issue as in `TableHookDispatcher#createTable`.

core/src/main/java/org/apache/gravitino/hook/ModelHookDispatcher.java
```diff
@@ -69,13 +72,17 @@ public Model registerModel(NameIdentifier ident, String comment, Map<String, Str
...
-          ident.namespace().level(0),
-          NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.MODEL),
-          PrincipalUtils.getCurrentUserName(),
-          Owner.Type.USER);
+    try {
+      OwnerDispatcher ownerManager = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerManager != null) {
+        ownerManager.setOwner(
+            ident.namespace().level(0),
+            NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.MODEL),
```
mchades: Same normalization issue as in `TableHookDispatcher#createTable`.

core/src/main/java/org/apache/gravitino/hook/ModelHookDispatcher.java
```diff
@@ -157,13 +164,17 @@ public Model registerModel(
...
-      ownerManager.setOwner(
-          ident.name(),
-          NameIdentifierUtil.toMetadataObject(ident, Entity.EntityType.MODEL),
-          PrincipalUtils.getCurrentUserName(),
-          Owner.Type.USER);
+    try {
+      OwnerDispatcher ownerManager = GravitinoEnv.getInstance().ownerDispatcher();
+      if (ownerManager != null) {
+        ownerManager.setOwner(
+            ident.name(),
```
mchades: (Pre-existing, unrelated to this PR) The first argument to `setOwner` should be the metalake name (`ident.namespace().level(0)`), not the model name (`ident.name()`).

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergNamespaceHookDispatcher.java
mehakmeet: Yea, I was initially thinking that it's required for the set owner call only so wrapped them in one try catch, but probably better to propagate this exception as the entity won't be existing in Gravitino if this fails.

core/src/main/java/org/apache/gravitino/hook/MetalakeHookDispatcher.java
mehakmeet: Yea, addUser seemed like a different type of exception so I let it be.
We have 2 cases of exception `UserAlreadyExistsException` then we can let the setOwner call be as is, other exceptions means the user wasn't able to be added to the metalake so setOwner will fail regardless, we can raise a warning here too and return. User can be added afterwards as well since the metalake creation has already happened.

core/src/main/java/org/apache/gravitino/hook/TableHookDispatcher.java
mehakmeet: Good catch!

mehakmeet:
@bharos and @mchades I understood the normalization issue clearly now. I have come up with a fix, please tell me if this makes sense:

Wiring | Manager stores | Event sees | Hook sees | setOwner uses
-- | -- | -- | -- | --
OLD (pre-PR): Event → Normalize → Hook → Manager | normalized | raw | normalized | normalized (correct)
Current PR (Cat/Sch/Meta): Hook → Normalize → Event → Manager | normalized | normalized | raw | raw (BUG)
Current PR (Tbl/Fset/Top/Mod): Hook → Event → Normalize → Manager | normalized | raw | raw | raw (BUG)
Option A (rewire): Normalize → Hook → Event → Manager | normalized | normalized (changed) | normalized | normalized (correct)

Does Wrapping NormalizeDispatcher outermost makes sense? I think it will still have some gaps like the events also seeing the normalized identifier now. 
Also I see one more issue in this, that if normalization fails due to invalid naming we won't be able to raise a failure event from the event dispatcher as it's the outermost layer. 
But, we can have a manual normalization solution which can fill these gaps:

Wiring | Manager stores | Event sees | Hook sees | setOwner uses
-- | -- | -- | -- | --
Option B (manual): Hook → Event → Normalize → Manager + duplicate normalize logic in Hook for setOwner | normalized | raw (preserves OLD) | raw | normalized via manual applyCapabilities (correct)

core/src/main/java/org/apache/gravitino/hook/ModelHookDispatcher.java
```diff
@@ -69,13 +74,23 @@ public Model registerModel(NameIdentifier ident, String comment, Map<String, Str
...
+            CapabilityHelpers.applyCapabilities(
+                ident, Capability.Scope.MODEL, GravitinoEnv.getInstance().catalogManager());
+        ownerManager.setOwner(
+            normalizedIdent.namespace().level(0),
+            NameIdentifierUtil.toMetadataObject(normalizedIdent, Entity.EntityType.MODEL),
+            PrincipalUtils.getCurrentUserName(),
+            Owner.Type.USER);
+      }
+    } catch (Exception e) {
+      LOG.warn("Failed to set owner for model {}, model exists without owner", ident, e);
```
mchades: @roryqi is this behavior make sense?

mchades (dismissed):
LGTM. only one minor comment. see https://github.com/apache/gravitino/pull/10834/changes#r3165714793

core/src/main/java/org/apache/gravitino/hook/ModelHookDispatcher.java
roryqi: I prefer not catching the exceptions. We can modify this place after we have a discussion conclusion.

mehakmeet: Ideally, we should propagate some sort of warning or response that isn't similar to full success but that requires changes to the Response classes and for Iceberg rest server it's even trickier. This we could handle in a separate PR?
 
But, we should also not propagate an exception as the entity gets created but the owner setting fails, if the user sees an error for this and tries to recreate the same entity they will see entityAlreadyExists exception.

roryqi: He should drop the incomplete entity and recreate the entity.

mehakmeet: Yes, agreed.

Is this good for now for this PR? Something we can think about in the followup on how to propagate different response for partial failure, failure and success?

roryqi: Could you remove `try catch` for the setting owners?

mehakmeet: Done

roryqi (approved):

roryqi:
I am ok for current change. @mchades Do you have other suggestion?

mchades (approved):
Thanks for your contributions! @mehakmeet

roryqi:
@mehakmeet Because we use the internal dispatcher for setting owner now. https://github.com/apache/gravitino/pull/12893 , we won't produce a set owner event. I will revert the dispatcher order. Because it make more sense after setting owner successfully.

PR #10838, opened 2026-04-21
[#10837] fix(web-v2): Preserve hidden properties when editing a catalog

### LINKED ISSUE #10837: [Bug report] web-v2: Editing a catalog removes hidden properties (e.g. warehouse), causing 400 error
### Version

main branch

### Describe what's wrong

In the web-v2 UI, editing a catalog that has optional (non-required) properties like warehouse causes those properties to be silently removed. This results in a 400 error from the backend when warehouse is required for the catalog to function correctly.

The root cause is in CreateCatalogDialog.js. During edit mode, the isHidden() function's default case hides all non-required properties that aren't region or location. The getSubmitData() function then filters out hidden props via .filter(item => !isHidden(item)), so they are excluded from the submitted data. When genUpdates() compares the original catalog properties against the submitted properties, it generates a removeProperty("warehouse") update, which the backend rejects with a 400 error.

Backend returns HTTP 400 when attempting to save the edited catalog. The warehouse property is removed from the update payload even though the user did not change or remove it.

### How to reproduce

How to reproduce:

Start Gravitino server and web-v2 UI on main branch
Create a lakehouse-iceberg catalog with hive backend, setting uri and warehouse properties
Click edit on the catalog
Change only the comment field (or any visible field)
Click submit
Observe 400 error — the warehouse property was removed from the update request

### Additional context

https://github.com/user-attachments/assets/a1ecbf84-5a7e-4a55-8e44-82db2b16ca03

bharos:

When editing a catalog in the web-v2 UI, hidden `defaultProps` (properties that `isHidden()` returns `true` for) are now backfilled from the cached catalog data before computing the update diff. This prevents `genUpdates()` from generating spurious `removeProperty` operations for properties the user never touched.

During catalog edit, `isHidden()` hides all non-required, non-region/location default properties (e.g. `warehouse` for lakehouse-iceberg with hive backend). `getSubmitData()` then filters these out via `.filter(item => !isHidden(item))`, so they are absent from the submitted payload. `genUpdates()` sees these keys in the original catalog but not in the new data, and generates `removeProperty` updates — which causes the backend to return a 400 error.

The fix adds a targeted backfill step in `handleSubmit` that iterates over all `defaultProps`, and for each one that is hidden and missing from the submit data, copies the original value from `cacheData.properties`.

Fix: #10837

No API changes. Catalog edit in web-v2 now correctly preserves properties that are not displayed in the edit form.

Manually tested by:
1. Creating a lakehouse-iceberg catalog (hive backend) with warehouse and uri
2. Editing the catalog (changing only the comment)
3. Verifying the edit succeeds without a 400 error
4. Verifying the warehouse property is preserved after edit

### DIFF
--- a/web-v2/web/src/app/catalogs/rightContent/CreateCatalogDialog.js
+++ b/web-v2/web/src/app/catalogs/rightContent/CreateCatalogDialog.js
@@ -294,6 +294,17 @@ export default function CreateCatalogDialog({ ...props }) {
         setConfirmLoading(true)
         const submitData = getSubmitData()
         if (editCatalog) {
+          // Backfill hidden default props that the form doesn't render during edit
+          const allDefaultProps =
+            (catalogType === 'model'
+              ? providerBase['model'].defaultProps
+              : providerBase[currentProvider]?.defaultProps) || []
+          allDefaultProps.forEach(prop => {
+            if (isHidden(prop) && !(prop.key in submitData.properties) && cacheData?.properties?.[prop.key] != null) {
+              submitData.properties[prop.key] = cacheData.properties[prop.key]
+            }
+          })
+
           // update catalog
           const reqData = { updates: genUpdates(cacheData, submitData) }
           if (reqData.updates.length) {


LauraXia123 (approved):
LGTM

PR #10840, opened 2026-04-22
[#10839] feat(server): Add health check endpoints (liveness, readiness, aggregate)

### LINKED ISSUE #10839: [Improvement] Add health check endpoints (liveness, readiness, aggregate)
## Problem

Gravitino does not expose standard health check endpoints. Modern Java services — including Apache Polaris (via Quarkus), Spring Boot applications, and Micronaut applications — ship liveness, readiness, and aggregate health endpoints out of the box. Gravitino runs on raw Jetty + Jersey and has none.

This creates friction on day one of any enterprise deployment:

- **Kubernetes probes.** `livenessProbe` and `readinessProbe` expect HTTP endpoints that distinguish "restart this pod" from "take it out of rotation." The current `/api/version` endpoint conflates the two and, more importantly, doesn't actually verify Gravitino's ability to serve traffic.
- **Load balancers and GTMs.** Standard enterprise traffic managers poll a configurable health endpoint and route based on HTTP status. Some GTM onboarding standards also hardcode well-known paths like `/health` or `/health.html`.
- **APM tooling.** Datadog, New Relic, and similar platforms auto-discover health endpoints following MicroProfile or Spring conventions.
- **Bakeoff checklists.** Competitive evaluations mark Gravitino down on "operational readiness" because health endpoints are considered table stakes.

## Proposal

Add three endpoints under `/api/health`, following [MicroProfile Health](https://microprofile.io/specifications/microprofile-health/) semantics (the same pattern Quarkus, Spring Boot, and Polaris use), plus two root-level aliases for compatibility with enterprise GTM onboarding standards:

| Endpoint | Purpose | Healthy | Unhealthy |
|---|---|---|---|
| `GET /api/health/live` | Liveness — is the HTTP thread able to respond? | 200 | (no response / timeout → K8s restarts pod) |
| `GET /api/health/ready` | Readiness — is the entity store reachable? | 200 | 503 (LB/GTM routes away) |
| `GET /api/health` | Aggregate of live + ready, for consumers expecting a single URL | 200 | 503 |
| `GET /health`, `GET /health.html` | Root-level aliases forwarding to `/api/health`, for GTMs that hardcode well-known paths | 200 | 503 |

All `/api/health*` endpoints return a JSON body describing the individual checks:

```json
{
  "code": 0,
  "status": "UP",
  "checks": [
    { "name": "httpServer", "status": "UP", "details": {} },
    { "name": "entityStore", "status": "UP", "details": {} }
  ]
}
```

## Design notes

- **Liveness does not touch downstream systems.** If the process is deadlocked or OOM, the probe fails by not returning. This is the K8s-recommended pattern.
- **Readiness probes the entity store** via a bounded-timeout call to `EntityStore.exists()` against a sentinel identifier. The call exercises the JDBC connection without requiring any entity to be present.
- **Failure details are minimal by design.** We report the exception class name, not full stack traces, to avoid leaking internal state over an unauthenticated endpoint.
- **Endpoints must be reachable without authentication** for K8s probes and external GTMs to work. The `AuthenticationFilter` exempts `/api/health*`; root-level aliases are outside the `/api/*` path spec the filter is registered on.

## Out of scope for this issue

- Per-catalog backend health (belongs in a separate `/api/health/detailed` or `/api/catalogs/{name}/health` — catalog outages should not pull Gravitino itself out of rotation).
- Prometheus-formatted `/metrics` endpoint (separate effort).
- Startup probes (K8s 1.16+ — can be added if a slow-start subsystem is identified).
- Dedicated management port (Polaris runs health/metrics on port 8182 separately from API on 8181; worth considering as a follow-up).

markhoerth:

Adds three health check endpoints following MicroProfile Health semantics:

- `GET /api/health/live` — liveness, returns 200 when the HTTP thread is responsive
- `GET /api/health/ready` — readiness, returns 200 when the entity store is reachable, 503 otherwise
- `GET /api/health` — aggregate, returns 200 when both checks pass
- `GET /health`, `GET /health/live`, `GET /health/ready` — root-level aliases forwarding to the corresponding `/api/health/*` endpoints, for enterprise GTM standards that hardcode well-known root paths

Also exempts `/api/health*` and `/health/*` from `AuthenticationFilter` so Kubernetes probes and load balancers can reach the endpoints without credentials under any configured authenticator.

Adds one new optional server configuration key:
- `gravitino.server.health.entityStore.probeTimeoutMs` (default: 2000 ms) — timeout for the entity-store readiness probe used by `/api/health/ready`

Modern Java services (Apache Polaris, Spring Boot, Quarkus, Micronaut) ship these endpoints by default. Gravitino runs on raw Jetty and does not, which blocks standard Kubernetes probe configuration, load balancer health checks, and enterprise GTM integration. This is a parity gap that surfaces on day one of enterprise deployments.

Fix: #10839

Yes — adds three new public REST endpoints under `/api/health`. No existing endpoint behavior is changed.

- New unit tests in `TestHealthOperations` covering liveness, readiness (happy path, entity store uninitialized, entity store throws, timeout), and aggregate status.
- New unit tests in `TestHealthAliasServlet` verifying forward success for `/health`, `/health/live`, `/health/ready` and null-dispatcher 503 path.
- New unit tests in `TestAuthenticationFilter` verifying health paths bypass authentication and non-health paths (including paths that merely contain "health", e.g. `/api/metalakes/health_metalake`) continue to require authentication.
- Manual verification against a running Gravitino instance: all three endpoints return 200 with expected JSON bodies; `/api/version` continues to work unchanged.

### Notes for reviewers

- **Auth filter exemption is hardcoded** to `/api/health*` and `/health/*`, matching how Spring Boot and Quarkus hardcode their well-known health paths. Happy to make this config-driven in a follow-up if preferred.
- **Bounded timeout on entity store probe.** The readiness check runs `EntityStore.exists()` with a configurable ceiling (default 2 s) via a dedicated `ThreadPoolExecutor` to prevent a hanging JDBC connection from tying up Jetty worker threads. `RejectedExecutionException` is caught and returns DOWN so a saturated probe pool never blocks the request thread.
- **Response body format.** `HealthResponse` extends `BaseResponse` and keeps `code: 0` even in 503 responses — the HTTP status is the probe signal, the body is diagnostic. This is intentional and differs from `ErrorResponse` usage.
- **`/api/version` is conventionally also unauthenticated** across most Java services (used for client discovery before auth negotiation). Not changed in this PR but worth considering as a follow-up.

### DIFF
--- a/conf/gravitino.conf.template
+++ b/conf/gravitino.conf.template
@@ -19,7 +19,7 @@
 
 # THE CONFIGURATION FOR Gravitino SERVER
 gravitino.server.shutdown.timeout = 3000
-# Timeout in milliseconds for the entity-store liveness probe used by health endpoints
+# Timeout in milliseconds for the entity-store readiness probe used by /api/health/ready
 # gravitino.server.health.entityStore.probeTimeoutMs = 2000
 
 # THE CONFIGURATION FOR Gravitino WEB SERVER
--- a/conf/gravitino.conf.template
+++ b/conf/gravitino.conf.template
@@ -19,6 +19,8 @@
 
 # THE CONFIGURATION FOR Gravitino SERVER
 gravitino.server.shutdown.timeout = 3000
+# Timeout in milliseconds for the entity-store liveness probe used by health endpoints
+# gravitino.server.health.entityStore.probeTimeoutMs = 2000
 
 # THE CONFIGURATION FOR Gravitino WEB SERVER
 # The host name of the built-in web server
--- a/docs/open-api/health.yaml
+++ b/docs/open-api/health.yaml
@@ -37,7 +37,7 @@ paths:
           $ref: "#/components/responses/HealthResponse"
         "4XX":
           $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
-        "5xx":
+        "5XX":
           $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
 
   /health/live:
@@ -56,7 +56,7 @@ paths:
           $ref: "#/components/responses/HealthResponse"
         "4XX":
           $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
-        "5xx":
+        "5XX":
           $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
 
   /health/ready:
@@ -78,7 +78,7 @@ paths:
           $ref: "#/components/responses/HealthResponse"
         "4XX":
           $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
-        "5xx":
+        "5XX":
           $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
 
 components:
--- a/docs/open-api/health.yaml
+++ b/docs/open-api/health.yaml
@@ -1,0 +1,168 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+---
+
+paths:
+  /health:
+    get:
+      tags:
+        - health
+      summary: Aggregate health
+      operationId: aggregateHealth
+      description: >
+        Returns aggregate health status combining liveness and readiness checks.
+        Returns 200 when all checks pass; 503 when any check fails.
+        This endpoint is exempt from authentication so probes can reach it without credentials.
+      security: []
+      responses:
+        "200":
+          $ref: "#/components/responses/HealthResponse"
+        "503":
+          $ref: "#/components/responses/HealthResponse"
+        "4XX":
+          $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
+        "5xx":
+          $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
+
+  /health/live:
+    get:
+      tags:
+        - health
+      summary: Liveness check
+      operationId: livenessCheck
+      description: >
+        Returns 200 as long as the HTTP server thread can respond.
+        Use this for Kubernetes liveness probes to determine whether to restart a pod.
+        This endpoint is exempt from authentication so probes can reach it without credentials.
+      security: []
+      responses:
+        "200":
+          $ref: "#/components/responses/HealthResponse"
+        "4XX":
+          $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
+        "5xx":
+          $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
+
+  /health/ready:
+    get:
+      tags:
+        - health
+      summary: Readiness check
+      operationId: readinessCheck
+      description: >
+        Returns 200 when the entity store is reachable within the configured probe timeout.
+        Returns 503 when the entity store is unavailable or slow.
+        Use this for Kubernetes readiness probes to control traffic routing.
+        This endpoint is exempt from authentication so probes can reach it without credentials.
+      security: []
+      responses:
+        "200":
+          $ref: "#/components/responses/HealthResponse"
+        "503":
+          $ref: "#/components/responses/HealthResponse"
+        "4XX":
+          $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
+        "5xx":
+          $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
+
+components:
+  responses:
+    HealthResponse:
+      description: Health check result
+      content:
+        application/vnd.gravitino.v1+json:
+          schema:
+            $ref: "#/components/schemas/HealthResponse"
+          examples:
+            HealthUp:
+              $ref: "#/components/examples/HealthUp"
+            HealthDown:
+              $ref: "#/components/examples/HealthDown"
+
+  schemas:
+    HealthResponse:
+      type: object
+      required:
+        - code
+        - status
+        - checks
+      properties:
+        code:
+          type: integer
+          description: Response code, 0 for success
+          example: 0
+        status:
+          type: string
+          enum:
+            - UP
+            - DOWN
+          description: Aggregate health status
+        checks:
+          type: array
+          description: Per-component check results
+          items:
+            $ref: "#/components/schemas/HealthCheck"
+
+    HealthCheck:
+      type: object
+      required:
+        - name
+        - status
+      properties:
+        name:
+          type: string
+          description: Name of the component being checked
+          example: entityStore
+        status:
+          type: string
+          enum:
+            - UP
+            - DOWN
+          description: Status of the component
+        details:
+          type: object
+          additionalProperties:
+            type: string
+          description: Optional diagnostic details about the check result
+
+  examples:
+    HealthUp:
+      value:
+        code: 0
+        status: UP
+        checks:
+          - name: httpServer
+            status: UP
+            details: {}
+          - name: entityStore
+            status: UP
+            details: {}
+
+    HealthDown:
+      value:
+        code: 0
+        status: DOWN
+        checks:
+          - name: httpServer
+            status: UP
+            details: {}
+          - name: entityStore
+            status: DOWN
+            details:
+              reason: timeout
--- a/server-common/src/main/java/org/apache/gravitino/server/ServerConfig.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/ServerConfig.java
@@ -35,6 +35,15 @@ public class ServerConfig extends Config {
           .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
           .createWithDefault(3 * 1000);
 
+  public static final ConfigEntry<Long> HEALTH_ENTITY_STORE_PROBE_TIMEOUT_MS =
+      new ConfigBuilder("gravitino.server.health.entityStore.probeTimeoutMs")
+          .doc(
+              "Timeout in milliseconds for the entity-store liveness probe used by health endpoints")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .longConf()
+          .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(2000L);
+
   public ServerConfig(boolean loadDefaults) {
     super(loadDefaults);
   }
--- a/server-common/src/main/java/org/apache/gravitino/server/ServerConfig.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/ServerConfig.java
@@ -38,7 +38,7 @@ public class ServerConfig extends Config {
   public static final ConfigEntry<Long> HEALTH_ENTITY_STORE_PROBE_TIMEOUT_MS =
       new ConfigBuilder("gravitino.server.health.entityStore.probeTimeoutMs")
           .doc(
-              "Timeout in milliseconds for the entity-store liveness probe used by health endpoints")
+              "Timeout in milliseconds for the entity-store readiness probe used by /api/health/ready")
           .version(ConfigConstants.VERSION_1_3_0)
           .longConf()
           .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
--- a/server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
+++ b/server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
@@ -47,6 +47,7 @@ import org.apache.gravitino.policy.PolicyDispatcher;
 import org.apache.gravitino.server.authentication.ServerAuthenticator;
 import org.apache.gravitino.server.authorization.GravitinoAuthorizerProvider;
 import org.apache.gravitino.server.web.ConfigServlet;
+import org.apache.gravitino.server.web.HealthAliasServlet;
 import org.apache.gravitino.server.web.HttpServerMetricsSource;
 import org.apache.gravitino.server.web.JettyServer;
 import org.apache.gravitino.server.web.JettyServerConfig;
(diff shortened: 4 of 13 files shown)


markhoerth:
Added /health and /health.html as root-level servlet aliases that forward to /api/health. These satisfy enterprise GTM onboarding standards that hardcode well-known root-level health paths (confirmed with an enterprise customer whose GTM standard requires this). The canonical endpoint remains /api/health; aliases are a thin compatibility layer. Root-level paths are outside the /api/* spec the auth filter is registered on, so no additional filter changes needed. All 5 endpoints verified returning 200 with expected JSON bodies on a local build.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
```diff
@@ -0,0 +1,193 @@
...
+                      NameIdentifier.of(HEALTH_PROBE_SENTINEL), EntityType.METALAKE);
+                } catch (IOException e) {
+                  throw new RuntimeException(e);
+                }
+              });
+      future.get(ENTITY_STORE_PROBE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      return up(CHECK_ENTITY_STORE, Collections.emptyMap());
+    } catch (TimeoutException e) {
+      LOG.warn("Entity store probe timed out after {}ms", ENTITY_STORE_PROBE_TIMEOUT_MS);
+      return down(CHECK_ENTITY_STORE, "reason", "timeout");
```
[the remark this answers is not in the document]
jerryshao: Fixed. Added a dedicated `ThreadPoolExecutor` (1 core / 4 max threads, bounded queue of 20) passed explicitly to `supplyAsync`. The future is also hoisted before the `try` block so it can be cancelled in the `TimeoutException` and `InterruptedException` catch clauses.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
```diff
@@ -0,0 +1,193 @@
...
+        cause = cause.getCause();
+      }
+      if (cause == null) {
+        cause = e;
+      }
+      LOG.warn("Entity store probe failed: {}", cause.toString());
+      return down(CHECK_ENTITY_STORE, "reason", cause.getClass().getSimpleName());
+    } catch (Exception e) {
+      LOG.warn("Entity store probe encountered unexpected error", e);
+      return down(CHECK_ENTITY_STORE, "reason", e.getClass().getSimpleName());
```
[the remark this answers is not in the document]
jerryshao: Fixed. Added an explicit `catch (InterruptedException)` block that calls `future.cancel(true)` to stop the background probe and `Thread.currentThread().interrupt()` to restore the interrupt flag before returning DOWN. This ensures Jetty's thread management sees the interrupt after the response is sent.

server/src/main/java/org/apache/gravitino/server/web/HealthAliasServlet.java
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+    dispatcher.forward(req, resp);
+  }
+
+  @Override
+  protected void doHead(HttpServletRequest req, HttpServletResponse resp)
+      throws ServletException, IOException {
+    doGet(req, resp);
+  }
```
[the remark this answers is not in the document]
jerryshao: Fixed. Removed the `doHead` override entirely. `HttpServlet.doHead()` calls `doGet()` internally with a body-suppressing response wrapper, which is the correct behaviour here.

server/src/main/java/org/apache/gravitino/server/web/HealthAliasServlet.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+      throws ServletException, IOException {
+    RequestDispatcher dispatcher = req.getRequestDispatcher(CANONICAL_HEALTH_PATH);
+    if (dispatcher == null) {
+      resp.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "health dispatcher unavailable");
+      return;
+    }
+    dispatcher.forward(req, resp);
+  }
```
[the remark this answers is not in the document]
jerryshao: Fixed. Added `TestHealthAliasServlet` covering two cases: (1) forward success — verifies `RequestDispatcher.forward()` is called with the correct request/response; (2) null dispatcher — verifies `SC_SERVICE_UNAVAILABLE` is returned when `getRequestDispatcher` returns null.

server/src/main/java/org/apache/gravitino/server/GravitinoServer.java
```diff
@@ -175,6 +176,11 @@ protected void configure() {
     server.addServlet(servlet, API_ANY_PATH);
     Servlet configServlet = new ConfigServlet(serverConfig);
     server.addServlet(configServlet, "/configs");
+
+    // Root-level alias for enterprise GTMs that require probes at well-known root paths.
+    // Forwards /health, /health/live, /health/ready to the canonical /api/health/* endpoints.
+    server.addServlet(new HealthAliasServlet(), "/health/*");
```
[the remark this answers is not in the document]
jerryshao: No, I delibrately removed this. I will modify the PR description.

conf/gravitino.conf.template
```diff
@@ -19,6 +19,8 @@
 
 # THE CONFIGURATION FOR Gravitino SERVER
 gravitino.server.shutdown.timeout = 3000
+# Timeout in milliseconds for the entity-store liveness probe used by health endpoints
```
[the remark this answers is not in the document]
jerryshao: Fixed. Updated the comment to read "entity-store readiness probe used by /api/health/ready" to accurately reflect which endpoint this timeout controls.

docs/open-api/health.yaml
```diff
@@ -0,0 +1,168 @@
...
+      responses:
+        "200":
+          $ref: "#/components/responses/HealthResponse"
+        "503":
+          $ref: "#/components/responses/HealthResponse"
+        "4XX":
+          $ref: "./openapi.yaml#/components/responses/BadRequestErrorResponse"
+        "5xx":
+          $ref: "./openapi.yaml#/components/responses/ServerErrorResponse"
+
```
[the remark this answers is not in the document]
jerryshao: Fixed. Changed all three `5xx` keys to `5XX` to match the OpenAPI patterned-field syntax.

server-common/src/main/java/org/apache/gravitino/server/ServerConfig.java
```diff
@@ -35,6 +35,15 @@ public class ServerConfig extends Config {
...
           .createWithDefault(3 * 1000);
 
+  public static final ConfigEntry<Long> HEALTH_ENTITY_STORE_PROBE_TIMEOUT_MS =
+      new ConfigBuilder("gravitino.server.health.entityStore.probeTimeoutMs")
+          .doc(
+              "Timeout in milliseconds for the entity-store liveness probe used by health endpoints")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .longConf()
+          .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(2000L);
```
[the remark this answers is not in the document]
jerryshao: Fixed. Updated the doc string to read "entity-store readiness probe used by /api/health/ready".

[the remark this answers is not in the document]
jerryshao: Acknowledged. The PR description's 'No property keys added or removed' checkbox is incorrect — this PR does add `gravitino.server.health.entityStore.probeTimeoutMs`. I will update the PR description to reflect this.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
```diff
@@ -0,0 +1,239 @@
...
+  @Produces("application/vnd.gravitino.v1+json")
+  @Timed(name = "health." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "health", absolute = true)
+  public Response health() {
+    List<HealthCheckDTO> checks = new ArrayList<>(2);
+    checks.add(up(CHECK_HTTP_SERVER, Collections.emptyMap()));
+    checks.add(checkEntityStore());
+
+    HealthCheckDTO.Status overall =
+        checks.stream().anyMatch(c -> c.getStatus() == HealthCheckDTO.Status.DOWN)
```
yuqi1129, line 138: Since `CHECK_HTTP_SERVER` is always up in L129, why not just check the stauts of entity store?

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
```diff
@@ -0,0 +1,239 @@
...
+    } catch (RejectedExecutionException e) {
+      LOG.warn("Entity store probe rejected — health executor queue full");
+      return down(CHECK_ENTITY_STORE, "reason", "probe-rejected");
+    }
+
+    try {
+      future.get(timeoutMs, TimeUnit.MILLISECONDS);
+      return up(CHECK_ENTITY_STORE, Collections.emptyMap());
+
+    } catch (TimeoutException e) {
```
yuqi1129, line 175: Is there any consideration that we use a separate thread pool to handle health checks?

jerryshao: What't the meaning? You want or you don't want to use a separate thread pool?

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
jerryshao: Yeah, it is correct. But the logic here is more consistent if we add more checks in the future.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
yuqi1129: Since it's already in the Jetty thread pool and will be blocked until the check finishes, I'm not seeing the need for a separate thread to handle it.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
```diff
@@ -0,0 +1,239 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(HealthOperations.class);
+
+  private static final AtomicInteger PROBE_THREAD_COUNTER = new AtomicInteger();
+
+  private static final ExecutorService HEALTH_PROBE_EXECUTOR =
+      new ThreadPoolExecutor(
+          1,
+          4,
+          60L,
```
mchades: Could you please add short comments for these parameters? Or use static variables.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
jerryshao: The benefit here is to have a timeout control so the caller has a presumption that the api will definitely return before that time. Otherwise, like you mentioned, using thread pool doesn't have lots of value.

jerryshao: I don't have a strong opinion on it, what is your preference? @yuqi1129

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
```diff
@@ -0,0 +1,239 @@
...
+          TimeUnit.SECONDS,
+          new LinkedBlockingQueue<>(20),
+          r -> {
+            Thread t = new Thread(r, "health-probe-" + PROBE_THREAD_COUNTER.incrementAndGet());
+            t.setDaemon(true);
+            return t;
+          },
+          new ThreadPoolExecutor.AbortPolicy());
```
mchades, line 92: `HEALTH_PROBE_EXECUTOR` is a static thread pool but is never shut down. Although the threads are marked as daemon threads, the pool itself should be explicitly shut down when the servlet is destroyed to allow in-flight probes to complete gracefully.

Consider overriding `destroy()`:

```java
@Override
public void destroy() {
  HEALTH_PROBE_EXECUTOR.shutdown();
  try {
    if (!HEALTH_PROBE_EXECUTOR.awaitTermination(5, TimeUnit.SECONDS)) {
      HEALTH_PROBE_EXECUTOR.shutdownNow();
    }
  } catch (InterruptedException e) {
    HEALTH_PROBE_EXECUTOR.shutdownNow();
    Thread.currentThread().interrupt();
  }
  super.destroy();
}
```

jerryshao: OK, I see. My feeling is that this is OK, JVM will destroy it when it exits. Let me see how to fix.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
yuqi1129: Yeah, if the intention is to have a timeout mechanism for this API, it's acceptable for me.

jerryshao: OK, let's leave as it is.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
jerryshao: This is the feedback from Claude Code @mchades :

HealthOperations is a JAX-RS resource discovered by Jersey package scanning — it's not registered as a servlet with Jetty via addServlet(). The
  javax.servlet.Servlet.destroy() contract is only honoured by the servlet container (Jetty) for classes it directly manages. Jersey manages JAX-RS resource instances through
  HK2 and does not call destroy() on them.

  So adding destroy() here gives a false sense of lifecycle management — the method would be dead code.

  The right response to the reviewer is to explain this and note that daemon threads are the correct solution: when the JVM exits (or when Jetty stops its thread), daemon
  threads are killed automatically with no leaked resources.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
jerryshao: Fixed. Extracted the four constructor arguments into named static constants: `PROBE_CORE_THREADS`, `PROBE_MAX_THREADS`, `PROBE_KEEP_ALIVE_SECONDS`, and `PROBE_QUEUE_CAPACITY`.

server/src/main/java/org/apache/gravitino/server/web/rest/HealthOperations.java
jerryshao: Thanks for the suggestion. However, `HealthOperations` is a JAX-RS resource discovered by Jersey package scanning — it is not registered as a servlet directly with Jetty via `addServlet()`. The `javax.servlet.Servlet.destroy()` lifecycle method is only called by the servlet container on classes it directly manages; Jersey manages JAX-RS resource instances through HK2 and does not invoke `destroy()` on them. Adding the override would be dead code and give a false sense of lifecycle management. The executor threads are marked as daemon threads, so they are cleaned up automatically when the JVM exits without leaking resources.

mchades (approved):
LGTM

yuqi1129 (approved):

PR #10848, opened 2026-04-22
[#10411] feat(core): Extend metadata object owner to support group

### LINKED ISSUE #10411: [Subtask] Extend Metadata Object Owner to Support Group
### Describe the subtask

In the previous "Set Owner for Group" design, the ownerfield was implicitly assumed to point to a User. In practical scenarios, however, the logical owner of a resource (e.g., a Schema, Table) or even of a Group itself​ is more likely to be a team (Group)​ rather than an individual.

### Parent issue

#10404

bharos:

Remove the USER-only restriction in `OwnerManager.setOwner()` and add GROUP support to `notifyOwnerChange()`, enabling groups to be set as metadata object owners.

Changes:
- Remove `Preconditions.checkArgument(ownerType == Owner.Type.USER)` guard in `OwnerManager.setOwner()`
- Extend `notifyOwnerChange()` to resolve group entity IDs 
- Fix `GroupEntity` namespace in test (was using `ofUserNamespace`, corrected to `ofGroupNamespace`)
- Replace `testGroupTypeOwnerNotSupported` with tests that verify group ownership works

The owner field was implicitly assumed to point to a User. In practice, the logical owner of a resource (Schema, Table, etc.) is often a team (Group) rather than an individual. The downstream layers (REST API, DTOs, relational store) already support GROUP — only `OwnerManager` had a guard blocking it.

Fix: #10411

Yes. The `setOwner` API now accepts `Owner.Type.GROUP` in addition to `Owner.Type.USER`.

Unit tests in `TestOwnerManager`:
- `testGroupTypeOwner`: Sets group as owner, verifies, then replaces with user owner (exercises `notifyOwnerChange` GROUP path)
- `testSetNonExistentGroupAsOwner`: Verifies `NotFoundException` for non-existent group

### DIFF
--- a/core/src/test/java/org/apache/gravitino/authorization/TestOwnerManager.java
+++ b/core/src/test/java/org/apache/gravitino/authorization/TestOwnerManager.java
@@ -67,9 +67,13 @@ import org.apache.gravitino.storage.RandomIdGenerator;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
 import org.mockito.Mockito;
 
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public class TestOwnerManager {
   private static final String JDBC_STORE_PATH =
       "/tmp/gravitino_jdbc_entityStore_" + UUID.randomUUID().toString().replace("-", "");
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/gravitino/authorization/TestOwnerManager.java
```diff
@@ -67,9 +67,13 @@
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
 import org.mockito.Mockito;
 
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
bharos, line 76: The tests share a single in-memory H2 entity store that persists state across methods. testOwner() asserts the metalake has no owner initially, but testGroupTypeOwner() sets an owner on the same metalake. Without ordering, JUnit could run them in any order, causing testOwner() to fail if it runs second.

roryqi (approved):
LGTM.

PR #10863, opened 2026-04-24
[#10862] improvement(ci): Pre-install JDK 24 to stabilize Trino connector build

### LINKED ISSUE #10862: [Improvement] Fix unstable JDK 24 download in CI for Trino connector build
## What would you like to be improved?

Trino connector submodules require Java 24 via Gradle toolchain. CI workflows only install JDK 17, so Gradle auto-downloads JDK 24 via the Foojay resolver plugin at build time. This download is unreliable — it occasionally fails due to network timeouts or rate limits, causing flaky CI failures.

## How should we improve?

Two strategies:
1. **Pre-install JDK 24** in CI via `actions/setup-java@v4` (before JDK 17), then register its path with Gradle via `~/.gradle/gradle.properties`. This eliminates the Foojay download for workflows that must build Trino (5 workflows).
2. **Add `-PskipTrinoConnector` property** — conditionally exclude all Trino subprojects from `settings.gradle.kts`. Used in `gvfs-fuse-build-test.yml` where Trino is not needed.

A shared composite action `.github/actions/setup-java-toolchains` encapsulates strategy 1 to avoid repeating the pattern across workflows.

diqiu50:

- Add reusable composite action `.github/actions/setup-java-toolchains` that pre-installs JDK 24 alongside JDK 17 and registers JDK 24 with Gradle toolchain via `~/.gradle/gradle.properties`, eliminating unreliable Foojay downloads at build time.
- Apply the composite action to 5 workflows that build Trino connector modules: `build.yml`, `trino-integration-test.yml`, `trino-multi-version-test.yml`, `frontend-integration-test.yml`, `cron-integration-test.yml`.
- Add `skipTrinoConnector` property to `settings.gradle.kts` to conditionally exclude all Trino subprojects from the Gradle project graph. Use it in `gvfs-fuse-build-test.yml` which has no Trino dependency.

Fix: #10862

Trino connector submodules require Java 24 via Gradle toolchain. When only JDK 17 is installed in CI, Gradle auto-downloads JDK 24 via the Foojay resolver plugin, which is unreliable and causes flaky CI failures.

No.

CI workflows validation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,11 +106,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-java@v4
+      - uses: ./.github/actions/setup-java-toolchains
         with:
           java-version: 17
-          distribution: 'temurin'
-          cache: 'gradle'
 
       - name: Build with Gradle
         run: |
@@ -164,11 +162,9 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-java@v4
+      - uses: ./.github/actions/setup-java-toolchains
         with:
           java-version: ${{ matrix.java-version }}
-          distribution: 'temurin'
-          cache: 'gradle'
 
       - name: Test publish to local
         run: ./gradlew publishToMavenLocal -PskipWeb=true -x test
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -69,7 +69,9 @@ include("iceberg:iceberg-rest-server")
 include("lance:lance-common")
 include("lance:lance-rest-server")
 include("authorizations:authorization-ranger", "authorizations:authorization-common", "authorizations:authorization-chain")
-if (gradle.startParameter.projectProperties["skipTrinoConnector"]?.toBoolean() != true) {
+val skipTrinoConnector: Boolean =
+  gradle.startParameter.projectProperties["skipTrinoConnector"]?.toBoolean() ?: false
+if (!skipTrinoConnector) {
   include(
     "trino-connector:trino-connector",
     "trino-connector:trino-connector-435-439",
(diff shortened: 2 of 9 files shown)


settings.gradle.kts
```diff
@@ -69,16 +69,20 @@ include("iceberg:iceberg-rest-server")
...
-  "trino-connector:trino-connector",
-  "trino-connector:trino-connector-435-439",
-  "trino-connector:trino-connector-440-445",
-  "trino-connector:trino-connector-446-451",
-  "trino-connector:trino-connector-452-468",
-  "trino-connector:trino-connector-469-472",
-  "trino-connector:trino-connector-473-478",
-  "trino-connector:integration-test"
-)
+if (gradle.startParameter.projectProperties["skipTrinoConnector"]?.toBoolean() != true) {
```
yuqi1129: What's the default value of it?

yuqi1129: Do we need to set this to true in the backend CI pipeline?

diqiu50: false

.github/workflows/build.yml
yuqi1129, line 216: Do you need to add `-PskipTrinoConnector=true` here?

settings.gradle.kts
```diff
@@ -69,16 +69,22 @@ include("iceberg:iceberg-rest-server")
...
-  "trino-connector:trino-connector-440-445",
-  "trino-connector:trino-connector-446-451",
-  "trino-connector:trino-connector-452-468",
-  "trino-connector:trino-connector-469-472",
-  "trino-connector:trino-connector-473-478",
-  "trino-connector:integration-test"
-)
+val skipTrinoConnector: Boolean =
+  gradle.startParameter.projectProperties["skipTrinoConnector"]?.toBoolean() ?: false
+if (!skipTrinoConnector) {
```
yuqi1129, line 74: It's advised to add some description about this parameter in `how-to-build.md`

yuqi1129 (approved):

PR #10867, opened 2026-04-24
[#10412] feat(server-common): Support group ownership in JcasbinAuthorizer

### LINKED ISSUE #10412: [Subtask]  Enhance authorization to evaluate group-based ownership and role membership
### Describe the subtask

Problem Description

Currently, Gravitino's authorization checks are performed primarily by directly matching the user's identity against permissions or role assignments. However, in a practical system with group support, a user's effective permissions should also derive from their group memberships​ in two key aspects:
Group-based Ownership: A user may need to act as the "owner" of a resource not because they are the direct user:owner, but because they belong to a group:that is the resource's owner. The current logic does not evaluate this indirect ownership via group membership.
Role Inheritance via Groups: A role (a collection of privileges) is often granted to a group, and all members of that group should inherit that role's permissions. The current authorization engine likely lacks this indirect evaluation path.
This limitation makes the newly introduced group ownership and group-based role grants ineffective during permission checks, breaking core use cases for team-based management.

Goal

Extend the core authorization logic to evaluate a user's permissions by considering the groups they belong to, specifically for:
Determining if a user effectively "owns" a resource (because their group owns it).
Determining if a user has a specific privilege (because a role granting that privilege is assigned to one of their groups)

### Parent issue

#10404

bharos:

Extend `JcasbinAuthorizer` to recognize group-based ownership. When a metadata object is owned by a group, all members of that group are now treated as owners and granted owner privileges.

Key changes:
- **`OwnerInfo` inner class**: Replaces the raw `Long` owner ID in the cache with a struct that stores `id`, `type` (USER/GROUP), and `name`.
- **`ownerRel` cache type**: Changed from `Cache<Long, Optional<Long>>` to `Cache<Long, Optional<OwnerInfo>>`.
- **`loadOwnerPolicy()`**: Now handles `GroupEntity` alongside `UserEntity` when populating the owner cache.
- **`checkOwnership()`**: New method that resolves both user and group owners — for USER owners it compares entity IDs, for GROUP owners it checks whether the principal's groups include the owning group.
- **`isOwner()` and `authorizeByJcasbin()`**: Refactored to delegate to `checkOwnership()`, eliminating duplicated ownership logic.
- **Documentation**: Removed "group ownership not supported" info boxes from `docs/security/access-control.md` and added group ownership bullet.

Currently, when a metadata object's owner is set to a group (supported since #10848), the `JcasbinAuthorizer` does not recognize group ownership — only individual user ownership is checked. This means group members are denied owner privileges even when the group is the registered owner.

This PR is part 2 of a 3-PR series for #10412:
1. **#10848** — Core/API: `OwnerManager.setOwner` accepts GROUP (**merged**)
2. **This PR** — Enforcement: `JcasbinAuthorizer` recognizes group ownership
3. **Planned** — Role inheritance: `loadRolePrivilege` queries `ROLE_GROUP_REL`

Fix: #10412

Yes. Users who assign group ownership to metadata objects will now have all group members recognized as owners by the JCasbin authorization plugin.

Unit tests in `TestJcasbinAuthorizer`:
- `testAuthorizeByGroupOwner`: Verifies a principal whose groups include the owning group is recognized as owner, a non-member group is denied, and clearing ownership returns false.
- Existing tests (`testIsOwner`, `testOwnerCacheInvalidation`, `testCacheInitialization`) updated for `OwnerInfo` type.
- All 8 tests pass: `./gradlew :server-common:test --tests "...TestJcasbinAuthorizer"`


server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -601,4 +609,45 @@ private void loadPolicyByRoleEntity(RoleEntity roleEntity) {
...
+  private boolean checkOwnership(Principal principal, String metalake, Long metadataId) {
+    Optional<OwnerInfo> ownerOpt = ownerRel.getIfPresent(metadataId);
+    if (ownerOpt == null || !ownerOpt.isPresent()) {
+      return false;
+    }
+    OwnerInfo owner = ownerOpt.get();
+    if (owner.type == Entity.EntityType.USER) {
+      try {
+        UserEntity userEntity = getUserEntity(principal.getName(), metalake);
+        return Objects.equals(userEntity.id(), owner.id);
```
roryqi, line 632: Could we compare the user name and owner name directly to avoid extra IO operations here?

roryqi: We add a user A.
We set a user A to table B's owner.
We remove user A.
We add user A again.
The cache will cache invalid owner data if we use name here.

bharos: Good point. Then in the group name as well, I think I should actually involve an extra IO operation, for same use-case of removing and adding same name for group ?

bharos: @roryqi I added a comment explaining the extra IO and changed the logic for GROUP to also fetch the ID and compare. I hope this is correct

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -601,4 +610,65 @@ private void loadPolicyByRoleEntity(RoleEntity roleEntity) {
...
+      } catch (Exception e) {
+        LOG.debug("Can not get user entity for ownership check", e);
+        return false;
+      }
+    } else if (owner.type == Entity.EntityType.GROUP) {
+      if (principal instanceof UserPrincipal) {
+        for (UserGroup group : ((UserPrincipal) principal).getGroups()) {
+          if (group.getGroupname().equals(owner.name)) {
+            try {
+              GroupEntity groupEntity =
```
roryqi: Could we batchGet group ids?

bharos: Used entityStore.batchGet() for the group ownership check. Note that JDBCBackend.batchGet() for GROUP/USER/ROLE currently loops individually ([see the TODO at JDBCBackend.java L357](https://github.com/apache/gravitino/blob/f4d61784c996c24918b6bab4f27926e77af628a6/core/src/main/java/org/apache/gravitino/storage/relational/JDBCBackend.java#L355)). True SQL-level batch support for these types can be added in a follow-up PR.

roryqi (approved):
LGTM.

Issue #10404, opened 2026-03-12
[EPIC] Support Group-Based Access Control


roryqi:
### Describe the proposal

Problem Description

Currently, Gravitino's authorization system only supports permission checks at the level of an individual user. However, in real-world enterprise deployments, access control is typically managed with user groups (or roles)​ as the core unit. Many users' identity information (e.g., from LDAP/AD) inherently includes group attributes.

Supporting only user-level authorization leads to:
Management Overhead: Permissions must be granted individually to each user within a group, making it impossible to achieve efficient "grant once, apply to all group members" management.
Mismatch with Operational Practices: Administrators prefer to grant permissions to entities like the "developers group," "analysts group," or "ops group" rather than directly to individual employee accounts.
Underutilization of Existing Identity Data: The existing group information from identity sources is not leveraged during authorization.
Goal

Extend Gravitino's authorization model to support permission evaluation based on a user's group memberships, in addition to the existing user-level model. This will enable more flexible and enterprise-ready unified access control.

### Task list

- [ ]

bharos:
Building on [PR #10848](https://github.com/apache/gravitino/pull/10848) (which adds group support in setOwner/getOwner), next I'd like to implement the authorizer-side enforcement so isOwner() recognizes group owners.

**The problem:**
loadOwnerPolicy() only processes UserEntity owners. If a GroupEntity is the owner, the ownerRel cache is never populated and isOwner() always returns false.

**Proposed fix:**
The ownerRel cache currently stores `metadataId → Optional<userId>`. I'd change it to store an `OwnerInfo (id, type, name)` so it can represent both user and group owners. In `isOwner()`, group ownership is resolved by checking the user's JWT groups (UserPrincipal.getGroups()).

I plan to store the `group name` in the cache (alongside id and type) because `UserPrincipal.getGroups()` carries `names` from the JWT token, `NOT Gravitino entity IDs`. This avoids a DB lookup on every isOwner() call. First wanted to confirm if  this approach acceptable ?

cc @roryqi  @yuqi1129 @jerryshao

roryqi:
> Building on [PR #10848](https://github.com/apache/gravitino/pull/10848) (which adds group support in setOwner/getOwner), next I'd like to implement the authorizer-side enforcement so isOwner() recognizes group owners.
> 
> **The problem:** loadOwnerPolicy() only processes UserEntity owners. If a GroupEntity is the owner, the ownerRel cache is never populated and isOwner() always returns false.
> 
[...]

PR #10868, opened 2026-04-24
[#10412] feat(server-common): Support group-inherited roles in JcasbinAuthorizer

### LINKED ISSUE #10412: [Subtask]  Enhance authorization to evaluate group-based ownership and role membership

bharos:

Extend `JcasbinAuthorizer` to support role-based authorization inherited through group membership. This is PR #3 of the [group-based access control epic (#10404)](https://github.com/apache/gravitino/issues/10404).

**Changes:**
1. **`loadRolePrivilege`**: After loading a user's direct roles via `ROLE_USER_REL`, iterate the principal's groups and load each group's assigned roles into the jCasbin enforcers. Roles inherited from groups are treated identically to directly-assigned roles (same enforcer mappings, same async policy loading, same caching).
2. **`isSelf` (ROLE branch)**: After checking direct user-role assignment, also check if the role is assigned to any of the user's groups by loading the `GroupEntity` and checking its `roleIds()`.
3. **`addRoleForUserAndLoadPolicies`**: Extracted helper method to avoid duplication between user-direct and group-inherited role loading paths.
4. **`resolveCurrentUserGroups`**: Extracted helper that batch-resolves the principal's groups via `entityStore.batchGet()`. Stale/deleted groups are silently skipped (batchGet is lenient for GROUP entities since #10933).

**Design decisions:**
- Uses `GroupEntity.roleIds()`/`roleNames()` rather than `ROLE_GROUP_REL` direction in `SupportsRelationOperations` (which only supports role-to-groups, not group-to-roles). This avoids core module changes.
- Groups are batch-fetched via `entityStore.batchGet()`, which silently skips missing entities for GROUP type (lenient behavior added in #10933).
- No cache invalidation changes needed — `handleRolePrivilegeChange(roleId)` already invalidates by roleId regardless of whether the role was user-direct or group-inherited.

Currently, `JcasbinAuthorizer` only loads roles directly assigned to users. If a user belongs to a group that has a role with certain privileges, those privileges are not enforced. This means group-based role management does not work for authorization, which is a core requirement for the group-based access control feature.

Fix: #10412

No API changes. Users who assign roles to groups will now see those role privileges correctly enforced for all members of the group during authorization.

10 new unit tests in `TestJcasbinAuthorizer` (15 total)
All 15 tests pass. Formatted with `spotlessApply`.


server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -504,40 +516,106 @@ private void loadRolePrivilege(
...
+   * after processing both direct and group-inherited roles.
+   */
+  private void addRoleForUserAndLoadPolicies(
+      Long userId,
+      String metalake,
+      Long roleId,
+      String roleName,
+      List<CompletableFuture<Void>> loadRoleFutures,
+      EntityStore entityStore) {
+    allowEnforcer.addRoleForUser(String.valueOf(userId), String.valueOf(roleId));
```
roryqi, line 603: What will it happen if a group remove a role? What will it happen if the role belongs to the user and group at the same time?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -504,40 +516,106 @@ private void loadRolePrivilege(
...
+   */
+  private List<GroupEntity> resolveCurrentUserGroups(String metalake, EntityStore entityStore) {
+    Principal principal = PrincipalUtils.getCurrentPrincipal();
+    List<GroupEntity> result = new ArrayList<>();
+    if (!(principal instanceof UserPrincipal)) {
+      return result;
+    }
+    for (UserGroup group : ((UserPrincipal) principal).getGroups()) {
+      try {
+        NameIdentifier groupIdent = NameIdentifierUtil.ofGroup(metalake, group.getGroupname());
```
roryqi: Could u batch get the group entities?

bharos: @roryqi 
I had a concern for the batchGet call as I had mentioned in the PR description:
```
Groups are fetched individually (not via batchGet) so that a single stale or deleted group does not block checking the remaining ones.
```
So, if there's a group that's present in the auth token, but it's removed from Gravitino entity store, the current logic fails the entire batch get operation.

Of course there's TODO as well to implement "true" batchGet as current implementation does individual GET calls, but that's a separate issue.

For now, I think we can fix the batchGet to at least continue processing in case of stale/non-existent entries so that entire request doesn't fail. I created a separate PR now for this : https://github.com/apache/gravitino/pull/10933

if this looks good and we merge it, I can convert the get calls in current PR to use batchGet

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
bharos: Both cases are handled correctly by the existing invalidation mechanism and verified by two new edge-case tests:

case `Group removes a role`: `revokeRolesFromGroup` triggers `handleRolePrivilegeChange(roleId)` → `loadedRoles.invalidate(roleId)` → the synchronous Caffeine removal listener calls `deleteRole(roleId)` on both enforcers, wiping all g-rows and p-rows for that role. The next `doAuthorize` call re-resolves groups from the DB, finds the role is no longer assigned, and does not reload it. Test: `testGroupRoleRevokedDeniesAccess`.

case `Role belongs to user and group simultaneously`: `addRoleForUserAndLoadPolicies` is called once from the direct-role loop and once from the group-role loop, but `addRoleForUser` and `addPolicy` are both idempotent in jCasbin — duplicate calls are no-ops. When the group-side is revoked, `handleRolePrivilegeChange` wipes the role entirely, but the next authorize call re-resolves both direct roles (via ROLE_USER_REL) and group roles, so the direct assignment re-adds the g-row and reloads the policies. Test:`testRoleSharedByUserAndGroupSurvivesGroupRevocation`.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
bharos: Merged the PR https://github.com/apache/gravitino/pull/10933 and now using batchGet method to get group entities

bharos:
@roryqi Thanks for your review comments. PTAL at the updated PR!

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
roryqi: What will it happen if the user is removed from the group?
 What will it happen If the role belongs to the user and group at the same time and we revoke the role from the user. But the group stills is granted the role.

bharos: Q1: What happens if a user is removed from a group?

Group membership comes from the JWT token (PrincipalUtils.getCurrentPrincipal().getGroups()). When the IdP removes the user from the group, subsequent tokens won't include it, so resolveCurrentUserGroups skips it and the group's roles are not re-added.

However, stale g-rows from a previous request persist in the enforcer temporarily — until the role's cache entry is evicted (cache TTL expiry, or any handleRolePrivilegeChange call that invalidates it). There is no immediate signal from the IdP into Gravitino for group membership changes (unlike revokeRolesFromGroup which fires handleRolePrivilegeChange right away). I've added a TODO with a proposed fix (deleteUser() at the start of loadRolePrivilege to clear stale g-rows before re-resolving) and a test that documents this transient limitation - see testUserRemovedFromGroupAtIdpDeniesAccess.

Q2: Role belongs to user and group — revoke from user, group keeps it

This works correctly. When revokeRolesFromUser fires, handleRolePrivilegeChange(roleId) invalidates the cache → the removal listener calls deleteRole on both enforcers, removing all g-rows and p-rows for that role. On the next authorize call, loadRolePrivilege re-resolves both direct roles (user no longer has it) and group roles (group still has it) — so the role is re-loaded via the group path and access is preserved.
Added new test: testRoleSharedByUserAndGroupSurvivesUserRevocation

Also added couple of tests to verify deny vs allow case across user and group: `testDenyRoleOnUserOverridesAllowFromGroup` and `testDenyRoleFromGroupOverridesAllowOnUser`

bharos: @roryqi  In followup PR, if you agree with the approach, for the Q1 part, I can do:

At the start of [loadRolePrivilege](https://github.com/apache/gravitino/blob/3ca87d839eb5eeee6571b9ec95977e51ee3c78db/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java#L490), call `allowEnforcer.deleteUser(userId)` and `denyEnforcer.deleteUser(userId)` before re-adding roles. This removes all existing g-rows (user→role mappings) for the user, then the lambda re-resolves current groups from the JWT and re-adds only the roles the user is actually entitled to.

Today, stale g-rows persist because `addRoleForUser` is idempotent - it skips if the mapping already exists but never removes old ones. With `deleteUser()` first, it becomes `clear → re-resolve → re-add`. Since this is an in-memory operation on the enforcer's policy list, I anticipate no performance degradation.

roryqi: Q1: Cache TTL isn't easy to control. It will cause some undefined results. Users can't know the result will take effect. The worst case is that the result will still exist in the cache because other users still access the related data.

roryqi: > @roryqi In followup PR, if you agree with the approach, for the Q1 part, I can do:
> 
> At the start of [loadRolePrivilege](https://github.com/apache/gravitino/blob/3ca87d839eb5eeee6571b9ec95977e51ee3c78db/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java#L490), call `allowEnforcer.deleteUser(userId)` and `denyEnforcer.deleteUser(userId)` before re-adding roles. This removes all existing g-rows (user→role mappings) for the user, then the lambda re-resolves current groups from the JWT and re-adds only the roles the user is actually entitled to.
> 
> Today, stale g-rows persist because `addRoleForUser` is idempotent - it skips if the mapping already exists but never removes old ones. With `deleteUser()` first, it becomes `clear → re-resolve → re-add`. Since this is an in-memory operation on the enforcer's policy list, I anticipate no performance degradation.

If group membership changes, we should delete the user first. If not, we shouldn't delete the user.  If followup PR is merged in this version, I am OK.

roryqi: Q2: You should add notifyRoleGroupRelChange referring to notifyRoleUserRelChange.

bharos: For Q1, yes I can do that, but it is better to always delete the user, as it's just an in-memory operation on the enforcer. Deletion of user in enforcer will not result in any extra DB calls, role policies stay cached in loadedRoles.
> If not, we shouldn't delete the user.

This is hard to find, as group info is only present in user token. Otherwise, we will need to maintain a new cache: userId → Set<groupNames> from previous token

WDYT

bharos: For Q2, we are already calling notifyRoleUserRelChange in case of group mapping changes, so functionality works correctly even now. But for more clarity, I can add notifyRoleGroupRelChange, but with same logic as notifyRoleUserRelChange

roryqi: > For Q1, yes I can do that, but it is better to always delete the user, as it's just an in-memory operation on the enforcer. Deletion of user in enforcer will not result in any extra DB calls, role policies stay cached in loadedRoles.
> 
> > If not, we shouldn't delete the user.
> 
> This is hard to find, as group info is only present in user token. Otherwise, we will need to maintain a new cache: userId → Set from previous token
> 
> WDYT

We can record the group names for the user.
For example,
when token contains user1, group1 and group2, we can record user1<List<group1, group2>>. 

Another question,
If we delete a group from the metalake, how to handle this case?

bharos: I think maintaining a `userId → List<groupNames>` cache may not be future-proof. In a multi-node deployment, this cache is node-local and not backed by any DB entity (group membership comes from the IdP token, not Gravitino's DB). So it can't be invalidated via DB versioning (as we plan to do in the multi-node cache support) , each node would only self-correct after it independently sees a new token from the same user.

I am considering an `add-first-then-prune` approach instead:

- Collect all currently valid roleIds (from direct roles + group-inherited roles) into a desiredRoleIds set
- Add them all via addRoleForUser (idempotent — no-op if already present)
- After CompletableFuture.allOf().join(), call getRolesForUser(userId) on the enforcer
- For any role in the enforcer that's NOT in desiredRoleIds, call deleteRoleForUser(userId, role) on both enforcers

Why this is better:
- No extra cache — no memory overhead, no TTL tuning, no stale entries for inactive users
- Multi-node friendly — each node independently converges to the correct enforcer state using only the current token. No shared state or cross-node coordination needed
- No correctness issues - prune runs after join(), and is idempotent

bharos: >"If we delete a group from the metalake, how to handle this case?" 

For this, the `add-first-then-prune approach` handles this automatically. When a group is deleted from the metalake, `resolveCurrentUserGroups` calls `entityStore.batchGet()` which won't find the deleted group, so its roles won't be in `desiredRoleIds`. The prune step will then remove those stale role mappings from the enforcer on the next request.

This is another advantage over the `userId → List<groups> cache `approach — with group-cache, the token still lists the deleted group, so the cache wouldn't detect any change. We'd need a separate hook in `removeGroup()` to handle that case.

bharos: Example of the `add-then-prune` approach for followup PR : https://github.com/apache/gravitino/pull/10984/changes/99f5bf4edfee78279ef8e399356d3f63cd222cfa

@roryqi  Please LMK if you think this is reasonable and better approach than user->group cache

roryqi: Add-then-prune seems ok for me.

bharos: Closed the follow-up PR, and added the change in this same PR for "add-then-prune" approach.
Thanks for checking the possible approaches here @roryqi

roryqi (commented):
Could u add some cases for new optimization about pruning group roles.

bharos:
> Could u add some cases for new optimization about pruning group roles.

Yes, the existing tests already cover the pruning optimization well. I had modified the tests that are affected accordingly as well in https://github.com/apache/gravitino/pull/10868/commits/49c58f9b34193c365b34f9c2b1042853947557f9

- testUserRemovedFromGroupAtIdpDeniesAccess — directly exercises the prune step (no handleRolePrivilegeChange between assertions; pruning is the only mechanism removing the stale g-row)

- testMultipleGroupsPartialRevocationRetainsAccess — verifies selective pruning (only the stale role is removed)

- testRoleSharedByUserAndGroupSurvivesGroupRevocation / SurvivesUserRevocation — verify pruning doesn't over-remove roles still valid via another path

- testGroupRoleRevokedDeniesAccess — pruning after explicit revocation

roryqi (approved):
LGTM.

PR #10871, opened 2026-04-25
[#10172] improvement(server): Guard catch-block request dereferences in REST handlers

### LINKED ISSUE #10172: [Improvement] Catch-path request dereference can mask original exceptions in REST handlers

sachinnn99:

Multiple REST handlers dereference `request` (e.g. `request.getName()`, `request.getRoleNames()`, `request.getJobTemplate().name()`) inside `catch (Exception e)` blocks. When `request` is `null` (e.g. from an empty or malformed POST body), the catch block throws a secondary `NullPointerException` that masks the real failure.

This PR replaces all catch-path request dereferences with null-safe precomputed local variables across 16 REST handler files:

- Simple `getName()` pattern (11 files): Table, Fileset, Function, Model, Schema, Catalog, User, Group, Role, Policy, Tag, Topic
- `getRoleNames()` pattern: PermissionOperations (4 catch sites)
- Nested `getJobTemplate().name()`: JobOperations
- `isInUse()` / `isEnable()` pattern: MetalakeOperations, CatalogOperations, PolicyOperations
- `getNames()` pattern: StatisticOperations

**Note on StatisticOperations scope:** The issue lists 4 methods (`updateStatistics`, `dropStatistics`, `updatePartitionStatistics`, `dropPartitionStatistics`), but only `dropStatistics` is actually vulnerable — the other 3 already use null-safe helpers (`getStatisticNames`, `getPartitionNames`, `getDropPartitionNames`).

When `request` is null and an exception occurs in the method body, the catch block tries to dereference `request.get*()`, throwing a secondary NPE that masks the original exception. This makes debugging harder as the real error is lost. The fix ensures catch blocks always preserve the primary error through existing `ExceptionHandlers`.

Additionally, `JobOperations.registerJobTemplate` and `PolicyOperations.setPolicy` had pre-try `request` dereferences (outside the try-catch entirely), meaning a null request would NPE before the catch block could handle it at all.

Reproducible by sending an empty body to any affected endpoint:
```
POST /api/metalakes/m/catalogs/c/schemas
Accept: application/vnd.gravitino.v1+json
(empty body)
```

Fix: #10172

No. Error handling behavior is preserved — endpoints return the same error responses, but the original exception is no longer masked by a secondary NPE.

- Added null-request-body regression tests covering all main dereference patterns:
  - `TestGroupOperations.testAddGroupWithNullRequest` — `getName()` pattern
  - `TestPermissionOperations.testGrantRolesToUserWithNullRequest` — `getRoleNames()` pattern
  - `TestCatalogOperations.testCreateCatalogWithNullRequest` — `getName()` pattern in CatalogOperations
  - `TestJobOperations.testRegisterJobTemplateWithNullRequest` — nested `getJobTemplate().name()` pattern
- All existing tests pass: `./gradlew :server:test -PskipITs`

### DIFF
--- a/server/src/test/java/org/apache/gravitino/server/web/rest/TestCatalogOperations.java
+++ b/server/src/test/java/org/apache/gravitino/server/web/rest/TestCatalogOperations.java
@@ -588,6 +588,22 @@ public class TestCatalogOperations extends BaseOperationsTest {
     Assertions.assertEquals(RuntimeException.class.getSimpleName(), errorResponse1.getType());
   }
 
+  @Test
+  public void testCreateCatalogWithNullRequest() {
+    Response resp =
+        target("/metalakes/metalake1/catalogs")
+            .request(MediaType.APPLICATION_JSON_TYPE)
+            .accept("application/vnd.gravitino.v1+json")
+            .post(Entity.entity("null", MediaType.APPLICATION_JSON_TYPE));
+
+    Assertions.assertEquals(INTERNAL_SERVER_ERROR.getStatusCode(), resp.getStatus());
+
+    ErrorResponse errorResponse = resp.readEntity(ErrorResponse.class);
+    Assertions.assertNotNull(
+        errorResponse, "Null request should return a structured error, not crash the handler");
+    Assertions.assertEquals(ErrorConstants.INTERNAL_ERROR_CODE, errorResponse.getCode());
+  }
+
   private static TestCatalog buildCatalog(String metalake, String catalogName) {
     CatalogEntity entity =
         CatalogEntity.builder()
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java
@@ -161,8 +161,9 @@ public class CatalogOperations {
           });
 
     } catch (Exception e) {
+      String catalogName = request != null ? request.getName() : "";
       return ExceptionHandlers.handleCatalogException(
-          OperationType.CREATE, request.getName(), metalake, e);
+          OperationType.CREATE, catalogName, metalake, e);
     }
   }
 
@@ -198,7 +199,8 @@ public class CatalogOperations {
           });
 
     } catch (Exception e) {
-      LOG.info("Failed to test connection for catalog: {}.{}", metalake, request.getName());
+      String catalogName = request != null ? request.getName() : "";
+      LOG.info("Failed to test connection for catalog: {}.{}", metalake, catalogName);
       return ExceptionHandlers.handleTestConnectionException(e);
     }
   }
@@ -219,7 +221,8 @@ public class CatalogOperations {
       CatalogSetRequest request) {
     LOG.info("Received set request for catalog: {}.{}", metalake, catalogName);
 
-    OperationType op = request.isInUse() ? OperationType.ENABLE : OperationType.DISABLE;
+    OperationType op =
+        request != null && request.isInUse() ? OperationType.ENABLE : OperationType.DISABLE;
 
     try {
       return Utils.doAs(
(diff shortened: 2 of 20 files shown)


sachinnn99:
This builds on the same approach as #10183. Changes beyond that PR:

- Guards nested fields (e.g. `getJobTemplate().name()` in `JobOperations`)
- Fixes pre-try dereferences in `JobOperations.registerJobTemplate` and `PolicyOperations.setPolicy` that would NPE before the catch block
- Adds regression tests for the `getName()`, `getRoleNames()`, and `getJobTemplate().name()` patterns

server/src/test/java/org/apache/gravitino/server/web/rest/TestCatalogOperations.java
```diff
@@ -588,6 +588,22 @@ public void testSetCatalog() {
     Assertions.assertEquals(RuntimeException.class.getSimpleName(), errorResponse1.getType());
   }
 
+  @Test
+  public void testCreateCatalogWithNullRequest() {
+    Response resp =
+        target("/metalakes/metalake1/catalogs")
+            .request(MediaType.APPLICATION_JSON_TYPE)
+            .accept("application/vnd.gravitino.v1+json")
+            .post(Entity.entity("null", MediaType.APPLICATION_JSON_TYPE));
```
jerryshao, line 597: As I mentioned in other PR, parsing "null" will not normally happen, I really doubt the necessity of such a fix.

Another thing, if we really want to fix this, it is better to fix this in the object mapper. We should probably not allow such deserialization.

sachinnn99: Thanks @jerryshao. I looked into a centralized fix as you suggested:

- Jackson has no `DeserializationFeature` for rejecting null top-level POJO deserialization
- `ContainerRequestFilter` runs before deserialization, so it can't catch `"null"` JSON (the body is present, it just deserializes to Java `null`)
- The viable centralized options are Bean Validation (`@NotNull` + `jersey-bean-validation`) or a custom `MessageBodyReader` — both are worth exploring but are larger scope than this issue

I'll close this PR for now. If you'd be open to a centralized approach, I can open a separate issue for that.

jerryshao: Yes, I would prefer a centralized solution.

Issue #604, opened 2023-10-26
[EPIC] Add generic view semantics support in Gravitino


jerryshao:
### Describe the proposal

This epic tracks the works of supporting a generic view semantic in Gravitino. Please split into subtasks to track the design and implementation works.

PR #10879, opened 2026-04-27
[#10878] feat(core): add relational storage for logical views with versioning

### LINKED ISSUE #10878: [Subtask] feat(core): add relational storage for logical views with versioning
### Describe the subtask

Implement the relational storage layer for logical views, including `ViewEntity`, `ViewVersionInfoPO`, mappers/providers, and `ViewMetaService` CRUD with versioning.

### Parent issue

#604

mchades:

Adds the relational storage layer for the Logical View Management feature:

- New `ViewEntity` (implements `View`, `HasIdentifier`) with `defaultCatalog` / `defaultSchema` aligned with main's view API.
- New `ViewVersionInfoPO` + `ViewVersionInfoMapper` with base + PostgreSQL SQL providers.
- `ViewPO` / `ViewMetaMapper` providers extended to read/write `audit_info`, `default_catalog`, `default_schema` and join the current `view_version_info` row.
- `ViewMetaService` rewritten for full CRUD with version management and JSON serialization of columns/properties/representations; soft-delete support.
- `JDBCBackend` routes `ViewEntity` inserts through `ViewMetaService`; bumps `CURRENT_SCRIPT_VERSION` to `1.3.0`.
- `ViewOperationDispatcher` updated with a `createView` stub matching main's `ViewCatalog#createView(..., String defaultCatalog, String defaultSchema, Map)` signature; full CRUD dispatchers will land in a follow-up PR.

This is the storage portion of the Logical View Management EPIC (#604). The API and DDL are already on main; this PR wires the persistence layer so views can be stored in Gravitino's metadata store with versioned snapshots.

Fix: #10878

No new user-facing APIs in this PR. `CURRENT_SCRIPT_VERSION` is bumped from `1.2.0` to `1.3.0` to match the schema files already present on main; deployments will need to apply the `upgrade-1.2.0-to-1.3.0-*.sql` script (already in the repo).

- New unit tests in `TestViewMetaService` covering insert / load / update (with new version row) / soft delete / list-by-namespace / multi-representation round-trip.
- Existing `TestViewOperationDispatcher` extended for the updated `createView` signature.
- Verified locally: `./gradlew :core:test --tests "*TestViewMetaService*" --tests "*TestViewOperationDispatcher*" -PskipITs` — all 16 tests pass.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
@@ -29,7 +29,8 @@ import org.apache.gravitino.exceptions.NoSuchEntityException;
 import org.apache.gravitino.exceptions.NoSuchViewException;
 import org.apache.gravitino.lock.LockType;
 import org.apache.gravitino.lock.TreeLockUtils;
-import org.apache.gravitino.meta.GenericEntity;
+import org.apache.gravitino.meta.AuditInfo;
+import org.apache.gravitino.meta.ViewEntity;
 import org.apache.gravitino.rel.View;
 import org.apache.gravitino.storage.IdGenerator;
 import org.slf4j.Logger;
@@ -106,7 +107,7 @@ public class ViewOperationDispatcher extends OperationDispatcher implements View
 
     // Check if view exists in entity store
     try {
-      GenericEntity viewEntity = store.get(ident, Entity.EntityType.VIEW, GenericEntity.class);
+      ViewEntity viewEntity = store.get(ident, Entity.EntityType.VIEW, ViewEntity.class);
       return EntityCombinedView.of(catalogView, viewEntity).withImported(true);
     } catch (NoSuchEntityException e) {
       // View not in store yet
@@ -135,12 +136,26 @@ public class ViewOperationDispatcher extends OperationDispatcher implements View
 
     LOG.info("Auto-importing view {} into Gravitino entity store", ident);
     long uid = idGenerator.nextId();
-    GenericEntity newViewEntity =
-        GenericEntity.builder()
+    View catalogView = entityCombinedView.viewFromCatalog();
+    AuditInfo auditInfo =
+        AuditInfo.builder()
+            .withCreator(catalogView.auditInfo().creator())
+            .withCreateTime(catalogView.auditInfo().createTime())
+            .withLastModifier(catalogView.auditInfo().lastModifier())
+            .withLastModifiedTime(catalogView.auditInfo().lastModifiedTime())
+            .build();
+    ViewEntity newViewEntity =
+        ViewEntity.builder()
             .withId(uid)
             .withName(ident.name())
             .withNamespace(ident.namespace())
-            .withEntityType(Entity.EntityType.VIEW)
+            .withComment(catalogView.comment())
+            .withColumns(catalogView.columns())
+            .withRepresentations(catalogView.representations())
+            .withDefaultCatalog(catalogView.defaultCatalog())
+            .withDefaultSchema(catalogView.defaultSchema())
+            .withProperties(catalogView.properties())
+            .withAuditInfo(auditInfo)
             .build();
     try {
       store.put(newViewEntity, false /* overwrite */);
--- a/core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
@@ -21,17 +21,22 @@ package org.apache.gravitino.catalog;
 import static org.apache.gravitino.utils.NameIdentifierUtil.getCatalogIdentifier;
 
 import java.io.IOException;
+import java.util.Map;
+import javax.annotation.Nullable;
 import org.apache.gravitino.Entity;
 import org.apache.gravitino.EntityStore;
 import org.apache.gravitino.GravitinoEnv;
 import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.Namespace;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
 import org.apache.gravitino.exceptions.NoSuchViewException;
 import org.apache.gravitino.lock.LockType;
 import org.apache.gravitino.lock.TreeLockUtils;
-import org.apache.gravitino.meta.AuditInfo;
-import org.apache.gravitino.meta.ViewEntity;
+import org.apache.gravitino.meta.GenericEntity;
+import org.apache.gravitino.rel.Column;
+import org.apache.gravitino.rel.Representation;
 import org.apache.gravitino.rel.View;
+import org.apache.gravitino.rel.ViewChange;
 import org.apache.gravitino.storage.IdGenerator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -90,6 +95,33 @@ public class ViewOperationDispatcher extends OperationDispatcher implements View
     return entityCombinedView;
   }
 
+  @Override
+  public NameIdentifier[] listViews(Namespace namespace) {
+    throw new UnsupportedOperationException("Listing views is not supported yet");
+  }
+
+  @Override
+  public boolean dropView(NameIdentifier ident) {
+    throw new UnsupportedOperationException("Dropping a view is not supported yet");
+  }
+
+  @Override
+  public View createView(
+      NameIdentifier ident,
+      String comment,
+      Column[] columns,
+      Representation[] representations,
+      @Nullable String defaultCatalog,
+      @Nullable String defaultSchema,
+      Map<String, String> properties) {
+    throw new UnsupportedOperationException("Creating a view is not supported yet");
+  }
+
+  @Override
+  public View alterView(NameIdentifier ident, ViewChange... changes) {
+    throw new UnsupportedOperationException("Altering a view is not supported yet");
+  }
+
   /**
    * Internal method to load view and check if it exists in entity store.
    *
@@ -107,7 +139,7 @@ public class ViewOperationDispatcher extends OperationDispatcher implements View
 
     // Check if view exists in entity store
     try {
-      ViewEntity viewEntity = store.get(ident, Entity.EntityType.VIEW, ViewEntity.class);
+      GenericEntity viewEntity = store.get(ident, Entity.EntityType.VIEW, GenericEntity.class);
       return EntityCombinedView.of(catalogView, viewEntity).withImported(true);
     } catch (NoSuchEntityException e) {
       // View not in store yet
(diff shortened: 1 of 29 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/mapper/ViewMetaMapper.java
```diff
@@ -21,26 +21,69 @@
...
+        @Result(property = "deletedAt", column = "deleted_at"),
+        @Result(
+            property = "viewVersionInfoPO",
+            javaType = ViewVersionInfoPO.class,
+            column =
+                "{id,version_metalake_id,version_catalog_id,version_schema_id,version_view_id,"
+                    + "version,view_comment,columns,properties,default_catalog,default_schema,"
+                    + "representations,version_audit_info,version_deleted_at}",
+            one = @One(resultMap = "mapToViewVersionInfoPO"))
+      })
```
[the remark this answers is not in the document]
mchades: Checked this against the current runtime rather than changing it speculatively. `TestViewMetaService` still passes for insert/get/list/update on the supported relational backends, so this mapping is working as-is in the environments this PR targets. Since this PR is scoped to the ViewEntity storage changes, I am leaving the mapper shape unchanged here.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
```diff
@@ -158,104 +151,81 @@ public GenericEntity getViewByIdentifier(NameIdentifier identifier) {
...
 
-    if (viewPO == null) {
-      throw new NoSuchEntityException(
-          NoSuchEntityException.NO_SUCH_ENTITY_MESSAGE,
-          Entity.EntityType.VIEW.name().toLowerCase(),
-          viewName);
-    }
-    return viewPO;
+    return versionDeletedCount + metaDeletedCount;
   }
```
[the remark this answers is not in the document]
mchades: This observation is correct in isolation, but the same delete-budget pattern is already used in the other relational meta services as well. I do not want to special-case view cleanup in this PR and make it diverge from table/fileset/function behavior. If we want a single total budget plus one transaction, that should be handled as a broader follow-up across the shared pattern.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
```diff
@@ -158,104 +151,81 @@ public GenericEntity getViewByIdentifier(NameIdentifier identifier) {
...
       metricsSource = GRAVITINO_RELATIONAL_STORE_METRIC_NAME,
       baseMetricName = "updateViewByIdentifier")
-  public <E extends HasIdentifier> GenericEntity updateView(
+  public <E extends Entity & HasIdentifier> ViewEntity updateView(
       NameIdentifier ident, Function<E, E> updater) throws IOException {
     NameIdentifierUtil.checkView(ident);
 
     ViewPO oldViewPO = viewPOFetcher().apply(ident);
+    ViewEntity oldViewEntity = fromViewPO(oldViewPO, ident.namespace());
+    ViewEntity newEntity = (ViewEntity) updater.apply((E) oldViewEntity);
```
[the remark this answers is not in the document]
mchades: The type-safety concern is valid, but this updater signature currently follows the same pattern used by the other relational meta services too. For this PR I fixed the concrete runtime mismatch on the view rename path instead of tightening only the view service API and leaving the rest of the pattern inconsistent. I prefer to keep that signature change as a broader follow-up if we decide to apply it across table/fileset/function/view together.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
```diff
@@ -158,104 +151,81 @@ public GenericEntity getViewByIdentifier(NameIdentifier identifier) {
...
+            .withRepresentations(
+                new Representation[] {
+                  SQLRepresentation.builder().withDialect("unknown").withSql("placeholder").build()
+                })
+            .withAuditInfo(AuditInfo.EMPTY)
             .build();
-
-    insertView(viewPO, overwrite);
+    insertView(viewEntity, overwrite);
   }
```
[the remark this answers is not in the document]
mchades: Fixed by removing the `GenericEntity`-based view insert path entirely. View metadata is now persisted only through `ViewEntity`, so this code no longer synthesizes placeholder columns or representations for views.

core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
```diff
@@ -135,22 +170,45 @@ private EntityCombinedView importView(NameIdentifier ident) throws NoSuchViewExc
...
+      return EntityCombinedView.of(catalogView, newViewEntity).withImported(true);
     } catch (Exception e) {
       // Log but don't fail - view import is best-effort
       LOG.warn("Failed to import view {} into entity store: {}", ident, e.getMessage());
-      return EntityCombinedView.of(entityCombinedView.viewFromCatalog()).withImported(false);
+      return EntityCombinedView.of(catalogView).withImported(false);
     }
   }
+
+  private ViewEntity buildViewEntityForImport(Long uid, NameIdentifier ident, View view) {
```
jerryshao, line 186: What's the meaning of `buildViewEntityForImport`?

mchades: `buildViewEntityForImport` is the materialization step in the `loadView` auto-import path: it converts a catalog `View` into a persisted `ViewEntity` with Gravitino-generated id/identifier.

It is similar in purpose to table import, but currently lighter than `importTable` (no string-identifier reconciliation/rename correction path, and import here is best-effort).

Given `ViewOperationDispatcher` currently only supports `loadView` + auto-import, we expect to clean up/refactor this helper when full view CRUD lands in follow-up PRs.

core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
```diff
@@ -135,22 +170,45 @@ private EntityCombinedView importView(NameIdentifier ident) throws NoSuchViewExc
...
-            .withEntityType(Entity.EntityType.VIEW)
-            .build();
+    View catalogView = entityCombinedView.viewFromCatalog();
+    ViewEntity newViewEntity = buildViewEntityForImport(uid, ident, catalogView);
     try {
       store.put(newViewEntity, false /* overwrite */);
       LOG.info("Successfully imported view {} into entity store with id {}", ident, uid);
-      return EntityCombinedView.of(entityCombinedView.viewFromCatalog(), newViewEntity)
-          .withImported(true);
+      return EntityCombinedView.of(catalogView, newViewEntity).withImported(true);
```
jerryshao, line 178: If we import the view from external source to Gravitino, shall we still need to combine them, I think they're the same, am I right?

core/src/main/java/org/apache/gravitino/storage/relational/po/ViewPO.java
```diff
@@ -18,23 +18,119 @@
...
+import org.apache.gravitino.meta.ViewEntity;
+import org.apache.gravitino.rel.Column;
+import org.apache.gravitino.rel.Representation;
+import org.apache.gravitino.rel.SQLRepresentation;
+import org.apache.gravitino.storage.relational.service.EntityIdService;
 
-@Getter
-@EqualsAndHashCode
-@ToString
 public class ViewPO {
```
jerryshao, line 50: Why don't you use lombok to simplify the code?

core/src/main/java/org/apache/gravitino/storage/relational/po/ViewPO.java
```diff
@@ -18,23 +18,119 @@
...
+        metalakeId,
+        catalogId,
+        schemaId,
+        auditInfo,
+        currentVersion,
+        lastVersion,
+        deletedAt);
+  }
 
   public static class Builder {
```
jerryshao: Lombok can also generate the `Builder`.

core/src/main/java/org/apache/gravitino/catalog/ViewOperationDispatcher.java
mchades: Different catalogs support views to varying degrees, so combining them with viewEntity is necessary. Specific handling will be added in subsequent PRs.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
```diff
@@ -74,188 +80,118 @@ public Long getViewIdBySchemaIdAndName(Long schemaId, String viewName) {
...
+    try {
+      ViewPO newViewPO = updateViewPO(oldViewPO, newEntity);
+      SessionUtils.doMultipleWithCommit(
+          () ->
+              SessionUtils.doWithoutCommit(
+                  ViewVersionInfoMapper.class,
+                  mapper -> mapper.insertViewVersionInfo(newViewPO.getViewVersionInfoPO())),
+          () ->
+              SessionUtils.doWithoutCommit(
+                  ViewMetaMapper.class, mapper -> mapper.updateViewMeta(newViewPO, oldViewPO)));
```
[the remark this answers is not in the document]
mchades: fixed

core/src/main/java/org/apache/gravitino/storage/relational/po/ViewPO.java
mchades: fixed

core/src/main/java/org/apache/gravitino/storage/relational/po/ViewPO.java
```diff
@@ -18,94 +18,246 @@
...
+  }
+
+  private static Map<String, Object> toMap(Representation rep) {
+    if (rep instanceof SQLRepresentation) {
+      SQLRepresentation sql = (SQLRepresentation) rep;
+      ImmutableMap.Builder<String, Object> m = ImmutableMap.builder();
+      m.put("type", Representation.TYPE_SQL);
+      m.put("dialect", sql.dialect());
+      m.put("sql", sql.sql());
+      return m.build();
```
jerryshao: How does this `Representation` deserialize in DTO tranferring? I think we should align the json serde with DTO. Here the hard-coded `toMap` is very hard to maintain.

jerryshao: Or, you can figure out a more explicit way to do this.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
```diff
@@ -74,188 +80,127 @@ public Long getViewIdBySchemaIdAndName(Long schemaId, String viewName) {
...
+          () ->
+              SessionUtils.doWithoutCommit(
+                  ViewVersionInfoMapper.class,
+                  mapper -> mapper.insertViewVersionInfo(newViewPO.getViewVersionInfoPO())),
+          () -> {
+            updateResult.set(
+                SessionUtils.getWithoutCommit(
+                    ViewMetaMapper.class, mapper -> mapper.updateViewMeta(newViewPO, oldViewPO)));
+            if (updateResult.get() == 0) {
+              throw new RuntimeException("Failed to update the entity: " + ident);
```
jerryshao, line 166: Shall we throw a `NoSuchEntityException` here?

mchades: `updateViewMeta(...)` returning 0 here does not necessarily mean the view is missing; it can also be an optimistic-concurrency conflict (`current_version` mismatch).

We intentionally throw a runtime exception inside the transaction to force rollback, then map it to `IOException` for consistency with other update paths (e.g., `TableMetaService` and `FilesetMetaService`).

core/src/main/java/org/apache/gravitino/storage/relational/po/ViewPO.java
mchades: fixed

jerryshao (approved):

Issue #10881, opened 2026-04-28
[FEATURE] Local authentication in gravitino


lasdf1234:
### Describe the feature

The gravitino fulfills the function of Local authentication and enables local account authentication.

jerryshao:
Note, this issue is only for the design doc, for other tasks you need to create the subtask issues accordingly.

lasdf1234:
Got I have created sub-tasks for this issue and all of them have been completed. I am now awaiting the review of the code.

lasdf1234:
All the requirements have been fully developed.

PR #10883, opened 2026-04-28
[#10881] feat(authentication): Add local authentication design document

lasdf1234:

This PR adds a new design document, `design-docs/gravitino-local-idp.md`, for Local IDP support in Gravitino.

The document covers:
- the motivation and goals for introducing a built-in local IDP mode
- the Basic authentication model and related configuration
- password hashing design based on Argon2id
- the local identity data model, including `local_user_meta`, `local_group_meta`, and `local_group_user_rel`
- the logical mapping between local identity tables and existing `user_meta` / `group_meta`
- bootstrap account behavior
- authentication flow and transport security considerations
- local user/group/password management APIs

Gravitino currently relies on external identity providers for OAuth-based authentication. That works well in enterprise environments, but it creates friction for POC, offline, isolated, and emergency-fallback scenarios.

This design document captures a lightweight Local IDP proposal so the feature scope, storage model, authentication flow, and management APIs are clearly defined before implementation.

Fix: N/A

Yes.

It adds a new design document describing the proposed Local IDP feature, including:
- a new `basic` authenticator mode
- a proposed `gravitino.authenticator.basic.algorithm` configuration key
- new local identity management APIs for users, groups, group membership, and password changes

Documentation-only change.
Reviewed the document structure and aligned it with the existing `design-docs` style.


design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `group` | `String` | Yes | Group name |
+
+#### 9.1.8 Group Add User
+
+| Item | Value |
+|---|---|
+| Method | `PUT` |
+| Path | `/api/permissions/groups/{group}/users/add` |
```
roryqi: This isn't related about permission. Could u use accurate name?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+
+Because local IdP identities are global rather than metalake-scoped, these management interfaces do
+not include `{metalake}` in their paths.
+
+#### 9.1.1 Get Local User Info
+
+| Item | Value |
+|---|---|
+| Method | `GET` |
+| Path | `/api/users/{user}` |
```
roryqi: Maybe we would better have a prefix, because we should let users know this interface belongs to local authentication.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `user` | `String` | Yes | User name |
+
+**Response**
+
+| Field | Type | Description |
+|---|---|---|
+| `name` | `String` | User name |
+| `audit` | `AuditDTO` | Audit information |
```
roryqi: Why do we need audit information here?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `user` | `String` | Yes | User name |
+
+#### 9.1.4 Change Password
+
+| Item | Value |
+|---|---|
+| Method | `PUT` |
+| Path | `/api/users/{user}/password` |
```
roryqi: Do u need use the `password` here?   Is `/api/users/{user}` enough?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `user` | `String` | Yes | User name |
+| `password` | `String` | Yes | New password |
+
+**Spec**
+
+- Only service admin can change any account password.
+- The password cannot contain a colon (`:`). (RFC 7617)
+- The password must be at least 12 characters long and at most 64 characters long. (OWASP)
```
roryqi: What's meaning of OWASP?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `group` | `String` | Yes | Group name |
+
+**Response**
+
+| Field | Type | Description |
+|---|---|---|
+| `name` | `String` | Group name |
+| `audit` | `AuditDTO` | Audit information |
```
roryqi: Do we need audit info here?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+The local IdP is intentionally lightweight, but the following constraints remain important:
+
+- passwords are stored as hashes, never plaintext
+- Basic authentication should be used only over HTTPS
+- bootstrap credentials must be rotated after installation
+- all metadata tables use soft deletion for traceability and operational safety
+- local IdP mode is recommended for POC, offline rather than as a
+  replacement for enterprise-grade external identity systems
+
+---
```
roryqi, line 779: How did the Trino handle the Basic authentication?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+Because local IdP identities are global rather than metalake-scoped, these management interfaces do
+not include `{metalake}` in their paths.
+
+#### 9.1.1 Get Local User Info
+
+| Item | Value |
+|---|---|
+| Method | `GET` |
+| Path | `/api/users/{user}` |
+| Permission | Only service admin / owner can execute |
```
roryqi: Are these interfaces optional? Do we have a configuration option to control them?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+
+- Only service admin can change any account password.
+- The password cannot contain a colon (`:`). (RFC 7617)
+- The password must be at least 12 characters long and at most 64 characters long. (OWASP)
+
+**Error codes**
+
+| Error case | HTTP status |
+|---|---|
+| Password verification failed | `401` |
```
roryqi: I have concern about these error codes. Are they necessary?

lasdf1234: Yes, I removed the 428 and other less commonly used HTTP codes, and only retained codes like 401, which are common authentication failure codes.

roryqi: Especially for 422 Unprocessable Content, Is it common usage?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `group` | `String` | Yes | Group name |
+| `userNames` | `Array<String>` | Yes | User names. Must not be null or empty |
+---
+
+## 10. Security Considerations
+
+The local IdP is intentionally lightweight, but the following constraints remain important:
```
roryqi: Could u give me the user process, bootstrap process and verification?

design-docs/gravitino-local-idp.md
lasdf1234: These interfaces are currently fixed. Should we add a feature to determine whether basic authentication is enabled or not, in order to decide whether these interfaces can be accessed?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+## 9. Administrative Operations
+
+The local IdP must support the following administrator-managed operations:
+
+- add user
+- remove user
+- add group
+- remove group
+- add user to group
+- remove user from group
```
roryqi, line 466: Which group do the users belong to?

design-docs/gravitino-local-idp.md
lasdf1234: I will conduct further tests on this. If the use of Trino requires code modification for verification, I will open a separate issue/pr to address this problem.

design-docs/gravitino-local-idp.md
lasdf1234: Yes, I think it's necessary. I referred to the previous interface of gravitino and kept it consistent with the previous design style.

design-docs/gravitino-local-idp.md
lasdf1234: A general cryptography association can remove this. I'll submit the code.

design-docs/gravitino-local-idp.md
lasdf1234: OK, remove "password" at the end.

design-docs/gravitino-local-idp.md
lasdf1234: It is consistent with the previous similar interface, so "audit info" has been added.

design-docs/gravitino-local-idp.md
lasdf1234: Yes, I added the "/basic/auth" prefix to the code and it has been submitted.

design-docs/gravitino-local-idp.md
lasdf1234: The permissions are somewhat redundant, so I removed this prefix.

design-docs/gravitino-local-idp.md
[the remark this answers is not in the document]
lasdf1234: Got resolved it.I changed the prefix.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+CREATE TABLE IF NOT EXISTS `local_user_meta` (
+    `user_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'user id',
+    `user_name` VARCHAR(128) NOT NULL COMMENT 'username',
+    `password_hash` VARCHAR(256) NOT NULL COMMENT 'hashed password',
+    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'user audit info',
+    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'user current version',
+    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'user last version',
+    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'user deleted at',
+    PRIMARY KEY (`user_id`),
+    UNIQUE KEY `uk_un_del` (`user_name`, `deleted_at`)
```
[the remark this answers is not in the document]
lasdf1234: A very good suggestion. I changed it to varchar(1024) and then submitted the code.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+|---|---|
+| Password verification failed | `401` |
+| Account doesn't exist | `404` |
+| Password same with old | `422` |
+
+#### 9.1.5 Get Local Group Info
+
+| Item | Value |
+|---|---|
+| Method | `GET` |
```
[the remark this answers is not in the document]
lasdf1234: Got I have revised the document again. Only the serviceAdmin has the permission to change the password.

design-docs/gravitino-local-idp.md
roryqi: Maybe we can keep it simple first, we can add it when required for the user.

design-docs/gravitino-local-idp.md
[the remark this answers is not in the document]
lasdf1234: Got The ambiguous parts have been revised.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,568 @@
...
+- after initial login, the administrator is expected to change the password immediately
+
+This design favors usability for POC scenarios. The default password is intentionally simple so that
+users can access the system instantly, but it must be treated as a bootstrap credential rather than
+as a secure long-term password.
+
+---
+
+## 7. Authentication Flow
+
```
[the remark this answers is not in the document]
lasdf1234: Got it. I have already revised this part of the design.

design-docs/gravitino-local-idp.md
lasdf1234: "Add group" refers to creating a group, while "Remove group" refers to deleting an existing group.

design-docs/gravitino-local-idp.md
lasdf1234: I have included this part of the content in sections 6.2 and 7.1.

lasdf1234: The code has been submitted and the design document has been revised.

design-docs/gravitino-local-idp.md
lasdf1234: Would it be possible to remove the audit information from the local auth-related interfaces?

design-docs/gravitino-local-idp.md
roryqi: You should a segment to explain this point.

design-docs/gravitino-local-idp.md
roryqi: This interfaces should be optional.

design-docs/gravitino-local-idp.md
roryqi: We can remove them first. We can add them if required by users.

design-docs/gravitino-local-idp.md
lasdf1234: #### 9.1.1 Get Local User Info The response of this interface contains the "groups" information, which informs the interface caller which user groups this user belongs to.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+## 6. Bootstrap Experience
+
+To keep local IdP mode usable immediately after installation, Gravitino should provision a default
+administrator account on first startup.
+
+### 6.1 Initial Administrator
+
+- only one bootstrap account is created by default: **service admin**
+- the default password is **123456**
+- after initial login, the administrator is expected to change the password immediately
```
roryqi: This is used for UI login. REST API is hard to judge the first login.

design-docs/gravitino-local-idp.md
lasdf1234: image

design-docs/gravitino-local-idp.md
roryqi: You don't get the point about process. You should add user process instead of implementation process.

design-docs/gravitino-local-idp.md
lasdf1234: The web filter can filter such requests. If it is the "serviceAdmin" account and there is no password for "serviceAdmin" in the database, then only the "serviceAdmin" account is allowed to call the interface for changing the password.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+
+Add a dedicated configuration for password hashing.
+
+To align with Gravitino's current configuration naming convention, authenticator-specific settings
+should use the `gravitino.authenticator.<mode>.*` prefix rather than
+`gravitino.authenticators.<mode>.*`.
+
+| Key | Default | Optional Values |
+|---|---|---|
+| `gravitino.authenticator.basic.algorithm` | `Argon2id` | `Argon2id` |
```
roryqi: Is `gravitino.authenticator.basic.algorithm` a good name?

design-docs/gravitino-local-idp.md
roryqi: This comment isn't resolved. You shouldn't mark this resolved.

lasdf1234: Source of the specification: Initially defined in RFC 4918 of the WebDAV extension, it has now been incorporated into the core specification of HTTP semantics and content RFC 9110, becoming an Internet standard.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Design of Local IDP Support in Gravitino
```
roryqi: Local IDP will make others confused. I prefer using built-in IDP.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+
+```sql
+CREATE TABLE IF NOT EXISTS `local_group_user_rel` (
+    `id` BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT COMMENT 'auto increment id',
+    `group_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'local group id',
+    `user_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'local user id',
+    `audit_info` MEDIUMTEXT NOT NULL COMMENT 'relation audit info',
+    `current_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'relation current version',
+    `last_version` INT UNSIGNED NOT NULL DEFAULT 1 COMMENT 'relation last version',
+    `deleted_at` BIGINT(20) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'relation deleted at',
```
roryqi, line 232: Do use have GC thread to clean the legacy items?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+
+Because local IdP identities are global rather than metalake-scoped, these management interfaces do
+not include `{metalake}` in their paths and use the `/api/auth/basic` prefix.
+
+#### 9.1.1 Get Local User Info
+
+| Item | Value |
+|---|---|
+| Method | `GET` |
+| Path | `/api/auth/basic/users/{user}` |
```
roryqi: Could u refer to other systems? Which name will be better?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+| `name` | `String` | User name |
+| `audit` | `AuditDTO` | Audit information |
+| `groups` | `Array<String>` | Groups to which the user belongs |
+
+#### 9.1.2 Add Local User
+
+| Item | Value |
+|---|---|
+| Method | `POST` |
+| Path | `/api/auth/basic/users/{user}` |
```
roryqi: Why do we need `{user}` here? It isn't consistent with our other urls.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `user` | `String` | Yes | User name |
+
+#### 9.1.4 Change Password
+
+| Item | Value |
+|---|---|
+| Method | `PUT` |
+| Path | `/api/auth/basic/users/{user}/password` |
```
roryqi: Do we need `password` here?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+| `name` | `String` | Group name |
+| `audit` | `AuditDTO` | Audit information |
+| `users` | `Array<String>` | All users in the group |
+
+#### 9.1.6 Add Local Group
+
+| Item | Value |
+|---|---|
+| Method | `POST` |
+| Path | `/api/auth/basic/groups/{group}` |
```
roryqi: Why do we need `{group}` here?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `group` | `String` | Yes | Group name |
+
+#### 9.1.8 Group Add User
+
+| Item | Value |
+|---|---|
+| Method | `PUT` |
+| Path | `/api/auth/basic/groups/{group}/users/add` |
```
roryqi: This url is confused.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,574 @@
...
+|---|---|---|---|
+| `group` | `String` | Yes | Group name |
+| `userNames` | `Array<String>` | Yes | User names. Must not be null or empty |
+
+#### 9.1.9 Group Remove User
+
+| Item | Value |
+|---|---|
+| Method | `DELETE` |
+| Path | `/api/auth/basic/groups/{group}/users/remove` |
```
roryqi: Method is delete and you use the verb `remove`. This isn't a good url.

design-docs/gravitino-local-idp.md
roryqi: Which are other systems using 422?

design-docs/gravitino-local-idp.md
roryqi: cc @jerryshao WDYT?

design-docs/gravitino-local-idp.md
lasdf1234: Got resolved it.

design-docs/gravitino-local-idp.md
lasdf1234: Got resolved it.

design-docs/gravitino-local-idp.md
lasdf1234: Got resolved it

design-docs/gravitino-local-idp.md
lasdf1234: Got resolved it.

design-docs/gravitino-local-idp.md
lasdf1234: Got resolved it.

design-docs/gravitino-local-idp.md
lasdf1234: yes

lasdf1234: I have already incorporated this part of the content into the design document.

design-docs/gravitino-local-idp.md
lasdf1234: Or this document could be renamed as "Local Authentication"? And this place should be renamed as "Built-in IDP"

design-docs/gravitino-local-idp.md
lasdf1234: gravitino.authenticator.basic.password-hash-algorithm I have changed config name to this one.

roryqi: Could u borrow other system's experience? Notice, we should use camel style.

design-docs/gravitino-local-idp.md
lasdf1234: Got I have add in document : "### 6.3 Example Basic Authentication Bootstrap Flow"

design-docs/gravitino-local-idp.md
lasdf1234: Systems That Use HTTP 422 Status Code
HTTP 422 (Unprocessable Entity) originated as a WebDAV extension but is now widely adopted in modern RESTful APIs for semantic validation errors.

🌐 API Endpoints / Gateways & Cloud Platforms
Mainstream Web Frameworks (provide built-in support for returning 422):

Laravel (PHP): Returns 422 automatically for AJAX request validation failures.

Rails (Ruby) & Devise: Returns 422 when Turbo-driven form validations fail.

Drupal (PHP): Provides UnprocessableHttpEntityException class for throwing 422 exceptions.

Yii (PHP): Offers yii\web\UnprocessableEntityHttpException.

[ASP.NET](https://asp.net/) Core (C#): Includes UnprocessableEntity() result type.

Apache HttpClient (Java): Defines SC_UNPROCESSABLE_ENTITY constant.

Ruby: Gem::Net::HTTPUnprocessableEntity class for handling 422.

Specific Services & Components:

Red Hat Satellite: Returns 422 via its API when creating host entries with invalid parameters.

CnosDB: Uses 422 in its REST API to indicate operation execution failure.

FOLIO: Returns 422 with "EMAIL_ALREADY_EXIST" error when a user registers with an existing email.

NHS e-Referral Service: Returns 422 in its FHIR API when a request is unprocessable due to business logic (e.g., invalid relationship).

📡 Web Servers & Storage Systems
WebDAV servers: Use 422 when handling complex or malformed XML request bodies.

NetApp E-Series (SANtricity): Returns 422 when enabling/disabling AutoSupport with invalid parameters.

Dell PowerProtect Data Manager (versions <19.18): Shows a 422 error in the UI when configuring network with an invalid "search domain" value.

IBM Systems (AIX, z/OS): 422 may appear as a specific system error code.

📱 Operating Systems & Application Layer
Microsoft AD FS: Web Application Proxy (WAP) servers log Event ID 422 when idle connection timeout occurs (e.g., 100 seconds of inactivity).

Bittitan: Reports 422 Unprocessable Entity errors when interacting with Exchange WebDAV API on unprocessable mailbox folders.

OutSystems (mobile app framework): Returns 422 when data sent to an API fails validation

design-docs/gravitino-local-idp.md
lasdf1234: Got i have fixed it.

design-docs/gravitino-local-idp.md
roryqi: Do they use 422 to present that old password same as the new one.

design-docs/gravitino-local-idp.md
lasdf1234: image
The industry standard is to use "/api/users", or simply call it "/api/users?" This is more in line with the usual practice.

design-docs/gravitino-local-idp.md
roryqi: Prefer not using local IDP. Local authentication or built-in IDP is ok for me.

design-docs/gravitino-local-idp.md
roryqi: Could u avoid using `auth` prefix? We have already have API `/authn/me`. It will confuse users.

design-docs/gravitino-local-idp.md
lasdf1234: image
The core reason for using 422 is: when the request format is correct but the data is logically invalid, 422 is the most appropriate status code – unlike 400 Bad Request, which only indicates a malformed syntax.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,615 @@
...
+|---|---|---|---|
+| `group` | `String` | Yes | Group name |
+| `userNames` | `Array<String>` | Yes | User names. Must not be null or empty |
+
+#### 9.1.9 Group Remove Users
+
+| Item | Value |
+|---|---|
+| Method | `DELETE` |
+| Path | `/api/auth/basic/groups/{group}/users` |
```
roryqi: Could u refer to our other API design? For example, Tag operations?

design-docs/gravitino-local-idp.md
lasdf1234: The industry commonly uses "/api/{optional}/users" to define interfaces related to users and user groups.Keycloak uses 'realms' because there is a concept of 'realms' in Keycloak. Similarly, Ranger also uses the x+entity abstraction to create a layer of concept to name APIs. I would choose '/api/idp' in Gravitino. Do you think so?

design-docs/gravitino-local-idp.md
lasdf1234: OK, then modify it according to local authentication.

design-docs/gravitino-local-idp.md
lasdf1234: image
I think gravitino.authenticator.basic.password-hashing.algorithm this is the best, do you think so?

design-docs/gravitino-local-idp.md
lasdf1234: got resolved it

design-docs/gravitino-local-idp.md
roryqi: Prefer `passwordHasher`

design-docs/gravitino-local-idp.md
lasdf1234: I referred to UserOperations and also had the AI review it several times for me.

design-docs/gravitino-local-idp.md
roryqi: Ok for me.

design-docs/gravitino-local-idp.md
roryqi: I can't find it.

design-docs/gravitino-local-idp.md
lasdf1234: Got I have resolved and push code.

design-docs/gravitino-local-idp.md
roryqi: Maybe you can use 400 Bad Request here. Because it represents `illegalArguements` in our system instead of format exception. I prefer using 400. You can listen to others' opinion.

design-docs/gravitino-local-idp.md
lasdf1234: modify to /api/idp

design-docs/gravitino-local-idp.md
lasdf1234: Got i have fixed it

design-docs/gravitino-local-idp.md
lasdf1234: Ctrl + F you can find it

design-docs/gravitino-local-idp.md
lasdf1234: Got

design-docs/gravitino-local-idp.md
roryqi: You can refer to TagOperations API design instead of Permissions API.

design-docs/gravitino-local-idp.md
roryqi: Image
You don't have.

design-docs/gravitino-local-idp.md
roryqi: Could u point out which you added to the document?

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+
+The surveyed systems show that local authentication support typically converges on
+username/password-based flows. For Gravitino, simplicity matters more than protocol richness:
+
+- it shortens time-to-first-use,
+- it is easy to explain and operate,
+- it fits POC and offline scenarios well,
+- and it avoids introducing token lifecycle complexity into the server.
+
+For these reasons, the initial local IdP implementation uses:
```
roryqi: local authentication.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+
+- **File-based storage** requires a server restart to add users or rotate passwords.
+- **Database storage** supports normal metadata-style CRUD operations and matches Gravitino's
+  existing persistence model.
+
+Database-backed storage is the most practical choice for a built-in local IdP.
+
+### 3.4 Module Layout
+
+The local IdP feature should be implemented as an independent Gravitino module rather than being
```
roryqi: local authentication.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+scattered directly across the existing `server-common`, `server`, and `core` modules.
+
+The recommended module name is:
+
+- `authenticators:authenticator-local-idp`
```
roryqi: You shouldn't access core module.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+  "userNames": ["alice", "bob"]
+}' http://localhost:8090/api/idp/groups/engineering/users
+```
+
+#### 9.1.9 Remove users from a local group
+
+You can remove users from a local group by providing the group name in the path and the target
+user names in the request body.
+
+The request path for REST API is `/api/idp/groups/{group}/users`.
```
roryqi: Strange URL.You can refer to other URL desigin.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+
+```shell
+curl -X DELETE -H "Accept: application/vnd.gravitino.v1+json" \
```
roryqi: This is strange for me, too. DELETE usually doesn't contain request body.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+http://localhost:8090/api/idp/groups/engineering
+```
+
+#### 9.1.8 Add users to a local group
+
+You can add users to a local group by providing the group name in the path and the target user
+names in the request body.
+
+The request path for REST API is `/api/idp/groups/{group}/users`.
```
roryqi: I am not sure whether this is ok. Maybe collect others' suggestion.

design-docs/gravitino-local-idp.md
```diff
@@ -0,0 +1,679 @@
...
+- the user name cannot contain a colon (`:`); the Basic credential should be parsed by splitting on
+  the first colon, so the password may contain additional colons (RFC 7617)
+- the password must be at least 12 characters long and at most 64 characters long
+
+The password reset API should return:
+
+| Error case | HTTP status |
+|---|---|
+| Account doesn't exist | `404` |
+| Password same with old | `422` |
```
roryqi: I don't prefer 422. Need to listen to others' suggestion.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,679 @@
...
+   ```properties
+   gravitino.authenticators=basic
+   gravitino.authenticator.basic.passwordHasher=Argon2id
+   gravitino.authorization.serviceAdmins=adminUser
+   ```
+
+2. Start Gravitino.
+
+3. On first startup, no active `adminUser` record exists yet in `local_user_meta`, so Gravitino
+   accepts the bootstrap credential `adminUser:123456` only for the bootstrap login and immediate
```
roryqi: How to determine bootstrap login, this is not easy for REST.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,679 @@
...
+The local IdP requires three new tables:
+
+1. `local_user_meta` — local user records
+2. `local_group_meta` — local group records
+3. `local_group_user_rel` — user/group membership mapping
+
+These tables follow Gravitino's existing metadata table conventions:
+
+- numeric primary keys,
+- `audit_info`,
```
roryqi: Is audit info required?

design-docs/gravitino-local-idp.md
lasdf1234: https://github.com/trinodb/trino/pull/29132  have found this PR for Trino that can support this scenario. This PR was merged on April 29, 2026 at 6:00 (Beijing time).

design-docs/gravitino-local-idp.md
lasdf1234: Got I'll remove it.

design-docs/gravitino-local-authentication.md
lasdf1234: Got i have resolved it.

design-docs/gravitino-local-authentication.md
lasdf1234: For basic authentication, the user is stateless. Therefore, logic can be added in the filter: check the local_user_meta table. If the serviceAdmin does not have any data in this table, then the serviceAdmin must not have been registered in the built-in IDP. At this time, the filter will intercept all requests except for "change password" and prompt the serviceAdmin account to change the password.

design-docs/gravitino-local-idp.md
lasdf1234: I referred to the similar interface of gravitino-enterprise and changed it to a PUT interface.

design-docs/gravitino-local-idp.md
lasdf1234: I referred to the similar interface of gravitino-enterprise and changed it to a PUT interface.

design-docs/gravitino-local-idp.md
lasdf1234: I have made changes by referring to other documents.

design-docs/gravitino-local-idp.md
lasdf1234: The design has been modified.

design-docs/gravitino-local-idp.md
lasdf1234: Got I have been modified

design-docs/gravitino-local-idp.md
lasdf1234: Got I have been modified

design-docs/gravitino-local-idp.md
lasdf1234: Got I have been modified

design-docs/gravitino-local-idp.md
lasdf1234: Got I have been modified.These interfaces can only be called when the user uses "basic" seriously.

design-docs/gravitino-local-idp.md
lasdf1234: Got resolved

design-docs/gravitino-local-idp.md
lasdf1234: image
Got I have resolved it.

design-docs/gravitino-local-idp.md
roryqi: You should add the segment to explain this point.

design-docs/gravitino-local-authentication.md
roryqi: Collect others' suggestion.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+4. **Avoid vendor lock-in**: let users run Gravitino in environments where third-party IdPs are
+   impractical, undesirable, or cost-prohibitive.
+
+---
+
+## 3. Proposal
+
+### 3.1 Authentication Model
+
+The local authentication is introduced as a new Gravitino authenticator mode: **basic**.
```
jerryshao, line 65: The module name `basic` is not meaningful, you'd better have a more meaningful module name.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+#### 9.1.1 Get a local user
+
+You can get a local user by its name. The response should include the user name and current group
+memberships.
+
+The request path for REST API is `/api/idp/users/{user}`.
+
+```shell
+curl -X GET -H "Accept: application/vnd.gravitino.v1+json" \
+http://localhost:8090/api/idp/users/alice
```
roryqi, line 508: What's the response?

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+- the user name cannot contain a colon (`:`); the Basic credential should be parsed by splitting on
+  the first colon, so the password may contain additional colons (RFC 7617)
+- the password must be at least 12 characters long and at most 64 characters long
+
+The password reset API should return:
+
+| Error case | HTTP status |
+|---|---|
+| Account doesn't exist | `404` |
+| Password same with old | `422` |
```
roryqi: I have concern about this error code.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+Database-backed storage is the most practical choice for local authentication.
+
+### 3.4 Module Layout
+
+The local authentication feature should be implemented as an independent Gravitino module.
+
+The recommended module name is:
+
+- `authenticators:authenticator-local-authentication`
```
jerryshao: This is not aligned with one mentioned above. Also, the name is too long.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+curl -X DELETE -H "Accept: application/vnd.gravitino.v1+json" \
+http://localhost:8090/api/idp/groups/engineering
+```
+
+#### 9.1.8 Add users to a local group
+
+You can add users to a local group by providing the group name in the path and the target user
+names in the request body.
+
+The request path for REST API is `/api/idp/groups/{group}/users/add`.
```
roryqi: This seems that you have a user called `add`.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+- and `deleted_at` for soft deletion.
+
+Soft-deleted rows in `local_user_meta` and `local_group_meta` should be cleaned asynchronously by
+Gravitino's GC thread, following the same lifecycle management pattern used by other metadata
+tables. When a local user or local group is physically removed by the GC thread, the implementation
+should also clean the corresponding soft-deleted rows in `local_group_user_rel` to avoid leaving
+orphaned membership records.
+
+Unlike Gravitino's existing `user_meta` and `group_meta` tables, `local_user_meta` and
+`local_group_meta` are intentionally designed as **global identity tables** and therefore **do not
```
jerryshao: Make the table name more meaningful, `local_xxx` will easily be confused with existing table names.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+To keep local authentication usable immediately after installation, Gravitino should provision a default
+administrator account on first startup.
+
+### 6.1 Initial Administrator
+
+- only one bootstrap account is created by default: the configured **service admin** account (for
+  example, **adminUser**)
+- the default password is **123456**
+- the default password **123456** is intended only for the initial bootstrap login and the immediate
```
jerryshao: I would suggest to change the password to like "gravitino".

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+The bootstrap process should be:
+
+1. When the `basic` authenticator is enabled for the first time, check whether the configured
+   bootstrap **service admin** account already exists in `local_user_meta`.
+2. If the bootstrap account does not exist, the web filter allows the configured bootstrap service
+   admin account (for example, **adminUser**) with password **123456** to pass Basic verification
+   only for the bootstrap login and immediate password reset flow.
+3. Reject other management operations until the bootstrap password has been reset successfully.
+4. During the first successful password reset, create the bootstrap service admin record and store
```
jerryshao: How to reset the password, does it happen automatically or the service admin should do it manually?

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+The following end-to-end flow shows how a user can start from a fresh Gravitino deployment, enable
+Basic authentication, configure the bootstrap service admin identity, and immediately rotate the
+bootstrap password.
+
+1. Deploy Gravitino with the `basic` authenticator enabled:
+
+   ```properties
+   gravitino.authenticators=basic
+   gravitino.authenticator.basic.passwordHasher=Argon2id
```
jerryshao: Do you need to make this a configuration? I don't think it is necessary. If you want to make it a configuration, you need to provide different choices, do you support this?

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+4. Call the password reset API with HTTP Basic authentication:
+
+   ```bash
+   curl -X PUT \
+     -u 'adminUser:123456' \
+     -H 'Content-Type: application/json' \
+     https://<gravitino-host>/api/idp/users/adminUser \
+     -d '{
+       "password": "ChangeMeToAStrongPassword"
```
jerryshao: Is it safe to transfer the password without hashing?

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+You can remove a local group by its name. This operation should soft-delete the group record rather
+than physically remove it immediately.
+
+The request path for REST API is `/api/idp/groups/{group}`.
+
+```shell
+curl -X DELETE -H "Accept: application/vnd.gravitino.v1+json" \
+http://localhost:8090/api/idp/groups/engineering
+```
```
jerryshao, line 674: Can we remove a group if this group still has users?

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,672 @@
...
+
+You can remove users from a local group by providing the group name in the path and the target
+user names in the request body.
+
+The request path for REST API is `/api/idp/groups/{group}/users/remove`.
+
+```shell
+curl -X PUT -H "Accept: application/vnd.gravitino.v1+json" \
+-H "Content-Type: application/json" -d '{
+  "userNames": ["alice"]
```
jerryshao: Simplify all the `userNames` to `users`.

design-docs/gravitino-local-authentication.md
lasdf1234: Got it. I removed "/users" from the URL and have submitted the code.

design-docs/gravitino-local-authentication.md
lasdf1234: Got I have fixed it

design-docs/gravitino-local-authentication.md
lasdf1234: Got I have supplemented the responses for all the interfaces.

design-docs/gravitino-local-authentication.md
lasdf1234: Yes, transmitting passwords in plain text is secure. However, certain conditions must be met: 1. HTTPS 2. TLS 1.2 or a higher version 3. The server does not log passwords (log desensitization).
OWASP and NIST also recommend using this solution.

design-docs/gravitino-local-authentication.md
lasdf1234: Yes, when there are still users in the user group, the group can be removed. This interface will first remove the relationship between group and users, and then remove the group. The design of this interface is based on the design of the "remove role" interface. 
Spec have also been incorporated into the design document.

design-docs/gravitino-local-authentication.md
lasdf1234: Got, 'users' is better. I have modified the document.

design-docs/gravitino-local-authentication.md
lasdf1234: Got, Currently only supports this one encryption algorithm and there are no other options available. This configuration can be removed.
The design document has been modified.

design-docs/gravitino-local-authentication.md
lasdf1234: May be idp_user_meta/idp_group_meta is more meaningful.The URL of the newly added interface also includes "idp".

design-docs/gravitino-local-authentication.md
lasdf1234: This is the authenticator mode of gravitino, not the name of a module.

design-docs/gravitino-local-authentication.md
lasdf1234: Got, 'authenticators/authenticator-basic' is better. It is shorter and more in line with the semantic meaning of the context.

design-docs/gravitino-local-authentication.md
lasdf1234: The process has been changed. An 'interactive script' will be used, and the password will be changed by the service admin actively.

design-docs/gravitino-local-authentication.md
lasdf1234: Got, the document has been modified.

design-docs/gravitino-local-authentication.md
lasdf1234: I plan to adopt a different approach by using 'interactive script' to initialize the authentication of serviceAdmin. This will generate an SQL statement that will be directly written into the database.

design-docs/gravitino-local-idp.md
lasdf1234: I plan to adopt a different approach by using 'interactive script' to initialize the authentication of serviceAdmin. This will generate an SQL statement that will be directly written into the database.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,750 @@
...
+### 6.1 Initialization Script Inputs
+
+After Gravitino is installed, the installer should run an interactive script and provide:
+
+- the service admin name
+- the service admin password
+- the JDBC URL
+- the Gravitino database name
+- the JDBC user name
+- the JDBC password
```
jerryshao: This can be read from the configuration. If user they want to start Gravitino, they have to configure the JDBC information, so that can be shared here, no need to let user to input again.

jerryshao: Besides, I think this script is only needed for local authentication, right?

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,750 @@
...
+   - the service admin name must not contain a colon (`:`)
+   - the password must satisfy the local authentication password policy
+   - the JDBC parameters must be non-empty and syntactically valid
+5. Connect to the configured JDBC backend.
+6. Check whether an active row for the target service admin already exists in `idp_user_meta`. If
+   it already exists, abort rather than overwrite it implicitly.
+7. Hash the password with Argon2id.
+8. Generate the `INSERT` statement required for the target JDBC backend and execute it immediately
+   so the service admin record is written into `idp_user_meta`.
+9. Exit successfully only after the insert has been committed.
```
jerryshao: What happens if the initialization script is not running before the start?

jerryshao: Did you sync with @danhuawang and @geyanggang about how to use it in the docker and k8s environment?

jerryshao: Frankly saying, I'm in favor of the previous solution to have a default password, and let service admin to reset it for the first time.

To avoid hardcode issue, we can leave initial password as **blank**, then ask service admin to reset to a valid one.

Using a separate script and interactive solution makes things complicated, and hard to use in a container deployment.

design-docs/gravitino-local-authentication.md
jerryshao: This may be dangerous, you can have a query parameter like `force` to let the user choose. You can refer to schema / catalog deletion implementation.

design-docs/gravitino-local-authentication.md
```diff
@@ -0,0 +1,750 @@
...
+either unavailable or unnecessarily heavy. The key design choices are:
+
+- **HTTP Basic authentication**
+- **username/password credentials**
+- **database-backed storage**
+- **Argon2id password hashing**
+- **interactive service admin initialization**
+
+The result is a self-contained authentication path that is easy to deploy, fast to evaluate, and
+well aligned with Gravitino's lightweight quick-start experience.
```
jerryshao, line 793: Can you add a chapter to describe you work plan and checklist, it will be helpful for reviewer and AI to check.

design-docs/gravitino-local-authentication.md
lasdf1234: 1. if the initialization script is not running, the filter will block all requests.(Due to service admin has no password)
2. Got, I only discussed this matter with Rory. Now I'm going to discuss it with @danhuawang and @geyanggang

design-docs/gravitino-local-authentication.md
lasdf1234: 1.Yes, JDBC information can be read from other configuration files.
2.Yes, this script is only for local authentication.

design-docs/gravitino-local-authentication.md
lasdf1234: Got I will follow the style of other design documents to complete the plan and checklist.

design-docs/gravitino-local-authentication.md
danhuawang: I agree with the idea leaving initial password as blank, then ask service admin to reset it for the first time access. Using a separate script only for password init is not convenient.

design-docs/gravitino-local-idp.md
lasdf1234: My document has been updated.

lasdf1234: you can find it in 8.3

design-docs/gravitino-local-authentication.md
lasdf1234: The new initialization process has been completed. @roryqi @danhuawang @geyanggang all agree with this approach.

jerryshao (approved):

PR #10895, opened 2026-04-29
[#9087] feat(lance): support credential vending for Lance REST Server describeTable

### LINKED ISSUE #9087: [Subtask] Support authentication and authorization for Lance REST server
### Describe the subtask

Please see it.

### Parent issue

#8889

sunyuhan1998:

Add temporary credential vending support to the Lance REST Server's `describeTable` API, enabling clients to receive short-lived storage credentials instead of relying on long-term access keys.

Key changes:

1. **`CredentialPropertyUtils`** — Add Lance-specific credential property transformation: convert `S3TokenCredential`, `GCSTokenCredential`, and `S3SecretKeyCredential` into Lance `storage_options` format (`aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`, `expires_at_millis`).

2. **`GravitinoLanceTableOps.describeTable()`** — Integrate with `CatalogCredentialManager` to generate per-request temporary credentials based on the caller's privilege level (READ or WRITE). Merge vended credentials into `DescribeTableResponse.storageOptions`.

3. **`LanceTableOperations` (REST layer)** — Accept `vendCredentials` flag from `DescribeTableRequest`, determine caller privilege via `MetadataAuthzHelper`, and pass it down to the ops layer.

4. **Tests** — Unit tests for credential transformation, ops layer credential merging, REST layer vending flag handling, and end-to-end IT with `LanceDummyCredentialProvider` via Java SPI.

Currently the Lance REST Server returns static storage configuration without generating temporary credentials. When Lance clients access table data on S3/GCS, they need long-term access keys embedded in catalog properties, which poses a security risk. This aligns with the credential vending pattern already used by the Iceberg REST Server.

Fix: #9087

No. Credential vending is opt-in via the `credential-providers` catalog property and the `vendCredentials` request flag. Existing behavior is preserved when these are not configured.

- Unit tests: `TestCredentialPropertiesUtils`, `TestGravitinoLanceTableOpsCredentialVending`, `TestLanceNamespaceOperations` (vend credentials tests)
- IT: `LanceRESTServiceIT.testDescribeTableWithCredentialVending` using `LanceDummyCredentialProvider` via SPI

### DIFF
--- a/common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
+++ b/common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
@@ -51,6 +51,24 @@ public class CredentialPropertyUtils {
 
   private static final String GCS_OAUTH_2_TOKEN_EXPIRES_AT = "gcs.oauth2.token-expires-at";
 
+  // Lance storage_options key constants
+  @VisibleForTesting static final String LANCE_S3_ACCESS_KEY_ID = "aws_access_key_id";
+  @VisibleForTesting static final String LANCE_S3_SECRET_ACCESS_KEY = "aws_secret_access_key";
+  @VisibleForTesting static final String LANCE_S3_SESSION_TOKEN = "aws_session_token";
+  @VisibleForTesting static final String LANCE_OSS_ACCESS_KEY_ID = "oss_access_key_id";
+  @VisibleForTesting static final String LANCE_OSS_SECRET_ACCESS_KEY = "oss_secret_access_key";
+  @VisibleForTesting static final String LANCE_OSS_SECURITY_TOKEN = "oss_security_token";
+  @VisibleForTesting static final String LANCE_GCS_TOKEN = "google_storage_token";
+  @VisibleForTesting static final String LANCE_AZURE_SAS_TOKEN = "azure_storage_sas_token";
+
+  @VisibleForTesting
+  static final String LANCE_AZURE_STORAGE_ACCOUNT_NAME = "azure_storage_account_name";
+
+  @VisibleForTesting
+  static final String LANCE_AZURE_STORAGE_ACCOUNT_KEY = "azure_storage_account_key";
+
+  @VisibleForTesting static final String LANCE_EXPIRES_AT_MILLIS = "expires_at_millis";
+
   private static Map<String, String> icebergCredentialPropertyMap =
       ImmutableMap.<String, String>builder()
           .put(GCSTokenCredential.GCS_TOKEN_NAME, ICEBERG_GCS_TOKEN)
@@ -73,6 +91,31 @@ public class CredentialPropertyUtils {
           .put(AwsIrsaCredential.ACCESS_KEY_ID, ICEBERG_S3_ACCESS_KEY_ID)
           .put(AwsIrsaCredential.SECRET_ACCESS_KEY, ICEBERG_S3_SECRET_ACCESS_KEY)
           .put(AwsIrsaCredential.SESSION_TOKEN, ICEBERG_S3_TOKEN)
+          .build();
+
+  private static Map<String, String> lanceCredentialPropertyMap =
+      ImmutableMap.<String, String>builder()
+          .put(S3SecretKeyCredential.GRAVITINO_S3_STATIC_ACCESS_KEY_ID, LANCE_S3_ACCESS_KEY_ID)
+          .put(
+              S3SecretKeyCredential.GRAVITINO_S3_STATIC_SECRET_ACCESS_KEY,
+              LANCE_S3_SECRET_ACCESS_KEY)
+          .put(S3TokenCredential.GRAVITINO_S3_TOKEN, LANCE_S3_SESSION_TOKEN)
+          .put(OSSTokenCredential.GRAVITINO_OSS_SESSION_ACCESS_KEY_ID, LANCE_OSS_ACCESS_KEY_ID)
+          .put(
+              OSSTokenCredential.GRAVITINO_OSS_SESSION_SECRET_ACCESS_KEY,
+              LANCE_OSS_SECRET_ACCESS_KEY)
+          .put(OSSTokenCredential.GRAVITINO_OSS_TOKEN, LANCE_OSS_SECURITY_TOKEN)
+          .put(GCSTokenCredential.GCS_TOKEN_NAME, LANCE_GCS_TOKEN)
+          .put(ADLSTokenCredential.GRAVITINO_ADLS_SAS_TOKEN, LANCE_AZURE_SAS_TOKEN)
+          .put(
+              ADLSTokenCredential.GRAVITINO_AZURE_STORAGE_ACCOUNT_NAME,
+              LANCE_AZURE_STORAGE_ACCOUNT_NAME)
+          .put(
+              AzureAccountKeyCredential.GRAVITINO_AZURE_STORAGE_ACCOUNT_KEY,
+              LANCE_AZURE_STORAGE_ACCOUNT_KEY)
+          .put(AwsIrsaCredential.ACCESS_KEY_ID, LANCE_S3_ACCESS_KEY_ID)
+          .put(AwsIrsaCredential.SECRET_ACCESS_KEY, LANCE_S3_SECRET_ACCESS_KEY)
+          .put(AwsIrsaCredential.SESSION_TOKEN, LANCE_S3_SESSION_TOKEN)
           .build();
 
   /**
@@ -135,6 +178,24 @@ public class CredentialPropertyUtils {
                 !credentialPropertyKeys.contains(entry.getKey())
                     && !entry.getKey().startsWith(ICEBERG_ADLS_TOKEN));
     return filteredProperties;
+  }
+
+  /**
+   * Transforms a specific credential into a map of Lance storage_options properties.
+   *
+   * @param credential the credential to be transformed into Lance storage_options properties
+   * @return a map of Lance storage_options properties derived from the credential
+   */
+  public static Map<String, String> toLanceProperties(Credential credential) {
+    Map<String, String> properties =
+        new HashMap<>(transformProperties(credential.credentialInfo(), lanceCredentialPropertyMap));
+
+    long expireTimeInMs = credential.expireTimeInMs();
+    if (expireTimeInMs > 0) {
+      properties.put(LANCE_EXPIRES_AT_MILLIS, String.valueOf(expireTimeInMs));
+    }
+
+    return properties;
   }
 
   private static Map<String, String> transformProperties(
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -164,7 +164,8 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
           buildVendedStorageOptions(catalogName, catalog, table, credentialPrivilege));
     } else {
       response.setStorageOptions(
-          LancePropertiesUtils.resolveLanceStorageOptions(catalog.properties(), table.properties()));
+          LancePropertiesUtils.resolveLanceStorageOptions(
+              catalog.properties(), table.properties()));
     }
     return response;
   }
@@ -190,12 +191,13 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
 
     Credential credential;
     try {
-      credential = credManager.getCredential(context);
+      credential = credManager.getCredentialByPath(tableLocation, context);
     } catch (IllegalArgumentException e) {
       LOG.debug(
           "No credential providers configured for catalog {}, falling back to original storage options",
           catalogName);
-      return LancePropertiesUtils.resolveLanceStorageOptions(catalog.properties(), table.properties());
+      return LancePropertiesUtils.resolveLanceStorageOptions(
+          catalog.properties(), table.properties());
     }
 
     if (credential == null) {
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -29,18 +29,27 @@ import static org.apache.gravitino.rel.Column.DEFAULT_VALUE_NOT_SET;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.gravitino.Catalog;
 import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.credential.CatalogCredentialManager;
+import org.apache.gravitino.credential.Credential;
+import org.apache.gravitino.credential.CredentialPrivilege;
+import org.apache.gravitino.credential.CredentialPropertyUtils;
+import org.apache.gravitino.credential.PathBasedCredentialContext;
 import org.apache.gravitino.exceptions.NoSuchTableException;
 import org.apache.gravitino.lance.common.ops.LanceTableOperations;
 import org.apache.gravitino.lance.common.ops.gravitino.GravitinoLanceTableAlterHandler.AlterColumnsGravitinoLance;
@@ -50,6 +59,8 @@ import org.apache.gravitino.lance.common.utils.LancePropertiesUtils;
 import org.apache.gravitino.rel.Column;
 import org.apache.gravitino.rel.Table;
 import org.apache.gravitino.rel.TableChange;
+import org.apache.gravitino.utils.PrincipalUtils;
+import org.lance.namespace.errors.ServiceUnavailableException;
 import org.lance.namespace.errors.TableNotFoundException;
 import org.lance.namespace.model.AlterTableAlterColumnsRequest;
 import org.lance.namespace.model.AlterTableDropColumnsRequest;
@@ -75,6 +86,9 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
 
   private final GravitinoLanceNamespaceWrapper namespaceWrapper;
 
+  private final ConcurrentHashMap<String, CatalogCredentialManager> credentialManagers =
+      new ConcurrentHashMap<>();
+
   private enum CreateMode {
     CREATE,
     EXIST_OK,
@@ -98,9 +112,23 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
     this.namespaceWrapper = namespaceWrapper;
   }
 
+  public void close() {
+    for (Map.Entry<String, CatalogCredentialManager> entry : credentialManagers.entrySet()) {
+      try {
+        entry.getValue().close();
+      } catch (Exception e) {
+        LOG.warn("Error closing credential manager for catalog {}", entry.getKey(), e);
+      }
+    }
+    credentialManagers.clear();
+  }
+
   @Override
   public DescribeTableResponse describeTable(
-      String tableId, String delimiter, Optional<Long> version) {
+      String tableId,
+      String delimiter,
+      Optional<Long> version,
+      @Nullable CredentialPrivilege credentialPrivilege) {
     if (!version.isEmpty()) {
       throw new UnsupportedOperationException(
           "Describing specific table version is not supported. It should be null to indicate the"
@@ -131,9 +159,58 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
         Optional.ofNullable(table.properties().get(LANCE_TABLE_VERSION))
             .map(Long::valueOf)
             .orElse(null));
-    response.setStorageOptions(
-        LancePropertiesUtils.resolveLanceStorageOptions(catalog.properties(), table.properties()));
+    if (credentialPrivilege != null) {
+      response.setStorageOptions(
+          buildVendedStorageOptions(catalogName, catalog, table, credentialPrivilege));
+    } else {
+      response.setStorageOptions(
+          LancePropertiesUtils.resolveLanceStorageOptions(catalog.properties(), table.properties()));
+    }
     return response;
+  }
+
+  private Map<String, String> buildVendedStorageOptions(
+      String catalogName, Catalog catalog, Table table, CredentialPrivilege credentialPrivilege) {
+    String tableLocation = table.properties().get(LANCE_LOCATION);
+    Preconditions.checkArgument(
+        tableLocation != null && !tableLocation.isEmpty(),
+        "Table location is required for credential vending");
+
+    ImmutableSet<String> paths = ImmutableSet.of(tableLocation);
+    String userName = PrincipalUtils.getCurrentUserName();
+
+    PathBasedCredentialContext context =
+        credentialPrivilege == CredentialPrivilege.WRITE
+            ? new PathBasedCredentialContext(userName, paths, ImmutableSet.of())
+            : new PathBasedCredentialContext(userName, ImmutableSet.of(), paths);
+
+    CatalogCredentialManager credManager =
+        credentialManagers.computeIfAbsent(
+            catalogName, name -> new CatalogCredentialManager(name, catalog.properties()));
+
+    Credential credential;
+    try {
+      credential = credManager.getCredential(context);
+    } catch (IllegalArgumentException e) {
+      LOG.debug(
+          "No credential providers configured for catalog {}, falling back to original storage options",
+          catalogName);
+      return LancePropertiesUtils.resolveLanceStorageOptions(catalog.properties(), table.properties());
+    }
+
+    if (credential == null) {
+      LOG.warn("Failed to generate credential for table: {}", tableLocation);
+      throw new ServiceUnavailableException(
+          "Couldn't generate credential for table: " + tableLocation);
+    }
+
+    Map<String, String> baseStorageOptions =
+        LancePropertiesUtils.resolveLanceStorageOptions(catalog.properties(), table.properties());
+    Map<String, String> vendedProperties = CredentialPropertyUtils.toLanceProperties(credential);
+
+    Map<String, String> mergedOptions = new HashMap<>(baseStorageOptions);
+    mergedOptions.putAll(vendedProperties);
+    return mergedOptions;
   }
 
   @Override
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -41,7 +41,6 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import javax.annotation.Nullable;
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.gravitino.Catalog;
 import org.apache.gravitino.NameIdentifier;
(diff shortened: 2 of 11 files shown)


sunyuhan1998:
Hi @roryqi @yuqi1129, the dependent PR #10637 has been merged into main. This PR has been rebased onto the latest main and all conflicts are resolved. Could you please help review when you have time? Thanks!

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -51,6 +51,24 @@ public class CredentialPropertyUtils {
 
   private static final String GCS_OAUTH_2_TOKEN_EXPIRES_AT = "gcs.oauth2.token-expires-at";
 
+  // Lance storage_options key constants
+  @VisibleForTesting static final String LANCE_S3_ACCESS_KEY_ID = "aws_access_key_id";
```
yuqi1129, line 55: Those values are only used for Lance. Is it more proper to move them to the lance module?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -90,16 +96,37 @@ public Response describeTable(
       DescribeTableRequest request) {
     try {
       validateDescribeTableRequest(request);
+      boolean vendCredentials =
+          request.getVendCredentials() == null || Boolean.TRUE.equals(request.getVendCredentials());
```
yuqi1129, line 100: So, by default, we will get the credentials in the storage options?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -90,16 +96,37 @@ public Response describeTable(
...
       return Response.ok(response).build();
     } catch (Exception e) {
       return LanceExceptionMapper.toRESTResponse(tableId, e);
     }
   }
 
+  private CredentialPrivilege getCredentialPrivilege(String tableId, String delimiter) {
+    String[] parts = tableId.split(Pattern.quote(delimiter));
+    if (parts.length != 3) {
+      return CredentialPrivilege.READ;
```
yuqi1129: Why is it `CredentialPrivilege.READ` when the length of `parts` is not 3?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -90,16 +96,37 @@ public Response describeTable(
...
+    }
+
+    String metalake = lanceNamespace.config().getGravitinoMetalake();
+    NameIdentifier identifier = NameIdentifier.of(metalake, parts[0], parts[1], parts[2]);
+    boolean writable =
+        MetadataAuthzHelper.checkAccess(
+            identifier,
+            Entity.EntityType.TABLE,
+            AuthorizationExpressionConstants.FILTER_MODIFY_TABLE_AUTHORIZATION_EXPRESSION);
```
yuqi1129, line 130: Is this to check whether the current user has the `write` privileges? How can we get the current user?

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
sunyuhan1998: These constants and the `lanceCredentialPropertyMap` / `toLanceProperties()` are indeed Lance-specific, but `CredentialPropertyUtils` follows a centralized pattern where all table format credential transformations live in one class. The Iceberg equivalents (`ICEBERG_S3_ACCESS_KEY_ID`, `toIcebergProperties()`, etc.) are in the same class. Keeping them together seems consistent with the existing pattern. Would it be acceptable to keep it as is, or do you think moving them to the lance module would be better?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
sunyuhan1998: The `parts` should always have length 3 (catalog/schema/table) in normal flow. When it doesn't, using `READ` is a conservative safe default — it's better to under-provision than over-provision credentials. That said, this branch should never be reached in practice since upstream validation already enforces the 3-level structure. Would it be better to throw an explicit exception here instead?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
sunyuhan1998: Yes, `MetadataAuthzHelper.checkPermission()` checks whether the current authenticated user has write privilege on the table. The current user is obtained from the Gravitino authentication context — when the request arrives, the identity is already established and bound to the thread context. During credential generation, `CatalogCredentialManager` calls `PrincipalUtils.getCurrentUserName()` to retrieve the authenticated username and generates credentials scoped to that user's permissions.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
sunyuhan1998: Yes, by default credentials will be included in the storage options. When `vendCredentials` is not specified in the request (i.e., `null`), it defaults to `true`. Credentials are only skipped when the client explicitly sets `vendCredentials = false`. This follows the same convention as the Lance official namespace implementation (see [`unwrap_or(true)`](https://github.com/lance-format/lance/blob/main/rust/lance-namespace-impls/src/dir.rs#L1280)).

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
yuqi1129: OK, let me check it again.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: I believe an explicit exception will be better.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: I see.

sunyuhan1998:
Apologies for the force push. Here's a summary of the changes in the latest revision based on review feedback:

1. **`LanceTableOperations.java`** — Throw `IllegalArgumentException` when namespace is not 3-level, instead of silently returning `CredentialPrivilege.READ`.

2. **`LanceRESTServiceIT.java`** — Fix `org.junit.Assume.assumeTrue` (JUnit 4) to `org.junit.jupiter.api.Assumptions.assumeTrue` (JUnit 5) for the deploy-mode skip guard.

3. **`GravitinoLanceTableOperations.java`** — Add `close()` method to properly close cached `CatalogCredentialManager` instances (resource leak fix). Also moved `ALTER_HANDLERS` (static) before instance fields per code style guidelines.

4. **`GravitinoLanceNamespaceWrapper.java`** — Call `tableOperations.close()` in `close()` to wire up the lifecycle (following the same pattern as Iceberg REST's `CatalogWrapperForREST`).

@yuqi1129  Could you please review this again at your convenience? Thanks a lot.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
sunyuhan1998: done

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -90,16 +96,37 @@ public Response describeTable(
       DescribeTableRequest request) {
     try {
       validateDescribeTableRequest(request);
+      boolean vendCredentials =
+          request.getVendCredentials() == null || Boolean.TRUE.equals(request.getVendCredentials());
+      CredentialPrivilege privilege =
+          vendCredentials ? getCredentialPrivilege(tableId, delimiter) : null;
```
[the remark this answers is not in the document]
sunyuhan1998: We intentionally default `vendCredentials` to `true` to align with the Lance official namespace implementation

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
```diff
@@ -131,10 +148,61 @@ public DescribeTableResponse describeTable(
...
+
+    PathBasedCredentialContext context =
+        credentialPrivilege == CredentialPrivilege.WRITE
+            ? new PathBasedCredentialContext(userName, paths, ImmutableSet.of())
+            : new PathBasedCredentialContext(userName, ImmutableSet.of(), paths);
+
+    CatalogCredentialManager credManager =
+        credentialManagers.computeIfAbsent(
+            catalogName, name -> new CatalogCredentialManager(name, catalog.properties()));
+
```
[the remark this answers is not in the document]
sunyuhan1998: We've addressed this in the latest revision: `GravitinoLanceTableOperations` now has a `close()` method that properly closes all cached `CatalogCredentialManager` instances, and `GravitinoLanceNamespaceWrapper.close()` calls it during service shutdown. This follows the same pattern as Iceberg REST's `CatalogWrapperForREST`. For catalog property changes, the current behavior (requires server restart) is consistent with the Iceberg REST implementation.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
```diff
@@ -68,13 +79,16 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
...
 
+  private final GravitinoLanceNamespaceWrapper namespaceWrapper;
+
+  private final ConcurrentHashMap<String, CatalogCredentialManager> credentialManagers =
+      new ConcurrentHashMap<>();
+
   private static final Map<Class<?>, GravitinoLanceTableAlterHandler<?, ?>> ALTER_HANDLERS =
       Map.of(
           AlterTableDropColumnsRequest.class, new DropColumns(),
           AlterTableAlterColumnsRequest.class, new AlterColumnsGravitinoLance());
```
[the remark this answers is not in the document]
sunyuhan1998: Fixed in the latest revision. `ALTER_HANDLERS` is now placed before the instance fields.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -960,6 +962,63 @@ void testDeclareTable() {
     Assertions.assertFalse(new File(anotherLocation).exists());
   }
 
+  @Test
+  void testDescribeTableWithCredentialVending() {
+    // Skip in deploy mode: the dummy credential provider is only available in test classpath
+    org.junit.Assume.assumeTrue(
+        "Credential vending IT only runs in embedded mode",
+        ITUtils.EMBEDDED_TEST_MODE.equals(testMode));
```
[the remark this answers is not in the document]
sunyuhan1998: Fixed in the latest revision. Changed to `org.junit.jupiter.api.Assumptions.assumeTrue(...)`.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/integration/test/LanceRESTServiceIT.java
```diff
@@ -960,6 +962,63 @@ void testDeclareTable() {
...
+
+    // Create a catalog with credential-providers configured to use the dummy provider
+    String credCatalogName = GravitinoITUtils.genRandomName("lance_cred_catalog");
+    Map<String, String> credCatalogProps =
+        new HashMap<>() {
+          {
+            put("credential-providers", LanceDummyCredentialProvider.CREDENTIAL_TYPE);
+          }
+        };
```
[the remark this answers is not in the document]
sunyuhan1998: We'd prefer to keep the current style for consistency with other test methods in this file that use the same pattern.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -136,6 +136,10 @@ protected void configure() {
   public static void setup() {
     when(namespaceWrapper.asNamespaceOps()).thenReturn(namespaceOps);
     when(namespaceWrapper.asTableOps()).thenReturn(tableOps);
+    org.apache.gravitino.lance.common.config.LanceConfig lanceConfig =
+        mock(org.apache.gravitino.lance.common.config.LanceConfig.class);
+    when(namespaceWrapper.config()).thenReturn(lanceConfig);
+    when(lanceConfig.getGravitinoMetalake()).thenReturn("test-metalake");
```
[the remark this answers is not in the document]
sunyuhan1998: The FQN for `LanceConfig` is intentional — adding the import would cause a class name conflict with the existing `LanceNamespaceOperations` in the same file (from different packages), which broke tests in a previous revision. We'll keep the FQN here to avoid the collision.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
```diff
@@ -98,9 +112,23 @@ public GravitinoLanceTableOperations(GravitinoLanceNamespaceWrapper namespaceWra
...
+    credentialManagers.clear();
+  }
+
   @Override
   public DescribeTableResponse describeTable(
-      String tableId, String delimiter, Optional<Long> version) {
+      String tableId,
+      String delimiter,
+      Optional<Long> version,
+      @Nullable CredentialPrivilege credentialPrivilege) {
```
roryqi: Could u avoid null value here?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -90,16 +96,38 @@ public Response describeTable(
       DescribeTableRequest request) {
     try {
       validateDescribeTableRequest(request);
+      boolean vendCredentials =
+          request.getVendCredentials() == null || Boolean.TRUE.equals(request.getVendCredentials());
```
roryqi, line 100: You can reuse `Boolean.TRUE.equals(request.getVendCredentials());` directly. It can handle null value.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -90,16 +96,38 @@ public Response describeTable(
...
+      CredentialPrivilege privilege =
+          vendCredentials ? getCredentialPrivilege(tableId, delimiter) : null;
```
roryqi, line 102: Use null value is not good design. For many internal code, we don't handle null value. 
Usually we use optional instead.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
```diff
@@ -131,10 +159,61 @@ public DescribeTableResponse describeTable(
...
+          buildVendedStorageOptions(catalogName, catalog, table, credentialPrivilege));
+    } else {
+      response.setStorageOptions(LancePropertiesUtils.getLanceStorageOptions(table.properties()));
+    }
+
     return response;
   }
 
+  private Map<String, String> buildVendedStorageOptions(
+      String catalogName, Catalog catalog, Table table, CredentialPrivilege credentialPrivilege) {
```
roryqi: Could u put `CatalogCredentialManager` into Catalog? Could u create a catalogCredentialWrapper for this?

roryqi: Why do use `catalogName` instead of `catalog.name()`? It is duplicated.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
sunyuhan1998: Done. We have changed the parameter to `Optional<CredentialPrivilege>` in the `LanceTableOperations` interface and updated all callers/implementations.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
sunyuhan1998: The extra null check (`request.getVendCredentials() == null ||`) is intentional — it defaults `vendCredentials` to `true` when not specified, following the Lance official reference implementation(see [`unwrap_or(true)`](https://github.com/lance-format/lance/blob/main/rust/lance-namespace-impls/src/dir.rs#L1280))where `unwrap_or(true)` is used. That said, we are open to changing the default to `false` if you think that is more appropriate. What do you think?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
sunyuhan1998: Done in the latest commit. We have refactored to use `Optional<CredentialPrivilege>` throughout the call chain — from the REST layer through the interface down to the implementation.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
sunyuhan1998: Thanks for the suggestion. Our Lance server accesses catalogs through `GravitinoClient.loadCatalog()`, which returns a client-side proxy `Catalog`. In the Gravitino core, `BaseCatalog` already has a built-in `catalogCredentialManager()`, but we cannot access it from the client proxy since it is not part of the `Catalog` interface.

That is why we currently cache `CatalogCredentialManager` separately in a `ConcurrentHashMap` keyed by catalog name. To follow your suggestion, we could create a wrapper like:

```java
class LanceCatalogWrapper {
    private final Catalog catalog;
    private final CatalogCredentialManager credentialManager;

    LanceCatalogWrapper(Catalog catalog) {
        this.catalog = catalog;
        this.credentialManager = new CatalogCredentialManager(catalog.name(), catalog.properties());
    }
}
```

And modify `GravitinoLanceNamespaceWrapper` to return this wrapper instead of raw `Catalog`. This would make the association explicit. Would you like us to go with this approach? Or do you have a different design in mind?

sunyuhan1998: Done. We have replaced `catalogName` with `catalog.name()` as the cache key and removed the redundant `catalogName` variable.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
roryqi: Does the spec define the default value?

sunyuhan1998: Based on the interface specification of `lance-namespace`, it only stipulates that when the configuration **vend_credentials** is set to `true`, the server must return vended credentials. If this attribute is not explicitly specified, the server may determine the default behavior at its discretion; refer to: https://github.com/lance-format/lance-namespace/blob/81d3adf153aab117c43aabc474ce462831965e04/docs/src/spec.yaml#L2556-L2559 for details. However, looking at the current implementation of the Lance client, it does not carry or specify the `vend_credentials` parameter by default. This means the decision is delegated to the server implementation.

roryqi: I prefer that it keeps consistent with Iceberg REST behavior. So we could set false as the default value.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
roryqi: Revise the Lance server design. I found that Lance server only serves as the HTTP proxy. It forward all the request to the Gravitino server. 
Based on this point, we should use HTTP client to get the credential from the Gravitino server, Gravitino server generates the credential.
Now this proposal will generate the credential in the Lance REST server.

roryqi: Revise the Lance server design. I found that Lance server only serves as the HTTP proxy. It forward all the request to the Gravitino server. 
Based on this point, we should use HTTP client to get the credential from the Gravitino server, Gravitino server generates the credential.
Now this proposal will generate the credential in the Lance REST server.

FANNG1: I’d prefer to implement credential vending in a backend-agnostic way, rather than tying it to the current Gravitino-backed Lance namespace. That would align better with the approach used by the Iceberg REST server and would allow credential vending to work with other Lance namespace backends as well.

sunyuhan1998: Thanks for the feedback. Let me address both comments together since they relate to the same architectural question.

**Regarding the credential generation approach:**

We investigated how the Iceberg REST server handles credential vending. In [`CatalogWrapperForREST`](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java#L129), the `CatalogCredentialManager` is created locally and credentials are generated in-process via Java SPI-loaded `CredentialProvider`s — there is no HTTP call to the Gravitino server for credentials. Our current implementation follows the same pattern.

**Regarding the backend-agnostic concern (agree with @FANNG1):**

We agree that credential vending should not be tied to the Gravitino-backed implementation. We plan to refactor as follows:

1. **Create `LanceCredentialVendingService`** in `lance-common` — a Gravitino-agnostic service that handles credential generation. It takes catalog name, catalog properties, table location, and privilege as input, and returns credential properties for Lance `storage_options`.

2. **Simplify `GravitinoLanceTableOperations`** — it will only handle table operations (proxy to Gravitino server) and return the base `DescribeTableResponse` without credentials. All credential-related code will be removed.

3. **Credential injection at the wrapper/REST layer** — similar to how Iceberg's `CatalogWrapperForREST` injects credentials at the wrapper level (not inside individual table operations), we will orchestrate credential injection at the namespace wrapper or REST layer.

This way, any Lance namespace backend can use the `LanceCredentialVendingService` by simply providing catalog properties and table location, without depending on the Gravitino backend.

@roryqi @FANNG1 Does this approach make sense to you? Any concerns or suggestions?

roryqi: There are some difference between Lance and Iceberg. Lance used HTTP client to access the Gravitino server instead an internal class. If we finish the credential vending in the Lance server, Gravitino Lance catalog can't reuse the work.

roryqi: We only supports one Lance namespace backend.

FANNG1: Thanks @sunyuhan1998 ,  it make senses to me.  As how to support credential vending for Gravitino generic catalog, seems is another issue, we'd better not mixed it.

sunyuhan1998: Understood. We'll refactor to move credential generation to the Gravitino server side. Here's the overall approach:

1. `GenericCatalogOperations` implements `SupportsPathBasedCredentials` to provide storage paths for Lance tables, following the `FilesetCatalogOperations` pattern.
2. `GenericCatalog` overrides `propertiesWithCredentialProviders()` to auto-detect credential providers, following the `IcebergCatalog` pattern.
3. Lance server obtains credentials via the Gravitino credential API, converts to Lance `storage_options`, and removes local credential generation.
4. `vendCredentials` defaults to `false`.

One question: our current implementation generates **table-level (path-scoped)** credentials — each credential is scoped to the specific table's storage path with READ/WRITE distinction via `PathBasedCredentialContext`. To preserve this granularity on the server side, we'd need TABLE-level credential REST API support (adding `TABLE` to the supported types). Alternatively, we could use the existing **catalog-level** credential API (`catalog.supportsCredentials().getCredentials()`) which is simpler but generates broader credentials without per-table path scoping.

@roryqi @FANNG1 Which granularity do you prefer — catalog-level (simpler) or table-level (more secure, requires extending the credential API)?

roryqi: Why do we need extend the credential API？

sunyuhan1998: > Why do we need extend the credential API？

If we use `catalog.supportsCredentials().getCredentials()` directly, we get catalog-level credentials and don't need to extend the credential API at all. We'd only need TABLE-level support if per-table path scoping is required.

roryqi: It is reasonable for adding `table.supporsCredentials().getCredentials()`.

sunyuhan1998: > It is reasonable for adding `table.supporsCredentials().getCredentials()`.

Got it. We'll proceed with the table-level approach:

1. `GenericCatalogOperations` implements `SupportsPathBasedCredentials`
2. Server credential REST API adds `TABLE` type support
3. Client-side `RelationalTable` implements `SupportsCredentials`
4. Lance server uses `table.supportsCredentials().getCredentials()` and removes local credential generation

~~We'll rework this PR accordingly. Thanks for the guidance~~

Considering that some core items still need to be further clarified, the development work will temporarily not be advanced. It will be uniformly advanced after all relevant items are confirmed.

FANNG1: I may not have explained this clearly earlier, which may have caused some confusion. I’m -1 on fetching credentials through the Gravitino API. That would effectively tie credential vending to a Gravitino-API-based Lance namespace backend. If we want to support other Lance namespace backends in the future, or even a Gravitino-internal namespace backend implementation later (for example, if authentication/authorization is refactored), this approach would become a limitation. I’d prefer to keep credential vending backend-agnostic rather than coupling it to the current Gravitino API path.   cc @yuqi1129 @jerryshao , expecting your thought.

sunyuhan1998:
Hi, currently we still have some items that have not yet been decided. One of the core issues is: should the implementation of credential generation be placed within the Lance REST Server, or in the Gravitino Server? I'd like to hear everyone's thoughts. @FANNG1 @roryqi @yuqi1129 @jerryshao

jerryshao:
@sunyuhan1998 We still haven't figured out the right next step for Lance REST service, like using internal interface or call by Gravitino client, enforcing the RBAC in lance REST side or Gravitino API side, or even further, like how to manage and evolve this module. So, I would suggest that instead of implementing this, let's discuss such fundamental issues beforehand.

FANNG1:
Agree with @jerryshao. Before diving into implementation details, we need to align on the holistic picture of how Lance support should evolve in Gravitino — covering fundamental design decisions such as internal interface vs. Gravitino client, where RBAC enforcement should live, and the overall module management strategy. I have some initial thoughts on these questions and plan to organize them into a proposal in the next few days for everyone to discuss.

roryqi:
@sunyuhan1998 Hi, Since https://github.com/apache/gravitino/pull/11113 is merged, we can continue this pull request.

yuqi1129:
Yeah,  Lance Directory name namespace now supports credential-vending, so this direction is a worthy aspect for the Lance REST service.

PR #10904, opened 2026-04-29
[#10903] improvement(iceberg): Infer client io-impl from location

### LINKED ISSUE #10903: [Improvement] Supports switchingFileIO for Iceberg REST server

Supports switchingFileIO for Iceberg REST server

### How should we improve?

Supports switchingFileIO for Iceberg REST server

roryqi:

Infer client io-impl from location.

If the user unsets the FileIO in the catalog properties,  we use ResolvingFileIO as the default value.

Fix: #10903

No need.

Added UT.


roryqi:
@bharos @diqiu50 @yuqi1129 Could u help me review this pull request?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -160,6 +166,7 @@ public LoadTableResponse loadTable(
   public LoadTableResponse registerTable(
       Namespace namespace, RegisterTableRequest request, boolean requestCredential) {
     LoadTableResponse loadTableResponse = super.registerTable(namespace, request);
+    loadTableResponse = rewriteTableFileIOByLocation(loadTableResponse, requestCredential);
```
bharos: Why do we use requestCredential here instead of useSwitchingFileIO ?

bundles/gcp/src/main/java/org/apache/gravitino/gcs/credential/GCSTokenProvider.java
```diff
@@ -28,6 +28,11 @@
  */
 public class GCSTokenProvider extends CredentialProviderDelegator<GCSTokenCredential> {
 
+  @Override
+  public boolean supportsScheme(String scheme) {
```
bharos, line 32: IcebergCatalogUtil.resolveFileIOImplByLocation()  maps both "gs" and "gcs" to GCSFileIO, so should we make "gcs" as well here ?

core/src/main/java/org/apache/gravitino/credential/CatalogCredentialManager.java
```diff
@@ -51,15 +54,60 @@ public Credential getCredential(String credentialType, CredentialContext context
...
+    }
+
+    return getCredential(matchedCredentialType, filteredContext.get());
+  }
+
+  private String extractScheme(String path) {
+    Preconditions.checkArgument(path != null, "Path should not be null");
+    String scheme = URI.create(path).getScheme();
+    if (scheme == null) {
+      return "file://";
```
bharos: The :// is not part of the URI scheme , it's the separator. The correct fallback value is "file" ?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/io/SwitchingFileIO.java
```diff
@@ -0,0 +1,105 @@
...
+        .values()
+        .forEach(
+            fileIO -> {
+              try {
+                fileIO.close();
+              } catch (Exception e) {
+                throw new RuntimeException("Failed to close delegated FileIO", e);
+              }
+            });
+    delegateByImpl.clear();
```
bharos: If the first delegate's close() throws, the RuntimeException propagates and remaining delegates are never closed (resource leak). Also, clear() is never called. Should catch and accumulate exceptions, then throw a composite at the end.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/io/SwitchingFileIO.java
```diff
@@ -0,0 +1,105 @@
...
+  @Override
+  public Map<String, String> properties() {
+    return new HashMap<>(properties);
+  }
+
+  @Override
+  public void close() {
+    delegateByImpl
+        .values()
+        .forEach(
```
bharos: nit:  Between values().forEach(close) and delegateByImpl.clear(), a concurrent thread could call computeIfAbsent creating a new delegate. That delegate would then be cleared without being closed. Low probability (only during shutdown), but please check if we need to do anything against this

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
roryqi: My mistake. Fixed.

bundles/gcp/src/main/java/org/apache/gravitino/gcs/credential/GCSTokenProvider.java
roryqi: Make sense. Corrected.

core/src/main/java/org/apache/gravitino/credential/CatalogCredentialManager.java
roryqi: OK.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/io/SwitchingFileIO.java
roryqi: OK.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/io/SwitchingFileIO.java
roryqi: OK.

bharos:
@roryqi  do we support custom credential providers or know of any implementations ?
With this change,  even with explicit io-impl = S3FileIO, a custom CredentialProvider that hasn't overridden `supportsScheme()` (default returns false) will hit "No credential provider found for path ... with scheme s3" as soon as the client requests credentials.

Do we need to consider this or document anything as a behavior change in case we know of custom provider implementations

roryqi:
> @roryqi do we support custom credential providers or know of any implementations ? With this change, even with explicit io-impl = S3FileIO, a custom CredentialProvider that hasn't overridden `supportsScheme()` (default returns false) will hit "No credential provider found for path ... with scheme s3" as soon as the client requests credentials.
> 
> Do we need to consider this or document anything as a behavior change in case we know of custom provider implementations

I don't know that users have implement any custom providers. Maybe I should change the default value  to true for the compatible changes.

roryqi:
Maybe we should reuse [ResolvingFileIO](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/io/ResolvingFileIO.java) instead of creating a new FileIO.

bharos:
> Maybe we should reuse [ResolvingFileIO](https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/io/ResolvingFileIO.java) instead of creating a new FileIO.

That's better! Thanks

common/src/main/java/org/apache/gravitino/credential/CredentialProvider.java
```diff
@@ -49,6 +49,19 @@ public interface CredentialProvider extends Closeable {
    */
   String credentialType();
 
+  /**
+   * Checks whether this provider supports generating credentials for the given URI scheme.
+   *
+   * <p>By default, it returns {@code false}. Path-based credential flows should only use providers
```
bharos: Fix the javadoc

roryqi: Fixed.

bharos (approved):

Issue #10907, opened 2026-04-30
[Improvement] Improvement Jcasbin authoriation performance when roles with large grants


yuqi1129:
### What would you like to be improved?

When a role with large grants like 50, 100, or 500, the JCasbin authorizer will be CPU-bound and the performance gets lower and lower

PR #10908, opened 2026-04-30
[#10907] improvement(auth): Optimize JCasbin policy lookup

yuqi1129:

This PR optimizes the JCasbin authorization path by adding a per-role policy index in JcasbinAuthorizer.

The main changes are:

- Build and maintain a per-role policy index when role privileges are loaded.
- Cache the current request's user role IDs in AuthorizationRequestContext and use them for authorization checks.
- Resolve allow/deny authorization with hash lookups across the user's current roles instead of calling JCasbin enforce for every privilege probe.
- Remove the role's policy index when the role privilege cache entry is invalidated.
- Update TestJcasbinAuthorizer to verify that changed user-role assignments take effect immediately.

The previous JCasbin authorization path used enforce for each authorization probe, which scans the loaded policy set and becomes expensive when the number of roles and policies grows.

It also depended on role links cached inside the enforcer. If a user's role assignment changed while a role's policies were still cached, authorization could still be evaluated against stale role links.

With this change, each authorization request loads the user's current role list once, then checks the indexed policies for only those roles. This reduces the hot-path lookup cost and makes user-role assignment changes visible without waiting for role privilege cache invalidation.

Closes: #10907

No.

./gradlew :server-common:test --tests org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer -PskipITs -PskipDockerTests=false

### DIFF
--- a/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
+++ b/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
@@ -18,11 +18,14 @@
 package org.apache.gravitino.authorization;
 
 import java.security.Principal;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
+import javax.annotation.Nullable;
 import org.apache.gravitino.MetadataObject;
 
 public class AuthorizationRequestContext {
@@ -35,6 +38,13 @@ public class AuthorizationRequestContext {
 
   /** Used to determine whether the role has already been loaded. */
   private final AtomicBoolean hasLoadRole = new AtomicBoolean();
+
+  /**
+   * Role IDs of the current principal in the metalake under check. Populated once per request from
+   * {@code loadRole} and reused by every {@code authorize}/{@code deny} call so we don't re-derive
+   * the user→role linkage on each enforcer probe.
+   */
+  private volatile Set<Long> userRoleIds = Collections.emptySet();
 
   private volatile String originalAuthorizationExpression;
 
@@ -101,6 +111,14 @@ public class AuthorizationRequestContext {
 
   public void setOriginalAuthorizationExpression(String originalAuthorizationExpression) {
     this.originalAuthorizationExpression = originalAuthorizationExpression;
+  }
+
+  public Set<Long> getUserRoleIds() {
+    return userRoleIds;
+  }
+
+  public void setUserRoleIds(@Nullable Set<Long> userRoleIds) {
+    this.userRoleIds = userRoleIds == null ? Collections.emptySet() : userRoleIds;
   }
 
   public static class AuthorizationKey {
--- a/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
+++ b/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
@@ -19,6 +19,7 @@ package org.apache.gravitino.authorization;
 
 import java.security.Principal;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -113,12 +114,37 @@ public class AuthorizationRequestContext {
     this.originalAuthorizationExpression = originalAuthorizationExpression;
   }
 
+  /**
+   * Returns the user role IDs associated with the current authorization request.
+   *
+   * <p>This context is request-scoped and is expected to be used only for the lifetime of a single
+   * authorization request.
+   *
+   * <p>The returned set is immutable and safe to iterate without defensive copying.
+   *
+   * @return the user role IDs for the current request, or an empty set if none have been set
+   */
   public Set<Long> getUserRoleIds() {
     return userRoleIds;
   }
 
+  /**
+   * Sets the user role IDs associated with the current authorization request.
+   *
+   * <p>This context is request-scoped and is expected to be used only for the lifetime of a single
+   * authorization request.
+   *
+   * <p>The provided set is defensively copied before being stored, so subsequent caller-side
+   * mutations are not reflected in this context.
+   *
+   * @param userRoleIds the user role IDs for the current request; if {@code null}, an empty set is
+   *     stored
+   */
   public void setUserRoleIds(@Nullable Set<Long> userRoleIds) {
-    this.userRoleIds = userRoleIds == null ? Collections.emptySet() : userRoleIds;
+    this.userRoleIds =
+        userRoleIds == null || userRoleIds.isEmpty()
+            ? Collections.emptySet()
+            : Collections.unmodifiableSet(new HashSet<>(userRoleIds));
   }
 
   public static class AuthorizationKey {
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -535,6 +535,9 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
               roleIds.add(roleId);
               allowEnforcer.addRoleForUser(String.valueOf(userId), String.valueOf(roleId));
               denyEnforcer.addRoleForUser(String.valueOf(userId), String.valueOf(roleId));
+              if (loadedRoles.getIfPresent(roleId) != null) {
+                continue;
+              }
               CompletableFuture<Void> loadRoleFuture =
                   CompletableFuture.runAsync(
                       () -> {
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -40,6 +40,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Configs;
@@ -80,12 +81,6 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
   /** Jcasbin deny enforcer is used for metadata authorization. */
   private Enforcer denyEnforcer;
 
-  /** allow internal authorizer */
-  private InternalAuthorizer allowInternalAuthorizer;
-
-  /** deny internal authorizer */
-  private InternalAuthorizer denyInternalAuthorizer;
-
   /**
    * loadedRoles caches the indexed privileges for each loaded role. When a role's privileges are
    * updated, the role should be removed from this cache.
@@ -109,9 +104,7 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
 
     // Initialize enforcers before the caches that reference them in removal listeners
     allowEnforcer = new SyncedEnforcer(getModel("/jcasbin_model.conf"), new GravitinoAdapter());
-    allowInternalAuthorizer = new InternalAuthorizer(allowEnforcer, AuthorizationMode.ALLOW);
     denyEnforcer = new SyncedEnforcer(getModel("/jcasbin_model.conf"), new GravitinoAdapter());
-    denyInternalAuthorizer = new InternalAuthorizer(denyEnforcer, AuthorizationMode.DENY);
 
     loadedRoles =
         Caffeine.newBuilder()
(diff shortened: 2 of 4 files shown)


jerryshao:
Why don't fix this in main branch?

yuqi1129:
> Why don't fix this in main branch?

There is already a draft PR target for the main branch. I directly raised it to branch-1.2 to accelerate the release about 1.2.1. If this does not matter too much, we can merge it into the main and then use the cherry-pick procedure.

jerryshao:
> > Why don't fix this in main branch?
> 
> There is already a draft PR target for the main branch. I directly raised it to branch-1.2 to accelerate the release about 1.2.1. If this does not matter too much, we can merge it into the main and then use the cherry-pick procedure.

I see. My feeling is that this fix is a little hacky. @roryqi can you help to review?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -89,6 +94,16 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
...
 
+  /**
+   * Per-role policy index that mirrors what we hand to the JCasbin enforcers. Lookups in the hot
+   * path go through this map so an authorize call costs O(roles_per_user) hash probes instead of
+   * scanning every policy line in the enforcer. Lifecycle is tied to {@link #loadedRoles}: entries
+   * are written before {@code loadedRoles.put} and removed from {@code loadedRoles}'s removal
+   * listener so the two caches can never disagree on what a role grants.
+   */
+  private final ConcurrentHashMap<Long, Map<PolicyKey, Effect>> roleIndex =
+      new ConcurrentHashMap<>();
```
jerryshao: Do we have concurrency issue here?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -407,10 +423,12 @@ public void close() throws IOException {
 
   private class InternalAuthorizer {
 
-    Enforcer enforcer;
+    private final Enforcer enforcer;
+    private final boolean denyMode;
 
-    public InternalAuthorizer(Enforcer enforcer) {
+    InternalAuthorizer(Enforcer enforcer, boolean denyMode) {
```
jerryshao: Using `true` or `false` for different constructors is not so clear for the reviewer.

yuqi1129: OK, Let me optimize it.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
yuqi1129: Yes, there is a concurrency issue here indeed. The problem is not concurrent put for the same role itself, since ConcurrentHashMap can tolerate that. The issue is that roleIndex and loadedRoles are maintained as two separate states, so load and invalidation are not atomic.

One possible race is:

  1. Thread A sees loadedRoles does not contain role1 and starts loading role1 privileges.
  2. Thread B also loads or refreshes role1 and puts the latest index into roleIndex, then puts role1 into loadedRoles.
  3. A role privilege change happens and handleRolePrivilegeChange(role1) invalidates loadedRoles, whose removal listener removes role1 from roleIndex.
  4. Thread A finishes later and puts the old role index back into roleIndex, then puts role1 into loadedRoles again.

  After that, the stale role policy index may be resurrected even though the role was invalidated.

  Another possible race is that roleIndex is removed after it is put, but before loadedRoles.put(roleId, true). Then loadedRoles says the role is loaded while roleIndex no longer has the index, causing later authorization to skip reload and return a false negative.

I will fix this by avoiding two independent states for the same role. A better approach is to store the per-role policy index directly in the role cache, so “loaded” and “policy index” are updated/invalidated as one cache entry.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
yuqi1129: Updated. I replaced the boolean constructor flag with an explicit AuthorizationMode enum, so the call sites are now:

- new InternalAuthorizer(allowEnforcer, AuthorizationMode.ALLOW)
- new InternalAuthorizer(denyEnforcer, AuthorizationMode.DENY)

This should make the intent clear without requiring reviewers to remember what true/false means.

roryqi:
It seems that we can't remove Jcasbin, can we?

yuqi1129:
> It seems that we can't remove Jcasbin, can we

What do you mean here?

roryqi:
> > It seems that we can't remove Jcasbin, can we
> 
> What do you mean here?

The PR seems to do the duplicated with Jcasbin's function. If we follow the style of this pull request, we should implement check logic instead of using Jcasbin.

yuqi1129:
> > > It seems that we can't remove Jcasbin, can we
> > 
> > 
> > What do you mean here?
> 
> The PR seems to do the duplicated with Jcasbin's function. If we follow the style of this pull request, we should implement check logic instead of using Jcasbin.

I agree on the point that this change is duplicated with JCasbin's internal function to some extent. It replaces Jasbin's low-quality authorization method and tries to use the map to optimize the auth logic. 

This is just an improvement for a specific method, and Jcasbin still plays a key role in expression analysis, group, and grant management.  

Still, we need to thoroughly look into the JCasbin framework and try to optimize it one by one.  If you have a good idea, please do not hesitate to raise it.

core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
```diff
@@ -103,6 +113,14 @@ public void setOriginalAuthorizationExpression(String originalAuthorizationExpre
     this.originalAuthorizationExpression = originalAuthorizationExpression;
   }
 
+  public Set<Long> getUserRoleIds() {
+    return userRoleIds;
+  }
+
+  public void setUserRoleIds(@Nullable Set<Long> userRoleIds) {
+    this.userRoleIds = userRoleIds == null ? Collections.emptySet() : userRoleIds;
+  }
```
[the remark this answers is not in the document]
yuqi1129: Addressed. setUserRoleIds now defensively copies the provided set and stores an unmodifiable set, so caller-side mutation cannot affect the request context and downstream code cannot mutate the returned set. Added a unit test for both behaviors.

core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
```diff
@@ -103,6 +113,14 @@ public void setOriginalAuthorizationExpression(String originalAuthorizationExpre
     this.originalAuthorizationExpression = originalAuthorizationExpression;
   }
 
+  public Set<Long> getUserRoleIds() {
+    return userRoleIds;
+  }
+
```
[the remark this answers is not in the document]
yuqi1129: Addressed. I added JavaDoc for getUserRoleIds and setUserRoleIds describing the request-scoped lifecycle, immutable returned set, and null handling.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -482,36 +528,37 @@ private void loadRolePrivilege(
...
+                                return loadPolicyByRoleEntity(roleEntity);
+                              } catch (Exception e) {
+                                throw new RuntimeException(
+                                    "Failed to load role: " + role.name(), e);
+                              }
+                            });
+                      },
+                      executor);
               loadRoleFutures.add(loadRoleFuture);
             }
```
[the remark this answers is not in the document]
yuqi1129: Addressed. I added a loadedRoles.getIfPresent(roleId) fast path before creating the async load task, so cached roles no longer pay the CompletableFuture scheduling cost on every request. I also added testCachedRolePolicyIndexSkipsRoleEntityLoad to verify cached roles do not reload RoleEntity.

roryqi:
> > > > It seems that we can't remove Jcasbin, can we
> > > 
> > > 
> > > What do you mean here?
> > 
> > 
> > The PR seems to do the duplicated with Jcasbin's function. If we follow the style of this pull request, we should implement check logic instead of using Jcasbin.
> 
> I agree on the point that this change is duplicated with JCasbin's internal function to some extent. It replaces Jasbin's low-quality authorization method and tries to use the map to optimize the auth logic.
> 
> This is just an improvement for a specific method, and Jcasbin still plays a key role in expression analysis, group, and grant management.
> 
> Still, we need to thoroughly look into the JCasbin framework and try to optimize it one by one. If you have a good idea, please do not hesitate to raise it.

We can only use expression layer. We can implement our enforcer from your pull request. We don't need load policy to the Jcasbin enforcer, too.

jerryshao:
Can we figure out a solution quickly, so this will not block the 1.2.1 release and user delivery?

jerryshao:
If this is just a temporary fix for branch-1.2, the main branch will have a different solution. I think we can have a quick fix here, just make sure the fix is correct. What do you think? @yuqi1129 @roryqi

yuqi1129:
I have verified that the case mentioned by @roryqi is okay with the current changes and added tests to cover it.

yuqi1129:
We need to merge https://github.com/apache/gravitino/pull/10930 once the CI in this PR pass

jerryshao:
OK, since 1.2.0-hotfix cannot trigger the CI. We'd better fix it here and cherry-pick to that branch manually.

jerryshao:
@roryqi can you take another look on this PR?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -440,20 +447,59 @@ private boolean loadPrivilegeAndAuthorize(
...
-          privilege);
+      Set<Long> roleIds = requestContext.getUserRoleIds();
+      if (roleIds.isEmpty()) {
+        return false;
+      }
+      PolicyKey key = new PolicyKey(metadataObject.type().name(), metadataId, privilege);
+      if (authorizationMode == AuthorizationMode.DENY) {
+        for (Long roleId : roleIds) {
+          Map<PolicyKey, Effect> idx = loadedRoles.getIfPresent(roleId);
+          if (idx != null && idx.get(key) == Effect.DENY) {
```
roryqi: This seems to break the old logic.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -440,20 +447,59 @@ private boolean loadPrivilegeAndAuthorize(
...
-          String.valueOf(userId),
-          String.valueOf(metadataObject.type()),
-          String.valueOf(metadataId),
-          privilege);
+      Set<Long> roleIds = requestContext.getUserRoleIds();
+      if (roleIds.isEmpty()) {
+        return false;
+      }
+      PolicyKey key = new PolicyKey(metadataObject.type().name(), metadataId, privilege);
+      if (authorizationMode == AuthorizationMode.DENY) {
```
roryqi: We shouldn't have authorizationMode in my view.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
yuqi1129: You can describe clearly what the problem is, and about this point, you are more familiar with it than me. if it has problems, you can directly give me advice.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
yuqi1129: This is to replace the original `true/false` for `allowEnforcer.`.  The original is just a boolean value which indicates whether it's an ' allowEnforcer `allowEnforcer`or not.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
yuqi1129: What is the old logic? Deny has the highest priority.

jerryshao:
@roryqi please review and push forward the PR.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
roryqi: Image
I can't get your point. They should reuse the common enforcer logic.

yuqi1129: > They should reuse the common enforcer logic.

Optimized

roryqi (dismissed):

yuqi1129:
It seems that I can handle such cases, and I have added related tests.

server-common/src/test/java/org/apache/gravitino/server/authorization/jcasbin/TestJcasbinAuthorizer.java
```diff
@@ -1033,6 +1034,145 @@ public void testAuthorizeAndDenyShareSameResolveAcrossPrivileges() throws Except
...
+    Principal currentPrincipal = PrincipalUtils.getCurrentPrincipal();
+    NameIdentifier userIdent = NameIdentifierUtil.ofUser(METALAKE, USERNAME);
+    Long roleId = 3001L;
+    RoleEntity denyRole =
+        getRoleEntity(
+            roleId,
+            "denyRunJobRole" + roleId,
+            ImmutableList.of(
+                makeSecurableObject(
+                    "testMetalake", MetadataObject.Type.METALAKE, roleId, RUN_JOB, "DENY")));
```
roryqi: You should add an another role with an object allow run job, too.

roryqi (approved):

PR #10912, opened 2026-04-30
[#10770] feat(cache): Add 1.3.0 relational schema

### LINKED ISSUE #10770: [Subtask] Adjust SQL DDL to support authorization cache.

yuqi1129:

Add the 1.3.0 relational schema and upgrade scripts for H2, MySQL, and PostgreSQL, and bump the current script version to 1.3.0.

This is the database schema foundation for the cache improvement work. It only includes SQL and version changes.

Fix: #10770

No user-facing API change. This updates backend relational storage schema scripts.

Ran targeted relational tests in the stacked branches after rebasing on this schema branch.


scripts/h2/schema-1.3.0-h2.sql
```diff
@@ -560,3 +565,15 @@ CREATE INDEX IF NOT EXISTS `idx_table_metrics_composite`
   ON `table_metrics`(`table_identifier`, `table_partition`, `metric_ts`);
 CREATE INDEX IF NOT EXISTS `idx_job_metrics_identifier_metric_ts`
   ON `job_metrics`(`job_identifier`, `metric_ts`);
+
+CREATE TABLE IF NOT EXISTS `entity_change_log` (
+  `id`            BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+  `metalake_name` VARCHAR(128)    NOT NULL,
+  `entity_type`   VARCHAR(32)     NOT NULL,
+  `full_name`     VARCHAR(512)    NOT NULL,
```
roryqi: Why do u use full name here?

yuqi1129: Do you have a good suggestion here? I need to recode the name of an entity( in fact, full name).

roryqi: Should we use entity id here? Name will change.

yuqi1129: This is the change log, and it's okay to store the name. Using ID will make us query the database to get the names, so I'm relunctant to use ID here.

roryqi: Name will involve the order issues. If two tables rename the same table, what will it happen?

yuqi1129: The table will record the old name and invalid the cache related to it

roryqi: Just feel that id will be better than name. Name relies on the correct order. It's ok for us now at least.

jerryshao (approved):

PR #10914, opened 2026-04-30
[#10916] feat(cache): Add entity change log mapper

### LINKED ISSUE #10916: [Subtask] Add mapper and services for entity_change_log
### Describe the subtask

This issues track the support for entity_change_log mappers and the related service.

### Parent issue

#9898

yuqi1129:

Add the entity_change_log mapper, operation type handling, mapper registration, write-path changelog inserts for entity rename/drop, and focused mapper tests.

This PR is stacked on #10912.

This is the entity change log foundation for cache invalidation. It depends on the 1.3.0 schema PR and is independent from the auth mapper PR.

Fix: #10916

No user-facing API change. This adds internal relational mapper support for targeted cache invalidation.

./gradlew :core:test --tests org.apache.gravitino.storage.relational.mapper.provider.base.TestAuthMappers


jerryshao (commented):
Review of entity change log mapper and write-path changelog inserts. A few correctness issues to address.

core/src/main/java/org/apache/gravitino/storage/relational/po/cache/OperateType.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * <p>The numeric {@code code} is what is persisted in the database. Codes are stable forever and
+ * must never be reused: removing a value from this enum is allowed, but the same numeric code
+ * cannot subsequently be assigned to a different meaning, otherwise old log rows would be
+ * misinterpreted by readers.
+ */
+public enum OperateType {
+  ALTER(1),
+  DROP(2),
+  INSERT(3);
```
jerryshao: **`INSERT` is defined but never used in production.**

`ALTER` and `DROP` are the only values emitted by the service layer. `INSERT(3)` appears only in the mapper tests. The class Javadoc says the log tracks 'create, alter, drop', suggesting CREATE was planned.

Two options:
- If CREATE events don't need cache invalidation, remove this value (the Javadoc says removing a value is allowed).
- If CREATE events should be tracked, add `insertChange(..., OperateType.INSERT)` calls to the `insert*` methods of each MetaService.

As-is it's a dead value that will confuse future readers.

core/src/main/java/org/apache/gravitino/storage/relational/po/cache/OperateType.java
```diff
@@ -0,0 +1,52 @@
...
+
+  OperateType(int code) {
+    this.code = code;
+  }
+
+  public int getCode() {
+    return code;
+  }
+
+  public static OperateType fromCode(int code) {
```
jerryshao, line 51: **`fromCode` is O(n) — consider a static lookup map.**

Currently does a linear scan through `values()` on every read. With three values today it is fine, but as the enum grows or is called frequently on the cache-invalidation hot path this will degrade. A `static final Map<Integer, OperateType> BY_CODE` built in a static initializer makes lookups O(1) and removes the `for` loop entirely.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
```diff
@@ -0,0 +1,55 @@
...
+      @Param("metalakeName") String metalakeName,
+      @Param("entityType") String entityType,
+      @Param("fullName") String fullName,
+      @Param("operateType") OperateType operateType) {
+    return "INSERT INTO "
+        + ENTITY_CHANGE_LOG_TABLE_NAME
+        + " (metalake_name, entity_type, entity_full_name, operate_type, created_at)"
+        + " VALUES (#{metalakeName}, #{entityType}, #{fullName}, #{operateType},"
+        + " (UNIX_TIMESTAMP() * 1000.0)"
+        + " + EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000)";
```
jerryshao, line 67: **MySQL-specific timestamp expression inherited by H2.**

`EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3))` is MySQL syntax. H2 in `MODE=MYSQL` does not fully implement fractional-seconds `CURRENT_TIMESTAMP(3)`, so H2 tests could fail or silently produce wrong `created_at` values.

Consider overriding `insertEntityChange` in `EntityChangeLogH2Provider` with a simpler H2-compatible expression (e.g. `CURRENT_TIMESTAMP * 1000`), keeping the precise formula only in the MySQL provider.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
```diff
@@ -230,6 +258,14 @@ private boolean deleteView(Long viewId) {
                 mapper ->
                     mapper.softDeleteTagMetadataObjectRelsByMetadataObject(
                         viewId, MetadataObject.Type.VIEW.name()));
+            SessionUtils.doWithoutCommit(
```
jerryshao, line 265: **Changelog insert is inside the tag-deletion lambda and is missing the `deleteResult > 0` guard.**

Unlike every other service in this PR, this `insertChange` call is appended inside the lambda that does `softDeleteTagMetadataObjectRelsByMetadataObject`, rather than being a separate top-level lambda in `doMultipleWithCommit`. More importantly, `TableMetaService.deleteTable` and `ModelMetaService.deleteModel` both guard the insert with `if (deleteResult.get() > 0)`, but there is no such guard here — the changelog row is always written regardless of whether the view existed.

Please:
1. Move the `SessionUtils.doWithoutCommit(EntityChangeLogMapper.class, ...)` out to a new separate top-level lambda in `doMultipleWithCommit`.
2. Guard it with `if (deleteResult.get() > 0)` to match the pattern.

core/src/main/java/org/apache/gravitino/storage/relational/service/TopicMetaService.java
```diff
@@ -298,7 +331,17 @@ public boolean deleteTopic(NameIdentifier identifier) {
                 PolicyMetadataObjectRelMapper.class,
                 mapper ->
                     mapper.softDeletePolicyMetadataObjectRelsByMetadataObject(
-                        topicId, MetadataObject.Type.TOPIC.name())));
+                        topicId, MetadataObject.Type.TOPIC.name())),
+        () -> {
```
jerryshao, line 337: **Inconsistent conditional guard on changelog insert for `deleteTopic`.**

`TableMetaService.deleteTable` wraps its changelog insert in `if (deleteResult.get() > 0)`. Here `deleteTopic` always inserts the changelog row unconditionally — there is no `deleteResult` tracking and the method always returns `true`. The same unconditional pattern is also in `FilesetMetaService.deleteFileset`.

Either:
- Track a delete count and guard the insert (like `TableMetaService`), or
- Accept the unconditional pattern and apply it uniformly in `TableMetaService` / `ModelMetaService` too.

Pick one and apply it consistently across all services.

core/src/test/java/org/apache/gravitino/storage/relational/mapper/provider/base/TestEntityChangeLogMapper.java
```diff
@@ -0,0 +1,169 @@
...
+    List<EntityChangeRecord> rows = entityChangeLogMapper.selectChanges(0L, 100);
+    Assertions.assertEquals(3, rows.size());
+    Assertions.assertTrue(rows.get(0).getId() < rows.get(1).getId());
+    Assertions.assertTrue(rows.get(1).getId() < rows.get(2).getId());
+  }
+
+  private void forceCreatedAt(String fullName, long createdAt) throws SQLException {
+    try (Statement statement = sharedSession.getConnection().createStatement()) {
+      statement.execute(
+          "UPDATE entity_change_log SET created_at = "
```
jerryshao: **SQL injection in test helper.**

`fullName` is concatenated directly into SQL:
```java
"WHERE entity_full_name = '" + fullName + "'"
```
Use a `PreparedStatement` instead:
```java
PreparedStatement ps = sharedSession.getConnection()
    .prepareStatement("UPDATE entity_change_log SET created_at = ? WHERE entity_full_name = ?");
ps.setLong(1, createdAt);
ps.setString(2, fullName);
ps.executeUpdate();
```

core/src/main/java/org/apache/gravitino/storage/relational/mapper/EntityChangeLogMapper.java
```diff
@@ -0,0 +1,53 @@
...
+  @InsertProvider(type = EntityChangeLogSQLProviderFactory.class, method = "insertEntityChange")
+  void insertChange(
+      @Param("metalakeName") String metalakeName,
+      @Param("entityType") String entityType,
+      @Param("fullName") String fullName,
+      @Param("operateType") OperateType operateType);
+
+  @DeleteProvider(type = EntityChangeLogSQLProviderFactory.class, method = "pruneOldEntityChanges")
+  void pruneOldEntries(@Param("before") long before);
+}
```
jerryshao, line 52: Like I mentioned in another PR, the method name should be aligned with method string; otherwise it is errorprone.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/EntityChangeLogSQLProviderFactory.java
```diff
@@ -0,0 +1,94 @@
...
+            .getDatabaseId();
+    JDBCBackendType jdbcBackendType = JDBCBackendType.fromString(databaseId);
+    return ENTITY_CHANGE_LOG_SQL_PROVIDER_MAP.get(jdbcBackendType);
+  }
+
+  static class EntityChangeLogMySQLProvider extends EntityChangeLogBaseSQLProvider {}
+
+  static class EntityChangeLogH2Provider extends EntityChangeLogBaseSQLProvider {}
+
+  static class EntityChangeLogPostgreSQLProvider extends EntityChangeLogBaseSQLProvider {
```
jerryshao: You'd better put this class into postges package.

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
```diff
@@ -365,7 +398,17 @@ public boolean deleteCatalog(NameIdentifier identifier, boolean cascade) {
...
+          () -> {
+            SessionUtils.doWithoutCommit(
+                EntityChangeLogMapper.class,
+                mapper ->
+                    mapper.insertChange(
+                        metalakeName,
+                        Entity.EntityType.CATALOG.name(),
+                        NameIdentifierUtil.ofCatalog(metalakeName, catalogName).toString(),
+                        OperateType.DROP));
+          });
```
jerryshao, line 414: Will the deletion of catalog also insert the drop record for schemas and tables?

core/src/main/java/org/apache/gravitino/storage/relational/session/SqlSessionFactoryHelper.java
```diff
@@ -110,6 +112,9 @@ public void init(Config config) {
       // Initialize the configuration
       Configuration configuration = new Configuration(environment);
       configuration.setDatabaseId(jdbcType.name().toLowerCase());
+      configuration
+          .getTypeHandlerRegistry()
+          .register(OperateType.class, new OperateTypeTypeHandler());
```
jerryshao: `OperateTypeTypeHandler` or `OperateTypeHandler` ?

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
yuqi1129: The `(UNIX_TIMESTAMP() * 1000.0) + EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000` expression is the established codebase convention for `created_at`/`deleted_at` columns — it appears in 27+ base providers (`TableMetaBaseSQLProvider`, `FilesetVersionBaseSQLProvider`, `TagMetaBaseSQLProvider`, etc.). H2 in `MODE=MYSQL` parses it correctly; `TestEntityChangeLogMapper` runs against H2 and the timestamp assertion passes within the JVM-clock window. Splitting it per-DB here would diverge from the existing pattern — I would prefer to keep it consistent and address the H2/MySQL fractional-second alignment project-wide if it ever becomes a real issue. WDYT?

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
yuqi1129: No — this PR only writes a single `CATALOG` DROP record on cascade. The cache layer treats catalog-level invalidation as a prefix invalidation: when it sees `CATALOG drop <metalake>.<catalog>`, it evicts every entry whose key begins with `<metalake>.<catalog>.*` (schemas, tables, topics, filesets, views under it). Writing per-child DROP rows would multiply changelog volume on cascade deletes without changing cache correctness. Happy to add a comment in `CatalogMetaService.deleteCatalog` clarifying this if it helps future readers.

core/src/main/java/org/apache/gravitino/storage/relational/service/TopicMetaService.java
yuqi1129: Changed

core/src/main/java/org/apache/gravitino/storage/relational/po/cache/OperateType.java
yuqi1129: changed.

core/src/main/java/org/apache/gravitino/storage/relational/po/cache/OperateType.java
yuqi1129: Remove `INSERT` as `INSERT` is unnecessary, as it's only used for cache invalidation.

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
jerryshao: did you fix this issue?

jerryshao (commented):
**Re-review after the latest push** — several issues from the prior round are now fixed (ViewMetaService guard confirmed fixed, OperateType.fromCode O(1) map, SQL injection in test helper, @BeforeEach cache clear, INSERT OperateType removed, TopicMetaService deleteResult guard added). Below are the remaining items.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/EntityChangeLogMapper.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.ibatis.annotations.DeleteProvider;
+import org.apache.ibatis.annotations.InsertProvider;
+import org.apache.ibatis.annotations.Param;
+import org.apache.ibatis.annotations.SelectProvider;
+
+/**
+ * A MyBatis Mapper for entity_change_log table operations.
+ *
+ * <p>This append-only log tracks structural changes to entities (create, alter, drop) and is used
+ * by the entity change poller to drive targeted invalidation of the metadataIdCache on HA peer
```
jerryshao: Stale Javadoc — `INSERT` was removed from `OperateType`, so this should now read **"alter, drop"** (not "create, alter, drop"). Same wording appears in `OperateType`'s own class Javadoc; please update both.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/EntityChangeLogSQLProviderFactory.java
```diff
@@ -0,0 +1,69 @@
...
+        SqlSessionFactoryHelper.getInstance()
+            .getSqlSessionFactory()
+            .getConfiguration()
+            .getDatabaseId();
+    JDBCBackendType jdbcBackendType = JDBCBackendType.fromString(databaseId);
+    return ENTITY_CHANGE_LOG_SQL_PROVIDER_MAP.get(jdbcBackendType);
+  }
+
+  static class EntityChangeLogMySQLProvider extends EntityChangeLogBaseSQLProvider {}
+
```
jerryshao, line 50: The codebase convention for DB-specific providers is separate top-level classes in their own files under `provider/mysql/` and `provider/h2/` (see `TableMetaMySQLProvider`, `FilesetVersionH2Provider`, etc.). `EntityChangeLogMySQLProvider` and `EntityChangeLogH2Provider` as private inner classes here breaks that pattern. Please extract them to `provider/mysql/EntityChangeLogMySQLProvider.java` and `provider/h2/EntityChangeLogH2Provider.java`.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
```diff
@@ -0,0 +1,55 @@
...
+public class EntityChangeLogBaseSQLProvider {
+
+  public String selectEntityChanges(
+      @Param("createdAtFrom") long createdAtFrom, @Param("maxRows") int maxRows) {
+    return "SELECT id, metalake_name as metalakeName, entity_type as entityType,"
+        + " entity_full_name as fullName, operate_type as operateType, created_at as createdAt"
+        + " FROM "
+        + ENTITY_CHANGE_LOG_TABLE_NAME
+        + " WHERE created_at >= #{createdAtFrom} ORDER BY created_at, id LIMIT #{maxRows}";
+  }
```
jerryshao, line 46: The `>=` bound is **inclusive**, but the parameter name `createdAtFrom` ("from") implies the same. Callers must be careful: advancing the cursor by `lastRecord.createdAt` will re-read the last row on the next poll. If the poller increments `createdAtFrom` by 1 (millisecond), rows within the same millisecond at the boundary could be missed. Since `ORDER BY created_at, id` is now in place, the correct cursor strategy is to track both `lastCreatedAt` and `lastSeenId` and filter client-side, or use `(created_at, id) > (#{lastCreatedAt}, #{lastId})`. Please add a comment here documenting the expected cursor-advance contract so the poller implementation isn't left to guess.

core/src/main/java/org/apache/gravitino/storage/relational/service/MetalakeMetaService.java
```diff
@@ -173,19 +176,37 @@ public <E extends Entity & HasIdentifier> BaseMetalake updateMetalake(
...
+                  SessionUtils.getWithoutCommit(
+                      MetalakeMetaMapper.class,
+                      mapper -> mapper.updateMetalakeMeta(newMetalakePO, oldMetalakePO))),
+          () -> {
+            if (isRenamed && updateResult.get() > 0) {
+              SessionUtils.doWithoutCommit(
+                  EntityChangeLogMapper.class,
+                  mapper ->
+                      mapper.insertEntityChange(
+                          oldFullName,
```
jerryshao: Minor but confusing: `oldFullName` is passed as the `metalakeName` parameter. For metalakes the full name *equals* the name, so the value is correct — but the intent is unclear at a glance. Prefer passing `oldMetalakeEntity.name()` explicitly as the first argument (and keep `oldFullName` only for the `fullName` param) to make the roles obvious without relying on the reader knowing the metalake naming invariant.

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
```diff
@@ -322,8 +346,17 @@ public boolean deleteCatalog(NameIdentifier identifier, boolean cascade) {
                   mapper -> mapper.softDeleteStatisticsByCatalogId(catalogId)),
           () ->
               SessionUtils.doWithoutCommit(
-                  ViewMetaMapper.class,
-                  mapper -> mapper.softDeleteViewMetasByCatalogId(catalogId)));
+                  ViewMetaMapper.class, mapper -> mapper.softDeleteViewMetasByCatalogId(catalogId)),
+          () -> {
```
jerryshao, line 353: This intentionally writes only a single `CATALOG DROP` record on cascade delete rather than per-child records (schemas, tables, topics, etc.). That design was explained in the review thread but isn't captured in the code. Please add a comment like:

```java
// Only one CATALOG DROP record is written here; the cache layer performs prefix
// eviction on '<metalake>.<catalog>.*', covering all descendants automatically.
// Writing per-child DROP rows would multiply changelog volume without benefit.
```

Without this comment, a future maintainer will almost certainly "fix" the missing child records.

core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
```diff
@@ -273,46 +312,65 @@ public boolean deleteFileset(NameIdentifier identifier) {
...
         () ->
-            SessionUtils.doWithoutCommit(
-                FilesetMetaMapper.class,
-                mapper -> mapper.softDeleteFilesetMetasByFilesetId(filesetId)),
-        () ->
+            deleteResult.set(
+                SessionUtils.getWithoutCommit(
+                    FilesetMetaMapper.class,
+                    mapper -> mapper.softDeleteFilesetMetasByFilesetId(filesetId))),
+        () -> {
```
jerryshao, line 330: This restructuring changes behavior: previously all secondary soft-deletes (FilesetVersion, Owner, SecurableObject, Tag, Statistic, Policy) ran unconditionally as separate lambdas; now they only execute when `deleteResult.get() > 0`. This is actually an improvement (avoids orphaned cleanup when the primary delete is a no-op), but it is an unintended side-effect of the changelog work. Since `getFilesetPOByIdentifier` already guarantees the entity exists before this code runs, it is safe — but please acknowledge this behavior change in the PR description so reviewers don't miss it.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
yuqi1129: Add comments

core/src/test/java/org/apache/gravitino/storage/relational/mapper/provider/base/TestEntityChangeLogMapper.java
```diff
@@ -0,0 +1,150 @@
...
+    Mockito.when(config.get(Configs.ENTITY_RELATIONAL_JDBC_BACKEND_MAX_CONNECTIONS)).thenReturn(20);
+    Mockito.when(config.get(Configs.ENTITY_RELATIONAL_JDBC_BACKEND_WAIT_MILLISECONDS))
+        .thenReturn(1000L);
+
+    backend = new JDBCBackend();
+    backend.initialize(config);
+
+    sharedSession = SqlSessionFactoryHelper.getInstance().getSqlSessionFactory().openSession(true);
+    entityChangeLogMapper = sharedSession.getMapper(EntityChangeLogMapper.class);
+  }
```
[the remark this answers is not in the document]
yuqi1129: fixed

core/src/main/java/org/apache/gravitino/storage/relational/service/ViewMetaService.java
yuqi1129: fixed

core/src/main/java/org/apache/gravitino/storage/relational/mapper/EntityChangeLogSQLProviderFactory.java
yuqi1129: No need.

core/src/main/java/org/apache/gravitino/storage/relational/service/FilesetMetaService.java
yuqi1129: changed.

core/src/main/java/org/apache/gravitino/storage/relational/service/CatalogMetaService.java
yuqi1129: solved.

jerryshao (approved):

PR #10917, opened 2026-04-30
[#10910] feat(catalog-glue): Add integration tests for glue catalog

### LINKED ISSUE #10910: [Subtask] Add integration tests for glue catalog
### Describe the subtask

Add integration test coverage for table-level features in `catalog-glue`, including Hive table distribution/sort/comment/types and Iceberg table operations. Replace LocalStack-based integration tests with `GravitinoMotoContainer` since LocalStack free tier does not support the Glue API (returns HTTP 501).

### Parent issue

diqiu50:

Add integration test coverage for table-level features in catalog-glue, including Hive table distribution/sort/comment/types and Iceberg table operations.

Verify table-level features (Hive distribution, sort columns, comments, various types, Iceberg operations) work correctly with the AWS Glue backend.

Fix: #10910

No.

Unit tests and integration tests.


catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/integration/test/AbstractGlueCatalogIT.java
```diff
@@ -0,0 +1,661 @@
...
+    if (ops != null) {
+      ops.close();
+    }
+  }
+
+  @AfterEach
+  void cleanupSchema() {
+    if (currentSchema != null) {
+      try {
+        ops.dropSchema(schemaIdent(currentSchema), true);
```
yuqi1129, line 100: `dropSchema(schema, true)` does not actually delete tables in the current Glue implementation; it skips the non-empty check and then calls Glue `DeleteDatabase`, which real AWS Glue rejects for non-empty databases. Because this cleanup swallows the exception, any test that fails after creating a table can leak the database/table in the AWS account. Please make cleanup explicitly list/drop tables before deleting the schema, or implement real cascade semantics before relying on this.

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/integration/test/AwsGlueCatalogIT.java
```diff
@@ -0,0 +1,57 @@
...
+ * Required environment variables:
+ *
+ * <ul>
+ *   <li>{@code AWS_ACCESS_KEY_ID}
+ *   <li>{@code AWS_SECRET_ACCESS_KEY}
+ *   <li>{@code AWS_DEFAULT_REGION} (e.g. {@code us-east-1})
+ *   <li>{@code GLUE_CATALOG_ID} (12-digit AWS account ID; optional)
+ * </ul>
+ */
+@EnabledIfEnvironmentVariable(named = "AWS_ACCESS_KEY_ID", matches = ".+")
```
yuqi1129, line 40: This enables the real AWS IT when only `AWS_ACCESS_KEY_ID` is present, but `catalogConfig()` always passes both static credential keys and `GlueClientProvider` rejects a partial pair. If `AWS_SECRET_ACCESS_KEY` is missing, the test will fail during setup instead of being skipped. Please gate on both environment variables (or omit both keys and rely on the default AWS credential chain).

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/integration/test/AbstractGlueCatalogIT.java
diqiu50: The glue API succeeds on non-empty databases — tables are cleaned up asynchronously by the service. No leak here.

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/integration/test/AwsGlueCatalogIT.java
diqiu50: Checking AWS_ACCESS_KEY_ID to enable the test is enough. If a user configures the access key but not the secret key, it’s fine for the test to fail.

yuqi1129 (approved):
LGTM

PR #10930, opened 2026-04-30
[#10907] fix(auth): Backport JCasbin policy lookup optimization to 1.2.0-hotfix

yuqi1129:

This PR backports the optimization from #10908 to `1.2.0-hotfix`.

The backport includes:
- caching the current request's user role IDs in `AuthorizationRequestContext`
- building and using a per-role policy index in `JcasbinAuthorizer`
- keeping user-role assignment changes immediately visible during authorization
- backporting the related regression tests

The previous JCasbin authorization path relied on `enforce` for each privilege probe and could still observe stale user-role links cached in the enforcer.

This backport keeps the hotfix branch aligned with the optimized authorization path from branch-1.2, reduces the lookup cost on the authorization hot path, and makes role assignment changes take effect immediately.

Fix: #10907

No.

`./gradlew --no-daemon :server-common:test --tests org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer :core:test --tests org.apache.gravitino.authorization.TestAuthorizationRequestContext -PskipITs -PskipDockerTests=false`

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -40,6 +40,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.Configs;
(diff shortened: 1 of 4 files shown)


server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -440,20 +447,59 @@ private boolean loadPrivilegeAndAuthorize(
...
-          String.valueOf(userId),
-          String.valueOf(metadataObject.type()),
-          String.valueOf(metadataId),
-          privilege);
+      Set<Long> roleIds = requestContext.getUserRoleIds();
+      if (roleIds.isEmpty()) {
+        return false;
+      }
+      PolicyKey key = new PolicyKey(metadataObject.type().name(), metadataId, privilege);
+      if (authorizationMode == AuthorizationMode.DENY) {
```
roryqi: This seems to break old logic.

PR #10932, opened 2026-04-30
Fix Helm chart init containers against current Gravitino image

markhoerth:
### Problem

When `helm install` is run with `postgresql.enabled=true` or `mysql.enabled=true`, the Gravitino server pod enters CrashLoopBackOff and never starts. The failure is in the `sqlfile` init container with `Permission denied` errors.

Root cause: the `sqlfile` init container runs the `apache/gravitino` image and copies bundled schema scripts and the server JAR out of it. The container's script hardcodes paths under `/opt/gravitino/`, but the image actually places those files under `/root/gravitino/`. The `cp` and `ls` commands fail, the init container exits with an error, and the pod never reaches the main container.

The default H2 install works because it doesn't run the failing init containers.

### Fix

1. New `gravitinoHome` values key (defaults to `/root/gravitino`) so paths into the image's filesystem are configurable instead of hardcoded. The `sqlfile` init container and the main container both consume it via `GRAVITINO_HOME`.
2. Init and main container security contexts default to `runAsUser: 0`. The image's `/root` directory is mode `0700`, so init containers need root to traverse into `/root/gravitino/scripts/`, and the main container's launch script needs to write to `/root/gravitino/libs/` and `/root/gravitino/logs/` at startup.

Both are workarounds for the image's current layout. A proper fix at the image layer (move bundled files to a world-traversable path, or chown to a non-root user) would let everything run as non-root. Out of scope for this PR.

### Tests

Tested on Docker Desktop Kubernetes against `apache/gravitino:1.3.0-SNAPSHOT`. Three scenarios, all reach `1/1 Running` and serve `GET /api/version` successfully:

- Default install (H2)
- `--set postgresql.enabled=true`
- `--set mysql.enabled=true`

### DIFF
--- a/dev/charts/gravitino/templates/deployment.yaml
+++ b/dev/charts/gravitino/templates/deployment.yaml
@@ -64,13 +64,16 @@ spec:
             - /bin/bash
             - -c
             - |
-              cp -r /opt/gravitino/scripts/*  /tmp/scripts/
-              VERSION=$(ls /opt/gravitino/libs/gravitino-server-* | grep -oP '[0-9]+\.[0-9]+\.[0-9]+'|head -1)
+              cp -r ${GRAVITINO_HOME}/scripts/*  /tmp/scripts/
+              VERSION=$(ls ${GRAVITINO_HOME}/libs/gravitino-server-* | grep -oP '[0-9]+\.[0-9]+\.[0-9]+'|head -1)
               echo $VERSION > /tmp/scripts/version.txt
           resources:
             {{- toYaml .Values.initResources | nindent 12 }}
           securityContext:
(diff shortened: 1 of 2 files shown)


markhoerth:
Closing — after looking at the file's commit history, this isn't a chart bug. The 1.3.0-SNAPSHOT image hasn't been rebuilt since #10681 landed (image built Apr 7, that PR landed Apr 15), so the chart on main expects the new layout from #10681 but the published SNAPSHOT image still has the old layout. A fresh SNAPSHOT build resolves this with no chart change needed. Released versions of the chart and image are matched and unaffected.

PR #10975, opened 2026-05-06
[Improvement] Add tlsConfigurer for Gravitino #9836

### LINKED ISSUE #9836: [Improvement] Add tlsConfigurer for Gravitino

In many scenarios, we need to trust clients. A recommended approach is to issue SSL certificates to trusted clients, ensuring that only authorized clients can access the service.

### How should we improve?

We can refer to https://github.com/apache/iceberg/pull/13190

Octavi00:
Co-authored with @Vonna9 and @jujusjuicebox

[Improvement] Add tlsConfigurer for Gravitino #9836
   

The following changes are unit tests to test the proposed SSL certificate authentication for clients.

The following changes were added to replicate the issue, which was claimed to be the lack of mTLS, and we decided to see if improvements toward SSL certificates were actually needed. To be able to confirm or deny such, we decided to add a couple of tests to the common/src/test/java/org/apache/gravitino/server/web/TestJettyServer.java to prove if SSL certificates fail in either keystores or during client authentications. We found that mTLS is already automatically supported by the Gravitino Server, and clients are able to respond to the mTLS request. In such a case, that means this Improvement wasn't actually needed and was most likely addressed in a separate pr.

Fix: #9836 

No

Using the command 
./gradlew :server-common:test --tests "org.apache.gravitino.server.web.TestJettyServer", the build succeeded with the newly added tests proving mTLS support.


roryqi:
You can refer to https://github.com/apache/iceberg/pull/13190 You need to modify the client code.

Octavi00:
Hello @roryqi, me and @Vonna9 spent some time implementing the TLS functionality on the client side of the code. At first, since the issue stated we needed a way to trust clients, our understanding was that we were from the perspective of the server, which is why we felt confused about the server functionality already being there, and just added a test to prove so. Now, what we've done is add in a TLSConfigurer Java file in the same fashion as the reference pr, and implemented it in a very similar method inside the HTTPClient.java file. We found quite a few bigger differences between the gravitino codebase and the iceberg codebase, which led to some changes I would really like to explain. The most important thing is that in order to allow testing the TLS integration from the TestHTTPClient file, I removed the private keyword from the configureConnectionManager() function. In doing so, the function remains package private so it's accessible within the same Java package that is used in the test file. I wanted to make sure you were aware of this to ensure it isn't sacrificing any security boundaries you intend to keep. I also added the properties variable as a parameter in the function, as previously the function would only rely on the GravitinoClientConfiguration object defined functions; however, properties isn't accessible from the object. I chose to keep that object separate from the TLS configuration instead of implementing it within the Gravitino client object, but it could also easily be moved into it for a cleaner HTTPClient file. Testing-wise, I implemented all of the tests in a similar fashion from the referenced pr, and added one more to ensure previous functionality without a TLS configuration. All tests currently pass, and the server compiles and runs. I feel that we are ready for a review now, and please let us know if there's a different direction you'd like us to take our implementation. 

Here are some photos of our testing

test file passes
image

server compiles
image

server running terminal notification
image


Thank you

clients/client-java/src/main/java/org/apache/gravitino/client/HTTPClient.java
```diff
@@ -730,9 +735,65 @@ private static HttpClientConnectionManager configureConnectionManager(
 
     ConnectionConfig connectionConfig = configureConnectionConfig(clientConfiguration);
     connectionManagerBuilder.setDefaultConnectionConfig(connectionConfig);
+
+    // Configure custom TLS settings, if provided
+    TLSConfigurer tlsConfigurer = loadTlsConfigurer(properties);
```
roryqi: Properties are flexible. It can't give enough constraint to client.
I prefer add a `withTlsConfigurer` method for the builder instead of using reflection.

clients/client-java/src/main/java/org/apache/gravitino/client/TLSConfigurer.java
```diff
@@ -0,0 +1,72 @@
...
+
+/** Configures TLS settings for the HTTP client. */
+public interface TLSConfigurer {
+
+  /**
+   * Initializes the TLS configurer with client properties.
+   *
+   * @param properties client configuration properties
+   */
+  default void initialize(Map<String, String> properties) {}
```
roryqi: Could we avoid use initialize methods? The reason is like above. Our TLSConfugurer shouldn't initialize by properties.

clients/client-java/src/main/java/org/apache/gravitino/client/TLSConfigurer.java
```diff
@@ -0,0 +1,72 @@
...
+    return HttpsSupport.getDefaultHostnameVerifier();
+  }
+
+  /**
+   * Returns the supported TLS protocols.
+   *
+   * @return supported TLS protocols
+   */
+  default String[] supportedProtocols() {
+    return null;
```
roryqi, line 53: Could u return a empty array instead of null value? It's hard to handle null value.

clients/client-java/src/main/java/org/apache/gravitino/client/TLSConfigurer.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Returns the supported cipher suites.
+   *
+   * @return supported cipher suites
+   */
+  default String[] supportedCipherSuites() {
```
roryqi, line 61: Ditto.

roryqi (commented):
I prefer you can test it  by the e2e test. Hopes tls configurer works.

Octavi00:
Hello @roryqi, i made some big changes as requested, and I will list most of them so it's easier to read. Instead of using an initialize method as previously done so in Iceberg, I went with a withTLSConfigurer method. This required large modifications and led to some more files needing adjustments. 

To start, due to the fact that a user makes a gravitino client base, and the flow is base -> adminClient or client -> http
- All 4 files had the .withTlsConfigurer() function added.

To maintain all prior test cases + ensure existing code is unaffected: 
- The same default constructor was kept by initializing the new TLSConfigurer field to null
* Possible future revisions/additions might not want to keep making several different constructors, but rather an object that holds configurators such as TLS within it to avoid large, redundant code. Regardless, making another constructor wasn't reasonably avoidable in this case.

-E2E was locally tested using a custom test file and a custom test, which can be seen below, and passes the expected behaviour
Screenshot 2026-05-28 224619

-gravitino.conf.template was updated to make an easier user experience when setting up HTTPS when using TLS

- Two simple TLS unit tests were made inside TestHTTPClient to replace prior unit tests 
* Maybe will want similar ones within the 3 gravitino objects for new constructors, but all prior behaviour remains the same 

Thank you

nevzheng:
Hi @Octavi00 — @roryqi asked me to help review this PR. I’m also working on the related TLS changes in #12011, so I want to be transparent about that context.

Before beginning a detailed review or comparing the approaches, could you let us know whether you and the other contributors are still interested in continuing this work? This PR currently conflicts with `main` and has not been updated recently.

If you would like to continue, I’m happy to engage here and help work through the next steps. A good first step would be to rebase the PR onto the current `main` branch and resolve the conflicts. We can then review the updated scope and decide how best to proceed.

Could you please let us know by end of day Friday, July 17? Thanks!

Octavi00:
Hi @nevzheng, yes me and @Vonna9 would be willing to continue working on this issue. It's a somewhat old branch, and I know a lot has changed with the V2 web switch, so we will get to rebasing it by the end of Friday if that's alright. I'm predicting the first changes we will need to make will involve redundant/problematic code if there have been large changes to the Java server.

nevzheng:
@Octavi00! Great. I’ll leave it to you. I’m subscribed to the thread. Tag me for review / questions. Thanks!

server-common/src/test/java/org/apache/gravitino/server/web/TestJettyServer.java
```diff
@@ -95,4 +95,40 @@ public void testStopWithNullServer() {
   public void testStartWithoutInitialise() throws InterruptedException {
     assertThrows(RuntimeException.class, () -> jettyServer.start());
   }
+
+  @Test
+  public void testInitializeHttpsWithClientAuthRequiresTrustStore() throws IOException {
```
nevzheng: **[Blocking] The PR does not currently satisfy the functional end-to-end requirement of #9836.**

The added tests only verify that client and server objects can be constructed. They do not establish a TLS connection, perform a handshake, present a client certificate, or make an actual request.

Please add functional end-to-end coverage that proves the feature works, including both success and failure cases. At minimum:

- A server requiring client authentication accepts a request from a client configured with a trusted certificate and private key.
- The same server rejects a client that does not provide a certificate.
- The server rejects a client whose certificate is not trusted.
- The client rejects an untrusted server certificate.
- Existing behavior still works when no custom TLS configuration is supplied and client authentication is not required.

The negative client-certificate tests should still trust the server certificate so that they fail specifically because client authentication failed.

This coverage is required to demonstrate that the PR actually implements the requested mTLS behavior, rather than only adding configuration plumbing.

clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoAdminClient.java
```diff
@@ -65,7 +66,28 @@ private GravitinoAdminClient(
...
+   *
+   * @param uri The base URI for the Gravitino API.
+   * @param authDataProvider The provider of the data which is used for authentication.
+   * @param checkVersion Whether to check the version of the Gravitino server. Gravitino does not
+   *     support the case that the client-side version is higher than the server-side version.
+   * @param headers The base header for Gravitino API.
+   * @param properties A map of properties (key-value pairs) used to configure the Gravitino client.
+   * @param tlsConfigurer The TLS configurer for the Gravitino client.
+   */
+  private GravitinoAdminClient(
```
nevzheng, line 65: **[Non-blocking design] Inject HTTP behavior instead of extending the Gravitino client hierarchy**

My main concern is that this change extends `GravitinoAdminClient`, `GravitinoClient`, and `GravitinoClientBase` with a transport-specific concern. We are not adding new Gravitino client behavior—we are only changing how the underlying HTTP connection is constructed.

I suggest keeping `TLSConfigurer` entirely within `HTTPClient` and introducing a shared client-construction path that can configure the HTTP builder. For example:

```java
GravitinoClientFactory factory =
    GravitinoClientFactory.builder(uri)
        .configureHttp(http -> http.withTlsConfigurer(tlsConfigurer))
        .build();

GravitinoClient client = factory.createClient(metalakeName);
AdminClient admin = factory.createAdminClient();
```

The implementation could use a non-null, no-op HTTP configuration hook:

```java
private Consumer<HTTPClient.Builder> httpConfigurer = ignored -> {};

public Builder configureHttp(Consumer<HTTPClient.Builder> configurer) {
  this.httpConfigurer = Objects.requireNonNull(configurer);
  return this;
}

private RESTClient createRestClient() {
  HTTPClient.Builder http =
      HTTPClient.builder(properties)
          .uri(uri)
          .withAuthDataProvider(authDataProvider)
          .withHeaders(headers);

  httpConfigurer.accept(http);
  return http.build();
}
```

This gives us one HTTP extension point without adding `TLSConfigurer` fields, constructor parameters, or duplicated builder methods to every Gravitino client type. It also avoids nullable TLS state: the default behavior is simply a no-op HTTP configuration.

As a follow-up, URI, authentication, headers, and other HTTP settings could move into this same transport-construction path, and the concrete Gravitino clients could accept only the resulting `RESTClient`.

I am okay accepting the current structure in this PR, but I recommend at least removing the unused protected `tlsConfigurer` field now. The functional focus of this change should remain configuring `HTTPClient` with the `SSLContext` required for client-certificate authentication.

nevzheng (changes requested):

nevzheng (commented):
Hi, thanks for the PR.

**Blocking:** This needs committed end-to-end mTLS tests that perform a real handshake and validate the behavior: a trusted client certificate succeeds, while a missing or untrusted client certificate is rejected.

**Architecture:** I don’t think `TLSConfigurer` should be added to `GravitinoAdminClient`, `GravitinoClient`, or `GravitinoClientBase`. We are injecting behavior into the HTTP transport, not extending the Gravitino client API.

Please add or extend a shared builder so it can configure the `HTTPClient.Builder` directly:

```java
builder.configureHttp(
    http -> http.withTlsConfigurer(tlsConfigurer));
```

TLS then remains owned by `HTTPClient`, without being passed through the Gravitino client hierarchy. As a follow-up, other HTTP-specific properties such as headers and authentication can move into the same construction path.

nevzheng:
@Octavi00 
Hi, thanks for the PR.

**Blocking:** This needs committed end-to-end mTLS tests that perform a real handshake and validate the behavior: a trusted client certificate succeeds, while a missing or untrusted client certificate is rejected.

**Architecture:** I don’t think `TLSConfigurer` should be added to `GravitinoAdminClient`, `GravitinoClient`, or `GravitinoClientBase`. We are injecting behavior into the HTTP transport, not extending the Gravitino client API.

Please add or extend a shared builder so it can configure the `HTTPClient.Builder` directly:

```java
builder.configureHttp(
    http -> http.withTlsConfigurer(tlsConfigurer));
```

TLS then remains owned by `HTTPClient`, without being passed through the Gravitino client hierarchy. As a follow-up, other HTTP-specific properties such as headers and authentication can move into the same construction path.

roryqi:
@Octavi00 Comments seem not to be addressed. If you don't have time, I will review this pull request https://github.com/apache/gravitino/pull/12011 and add you as the collaborator.

Octavi00:
> @Octavi00 Comments seem not to be addressed. If you don't have time, I will review this pull request https://github.com/apache/gravitino/pull/12011 and add you as the collaborator.

Hi, my team had planned to meet today in the afternoon to start making the changes. I didnt want to jump on it without including them, but we are definitely still interested in finishing up this pr.

Octavi00:
@nevzheng I have implemented an http injection method as requested, and I wanted to ask if you'd still like to see GravitinoAdminClient and GravitinoClient keep their original constructor and add the new httpConfigurer in another constructor? I had assumed that wouldn't be necessary, as the new parameter httpConfigurer is optional, but I still kept the original constructor for GravitinoClientBase since I feel that it's already in use by other subclasses, and there may be a large change if I were to force all bases to be constructed with the new parameter.

My second question is about testing. Would you want me to create some new pkcs12 certs and drop them in the resources folder within the Jetty server test folder? To me, this seems reasonable as long as I include "test" within their individual names, and that way they can be reused elsewhere if needed in the future. Otherwise, I think the second option would be to create them at runtime and delete them afterwards, but that seems like a spot for failure and would just be more code to execute during tests?

Thanks

nevzheng:
@Octavi00 Please keep the existing protected `GravitinoClientBase` constructor and have it delegate to the new constructor with a no-op HTTP configuration. This preserves compatibility for subclasses. There is no need to retain the old private constructors in `GravitinoClient` or `GravitinoAdminClient`. We can handle the test-certificate question separately.

nevzheng:
@Octavi00 I checked the current repository and this PR and did not find existing test keystore or truststore fixtures to reuse. Please create static PKCS12 fixtures for the integration tests rather than generating and deleting certificates at runtime. Store them under the `src/test/resources/tls` directory of the module that owns the integration test. Clearly mark the filenames, aliases, certificate subjects, and passwords as test-only, and include brief regeneration instructions. The integration tests should use these fixtures to perform real TLS handshakes.

Octavi00:
Hi @nevzheng, I've applied the mTLS tests as you requested. I created the necessary certs plus added a readme and a regen script for easy understanding. Everything is passing on my end, and nothing is catching either my eye or my group's eye, so I think it's ready for another review. Please let me know if anything stands out. Thank you.

clients/client-java/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ dependencies {
   implementation(project(":common")) {
     exclude(group = "org.apache.logging.log4j")
   }
+  implementation(project(":catalogs:hive-metastore-common"))
```
nevzheng: **issue (blocking):** Remove the unused Hive dependency from the Java client

This PR adds `implementation(project(":catalogs:hive-metastore-common"))` to `clients/client-java/build.gradle.kts`, but nothing in the change uses that module. Please remove it — it pulls catalog/Hive dependencies into the public client artifact.

clients/client-java/src/test/java/org/apache/gravitino/client/TestHTTPClient.java
```diff
@@ -356,6 +356,23 @@ public void testSocketTimeout() throws IOException {
...
 
   @Test
+  public void testBuildWithoutTLSConfigurer() {
+    HTTPClient client = HTTPClient.builder(ImmutableMap.of()).uri("http://localhost").build();
+
+    Assertions.assertNotNull(client);
+  }
+
+  @Test
+  public void testBuildWithTLSConfigurer() {
```
nevzheng: **blocking:** Please test the Gravitino client TLS path

The current tests exercise Jetty via `java.net.http.HttpClient` with a hand-built `SSLContext`. That does not cover `HTTPClient.withTlsConfigurer` / `configureHttp(...)`.

Please add CUJ-level tests on the Gravitino client path. Example shape:

```java
TLSConfigurer tlsConfigurer =
    new TLSConfigurer() {
      @Override
      public SSLContext sslContext() {
        // load client truststore (+ optional client keystore) from test fixtures
        return sslContext;
      }
    };

try (HTTPClient client =
    HTTPClient.builder(ImmutableMap.of())
        .uri("https://localhost:" + port)
        .withTlsConfigurer(tlsConfigurer)
        .build()) {
  // assert success or handshake failure per case below
}
```

Same idea via the public builder is also fine:

```java
GravitinoClient.builder("https://localhost:" + port)
    .configureHttp(http -> http.withTlsConfigurer(tlsConfigurer))
    .build();
```

### Test matrix

| # | Server | Client truststore | Client keystore | Expected |
|---|---|---|---|---|
| 1 | HTTPS, client auth **required** | `test-client-truststore.p12` | `test-trusted-client-keystore.p12` | **Success** |
| 2 | HTTPS, client auth **required** | `test-client-truststore.p12` | _(none)_ | **Handshake failure** |
| 3 | HTTPS, client auth **required** | `test-client-truststore.p12` | `test-untrusted-client-keystore.p12` | **Handshake failure** |
| 4 | HTTPS, client auth **off** | `test-client-truststore.p12` | _(none)_ | **Success** |
| 5 | HTTPS, client auth **off** | `test-untrusted-server-truststore.p12` | _(none)_ | **Handshake failure** |

clients/client-java/src/main/java/org/apache/gravitino/client/TLSConfigurer.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.apache.gravitino.client;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLContext;
+import org.apache.hc.client5.http.ssl.HttpsSupport;
+import org.apache.hc.core5.ssl.SSLContexts;
+
+/** Configures TLS settings for the HTTP client. */
+public interface TLSConfigurer {
```
nevzheng, line 27: **praise:** Nice progress on the `configureHttp` / `TLSConfigurer` injection and the PKCS12 + Jetty mTLS fixtures.

**suggestion (non-blocking):** Follow-up — better client UX for TLS setup

`TLSConfigurer` is a solid extension point, but callers still have to assemble an `SSLContext` by hand. As a follow-up, please add a small helper so the common CUJ is just paths + passwords, for example:

```java
TLSConfigurer tls =
    TLSConfigurers.builder()
        .trustStore(Path.of("/path/to/truststore.p12"), "changeit")
        .keyStore(Path.of("/path/to/keystore.p12"), "changeit") // optional, for mTLS
        .storeType("PKCS12") // optional, default PKCS12
        .build();

GravitinoClient.builder("https://gravitino.example.com")
    .configureHttp(http -> http.withTlsConfigurer(tls))
    .build();
```

Ideally also support the same via env vars / client config, plus a short doc section for:
1. HTTPS with custom truststore
2. mTLS with client keystore + truststore

If you prefer not to do that in this PR, please file a ticket describing that follow-up.

nevzheng (changes requested):
Thanks for the work so far.

**Changes requested:**
1. Remove the unused `hive-metastore-common` dependency from the Java client.
2. Add CUJ-level positive/negative tests through the Gravitino client TLS path (`withTlsConfigurer` / `configureHttp`), not only JDK `HttpClient`.
3. Follow-up (or file a ticket): better client UX so callers can configure TLS from keystore/truststore paths instead of hand-building an `SSLContext`.

Octavi00:
Hi @nevzheng, while I was writing some CUJ level tests as you've requested, I've come across a couple of questions. The first being if it'd be appropriate to create a whole new test file named TestHTTPClientTLS.java inside the clients/client-java/src/test/java/org/apache/gravitino/client path, which would house the newer TLS specific tests, instead of adding them into the already large HTTP test file. The second question would be: since I'd also need to use the same .p12 certificates as in the server-common test, would it make sense to move the certificates into a folder further outside of the path so that they can be globally accessed for testing purposes, or should I drop a copy of the certificates folder into the client test resource path?

nevzheng:
Hi @Octavi00 — thanks for checking in.

The goal is an integration-style test of the **client → server** path with the correct TLS configuration: `HTTPClient.withTlsConfigurer` / `configureHttp(...)` talking to a TLS-enabled server, covering the success and failure cases from the earlier matrix. Focus on getting that working first; we can review the specifics afterward.

To unblock your two questions:

1. **New test file** — yes, please add a dedicated file such as `TestHTTPClientTLS.java` under `clients/client-java/src/test/java/org/apache/gravitino/client/` rather than growing `TestHTTPClient`. That keeps the TLS CUJs easy to find and maintain.

2. **Certificate fixtures** — copy the PKCS12 fixtures into `clients/client-java/src/test/resources/tls/` (same layout as `server-common`). Duplication is fine for small test-only stores. Please don’t invent a repo-global shared path for this. `client-java` already depends on `server-common` for tests, so you can start a Jetty HTTPS server from the client module and point the Gravitino HTTP client at it.

You’re empowered to make the remaining local decisions in pursuit of that goal — I trust your judgment. Tag me when you have something ready for another look.

Octavi00:
Hi @nevzheng, just wanted to check in about my other comment I had left for adding a better client UX. From my last commit, I had added some helper functions within the TLSConfigurers file to suit your request, but im not sure if you'd like for me to take a deeper step and do the same for HTTPS in this pr?

roryqi:
Could u fix the Ci failure first?

nevzheng:
Hi @Octavi00,

We've been in process for a while, and I think having some structure would help everyone involved.

I'm noticing the original issue is an XL task that should be broken down into more maintainable deliverables + PRs. I think that would make each PR easier to review, and help everyone with a sense of progress.

For me as reviewers: Seeing so many. files w/o a narrative or milestone attached to it is hard to review

For authors: its frustrating because we end up with large PRs that don't get feedback or input on.

let me follow up on the ticket with a proposal about how we can land this work more effectively. I will tag you.

nevzheng:
Deprecating this PR in favor of the new milestones + break down. Please keep the remote branch for reference as you split the PR.

PR #10984, opened 2026-05-06
[#10868] feat: Prune stale group-inherited role mappings from enforcer

bharos:

Follow-up to #10868. Adds add-first-then-prune logic to `loadRolePrivilege` to clean up stale g-rows (user→role mappings) in the jCasbin enforcer when a user's group membership changes at the IdP level.

**Changes:**
- Collect all currently valid roleIds into a `desiredRoleIds` set during role loading
- After all role futures complete, call `getRolesForUser` on the enforcer and remove any role not in the desired set via `deleteRoleForUser`
- Update tests to reflect that stale g-rows are now pruned immediately

Without this change, when a user is removed from a group at the IdP level (e.g., LDAP/OIDC), the stale role mappings persist in the enforcer until the cache TTL expires. This means the user retains permissions from the removed group for up to the cache expiration period.

The add-first-then-prune approach was chosen over alternatives:
- **`deleteUser` then re-add**: Creates a false-denial window where concurrent requests see zero roles
- **userId→groups cache**: Not future-proof for multi-node (group membership comes from IdP token, not DB, so can't be invalidated via DB versioning)
- **Add-then-prune** (this PR): No false-denial window, no extra cache, multi-node friendly — each node converges independently

Follow-up for: #10868

No. Internal authorization enforcement behavior only.

- Updated `testAuthorize` — verifies that when a user's role changes, the old role is pruned immediately
- Updated `testUserRemovedFromGroupAtIdpDeniesAccess` — verifies that removing a user from a group at IdP level immediately denies access (no longer a known limitation)
- All 20 existing tests pass: `./gradlew :server-common:test --tests "*.TestJcasbinAuthorizer"`


bharos:
Closing this follow-up PR — the add-then-prune change has been incorporated directly into #10868 per reviewer feedback.

PR #10993, opened 2026-05-08
[#10956] fix(catalog-glue): Move AWS Glue integration tests to integration/test path

### LINKED ISSUE #10956: [Improvement] Prevent accidental AWS Glue API calls during local testing.

The current location of `TestAwsGlueSchema.java` and `TestAwsGlueTable.java` causes them to run during standard unit test builds if a developer has ambient AWS credentials configured locally. 

Because the files reside in `catalogs/catalog-glue/src/test/java/...`, they bypass the Gradle `exclude("**/integration/test/**")` rule tied to the `skipITs` flag. If a contributor runs a standard `./gradlew catalogs:catalot-glue:test -PskipITs` while having `AWS_ACCESS_KEY_ID` set in their terminal profile, JUnit will automatically execute these live cloud integration tests. 

This can lead to unintended resource creation in personal or organizational AWS accounts, trigger security alerts, or cause accidental API costs for local developers.

### How should we improve?

To improve the local developer experience and ensure live AWS tests are only run when explicitly intended, we should resolve the boundary between the unit and integration tests. I propose two potential paths:

- Move `TestAwsGlueSchema.java` and `TestAwsGlueTable.java`  into a `integration/test/...` directory. This properly classifies them as integration tests and perfectly aligns with the existing `skipITs` Gradle logic, ensuring the test is safely skipped by default for local builds.

- Keep the test in the unit test directory, but decouple it from the real AWS network endpoint. We can refactor the test to use Mockito to mock the `GlueClient`.

sachinnn99:

Move `TestAwsGlueSchema` and `TestAwsGlueTable` from the unit test directory into the `integration/test/` subdirectory so they are excluded by the existing `skipITs` Gradle rule. Widen the abstract base classes (`AbstractGlueSchemaTest`, `AbstractGlueTableTest`) to `public` with `@VisibleForTesting` for cross-package access.

These tests hit real AWS Glue endpoints and run automatically when a developer has ambient AWS credentials (`AWS_ACCESS_KEY_ID`) set. Because they reside outside the `integration/test/` path, the `exclude("**/integration/test/**")` rule in `build.gradle.kts` does not filter them out during `./gradlew test -PskipITs`. This can cause unintended resource creation, security alerts, or API costs.

Fixes #10956

No.

Ran `./gradlew :catalogs:catalog-glue:test -PskipITs` and verified `TestAwsGlueSchema` and `TestAwsGlueTable` no longer appear in the test report while all other unit tests pass.

### DIFF
--- a/catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/AbstractGlueSchemaTest.java
+++ b/catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/AbstractGlueSchemaTest.java
@@ -34,7 +34,7 @@ import software.amazon.awssdk.services.glue.model.Database;
  * <p>Subclasses provide a {@link Database} object however they like (SDK builder, real AWS API,
  * etc.). The test scenarios are defined once here and shared across all implementations.
  */
-abstract class AbstractGlueSchemaTest {
+public abstract class AbstractGlueSchemaTest {
 
   /**
    * Returns a Glue {@link Database} with the given fields. Subclasses may create this via the SDK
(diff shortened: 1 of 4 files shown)


diqiu50:
@sachinnn99 Please resolve the conflict

sachinnn99:
Closing — upstream PR #10917 already addressed the original concern by gating the AWS-calling tests with `@EnabledIfEnvironmentVariable` (skipped by default) and adding Moto-based containerized integration tests. The `skipITs`-based approach from this PR is no longer needed.

PR #11013, opened 2026-05-09
[#11003] feat(catalog-lakehouse-paimon): add paimon view CRUD support and interoperability tests

### LINKED ISSUE #11003: [Subtask] Add Paimon catalog view CRUD support
### Describe the subtask

Add Paimon catalog view CRUD support

### Parent issue

#604

mchades:

- Added Paimon view CRUD support in Gravitino (`list/load/create/alter/rename/drop`).
- Added Paimon view metadata conversion and extracted view logic into a dedicated composed ops class.
- Exposed Paimon catalog as `ViewCatalog`.
- Added tests:
  - unit tests for view ops/wiring
  - integration tests for:
    - Spark create view -> Gravitino load view
    - Gravitino create view -> Spark query view

Paimon catalog did not fully support view CRUD in Gravitino. This PR completes Paimon view capability and adds interoperability coverage with Spark.

Fix: #11003

Yes.
- Paimon catalog now supports `ViewCatalog` operations in Gravitino.
- Added internal reserved hidden properties for view default catalog/schema handling.

- `./gradlew :catalogs:catalog-lakehouse-paimon:spotlessApply`
- `./gradlew :catalogs:catalog-lakehouse-paimon:test -PskipITs`
- `SKIP_DOCKER_TESTS=false ./gradlew :catalogs:catalog-lakehouse-paimon:test -PskipDockerTests=false --tests org.apache.gravitino.catalog.lakehouse.paimon.integration.test.CatalogPaimonHiveIT.testSparkCreateViewAndLoadByGravitino --tests org.apache.gravitino.catalog.lakehouse.paimon.integration.test.CatalogPaimonHiveIT.testGravitinoCreateViewAndReadBySpark`

### DIFF
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalog.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalog.java
@@ -31,6 +31,7 @@ import org.apache.gravitino.connector.PropertiesMetadata;
 import org.apache.gravitino.connector.capability.Capability;
 import org.apache.gravitino.credential.CredentialConstants;
 import org.apache.gravitino.credential.JdbcCredential;
+import org.apache.gravitino.rel.ViewCatalog;
 
 /**
  * Implementation of {@link Catalog} that represents an Apache Paimon catalog in Apache Gravitino.
@@ -63,6 +64,11 @@ public class PaimonCatalog extends BaseCatalog<PaimonCatalog> {
   @Override
   protected CatalogOperations newOps(Map<String, String> config) {
     return new PaimonCatalogOperations();
+  }
+
+  @Override
+  public ViewCatalog asViewCatalog() {
+    return (ViewCatalog) ops();
   }
 
   @Override
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java
@@ -49,14 +49,20 @@ import org.apache.gravitino.exceptions.NoSuchCatalogException;
 import org.apache.gravitino.exceptions.NoSuchColumnException;
 import org.apache.gravitino.exceptions.NoSuchSchemaException;
 import org.apache.gravitino.exceptions.NoSuchTableException;
+import org.apache.gravitino.exceptions.NoSuchViewException;
 import org.apache.gravitino.exceptions.NonEmptySchemaException;
 import org.apache.gravitino.exceptions.SchemaAlreadyExistsException;
 import org.apache.gravitino.exceptions.TableAlreadyExistsException;
+import org.apache.gravitino.exceptions.ViewAlreadyExistsException;
 import org.apache.gravitino.meta.AuditInfo;
 import org.apache.gravitino.rel.Column;
+import org.apache.gravitino.rel.Representation;
 import org.apache.gravitino.rel.TableCatalog;
 import org.apache.gravitino.rel.TableChange;
 import org.apache.gravitino.rel.TableChange.RenameTable;
+import org.apache.gravitino.rel.View;
+import org.apache.gravitino.rel.ViewCatalog;
+import org.apache.gravitino.rel.ViewChange;
 import org.apache.gravitino.rel.expressions.NamedReference;
 import org.apache.gravitino.rel.expressions.distributions.Distribution;
 import org.apache.gravitino.rel.expressions.distributions.Distributions;
@@ -77,11 +83,13 @@ import org.slf4j.LoggerFactory;
  * Implementation of {@link CatalogOperations} that represents operations for interacting with the
  * Apache Paimon catalog in Apache Gravitino.
  */
-public class PaimonCatalogOperations implements CatalogOperations, SupportsSchemas, TableCatalog {
+public class PaimonCatalogOperations
+    implements CatalogOperations, SupportsSchemas, TableCatalog, ViewCatalog {
 
   public static final Logger LOG = LoggerFactory.getLogger(PaimonCatalogOperations.class);
 
   @VisibleForTesting public PaimonCatalogOps paimonCatalogOps;
+  @VisibleForTesting PaimonViewCatalogOps paimonViewCatalogOps;
 
   private static final String NO_SUCH_SCHEMA_EXCEPTION =
       "Paimon schema (database) %s does not exist.";
@@ -117,6 +125,9 @@ public class PaimonCatalogOperations implements CatalogOperations, SupportsSchem
     resultConf.putAll(gravitinoConfig);
 
     this.paimonCatalogOps = new PaimonCatalogOps(new PaimonConfig(resultConf));
+    this.paimonViewCatalogOps =
+        new PaimonViewCatalogOps(
+            paimonCatalogOps, this::buildPaimonNameIdentifier, this::schemaExists);
   }
 
   /**
@@ -463,6 +474,84 @@ public class PaimonCatalogOperations implements CatalogOperations, SupportsSchem
     }
     LOG.info("Purged Paimon table {}.", identifier);
     return true;
+  }
+
+  /**
+   * Lists all views under the specified namespace.
+   *
+   * @param namespace The namespace to list views for.
+   * @return An array of {@link NameIdentifier} representing the views in the namespace.
+   * @throws NoSuchSchemaException If the schema with the provided namespace does not exist.
+   */
+  @Override
+  public NameIdentifier[] listViews(Namespace namespace) throws NoSuchSchemaException {
+    return paimonViewCatalogOps.listViews(namespace);
+  }
+
+  /**
+   * Loads the view with the provided identifier.
+   *
+   * @param identifier The identifier of the view to load.
+   * @return The loaded {@link View} instance representing the view metadata.
+   * @throws NoSuchViewException If the view with the provided identifier does not exist.
+   */
+  @Override
+  public View loadView(NameIdentifier identifier) throws NoSuchViewException {
+    return paimonViewCatalogOps.loadView(identifier);
+  }
+
+  /**
+   * Creates a new view with the provided metadata.
+   *
+   * @param identifier The identifier of the view to create.
+   * @param comment The view comment.
+   * @param columns The output columns of the view.
+   * @param representations The view SQL representations.
+   * @param defaultCatalog The default catalog for unqualified identifiers.
+   * @param defaultSchema The default schema for unqualified identifiers.
+   * @param properties The view properties.
+   * @return The created view metadata.
+   * @throws NoSuchSchemaException If the target schema does not exist.
+   * @throws ViewAlreadyExistsException If the view already exists.
+   */
+  @Override
+  public View createView(
+      NameIdentifier identifier,
+      String comment,
+      Column[] columns,
+      Representation[] representations,
+      String defaultCatalog,
+      String defaultSchema,
+      Map<String, String> properties)
+      throws NoSuchSchemaException, ViewAlreadyExistsException {
+    return paimonViewCatalogOps.createView(
+        identifier, comment, columns, representations, defaultCatalog, defaultSchema, properties);
+  }
+
+  /**
+   * Alters an existing view according to the provided changes.
+   *
+   * @param identifier The identifier of the view to alter.
+   * @param changes The changes to apply.
+   * @return The updated view metadata.
+   * @throws NoSuchViewException If the view does not exist.
+   * @throws IllegalArgumentException If any change type is unsupported by Paimon.
+   */
+  @Override
+  public View alterView(NameIdentifier identifier, ViewChange... changes)
+      throws NoSuchViewException, IllegalArgumentException {
+    return paimonViewCatalogOps.alterView(identifier, changes);
+  }
+
+  /**
+   * Drops the view with the provided identifier.
+   *
+   * @param identifier The identifier of the view to drop.
+   * @return true if the view is successfully dropped, false if the view does not exist.
+   */
+  @Override
+  public boolean dropView(NameIdentifier identifier) {
+    return paimonViewCatalogOps.dropView(identifier);
   }
 
   @Override
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
@@ -61,7 +61,11 @@ public class PaimonTablePropertiesMetadata extends BasePropertiesMetadata {
             stringImmutablePropertyEntry(
                 ROWKIND_FIELD, "The table rowkind field", false, null, false, false),
             stringReservedPropertyEntry(PRIMARY_KEY, "The table primary key", false),
-            stringReservedPropertyEntry(PARTITION, "The table partition", false));
+            stringReservedPropertyEntry(PARTITION, "The table partition", false),
+            stringReservedPropertyEntry(
+                PaimonView.DEFAULT_CATALOG_PROPERTY, "The view default catalog", true),
+            stringReservedPropertyEntry(
+                PaimonView.DEFAULT_SCHEMA_PROPERTY, "The view default schema", true));
     PROPERTIES_METADATA = Maps.uniqueIndex(propertyEntries, PropertyEntry::getName);
   }
 
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
@@ -1,0 +1,214 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.catalog.lakehouse.paimon;
+
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.meta.AuditInfo;
+import org.apache.gravitino.rel.Column;
+import org.apache.gravitino.rel.Representation;
+import org.apache.gravitino.rel.SQLRepresentation;
+import org.apache.gravitino.rel.View;
+import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.types.DataField;
+import org.apache.paimon.view.ViewImpl;
+
+/** Represents a Gravitino view object converted from Apache Paimon view metadata. */
+final class PaimonView implements View {
+
+  static final String DEFAULT_CATALOG_PROPERTY = "gravitino.view.default-catalog";
+  static final String DEFAULT_SCHEMA_PROPERTY = "gravitino.view.default-schema";
+
+  private static final String DEFAULT_DIALECT = "default";
+
+  private final String name;
+  @Nullable private final String comment;
+  private final Column[] columns;
+  private final Representation[] representations;
+  @Nullable private final String defaultCatalog;
+  @Nullable private final String defaultSchema;
+  private final Map<String, String> properties;
+  private final AuditInfo auditInfo;
+
+  private PaimonView(
+      String name,
+      @Nullable String comment,
+      Column[] columns,
+      Representation[] representations,
+      @Nullable String defaultCatalog,
+      @Nullable String defaultSchema,
+      Map<String, String> properties,
+      AuditInfo auditInfo) {
+    this.name = name;
+    this.comment = comment;
+    this.columns = Arrays.copyOf(columns, columns.length);
+    this.representations = Arrays.copyOf(representations, representations.length);
+    this.defaultCatalog = defaultCatalog;
+    this.defaultSchema = defaultSchema;
+    this.properties = new HashMap<>(properties);
+    this.auditInfo = auditInfo;
+  }
+
+  static PaimonView fromPaimonView(org.apache.paimon.view.View view) {
+    Map<String, String> options =
+        view.options() == null ? new HashMap<>() : new HashMap<>(view.options());
+
+    String defaultCatalog = options.remove(DEFAULT_CATALOG_PROPERTY);
+    String defaultSchema = options.remove(DEFAULT_SCHEMA_PROPERTY);
+
+    return new PaimonView(
+        view.name(),
+        view.comment().orElse(null),
+        GravitinoPaimonColumn.fromPaimonRowType(view.rowType()).toArray(new Column[0]),
+        toRepresentations(view.query(), view.dialects()),
+        defaultCatalog,
+        defaultSchema,
+        options,
+        AuditInfo.EMPTY);
+  }
+
+  static org.apache.paimon.view.View toPaimonView(
+      NameIdentifier ident,
+      @Nullable String comment,
+      Column[] columns,
+      Representation[] representations,
+      @Nullable String defaultCatalog,
+      @Nullable String defaultSchema,
+      Map<String, String> properties) {
+    Preconditions.checkArgument(ident != null, "View identifier must not be null");
+
+    Representation[] safeRepresentations =
+        representations == null ? new Representation[0] : representations;
+    Preconditions.checkArgument(
+        safeRepresentations.length > 0, "representations must not be null or empty");
+
+    Column[] safeColumns = columns == null ? new Column[0] : columns;
+    List<DataField> fields = new ArrayList<>(safeColumns.length);
+    for (int index = 0; index < safeColumns.length; index++) {
+      fields.add(GravitinoPaimonColumn.toPaimonColumn(index, safeColumns[index]));
+    }
+
+    Map<String, String> dialectQueries = new HashMap<>();
+    String query = null;
+    for (Representation representation : safeRepresentations) {
+      Preconditions.checkArgument(
+          representation instanceof SQLRepresentation, "Paimon only supports SQL representations");
+      SQLRepresentation sqlRepresentation = (SQLRepresentation) representation;
+      if (query == null) {
+        query = sqlRepresentation.sql();
+      }
+
+      dialectQueries.put(sqlRepresentation.dialect(), sqlRepresentation.sql());
+    }
+
+    Preconditions.checkArgument(
+        query != null && !query.isEmpty(), "View SQL must not be null or empty");
+
+    Map<String, String> options = properties == null ? new HashMap<>() : new HashMap<>(properties);
+    if (defaultCatalog != null) {
+      options.put(DEFAULT_CATALOG_PROPERTY, defaultCatalog);
+    } else {
+      options.remove(DEFAULT_CATALOG_PROPERTY);
+    }
+
+    if (defaultSchema != null) {
+      options.put(DEFAULT_SCHEMA_PROPERTY, defaultSchema);
+    } else {
+      options.remove(DEFAULT_SCHEMA_PROPERTY);
+    }
+
+    String[] namespaceLevels = ident.namespace().levels();
+    Preconditions.checkArgument(namespaceLevels.length > 0, "View namespace must not be empty");
+    Identifier paimonIdentifier =
+        Identifier.create(namespaceLevels[namespaceLevels.length - 1], ident.name());
+
+    return new ViewImpl(paimonIdentifier, fields, query, dialectQueries, comment, options);
+  }
+
+  @Override
+  public String name() {
+    return name;
+  }
+
+  @Override
+  public String comment() {
+    return comment;
+  }
+
+  @Override
+  public Column[] columns() {
+    return Arrays.copyOf(columns, columns.length);
+  }
+
+  @Override
+  public Representation[] representations() {
+    return Arrays.copyOf(representations, representations.length);
+  }
+
+  @Override
+  public String defaultCatalog() {
+    return defaultCatalog;
+  }
+
+  @Override
+  public String defaultSchema() {
+    return defaultSchema;
+  }
+
+  @Override
+  public Map<String, String> properties() {
+    return Collections.unmodifiableMap(properties);
+  }
+
+  @Override
+  public AuditInfo auditInfo() {
+    return auditInfo;
+  }
+
+  private static Representation[] toRepresentations(String query, Map<String, String> dialects) {
+    Preconditions.checkArgument(
+        query != null && !query.isEmpty(), "Paimon view query must not be null or empty");
+
+    Map<String, String> safeDialects =
+        dialects == null ? Collections.emptyMap() : new HashMap<>(dialects);
+    List<Representation> sqlRepresentations = new ArrayList<>();
+    safeDialects.forEach(
+        (dialect, sql) ->
+            sqlRepresentations.add(
+                SQLRepresentation.builder().withDialect(dialect).withSql(sql).build()));
+
+    boolean queryCovered =
+        safeDialects.values().stream()
+            .anyMatch(dialectQuery -> Objects.equals(dialectQuery, query));
+    if (sqlRepresentations.isEmpty() || !queryCovered) {
+      sqlRepresentations.add(
+          SQLRepresentation.builder().withDialect(DEFAULT_DIALECT).withSql(query).build());
+    }
+
+    return sqlRepresentations.toArray(new Representation[0]);
+  }
+}
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
@@ -25,7 +25,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import javax.annotation.Nullable;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.meta.AuditInfo;
@@ -43,7 +42,7 @@ final class PaimonView implements View {
   static final String DEFAULT_CATALOG_PROPERTY = "gravitino.view.default-catalog";
   static final String DEFAULT_SCHEMA_PROPERTY = "gravitino.view.default-schema";
 
-  private static final String DEFAULT_DIALECT = "default";
+  private static final String PAIMON_VIEW_QUERY = "query";
 
   private final String name;
   @Nullable private final String comment;
@@ -94,17 +93,15 @@ final class PaimonView implements View {
   static org.apache.paimon.view.View toPaimonView(
       NameIdentifier ident,
       @Nullable String comment,
-      Column[] columns,
+      @Nullable Column[] columns,
       Representation[] representations,
       @Nullable String defaultCatalog,
       @Nullable String defaultSchema,
       Map<String, String> properties) {
     Preconditions.checkArgument(ident != null, "View identifier must not be null");
-
-    Representation[] safeRepresentations =
-        representations == null ? new Representation[0] : representations;
     Preconditions.checkArgument(
-        safeRepresentations.length > 0, "representations must not be null or empty");
+        representations != null && representations.length > 0,
+        "representations must not be null or empty");
 
     Column[] safeColumns = columns == null ? new Column[0] : columns;
     List<DataField> fields = new ArrayList<>(safeColumns.length);
@@ -114,19 +111,25 @@ final class PaimonView implements View {
 
     Map<String, String> dialectQueries = new HashMap<>();
     String query = null;
-    for (Representation representation : safeRepresentations) {
+    for (Representation representation : representations) {
       Preconditions.checkArgument(
           representation instanceof SQLRepresentation, "Paimon only supports SQL representations");
       SQLRepresentation sqlRepresentation = (SQLRepresentation) representation;
-      if (query == null) {
+      if (PAIMON_VIEW_QUERY.equalsIgnoreCase(sqlRepresentation.dialect())) {
+        Preconditions.checkArgument(
+            query == null,
+            "Only one representation with dialect '%s' is allowed",
+            PAIMON_VIEW_QUERY);
         query = sqlRepresentation.sql();
+      } else {
+        dialectQueries.put(sqlRepresentation.dialect(), sqlRepresentation.sql());
       }
-
-      dialectQueries.put(sqlRepresentation.dialect(), sqlRepresentation.sql());
     }
 
     Preconditions.checkArgument(
-        query != null && !query.isEmpty(), "View SQL must not be null or empty");
+        query != null && !query.isEmpty(),
+        "View representation with dialect '%s' must not be null or empty",
+        PAIMON_VIEW_QUERY);
 
     Map<String, String> options = properties == null ? new HashMap<>() : new HashMap<>(properties);
     if (defaultCatalog != null) {
@@ -195,19 +198,17 @@ final class PaimonView implements View {
 
     Map<String, String> safeDialects =
         dialects == null ? Collections.emptyMap() : new HashMap<>(dialects);
-    List<Representation> sqlRepresentations = new ArrayList<>();
-    safeDialects.forEach(
-        (dialect, sql) ->
-            sqlRepresentations.add(
-                SQLRepresentation.builder().withDialect(dialect).withSql(sql).build()));
+    List<Representation> sqlRepresentations = new ArrayList<>(safeDialects.size() + 1);
+    sqlRepresentations.add(
+        SQLRepresentation.builder().withDialect(PAIMON_VIEW_QUERY).withSql(query).build());
 
-    boolean queryCovered =
-        safeDialects.values().stream()
-            .anyMatch(dialectQuery -> Objects.equals(dialectQuery, query));
-    if (sqlRepresentations.isEmpty() || !queryCovered) {
-      sqlRepresentations.add(
-          SQLRepresentation.builder().withDialect(DEFAULT_DIALECT).withSql(query).build());
-    }
+    safeDialects.forEach(
+        (dialect, sql) -> {
+          if (!PAIMON_VIEW_QUERY.equalsIgnoreCase(dialect)) {
+            sqlRepresentations.add(
+                SQLRepresentation.builder().withDialect(dialect).withSql(sql).build());
+          }
+        });
 
     return sqlRepresentations.toArray(new Representation[0]);
   }
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import javax.annotation.Nullable;
 import org.apache.gravitino.NameIdentifier;
@@ -122,7 +123,12 @@ final class PaimonView implements View {
             PAIMON_VIEW_QUERY);
         query = sqlRepresentation.sql();
       } else {
-        dialectQueries.put(sqlRepresentation.dialect(), sqlRepresentation.sql());
+        String normalizedDialect = sqlRepresentation.dialect().toLowerCase(Locale.ROOT);
+        Preconditions.checkArgument(
+            !dialectQueries.containsKey(normalizedDialect),
+            "Only one representation per dialect is allowed (case-insensitive). Found duplicate: %s",
+            sqlRepresentation.dialect());
+        dialectQueries.put(normalizedDialect, sqlRepresentation.sql());
       }
     }
 
--- a/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
+++ b/catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
@@ -209,10 +209,6 @@ final class PaimonViewCatalogOps {
 
   private View replaceView(NameIdentifier identifier, ViewChange... changes)
       throws NoSuchViewException, IllegalArgumentException {
-    Preconditions.checkArgument(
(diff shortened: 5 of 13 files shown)


catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
```diff
@@ -61,7 +61,11 @@ public class PaimonTablePropertiesMetadata extends BasePropertiesMetadata {
             stringImmutablePropertyEntry(
                 ROWKIND_FIELD, "The table rowkind field", false, null, false, false),
             stringReservedPropertyEntry(PRIMARY_KEY, "The table primary key", false),
-            stringReservedPropertyEntry(PARTITION, "The table partition", false));
+            stringReservedPropertyEntry(PARTITION, "The table partition", false),
+            stringReservedPropertyEntry(
+                PaimonView.DEFAULT_CATALOG_PROPERTY, "The view default catalog", true),
+            stringReservedPropertyEntry(
+                PaimonView.DEFAULT_SCHEMA_PROPERTY, "The view default schema", true));
```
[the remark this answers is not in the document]
mchades: This reuse is intentional and will be split into separate components if necessary in the future.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,300 @@
...
+    NameIdentifier targetIdentifier = paimonIdentifierBuilder.apply(finalIdentifier);
+
+    try {
+      if (sourceIdentifier.equals(targetIdentifier)) {
+        paimonCatalogOps.dropView(sourceIdentifier.toString());
+        paimonCatalogOps.createView(sourceIdentifier.toString(), paimonView);
+      } else {
+        paimonCatalogOps.createView(targetIdentifier.toString(), paimonView);
+        paimonCatalogOps.dropView(sourceIdentifier.toString());
+      }
```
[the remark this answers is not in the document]
mchades: The current behavior follows the engine behavior of Paimon Spark and Flink.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
```diff
@@ -0,0 +1,214 @@
...
+            sqlRepresentations.add(
+                SQLRepresentation.builder().withDialect(dialect).withSql(sql).build()));
+
+    boolean queryCovered =
+        safeDialects.values().stream()
+            .anyMatch(dialectQuery -> Objects.equals(dialectQuery, query));
+    if (sqlRepresentations.isEmpty() || !queryCovered) {
+      sqlRepresentations.add(
+          SQLRepresentation.builder().withDialect(DEFAULT_DIALECT).withSql(query).build());
+    }
```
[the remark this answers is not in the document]
mchades: Addressed in the latest commit. I removed the synthetic placeholder dialect behavior and switched to a dedicated `query` representation contract, with filtering on write-back so no synthetic/default dialect is persisted into Paimon dialects.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
```diff
@@ -0,0 +1,214 @@
...
+    for (Representation representation : safeRepresentations) {
+      Preconditions.checkArgument(
+          representation instanceof SQLRepresentation, "Paimon only supports SQL representations");
+      SQLRepresentation sqlRepresentation = (SQLRepresentation) representation;
+      if (query == null) {
+        query = sqlRepresentation.sql();
+      }
+
+      dialectQueries.put(sqlRepresentation.dialect(), sqlRepresentation.sql());
+    }
```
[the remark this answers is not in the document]
mchades: Addressed. The canonical query is now selected deterministically from the dedicated `query` representation (case-insensitive match), rather than relying on input array order.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/TestPaimonViewCatalogOps.java
```diff
@@ -0,0 +1,328 @@
...
+      Identifier identifier, org.apache.paimon.view.View view) {
+    return new ViewImpl(
+        identifier,
+        view.rowType().getFields(),
+        view.query(),
+        new HashMap<>(view.dialects()),
+        view.comment().orElse(null),
+        new HashMap<>(view.options()));
+  }
+
```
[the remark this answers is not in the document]
mchades: Addressed. I extracted the duplicated in-memory view-support catalog logic into a shared helper (`PaimonViewTestCatalogHelper`) and updated the tests to reuse it.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,300 @@
...
+          Arrays.stream(changes)
+              .filter(viewChange -> !(viewChange instanceof ViewChange.RenameView))
+              .map(String::valueOf)
+              .collect(Collectors.joining("\n"));
+      Preconditions.checkArgument(
+          StringUtils.isEmpty(otherChanges),
+          String.format(
+              "The operation to change the view name cannot be performed together with other operations. "
+                  + "The list of operations that you cannot perform includes: \n%s",
+              otherChanges));
```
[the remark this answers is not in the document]
mchades: Addressed. I replaced eager `String.format(...)` construction in `Preconditions.checkArgument` with the varargs format-args overload so formatting only happens on failure.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/integration/test/CatalogPaimonBaseIT.java
```diff
@@ -1040,6 +1111,30 @@ void testTimeTypePrecision() throws org.apache.paimon.catalog.Catalog.TableNotEx
...
+              Column.of("name", Types.StringType.get(), "name column")
+            },
+            table_comment,
+            createProperties());
+
+    spark.sql(
+        String.format(
+            "INSERT INTO paimon.%s.%s VALUES (1, 'name_1'), (2, 'name_2')", schemaName, tableName));
+    return tableIdentifier;
+  }
```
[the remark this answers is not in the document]
mchades: Addressed. I added an explicit `Preconditions.checkNotNull(spark, ...)` guard in the helper to fail fast with a clear message if the helper is ever used outside the gated Hive/Spark path.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,300 @@
...
+      return renameView(identifier, (ViewChange.RenameView) renameViewOpt.get());
+    }
```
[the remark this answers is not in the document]
mchades: Addressed. Multiple `RenameView` operations are now explicitly rejected (`renameViewChanges.size() == 1`), and I added a dedicated test to cover this behavior.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/ops/TestPaimonCatalogOps.java
```diff
@@ -167,6 +174,54 @@ void testLoadListAndDropTableOperations() throws Exception {
     Assertions.assertNotNull(paimonCatalogOps.loadTable(IDENTIFIER.toString()));
   }
 
+  @Test
+  void testCreateLoadAlterRenameAndDropViewOperations() throws Exception {
+    paimonCatalogOps.catalog =
+        PaimonViewTestCatalogHelper.createViewSupportedCatalog(paimonCatalogOps.catalog);
```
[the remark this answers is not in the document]
mchades: We intentionally keep this test setup aligned with the table-side test style in this class and avoid introducing another test-only accessor subclass in ops tests. The assignment is test-scoped and follows the existing pattern used in TestPaimonCatalogOps. No production behavior is affected.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,307 @@
...
+
+    try {
+      if (sourceIdentifier.equals(targetIdentifier)) {
+        // Paimon does not provide a native replace-view API. To align with Paimon Spark's
+        // CREATE OR REPLACE VIEW behavior, replace on the same identifier is executed as
+        // drop-then-create. This path is intentionally non-atomic: if create fails after drop,
+        // the original view remains dropped and no rollback is attempted.
+        paimonCatalogOps.dropView(sourceIdentifier.toString());
+        paimonCatalogOps.createView(sourceIdentifier.toString(), paimonView);
+      } else {
```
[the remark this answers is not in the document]
mchades: Addressed in 2e77bed36. Added explicit non-atomic behavior comments and failure-path warning for cross-identifier replace: target is created first; if source drop fails, both views may temporarily coexist.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,307 @@
...
+            columns,
+            representations,
+            defaultCatalog,
+            defaultSchema,
+            properties);
+
+    try {
+      paimonCatalogOps.createView(nameIdentifier.toString(), paimonView);
+    } catch (Catalog.DatabaseNotExistException e) {
+      throw new NoSuchSchemaException(e, NO_SUCH_SCHEMA_EXCEPTION, identifier);
```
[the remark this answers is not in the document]
mchades: Addressed in 2e77bed36. The create-view DatabaseNotExist path now reports the schema identifier (instead of the full view identifier) in the NoSuchSchemaException message.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,307 @@
...
+          throw new IllegalArgumentException(
+              "Unsupported view change type: " + change.getClass().getSimpleName());
+        }
+      }
+      paimonCatalogOps.alterView(
+          paimonIdentifierBuilder.apply(identifier).toString(), paimonViewChanges);
+    } catch (Catalog.ViewNotExistException e) {
+      throw new NoSuchViewException(e, NO_SUCH_VIEW_EXCEPTION, identifier);
+    } catch (Catalog.DialectAlreadyExistException | Catalog.DialectNotExistException e) {
+      throw new IllegalArgumentException(e);
```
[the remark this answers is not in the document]
mchades: Addressed in 2e77bed36. I narrowed the alter-view exception surface: removed unreachable dialect catches in PaimonViewCatalogOps, and PaimonCatalogOps.alterView now throws ViewNotExistException only while wrapping dialect exceptions as IllegalArgumentException.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonView.java
```diff
@@ -0,0 +1,215 @@
...
+      SQLRepresentation sqlRepresentation = (SQLRepresentation) representation;
+      if (PAIMON_VIEW_QUERY.equalsIgnoreCase(sqlRepresentation.dialect())) {
+        Preconditions.checkArgument(
+            query == null,
+            "Only one representation with dialect '%s' is allowed",
+            PAIMON_VIEW_QUERY);
+        query = sqlRepresentation.sql();
+      } else {
+        dialectQueries.put(sqlRepresentation.dialect(), sqlRepresentation.sql());
+      }
```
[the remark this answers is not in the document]
mchades: Addressed in 2e77bed36. Non-query dialect representations are normalized to lowercase and enforced unique case-insensitively before writing to the dialect map, avoiding inconsistent dialect keys.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/integration/test/CatalogPaimonBaseIT.java
```diff
@@ -1040,6 +1145,59 @@ void testTimeTypePrecision() throws org.apache.paimon.catalog.Catalog.TableNotEx
...
+      }
+
+      current = current.getCause();
+    }
+
+    Assertions.fail(
+        String.format(
+            "Expected unsupported view interoperability for backend %s, but got: %s",
+            TYPE, throwable));
+  }
```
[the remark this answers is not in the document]
mchades: Addressed in 2e77bed36. I removed the loose substring-based unsupported assertion path and switched these Spark/Gravitino view interop checks to explicit UnsupportedOperationException assertions.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/integration/test/CatalogPaimonBaseIT.java
```diff
@@ -858,6 +862,107 @@ void testOperationDataOfPaimonTable() {
...
+              () -> {
+                spark.sql(
+                    String.format(
+                        "CREATE VIEW paimon.%s AS SELECT id, name FROM paimon.%s",
+                        viewIdentifier, tableIdentifier));
+                viewCatalog.loadView(NameIdentifier.of(schemaName, viewName));
+              });
+      assertViewInteropUnsupported(throwable);
+      return;
+    }
```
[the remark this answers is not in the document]
mchades: Addressed in 2e77bed36. I removed the TYPE-based gating helper and made the related Spark view interop tests assert unsupported behavior directly, so non-Hive subclasses no longer bypass these checks.

jerryshao (commented):
Thanks for the well-structured PR — the composition approach with `PaimonViewCatalogOps`, the defensive copy discipline in `PaimonView`, and the two-direction Spark interop tests are all good. Left inline comments below on a few issues, one of which is a logic bug (dead code in `replaceView`).

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,318 @@
...
+            identifier,
+            replaceView.getComment(),
+            replaceView.getColumns(),
+            replaceView.getRepresentations(),
+            replaceView.getDefaultCatalog(),
+            replaceView.getDefaultSchema(),
+            finalProperties);
+
+    NameIdentifier sourceIdentifier = paimonIdentifierBuilder.apply(identifier);
+    NameIdentifier targetIdentifier = paimonIdentifierBuilder.apply(identifier);
```
jerryshao: **Bug: `else` branch is dead code.** `sourceIdentifier` and `targetIdentifier` are both computed by calling `paimonIdentifierBuilder.apply(identifier)` with the same `identifier`, so they are always equal and the `else` branch can never be reached. The comment describing a "cross-identifier replace" is therefore misleading. The `targetIdentifier` variable is also redundant — just use `sourceIdentifier` directly. Remove the `else` branch and simplify to a straightforward drop-then-create.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
```diff
@@ -61,7 +61,11 @@ public class PaimonTablePropertiesMetadata extends BasePropertiesMetadata {
             stringImmutablePropertyEntry(
                 ROWKIND_FIELD, "The table rowkind field", false, null, false, false),
             stringReservedPropertyEntry(PRIMARY_KEY, "The table primary key", false),
-            stringReservedPropertyEntry(PARTITION, "The table partition", false));
+            stringReservedPropertyEntry(PARTITION, "The table partition", false),
+            stringReservedPropertyEntry(
```
jerryshao, line 65: `DEFAULT_CATALOG_PROPERTY` and `DEFAULT_SCHEMA_PROPERTY` are view-only concepts but are being registered inside `PaimonTablePropertiesMetadata`. This mixes table and view concerns and could cause these reserved properties to surface in table-level validation paths. Consider a dedicated `PaimonViewPropertiesMetadata`, or at minimum add a comment explaining why they are registered here.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java
```diff
@@ -77,11 +83,13 @@
...
  * Apache Paimon catalog in Apache Gravitino.
  */
-public class PaimonCatalogOperations implements CatalogOperations, SupportsSchemas, TableCatalog {
+public class PaimonCatalogOperations
+    implements CatalogOperations, SupportsSchemas, TableCatalog, ViewCatalog {
 
   public static final Logger LOG = LoggerFactory.getLogger(PaimonCatalogOperations.class);
 
   @VisibleForTesting public PaimonCatalogOps paimonCatalogOps;
+  @VisibleForTesting PaimonViewCatalogOps paimonViewCatalogOps;
```
jerryshao, line 92: `paimonCatalogOps` just above is `@VisibleForTesting public`, but `paimonViewCatalogOps` has neither the annotation nor `public`. Since it is accessed from `TestPaimonViewCatalogOps`, it should be `@VisibleForTesting` for consistency. As-is the field is package-private by accident rather than by intent.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/ops/PaimonCatalogOps.java
```diff
@@ -104,7 +110,41 @@ public void renameTable(String fromTableName, String toTableName)
...
+  public void createView(String viewName, View view)
+      throws ViewAlreadyExistException, DatabaseNotExistException {
+    catalog.createView(viewIdentifier(viewName), view, false);
+  }
+
+  public void alterView(String viewName, List<ViewChange> changes) throws ViewNotExistException {
+    try {
+      catalog.alterView(viewIdentifier(viewName), changes, false);
+    } catch (DialectAlreadyExistException | DialectNotExistException e) {
+      throw new IllegalArgumentException(e);
```
jerryshao: The original `DialectAlreadyExistException` / `DialectNotExistException` message is lost when wrapping. Consider including context:
```java
    "Cannot alter view " + viewName + ": " + e.getMessage(), e);
```

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,318 @@
...
+          ViewChange.SetProperty setProperty = (ViewChange.SetProperty) change;
+          paimonViewChanges.add(
+              org.apache.paimon.view.ViewChange.setOption(
+                  setProperty.getProperty(), setProperty.getValue()));
+        } else if (change instanceof ViewChange.RemoveProperty) {
+          ViewChange.RemoveProperty removeProperty = (ViewChange.RemoveProperty) change;
+          paimonViewChanges.add(
+              org.apache.paimon.view.ViewChange.removeOption(removeProperty.getProperty()));
+        } else {
+          throw new IllegalArgumentException(
```
jerryshao, line 294: `alterViewWithoutReplace` silently rejects `ViewChange.UpdateComment` and any dialect changes with a generic `IllegalArgumentException`. Callers can't discover this limitation without reading the source. At minimum, document on `alterView` which `ViewChange` subtypes are unsupported (or throw a more descriptive exception message that names the supported types).

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,318 @@
...
+      throw new NoSuchViewException(e, NO_SUCH_VIEW_EXCEPTION, identifier);
+    } catch (Catalog.ViewAlreadyExistException e) {
+      throw new IllegalArgumentException(
+          String.format(VIEW_ALREADY_EXISTS_EXCEPTION, renamedIdentifier), e);
+    }
+  }
+
+  private View replaceView(NameIdentifier identifier, ViewChange... changes)
+      throws NoSuchViewException, IllegalArgumentException {
+    Preconditions.checkArgument(
```
jerryshao: This `noneMatch(RenameView)` precondition is redundant. `replaceView` is only ever called from `alterView` after it has already handled and returned early on any `RenameView` changes. The guard can be removed.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/utils/PaimonViewTestCatalogHelper.java
```diff
@@ -0,0 +1,182 @@
...
+            if (!updatedDialects.containsKey(updateDialect.dialect())) {
+              throw new Catalog.DialectNotExistException(identifier, updateDialect.dialect());
+            }
+            updatedDialects.put(updateDialect.dialect(), updateDialect.query());
+          } else if (change instanceof ViewChange.DropDialect) {
+            ViewChange.DropDialect dropDialect = (ViewChange.DropDialect) change;
+            if (!updatedDialects.containsKey(dropDialect.dialect())) {
+              throw new Catalog.DialectNotExistException(identifier, dropDialect.dialect());
+            }
+            updatedDialects.remove(dropDialect.dialect());
```
jerryshao, line 124: When a dialect entry is updated (e.g., `updateDialect("spark", ...)`), the main `query` field is overwritten with `updatedDialects.values().iterator().next()` — an arbitrary first dialect. In real Paimon the main `query` and per-dialect entries are independent fields. This mismatch means the in-memory test catalog behaves differently from real Paimon, which could cause unit tests to pass even when the actual implementation has a bug.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for calling this out. In the current Gravitino ViewChange API, callers can only construct rename, setProperty, removeProperty, and replaceView. Comment/dialect updates are intentionally represented through replaceView, so they are not expected in alterViewWithoutReplace. For unknown custom ViewChange implementations, we still fail fast with IllegalArgumentException. Given this API contract, we would like to keep the current behavior in this PR.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
mchades: Good point. We considered splitting this into a dedicated view metadata class, but the current connector API only exposes a single properties metadata hook (tablePropertiesMetadata) and does not provide a separate view-properties metadata entry. We register these two reserved keys here so view default catalog/schema options can still be validated through the same reserved-property path. Given the scope of this PR, we would like to keep the current structure and avoid a broader metadata refactor here.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalogOperations.java
mchades: Thanks for the review. This field is already annotated with @VisibleForTesting in the current code. We intentionally keep paimonViewCatalogOps package-private (instead of public) to minimize visibility while still allowing tests in the same package to access it. So the package-private visibility here is by intent.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,318 @@
...
+        } catch (RuntimeException e) {
+          LOG.warn(
+              "Cross-identifier replace of Paimon view is non-atomic: target view {} was created, "
+                  + "but dropping source view {} failed. Both views may exist until cleanup.",
+              targetIdentifier,
+              sourceIdentifier,
+              e);
+          throw e;
+        }
+      }
```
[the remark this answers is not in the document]
mchades: Addressed in commit 2196492bb: removed the dead branch in replaceView and kept a single drop-then-create flow for the same identifier.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,318 @@
...
+    org.apache.paimon.view.View paimonView =
+        PaimonView.toPaimonView(
+            identifier,
+            replaceView.getComment(),
+            replaceView.getColumns(),
+            replaceView.getRepresentations(),
+            replaceView.getDefaultCatalog(),
+            replaceView.getDefaultSchema(),
+            finalProperties);
+
```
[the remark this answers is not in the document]
mchades: Addressed in commit 2196492bb: added focused ReplaceView unit coverage in TestPaimonViewCatalogOps, including replacement of columns, representations, and comment while preserving properties.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,318 @@
...
+      views = paimonCatalogOps.listViews(schemaIdentifier.name());
+    } catch (Catalog.DatabaseNotExistException e) {
+      throw new NoSuchSchemaException(e, NO_SUCH_SCHEMA_EXCEPTION, namespace.toString());
+    }
+
+    return views.stream()
+        .map(
+            viewIdentifier -> NameIdentifier.of(ArrayUtils.add(namespace.levels(), viewIdentifier)))
+        .toArray(NameIdentifier[]::new);
+  }
```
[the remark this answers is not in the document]
mchades: Addressed in commit 2196492bb: added listViews/dropView unit coverage, including the false return when dropping a non-existent view.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Addressed in commit 2196492bb: removed the dead branch in replaceView and kept a single drop-then-create flow for the same identifier.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/ops/PaimonCatalogOps.java
mchades: Addressed in commit 2196492bb: alterView now includes view-name context and original cause message when wrapping DialectAlreadyExistException/DialectNotExistException, with regression test coverage.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Addressed in commit 2196492bb: removed the redundant noneMatch(RenameView) precondition in replaceView and added a regression test to keep behavior intact.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/utils/PaimonViewTestCatalogHelper.java
mchades: Addressed in commit 2196492bb: aligned the in-memory Paimon test helper so updateDialect does not overwrite the primary query, and added regression assertion.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,295 @@
...
+  View loadView(NameIdentifier identifier) throws NoSuchViewException {
+    org.apache.paimon.view.View view;
+    try {
+      NameIdentifier viewIdentifier = paimonIdentifierBuilder.apply(identifier);
+      view = paimonCatalogOps.loadView(viewIdentifier.toString());
+    } catch (Catalog.ViewNotExistException e) {
+      throw new NoSuchViewException(e, NO_SUCH_VIEW_EXCEPTION, identifier);
+    }
+
+    LOG.info("Loaded Paimon view {}.", identifier);
```
jerryshao: These kinds of log seems not so useful. We already have such log in the REST endpoint, print here again without more useful information is not necessary.

jerryshao (commented):
Well-structured PR — the separation of view logic into `PaimonViewCatalogOps` is a good design choice and unit test coverage for the ops layer is solid. A few correctness issues to address before merge, details inline.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalog.java
```diff
@@ -65,6 +66,11 @@ protected CatalogOperations newOps(Map<String, String> config) {
     return new PaimonCatalogOperations();
   }
 
+  @Override
+  public ViewCatalog asViewCatalog() {
+    return (ViewCatalog) ops();
```
jerryshao, line 71: `ops()` is typed as `CatalogOperations`, so this cast is unchecked. If `newOps()` is ever overridden in a subclass or the class hierarchy changes, this throws `ClassCastException` at runtime with no diagnostic. At minimum add a state guard:
```java
CatalogOperations ops = ops();
Preconditions.checkState(ops instanceof ViewCatalog,
    "PaimonCatalogOperations must implement ViewCatalog");
return (ViewCatalog) ops;
```

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
```diff
@@ -61,7 +61,11 @@ public class PaimonTablePropertiesMetadata extends BasePropertiesMetadata {
...
+                PaimonView.DEFAULT_CATALOG_PROPERTY, "The view default catalog", true),
```
jerryshao, line 66: `DEFAULT_CATALOG_PROPERTY` and `DEFAULT_SCHEMA_PROPERTY` are view-internal properties but they're being registered in the **table** properties metadata. This means they appear as reserved property keys when validating table operations — a `CREATE TABLE` call that happens to include one of these keys would be silently rejected, even though they're view-only. These should live in a dedicated `PaimonViewPropertiesMetadata`, or at minimum be removed from this class.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,295 @@
...
+    Preconditions.checkArgument(
+        schemaExistsChecker != null, "Schema exists checker must not be null");
+
+    this.paimonCatalogOps = paimonCatalogOps;
+    this.paimonIdentifierBuilder = paimonIdentifierBuilder;
+    this.schemaExistsChecker = schemaExistsChecker;
+  }
+
+  NameIdentifier[] listViews(Namespace namespace) throws NoSuchSchemaException {
+    String[] levels = namespace.levels();
```
jerryshao, line 77: `levels[levels.length - 1]` throws `ArrayIndexOutOfBoundsException` if `namespace.levels()` is empty. The `TableCatalog` implementations in this repo consistently guard this. Add:
```java
Preconditions.checkArgument(levels.length > 0, "Namespace must not be empty");
```

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,295 @@
...
+    try {
+      paimonCatalogOps.createView(nameIdentifier.toString(), paimonView);
+    } catch (Catalog.DatabaseNotExistException e) {
+      throw new NoSuchSchemaException(e, NO_SUCH_SCHEMA_EXCEPTION, schemaIdentifier);
+    } catch (Catalog.ViewAlreadyExistException e) {
+      throw new ViewAlreadyExistsException(e, VIEW_ALREADY_EXISTS_EXCEPTION, identifier);
+    }
+
+    LOG.info("Created Paimon view {}.", identifier);
+    return PaimonView.fromPaimonView(paimonView);
```
jerryshao, line 137: This returns `fromPaimonView(paimonView)` where `paimonView` is the **locally-constructed** Paimon view object, not the one fetched back from the catalog. If the underlying Paimon catalog normalizes properties, adds audit metadata, or transforms anything on write, the returned view will differ from what was actually persisted. For consistency with `alterView` (which calls `loadView` after each modification), `createView` should also call `loadView(identifier)` after the catalog write.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,295 @@
...
+
+  private View renameView(NameIdentifier identifier, ViewChange.RenameView renameView)
+      throws NoSuchViewException, IllegalArgumentException {
+    NameIdentifier sourceIdentifier = paimonIdentifierBuilder.apply(identifier);
+    NameIdentifier renamedIdentifier =
+        NameIdentifier.of(identifier.namespace(), renameView.getNewName());
+    NameIdentifier targetIdentifier = paimonIdentifierBuilder.apply(renamedIdentifier);
+    try {
+      paimonCatalogOps.renameView(sourceIdentifier.toString(), targetIdentifier.toString());
+      return loadView(renamedIdentifier);
```
jerryshao: If `renameView` succeeds but the subsequent `loadView(renamedIdentifier)` fails (e.g. due to a transient I/O error), a `NoSuchViewException` is thrown even though the rename completed. The caller has no way to distinguish "rename failed" from "rename succeeded but load failed", and retrying would attempt to rename a view that no longer exists at the source. Consider wrapping the `loadView` call separately and mapping its failure to a more diagnostic exception.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,295 @@
...
+    } catch (Catalog.ViewNotExistException e) {
+      throw new NoSuchViewException(e, NO_SUCH_VIEW_EXCEPTION, identifier);
+    } catch (Catalog.ViewAlreadyExistException e) {
```
jerryshao, line 198: When the rename target already exists, `Catalog.ViewAlreadyExistException` is mapped to `IllegalArgumentException`. Gravitino already has `ViewAlreadyExistsException` (used correctly in `createView`). Using it here too would give callers a typed exception to handle consistently with the rest of the `ViewCatalog` API, rather than parsing the plain `IllegalArgumentException` message.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,295 @@
...
+            finalProperties);
+
+    NameIdentifier sourceIdentifier = paimonIdentifierBuilder.apply(identifier);
+
+    try {
+      // Paimon does not provide a native replace-view API. To align with Paimon Spark's
+      // CREATE OR REPLACE VIEW behavior, replace is executed as drop-then-create.
+      // This path is intentionally non-atomic: if create fails after drop,
+      // the original view remains dropped and no rollback is attempted.
+      paimonCatalogOps.dropView(sourceIdentifier.toString());
```
jerryshao, line 250: The comment correctly identifies this as non-atomic, but the consequence is severe: **if `createView` fails after `dropView`, the original view is permanently lost with no rollback**. This should be surfaced more prominently — e.g. a `// WARNING:` prefix, a `LOG.warn` at the start of this path, and an explicit callout in the PR description as a known limitation. The current comment tone reads like a casual implementation note rather than a data-loss risk.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/integration/test/CatalogPaimonBaseIT.java
```diff
@@ -858,6 +861,55 @@ void testOperationDataOfPaimonTable() {
...
   }
 
+  @Test
+  void testSparkCreateViewAndLoadByGravitino() {
+    String viewName = GravitinoITUtils.genRandomName("spark_create_view");
+    String viewIdentifier = String.join(".", schemaName, viewName);
+
+    ViewCatalog viewCatalog = catalog.asViewCatalog();
+    Assertions.assertThrows(
+        UnsupportedOperationException.class,
```
jerryshao, line 871: Both `testSparkCreateViewAndLoadByGravitino` (here) and `testGravitinoCreateViewAndReadBySpark` (line 894) wrap multiple statements inside a single `assertThrows`. Because the first statement throws `UnsupportedOperationException` immediately in non-Hive catalog contexts, the second statement is **never executed** — the assertion passes trivially without testing the second operation.

- Here: if `spark.sql(...)` throws, `viewCatalog.loadView(...)` is never called, so we don't actually verify that Gravitino loads a Spark-created view.
- Line 894: if `createView(...)` throws, the Spark `SELECT` is never attempted.

The base-class tests should either be annotated `@Disabled` with an explanatory comment, use `Assumptions.assumeTrue(supportsViewCatalog())` to skip gracefully, or be removed entirely and left only in the concrete `CatalogPaimonHiveIT` subclass.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonTablePropertiesMetadata.java
mchades: Thanks for raising this again. This was discussed in an earlier thread: https://github.com/apache/gravitino/pull/11013#discussion_r3264688555. As noted there, the current connector API only exposes a single properties metadata hook (`tablePropertiesMetadata`) and does not provide a dedicated view-properties metadata entry yet. We intentionally keep `DEFAULT_CATALOG_PROPERTY` and `DEFAULT_SCHEMA_PROPERTY` in the shared metadata so view default catalog/schema options can still go through the same reserved-property validation path. Given this PR's scope, we prefer to keep this structure and avoid a broader metadata API refactor here. We can follow up with a dedicated split once the connector API supports view properties metadata separately.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for the suggestion. In the current connector flow, listViews is invoked with a schema namespace validated by upper layers, so an empty namespace is not expected to reach this path. We also keep the behavior aligned with listTables in the same catalog operations implementation, which currently follows the same pattern. Given that contract and consistency, we prefer not to add an extra guard in this PR.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonCatalog.java
mchades: Thanks for the suggestion. We prefer to keep the current implementation as-is for this PR. The direct cast pattern is already used in our catalog implementations and is consistent with the current project style. Given the current class contract and scope here, we don't plan to add an extra state guard in this change.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for the suggestion. For consistency with the existing table create path in this connector, we currently keep create operations as write-then-return-local-metadata (without an immediate read-after-write). The current createView behavior follows the same pattern. We also have not observed additional persisted fields in this flow that require a post-create load to be reflected in the returned metadata. Given this PR scope, we prefer to keep the current behavior.

catalogs/catalog-lakehouse-paimon/src/test/java/org/apache/gravitino/catalog/lakehouse/paimon/integration/test/CatalogPaimonBaseIT.java
mchades: Thanks for the review. In the base IT, these two cases intentionally validate the unsupported-view behavior for non-Hive backends. Once the first view operation fails with UnsupportedOperationException, we do not continue to the second operation by design. The actual cross-engine view interoperability behavior is covered in the Hive-specific overrides (CatalogPaimonHiveIT), where view support is enabled and both steps are exercised end-to-end.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for the comment. Updated in this PR: I removed the redundant log in the loadView path, since the REST layer already records the request-level context.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for the suggestion. Addressed in this PR: renameView now separates rename from post-rename load. If rename succeeds but loading the renamed view fails, we now throw a diagnostic IllegalStateException that clearly indicates "rename succeeded but load failed", instead of returning a misleading NoSuchViewException. I also added a dedicated unit test for this path.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for the suggestion. Addressed in this PR: when rename target already exists, we now map that case to ViewAlreadyExistsException for typed handling consistency with createView. I also added a unit test that validates this behavior.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: Thanks for the comment. Updated in this PR to make the risk explicit: I changed the inline note to a WARNING-style comment and added a LOG.warn before the non-atomic drop-and-create replace path, so the potential data-loss risk is clearly surfaced.

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
jerryshao: You should remove the log in all the places, not just here.

jerryshao:
I have no further comments, @diqiu50 can you please help to review?

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
mchades: fixed

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,310 @@
...
+    } catch (Catalog.ViewNotExistException e) {
+      throw new NoSuchViewException(e, NO_SUCH_VIEW_EXCEPTION, identifier);
+    }
+
+    try {
+      paimonCatalogOps.createView(sourceIdentifier.toString(), paimonView);
+    } catch (Catalog.ViewAlreadyExistException e) {
+      LOG.warn(
+          "Replacing Paimon view {} is non-atomic (drop-then-create). "
```
diqiu50, line 259: The log level should be error

catalogs/catalog-lakehouse-paimon/src/main/java/org/apache/gravitino/catalog/lakehouse/paimon/PaimonViewCatalogOps.java
```diff
@@ -0,0 +1,310 @@
...
+              + "Create failed after drop and the original view may be lost.",
+          identifier,
+          e);
+      throw new IllegalArgumentException(
+          String.format(VIEW_ALREADY_EXISTS_EXCEPTION, identifier), e);
+    } catch (Catalog.DatabaseNotExistException e) {
+      LOG.warn(
+          "Replacing Paimon view {} is non-atomic (drop-then-create). "
```
diqiu50, line 267: The same problem

mchades: fixed

diqiu50 (approved):
LGTM

PR #10913, opened 2026-04-30
[#10915] feat(cache): Add auth cache mappers

yuqi1129:
### What changes were proposed in this pull request?

Add auth-related mapper methods, provider SQL, auth PO classes, write-path updated_at touches, and focused mapper tests.

This PR is stacked on #10912.

### Why are the changes needed?

This is the auth mapper foundation for cache improvement. It depends on the 1.3.0 schema PR and is independent from the entity change log PR.

Fix: #10915 

### Does this PR introduce _any_ user-facing change?

No user-facing API change. This adds internal relational mapper support for auth cache consistency.

### How was this patch tested?

./gradlew :core:test --tests org.apache.gravitino.storage.relational.mapper.provider.base.TestAuthMappers

PR #11036, opened 2026-05-11
[#11037] feat(core): nested namespace persistence and owner batch in relational service

### LINKED ISSUE #11037: [FEATURE] Add relational nested-namespace naming bridge and schema meta support
### Describe the feature

Add relational nested-namespace naming bridge and schema meta support

### Motivation

Add relational nested-namespace naming bridge and schema meta support

### Describe the solution

Introduces HierarchicalSchemaUtil and RelationalSchemaNamingBridge for ASCII-1 hierarchical schema segments, extends JDBCBackend and SchemaMetaService, and updates schema SQL providers/mappers accordingly.

### Additional context

_No response_

roryqi:

    - Hierarchical schema config/util; schema meta service hierarchical naming
    - Owner batch soft-delete/insert SQL (OwnerRelDeletion) and OwnerMetaService.batchSetOwners
    - Schema/owner mappers and SQL providers; StatisticPO adjustments
    - PO converter tests

Fix: #11037

No. 

Added UTs.


roryqi:
I have addressed the copilot, claude and cursor comments.  I use Claude to review the code. The code is generally ok.

core/src/main/java/org/apache/gravitino/storage/relational/po/OwnerRelDeletion.java
```diff
@@ -16,7 +16,15 @@
...
 
-package org.apache.gravitino.flink.connector.integration.test.hive;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
 
-public class FlinkHiveCatalogIT118 extends FlinkHiveCatalogIT {}
+/** Key for batch soft-deleting owner rows by metadata object. */
+@Getter
+@AllArgsConstructor
+public class OwnerRelDeletion {
```
jerryshao: `OwnerRelForDeletion`? Also, where the `Rel` takes affect, what I can see in this class is just the metadata object id and type.

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
```diff
@@ -172,4 +173,49 @@ public void setOwner(
...
+
+    SessionUtils.doMultipleWithCommit(
+        () ->
+            SessionUtils.doWithoutCommit(
+                OwnerMetaMapper.class,
+                mapper -> mapper.batchSoftDeleteOwnerRelByMetadataObjects(deletions)),
+        () ->
+            SessionUtils.doWithoutCommit(
+                OwnerMetaMapper.class, mapper -> mapper.batchInsertOwnerRels(ownerRelPOs)));
+  }
```
jerryshao, line 220: Why do we delete then insert?

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -162,18 +168,54 @@ public List<SchemaEntity> listSchemasByNamespace(Namespace namespace) {
...
+                  .withComment(null)
+                  .withProperties(Collections.emptyMap())
+                  .withAuditInfo(schemaEntity.auditInfo())
+                  .build();
+          rowsToInsert.add(ancestor);
+        }
+        rowsToInsert.add(schemaEntity);
+      }
 
       SessionUtils.doWithCommit(
```
jerryshao, line 199: Do we need to use transaction for these db operations?

core/src/main/java/org/apache/gravitino/utils/HierarchicalSchemaUtil.java
```diff
@@ -0,0 +1,166 @@
...
+      int lastIdx = current.lastIndexOf(separator);
+      current = current.substring(0, lastIdx);
+      if (current.isEmpty()) {
+        break;
+      }
+      result.add(current);
+    }
+    return result;
+  }
+}
```
jerryshao, line 166: You should have tests for this class.

core/src/main/java/org/apache/gravitino/storage/relational/service/OwnerMetaService.java
roryqi: We need remove the old owner of the metadata object then set the new owner to the metadata object.
For removing the old owner, we delete a old data. 
For setting the new owner, we insert a new data.

core/src/main/java/org/apache/gravitino/storage/relational/po/OwnerRelDeletion.java
roryqi: We will use this in the batch deletion operations. We pass the parameter a list of ` OwnerRelDeletion` objects.

OK to rename.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
roryqi: I think we should use transaction for these db operations.
I wrapped the operations into a insert statement and wrap it by the method `doWithCommit`.

core/src/main/java/org/apache/gravitino/utils/HierarchicalSchemaUtil.java
roryqi: Added.

roryqi:
Have addressed the last round comments.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -162,18 +168,54 @@ public List<SchemaEntity> listSchemasByNamespace(Namespace namespace) {
...
+      } else {
+        String[] parts = schemaName.split(Pattern.quote(physicalSep), -1);
+        for (int nSeg = 1; nSeg < parts.length; nSeg++) {
+          String ancestorPhysical = String.join(physicalSep, Arrays.copyOf(parts, nSeg));
+          SchemaEntity ancestor =
+              SchemaEntity.builder()
+                  .withId(nextIdForNestedAncestor())
+                  .withName(ancestorPhysical)
+                  .withNamespace(schemaEntity.namespace())
+                  .withComment(null)
```
yuqi1129, line 190: It seems that this will overwrite any existing parents with non empty properties and comment, is it expected?

roryqi: I corrected this place. AI changed the logic. I don't notice this point. I added the check parent schema existence logic back.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
roryqi: I corrected this place. AI changed the logic. I don't notice this point. I merge all the insert operations into one operation.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -162,18 +169,69 @@ public List<SchemaEntity> listSchemasByNamespace(Namespace namespace) {
...
-      fillSchemaPOBuilderParentEntityId(builder, schemaEntity.namespace());
+      // Callers above this service (e.g. JDBCBackend + naming bridge) pass storage-form schema
+      // names: nested paths use the internal physical separator, not the external logical one.
+      String physicalSep = HierarchicalSchemaUtil.physicalSeparator();
+      String schemaName = schemaEntity.name();
+      List<SchemaEntity> rowsToInsert = new ArrayList<>();
+      if (schemaName == null || !schemaName.contains(physicalSep)) {
+        rowsToInsert.add(schemaEntity);
+      } else {
+        String[] parts = schemaName.split(Pattern.quote(physicalSep), -1);
```
yuqi1129, line 182: If possible, can you give a short example to demonstrate the logic here and L203. I have no more comments except this minor one.

roryqi: Added the comment.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/OwnerMetaBaseSQLProvider.java
```diff
@@ -124,6 +125,35 @@ public String insertOwnerRel(@Param("ownerRelPO") OwnerRelPO ownerRelPO) {
...
+        + "</foreach>"
+        + "</script>";
+  }
+
+  public String batchSoftDeleteOwnerRelByMetadataObjects(
+      @Param("deletions") List<OwnerRelForDeletion> deletions) {
+    return "<script>"
+        + "UPDATE "
+        + OWNER_TABLE_NAME
+        + " SET deleted_at = (UNIX_TIMESTAMP() * 1000.0)"
```
jerryshao, line 147: **MySQL soft-delete timestamp produces a `DOUBLE`, not `BIGINT`**

`UNIX_TIMESTAMP()` (no argument) returns a plain integer — no sub-second precision. Adding `EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000` makes the expression a `DOUBLE`, which is then stored into the `BIGINT` `deleted_at` column via an implicit cast. MySQL silently truncates, but this is fragile and inconsistent with the PostgreSQL override which uses `CAST(... AS BIGINT)` explicitly.

Consider wrapping with `FLOOR()` to make the truncation explicit:
```sql
SET deleted_at = FLOOR((UNIX_TIMESTAMP() * 1000.0)
    + EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000)
```

roryqi: This truncates are acceptable.

jerryshao, line 147: **MySQL soft-delete timestamp produces a `DOUBLE`, not `BIGINT`**

`UNIX_TIMESTAMP()` (no argument) returns a plain integer — no sub-second precision. Adding `EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000` makes the expression a `DOUBLE`, which is stored into the `BIGINT` `deleted_at` column via an implicit cast. MySQL silently truncates, but this is fragile and inconsistent with the PostgreSQL override which uses `CAST(... AS BIGINT)` explicitly.

Consider wrapping with `FLOOR()` to make the truncation explicit:
```sql
SET deleted_at = FLOOR((UNIX_TIMESTAMP() * 1000.0)
    + EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000)
```

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -162,18 +169,69 @@ public List<SchemaEntity> listSchemasByNamespace(Namespace namespace) {
...
+            fillSchemaPOBuilderParentEntityId(leafBuilder, leafRow.namespace());
+            SchemaPO leafPO = POConverters.initializeSchemaPOWithVersion(leafRow, leafBuilder);
+            List<SchemaPO> schemaPosToInsert = new ArrayList<>(missingAncestorPOs);
+            schemaPosToInsert.add(leafPO);
             if (overwrite) {
-              mapper.insertSchemaMetaOnDuplicateKeyUpdate(po);
+              mapper.batchInsertSchemaMetaOnDuplicateKeyUpdate(schemaPosToInsert);
             } else {
-              mapper.insertSchemaMeta(po);
+              mapper.batchInsertSchemaMeta(schemaPosToInsert);
```
jerryshao, line 236: **Ancestor insert race condition when `overwrite=false`**

Two concurrent inserts of `A:B:C` and `A:B:D` will both call `batchSelectSchemaByIdentifier`, both find ancestor `A` missing, and both attempt `batchInsertSchemaMeta`. The second caller gets a duplicate-key exception. The `overwrite=true` path handles this correctly via upsert, but the `false` path does not.

Catalog bootstrap scenarios (where many hierarchical schemas are imported in parallel) are particularly vulnerable. At minimum, document the known limitation. Ideally, catch the duplicate-key exception on ancestor rows specifically and treat it as a benign race.

core/src/main/java/org/apache/gravitino/utils/HierarchicalSchemaUtil.java
```diff
@@ -0,0 +1,166 @@
...
+   * <p>Example: {@code "A:B:C"} with separator {@code ":"} → {@code ["A:B:C", "A:B", "A"]}
+   *
+   * <p>This is used for privilege inheritance: a privilege on an ancestor schema is inherited by
+   * all descendant schemas.
+   *
+   * @param schemaName the schema name to compute scopes for
+   * @param separator the external separator
+   * @return the schema name and all ancestor names, from most specific to outermost
+   */
+  public static List<String> allScopes(String schemaName, String separator) {
```
jerryshao, line 150: **`allScopes()` is public but not called anywhere in this PR**

If this is pre-built for the upcoming privilege-inheritance layer, add a comment explaining the intent (e.g. `// used by privilege evaluation for ancestor scope resolution`). Otherwise, defer it until it is actually needed to keep the public API surface minimal.

jerryshao (commented):
## Review Summary (post-merge with main)

Three inline comments are attached to lines changed directly by this PR. The following issues were found in code already merged to `main` via prior PRs — they can't be raised as inline comments here but are worth tracking separately:

**`selectChangedOwners` (introduced in #10913, already in `main`)**

1. **Hardcoded `LIMIT 1000` with no pagination path** — if there are >1000 owner changes since `updatedAtFrom`, the poller silently drops the remainder. `EntityChangeLogMapper.selectEntityChanges` uses a `maxRows` parameter; this query should do the same, or the caller must document that it pages by advancing `updatedAtFrom` to the max value in the last batch.

2. **No metalake scope** — returns changed owners across all metalakes. `selectEntityChanges` scopes by `metalakeName`; the contract here should either match or be explicitly documented as intentionally global.

**`OwnerInfo.ownerId` (already in `main`)**

3. **Primitive `long` risks NPE from MyBatis** — `selectOwnerByMetadataObjectIdAndType` returns `null` when no row matches. If MyBatis tries to set a null `owner_id` column value into the primitive `long` field it will throw `NullPointerException`. Use `Long` (boxed) or add a `NOT NULL` guarantee at the DB level.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
roryqi: Upper layer has the tree lock to avoid this case.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/OwnerMetaBaseSQLProvider.java
roryqi: This is acceptable. We don't align to the PostgresSQL fully.

core/src/main/java/org/apache/gravitino/utils/HierarchicalSchemaUtil.java
roryqi: This will be used for the next pull requests. I will add the comments.

roryqi:
> ## Review Summary (post-merge with main)
> Three inline comments are attached to lines changed directly by this PR. The following issues were found in code already merged to `main` via prior PRs — they can't be raised as inline comments here but are worth tracking separately:
> 
> **`selectChangedOwners` (introduced in #10913, already in `main`)**
> 
> 1. **Hardcoded `LIMIT 1000` with no pagination path** — if there are >1000 owner changes since `updatedAtFrom`, the poller silently drops the remainder. `EntityChangeLogMapper.selectEntityChanges` uses a `maxRows` parameter; this query should do the same, or the caller must document that it pages by advancing `updatedAtFrom` to the max value in the last batch.
> 2. **No metalake scope** — returns changed owners across all metalakes. `selectEntityChanges` scopes by `metalakeName`; the contract here should either match or be explicitly documented as intentionally global.
> 
> **`OwnerInfo.ownerId` (already in `main`)**
> 
> 3. **Primitive `long` risks NPE from MyBatis** — `selectOwnerByMetadataObjectIdAndType` returns `null` when no row matches. If MyBatis tries to set a null `owner_id` column value into the primitive `long` field it will throw `NullPointerException`. Use `Long` (boxed) or add a `NOT NULL` guarantee at the DB level.

This is out of this pull request scope.

jerryshao (approved):

PR #11041, opened 2026-05-11
[#11040] fix(core): make Iceberg import idempotent under concurrent load

### LINKED ISSUE #11040: [Improvement] Iceberg catalog concurrent import can fail with EntityAlreadyExistsException in HA deployments

In the Iceberg catalog path, concurrent schema/table import can fail in multi-node deployments when another node has already imported the same entity into Gravitino `EntityStore`.

### How should we improve?

Make concurrent import idempotent by reusing the existing Gravitino entity when `EntityAlreadyExistsException` is raised during import, instead of failing the request.

yuqi1129:

Make Iceberg schema and table import idempotent when multiple Gravitino nodes try to import the same object concurrently. If `EntityStore.put(..., true)` reports that the entity already exists, the dispatcher now reuses the existing Gravitino entity instead of failing the request.

Also reconcile Gravitino `EntityStore` metadata after Iceberg REST Catalog table/view `drop` and `rename` operations. After the Iceberg backend operation succeeds, the hook checks the current backend state and then either imports the table/view again or removes the stale Gravitino entity.

In HA deployments, another node can finish the same import first. The current code treats that race as an error, even though the entity is already present in Gravitino. That makes already-existing schema/table imports fail spuriously.

Iceberg REST requests can also be served by different Gravitino nodes. Because the current TreeLock is process-local, another node may drop or recreate the same Iceberg table/view between the backend operation and the hook's direct `EntityStore` update/delete. Without reconciliation, Gravitino may keep stale/orphan table or view metadata that no longer matches the Iceberg backend. This PR does not introduce a distributed TreeLock; it adds a lightweight backend-state reconciliation step for the high-risk IRC rename/drop paths.

Fix: #11040

Yes. Concurrent import/load requests become idempotent instead of failing with `EntityAlreadyExistsException` in the import path. IRC table/view rename and drop operations also reduce stale/orphan Gravitino metadata in multi-node deployments by reconciling against the Iceberg backend state.

`./gradlew :core:test --tests org.apache.gravitino.catalog.TestTableOperationDispatcher --tests org.apache.gravitino.catalog.TestSchemaOperationDispatcher -PskipITs`

`./gradlew :iceberg:iceberg-rest-server:test --tests org.apache.gravitino.iceberg.service.dispatcher.TestIcebergTableHookDispatcher --tests org.apache.gravitino.iceberg.service.dispatcher.TestIcebergViewHookDispatcher -PskipITs`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
@@ -192,13 +192,27 @@ public class SchemaOperationDispatcher extends OperationDispatcher implements Sc
         TreeLockUtils.doWithTreeLock(ident, LockType.READ, () -> internalLoadSchema(ident));
 
     if (!schema.imported()) {
-      TreeLockUtils.doWithTreeLock(
-          NameIdentifier.of(ident.namespace().levels()),
-          LockType.WRITE,
-          () -> {
-            importSchema(ident);
-            return null;
-          });
+      try {
+        TreeLockUtils.doWithTreeLock(
+            NameIdentifier.of(ident.namespace().levels()),
+            LockType.WRITE,
+            () -> {
+              importSchema(ident);
+              return null;
+            });
+      } catch (EntityAlreadyExistsException e) {
+        // HA race: another Gravitino node concurrently imported this schema. Reload from the
+        // entity store to verify the entity stored by the winning node is consistent.
+        LOG.info(
+            "Schema {} was concurrently imported by another node; reloading from store.", ident);
+        EntityCombinedSchema reloaded =
+            TreeLockUtils.doWithTreeLock(ident, LockType.READ, () -> internalLoadSchema(ident));
+        if (!reloaded.imported()) {
+          throw new UnsupportedOperationException(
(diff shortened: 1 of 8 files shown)


roryqi:
Does distributed tree lock solve this issue? This seems a dangerous point fix. Because we need to delete the data.

yuqi1129:
> Does distributed tree lock solve this issue? This seems a dangerous point fix. Because we need to delete the data.

It may be. I believe this problem still exists if IRC was deployed in the standalone mode, in this way,  the distributed tree lock can't handle this scenario.

If IRC and Gravitino run within a single JVM and are all protected by a distributed tree lock, I suppose we may solve it; anyway, I have no detailed data about the result when the distributed tree lock is supported.

roryqi:
> > Does distributed tree lock solve this issue? This seems a dangerous point fix. Because we need to delete the data.
> 
> It may be. I believe this problem still exists if IRC was deployed in the standalone mode, in this way, the distributed tree lock can't handle this scenario.
> 
> If IRC and Gravitino run within a single JVM and are all protected by a distributed tree lock, I suppose we may solve it; anyway, I have no detailed data about the result when the distributed tree lock is supported.

Could we fix this issue in the dispatcher layer instead of entity store layer?

roryqi (approved):
LGTM.

PR #11042, opened 2026-05-11
[#11039] fix(iceberg): Fast fail missing OSS table metadata

### LINKED ISSUE #11039: [Bug] Iceberg REST retries Aliyun OSS NoSuchKey and holds Gravitino tree locks for missing metadata files
### What happened

When an Iceberg table registered in Gravitino points to an Aliyun OSS metadata JSON that no longer exists, Gravitino's Iceberg REST service can spend a long time retrying the missing object. During that time, Gravitino table/tag/policy requests may hold tree locks for more than 30 seconds, causing the UI and metadata APIs for the same schema to appear stuck.

### Why this matters

A single stale or externally damaged Iceberg table can degrade a whole schema's metadata browsing experience. This is especially visible in the web UI, which may issue table detail, tags, policies, and existence checks close together.

### Observed behavior

For a table similar to `metalake.catalog.schema.table`, where the JDBC Iceberg catalog row still pointed at a missing metadata file, the Gravitino server log repeatedly showed:

```text
Retrying task after failure: sleepTimeMs=... The specified key does not exist.
<Key>.../metadata/00000-...metadata.json</Key>
```

The same time window also showed tree-lock watchdog warnings similar to:

```text
Thread with identifier ThreadIdentifier{thread=Thread[Gravitino-webserver-...,5,main], ident=metalake.catalog.schema.table} holds the lock node TreeLockNode{ident=/,...} for more than 30s ... io-connection hangs
```

After unregistering/dropping the stale table entry, the same API returned quickly with `NoSuchTableException`/404.

### Expected behavior

Missing Aliyun OSS metadata objects should be treated as a non-retryable not-found condition and should fail quickly. A stale table should not cause long retry loops that hold Gravitino tree locks and block unrelated metadata browsing in the same schema.

### Root cause analysis

Gravitino 1.2.0 uses Iceberg 1.10.1. Iceberg's `BaseMetastoreTableOperations.refreshFromMetadataLocation` stops retrying on Iceberg `NotFoundException`, but the Aliyun OSS input stream path appears to propagate Aliyun SDK `OSSException` for `NoSuchKey` directly from `client.getObject(...)`. That exception is not classified as Iceberg `NotFoundException`, so Iceberg's generic retry path keeps retrying.

By contrast, Iceberg's AWS S3 input stream maps `NoSuchKeyException` to `NotFoundException`, allowing the retry loop to stop quickly.

Relevant upstream source references:

- Iceberg retry stops on `NotFoundException`: `BaseMetastoreTableOperations.refreshFromMetadataLocation`
- Iceberg Aliyun OSS stream reads with `client.getObject(...)` without translating `NoSuchKey`: `OSSInputStream.openStream`
- Iceberg AWS S3 stream translates `NoSuchKeyException` to `NotFoundException`: `S3InputStream.openStream`
- Gravitino table operations hold tree locks around load/drop paths: `TableOperationDispatcher.loadTable` and `dropTable`

### Possible fixes

1. In Gravitino's Iceberg REST wrapper, classify Aliyun OSS `NoSuchKey`/`NoSuchBucket` failures from metadata loading as a not-found table condition instead of allowing generic retries to continue.
2. Add a regression test around Iceberg REST `loadTable`/`tableExists` behavior when metadata loading fails with Aliyun OSS `NoSuchKey`.
3. Longer term, consider avoiding heavyweight `loadTable` calls for shallow existence checks used by UI authorization/tag/policy flows.

I am preparing a small PR for the fast-fail behavior.

atovk:

This PR adds an OSS metadata fast-fail path in `IcebergCatalogWrapper`:

- Before `loadTable` enters Iceberg's full catalog load path, use `SupportsMetadataLocation` to get the table metadata file location.
- For `oss://` metadata locations with a configured Iceberg `FileIO`, check `newInputFile(location).exists()` first.
- If the OSS metadata file is missing, `loadTable` throws `NoSuchTableException` and `tableExists` returns `false` without calling `catalog.loadTable`.
- Add regression tests that verify missing OSS metadata does not enter the catalog load path.

A stale Iceberg catalog entry can point to an OSS metadata file that has already been removed. In that case, Iceberg's Aliyun OSS metadata read path may spend a long time retrying `NoSuchKey` before the failure surfaces. Gravitino table operations hold tree locks while dispatching those catalog calls, so the retry loop can make UI/API table operations appear stuck and can block related operations behind the same lock path.

Using the OSS `exists()` path lets Gravitino detect the missing metadata file before entering Iceberg's metadata read retry loop and return the normal no-such-table result quickly.

Fix: #11039

For stale OSS-backed Iceberg table entries whose metadata file is already missing, table load/existence checks now fail fast with the existing no-such-table behavior instead of waiting for the full Iceberg metadata read retry loop.

- JDK 17: `./gradlew --no-daemon --max-workers=1 :iceberg:iceberg-common:test -PskipITs`
- JDK 17: `./gradlew --no-daemon --max-workers=1 :iceberg:iceberg-rest-server:test -PskipITs`


roryqi:
Should we fix this issue in the Iceberg side instead of Gravitino side?

atovk:
> Should we fix this issue in the Iceberg side instead of Gravitino side?

 I think we should keep the fix on the Gravitino side for this PR, and I will also follow up in Iceberg via apache/iceberg#16299 (https://github.com/apache/iceberg/issues/16299).

  Gravitino already knows the table metadata location through SupportsMetadataLocation, so it can short-circuit before entering Iceberg’s metadata read path. That gives us a fast and consistent failure for loadTable and tableExists, instead of waiting for Iceberg’s metadata read retry loop
  when the metadata file is missing.

  The Iceberg issue is still valid and complementary: some FileIO read paths, such as Aliyun OSS and Dell ECS, should translate missing-object read failures into Iceberg NotFoundException consistently. I’ll address that upstream in Iceberg, but it depends on an Iceberg change and a later
  dependency upgrade in Gravitino.

  So for this PR, I would keep the storage-neutral Gravitino fix, and use apache/iceberg#16299 as the upstream follow-up for the underlying FileIO/read-path behavior.

atovk:
@jerryshao  PTAL

PR #11046, opened 2026-05-11
[#11048] feat(core): batch SQL for JDBCBackend.batchGet GROUP entity

### LINKED ISSUE #11048: [Subtask] Add batchGet operation support for GROUP
### Describe the subtask

Currently batchGet gets individual GROUPs in separate DB calls. Group them to single call

### Parent issue

https://github.com/apache/gravitino/issues/10404

bharos:

Replace the per-identifier loop for `GROUP` in `JDBCBackend.batchGet` with a single batched SQL SELECT that returns groups and their roles in one round-trip.

- New `listExtendedGroupPOsByMetalakeIdAndNames` in `GroupMetaBaseSQLProvider` (with H2 / PostgreSQL overrides) — mirrors the existing `listExtendedGroupPOsByMetalakeId` JOIN + `JSON_ARRAYAGG` template, filtered by `group_name IN (...)` via MyBatis `<script><foreach>` (portable IN-clause, soft-delete aware, uses indexed `(metalake_id, group_name)`).
- New `GroupMetaService.batchGetGroupByIdentifier`: cross-id same-namespace check, single transaction via `SessionUtils.doWithCommitAndFetchResult`, reuses `ExtendedGroupPO` + `POConverters.fromExtendedGroupPO` to attach role names/ids — no extra role round-trip.
- `JDBCBackend.batchGet` `GROUP` case dispatches to the new path.

Reduces N round-trips to a single JOIN query on the GROUP batch path.

Fix: #11048

No. Behavior contract is unchanged and consistent with sibling batches: missing groups skipped (no exception), duplicates deduped, order not preserved, all identifiers must share a namespace.

- New unit test `testBatchGetGroups` covers happy-path multi-group retrieval and role attachment.
- Existing `testBatchGetGroupsPartialResults` covers missing groups.
- `./gradlew :core:test --tests "...TestJDBCBackendBatchGet" --tests "...TestGroupMetaService" --tests "...TestRoleMetaService" -PskipITs` → pass.
- `./gradlew :core:spotlessApply` → clean.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/service/GroupMetaService.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/service/GroupMetaService.java
@@ -111,6 +111,48 @@ public class GroupMetaService {
 
   @Monitored(
       metricsSource = GRAVITINO_RELATIONAL_STORE_METRIC_NAME,
+      baseMetricName = "batchGetGroupByIdentifier")
+  public List<GroupEntity> batchGetGroupByIdentifier(List<NameIdentifier> identifiers) {
+    if (identifiers == null || identifiers.isEmpty()) {
+      return Collections.emptyList();
+    }
+
+    NameIdentifier firstIdent = identifiers.get(0);
+    Namespace namespace = firstIdent.namespace();
+    String metalake = NameIdentifierUtil.getMetalake(firstIdent);
+
+    for (NameIdentifier identifier : identifiers) {
+      AuthorizationUtils.checkGroup(identifier);
+      Preconditions.checkArgument(
+          identifier.namespace().equals(namespace),
+          "All group identifiers must belong to the same namespace, expected %s but got %s",
+          namespace,
+          identifier.namespace());
+    }
+
+    long metalakeId =
+        EntityIdService.getEntityId(NameIdentifier.of(metalake), Entity.EntityType.METALAKE);
+    List<String> groupNames =
+        identifiers.stream().map(NameIdentifier::name).collect(Collectors.toList());
+
+    return SessionUtils.doWithCommitAndFetchResult(
+        GroupMetaMapper.class,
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/gravitino/storage/relational/service/GroupMetaService.java
```diff
@@ -109,6 +109,48 @@ public GroupEntity getGroupByIdentifier(NameIdentifier identifier) {
...
+    return SessionUtils.doWithCommitAndFetchResult(
+        GroupMetaMapper.class,
+        mapper -> {
+          List<GroupPO> groupPOs =
+              mapper.batchSelectGroupMetaByMetalakeIdAndNames(metalakeId, groupNames);
+          return groupPOs.stream()
+              .map(
+                  groupPO -> {
+                    List<RolePO> rolePOs =
+                        RoleMetaService.getInstance().listRolesByGroupId(groupPO.getGroupId());
```
roryqi: This operation will cost a lot. Group entity only needs role names and role ids. Could we a use a batch operation to get all the group's role ids and role names?

roryqi: Maybe you can refer to fromExtendedGroupPO. We can join the table group and group_role_rel, filter the results using the list of group ids.

bharos:
Thanks @roryqi — refactored in 00c247c to use a single JOIN + JSON_ARRAYAGG query that mirrors listExtendedGroupPOsByMetalakeId (filtered by group_name IN (...)). Reuses ExtendedGroupPO + POConverters.fromExtendedGroupPO, so there is now only one round-trip and no new PO/converter code. All batch-get + group/role meta tests pass.

roryqi (approved):
LGTM.

PR #11047, opened 2026-05-11
[#4571] fix(catalog): jdbc-password from catalog API responses exposure

### LINKED ISSUE #4571: [Bug report] Backend API send plaintext password to frontend
### Version

main branch

### Describe what's wrong

Nowadays, backend API such as GET /metalake/{metalake}/catalog/{catalog} will return SOL password (jdbc-password) in plaintext. As the picture shows.  This might cause some potential security issues.

image

### How to reproduce

curl api/metalake/{metalake}/catalog/{catalog} to the server

### Additional context

_No response_

eduohe:

Sensitive catalog properties (`jdbc-password`, S3/OSS/Azure secret keys, token secrets) are now marked `hidden=true`. `BaseCatalog.properties()` already strips hidden properties before the REST response is built, so those values are no longer returned by `GET /api/metalakes/{metalake}/catalogs/{catalog}`.

To allow engine connectors (Trino, Spark) to still access these credentials, a new dedicated endpoint is added:

```
GET /api/metalakes/{metalake}/catalogs/{catalog}/credentials
```

This endpoint returns only the hidden properties and requires OWNER or higher privilege. Engine connectors are updated to fetch credentials from this endpoint and merge them into the catalog properties before configuring the underlying connector (e.g. mapping `jdbc-password` → `connection-password` for the PostgreSQL connector).

Affected catalogs:
- `catalog-jdbc-common` (MySQL, PostgreSQL, Doris, StarRocks) — `jdbc-password`
- `catalog-lakehouse-iceberg` — `jdbc-password`, S3/OSS/Azure secret keys
- `catalog-lakehouse-paimon` — `jdbc-password`, S3/OSS secret keys, DLF token

`GET /api/metalakes/{metalake}/catalogs/{catalog}` was returning sensitive credentials (passwords, secret keys) in plaintext. Any client with catalog read access could retrieve them. The fix hides those values from the standard catalog response while keeping them accessible to privileged callers through the new `/credentials` endpoint.

Fix: #4571

Yes — sensitive credential properties (`jdbc-password`, secret keys) are no longer included in `GET .../catalogs/{catalog}` responses. Clients that were reading credentials from that response must instead call `GET .../catalogs/{catalog}/credentials`.

Engine connectors (Trino, Spark) are updated automatically and require no configuration changes.

- Unit tests asserting hidden properties are absent from `catalog.properties()` and present in `credentialProperties()` for JDBC, Iceberg, and Paimon catalogs.
- Manually verified:
  1. `GET .../catalogs/catalog_postgres` — no `jdbc-password` in response.
  2. `GET .../catalogs/catalog_postgres/credentials` — returns `{"jdbc-password": "postgres", ...}`.
  3. Trino federated query across `catalog_hive` + `catalog_postgres` — returns results correctly (credentials fetched transparently via the new endpoint).

### DIFF
--- a/catalogs/catalog-jdbc-common/src/test/java/org/apache/gravitino/catalog/jdbc/TestJdbcCatalogPropertiesMetadata.java
+++ b/catalogs/catalog-jdbc-common/src/test/java/org/apache/gravitino/catalog/jdbc/TestJdbcCatalogPropertiesMetadata.java
@@ -40,6 +40,44 @@ import org.junit.jupiter.api.Test;
 public class TestJdbcCatalogPropertiesMetadata {
 
   @Test
+  public void testPasswordIsHidden() {
+    JdbcCatalogPropertiesMetadata propertiesMetadata = new JdbcCatalogPropertiesMetadata();
+
+    Assertions.assertTrue(propertiesMetadata.isHiddenProperty(JdbcConfig.PASSWORD.getKey()));
+  }
+
+  @Test
+  public void testPasswordIsNotReturnedInCatalogProperties() {
+    Map<String, String> createProperties =
+        ImmutableMap.<String, String>builder()
+            .put(JdbcConfig.JDBC_URL.getKey(), "jdbc:sqlite::memory:")
+            .put(JdbcConfig.JDBC_DRIVER.getKey(), "org.sqlite.JDBC")
+            .put(JdbcConfig.USERNAME.getKey(), "user")
+            .put(JdbcConfig.PASSWORD.getKey(), "secret")
+            .build();
+
+    CatalogEntity catalogEntity =
+        CatalogEntity.builder()
+            .withId(1L)
+            .withName("jdbc_catalog")
+            .withNamespace(Namespace.of("metalake"))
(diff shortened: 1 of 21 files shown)


yuqi1129:
@eduohe 

Thanks for your contribution. I'm afraid we can't mark properties like `jdbc-password` as hidden directly, as engines like Trino and Spark will depend on catalog properties to get credentials for those catalogs, so we'd better come up with another way to solve the problem.

eduohe:
Thanks for the reply @yuqi1129 , I'll work on it

eduohe:
@yuqi1129 thanks for the feedback. This updated PR addresses this by adding a dedicated GET .../catalogs/{catalog}/credentials endpoint that returns the hidden properties. The Trino and Spark connectors were also updated to fetch credentials from that endpoint and merge them before building the engine connector config, so connector functionality is fully preserved. jdbc-password is now hidden from the public catalog response without breaking anything downstream.

I also checked it the gravitino playground to confirm that the response doesn't contain sensitive infos.
gravitino_playground_api_response

PR #11052, opened 2026-05-12
[#11051] improvement(deps): Exclude Derby from hive-metastore-libs runtime classpath

### LINKED ISSUE #11051: [Improvement] Exclude Derby from hive-metastore-libs runtime classpath

Apache Derby is transitively pulled into the runtime classpath of `hive-metastore2-libs` and `hive-metastore3-libs` through Hive Metastore dependencies. Derby is only needed as an embedded database backend for Hive Metastore in test environments — production deployments use MySQL or PostgreSQL as the HMS backend.

Currently:
- `hive-metastore2-libs` ships Derby 10.10.2.0 in its runtimeClasspath
- `hive-metastore3-libs` ships Derby 10.14.1.0 in its runtimeClasspath

These jars end up in the production distribution (`distribution/package/catalogs/hive/libs/`) unnecessarily, increasing the attack surface and distribution size.

### How should we improve?

Exclude `org.apache.derby` from the Hive Metastore transitive dependencies in both modules. This is consistent with how other modules (`iceberg-common`, `hive-metastore-common`, `catalog-lakehouse-paimon`) already handle this dependency.

geyanggang:

Exclude `org.apache.derby` from the Hive Metastore transitive dependencies in `hive-metastore2-libs` and `hive-metastore3-libs`.

Derby is transitively pulled into the runtime classpath of both hive-metastore-libs modules through Hive Metastore. It is only needed as an embedded database for HMS in tests — production uses MySQL/PostgreSQL. Other modules (`iceberg-common`, `hive-metastore-common`, `catalog-lakehouse-paimon`) already exclude Derby from runtime. This change makes the two libs modules consistent and removes an unnecessary dependency from the production distribution.

Fix: #11051 

No. Derby was never intentionally included in the production distribution and is not used at runtime.

- Verified `runtimeClasspath` of both modules no longer contains Derby via `./gradlew :catalogs:hive-metastore2-libs:dependencies --configuration runtimeClasspath`
- Build passes for both affected modules

### DIFF
--- a/catalogs/hive-metastore2-libs/build.gradle.kts
+++ b/catalogs/hive-metastore2-libs/build.gradle.kts
@@ -77,6 +77,7 @@ dependencies {
     exclude(group = "junit")
     exclude(group = "org.apache.ant")
     exclude(group = "org.apache.avro")
+    exclude(group = "org.apache.derby")
     exclude(group = "org.apache.hadoop", module = "hadoop-yarn-server-resourcemanager")
     exclude(group = "org.apache.hbase")
     exclude(group = "org.apache.logging.log4j")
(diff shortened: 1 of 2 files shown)


diqiu50 (approved):
LGTM

PR #11060, opened 2026-05-12
[#10983] feat(lance): Upgrade lance-core to 6.0.0 and lance-namespace to 0.7.5

### LINKED ISSUE #10983: [Improvement] Upgrade LanceDB dependency in Gravitino

Hi, I noticed the LanceDB dependencies are a bit old: 

```
lance = "0.39.0"
lance-namespace = "0.0.20"
```
in https://github.com/apache/gravitino/blob/main/gradle/libs.versions.toml#L31-L32  .

Shall we update to latest dependency - https://mvnrepository.com/artifact/org.lance/lance-namespace-core/versions? Note the latest should be 0.7.5. Thanks!

### How should we improve?

_No response_

yuqi1129:

Upgrades the Lance dependencies:
- `lance-core`: `2.0.1` -> `6.0.0`
- `lance-namespace-core`: `0.4.5` -> `0.7.5`

API-breaking changes addressed:
1. **`Transaction` -> `SourcedTransaction`** in `LancePartitionStatisticStorage`: `Dataset.newTransactionBuilder()` now returns `SourcedTransaction.Builder` instead of `Transaction.Builder`.
2. **`CreateEmptyTableRequest/Response` -> `DeclareTableRequest/Response`**: `createEmptyTable()` was removed from the `LanceNamespace` interface; all usages migrated to `declareTable()`.
3. **`TableApi.createTable()` signature**: Two new optional parameters (`properties`, `storageOptions`) added; pass `null` for both since Gravitino reads from HTTP headers.
4. **`AlterColumnsEntry.rename` is now `JsonNullable<String>`**: Added a JAX-RS `@Provider` (`JsonNullableMapperProvider`) that registers `JsonNullableModule` with the `ObjectMapper` used by the lance-rest-server.
5. **Typed Lance namespace errors**: Updated Lance REST integration assertions to validate `LanceNamespaceException#getCode()` instead of matching old serialized JSON error strings.

Testability fix:
- Introduced a package-private `DatasetHolder` wrapper class in `LancePartitionStatisticStorage` so the Caffeine cache holds `DatasetHolder` instead of `Dataset` directly. Newer `lance-core` versions expose many JNI native methods that prevent Mockito's inline mock-maker from instrumenting `Dataset` directly.

Keeping Lance dependencies up to date ensures we receive bug fixes, performance improvements, and security patches while keeping Gravitino compatible with current Lance namespace clients.

Fix: #10983

No. The REST API contract is unchanged. The internal migration from `createEmptyTable` to `declareTable` is transparent to users.

- `./gradlew :core:compileJava :lance:lance-common:compileJava :lance:lance-rest-server:compileJava -PskipWeb=true`
- `./gradlew :lance:lance-rest-server:spotlessApply :core:test --tests org.apache.gravitino.stats.storage.TestLancePartitionStatisticStorage :lance:lance-common:test :lance:lance-rest-server:test -PskipWeb=true -PskipDockerTests=true`


yuqi1129:
@FANNG1 

FANNG1:
Thanks for adding the matrix runners and documenting how to re-verify the Lance compatibility ranges. I think the CI strategy here needs to be called out explicitly before we merge this.

My concern is that the full compatibility matrix should probably not become part of the default PR validation path. These tests are expensive, depend on external package ecosystems, and can become noisy when PyPI / Maven resolution changes upstream. That makes them a poor fit for the main per-PR signal.

What seems more sustainable is a two-layer approach:
1. Keep one pinned, known-good Lance combination in the normal CI path as the regression guard.
2. Keep the full multi-version matrix as a separate automation path: scheduled workflow, release-time validation, and manual dispatch when we need to re-verify or extend the compatibility claims.

That would still keep the matrix exercised continuously enough to avoid script rot, but without making every PR pay the cost or absorb upstream ecosystem flakiness. If we go this direction, I also think the matrix runner should avoid floating dependencies by default (for example, not resolving `lance-namespace` as unpinned `latest`).

Can we align on that execution model in this PR? Right now the tasks and docs are being added, but the intended automation boundary is still unclear.

yuqi1129:
@FANNG1 

Thanks for your reply. 

By default, we will only run one version of  `lance-spark` and `lance-ray` to verify whether it works fine. As you point out, there are several versions of `lance-spark` and `lance-ray`, and it aims to provide a full test in future upgrades and in a possible CI environment.  Is there acceptable for us?

FANNG1:
Thanks, that makes sense to me regarding the default CI path only exercising one known-good version.

One follow-up on the matrix runner itself: would you mind pinning the default `lance-namespace` version there as well, instead of resolving `latest`? Since the normal CI path is intentionally single-version and reproducible, it would be good if the matrix driver were reproducible by default too, rather than drifting with PyPI.

Also, what do you think about the PR scope? The dependency upgrade / API adaptation part and the compatibility-matrix tooling/docs feel somewhat separable. I am okay either way, but if more CI wiring is planned later, splitting those concerns might make review and future maintenance a bit easier.

yuqi1129:
> would you mind pinning the default lance-namespace version there as well, instead of resolving latest? 

Of course, I can pin it to a fixed version.

> The dependency upgrade / API adaptation part and the compatibility-matrix tooling/docs feel somewhat separable.

It's acceptable to me for both. If you prefer to exclude multiple version support in this PR, I can remove it now.

yuqi1129:
@FANNG1 
I have split it.

gradle/libs.versions.toml
```diff
@@ -28,8 +28,8 @@ guava = "32.1.3-jre"
 lombok = "1.18.20"
 slf4j = "2.0.16"
 log4j = "2.25.4"
-lance = "2.0.1"
-lance-namespace = "0.4.5"
+lance = "4.0.1"
+lance-namespace = "0.7.5"
```
roryqi, line 32: 0.7.6 is released. Could we use 0.7.6?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -164,7 +163,7 @@ public Response createEmptyTable(
       // conflict.
       props.putAll(headerProps);
 
-      CreateEmptyTableResponse response =
+      DeclareTableResponse response =
```
FANNG1: Does it have compatiable problem with old clients?

docs/lance-rest-integration.md
```diff
@@ -22,12 +22,17 @@ The following table outlines the tested compatibility between Gravitino versions
 
 | Gravitino Version (Lance REST) | Supported lance-spark Versions | Supported lance-ray Versions |
 |--------------------------------|--------------------------------|------------------------------|
-| 1.1.1 - 1.2.1                  | 0.0.10 - 0.0.15               | 0.0.6 - 0.0.8                |
-| 1.3.0                          | 0.1.0 - 0.2.0                 | 0.0.6 - 0.2.0                |
+| 1.1.1 - 1.2.1                  | 0.0.10 - 0.0.15                | 0.0.6 - 0.0.8                |
+| 1.3.0                          | 0.1.0 - 0.4.0                  | 0.3.0 - 0.4.2                |
```
FANNG1: after upgrading,  lance ray `0.0.6 - 0.2.0 ` is not supported any more?

gradle/libs.versions.toml
yuqi1129: It's okay for me to use both as they are released within a day.
image

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: Yes, `DeclareTableResponse` has been removed from API.

docs/lance-rest-integration.md
yuqi1129: Confirmed unsupported. Spun up Gravitino 1.3.0 (post-upgrade) and ran the lance-ray IT against each version with `lance-namespace==0.7.5` pinned. The PyPI install actually succeeds for all of these; the failures show up at runtime:

- **0.1.0**: `TypeError: write_lance() got an unexpected keyword argument 'namespace_impl'`. 0.1.0's `write_lance(...)` takes `namespace: LanceNamespace` directly; the IT (and 0.2.0+) use the new `namespace_impl: str` + `namespace_properties: dict` form, so the call doesn't even reach the server.
- **0.2.0**: `ImportError: cannot import name 'LanceNamespaceStorageOptionsProvider' from 'lance'`. The IT call signature matches, but `lance_ray.utils.create_storage_options_provider` does `from lance import LanceNamespaceStorageOptionsProvider`, and that symbol no longer exists in the `pylance` 6.0.0 wheel that `lance-namespace==0.7.5` pulls in.
- **0.3.0** (positive control): PASS — 1000-row Ray dataset round-trips through write_lance / read_lance.

I updated `docs/lance-rest-integration.md` in #11119 to replace the inaccurate "pip resolves them with an older lance-namespace" line with these two concrete failure modes, so future readers know exactly what breaks.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
FANNG1: seems we could remove the createEmptyTable in this version, since it's too old and could't be used by old version (the response is changed from `CreateEmptyTableResponse ` to `DeclareTableResponse `), WDYT?

docs/lance-rest-integration.md
FANNG1: seems we shouldn't pin `lance-namespace==0.7.5`,  `lance-ray` will choose corresponding  `lance-namespace` version.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
yuqi1129: It also okay for me.

yuqi1129: removed.

docs/lance-rest-integration.md
yuqi1129: changed.

FANNG1:
Current issues after upgrading Lance Namespace to 0.7.x:

1. CreateTable does not handle the new 0.7.x REST parameters. In 0.7.x, properties and storage_options are passed as JSON-encoded query parameters, but the current server side still only follows the old behavior, so values sent by new clients will be ignored.

2. DescribeTable does not handle check_declared. In 0.7.x, check_declared is passed as a query parameter, and the response should populate is_only_declared when requested. Currently this is not parsed/returned.

3. DeclareTable should preserve/use request.properties and return the new response fields consistently if supported, especially properties and managed_versioning.

From the client compatibility side, additive response fields should be safe. I checked the Lance generated clients: Java disables FAIL_ON_UNKNOWN_PROPERTIES, Rust serde models do not use deny_unknown_fields, and Python generated models only extract known keys in from_dict. So older clients should ignore these new response fields instead of failing.

The main compatibility gap is server-side request handling: after this dependency upgrade, the REST server should accept and process the new 0.7.x request parameters rather than silently dropping them.

FANNG1:
What do you think about the handling of these compatibility gaps? Should we address them in this PR, or would you prefer to keep this PR focused on the dependency upgrade and fix them in a follow-up PR? Either way works for me.

yuqi1129:
> What do you think about the handling of these compatibility gaps? Should we address them in this PR, or would you prefer to keep this PR focused on the dependency upgrade and fix them in a follow-up PR? Either way works for me.

Let me evaluate the change. If it's too large to include them, I would prefer to use another PR to track it.

beinan:
Many thanks @yuqi1129 for working on this!  Now we (Microsoft MST) are blocked by this feature,  can we have this PR merged soon? 

The most recent release version of lance is 6.0.0 https://github.com/lance-format/lance/releases/tag/v6.0.0 can we also bump the version of lance-core to 6.0.0 also?  I think it's backward compatible.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
```diff
@@ -39,7 +39,7 @@ public class LanceConstants {
 
   public static final String LANCE_TABLE_VERSION = "lance.version";
   // Mark whether it is to create an empty Lance table(no data files)
-  public static final String LANCE_TABLE_CREATE_EMPTY = "lance.create-empty";
+  public static final String LANCE_TABLE_DECLARED = "lance.declared";
```
jerryshao, line 42: The change here will break the compatibility, right?

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LancePropertiesUtils.java
```diff
@@ -68,4 +68,14 @@ public static Map<String, String> resolveLanceStorageOptions(
...
+
+  public static Map<String, String> toTableProperties(Map<String, String> storageOptions) {
+    if (storageOptions == null) {
+      return Map.of();
+    }
+
+    return storageOptions.entrySet().stream()
+        .collect(
+            Collectors.toMap(e -> LANCE_STORAGE_OPTIONS_PREFIX + e.getKey(), Map.Entry::getValue));
+  }
```
jerryshao, line 88: We can simplify as: `storageOptions == null ? Collections.emptyMap() : ...`

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
yuqi1129: Yes. several breaking changes, including that we no longer support `createEmptyTable`.

core/src/main/java/org/apache/gravitino/stats/storage/LancePartitionStatisticStorage.java
```diff
@@ -203,11 +204,12 @@ public LancePartitionStatisticStorage(Map<String, String> properties) {
...
-                      (RemovalListener<Long, Dataset>)
+                  .executor(Runnable::run)
+                  .removalListener(
+                      (RemovalListener<Long, DatasetHolder>)
                           (key, value, cause) -> {
                             if (value != null) {
-                              value.close();
+                              closeDatasetHolder(value);
                             }
                           })
```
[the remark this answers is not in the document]
yuqi1129: Updated in 4b4442d48: removed the synchronous Caffeine executor, made holder close idempotent, and explicitly closes cached holders during storage shutdown before closing the allocator.

clients/client-python/tests/integration/test_lance_ray.py
```diff
@@ -0,0 +1,335 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        gravitino_home = os.environ.get("GRAVITINO_HOME")
+        if not gravitino_home:
+            raise RuntimeError(
+                "GRAVITINO_HOME must be set to the distribution package directory"
+            )
```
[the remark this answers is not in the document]
yuqi1129: Updated in 4b4442d48: `setUpClass` now uses the existing `_get_gravitino_home()` behavior instead of raising a RuntimeError directly.

core/build.gradle.kts
```diff
@@ -54,6 +54,13 @@ dependencies {
     exclude(group = "com.fasterxml.jackson.jaxrs", module = "jackson-jaxrs-json-provider") // using gravitino's version
     exclude(group = "org.apache.httpcomponents.client5", module = "*") // provided by gravitino
     exclude(group = "org.lance", module = "lance-namespace-core") // This is unnecessary in the core module
+    // Same rationale as lance-namespace-core: lance-core 4.0.1 declares
+    // lance-namespace-apache-client as a transitive, but core never calls into it.
+    // Leaving it on the main classpath shadows the lance-rest aux service's own
+    // lance-namespace-apache-client (loaded via lance-rest-server/libs/), and
+    // because the aux classloader is parent-first, the older transitive wins
+    // on request deserialization (e.g. dropping fields like `check_declared`).
+    exclude(group = "org.lance", module = "lance-namespace-apache-client")
```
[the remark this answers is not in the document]
yuqi1129: Updated in 4b4442d48.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LancePropertiesUtils.java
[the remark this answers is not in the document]
yuqi1129: Updated in 4b4442d48: added Javadoc for the public helper.

yuqi1129: Updated in 4b4442d48: simplified the null handling and added Javadoc for `toTableProperties`.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -282,14 +282,14 @@ Table createTableInternal(
...
     boolean createEmpty =
-        Optional.ofNullable(properties.get(LanceConstants.LANCE_TABLE_CREATE_EMPTY))
+        Optional.ofNullable(properties.get(LanceConstants.LANCE_TABLE_DECLARED))
             .map(Boolean::parseBoolean)
             .orElse(false);
     if (createEmpty) {
-      // For create empty table, we just create the table metadata in Gravitino without creating
-      // the underlying Lance dataset.
+      // For declare table, we just create the table metadata in Gravitino without creating the
+      // underlying Lance dataset.
```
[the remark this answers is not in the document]
yuqi1129: Updated in ee93f7164: renamed the local variable to `declaredOnly`.

clients/client-python/setup.py
```diff
@@ -56,6 +56,7 @@
     install_requires=open("requirements.txt").read(),
     extras_require={
         "dev": open("requirements-dev.txt").read(),
+        "lance": open("requirements-lance.txt").read(),
     },
```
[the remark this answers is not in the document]
yuqi1129: Updated in ee93f7164: added `requirements-lance.txt` to `MANIFEST.in`.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/utils/LanceConstants.java
```diff
@@ -39,7 +39,7 @@ public class LanceConstants {
 
   public static final String LANCE_TABLE_VERSION = "lance.version";
   // Mark whether it is to create an empty Lance table(no data files)
```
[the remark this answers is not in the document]
yuqi1129: Updated in ee93f7164: refreshed the comment to describe metadata-only declare-table semantics.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestLanceNamespaceOperations.java
```diff
@@ -115,6 +116,7 @@ protected Application configure() {
     resourceConfig.property(CommonProperties.FEATURE_AUTO_DISCOVERY_DISABLE, true);
```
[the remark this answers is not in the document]
yuqi1129: Updated in ee93f7164: removed the stale `extractPropertiesFromBody` reference from the test setup comment.

jerryshao:
I have no further comments. @FANNG1 can you please help to review?

FANNG1:
## Review: handling of new lance-namespace fields

I reviewed how this PR handles the fields introduced between lance-namespace `0.4.5` and `0.7.5`. Most of the new fields (`CreateTable` query params, `DescribeTableResponse.properties/managed_versioning/is_only_declared`, `DescribeTableRequest.check_declared`, `DeclareTableRequest/Response.properties`, `AlterColumnsEntry.rename` -> `JsonNullable<String>`) are handled correctly. A few concerns below.

### Main issue: `is_only_declared` semantics do not match the Lance spec

`GravitinoLanceTableOperations.describeTable()` fills `is_only_declared` by reading the `lance.declared` property:

```java
if (checkDeclared) {
    response.setIsOnlyDeclared(
        Boolean.parseBoolean(table.properties().getOrDefault(LANCE_TABLE_DECLARED, "false")));
}
```

But `lance.declared`:
- is written exactly once, inside `declareTable()`;
- is registered as **immutable** in `LanceTableDelegator`, so it can never change afterwards.

The Lance spec (v0.7.0) defines `is_only_declared` as a **current state** — "the table is only declared in the namespace and has no underlying storage data/metadata files yet" — and expects it to flip to `false` once the table is materialized by the first data write.

Since the lance-rest-server does not implement `InsertIntoTable`, materialization happens client-side (e.g. lance-ray writes directly to storage) and the namespace never learns about it, so `lance.declared` stays `true` forever. **As a result, a table that was declared and then populated with real data will still report `is_only_declared=true` from `describeTable(check_declared=true)`, which is incorrect.**

Suggestion: when `check_declared=true`, probe the table location for an actual Lance manifest instead of reading the static property — or at minimum document this limited semantics for Gravitino.

### `ListTables` / `ListAllTables` ignore `include_declared` (v0.7.1)

`ListTablesRequest` has an `includeDeclared` field in 0.7.5, but `LanceNamespaceOperations.listTables` only accepts `delimiter / page_token / limit`. A client that explicitly sends `include_declared=false` is silently ignored and still receives declared-only tables.

### `lance.create-empty` -> `lance.declared` is a silent breaking change

The property key is renamed, but this key is persisted in Gravitino's metadata store. After an upgrade, existing tables carrying `lance.create-empty=true` will no longer be recognized by the `declaredOnly` check. The PR description states "No user-facing change", which is not accurate for existing deployments. Suggest a compatibility fallback:

```java
boolean declaredOnly = Boolean.parseBoolean(
    properties.getOrDefault(LANCE_TABLE_DECLARED,
        properties.getOrDefault(LANCE_TABLE_CREATE_EMPTY, "false")));
```

### Minor: `DescribeTableResponse.properties` exposes internal properties

`describeTable` assigns `table.properties()` to both `metadata` and `properties`. Per the v0.5.0 semantics, `properties` is meant to be namespace-managed business-logic properties, whereas `table.properties()` contains internal implementation details (`lance.creation-mode`, `lance.declared`, ...). Consider filtering before returning.

### Minor: `createTable` query-parameter precedence is undocumented

`props` (headers) -> `queryProperties` -> `queryStorageOptions` are applied via successive `putAll`, so query parameters silently override header values. Worth noting the precedence in the Javadoc.

yuqi1129:
> Main issue: is_only_declared semantics do not match the Lance spec

This problem will be solved in #11176

FANNG1:
## The lance-ray compatibility rationale in the docs is incorrect

I empirically tested the lance-ray version compatibility claim in `docs/lance-rest-integration.md`. The conclusion (0.3.0/0.4.2 work, <=0.2.0 don't) holds for a default `pip install`, but **the stated reason is wrong, and "0.2.0 not supported" is a dependency-resolution artifact rather than a real incompatibility.**

### What the doc says

> lance-ray 0.2.0 and earlier are *not* supported on 1.3.0 because pip resolves them with an older `lance-namespace` whose request schema is incompatible with the upgraded server-side `lance-namespace-core` (0.7.5+).

### Test 1 — dependency resolution (`pip install --dry-run`, fresh resolve)

Every lance-ray version resolves to the **same, newest** stack:

| lance-ray | resolved pylance | resolved lance-namespace |
|---|---|---|
| 0.0.6 / 0.0.8 / 0.1.0 / 0.2.0 / 0.3.0 / 0.4.2 | 6.0.1 | **0.7.7** |

pip never resolves an *older* lance-namespace — all the constraints are lower bounds (`>=`), so pip always picks the newest. The premise "pip resolves them with an older lance-namespace" is factually wrong.

### Test 2 — functional `write_lance` + `read_lance` (local `dir` namespace)

| lance-ray | with pylance 6.0.1 (pip default) |
|---|---|
| 0.2.0 | FAIL — `ImportError: cannot import name 'LanceNamespaceStorageOptionsProvider' from 'lance'` |
| 0.3.0 | OK |
| 0.4.2 | OK |

The break has nothing to do with the lance-namespace wire schema. lance-ray 0.2.0's code imports `LanceNamespaceStorageOptionsProvider` from `lance` (pylance), and that symbol was **removed in pylance 6.x**.

### Test 3 — lance-ray 0.2.0 with a matching pylance

`LanceNamespaceStorageOptionsProvider` is present in pylance 3.x/4.x, absent in 6.x. So:

```
lance-ray 0.2.0 + pylance 4.0.1 + lance-namespace 0.7.7  ->  WRITE+READ OK (100 rows)
```

**lance-ray 0.2.0 is not inherently incompatible** — it works once pylance is pinned to 3.x/4.x.

### Root cause

This is a 3-way version problem (`lance-ray` <-> `pylance` <-> `lance-namespace`), not the 2-way matrix the doc presents. lance-ray 0.2.0 declares `pylance>=2.0.0` with **no upper bound**, so a default `pip install` drags in pylance 6.x and breaks it. `requirements-lance.txt` pins `lance-ray` and `ray` but **not `pylance`**, so the resolution is left to drift — that is what the test environment actually exercises.

### Suggestions

1. Pin `pylance` explicitly in `requirements-lance.txt` alongside `lance-ray`, otherwise the resolved stack is non-deterministic.
2. Rewrite the compatibility note: the constraint comes from `lance-ray` <-> `pylance` API compatibility, not from a lance-namespace schema mismatch. Ideally document the compatible `pylance` range per `lance-ray` version.
3. Separately, note the real schema risk in the **opposite** direction: pip installs the lance-namespace **0.7.7** client, while this PR's server uses lance-namespace-core **0.7.5** — the client is newer than the server. Consider bumping the server to 0.7.7 to match.

FANNG1:
> > Main issue: is_only_declared semantics do not match the Lance spec
>
> This problem will be solved in #11176

@yuqi1129 Thanks for the pointer. Looking at #11176, it refreshes the table **column schema** from the Lance dataset on `loadTable`, while this concern is about the **`is_only_declared`** field in the `describeTable` REST response (derived from the immutable `lance.declared` property). They look related but seem like different problems, so they may not necessarily be fixed in the same PR. I've filed a separate issue to track it: #11178 — feel free to fold it into #11176 if it fits, otherwise it can be tracked independently.

FANNG1:
LGTM

FANNG1 (approved):

PR #11061, opened 2026-05-13
[#11063] fix(lance): manage storage config in Gravitino

### LINKED ISSUE #11063: [Improvement] Let Gravitino manage Lance storage configuration to simplify Spark integration

When Spark queries Lance tables through Gravitino Lance REST, the Spark side still needs to provide the underlying storage configuration explicitly, such as:

```text
spark.sql.catalog.lance.storage.endpoint=http://127.0.0.1:9000
spark.sql.catalog.lance.storage.aws_allow_http=true
spark.sql.catalog.lance.storage.access_key_id=minioadmin
spark.sql.catalog.lance.storage.secret_access_key=minioadmin
spark.sql.catalog.lance.storage.region=us-east-1
```

This makes the integration much more complicated than expected. In practice, users already register the Lance catalog in Gravitino, and Gravitino already knows the table location, such as `s3://contacts/raw/lance`. However, Spark still has to repeat the storage endpoint and credentials to make Lance queries work.

This leads to several problems:

1. Spark configuration is too verbose for a simple federated query scenario.
2. Storage credentials are duplicated between Gravitino and Spark.
3. Users need to understand internal storage details even though they are already managed by Gravitino.
4. The integration experience is inconsistent with the goal of using Gravitino as the metadata and access entry point.

For Spark + Gravitino + Lance integration, it would be better if users only needed to configure the Gravitino/Lance REST endpoint and catalog identity, without passing storage credentials again on the Spark side.

### How should we improve?

Gravitino should manage and provide the Lance storage configuration for Spark queries, instead of requiring Spark to pass the storage parameters manually.

Possible directions:

1. Allow the Lance catalog properties in Gravitino to include the required storage configuration, such as endpoint, region, path-style access, and credential provider information.
2. Make Lance REST read and use these managed properties when serving Spark requests.
3. If credentials need to be delegated, provide a Gravitino-managed mechanism similar in spirit to the Iceberg REST flow, so Spark does not need direct static storage credentials.
4. Keep the Spark-side configuration minimal, ideally only requiring:
   - Lance Spark extension
   - Lance catalog type/URI
   - Gravitino/Lance REST endpoint
   - metalake/catalog identification if needed

Expected result:

Users can query Lance tables from Spark with much simpler configuration, while Gravitino remains the single place to manage storage-related settings for Lance catalogs.

FANNG1:

This PR moves Lance storage configuration into Gravitino-managed catalog properties and teaches Lance REST to use those catalog defaults when serving Spark clients. It removes the need to pass spark.sql.catalog.lance.storage.* in the demo path, while keeping table-level lance.storage.* overrides supported.

Spark currently has to repeat MinIO/S3 settings that are already known to Gravitino, which makes Lance integration verbose and error-prone. This change keeps Gravitino as the source of truth for Lance storage configuration and simplifies Spark usage for Gravitino-managed Lance catalogs.

Fix: #11063

Yes.

- lance.storage.* is now supported as a catalog-level property family for generic lakehouse catalogs.
- Lance REST responses now resolve storage options from catalog defaults when table-level values are absent.
- The Spark demo no longer requires explicit Lance storage conf entries.
- Table-level lance.storage.* remains supported as an override.

- ./gradlew :lance:lance-common:test --tests org.apache.gravitino.lance.common.utils.TestLancePropertiesUtils -PskipITs -PskipDockerTests=true
- ./gradlew :catalogs:catalog-lakehouse-generic:test --tests org.apache.gravitino.catalog.lakehouse.generic.TestPropertiesMetadata -PskipITs -PskipDockerTests=true
- ./gradlew :lance:lance-rest-server:compileTestJava -PskipITs -PskipDockerTests=true


FANNG1:
Superseded by #11070 on a clean branch containing only the Lance storage-config fix.

PR #11074, opened 2026-05-13
[#11103] feat(core): Add the hierarchy conversion layer

### LINKED ISSUE #11103: [Subtask] Add the hierarchy convention layer
### Describe the subtask

Add the hierarchy convention layer

### Parent issue

Add the hierarchy convention layer

roryqi:

Add abstract class BasePOStorageOps to let all the PO logic to one class

Add SchemaPOStorageOps,TablePOStorageOps,FunctionPOStorageOps,ViewPOStorageOps.

Add Hierachical convention class

Fix: #11103 

No.

Added tests.


roryqi:
@jerryshao @yuqi1129 Could u help me review this pull request?

jerryshao:
Is it ready to review?

roryqi:
> Is it ready to review?

Yes, it is ready to review.

jerryshao:
The overall direction here is good — extracting per-entity PO logic into dedicated classes and using a decorator for hierarchical name translation is a clean separation. A few observations on the design that might be worth considering.

**The core tension**

`BasePOStorageOps.getPO/listPOs` currently combines two orthogonal concerns:
1. **Name translation** (hierarchical logical ↔ physical) — handled well by `HierarchicalConventionPOStorageOps`
2. **Query routing** (cache-enabled → ID-based; no cache → full-name JOIN) — currently embedded via `GravitinoEnv.getInstance().cacheEnabled()` in `BasePOStorageOps`

Mixing these in the base class causes the `Capability` enum + runtime `UnsupportedOperationException` pattern, the `GravitinoEnv` static dependency leaking into a storage layer class, and the issue in `MetadataObjectService.TYPE_TO_STORAGE_OPS_MAP` where raw `*POStorageOps` instances bypass the naming convention.

**An alternative: expose two explicit SQL methods, route in the service**

```java
// BasePOStorageOps: pure SQL delegation, no routing logic
abstract class BasePOStorageOps<PO, Mapper> {
    // ID-based (cache-enabled path)
    public PO getPOByParentId(Mapper mapper, Long parentId, String name) { ... }
    public List<PO> listPOsByParentId(Mapper mapper, Long parentId) { ... }

    // Name-based (full join, no-cache path)
    public PO getPOByFullName(Mapper mapper, NameIdentifier ident) { ... }
    public List<PO> listPOsByNSFullName(Mapper mapper, Namespace ns) { ... }
}
```

Routing moves back to the service as a single private helper per operation — not repeated per-method:

```java
// SchemaMetaService
private SchemaPO getSchemaPO(NameIdentifier ident) {
    return SessionUtils.getWithoutCommit(SchemaMetaMapper.class, mapper ->
        cacheEnabled()
            ? ops.getPOByParentId(mapper, getCatalogId(ident), ident.name())
            : ops.getPOByFullName(mapper, ident));
}
```

Since `ops` is still a `HierarchicalConventionPOStorageOps`, both paths get name translation automatically. `MetadataObjectService` can store wrapped instances in its map instead of raw ops, fixing the naming convention bypass.

This eliminates the `Capability` enum, the `GravitinoEnv` dependency in the base class, and the dispatcher `getPO(Mapper, NameIdentifier)`. Each layer has one job: name translation in the decorator, SQL in the ops, routing in the service.

**What stays the same**

`HierarchicalConventionPOStorageOps` as a decorator and the `*POStorageOps` concrete classes are the right abstraction — they just shouldn't know about cache state.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConventionPOStorageOps.java
```diff
@@ -0,0 +1,195 @@
...
+ * the SQL still receives storage-form values.
+ *
+ * @param <PO> persistent object type
+ * @param <Mapper> MyBatis mapper type
+ */
+public class HierarchicalConventionPOStorageOps<PO, Mapper> extends BasePOStorageOps<PO, Mapper> {
+
+  private final BasePOStorageOps<PO, Mapper> delegate;
+  private final UnaryOperator<PO> readRewriter;
+  private final UnaryOperator<PO> writeRewriter;
```
jerryshao: These two variable names are confusing. What is the meaning of `writeWriter`? You'd better pick a better name.

roryqi: OK, I will rename them to `LogicalToPhyiscalRewriter` and `PhysicalToLogicalRewriter`.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConventionPOStorageOps.java
```diff
@@ -0,0 +1,195 @@
...
+      return name;
+    }
+    String sep = HierarchicalSchemaUtil.schemaSeparator();
+    if (!name.contains(sep)) {
+      return name;
+    }
+    return HierarchicalSchemaUtil.logicalToPhysical(name, sep);
+  }
+
+  private static NameIdentifier apiIdentifierToStorage(NameIdentifier apiIdentifier) {
```
jerryshao: What is `apiIdentifier`, can you figure out a better name ?

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -327,6 +327,26 @@ public void insertRelation(
...
+    if (srcIdentifiers == null || srcIdentifiers.isEmpty()) {
+      return;
+    }
+    backend.batchInsertRelations(
+        relType, srcIdentifiers, srcType, dstIdentifier, dstType, override);
+    for (NameIdentifier ident : srcIdentifiers) {
+      cache.invalidate(ident, srcType, relType);
+    }
+    cache.invalidate(dstIdentifier, dstType, relType);
+  }
```
jerryshao, line 348: Why do we need to add this?

roryqi:
> The overall direction here is good — extracting per-entity PO logic into dedicated classes and using a decorator for hierarchical name translation is a clean separation. A few observations on the design that might be worth considering.
> 
> **The core tension**
> 
> `BasePOStorageOps.getPO/listPOs` currently combines two orthogonal concerns:
> 
> 1. **Name translation** (hierarchical logical ↔ physical) — handled well by `HierarchicalConventionPOStorageOps`
> 2. **Query routing** (cache-enabled → ID-based; no cache → full-name JOIN) — currently embedded via `GravitinoEnv.getInstance().cacheEnabled()` in `BasePOStorageOps`
> 
> Mixing these in the base class causes the `Capability` enum + runtime `UnsupportedOperationException` pattern, the `GravitinoEnv` static dependency leaking into a storage layer class, and the issue in `MetadataObjectService.TYPE_TO_STORAGE_OPS_MAP` where raw `*POStorageOps` instances bypass the naming convention.
> 
> **An alternative: expose two explicit SQL methods, route in the service**
> 
> ```java
> // BasePOStorageOps: pure SQL delegation, no routing logic
> abstract class BasePOStorageOps<PO, Mapper> {
>     // ID-based (cache-enabled path)
>     public PO getPOByParentId(Mapper mapper, Long parentId, String name) { ... }
>     public List<PO> listPOsByParentId(Mapper mapper, Long parentId) { ... }
> 
>     // Name-based (full join, no-cache path)
>     public PO getPOByFullName(Mapper mapper, NameIdentifier ident) { ... }
>     public List<PO> listPOsByNSFullName(Mapper mapper, Namespace ns) { ... }
> }
> ```
> 
> Routing moves back to the service as a single private helper per operation — not repeated per-method:
> 
> ```java
> // SchemaMetaService
> private SchemaPO getSchemaPO(NameIdentifier ident) {
>     return SessionUtils.getWithoutCommit(SchemaMetaMapper.class, mapper ->
>         cacheEnabled()
>             ? ops.getPOByParentId(mapper, getCatalogId(ident), ident.name())
>             : ops.getPOByFullName(mapper, ident));
> }
> ```
> 
> Since `ops` is still a `HierarchicalConventionPOStorageOps`, both paths get name translation automatically. `MetadataObjectService` can store wrapped instances in its map instead of raw ops, fixing the naming convention bypass.
> 
> This eliminates the `Capability` enum, the `GravitinoEnv` dependency in the base class, and the dispatcher `getPO(Mapper, NameIdentifier)`. Each layer has one job: name translation in the decorator, SQL in the ops, routing in the service.
> 
> **What stays the same**
> 
> `HierarchicalConventionPOStorageOps` as a decorator and the `*POStorageOps` concrete classes are the right abstraction — they just shouldn't know about cache state.

I got your point. 
I will try solve this issue. 
I want to extract a pattern that resolve the duplicated code for cached operations. Maybe I should extract a helper class for them.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: I will set owners for multiple metadata objects. So I need to insert multiple relations. Although we already have the update multiple relations, it can't cover the cases when we need to insert multiple records.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConventionPOStorageOps.java
roryqi: Maybe I should use the keywords `physical` and `logical`, too. Thanks for input.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -327,6 +327,26 @@ public void insertRelation(
...
+      throws IOException {
+    if (srcIdentifiers == null || srcIdentifiers.isEmpty()) {
+      return;
+    }
+    backend.batchInsertRelations(
+        relType, srcIdentifiers, srcType, dstIdentifier, dstType, override);
+    for (NameIdentifier ident : srcIdentifiers) {
+      cache.invalidate(ident, srcType, relType);
+    }
+    cache.invalidate(dstIdentifier, dstType, relType);
```
yuqi1129, line 347: Is it more proper to invalidate the cache first and then operate the database?

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaPOStorageOps.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public SchemaPO getPO(SchemaMetaMapper mapper, Long parentId, String name) {
+    return mapper.selectSchemaMetaByCatalogIdAndName(parentId, name);
+  }
+
+  @Override
+  public SchemaPO getPOByFullName(SchemaMetaMapper mapper, NameIdentifier identifier) {
+    Namespace namespace = identifier.namespace();
```
yuqi1129, line 55: There will be performance degradation if you are using the full qualified name when cache is enabled, we can utilize the entity ID to retrieve entities.

core/src/main/java/org/apache/gravitino/storage/relational/service/BasePOStorageOps.java
```diff
@@ -0,0 +1,81 @@
...
+    throw new UnsupportedOperationException(
+        "listPOs by parent id is not supported by " + getClass().getSimpleName());
+  }
+
+  public List<PO> listPOs(Mapper mapper, Namespace namespace, List<String> names) {
+    throw new UnsupportedOperationException(
+        "listPOs by namespace and names is not supported by " + getClass().getSimpleName());
+  }
+
+  public List<PO> listPOs(Mapper mapper, List<Long> uuids) {
```
yuqi1129: uuids ?

core/src/main/java/org/apache/gravitino/storage/relational/service/MetadataObjectService.java
```diff
@@ -96,6 +96,14 @@ public class MetadataObjectService {
                   MetadataObjectService::getJobTemplateObjectsFullName)
               .build();
 
+  static final Map<MetadataObject.Type, BasePOStorageOps<?, ?>> TYPE_TO_STORAGE_OPS_MAP =
+      ImmutableMap.<MetadataObject.Type, BasePOStorageOps<?, ?>>builder()
+          .put(MetadataObject.Type.SCHEMA, new SchemaPOStorageOps())
```
yuqi1129: What about other types?

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: I take a look. Other relation ops follows this pattern, too. Do we need to modify them together?

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaPOStorageOps.java
roryqi: I have added `POStorageReadRouting` to handle this issue.

core/src/main/java/org/apache/gravitino/storage/relational/service/BasePOStorageOps.java
roryqi: Yes, I should use entityIds. It may be readable.

core/src/main/java/org/apache/gravitino/storage/relational/service/MetadataObjectService.java
roryqi: The StorageOps only cover the schema related POs.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
yuqi1129: @diqiu50 
Please take a look, and I recall that you have modified this point.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaPOStorageOps.java
```diff
@@ -0,0 +1,117 @@
...
+  public void batchInsertPOs(SchemaMetaMapper mapper, List<SchemaPO> schemaPOs, boolean overwrite) {
+    if (overwrite) {
+      mapper.batchInsertSchemaMetaOnDuplicateKeyUpdate(schemaPOs);
+    } else {
+      mapper.batchInsertSchemaMeta(schemaPOs);
+    }
+  }
+
+  @Override
+  public Integer updatePO(SchemaMetaMapper mapper, SchemaPO oldPO, SchemaPO newPO) {
```
yuqi1129: `BasePOStorageOps.updatePO` defines the argument order as `(newPO, oldPO)`, and `HierarchicalConversionPOStorageOps.updatePO` preserves that order when delegating. This override reverses the semantic names and calls `mapper.updateSchemaMeta(oldPO, newPO)`, but `SchemaMetaMapper.updateSchemaMeta` expects `newSchemaMeta` first and `oldSchemaMeta` second.

That makes schema updates use the old row as the SET values and the new row as the WHERE predicate, so a rename/property update can silently update zero rows or fail incorrectly. Please swap the parameters here to `updatePO(..., SchemaPO newPO, SchemaPO oldPO)` and call `mapper.updateSchemaMeta(newPO, oldPO)`, and add a schema update/rename test that catches this.

core/src/main/java/org/apache/gravitino/storage/relational/service/MetadataObjectService.java
yuqi1129: These storage ops are not wrapped by `HierarchicalConversionPOStorageOps`, so `getSchemaObjectsFullName` reads `schema_meta.schema_name` in physical form. For a nested schema stored as `ns_a\u0001ns_b`, this method returns `catalog.ns_a\u0001ns_b` instead of the logical/API name `catalog.ns_a:ns_b`; table/view/function full names inherit the same physical schema segment through `getSchemaObjectsFullName`.

This leaks the internal separator into metadata-object names used by owner/tag/policy/auth paths. Please apply the same physical-to-logical conversion here, or use wrapped ops for SCHEMA/TABLE/VIEW/FUNCTION in `TYPE_TO_STORAGE_OPS_MAP`, and add coverage for `MetadataObjectService.getSchemaObjectsFullName` / table full names with nested schemas.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -230,11 +198,7 @@ public void insertSchema(SchemaEntity schemaEntity, boolean overwrite) throws IO
             SchemaPO leafPO = POConverters.initializeSchemaPOWithVersion(leafRow, leafBuilder);
```
yuqi1129, line 198: This PR now creates separate ancestor schema rows for nested schemas, but `deleteSchema` still only checks and deletes objects whose `schema_id` equals the selected row. Deleting `ns_a` therefore does not consider descendant schema rows such as `ns_a:ns_b` / `ns_a:ns_b:leaf`, because those are separate rows under the same catalog with different schema ids.

As a result, non-cascade delete can remove an ancestor while child schemas still exist, and cascade delete removes only the ancestor's own objects, leaving descendant schema rows and their tables/filesets/etc. behind. The delete path needs to include descendant schemas by logical/physical prefix, or reject deleting a schema that has nested children unless cascade handles the whole subtree. Please add tests for deleting `ns_a` when `ns_a:ns_b` exists, both cascade=false and cascade=true.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaPOStorageOps.java
roryqi: OK, I will fix this issue.

core/src/main/java/org/apache/gravitino/storage/relational/service/MetadataObjectService.java
roryqi: OK, I will fix this issue.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
roryqi: OK, I will fix this issue.

core/src/main/java/org/apache/gravitino/storage/relational/service/BasePOStorageOps.java
```diff
@@ -0,0 +1,81 @@
...
+    throw new UnsupportedOperationException(
+        "getPOByFullName is not supported by " + getClass().getSimpleName());
+  }
+
+  public List<PO> listPOsByNSFullName(Mapper mapper, Namespace namespace) {
+    throw new UnsupportedOperationException(
+        "listPOsByNSFullName is not supported by " + getClass().getSimpleName());
+  }
+
+  protected abstract Entity.EntityType entityType();
```
jerryshao: Where do we use this `entityType()` method?

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConversionPOStorageOps.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.Entity;
+import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.Namespace;
+import org.apache.gravitino.utils.HierarchicalSchemaUtil;
+
+/**
+ * Wraps a {@link BasePOStorageOps} to bridge the hierarchical schema naming convention. Identifiers
```
jerryshao: Conversion?

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConversionPOStorageOps.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    return pos.stream().map(this::applyRead).collect(Collectors.toList());
+  }
+
+  private List<PO> applyWrite(List<PO> pos) {
+    if (pos == null || pos.isEmpty()) {
+      return pos;
+    }
+    return pos.stream().map(logicalToPhysicalRewriter).collect(Collectors.toList());
+  }
```
jerryshao: I would suggest to unify the term, you have read/write, physical/logic two expression semantics, why read should bind to physicalToLogical, and write should bind to logicalToPhysical? This is implicitly defined by you. You should only keep one semantic for all.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConversionPOStorageOps.java
roryqi: OK, I feel we can use naming convention here. So I don't modify this at first. You feel that it is not good. I can change this point.

core/src/main/java/org/apache/gravitino/storage/relational/service/BasePOStorageOps.java
roryqi: I can remove it now. The related logic is extracted to other classes now.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConversionPOStorageOps.java
roryqi: OK, I will unify them.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
roryqi: If the schema has the child schema, we shouldn't support it. I should throw an exception direclty.
I want to optmize this point in the next pull request. I don't want to make this pull request bigger.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConversionPOStorageOps.java
```diff
@@ -0,0 +1,197 @@
...
+    return convertPhysicalToLogical(delegate.listPOs(mapper, parentId));
+  }
+
+  @Override
+  public List<PO> listPOs(Mapper mapper, Namespace logicalNamespace, List<String> names) {
+    Namespace physicalNamespace = logicalToPhysicalNamespace(logicalNamespace);
+    List<String> physicalNames =
+        names.stream()
+            .map(HierarchicalConversionPOStorageOps::toPhysicalIfHierarchical)
+            .collect(Collectors.toList());
```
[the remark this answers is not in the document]
roryqi: Physical separator won't occur in the table,view,function name. It isn't visible character.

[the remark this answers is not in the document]
roryqi: This case won't occur.

core/src/main/java/org/apache/gravitino/storage/relational/service/MetadataObjectService.java
```diff
@@ -96,6 +97,14 @@ public class MetadataObjectService {
                   MetadataObjectService::getJobTemplateObjectsFullName)
               .build();
 
+  static final Map<MetadataObject.Type, BasePOStorageOps<?, ?>> TYPE_TO_STORAGE_OPS_MAP =
+      ImmutableMap.<MetadataObject.Type, BasePOStorageOps<?, ?>>builder()
```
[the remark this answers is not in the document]
roryqi: I used another way.

core/src/main/java/org/apache/gravitino/storage/relational/service/HierarchicalConversionPOStorageOps.java
```diff
@@ -0,0 +1,197 @@
...
+  @Override
+  public Integer updatePO(Mapper mapper, PO newPO, PO oldPO) {
+    return delegate.updatePO(
+        mapper, convertLogicalToPhysical(newPO), convertLogicalToPhysical(oldPO));
+  }
+
+  @Override
+  public PO getPO(Mapper mapper, Long parentId, String name) {
+    return convertPhysicalToLogical(
+        delegate.getPO(mapper, parentId, toPhysicalIfHierarchical(name)));
```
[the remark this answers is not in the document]
roryqi: This won't occur.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: If required, I can modify it in the next pull request.

roryqi:
@yuqi1129 Could u take a look again?

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
```diff
@@ -623,4 +522,39 @@ private static long nextIdForNestedAncestor() {
...
+    String name = po.getSchemaName();
+    if (name == null || !name.contains(HierarchicalSchemaUtil.schemaSeparator())) {
+      return po;
+    }
+    return copySchemaPOWithName(
+        po,
+        HierarchicalSchemaUtil.logicalToPhysical(name, HierarchicalSchemaUtil.schemaSeparator()));
+  }
+
+  private static SchemaPO copySchemaPOWithName(SchemaPO po, String name) {
```
diqiu50, line 546: Why don't we changed po instead of copy it

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
diqiu50: Cache invalidation after the DB operation is appropriate.

yuqi1129 (commented):
I have no more comments here.

core/src/main/java/org/apache/gravitino/storage/relational/service/SchemaMetaService.java
roryqi: We don't provide the interface to modify the PO.

roryqi:
@jerryshao Could u take an another look for this pull request?

jerryshao (approved):

PR #11102, opened 2026-05-14
[#11101] feat(iceberg-rest-server): Add health check endpoints to the Iceberg REST Catalog server

### LINKED ISSUE #11101: [Improvement] Add health check endpoints to the Iceberg REST Catalog server

The main Gravitino server exposes MicroProfile-style health endpoints (`/api/health`, `/api/health/live`, `/api/health/ready`) that Kubernetes probes, load balancers, and global traffic managers (GTMs) can use to monitor server availability. The Iceberg REST Catalog (IRC) server, which runs on its own Jetty instance (port 9001), has no equivalent — operators running IRC in standalone or auxiliary mode have no standard way to probe its health.

### How should we improve?

Add the following endpoints to the IRC server:

- `GET /iceberg/health/live` — liveness probe; returns 200 as long as the HTTP thread can respond
- `GET /iceberg/health/ready` — readiness probe; returns 200 when `IcebergCatalogWrapperManager` is initialized, 503 otherwise
- `GET /iceberg/health` — aggregate; returns 200 only when both liveness and readiness pass

Register GTM root-level aliases (`/health/*`, `/health.html`) that forward to `/iceberg/health*`, so enterprise GTMs that require probes at well-known root paths work without configuration changes.

All health endpoints bypass authentication so infrastructure probes work without credentials.

jerryshao:

Adds MicroProfile-style health check endpoints to the Iceberg REST Catalog (IRC) server, mirroring the existing endpoints on the main Gravitino server:

- `GET /iceberg/health/live` — liveness probe; always 200 while the HTTP thread is responsive
- `GET /iceberg/health/ready` — readiness probe; 200 when `IcebergCatalogWrapperManager` is initialized, 503 otherwise
- `GET /iceberg/health` — aggregate; 200 only when both pass

Root-level GTM aliases (`/health/*`, `/health.html`) are registered unconditionally (standalone and auxiliary mode) and forward to `/iceberg/health*`.

All health paths bypass `IcebergAuthenticationFilter` so infrastructure probes work without credentials.

`HealthAliasServlet` is moved from `:server` to `:server-common` so it can be shared by both modules. `AuthenticationFilter.isHealthCheckRequest` is widened from `private` to `protected` to allow the override in `IcebergAuthenticationFilter`.

The IRC server runs on its own Jetty instance and has no health endpoints, making it impossible for Kubernetes probes, load balancers, and enterprise GTMs to monitor its availability.

Fix: #11101

Yes — three new REST endpoints are available on the IRC server: `GET /iceberg/health`, `GET /iceberg/health/live`, `GET /iceberg/health/ready`, plus root-level GTM aliases `/health`, `/health/live`, `/health/ready`, `/health.html`.

- `TestIcebergHealthOperations` — 5 unit tests covering all UP/DOWN combinations for liveness, readiness, and aggregate endpoints
- `TestIcebergAuthenticationFilter` — 2 new parametrized tests verifying `/iceberg/health*` paths bypass auth and non-health paths do not
- `TestHealthAliasServlet` — extended with custom-prefix cases covering `/iceberg` forwarding

### DIFF
--- a/clients/client-python/gravitino/dto/requests/owner_set_request.py
+++ b/clients/client-python/gravitino/dto/requests/owner_set_request.py
@@ -38,8 +38,6 @@ class OwnerSetRequest(RESTRequest):
         self._type = owner_type
 
     def validate(self) -> None:
-        Precondition.check_string_not_empty(
-            self._name, '"name" field is required'
-        )
+        Precondition.check_string_not_empty(self._name, '"name" field is required')
         if self._type is None:
             raise IllegalArgumentException('"type" field is required')
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
@@ -21,6 +21,8 @@ package org.apache.gravitino.iceberg.service;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import javax.servlet.ServletRequest;
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.server.authentication.AuthenticationFilter;
@@ -39,6 +41,21 @@ import org.eclipse.jetty.http.HttpStatus;
 public class IcebergAuthenticationFilter extends AuthenticationFilter {
 
   private static final ObjectMapper MAPPER = IcebergObjectMapper.getInstance();
+
+  @Override
+  protected boolean isHealthCheckRequest(ServletRequest request) {
+    if (super.isHealthCheckRequest(request)) {
+      return true;
+    }
+    if (!(request instanceof HttpServletRequest)) {
+      return false;
+    }
+    String path = ((HttpServletRequest) request).getRequestURI();
+    if (path == null) {
+      return false;
+    }
+    return path.equals("/iceberg/health") || path.startsWith("/iceberg/health/");
+  }
 
   @Override
   protected void sendAuthErrorResponse(HttpServletResponse response, Exception exception)
--- a/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
+++ b/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
@@ -47,13 +47,8 @@ public class IcebergAuthenticationFilter extends AuthenticationFilter {
     if (super.isHealthCheckRequest(request)) {
       return true;
     }
-    if (!(request instanceof HttpServletRequest)) {
-      return false;
-    }
+
     String path = ((HttpServletRequest) request).getRequestURI();
-    if (path == null) {
-      return false;
-    }
     return path.equals("/iceberg/health") || path.startsWith("/iceberg/health/");
   }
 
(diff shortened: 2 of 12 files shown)


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergHealthOperations.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return up(CHECK_CATALOG_WRAPPER_MANAGER, Collections.emptyMap());
+  }
+
+  /** Visible for testing — subclasses override to inject a different manager instance. */
+  IcebergCatalogWrapperManager getCatalogWrapperManager() {
+    return catalogWrapperManager;
+  }
+
+  private static HealthCheckDTO up(String name, java.util.Map<String, String> details) {
```
xxubai: nit: 
```suggestion
  private static HealthCheckDTO up(String name, Map<String, String> details) {
```

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
```diff
@@ -40,6 +42,21 @@ public class IcebergAuthenticationFilter extends AuthenticationFilter {
 
   private static final ObjectMapper MAPPER = IcebergObjectMapper.getInstance();
 
+  @Override
+  protected boolean isHealthCheckRequest(ServletRequest request) {
+    if (super.isHealthCheckRequest(request)) {
+      return true;
+    }
+    if (!(request instanceof HttpServletRequest)) {
```
xxubai: Already checked in `super.isHealthCheckRequest`

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
```diff
@@ -40,6 +42,16 @@ public class IcebergAuthenticationFilter extends AuthenticationFilter {
...
   private static final ObjectMapper MAPPER = IcebergObjectMapper.getInstance();
 
+  @Override
+  protected boolean isHealthCheckRequest(ServletRequest request) {
+    if (super.isHealthCheckRequest(request)) {
+      return true;
+    }
+
+    String path = ((HttpServletRequest) request).getRequestURI();
+    return path.equals("/iceberg/health") || path.startsWith("/iceberg/health/");
```
roryqi, line 52: Is it clear use `path.startsWith("/iceberg/health/"` directly. `path.equals("/iceberg/health")` is subset of `path.startsWith("/iceberg/health/")`.

roryqi (commented):
Just a minor.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
jerryshao: No, it is not a subset. The latter is "/iceberg/health/" that has a "/" in the end. So these two expressions cannot be merged into one.

If we simplify to `path.startsWith("/iceberg/health")`, it will match to `/iceberg/health-foo`. So, I think the current check is correct.

xxubai (approved):
LGTM

clients/client-python/gravitino/dto/requests/owner_set_request.py
```diff
@@ -38,8 +38,6 @@ def __init__(self, name: str, owner_type: Owner.Type):
         self._type = owner_type
 
     def validate(self) -> None:
-        Precondition.check_string_not_empty(
-            self._name, '"name" field is required'
-        )
+        Precondition.check_string_not_empty(self._name, '"name" field is required')
```
yuqi1129, line 41: Why do we change the format here?

jerryshao: It is a style issue that wasn't fixed in the previous issue.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergAuthenticationFilter.java
roryqi: OK. got it. Thanks.

roryqi (approved):
LGTM.

PR #10996, opened 2026-05-08
[#10772] feat(authorization): refactor JcasbinAuthorizer with version-validated cache

yuqi1129:
### What changes were proposed in this pull request?

> **Prerequisite:** #11117 (eventual-consistency invalidation infrastructure) must land first. This PR will be rebased on `main` after that, at which point the diff narrows to the role-loading rewrite described below.

This is the final piece of the #10812 split. It rewrites `JcasbinAuthorizer`'s strong-consistency caching layer on top of the lookups + change-poller infrastructure introduced in #11117:

- **Per-request dedup** via `AuthorizationRequestContext`: user identity, group identity and the role-load step are deduplicated within a single HTTP request.
- **Version-validated shared caches** (`userRoleCache`, `groupRoleCache`, `loadedRoles`): cache hits revalidate against `user_meta.updated_at` / `group_meta.updated_at` / `role_meta.updated_at` before being trusted, replacing TTL-only invalidation.
- **`JcasbinRoleLoader`** (new): owns the 4-step role load — version-checked user-direct roles, version-checked group-inherited roles, prune stale g-rows, then batch version-check + reload of `role_meta`.
- **`JcasbinLoadedRolesCache`** (new): wraps a `GravitinoCache` with a removal listener that synchronously deletes the role's JCasbin policies, so eviction and policy state stay in sync.
- Removes the `Executor` + `CompletableFuture` async role-load path; role loading is now synchronous via `AuthorizationRequestContext.loadRole(Runnable)`.
- `isMetalakeUser` now takes `AuthorizationRequestContext` so the per-request `UserUpdatedAt` cache populated by `authorize`/`isOwner` is reused, removing a redundant DB round-trip.

### Why are the changes needed?

Fix: #10772

### Does this PR introduce _any_ user-facing change?

No. New configs (`gravitino.authorization.jcasbin.*`) have safe defaults.

### How was this patch tested?

- 12 new unit tests across `TestGravitinoCache` (TTL, max-size eviction, prefix edge cases) and `TestAuthorizationRequestContext` (request-scoped dedup for user/owner/metadata-id, including absent-result caching).
- New `TestJcasbinAuthorizer` group-role / version-validated cases: group-inherited role, IdP group removal, multi-group partial revocation, deny-wins over group-inherited allow, role-shared-by-user-and-group survives single-side revocation.
- All existing `TestJcasbinAuthorizer`, `TestJcasbinAuthorizerCacheHelpers` and `TestPassThroughAuthorizer` suites still pass.

PR #11117, opened 2026-05-15
[#10772] feat(authz): Eventual-consistency invalidation for JcasbinAuthorizer caches

### LINKED ISSUE #10772: [Subtask] Refactor cache in JcasbinAuthorizer

yuqi1129:

Introduces the HA invalidation infrastructure that the JcasbinAuthorizer cache refactor (#10772) needs as a prerequisite. The existing role-loading path stays on its current TTL-only behavior; the version-validated user / group / role caches and the JcasbinRoleLoader ride on top of this PR in a follow-up.

**New classes:**

- JcasbinAuthorizationLookups: two-tier metadata-id and owner lookup facade (per-request dedup via AuthorizationRequestContext, shared Caffeine-backed GravitinoCache fallback, DB on miss). Owners are now fetched directly from owner_meta via OwnerMetaMapper instead of the OWNER_REL relation query, so the cache key space matches the entity-change-log key space.
- JcasbinChangePoller: scheduled thread that drains entity_change_log and owner_meta change rows since a high-water cursor and invalidates the affected cache keys. Documents the id-cursor in-flight-commit trade-off and the writer-side pre-mutation-name contract.

**Modified:**

- JcasbinAuthorizer: wires the two GravitinoCaches + lookups + poller in initialize/close; routes isOwner, OWNER-privilege authorization, and handleMetadataOwnerChange through the lookups; drops the in-class OwnerInfo (replaced by org.apache.gravitino.storage.relational.po.auth.OwnerInfo) and collapses loadOwnerPolicy + checkOwnership into ownerMatchesUserOrGroups; implements handleEntityNameIdMappingChange.
- GravitinoAuthorizer: adds the handleEntityNameIdMappingChange default method.
- Configs: adds metadataIdCacheSize and changePollIntervalSecs.

This is a prerequisite split of #10996 (tracking issue #10772) to keep that follow-up PR scoped to the role-loading rewrite alone.

Fix: #10772

No. New configs ship with safe defaults and existing behavior is preserved for the role-loading path.

- TestJcasbinAuthorizer: existing cases (owner mocks adjusted to the new OwnerMetaMapper.selectOwnerByMetadataObjectIdAndType path).
- TestJcasbinAuthorizationLookups: buildCacheKey / isContainerType / cascade-prefix hierarchy.
- TestJcasbinChangePoller: metadataObjectFromChangeLog builds cache keys matching the writer-side full-name format.

```
./gradlew :server-common:test --tests org.apache.gravitino.server.authorization.jcasbin.TestJcasbinChangePoller --tests org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer
./gradlew :core:test --tests org.apache.gravitino.authorization.TestAuthorizationUtils
```


server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
```diff
@@ -0,0 +1,233 @@
...
+      }
+
+      MetadataObject mdObj = metadataObjectFromChangeLog(metalake, fullName, mdType);
+      String cacheKey = JcasbinAuthorizationLookups.buildCacheKey(metalake, mdObj);
+
+      if (JcasbinAuthorizationLookups.isNonLeaf(mdType)) {
+        metadataIdCache.invalidateByPrefix(cacheKey);
+      } else {
+        metadataIdCache.invalidate(cacheKey);
+      }
```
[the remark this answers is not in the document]
yuqi1129: This is a corner performance improvment, and can be postponed.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizationLookups.java
```diff
@@ -0,0 +1,157 @@
...
+ * request and later ones — hit the cache. The two underlying caches are invalidated externally by
+ * {@link JcasbinChangePoller} (HA peers) and by the {@link
+ * org.apache.gravitino.authorization.GravitinoAuthorizer#handleMetadataOwnerChange} / {@link
+ * org.apache.gravitino.authorization.GravitinoAuthorizer#handleEntityNameIdMappingChange} hooks
+ * (local mutations).
+ */
+public class JcasbinAuthorizationLookups {
+
+  /** Key separator for path-based cache keys. */
+  static final String KEY_SEP = "::";
```
roryqi: Why do we need KEY_SEP? Could we use a  nameIdentifier?

roryqi: "::" could be used for hierachical schema separator.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
```diff
@@ -0,0 +1,233 @@
...
+   * TTL eviction. Acceptable for the eventual-consistency caches targeted here; revisit if we ever
+   * route strong-consistency data through this poller.
+   */
+  public void start() {
+    ownerPollHighWaterId =
+        nullToZero(
+            SessionUtils.getWithoutCommit(
+                OwnerMetaMapper.class, OwnerMetaMapper::selectMaxChangeId));
+    entityPollHighWaterId =
+        nullToZero(
```
roryqi: Could u use `getDefault` here?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizationLookups.java
jerryshao: Can we unify this separator with the one introduced by @roryqi 's PR for internal use, rather than define a new one? My feeling is that there're different separators for different purposes, it is difficult to maintain.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizationLookups.java
```diff
@@ -0,0 +1,157 @@
...
+        cacheKey,
+        k -> {
+          Optional<Long> cached = metadataIdCache.getIfPresent(k);
+          if (cached.isPresent()) {
+            return cached.get();
+          }
+          Long id = MetadataIdConverter.getID(metadataObject, metalake);
+          metadataIdCache.put(k, id);
+          return id;
+        });
```
jerryshao: Is this function call atomic? It is be atomic, otherwise it will have concurrency issue.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizationLookups.java
```diff
@@ -0,0 +1,157 @@
...
+            return cached.get();
+          }
+          OwnerInfo ownerInfo =
+              SessionUtils.getWithoutCommit(
+                  OwnerMetaMapper.class,
+                  m -> m.selectOwnerByMetadataObjectIdAndType(id, metadataType.name()));
+          Optional<OwnerInfo> owner = ownerInfo == null ? Optional.empty() : Optional.of(ownerInfo);
+          ownerRelCache.put(id, owner);
+          return owner;
+        });
```
jerryshao: Also here.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizationLookups.java
yuqi1129: I see.

core/src/main/java/org/apache/gravitino/utils/HierarchicalSchemaUtil.java
```diff
@@ -41,8 +41,8 @@
  */
 public final class HierarchicalSchemaUtil {
 
-  /** The internal separator used in EntityStore for HierarchicalSchema names. */
-  private static final String PHYSICAL_SEPARATOR = "\u0001";
+  /** The internal separator used in EntityStore and internal path-based keys. */
+  private static final String INTERNAL_SEPARATOR = "\u0001";
```
roryqi: It may not be suitable if you want to use this. This class can only defines HierarchicalSchema separator.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizationLookups.java
yuqi1129: I have changed it.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
yuqi1129: modified.
/

core/src/main/java/org/apache/gravitino/utils/HierarchicalSchemaUtil.java
yuqi1129: Okay, let me define a new value for that.

core/src/main/java/org/apache/gravitino/authorization/GravitinoAuthorizer.java
```diff
@@ -166,4 +166,17 @@ default void handleRolePrivilegeChange(String metalake, String roleName) {
...
+   * Implementations evict the cache key for the given entity and all of its descendants.
+   *
+   * @param metalake the metalake name
+   * @param nameIdentifier the entity name identifier
+   * @param type the entity type
+   */
+  default void handleEntityNameIdMappingChange(
+      String metalake, NameIdentifier nameIdentifier, Entity.EntityType type) {
+    // default no-op for backward compatibility
+  }
```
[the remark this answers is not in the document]
jerryshao: Did you fix all the copilot comments here?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
```diff
@@ -0,0 +1,261 @@
...
+            OwnerMetaMapper.class, m -> m.selectChangedOwners(ownerPollHighWaterId));
+
+    long maxSeenId = ownerPollHighWaterId;
+    for (ChangedOwnerInfo change : changes) {
+      ownerRelCache.invalidate(change.getMetadataObjectId());
+      if (change.getId() > maxSeenId) {
+        maxSeenId = change.getId();
+      }
+    }
+    ownerPollHighWaterId = maxSeenId;
```
jerryshao: Should this code block be done in a mutex to guarantee the atomicity?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
```diff
@@ -0,0 +1,261 @@
...
+  private void invalidateCoalescedKeys(Set<String> prefixes, Set<String> leafKeys) {
+    for (String prefix : prefixes) {
+      metadataIdCache.invalidateByPrefix(prefix);
+    }
+    for (String leafKey : leafKeys) {
+      if (prefixes.stream().noneMatch(leafKey::startsWith)) {
+        metadataIdCache.invalidate(leafKey);
+      }
+    }
+  }
```
jerryshao, line 319: Also here.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
yuqi1129: Modified.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
yuqi1129: Modified, and it will only be called within a `synchronized` block.

core/src/main/java/org/apache/gravitino/authorization/GravitinoAuthorizer.java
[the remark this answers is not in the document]
yuqi1129: Yes, all have been resolved.

core/src/main/java/org/apache/gravitino/authorization/AuthorizationUtils.java
```diff
@@ -325,6 +325,7 @@ public static void checkPrivilege(
 
   public static void authorizationPluginRemovePrivileges(
       NameIdentifier ident, Entity.EntityType type, List<String> locations) {
+    notifyEntityNameIdMappingChange(ident, type);
```
roryqi: This is related about authorization plugin. If plugin is null, it won't be called.

yuqi1129: Changed.

core/src/main/java/org/apache/gravitino/authorization/OwnerManager.java
```diff
@@ -134,44 +135,47 @@ public void setOwner(
...
+              .get(
+                  NameIdentifierUtil.ofGroup(metalake, owner.name()),
+                  Entity.EntityType.GROUP,
+                  GroupEntity.class);
+      return groupEntity.id();
+    } else {
+      throw new IllegalArgumentException("Unsupported owner type: " + owner.type());
+    }
+  }
+
```
jerryshao, line 199: What is the purpose of this change? I see it split into several methods, with no logic changes.

core/src/main/java/org/apache/gravitino/cache/GravitinoCache.java
```diff
@@ -38,6 +40,24 @@ public interface GravitinoCache<K, V> extends Closeable {
...
+   */
+  default V get(K key, Function<K, V> loader) {
+    Optional<V> value = getIfPresent(key);
+    if (value.isPresent()) {
+      return value.get();
+    }
+    V loaded = Objects.requireNonNull(loader.apply(key), "Cache loader must not return null");
+    put(key, loaded);
+    return loaded;
+  }
```
jerryshao, line 61: This method is not thread safe; should you guarantee it?

core/src/main/java/org/apache/gravitino/authorization/OwnerManager.java
yuqi1129: It just extracts a method and makes the method more readable. No change is introduced.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
jerryshao: Does adding `synchronized` really solve the problem? You're invalidating the `ownerRelCache` one by one, in the meantime, the authorizer will use this cache, so it will get the data that haven't yet been invalidated. Is this the problem?

You should answer me the questions, rather than tell me that it is fixed or not.

jerryshao: The lock should be on the cache, not on this class object.

yuqi1129: > Does adding synchronized really solve the problem? You're invalidating the ownerRelCache one by one, in the meantime, the authorizer will use this cache, so it will get the data that haven't yet been invalidated. Is this the problem?

I added a read/write lock to avoid this problem, and I could not come up with a better solution now.  The biggest risk is that it will degrade the performance to some extent. I will optimize it if it indeed affects too much.

core/src/main/java/org/apache/gravitino/cache/GravitinoCache.java
yuqi1129: Add comments to explain the possible problem.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
yuqi1129: I would like to add more details here:

There are actually two distinct races here, worth separating:

 **(1) Multi-node staleness window.** This is inherent to the eventual-consistency design and **no lock can fix it**. After a change is committed on node A, node B keeps returning stale data until B's poller runs (~poll interval, seconds). The "stale read while a batch is in progress" you describe is a small slice (~ms) of that larger window. It self-heals: after the batch, the affected keys are gone from the cache, so the next read pulls fresh data. If a caller needs read-after-write consistency, they must bypass the cache — nothing on node B can make node A's commit visible faster.

  **(2) Single-node read/invalidate race.** This is what `synchronized` alone *couldn't* address (good catch) and what the cache-level RW lock added in a follow-up commit now does. Without the cache lock, a reader and the poller's per-key `invalidate(k)` loop interleave, so a reader can observe a half-applied batch (some keys evicted, others still hot). The cache lock makes the batch atomic from a reader's POV.

Cost is small: the poller runs every few seconds with small batches, contention is rare, and a fair RW lock under no contention is on the order of nanoseconds. Given problem (1) dominates the staleness budget, dropping the lock would also be defensible. I lean toward keeping it — the atomicity guarantee makes the
cache layer's invalidation semantics are easy to reason about and to test, at a cost we won't measure on the hot path.

jerryshao (approved):

PR #11119, opened 2026-05-15
[#10983] test(lance): add Lance compatibility matrix automation

### LINKED ISSUE #10983: [Improvement] Upgrade LanceDB dependency in Gravitino

yuqi1129:

This PR is stacked on #11060 and should be merged after #11060.

It adds a separate Lance compatibility matrix validation path:
- Add `:lance:lance-rest-server:lanceSparkMatrixTest` to run `LanceSparkRESTServiceIT` against multiple `lance-spark-bundle` versions.
- Add `:clients:client-python:lanceRayMatrixTest` and `scripts/run_lance_ray_matrix.py` to run the Lance Ray integration test against multiple `lance-ray` versions with isolated per-version virtualenvs.
- Add a standalone `Lance Compatibility Matrix Test` GitHub Actions workflow, triggered by `workflow_dispatch` and weekly schedule only.
- Document how to reproduce the Lance Spark and Lance Ray matrix locally.

The default PR validation should keep using a single pinned, known-good Lance dependency combination so normal PR checks stay stable and reasonably fast.

The full compatibility matrix is still useful for catching Lance ecosystem compatibility drift, but it should run as a separate automation path instead of being part of every PR validation.

Fix: #10983

No user-facing API change.

This adds developer/CI validation tooling and documentation for Lance compatibility testing.

- `./gradlew :lance:lance-rest-server:spotlessApply :clients:client-python:pylint -PskipITs`
- `./gradlew :lance:lance-rest-server:test -PskipITs`
- `git diff --cached --check`

The full multi-version matrix was not run locally because it downloads and runs multiple external Lance dependency combinations.


clients/client-python/build.gradle.kts
```diff
@@ -239,6 +239,81 @@ tasks {
...
+      condaPython != null && condaPython.exists() -> condaPython.absolutePath
+      else -> "python3"
+    }
+    val script = projectDir.resolve("scripts/run_lance_ray_matrix.py")
+    val gravitinoHome = file("${project.rootDir}/distribution/package")
+
+    doFirst {
+      gravitinoServer("start")
+    }
+    doLast {
```
FANNG1, line 285: `gravitinoServer("start")` is placed in a `doFirst` block, but `gravitinoServer("stop")` only lives inside the `finally` of a `doLast` block. In Gradle, if a `doFirst` action throws (e.g. the server fails to start or the script is missing), Gradle aborts the task immediately and **never executes any `doLast` blocks** — the `finally` is unreachable. The server process would be orphaned for the rest of the CI session.

Consider moving the server lifecycle entirely into `doLast` so the `finally` can actually guard it:

```kotlin
doLast {
    gravitinoServer("start")
    try {
        val proc = ProcessBuilder(args).inheritIO().start()
        val exit = proc.waitFor()
        if (exit != 0) throw GradleException("lance-ray matrix failed with exit code $exit")
    } finally {
        gravitinoServer("stop")
    }
}
```

lance/lance-rest-server/build.gradle.kts
```diff
@@ -169,6 +200,64 @@ tasks {
...
+          layout.buildDirectory.dir("reports/lance-spark-matrix/$versionSlug")
+        )
+        junitXml.outputLocation.set(
+          layout.buildDirectory.dir("test-results/lance-spark-matrix/$versionSlug")
+        )
+      }
+    }
+  }
+
+  register("lanceSparkMatrixTest") {
```
FANNG1, line 255: `lanceSparkMatrixTest` depends on all per-version `Test` tasks via a plain `dependsOn` list with no `mustRunAfter` ordering between them. With `--parallel` (or `org.gradle.parallel=true` in `gradle.properties`), Gradle can dispatch multiple per-version tasks concurrently. Each `LanceSparkRESTServiceIT` spins up an embedded `MiniGravitino` instance. Since `findAvailablePort` is a scan rather than an atomic OS-level reservation (TOCTOU), two tasks can both observe the same port as free and then race to bind it, producing intermittent `address already in use` errors that look like test failures rather than infrastructure conflicts.

Consider serializing the tasks with `mustRunAfter`, or set `maxParallelForks = 1` on each per-version `Test` task.

clients/client-python/scripts/run_lance_ray_matrix.py
```diff
@@ -0,0 +1,329 @@
...
+    venv_python = venv_dir / "bin" / "python"
+    if not venv_python.exists():
+        venv_dir.parent.mkdir(parents=True, exist_ok=True)
+        rc = run([python, "-m", "venv", str(venv_dir)]).returncode
+        if rc != 0:
+            raise RuntimeError(f"Failed to create venv at {venv_dir}")
+    return venv_python
+
+
+def install_deps(
```
FANNG1, line 160: `ensure_venv` correctly skips venv creation when the venv already exists, but `install_deps` is called unconditionally after it every run regardless. All three `pip` invocations (upgrade pip/wheel, install lance-ray + ray + lance-namespace, install editable gravitino) re-run even when the venv is fully populated from a prior cached run — so the "cached across runs" claim in the docstring only holds for the venv skeleton, not the packages.

This also makes the matrix fragile in restricted network environments: a run that should succeed from cache will fail at the `pip install` step if PyPI is unavailable.

A simple fix is to write a sentinel file (e.g. `.installed-<version>`) inside the venv dir after a successful `install_deps`, and skip the function when that sentinel exists.

.github/workflows/lance-compatibility-matrix-test.yml
```diff
@@ -0,0 +1,96 @@
...
+  workflow_dispatch:
+    inputs:
+      lance_spark_versions:
+        description: "Comma-separated lance-spark-bundle versions"
+        required: false
+        default: "0.2.0,0.4.0"
+      lance_ray_versions:
+        description: "Comma-separated lance-ray versions"
+        required: false
+        default: "0.4.2,0.3.0"
```
FANNG1, line 13: The workflow default `"0.4.2,0.3.0"` only covers 2 versions, but `DEFAULT_VERSIONS` in `run_lance_ray_matrix.py` is `["0.4.2", "0.4.1", "0.4.0", "0.3.0"]` (4 versions).

When the Gradle task is invoked without `-PlanceRayVersions=`, it does **not** pass `--versions` to the script, so the script falls back to its own `DEFAULT_VERSIONS` (4 versions). The weekly CI workflow, however, pins only 2 — meaning 0.4.1 and 0.4.0 are **never exercised in CI**, only in local runs.

Either expand the CI default to `"0.4.2,0.4.1,0.4.0,0.3.0"` to match `DEFAULT_VERSIONS`, or align `DEFAULT_VERSIONS` in the script with the workflow default so both paths test the same set.

clients/client-python/scripts/run_lance_ray_matrix.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+def _deps_sentinel(venv_dir: Path, version: str, ray_spec: str, lance_namespace_spec: str) -> Path:
+    """Return the path to the sentinel file that marks a fully-installed venv.
+
+    The sentinel encodes all dep specs so any change in pinned versions forces
+    a reinstall, while an identical set of specs skips all pip invocations.
+    """
+    key = f"{version}|{ray_spec}|{lance_namespace_spec}"
+    digest = hashlib.sha1(key.encode()).hexdigest()[:12]
```
FANNG1: `hashlib.sha1()` without `usedforsecurity=False` raises `ValueError` on FIPS-enabled systems (RHEL 9, Ubuntu in FIPS mode, etc.). Python 3.9+ accepts the flag:

```python
digest = hashlib.sha1(key.encode(), usedforsecurity=False).hexdigest()[:12]
```

.github/workflows/lance-compatibility-matrix-test.yml
yuqi1129: Okay. I will fix it.

clients/client-python/scripts/run_lance_ray_matrix.py
yuqi1129: I accept the suggestion.

clients/client-python/scripts/run_lance_ray_matrix.py
```diff
@@ -0,0 +1,355 @@
...
+import os
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import List
+
+# Default version set tracks what the Compatibility Matrix in
+# docs/lance-rest-integration.md claims to support. Keep these in sync.
+DEFAULT_VERSIONS = ["0.4.2", "0.4.1", "0.4.0", "0.3.0"]
```
FANNG1: `DEFAULT_VERSIONS` includes all 4 patch versions (0.4.2, 0.4.1, 0.4.0, 0.3.0), but the docs only claim to explicitly verify `{0.3.0, 0.4.2}` — and `lance-spark` only tests the two boundary versions (0.2.0, 0.4.0).

Is it necessary to test every patch version? Patch releases shouldn't introduce breaking changes. If there are no known API differences between 0.4.0 → 0.4.1 → 0.4.2, testing only the boundary versions `["0.4.2", "0.3.0"]` would align with the docs and halve the CI runtime. If the intermediate versions do have distinct behaviors worth validating, the docs' "explicitly verified" list should also reflect that.

docs/lance-rest-integration.md
```diff
@@ -23,20 +23,66 @@ The following table outlines the tested compatibility between Gravitino versions
...
+- **`lance-spark` 0.1.0 and 0.1.1 are not supported on Gravitino 1.3.0.**
+  Those bundles create tables by calling the legacy
+  `POST /lance/v1/table/{id}/create-empty` endpoint, which 1.3.0 no longer
+  exposes — the table-declaration path was consolidated onto
+  `POST /lance/v1/table/{id}/declare` (`LanceTableOperations#declareTable`)
+  when the deprecated `createEmptyTable` API was removed during the
+  `lance-namespace-core` 0.7.5 upgrade. Running 0.1.x against 1.3.0 surfaces
+  as `404 Not Found` on every table-creation flow; the small set of
+  list/describe-only tests still works, but any write path will fail.
+- **`lance-ray` 0.2.0 and earlier are not supported on Gravitino 1.3.0.** They
```
FANNG1: The section header says "`lance-ray` 0.2.0 and earlier are **not supported**", but the bullet immediately ends with "You can still use lance-ray 0.2.0 with Gravitino 1.3.0 by pinning pylance to 3.x or 4.x" — which contradicts the header. The compatibility table also already lists 0.2.0 as "conditionally supported".

The two cases are fundamentally different and should be documented separately:
- **0.1.0**: truly not supported — the `write_lance()` API signature is incompatible, no workaround exists.
- **0.2.0**: conditionally supported — the issue is in `pylance` (not Gravitino), and pinning `pylance` to 3.x/4.x makes it work.

Grouping them under "not supported" is misleading. Consider splitting into two bullets to match the table's intent.

.github/workflows/lance-compatibility-matrix-test.yml
```diff
@@ -0,0 +1,96 @@
+name: Lance Compatibility Matrix Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      lance_spark_versions:
+        description: "Comma-separated lance-spark-bundle versions"
+        required: false
+        default: "0.2.0,0.4.0"
```
FANNG1, line 9: The default skips from 0.2.0 to 0.4.0 with no 0.3.0. The docs state the supported range as "0.2.0 - 0.4.0", which implies 0.3.0 should also be covered.

Two questions:
1. Does `lance-spark-bundle 0.3.0` actually exist on Maven Central? Looking at the version history in this PR (0.1.0, 0.1.1, 0.2.0, 0.4.0), 0.3.x has never appeared, which suggests it may not have been released.
2. If 0.3.0 was never released, the docs should use a discrete version list `{0.2.0, 0.4.0}` rather than the range `0.2.0 - 0.4.0`, since the range implies 0.3.0 is also covered.

FANNG1:
LGTM except minor comments

.github/workflows/lance-compatibility-matrix-test.yml
yuqi1129: lance-spark-bundle has never been released.

docs/lance-rest-integration.md
yuqi1129: Split the documentation into separate bullets: lance-ray 0.1.0 is unsupported due to the write_lance API mismatch, while 0.2.0 is conditionally supported when pylance is pinned to 3.x or 4.x.

clients/client-python/scripts/run_lance_ray_matrix.py
yuqi1129: Updated DEFAULT_VERSIONS to ["0.4.2", "0.3.0"] and aligned the local runner example with the workflow defaults.

.github/workflows/lance-compatibility-matrix-test.yml
yuqi1129: Confirmed that lance-spark-bundle 0.3.0 was not released. Updated the compatibility table to use the discrete set {0.2.0, 0.4.0} and removed the stale 0.3.0 Gradle example.

.github/workflows/lance-compatibility-matrix-test.yml
```diff
@@ -0,0 +1,96 @@
...
+      lance_ray_versions:
+        description: "Comma-separated lance-ray versions"
+        required: false
+        default: "0.4.2,0.4.1,0.4.0,0.3.0"
```
FANNG1: `DEFAULT_VERSIONS` in `run_lance_ray_matrix.py` was simplified to `["0.4.2", "0.3.0"]`, but the workflow default was not updated and still lists 4 versions (`"0.4.2,0.4.1,0.4.0,0.3.0"`). This creates a mismatch in the opposite direction from before: CI tests 4 versions, while a local `./gradlew :clients:client-python:lanceRayMatrixTest` (without `-PlanceRayVersions`) only tests 2.

Should this be `"0.4.2,0.3.0"` to match the new `DEFAULT_VERSIONS`?

yuqi1129: Yes. We need to change both workflow defaults to `0.4.2,0.3.0` so scheduled CI and local Gradle runs use the same matrix. I checked PyPI: `lance-ray` 0.4.0 and 0.4.1 were not published, while 0.4.2 and 0.3.0 are published releases. We only need one representative release per supported minor line here.

yuqi1129: done.

yuqi1129: done

FANNG1 (approved):

FANNG1:
LGTM

Issue #11068, opened 2026-05-13
[Subtask] Add built-in IdP group meta storage model


lasdf1234:
### Describe the subtask

Add the built-in IdP group metadata relational storage model under `plugins:idp-basic`, including the group persistent object, MyBatis mapper, SQL provider factory, backend-specific SQL providers, mapper registration updates, and corresponding tests.

Related PR: https://github.com/lasdf1234/gravitino/pull/6

### Parent issue

https://github.com/apache/gravitino/issues/10959

PR #11127, opened 2026-05-18
[#11068] feat(authn): Add built-in IdP group meta storage model

lasdf1234:

This PR adds the built-in IdP group metadata relational storage model under `plugins:idp-basic`.

It introduces:
- the `IdpGroupPO` persistent object for built-in IdP groups
- the `IdpGroupMetaMapper` MyBatis mapper for querying, inserting, soft deleting, and cleaning legacy group metadata
- the `IdpGroupMetaSQLProviderFactory` together with base, H2, and PostgreSQL SQL providers for group metadata
- mapper package registration updates so the built-in IdP group mapper is loaded by the `idp-basic` plugin
- tests covering the group PO, SQL providers, and storage behavior

This PR keeps the new group metadata storage implementation aligned with the existing built-in IdP user metadata code by:
- following the same PO, mapper, and SQL provider structure
- keeping MySQL in the base provider and using dedicated H2 and PostgreSQL providers for dialect differences
- using the shared storage test base and mapper-driven test setup already used by the user metadata tests

The built-in IdP needs a relational group metadata model so group records can be persisted and queried consistently across the supported JDBC backends.

Part of apache/gravitino#11068

No.

```bash
./gradlew --no-daemon :plugins:idp-basic:test -PskipITs -PskipDockerTests=false \
  --tests org.apache.gravitino.idp.storage.po.TestIdpGroupPO \
  --tests org.apache.gravitino.idp.storage.mapper.TestIdpBasicMapperPackageProvider \
  --tests org.apache.gravitino.idp.storage.mapper.TestIdpGroupMetaSQLProviderFactory \
  --tests org.apache.gravitino.idp.storage.mapper.provider.base.TestIdpGroupMetaBaseSQLProvider \
  --tests org.apache.gravitino.idp.storage.mapper.provider.h2.TestIdpGroupMetaH2Provider \
  --tests org.apache.gravitino.idp.storage.mapper.provider.postgresql.TestIdpGroupMetaPostgreSQLProvider \
  --tests org.apache.gravitino.idp.storage.mapper.TestIdpGroupMetaStorage
```

`TestIdpGroupMetaStorage` covers H2, MySQL, and PostgreSQL when `-PskipDockerTests=false` is set.


plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/provider/base/IdpGroupMetaBaseSQLProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public String deleteIdpGroupMetasByLegacyTimeline(
+      @Param("legacyTimeline") Long legacyTimeline, @Param("limit") int limit) {
+    return "DELETE FROM "
+        + IdpGroupMetaMapper.IDP_GROUP_TABLE_NAME
+        + " WHERE deleted_at > 0 AND deleted_at < #{legacyTimeline} LIMIT #{limit}";
+  }
+
+  protected String currentTimeMillisExpression() {
+    return "(UNIX_TIMESTAMP() * 1000.0) + EXTRACT(MICROSECOND FROM CURRENT_TIMESTAMP(3)) / 1000";
```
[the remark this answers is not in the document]
lasdf1234: Consistent with the original code

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/storage/po/TestIdpGroupPO.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Test
+  public void testBuilderReuseDoesNotMutateBuiltObject() {
+    var builder =
+        IdpGroupPO.builder()
+            .withGroupId(1L)
+            .withGroupName("engineering")
+            .withCurrentVersion(1L)
+            .withLastVersion(1L)
+            .withDeletedAt(0L);
```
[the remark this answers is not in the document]
lasdf1234: This project uses version 10 or higher.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/provider/base/IdpGroupMetaBaseSQLProvider.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public String deleteIdpGroupMetasByLegacyTimeline(
+      @Param("legacyTimeline") Long legacyTimeline, @Param("limit") int limit) {
+    return "DELETE FROM "
+        + IdpGroupMetaMapper.IDP_GROUP_TABLE_NAME
+        + " WHERE deleted_at > 0 AND deleted_at < #{legacyTimeline} LIMIT #{limit}";
+  }
+
+  protected String currentTimeMillisExpression() {
+    return "(UNIX_TIMESTAMP() * 1000.0)";
```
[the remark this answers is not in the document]
lasdf1234: Consistent with the original code

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/provider/postgresql/IdpGroupMetaPostgreSQLProvider.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  @Override
+  public String deleteIdpGroupMetasByLegacyTimeline(
+      @Param("legacyTimeline") Long legacyTimeline, @Param("limit") int limit) {
+    return "DELETE FROM "
+        + IdpGroupMetaMapper.IDP_GROUP_TABLE_NAME
+        + " WHERE group_id IN (SELECT group_id FROM "
+        + IdpGroupMetaMapper.IDP_GROUP_TABLE_NAME
+        + " WHERE deleted_at > 0 AND deleted_at < #{legacyTimeline} LIMIT #{limit})";
```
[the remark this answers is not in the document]
lasdf1234: Your statement is incorrect. It's not about deleting data in sequence; it's about deleting the data completely. This suggestion has been made many times before, so please don't repeat it!

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpGroupMetaSQLProviderFactory.java
```diff
@@ -0,0 +1,109 @@
...
+              "No IdP group SQL provider registered for backend %s (databaseId: %s)",
+              jdbcBackendType, databaseId));
+    } catch (IllegalArgumentException e) {
+      throw new IllegalStateException(
+          String.format(
+              "Unsupported IdP group SQL provider databaseId: %s, supported backends: %s",
+              databaseId, IDP_GROUP_META_SQL_PROVIDER_MAP.keySet()),
+          e);
+    }
+  }
```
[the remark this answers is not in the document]
lasdf1234: The core module also did not include this test class.IDP also doesn't plan to do it.

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/storage/mapper/TestIdpUserMetaStorage.java
```diff
@@ -42,7 +42,16 @@ class TestIdpUserMetaStorage extends AbstractIdpMetaStorageTest {
...
+    IdpUserPO firstUser =
+        IdpUserPO.builder()
+            .withUserId(1L)
+            .withUserName("alice")
+            .withPasswordHash("hash-a")
+            .withCurrentVersion(1L)
+            .withLastVersion(0L)
+            .withDeletedAt(0L)
+            .build();
+    idpUserMetaMapper.insertIdpUser(firstUser);
```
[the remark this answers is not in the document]
lasdf1234: This is not the content of this PR.

lasdf1234:
cc @roryqi  Could you please review this PR for me? THX

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/po/IdpGroupPO.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Getter
+@EqualsAndHashCode
+@ToString
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@Builder(setterPrefix = "with")
+public class IdpGroupPO {
+  private Long groupId;
+  private String groupname;
```
roryqi: You should use groupName here. 
Why do we use username instead of userName?
Because username is one word. `group name` is two words.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpGroupMetaSQLProviderFactory.java
```diff
@@ -0,0 +1,109 @@
...
+      new IdpGroupMetaPostgreSQLProvider();
+
+  private static final Map<JDBCBackendType, IdpGroupMetaBaseSQLProvider>
+      IDP_GROUP_META_SQL_PROVIDER_MAP =
+          ImmutableMap.of(
+              JDBCBackendType.MYSQL, IDP_GROUP_META_BASE_PROVIDER,
+              JDBCBackendType.H2, IDP_GROUP_META_H2_PROVIDER,
+              JDBCBackendType.POSTGRESQL, IDP_GROUP_META_POSTGRESQL_PROVIDER);
+
+  static IdpGroupMetaBaseSQLProvider getProvider(String databaseId) {
```
roryqi: Could u extract a util method? It's similar to user meta mapper. We reuse the method.
It will better that we can extract abstract class similar to BaseSQLProviderFactory.

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/storage/mapper/TestIdpGroupMetaStorage.java
```diff
@@ -0,0 +1,174 @@
...
+
+import java.io.IOException;
+import org.apache.gravitino.idp.storage.po.IdpGroupPO;
+import org.apache.gravitino.storage.relational.session.SqlSessionFactoryHelper;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@Tag("gravitino-docker-test")
+class TestIdpGroupMetaStorage extends AbstractIdpMetaStorageTest {
```
roryqi, line 39: Will it run the user tests iff we extends `AbstractIdpMetaStorageTest` directly. 
You have two options:
1) Remove user tests from `AbstractIdpMetaStorageTest`, add a user test class TestIdpUserMetaStorage
2) Put all the test cases to `AbstractIdpMetaStorageTest` and rename `AbstractIdpMetaStorageTest` to `IdpMetaStorageTest`.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/provider/postgresql/IdpGroupMetaPostgreSQLProvider.java
```diff
@@ -0,0 +1,42 @@
...
+  protected String currentTimeMillisExpression() {
+    return "CAST(EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000 AS BIGINT)";
+  }
+
+  @Override
+  public String deleteIdpGroupMetasByLegacyTimeline(
+      @Param("legacyTimeline") Long legacyTimeline, @Param("limit") int limit) {
+    return "DELETE FROM "
+        + IdpGroupMetaMapper.IDP_GROUP_TABLE_NAME
+        + " WHERE group_id IN (SELECT group_id FROM "
```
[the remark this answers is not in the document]
lasdf1234: Both "group_id" and "user_id" are primary keys, so there is no such issue.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/po/IdpGroupPO.java
lasdf1234: Got groupName rename to groupname

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/storage/mapper/TestIdpGroupMetaStorage.java
lasdf1234: I chose the first option. TestIdpUserMetaStorage and TestIdpGroupMetaStorage will run the test case.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpGroupMetaSQLProviderFactory.java
lasdf1234: Got IdpBaseSQLProviderFactory has been created.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/po/IdpGroupPO.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+
+@Getter
+@EqualsAndHashCode
+@ToString
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
```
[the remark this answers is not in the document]
lasdf1234: Be consistent with the previous code.

lasdf1234:
cc @roryqi All the codes have been modified. Please review them for me.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/provider/base/IdpGroupMetaBaseSQLProvider.java
```diff
@@ -0,0 +1,84 @@
...
+        + " current_version as currentVersion,"
+        + " last_version as lastVersion, deleted_at as deletedAt"
+        + " FROM "
+        + IdpGroupMetaMapper.IDP_GROUP_TABLE_NAME
+        + " WHERE deleted_at = 0 "
+        + "<foreach collection='groupNames' item='groupName'"
+        + " open='AND group_name IN (' separator=',' close=')'>"
+        + "#{groupName}"
+        + "</foreach>"
+        + "</script>";
```
yuqi1129, line 50: `selectIdpGroups(emptyList)` returns **every** active group instead of an empty result — when `groupNames` is empty the `<foreach open='AND group_name IN ('...>` block is skipped entirely, so the WHERE clause reduces to `deleted_at = 0`.

This matches the pre-existing `selectIdpUsers` behavior (see `TestIdpUserMetaStorage.java:92-94`) and `TestIdpGroupMetaStorage.java:85-87` explicitly pins it, so it isn't a regression. But it's a footgun: typical SQL conventions treat an empty `IN ()` as "matches nothing", and a future caller doing `selectIdpGroups(filteredNames)` where `filteredNames` is empty after some upstream filter would silently get every active group instead of zero rows.

Two paths if you want to keep this PR scoped:

1. Add a one-line javadoc on `IdpGroupMetaMapper.selectIdpGroups` (and ideally on `selectIdpUsers` too) saying "an empty list returns all active groups; pass `null` for an explicit error".
2. Or change the SQL to `<foreach ... open=' AND group_name IN ('><if test="false">...` — well actually the cleaner version is `<if test="groupNames != null and !groupNames.isEmpty()">AND group_name IN <foreach...></if>` plus a `<choose>` that yields `1=0` when the list is empty if you want "empty list means no match".

Option 1 is cheap and preserves the existing convention.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpBaseSQLProviderFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+abstract class IdpBaseSQLProviderFactory<T> {
+  private final T mysqlProvider;
+  private final T h2Provider;
+  private final T postgresqlProvider;
+  private final Map<JDBCBackendType, T> providerMap;
+  private final String providerName;
+
+  protected IdpBaseSQLProviderFactory(
+      String providerName, T mysqlProvider, T h2Provider, T postgresqlProvider) {
```
yuqi1129: The parameter named `mysqlProvider` here actually receives `new IdpGroupMetaBaseSQLProvider()` / `new IdpUserMetaBaseSQLProvider()` at the call sites — i.e. the **base** class is being passed as the "mysql" slot, not a MySQL-specific subclass. That works because `(UNIX_TIMESTAMP() * 1000.0)` in the base provider happens to be MySQL-flavored, but for a new reader it's surprising: looking at the H2 / PostgreSQL subclasses you naturally expect a `IdpGroupMetaMySQLProvider` next to them and there isn't one.

Two ways to make this self-documenting (pick whichever feels less invasive):

1. **Cheap** — rename the constructor parameter to `mysqlOrBaseProvider` (or `mysqlAndDefaultProvider`) and add one javadoc line saying "the base SQL provider is the MySQL provider by convention; H2 / PostgreSQL extend it with backend-specific overrides".
2. **Bigger** — rename `IdpGroupMetaBaseSQLProvider` → `IdpGroupMetaMySQLProvider` (and the user one too), then have an actual `IdpGroupMetaBaseSQLProvider` abstract class without the MySQL-specific default. That makes the constructor call read naturally: `super("...", new IdpGroupMetaMySQLProvider(), new IdpGroupMetaH2Provider(), new IdpGroupMetaPostgreSQLProvider())`.

Option 1 is cheap and worth doing here since this base class is now shared by both user and group factories — the confusion compounds once a third subclass shows up.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/provider/base/IdpGroupMetaBaseSQLProvider.java
lasdf1234: Got. Add java doc. Resolved.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpBaseSQLProviderFactory.java
lasdf1234: The code needs to be consistent with that of the Core module. No additional "MySQLProvider" is required.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpBaseSQLProviderFactory.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  protected final T currentProviderInstance() {
+    return resolveProvider(currentDatabaseId());
+  }
+
+  protected final T resolveProvider(String databaseId) {
+    if (databaseId == null) {
+      throw new IllegalStateException(
+          String.format("MyBatis databaseId is not configured for %s.", providerName));
```
roryqi: Could u use `getClass().getSimpleName()` here instead of `providerName` here? Could u remove `providerName` in the constructor method?

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpGroupMetaSQLProviderFactory.java
```diff
@@ -0,0 +1,80 @@
...
+  public static IdpGroupMetaBaseSQLProvider postgresqlProvider() {
+    return INSTANCE.postgresqlProviderInstance();
+  }
+
+  static IdpGroupMetaBaseSQLProvider getProvider(String databaseId) {
+    return INSTANCE.resolveProvider(databaseId);
+  }
+
+  public static String selectIdpGroup(@Param("groupName") String groupName) {
+    return INSTANCE.currentProviderInstance().selectIdpGroup(groupName);
```
roryqi: This pattern seems weird. 
It is weird to see `instance.stance`. 
There are some possible options, I am not sure which is better.
1. Could u make some methods become method instead of static methods.
2. Could u construct a instance currentProviderInstance directly?

roryqi: If you have no idea, I can help you rewrite the code.

lasdf1234: Got I use currentProvider() instead of INSTANCE

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpBaseSQLProviderFactory.java
lasdf1234: Got,This is truly an excellent method. It has been revised and the code has become much more concise.

lasdf1234:
cc @roryqi  @yuqi1129  All the comments have been resolved. If you have time, please run the CI test for me. Thank you.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpBaseSQLProviderFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ * under the License.
+ */
+
+package org.apache.gravitino.idp.storage.mapper;
+
+import java.util.Map;
+import org.apache.gravitino.storage.relational.JDBCBackend.JDBCBackendType;
+import org.apache.gravitino.storage.relational.session.SqlSessionFactoryHelper;
+
+final class IdpBaseSQLProviderFactory {
```
roryqi: Could u rename this class to `SQLProviderFactoryHelper`?

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/storage/mapper/AbstractIdpMetaStorageTest.java
```diff
@@ -88,6 +84,7 @@ protected void init(String type) throws IOException {
     backend = new JDBCBackend();
     backend.initialize(config);
     sharedSession = SqlSessionFactoryHelper.getInstance().getSqlSessionFactory().openSession(true);
+    idpGroupMetaMapper = sharedSession.getMapper(IdpGroupMetaMapper.class);
```
roryqi: Could u remove mapper initialization to sub classes?

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/storage/mapper/IdpBaseSQLProviderFactory.java
lasdf1234: Got resolved.

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/storage/mapper/AbstractIdpMetaStorageTest.java
lasdf1234: Got remove mapper init into child class.

roryqi (approved):

PR #10724, opened 2026-04-09
[#10588] doc(view): add design doc for view management

mchades:
### What changes were proposed in this pull request?

add design doc for view management

### Why are the changes needed?

Fix: #10588

### Does this PR introduce _any_ user-facing change?

no

### How was this patch tested?

no need

PR #11144, opened 2026-05-19
  [#11007] docs(authz): Add design doc for view privilege control

### LINKED ISSUE #11007: [Subtask] View privilege support
### Describe the subtask

Complete/enforce view-level privileges (grant/revoke/check).

### Parent issue

#604

sgedward:

  Adds a new design document at `design-docs/gravitino-view-privilege.md` covering view privilege control as a sub-design under the parent [View Management
  design](https://docs.google.com/document/d/1qKZMcY5ifgZF-BjGF2FwYBNWyTwqrCDLaGW_D2jD_LY).

  The doc proposes:
  - **Adding `ALTER_VIEW` privilege** (with bit `1L << 33`, allow/deny variants)
  - **Wiring `@AuthorizationExpression`** on all five `ViewOperations.java` endpoints, mirroring the `TableOperations.java` pattern
  - **Adding `ViewHookDispatcher`** to set the creator as view owner on `createView`, unblocking the `VIEW::OWNER` clause already present in `LOAD_VIEW_AUTHORIZATION_EXPRESSION`
  - **Adding visibility filtering** to `listViews` via `FILTER_VIEW_AUTHORIZATION_EXPRESSION` (currently consumed only by the Iceberg-REST path)
  - **Two new expression constants**: `CREATE_VIEW_AUTHORIZATION_EXPRESSION` and `ALTER_VIEW_AUTHORIZATION_EXPRESSION`

  The doc **diverges from the parent View Management design doc** on one point: no `DROP_VIEW` privilege is introduced. View drops use the hierarchical owner-based expression form,
  consistent with `dropTable` / `dropFileset` / `dropFunction`. Rationale in the doc — would appreciate community review on this specifically.

Issue #11007 tracks the completion of view privilege support. However, it would be better to summarize the changes and aggregate them into a design document to share with the community for confirmation before proceeding to the implementation phase.

  Current state has `CREATE_VIEW` / `SELECT_VIEW` defined in `Privilege.Name` but never enforced on the generic REST path. Alter and drop operations have no privilege protection. Only the
   Iceberg-REST path (`IcebergViewOperations.java`) is fully wired today, making the privilege model inconsistent across catalog types.

  Fix: #11007

  No code changes in this PR — design doc only. The implementation work that follows the design will introduce:

  1. A new privilege type `ALTER_VIEW` (grantable like `MODIFY_TABLE`)
  2. Enforcement of `CREATE_VIEW` / `SELECT_VIEW` / `ALTER_VIEW` on Gravitino REST endpoints for non-Iceberg catalogs (previously unenforced)
  3. Behavioral change: `listViews` will return only views the caller has privileges on (currently returns all views in the schema)
  4. Hierarchical owner-based drop authorization for views (currently no privilege check at all)

  N/A — this is a design document. Reviewers should evaluate the proposed approach and the rationale for diverging from the parent doc on `DROP_VIEW`.

### DIFF
--- a/design-docs/gravitino-view-privilege.md
+++ b/design-docs/gravitino-view-privilege.md
@@ -1,0 +1,241 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Design of View Privilege Control in Gravitino
+
+## Background
+
+Apache Gravitino's [View Management design](https://docs.google.com/document/d/1qKZMcY5ifgZF-BjGF2FwYBNWyTwqrCDLaGW_D2jD_LY) provides unified view management across heterogeneous catalogs (Iceberg, Paimon, HMS, JDBC, and Gravitino-managed catalogs). The parent design assigns view privilege control to "Phase IV. Governance & Security" and lists four required privileges: `CREATE_VIEW`, `SELECT_VIEW`, `ALTER_VIEW`, `DROP_VIEW`.
+
+The existing Gravitino access control framework covers catalogs, schemas, tables, filesets, topics, models, tags, policies, jobs, and functions. Views are a partially-defined object type in this model — the privilege types exist in the API but are not enforced end-to-end, leaving view access control in an inconsistent state:
+
+- **Privilege type set is incomplete.** Only `CREATE_VIEW` and `SELECT_VIEW` exist in `Privilege.Name` (see `api/src/main/java/org/apache/gravitino/authorization/Privilege.java:146-148`). `ALTER_VIEW`-equivalent and `DROP_VIEW`-equivalent are missing, so view alter and drop operations have no privilege to check against.
+- **REST endpoints do not enforce privileges.** `ViewOperations.java` (the generic view REST endpoints) carries zero `@AuthorizationExpression` annotations on its five endpoints (`listViews`, `createView`, `loadView`, `alterView`, `dropView` at lines 67, 93, 132, 158, 191 respectively). Compare with `TableOperations.java` where every endpoint is wired (lines 83, 119, 167, 204, 241).
+- **No visibility filtering on `listViews`.** `ViewOperations.listViews` (lines 71–94) returns the raw `dispatcher.listViews(viewNS)` result with no filter applied. Compare with `TableOperations.listTables` (lines 86–112) which calls `MetadataAuthzHelper.filterByExpression(..., FILTER_TABLE_AUTHORIZATION_EXPRESSION, ...)` to drop entries the caller has no privilege on. The `FILTER_VIEW_AUTHORIZATION_EXPRESSION` constant already exists in `AuthorizationExpressionConstants.java:119` but is consumed only by the Iceberg-REST path (`IcebergViewOperations.java:350`); the generic path never references it. Result: any user who can reach the list endpoint sees every view in the schema, regardless of per-view privileges.
+- **No ownership tracking for views.** `core/src/main/java/org/apache/gravitino/hook/` contains 12 `*HookDispatcher` classes (Catalog, Schema, Table, Fileset, Function, Model, …) but no `ViewHookDispatcher`. As a result, view owners are never set on creation, and the `VIEW::OWNER` clause already present in `LOAD_VIEW_AUTHORIZATION_EXPRESSION` (`AuthorizationExpressionConstants.java:97`) never resolves to true for the view creator.
+- **Engine-side ACL translation is silent.** The Ranger plugin (`authorizations/authorization-ranger/`) contains zero references to `MetadataObject.Type.VIEW` or any view privilege — view grants are not translated to Ranger ACLs.
+- **Iceberg-REST is the only path that enforces view auth today.** `IcebergViewOperations.java` (lines 90, 126, 164, 201, 243, 282) is fully wired with `@AuthorizationExpression` and `ICEBERG_LOAD_VIEW_AUTHORIZATION_EXPRESSION`. This means the same conceptual operation has two enforcement paths with different coverage depending on the catalog type, which is confusing for users and operators.
+
+The result is a privilege model where granting `SELECT_VIEW` to a user has visible effect on Iceberg-REST views but no effect on views served via the generic REST path, and where alter/drop operations on any view are unprotected.
+
+**Current state — two parallel paths, only one enforces view auth:**
+
+```
+                       ┌────────────────────────────────────────┐
+                       │       Authorization framework          │
+                       │  (roles, grants, owner, expressions)   │
+                       └──────────────┬─────────────────────────┘
+                                      │ checked at REST layer
+                                      │
+       ┌──────────────────────────────┴──────────────────────────────┐
+       │                                                             │
+       ▼                                                             ▼
+  IcebergViewOperations                                       ViewOperations
+  (Iceberg REST spec path)                                    (Gravitino REST path)
+       │ @AuthorizationExpression                                  │ (no @AuthorizationExpression)
+       │ on every endpoint  ✅                                      │ on any endpoint     ❌
+       ▼                                                            ▼
+  Iceberg REST handlers                                       ViewNormalizeDispatcher
+                                                                    │
+                                                                    │ (no ViewHookDispatcher
+                                                                    │  → view owner never set ❌
+                                                                    │  → VIEW::OWNER clause
+                                                                    │    never resolves true)
+                                                                    ▼
+                                                              ViewOperationDispatcher
+                                                                    │
+                                                                    ▼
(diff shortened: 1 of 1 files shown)


design-docs/gravitino-view-privilege.md
```diff
@@ -0,0 +1,241 @@
...
+
+5. **Backward Compatibility**: When authorization is disabled, all view REST endpoints behave identically to today. Existing `CREATE_VIEW` and `SELECT_VIEW` privileges retain their current bit values and semantics. No client or DTO changes. The Iceberg-REST view authorization path (`IcebergViewOperations.java`) is unchanged.
+
+---
+
+## Non-Goals
+
+1. **Data-Level Access Control**: View privileges in Gravitino govern metadata access only. Whether the user can read the underlying tables referenced by the view SQL is enforced by the compute engine and the underlying catalog's permission system. The parent design doc commits to this scope split explicitly.
+
+2. **DEFINER / INVOKER Enforcement**: `securityConfig.securityMode` is stored by Gravitino as a metadata annotation and passed through to the engine. Gravitino does not enforce DEFINER/INVOKER semantics. This is the parent design doc's position and is unchanged here.
```
mchades: Gravitino currently does not explicitly store the `securityConfig.securityMode`. Should we consider adding this to the view field or view properties? Additionally, are these values mutable or immutable?

design-docs/gravitino-view-privilege.md
```diff
@@ -0,0 +1,241 @@
...
+                                                          Catalog connector (Iceberg/
+                                                          Paimon/HMS/JDBC/managed)
+
+  Ranger plugin: zero references to MetadataObject.Type.VIEW ❌
+  (view grants in Gravitino are not translated to Ranger ACLs)
+```
+
+---
+
+## Goals
```
mchades, line 73: The design should take this into consideration: https://github.com/apache/gravitino/pull/10724#discussion_r3056018270

Image

design-docs/gravitino-view-privilege.md
sgedward: @mchades 

Thanks for pointing that out; I’ll move this to out of scope for now. I’m thinking it could stay in the view field since it’s a simple enum, and not all data engines support the mode—plus, it’s likely a catalog-managed feature. We can have the client-side check eligibility before sending it to the server.

Regarding the values, maybe we can come back to this later. I’m leaning towards "mutable" since MySQL supports the alter mode. However, the Iceberg change is still ongoing, and I’m not sure if IRC is immutable or mutable. We can circle back once they have an implementation. Or maybe I missed something? What do you think? Feel free to point me toward any references.

mchades: Although different engines have varying degrees of support for this attribute, it is a key design for view privilege, and we should not skip it in the design of view privilege. Since we also maintain connectors for Trino/Flink/Spark, we need to consider comprehensively in this design document how to support it.

sgedward: Hi @mchades,

Thanks for pointing out the information here. I’ve updated the design docs, especially the security mode sections.

I have divided the security mode into three components:
1. The connectors
2. Gravitino
3. The underlying data engines

I separately covered how each section handles various situations and combinations. I have encountered one difficulty: Flink and Spark do not have a caller identity, only a session identity (or maybe I am missing something; please let me know). This means that when we switch the security mode to invoker mode, the caller can still use the Flink or Spark connector’s session identity (which is the service identity) to access the view. 

This seems to defeat the purpose of invoker privilege mode. For now, I will call this out as an exception and treat the session identity the same way as the user or caller identity, as is done in Trino.

mchades:
@roryqi could you please also help to review the design when you are free? Thanks!

roryqi:
I will review this document tonight.

design-docs/gravitino-view-privilege.md
```diff
@@ -0,0 +1,339 @@
...
+
+2. **INVOKER's accuracy depends on the cell**:
+   - **Pass-through cells** (native-dialect path to MySQL/Postgres/HMS): the data engine enforces with whatever identity the connector's JDBC/Thrift connection uses. Identity propagation between connector and data engine is a connector / deployment concern outside this design — typically per-user JDBC credentials, proxy authentication, or Kerberos delegation.
+   - **Gravitino-managed cells** (data lakes, or non-native dialect on a JDBC data engine): Gravitino enforces using session identity. Accurate per-user on Trino with `forwardUser`; degraded to service-identity enforcement on Spark/Flink today. The degradation is principled — INVOKER means "whoever Gravitino sees as the caller" — and matches deployment reality, since operators today grant `spark_svc` / `flink_svc` the privileges they want Spark/Flink jobs to have. When per-session forwarding lands in those connectors, INVOKER becomes per-user automatically with no design changes here.
+
+---
+
+#### Identity Mapping (Pass-through Mode Only)
```
roryqi: Why do we need this?

sgedward: This might be due to my knowledge gap in Gravitino, but I assume the Gravitino user principal may not be the same as the JDBC database connection user. Consequently, I manually managed the principal mapping. If this assumption is incorrect, I can remove that logic. Please feel free to point me in the right direction.

roryqi: We don't have a mature solution now. It's a complex solution.

sgedward: > We don't have a mature solution now. It's a complex solution.

Got it. If that’s the case, I can pass the Gravitino user principal as the JDBC connection user and put the complex solution out of scope for now. What do you think?

roryqi: We can only focus on the metadata authentication now.

sgedward: Updated the design doc and marked the metadata <--> data engine identity mapping out of scope for now.

design-docs/gravitino-view-privilege.md
```diff
@@ -0,0 +1,324 @@
...
+   - Applies `FILTER_VIEW_AUTHORIZATION_EXPRESSION` (`AuthorizationExpressionConstants.java:119`) to the result set — only views the user has `SELECT_VIEW`, `ALTER_VIEW`, `CREATE_VIEW` (at schema, catalog, or metalake level), or ownership on are returned.
+
+2. **`loadView`**
+   - Requires `USE_CATALOG` + `USE_SCHEMA`, plus `SELECT_VIEW`, `ALTER_VIEW`, `CREATE_VIEW`, or view ownership (consistent with `loadTable`).
+   - If the user lacks privileges, the authorization framework denies access.
+
+3. **`createView`**
+   - Requires `USE_CATALOG` + `USE_SCHEMA` + (`CREATE_VIEW` at schema, catalog, or metalake, or schema ownership).
+
+4. **`alterView`**
```
roryqi, line 172: Do we support to rename a view ? Do we support rename a view crossing different namespace?

design-docs/gravitino-view-privilege.md
```diff
@@ -0,0 +1,324 @@
...
+| **Spark** (singleton service id)      | DEFINER ✅ owner / INVOKER ⚠️ service                     | DEFINER ✅ / INVOKER ✅ (data engine enforces)                                            | DEFINER ✅ owner / INVOKER ⚠️ service                                          |
+| **Flink** (singleton service id)      | DEFINER ✅ owner / INVOKER ⚠️ service                     | DEFINER ✅ / INVOKER ✅ (data engine enforces)                                            | DEFINER ✅ owner / INVOKER ⚠️ service                                          |
+
+Two consequences of this design:
+
+1. **DEFINER (the default) is fully enforced in every cell from V1.** Whether the data engine has a native view-security concept (MySQL/Postgres/HMS) or not (data lakes), and whether the connector forwards real-user identity (Trino with `forwardUser`) or only a session/service identity (Spark/Flink), Gravitino's lookup of `view.owner()` from `owner_meta` and subsequent authorization is independent of the calling connection. DEFINER is the default precisely because it is comprehensively enforceable today.
```
roryqi, line 283: Why is definer default mode?

sgedward: There are two reasons:
1. The [view design doc](https://docs.google.com/document/d/1qKZMcY5ifgZF-BjGF2FwYBNWyTwqrCDLaGW_D2jD_LY/edit?tab=t.0) table information schema uses DEFINER as the default:
```
 `security_mode` VARCHAR(32) NOT NULL DEFAULT 'DEFINER' COMMENT 'DEFINER or INVOKER, immutable in V1',
 ```
2. The MySQL default for `CREATE VIEW my_view AS SELECT ...;` is DEFINER mode. It is easier to align with the data engine.

design-docs/gravitino-view-privilege.md
sgedward: That’s a good call. I don’t think we should support it in `ALTER VIEW` to keep things simple. If a user wants to rename a view, they can just drop and recreate it. I will outline the limitation here.

design-docs/gravitino-view-privilege.md
roryqi: cc @mchades Most of computes won't support definer mode now.

sgedward: Sorry, I think there might have been a misunderstanding on my end. I thought we needed input from @mchades first.

The core issue here seems to be a misalignment between our current support and MySQL's behavior (which uses `DEFINER` as the default). Given that, should we proceed with Definer mode, or do you think we should take a different approach?

design-docs/gravitino-view-privilege.md
sgedward: @roryqi  I think `ALTER VIEW` is still a broad term for supporting the main feature here. I have scoped out `RENAME` support for the current `ALTER_VIEW` feature.

sgedward:
Hi @roryqi and @mchades,

Should we share the design document for community review now? The primary outstanding question is whether definer mode should be set as the default.

Thanks!

design-docs/gravitino-view-privilege.md
roryqi: I prefer using invoker mode as the default mode.

PR #11152, opened 2026-05-19
[#11220] docs(spec): Add design documents for async hard deletion

roryqi:

Adds the design document `design-docs/async-iceberg-rest-hard-deletion.md` for **asynchronous hard deletion** in the Gravitino Iceberg REST server.

The design makes `DELETE …?purgeRequested=true` return quickly by dropping the catalog entry on the request thread and deleting data/metadata files in the background via a JDBC-backed job table (`iceberg_purge_job`) and a multi-replica worker pool (lease via `FOR UPDATE SKIP LOCKED`, CAS fallback elsewhere), with retry/backoff and a `DEAD_LETTER` state. The synchronous path is retained behind `async-purge.enabled=false` purely as a rollback.

Resolved review/open questions in this revision:
- **Name reuse during purge (§5.13, closes §8.1 / Jerry's Q1):** the active job row acts as a tombstone — `createTable`/`register` at the same identifier returns **409** until the job is terminal, closing the PRD §2.2 same-name/same-location recreate attack. Re-register-as-recovery atomically cancels the matching job (`CANCELLED`), removing the worker-deletes-recovered-files race.
- **Purge visibility (§5.14, Jerry's follow-up):** how to tell a table is mid-purge / done / dead-lettered — a wire-compatible informative `409` message, a read-only `purge-jobs` management endpoint, metrics, and events, with `DEAD_LETTER` discoverable across all three.
- Marked §8.1 (Q1) and §8.4 (Q4) resolved; updated schema (`CANCELLED` state, `idx_object`), failure model, wire-compat notes, and task breakdown.

Synchronous purge on the Jetty request thread exceeds HTTP timeouts, saturates the thread pool, and leaks files with no retry/audit on large tables. This document captures the async design before implementation.

No (design document only). The design it describes would change drop semantics (recreate blocked with 409 until cleanup completes) — called out in the doc's release-notes section.

N/A — documentation only.

### DIFF
--- a/design-docs/async-iceberg-rest-hard-deletion.md
+++ b/design-docs/async-iceberg-rest-hard-deletion.md
@@ -77,8 +77,9 @@ different failure model — kept separate.
 ## 3. Non-Goals
 
 1. **Native soft-delete / undrop semantics**: This design delivers async
-   *hard* deletion only. Full soft-delete / undrop is a follow-up; §5.7
-   records the seam it builds on.
+   *hard* deletion only — once a drop is accepted the files are destroyed
+   with no undrop or recovery path. Full soft-delete / undrop is a separate
+   follow-up.
 2. **Purge cancellation**: User-initiated cancellation of in-flight purges
    is out of scope; it needs a control API and lifecycle states we can add
    later without breaking this design.
@@ -160,11 +161,7 @@ false`. The header name uses canonical HTTP casing (each word capitalized).
    `metadata_location`, and streams through every reachable file deleting
    it, retrying transient failures with backoff.
 5. On success the job is `SUCCEEDED`; on terminal failure it lands in
-   `FAILED` for operator inspection. Listeners observe
-   `IcebergPurge{Started,Completed,Failed}Event` throughout.
-6. *(Recovery)* If the drop was a mistake, the table can be re-registered at
-   the stored metadata location any time before the worker deletes the
-   files; this also cancels the matching job (§5.7).
+   `FAILED` for operator inspection (queryable in the job table, §5.9).
 
 ### 5.3 Request-path interaction
 
@@ -212,7 +209,7 @@ CREATE TABLE IF NOT EXISTS `iceberg_purge_job` (
   `metadata_location` VARCHAR(1024) NOT NULL,
   `file_io_impl`      VARCHAR(256)  NOT NULL,
   `file_io_props`     MEDIUMTEXT    NOT NULL COMMENT 'JSON',
-  `state`             VARCHAR(16)   NOT NULL COMMENT 'PENDING|RUNNING|SUCCEEDED|FAILED|CANCELLED',
+  `state`             VARCHAR(16)   NOT NULL COMMENT 'PENDING|RUNNING|SUCCEEDED|FAILED',
   `attempts`          INT(10)       NOT NULL DEFAULT 0,
   `last_error`        TEXT          NULL,
   `heartbeat_at`      BIGINT(20)    NULL COMMENT 'last heartbeat from the worker; NULL when unclaimed',
@@ -309,7 +306,6 @@ success. Backoff: `min(maxBackoff, base * 2^attempts)` with jitter.
 | Metadata load failed, transient     | back to `PENDING`, `attempts++`, `next_attempt_at = now + backoff`, `heartbeat_at=NULL`   |
 | Metadata load failed, terminal      | `attempts++`; if `attempts >= max-attempts` → `FAILED`                                   |
 | Worker killed mid-job               | `RUNNING` row's heartbeat goes stale; another worker reclaims; deletes are idempotent     |
-| Recovered via re-register (§5.7)    | Job CAS'd to `CANCELLED`; the worker only claims non-terminal rows, so files stay intact  |
 
 Restart handling falls out of the durable job table plus the heartbeat
 model — no separate mechanism:
@@ -330,7 +326,7 @@ heartbeat ages past `heartbeatTimeout` before the rebooted process reclaims
 them. We keep `heartbeatTimeout` modest so the resume delay is small; with no
 owner column there is nothing extra to sweep on boot.
 
-### 5.7 Name reuse during purge, and recovery
+### 5.7 Name reuse during purge
 
 While a purge job for an identifier is non-terminal (`PENDING` or
 `RUNNING`), that identifier is treated as **still occupied** even though its
@@ -338,11 +334,11 @@ catalog entry is already gone. The active `iceberg_purge_job` row *is* the
 tombstone — no second table is needed. The REST server consults the store on
 the request thread (one indexed lookup via `idx_object`):
 
-| Operation                                   | Active job exists                         | No active job (`SUCCEEDED`/`FAILED`/`CANCELLED`/none) |
-|---------------------------------------------|-------------------------------------------|-------------------------------------------------------|
-| `loadTable` / `HEAD`, `alterTable` / commit | `404 NoSuchTableException`                | `404`                                                 |
-| `createTable` (same identifier)             | **`409 Conflict`** — being purged         | succeeds                                              |
-| `registerTable` (same identifier)           | **`409 Conflict`**, except recovery below | succeeds                                              |
+| Operation                                   | Active job exists                 | No active job (`SUCCEEDED`/`FAILED`/none) |
+|---------------------------------------------|-----------------------------------|-------------------------------------------|
+| `loadTable` / `HEAD`, `alterTable` / commit | `404 NoSuchTableException`        | `404`                                     |
+| `createTable` (same identifier)             | **`409 Conflict`** — being purged | succeeds                                  |
+| `registerTable` (same identifier)           | **`409 Conflict`**                | succeeds                                  |
 
 A `SUCCEEDED` job (even before pruning) no longer blocks reuse — the files
 are gone. A `FAILED` job also stops blocking, so a failed cleanup never
@@ -352,33 +348,17 @@ attack: a recreate cannot expose the dropped table's data, because it is
 refused until the files are deleted. A **namespace** carries no data, so it
 can be recreated freely; tables inside it are guarded individually.
 
-**Recovery.** Because the data / metadata files are untouched until the
-worker runs, a table dropped by mistake can be recovered before its files
-are deleted by re-registering at the stored location:
+Once a drop is accepted the table is gone for good: this is *hard* delete,
+so there is deliberately no undrop or re-register recovery path. A client
+that wants a grace period before data is destroyed should use a future
+soft-delete feature (a non-goal here, §3), not this path.
 
-```
-POST /v1/{prefix}/namespaces/{ns}/register
-{ "name": "<table>", "metadata-location": "<stored metadata_location>" }
-```
-
-Re-registering atomically CAS's the matching active job to `CANCELLED` and
-restores the catalog entry in the same transaction — lifting the tombstone
-and removing the worker-deletes-the-recovered-table race (the worker only
-claims non-terminal rows). This is **recovery-scoped** cancellation only; a
-general "cancel my purge" API stays out of scope (§3).
-
-Once the worker deletes the files the table is unrecoverable — the intended
-semantics of *hard* delete. A future soft-delete / undrop feature layers on
-top by deferring file deletion for a retention window (delaying
-`next_attempt_at`) on the **same** job row, executor, and scheduler, so
-register-table recovery succeeds throughout the window.
-
-**Concurrency.** The conflict check, the worker's claim, and recovery's
-cancel all serialize on the row's `state`. Recovery's CAS to `CANCELLED` is
-guarded by `state='PENDING'`, and the worker's claim CAS moves
-`PENDING → RUNNING`; they target the same `PENDING` row, so at most one wins.
-A job already `RUNNING` cannot be silently cancelled — recovery's CAS matches
-no row and returns `409` (the caller retries once the job terminates).
+**Concurrency.** The create/register conflict check and the worker's claim
+both serialize on the row's `state`. A `createTable` / `registerTable` that
+observes a non-terminal job (`PENDING` or `RUNNING`) returns `409`; one that
+observes a terminal job (`SUCCEEDED` / `FAILED`) or no job proceeds. The
+worker's claim CAS moves `PENDING → RUNNING`, so the conflict check and the
+claim never disagree about whether the identifier is still occupied.
 
 ### 5.8 Object coverage: tables now, views later
 
@@ -433,13 +413,13 @@ per-request **client** choice via the `X-Gravitino-Async-Purge` header
 (default async; `false` selects synchronous deletion). The server-side keys
 only tune the worker pool and retries:
 
-| Key                                                            | Default  | Description                                             |
-|----------------------------------------------------------------|----------|---------------------------------------------------------|
-| `gravitino.iceberg-rest.async-purge.worker-threads`            | `4`      | Worker pool size per server.                            |
-| `gravitino.iceberg-rest.async-purge.poll-interval-ms`          | `5000`   | Worker poll interval.                                   |
-| `gravitino.iceberg-rest.async-purge.heartbeat-timeout-ms`      | `300000` | Age after which a job with no heartbeat is reclaimable. |
-| `gravitino.iceberg-rest.async-purge.max-attempts`              | `5`      | Attempts before `FAILED`.                               |
-| `gravitino.iceberg-rest.async-purge.completed-retention-hours` | `168`    | How long `SUCCEEDED` rows are retained before pruning.  |
+| Key                                                           | Default  | Description                                                                  |
+|---------------------------------------------------------------|----------|------------------------------------------------------------------------------|
+| `gravitino.iceberg-rest.async-purge.worker-threads`           | `4`      | Worker pool size per server.                                                 |
+| `gravitino.iceberg-rest.async-purge.poll-interval-ms`         | `5000`   | Worker poll interval.                                                        |
+| `gravitino.iceberg-rest.async-purge.heartbeat-timeout-ms`     | `300000` | Age after which a job with no heartbeat is reclaimable.                      |
+| `gravitino.iceberg-rest.async-purge.max-attempts`             | `5`      | Attempts before `FAILED`.                                                    |
+| `gravitino.iceberg-rest.async-purge.terminal-retention-hours` | `168`    | How long terminal (`SUCCEEDED` / `FAILED`) rows are retained before pruning. |
 
 File-delete parallelism, claim batch size, and backoff base/ceiling are
 internal constants with sensible defaults; they become config keys only if a
@@ -451,8 +431,8 @@ deployment demonstrates the need.
   unchanged.
 - The worker uses FileIO credentials snapshotted at enqueue time.
   Credentials that expire before the worker runs (STS tokens, …) are
-  refreshed from the catalog on each heartbeat, so long-running purges (and
-  future long-retention soft delete) keep valid credentials.
+  refreshed from the catalog on each heartbeat, so long-running purges keep
+  valid credentials.
 - `iceberg_purge_job` may contain credentials in `file_io_props`; the
   existing Gravitino DB encryption / access controls apply.
 
@@ -489,7 +469,6 @@ only as a rollback flag.
 - [ ] Add observability (§5.9): metrics and the informative `ErrorResponse` message on the §5.7 `409`; operator read path in 1.3 is direct DB query
 - [ ] Namespace-drop cascade for tables (one job per contained table)
 - [ ] Enforce tombstone semantics (§5.7): on `createTable`/`registerTable`, reject with `409` when an active job exists (`idx_object` lookup on the request thread)
-- [ ] Add register-table recovery (§5.7): atomically CAS the matching active job to `CANCELLED` and restore the catalog entry; documentation
 
 ### Phase 1 (1.3): Testing
 - [ ] Unit tests: enqueue, state machine, worker claiming/heartbeat/contention (H2)
@@ -518,9 +497,7 @@ only as a rollback flag.
     synchronous purge; thrown errors surface as 5xx.
   - `TestIcebergPurgeTombstone` (§5.7) — `createTable`/`register` at an
     identifier with an active job returns `409`, succeeds once
-    `SUCCEEDED`/`FAILED`; `loadTable`/`alterTable` return `404`; re-register
-    CAS's the job to `CANCELLED` and the worker then skips it; the
-    recovery-vs-worker CAS race resolves to a single winner.
+    `SUCCEEDED`/`FAILED`; `loadTable`/`alterTable` return `404`.
 - Integration (`gravitino-docker-test`):
   - Drop a table; REST returns < 500 ms; worker eventually clears every file.
   - Restart REST mid-purge (`kill -9`); purge resumes.
--- a/design-docs/async-iceberg-rest-hard-deletion.md
+++ b/design-docs/async-iceberg-rest-hard-deletion.md
@@ -303,13 +303,13 @@ synchronous purge has the same "best effort" stance. A job fails only if the
 **metadata phase** fails. `NotFoundException` from `deleteFile` counts as
 success. Backoff: `min(maxBackoff, base * 2^attempts)` with jitter.
 
-| Outcome                            | Action                                                              |
-| ---------------------------------- | ------------------------------------------------------------------- |
-| All files deleted (or already gone) | `state='SUCCEEDED'`                                                |
-| Metadata load failed, transient    | back to `PENDING`, `attempts++`, `next_attempt_at = now + backoff`, `heartbeat_at=NULL` |
-| Metadata load failed, terminal     | `attempts++`; if `attempts >= max-attempts` → `FAILED`              |
-| Worker killed mid-job              | `RUNNING` row's heartbeat goes stale; another worker reclaims; deletes are idempotent |
-| Recovered via re-register (§5.7)   | Job CAS'd to `CANCELLED`; the worker only claims non-terminal rows, so files stay intact |
+| Outcome                             | Action                                                                                   |
+|-------------------------------------|------------------------------------------------------------------------------------------|
+| All files deleted (or already gone) | `state='SUCCEEDED'`                                                                       |
+| Metadata load failed, transient     | back to `PENDING`, `attempts++`, `next_attempt_at = now + backoff`, `heartbeat_at=NULL`   |
+| Metadata load failed, terminal      | `attempts++`; if `attempts >= max-attempts` → `FAILED`                                   |
+| Worker killed mid-job               | `RUNNING` row's heartbeat goes stale; another worker reclaims; deletes are idempotent     |
+| Recovered via re-register (§5.7)    | Job CAS'd to `CANCELLED`; the worker only claims non-terminal rows, so files stay intact  |
 
 Restart handling falls out of the durable job table plus the heartbeat
 model — no separate mechanism:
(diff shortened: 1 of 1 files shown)


design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+| `…purger.jdbc-async.max-attempts`                             | `5`       |
+| `…purger.jdbc-async.backoff-base-ms`                          | `30000`   |
+| `…purger.jdbc-async.backoff-max-ms`                           | `3600000` |
+| `…purger.jdbc-async.completed-retention-hours`                | `168`     |
+
+`synchronous` has no tunables.
+
+## 9. Wire compatibility
+
+The Iceberg REST spec does not require file deletion to be complete
```
markhoerth: Flagging this against PRD §2.2. The line "CREATE at the same identifier succeeds immediately" is the exact failure mode the PRD is asking us to prevent. PRD §2.2 quotes the Snowflake Open Catalog docs directly: "don't create a new table with the same name and location as the dropped table... a user could gain access to the original table's data when they shouldn't have permission." The PRD's position is that protecting against this shouldn't sit on the operator. The catalog should enforce it as an invariant by keeping the row in iceberg_tables with a tombstone marker, so the existing name collision check returns 409 until cleanup completes. Can we adjust the design to handle that?

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+2. File-deletion strategy is **pluggable** behind an `IcebergPurger` SPI.
+3. The default async implementation deletes every file the synchronous
+   purge would have deleted, retries transient failures, and survives
+   restarts.
+4. No change to the Iceberg REST wire protocol.
+5. Authorization runs on the **request thread**, never deferred.
+
+## 3. Non-Goals
+
+- Changing Gravitino-native soft-delete semantics.
```
markhoerth: The PRD has soft delete (R2) as the next requirement after async hard delete, built on the same row, executor, and scheduler that R1 introduces. Listing it as a non-goal here is fine for implementation order, but the row removal in §6.1 (see my comment on §9) makes recovery impossible. Once the row is gone, there's nothing to undrop from. Under this design R2 becomes a separate mechanism in V2 instead of a small extension on V1, which slips the migration story for Snowflake and Unity customers out of 1.3.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+### 6.1 Request-path interaction
+
+```java
+public void dropTable(IcebergRequestContext ctx, TableIdentifier id,
+                      boolean purgeRequested) {
+  IcebergCatalogWrapper w = catalogWrapperManager.getCatalogWrapper(ctx.catalogName());
+  if (!purgeRequested) { w.dropTable(id); return; }
+
+  TableMetadata metadata = w.loadTableMetadata(id);
+  w.dropTable(id);          // metadata-only drop in the catalog
```
markhoerth: See comments on §9 and §3. The decision to remove the row here is what creates the name reuse and R2 recovery issues.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+
+`IcebergTableOperationExecutor.dropTable` no longer knows how purge is
+implemented — it calls `purger.purgeTable(request)` and the configured
+plugin decides whether the work is synchronous, queued in a DB,
+dispatched externally, or skipped.
+
+The default plugin (`JdbcAsyncIcebergPurger`) persists a job row, returns
+immediately, and drains the queue from a background worker pool.
+
+## 5. The `IcebergPurger` SPI
```
markhoerth: What's driving the SPI design? The PRD doesn't ask for one, and §14.4 (waiting for a third-party plugin in production before stabilizing the interface) suggests we don't have one in flight. §13 argues the SPI is "strictly more general for the same core complexity," but I'd push back on that. The SPI surface, the discovery factory, the classpath loading path, and the IcebergPurgerContext are real added complexity compared to shipping one implementation. The PRD's competitive frame against Polaris is "simpler design with smaller bug surface," and the SPI moves the wrong direction. Could we ship one default async purger in V1, keep the synchronous path as a feature flag for rollback, and revisit the SPI when a second implementation has a real customer behind it?

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+          .fileIoProperties(w.fileIoProperties())
+          .requestUser(ctx.userPrincipal())
+          .build());
+}
+```
+
+Order matters: load metadata → drop catalog entry → call the purger. A
+purge job exists only for a table that is already gone from the catalog.
+
+### 6.2 Schema — `iceberg_purge_job`
```
markhoerth: If we end up keeping the row on iceberg_tables per the §6.1 / §9 discussion, do we still need a separate job table, or can the lease and retry fields live on the existing row alongside the tombstone columns? Smaller migration if we can collapse them into one.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+
+- `@AuthorizationExpression` on `dropTable` runs on the request thread,
+  unchanged.
+- Plugins use FileIO credentials snapshotted at request time. Plugins
+  that defer past credential lifetime (STS tokens, …) must refresh or
+  require static credentials — a documented plugin responsibility.
+- `iceberg_purge_job` may contain credentials in `file_io_props`; the
+  existing Gravitino DB encryption / access controls apply.
+
+## 11. Rollout
```
markhoerth: The PRD has async as the default in 1.3 for the enterprise POCs (Tier 1 timeouts commit to that release). The phased flip to jdbc-async after phase 3 here reads like the default change could land later. Want to confirm which release the default flip happens in before we lock the sequence.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+
+1. SPI + `SynchronousIcebergPurger` (parity with today). Default
+   `synchronous`. No behavior change.
+2. Land `JdbcAsyncIcebergPurger` + schema migration behind opt-in.
+3. Flip default to `jdbc-async`. `synchronous` stays as fallback.
+4. Document the SPI as a public extension point.
+
+## 12. Testing
```
markhoerth: Don't see the scale and concurrency testing from PRD §4 (R3) covered here. The Locust scenarios in gravitino-test/cloud/scale-test/, the high-file-count fixtures (medium ~50K, large ~500K files), the concurrent-drop tier, the Cucumber features for drop/soft-delete/cleanup/cleanup-failure, the timing assertion that drop returns within 5s regardless of table size, and the direct GCS/S3 listing to verify the storage prefix is empty after cleanup. Going into a separate design doc you own, or did it fall out? Want to make sure it doesn't slip, since the bug the PRD opens with is exactly the kind scale tests catch.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+   conditional updates? Need a decision before implementation.
+2. **Credential refresh in `jdbc-async`** — refresh from the catalog at
+   lease-renewal time, or require static credentials for catalogs that
+   opt into async purge?
+3. **Admin visibility** — expose an in-flight-jobs read endpoint? If
+   yes, on the SPI (`IcebergPurger#describeInFlight()`) or only on the
+   default plugin?
+4. **SPI stability** — mark `@Evolving` until at least one third-party
+   plugin is in production?
+5. **Namespace / view purge on the SPI in v1** — adding
```
markhoerth: PRD §2.2 covers this. Views go through the same mechanism (no data files, cleanup just removes the metadata.json), and schema drops cascade by tombstoning each contained table and view independently. Recovery is per-object, so undropping a schema doesn't auto-recover its contents. Happy to walk through it if useful.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,421 @@
...
+operational burden. Operators who want one write a plugin.
+
+**Enumerate files at enqueue time.** Defeats the latency goal; bloats
+rows.
+
+## 14. Open questions
+
+1. **Multi-server coordination on H2** — fall back to advisory locks or
+   conditional updates? Need a decision before implementation.
+2. **Credential refresh in `jdbc-async`** — refresh from the catalog at
```
markhoerth: Good catch, the PRD doesn't answer this either. Want to think it through together. What's the cost difference between refreshing creds at lease-renewal time vs requiring static creds for catalogs that opt into async? Leaning toward refresh so long-retention soft delete still works, but want your read before I commit it to the PRD.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: We can use register table and location to recover the table. It's possible.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: I can only make one implementation.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: We don't have a table called `iceberg_tables`. Do we mean that Jdbc catalog's table? Can we only support it in the JdbcCatalog?

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: I have concern about this semantics. It breaks the drop behavior semantics.
We can create the same identifier after dropping the identifier.  @jerryshao WDYT?

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: I will add the scaling tests.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: OK.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Ok, I can make it as the default implementation in 1.3.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,436 @@
...
+  view's `metadata.json`. The job row uses `object_type = 'VIEW'`.
+- **Namespace (schema) drops** cascade: the server enqueues an
+  independent purge job for each contained table and view. Failures and
+  retries are tracked per object, so a re-run never re-deletes an
+  already-cleaned object, and one object's failure does not block the
+  others.
+
+### 5.7 Recovery and the soft-delete (R2) seam
+
+Because the job stores `metadata_location` and the underlying data /
```
markhoerth: The math from §5.9 makes the recovery window concrete. With default poll-interval-ms=5000 and 8 parallel delete threads per job, the practical window between DROP and irrecoverable is on the order of 5 to 10 seconds for a typical table: ~5 seconds for the next worker poll, plus a fraction of a second to begin deleting reachable files. Once any file in the metadata tree is gone, register-table can no longer reconstruct the table.

For the major enterprise pitch, "you have approximately 5 seconds to undrop" is not a story. The competitive baseline for recovery is Databricks Unity Catalog at 7 days for managed tables, Snowflake Time Travel at 1 day default and up to 90 days, and Lakekeeper's configurable warehouse-level retention. PRD §3.4 sets a 24-hour default for the same reason: a recovery window has to be long enough for a human to notice the mistake and act.

Register-table at the stored metadata_location is a useful seam for R2 once retention exists, but it is not a recovery mechanism in R1 on its own. Without a retention timer holding the files for hours or days, §5.7 should not be described as a recovery surface in user-facing docs.

design-docs/async-iceberg-rest-hard-deletion.md
markhoerth: The issue is that without sufficient safeguards, a user can create the same identifier immediately after the drop and before deletion has finished. That is exactly what we must avoid.

A small extension to your existing design covers it: CREATE checks iceberg_purge_job for an outstanding job at (namespace, name) and returns 409 if one exists. The 409 applies only while a delete job is active; normal CREATE semantics resume afterward.

roryqi: If we let users use the register to recover the tables, user can rename the table if conflicted.
For example, 
user deletes  a table A with (location = "s3://123")
create a table A with (location= "s3://456")
User want to recovery the old table A. He can rename new table A to another name A'.

My concern is that we can't create the same identifier table after dropping. It breaks the drop semantics.

markhoerth: There is basically no recovery in hard delete, becuase in practice the worker has started deleting a table before recovery can be initiated. 

You can create the same identifier, it just needs to be after the first verison is deleted.  That does not break any semantics.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Do we need recovery mechanism in R1?

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Does hard deletion require recovery?

roryqi: We can't create the same identifier table during hard deletion period. It makes sense.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: We have a deferred deletion operations if users used a soft deletion.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,436 @@
...
+TableMetadata meta = TableMetadataParser.read(io, job.metadataLocation());
+Tasks.foreach(collectAllReachableFiles(meta))
+     .executeWith(deleteExecutor)
+     .retry(perFileRetries)
+     .suppressFailureWhenFinished()
+     .run(io::deleteFile);
+```
+
+A separate task renews the lease every `leaseTimeout / 3`. If the host
+dies, the lease expires and another replica reclaims the job.
```
jerryshao: How to handle server restart problem? Also, how to handle a multiple-server deployment situation?

design-docs/async-iceberg-rest-hard-deletion.md
jerryshao: > Flagging this against PRD §2.2. The line "CREATE at the same identifier succeeds immediately" is the exact failure mode the PRD is asking us to prevent. PRD §2.2 quotes the Snowflake Open Catalog docs directly: "don't create a new table with the same name and location as the dropped table... a user could gain access to the original table's data when they shouldn't have permission." The PRD's position is that protecting against this shouldn't sit on the operator. The catalog should enforce it as an invariant by keeping the row in iceberg_tables with a tombstone marker, so the existing name collision check returns 409 until cleanup completes. Can we adjust the design to handle that?

I think it makes sense to not to allow the creation of a table with the same name during deletion. But, we should have way to check this and let user know.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,436 @@
...
+  `lease_owner`       VARCHAR(128)  NULL,
+  `lease_expires_at`  BIGINT(20)    NULL,
+  `next_attempt_at`   BIGINT(20)    NOT NULL,
+  `created_at`        BIGINT(20)    NOT NULL,
+  `created_by`        VARCHAR(128)  NOT NULL,
+  `updated_at`        BIGINT(20)    NOT NULL,
+  PRIMARY KEY (`id`),
+  KEY `idx_state_next_attempt` (`state`, `next_attempt_at`)
+) ENGINE=InnoDB;
+```
```
jerryshao, line 238: Do we have a light-weighted solution other than db access? Introducing a table to manage the deletion job seems too overkill.

roryqi:
Updated the design doc to resolve the two open questions raised in review:
- **Name reuse (§5.13):** tombstone approach (Option A) — same-identifier recreate returns 409 until purge completes, closing the §2.2 recreate attack; re-register cancels the job.
- **Purge visibility (§5.14):** informative 409 message + read-only `purge-jobs` endpoint + metrics + events; `DEAD_LETTER` discoverable everywhere.

Also marked Q1/Q4 resolved and updated the schema, failure model, and task breakdown.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: I ever thought about light-weighted solution. 
We need a shared storage to solve this issue.
There are two options.
1. Database
2. Object storage
Object storage lacks atomic operation guarantee. It will make solution complex. You can see 113 line.

roryqi:
@jerryshao I have addressed the comments from the offline discussion.

jerryshao:
You should create an issue to track this PR.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+   feature flag purely for rollback, not as a parallel product surface.
+3. **Reliable deletion**: The async path deletes every file the
+   synchronous purge would have deleted, retries transient failures, and
+   survives restarts and replica failover.
+4. **Wire compatibility**: No change to the Iceberg REST wire protocol.
+5. **Request-thread authorization**: Authorization runs on the request
+   thread, never deferred.
+6. **Uniform object coverage**: Tables and namespace (schema) drops flow
+   through the same async cleanup mechanism. Views reuse the same mechanism
+   but are a planned follow-up, not part of the initial scope (§5.6).
```
jerryshao: Do we support namespace and view async deletion for now?

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+
+---
+
+## 3. Non-Goals
+
+1. **Native soft-delete / undrop semantics**: This design only delivers
+   async *hard* deletion. Full soft-delete / undrop semantics are a
+   follow-up requirement; §5.7 records the seam they build on so that work
+   stays a small extension rather than a separate V2 mechanism.
```
jerryshao: What is V2 mechanism here?

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: We don't support namespace and view async deletion for now.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+## 5. Proposal
+
+### 5.1 Overview
+
+```
+  DELETE …?purgeRequested=true
+            │
+            ▼
+  IcebergTableOperationExecutor.dropTable
+            │  X-Gravitino-Async-Purge header (default: async)
```
jerryshao: The default value is 'true' or 'async'? Also, check the sytle of http header, it should capitalize the first letter.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: It means that soft deletion mechanism. We designed the basic table to record some information. Soft deletion mechanism can extend the table.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+ worker pool (any replica)
+   claims job via FOR UPDATE SKIP LOCKED (heartbeat ownership)
+   → rebuild snapshot graph from metadata_location
+   → stream + delete every reachable file, retry w/ backoff
+   → SUCCEEDED | FAILED
+```
+
+Async is the default deletion path. The synchronous path is retained only
+as a per-request fallback that a client opts into with
+`X-Gravitino-Async-Purge: false`.
```
jerryshao: True and false are a pair, not 'async' and 'false'.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+  `object_type`       VARCHAR(16)   NOT NULL COMMENT 'TABLE|VIEW',
+  `metadata_location` VARCHAR(1024) NOT NULL,
+  `file_io_impl`      VARCHAR(256)  NOT NULL,
+  `file_io_props`     MEDIUMTEXT    NOT NULL COMMENT 'JSON',
+  `state`             VARCHAR(16)   NOT NULL COMMENT 'PENDING|RUNNING|SUCCEEDED|FAILED|CANCELLED; a RUNNING row is being deleted while its heartbeat_at stays fresh',
+  `attempts`          INT(10)       NOT NULL DEFAULT 0,
+  `max_attempts`      INT(10)       NOT NULL,
+  `last_error`        TEXT          NULL,
+  `heartbeat_at`      BIGINT(20)    NULL COMMENT 'last heartbeat from the processing worker; NULL when unclaimed',
+  `next_attempt_at`   BIGINT(20)    NOT NULL,
```
jerryshao: Do you need `attempts`, `max_attempts` and `next_attempt_at`?

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+client that needs strict synchronous deletion sends
+`X-Gravitino-Async-Purge: false`.
+
+The remaining server-side keys only tune the worker pool and retry behavior
+— operational concerns a client cannot set:
+
+| Key                                                  | Default   | Description                                                                              |
+| ---------------------------------------------------- | --------- | ---------------------------------------------------------------------------------------- |
+| `gravitino.iceberg-rest.async-purge.worker-threads`           | `4`       | Worker pool size per server. |
+| `gravitino.iceberg-rest.async-purge.delete-threads-per-job`   | `8`       | Parallelism of file deletes within a single job. |
```
jerryshao: What's the meaning/defintion of 'job' here?

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+
+The earlier draft proposed a pluggable `IcebergPurger` SPI. Review
+feedback (PR #11152) pushed back: the PRD does not ask for an SPI, no
+second implementation is in flight, and the competitive frame against
+Polaris is "simpler design with smaller bug surface." We re-evaluated and
+chose a single async implementation with a synchronous rollback flag.
+
+| Approach | Pros | Cons | Decision |
+|----------|------|------|----------|
+| Synchronous only (status quo) | Simplest; strongest "deleted means gone" guarantee | Exceeds HTTP timeouts, saturates Jetty, no retry/audit on large tables | **Rejected** — the problem we are solving |
```
yuqi1129: Format the table.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+
+The remaining server-side keys only tune the worker pool and retry behavior
+— operational concerns a client cannot set:
+
+| Key                                                  | Default   | Description                                                                              |
+| ---------------------------------------------------- | --------- | ---------------------------------------------------------------------------------------- |
+| `gravitino.iceberg-rest.async-purge.worker-threads`           | `4`       | Worker pool size per server. |
+| `gravitino.iceberg-rest.async-purge.delete-threads-per-job`   | `8`       | Parallelism of file deletes within a single job. |
+| `gravitino.iceberg-rest.async-purge.poll-interval-ms`         | `5000`    | Worker poll interval. |
+| `gravitino.iceberg-rest.async-purge.batch-size`               | `16`      | Jobs claimed per tick. |
```
jerryshao: What's the meaning of `batch-size` and `tick` here?

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -0,0 +1,716 @@
...
+This closes the PRD §2.2 attack directly: a same-name / same-location
+recreate can no longer expose the dropped table's data, because the
+recreate is refused until the data files are deleted. Telling *whether* a
+given identifier is mid-purge, done, or failed is the read-side
+concern handled in §5.14.
+
+The same rule will apply to **views** (`createView` / `replaceView` → `409`
+while a `VIEW` job is active) once view cleanup ships (§5.6). A **namespace**
+carries no data, so it can be recreated freely; any table created inside it
+is still guarded individually by its own job row.
```
jerryshao: Do we support view and namespace purge for now?

roryqi: We don't support them.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: OK, I can correct it.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: default is none. It means `async`. I can change this.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: I will remove this config options. We can use threads number.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: This is required for the transform from the  RUNNING state to FAILED state. 
`next_attempt_at` is related about jitter and backoff. I can remove it in the first version.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: The name is confused. I will change this name to `deleter-threads`. Add the proper description to explain this configuration option.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Fixed.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Iceberg used `X-Iceberg-Access-Delegation`.

roryqi:
I addressed the comments from the offline meeting. cc @jerryshao 

- **Renamed the DB table** `iceberg_purge_job` → `iceberg_cleanup_job` — "drop"/"purge" was misleading; the row tracks deferred *cleanup* work.
- **Repeated `dropTable` is now defined as idempotent**: the catalog entry is removed before the first `204`, so a second `DELETE …?purgeRequested=true` returns `404` and enqueues no second job (§5.3, §5.7).
- **Investigated recreate behavior**: a `createTable` with no explicit `location` reuses the same base prefix via `defaultWarehouseLocation(identifier)`, which justifies the §5.7 `409` tombstone; documented why the new table's UUID-named files never collide with the purge.
- **Added a `last_error` column** so a `FAILED` job's reason is queryable in the row, not only in worker logs (§5.4, §5.6, §5.9).
- **Decoupled the heartbeat from deletion**: a separate background task refreshes `heartbeat_at`, so a worker blocked in a long bulk delete keeps its lease (§5.5).
- **Switched to bulk deletion** (`CatalogUtil.deleteFiles` → `SupportsBulkOperations`) instead of per-file deletes.
- **Handle a full delete queue** via a bounded `ThreadPoolExecutor` + `CallerRunsPolicy` back-pressure — no unbounded queue, no dropped work (§5.5).
- **Tuned defaults**: `worker-threads` 4→2, `delete-threads` 16→4, added `delete-batch-size` (1000); renamed `terminal-retention-hours` → `retention-hours` and raised it to 720 (30 days).

roryqi:
> I addressed the comments from the offline meeting. cc @jerryshao
> 
> * **Renamed the DB table** `iceberg_purge_job` → `iceberg_cleanup_job` — "drop"/"purge" was misleading; the row tracks deferred _cleanup_ work.
> * **Repeated `dropTable` is now defined as idempotent**: the catalog entry is removed before the first `204`, so a second `DELETE …?purgeRequested=true` returns `404` and enqueues no second job (§5.3, §5.7).
> * **Investigated recreate behavior**: a `createTable` with no explicit `location` reuses the same base prefix via `defaultWarehouseLocation(identifier)`, which justifies the §5.7 `409` tombstone; documented why the new table's UUID-named files never collide with the purge.
> * **Added a `last_error` column** so a `FAILED` job's reason is queryable in the row, not only in worker logs (§5.4, §5.6, §5.9).
> * **Decoupled the heartbeat from deletion**: a separate background task refreshes `heartbeat_at`, so a worker blocked in a long bulk delete keeps its lease (§5.5).
> * **Switched to bulk deletion** (`CatalogUtil.deleteFiles` → `SupportsBulkOperations`) instead of per-file deletes.
> * **Handle a full delete queue** via a bounded `ThreadPoolExecutor` + `CallerRunsPolicy` back-pressure — no unbounded queue, no dropped work (§5.5).
> * **Tuned defaults**: `worker-threads` 4→2, `delete-threads` 16→4, added `delete-batch-size` (1000); renamed `terminal-retention-hours` → `retention-hours` and raised it to 720 (30 days).

Changed default header value to synchorous

jerryshao (approved):

PR #11174, opened 2026-05-20
[#11088] feat(authz): Optimize JCasbin cache consistency and role checks

### LINKED ISSUE #11088: [Improvement] Route isSelf(ROLE) through the version-validated userRoleCache

In `JcasbinAuthorizer.isSelf(EntityType, NameIdentifier)`, the `ROLE` branch resolves the user's role assignments by going straight to `EntityStore.relationOperations().listEntitiesByRelation(ROLE_USER_REL, ...)`:

```java
List<RoleEntity> entities =
    entityStore
        .relationOperations()
        .listEntitiesByRelation(
            SupportsRelationOperations.Type.ROLE_USER_REL,
            userNameIdentifier,
            Entity.EntityType.USER);
```

This bypasses the version-validated `userRoleCache` / `groupRoleCache` that the rest of `loadRolePrivilege` already uses. As a result, `isSelf(ROLE)` issues an extra DB query for every call, even when the same user's role list was just resolved a few microseconds earlier on the same request.

### How should we improve?

PR #10996 introduced the version-validated user-role / group-role caches but intentionally kept `isSelf(ROLE)` on the pre-existing direct-entity-store path to keep the diff focused. Switching `isSelf(ROLE)` to reuse the cache path would:

- eliminate the redundant query when role-management APIs are called frequently;
- keep `isSelf` semantics consistent with `authorize` / `isOwner`.

Suggested refactor for `isSelf(ROLE)`:

1. Call `loadUserInfo` to get the user id (and benefit from per-request dedup via `AuthorizationRequestContext`).
2. Call `loadUserRoles` for the direct role ids and `loadGroupRoles` per group for inherited role ids.
3. Check the target role id against the union.

Falls back to deny on any exception, same as today.

Acceptance criteria:

- `isSelf(ROLE)` no longer calls `entityStore.relationOperations().listEntitiesByRelation` directly.
- Existing `TestJcasbinAuthorizer.testIsSelfRoleViaGroup` and other `isSelf(ROLE)` tests still pass.
- Add a unit test that asserts repeated `isSelf(ROLE)` calls in the same request don't trigger redundant role-list DB queries.

yuqi1129:

This PR improves JCasbin authorization cache behavior and reduces DB access on hot authorization paths.

Main changes:

- Batch the current user, groups, user-role relations, group-role relations, and role version probes into one mapper query for `authorize()`.
- Route `isSelf(ROLE)` through the same version-validated user/group role caches used by `authorize()`.
- Add per-request cache priming in `AuthorizationRequestContext` to avoid repeated user/group/role probes within one request.
- Add local cache invalidation hooks for user-role and group-role relation changes.
- Add metadata-id / owner cache lookup helpers and poller-backed invalidation for multi-node deployments.
- Add mapper/service tests and JCasbin authorizer tests for cache reuse, invalidation, and authorization behavior.

After disabling the previous global cache behavior for multi-node correctness, JCasbin authorization started doing extra DB probes on hot paths. This PR restores most of the lost performance by batching version checks and making `isSelf(ROLE)` reuse the same version-validated caches as normal authorization.

It also makes local cache invalidation explicit when user-role or group-role relations change, which avoids stale role bindings after grant/revoke operations.

Fix: #11088

No.

- `./gradlew :core:spotlessApply :server-common:spotlessApply`
- `./gradlew :core:test --tests org.apache.gravitino.storage.relational.mapper.provider.base.TestAuthMappers -PskipITs`
- `./gradlew :server-common:test --tests org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer -PskipITs`
- `./gradlew :core:test --tests org.apache.gravitino.hook.TestAccessControlHookDispatcher -PskipITs`
- `./gradlew :clients:client-java:test --tests org.apache.gravitino.client.integration.test.authorization.FunctionAuthorizationIT -PskipDockerTests=false`


yuqi1129:
Table: 1000
user: 1000
role: 2000


**Global cache enable:**
image

**Global cache off:**
image



I'm struggling to find out the degradation.

yuqi1129:
@roryqi 
Could you help me review this one? Thanks.

roryqi (commented):
I found two cache-consistency issues that should be addressed before merging:

1. `selectChangedOwners` can permanently skip owner insert rows. The query now returns rows ordered by `(updated_at, id)` with `LIMIT 1000`, while `JcasbinChangePoller` advances `ownerPollHighWaterInsertId` to the maximum `id` seen in that page. Owner inserts still use `System.currentTimeMillis()` for `updated_at`, while owner soft-deletes use DB time. With clock skew or otherwise non-monotonic insert timestamps, a page can contain a higher `id` row while a lower `id` row remains outside the limited page and has `updated_at <= ownerPollHighWaterUpdatedAt`. After the insert cursor is advanced past it, the next poll filters it out by both cursors, so peer nodes never invalidate that `ownerRelCache` entry until TTL. Please either page the insert stream by `id` separately, or only advance the insert cursor across rows that are guaranteed to have been fully consumed in `id` order.

2. `metadataIdCache` keys are built before metadata-name case normalization. `resolveMetadataId` uses the incoming `MetadataObject` to build the cache key, but `MetadataIdConverter.getID` normalizes the `NameIdentifier` according to catalog capability before reading the entity. For a case-insensitive catalog, a request using mixed-case names can populate a cache alias that later invalidation paths, which are built from canonical/stored names, will not evict on rename/drop. That can leave stale metadata ids and stale authorization decisions until TTL. The cache key should be derived from the same normalized/canonical identifier used for lookup, or invalidation must cover all case aliases.

I verified `git diff --check` and `TestJcasbinChangePoller`; the mapper test did not run locally because this PR head hit an unrelated `lineage/server` compile failure before reaching `TestAuthMappers`.

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
```diff
@@ -138,6 +140,7 @@ public Group revokeRolesFromGroup(String metalake, List<String> roles, String gr
       throws NoSuchGroupException, IllegalRoleException, NoSuchMetalakeException {
     Group revokedGroup = dispatcher.revokeRolesFromGroup(metalake, roles, group);
     notifyRoleGroupRelChange(metalake, roles);
```
roryqi: Why do we use two methods instead of one method?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
```diff
@@ -87,16 +89,19 @@ public JcasbinChangePoller(
...
+   * advance {@code updated_at}; an id-only cursor would miss them on peer nodes.
    */
   public void start() {
-    ownerPollHighWaterId =
+    ChangedOwnerInfo maxOwnerChange =
+        SessionUtils.getWithoutCommit(
+            OwnerMetaMapper.class, OwnerMetaMapper::selectMaxChangedOwner);
+    if (maxOwnerChange != null) {
+      ownerPollHighWaterUpdatedAt = maxOwnerChange.getUpdatedAt();
+      ownerPollHighWaterUpdatedAtId = maxOwnerChange.getId();
```
roryqi, line 106: Must we have `UpdateAt` and `UpdateAtId`?

core/src/main/java/org/apache/gravitino/storage/relational/po/auth/AuthSubjectVersion.java
```diff
@@ -0,0 +1,80 @@
...
+ * </ul>
+ */
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class AuthSubjectVersion {
+
+  /** Discriminates the row's underlying subject kind. */
+  public enum Kind {
```
roryqi: Could we simplify the kind to two types?

roryqi (commented):
Do we have ways to simplify the code?

core/src/main/java/org/apache/gravitino/hook/AccessControlHookDispatcher.java
yuqi1129: Good catch — the two paired calls (`notifyRoleXxxRelChange` + `notifyXxxRoleRelChange`) were always issued together at every grant/revoke site. Combined them into `notifyUserRoleBindingChange` / `notifyGroupRoleBindingChange` wrappers. The single-notify methods remain for role-only operations (createRole, deleteRole, etc.) that have no user/group counterpart.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
yuqi1129: Both are needed. `updated_at` is BIGINT milliseconds, so multiple soft-delete updates within the same millisecond would tie. `updatedAtId` is the tiebreaker — together `(updatedAt, updatedAtId)` form one logical composite cursor. A pure `id` cursor does not work because soft-delete updates reuse the existing row id and only advance `updated_at`. Added a field-level comment to make this clearer.

core/src/main/java/org/apache/gravitino/storage/relational/po/auth/AuthSubjectVersion.java
yuqi1129: I looked into collapsing to 2 kinds (SUBJECT / ROLE) but it would actually be more complex. The consumer in `JcasbinAuthorizer#batchLoad` takes 4 distinct actions:

- `USER` → single `foundUser`
- `GROUP` → `foundGroups` map keyed by name
- `USER_ROLE` → `userRoleIds` set
- `GROUP_ROLE` → `groupRoleIdsByGroupId` map keyed by `parentGroupId`

Collapsing to 2 kinds would require an extra `subjectKind: USER/GROUP` field to disambiguate — replacing one enum field with two. The SQL has 4 `UNION ALL` branches that map 1:1 to the 4 kinds, so keeping 4 kinds is the simpler design.

roryqi: Could it be 
```
user {
string name
String id
List roles
}
```
?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
roryqi: Is it updated_at enough? It seems simpler?

yuqi1129: It seems that we need `updated_at` and `id` both when there are plenty of rows with the same `updated_at` , for example, If we drop a catalog, there can be as many as 10k metadata object owner changes at the same time. 

The pull logic is that we add a limit syntax to avoid memory and lock problems, so the ID is necessary.

core/src/main/java/org/apache/gravitino/storage/relational/po/auth/AuthSubjectVersion.java
yuqi1129: The original code is what you described above.  To reduce the number of database accesses to one (original at least 3), I changed it to what it is currently.

I have updated class name, strucutre and more comments.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
roryqi: OK for me.

roryqi (approved):

PR #11176, opened 2026-05-20
[#11027][#11178] fix(lance): Refresh Lance table schema from dataset on load

### LINKED ISSUE #9501: [Bug report] The structure of the lance table cannot be obtained.
### Version

tag v1.1.0

### Describe what's wrong

Here is how it is done:
1. Create the "lance" table using gravitino, and check that the "lance" table is displayed correctly through the web interface. 
2. Use Python LanceDB v0.21.1 to create a local Lance table and write data as shown in the following figure:

Image

Then, the existing "Lance" table was registered through the Gravitino API. By using the Gravitino web interface, it was found that the registered "Lance" table could be viewed normally, but the structure information of the "Lance" table could not be obtained. As shown in the following figure:

Image

The gravitino system has no error messages, and the lance-rest-server system also has no error messages.

### How to reproduce

gravitino env:
gravitino version 1.1.0 
jdk 17

mac python:
python 3.12
lanceDb 0.21.1

### Additional context

_No response_

### LINKED ISSUE #11027: [Bug report] The column of Lance table is empty in Gravitino Web UI for the table created by compute engine
### Version

main branch

### Describe what's wrong

When creating a Lance table through Spark SQL (CTAS: _CREATE TABLE ... AS SELECT_) via lance-rest-server, the Gravitino Web UI displays empty columns for the table, even though the actual Lance dataset on S3 contains data with the correct
  schema.                                                                                                                                                                                                                                          
                                                                               
  In contrast, directly calling the lance-rest-server` /v1/table/{id}/create` API (with arrowStreamBody) results in the columns being correctly displayed in Gravitino Web UI.                                                                       
                                                                               
  The root cause is that Spark CTAS uses StagingTableCatalog, which follows a staged creation flow:                                                                                                                                                
                                                                                                                                                                                                                                                 
1.    `stageCreate()` → calls `declareTable()` → mapped to lance-rest-server `/v1/table/{id}/create-empty` → only registers table metadata (name + location) in Gravitino with empty columns and _lance.create-empty=true_ property
2.  Spark writes data to S3 → `StagedCommit` directly writes to S3 via Lance Java SDK (CommitBuilder), embedding the actual schema into the Lance dataset                                                                                           
3. `StagedCommit.commit()` → finishes writing to S3, but never calls back to lance-rest-server to update the Gravitino metadata with actual columns
                                                                                                                                                                                                                                                   
  The staged path decouples metadata registration from data writing, and `StagedCommit.commit()` only writes to S3 without updating Gravitino's stored schema.

No error or exception is thrown. The table is created successfully and data can be read via Spark. The issue is that the Gravitino metadata store has empty columns, so the Web UI and any Gravitino client API (loadTable) show no column       
  information.

### How to reproduce

1. Set up Gravitino with lance-rest-server and a Lance catalog with S3 storage options                                                                                                                                                         
2. Connect Spark to lance-rest-server using the Lance Spark connector                                                                                                                                                                            
3. Execute a CTAS statement: _CREATE TABLE test_table USING lance AS SELECT 1 AS id, 'hello' AS name_
4. Check the table in Gravitino Web UI → columns are empty                                                                                                                                                                                       
5. Query the table via Spark → data and columns are correctly returned       
6. Compare: directly calling POST `/v1/table/{id}/create` with Arrow IPC stream body → columns are displayed correctly in Web UI

### Additional context

**Proposed fix:** Override `loadTable()` in LanceTableOperations to implement a self-healing mechanism: when _lance.create-empty=true_ is detected, read the actual schema from the Lance dataset on S3, update the Gravitino entity store with the      
  correct columns, and clear the _lance.create-empty_ flag. This is a one-time self-heal — subsequent `loadTable()` calls return the correct metadata via the normal path.                                                                             
                                                                                                                                                                                                                                                   
**Key files involved:**                                                                                                                                                                                                                              

- `catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java`

-  `catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableDelegator.java`                                                                                                                         

-  `lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java`

### LINKED ISSUE #11178: [Bug report] Lance describeTable always reports is_only_declared=true even after a declared table is materialized
### Version

main branch

### Describe what's wrong

In the Lance REST namespace, `DescribeTableResponse.is_only_declared` is populated by reading the `lance.declared` table property:

```java
// GravitinoLanceTableOperations.describeTable()
if (checkDeclared) {
    response.setIsOnlyDeclared(
        Boolean.parseBoolean(table.properties().getOrDefault(LANCE_TABLE_DECLARED, "false")));
}
```

The `lance.declared` property is written exactly once inside `declareTable()`, and is registered as **immutable** in `LanceTableDelegator`, so it can never change afterwards.

Per the Lance namespace spec (v0.7.0), `is_only_declared` is a **current state** — "the table is only declared in the namespace and has no underlying storage data yet" — and is expected to flip to `false` once the table is materialized by the first data write.

Because Gravitino derives it from an immutable property, a table that was declared via `declareTable` and then populated with real data will **permanently report `is_only_declared=true`**, which is incorrect.

N/A — semantic bug, no exception is thrown.

### How to reproduce

1. Declare a table via the Lance REST `declare` endpoint (`lance.declared=true`).
2. Write actual data into the table's location (e.g. via lance-ray).
3. Call `describeTable` with `check_declared=true`.
4. Observe `is_only_declared=true` even though the table now has storage data.

### Additional context

Discovered while reviewing #11060. This issue is related to #11176 (which refreshes the table column schema from the Lance dataset on `loadTable`), but addresses a different aspect: the `is_only_declared` field in the `describeTable` REST response, derived from the immutable `lance.declared` property. The two are related but distinct concerns and may not necessarily be addressed in the same PR. A correct fix likely requires `describeTable` (when `check_declared=true`) to probe the table location for an actual Lance manifest instead of reading the static property.

yuqi1129:

This PR updates Lance table loading in the Generic Lakehouse catalog to repair Gravitino table metadata from the underlying Lance dataset when needed.

The refresh behavior is:
- `DECLARED_AND_EMPTY` by default: refresh tables marked with `lance.declared=true` or tables with empty stored columns.
- `VERSION_CHECK`: open the Lance dataset during `loadTable`, compare its version with the stored `lance.version`, and refresh columns when the version changes.

When a declared table has been materialized, the refresh persists its actual columns and dataset version and removes `lance.declared`. As a result, Lance REST `describeTable` reports `is_only_declared=false` after the first data write.

This PR also supports the Lance `loadDetailedMetadata` option and documents the new catalog property `lance.schema-refresh-mode`.

Spark staged create/declare flows register a table in Gravitino before writing the actual Lance dataset. After the write, Gravitino can still contain empty columns and the immutable `lance.declared=true` marker. This causes the Web UI and client APIs to return an empty schema and causes `describeTable` to incorrectly report that the materialized table is still declaration-only.

Refreshing the stored metadata from the Lance dataset during `loadTable` repairs both states.

Fix: #11027
Fix: #11178
Fix: #9501

Yes.

A new Generic Lakehouse catalog property is added:
- `lance.schema-refresh-mode`
  - `DECLARED_AND_EMPTY`: the default; repairs declared tables and tables with empty stored columns.
  - `VERSION_CHECK`: refreshes table columns when the Lance dataset version changes.

Materialized declared tables now return their actual schema and report `is_only_declared=false`. No generic Table API parameter is added.

- Added unit tests for declared-table repair, empty stored schemas, dataset version changes, genuinely empty datasets, missing locations, dataset open failures, and concurrent metadata repair.
- Added Lance REST and Spark integration coverage for schema refresh and declared-table state.
- `./gradlew :catalogs:catalog-lakehouse-generic:spotlessApply :lance:lance-rest-server:spotlessApply :catalogs:catalog-lakehouse-generic:test --tests org.apache.gravitino.catalog.lakehouse.lance.TestLanceTableOperations --tests org.apache.gravitino.catalog.lakehouse.generic.TestPropertiesMetadata :lance:lance-rest-server:compileTestJava -PskipWeb=true -PskipDockerTests=true`
- `./gradlew :docs:build`


FANNG1 (commented):
I think this lazy refresh path needs to be idempotent under concurrent `loadTable` calls. The refresh decision and generated `TableChange`s are based on the initial `super.loadTable()` snapshot, but `super.alterTable()` applies those changes against the current entity in `EntityStore.update`. If two callers load the same declared/empty table at the same time, the first call can add the refreshed columns and remove `lance.declared`; the second call then applies its stale `addColumn` changes to an entity that already has those columns and fails with `Column ... already exists`. The same race can happen in `VERSION_CHECK` when two callers refresh the same old `lance.version`. Since this runs from `loadTable`, concurrent reads can intermittently fail. Could we make the metadata repair idempotent inside the update path, or re-check the current stored columns/version before applying the add-column changes?

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +120,62 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
...
+      }
+    }
+    addColumnChanges(changes, columns);
+    changes.add(
+        TableChange.setProperty(
+            LanceConstants.LANCE_TABLE_VERSION, String.valueOf(datasetVersion)));
+    changes.add(TableChange.removeProperty(LanceConstants.LANCE_TABLE_DECLARED));
+    // This is a metadata-only repair. Calling super avoids replaying these column changes against
+    // the Lance dataset that we just read from.
+    return super.alterTable(ident, changes.toArray(new TableChange[0]));
```
FANNG1: **Concurrent `loadTable` can fail with "Column already exists" (TOCTOU).** The snapshot read (`super.loadTable`), the `emptySchema`/delete-list computation, and this `super.alterTable` commit are not in one transaction. If two requests load the same unrepaired table concurrently (empty/declared table, or any `VERSION_CHECK` table whose version changed), both build `[deleteColumn(old)…, addColumn(new)…]`. After thread A commits, thread B's `store.update` runs `applyChanges` against the already-repaired entity, where its `addColumn` hits an existing column and `applyColumnChanges` throws `IllegalArgumentException("Column X already exists")`. That unchecked exception escapes `loadTable` (the `try/catch` above only wraps the dataset open/extract, not this call), so a plain read fails; a `store.update` retry would re-collide. Consider re-checking state inside the update function, or tolerating already-present columns.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +120,62 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
         catalogProperties == null ? Map.of() : ImmutableMap.copyOf(catalogProperties);
   }
 
+  @Override
+  public Table loadTable(NameIdentifier ident) throws NoSuchTableException {
```
FANNG1, line 164: **`dropTable`/`purgeTable` now inherit this refresh path.** Both call `loadTable(ident)` but only need `location`. With this override, dropping a table in `VERSION_CHECK` mode (or an empty/declared table) opens the dataset and may issue `super.alterTable` to "repair" a table that is about to be deleted, and opens the dataset twice (once here, once in `Dataset.drop`). Worse, if that repair write fails, `dropTable`'s generic `catch (Exception)` rewraps it as `"Failed to drop..."`, so a transient refresh error can block a drop that previously succeeded. Suggest having `dropTable`/`purgeTable` call `super.loadTable(ident)` (as `createTable`'s `EXIST_OK` path already does).

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +120,62 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
...
+    Table table = super.loadTable(ident);
```
FANNG1, line 165: **Altitude: the refresh's beneficiary isn't the path it taxes most.** The native Lance client's dataset-open path (`DatasetBuilder::from_namespace`, and the Append/Overwrite write path) sends `DescribeTableRequest { ..Default::default() }` (`load_detailed_metadata = None`) and only reads `response.location` — it opens the dataset itself for the real schema (`resolve_table_location` even sets `Some(false)`); `load_detailed_metadata = Some(true)` appears only in tests. So placing the refresh in the shared `loadTable` is pure overhead for the Lance `describeTable` path (which discards the response schema), while its real beneficiary is the Gravitino relational `loadTable` path (#11027). Consider driving it by intent instead of overriding the universal `loadTable`: thread `load_detailed_metadata` through `describeTable` (false/unset → no dataset open) and keep the relational-path refresh in a dedicated `loadTable(ident, refreshSchema)`-style entry. This also resolves the `dropTable`/`purgeTable` concern naturally.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +120,62 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
...
+    String location = table.properties().get(Table.PROPERTY_LOCATION);
+    if (StringUtils.isBlank(location)) {
+      return table;
+    }
+
+    Map<String, String> storageOptions =
+        LancePropertiesUtils.resolveLanceStorageOptions(catalogProperties, table.properties());
+    Column[] columns;
+    long datasetVersion;
+    try (Dataset dataset = openDataset(location, storageOptions)) {
```
FANNG1, line 195: **`RootAllocator` leak amplified onto the read path.** `openDataset` builds with `.allocator(new RootAllocator())`, which sets `selfManagedAllocator = false` in `OpenDatasetBuilder`, and `Dataset.close()` only closes the allocator when it is self-managed. So each `openDataset` call leaks a `RootAllocator` (off-heap memory). This pattern pre-exists in `handleLanceTableChange`/`createTableInternal`, but the PR now puts `openDataset` on the read path — in `VERSION_CHECK` mode every `loadTable` leaks one. Either close the allocator in a `finally`, or drop `.allocator(...)` so the builder self-manages and closes it on `Dataset.close()`.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +120,62 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
...
   }
 
+  @Override
+  public Table loadTable(NameIdentifier ident) throws NoSuchTableException {
+    Table table = super.loadTable(ident);
+    // Spark staged create can write the actual schema only to the Lance dataset path. Refresh
+    // Gravitino metadata when the stored table is declared-only, empty, or configured to track
+    // Lance dataset versions.
+    boolean declaredOnly = isDeclaredOnly(table);
+    boolean emptySchema = table.columns().length == 0;
```
FANNG1, line 170: **Minor: a genuinely-empty Lance dataset re-repairs on every load.** If the dataset truly has 0 columns, `emptySchema` stays `true`, so every load reopens the dataset and re-issues `super.alterTable` (setting version, removing the declared flag again) even though nothing changed. Edge case, but it's an unnecessary write per read.

FANNG1 (commented):
Thanks for the fix! A few correctness/robustness concerns plus one design (altitude) point, left inline. Severity-ordered: the `super.alterTable` concurrency issue and the `RootAllocator` leak are the most impactful.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Good point on the altitude. The refresh is intentional on the shared loadTable path because the main consumer (Spark DataSource V2) needs the columns immediately. For the native Lance client describeTable path the overhead is real but small. Tracking a lightweight variant that skips the refresh for metadata-only callers as a follow-up optimization.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +120,62 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
...
+          ident,
+          e);
+      return table;
+    }
+
+    List<TableChange> changes = new ArrayList<>();
+    if (!emptySchema) {
+      for (Column column : table.columns()) {
+        changes.add(TableChange.deleteColumn(new String[] {column.name()}, true));
+      }
```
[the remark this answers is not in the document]
yuqi1129: Fixed: added deleteColumn(ifExists=true) for each incoming column before addColumn when emptySchema=true, making the repair idempotent under concurrent loadTable calls.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -350,6 +413,59 @@ private org.apache.arrow.vector.types.pojo.Schema convertColumnsToArrowSchema(Co
...
+  }
+
+  private void addColumnChanges(List<TableChange> changes, Column[] columns) {
+    for (Column column : columns) {
+      changes.add(
+          TableChange.addColumn(
+              new String[] {column.name()},
+              column.dataType(),
+              column.comment(),
+              null,
```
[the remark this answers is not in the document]
yuqi1129: Fixed: using the 7-arg addColumn overload and passing column.defaultValue() as the last argument instead of null.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -350,6 +413,59 @@ private org.apache.arrow.vector.types.pojo.Schema convertColumnsToArrowSchema(Co
...
+              column.nullable(),
+              column.autoIncrement()));
+    }
+  }
+
+  private Column[] extractColumns(org.apache.arrow.vector.types.pojo.Schema arrowSchema) {
+    return arrowSchema.getFields().stream()
```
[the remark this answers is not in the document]
yuqi1129: Fixed: added import for org.apache.arrow.vector.types.pojo.Schema and replaced FQN with the simple name throughout.

catalogs/catalog-lakehouse-generic/src/test/java/org/apache/gravitino/catalog/lakehouse/lance/TestLanceTableOperations.java
```diff
@@ -100,6 +118,162 @@ public void testCreateTableWithInvalidMode() {
...
+              Function<TableEntity, TableEntity> updater = invocation.getArgument(3);
+              return updater.apply(tableEntity);
+            });
+
+    Dataset dataset = mock(Dataset.class);
+    when(dataset.getSchema())
+        .thenReturn(
+            new org.apache.arrow.vector.types.pojo.Schema(
+                List.of(
+                    Field.nullable("id", new ArrowType.Int(32, true)),
```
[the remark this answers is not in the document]
yuqi1129: Fixed: added import for org.apache.arrow.vector.types.pojo.Schema and replaced FQNs in the test file.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Fixed: added defensive deleteColumn(ifExists=true) for each incoming column when emptySchema=true so concurrent repairs are idempotent.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Fixed: dropTable and purgeTable now call super.loadTable so the schema-refresh path is not triggered for operations that only need location metadata.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Fixed: removed .allocator(new RootAllocator()) from both openDataset and Dataset.write so the builder manages its own allocator (selfManagedAllocator=true) and Dataset.close() releases the off-heap memory.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Fixed: added an early return when columns.length == 0 so a genuinely-empty dataset does not write a no-op repair that would re-trigger on every subsequent load.

FANNG1:
The current implementation makes Lance REST `describeTable` call the Gravitino `loadTable` path unconditionally, which now may open the Lance dataset and refresh/write Gravitino table columns. This seems too heavy for the default Lance REST describe path.

Gravitino Lance REST server does not support the Lance commit API today, so in most cases `describeTable` only needs to return lightweight metadata such as location, table properties, and storage options. Schema/detail loading should be tied to `DescribeTableRequest.load_detailed_metadata == true`. Even then, detailed metadata may include information outside Gravitino-managed metadata, so that behavior should probably be designed separately and is beyond this PR’s scope.

Could we keep the default describe path lightweight and avoid triggering schema refresh unless `load_detailed_metadata` is explicitly requested?

yuqi1129:
1. Add catalog properties `lance.lance.schema-refresh-mode` to control when Gravitino will refresh the schema of the Lance table. 
2. For Lance REST describe table API, if `loadDetailedMetadata` of `DescribeTableRequest` is false, we will not load the details of the table, such as column definition.

yuqi1129:
I looked into this more. The concern is valid: `load_detailed_metadata=false` currently only suppresses the schema in the response, but the Gravitino-backed implementation still goes through `loadTable`, and Lance `loadTable` may open the dataset and refresh persisted columns.

I considered a few ways to make the false path lightweight:

1. Add a narrow catalog capability such as `loadTableMetadata` and let the generic Lance catalog bypass schema refresh. This works technically, but it adds a new abstraction only for this REST describe path and feels awkward.
2. Read Gravitino table metadata directly from the entity store in the Lance REST implementation. This would avoid `loadTable`, but it assumes the Lance REST server and Gravitino server share the same JVM/store access, which is not a safe deployment assumption.
3. Add a metadata-only method to the public `TableCatalog` API. This is cleaner conceptually, but it is a broader API/design change and should not be rushed as part of this Lance upgrade/schema-refresh PR.

Given those tradeoffs, I would rather not add an inelegant lightweight describe implementation in this PR. We can keep the current behavior for now and design a proper metadata-only table load path separately if Lance REST needs a true lightweight `describeTable` path.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -113,6 +130,55 @@ public void setCatalogProperties(Map<String, String> catalogProperties) {
...
+      return table;
+    }
+
+    // A genuinely-empty Lance dataset has no schema to repair; skip to avoid a metadata write
+    // that would immediately repeat on the next load because emptySchema remains true.
+    if (columns.length == 0) {
+      return table;
+    }
+
+    return repairTableMetadata(ident, columns, datasetVersion);
```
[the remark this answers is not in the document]
yuqi1129: Fixed: metadata repair is now best-effort. If repair fails after loading the stored table and Lance schema, loadTable logs the failure and returns the stored Gravitino table metadata.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceTableOperations.java
```diff
@@ -96,10 +96,15 @@ public Response describeTable(
       boolean shouldCheckDeclared =
           Optional.ofNullable(checkDeclared)
               .orElse(request != null && Boolean.TRUE.equals(request.getCheckDeclared()));
+      // loadDetailedMetadata defaults to true when absent so callers that omit the field
+      // still receive the full schema in the response.
+      boolean shouldLoadDetailedMetadata =
+          request == null || !Boolean.FALSE.equals(request.getLoadDetailedMetadata());
       DescribeTableResponse response =
```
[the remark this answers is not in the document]
yuqi1129: Fixed: added REST coverage for DescribeTableRequest.loadDetailedMetadata=false and verified the false flag is propagated while the response schema remains absent.

yuqi1129:
@FANNG1 Can you plz help to take a look

FANNG1:
Thanks for checking this. I think there are still two related semantics issues here:

1. `load_detailed_metadata=false` still does not keep the Lance REST describe path lightweight. The REST layer now passes the flag down and suppresses `schema` in the response, but `GravitinoLanceTableOperations.describeTable` still calls `catalog.asTableCatalog().loadTable(...)` unconditionally. With this PR, the Generic Lance `loadTable` path may open the Lance dataset and refresh/write Gravitino table columns for declared/empty-schema tables, and also for `VERSION_CHECK`. So the expensive part still happens before we decide whether to set `schema` in the response. Also, the REST layer currently defaults absent `load_detailed_metadata` to true. For Gravitino Lance REST, I think the safer default is lightweight describe, and detailed metadata should only be loaded when `Boolean.TRUE.equals(request.getLoadDetailedMetadata())`.

2. The `DECLARED_ONLY` mode name/semantics are a bit unclear. The current implementation refreshes not only tables with `lance.declared=true`, but also any table whose stored Gravitino columns are empty. That may be useful for register/staged-create repair, where Gravitino has empty column metadata but the Lance dataset already has a real schema. However, the mode name reads like it only applies to declared tables. Could we either rename the mode or document explicitly that `DECLARED_ONLY` also repairs empty stored columns? One related edge case is a genuinely zero-column Lance dataset: because stored columns remain empty, `loadTable` will keep opening the dataset on every load and return without recording that it has already checked it.

Could we keep default Lance REST describe lightweight, and clarify the default schema-refresh mode semantics around empty stored columns?

FANNG1:
Thanks for looking into this. I agree that fully fixing the lightweight Lance REST `describeTable(load_detailed_metadata=false)` path is not straightforward in this PR.

The root issue is that Gravitino REST and Lance REST currently share the same `TableCatalog.loadTable` path. Gravitino REST needs the full table schema and may reasonably trigger schema loading/repair from the underlying storage, while Lance REST often only needs a lightweight probe for location/properties/storage options. Without a metadata-only table probing API, Lance REST cannot cleanly avoid the schema-loading/repair behavior attached to `loadTable`.

I created a follow-up issue to track the broader API/design improvement: #11442. I think it is reasonable to keep this PR focused on the Lance schema refresh behavior, and handle the true lightweight describe path separately through that issue.

yuqi1129:
> Thanks for checking this. I think there are still two related semantics issues here:
> 
> 1. `load_detailed_metadata=false` still does not keep the Lance REST describe path lightweight. The REST layer now passes the flag down and suppresses `schema` in the response, but `GravitinoLanceTableOperations.describeTable` still calls `catalog.asTableCatalog().loadTable(...)` unconditionally. With this PR, the Generic Lance `loadTable` path may open the Lance dataset and refresh/write Gravitino table columns for declared/empty-schema tables, and also for `VERSION_CHECK`. So the expensive part still happens before we decide whether to set `schema` in the response. Also, the REST layer currently defaults absent `load_detailed_metadata` to true. For Gravitino Lance REST, I think the safer default is lightweight describe, and detailed metadata should only be loaded when `Boolean.TRUE.equals(request.getLoadDetailedMetadata())`.
> 2. The `DECLARED_ONLY` mode name/semantics are a bit unclear. The current implementation refreshes not only tables with `lance.declared=true`, but also any table whose stored Gravitino columns are empty. That may be useful for register/staged-create repair, where Gravitino has empty column metadata but the Lance dataset already has a real schema. However, the mode name reads like it only applies to declared tables. Could we either rename the mode or document explicitly that `DECLARED_ONLY` also repairs empty stored columns? One related edge case is a genuinely zero-column Lance dataset: because stored columns remain empty, `loadTable` will keep opening the dataset on every load and return without recording that it has already checked it.
> 
> Could we keep default Lance REST describe lightweight, and clarify the default schema-refresh mode semantics around empty stored columns?


I renamed it to `DECLARED_AND_EMPTY`.  For the first one, we had already created a new PR to track it.

FANNG1:
The rename to `DECLARED_AND_EMPTY` looks good. One small logic point around `VERSION_CHECK`: if the dataset version has not changed, re-reading schema from storage is not useful even when Gravitino stored columns are empty.

Currently the early return is guarded by `!emptySchema`:

```java
if (refreshMode == SchemaRefreshMode.VERSION_CHECK
    && !declaredOnly
    && !emptySchema
    && !isDatasetVersionChanged(table, datasetVersion)) {
  return table;
}
```

Could we drop the `!emptySchema` condition? In `VERSION_CHECK`, once `lance.version` matches the dataset version, `loadTable` should be able to return the stored metadata without calling `dataset.getSchema()`, regardless of whether the stored columns are empty. If there is no stored `lance.version` (for example immediately after register), `isDatasetVersionChanged` already returns true, so the first load can still read schema and repair metadata.

yuqi1129:
> The rename to `DECLARED_AND_EMPTY` looks good. One small logic point around `VERSION_CHECK`: if the dataset version has not changed, re-reading schema from storage is not useful even when Gravitino stored columns are empty.
> 
> Currently the early return is guarded by `!emptySchema`:
> 
> ```java
> if (refreshMode == SchemaRefreshMode.VERSION_CHECK
>     && !declaredOnly
>     && !emptySchema
>     && !isDatasetVersionChanged(table, datasetVersion)) {
>   return table;
> }
> ```
> 
> Could we drop the `!emptySchema` condition? In `VERSION_CHECK`, once `lance.version` matches the dataset version, `loadTable` should be able to return the stored metadata without calling `dataset.getSchema()`, regardless of whether the stored columns are empty. If there is no stored `lance.version` (for example, immediately after register), `isDatasetVersionChanged` already returns true, so the first load can still read schema and repair metadata.

I did check this problem: as long as the column is empty for a Lance table, the Gravitino API will repeatedly check it. I did not change it, as I think an empty column table is rare. Anyway, your comments are valuable, and I will take them.

yuqi1129:
> > The rename to `DECLARED_AND_EMPTY` looks good. One small logic point around `VERSION_CHECK`: if the dataset version has not changed, re-reading schema from storage is not useful even when Gravitino stored columns are empty.
> > Currently the early return is guarded by `!emptySchema`:
> > ```java
> > if (refreshMode == SchemaRefreshMode.VERSION_CHECK
> >     && !declaredOnly
> >     && !emptySchema
> >     && !isDatasetVersionChanged(table, datasetVersion)) {
> >   return table;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could we drop the `!emptySchema` condition? In `VERSION_CHECK`, once `lance.version` matches the dataset version, `loadTable` should be able to return the stored metadata without calling `dataset.getSchema()`, regardless of whether the stored columns are empty. If there is no stored `lance.version` (for example, immediately after register), `isDatasetVersionChanged` already returns true, so the first load can still read schema and repair metadata.
> 
> I did check this problem: as long as the column is empty for a Lance table, the Gravitino API will repeatedly check it. I did not change it, as I think an empty column table is rare. Anyway, your comments are valuable, and I will take them.

done

yuqi1129:
Gently pin @FANNG1

FANNG1:
There is one edge case in the zero-column schema refresh path that can leave stale columns in Gravitino.

In `VERSION_CHECK` mode, suppose Gravitino currently stores:

```text
columns = [id, name]
lance.version = 8
```

Then an external Lance writer updates the underlying dataset to a new zero-column schema:

```text
Lance dataset columns = []
Lance dataset version = 9
```

On the next `loadTable`, the code detects the version change and reads the Lance schema. But because `columns.length == 0`, it goes through `recordCheckedEmptyVersion()`, which only updates `lance.version` and keeps the existing stored columns. The result becomes:

```text
columns = [id, name]  // stale
lance.version = 9
```

After that, future `VERSION_CHECK` loads see version `9 == 9` and return early without reading the schema, so the stale columns are permanently preserved.

I think this path should distinguish between:

1. stored columns are already empty and the dataset is also empty: record the checked version as the current code does;
2. stored columns are non-empty but the dataset schema is now empty: replace the stored columns with an empty list, or avoid recording the new version so a later load can still repair it.

A regression test for `VERSION_CHECK + non-empty stored columns + changed dataset version + empty dataset schema` would cover this.

FANNG1 (approved):

FANNG1:
LGTM

FANNG1:
This PR fixes https://github.com/apache/gravitino/issues/11178 too ?  could you update PR title and description?

yuqi1129:
> This PR fixes #11178 too ? could you update PR title and description?

done

PR #11180, opened 2026-05-21
[#11164] improvement(build): Clean distribution package before copying files

### LINKED ISSUE #11164: [Improvement] compileDistribution should clean stale jars from distribution/package before copying

`./gradlew compileDistribution` (`build.gradle.kts:781`) is a copy-only task — it depends on a list of `copy*` subtasks and a `doLast` block that copies `conf/`, `bin/`, etc. into `distribution/package/`, but it never **removes** files that are no longer produced by the build. Old jars from previous versions silently accumulate.

Concrete example hit on 2026-05-20 while validating the `lance-core` 4.0.1 → 6.0.0 upgrade:
- `distribution/package/libs/` ended up with **both** `lance-core-4.0.1.jar` and `lance-core-6.0.0.jar`.
- `distribution/package/lance-rest-server/libs/` had the same duplicate.
- `distribution/package/libs/lance-namespace-apache-client-0.5.2.jar` from a much older build was still present, shadowing the newer `0.7.5` that the aux service expected. This silently broke the `lance-ray` matrix integration test for **days** (request schema validation rejected `check_declared`, see #10983).

Because Java classpath order is undefined when two versions of the same jar are present, the bug surfaces non-deterministically — passes in CI (always fresh checkout) and fails locally after any version bump.

### How should we improve?

Make `compileDistribution` start from a clean output directory. Two options, pick one:

1. **(preferred — narrow)** In the `compileDistribution` task block, add a `doFirst { delete(projectDir.dir("distribution/package")) }` so only the distribution output is wiped (not other modules' `build/`). Cheap and self-contained.
2. **(alternative)** Register a small task like `cleanDistributionPackage` that deletes `distribution/package/`, then add `dependsOn(cleanDistributionPackage)` and make sure it runs before the `copy*` tasks (e.g. via `mustRunAfter`).

A repo-wide `./gradlew clean` would also fix it but forces a full Java rebuild — too heavy as a default.

Verify with:
```bash
./gradlew clean compileDistribution -PskipWeb=true -x test
# bump a version in gradle/libs.versions.toml, then:
./gradlew compileDistribution -PskipWeb=true -x test
# ls distribution/package/libs/ | grep <old-version>   # should be empty
```

ningsh7:

This change adds a dedicated `cleanDistributionPackage` task to remove stale distribution package outputs and subproject jar outputs before distribution artifacts are copied.

The task cleans:
- `distribution/package`
- `distribution/package-all`
- subproject `build/libs` directories

Copy tasks that write into the distribution package directories are wired to depend on `cleanDistributionPackage`. Jar-producing tasks and copy tasks that write into subproject `build/libs` are ordered to run after the clean task, so stale jars are removed before new artifacts are generated and copied.

`compileDistribution` previously copied new files into distribution package directories without removing files that were no longer produced by the current build. After dependency version bumps, stale jars could remain either in the distribution package outputs or in subproject `build/libs` directories that are used as copy sources.

This may cause non-deterministic local failures when old and new versions of the same jar coexist.

I avoided deleting the package directories directly in `compileDistribution.doFirst` because `compileDistribution` depends on several copy tasks. A `doFirst` action on `compileDistribution` may run after those dependent copy tasks and delete files that were just copied.

This approach avoids a full repository clean while still cleaning the jar/package outputs relevant to `compileDistribution`.

Fix: #11164

No. This only changes the local distribution build behavior.

- `./gradlew compileDistribution -PskipWeb=true -x test --dry-run`
  - Verified `:cleanDistributionPackage` is scheduled before jar/copy tasks.
  - Verified there is no dependency cycle.

- `./gradlew compileDistribution -PskipWeb=true -x test`
  - Build succeeded with `lance = "2.0.1"`.

- Changed `lance` in `gradle/libs.versions.toml` from `2.0.1` to `2.0.0`.

- Re-ran `./gradlew compileDistribution -PskipWeb=true -x test`
  - Build succeeded with `lance = "2.0.0"`.

- Verified the old Lance jar was removed:
  - `find lance/lance-rest-server/build/libs distribution/package distribution/package-all -name 'lance-core-2.0.1.jar' -print`
  - Result: no output.

- Verified the current Lance jar exists in expected locations:
  - `find lance/lance-rest-server/build/libs distribution/package distribution/package-all -name 'lance-core-2.0.0.jar' -print`
  - Result includes:
    - `lance/lance-rest-server/build/libs/lance-core-2.0.0.jar`
    - `distribution/package/libs/lance-core-2.0.0.jar`
    - `distribution/package/lance-rest-server/libs/lance-core-2.0.0.jar`
    - `distribution/package-all/libs/lance-core-2.0.0.jar`
    - `distribution/package-all/lance-rest-server/libs/lance-core-2.0.0.jar`

- Restored `lance` in `gradle/libs.versions.toml` from `2.0.0` to `2.0.1`.

- Re-ran `./gradlew compileDistribution -PskipWeb=true -x test`
  - Build succeeded with `lance = "2.0.1"`.

- Verified the temporary Lance jar was removed:
  - `find lance/lance-rest-server/build/libs distribution/package distribution/package-all -name 'lance-core-2.0.0.jar' -print`
  - Result: no output.

- Verified the restored Lance jar exists in expected locations:
  - `find lance/lance-rest-server/build/libs distribution/package distribution/package-all -name 'lance-core-2.0.1.jar' -print`

- `./gradlew spotlessApply`
- `git diff --check`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,6 +69,9 @@ if (scalaVersion !in listOf("2.12", "2.13")) {
 }
 
 val skipWeb: Boolean = (project.findProperty("skipWeb") as? String)?.toBoolean() ?: false
+val distributionPackageDir = layout.projectDirectory.dir("distribution/package")
+val distributionPackageAllDir = layout.projectDirectory.dir("distribution/package-all")
+val subprojectJarOutputDirs = subprojects.map { it.layout.buildDirectory.dir("libs") }
 
 project.extra["extraJvmArgs"] =
   listOf(
@@ -790,6 +793,12 @@ jacoco {
 tasks {
   val projectDir = layout.projectDirectory
   val outputDir = projectDir.dir("distribution")
+  val cleanDistributionPackage by registering(Delete::class) {
+    group = "gravitino distribution"
+    delete(distributionPackageDir, distributionPackageAllDir)
+    delete(subprojectJarOutputDirs)
+  }
+
   val compileDistribution by registering {
     val dependencies =
       mutableListOf(
@@ -808,10 +817,12 @@ tasks {
       dependencies.add(":web:web:build")
       dependencies.add(":web-v2:web:build")
     }
+    dependsOn(cleanDistributionPackage)
     dependsOn(dependencies)
 
     group = "gravitino distribution"
-    outputs.dir(projectDir.dir("distribution/package"))
+    outputs.dir(distributionPackageDir)
+    outputs.dir(distributionPackageAllDir)
     doLast {
       copy {
         from(projectDir.dir("conf")) { into("package/conf") }
@@ -867,8 +878,8 @@ tasks {
 
       // Copy the all directory distribution/package to distribution/package-all
       copy {
-        from(projectDir.dir("distribution/package"))
-        into(projectDir.dir("distribution/package-all"))
+        from(distributionPackageDir)
+        into(distributionPackageAllDir)
       }
 
       // remove catalogs-contrib modules from distribution/package
@@ -978,6 +989,7 @@ tasks {
 
   val assembleDistribution by registering(Tar::class) {
     dependsOn(
+      compileDistribution,
       ":trino-connector:trino-connector-435-439:assembleTrinoConnector",
       ":trino-connector:trino-connector-440-445:assembleTrinoConnector",
       ":trino-connector:trino-connector-446-451:assembleTrinoConnector",
@@ -990,7 +1002,7 @@ tasks {
     group = "gravitino distribution"
     finalizedBy("checksumDistribution")
     into("${rootProject.name}-$version-bin")
-    from(compileDistribution.map { it.outputs.files.single() })
+    from(distributionPackageDir)
     compression = Compression.GZIP
     archiveFileName.set("${rootProject.name}-$version-bin.tar.gz")
     destinationDirectory.set(projectDir.dir("distribution"))
@@ -1207,6 +1219,32 @@ tasks {
 
   clean {
     dependsOn(cleanDistribution)
+  }
+}
+
+gradle.projectsEvaluated {
+  val cleanDistributionPackageTask = rootProject.tasks.named("cleanDistributionPackage")
+  val distributionPackagePaths =
+    listOf(distributionPackageDir, distributionPackageAllDir)
+      .map { it.asFile.toPath().toAbsolutePath().normalize() }
+  val subprojectJarOutputPaths =
+    subprojectJarOutputDirs.map { it.get().asFile.toPath().toAbsolutePath().normalize() }
+
+  allprojects {
+    tasks.withType<Jar>().configureEach {
+      mustRunAfter(cleanDistributionPackageTask)
+    }
+
+    tasks.withType<Copy>().configureEach {
+      val copyDestinationDir = destinationDir ?: return@configureEach
+      val destinationPath = copyDestinationDir.toPath().toAbsolutePath().normalize()
+      if (distributionPackagePaths.any { destinationPath.startsWith(it) }) {
+        dependsOn(cleanDistributionPackageTask)
+      }
+      if (subprojectJarOutputPaths.any { destinationPath.startsWith(it) }) {
+        mustRunAfter(cleanDistributionPackageTask)
+      }
+    }
   }
 }
 


build.gradle.kts
```diff
@@ -778,6 +780,11 @@ jacoco {
 tasks {
   val projectDir = layout.projectDirectory
   val outputDir = projectDir.dir("distribution")
+  val cleanDistributionPackage by registering(Delete::class) {
+    group = "gravitino distribution"
+    delete(distributionPackageDir, distributionPackageAllDir)
```
yuqi1129, line 798: I prefer to clean all jars in all submodules besides the distribution package. The result looks like the following command.
```
./gradlew clean && ./gradlew compileDistribution 
```

ningsh7: @yuqi1129 Thanks for the suggestion. I updated the patch to clean the subproject `build/libs` outputs as well as `distribution/package` and `distribution/package-all`. Could you please take another look when you have time?

yuqi1129:
```java 
val compileDistribution by registering {
      // Add clean task
      dependsOn("clean")
      val dependencies = ...
```
I mean, we can only add a single line above to solve the problem, can you try to verify it?

ningsh7: > ```java
> val compileDistribution by registering {
>       // Add clean task
>       dependsOn("clean")
>       val dependencies = ...
> ```
> 
> I mean, we can only add a single line above to solve the problem, can you try to verify it?

@yuqi1129 Sure! I have verified your suggestion. Adding only `dependsOn("clean")` to `compileDistribution` looks fine in dry-run:

- `./gradlew compileDistribution -PskipWeb=true -x test --dry-run`
- `:cleanDistribution` and `:clean` were listed before the copy/jar tasks
- No dependency cycle was reported

However, the actual execution order was different:

- `./gradlew compileDistribution -PskipWeb=true -x test`
- Some copy tasks were executed before `clean`, for example:
  - `:copySubprojectDependencies`
  - `:copyCatalogLibAndConfigs`
  - `:lance:lance-rest-server:copyLibs`
- Then Gradle executed:
  - `:cleanDistribution`
  - `:clean`

This deleted files that had already been copied into `distribution/package`. The build still ended with `BUILD SUCCESSFUL`, but the generated distribution package was incomplete.

I checked the Lance REST server jars afterwards:

- `find distribution/package distribution/package-all -name 'lance-core-2.0.1.jar' -print`
  - Result: no output
- `find distribution/package -maxdepth 3 -type d -path '*/lance-rest-server/libs' -print`
  - Result: no output

So adding only `dependsOn("clean")` does not solve the issue. It has a similar ordering risk to deleting files in `compileDistribution.doFirst`: the clean task may run after some dependent
copy tasks and remove files that were already copied.

I restored the current implementation and verified it again:

- `./gradlew compileDistribution -PskipWeb=true -x test`
- `find distribution/package distribution/package-all -name 'lance-core-2.0.1.jar' -print`

The result included:

- `distribution/package/libs/lance-core-2.0.1.jar`
- `distribution/package/lance-rest-server/libs/lance-core-2.0.1.jar`
- `distribution/package-all/libs/lance-core-2.0.1.jar`
- `distribution/package-all/lance-rest-server/libs/lance-core-2.0.1.jar`

Therefore, I do not think the single-line `dependsOn("clean")` approach is safe here.

yuqi1129 (commented):
Potential regression: `compileDistribution` now declares two outputs (`distribution/package` and `distribution/package-all`), but `assembleDistribution` still does `compileDistribution.map { it.outputs.files.single() }`. That assumes exactly one output and will fail once the provider is resolved. Could we switch `assembleDistribution` to an explicit source such as `from(distributionPackageDir)` instead of relying on `single()`?

ningsh7:
> Potential regression: `compileDistribution` now declares two outputs (`distribution/package` and `distribution/package-all`), but `assembleDistribution` still does `compileDistribution.map { it.outputs.files.single() }`. That assumes exactly one output and will fail once the provider is resolved. Could we switch `assembleDistribution` to an explicit source such as `from(distributionPackageDir)` instead of relying on `single()`?

@yuqi1129 Good point, I have changed `assembleDistribution` to use `from(distributionPackageDir)` explicitly instead of `compileDistribution.outputs.files.single()`.

I also added `dependsOn(compileDistribution)` because the previous provider-based `from(...)` carried an implicit dependency on `compileDistribution`. After switching to the explicit directory source, the dry-run showed `compileDistribution` was no longer scheduled unless we declare that dependency explicitly.

Verified with `assembleDistribution --dry-run` and the real `assembleDistribution` build.

Please check again thanks!

yuqi1129 (approved):

yuqi1129:
@ningsh7 
I'm going to merge this one, thanks for your contributions~ @ningsh7

Issue #10978, opened 2026-05-06
[Bug report] Trino/Spark throws a ForbiddenException when re-running a query after a 90-minute wait


danhuawang:
### Version

main branch

### Describe what's wrong

After an idle period of 90 minutes following an initial Trino CLI query, subsequent executions of the same query result in a ForbiddenException.

```
trino> SELECT * FROM remote_irc.gt_iceberg_test_update.tb01 ORDER BY name;
    name    | salary 
------------+--------
 alice      |     16 
 alice2     |     16 
 bob_update |     14 
 lisa       |     15 
 lisa2      |     15 
 tom        |     12 
(6 rows)

Query 20260506_075018_00095_ffihd, FINISHED, 1 node
http://34.138.225.171:8080/ui/query.html?20260506_075018_00095_ffihd
Splits: 8 total, 8 done (100.00%)
CPU Time: 0.0s total,   153 rows/s, 49.9KB/s, 39% active
Per Node: 0.0 parallelism,     2 rows/s,   668B/s
Parallelism: 0.0
Peak Memory: 556B
2.99 [6 rows, 1.95KB] [2 rows/s, 668B/s]

trino> SELECT * FROM remote_irc.gt_iceberg_test_update.tb01 ORDER BY name;
Query 20260506_092941_00096_ffihd failed: Failed to load view 'tb01'
io.trino.spi.TrinoException: Failed to load view 'tb01'
	at io.trino.plugin.iceberg.catalog.rest.TrinoRestCatalog.getIcebergView(TrinoRestCatalog.java:732)
	at io.trino.plugin.iceberg.catalog.rest.TrinoRestCatalog.getView(TrinoRestCatalog.java:699)
	at io.trino.plugin.iceberg.IcebergMetadata.getView(IcebergMetadata.java:3261)
	at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorMetadata.getView(ClassLoaderSafeConnectorMetadata.java:717)
	at io.trino.tracing.TracingConnectorMetadata.getView(TracingConnectorMetadata.java:853)
	at io.trino.metadata.MetadataManager.getViewInternal(MetadataManager.java:1610)
	at io.trino.metadata.MetadataManager.getView(MetadataManager.java:1548)
	at io.trino.tracing.TracingMetadata.getView(TracingMetadata.java:921)
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitTable(StatementAnalyzer.java:2303)
[...]

PR #11182, opened 2026-05-21
[SPARK] Add bearer token file authentication mode

### LINKED ISSUE #11181: [SPARK] Support bearer token file authentication in Spark connector
## Motivation

The Spark connector currently supports `simple`, `oauth2` and `kerberos` authentication when accessing a Gravitino server.

The existing `oauth2` mode requests a token through the configured OAuth2 token endpoint, which works for client-credentials style deployments. It does not cover runtime environments where a trusted platform component has already obtained a short-lived user or workload Bearer token and refreshes it outside the Spark connector.

Common examples include:

- Kubernetes notebook runtimes where a sidecar refreshes a short-lived token.
- Enterprise platforms where a runtime credential service writes a token file for the local Spark driver.
- Workload identity systems where Spark should reuse an existing Bearer token instead of holding OAuth client credentials.

In these cases, the Spark connector only needs to read the local token and send it as:

```text
Authorization: Bearer <token>
```

The Gravitino server still validates the token and enforces authorization. The server does not read the token file.

## Proposal

Add a Spark connector authentication mode:

```properties
spark.sql.gravitino.authType=bearer-token-file
spark.sql.gravitino.bearer.tokenFile=/path/to/token
```

Behavior:

- The token file is local to the Spark driver.
- The file may contain either the raw token or a `Bearer ` prefixed value.
- The connector reads the token file when generating authentication data for Gravitino client requests.
- The implementation uses the existing `CustomTokenProvider` extension point.
- The feature is generic and does not depend on JupyterHub, Kubernetes, a specific identity provider, or a specific token issuer.

## Why not use the existing OAuth2 mode?

The existing OAuth2 mode requires the Spark connector to call the OAuth2 token endpoint with configured credentials. In the runtime-token-file pattern, token acquisition and refresh are handled by a trusted runtime component, and the connector should not need OAuth client credentials or refresh tokens.

## Scope

This issue only proposes a Spark connector client-side authentication mode. It does not propose any server-side token issuance behavior or any deployment-specific sidecar implementation.

fzhsbc:
## What is changed

This PR adds a `bearer-token-file` authentication mode to the Spark connector.

New configuration:

```properties
spark.sql.gravitino.authType=bearer-token-file
spark.sql.gravitino.bearer.tokenFile=/path/to/token
```

The connector reads an existing Bearer token from the configured local file and sends it to Gravitino using the existing custom token provider path:

```text
Authorization: Bearer <token>
```

The token file may contain either the raw token or a `Bearer ` prefixed value.

## Why

Some runtime environments refresh short-lived user or workload tokens outside the Spark connector, for example with a trusted sidecar or workload identity component. The existing OAuth2 mode requires the connector to request a token from an OAuth2 endpoint with configured credentials, which is not appropriate for those deployments.

This change keeps token acquisition outside the Spark connector and only adds a generic way for the connector to reuse an existing local Bearer token.

## Notes

- The Gravitino server does not read the token file.
- The token file is local to the Spark driver runtime.
- The feature is not tied to JupyterHub, Kubernetes, or any specific identity provider.
- The implementation uses the existing `CustomTokenProvider` extension point.

## Tests

Added unit tests for:

- Reading a raw token from a file.
- Reading a `Bearer ` prefixed token from a file.
- Rejecting an empty token file.
- Verifying the Spark config key.

Closes #11181.

Related to #10978, but this PR does not attempt to close it because that issue appears to involve Iceberg REST / authorization cache behavior beyond Spark connector token-file authentication.

### DIFF
--- a/docs/spark-connector/spark-authentication-with-gravitino.md
+++ b/docs/spark-connector/spark-authentication-with-gravitino.md
@@ -7,11 +7,11 @@ license: "This software is licensed under the Apache License version 2."
 
 ## Overview
 
-Spark connector supports `simple` `oauth2` and `kerberos` authentication when accessing Gravitino server.
+Spark connector supports `simple`, `oauth2`, `bearer-token-file` and `kerberos` authentication when accessing Gravitino server.
 
 | Property                     | Type   | Default Value | Description                                                                                                         | Required | Since Version    |
 |------------------------------|--------|---------------|---------------------------------------------------------------------------------------------------------------------|----------|------------------|
-| spark.sql.gravitino.authType | string | `simple`      | The authentication mechanisms when communicating with Gravitino server, supports `simple`, `oauth2` and `kerberos`. | No       | 0.7.0-incubating |
+| spark.sql.gravitino.authType | string | `simple`      | The authentication mechanisms when communicating with Gravitino server, supports `simple`, `oauth2`, `bearer-token-file` and `kerberos`. | No       | 0.7.0-incubating |
 
 ## Simple mode
 
@@ -30,6 +30,14 @@ In the OAuth2 mode, you could use the following configuration to fetch an OAuth2
 | spark.sql.gravitino.oauth2.tokenPath  | string | None          | The path of token interface in OAuth2 server. | Yes, for OAuth2 mode | 0.7.0-incubating |
 | spark.sql.gravitino.oauth2.credential | string | None          | The credential to request the OAuth2 token.   | Yes, for OAuth2 mode | 0.7.0-incubating |
 | spark.sql.gravitino.oauth2.scope      | string | None          | The scope to request the OAuth2 token.        | Yes, for OAuth2 mode | 0.7.0-incubating |
+
+## Bearer token file mode
+
+In bearer token file mode, Spark connector reads an existing Bearer token from a local file when accessing Gravitino server. This is useful when another trusted runtime component, such as a sidecar, refreshes the token file outside the Spark connector.
+
+| Property                                | Type   | Default Value | Description                                                                                  | Required                         | Since Version |
+|-----------------------------------------|--------|---------------|----------------------------------------------------------------------------------------------|----------------------------------|---------------|
+| spark.sql.gravitino.bearer.tokenFile    | string | None          | The local file path containing a Bearer token. The file may contain either the raw token or a `Bearer ` prefixed value. | Yes, for bearer-token-file mode | TBD           |
 
 ## Kerberos mode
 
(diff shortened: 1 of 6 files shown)


markhoerth:
This has been sitting without review since May, which is a shame, because the deployment pattern it describes is real and I ran into the same one independently. I have a branch that overlaps with this PR and I would rather merge with it than compete with it.

What I built separately: the same `CustomTokenProvider` path, plus per-identity caching in `GravitinoCatalogManager`, because the connector's catalog cache is keyed on catalog name alone and never evicts. Once more than one identity can reach a shared driver, the first session to touch a catalog populates that cache and later sessions are served its entry, including the REST client it carries. To be clear about the exposure, that is not a live hole on main: there is only ever one client per driver today, so there is no second identity to confuse. The half that does bite today is that the cache never evicts, so a grant revoked in Gravitino keeps being honoured for the life of the driver under any auth type. The cache-key half is inert until a per-user credential mode exists, which is exactly what this PR adds, and that is my argument for landing both together rather than either alone.

On configuration naming, I think you are right and I was wrong. I had used a bare `spark.sql.gravitino.tokenFile`. Yours namespaces under the auth type, which is the convention `AuthProperties` already follows with `basic.username` and `oauth2.serverUri`. Concretely I would suggest:

```properties
spark.sql.gravitino.authType=token
spark.sql.gravitino.token.file=/path/to/token
spark.sql.gravitino.token.value=<inline token>
```

The one place I would push back is the auth type string itself, `bearer-token-file`. Two reasons:

1. That string lands in `AuthProperties`, which the Trino and Flink connectors share. The other four are named for the mechanism (`simple`, `basic`, `oauth2`, `kerberos`) rather than for where the credential came from. `bearer-token-file` bakes the source into the mechanism, which forces a second auth type the first time someone wants to pass a token inline instead of through a file.
2. An inline token is worth supporting. Not every runtime writes a file, and it is a two-line addition once `token.file` and `token.value` sit under one auth type, with the file taking precedence.

Keeping the property suffixes in `AuthProperties` rather than in `GravitinoSparkConfig` would also let Trino reuse them without redefining the strings.

Two things your PR does that mine does not, and that should survive any merge: stripping an existing `Bearer ` prefix from the file contents, and rejecting an empty token file. I would also suggest resolving the token on every request rather than once, so a sidecar rotating the file is picked up without restarting the driver, and reading the active `SparkSession` configuration in preference to the `SparkConf` so a shared driver can carry a different token per session. `HTTPClient` calls `authDataProvider.getTokenData()` inline while assembling each request, on the caller's thread, so that works without a refresh thread.

How would you like to handle this? I am happy to either send you a patch against your branch, or open a PR that leaves the auth type entirely to you and carries only the caching half. Your call, it is your issue and your PR.

PR #11186, opened 2026-05-21
[#11185] feat(spark-connector): Add Glue catalog support for Spark connector

### LINKED ISSUE #11185: [Subtask] Add Spark connector support for Glue catalog
### Describe the subtask:

Implement Spark connector support for the Gravitino Glue catalog, enabling Spark to read and write Glue-managed tables (Hive, Iceberg) through Gravitino. Includes:
- GravitinoGlueCatalog: Spark V2 catalog backed by Gravitino Glue catalog, with version-specific implementations for Spark 3.3/3.4/3.5
- GluePropertiesConverter: converts catalog/table properties between Gravitino and Spark
- Integration tests (SparkAwsGlueCatalogIT33/34/35) covering Hive, Iceberg, and mixed table types against a real AWS Glue endpoint

### Parent issue:

diqiu50:

- Add `GravitinoGlueCatalog` (and Spark 3.3/3.4/3.5 subclasses) as the Spark V2 catalog implementation backed by the Gravitino Glue catalog
- Add `GluePropertiesConverter` to map catalog/table properties between Gravitino and Spark
- Add `SparkAwsGlueCatalogIT33/34/35` integration tests covering Hive-format tables, Iceberg tables, partitioning, CTAS, and ALTER TABLE against a real AWS Glue endpoint

AWS Glue Data Catalog is the default metadata store for Amazon EMR, Athena, and Lake Formation. Without this, Spark users cannot use Gravitino as a unified metadata layer for Glue-managed tables.

Fix: #11185

Yes. Users can now configure Gravitino as the Spark catalog (`spark.sql.catalog.<name>=org.apache.gravitino.spark.connector.glue.GravitinoGlueCatalogSpark3X`) to access AWS Glue tables through Spark.

Integration tests `SparkAwsGlueCatalogIT33/34/35` run against a real AWS Glue endpoint (skipped by default, enabled by `AWS_ACCESS_KEY_ID` env var). All 33 active tests pass for Spark 3.3, 3.4, and 3.5.


spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,484 @@
...
+              ? sparkTable
+              : loadIcebergSparkTable(identifier, icebergCatalog);
+      return new SparkIcebergTable(
+          identifier,
+          gravitinoTable,
+          (SparkTable) icebergSparkTable,
+          icebergCatalog,
+          propertiesConverter,
+          sparkTransformConverter,
+          sparkTypeConverter);
```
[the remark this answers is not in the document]
diqiu50: fallback to iceberg is required

spark-connector/v3.5/spark/build.gradle.kts
```diff
@@ -136,6 +136,16 @@ dependencies {
...
   testImplementation(libs.postgresql.driver)
   testImplementation(libs.testcontainers)
+  testImplementation(libs.hadoop3.aws)
+  // hadoop-aws declares hadoop-client-api as provided; add it explicitly so S3AFileSystem can load
+  // org.apache.hadoop.fs.impl.prefetch.PrefetchingStatistics (added in 3.3.5) at runtime.
+  testImplementation(libs.hadoop3.client.api)
+  // Iceberg's GlueCatalog references several AWS SDK modules at runtime; must be on test classpath
+  testImplementation(libs.aws.glue)
+  testImplementation(libs.aws.sts)
+  testImplementation(libs.aws.s3)
```
yuqi1129: sort this part.

spark-connector/v3.4/spark/build.gradle.kts
```diff
@@ -134,6 +134,15 @@ dependencies {
...
   testImplementation(libs.testcontainers)
+  testImplementation(libs.hadoop3.aws)
+  // hadoop-aws declares hadoop-client-api as provided; add it explicitly so S3AFileSystem can load
+  // org.apache.hadoop.fs.impl.prefetch.PrefetchingStatistics (added in 3.3.5) at runtime.
+  testImplementation(libs.hadoop3.client.api)
+  // Iceberg's GlueCatalog references several AWS SDK modules at runtime; must be on test classpath
+  testImplementation(libs.aws.glue)
+  testImplementation(libs.aws.sts)
+  testImplementation(libs.aws.s3)
+  testImplementation(libs.aws.kms)
```
yuqi1129: ditto

spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/integration/test/util/SparkUtilIT.java
```diff
@@ -134,12 +135,45 @@ protected List<String> getTableMetadata(String getTableMetadataSql) {
...
+      identifier = Identifier.of(new String[] {parts[1]}, parts[2]);
+      CatalogPlugin catalog = catalogManager.catalog(parts[0]);
+      tableCatalog = (TableCatalog) catalog;
+    } else {
+      throw new IllegalArgumentException("Invalid table name format: " + tableName);
+    }
+    try {
+      return SparkTableInfo.create(tableCatalog.loadTable(identifier));
+    } catch (NoSuchTableException e) {
+      throw new RuntimeException(e);
```
yuqi1129, line 175: Why do we need to catch `NoSuchTableException`? Is it just the subclass of `RuntimeException`?

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCredentialsProvider.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * AWS credentials provider for Iceberg {@code GlueCatalog} that reads static credentials from a
+ * properties map.
+ *
+ * <p>Iceberg 1.10+ loads credentials via {@code client.credentials-provider}. This class is
+ * instantiated dynamically by Iceberg's {@code AwsClientProperties} using the {@code create(Map)}
+ * static factory. The properties map contains the {@code client.credentials-provider.*} entries
+ * with their prefix stripped, i.e. {@code access-key-id} and {@code secret-access-key}.
+ */
+public class GravitinoGlueCredentialsProvider implements AwsCredentialsProvider {
```
yuqi1129, line 38: I remember that there already exist several similar classes. Can you try to reuse them?

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,350 @@
...
+   * Hive tables are loaded from Derby, syncing from Glue first if the entry is missing.
+   */
+  @Override
+  protected Table loadSparkTable(Identifier ident) {
+    try {
+      org.apache.gravitino.rel.Table gravitinoTable = loadGravitinoTable(ident);
+      if (isIcebergTable(gravitinoTable)) {
+        return loadIcebergSparkTable(ident, getOrCreateIcebergGlueCatalog());
+      }
+      syncNamespaceToDerby(ident.namespace());
```
yuqi1129: What is it used for? Why do we sync the namespace to Derby?

yuqi1129: `Derby` is only used for testing, I'm curious about why do we add those logic in the code.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GluePropertiesConverter.java
```diff
@@ -0,0 +1,152 @@
...
+ * <p>This converter handles the property mapping for:
+ *
+ * <ul>
+ *   <li>Non-Iceberg tables: pass through AWS credentials and region to HiveTableCatalog
+ *   <li>Iceberg tables: map Gravitino Glue properties to Iceberg's GlueCatalog configuration
+ * </ul>
+ */
+public class GluePropertiesConverter implements PropertiesConverter {
+
+  public static final String GLUE_CATALOG_IMPL = "org.apache.iceberg.aws.glue.GlueCatalog";
```
yuqi1129: Can we use GlueCatalog.class.getName()

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GluePropertiesConverter.java
```diff
@@ -0,0 +1,152 @@
...
+   * <p>For Glue-backed Hive tables, Spark uses the AWS SDK directly through the Hive metastore
+   * compatibility layer. The properties are passed through to enable Glue API access.
+   */
+  @Override
+  public Map<String, String> toSparkCatalogProperties(Map<String, String> properties) {
+    Preconditions.checkArgument(properties != null, "Glue Catalog properties should not be null");
+    HashMap<String, String> all = new HashMap<>();
+    String region = properties.get(AWS_REGION);
+    if (StringUtils.isNotBlank(region)) {
+      all.put("aws.region", region);
```
yuqi1129: You may need to add a map to maintain the mapping of Spark key to Gravitino key.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
diqiu50: Spark Hive tables cannot retrieve metadata using the AWS Glue SDK, so we synchronize the metadata retrieved by Gravitino from Glue into Derby to make it accessible for Spark.

spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/integration/test/util/SparkUtilIT.java
diqiu50: TableCatalog.loadTable() declares throws NoSuchTableException in Spark's API,  It' not a subclass of Runtime Exception ,Without it the code won't compile.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCredentialsProvider.java
diqiu50: This two class live in separate modules.  moving to catalog-common would required adding the AWS-SDK as  a dependency to that module

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
diqiu50: Using AWSGlueDataCatalogHiveClientFactory to replace the Hive metastore client in Spark is not feasible for Gravitino because:

  1. Requires patching Hive: The hive.metastore.client.factory.class mechanism does not exist in standard Apache Hive. It requires
  manually applying the AWS-private patch HIVE-12679 to Hive (the fork embedded in Spark) and rebuilding it from source.
  2. JAR not in Maven Central: aws-glue-datacatalog-spark-client depends on the patched Hive  and is not published to
  Maven Central. It must be manually compiled and distributed to every Spark node's classpath.
  3. EMR-only in practice: This approach works out of the box only on AWS EMR, where Amazon pre-applies the patch and pre-installs the
  JARs. It cannot be used in standard open-source Spark environments.

  As an Apache project, Gravitino cannot depend on non-Maven-Central artifacts or require users to maintain a patched Hive build. The
  current Derby sync approach, while more complex in code, works with standard unmodified Spark and Hive distributions and imposes no extra runtime dependencies on users.

yuqi1129 (commented):
Review comments on the Glue Spark catalog implementation.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+ *
+ * <ul>
+ *   <li>Non-Iceberg tables (Hive, Delta, Parquet): routed to HiveTableCatalog for I/O
+ *   <li>Iceberg tables: routed to Iceberg's GlueCatalog for I/O
+ * </ul>
+ *
+ * <p>Table routing is based on the {@code table-format} property in Glue table parameters. Tables
+ * with {@code table-format=ICEBERG} are delegated to the Iceberg backend.
+ *
+ * <p>Derby sync: Gravitino creates/modifies tables in AWS Glue. HiveTableCatalog (used as
```
yuqi1129: This should not be described as Derby-specific. The catalog is writing a shadow entry into whatever Hive metastore `HiveTableCatalog` is configured to use; embedded Derby is only one possible backend. Please rename the methods/comments to Spark/Hive metastore shadow metadata and document that Glue/Gravitino remains the source of truth.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+  @Override
+  protected Table loadSparkTable(Identifier ident) {
+    try {
+      org.apache.gravitino.rel.Table gravitinoTable = loadGravitinoTable(ident);
+      if (isIcebergTable(gravitinoTable)) {
+        return loadIcebergSparkTable(ident, getOrCreateIcebergGlueCatalog());
+      }
+      syncNamespaceToDerby(ident.namespace());
+      try {
+        return sparkCatalog.loadTable(ident);
```
yuqi1129: This only syncs the Hive metastore shadow entry when `loadTable` misses. If the shadow entry already exists but Glue/Gravitino metadata changed (for example after alterTable, schema changes, location/serde changes, or a previous stale run), Spark will keep using the old `HiveTable.catalogTable()` metadata. Please refresh/recreate the shadow entry after metadata-changing operations, or validate that the existing entry matches the Gravitino table before returning it.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+  /**
+   * Overrides dropTable to also remove the table from Derby. Without this, Derby accumulates stale
+   * entries that cause TableAlreadyExistsException on the next createTable call for the same name.
+   * Iceberg tables are never registered in Derby, so Derby cleanup is skipped for them.
+   */
+  @Override
+  public boolean dropTable(Identifier ident) {
+    try {
+      if (!isIcebergTable(loadGravitinoTable(ident))) {
+        dropFromDerby(ident);
```
yuqi1129: The local shadow entry is dropped before the authoritative Glue/Gravitino drop succeeds. If `super.dropTable` fails, the source table still exists but Spark has already lost its local Hive metastore entry. Please perform the shadow cleanup only after `super.dropTable(ident)` returns true.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+
+  /**
+   * Overrides purgeTable to also remove the table from Derby. Iceberg tables are never registered
+   * in Derby, so Derby cleanup is skipped for them.
+   */
+  @Override
+  public boolean purgeTable(Identifier ident) {
+    try {
+      if (!isIcebergTable(loadGravitinoTable(ident))) {
+        dropFromDerby(ident);
```
yuqi1129: Same ordering issue as `dropTable`, and it is more visible for Glue because the Glue catalog operations do not override `purgeTable` today. `super.purgeTable(ident)` can throw `UnsupportedOperationException` after this code has already removed the local shadow entry. Please call the authoritative operation first and clean the shadow entry only after success, or do not override purge until Glue supports it.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
```diff
@@ -0,0 +1,417 @@
...
+      props.put("hive.output-format", outputFormat);
+    }
+    String serdeLib = gravitinoProps.get(GlueConstants.SERDE_LIB);
+    if (serdeLib != null) {
+      props.put("hive.serde", serdeLib);
+    }
+
+    syncNamespaceToDerby(ident.namespace());
+    try {
+      sparkCatalog.createTable(ident, schema, sparkPartitions, props);
```
yuqi1129: The shadow table is created with only schema, partition transforms, location, input/output format, and serde. It drops table comments, column comments, serde parameters, and other table properties that may be needed by Spark/Hive readers or writers. Please either sync the full Hive-compatible metadata from Glue/Gravitino, or document and test the intentionally supported subset.

spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/integration/test/glue/SparkGlueCatalogIT.java
```diff
@@ -0,0 +1,582 @@
...
+   * non-Iceberg (PARQUET) tables via the Glue catalog path. The rename operation triggers
+   * tableExists() which calls tableCatalog.loadTable() → GravitinoGlueCatalog.loadTable() →
+   * loadSparkTable() → HiveTableCatalog.loadTable(ident), but the table lookup may fail because
+   * Derby and Glue are out of sync for renamed tables. This requires further investigation into the
+   * HiveTableCatalog.loadTable() behavior and the renameTable dispatch chain in
+   * RenameTableExec.apply().
+   */
+  @Test
+  @Override
+  protected void testRenameTable() {
```
yuqi1129: This turns a known failing/unsupported rename path into a passing empty test. Please use `@Disabled` with the tracking issue/reason, or introduce a capability switch such as `supportsRenameTable()` so the report clearly shows the case is skipped instead of passed.

spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/integration/test/glue/SparkGlueEnvIT.java
```diff
@@ -0,0 +1,358 @@
...
+    if (serverConfig != null) {
+      // Gravitino server already started by an external process (e.g., real AWS integration
+      // test environment). Use existing config.
+      gravitinoPort = getGravitinoServerPort();
+    } else {
+      // Start the embedded Gravitino server. SparkEnvIT.startIntegrationTest() is an empty
+      // @BeforeAll override that prevents JUnit from auto-invoking BaseIT.startIntegrationTest().
+      // We call startServer() directly (a non-@BeforeAll method) to avoid the virtual-dispatch
+      // problem that would occur with reflection + Method.invoke().
+      try {
```
yuqi1129: Catching all embedded-server startup failures and falling back to port 8090 hides the real failure and makes the test fail later with a less useful connection error. Since the external-server path is already represented by `serverConfig != null`, please let embedded startup failures propagate here.

spark-connector/v3.3/spark/src/test/java/org/apache/gravitino/spark/connector/integration/test/glue/SparkAwsGlueCatalogIT33.java
```diff
@@ -0,0 +1,71 @@
...
+ * Required environment variables:
+ *
+ * <ul>
+ *   <li>{@code AWS_ACCESS_KEY_ID}
+ *   <li>{@code AWS_SECRET_ACCESS_KEY}
+ *   <li>{@code AWS_DEFAULT_REGION} (optional, defaults to us-east-1)
+ *   <li>{@code S3_BUCKET_NAME} (optional, uses default bucket if not set)
+ * </ul>
+ */
+@EnabledIfEnvironmentVariable(named = "AWS_ACCESS_KEY_ID", matches = ".+")
```
yuqi1129, line 36: This enables the real AWS test when only `AWS_ACCESS_KEY_ID` is set, but the setup also requires `AWS_SECRET_ACCESS_KEY` and uses a real S3 bucket. Please require the secret key too, and strongly consider requiring `S3_BUCKET_NAME` instead of falling back to the fixed default bucket. The same applies to the Spark 3.4 and 3.5 variants.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/glue/GravitinoGlueCatalog.java
diqiu50: Remove Derby

spark-connector/v3.5/spark/build.gradle.kts
```diff
@@ -27,6 +29,51 @@ repositories {
   mavenCentral()
 }
 
+val glueHiveJarsDir = "$buildDir/tmp/glue-hive-jars"
+val glueLibsBase =
+  "https://raw.githubusercontent.com/datastrato/spark-hive-glue-libs/main/spark3/glue-3.4.0"
+val mavenCentral = "https://repo1.maven.org/maven2"
```
yuqi1129: What this one is used for?

spark-connector/v3.5/spark/build.gradle.kts
```diff
@@ -27,6 +29,51 @@ repositories {
...
+
+// Jars for Spark's IsolatedClientLoader: patched Hive 2.3.10 + Glue datacatalog client.
+// aws-java-sdk-glue 1.12.31 requires PropertyNamingStrategy$PascalCaseStrategy which was
+// removed in Jackson 2.12. Jackson 2.11.x is included here so the isolated classloader
+// uses a compatible version instead of the app-level Jackson bundled with Spark (2.14+).
+val glueHiveJarDownloads =
+  mapOf(
```
yuqi1129: So, we will maintain all those files in the `glueLibsBase`?

spark-connector/v3.5/spark/build.gradle.kts
diqiu50: A patch to add AWS Glue support for Spark Hive JARs, which we use in integration tests to launch Spark environments.

spark-connector/v3.5/spark/build.gradle.kts
diqiu50: Yes

yuqi1129:
I have no more comments, and please solve the CI problem.

yuqi1129 (approved):

yuqi1129:
@roryqi @diqiu50 
Do you have time to take a look?

roryqi:
> @roryqi @diqiu50 Do you have time to take a look?

No comments.

Issue #11170, opened 2026-05-20
[Bug] Audit log subsystem: file management, list response capture, client IP, and formatter payload gaps


jerryshao:
### Version
main branch

### Describe what's wrong

Four related defects in the audit log subsystem affect the correctness and completeness of audit records.

---

**1. Audit log file management is broken (no rotation, unreliable flushing)**

`FileAuditWriter` opens a `FileOutputStream` with `append=true` and writes forever — no size cap, no time-based roll, no compression, no retention policy. Additionally, `tryFlush()` only fires when a new event arrives and the configured interval has elapsed, with no background timer. Combined: the log grows without bound, and under low traffic events can sit in the OS buffer indefinitely before reaching disk.

**2. List event responses are not captured**

`IcebergListTableEvent`, `IcebergListNamespacesEvent`, `IcebergListViewEvent`, and native equivalents (`ListMetalakeEvent`, `ListCatalogEvent`, `ListSchemaEvent`, `ListTableEvent`) store only the parent identifier. The response payload returned to the caller is never saved on the event object, so the audit log cannot answer "what was user X shown at time T?" — critical when RBAC filtering causes different users to see different results for the same query.

**3. Client IP is not captured correctly on either API surface**

On the Gravitino REST surface, `BaseEvent.remoteAddress()` returns the hardcoded string `"unknown"` and none of the event subclasses override it. On the Iceberg REST surface, `IcebergRequestContext` reads `httpRequest.getRemoteHost()` directly, recording the proxy IP rather than the real client behind any load balancer. In both cases, forensic correlation between Gravitino audit logs and network logs is not possible.

**4. Event payload data is captured but discarded by the formatter**

`SimpleFormatterV2` emits only 7 fixed fields. Rich data already present on event objects — `LoadTableResponse` (including vended credentials), `CatalogInfo`, request/response pairs — is never serialized. Critically, there is no way to tell from an audit row whether a `LoadTable` operation resulted in credential vending.

---

### Error message and/or stacktrace

**Finding 1** — `FileAuditWriter.tryFlush()`:
```java
private void tryFlush() {
    Instant now = Instant.now();
    if (now.isAfter(nextFlushTime)) {
        nextFlushTime = now.plusSeconds(flushIntervalSecs);
        doFlush();
    }
}
```
Flush is only triggered on write, never on a background timer.

**Finding 2** — `IcebergListTableEvent` constructor takes no response payload:
```java
public IcebergListTableEvent(
[...]

PR #11191, opened 2026-05-21
[#11170] fix(audit): Fix file rotation, client IP capture, and log format

jerryshao:

Fixes three audit log bugs from #11170:

**Problem 1 — FileAuditWriter has no file rotation, compression, or retention**
- Rewrite `FileAuditWriter` to delegate all file management to Log4j2 via a
  dedicated SLF4J logger named `gravitino.audit`
- Add `audit_file` rolling appender group to `conf/log4j2.properties.template`:
  256 MB or daily rotation, gzip compression, 30-day retention
- Deprecated writer properties (`fileName`, `append`, `flushIntervalSecs`) now
  emit a `WARN` log at startup and are otherwise ignored
- Update `gravitino-server-config.md` with migration guidance

**Problem 3 — remoteAddress() always returns "unknown" for Gravitino server events**
- Add `RequestContext` ThreadLocal in `core/` to hold the client IP on the servlet thread
- Read the value at `Event` construction time (not at listener processing time),
  so async listener threads safely read a `final` field instead of ThreadLocal
- Add `RequestContextFilter` in `server/` to set/clear the value around each
  request, honoring `X-Forwarded-For` for reverse-proxy deployments
- Apply the same `X-Forwarded-For` logic to `IcebergRequestContext`

**Problem 4 — SimpleAuditLogV2 uses second-precision timestamps and omits customInfo**
- Replace `new SimpleDateFormat(...)` per call with a static thread-safe
  `DateTimeFormatter` at millisecond precision (`yyyy-MM-dd HH:mm:ss.SSS`)
- Append `customInfo()` as an 8th tab-separated field (empty string when absent)

## How was this patch tested?

- `TestRequestContext` — ThreadLocal set/get/clear/thread-isolation
- `TestEventRemoteAddress` — value captured at construction time, survives ThreadLocal clear
- `TestRequestContextFilter` — X-Forwarded-For (single/multi-entry), fallback to `getRemoteAddr()`, cleanup in finally including on exception
- `TestSimpleAuditLogV2` — millisecond precision, 8-field output, customInfo presence/absence
- `TestFileAuditWriter` — doWrite publishes to audit logger, deprecated keys emit WARN, clean properties emit no WARN, close is no-op
- `TestAuditManager` — default writer/formatter types, batch dispatch smoke test


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/listener/api/event/IcebergRequestContext.java
```diff
@@ -58,13 +58,21 @@ public IcebergRequestContext(HttpServletRequest httpRequest, String catalogName)
...
+    this.remoteHostName = resolveClientAddress(httpRequest);
     this.httpHeaders = IcebergRESTUtils.getHttpHeaders(httpRequest);
     this.catalogName = catalogName;
     this.userName = PrincipalUtils.getCurrentUserName();
     this.requestCredentialVending = requestCredentialVending;
   }
 
+  private static String resolveClientAddress(HttpServletRequest request) {
+    String xff = request.getHeader("X-Forwarded-For");
+    if (xff != null && !xff.isEmpty()) {
```
roryqi: StringUtils.isNotBlank? What's meaning of `xff`?

jerryshao: It is short for `X-Forwarded-For`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/listener/api/event/IcebergRequestContext.java
```diff
@@ -58,13 +58,21 @@ public IcebergRequestContext(HttpServletRequest httpRequest, String catalogName)
...
+      return xff.split(",")[0].trim();
+    }
+    return request.getRemoteHost();
```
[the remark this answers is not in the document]
jerryshao: We use `getRemoteHost` before, changing to `getRemoteAddr` will change the behavior, I would not change this.

server-common/src/main/java/org/apache/gravitino/server/web/RequestContextFilter.java
```diff
@@ -0,0 +1,75 @@
...
+ * X-Forwarded-For} header (a de-facto standard header set by reverse proxies; note that it is
+ * trusted unconditionally — deployments where the server is reachable directly, without a trusted
+ * reverse proxy, should be aware that clients can spoof this header). If the header is absent,
+ * {@link HttpServletRequest#getRemoteAddr()} is used instead.
+ *
+ * <p>The stored value is always cleared in a {@code finally} block to prevent thread-pool leaks.
+ */
+public class RequestContextFilter implements Filter {
+
+  private static final String X_FORWARDED_FOR = "X-Forwarded-For";
```
roryqi, line 47: Could we extract this constant to the module `server-common` to reuse for the other auxiliary services modules?

core/src/main/java/org/apache/gravitino/listener/api/event/Event.java
```diff
@@ -21,11 +21,33 @@
...
+ */
 @DeveloperApi
 public abstract class Event extends BaseEvent {
+
+  private final String remoteAddress;
+
   protected Event(String user, NameIdentifier identifier) {
     super(user, identifier);
+    String addr = RequestContext.getRemoteAddress();
+    this.remoteAddress = addr != null ? addr : "unknown";
```
roryqi: StringUtils.isBlanke?

roryqi:
Two minors. No other comments.

roryqi (approved):
LGTM.

PR #11199, opened 2026-05-22
[#11170] fix(audit): Add ListEvent interface with resultCount to list events and surface count in audit log

### LINKED ISSUE #11170: [Bug] Audit log subsystem: file management, list response capture, client IP, and formatter payload gaps

jerryshao:

Introduces a `ListEvent` marker interface that exposes `resultCount()` on all
list-operation success event classes (core and Iceberg REST server). All event
dispatchers are updated to pass the actual result count when constructing these
events. `SimpleAuditLogV2.toString()` now surfaces the count as `{count=N}` in
the custom-info audit-log field.

Backward compatibility is preserved: the original no-count constructors are
deprecated and delegate to `-1`; the audit formatter omits the count field when
`resultCount()` returns `-1`.

Without the result count in audit logs it is impossible to distinguish an
empty list response from a non-empty one, making audit trails less useful for
security analysis and capacity planning.

Fix: #11170

- New `ListEvent` interface added to the public `@DeveloperApi` surface.
- All `List*Event` classes gain a new preferred constructor that accepts a result
  count; the old constructors are deprecated but still compile and run.
- Audit log lines for list operations now include `{count=N}` in the last field.

- `TestSimpleAuditLogV2`: new tests cover count in log output, zero count, null
  identifier, count merged with custom info, deprecated constructor omitting
  count, and non-list events.
- `TestFileAuditWriter`: end-to-end log-line tests for `ListTableEvent`,
  `ListMetalakeEvent`, and `ListCatalogEvent`.
- Existing dispatcher and event tests updated to use the new constructors.


mchades (approved):
lgtm.

PR #11213, opened 2026-05-25
[#11212] test(authz): add CI workflow for multi-instance consistency

### LINKED ISSUE #11212: [Improvement] Add a CI workflow for cross-instance authorization consistency

We have a shell-based multi-instance consistency test at `dev/ci/test_multi_instance_consistency.sh` (94 assertions across 14 phases) that exercises the `JcasbinAuthorizer` cache propagation between two Gravitino instances backed by the same SQL backend. Today it has to be run manually against locally-started instances, which means:

- Regressions in the authz cache invalidation paths (owner change, role grant/revoke, role-privilege change, role/user deletion, etc.) only get caught when a developer remembers to run the script.
- The recent `versionCheckAndLoadRoles` fix for the role-delete propagation gap (follow-up to #10996) has no automated coverage gate.
- New caches / pollers added under `server-common/.../jcasbin/` are not exercised end-to-end across instances.

### How should we improve?

Add a dedicated GitHub Actions workflow that:

1. Starts a MySQL 8.0.33 service container (so both Gravitino instances share one real backend — H2 is not viable for multi-instance).
2. Applies `scripts/mysql/schema-1.3.0-mysql.sql`.
3. Builds `compileDistribution`, then stages two copies of `distribution/package` configured with:
   - Same JDBC URL → MySQL service.
   - `gravitino.cache.enabled = false` (isolates the JcasbinAuthorizer caches from the entity-store cache).
   - `gravitino.authorization.enable = true`.
   - `admin` as service admin.
   - Distinct ports (A: 8090/9001/9101, B: 8190/9011/9111).
4. Starts both instances and runs `dev/ci/test_multi_instance_consistency.sh` against them.
5. Uploads logs + confs on failure.

Triggers: push / PR on `main` and `branch-*`, plus manual `workflow_dispatch`. A `paths-filter` keeps it from running for unrelated changes (web-only, docs-only, etc.).

yuqi1129:

Adds a new GitHub Actions workflow at \`.github/workflows/multi-instance-consistency-test.yml\`
that brings up two Gravitino instances backed by a shared MySQL service container and runs
the shell-based consistency suite against them.

The complex instance-staging logic (JDBC download, conf patching, PID-grep fix, server
start + health-wait) lives in \`dev/ci/setup_multi_instance.sh\`, following the same pattern
as \`util_free_space.sh\` and \`test_multi_instance_consistency.sh\`. The YAML step is a single
\`bash dev/ci/setup_multi_instance.sh\` call.

The consistency suite at \`dev/ci/test_multi_instance_consistency.sh\` is a 19-phase /
128-assertion script that exercises the \`JcasbinAuthorizer\` caches (\`ownerRelCache\`,
\`metadataIdCache\`, \`userRoleCache\`, \`loadedRoles\`, \`groupRoleCache\`) and the
\`JcasbinChangePoller\` invalidation paths across two instances:

| Phase | What it covers |
|---|---|
| A / B' | Owner read-back + warm-cache invalidation (eventual, poll-driven) |
| C / D | Role grant / revoke immediately effective on the peer |
| E | Role-privilege add/remove immediately effective on the peer |
| F | Role hard-delete clears peer's \`loadedRoles\` cache |
| G / H | User delete + same-name re-create has no stale role inheritance |
| I | Role re-create with same name has no stale privilege leak |
| J | Cross-metalake isolation (role/owner does not leak) |
| K | DENY condition overrides ALLOW on a different role |
| L | Group lifecycle: create / grant / revoke / delete visible on peer |
| N | \`setOwner\` propagation works for non-METALAKE entities (TAG) |
| O | Multi-role partial revoke only affects the revoked role |
| P | Owner change on TAG with warm cache on the peer |
| Q | Burst of \`setOwner\` calls within one poll window converges |
| R | Re-granting an already-held role is idempotent |
| S | GROUP-type owner propagated correctly via \`ownerRelCache\` |
| T | Tag delete + same-name re-create has no stale owner (owner_meta invalidation) |
| U | GET \`/users/{user}\` on peer reflects role grant/revoke from the other instance |
| V | GET \`/roles/{role}\` on peer reflects privilege changes from the other instance |
| W | Bidirectional: grant/revoke on B is immediately effective on A |

Both Gravitino instances run with:

\`\`\`
gravitino.cache.enabled                = false    # isolates from entity-store cache
gravitino.authorization.enable         = true
gravitino.authorization.serviceAdmins  = admin
gravitino.entity.store.relational.jdbcUrl = <shared MySQL>
\`\`\`

A \`paths-filter\` job keeps the workflow from running for unrelated changes.

The script exercises the \`JcasbinAuthorizer\` cache layer end-to-end across instances;
without a CI gate, regressions in these paths only surface when someone runs the script
by hand. The recent \`versionCheckAndLoadRoles\` fix for the role-delete propagation gap
(follow-up to #10996) in particular has no automated coverage.

Fix: #11212

No.

The script itself is the test. It passes 128/128 in CI against two instances:

\`\`\`
== Summary ==
Passed: 128
Failed: 0
\`\`\`

The workflow file is structured per the existing access-control / integration test
workflows in this repo.


.github/workflows/multi-instance-consistency-test.yml
```diff
@@ -0,0 +1,231 @@
...
+            -o "$JDBC_JAR"
+          # Verify integrity against Maven Central's published SHA-1.
+          EXPECTED_SHA1="$(curl -fsSL \
+            https://repo1.maven.org/maven2/com/mysql/mysql-connector-j/8.0.33/mysql-connector-j-8.0.33.jar.sha1)"
+          echo "${EXPECTED_SHA1}  ${JDBC_JAR}" | sha1sum --check
+
+          # Second instance is a full copy with its own conf/, logs/, data/.
+          cp -a distribution/package distribution/package-b
+
+          configure_instance() {
```
roryqi: Should we use shell instead of yaml to maintain the complex logic?

yuqi1129: Done. Extracted all the staging + startup logic (JDBC download, `configure_instance()`, `patch_pid_grep()`, server start + health-wait) into `dev/ci/setup_multi_instance.sh`, following the same pattern as `util_free_space.sh` and `test_multi_instance_consistency.sh`. The YAML step is now a single `bash dev/ci/setup_multi_instance.sh` call.

roryqi:
Could u check whether it will involve the flaky tests?

yuqi1129:
> Could u check whether it will involve the flaky tests?

I'm working on it.

yuqi1129:
> flaky

Add more tests to cover flaky and metadata changes.

dev/ci/test_multi_instance_consistency.sh
```diff
@@ -0,0 +1,1051 @@
...
+expect "A: $USER_LISI (metalake owner) deletes tag $TAG_S" 200
+
+api "$INSTANCE_A" "$USER_LISI" DELETE "/api/metalakes/$METALAKE/groups/$GROUP_S"
+expect "A: delete group $GROUP_S" 200
+
+# ---- T. metadataIdCache invalidation: tag delete + same-name re-create -----
+
+section "T. Stale-owner safety on tag delete+re-create (eventual, ~${POLL_WAIT_SECS}s)"
+# When a tag is deleted on A and re-created with the same name, B must not
+# allow the old owner (wangwu) to exercise ownership on the new entity.
```
roryqi: wangwu? Could u give a better name?

roryqi:
No other comments except for names.

roryqi (approved):
LGTM.

PR #11218, opened 2026-05-25
[#10474] feat(lock): Activate latent OCC in entity store to fix HA write races

### LINKED ISSUE #10474: [Improvement] Address TreeLock limitations for Gravitino HA deployment
#### Background

Gravitino uses `TreeLock` (implemented in `core/src/main/java/org/apache/gravitino/lock/`) to ensure consistency and atomicity of metadata operations. The lock follows a hierarchical read-write locking strategy — it acquires read locks on all ancestor nodes and a write/read lock on the target node itself. For example, renaming a table `metalake.catalog.db.table1` acquires:

```
/                         → readLock
/metalake                 → readLock
/metalake/catalog         → readLock
/metalake/catalog/db      → writeLock   (parent of the renamed resource)
```

This design is managed by `LockManager`, which maintains an in-memory tree of `TreeLockNode`s backed by Java's `ReadWriteLock`.

#### Problem

`TreeLock` is **entirely in-process**. Each Gravitino server instance has its own `LockManager` with its own isolated lock tree in JVM memory. This fundamentally breaks down in **High Availability (HA) mode**, where multiple Gravitino server instances operate behind a load balancer:

1. **No cross-process coordination**: A write lock acquired on server A has zero visibility to server B. Two servers can concurrently modify the same metadata resource (e.g., both renaming the same table), causing data races and inconsistency.
2. **Race conditions on shared storage**: The backend metadata store (relational database) is shared across all nodes, but the locking layer protecting it is per-node, so conflicting writes can corrupt metadata state.
3. **Dead-lock detection is per-JVM**: The built-in dead-lock checker in `LockManager` only inspects threads within the local JVM; cross-node lock contention is entirely invisible.

#### Scope of Impact

`TreeLockUtils.doWithTreeLock(...)` is called across the majority of managers:

- `CatalogManager`, `MetalakeManager`
- `TableOperationDispatcher`, `SchemaOperationDispatcher`, `FilesetOperationDispatcher`, `TopicOperationDispatcher`, `ViewOperationDispatcher`, `ModelOperationDispatcher`, `PartitionOperationDispatcher`
- `TagManager`, `PolicyManager`, `StatisticManager`, `AccessControlManager`, `PermissionManager`, `OwnerManager`

Any metadata operation in these managers is potentially unsafe under concurrent HA traffic.

---

### How should we improve?

Two broad approaches are worth evaluating. Each has significant trade-offs:

**Option A — Distributed Lock**
Replace or wrap `TreeLock` with a distributed locking mechanism (e.g., ZooKeeper, etcd, or Redis-based). This preserves the existing hierarchical lock semantics and requires minimal changes to call sites.

- Pros: Semantically equivalent to current model; relatively bounded change surface.
- Cons: Introduces an external dependency; adds network round-trips on every metadata operation; requires handling distributed lock lease expiry, fencing tokens, and failure scenarios; potential new SPOF.

**Option B — Remove TreeLock; rely on storage-level consistency**
Eliminate `TreeLock` entirely and push concurrency control into the storage layer using optimistic locking (version-based CAS), database-level row locks, or serializable transactions.

- Pros: No external dependency; naturally correct in HA because all nodes share the same storage.
- Cons: Significant refactoring across all managers; requires retry logic for conflicts; may increase database load; harder to express coarse-grained parent-level write locks (e.g., "lock the whole schema while renaming a table").

**Option C....**

---

### Expected outcome

A concrete solution proposal covering:

1. Detailed analysis of the current `TreeLock` implementation and its limitations in HA mode.
2. Proposed solution with:
   - Design overview and key components/interfaces to change.
   - Analysis of correctness (does it fully eliminate the race conditions?).
   - Analysis of performance impact.
   - Migration strategy (how to transition from the current design without breaking existing behavior).
3. Optionally, a proof-of-concept implementation or pseudocode for the critical path.

urvishdesai:

Resolves #10474 — TreeLock is JVM-local, so concurrent writes from two Gravitino server nodes can silently corrupt entity state. This PR activates the **optimistic concurrency control (OCC) that already exists latently** in the entity store and adds a retry loop in the lock layer to make it HA-safe.

**The key insight**: every entity mapper already emits `WHERE current_version = #{old.currentVersion}` in its UPDATE SQL. If two nodes write concurrently, exactly one will match and advance the version; the other returns 0 rows. Before this PR, returning 0 rows silently threw a generic `IOException("Failed to update the entity")`. This PR converts that 0-row result into a typed `OptimisticLockException` and retries the full read-modify-write cycle (up to 3 times with 10/20/40 ms exponential back-off) inside `TreeLockUtils.doWithTreeLock`.

```sql
UPDATE catalog_meta SET ..., current_version = #{new.currentVersion}
WHERE  catalog_id = #{old.catalogId}
AND    current_version = #{old.currentVersion}
AND    deleted_at = 0
```

**Changes:**
- **`OptimisticLockException.java`** (new) — typed runtime exception extending `GravitinoRuntimeException`
- **`TreeLockUtils.java`** — retry loop for WRITE operations only; READ path unchanged; max 3 retries, 10/20/40 ms back-off
- **11 entity service classes** — replace generic `IOException` throw with `throw new OptimisticLockException(...)` when UPDATE returns 0 rows
- **Tests** — `TestTreeLockUtils` gets 3 new OCC test cases; `TestFilesetMetaService` updated for new exception type

`TreeLock` uses `ReentrantReadWriteLock` — a JVM-local primitive. In a multi-node (HA) deployment, two servers can acquire write locks on the same entity independently and race to commit, with no cross-node coordination. The existing WHERE clause in every UPDATE already provides DB-level atomicity, but the 0-row signal was discarded. This PR surfaces and acts on that signal.

Fix: #(issue)

No. The retry is transparent. After retries are exhausted `OptimisticLockException` propagates to the REST layer and returns HTTP 409 Conflict — a cleaner signal than the previous opaque 500.

- `./gradlew :core:test -PskipITs` — all existing tests pass
- New unit tests in `TestTreeLockUtils` cover:
  - Retry succeeds on second attempt
  - READ path does not retry on `OptimisticLockException`
  - Retry exhaustion re-throws after `OCC_MAX_RETRIES + 1` attempts
- `TestFilesetMetaService` updated to assert the new exception type

---

### Problem

`TreeLock` is JVM-local. In an HA setup with N Gravitino nodes, each node has its own lock tree. Two nodes can hold write locks on the same entity simultaneously and race to commit — silently corrupting state.

### This Approach — Latent OCC Activation (Option B)

**Core idea**: leverage the `current_version` column already present in every entity table. The DB UPDATE becomes the arbiter:

```sql
UPDATE gravitino_meta_schema
SET    current_version = #{new.currentVersion}, ...
WHERE  schema_id = #{new.schemaId}
  AND  current_version = #{old.currentVersion}   -- ← already in the SQL
```

Only one concurrent write can advance `N → N+1`. The loser gets 0 rows affected, catches `OptimisticLockException`, re-reads fresh state, and retries. TreeLock remains for single-node ordering; OCC adds cross-node safety at zero extra schema cost.

```
Node A                    DB                     Node B
  │── lock(WRITE) ──────────────────────────────────│
  │── SELECT version=5 ──▶                          │
  │                        ◀── SELECT version=5 ────│
  │── UPDATE WHERE v=5 ──▶ (v→6, 1 row affected)    │
  │                        ◀── UPDATE WHERE v=5 ─── │ (0 rows → OptimisticLockException)
  │                                                 │── retry: SELECT version=6
  │                                                 │── UPDATE WHERE v=6 ──▶ (v→7, 1 row)
```

**Retry policy**: 3 attempts max, 10/20/40 ms exponential back-off, WRITE path only.

**Limitations (acknowledged)**:
- Metalakes and catalogs set `nextVersion = lastVersion` (no increment), so OCC falls back to the full-field WHERE clause already present — slightly weaker but still safe under the DB's row-level locking.
- Parent-level create/drop isolation (e.g., two nodes racing to create tables in the same schema) is a pre-existing HA gap not addressed here. A follow-up can add an existence check inside the child INSERT transaction.

### Comparison with Other Proposals

| Dimension | **This PR (OCC)** | soulmachine's PR #11020 (JdbcLockBackend) | Option A (External ZK/Redis) |
|---|---|---|---|
| Schema changes | None — uses existing `current_version` | New `gravitino_lock` table | None |
| Extra DB round-trips per write | 0 on happy path (retry only on conflict) | +2 always (`SELECT FOR UPDATE` + release) | 0 |
| HA safe | ✅ Yes | ✅ Yes | ✅ Yes |
| New dependencies | None | None | ZooKeeper or Redis |
| Failure mode | Retry → 409 after exhaustion | Timeout → blocked threads | Network partition |
| Lines of code | ~150 | ~500+ (SPI + two backends) | ~1000+ |
| Contention model | Optimistic (low-write-conflict workloads) | Pessimistic (serialises all writes) | Pessimistic |
| Extensibility | Add retry tuning config | Add new LockBackend implementations | External cluster tuning |

**When OCC wins**: Gravitino's metadata workloads are typically read-heavy with infrequent concurrent writes to the same entity. OCC pays zero overhead on the common path and only retries on actual conflict.

**When pessimistic lock wins**: Under sustained high write concurrency to the same entity, OCC retry storms can increase latency. soulmachine's backend would be the right follow-up for those extreme cases.

Our OCC approach is the minimal, zero-dependency first step. It can coexist with soulmachine's SPI — a future `OccLockBackend` could be one implementation of the same interface.


jerryshao:
Hi @urvishdesai thanks a lot for your proposal. I think the purpose of the treelock doesn't only control the concurrency and consistency of DB operations. In the meantime, it also keeps the serial operations of the underlying data source. Here, in your proposal, you only take care of the DB operation. I think it is not enough.

In the meantime, the single row OCC solution can only take care of a single line update. How do we handle multi-table updates?

PR #11222, opened 2026-05-26
[#11221] fix(bin): Scope launcher PID lookup by home

### LINKED ISSUE #11221: [Improvement] gravitino.sh should detect its own JVM by GRAVITINO_HOME, not just process name

`bin/gravitino.sh.template` decides "is this install already running?" via this grep in `found_gravitino_server_pid()`:

```sh
process_name='GravitinoServer';
RUNNING_PIDS=$(ps x | grep ${process_name} | grep -v grep | awk '{print $1}');
```

This matches **any** Java process containing the string `GravitinoServer` anywhere in its command line — across the whole host. The match isn't scoped to the install (`GRAVITINO_HOME`) that the script belongs to. Concretely:

- Two co-located Gravitino installs (e.g. one staged at `/path/A` and another at `/path/B`, sharing one SQL backend for multi-instance testing) cannot both be started by their own `bin/gravitino.sh`. The second `start` invocation sees the first install's JVM and short-circuits with `Gravitino Server is already running` without ever launching — its `logs/` directory isn't even created.
- This bit the new `.github/workflows/multi-instance-consistency-test.yml` (#11212), which had to work around it by patching the line at staging time.
- `stop` / `restart` / `status` have the same blind spot in the other direction: they may operate on a sibling install's JVM instead of their own.

### How should we improve?

Tighten the grep to require the running JVM's command line to also contain this install's `GRAVITINO_HOME`. The JVM is launched as `java ... -cp ${GRAVITINO_CLASSPATH} GravitinoServer`, and `GRAVITINO_CLASSPATH` contains paths derived from `GRAVITINO_HOME` (`${GRAVITINO_HOME}/libs/*`, `${GRAVITINO_HOME}/conf`, …), so the install path is always present in the command line of every JVM that this script started.

Minimal patch in `bin/gravitino.sh.template`:

```diff
 function found_gravitino_server_pid() {
   process_name='GravitinoServer';
-  RUNNING_PIDS=$(ps x | grep ${process_name} | grep -v grep | awk '{print $1}');
+  RUNNING_PIDS=$(ps x | grep ${process_name} | grep "${GRAVITINO_HOME}" | grep -v grep | awk '{print $1}');
```

Effect: each install only ever finds its own JVM, regardless of how many sibling installs are running on the same host.

The same change applies to the aux-service launchers if they have a parallel pattern (`bin/gravitino-iceberg-rest-server.sh.template`, `bin/gravitino-lance-rest-server.sh.template`) — please grep for the same `RUNNING_PIDS=$(ps x | grep ...)` shape and apply consistently.

The CI workflow in #11212 can drop its in-line patch once this lands.

yuqi1129:

This PR scopes launcher PID detection to the current `GRAVITINO_HOME` in:

- `bin/gravitino.sh.template`
- `bin/gravitino-iceberg-rest-server.sh.template`
- `bin/gravitino-lance-rest-server.sh.template`

The PID lookup now matches both the server process name and the current installation path, and uses `ps xww` to avoid truncated command lines.

The old lookup only matched the server class name, so sibling Gravitino installations on the same host could be mistaken for the current installation. This could make `start`, `stop`, `restart`, and `status` operate on the wrong JVM.

Fix: #11221

Yes. Launcher scripts now detect only JVMs started from the same Gravitino installation directory.

- `bash -n bin/gravitino.sh.template bin/gravitino-iceberg-rest-server.sh.template bin/gravitino-lance-rest-server.sh.template`
- `git diff --check`


.github/workflows/backend-integration-test-action.yml
```diff
@@ -30,7 +30,7 @@ jobs:
   start-runner:
     name: JDK${{ inputs.java-version }}-${{ inputs.test-mode }}-${{ inputs.backend }}
     runs-on: ubuntu-22.04
-    timeout-minutes: 90
+    timeout-minutes: 120i
```
mchades: What is the meaning of the suffix "i" in the number 120?

mchades (dismissed):
lgtm.

diqiu50 (dismissed):
LGTM

diqiu50 (approved):
LGTM

PR #11254, opened 2026-05-27
[#11232][#11233] improvement(core): Global EntityChangeLogPoller with CatalogManager cache invalidation and automatic retention cleanup

### LINKED ISSUE #11232: [Improvement] Use entity change log to keep CatalogManager catalog cache eventually consistent in HA deployments

After the global entity cache, `CatalogManager` still maintains its own local catalog cache for catalog entities and their corresponding class loaders.

In a multi-node deployment, catalog changes on one server may not invalidate the `CatalogManager` cache on peer servers immediately. This can leave peer nodes using stale catalog entities or stale catalog class loaders until local cache eviction happens.

### How should we improve?

Use the existing `entity_change_log` mechanism to invalidate the `CatalogManager` catalog cache across server instances and provide eventual consistency.

Suggested behavior:

- When catalog-related change log records are consumed on a peer node, invalidate the corresponding `CatalogManager` cache entry.
- Reload the catalog and class loader on the next access.
- Preserve the current eventual consistency model. The stale window should be bounded by the entity change polling interval, which appears to be around 3 seconds by default.
- Handle catalog alter, rename, and drop cases carefully, including invalidating both old and new identifiers when needed.
- Add tests for multi-node style cache invalidation behavior.

### LINKED ISSUE #11233: [Improvement] Add automatic retention cleanup for entity_change_log

`entity_change_log` is currently append-only in runtime. Although there is low-level mapper support for pruning old rows, Gravitino does not have an automatic cleanup mechanism for expired change log records stored in the database.

As the table keeps growing, full change consumption or catch-up scans become slower, and the database keeps consuming more storage for stale change log records.

### How should we improve?

Add automatic retention cleanup for `entity_change_log`.

Suggested behavior:

- Keep one day of change log records by default.
- Automatically delete records older than the retention period.
- Reuse the existing pruning capability where possible, such as `EntityChangeLogMapper.pruneOldEntityChanges`.
- Make the retention period and cleanup interval configurable if needed.
- Ensure cleanup does not break HA cache invalidation semantics or poller startup/catch-up behavior.
- Add tests for the cleanup logic and supported JDBC SQL providers.

yuqi1129:

1. Extract entity change log polling into a shared `EntityChangeLogPoller` in `core` with a listener-based dispatch pattern, replacing the per-consumer polling in `JcasbinChangePoller`.
2. Add `CatalogChangeLogListener` that invalidates `CatalogManager`'s local catalog cache when catalog ALTER/RENAME/DROP changes are consumed, providing eventual consistency across HA nodes.
3. Add automatic retention cleanup for `entity_change_log` — expired rows are pruned periodically (default: 1-hour interval, 1-day retention), with configurable settings.
4. Write change log on all catalog updates (not just renames) so that property-only changes also trigger cross-node cache invalidation.

In a multi-node deployment, catalog changes on one server do not invalidate the `CatalogManager` cache on peer servers until local TTL eviction. This can leave peers using stale catalog entities or class loaders. Additionally, `entity_change_log` grows unboundedly without automatic cleanup.

Fix: #11232
Fix: #11233

Yes — three new configuration keys:
- `gravitino.entityChangeLog.pollIntervalSecs` (default 3)
- `gravitino.entityChangeLog.retentionSecs` (default 86400)
- `gravitino.entityChangeLog.cleanupIntervalSecs` (default 3600)

- `TestEntityChangeLogPoller` — unit tests for dispatch, cursor advancement, listener failure semantics, retention pruning, and cleanup interval gating.
- `TestCatalogManager.testCatalogChangeLogListenerInvalidatesCatalogCache` — verifies cache invalidation on entity change.
- `TestEntityChangeLogService` — extended to cover ALTER (non-rename) change log writing.
- `TestJcasbinChangePoller` — updated for listener-based entity change delivery.


core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
```diff
@@ -0,0 +1,76 @@
...
+      LOG.debug("Invalidating catalog cache due to entity change log: {}", ident);
+      catalogManager.getCatalogCache().invalidate(ident);
+    }
+  }
+
+  private NameIdentifier catalogIdentifier(EntityChangeRecord change) {
+    String[] names = change.getFullName().split("\\.");
+    if (names.length != 2) {
+      LOG.warn("Invalid catalog full name in entity change log: {}", change.getFullName());
+      return null;
```
jerryshao: Use `Optional` rather than null for invalid name identifier.

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Locale;
+import org.apache.gravitino.Entity.EntityType;
+import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.storage.relational.EntityChangeLogListener;
+import org.apache.gravitino.storage.relational.po.cache.EntityChangeRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Invalidates {@link CatalogManager}'s local catalog cache from {@code entity_change_log}. */
+public class CatalogChangeLogListener implements EntityChangeLogListener {
```
jerryshao, line 37: Is this listener a sync or async operation?

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -354,6 +358,34 @@ public void addCatalogCacheRemoveListener(Consumer<NameIdentifier> listener) {
...
+   *     invalidation), false if the change originated from a remote node
+   */
+  boolean consumeLocalMutation(NameIdentifier ident) {
+    boolean[] consumed = {false};
+    localMutationCounts.computeIfPresent(
+        ident,
+        (k, count) -> {
+          consumed[0] = true;
+          return count.decrementAndGet() <= 0 ? null : count;
+        });
```
jerryshao, line 423: Do we have a threading issue for this lambda function call?

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -0,0 +1,238 @@
...
+  /**
+   * Initializes the high-water cursor to the current DB tail and schedules periodic polling.
+   *
+   * <p>Startup does not scan historical changes because this process has no warm local cache yet.
+   */
+  public void start() {
+    entityPollHighWaterId =
+        getOrDefault(
+            SessionUtils.getWithoutCommit(
+                EntityChangeLogMapper.class, EntityChangeLogMapper::selectMaxChangeId));
```
jerryshao, line 130: What if the system get restarted, will it poll from start or from the last record?

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -436,6 +444,16 @@ public void setGravitinoAuthorizer(GravitinoAuthorizer gravitinoAuthorizer) {
...
 
+  /**
+   * Get the global entity change log poller if the current entity store supports it.
+   *
+   * @return the global entity change log poller, or null when it is not available
+   */
+  @Nullable
+  public EntityChangeLogPoller entityChangeLogPoller() {
+    return entityChangeLogPoller;
+  }
```
jerryshao: The `EntityChangeLogPoller` seems like a sub component for entity store, shall we put this into entity store?

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -565,6 +597,18 @@ private void initGravitinoServerComponents() {
...
+    this.catalogChangeLogListener = null;
+    if (entityStore instanceof RelationalEntityStore) {
+      this.entityChangeLogPoller =
+          new EntityChangeLogPoller(
+              config.get(Configs.ENTITY_CHANGE_LOG_POLL_INTERVAL_SECS),
+              TimeUnit.SECONDS.toMillis(config.get(Configs.ENTITY_CHANGE_LOG_RETENTION_SECS)),
+              TimeUnit.SECONDS.toMillis(
+                  config.get(Configs.ENTITY_CHANGE_LOG_CLEANUP_INTERVAL_SECS)));
+      this.catalogChangeLogListener = new CatalogChangeLogListener(catalogManager);
+      this.entityChangeLogPoller.registerListener(catalogChangeLogListener);
```
jerryshao: The registration can be in `CatalogManager`, WDYT?

jerryshao:
I left all the comments here, but you updated without any comments. I'm asking you **questions**, I want you to reply with your thoughts. I hate changing all the things silently with AI.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: I will start from the last record, see `selectMaxChangeId`.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
yuqi1129: ConcurrentHashMap#computeIfPresent is thread-safe and guarded by the CAS mechanism, so it's safe enough, correct me if I misunderstood your word.

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
yuqi1129: For the poll thread it self, it's sync, and it's async for Gravitino API operations.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
yuqi1129: `EntityChangeLogPoller` will be shared by Jasbin and the server, and it's not binding to the entity store scope. It may be like a storage service that relies on the entity store.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
yuqi1129: `entityChangeLogPoller` will be used both by `Jasbin`(server-common) and `catalogManager`.
 Yeah, for the`entityChangeLogPoller` used by the `catalogManager`, we can move it there, and the same goes for `JcasbinAuthorizer`.

yuqi1129:
> I left all the comments here, but you updated without any comments. I'm asking you **questions**, I want you to reply with your thoughts. I hate changing all the things silently with AI.

Sorry, I got it and will not repeat this problem again.

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
```diff
@@ -0,0 +1,103 @@
...
+          LOG.debug("Skipping catalog cache invalidation for local mutation: {}", ident);
+          continue;
+        }
+
+        LOG.debug("Invalidating catalog cache due to entity change log: {}", ident);
+        catalogManager.getCatalogCache().invalidate(ident);
+      } catch (RuntimeException e) {
+        LOG.warn(
+            "Failed to process catalog change log record: fullName={}, entityType={}",
+            change == null ? null : change.getFullName(),
```
jerryshao: In which case the change will be `null`?

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  private boolean isCatalogChange(EntityChangeRecord change) {
+    if (change == null || change.getEntityType() == null) {
+      return false;
+    }
+    return EntityType.CATALOG.name().equals(change.getEntityType().toUpperCase(Locale.ROOT));
+  }
+
+  private Optional<NameIdentifier> catalogIdentifier(EntityChangeRecord change) {
```
jerryshao, line 90: Should you change the `nullability` of the `change` here based on your above code?

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param catalogManager the catalog manager whose local cache should be invalidated
+   */
+  public CatalogChangeLogListener(CatalogManager catalogManager) {
+    this.catalogManager = catalogManager;
+  }
+
+  @Override
+  public void onEntityChange(List<EntityChangeRecord> changes) {
+    for (EntityChangeRecord change : changes) {
```
jerryshao, line 54: If the `change` can be null, why don't you check it here in the first line?

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Creates an {@link EntityChangeLogPoller}.
+   *
+   * @param pollIntervalSecs interval between successive polling cycles
+   * @param retentionMs entity change retention in milliseconds, or 0 to disable cleanup
+   * @param cleanupIntervalMs interval between successive cleanup attempts in milliseconds
+   */
+  public EntityChangeLogPoller(long pollIntervalSecs, long retentionMs, long cleanupIntervalMs) {
+    this(pollIntervalSecs, retentionMs, cleanupIntervalMs, System::currentTimeMillis);
+  }
```
jerryshao, line 83: What is the usage of these two constructors? Are they only for tests?

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -0,0 +1,248 @@
...
+    this.clockMs = clockMs;
+  }
+
+  /**
+   * Registers a listener to receive future entity change batches.
+   *
+   * @param listener the listener to register
+   */
+  public void registerListener(EntityChangeLogListener listener) {
+    Preconditions.checkNotNull(listener, "listener cannot be null");
```
jerryshao: Don't use checkNotNull.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    List<EntityChangeRecord> dispatchedChanges = Collections.unmodifiableList(changes);
+    for (EntityChangeLogListener listener : listeners) {
+      try {
+        listener.onEntityChange(dispatchedChanges);
+      } catch (Exception e) {
+        LOG.warn("Entity change listener {} failed", listener.getClass().getName(), e);
+      }
+    }
```
jerryshao, line 192: Why don't you merge these for..loop together?

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -0,0 +1,248 @@
...
+      try {
+        if (!scheduler.awaitTermination(5, TimeUnit.SECONDS)) {
+          scheduler.shutdownNow();
+        }
+      } catch (InterruptedException e) {
+        scheduler.shutdownNow();
+        Thread.currentThread().interrupt();
+      }
+    }
+  }
```
jerryshao: Put public methods on the top of private methods.

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -184,6 +184,34 @@ private Configs() {}
...
+          .checkValue(value -> value >= 0, ConfigConstants.NON_NEGATIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(DEFAULT_ENTITY_CHANGE_LOG_RETENTION_SECS);
+
+  public static final ConfigEntry<Long> ENTITY_CHANGE_LOG_CLEANUP_INTERVAL_SECS =
+      new ConfigBuilder("gravitino.entityChangeLog.cleanupIntervalSecs")
+          .doc("The interval in seconds for pruning expired entity change logs")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .longConf()
+          .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(DEFAULT_ENTITY_CHANGE_LOG_CLEANUP_INTERVAL_SECS);
```
jerryshao, line 213: Update the docs accordingly.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
jerryshao: I don't agree. This module is highly bound to the storage module, and cannot work without an entity store. It makes more sense to put this module into the entity store.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -565,6 +592,19 @@ private void initGravitinoServerComponents() {
...
+          new EntityChangeLogPoller(
+              config.get(Configs.ENTITY_CHANGE_LOG_POLL_INTERVAL_SECS),
+              TimeUnit.SECONDS.toMillis(config.get(Configs.ENTITY_CHANGE_LOG_RETENTION_SECS)),
+              TimeUnit.SECONDS.toMillis(
+                  config.get(Configs.ENTITY_CHANGE_LOG_CLEANUP_INTERVAL_SECS)));
+      // Registration and local-mutation tracking are encapsulated in CatalogManager so that
+      // CatalogChangeLogListener remains an implementation detail of that class rather than
+      // being wired externally here.
+      catalogManager.registerToPoller(entityChangeLogPoller);
+    }
```
jerryshao: This entityChangeLogPoller is just a side module of the entity store. I don't want to put into the Gravitino as a global component.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinChangePoller.java
```diff
@@ -47,19 +47,15 @@
...
- * a high-water-mark cursor and invalidates the affected keys. Other Gravitino nodes therefore
- * observe ALTER/DROP and owner changes within one poll interval.
+ * <p>This class polls {@code owner_meta} itself and receives {@code entity_change_log} batches from
+ * the global entity change log poller.
  *
- * <p>Both polls run on every tick — a failure in one does not stop the other.
+ * <p>Other Gravitino nodes therefore observe ALTER/DROP and owner changes within one poll interval.
  */
-public class JcasbinChangePoller implements AutoCloseable {
+public class JcasbinChangePoller implements EntityChangeLogListener, AutoCloseable {
```
jerryshao: Is it a listener or a poller?

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: The constructor in L66 is only used for testing, and this one will be used in `GravitinoEnv`. I will add `VisibleForTesting` annotation in L66.

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
yuqi1129: It can't be null strictly based on the mapper, and I will remove the null check here.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
yuqi1129: Let me think twice about it

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: I can't get your points. L164 is about `EntityChangeRecord`, and L171 is for `EntityChangeLogListener.

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
yuqi1129: Fixed, removed all unnecessary null checks

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: done

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: moved.

core/src/main/java/org/apache/gravitino/catalog/CatalogChangeLogListener.java
yuqi1129: No changes needed as `change` can't be null here.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
jerryshao: I see. I missed them together.

core/src/main/java/org/apache/gravitino/Configs.java
yuqi1129: done

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
yuqi1129: I added an interface `SupportsEntityChangeLog` and let `RelationalEntityStore` implement it.

jerryshao:
I have no further comment, let Copilot review again.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -328,6 +339,19 @@ public CatalogManager(Config config, EntityStore store, IdGenerator idGenerator)
...
+    // changes made by this node are not redundantly re-invalidated. Registration is the last step
+    // of
+    // the constructor so the listener never sees a half-initialized manager (catalogCache and
+    // localMutationCounts are already set). CatalogChangeLogListener stays an implementation detail
+    // of this class rather than being wired externally.
+    if (store instanceof SupportsEntityChangeLog) {
+      this.trackLocalMutations = true;
+      ((SupportsEntityChangeLog) store)
+          .registerEntityChangeLogListener(new CatalogChangeLogListener(this));
+    }
```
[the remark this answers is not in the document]
yuqi1129: Fixed in 0a47959e6. `CatalogManager` now keeps the registered `CatalogChangeLogListener` instance and unregisters it in `close()`. It also disables local-mutation tracking and clears pending markers during close.

core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -812,6 +881,7 @@ public boolean dropCatalog(NameIdentifier ident, boolean force)
             // Invalidate after store.delete() to prevent a background thread from repopulating
             // the cache with stale data between invalidate and delete.
             boolean deleted = store.delete(ident, EntityType.CATALOG, true);
+            markLocalMutation(ident);
             catalogCache.invalidate(ident);
```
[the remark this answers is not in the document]
yuqi1129: Fixed the delete path in 0a47959e6: `CatalogManager` now marks a local mutation only when `store.delete(...)` returns `true`. For the update paths, `CatalogMetaService.updateCatalog()` throws an `IOException` when no row is updated, so callers do not reach `markLocalMutation()`. Added regression coverage for the false-delete case.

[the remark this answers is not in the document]
yuqi1129: All were fixed.

roryqi (approved):

PR #11272, opened 2026-05-28
[#10556] improvement(docker): Move container entrypoint to docker/ and run JVM in foreground

### LINKED ISSUE #10556: [Improvement] Unclear Usage Scenarios for start-gravitino.sh and gravitino.sh start Leads to Confusion

The usage scenarios for start-gravitino.sh and gravitino.sh start are not clearly defined, leading to confusion. 
But they actually exist in the image bin folder.

Users are unsure which script to use for starting the Gravitino server, and the intended purpose of each is ambiguous. This overlap makes the startup process error‑prone and complicates automation or troubleshooting. A clear distinction (e.g., one for standalone startup, one for service management) should be documented or the scripts unified to avoid redundancy.

Image

### How should we improve?

Prefer to use more clear script name.

jerryshao:

- Extract the Docker container entrypoint into `dev/docker/gravitino/docker-entrypoint.sh`,
  separated from user-facing scripts in `bin/`.
- Replace the `tail -f /dev/null` keep-alive hack (appended to `gravitino.sh` at build time)
  with a proper foreground server launch via `gravitino.sh run`.
- `bin/start-gravitino.sh` is kept as a deprecated one-release backward-compatibility stub
  that delegates to the new entrypoint.
- `rewrite_gravitino_server_config.py` is moved from `bin/` to `docker/` alongside the entrypoint.

The previous `tail -f /dev/null` approach was fragile: it was appended directly to `gravitino.sh`
at image build time, affecting all invocations of that script. Running the JVM in the foreground
is the standard Docker pattern and more robust for signal handling on container stop.

Fix: #10556

- `bin/start-gravitino.sh` now prints a deprecation warning and delegates to `docker/docker-entrypoint.sh`.
  Users overriding the container entrypoint or calling this script directly should update to
  reference `docker/docker-entrypoint.sh`.

Built a local Docker image and verified:
- Server starts in the foreground via `gravitino.sh run`
- `SKIP_CONFIG_REWRITE=true` correctly skips config rewrite
- `bin/start-gravitino.sh` prints the deprecation warning and delegates correctly
- `docker stop` sends SIGTERM and the container exits cleanly


jerryshao:
@danhuawang can you help to review this PR?

mchades (approved):

PR #11280, opened 2026-05-28
[MINOR] docs: Add Spark REST catalog registration design

FANNG1:

Add a design document for automatically registering Spark lakehouse REST catalogs from Gravitino
REST servers:

`design-docs/spark-rest-catalog-registration.md`

V1 is scoped to Iceberg on Spark. The design covers:

- A server-authoritative catalog-listing endpoint, `GET {iceberg-rest-base}/gravitino/v1/management/catalogs`,
  returning catalog names only, so Spark never guesses names (they match Gravitino names under
  `dynamic-config-provider` but not under `static-config-provider`).
- A single, format-agnostic `GravitinoLakehouseRESTDiscoveryPlugin` that lists catalogs and writes
  `spark.sql.catalog.*` entries. Each lakehouse format plugs in as a provider (Iceberg in V1, Lance
  later) with its own engine runtime and `<format>Rest.*` config; a provider activates only when its
  `uri` is set, so the URI is the switch (no `enable*` flag) and only the providers a user needs are
  on the classpath.
- A single plugin-level `CatalogRegistrationPolicy` (with a `format` argument) deciding which
  catalogs to register and under what Spark name; the default keeps a 1:1 identity.
- Deterministic precedence: the discovery plugin must precede `GravitinoSparkPlugin` (validated,
  fail-fast) and user-written `spark.sql.catalog.*` always wins.
- Listing-call authentication limits (static `token`/`header.*` in V1) and storage-credential
  handling for both vended and non-vended catalogs.
- A Lance provider sketch reusing the same mechanism (future work, not V1).

Spark users must hand-write catalog configuration for every REST catalog, duplicating what the
Gravitino REST server already manages and requiring an edit whenever catalogs change. This design
removes that per-catalog configuration while keeping existing `GravitinoSparkPlugin` behavior
unchanged when the discovery plugin is not added.

Fix: N/A

No. This is a design-document-only change. It proposes new configuration keys for future
implementation, e.g. `spark.sql.gravitino.icebergREST.uri`,
`spark.sql.gravitino.icebergREST.catalogProperties.<key>`, and
`spark.sql.gravitino.REST.registrationPolicy`.

Documentation-only change.


design-docs/unified-engine-access.md
```diff
@@ -0,0 +1,361 @@
...
+  - lance
+  - iceberg
+  - engine-access-mode
+  - native catalog
+license: "This software is licensed under the Apache License version 2."
+---
+
+## Background
+
+Gravitino can manage multiple lakehouse catalogs and lets compute engines access the underlying table
```
jerryshao: You'd better move this design doc to the `design-docs` folder.

FANNG1: Moved the design doc to `design-docs/unified-engine-access.md`.

design-docs/unified-engine-access.md
```diff
@@ -0,0 +1,361 @@
...
+spark.sql.gravitino.lakehouse-lance.engine-access-mode=native
+spark.sql.gravitino.enableIcebergSupport=true
+spark.sql.gravitino.enableLanceSupport=true
+```
+
+The Spark connector then automatically registers Iceberg and Lance catalogs based on the switches.
+Under `auto`, Iceberg uses the Gravitino catalog by default; Lance uses the native catalog.
+
+`spark.sql.gravitino.enableLanceSupport` defaults to `false` to avoid loading Lance catalogs or
+extensions when the user has not explicitly included the Lance Spark connector dependency.
```
jerryshao: How do we support multiple-catalog configuration for Iceberg and Lance?

FANNG1: Thanks for raising this. This is an important point; I need to think through the overall mapping model again, especially how Spark native access should map Gravitino catalogs to the catalogs exposed by Iceberg/Lance REST servers.

jerryshao:
Overall, the general feeling is that the design is mainly to simplify the configuration. But the doc mentions the engine-native, access modes, which may be confused with query engine optimizations or access control things.

Besides, this invents several new terms like engine access mode, native, auto; all these make it hard for users to understand and use, and increase the explanation efforts. I think we'd better simplify the design and make users easy to understand the purpose.

design-docs/unified-engine-access.md
```diff
@@ -0,0 +1,401 @@
...
+## Non-Goals
+
+1. In the first phase, full Lance support across all engines is not required simultaneously.
+
+## Core Design
+
+A new engine-side, provider-level access mode configuration is introduced:
+
+```text
+spark.sql.gravitino.<provider>.engine-access-mode = auto | gravitino | native
```
yuqi1129: You may clarify what `provider` means here? According to the document,  it may refer to catalog providers in the Gravitino catalog. Moreover, you may need to add a document about what kinds of providers it supports.

design-docs/unified-engine-access.md
```diff
@@ -0,0 +1,401 @@
...
+
+```text
+name = iceberg
+type = RELATIONAL
+provider = lakehouse-iceberg
+
+catalog-backend = rest
+uri = http://127.0.0.1:9001/iceberg/
+warehouse = iceberg
+data-access = vended-credentials
```
yuqi1129: Are those configurations for all the catalogs with type `Iceberg`? How can we define catalog-specific configurations?

design-docs/unified-engine-access.md
```diff
@@ -0,0 +1,401 @@
...
+spark.sql.gravitino.uri=http://127.0.0.1:8090
+spark.sql.gravitino.metalake=test
+```
+
+Optional overrides:
+
+```text
+spark.sql.gravitino.lakehouse-iceberg.engine-access-mode=native
+spark.sql.gravitino.lakehouse-generic.engine-access-mode=native
+spark.sql.gravitino.enableIcebergSupport=true
```
yuqi1129: Why do we need to add these extra configurations? What's the problem if we set it to constant value `true`?

FANNG1: Good question. I updated the design to make this explicit.

For Iceberg, REST access is not always equivalent to the existing Gravitino Spark catalog behavior. Enabling it replaces the existing Gravitino Spark catalog registration for matching Iceberg catalogs, so it should be an explicit opt-in to preserve backward compatibility.

For Lance, REST registration needs to load Lance Spark catalog and extension classes. Users may not have the Lance Spark runtime on the classpath, so making it always true could break existing Spark applications. Therefore Lance REST access is also explicitly enabled with `spark.sql.gravitino.lance.enableRestAccess=true`.

design-docs/unified-engine-access.md
FANNG1: Thanks, addressed by simplifying the design. I removed the generic provider-level access-mode framework and clarified the supported Gravitino catalog providers in V1.

The current design only covers:
- `lakehouse-iceberg`
- `lakehouse-generic` with catalog-level `format=lance`

Other providers keep existing behavior and are not part of V1 REST catalog registration.

design-docs/unified-engine-access.md
FANNG1: Thanks for pointing this out. I reworked the design to make the multi-catalog mapping explicit.

For Iceberg, Spark registers one Spark catalog per Gravitino Iceberg catalog. These catalogs can share the same Gravitino Iceberg REST server URI, and Spark uses `warehouse=<gravitino-catalog-name>` to select the target catalog exposed by the Iceberg REST server.

For Lance, Spark also registers one Spark catalog per Gravitino Lance catalog. These catalogs can share the same Lance REST service URI, and Spark uses `parent=<gravitino-catalog-name>` to select the target catalog in the Lance REST Namespace.

The doc now also calls out that Iceberg REST server catalog names must match Gravitino catalog names. `dynamic-config-provider` is the recommended setup; with `static-config-provider`, users need to configure matching names manually.

design-docs/unified-engine-access.md
FANNG1: Yes, in V1 `spark.sql.gravitino.iceberg.enableRestAccess=true` applies to all Gravitino `lakehouse-iceberg` catalogs visible to the Spark connector.

The connector still registers one Spark catalog per Gravitino catalog, using the same catalog name. Catalog-specific routing is handled through generated Spark catalog properties, especially `warehouse=<gravitino-catalog-name>` for Iceberg. Per-catalog enable/disable override is not included in V1 and is documented as future work.

FANNG1:
Thanks, agreed. I simplified the design in this direction.

The updated doc no longer uses the generic engine-native/access-mode framework, and removes the `engine-access-mode`, `native`, and `auto` terminology. The scope is now Spark-only REST catalog registration, with explicit Spark configs:

- `spark.sql.gravitino.iceberg.enableRestAccess`
- `spark.sql.gravitino.iceberg.restUri`
- `spark.sql.gravitino.lance.enableRestAccess`
- `spark.sql.gravitino.lance.restUri`

I also renamed the doc to `design-docs/spark-rest-catalog-registration.md` and updated the PR title/description to match the simplified scope.

FANNG1:
@jerryshao @yuqi1129 PTAL, thx

jerryshao:
> @jerryshao @yuqi1129 PTAL, thx

Sure, I will take a review today.

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,315 @@
...
+
+Optional REST service URI overrides:
+
+| Configuration | Default |
+|---------------|---------|
+| `spark.sql.gravitino.iceberg.restUri` | Inferred from `spark.sql.gravitino.uri` |
+| `spark.sql.gravitino.lance.restUri` | Inferred from `spark.sql.gravitino.uri` |
+
+The Iceberg switch applies to all Gravitino Iceberg catalogs visible to the Spark connector in V1.
+Per-catalog REST registration override is future work.
```
jerryshao: Can you give me an example of how to use this feature, and a comparison before and after?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,315 @@
...
+example, `uri` may be a JDBC URL, Hive Metastore URI, or upstream Iceberg REST URI. Spark should not
+use this catalog property as the Iceberg REST client URI.
+
+Generated Spark configuration:
+
+```text
+spark.sql.catalog.iceberg_prod=org.apache.iceberg.spark.SparkCatalog
+spark.sql.catalog.iceberg_prod.type=rest
+spark.sql.catalog.iceberg_prod.uri=<resolved-iceberg-rest-uri>
+spark.sql.catalog.iceberg_prod.warehouse=iceberg_prod
```
jerryshao: What if the catalog name we what is a different catalog in the Spark side? How do we support this?

design-docs/spark-rest-catalog-registration.md
FANNG1: Good point. I will add a before/after example to make the behavior clearer.

Assume there is a Gravitino Iceberg catalog:

```text
name = iceberg_jdbc
type = RELATIONAL
provider = lakehouse-iceberg

catalog-backend = jdbc
uri = jdbc:postgresql://127.0.0.1:5432/iceberg
warehouse = s3://warehouse/iceberg_jdbc
```

Also assume the Gravitino Iceberg REST server exposes a catalog named `iceberg_jdbc`, and its backend configuration is consistent with this Gravitino Iceberg catalog.

This is naturally satisfied when the Iceberg REST server uses the dynamic config provider, because it loads catalog configuration from Gravitino and exposes catalogs by Gravitino catalog name. This is the recommended scenario. If the static config provider is used, users must manually keep the REST server catalog name and backend configuration consistent with the Gravitino catalog.

Before enabling this option, Spark keeps the existing Gravitino Spark connector behavior:

```text
spark.sql.gravitino.iceberg.enableRestAccess=false
```

Catalog metadata access goes through the Gravitino REST protocol. For Iceberg table access, the Iceberg metadata used by the IO path, such as table metadata and manifest metadata, is still loaded through the Iceberg JDBC catalog implementation behind the Gravitino Iceberg catalog.

After enabling this option:

```text
spark.sql.gravitino.iceberg.enableRestAccess=true
spark.sql.gravitino.iceberg.restUri=http://127.0.0.1:9001/iceberg/
```

The Spark connector registers a Spark Iceberg REST catalog using the Gravitino catalog name:

```text
spark.sql.catalog.iceberg_jdbc=org.apache.iceberg.spark.SparkCatalog
spark.sql.catalog.iceberg_jdbc.type=rest
spark.sql.catalog.iceberg_jdbc.uri=http://127.0.0.1:9001/iceberg/
spark.sql.catalog.iceberg_jdbc.warehouse=iceberg_jdbc
```

Then Spark accesses Iceberg metadata through the Iceberg REST protocol. The Spark side no longer needs direct exposure to the Iceberg JDBC catalog configuration; it only needs the Iceberg REST endpoint and the Gravitino catalog name. The backend catalog configuration is owned by the Iceberg REST server.

design-docs/spark-rest-catalog-registration.md
FANNG1: Thanks for pointing this out. I think this corresponds to the case where the Gravitino Iceberg catalog name and the Iceberg REST server catalog name are different.

Technically, we could support this by adding a catalog-level alias property, for example:

```text
iceberg-rest-catalog-name = <iceberg-rest-server-catalog-name>
```

Then Spark would still register the Spark catalog using the Gravitino catalog name, but generate:

```text
spark.sql.catalog.<gravitino-catalog-name>.warehouse=<iceberg-rest-server-catalog-name>
```

instead of:

```text
spark.sql.catalog.<gravitino-catalog-name>.warehouse=<gravitino-catalog-name>
```

However, this adds another name-mapping layer for users to understand and maintain. I am not sure whether this is a common requirement. The current design assumes the Gravitino Iceberg REST server exposes catalogs by Gravitino catalog name, which is naturally satisfied by the dynamic config provider and is also the recommended deployment model.

So my preference is to keep V1 simple and not support a separate REST catalog alias first.

What do you think? Do you think the alias case is important enough to include in V1?

FANNG1:
Hi @jerryshao @yuqi1129, I've submitted a draft PR (#11838) implementing the Iceberg portion of this design — `spark.sql.gravitino.iceberg.enableRestAccess` auto-registration with `dynamic-config-provider` integration tests and docs. Happy to continue any design-level discussion here, or if the direction looks reasonable, we can shift to reviewing the implementation directly on the PR. Let me know which works better for you.

design-docs/spark-rest-catalog-registration.md
jerryshao: My question is that:

1) What if user they want a different catalog name for Spark other than the one in Gravitino? Not sure if it is a valid scenario
2) If we have multiple IRC, how do we support this? What if the catalog name is conflicted?
3) What if they only want to register some catalogs (not all) selectively?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,315 @@
...
+
+The Iceberg switch applies to all Gravitino Iceberg catalogs visible to the Spark connector in V1.
+Per-catalog REST registration override is future work.
+
+## Registration Precedence
+
+For each Spark catalog name, the connector should register exactly one Spark catalog implementation.
+When `spark.sql.gravitino.iceberg.enableRestAccess=true`, REST catalog registration replaces the
+existing Gravitino Spark catalog registration for matching `lakehouse-iceberg` catalogs and keeps the
+same Spark catalog name.
```
jerryshao: What if users already have exsting catalog configurations in the Spark conf. And then, they update the conf to enable this feature, who takes precedence?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,315 @@
...
+
+1. V1 is Spark only.
+2. Iceberg REST catalog registration is controlled by
+   `spark.sql.gravitino.iceberg.enableRestAccess` and applies to all Gravitino Iceberg catalogs
+   visible to the Spark connector. Per-catalog override is future work.
+3. Lance REST catalog registration requires `spark.sql.gravitino.lance.enableRestAccess=true` and
+   `lakehouse-generic + format=lance`.
+4. V1 supports `lakehouse-iceberg` and `lakehouse-generic + format=lance` only.
+5. A dedicated `lakehouse-lance` provider can be discussed after the
+   `lakehouse-generic + format=lance` path is validated.
```
jerryshao: I'm thinking about the actual benefits of this feature. Seems like introducing this will simplify the Spark side catalog configuration for IRC. But, my rough feeling is that the simplification is quite limited; we still need to configure some additional confs that are not covered by this auto-mapping solution.

In the meantime, introducing a new concept that is different from Spark's default behavior is also a point that concerns me.

Please correct me if I'm wrong.

roryqi:
A little concern about this feature: It seems bring more complexity than pain point reduce. It has many different cases. It seems hard to solve the cases.

lasdf1234:
Many special cases need to be taken into consideration.
Another question I have is: If this function is to be implemented, will it be extended to the connectors for Flink and Trino in the future?

design-docs/spark-rest-catalog-registration.md
FANNG1: For the first case, I think user-provided Spark catalog configuration should take precedence. If `spark.sql.catalog.<name>` is already configured by the user, the auto-registration logic should not override it. In that case, users can still use any Spark-side catalog name they want by configuring Spark manually.

For the second and third cases, technically we could support them by adding catalog-level properties, for example a Spark catalog alias, per-catalog REST URI, or an include/exclude flag. But that would make both the Gravitino catalog model and the Spark-side behavior more complicated.

This keeps the feature focused on reducing the common duplicated Spark catalog configuration, without introducing another layer of catalog-level mapping rules too early.

FANNG1:
Thanks for all the review comments and suggestions.

Let me first explain our use case.

We are building a multi-tenant compute platform. Each tenant is assigned separate storage resources, such as buckets and AK/SK credentials. For each tenant, we create separate Iceberg REST catalogs and Lance catalogs. As the number of tenants grows, especially when new tenants are added, maintaining client-side Spark catalog configuration becomes difficult. In some environments, upgrading or changing client-side configuration is also not easy.

So dynamic discovery and automatic registration of Iceberg and Lance REST catalogs is a real pain point for us.

I agree that the current proposal can make setup easier under certain conditions, but it also introduces some understanding cost and complexity when we try to extend it to more cases.

Another possible direction is to add a `catalogs` API to the Iceberg/Lance REST catalog service. The API could return the catalogs managed by the REST service, and the client side could discover and register them automatically from that API. This direction may also be easier to extend to other connectors, such as Flink and Trino, because the discovery logic would live closer to the REST catalog service instead of being tied to Spark-specific Gravitino catalog mapping.

If the current Gravitino-driven mapping approach feels too complicated, what do you think about this alternative direction?

jiangxt2:
One thought on simplifying the design: instead of adding user-facing configs like `enableRestAccess`, the connector could transparently probe REST endpoint reachability and silently prefer REST when available. If the user has already configured `spark.sql.catalog.<name>`, the connector skips that catalog entirely and respects the users config. For unconfigured catalogs, probe the endpoint — if reachable, register as REST, otherwise fall back to proxy.

FANNG1:
Updated the design based on the earlier feedback. Main changes:

- Dropped the `enableRestAccess` flag. A separate plugin (`GravitinoIcebergRestSparkPlugin`) is now the switch — its presence enables the feature, the existing `GravitinoSparkPlugin` behavior is unchanged, and the Iceberg runtime stays isolated.
- Discovery is server-authoritative: a new `GET /gravitino/v1/catalogs` on the Iceberg REST server lists the catalogs, so Spark never guesses names — this removes the `static-config-provider` name-mapping problem.
- Registration is driven by a name-only `CatalogRegistrationPolicy`, with deterministic precedence: plugin ordering is validated (fail-fast) and user-written `spark.sql.catalog.*` always wins.

Could you take another look? @jerryshao @yuqi1129 @roryqi @lasdf1234 @jiangxt2

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,338 @@
...
+authorization-filtered, so every authenticated caller sees all catalog names. Filtering it with
+`MetadataAuthzHelper`, as the Iceberg server already does for namespace listing, is a prerequisite —
+not a follow-up.
+
+---
+
+## Limitations and Future Work
+
+1. **Flink and Trino**: The server side is reusable as-is; only registration differs. Flink allows
+   one catalog store per session, so it would resolve REST catalogs lazily inside
```
lasdf1234, line 356: Flink and Trino can indeed be implemented in future versions, but there are some issues that I have doubts about:
1. If an attribute is available in Spark but needs to be disabled in Flink or Trino, how should this situation be handled?
2. If an attribute is available in Spark but has a different name in Flink or Trino, how can this be resolved?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,338 @@
...
+  ]
+}
+```
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `catalogs[].name` | string | Catalog name as accepted by this server's `warehouse` parameter |
+| `catalogs[].properties` | map | Reserved for non-sensitive, client-relevant metadata; V1 defines no keys. Per-catalog client configuration already arrives via `GET /v1/config?warehouse=<name>`, so this endpoint only enumerates names |
+
+### GravitinoIcebergRestSparkPlugin
```
roryqi: Could we have a plugin which called `GravitinoLakehouseRESTDiscoveryPlugin`?  It will be difficult to maintain if every lakehouse has a dedicated plugin. WDYT?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,338 @@
...
+    { "name": "iceberg_prod", "properties": {} },
+    { "name": "iceberg_audit", "properties": {} }
+  ]
+}
+```
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `catalogs[].name` | string | Catalog name as accepted by this server's `warehouse` parameter |
+| `catalogs[].properties` | map | Reserved for non-sensitive, client-relevant metadata; V1 defines no keys. Per-catalog client configuration already arrives via `GET /v1/config?warehouse=<name>`, so this endpoint only enumerates names |
```
roryqi: Could u elaborate more about properties? Why do we need this?

design-docs/spark-rest-catalog-registration.md
FANNG1: Good call — done. The design now uses a single format-agnostic `GravitinoLakehouseRESTDiscoveryPlugin` instead of one plugin per format. Each lakehouse format is a **provider** (Iceberg in V1, Lance later) with its own engine runtime and `<format>Rest.*` config. A provider activates only when its `uri` is set — the URI is the switch, so there's no `enable*` flag — and dependency isolation is kept by loading providers from the classpath (fail-fast if a `uri` is set but its provider jar is missing). So `spark.plugins` lists one plugin and there's a single ordering rule. The registration policy is now a single plugin-level setting shared across providers, with a `format` argument on its methods so one policy can still apply format-specific rules.

design-docs/spark-rest-catalog-registration.md
FANNG1: It was only a placeholder for future extension, not used in V1, so I removed it. The response returns names only; per-catalog config already comes from `GET /v1/config?warehouse=<name>`, and since JSON extends compatibly a field can be added later if an engine ever needs it.

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,360 @@
...
+
+The provider issues the listing request with Iceberg's own `RESTClient`
+(`org.apache.iceberg.rest.HTTPClient`, the one `RESTCatalog` uses), not the Gravitino client: the
+target is the Iceberg REST server, the Iceberg runtime is already on its classpath, and it parses
+`ErrorResponse` the same way table calls do (authentication below). It uses the
+`spark.sql.gravitino.icebergRest.*` prefix, disjoint from the existing plugin's keys:
+
+| Configuration | Required | Default | Description |
+|---------------|----------|---------|-------------|
+| `spark.sql.gravitino.icebergRest.uri` | Yes | None | Base URI of the Iceberg REST server, e.g. `http://127.0.0.1:9001/iceberg/`; setting it activates the Iceberg provider |
```
roryqi: `icebergRest` -> `icebergREST`?

roryqi:
Just a minor.

design-docs/spark-rest-catalog-registration.md
FANNG1: Good questions. The key point is that the shared layer carries no engine client properties, so there is nothing to disable or rename there:

- The listing endpoint returns catalog **names only**. Client config comes from Iceberg's `GET /v1/config`, in Iceberg's own property vocabulary, which the Iceberg REST catalog for Spark/Flink/Trino all consume identically — no cross-engine mapping needed.
- What is engine-specific is only the catalog **wiring** (Spark: `spark.sql.catalog.*` + `type=rest`; Trino: `CREATE CATALOG ... USING iceberg` + `iceberg.catalog.type=rest`). Each engine's registration emits that in its own vocabulary, so it produces only the keys that engine supports (your #1) and uses that engine's names (your #2) — there is no shared property set to translate.
- `catalogProperties` is Spark-namespaced (copied to `spark.sql.catalog.<name>.<key>`); Flink/Trino would have their own equivalent input, so a Spark-only or differently-named property simply is not part of another engine's input.

lasdf1234: By using different configurations, different engines can be enabled, is that what you mean?
So, if it's a shared configuration, what would the key be like?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,360 @@
...
+For each accepted catalog it generates:
+
+```text
+spark.sql.catalog.<sparkName>=org.lance.spark.LanceNamespaceSparkCatalog
+spark.sql.catalog.<sparkName>.impl=rest
+spark.sql.catalog.<sparkName>.uri=<spark.sql.gravitino.lanceRest.uri>
+spark.sql.catalog.<sparkName>.parent=<advertisedName>
+```
+
+`parent` is Lance's `warehouse`, and the four core keys are the implementation key, `impl`, `uri`,
```
yuqi1129: "`parent` is Lance's `warehouse`", what's the meaning of `warehouse` here? I remembered that it should be catalog name here.

FANNG1: You're right that the value is the catalog name. `parent` is set to `<advertisedName>` (the catalog name); the phrase only meant it plays the same catalog-selecting role that `warehouse` plays in the Iceberg generated config. I've reworded the doc to say that explicitly.

design-docs/spark-rest-catalog-registration.md
FANNG1: Good point — renamed `icebergRest`/`lanceRest`/`rest` to `icebergREST`/`lanceREST`/`REST` across the config keys to match the plugin name.

design-docs/spark-rest-catalog-registration.md
FANNG1: Right — each engine is enabled through its own native config; there is no shared cross-engine key. The same setting is just named per engine. For the Iceberg REST server URI:

- Spark: `spark.sql.gravitino.icebergREST.uri`
- Trino: `iceberg.rest-catalog.uri` (with `connector.name=iceberg`, `iceberg.catalog.type=rest`)
- Flink: a `uri` catalog option on the Gravitino catalog store

The only shared part is server-side (the listing endpoint + `/v1/config`), which lives on the REST server, not in engine config.

roryqi:
No other comments.

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,361 @@
...
+  client side — through environment variables or the storage SDK's default credential chain, or
+  per-catalog sub-keys such as `spark.sql.catalog.<name>.s3.access-key-id`.
+
+### Plugin ordering and precedence
+
+When both plugins are configured, `GravitinoLakehouseRESTDiscoveryPlugin` **must be listed first**:
+
+```text
+spark.plugins=org.apache.gravitino.spark.connector.plugin.GravitinoLakehouseRESTDiscoveryPlugin,\
+              org.apache.gravitino.spark.connector.plugin.GravitinoSparkPlugin
```
yuqi1129, line 240: Why do we need this plugin(GravitinoSparkPlugin) for the Iceberg REST service? Does Lance REST need it?

FANNG1: No. `GravitinoSparkPlugin` and the REST discovery plugin serve two independent access paths:

- `GravitinoSparkPlugin` accesses catalogs registered in Gravitino through Gravitino's **generic catalog API** — Hive, Iceberg, JDBC, Paimon.
- `GravitinoLakehouseRESTDiscoveryPlugin` accesses catalogs through the **lakehouse format's own REST protocol**. It is self-contained: it only needs `icebergREST.uri` (or `lanceREST.uri`), not `spark.sql.gravitino.uri`/`metalake`, and it does not create a Gravitino client.

So a session that only needs Iceberg-over-IRC, or only Lance, configures the discovery plugin alone. `GravitinoSparkPlugin` is only added when the same session **also** needs Hive/JDBC/Paimon or non-REST Iceberg catalogs — that is the only case the ordering rule applies to.

For Lance specifically it is not even an option: `GravitinoSparkPlugin` has no Lance catalog implementation (`CatalogNameAdaptor` covers hive / lakehouse-iceberg / lakehouse-paimon / jdbc only), so it would just log `Skip registering <name> because lakehouse-generic is not supported yet`. REST discovery is the only Spark path for Lance.

The section reads as if the plugin were required, which is my doc's fault — I'll retitle it "Coexistence with GravitinoSparkPlugin (optional)" and state the two access paths explicitly up front.

yuqi1129: I see. I have no more comments.

FANNG1:
@jerryshao do you have any other comments?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,361 @@
...
+plugin, `GravitinoLakehouseRESTDiscoveryPlugin`, that consumes it — plus one **ordering rule** that
+makes the interaction with `GravitinoSparkPlugin` deterministic. Each lakehouse format plugs into
+the plugin as a provider; V1 ships Iceberg.
+
+### Catalog-listing endpoint
+
+#### GET `{iceberg-rest-base}/gravitino/v1/catalogs`
+
+Placed outside the Iceberg REST specification's `/v1/` namespace (default deployment:
+`http://<host>:9001/iceberg/gravitino/v1/catalogs`) to mark it as a Gravitino-private extension.
```
jerryshao: I would suggest we use `management`, we will also add other APIs under this scope.

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,361 @@
...
+
+The registration policy and user-configuration precedence below are shared by every provider; the
+listing client, generated entries, and credential handling are provider-specific.
+
+#### Registration policy interface
+
+A single plugin-level policy, selected with `spark.sql.gravitino.REST.registrationPolicy`, applies
+to every provider (the default implementation applies when unset). Each method receives the `format`
+that advertised the catalog — the same token as the config prefix, `"iceberg"` or `"lance"` — so one
+policy can still apply format-specific rules without a policy per format.
```
jerryshao, line 131: What is the meaning of `so one
policy can still apply format-specific rules without a policy per format.`? Can you explain more?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,361 @@
...
+  boolean shouldRegister(String format, String catalogName);
+
+  /**
+   * The Spark catalog name to register an accepted catalog under. Defaults to the advertised name.
+   *
+   * @param format the lakehouse format that advertised the catalog
+   * @param catalogName the accepted catalog name
+   * @return the Spark catalog name
+   */
+  default String sparkCatalogName(String format, String catalogName) {
```
jerryshao: Can this interface be used by Trino and Flink? If so, we should make it more general to avoid mentioning the "Spark" only.

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,361 @@
...
+  "catalogs": [
+    { "name": "iceberg_prod" },
+    { "name": "iceberg_audit" }
+  ]
+}
+```
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `catalogs[].name` | string | Catalog name as accepted by this server's `warehouse` parameter |
```
jerryshao, line 97: Is returning the catalog name enough? Do you need more information about catalog properties?

design-docs/spark-rest-catalog-registration.md
```diff
@@ -0,0 +1,361 @@
...
+The default implementation registers every advertised catalog under its advertised name, keeping a
+1:1 identity between the Spark catalog name and the REST server catalog name. Deployments that need
+to register a subset, rename catalogs on the Spark side, or treat formats differently implement the
+interface and point `registrationPolicy` at their class.
+
+#### User configuration precedence
+
+User configuration has the highest priority; the plugin fills in only what the user left unset.
+Precedence for each registered catalog, high to low, enforced by the plugin so no policy can weaken
+it:
```
jerryshao, line 170: What if both are set: the user's static configurations and this plugin?

design-docs/spark-rest-catalog-registration.md
FANNG1: Agreed, updated to http://<host>:9001/iceberg/gravitino/v1/management/catalogs

design-docs/spark-rest-catalog-registration.md
FANNG1: Names are enough for V1 — the generated entry needs only the name plus the shared `uri`, and the rest comes from `GET /v1/config?warehouse=<name>` that `RESTCatalog` fetches at init anyway.

Each entry is an object (`{"name": "..."}`), so fields can be added compatibly later. Any specific property in mind?

design-docs/spark-rest-catalog-registration.md
FANNG1: They can be used together, and the user’s static configuration always takes precedence. If `spark.sql.catalog.<name>` is explicitly set, the plugin skips that catalog entirely. If only per-catalog sub-properties are set, the plugin fills in the missing properties without overwriting user values.

design-docs/spark-rest-catalog-registration.md
FANNG1: It means the plugin configures only one `CatalogRegistrationPolicy` implementation. The `format` argument tells that policy which provider advertised the catalog, such as `iceberg` or `lance`, so the implementation can apply different filtering or alias rules for each format when needed. Therefore, users do not need to configure separate `icebergRegistrationPolicy` and `lanceRegistrationPolicy` classes.

design-docs/spark-rest-catalog-registration.md
FANNG1: renamed to `registeredCatalogName`

FANNG1:
@jerryshao comments are addressed, please take a look again.

jerryshao:
I have no comments, let's merge it first.

jerryshao (approved):

PR #11288, opened 2026-05-28
[#11287] feat(spark-connector): Add view support for Hive catalog

### LINKED ISSUE #11287: [FEATURE] Add view support to Spark connector
### Describe the feature

Support reading Gravitino views via Spark SQL on Gravitino Hive catalogs.

### Motivation

Gravitino exposes a ViewCatalog API and the Hive backend already stores
views in HMS. The Spark connector should allow users to query these views
via Spark SQL.

### Describe the solution

Route view loading through ViewCatalog API in BaseCatalog, and implement
SparkHiveView as a V2 Table that executes the view SQL via SparkSession.

### Additional context

No additional context.

diqiu50:

Add read-only view support to the Gravitino Spark Hive connector:

- `HiveViewCatalogOperations`: accept Spark dialect in view creation and loading.
- `BaseCatalog`: fall through to view lookup in `loadTable` when identifier is
  not a table; check view existence in `tableExists`.
- `SparkHiveView`: new V2 Table that executes view SQL via SparkSession on the
  driver and returns rows as a single in-memory partition.
- `GravitinoHiveCatalog`: override `createSparkView` to return `SparkHiveView`.

Fix: #11287

Yes — users can now SELECT from Gravitino views via Spark SQL on Hive catalogs.
Note: CREATE VIEW via Spark SQL is not supported due to a Spark V2 named catalog
limitation; views must be created through the Gravitino API.

- Unit test: `testCreateViewAcceptsSparkDialect` in `TestHiveCatalogOperations`
- Integration tests: 4 new cases in `SparkHiveCatalogIT`:
  `testSelectHiveView`, `testTableExistsForView`, `testShowTablesExcludesViews`,
  `testCreateViewViaSql`

### DIFF
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
@@ -52,7 +52,7 @@ import org.apache.gravitino.rel.View;
 @ToString
 public class HiveView implements View {
 
-  private static final String SPARK_VERSION_KEY = "spark.sql.create.version";
+  static final String SPARK_VERSION_KEY = "spark.sql.create.version";
   static final String SPARK_DEFAULT_CATALOG_KEY = "gravitino.view.default.catalog";
   static final String SPARK_DEFAULT_SCHEMA_KEY = "gravitino.view.default.schema";
   static final String FLINK_PROPERTY_PREFIX = "flink.";
@@ -132,14 +132,6 @@ public class HiveView implements View {
       return Dialects.FLINK;
     }
     return Dialects.HIVE;
-  }
-
-  /**
-   * Returns true if {@code properties} contains the Spark dialect marker key {@code
-   * spark.sql.create.version}, which {@link #detectDialect} requires to identify a view as Spark.
-   */
-  static boolean hasSparkMarker(Map<String, String> properties) {
-    return properties.containsKey(SPARK_VERSION_KEY);
   }
 
   /** Builder for {@link HiveView}. */
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
@@ -53,6 +53,8 @@ import org.apache.gravitino.rel.View;
 public class HiveView implements View {
 
   static final String SPARK_VERSION_KEY = "spark.sql.create.version";
+  static final String SPARK_DEFAULT_CATALOG_KEY = "gravitino.view.default.catalog";
+  static final String SPARK_DEFAULT_SCHEMA_KEY = "gravitino.view.default.schema";
   private static final String TRINO_VIEW_MARKER_KEY = "presto_view";
   private static final String TRINO_VIEW_PREFIX = "/* Presto View:";
 
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
@@ -53,8 +53,6 @@ import org.apache.gravitino.rel.View;
 public class HiveView implements View {
 
   static final String SPARK_VERSION_KEY = "spark.sql.create.version";
-  static final String VIEW_DEFAULT_CATALOG_KEY = "gravitino.view.default.catalog";
-  static final String VIEW_DEFAULT_SCHEMA_KEY = "gravitino.view.default.schema";
   static final String FLINK_PROPERTY_PREFIX = "flink.";
   private static final String TRINO_VIEW_MARKER_KEY = "presto_view";
   private static final String TRINO_VIEW_PREFIX = "/* Presto View:";
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
@@ -52,7 +52,7 @@ import org.apache.gravitino.rel.View;
 @ToString
 public class HiveView implements View {
 
-  private static final String SPARK_VERSION_KEY = "spark.sql.create.version";
+  static final String SPARK_VERSION_KEY = "spark.sql.create.version";
   private static final String TRINO_VIEW_MARKER_KEY = "presto_view";
   private static final String TRINO_VIEW_PREFIX = "/* Presto View:";
 
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
@@ -52,7 +52,7 @@ import org.apache.gravitino.rel.View;
 @ToString
 public class HiveView implements View {
 
-  static final String SPARK_VERSION_KEY = "spark.sql.create.version";
+  private static final String SPARK_VERSION_KEY = "spark.sql.create.version";
   static final String SPARK_DEFAULT_CATALOG_KEY = "gravitino.view.default.catalog";
   static final String SPARK_DEFAULT_SCHEMA_KEY = "gravitino.view.default.schema";
   static final String FLINK_PROPERTY_PREFIX = "flink.";
@@ -132,6 +132,14 @@ public class HiveView implements View {
       return Dialects.FLINK;
     }
     return Dialects.HIVE;
+  }
+
+  /**
+   * Returns true if {@code properties} contains the Spark dialect marker key {@code
+   * spark.sql.create.version}, which {@link #detectDialect} requires to identify a view as Spark.
+   */
+  static boolean hasSparkMarker(Map<String, String> properties) {
+    return properties.containsKey(SPARK_VERSION_KEY);
   }
 
   /** Builder for {@link HiveView}. */
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
@@ -55,6 +55,7 @@ public class HiveView implements View {
   static final String SPARK_VERSION_KEY = "spark.sql.create.version";
   static final String SPARK_DEFAULT_CATALOG_KEY = "gravitino.view.default.catalog";
   static final String SPARK_DEFAULT_SCHEMA_KEY = "gravitino.view.default.schema";
+  static final String FLINK_PROPERTY_PREFIX = "flink.";
   private static final String TRINO_VIEW_MARKER_KEY = "presto_view";
   private static final String TRINO_VIEW_PREFIX = "/* Presto View:";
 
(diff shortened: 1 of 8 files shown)


spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/hive/SparkHiveView.java
```diff
@@ -0,0 +1,231 @@
...
+
+    @Override
+    public InputPartition[] planInputPartitions() {
+      try {
+        InternalRow[] rows = spark.sql(viewSql).queryExecution().executedPlan().executeCollect();
+        return new InputPartition[] {new ViewPartition(rows)};
+      } catch (Exception e) {
+        throw new RuntimeException(
+            String.format("Failed to execute view SQL [%s]: %s", viewSql, e.getMessage()), e);
+      }
```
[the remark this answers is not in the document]
diqiu50: The function doest not declared  throw exceptions

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
```diff
@@ -421,6 +435,45 @@ public boolean dropNamespace(String[] namespace, boolean cascade)
...
+   * @throws NoSuchTableException if no view exists or views are not supported by this catalog
+   */
+  protected Table loadViewAsTable(Identifier ident) throws NoSuchTableException {
+    View gravitinoView;
+    try {
+      gravitinoView =
+          gravitinoCatalogClient
+              .asViewCatalog()
+              .loadView(NameIdentifier.of(getDatabase(ident), ident.name()));
+    } catch (NoSuchViewException | UnsupportedOperationException e) {
```
mchades: `loadView()` can throw `ForbiddenException` (an unchecked exception) when the caller lacks `LOAD_VIEW` privilege. This will bubble out of `loadTable()` which only declares `throws NoSuchTableException`, breaking the API contract. `tableExists()` already handles `ForbiddenException` explicitly — the same treatment is needed here:

```java
} catch (NoSuchViewException | UnsupportedOperationException | ForbiddenException e) {
    throw new NoSuchTableException(ident);
}
```

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -363,14 +369,17 @@ private HiveView toHiveView(
...
     String detectedDialect = HiveView.detectDialect(representationSql, params);
-    if (!Dialects.HIVE.equalsIgnoreCase(detectedDialect)
-        && !Dialects.FLINK.equalsIgnoreCase(detectedDialect)) {
-      // TODO(design-docs/gravitino-logical-view-management.md): support loading trino/spark HMS
-      // views.
-      throw new UnsupportedOperationException(
-          String.format(
-              "Hive catalog currently supports only '%s' and '%s' view dialects, but found '%s' for view %s",
-              Dialects.HIVE, Dialects.FLINK, detectedDialect, ident));
+    switch (detectedDialect.toLowerCase(java.util.Locale.ROOT)) {
```
mchades: `java.util.Locale.ROOT` is a fully-qualified class name used inside a method body, which violates the project coding style (AGENTS.md: prefer normal imports over FQN). Please add `import java.util.Locale;` and replace all three occurrences (lines 372, 420, 444) with `Locale.ROOT`.

spark-connector/spark-common/src/test/java/org/apache/gravitino/spark/connector/integration/test/hive/SparkHiveCatalogIT.java
```diff
@@ -517,4 +525,160 @@ void testCreateTableWithTimestamp() {
...
+    String tableName = "view_base_table";
+    String viewName = "test_hive_view";
+    String schemaName = getDefaultDatabase();
+
+    dropTableIfExists(tableName);
+    createSimpleTable(tableName);
+    sql(String.format("INSERT INTO %s VALUES (1, '1', 1),(2, '2', 2),(3, '3', 3)", tableName));
+
+    ViewCatalog viewCatalog =
+        client.loadMetalake("test").loadCatalog(getCatalogName()).asViewCatalog();
```
mchades: These three tests all repeat the same boilerplate `client.loadMetalake("test").loadCatalog(getCatalogName()).asViewCatalog()` chain to create views via the Gravitino API. Since `SparkEnvIT.metalakeName` is `private`, the metalake name had to be hardcoded as `"test"`.

Consider adding a helper method in `SparkEnvIT` (e.g. `protected Catalog getGravitinoCatalog()`) that encapsulates this lookup, so subclasses do not need to repeat the chain or hard-code the metalake name.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -418,27 +429,31 @@ private SQLRepresentation validateSQLRepresentation(
             defaultSchema,
             ident);
         return selected;
+      case Dialects.SPARK:
```
mchades, line 451: When `representation.dialect()` is `"spark"`, `detectDialect` identifies the view as Spark on reload only if `spark.sql.create.version` is present in the HMS table parameters — but `createView()` does not validate that this key exists. If a caller omits it, the view silently round-trips as Hive dialect and `defaultCatalog`/`defaultSchema` will not be restored on load.

Consider adding a validation here for the Spark case: require `spark.sql.create.version` to be present in `params`, and throw `IllegalArgumentException` with a clear message if it is missing.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/hive/SparkHiveView.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Override
+    public StructType readSchema() {
+      return schema;
+    }
+
+    @Override
+    public InternalRow[] rows() {
+      try {
+        return spark.sql(viewSql).queryExecution().executedPlan().executeCollect();
```
mchades, line 185: `executeCollect()` executes the physical plan and collects all resulting `InternalRow`s into driver memory. For a view like `SELECT * FROM large_table`, every row from the base table will be pulled to the driver JVM heap, which can cause `OutOfMemoryError` for large datasets.

The class Javadoc currently says *"never serializes rows to executors"*, which reads like an advantage rather than a limitation. Please add a warning on the class or on `rows()`, e.g.:
```java
// WARNING: executeCollect() materializes the entire result set to driver memory.
// Suitable only for small/bounded views; large result sets may cause OutOfMemoryError.
```

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
```diff
@@ -52,7 +52,9 @@
 @ToString
 public class HiveView implements View {
 
-  private static final String SPARK_VERSION_KEY = "spark.sql.create.version";
+  static final String SPARK_VERSION_KEY = "spark.sql.create.version";
```
mchades, line 55: `SPARK_VERSION_KEY` was widened from `private` to package-private, but no other class in this PR references `HiveView.SPARK_VERSION_KEY`. `SPARK_DEFAULT_CATALOG_KEY` and `SPARK_DEFAULT_SCHEMA_KEY` are correctly package-private (they are referenced in `HiveViewCatalogOperations`), but `SPARK_VERSION_KEY` is only used inside `HiveView` itself. Please revert it to `private`.

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
```diff
@@ -312,11 +315,22 @@ public boolean tableExists(Identifier ident) {
       loadGravitinoTable(ident);
       return true;
     } catch (NoSuchTableException e) {
-      return false;
+      // fall through to view check
```
mchades, line 318: `tableExists()` now returns `true` for views as well as tables, but this behavioral change is not reflected in the Javadoc. Please update it to document that the method checks both tables and views.

mchades, line 318: Spark 3.4+ has a `ViewCatalog` interface (`org.apache.spark.sql.connector.catalog.ViewCatalog` in `spark-catalyst`) that enables `SHOW VIEWS`, `CREATE VIEW`, `DROP VIEW` etc. via Spark SQL. Since this project already has version-specific subclasses (`GravitinoHiveCatalogSpark34`, etc.) for Spark 3.4/3.5, it seems feasible to implement `ViewCatalog` there without breaking 3.3 support. Is there a reason it is not included in this PR? Should it be tracked as a follow-up?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
diqiu50: SPARK_VERSION_KEY is used by HiveViewCatalogOperations

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
diqiu50: CREATE VIEW via SQL is not supported for V2 named catalogs. Spark's
ResolveSessionCatalog unconditionally throws for any non-session catalog
(see the "case CreateView(ResolvedIdentifier(catalog, _), ...)" branch), regardless of
 whether the catalog implements ViewCatalog.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
```diff
@@ -52,7 +52,10 @@
...
+  static final String SPARK_DEFAULT_CATALOG_KEY = "gravitino.view.default.catalog";
+  static final String SPARK_DEFAULT_SCHEMA_KEY = "gravitino.view.default.schema";
```
mchades: Does Spark support these two properties? How are they used?

diqiu50: Spark does not support these two properties, removed

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
```diff
@@ -421,6 +435,45 @@ public boolean dropNamespace(String[] namespace, boolean cascade)
...
+   * Creates a catalog-specific Spark Table wrapping a Gravitino view. Subclasses that support views
+   * must override this method.
+   *
+   * @param ident the identifier
+   * @param gravitinoView the Gravitino view metadata
+   * @param sparkTable the underlying Spark table for IO (view expansion)
+   * @return a Spark Table representing the view
+   * @throws UnsupportedOperationException if views are not supported by this catalog
+   */
+  protected Table createSparkView(Identifier ident, View gravitinoView, Table sparkTable) {
```
mchades, line 472: better to use `convertToTable`?

diqiu50: I think this name matches its functionality. convertToTable is actually less clear.

mchades (approved):

PR #11294, opened 2026-05-29
[#10990] feat(iceberg-rest): Vended credential refresh for S3, GCS, OSS, and ADLS

### LINKED ISSUE #10990: [Improvement] IRC: Vended credential expiry, the new credential isn't automatically refreshed.

Single Spark job whose execution time exceeds the credential TTL (15 minutes) will encounter the token expiration issue as following.

```
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] INFO org.apache.spark.sql.execution.datasources.v2.DataWritingSparkTask - Writer for partition 0 is committing.
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] ERROR org.apache.spark.util.Utils - Aborting task
java.io.UncheckedIOException: Failed to close current writer
	at org.apache.iceberg.io.RollingFileWriter.closeCurrentWriter(RollingFileWriter.java:128)
	at org.apache.iceberg.io.RollingFileWriter.close(RollingFileWriter.java:156)
	at org.apache.iceberg.io.RollingDataWriter.close(RollingDataWriter.java:32)
	at org.apache.iceberg.spark.source.SparkWrite$UnpartitionedDataWriter.close(SparkWrite.java:762)
	at org.apache.iceberg.spark.source.SparkWrite$UnpartitionedDataWriter.commit(SparkWrite.java:744)
	at org.apache.spark.sql.execution.datasources.v2.WritingSparkTask.$anonfun$run$1(WriteToDataSourceV2Exec.scala:493)
	at org.apache.spark.util.Utils$.tryWithSafeFinallyAndFailureCallbacks(Utils.scala:1563)
	at org.apache.spark.sql.execution.datasources.v2.WritingSparkTask.run(WriteToDataSourceV2Exec.scala:509)
	at org.apache.spark.sql.execution.datasources.v2.WritingSparkTask.run$(WriteToDataSourceV2Exec.scala:448)
	at org.apache.spark.sql.execution.datasources.v2.DataWritingSparkTask$.run(WriteToDataSourceV2Exec.scala:514)
	at org.apache.spark.sql.execution.datasources.v2.V2TableWriteExec.$anonfun$writeWithV2$2(WriteToDataSourceV2Exec.scala:411)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:92)
	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:161)
	at org.apache.spark.scheduler.Task.run(Task.scala:139)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:554)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1529)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:557)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.io.IOException: software.amazon.awssdk.services.s3.model.S3Exception: The provided token has expired. (Service: S3, Status Code: 400, Request ID: 86DG6ABA4TWGNDGK, Extended Request ID: GOHaAMjUtJ43ryoqbg3/bRkhRAnFBGgw/1fK5LFZ6RaclQfbLJBZAD75jZ/4zBFPReAmTiNAu3o=) (SDK Attempt Count: 1)
	at org.apache.iceberg.shaded.org.apache.parquet.io.DelegatingPositionOutputStream.close(DelegatingPositionOutputStream.java:41)
	at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileWriter.close(ParquetFileWriter.java:1830)
	at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileWriter.end(ParquetFileWriter.java:1816)
	at org.apache.iceberg.parquet.ParquetWriter.close(ParquetWriter.java:261)
	at org.apache.iceberg.io.DataWriter.close(DataWriter.java:82)
	at org.apache.iceberg.io.RollingFileWriter.closeCurrentWriter(RollingFileWriter.java:126)
	... 19 more
Caused by: software.amazon.awssdk.services.s3.model.S3Exception: The provided token has expired. (Service: S3, Status Code: 400, Request ID: 86DG6ABA4TWGNDGK, Extended Request ID: GOHaAMjUtJ43ryoqbg3/bRkhRAnFBGgw/1fK5LFZ6RaclQfbLJBZAD75jZ/4zBFPReAmTiNAu3o=) (SDK Attempt Count: 1)
	at software.amazon.awssdk.services.s3.model.S3Exception$BuilderImpl.build(S3Exception.java:113)
	at software.amazon.awssdk.services.s3.model.S3Exception$BuilderImpl.build(S3Exception.java:61)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.utils.RetryableStageHelper.retryPolicyDisallowedRetryException(RetryableStageHelper.java:168)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:36)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:53)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:35)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:82)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:62)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:43)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:50)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:32)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:37)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:26)
	at software.amazon.awssdk.core.internal.http.AmazonSyncHttpClient$RequestExecutionBuilderImpl.execute(AmazonSyncHttpClient.java:210)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.invoke(BaseSyncClientHandler.java:103)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.doExecute(BaseSyncClientHandler.java:173)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.lambda$execute$1(BaseSyncClientHandler.java:80)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.measureApiCallSuccess(BaseSyncClientHandler.java:182)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:74)
	at software.amazon.awssdk.core.client.handler.SdkSyncClientHandler.execute(SdkSyncClientHandler.java:45)
	at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:53)
	at software.amazon.awssdk.services.s3.DefaultS3Client.putObject(DefaultS3Client.java:11883)
	at org.apache.iceberg.aws.s3.S3OutputStream.completeUploads(S3OutputStream.java:443)
	at org.apache.iceberg.aws.s3.S3OutputStream.close(S3OutputStream.java:269)
	at org.apache.iceberg.aws.s3.S3OutputStream.close(S3OutputStream.java:255)
	at org.apache.iceberg.shaded.org.apache.parquet.io.DelegatingPositionOutputStream.close(DelegatingPositionOutputStream.java:40)
	... 24 more
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] ERROR org.apache.spark.sql.execution.datasources.v2.DataWritingSparkTask - Aborting commit for partition 0 (task 4, attempt 0, stage 4.0)
[iceberg-s3fileio-delete-0] WARN org.apache.iceberg.aws.s3.S3FileIO - Encountered failure when deleting batch
software.amazon.awssdk.services.s3.model.S3Exception: The provided token has expired. (Service: S3, Status Code: 400, Request ID: EGCNEJ0A0MEESFRV, Extended Request ID: atx1n3nNimZpxeRkfSz6gOXSUCucA45pYWdTIfpmoyCzIn4WQlnUuEZ4HFMa1mXZCryWcpvnKoI=) (SDK Attempt Count: 1)
	at software.amazon.awssdk.services.s3.model.S3Exception$BuilderImpl.build(S3Exception.java:113)
	at software.amazon.awssdk.services.s3.model.S3Exception$BuilderImpl.build(S3Exception.java:61)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.utils.RetryableStageHelper.retryPolicyDisallowedRetryException(RetryableStageHelper.java:168)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:36)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:53)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:35)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:82)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:62)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:43)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:50)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:32)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:37)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:26)
	at software.amazon.awssdk.core.internal.http.AmazonSyncHttpClient$RequestExecutionBuilderImpl.execute(AmazonSyncHttpClient.java:210)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.invoke(BaseSyncClientHandler.java:103)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.doExecute(BaseSyncClientHandler.java:173)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.lambda$execute$1(BaseSyncClientHandler.java:80)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.measureApiCallSuccess(BaseSyncClientHandler.java:182)
	at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:74)
	at software.amazon.awssdk.core.client.handler.SdkSyncClientHandler.execute(SdkSyncClientHandler.java:45)
	at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:53)
	at software.amazon.awssdk.services.s3.DefaultS3Client.deleteObjects(DefaultS3Client.java:4020)
	at org.apache.iceberg.aws.s3.S3FileIO.deleteBatch(S3FileIO.java:340)
	at org.apache.iceberg.aws.s3.S3FileIO.lambda$deleteFiles$3(S3FileIO.java:276)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] WARN org.apache.iceberg.aws.s3.S3FileIO - Failed to delete object at path s3://irc-e2e-env1-metalake-b-warehouse/env1/metalake_b/catalog_1/dml2/spark_single_job_dst_54332f0c05a441c689913a4e97408fca/data/00000-4-bae6c1c6-9e0a-44a2-a58a-59edb5a59333-0-00001.parquet
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] WARN org.apache.iceberg.spark.source.SparkCleanupUtil - Deleted only 0 of 1 file(s) using bulk deletes (partition 0 (task 4, attempt 0, stage 4.0))
org.apache.iceberg.io.BulkDeletionFailureException: Failed to delete 1 files
	at org.apache.iceberg.aws.s3.S3FileIO.deleteFiles(S3FileIO.java:297)
	at org.apache.iceberg.spark.source.SparkCleanupUtil.bulkDelete(SparkCleanupUtil.java:101)
	at org.apache.iceberg.spark.source.SparkCleanupUtil.deletePaths(SparkCleanupUtil.java:93)
	at org.apache.iceberg.spark.source.SparkCleanupUtil.deleteFiles(SparkCleanupUtil.java:87)
	at org.apache.iceberg.spark.source.SparkCleanupUtil.deleteTaskFiles(SparkCleanupUtil.java:58)
	at org.apache.iceberg.spark.source.SparkWrite$UnpartitionedDataWriter.abort(SparkWrite.java:757)
	at org.apache.spark.sql.execution.datasources.v2.WritingSparkTask.$anonfun$run$6(WriteToDataSourceV2Exec.scala:505)
	at org.apache.spark.util.Utils$.tryWithSafeFinallyAndFailureCallbacks(Utils.scala:1574)
	at org.apache.spark.sql.execution.datasources.v2.WritingSparkTask.run(WriteToDataSourceV2Exec.scala:509)
	at org.apache.spark.sql.execution.datasources.v2.WritingSparkTask.run$(WriteToDataSourceV2Exec.scala:448)
	at org.apache.spark.sql.execution.datasources.v2.DataWritingSparkTask$.run(WriteToDataSourceV2Exec.scala:514)
	at org.apache.spark.sql.execution.datasources.v2.V2TableWriteExec.$anonfun$writeWithV2$2(WriteToDataSourceV2Exec.scala:411)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:92)
	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:161)
	at org.apache.spark.scheduler.Task.run(Task.scala:139)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:554)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1529)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:557)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] ERROR org.apache.spark.sql.execution.datasources.v2.DataWritingSparkTask - Aborted commit for partition 0 (task 4, attempt 0, stage 4.0)
[Executor task launch worker for task 0.0 in stage 4.0 (TID 4)] ERROR org.apache.spark.executor.Executor - Exception in task 0.0 in stage 4.0 (TID 4)
java.io.UncheckedIOException: Failed to close current writer
```

### How should we improve?

Support refresh for the S3FileIO held storage credentials.

FYI https://github.com/apache/iceberg/pull/15678/

lasdf1234:

This PR adds **Iceberg 1.11 vended credential refresh** for object storage in the Iceberg REST Catalog (IRC), and includes the **server-side Iceberg 1.11.0 upgrade** required for the feature.

**Vended credential refresh (IRC)**

- Add `IcebergRESTUtils.toRESTCredential(...)` to map Gravitino storage tokens to Iceberg 1.11 client properties, table-scoped refresh paths (`GET .../tables/{table}/credentials`), and storage `prefix` from table location.
- Extend `CredentialPropertyUtils.toIcebergProperties` with token expiry metadata for S3/IRSA, OSS, and ADLS (GCS expiry already supported).
- Wire credential refresh into IRC load-table, credentials API, and scan-plan responses in `CatalogWrapperForREST`.
- For `catalog-backend=rest`, proxy `/credentials` to the upstream IRC via `getRESTTableCredentials()` (uses `UserPrincipalForwardingAuthManager` from `RESTCatalog` properties); other backends use local `getLocalTableCredentials()` via `catalogCredentialManager`.

**Iceberg 1.11 upgrade (included in this branch)**

- Upgrade server-side Iceberg to **1.11.0** for IRC and lakehouse catalog.
- Adapt IRC REST handlers, plan-scan response building, and related Spark/Flink IT classpath expectations.

Iceberg 1.11 clients can refresh vended object-storage credentials before they expire during long-running jobs. Gravitino IRC must return expiry fields and table-scoped refresh endpoints for S3, GCS, OSS, and ADLS when temporary credentials are enabled.

When IRC proxies another IRC (`catalog-backend=rest`), `/credentials` must be fetched from the upstream catalog instead of being generated locally on the proxy.

The server must run Iceberg 1.11 to align with the REST spec and client refresh behavior.

Fix: #10990

- When `vended-credentials` is enabled, load-table / credentials / scan-plan responses may include cloud-specific expiry fields and refresh-endpoint properties (for example `client.refresh-credentials-endpoint`, `gcs.oauth2.refresh-credentials-endpoint`, `adls.refresh-credentials-endpoint`).
- Refresh paths use the relative form without a leading slash: `v1/{catalog}/namespaces/.../tables/.../credentials`.
- IRC and lakehouse Iceberg catalog documentation reflect **Iceberg 1.11.0** server support.

- [x] `./gradlew :common:test --tests org.apache.gravitino.credential.TestCredentialPropertiesUtils -PskipITs`
- [x] `./gradlew :iceberg:iceberg-rest-server:test --tests org.apache.gravitino.iceberg.service.TestIcebergRESTUtils --tests org.apache.gravitino.iceberg.service.TestCatalogWrapperForREST -PskipITs`
- [x] `./gradlew :iceberg:iceberg-rest-server:compileJava :iceberg:iceberg-rest-server:compileTestJava -PskipITs`


iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestIcebergRESTUtils.java
```diff
@@ -87,4 +98,119 @@ void testSerdeIcebergRESTObject() {
...
+    when(tableMetadata.location()).thenReturn("s3://bucket/t/");
+    Map<String, String> config =
+        IcebergRESTUtils.toRestCredential(
+                "aws", table, new S3TokenCredential("key", "secret", "token", 1234L), tableMetadata)
+            .config();
+
+    Assertions.assertEquals(
+        "v1/aws/namespaces/ns/tables/tbl/credentials",
+        config.get("client.refresh-credentials-endpoint"));
+  }
```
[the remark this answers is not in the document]
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -83,6 +101,63 @@ public String getValue() {
...
+    if (refreshProperty != null) {
+      config.put(
+          refreshProperty,
+          String.format(
+              "v1/%s/namespaces/%s/tables/%s/credentials",
+              RESTUtil.encodeString(catalogName),
+              RESTUtil.encodeNamespace(
+                  tableIdentifier.namespace(), NAMESPACE_SEPARATOR_URLENCODED_UTF_8),
+              RESTUtil.encodeString(tableIdentifier.name())));
+    }
```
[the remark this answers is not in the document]
lasdf1234: Resolved.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -49,7 +55,13 @@ public class CredentialPropertyUtils {
   @VisibleForTesting
   static final String ICEBERG_ADLS_ACCOUNT_KEY = "adls.auth.shared-key.account.key";
 
-  private static final String GCS_OAUTH_2_TOKEN_EXPIRES_AT = "gcs.oauth2.token-expires-at";
+  @VisibleForTesting
+  static final String ICEBERG_ADLS_SAS_TOKEN_EXPIRES_AT_MS_PREFIX = "adls.sas-token-expires-at-ms.";
```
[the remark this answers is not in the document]
lasdf1234: Based on your proposed approach, the code cannot be compiled successfully (the spotlessApply fails), so your suggestion cannot be adopted.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -83,6 +102,63 @@ public String getValue() {
...
+
+    String prefix = Strings.CS.appendIfMissing(tableMetadata.location(), "/");
```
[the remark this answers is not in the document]
lasdf1234: StringUtils has already been deprecated.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -82,24 +94,25 @@ public class CredentialPropertyUtils {
...
       Map<String, String> icebergGCSCredentialProperties =
           transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap);
       icebergGCSCredentialProperties.put(
-          GCS_OAUTH_2_TOKEN_EXPIRES_AT, String.valueOf(credential.expireTimeInMs()));
+          ICEBERG_GCS_TOKEN_EXPIRES_AT, String.valueOf(credential.expireTimeInMs()));
       return icebergGCSCredentialProperties;
-    }
-
-    if (credential instanceof ADLSTokenCredential) {
+    } else if (credential instanceof ADLSTokenCredential) {
```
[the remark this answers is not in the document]
lasdf1234: Got resovled.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -82,24 +94,28 @@ public class CredentialPropertyUtils {
...
-        || credential instanceof AwsIrsaCredential) {
-      return transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap);
-    }
-
-    if (credential instanceof GCSTokenCredential) {
+    if (credential instanceof S3TokenCredential || credential instanceof AwsIrsaCredential) {
+      Map<String, String> icebergProperties =
+          new HashMap<>(
+              transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap));
+      icebergProperties.put(
```
roryqi, line 126: Why not put this into the method `transformProperties`?

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -82,24 +94,28 @@ public class CredentialPropertyUtils {
...
+          ICEBERG_OSS_SECURITY_TOKEN_EXPIRES_AT_MS, String.valueOf(credential.expireTimeInMs()));
+      return icebergProperties;
+    } else if (credential instanceof GCSTokenCredential) {
       Map<String, String> icebergGCSCredentialProperties =
-          transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap);
+          new HashMap<>(
+              transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap));
       icebergGCSCredentialProperties.put(
-          GCS_OAUTH_2_TOKEN_EXPIRES_AT, String.valueOf(credential.expireTimeInMs()));
+          ICEBERG_GCS_TOKEN_EXPIRES_AT, String.valueOf(credential.expireTimeInMs()));
```
roryqi, line 135: Why do we change this?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
roryqi: Could u avoid using Strings.CS.?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -83,6 +102,63 @@ public String getValue() {
...
+   */
+  public static org.apache.iceberg.rest.credentials.Credential toRestCredential(
+      String catalogName,
+      TableIdentifier tableIdentifier,
+      Credential credential,
+      TableMetadata tableMetadata) {
+    Map<String, String> config =
+        new HashMap<>(CredentialPropertyUtils.toIcebergProperties(credential));
+
+    String refreshProperty = null;
```
roryqi: refreshEndpointProp.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -83,6 +102,63 @@ public String getValue() {
...
+   * Builds an Iceberg REST {@link org.apache.iceberg.rest.credentials.Credential} for load-table,
+   * scan-plan, or credentials API responses.
+   *
+   * @param catalogName IRC catalog name used in the refresh path
+   * @param tableIdentifier table receiving the credential
+   * @param credential Gravitino credential to vend
+   * @param tableMetadata table metadata used to derive the storage prefix
+   * @return Iceberg REST credential with prefix and config
+   */
+  public static org.apache.iceberg.rest.credentials.Credential toRestCredential(
```
roryqi: Rest -> REST.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
lasdf1234: Got resovled.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
lasdf1234: All the static constants in the CredentialPropertyUtils class start with "ICEBERG", except for this one which has an irregular naming convention.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -336,9 +333,7 @@ protected LoadTableResponse injectCredentialConfig(
         .build();
   }
 
-  private Credential getCredential(
-      LoadTableResponse loadTableResponse, CredentialPrivilege privilege) {
-    TableMetadata tableMetadata = loadTableResponse.tableMetadata();
+  private Credential getCredential(TableMetadata tableMetadata, CredentialPrivilege privilege) {
```
roryqi, line 408: If the catalog backend is a REST catalog backend, it should retrieve the credential from the catalog backend instead generating a credential in the local mode.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
roryqi: Seems not resolved.

lasdf1234: Got

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
lasdf1234: Got Resovled.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
roryqi: OK.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -198,32 +207,65 @@ public LoadTableResponse updateTable(
...
   }
 
+  private static LoadCredentialsResponse restTableCredentials(
+      RESTCatalog restCatalog, TableIdentifier identifier) {
+    try {
+      Map<String, String> properties = Maps.newHashMap(restCatalog.properties());
+      String credentialsPath =
+          ResourcePaths.forCatalogProperties(properties).table(identifier) + "/credentials";
+
+      try (AuthManager authManager = AuthManagers.loadAuthManager(restCatalog.name(), properties);
```
roryqi: Will it use `UserPrincipalForwardingAuthManager`?

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestCatalogWrapperForREST.java
```diff
@@ -101,6 +114,192 @@ void testIsLocalOrHdfsLocation() {
     Assertions.assertFalse(CatalogWrapperForREST.isLocalOrHdfsLocation("   "));
   }
 
+  @Test
+  void testRestTableCredentials() throws Exception {
```
roryqi: Rest -> REST.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -198,32 +207,65 @@ public LoadTableResponse updateTable(
...
+                  credential,
+                  loadTableResponse.tableMetadata()))
+          .build();
     } catch (ServiceUnavailableException e) {
       LOG.warn("Service unavailable when loading table credentials for table: {}", identifier, e);
       return ImmutableLoadCredentialsResponse.builder().build();
     }
   }
 
+  private static LoadCredentialsResponse restTableCredentials(
```
roryqi: rest -> REST.

roryqi:
I don't have other comments except above. You would better test remote REST backend by hand. You can use two real servers to test it.

lasdf1234:
> I don't have other comments except above. You would better test remote REST backend by hand. You can use two real servers to test it.

Got.After I test it with real scenarios, I will inform you.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
lasdf1234: Yes, it will use UserPrincipalForwardingAuthManager.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
freesinger: The local credential generation path is already cached. My concern here is the new REST-backend path: if `/credentials` is fetched remotely on demand, do we also want reuse/caching for that remote-fetch path?                                                                  
                                                                                                                                                                                                                                                                                      
Otherwise frequent loads may keep hitting the backend even while the returned credential is still valid.

roryqi: Remote server should have cache. We don't cache remote catalog data locally to avoid consistence isssue.

freesinger: 「Remote server should have cache」will solve my concern.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestCatalogWrapperForREST.java
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
lasdf1234: Got resolved.

roryqi:
Could u modify your title? You should start with `[#xxxx]` or `[MINOR]`. xxxx is the issue id.

lasdf1234:
> Could u modify your title? You should start with `[#xxxx]` or `[MINOR]`. xxxx is the issue id.

Got title has changed.

roryqi:
No other comments except for tests.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -198,32 +207,65 @@ public LoadTableResponse updateTable(
...
+    } catch (NoSuchTableException e) {
+      throw e;
+    } catch (Exception e) {
+      LOG.warn(
+          "Failed to load table credentials from REST catalog backend for table: {}",
+          identifier,
+          e);
+      return ImmutableLoadCredentialsResponse.builder().build();
+    }
+  }
```
jerryshao, line 288: Have you tested in both local IRC and federated IRC cases?

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
```diff
@@ -82,24 +94,25 @@ public class CredentialPropertyUtils {
...
-        || credential instanceof OSSSecretKeyCredential
-        || credential instanceof AzureAccountKeyCredential
-        || credential instanceof AwsIrsaCredential) {
-      return transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap);
-    }
-
-    if (credential instanceof GCSTokenCredential) {
+    if (credential instanceof S3TokenCredential || credential instanceof AwsIrsaCredential) {
+      Map<String, String> icebergProperties =
+          transformProperties(credential.credentialInfo(), icebergCredentialPropertyMap);
```
jerryshao: Why can't we extract these two lines out of if...else loop? They are duplicated for all the if...else branches.

FANNG1:
I found a few issues after checking this PR against the local Iceberg 1.11 code under `~/opensource/iceberg`:

1. `getRESTTableCredentials()` silently converts upstream REST failures into an empty credential response.

`getRESTTableCredentials()` catches all non-`NoSuchTableException` exceptions and returns an empty `LoadCredentialsResponse`. This differs from `getLocalTableCredentials()`, which only catches local credential-generation `ServiceUnavailableException`. If the upstream REST catalog returns auth/403/5xx, Gravitino hides the real error and the Iceberg client later fails inside `VendedCredentialsProvider.refreshCredential()` with `Invalid S3 Credentials: empty`, which is misleading and makes the original upstream failure hard to diagnose.

Please let upstream auth/403/5xx errors propagate to the REST response, and only downgrade truly expected/no-credential cases.

2. `AwsIrsaCredential` can emit a non-positive S3 token expiry.

The new `CredentialPropertyUtils.toIcebergProperties()` path treats `AwsIrsaCredential` like a temporary S3 credential and writes `s3.session-token-expires-at-ms`. Iceberg 1.11 parses this with `Long.parseLong` and skips the initial cached credential when the expiry is already before the prefetch time. Unlike `S3TokenCredential`, `AwsIrsaCredential` currently does not validate `expireTimeInMs > 0`, so `0` can cause the client to refresh immediately and repeatedly.

Please either validate `AwsIrsaCredential.expireTimeInMs() > 0` or reject non-positive expiry before adding the Iceberg expiry property.

3. ADLS refresh expiry key does not match the Iceberg expected host suffix.

Gravitino emits the SAS token key as `adls.sas-token.<account>.dfs.core.windows.net`, but the expiry key as `adls.sas-token-expires-at-ms.<account>`. Iceberg ADLS refresh uses `ADLSLocation.host()` for both keys, for example `account.dfs.core.windows.net`, and `VendedAdlsCredentialProvider` requires `adls.sas-token-expires-at-ms.<host>`. With a normal `abfss://container@account.dfs.core.windows.net/...` table location, refresh will fail because the expiry key is missing.

Please use the same full host suffix for both ADLS SAS token and expiry keys.

4. OSS refresh endpoint is advertised, but the local Iceberg OSS module has no refresh provider.

`IcebergRESTUtils.toRESTCredential()` adds `client.refresh-credentials-endpoint` for `OSSTokenCredential`. In the local Iceberg code, `AliyunProperties` only reads `client.access-key-id`, `client.access-key-secret`, and `client.security-token`; there is no OSS vended refresh endpoint/provider. So this currently provides initial temporary OSS credentials, but does not actually enable refresh.

Please remove the OSS refresh endpoint until Iceberg OSS supports it, or document/prove the supported client path.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
lasdf1234: Yes, all the scenarios have been tested.There are no issues with the Federation and local scenarios.

common/src/main/java/org/apache/gravitino/credential/CredentialPropertyUtils.java
lasdf1234: Thank you for your valuable suggestions. I have extracted the same logic from if... else branches.The code has been simplified.

lasdf1234:
> I found a few issues after checking this PR against the local Iceberg 1.11 code under `~/opensource/iceberg`:
> 
> 1. `getRESTTableCredentials()` silently converts upstream REST failures into an empty credential response.
> 
> `getRESTTableCredentials()` catches all non-`NoSuchTableException` exceptions and returns an empty `LoadCredentialsResponse`. This differs from `getLocalTableCredentials()`, which only catches local credential-generation `ServiceUnavailableException`. If the upstream REST catalog returns auth/403/5xx, Gravitino hides the real error and the Iceberg client later fails inside `VendedCredentialsProvider.refreshCredential()` with `Invalid S3 Credentials: empty`, which is misleading and makes the original upstream failure hard to diagnose.
> 
> Please let upstream auth/403/5xx errors propagate to the REST response, and only downgrade truly expected/no-credential cases.
> 
> 2. `AwsIrsaCredential` can emit a non-positive S3 token expiry.
> 
> The new `CredentialPropertyUtils.toIcebergProperties()` path treats `AwsIrsaCredential` like a temporary S3 credential and writes `s3.session-token-expires-at-ms`. Iceberg 1.11 parses this with `Long.parseLong` and skips the initial cached credential when the expiry is already before the prefetch time. Unlike `S3TokenCredential`, `AwsIrsaCredential` currently does not validate `expireTimeInMs > 0`, so `0` can cause the client to refresh immediately and repeatedly.
> 
> Please either validate `AwsIrsaCredential.expireTimeInMs() > 0` or reject non-positive expiry before adding the Iceberg expiry property.
> 
> 3. ADLS refresh expiry key does not match the Iceberg expected host suffix.
> 
> Gravitino emits the SAS token key as `adls.sas-token.<account>.dfs.core.windows.net`, but the expiry key as `adls.sas-token-expires-at-ms.<account>`. Iceberg ADLS refresh uses `ADLSLocation.host()` for both keys, for example `account.dfs.core.windows.net`, and `VendedAdlsCredentialProvider` requires `adls.sas-token-expires-at-ms.<host>`. With a normal `abfss://container@account.dfs.core.windows.net/...` table location, refresh will fail because the expiry key is missing.
> 
> Please use the same full host suffix for both ADLS SAS token and expiry keys.
> 
> 4. OSS refresh endpoint is advertised, but the local Iceberg OSS module has no refresh provider.
> 
> `IcebergRESTUtils.toRESTCredential()` adds `client.refresh-credentials-endpoint` for `OSSTokenCredential`. In the local Iceberg code, `AliyunProperties` only reads `client.access-key-id`, `client.access-key-secret`, and `client.security-token`; there is no OSS vended refresh endpoint/provider. So this currently provides initial temporary OSS credentials, but does not actually enable refresh.
> 
> Please remove the OSS refresh endpoint until Iceberg OSS supports it, or document/prove the supported client path.

1.Now the upstream exception will be thrown and the code has been modified.
2.The expiration logic of AwsIrsaCredential has been made consistent with that of S3 token.
3.The issue of ADLS refresh expiry key has also been resolved.
4.The oss module of iceberg does not support the "vended refresh". This issue has been noted in the comments of the relevant methods in gravitino, and the corresponding code has also been modified.

FANNG1:
I think there is still a REST proxy issue here. `filterCredentialProperties()` now keeps `client.refresh-credentials-endpoint`, `adls.refresh-credentials-endpoint`, and `gcs.oauth2.refresh-credentials-endpoint`, and the REST catalog path copies the filtered `FileIO.properties()` into the downstream `LoadTableResponse.config()` in create/stage-create/load-table.

Iceberg clients consume these keys directly from `LoadTableResponse.config()`: `RESTSessionCatalog.tableFileIO()` merges `response.config()` into the FileIO initialization properties, and the S3/GCS/ADLS vended credential providers use those refresh endpoints for later refreshes. For a proxied REST catalog, forwarding the upstream endpoint unchanged can therefore send clients to the wrong place: a relative upstream path will be resolved against the Gravitino catalog URI, while an absolute upstream URL lets clients bypass Gravitino and call the upstream `/credentials` endpoint directly, which can break auth/topology assumptions.

I think Gravitino should either strip upstream refresh endpoint properties from the proxied load-table config, or rewrite them to the current Gravitino endpoint (`v1/<catalog>/namespaces/.../tables/.../credentials`) instead of forwarding the upstream value unchanged.

lasdf1234:
> I think there is still a REST proxy issue here. `filterCredentialProperties()` now keeps `client.refresh-credentials-endpoint`, `adls.refresh-credentials-endpoint`, and `gcs.oauth2.refresh-credentials-endpoint`, and the REST catalog path copies the filtered `FileIO.properties()` into the downstream `LoadTableResponse.config()` in create/stage-create/load-table.
> 
> Iceberg clients consume these keys directly from `LoadTableResponse.config()`: `RESTSessionCatalog.tableFileIO()` merges `response.config()` into the FileIO initialization properties, and the S3/GCS/ADLS vended credential providers use those refresh endpoints for later refreshes. For a proxied REST catalog, forwarding the upstream endpoint unchanged can therefore send clients to the wrong place: a relative upstream path will be resolved against the Gravitino catalog URI, while an absolute upstream URL lets clients bypass Gravitino and call the upstream `/credentials` endpoint directly, which can break auth/topology assumptions.
> 
> I think Gravitino should either strip upstream refresh endpoint properties from the proxied load-table config, or rewrite them to the current Gravitino endpoint (`v1/<catalog>/namespaces/.../tables/.../credentials`) instead of forwarding the upstream value unchanged.

Got it. The code has been modified as per your instructions. Now it only transmits the "upstream refresh endpoint properties".

api/src/main/java/org/apache/gravitino/credential/AwsIrsaCredential.java
```diff
@@ -118,12 +118,15 @@ public String sessionToken() {
...
+  private void validate(
+      String accessKeyId, String secretAccessKey, String sessionToken, long expireTimeInMs) {
     Preconditions.checkArgument(
         StringUtils.isNotBlank(accessKeyId), "Access key Id should not be empty");
     Preconditions.checkArgument(
         StringUtils.isNotBlank(secretAccessKey), "Secret access key should not be empty");
     Preconditions.checkArgument(
         StringUtils.isNotBlank(sessionToken), "Session token should not be empty");
+    Preconditions.checkArgument(
+        expireTimeInMs > 0, "The expire time of AwsIrsaCredential should be greater than 0");
```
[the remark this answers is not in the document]
lasdf1234: It should be consistent with the other copywriting.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -198,32 +208,60 @@ public LoadTableResponse updateTable(
...
+              credentialsPath,
+              LoadCredentialsResponse.class,
+              Collections.emptyMap(),
+              ErrorHandlers.tableErrorHandler());
+    } catch (IOException e) {
+      throw new RESTException(
+          e,
+          "Failed to close REST client resources when loading credentials for table: %s",
+          identifier);
+    }
```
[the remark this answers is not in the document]
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -198,32 +208,60 @@ public LoadTableResponse updateTable(
    */
   public LoadCredentialsResponse getTableCredentials(
       TableIdentifier identifier, CredentialPrivilege privilege) {
+    if (isRESTCatalog()) {
+      return getRESTTableCredentials((RESTCatalog) getCatalog(), identifier);
+    } else {
+      return getLocalTableCredentials(identifier, privilege);
+    }
+  }
```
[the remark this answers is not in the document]
lasdf1234: I have modified java doc

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -40,7 +41,15 @@
 import javax.ws.rs.core.Response.Status;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.credential.ADLSTokenCredential;
+import org.apache.gravitino.credential.AwsIrsaCredential;
+import org.apache.gravitino.credential.Credential;
+import org.apache.gravitino.credential.CredentialPropertyUtils;
+import org.apache.gravitino.credential.GCSTokenCredential;
+import org.apache.gravitino.credential.OSSTokenCredential;
+import org.apache.gravitino.credential.S3TokenCredential;
```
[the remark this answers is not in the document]
lasdf1234: Got resovled.

api/src/main/java/org/apache/gravitino/credential/AwsIrsaCredential.java
```diff
@@ -118,12 +118,15 @@ public String sessionToken() {
...
+  private void validate(
+      String accessKeyId, String secretAccessKey, String sessionToken, long expireTimeInMs) {
     Preconditions.checkArgument(
         StringUtils.isNotBlank(accessKeyId), "Access key Id should not be empty");
     Preconditions.checkArgument(
         StringUtils.isNotBlank(secretAccessKey), "Secret access key should not be empty");
     Preconditions.checkArgument(
         StringUtils.isNotBlank(sessionToken), "Session token should not be empty");
+    Preconditions.checkArgument(
+        expireTimeInMs > 0, "The expire time of AwsIrsaCredential should great than 0");
```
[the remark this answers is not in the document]
lasdf1234: It should be consistent with the other copywriting.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -692,6 +759,13 @@ private LoadTableResponse createTableInternal(Namespace namespace, CreateTableRe
                   properties, key -> catalogPropertiesToClientKeys.contains(key)))
           // Keep only credential fields from FileIO properties before returning them to the client.
           .addAllConfig(CredentialPropertyUtils.filterCredentialProperties(properties))
+          .addAllConfig(
+              CredentialPropertyUtils.buildRefreshCredentialEndpoints(
+                  RESTUtil.encodeString(catalogCredentialManager.catalogName()),
+                  RESTUtil.encodeNamespace(
+                      ident.namespace(), IcebergRESTUtils.NAMESPACE_SEPARATOR_URLENCODED_UTF_8),
+                  RESTUtil.encodeString(ident.name()),
+                  CredentialPropertyUtils.filterCredentialProperties(properties)))
```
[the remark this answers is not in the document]
lasdf1234: Got resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -192,38 +201,92 @@ public LoadTableResponse updateTable(
...
+
+      if (authManager != null) {
+        try {
+          authManager.close();
+        } catch (Exception e) {
+          LOG.warn(
+              "Failed to close auth manager when loading credentials for table: {}", identifier, e);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
lasdf1234: Backtracking will cause problems.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestCatalogWrapperForREST.java
```diff
@@ -101,6 +118,292 @@ void testIsLocalOrHdfsLocation() {
...
+        "/",
+        exchange -> {
+          requestPath.set(exchange.getRequestURI().getPath());
+          byte[] body = upstreamJson.getBytes(StandardCharsets.UTF_8);
+          exchange.getResponseHeaders().add("Content-Type", "application/json");
+          exchange.sendResponseHeaders(200, body.length);
+          try (OutputStream os = exchange.getResponseBody()) {
+            os.write(body);
+          }
+        });
```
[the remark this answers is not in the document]
lasdf1234: Tests run on JDK 17 in CI with the full JDK; jdk.httpserver is available.
iceberg-rest-server is not JDK 8–targeted.

FANNG1:
Hi, thanks for the detailed PR! I noticed a potential gap in the REST catalog proxy path.

The current fix in `loadTableInternal` / `createTableInternal` / `stageTableCreateInternal` extracts credential config via `retrieveFileIOProperties(table.io())` → `fileIO.properties()`, then calls `buildRefreshCredentialEndpoints` based on those properties.

However, in Iceberg 1.11, when the upstream REST catalog returns credentials via `LoadTableResponse.credentials()` (the standard structured field), `RESTSessionCatalog` injects them through `SupportsStorageCredentials#setCredentials()` — **not** into `fileIO.properties()`. These two are separate fields in `S3FileIO` / `ResolvingFileIO` / `GCSFileIO`: `setCredentials()` writes to `storageCredentials`, while `properties()` only returns what was passed to `initialize()`.

So if the upstream follows the Iceberg 1.11 spec and uses `storage-credentials: [...]` in its load-table response, `retrieveFileIOProperties(table.io())` will return empty credential keys, `filterCredentialProperties` will filter them out, and `buildRefreshCredentialEndpoints` will return nothing. The downstream client's load-table response ends up with neither initial credentials nor a refresh endpoint.

The existing test `testLoadTableInjectsRefreshEndpoint` only covers the compat path (mocking `fileIO.properties()` with `s3.session-token`), not the standard `credentials` field path.

Could you share your thoughts on this? One possible fix direction would be to also read `((SupportsStorageCredentials) table.io()).credentials()` and either forward those credentials (with endpoint rewriting) to the downstream `LoadTableResponse.credentials()` field, or at least use them to drive `buildRefreshCredentialEndpoints`. Would be happy to discuss the best approach.

lasdf1234:
> Hi, thanks for the detailed PR! I noticed a potential gap in the REST catalog proxy path.
> 
> The current fix in `loadTableInternal` / `createTableInternal` / `stageTableCreateInternal` extracts credential config via `retrieveFileIOProperties(table.io())` → `fileIO.properties()`, then calls `buildRefreshCredentialEndpoints` based on those properties.
> 
> However, in Iceberg 1.11, when the upstream REST catalog returns credentials via `LoadTableResponse.credentials()` (the standard structured field), `RESTSessionCatalog` injects them through `SupportsStorageCredentials#setCredentials()` — **not** into `fileIO.properties()`. These two are separate fields in `S3FileIO` / `ResolvingFileIO` / `GCSFileIO`: `setCredentials()` writes to `storageCredentials`, while `properties()` only returns what was passed to `initialize()`.
> 
> So if the upstream follows the Iceberg 1.11 spec and uses `storage-credentials: [...]` in its load-table response, `retrieveFileIOProperties(table.io())` will return empty credential keys, `filterCredentialProperties` will filter them out, and `buildRefreshCredentialEndpoints` will return nothing. The downstream client's load-table response ends up with neither initial credentials nor a refresh endpoint.
> 
> The existing test `testLoadTableInjectsRefreshEndpoint` only covers the compat path (mocking `fileIO.properties()` with `s3.session-token`), not the standard `credentials` field path.
> 
> Could you share your thoughts on this? One possible fix direction would be to also read `((SupportsStorageCredentials) table.io()).credentials()` and either forward those credentials (with endpoint rewriting) to the downstream `LoadTableResponse.credentials()` field, or at least use them to drive `buildRefreshCredentialEndpoints`. Would be happy to discuss the best approach.

@FANNG1 Thanks your valuable suggestions and agreed on the gap.

Kept the existing fileIO.properties() path and added an Iceberg 1.11 path that reads credentials from SupportsStorageCredentials#credentials()

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/TestCatalogWrapperForREST.java
```diff
@@ -27,8 +27,13 @@
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
 
 import com.google.common.collect.ImmutableMap;
+import com.sun.net.httpserver.HttpServer;
+import java.io.OutputStream;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
```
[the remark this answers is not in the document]
lasdf1234: I kept HttpServer for consistency with existing IRC tests in this module;

api/src/main/java/org/apache/gravitino/credential/AwsIrsaCredential.java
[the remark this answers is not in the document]
lasdf1234: It should be consistent with the other copywriting. This has been mentioned many times before and should not be repeated.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergRESTUtils.java
```diff
@@ -83,6 +93,103 @@ public String getValue() {
...
+    }
+
+    List<org.apache.iceberg.rest.credentials.Credential> restCredentials = new ArrayList<>();
+    for (StorageCredential credential : credentials) {
+      Map<String, String> filteredConfig =
+          CredentialPropertyUtils.filterCredentialProperties(credential.config());
+      filteredConfig.putAll(buildRefreshProps(catalogName, tableIdentifier, filteredConfig));
+      restCredentials.add(
+          toRESTCredential(credential.prefix(), ImmutableMap.copyOf(filteredConfig)));
+    }
```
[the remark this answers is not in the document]
lasdf1234: Got fixed it.

FANNG1:
Thanks, the new `SupportsStorageCredentials#credentials()` path looks like the right direction and the added tests cover the structured `storage-credentials` case.

One remaining edge case: `buildStorageCreds()` currently normalizes every upstream `StorageCredential.prefix()` through `storageLocationPrefix()`, which appends `/` when the prefix does not already end with one. That is fine for table/location-style prefixes such as `s3://bucket/path`, but it changes Iceberg root scheme prefixes like `s3` into `s3/`.

Iceberg S3 vended credentials use `prefix("s3")` in its own tests, and `S3FileIO.clientForStoragePath()` matches credentials by checking `storagePath.startsWith(storagePrefix)`. After rewriting `s3` to `s3/`, a normal storage path like `s3://bucket/...` no longer matches that credential, so the client can fall back to the default S3 client instead of the vended credential.

Could we avoid normalizing upstream `StorageCredential.prefix()` values when forwarding them? The trailing-slash normalization is useful for locally generated table-location prefixes, but forwarded upstream credential prefixes should probably be preserved as-is, or at least root scheme prefixes like `s3` / `gs` should be left unchanged.

Minor nit while touching this area: `AwsIrsaCredential` now says `should great than 0`; should be `should be greater than 0`.

lasdf1234:
> should be greater than 0



> Thanks, the new `SupportsStorageCredentials#credentials()` path looks like the right direction and the added tests cover the structured `storage-credentials` case.
> 
> One remaining edge case: `buildStorageCreds()` currently normalizes every upstream `StorageCredential.prefix()` through `storageLocationPrefix()`, which appends `/` when the prefix does not already end with one. That is fine for table/location-style prefixes such as `s3://bucket/path`, but it changes Iceberg root scheme prefixes like `s3` into `s3/`.
> 
> Iceberg S3 vended credentials use `prefix("s3")` in its own tests, and `S3FileIO.clientForStoragePath()` matches credentials by checking `storagePath.startsWith(storagePrefix)`. After rewriting `s3` to `s3/`, a normal storage path like `s3://bucket/...` no longer matches that credential, so the client can fall back to the default S3 client instead of the vended credential.
> 
> Could we avoid normalizing upstream `StorageCredential.prefix()` values when forwarding them? The trailing-slash normalization is useful for locally generated table-location prefixes, but forwarded upstream credential prefixes should probably be preserved as-is, or at least root scheme prefixes like `s3` / `gs` should be left unchanged.
> 
> Minor nit while touching this area: `AwsIrsaCredential` now says `should great than 0`; should be `should be greater than 0`.

Regarding the first issue, I carefully reviewed the source code of iceberg and conducted local tests. I found that the "/ " at the end of "location" was not necessary, so I removed this logic and thus this problem no longer exists.
Regarding the second question, I have revised the prompt information, and the prompts for other token classes also align with your requirements.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/AwsIrsaCredentialGenerator.java
```diff
@@ -81,10 +81,12 @@ public AwsIrsaCredential generate(CredentialContext context) {
       AwsCredentials creds = baseCredentialsProvider.resolveCredentials();
       if (creds instanceof AwsSessionCredentials) {
         AwsSessionCredentials sessionCreds = (AwsSessionCredentials) creds;
-        long expiration =
-            sessionCreds.expirationTime().isPresent()
-                ? sessionCreds.expirationTime().get().toEpochMilli()
-                : 0L;
+        if (!sessionCreds.expirationTime().isPresent()) {
```
roryqi, line 84: isEmpty.

lasdf1234: Thanks for the suggestion. Optional.isEmpty() would read better, but we can’t use it in this module.

:bundles:aws is built as Java 8–compatible bytecode (release 8 via compatibleWithJDK8() in the root build.gradle.kts; :bundles is in jdk8CompatibleProjectPathPrefixes). Optional.isEmpty() was added in Java 11, so using it here would break compilation under our current Java 8 target.

roryqi: OK. Got it.

FANNG1:
LGTM

FANNG1 (approved):

roryqi:
No other comments.

roryqi (approved):

PR #11266, opened 2026-05-27
[#11265] feat(iceberg-rest): add async cleanup persistence layer

roryqi:
### What changes were proposed in this pull request?

First of three stacked PRs for async Iceberg hard deletion - the **persistence foundation only**:

- New IcebergConfig keys for the async-cleanup worker/retry knobs.
- iceberg_cleanup_job table schema + 1.2.0->1.3.0 migrations for H2/MySQL/PostgreSQL.
- IcebergCleanupJob value object with a nested State enum.
- IcebergCleanupJobStore - add jobs, take runnable jobs with heartbeat CAS, mark progress, and delete old finished jobs - with its MyBatis mapper/PO/SQL provider.
- core test-jar (testArtifacts) wiring so the store tests can reuse the relational entity-store backend.

Nothing is wired into the REST drop flow yet; that comes in the follow-up PRs (manager engine, then REST integration).

### Why are the changes needed?

Inline cleanup of a large Iceberg table blocks the drop request and risks client timeouts, and a crash mid-deletion leaves no record to resume. A durable, persisted job store is the foundation for an opt-in async cleanup. Splitting it out keeps each PR reviewable.

Fix: #11265

### Does this PR introduce _any_ user-facing change?

Adds new Iceberg REST config property keys (opt-in, unused until later PRs) and a new iceberg_cleanup_job table in the 1.3.0 schema/upgrade scripts. No API or default-behavior change.

### How was this patch tested?

New unit tests TestIcebergCleanupJob and TestIcebergCleanupJobStore (the latter runs the relational backend matrix via AbstractIcebergCleanupJobStoreBackendTest). Verified :iceberg:iceberg-common:testClasses and :iceberg:iceberg-rest-server:testClasses compile.

PR #11299, opened 2026-05-29
[#11298] feat(iceberg-rest): add async cleanup manager

### LINKED ISSUE #11298: [Subtask] Async cleanup manager for the Iceberg REST catalog
### Describe the subtask

Implement `IcebergCleanupManager`, the server-wide worker engine for async hard deletion in the Iceberg REST catalog. It pulls `PENDING` jobs from the `iceberg_cleanup_job` store (added in #11266), drives the cleanup state machine (claim → heartbeat → succeed / fail-after-max-attempts), deletes a dropped table's reachable files in bulk through a shared delete pool, renews claim heartbeats on a thread decoupled from deletion, and prunes finished rows past the retention window.

This is PR 2 of 3 for the feature: (1) persistence layer (#11266, merged), (2) cleanup manager engine (this subtask), (3) REST integration.

### Parent issue

#11265

roryqi:

- Add `IcebergCleanupManager`: a server-wide worker pool that polls the `iceberg_cleanup_job` store for `PENDING` jobs, claims them via a heartbeat lease, deletes the dropped table's reachable files in bulk through a shared `deleteExecutor` (with `CallerRunsPolicy` back-pressure), renews heartbeats on a separate scheduler thread, drives the retry state machine (transient failure → back to `PENDING`, give up at `max-attempts` → `FAILED`), and prunes finished rows past the retention window.
- Add `TestIcebergCleanupManager` unit tests, running against the H2/MySQL/PostgreSQL backend matrix via `TestJDBCBackend`.
- Sync the async hard-deletion design doc with the merged persistence layer (the `dropTable` enqueue snippet and the testing section).

Second of three stacked PRs for async hard deletion in the Iceberg REST catalog. The persistence layer (#11266) added the durable job store; this PR adds the worker engine that drains jobs and deletes files. REST integration follows in PR 3.

Fix: #11298

No. The manager is not yet wired into the REST drop path (that lands in PR 3), and the config keys (`async-cleanup.*`) were introduced with the persistence layer (#11266).

`./gradlew :iceberg:iceberg-rest-server:test --tests "org.apache.gravitino.iceberg.service.cleanup.TestIcebergCleanupManager" -PskipITs` — 7 tests pass (lifecycle to `SUCCEEDED`, transient-retry to `FAILED`, bulk-delete batching, already-deleted-file tolerance, reachable-file cleanup, and tombstone delegation).

### DIFF
--- a/design-docs/async-iceberg-rest-hard-deletion.md
+++ b/design-docs/async-iceberg-rest-hard-deletion.md
@@ -183,10 +183,10 @@ public void dropTable(IcebergRequestContext ctx, TableIdentifier id,
 
   TableMetadata metadata = w.loadTableMetadata(id);
   w.dropTable(id);                          // metadata-only drop in the catalog
-  cleanupManager.enqueue(
+  cleanupManager.addJob(
       new IcebergCleanupJob(
-          0L,                               // id assigned from IdGenerator at enqueue (§5.4)
-          catalogId,                        // resolved catalog entity id (§5.4)
+          0L,                               // id assigned by IdGenerator at enqueue
+          catalogId,                        // resolved catalog entity id
           id.namespace().toString(),
           id.name(),
           metadata.metadataFileLocation(),
@@ -337,8 +337,11 @@ statement can register its own provider without touching the rest of the design.
 1. **Poll & claim.** Read the candidate window above and CAS the first
    winnable row to `RUNNING` with a fresh `heartbeat_at`. If nothing is
    claimable, wait `poll-interval-secs` and repeat.
-2. **Load metadata.** `TableMetadataParser.read(io, metadata_location)`. A
-   transient failure releases the job for retry; a terminal one fails it
+2. **Load metadata.** `TableMetadataParser.read(io, metadata_location)`. The
+   root `metadata.json` is the only file whose absence means "the table is
+   already gone": a `NotFoundException` here completes the job (a prior attempt
+   already deleted it). Any other failure (e.g. corrupt/unreadable metadata)
+   releases the job for retry, failing it once `attempts` hits the ceiling
    (§5.6).
 3. **Stream & delete.** Walk the snapshot graph lazily, group reachable files
    into batches of `delete-batch-size`, and hand each batch to the shared
@@ -440,18 +443,19 @@ rows age out for reclaim.
 
 Per-file failures are logged but do not fail the whole job — the
 synchronous purge has the same "best effort" stance. A job fails only if the
-**metadata phase** fails. `NotFoundException` from `deleteFile` counts as
-success. A transient failure goes back to `PENDING` and is retried when a
-free thread next polls (§5.5), incrementing `attempts` each time; when `attempts`
-reaches `max-attempts` the job goes to `FAILED` instead of `PENDING`. A
-clearly terminal failure skips the retries and goes straight to `FAILED`.
+**metadata phase** fails. A `NotFoundException` counts as success rather than a
+failure: a missing root `metadata.json` means the table is already gone, and a
+missing data file, manifest, or manifest list (from `deleteFile` or while
+enumerating reachable files) is already deleted. Any other failure goes back to
+`PENDING` and is retried when a free thread next polls (§5.5), incrementing
+`attempts` each time; when `attempts` reaches `max-attempts` the job goes to
+`FAILED` instead of `PENDING`.
 
 | Outcome                                              | Action                                                                                |
 |------------------------------------------------------|---------------------------------------------------------------------------------------|
 | All files deleted (or already gone)                  | `state='SUCCEEDED'`                                                                    |
-| Transient failure, `attempts < max-attempts`         | `attempts++`, set `last_error`, back to `PENDING`, `heartbeat_at=0`; re-claimed later |
-| Transient failure, `attempts` reaches `max-attempts` | `attempts++`, set `last_error` → `FAILED` (gave up retrying)                            |
-| Terminal failure (e.g. metadata gone/corrupt)        | set `last_error` → `FAILED` immediately; retrying cannot help                          |
+| Failure, `attempts < max-attempts`                   | `attempts++`, set `last_error`, back to `PENDING`, `heartbeat_at=0`; re-claimed later  |
+| Failure, `attempts` reaches `max-attempts`           | `attempts++`, set `last_error` → `FAILED` (gave up retrying)                            |
 | Worker killed mid-job                                | `RUNNING` row's heartbeat goes stale; another worker reclaims; deletes are idempotent  |
 
 **After `SUCCEEDED`.** The files are gone, so the tombstone lifts at once:
--- a/design-docs/async-iceberg-rest-hard-deletion.md
+++ b/design-docs/async-iceberg-rest-hard-deletion.md
@@ -19,12 +19,12 @@
 
 # Design: Asynchronous Hard Deletion for the Gravitino Iceberg REST Server
 
-| Field    | Value                                                   |
-| -------- | ------------------------------------------------------- |
-| Status   | Complete                                                |
-| Authors  | @roryqi                                                 |
-| Created  | 2026-05-19                                              |
-| Module   | `iceberg/iceberg-rest-server`, `iceberg/iceberg-common` |
+| Field   | Value                                                   |
+| ------- | ------------------------------------------------------- |
+| Status  | Complete                                                |
+| Authors | @roryqi                                                 |
+| Created | 2026-05-19                                              |
+| Module  | `iceberg/iceberg-rest-server`, `iceberg/iceberg-common` |
 
 ---
 
@@ -98,14 +98,14 @@ simpler design with a smaller bug surface. We chose a single async
 implementation gated behind a per-request opt-in, with synchronous deletion
 as the default.
 
-| Approach                                                    | Pros                                                                                     | Cons                                                                                             | Decision                                                                |
-|-------------------------------------------------------------|------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------|
-| Synchronous only (status quo)                               | Simplest; strongest "deleted means gone" guarantee                                       | Exceeds HTTP timeouts, saturates Jetty, no retry/audit on large tables                           | **Rejected** — the problem we are solving                               |
-| Pluggable `IcebergPurger` SPI                               | Extensible without code changes                                                          | Real added surface (SPI, discovery factory, context) with **no** second implementation in flight | **Rejected** — revisit when a second implementation has a real customer |
-| Reuse `RelationalGarbageCollector`                          | Proven worker/scheduling pattern                                                         | Different IO surface (object store vs. JDBC) and failure model                                   | **Rejected** — share patterns, not code                                 |
-| External job system (Quartz / Temporal)                     | Mature scheduling, retries, observability                                                | Heavy operational burden on every operator                                                       | **Rejected** — disproportionate for one workload                        |
-| Enumerate files at enqueue time                             | Worker needs no metadata re-read                                                         | Slow on large tables (defeats the latency goal), bloats job rows                                 | **Rejected** — store `metadata_location`, re-read at run time           |
-| Object-store job markers (no DB table)                      | No schema / migration                                                                    | Hand-rolled lease + renewal, no indexed scheduling, fragments per-bucket                         | **Rejected** — higher net complexity (see below)                        |
+| Approach                                                             | Pros                                                                                     | Cons                                                                                             | Decision                                                                |
+| -------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------- |
+| Synchronous only (status quo)                                        | Simplest; strongest "deleted means gone" guarantee                                       | Exceeds HTTP timeouts, saturates Jetty, no retry/audit on large tables                           | **Rejected** — the problem we are solving                               |
+| Pluggable `IcebergPurger` SPI                                        | Extensible without code changes                                                          | Real added surface (SPI, discovery factory, context) with **no** second implementation in flight | **Rejected** — revisit when a second implementation has a real customer |
+| Reuse `RelationalGarbageCollector`                                   | Proven worker/scheduling pattern                                                         | Different IO surface (object store vs. JDBC) and failure model                                   | **Rejected** — share patterns, not code                                 |
+| External job system (Quartz / Temporal)                              | Mature scheduling, retries, observability                                                | Heavy operational burden on every operator                                                       | **Rejected** — disproportionate for one workload                        |
+| Enumerate files at enqueue time                                      | Worker needs no metadata re-read                                                         | Slow on large tables (defeats the latency goal), bloats job rows                                 | **Rejected** — store `metadata_location`, re-read at run time           |
+| Object-store job markers (no DB table)                               | No schema / migration                                                                    | Hand-rolled lease + renewal, no indexed scheduling, fragments per-bucket                         | **Rejected** — higher net complexity (see below)                        |
 | **JDBC job table + worker pool, async opt-in (synchronous default)** | Smallest bug surface; restart-safe and cluster-safe via CAS claim; one code path to test | No built-in extension point                                                                      | **Chosen**                                                              |
 
 **Why not an S3-only control plane?** Conditional writes (`If-None-Match`)
@@ -218,15 +218,15 @@ already terminal, so it is consistent with the §5.7 tombstone table.
 
 ```sql
 CREATE TABLE IF NOT EXISTS `iceberg_cleanup_job` (
-  `id`                BIGINT(20)    UNSIGNED NOT NULL,
+  `id`                BIGINT(20)    UNSIGNED NOT NULL COMMENT 'globally unique cleanup job id',
   `catalog_id`        BIGINT(20)    UNSIGNED NOT NULL COMMENT 'globally unique id of the owning catalog, stable across catalog rename',
-  `namespace`         VARCHAR(512)  NOT NULL,
-  `table_name`        VARCHAR(256)  NOT NULL,
-  `metadata_location` MEDIUMTEXT   NOT NULL,
-  `file_io_impl`      VARCHAR(256)  NOT NULL,
-  `file_io_props`     MEDIUMTEXT    NOT NULL COMMENT 'JSON',
+  `namespace`         VARCHAR(512)  NOT NULL COMMENT 'namespace of the table to be cleaned up',
+  `table_name`        VARCHAR(256)  NOT NULL COMMENT 'name of the table to be cleaned up',
+  `metadata_location` MEDIUMTEXT   NOT NULL COMMENT 'location of the table metadata file to purge',
+  `file_io_impl`      VARCHAR(256)  NOT NULL COMMENT 'FileIO implementation class used to access the table files',
+  `file_io_props`     MEDIUMTEXT    NOT NULL COMMENT 'JSON-encoded FileIO properties',
   `state`             VARCHAR(16)   NOT NULL COMMENT 'PENDING|RUNNING|SUCCEEDED|FAILED',
-  `attempts`          INT(10)       NOT NULL DEFAULT 0,
+  `attempts`          INT(10)       NOT NULL DEFAULT 0 COMMENT 'number of processing attempts made so far',
   `last_error`        VARCHAR(2048) NULL COMMENT 'truncated reason for the most recent failure, NULL until a job fails',
   `heartbeat_at`      BIGINT(20)    NOT NULL DEFAULT 0 COMMENT 'last heartbeat from the worker, 0 when not running',
   `created_by`        VARCHAR(128)  NOT NULL COMMENT 'principal that requested the drop (audit)',
@@ -343,29 +343,59 @@ statement can register its own provider without touching the rest of the design.
    already deleted it). Any other failure (e.g. corrupt/unreadable metadata)
    releases the job for retry, failing it once `attempts` hits the ceiling
    (§5.6).
-3. **Stream & delete.** Walk the snapshot graph lazily, group reachable files
-   into batches of `delete-batch-size`, and hand each batch to the shared
-   `deleteExecutor` for a **bulk** delete; `NotFoundException` counts as
-   deleted. The worker thread itself never issues the storage call directly,
-   and a *separate* background task keeps `heartbeat_at` fresh throughout
-   (see "Heartbeat is decoupled from deletion" below).
+3. **Delete leaves before parents.** Walk the snapshot graph and delete one
+   dependency level at a time — data files → manifests → manifest lists →
+   statistics → ancestor metadata → the root `metadata.json` **last** — only
+   advancing once the current level is fully gone. Within a level, files are
+   independent: group them into batches of `delete-batch-size` and hand each to
+   the shared `deleteExecutor` for a **bulk** delete; `NotFoundException` counts
+   as deleted. Leaf-first ordering keeps every crash recoverable: the root
+   pointer (and the manifests/manifest lists above any leaf still on disk)
+   survives, so a retry re-enumerates and finishes. Deleting a parent first
+   would strand its children — and a `metadata.json` removed early would be
+   misread as "table already gone", leaking everything beneath it. The worker
+   thread itself never issues the storage call directly, and a *separate*
+   background task keeps `heartbeat_at` fresh throughout (see "Heartbeat is
+   decoupled from deletion" below).
 4. **Finish.** Once every reachable file is gone the job is `SUCCEEDED`
    (§5.6). The thread then loops back to step 1.
 
-Execution mirrors `CatalogHandlers.purgeTable`, but **streams** the
-reachable files instead of materializing them. A large table can reference
-millions of data files, so the worker walks the snapshot graph lazily and
-deletes them **in bulk batches**, never one file at a time:
+Execution mirrors `CatalogHandlers.purgeTable`, deleting **in bulk batches**
+and never one file at a time. To keep every crash recoverable it deletes one
+dependency level at a time — leaves first, the root `metadata.json` last — and
+each `deleteAll` blocks until its level is fully gone before the next begins.
+Data files are the only unbounded level, so they are streamed and deleted one
+manifest at a time rather than collected up front — a million-file table never
+materializes its whole path set; only the far smaller manifest, manifest-list
+and metadata path lists are held at once:
 
 ```java
 TableMetadata meta = TableMetadataParser.read(io, job.metadataLocation());
-try (CloseableIterable<String> files = reachableFiles(meta)) {  // lazy, not materialized
-  Iterators.partition(files.iterator(), deleteBatchSize)        // bounded batches
+Set<String> manifests = new LinkedHashSet<>();
+for (Snapshot s : meta.snapshots()) {
+  for (ManifestFile m : s.allManifests(io)) {
+    if (!manifests.add(m.path())) continue;                  // dedup shared manifests
+    try (CloseableIterable<String> dataFiles =               // lazy, one batch resident at a time
+        ManifestFiles.readPaths(m, io, meta.specsById())) {
+      deleteAll(io, dataFiles);                              // leaves: streamed, then deleted
+    }
+  }
+}
+deleteAll(io, manifests);                                    // then their manifests
+deleteAll(io, ReachableFileUtil.manifestListLocations(table));
+deleteAll(io, ReachableFileUtil.statisticsFilesLocations(table));
+deleteAll(io, ancestorMetadata);                             // older metadata.json
+deleteAll(io, List.of(job.metadataLocation()));              // root pointer, deleted last
+
+// deleteAll: bulk-batch the (possibly lazy) iterable, then await it
+void deleteAll(FileIO io, Iterable<String> files) {
+  Iterators.partition(files.iterator(), deleteBatchSize)     // bounded batches
       .forEachRemaining(batch ->
-          deleteExecutor.execute(() ->
+          futures.add(deleteExecutor.submit(() ->
               // SupportsBulkOperations.deleteFiles when the FileIO supports it
               // (e.g. S3FileIO's batch-delete API), else concurrent per-file
-              CatalogUtil.deleteFiles(io, batch, "cleanup", /* bulk */ true)));
+              CatalogUtil.deleteFiles(io, batch, "cleanup", /* bulk */ true))));
+  awaitAll(futures);                                         // NotFoundException counts as deleted
 }
 ```
 
@@ -451,12 +481,12 @@ enumerating reachable files) is already deleted. Any other failure goes back to
 `attempts` each time; when `attempts` reaches `max-attempts` the job goes to
 `FAILED` instead of `PENDING`.
 
-| Outcome                                              | Action                                                                                |
-|------------------------------------------------------|---------------------------------------------------------------------------------------|
-| All files deleted (or already gone)                  | `state='SUCCEEDED'`                                                                    |
-| Failure, `attempts < max-attempts`                   | `attempts++`, set `last_error`, back to `PENDING`, `heartbeat_at=0`; re-claimed later  |
-| Failure, `attempts` reaches `max-attempts`           | `attempts++`, set `last_error` → `FAILED` (gave up retrying)                            |
-| Worker killed mid-job                                | `RUNNING` row's heartbeat goes stale; another worker reclaims; deletes are idempotent  |
+| Outcome                                    | Action                                                                                |
+| ------------------------------------------ | ------------------------------------------------------------------------------------- |
+| All files deleted (or already gone)        | `state='SUCCEEDED'`                                                                   |
+| Failure, `attempts < max-attempts`         | `attempts++`, set `last_error`, back to `PENDING`, `heartbeat_at=0`; re-claimed later |
+| Failure, `attempts` reaches `max-attempts` | `attempts++`, set `last_error` → `FAILED` (gave up retrying)                          |
+| Worker killed mid-job                      | `RUNNING` row's heartbeat goes stale; another worker reclaims; deletes are idempotent |
 
 **After `SUCCEEDED`.** The files are gone, so the tombstone lifts at once:
 `createTable` / `register` at the identifier succeed again (§5.7). The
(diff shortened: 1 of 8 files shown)


design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -183,21 +183,22 @@ public void dropTable(IcebergRequestContext ctx, TableIdentifier id,
...
-          .catalogName(ctx.catalogName())
-          .tableIdentifier(id)
-          .metadataLocation(metadata.metadataFileLocation())
-          .fileIoImpl(w.fileIoImpl())
-          .fileIoProperties(w.fileIoProperties())
-          .createdBy(ctx.userPrincipal())
-          .build());
+  cleanupManager.enqueue(
+      new IcebergCleanupJob(
+          0L,                               // id assigned from IdGenerator at enqueue (§5.4)
```
yuqi1129: The flag `(§5.4)` is not so proper, and we can't get what it refers to

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,319 @@
...
+
+  /** Starts worker threads and the heartbeat/prune scheduler. */
+  public void start() {
+    if (running) {
+      return;
+    }
+
+    running = true;
+    workers = Executors.newFixedThreadPool(workerThreads, daemon("iceberg-cleanup-worker"));
+    scheduler = Executors.newScheduledThreadPool(1, daemon("iceberg-cleanup-scheduler"));
```
yuqi1129: You need to declare the queue in the thread pool, or the size is unlimited.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,319 @@
...
+    for (int i = 0; i < workerThreads; i++) {
+      workers.submit(this::workerLoop);
+    }
+
+    long heartbeatIntervalMs = Math.max(1L, heartbeatTimeoutMs / 3L);
```
yuqi1129: `heartbeatIntervalMs` can be 1?

roryqi: Just a guard design. It can't be 1 actuallly.

design-docs/async-iceberg-rest-hard-deletion.md
```diff
@@ -217,33 +218,57 @@ already terminal, so it is consistent with the §5.7 tombstone table.
...
+  `heartbeat_at`      BIGINT(20)    NOT NULL DEFAULT 0 COMMENT 'last heartbeat from the worker, 0 when not running',
   `created_by`        VARCHAR(128)  NOT NULL COMMENT 'principal that requested the drop (audit)',
-  `updated_at`        BIGINT(20)    NOT NULL COMMENT 'last state change; drives poll ordering and terminal-row pruning',
+  `updated_at`        BIGINT(20)    NOT NULL COMMENT 'last state change, drives poll ordering and old finished-job cleanup',
   PRIMARY KEY (`id`),
   KEY `idx_state_updated` (`state`, `updated_at`),
-  KEY `idx_object` (`catalog_name`, `namespace`, `table_name`, `state`)
-) ENGINE=InnoDB;
+  KEY `idx_object` (`catalog_id`, `namespace`(255), `table_name`(128), `state`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin COMMENT 'async Iceberg table cleanup jobs';
```
jerryshao: Did you change here aligning with the SQL script?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,356 @@
...
+    // LinkedBlockingQueue.
+    workers =
+        new ThreadPoolExecutor(
+            workerThreads,
+            workerThreads,
+            0L,
+            TimeUnit.MILLISECONDS,
+            new ArrayBlockingQueue<>(workerThreads),
+            daemon("iceberg-cleanup-worker"));
+    scheduler = Executors.newScheduledThreadPool(1, daemon("iceberg-cleanup-scheduler"));
```
jerryshao: The scheduler not only do the prune thing, but also do the hearbeat thing, do you need to make the daemon name more meaningful?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,356 @@
...
+    for (int i = 0; i < workerThreads; i++) {
+      workers.submit(this::workerLoop);
+    }
```
jerryshao, line 162: This code snippet is better put together with workers definition.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Removed the `(§5.4)` flags; the inline comments now describe the fields directly (`id assigned by IdGenerator at enqueue`, `resolved catalog entity id`).

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Done. Replaced the implicit unbounded queue with an explicit `ThreadPoolExecutor` backed by a bounded `ArrayBlockingQueue`. We submit exactly `workerThreads` loops, so it's bounded only to avoid `newFixedThreadPool`'s unbounded `LinkedBlockingQueue`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,356 @@
...
+              "Manifest {} already deleted during async cleanup; skipping",
+              manifest.path(),
+              alreadyDeleted);
+        } catch (Exception e) {
+          throw new RuntimeException("Failed to read manifest " + manifest.path(), e);
+        }
+      }
+    }
+    return files;
+  }
```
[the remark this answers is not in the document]
roryqi: Fixed. Data files are now streamed one manifest at a time (`deleteDataFiles`): each manifest's `ManifestFiles.readPaths` iterable is handed straight to `deleteAll`, which partitions it lazily, so only one manifest's paths are resident at once. Only the much smaller manifest/manifest-list/metadata path lists are materialized.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,356 @@
...
+          ownedHeartbeats.remove(entry.getKey());
+        }
+      } catch (Throwable t) {
+        // scheduleAtFixedRate suppresses all future runs if a task ever throws, which would stop
+        // heartbeat renewal for the whole process. Swallow per-job faults (including Errors) so one
+        // bad job neither halts this pass nor the recurring task.
+        LOG.warn("Heartbeat update failed for job {}", entry.getKey(), t);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
roryqi: Addressed. The worker now checks ownership before each manifest and between deletion levels (`requireOwnership`) and stops early once the lease is lost. I also added a heartbeat-token CAS to `markSucceeded`/`recordFailure` (`WHERE ... AND heartbeat_at = ?`) so a reclaimed worker can't change a job a peer now owns.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,356 @@
...
+    scheduler = Executors.newScheduledThreadPool(1, daemon("iceberg-cleanup-scheduler"));
+    for (int i = 0; i < workerThreads; i++) {
+      workers.submit(this::workerLoop);
+    }
+
+    long heartbeatIntervalMs = Math.max(1L, heartbeatTimeoutMs / 3L);
+    scheduler.scheduleAtFixedRate(
+        this::refreshHeartbeats, heartbeatIntervalMs, heartbeatIntervalMs, TimeUnit.MILLISECONDS);
+    scheduler.scheduleAtFixedRate(this::prune, 1L, 1L, TimeUnit.HOURS);
+  }
```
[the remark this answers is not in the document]
roryqi: Done. `running` is now an `AtomicBoolean` and `start()` guards with `compareAndSet(false, true)`.

design-docs/async-iceberg-rest-hard-deletion.md
roryqi: Aligned the doc's CREATE TABLE with scripts/mysql/schema-1.3.0-mysql.sql -- added the per-column COMMENTs so they match.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Renamed to `iceberg-cleanup-heartbeat-prune`, with a comment noting the single scheduler thread runs both heartbeat renewal and pruning.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Moved the worker-submission loop directly under the `workers = new ThreadPoolExecutor(...)` definition, before the scheduler setup.

roryqi:
@yuqi1129 I have addressed all the comments. Could u help me take a look again?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,409 @@
...
+
+    // Delete children before parents, root metadata.json last. Each deleteAll blocks until its
+    // level is gone, so a crash always leaves the root (and the manifests above any surviving file)
+    // readable for a retry to rebuild from. Deleting a parent first would orphan its children.
+    //
+    // Data files are the only huge level, so they are streamed and deleted one manifest at a time
+    // rather than all collected first; only the smaller manifest/list/metadata paths are held.
+    //
+    // requireOwnership stops between levels if a peer reclaimed the job, so we never delete a
+    // parent
```
yuqi1129: Remove the line break between `parent` and `while`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,409 @@
...
+    return running.get() && ownedHeartbeats.containsKey(id);
+  }
+
+  private void requireOwnership(BooleanSupplier stillOwned) {
+    if (!stillOwned.getAsBoolean()) {
+      throw new OwnershipLostException();
+    }
+  }
+
+  private static void sleep(long ms) {
```
yuqi1129, line 374: Can we use awaitability to replace sleep?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Done — reflowed the comment so `parent while` stays on one line.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,409 @@
...
+    return false;
+  }
+
+  private void workerLoop() {
+    while (running.get()) {
+      try {
+        long now = System.currentTimeMillis();
+        Optional<IcebergCleanupJob> job =
+            store.takePendingJob(now, heartbeatTimeoutMs, candidateWindow);
+        if (!job.isPresent()) {
```
yuqi1129: isEmpty()

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,409 @@
...
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
+    }
+    deleteExecutor.shutdownNow();
+    // The job store is backed by the entity store's shared relational backend, which owns the
+    // connection pool lifecycle, so there is nothing to close here.
+  }
+
+  void cleanupFiles(FileIO io, String metadataLocation) {
```
yuqi1129, line 194: VisiableForTesting if it's only used by tests

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,409 @@
...
+    // while the new owner is still deleting its children.
+    Set<String> manifests = new LinkedHashSet<>();
+    deleteDataFiles(io, metadata, manifests, stillOwned);
+    requireOwnership(stillOwned);
+    deleteAll(io, manifests);
+    requireOwnership(stillOwned);
+    deleteAll(io, ReachableFileUtil.manifestListLocations(table));
+    requireOwnership(stillOwned);
+    deleteAll(io, ReachableFileUtil.statisticsFilesLocations(table));
+    requireOwnership(stillOwned);
```
yuqi1129: I do get what you are intending to convey here. Why do we check the owner in L209.

yuqi1129: also, can you try to polish the code here.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: awaitability is a test class. So we shouldn't use it.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Done — switched to `job.isEmpty()`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Done — the 2-arg `cleanupFiles` is only called by tests (production goes through the 3-arg overload with the ownership check), so annotated it `@VisibleForTesting`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: The data-file phase (`deleteDataFiles`) checks ownership before each manifest, but not after the last one — so the check before the manifests level catches a lease lost during that final manifest's deletion, preventing us from deleting a parent the new owner is still working under. I've made this uniform: every level now goes through `deleteOwned(...)`, which re-checks ownership before deleting.

roryqi: Polished — folded the repeated `requireOwnership(...)` calls into a `deleteOwned()` helper, so each level reads as a single 'delete this level if still owned' call.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Fixed in b73c5df0c. Removed the `Math.max(1L, ...)` guard; `async-cleanup.heartbeat-timeout-secs` is validated as a positive integer in seconds, so `heartbeatTimeoutMs / 3` is at least 333ms and cannot become 1ms.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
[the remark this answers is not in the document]
roryqi: Fixed in b73c5df0c. Added an explicit closed-state guard: concurrent `start()` is still protected by `compareAndSet`, and `start()` after `close()` now fails fast instead of running with a shut-down delete executor. Added coverage for that lifecycle case.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Keeping this as a production wait helper. Awaitility is test-scoped and is already used in `TestIcebergCleanupManager` to wait for async state transitions; using it in main code would add a test polling library to the production path.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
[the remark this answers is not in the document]
roryqi: Fixed in b73c5df0c. `deleteAll` now accepts the ownership supplier, checks it before submitting each batch and before waiting on each future, cancels queued futures when ownership is lost, and the manifest data-file path now uses that ownership-aware overload.

yuqi1129 (commented):
I have no more comments.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    // compareAndSet keeps concurrent or repeated start() calls from each allocating a pool.
+    if (!running.compareAndSet(false, true)) {
+      return;
+    }
+
+    // We submit exactly workerThreads loops, so the queue is never used; it is bounded only to
+    // avoid Executors.newFixedThreadPool's unbounded queue.
+    workers =
```
jerryshao: `this.workers`

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+            0L,
+            TimeUnit.MILLISECONDS,
+            new ArrayBlockingQueue<>(workerThreads),
+            daemon("iceberg-cleanup-worker"));
+    for (int i = 0; i < workerThreads; i++) {
+      workers.submit(this::workerLoop);
+    }
+
+    // One scheduler thread runs both periodic tasks: heartbeat renewal and row pruning.
+    scheduler = Executors.newScheduledThreadPool(1, daemon("iceberg-cleanup-heartbeat-prune"));
```
jerryshao: `this.scheduler`.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+    // The job store is backed by the entity store's shared relational backend, which owns the
+    // connection pool lifecycle, so there is nothing to close here.
+  }
+
+  @VisibleForTesting
+  void cleanupFiles(FileIO io, String metadataLocation) {
+    cleanupFiles(io, metadataLocation, () -> true);
+  }
+
+  void cleanupFiles(FileIO io, String metadataLocation, BooleanSupplier stillOwned) {
```
jerryshao: `stillOwned` is always true, any purpose to add this parameter?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+    Set<String> ancestorMetadata =
+        new LinkedHashSet<>(ReachableFileUtil.metadataFileLocations(table, true));
+    ancestorMetadata.remove(metadataLocation);
+    deleteAll(io, ancestorMetadata, stillOwned);
+    deleteAll(io, Collections.singletonList(metadataLocation), stillOwned);
+  }
+
+  void deleteAll(FileIO io, Iterable<String> files) {
+    deleteAll(io, files, () -> true);
+  }
```
jerryshao: Also here, what is the purpose of this method?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+  private void runJob(IcebergCleanupJob job) {
+    try {
+      FileIO io = CatalogUtil.loadFileIO(job.fileIOImpl(), job.fileIOProperties(), null);
+      cleanupFiles(io, job.metadataLocation(), () -> ownsJob(job.id()));
+      // markSucceeded/recordFailure CAS on our heartbeat token, so a worker that lost its lease
+      // cannot change a job a peer reclaimed. A null token means we already lost it; skip the call.
+      // If a heartbeat refresh bumps the token between this read and the CAS, the CAS just no-ops
+      // and the job is reclaimed and re-run later (which finds the files gone) -- harmless.
+      Long heartbeat = ownedHeartbeats.get(job.id());
+      if (heartbeat != null) {
```
jerryshao: **Race between `refreshHeartbeats` and `markSucceeded`/`recordFailure`**

`refreshHeartbeats` updates the DB heartbeat first (`store.heartbeat`), then writes the new token back into `ownedHeartbeats`. If the worker reads `ownedHeartbeats.get(job.id())` in this window it gets the stale token, and the subsequent CAS in `markSucceeded`/`recordFailure` fails. The job stays RUNNING until heartbeat timeout and is re-claimed and re-run harmlessly (a missing metadata.json is treated as done), but it is unnecessary work.

One fix: in `refreshHeartbeats`, update `ownedHeartbeats` *before* the DB call so the worker always sees a token that is ≥ the DB value.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+        store.markSucceeded(job.id(), heartbeat);
+      }
```
jerryshao: Return value of `markSucceeded` (and `recordFailure` a few lines below) is silently discarded. A failed CAS leaves the job stuck in RUNNING with no log entry. At minimum log a warning so the re-claim is observable.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+    if (workers != null) {
+      workers.shutdownNow();
+      try {
+        workers.awaitTermination(5, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
+    }
+    deleteExecutor.shutdownNow();
+    // The job store is backed by the entity store's shared relational backend, which owns the
```
jerryshao, line 190: `deleteExecutor.shutdownNow()` is fire-and-forget — in-flight delete batches are interrupted without any wait. Consider adding `deleteExecutor.awaitTermination(5, TimeUnit.SECONDS)` here, matching the pattern used for `workers` above.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,423 @@
...
+    this.workerThreads = config.get(IcebergConfig.ASYNC_CLEANUP_WORKER_THREADS);
+    int deleteThreads = config.get(IcebergConfig.ASYNC_CLEANUP_DELETE_THREADS);
+    this.deleteBatchSize = config.get(IcebergConfig.ASYNC_CLEANUP_DELETE_BATCH_SIZE);
+    this.pollIntervalMs = config.get(IcebergConfig.ASYNC_CLEANUP_POLL_INTERVAL_SECS) * 1000L;
+    this.heartbeatTimeoutMs =
+        config.get(IcebergConfig.ASYNC_CLEANUP_HEARTBEAT_TIMEOUT_SECS) * 1000L;
+    this.maxAttempts = config.get(IcebergConfig.ASYNC_CLEANUP_MAX_ATTEMPTS);
+    this.retentionMs = config.get(IcebergConfig.ASYNC_CLEANUP_RETENTION_HOURS) * 3_600_000L;
+    this.candidateWindow = Math.max(8, workerThreads * 4);
+    this.deleteExecutor =
```
jerryshao, line 106: The magic numbers `8` and `4` in `Math.max(8, workerThreads * 4)` are unexplained. A short comment on why this multiplier provides enough headroom (e.g. to absorb a full heartbeat interval of new candidates) would help future readers.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/cleanup/TestIcebergCleanupManager.java
```diff
@@ -0,0 +1,398 @@
...
+          void cleanupFiles(FileIO io, String metadataLocation, BooleanSupplier stillOwned) {
+            throw new RuntimeException("transient");
+          }
+        };
+    long id = store.addJob(sampleJob());
+    svc.start();
+    try {
+      Awaitility.await()
+          .atMost(10, TimeUnit.SECONDS)
+          .until(() -> store.stateOf(id) == IcebergCleanupJob.State.FAILED);
```
jerryshao, line 260: 3 retries at a 1-second poll interval needs at least ~3 s; 10 s is tight on a loaded CI machine. `atMost(15, TimeUnit.SECONDS)` would give more headroom without meaningfully slowing the suite.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/cleanup/AbstractIcebergCleanupJobStoreBackendTest.java
```diff
@@ -100,36 +100,46 @@ void testAddTakeSucceedLifecycle() {
...
-  void testMarkFailed() {
-    long id = store.addJob(sampleJob());
-    store.takePendingJob(System.currentTimeMillis(), 300_000L, 10);
-    Assertions.assertTrue(store.markFailed(id, "corrupt metadata"));
-    Assertions.assertEquals(IcebergCleanupJob.State.FAILED, store.stateOf(id));
-  }
-
   @TestTemplate
   void testTransientFailureRetriesThenFailsAtCeiling() {
     long id = store.addJob(sampleJob());
```
jerryshao, line 114: The old `testMarkFailed` (direct FAILED transition) was removed. `testTransientFailureRetriesThenFailsAtCeiling` covers the retry path end-to-end, but there is no longer a store-layer test verifying that `recordFailure` at exactly `maxAttempts` produces `FAILED` (the new `testTerminalUpdateNeedsOwnership` tests ownership but not the state machine boundary). Consider keeping a targeted store-level assertion for that edge case.

jerryshao (commented):
Overall this is well-structured and the core design is sound — the heartbeat CAS in SQL, the children-before-parents deletion ordering, and streaming manifests one at a time are all correct. A couple of correctness issues worth addressing, plus a few nits.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Added the logic.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Resolved.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: Added the comments.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/cleanup/TestIcebergCleanupManager.java
roryqi: Adjusted the value.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/cleanup/AbstractIcebergCleanupJobStoreBackendTest.java
roryqi: Added the case.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: I simplified the code. AI comments have some optimization suggestion.  It add more complexity. So I removed them.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: I simplified the code. AI comments have some optimization suggestion. It add more complexity. So I removed them.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: fixed.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: fixed.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
roryqi: This won't influence the correctness. I only add a simple solution. 

`refreshHeartbeats` now publishes the new heartbeat token to `ownedHeartbeats` before calling `store.heartbeat(...)`, so a worker finishing during the refresh window reads the
  new token instead of the stale one.

  If the heartbeat CAS fails, the entry is removed; if the store call throws, the in-memory token is rolled back. Added a regression test covering this refresh/finish interleaving.

jerryshao:
I have no further comments, let Copilot to have a review again.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/cleanup/IcebergCleanupManager.java
```diff
@@ -0,0 +1,388 @@
...
+      FileIO io = CatalogUtil.loadFileIO(job.fileIOImpl(), job.fileIOProperties(), null);
+      cleanupFiles(io, job.metadataLocation());
+      finishJob(id, heartbeat -> store.markSucceeded(id, heartbeat));
+    } catch (RuntimeException e) {
+      LOG.warn("Cleanup job {} failed transiently; will retry", id, e);
+      finishJob(id, heartbeat -> store.recordFailure(id, e.getMessage(), maxAttempts, heartbeat));
+    } finally {
+      ownedHeartbeats.remove(id);
+    }
+  }
```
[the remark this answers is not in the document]
roryqi: Good catch — fixed in 8bc3adbae. Wrapped the per-job `FileIO` in try-with-resources so its backing resources are released on every path (success, transient failure, and the early return in `cleanupFiles`). Iceberg's `FileIO.close()` overrides `Closeable.close()` with no checked exception, so it slots in without touching the existing catch/finally.

yuqi1129 (approved):

Issue #10905, opened 2026-04-29
[Improvement] Make the gravitino-trino-connector compatible with Starburst ? (fast crash method & integration questions)


Jesshuan:
### What would you like to be improved?

Hey, the gravitino community,

I'm wonderning if it could be a good idea to make the gravitino-trino-connector compatible with Starburst, version by version...
For business reasons,
I've just forked & commited a simple "fast" and "crash method" solution that works with Starburst, after few tries.
(Gravitino seems to be the only way for me to federate an old hive legacy catalog)

I know Starburst is not an open-source product, and perhaps the good way could be that each starburst user have to fork and maintain this compatibility (and i'm agree with that...) but among the differents 4 commits, you will see that a good part of the changes are some little changes that are fairly "agnostics" and perhaps benefits for future trino versions too.

So I'd like to hear your thoughts on these topics and see if it's worth coordinating these little changes that could be enought for this Starburst compatibility

My forked gravitino repo project 
(i've work on the 1.2.0-hotfix branch for now, to ensure that i was working with the 1.2.0 version with the current fix...):
[https://github.com/Jesshuan/gravitino/tree/1.2.0-hotfix]
(https://github.com/Jesshuan/gravitino/tree/1.2.0-hotfix)

The details with the 'merge diff view', for a best visibility:
[https://github.com/apache/gravitino/compare/1.2.0-hotfix...Jesshuan:gravitino:1.2.0-hotfix?expand=1]
(https://github.com/apache/gravitino/compare/1.2.0-hotfix...Jesshuan:gravitino:1.2.0-hotfix?expand=1)

### How should we improve?

Here the "crash method" that i've used (and my comments), fix by fix, to achieve a final good result for now:

- first, i made a little patch for the little **checkTrinoSpiVersion** method within the **GravitinoConnectorFactory** class, just for the string version compatibility. Yes, the Starburst versions are not integer but string.
=> could it be beneficial for the gravitino trino connector project too ?

- secondly, i have replaced the **io.trino:trino-spi:$trinoVersion** dependency with the 'private' starburst jar dependency equivalent, to prevent, i think, a lot of issues.. Of course, here, i suppose that each straburst user will have to take care of this themselves. We won't push a starburst jar within a gravitino trino project !... The solution for the user (or the starburst team ?) might be o maintain a hybrid version of the gravitino project with the starburst project.

Despite of this, perhaps these other changes could be benefit after this replacement...

[...]

PR #11302, opened 2026-05-30
[#10905] feat(trino-connector): make trino-connector compatible with Starburst/Trino, with 3 little fix and a new documentation page

Jesshuan:

This PR makes the Gravitino Trino connector build cleanly across both open-source Trino and Starburst SPI variants, and documents how a user can build it against a Starburst SPI, for his private usage (because Starburst is a commercial and private product).
This PR is the result of a little discussion about this idea at [#10905].

- GravitinoConnectorFactory: a little agnostic change into the declaration inside the checkTrinoSpiVersion method. The numeric version is still parsed with spiVersion.split("[^0-9]")[0], so non-numeric Starburst version strings (e.g. 476-e.0.26) are handled.

- **GravitinoDynamicFilter**: add the getPreferredDynamicFilterTimeout() method with an optionalLong empty as return. That method is absent from the baseline open-source trino-spi DynamicFilter interface but present in newer Trino/Starburst SPIs, where it is still dispatched at runtime by signature. A comment documents why @Override is intentionally omitted, to prevent a failure at build time for trino SPI users.

- **JsonCodec**: the previous implementation assumed a no-arg public constructor for BlockEncodingManager, which broke compatibility with Starburst distributions (which require a FeaturesConfig argument) and newer Trino branches (which expose a Set<BlockEncoding> variant for Guice multibindings).
This change introduces a dedicated instantiateBlockEncodingManager helper that probes known constructor signatures in order — no-arg, then FeaturesConfig, then Set<BlockEncoding> — before falling back to a generic reflective scan that attempts to fill unknown parameters with safe defaults (null for objects, JVM defaults for primitives).
This makes the Trino connector resilient to constructor signature differences across OSS Trino and Starburst versions without requiring version-specific code paths.

**Docs** : add docs/trino-connector/starburst-compatibility.md (with a link from docs/trino-connector/index.md) describing, for indication only, how to build the connector against a Starburst SPI jar, including the licensing rationale and caveats.

These changes keep the shared source cross-version compatible, to allow to the Starburst users to benefit and (eventually) maintain an alternative trino-connector adapted and built for a specific Starburst/Trino SPI version, without any code changes, but just with little changes during the trino-connector build time of the gravitino project.
The new doc page explains the Starburst build path for the TrinoStarburst users which the project cannot ship itself because the Starburst trino-spi jar is a proprietary, non-redistributable artifact.

Fix: [#10905]

No API or property-key changes. The only user-facing addition is a new documentation page, [Starburst compatibility](vscode-webview://07991g9jc13v4i7rtg7544n4okvc96gfo7q714pqcv6rf2b24682/docs/trino-connector/starburst-compatibility.md), linked from the Trino connector docs index.

Built locally with JDK 17: JAVA_HOME=/opt/jdk-17 ./gradlew :trino-connector:trino-connector:compileJava succeeds with no errors.
./gradlew :trino-connector:trino-connector:spotlessApply is clean.
./gradlew :docs:build (OpenAPI validation) passes.
and manually verified the documented Starburst build path for one Starburst-based Trino version in the 473–478 range.

### DIFF
--- a/docs/trino-connector/index.md
+++ b/docs/trino-connector/index.md
@@ -22,3 +22,4 @@ Apache Gravitino Trino connector index:
   - [Supported SQL](sql-support.md)
   - [UDF support](udf-support.md)
   - [Development](development.md)
+  - [Starburst compatibility](starburst-compatibility.md)
--- a/docs/trino-connector/starburst-compatibility.md
+++ b/docs/trino-connector/starburst-compatibility.md
@@ -1,0 +1,139 @@
+---
+title: "Apache Gravitino Trino connector with Starburst"
+slug: /trino-connector/starburst-compatibility
+keyword: gravitino connector trino starburst
+license: "This software is licensed under the Apache License version 2."
+---
+
+This page describes, **for indication only**, how you can build the Apache Gravitino
+Trino connector against the **Starburst** SPI — a commercial, proprietary product —
+instead of the open-source Trino SPI.
+
+:::caution Read this first — this is your responsibility, not Gravitino's
+Apache Gravitino is an open-source project licensed under the Apache License 2.0, and
+it builds **exclusively against the open-source `io.trino:trino-spi` artifacts** published
+on Maven Central.
+
+**Starburst Enterprise / Starburst Galaxy are commercial, proprietary products.** Their
+`trino-spi` jar is **not** an Apache artifact, is **not** redistributable by Gravitino, and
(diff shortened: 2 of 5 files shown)


Jesshuan:
@jerryshao 
Hey ! Indeed, this is my pull request against the main branch... Perhaps did you want to check that ?
Do not hesitate to give me some advices about this PR, if you have the time and you still think that's these agnostic changes could be a good deal for starbust users

PR #11308, opened 2026-06-01
[#10866] feat(charts): Support Gateway API HTTPRoute in Helm charts

### LINKED ISSUE #10866: [Improvement] Add Kubernetes Gateway API (HTTPRoute) support to Gravitino Helm charts

The Gravitino, Iceberg REST, and Lance REST Helm charts only support the legacy `networking.k8s.io/v1 Ingress` resource. The Kubernetes Gateway API (`gateway.networking.k8s.io`) is the successor to Ingress and is now the standard in modern Kubernetes environments. 

### How should we improve?

Add an opt-in `ingress.type` field (defaulting to `ingress` for backward compatibility) that, when set to `gateway`, renders a `HTTPRoute` resource instead of an `Ingress`

Applies to:
dev/charts/gravitino
dev/charts/gravitino-iceberg-rest-server
dev/charts/gravitino-lance-rest-server

somaz94:

Add an opt-in `ingress.type` field to the Gravitino, Iceberg REST, and Lance REST
Helm charts. It defaults to `ingress` (renders the existing `networking.k8s.io/v1`
Ingress, fully backward compatible). When set to `gateway`, the chart renders a
Kubernetes Gateway API `HTTPRoute` (`gateway.networking.k8s.io/v1`) instead of an
Ingress.

- New template `templates/httproute.yaml` in all three charts.
- Existing `templates/ingress.yaml` is now gated to skip when `type: gateway`, so
  the two resources are mutually exclusive.
- New `ingress.gateway` block (`apiVersion`, `parentRefs`) in `values.yaml`. The
  existing `hosts` / `paths` / `annotations` are reused for the HTTPRoute.
- helm-unittest coverage added for the Iceberg chart (`httproute_test.yaml`, plus a
  case asserting the Ingress is not rendered in gateway mode).

Note: `tls` and `className` are Ingress-only and are intentionally not consumed by
HTTPRoute — Gateway API terminates TLS at the parent Gateway listener.

The charts currently only support the legacy Ingress resource. The Gateway API is
the successor to Ingress and the standard in modern Kubernetes environments. This
lets users on a Gateway API stack expose Gravitino services without bolting an
Ingress controller alongside their Gateway.

Fix: #10866

Yes — a new opt-in chart value `ingress.type` (default `ingress`) and an
`ingress.gateway` block. Existing installations are unaffected because the default
preserves the current Ingress behavior.

- `helm unittest --with-subchart=false dev/charts/gravitino-iceberg-rest-server`
  → 8 suites, 27 tests pass (includes the new httproute suite).
- `helm template` on all three charts: `type: gateway` renders an HTTPRoute and no
  Ingress; default/`type: ingress` renders an Ingress and no HTTPRoute.
- `helm lint` passes for the iceberg and lance charts.

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8)

### DIFF
--- a/dev/charts/gravitino-iceberg-rest-server/templates/httproute.yaml
+++ b/dev/charts/gravitino-iceberg-rest-server/templates/httproute.yaml
@@ -21,13 +21,20 @@
 {{- $fullName := include "gravitino-iceberg-rest-server.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
 {{- $gateway := .Values.ingress.gateway | default dict -}}
+{{- $root := . -}}
+{{- /* Render one HTTPRoute per host so that, like the Ingress, each host only
+   matches its own paths (HTTPRoute hostnames apply to the whole resource). */}}
+{{- range $hostIndex, $hostEntry := .Values.ingress.hosts }}
+{{- if $hostIndex }}
+---
+{{- end }}
 apiVersion: {{ $gateway.apiVersion | default "gateway.networking.k8s.io/v1" }}
 kind: HTTPRoute
 metadata:
-  name: {{ $fullName }}
+  name: {{ $fullName }}{{ if $hostIndex }}-{{ $hostIndex }}{{ end }}
   labels:
-    {{- include "gravitino-iceberg-rest-server.labels" . | nindent 4 }}
-  {{- with .Values.ingress.annotations }}
+    {{- include "gravitino-iceberg-rest-server.labels" $root | nindent 4 }}
+  {{- with $root.Values.ingress.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
@@ -36,17 +43,12 @@ spec:
   parentRefs:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- if .Values.ingress.hosts }}
+  {{- if $hostEntry.host }}
   hostnames:
-    {{- range .Values.ingress.hosts }}
-    {{- if .host }}
-    - {{ .host | quote }}
-    {{- end }}
-    {{- end }}
+    - {{ $hostEntry.host | quote }}
   {{- end }}
   rules:
-    {{- range .Values.ingress.hosts }}
-    {{- range .paths }}
+    {{- range $hostEntry.paths }}
     {{- $pathType := .pathType | default "Prefix" }}
     {{- if or (eq $pathType "Prefix") (eq $pathType "PathPrefix") (eq $pathType "ImplementationSpecific") }}
     {{- $pathType = "PathPrefix" }}
@@ -61,5 +63,5 @@ spec:
         - name: {{ $fullName }}
           port: {{ $svcPort }}
     {{- end }}
-    {{- end }}
+{{- end }}
 {{- end }}
--- a/dev/charts/gravitino-iceberg-rest-server/templates/httproute.yaml
+++ b/dev/charts/gravitino-iceberg-rest-server/templates/httproute.yaml
@@ -1,0 +1,59 @@
+{{- /*
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+  */}}
+
+{{- if and .Values.ingress.enabled (eq (.Values.ingress.type | default "ingress") "gateway") -}}
+{{- $fullName := include "gravitino-iceberg-rest-server.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
+{{- $gateway := .Values.ingress.gateway | default dict -}}
+apiVersion: {{ $gateway.apiVersion | default "gateway.networking.k8s.io/v1" }}
+kind: HTTPRoute
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "gravitino-iceberg-rest-server.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with $gateway.parentRefs }}
+  parentRefs:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .Values.ingress.hosts }}
+  hostnames:
+    {{- range .Values.ingress.hosts }}
+    {{- if .host }}
+    - {{ .host | quote }}
+    {{- end }}
+    {{- end }}
+  {{- end }}
+  rules:
+    {{- range .Values.ingress.hosts }}
+    {{- range .paths }}
+    - matches:
+        - path:
+            type: {{ if eq (.pathType | default "") "Exact" }}Exact{{ else }}PathPrefix{{ end }}
+            value: {{ .path | quote }}
+      backendRefs:
+        - name: {{ $fullName }}
+          port: {{ $svcPort }}
+    {{- end }}
+    {{- end }}
+{{- end }}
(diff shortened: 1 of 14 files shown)


somaz94:
Gentle ping 🙏 — this has been open a few days without a review or a CI run. Could a maintainer take a look (and trigger the workflows if needed)? The change adds optional Gateway API `HTTPRoute` support to the Helm charts, fully gated behind a disabled-by-default flag. Happy to address any feedback.

danhuawang (commented):
Overall this is a well-structured PR — opt-in design (`ingress.type`), the two resources are made mutually exclusive, and the default preserves existing Ingress behavior so it's fully backward compatible. Nice work. A few observations:

---

### 1. Missing unit tests for the gravitino and lance charts

Only `gravitino-iceberg-rest-server` got `httproute_test.yaml` and the ingress gateway-mode test case. The `gravitino` and `gravitino-lance-rest-server` charts carry identical template logic but have no corresponding tests.

**Suggestion**: Add httproute unit tests for the other two charts so all three are covered (or note in the PR why only one suite was added).

---

### 2. `pathType` mapping silently collapses `ImplementationSpecific`

```yaml
type: {{ if eq (.pathType | default "") "Exact" }}Exact{{ else }}PathPrefix{{ end }}
```

Ingress supports three `pathType` values: `Exact`, `Prefix`, and `ImplementationSpecific`. Everything except `Exact` is mapped to `PathPrefix`, including `ImplementationSpecific` — whose semantics are "let the controller decide". Silently rewriting it to `PathPrefix` can change routing behavior for users who relied on it.

**Suggestion**: Document in `values.yaml` that Gateway mode only supports `Exact` and `Prefix` (the latter mapped to `PathPrefix`), or `{{ fail }}` on unsupported values so it's explicit rather than silent.

---

### 3. (Minor) Is exposing `gateway.apiVersion` as a configurable value necessary?

```yaml
gateway:
  apiVersion: gateway.networking.k8s.io/v1
```

`gateway.networking.k8s.io/v1` is currently the only GA version of HTTPRoute. Making `apiVersion` user-configurable doesn't add much beyond what hardcoding `v1` already gives, and an arbitrary override is at the user's own risk.

**Suggestion**: Either hardcode `v1`, or keep the field but add a comment clarifying that only `v1` is supported/tested.

---

### 4. (Minor) Template duplication across the three charts

The three `httproute.yaml` files are identical except for the `fullname`/`labels` helper names. This is an inherent Helm multi-chart limitation rather than a bug — just flagging in case a shared library chart was considered for the chart family.

---

### Summary

No blockers. The main actionable item is **#1 (test coverage for the other two charts)**; **#2** is a worthwhile edge-case clarification. #3 and #4 are minor/optional. Design and backward-compatibility handling look solid. 👍

somaz94:
Thanks for the thorough review @danhuawang! Addressed all points:

**1. Missing unit tests for the `gravitino` and `gravitino-lance-rest-server` charts**
Added `tests/httproute_test.yaml` to both charts (mirroring the iceberg suite: default-off, ingress-mode-off, full gateway render with parentRefs/hostnames/annotations/multi-path, and backend name/port). Also wired two new steps into `.github/workflows/chart-test.yaml` so `helm unittest` actually runs for all three charts in CI (previously only iceberg was tested).

**2. `pathType` silently collapsing `ImplementationSpecific` → `PathPrefix`**
The template no longer maps silently. It now explicitly maps:
- `Exact` → `Exact`
- `Prefix` / `PathPrefix` → `PathPrefix`
- `ImplementationSpecific` → `PathPrefix` (documented — Gateway API has no equivalent, so it is treated as a prefix match)
- anything else → `{{ fail }}` with a clear message (`RegularExpression` is intentionally not supported)

This is documented in each chart's `values.yaml`, and covered by new test cases (the `ImplementationSpecific` mapping and a `failedTemplate` assertion for an unsupported value).

**3. `gateway.apiVersion` configurability**
Kept it configurable for forward-compatibility, but added a comment in `values.yaml` noting that only `gateway.networking.k8s.io/v1` (the GA version) is supported and tested.

**4. Template duplication across the three charts**
Agreed it's not ideal — it stems from Helm's per-chart template layout (each `httproute.yaml` only differs by the `<chart>.fullname` / `<chart>.labels` includes). A shared library chart would be the proper fix; happy to open a follow-up issue for that if you'd like, to keep this PR scoped.

Local results: `helm unittest --with-subchart=false` passes for all three charts (gravitino 6, lance 6, iceberg 30), and `helm template` renders cleanly with the `fail` path verified. Could you re-trigger the CI workflows when you have a moment? Thanks again!

jerryshao:
@somaz94 can you please fix the conflicts, thanks.

somaz94:
@jerryshao Conflicts resolved — rebased onto the latest `main` (`1693ec5f`).

One note on the resolution: the only real conflict was in `.github/workflows/chart-test.yaml`. While this PR was open, upstream independently added the same `helm unittest` steps for the `gravitino` and `gravitino-lance-rest-server` charts that I had added per @danhuawang's review. Since they are identical, I dropped my redundant change, so this PR no longer touches `chart-test.yaml`. The new `tests/httproute_test.yaml` suites still get exercised, because the now-upstream CI wiring runs `helm unittest` on all three charts.

Local validation on the rebased branch:
- `helm unittest --with-subchart=false` — gravitino (36 tests), gravitino-iceberg-rest-server (30 tests), gravitino-lance-rest-server (29 tests), all pass
- `helm lint` — all three charts pass

Ready for another look. 🙏

jerryshao:
@danhuawang can you please help to review again?

danhuawang:
LGTM

somaz94:
Thanks @danhuawang! CI is green (the skipped checks are the unaffected modules). Would you be able to formally approve, or is there another committer I should loop in to help land this?

danhuawang:
> Thanks @danhuawang! CI is green (the skipped checks are the unaffected modules). Would you be able to formally approve, or is there another committer I should loop in to help land this?

Could you help check the Copilot comments if make sense? You can reply or commit for improvement.

somaz94:
@danhuawang Thanks for the review! I went through all of Copilot's comments — they were valid, so I pushed `8606c95` addressing each:

**1. `values.yaml` comment claiming `networking.k8s.io/v1`** (all 3 charts)
The Ingress template selects the apiVersion by cluster version (`v1` / `v1beta1` / `extensions/v1beta1`), so I reworded the comment to be version-neutral.

**2. HTTPRoute host/path semantics** (all 3 charts)
Good catch — merging every host into a single `spec.hostnames` while flattening all paths into `spec.rules` changed the per-host matching that Ingress guarantees. I now render **one HTTPRoute per host** (separate documents). A single host keeps the original resource name for backward compatibility; additional hosts get a `-<index>` suffix. Each route carries only its own hostname and its own paths.

**3. Missing gateway-mode Ingress-suppression test** (gravitino, lance)
Added the `ingress.enabled=true` + `type=gateway` → `hasDocuments: 0` case to both charts, matching the iceberg suite.

**4. Stray leading/trailing whitespace** in `gravitino/templates/ingress.yaml`
Cleaned up to match the other charts.

I also added a multi-host isolation test to the iceberg `httproute_test.yaml` to lock in the per-host behavior.

Verified locally: `helm unittest` passes on all three charts (37 / 31 / 30 tests), `helm template` confirms multi-host → one route per host / single-host → one route (unchanged name) / default `ingress` mode unaffected, and `helm lint` is clean.

danhuawang:
> @danhuawang Thanks for the review! I went through all of Copilot's comments — they were valid, so I pushed `8606c95` addressing each:
> 
> **1. `values.yaml` comment claiming `networking.k8s.io/v1`** (all 3 charts) The Ingress template selects the apiVersion by cluster version (`v1` / `v1beta1` / `extensions/v1beta1`), so I reworded the comment to be version-neutral.
> 
> **2. HTTPRoute host/path semantics** (all 3 charts) Good catch — merging every host into a single `spec.hostnames` while flattening all paths into `spec.rules` changed the per-host matching that Ingress guarantees. I now render **one HTTPRoute per host** (separate documents). A single host keeps the original resource name for backward compatibility; additional hosts get a `-<index>` suffix. Each route carries only its own hostname and its own paths.
> 
> **3. Missing gateway-mode Ingress-suppression test** (gravitino, lance) Added the `ingress.enabled=true` + `type=gateway` → `hasDocuments: 0` case to both charts, matching the iceberg suite.
> 
> **4. Stray leading/trailing whitespace** in `gravitino/templates/ingress.yaml` Cleaned up to match the other charts.
> 
> I also added a multi-host isolation test to the iceberg `httproute_test.yaml` to lock in the per-host behavior.
> 
> Verified locally: `helm unittest` passes on all three charts (37 / 31 / 30 tests), `helm template` confirms multi-host → one route per host / single-host → one route (unchanged name) / default `ingress` mode unaffected, and `helm lint` is clean.

@somaz94 Thanks for your contribution, it's good to me.    @jerryshao Could you help merge this PR?

jerryshao (approved):

somaz94:
Thanks for the review and approval, @jerryshao! This has been approved and CI is green with no unresolved threads. Would a committer be able to help merge it when convenient? Happy to rebase if anything drifts. 🙏

Issue #11307, opened 2026-06-01
[FEATURE] Support trusted delegated authorization for Trino using Gravitino RBAC


fzhsbc:
### Describe the feature

I would like to propose an official Gravitino integration path for enforcing Gravitino RBAC from Trino's authenticated session identity, especially for BI/JDBC/ODBC deployments.

Gravitino already has Trino connector authentication and session forwarding support, including `gravitino.client.session.forwardUser=true` for simple auth. That is useful, but it still leaves an important deployment gap:

- Trino is often the system that authenticates end users, for example through password, OAuth2, LDAP, Kerberos, or an enterprise gateway.
- BI tools such as PowerBI, Superset, Looker, JDBC, or ODBC clients may not be able to forward a Gravitino-audience OAuth token reliably to Gravitino.
- In those deployments, the trusted identity boundary is Trino's authenticated `Identity`, not a user name or token self-reported by the downstream client.
- Gravitino RBAC should still be the policy source for metadata objects, instead of maintaining a second policy system only in Trino.

### Motivation

For a multi-user Trino deployment backed by Gravitino, a common desired flow is:

```text
BI/JDBC/ODBC user -> Trino authentication -> Trino Identity -> Gravitino RBAC authorization -> query allowed/denied
```

The security property we need is: the user name sent to Gravitino for an authorization decision must be derived from Trino's authenticated session identity and accepted only from a trusted Trino service principal.

Forwarding a user token to Gravitino is not always practical because some BI/ODBC clients cannot provide a suitable Gravitino-audience OAuth token. Forwarding only a client-supplied user name is also not sufficient unless Trino has already authenticated and bound that name to the session identity.

### Describe the solution

A possible upstream-friendly design could include the following pieces:

1. A Gravitino-side delegated authorization API, for example:

```text
POST /api/metalakes/{metalake}/authorize
```

The request would contain the target end user, metadata object, privilege, and optional audit context such as Trino query ID. The endpoint should only accept requests from configured trusted service principals, such as a Trino coordinator service account.

2. A documented Trino integration using Trino's `SystemAccessControl` SPI.

[...]

PR #11309, opened 2026-06-01
[DOC] Add Trino authorization guidance

fzhsbc:

Add Trino connector authorization documentation that explains identity boundaries between the Trino end user, Trino service identity, and Gravitino user. The page documents session user forwarding, BI/JDBC/ODBC considerations, Trino access-control integration guidance, and a typical mapping from Trino operations to Gravitino RBAC privileges.

The Trino connector index is also updated to link both Authentication and Authorization pages.

In multi-user Trino deployments, connector authentication to Gravitino is not the same as query authorization. BI/JDBC/ODBC clients may not always be able to forward a Gravitino-audience OAuth token, so deployments need clear guidance on how to avoid trusting client-supplied user names and how to bind Trino's authenticated session identity to Gravitino policies.

Related to #11307.

Yes. It adds documentation only. No runtime behavior changes.

- Ran `git diff --check`.
- Reviewed the generated Markdown content and Trino connector index changes locally.

### DIFF
--- a/docs/trino-connector/authorization.md
+++ b/docs/trino-connector/authorization.md
@@ -1,0 +1,93 @@
+---
+title: "Apache Gravitino Trino connector Authorization"
+slug: /trino-connector/authorization
+keyword: gravitino connector trino authorization rbac access control
+license: "This software is licensed under the Apache License version 2."
+---
+
+## Authorization
+
+The Gravitino Trino connector can authenticate to the Gravitino server, but query
+authorization also depends on how Trino authenticates users and how the deployment
+binds the Trino session user to Gravitino policies.
+
+For multi-user deployments, distinguish the following identities:
+
+- **Trino end user**: the user authenticated by Trino for a query session.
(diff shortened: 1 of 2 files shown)


roryqi:
In my opinion, we just enable the Trino authentication. Then we enable the Gravitino authorization. We  don't need other steps.

diqiu50 (commented):
Thanks for the contribution!

I have a question about the purpose of this document. The `SystemAccessControl` integration described here doesn't exist in Gravitino yet, and the delegated authorization API proposed in #11307 hasn't been implemented either. Publishing this doc now risks misleading users who try to follow it.

Two directions come to mind:

- **If the goal is to highlight current gaps** — a short "Limitations" note inside `authentication.md` would be more appropriate than a standalone page.
- **If the goal is to guide users on what works today** — the doc needs concrete, working configuration examples that users can actually follow end-to-end.

Could you clarify the intent?

PR #11318, opened 2026-06-01
[#11285] fix(iceberg-rest): default JDBC catalog to strict mode

### LINKED ISSUE #11285: [Bug report] JDBC backend allows creating tables in non-existent namespaces instead of returning 404
### Version

main branch

### Describe what's wrong

When using the JDBC catalog backend, the `POST /v1/namespaces/{namespace}/tables` endpoint allows creating a table in a namespace that does not exist. According to the Iceberg REST catalog specification, this should return HTTP 404 (`NoSuchNamespaceError`).

The memory catalog backend correctly rejects this request with a 404, but the JDBC backend implicitly creates the namespace and successfully creates the table.

No error is raised — that is the problem. The test expects a `NoSuchNamespaceError` but the operation succeeds:

```
FAILED test_create_table_with_invalid_database[rest_catalog] - Failed: DID NOT RAISE <class 'pyiceberg.exceptions.NoSuchNamespaceError'>
```

### How to reproduce

1. Start Gravitino Iceberg REST server with JDBC catalog backend (e.g., SQLite or H2):

```properties
gravitino.iceberg-rest.catalog-backend = jdbc
gravitino.iceberg-rest.uri = jdbc:sqlite::memory:
gravitino.iceberg-rest.jdbc-driver = org.sqlite.JDBC
gravitino.iceberg-rest.warehouse = s3://warehouse/
```

2. Attempt to create a table in a namespace that does not exist:

```bash
# Verify namespace does not exist
curl http://localhost:8181/iceberg/v1/namespaces/invalid
# Should return 404

# Create table in non-existent namespace
curl -X POST http://localhost:8181/iceberg/v1/namespaces/invalid/tables \
  -H "Content-Type: application/json" \
  -d '{"name": "test_table", "schema": {"type": "struct", "fields": [{"id": 1, "name": "id", "type": "long", "required": false}]}}'
# Expected: 404 NoSuchNamespaceError
# Actual: 200 OK (table created, namespace implicitly created)
```

3. Alternatively, reproduce with pyiceberg:

```python
from pyiceberg.catalog.rest import RestCatalog

catalog = RestCatalog("rest", **{
    "auth": {"type": "noop"},
})

# This should raise NoSuchNamespaceError but succeeds
table = catalog.create_table(
    ("invalid", "test_table"),
    schema=...,
)
```

### Additional context

- The memory catalog backend (`gravitino.iceberg-rest.catalog-backend = memory`) correctly returns 404 for this case.
- This behavior violates the Iceberg REST catalog specification which states that `POST /v1/namespaces/{namespace}/tables` should return 404 if the namespace does not exist.
- The JDBC backend appears to implicitly create namespaces when creating tables, which can lead to unexpected namespace proliferation and makes it impossible for clients to detect typos in namespace names.

roryqi:

Default the Iceberg `jdbc.strict-mode` property to `true` for the JDBC catalog backend, in `IcebergCatalogUtil#loadJdbcCatalog` (via `putIfAbsent`, mirroring the existing `jdbc.schema-version` default). The value can still be overridden by an explicit config.

The JDBC backend implicitly created the namespace when a table (or view) was created under a non-existent namespace, returning `200 OK` instead of the HTTP `404 NoSuchNamespaceError` required by the Iceberg REST catalog spec. The memory backend already returns 404. This caused namespace proliferation, hid namespace-name typos, and was spec-non-compliant. With strict mode, `JdbcTableOperations`/`JdbcViewOperations` validate namespace existence before commit and throw `NoSuchNamespaceException`.

Fix: #11285

Yes. For the Iceberg REST service backed by the JDBC catalog, creating a table or view in a namespace that does not exist now fails with HTTP 404 instead of implicitly creating the namespace. Operators relying on the old implicit behavior can restore it by explicitly setting `gravitino.iceberg-rest.jdbc.strict-mode = false`. No property keys were added or removed (the property is provided by Iceberg).

- Unit tests in `TestIcebergCatalogUtil`: default strict mode rejects creating a table in a missing namespace with `NoSuchNamespaceException`; an explicit `jdbc.strict-mode=false` is not overridden and allows implicit creation.
- Integration test `IcebergRESTServiceIT#testCreateTableInNonExistentNamespace` exercises the end-to-end REST path and runs across the memory, hive, and JDBC backends.


catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergConstants.java
```diff
@@ -38,6 +38,8 @@ public class IcebergConstants {
   public static final String GRAVITINO_JDBC_SCHEMA_VERSION = "jdbc-schema-version";
   public static final String ICEBERG_JDBC_SCHEMA_VERSION = "jdbc.schema-version";
 
+  public static final String ICEBERG_JDBC_STRICT_MODE = "jdbc.strict-mode";
```
mchades, line 41: Do you need to update the user documentation?

roryqi: Good catch. Added a `gravitino.iceberg-rest.jdbc.strict-mode` row (default `true`, since 1.3.0) to the JDBC backend configuration table in `docs/iceberg-rest-service.md` in f657fb093.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/integration/test/IcebergRESTServiceIT.java
```diff
@@ -305,6 +305,37 @@ void testCreateTable() {
...
+                sql(
+                    String.format(
+                        "CREATE TABLE %s.create_in_absent(id bigint) using iceberg",
+                        namespaceName)));
+    // Creating a table in a namespace that does not exist must surface a NoSuchNamespace error
+    // (HTTP 404) instead of implicitly creating the namespace. This holds for the memory and hive
+    // backends, and for the JDBC backend now that strict mode is enabled by default.
+    Assertions.assertTrue(
+        isNoSuchNamespace(thrown), () -> "Expected a NoSuchNamespace error but got: " + thrown);
+  }
```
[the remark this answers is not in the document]
roryqi: Added `IcebergRESTServiceIT#testCreateViewInNonExistentNamespace` in f657fb093, which creates a source table in an existing namespace and then asserts that `CREATE VIEW` in a non-existent namespace surfaces a `NoSuchNamespace` error (gated by `@EnabledIf("isSupportsViewCatalog")`).

mchades (approved):

PR #11322, opened 2026-06-01
[MINOR] improvement(server): expose schema separator via the /configs endpoint

roryqi:

- Expose the `gravitino.schema.separator` server config through Gravitino's `/configs` endpoint (`ConfigServlet`) by adding it to the always-exposed `basicConfigEntries`.
- Update `TestConfigServlet` expectations and add a test verifying a custom separator value is reflected in the response.

The schema separator determines how hierarchical (multi-level) schema names are represented at the API boundary (e.g. `:` in `A:B:C`). Clients and the web UI need to discover the configured separator to correctly construct and parse hierarchical schema names. Previously it was only a server-side config and was not exposed to clients.

Yes. The `/configs` endpoint now includes the `gravitino.schema.separator` key (default `:`).

Added/updated unit tests in `TestConfigServlet`; `./gradlew :server:test --tests "org.apache.gravitino.server.web.TestConfigServlet"` passes (8 tests).

### DIFF
--- a/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
+++ b/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
@@ -51,7 +51,7 @@ public class TestConfigServlet {
     configServlet.doGet(null, res);
     verify(writer)
         .write(
-            "{\"gravitino.authorization.enable\":false,\"gravitino.authenticators\":[\"simple\"]}");
+            "{\"gravitino.schema.separator\":\":\",\"gravitino.authorization.enable\":false,\"gravitino.authenticators\":[\"simple\"]}");
     configServlet.destroy();
   }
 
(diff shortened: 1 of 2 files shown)


mchades (dismissed):

server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
```diff
@@ -111,7 +111,24 @@ public void testConfigServletWithVisibleConfigs() throws Exception {
     configServlet.doGet(null, res);
     verify(writer)
         .write(
-            "{\"gravitino.extended.custom.config\":\"test\",\"gravitino.authorization.enable\":false,\"gravitino.authenticators\":[\"simple\"]}");
+            "{\"gravitino.extended.custom.config\":\"test\",\"gravitino.schema.separator\":\":\",\"gravitino.authorization.enable\":false,\"gravitino.authenticators\":[\"simple\"]}");
```
jerryshao: The code is too long, you should fix them.

roryqi: Refactored in ca201070a: the test now parses the servlet's JSON response into a `Map` and asserts on individual config entries via a shared `fetchConfigs` helper, removing the long, order-dependent exact-match JSON strings.

jerryshao (approved):

PR #11343, opened 2026-06-02
[#11171] improvement(core): add JSON formatter for audit logs

### LINKED ISSUE #11171: [FEATURE] Add JSON formatter for audit logs
### Describe the feature

Add a `JsonAuditFormatter` to the audit subsystem alongside the existing `SimpleFormatter` and `SimpleFormatterV2` TSV formatters.

The new formatter must:
- Output one JSON object per audit row.
- Use ISO 8601 timestamps with millisecond precision and explicit timezone offset.
- Serialize all fields including the `customInfo` map, which is currently silently dropped by both TSV formatters despite being part of the `AuditLog` interface.
- Scrub sensitive HTTP headers (`Authorization`, `Cookie`, `X-Amz-Security-Token`) and credential-bearing properties (`s3.access-key-id`, `jdbc-password`) before serialization.

### Motivation

The two existing formatters (`SimpleFormatter`, `SimpleFormatterV2`) both produce tab-separated text, which has three practical problems:

1. **SIEM incompatibility.** Enterprise SIEMs (Splunk, Datadog, Sumo Logic, Elastic) parse JSON natively. TSV requires custom parsers that break on every schema change.

2. **Industry alignment.** Structured JSON is the standard for compliance audit output (AWS CloudTrail, Kubernetes audit, Snowflake, Databricks Unity Catalog, Apache Polaris, OpenTelemetry). TSV is an outlier that creates friction during compliance evaluations.

3. **Blocks structured data capture.** Response payloads from list operations and credential-vending events are structured data that cannot be represented in flat TSV. A JSON formatter is a prerequisite to capturing that data in audit rows.

### Describe the solution

- Ship a `JsonAuditFormatter` class implementing the existing `Formatter` interface.
- Make the JSON formatter the new default; operators requiring legacy TSV output can opt back in via configuration.
- Retain `SimpleFormatter` and `SimpleFormatterV2` as-is for backward compatibility.

### Additional context

The `customInfo` map is already present on the `AuditLog` interface but is not serialized by either current formatter. The JSON formatter should be the first to expose it as a structured field.

freesinger:
Implement audit formatter to provide structured JSON so SIEM consumers can parse audit events reliably, while preserving compatibility through the existing formatter configuration. Also add coverage for JSON serialization, redaction, and default formatter wiring.

This PR adds a structured JSON formatter for audit logs.

The main changes are:

1. Add `JsonAuditFormatter` to serialize each audit log entry as one JSON object per line.
2. Include all core audit fields in the JSON output, including structured `customInfo`.
3. Format `timestamp` in ISO 8601 with millisecond precision and an explicit timezone offset.
4. Redact sensitive values before serialization for the following keys:
   - `Authorization`
   - `Cookie`
   - `X-Amz-Security-Token`
   - `s3.access-key-id`
   - `jdbc-password`
5. Emit `resultCount` as a top-level JSON field for `ListEvent` when the count is available.
6. Update the Helm config template and server configuration documentation to reflect the new default formatter.
7. Add unit tests for JSON serialization, redaction, null identifier handling, list event count output, and default formatter wiring.

The existing audit formatters mainly emit tab-separated text, which is harder for SIEM systems and other downstream log processors to consume reliably.

This change is needed because:

1. Structured JSON is easier for SIEM systems to parse and index than TSV output.
2. `AuditLog.customInfo()` should be preserved in a structured way instead of being difficult to parse downstream.
3. Audit logs may include HTTP headers or credential-related properties, so sensitive values must be masked before being written.

Fix: #11171

Yes.

1. Audit logs now expose `customInfo` as structured JSON content instead of relying on TSV-compatible string formatting.
2. Sensitive values for specific headers and properties are redacted in the JSON audit output.

The patch was tested with targeted unit tests covering the new formatter and the default formatter wiring.

Executed test command:

```bash
./gradlew :core:test --tests org.apache.gravitino.audit.TestJsonAuditFormatter --tests org.apache.gravitino.audit.TestAuditManager --tests org.apache.gravitino.audit.TestFileAuditWriter
```

The tests cover:

1. Core JSON field serialization.
2. ISO 8601 timestamp serialization with millisecond precision.
3. Sensitive field redaction.
4. `ListEvent` `resultCount` serialization.
5. Null identifier handling.
6. Default audit formatter wiring through `AuditLogManager`.


core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -401,7 +401,7 @@ private Configs() {}
           .doc("Gravitino event log formatter class name")
           .version(ConfigConstants.VERSION_0_7_0)
           .stringConf()
-          .createWithDefault(SimpleFormatterV2.class.getName());
+          .createWithDefault(JsonAuditFormatter.class.getName());
```
jerryshao: We'd better not change the default format.

core/src/main/java/org/apache/gravitino/audit/JsonAuditFormatter.java
```diff
@@ -0,0 +1,114 @@
...
+  private static final ObjectMapper OBJECT_MAPPER = JsonMapper.builder().build();
+
+  private static final DateTimeFormatter TIMESTAMP_FORMATTER =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSXXX").withZone(ZoneId.systemDefault());
+
+  private static final String REDACTED_VALUE = "***";
+
+  private static final Set<String> MASKED_CUSTOM_INFO_KEYS =
+      ImmutableSet.of(
+          "authorization", "cookie", "x-amz-security-token", "s3.access-key-id", "jdbc-password");
```
jerryshao: Can you also do this redaction for the simple audit format?

core/src/main/java/org/apache/gravitino/Configs.java
freesinger: revert

core/src/main/java/org/apache/gravitino/audit/JsonAuditFormatter.java
freesinger: I extract an `AuditLogRedactor` class to do this stuff.

freesinger:
The Docker CI failure this PR run into is caused by `dev/docker/gravitino/gravitino-dependency.sh` assuming bundle jars still exist under `bundles/*/build/libs` after compileDistribution.                                                                                                            
                                                                                                                                                                                                                                                                                  
However, compileDistribution depends on cleanDistributionPackage, and cleanDistributionPackage deletes all subproject build/libs directories. Since bundle modules are excluded from copySubprojectLib, compileDistribution does not rebuild those bundle jars. As a result, the subsequent find command fails because `bundles/aliyun-bundle/build/libs` does not exist.                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                  
The fix is to explicitly rebuild the required bundle shadow jars after compileDistribution and before the find/copy steps.

I'm trying to fix this in another PR, could you take a look at this PR?  @jerryshao

jerryshao:
> The Docker CI failure this PR run into is caused by `dev/docker/gravitino/gravitino-dependency.sh` assuming bundle jars still exist under `bundles/*/build/libs` after compileDistribution.
> 
> However, compileDistribution depends on cleanDistributionPackage, and cleanDistributionPackage deletes all subproject build/libs directories. Since bundle modules are excluded from copySubprojectLib, compileDistribution does not rebuild those bundle jars. As a result, the subsequent find command fails because `bundles/aliyun-bundle/build/libs` does not exist.
> 
> The fix is to explicitly rebuild the required bundle shadow jars after compileDistribution and before the find/copy steps.
> 
> I'm trying to fix this in another PR, could you take a look at this PR? @jerryshao

OK.

jerryshao:
You should also bump the version in the Chart file to fix the Chart test issue.

yuqi1129:
> The Docker CI failure this PR run into is caused by `dev/docker/gravitino/gravitino-dependency.sh` assuming bundle jars still exist under `bundles/*/build/libs` after compileDistribution.
> 
> However, compileDistribution depends on cleanDistributionPackage, and cleanDistributionPackage deletes all subproject build/libs directories. Since bundle modules are excluded from copySubprojectLib, compileDistribution does not rebuild those bundle jars. As a result, the subsequent find command fails because `bundles/aliyun-bundle/build/libs` does not exist.
> 
> The fix is to explicitly rebuild the required bundle shadow jars after compileDistribution and before the find/copy steps.
> 
> I'm trying to fix this in another PR, could you take a look at this PR? @jerryshao

This problem should have been fixed by https://github.com/apache/gravitino/pull/11388

jerryshao:
I have no further comments. @freesinger can you fix some comments from Copilot?

docs/gravitino-server-config.md
```diff
@@ -259,7 +259,7 @@ Gravitino provides a default implementation to log basic audit information to a
 
 #### Audit log formatter
 
-The `Formatter` interface transforms an `Event` into an `AuditLog`. `SimpleFormatterV2` is the default implementation and requires no extra configuration. It produces a tab-separated line with the following fields: timestamp, user, operation type, identifier, operation status, event source, remote address, and custom info.
+The `Formatter` interface transforms an `Event` into an `AuditLog`. `SimpleFormatterV2` is the default implementation. `JsonAuditFormatter` is also available when structured JSON output is required. It emits one JSON object per line, serializes `customInfo`, and formats `timestamp` as ISO 8601 with millisecond precision and zone offset. Both simple and JSON formatters mask sensitive values such as `Authorization`, `Cookie`, `X-Amz-Security-Token`, `s3.access-key-id`, and `jdbc-password`.
```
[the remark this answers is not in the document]
freesinger: PR description has been updated to state that default behavior is unchange and default formatter is `SimpleFormatterV2`

dev/charts/gravitino/resources/config/gravitino.conf
```diff
@@ -173,7 +173,7 @@ gravitino.iceberg-rest.default-catalog-name = {{ .Values.icebergRest.dynamicConf
 # Audit log configuration
 gravitino.audit.enabled = {{ .Values.audit.enabled }}
 gravitino.audit.writer.className = {{ if (and .Values.audit .Values.audit.writer .Values.audit.writer.className) }}{{ .Values.audit.writer.className }}{{ else }}org.apache.gravitino.audit.FileAuditWriter{{- end }}
-gravitino.audit.formatter.className = {{ if (and .Values.audit .Values.audit.formatter .Values.audit.formatter.className) }}{{ .Values.audit.formatter.className }}{{ else }}org.apache.gravitino.audit.SimpleFormatter{{- end }}
+gravitino.audit.formatter.className = {{ if (and .Values.audit .Values.audit.formatter .Values.audit.formatter.className) }}{{ .Values.audit.formatter.className }}{{ else }}org.apache.gravitino.audit.v2.SimpleFormatterV2{{- end }}
```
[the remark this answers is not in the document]
freesinger: ditto

freesinger:
@jerryshao Copilot comments resolved.

jerryshao (approved):

PR #11344, opened 2026-06-02
[#11335] web(ui): setting web v2 ui as default

### LINKED ISSUE #11335: [Improvement] Setting web v2 UI as default

Setting web v2 UI as default in env config

### How should we improve?

_No response_

LauraXia123:
1. setting web v2 ui as default
2. update md and readme
image

N/A

Fix: #11335

N/A

manually

### DIFF
--- a/conf/gravitino-env.sh.template
+++ b/conf/gravitino-env.sh.template
@@ -33,6 +33,6 @@ GRAVITINO_VERSION=GRAVITINO_VERSION_PLACEHOLDER
 #                             # Appended to JAVA_OPTS by launch scripts; set GRAVITINO_MEM to change heap/metaspace sizes.
 
 # UI selection behaviour:
-# - By default this template does not set `GRAVITINO_USE_WEB_V2`, so the container or environment can control which UI is used.
-# - To force the v1 UI, explicitly set it to `false`; to enable the v2 UI, set it to `true` (for example: `export GRAVITINO_USE_WEB_V2=true`).
-# GRAVITINO_USE_WEB_V2=false
+# - Leave `GRAVITINO_USE_WEB_V1` unset, or set it to `false`, to use the v2 UI.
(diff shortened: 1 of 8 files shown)


conf/gravitino-env.sh.template
```diff
@@ -33,6 +33,6 @@ GRAVITINO_VERSION=GRAVITINO_VERSION_PLACEHOLDER
 #                             # Appended to JAVA_OPTS by launch scripts; set GRAVITINO_MEM to change heap/metaspace sizes.
 
 # UI selection behaviour:
-# - By default this template does not set `GRAVITINO_USE_WEB_V2`, so the container or environment can control which UI is used.
-# - To force the v1 UI, explicitly set it to `false`; to enable the v2 UI, set it to `true` (for example: `export GRAVITINO_USE_WEB_V2=true`).
-# GRAVITINO_USE_WEB_V2=false
\ No newline at end of file
+# - Leave `GRAVITINO_USE_WEB_V1` unset, or set it to `false`, to use the v2 UI.
+# - Set `GRAVITINO_USE_WEB_V1=true` only when you want to force the v1 UI.
+# GRAVITINO_USE_WEB_V1=false
```
jerryshao, line 38: Removing `GRAVITINO_USE_WEB_V2` and introducing a new `GRAVITINO_USE_WEB_V1` will break the compatibility if the user already uses `GRAVITINO_USE_WEB_V2`. I think we make `GRAVITINO_USE_WEB_V2` to `true` as a default value, no need to introduce a new  `GRAVITINO_USE_WEB_V1` env variable.

LauraXia123: Directly setting GRAVITINO_USE_WEB_V2 to true by default causes it to not take effect when set via container environment variables.

jerryshao: I think they're two separate problems; we'd better fix them one by one.

PR #11346, opened 2026-06-02
[#11345] improvement(deps): Exclude unused legacy netty dependency from hive-metastore2-libs

### LINKED ISSUE #11345: [Improvement] Exclude unused legacy netty dependency from hive-metastore2-libs

The `hive-metastore2-libs` module includes an unused transitive dependency `io.netty:netty:3.10.6.Final` pulled by `hadoop-mapreduce-client-core:3.3.6`. This legacy netty-3.x jar is not needed at runtime and increases the package size unnecessarily.

### How should we improve?

Add `exclude(group = "io.netty", module = "netty")` to the `hadoop3-mapreduce-client-core` dependency in `build.gradle.kts`.

geyanggang:

Exclude transitive i`o.netty:netty:3.10.6.Final` from `hadoop-mapreduce-client-core` in `build.gradle.kts`.

The legacy `netty-3.x `is an unused transitive dependency brought in by Hadoop. It is not required by Hive Metastore or Gravitino runtime and should be excluded to keep the dependency tree clean.

Fix: #11345 

No.

- Verified `io.netty:netty:3.10.6.Final` no longer appears in `:catalogs:hive-metastore2-libs:dependencies`
- `./gradlew :catalogs:hive-metastore2-libs:build` passes successfully

### DIFF
--- a/catalogs/hive-metastore2-libs/build.gradle.kts
+++ b/catalogs/hive-metastore2-libs/build.gradle.kts
@@ -62,6 +62,7 @@ dependencies {
     exclude(group = "com.google.code.findbugs")
     exclude(group = "com.google.guava")
     exclude(group = "com.sun.jersey")
+    exclude(group = "io.netty", module = "netty")
     exclude(group = "log4j")
     exclude(group = "org.apache.avro")
     exclude(group = "org.apache.logging.log4j")


jerryshao:
Do we need this `hadoop-mapreduce-client-core` for hms2 module?

geyanggang:
> Do we need this `hadoop-mapreduce-client-core` for hms2 module?

We should not need it. `hadoop-mapreduce-client-core` is an optional dependency of `hive-metastore:2.3.9`. It was carried over from the old Hadoop `2.10.2` setup and kept when we upgraded to Hadoop `3.3.6`. Since we only use the Hive metastore client for metadata operations, not MapReduce jobs, it should be safe to remove entirely. I'll verify with integration tests and update this PR to drop the whole dependency.

geyanggang:
> Do we need this `hadoop-mapreduce-client-core` for hms2 module?

After further investigation, hadoop-mapreduce-client-core is still needed at runtime — Hive Metastore 2.3.9 loads classes from it during initialization (CI integration tests confirmed: removing it causes Failed to connect to Hive Metastore). So I'm keeping the dependency and only excluding the unused transitive io.netty:netty (legacy netty-3.x) from it.

jerryshao (approved):

PR #11349, opened 2026-06-02
[#11348] feat(flink-connector): Flink connector view support to Iceberg and Paimon catalogs

### LINKED ISSUE #11348: [Improvement] Flink connector view support to Iceberg and Paimon catalogs

The Flink connector currently supports views only for the Hive catalog. Iceberg and Paimon catalogs already implement `asViewCatalog()` on the server side, so view operations should be supported there as well.

There is also a bug in `GravitinoPaimonCatalog.dropTable`: it calls `purgeTable` with no view fallback, so `DROP VIEW` on a Paimon catalog throws `TableNotExistException`.

### How should we improve?

1. Fix `GravitinoPaimonCatalog.dropTable` to fall back to `asViewCatalog().dropView()` when `purgeTable` returns false.
2. Add view integration tests for Iceberg (Hive + REST backends) and Paimon (Hive + JDBC backends).

diqiu50:

- Fix `GravitinoPaimonCatalog.dropTable` to fall back to `asViewCatalog().dropView()` when `purgeTable` returns false.
- Move view integration tests from `FlinkHiveCatalogIT` into shared `FlinkCommonIT`, gated by a `supportViewOperation()` hook.
- Enable view tests for Iceberg (Hive + REST) and Paimon (Hive + JDBC) backends.

The Flink connector only supported views for the Hive catalog. Iceberg and Paimon already implement `asViewCatalog()` server-side but lacked Flink-level coverage. `GravitinoPaimonCatalog.dropTable` also had a bug that caused `DROP VIEW` to throw `TableNotExistException`.

Fix: #11348

Yes. `CREATE/DROP/ALTER/SHOW VIEW` now work in Flink SQL for Iceberg and Paimon catalogs.

Added 9 view integration tests to `FlinkCommonIT`, run against Iceberg Hive/REST and Paimon Hive/JDBC backends.


flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalog.java
```diff
@@ -332,6 +333,16 @@ public void dropTable(ObjectPath tablePath, boolean ignoreIfNotExists)
     }
   }
 
+  /**
+   * Drops the table entry. Subclasses may override to use a different drop strategy (e.g., purge).
```
mchades, line 337: How can different drop strategies, such as purge, be controlled when the function parameter is limited to `NameIdentifier`?

api/src/main/java/org/apache/gravitino/rel/Dialects.java
```diff
@@ -39,5 +39,12 @@ public final class Dialects {
   /** The Apache Flink SQL dialect. */
   public static final String FLINK = "flink";
 
+  /**
+   * A dialect key for the canonical SQL representation of a view, used as a fallback when no
+   * engine-specific dialect (e.g. {@link #FLINK}, {@link #HIVE}) is available or preserved by the
+   * underlying metastore.
+   */
+  public static final String QUERY_DIALECT = "query";
```
mchades: Only the Paimon catalog utilizes `query`, so it is unnecessary to include it in the API module.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
```diff
@@ -103,6 +108,25 @@ protected AbstractCatalog realCatalog() {
...
+    String sql = view.getExpandedQuery();
+    return new Representation[] {
+      SQLRepresentation.builder().withDialect(Dialects.FLINK).withSql(sql).build(),
+      SQLRepresentation.builder().withDialect(Dialects.QUERY_DIALECT).withSql(sql).build()
+    };
+  }
+
+  @Override
+  protected boolean dropTableEntry(NameIdentifier ident) {
+    return catalog().asTableCatalog().purgeTable(ident);
```
mchades, line 130: why using purgeTable for dropTable?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
```diff
@@ -103,6 +108,25 @@ protected AbstractCatalog realCatalog() {
   // DDL — route through Gravitino (single source of truth)
   // ---------------------------------------------------------------------------
 
+  @Override
+  protected List<String> viewDialectFallbackOrder() {
+    return Arrays.asList(Dialects.FLINK, Dialects.QUERY_DIALECT);
```
mchades: should add HIVE dialect?

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/catalog/BaseCatalog.java
diqiu50: For Paimon, GravitinoPaimonCatalog overrides it to call purgeTable().

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
diqiu50: For Paimon managed tables, we use purgeTable() to ensure data files are cleaned up when a table is dropped from Flink

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
diqiu50: added

mchades (approved):

PR #11354, opened 2026-06-02
[MINOR] improvement(core): validate remote URI host in JobManager before download

yuqi1129:

Add shared remote URI validation and file-fetching utilities for server-side HTTP/HTTPS/FTP downloads. The validator rejects loopback, link-local, RFC 1918 private, IPv6 unique-local, multicast, unspecified, and known cloud metadata addresses by default.

Consolidate the duplicated `FetchFileUtils` implementations into `common` and use the shared implementation from `JobManager` and the Hive, Iceberg, Hadoop, and Paimon Kerberos clients. Administrators can explicitly disable unsafe-address blocking for trusted local or private endpoints.

The previous implementations passed configured or user-supplied URIs directly to the download client without checking the resolved address. This could allow the server to access internal services or cloud metadata endpoints. The duplicated download implementations also made it easy for security behavior to diverge across modules.

Yes. Remote file downloads now block unsafe addresses by default. Deployments that intentionally fetch from trusted local or private endpoints must disable the corresponding unsafe-address blocking configuration.

- `TestJobManager` covers blocked and explicitly allowed JobManager downloads.
- `TestRemoteUriValidator` covers loopback, link-local, RFC 1918 private, IPv6 unique-local, localhost, and cloud metadata addresses.
- `TestFetchFileUtils` covers shared local/remote fetching, unsafe-address configuration, concurrency, and error handling.


core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -808,6 +809,7 @@ static String fetchFileFromUri(String uri, File stagingDir, int timeoutInMs) {
         case "http":
         case "https":
         case "ftp":
+          validateRemoteUri(fileUri);
           FileUtils.copyURLToFile(fileUri.toURL(), destFile, timeoutInMs, timeoutInMs);
           break;
```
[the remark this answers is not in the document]
yuqi1129: Added comments.

roryqi:
FTP maybe a private address. Should we forbid it?

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -808,6 +809,7 @@ static String fetchFileFromUri(String uri, File stagingDir, int timeoutInMs) {
         case "http":
         case "https":
         case "ftp":
+          validateRemoteUri(fileUri);
```
jerryshao: There're several `FetchFileUtils` that have similar code. I think you should 1) Fix the similar issues for all of them; 2) Consolidate the similar code into one if possible.

core/src/main/java/org/apache/gravitino/job/JobManager.java
```diff
@@ -830,6 +832,57 @@ static String fetchFileFromUri(String uri, File stagingDir, int timeoutInMs) {
...
+
+  private static boolean isBlockedAddress(InetAddress address) {
+    // Covers loopback (127.x.x.x / ::1), link-local (169.254.x.x / fe80::/10 — includes AWS/GCP/
+    // Azure metadata), RFC-1918 private (10.x / 172.16-31.x / 192.168.x), multicast, unspecified.
+    if (address.isLoopbackAddress()
+        || address.isLinkLocalAddress()
+        || address.isSiteLocalAddress()
+        || address.isMulticastAddress()
+        || address.isAnyLocalAddress()) {
+      return true;
```
jerryshao: Why can't we support loopback or local address? Is there any issue using it? What if I have a local web server or ftp server? I think the requirement is quite fair.

yuqi1129: The main risk is SSRF. `localhost` or a private IP is resolved from the Gravitino server side, not from the user's machine.

So if a user passes `http://127.0.0.1:xxx` or `http://192.168.x.x`, Gravitino will try to access services on the Gravitino server itself or inside the server's internal network. This may expose local admin services, internal APIs, or cloud metadata endpoints.

I agree that some deployments may want to fetch files from a local HTTP/FTP server. But I think this should be disabled by default, and only enabled explicitly by an admin, for example with an allowlist/configuration. If we reach an agreement on this point, I can do as it is.

core/src/main/java/org/apache/gravitino/job/JobManager.java
yuqi1129: Got it.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/FetchFileUtils.java
jerryshao: Can we dedup this java class file into one?

catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/hive/HiveConstants.java
```diff
@@ -30,6 +30,8 @@ public class HiveConstants {
   public static final String PRINCIPAL = "kerberos.principal";
   public static final String CHECK_INTERVAL_SEC = "kerberos.check-interval-sec";
   public static final String FETCH_TIMEOUT_SEC = "kerberos.keytab-fetch-timeout-sec";
+  public static final String KEYTAB_FETCH_ALLOW_LOCAL_ADDRESS =
```
roryqi: Could we give a better name? Local address means 127.0.0.1 only. Actually it should include the 192.16.0.0 sub network.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/FetchFileUtils.java
yuqi1129: This file depends on `Hadoop` jars, which can't be shared between modules like `Paimon`, `Iceberg`, `fileset`.  Making `Hadoop` compile-only is optional; let me give it a try.

catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/hive/HiveConstants.java
yuqi1129: sure

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/FetchFileUtils.java
yuqi1129: I have used https://github.com/apache/gravitino/issues/11486 to track this issue, as different modules require Hadoop with different versions, and extracting a common module with Hadoop compile-only is risky and it's a major change, so I'm hesitant to move it forward.

catalogs/catalog-common/src/main/java/org/apache/gravitino/catalog/hive/HiveConstants.java
yuqi1129: Renamed it to `keytab-fetch-block-unsafe-address` (and `blockUnsafeAddress` for the server config). The validator also blocks multicast, unspecified, and cloud metadata addresses, so `unsafe address` reflects the complete set better than `local` or `private`. The blocking option now defaults to `true`; users can explicitly set it to `false` for trusted URIs.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/FetchFileUtils.java
jerryshao: How about using reflection to get rid of hadoop dependency issue? We just use very common hadoop FS API, so it should be compatible across versions.

core/src/main/java/org/apache/gravitino/Configs.java
```diff
@@ -551,6 +551,19 @@ private Configs() {}
           .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
           .createWithDefault(5 * 60 * 1000L); // Default is 5 minutes
 
+  public static final String JOB_REMOTE_URI_BLOCK_UNSAFE_ADDRESS_KEY =
+      "gravitino.job.remoteUri.blockUnsafeAddress";
```
jerryshao: It is better to make this 3-level name for configuration.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/FetchFileUtils.java
yuqi1129: Let me verify it locally, and it's not a minor change considering the release is approaching.

core/src/main/java/org/apache/gravitino/Configs.java
yuqi1129: changed.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/FetchFileUtils.java
yuqi1129: Removed all duplicated file `FetchFileUtils`. Other duplicated classes like `KerberosClient` can also be optimized, I would like to use https://github.com/apache/gravitino/issues/11486 to track them.

core/src/test/java/org/apache/gravitino/job/TestJobManager.java
```diff
@@ -914,6 +933,71 @@ public void testFetchFileFromUriWithMissingLocalFileShouldFail() throws IOExcept
...
+
+      String fetchedFile =
+          JobManager.fetchFileFromUri(
+              String.format("http://127.0.0.1:%d/artifact.jar", port), stagingDir, 1000, false);
+
+      Assertions.assertEquals("job artifact", Files.readString(Path.of(fetchedFile)));
+    } finally {
+      server.stop(0);
+    }
+  }
```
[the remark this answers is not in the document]
yuqi1129: Updated the PR description to reference `TestRemoteUriValidator` for URI validation coverage and `TestJobManager` for JobManager download behavior. The description now also reflects the shared `FetchFileUtils` consolidation and configuration changes.

roryqi:
one question:
Should it be a catalog property or a server configuration option?

yuqi1129:
> one question: Should it be a catalog property or a server configuration option?

Some catalogs used the method to fetch files, and the same goes for the Gravitino server, so both catalog properties and server configurations exist.

roryqi:
> > one question: Should it be a catalog property or a server configuration option?
> 
> Some catalogs used the method to fetch files, and the same goes for the Gravitino server, so both catalog properties and server configurations exist.

For me, server config option is enough. I don't wanna make it complex.

yuqi1129:
> > > one question: Should it be a catalog property or a server configuration option?
> > 
> > 
> > Some catalogs used the method to fetch files, and the same goes for the Gravitino server, so both catalog properties and server configurations exist.
> 
> For me, server config option is enough. I don't wanna make it complex.

I'm hesitant to do it, as configurations like download timeouts are separate. Unifying them did make things simple, and I have also thought about this point before. 

If this does not matter, I will merge all configurations into one.

yuqi1129:
> > > one question: Should it be a catalog property or a server configuration option?
> > 
> > 
> > Some catalogs used the method to fetch files, and the same goes for the Gravitino server, so both catalog properties and server configurations exist.
> 
> For me, server config option is enough. I don't wanna make it complex.

modified.

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * Removes the per-destination lock entry for the given file. Should be called when the
+   * destination file is deleted (for example on a Kerberos client {@code close()}) to prevent
+   * unbounded map growth.
+   *
+   * @param destFile the destination file whose lock entry should be removed
+   */
+  public static void removeLock(File destFile) {
+    SYMLINK_LOCKS.remove(destFile.toPath().toAbsolutePath().normalize().toString());
+  }
```
jerryshao, line 74: Who will call this method, I don't see any code calling this method.

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
```diff
@@ -0,0 +1,184 @@
...
+ * <p>The {@code hdfs} scheme is resolved reflectively against {@code
+ * org.apache.hadoop.fs.FileSystem} so that this class can live in the {@code common} module, which
+ * does not declare a compile-time dependency on Hadoop. Callers that never use {@code hdfs} URIs
+ * (for example the job manager and the Paimon catalog) simply pass {@code null} for the Hadoop
+ * configuration.
+ */
+public final class FetchFileUtils {
+
+  /** The server configuration that controls unsafe remote URI blocking. */
+  public static final String BLOCK_UNSAFE_REMOTE_URI_CONFIG = "gravitino.blockUnsafeRemoteUri";
```
jerryshao: "gravitino.fetchFile.blockUnsageRemoteUri"

common/src/main/java/org/apache/gravitino/utils/FileFetcher.java
```diff
@@ -0,0 +1,184 @@
...
+      // Replace via a temporary symlink + rename to minimize the window where the destination path
+      // is absent (which could otherwise cause a concurrent reader, e.g. loginUserFromKeytab, to
+      // fail). REPLACE_EXISTING is used here; on common local filesystems (ext4, xfs, APFS) a
+      // same-directory rename is effectively atomic at the OS level.
+      Path tmpPath = destPath.resolveSibling(destPath.getFileName() + ".symlink.tmp");
+      Files.deleteIfExists(tmpPath);
+      Files.createSymbolicLink(tmpPath, srcPath);
+      Files.move(tmpPath, destPath, StandardCopyOption.REPLACE_EXISTING);
+    }
+  }
```
jerryshao, line 146: If you wan this method thread-safe, why don't you simply add a synchronized keyword?

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
yuqi1129: The `close` method in class `KerberosClient` will call it.

yuqi1129: I will remove it as you suggested using the `synchronized ` keyword below.

common/src/main/java/org/apache/gravitino/utils/RemoteUriValidator.java
```diff
@@ -0,0 +1,92 @@
...
+      return true;
+    }
+
+    byte[] bytes = address.getAddress();
+    if (isCloudMetadataAddress(bytes)) {
+      return true;
+    }
+
+    return isIpv6UniqueLocalAddress(bytes);
+  }
```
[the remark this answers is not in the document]
yuqi1129: I will postpone this change later.

common/src/main/java/org/apache/gravitino/utils/FileFetcher.java
```diff
@@ -0,0 +1,184 @@
...
+
+      switch (scheme) {
+        case "http":
+        case "https":
+        case "ftp":
+          RemoteUriValidator.validate(
+              uri,
+              blockUnsafeRemoteUri,
+              String.format("'%s' to false", BLOCK_UNSAFE_REMOTE_URI_CONFIG));
+          FileUtils.copyURLToFile(uri.toURL(), destFile, timeoutMs, timeoutMs);
```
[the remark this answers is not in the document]
yuqi1129: This case is rather rare, and I will leave it unchanged.

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
yuqi1129: changed

common/src/main/java/org/apache/gravitino/utils/FileFetcher.java
yuqi1129: Changed as suggested.

docs/apache-hive-catalog.md
```diff
@@ -29,19 +29,19 @@ The Hive catalog supports creating, updating, and deleting databases and tables
...
-| `list-all-tables`                        | Whether to list all tables in a database, including non-Hive tables such as Iceberg, Paimon, and Hudi. When false, non-Hive tables are filtered out on a best-effort basis; see the note below for known limitations.                            | false         | No                           | 0.5.1         |
-| `default.catalog`                        | The default catalog name for the Hive3 metastore backend; this configuration is ignored when using a Hive2 metastore.                                                                                                                               | hive          | No                           | 1.1.0         |
+| Property Name                            | Description                                                                                                                                                                                                                                         | Default Value  | Required                     | Since Version |
```
roryqi, line 32: Do we need to change this document?

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
```diff
@@ -0,0 +1,161 @@
...
+    Files.deleteIfExists(tmpPath);
+    Files.createSymbolicLink(tmpPath, srcPath);
+    Files.move(tmpPath, destPath, StandardCopyOption.REPLACE_EXISTING);
+  }
+
+  /**
+   * Copies an {@code hdfs} file to the local destination reflectively, so that this class does not
+   * require a compile-time dependency on Hadoop.
+   */
+  private static void copyHdfsFileToLocal(URI uri, File destFile, @Nullable Object hadoopConf)
```
roryqi: Should we use `Optional` here?

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
```diff
@@ -0,0 +1,161 @@
...
+  private static volatile boolean blockUnsafeRemoteUri = true;
+
+  private FetchFileUtils() {}
+
+  /**
+   * Sets whether remote URIs that resolve to unsafe addresses should be blocked.
+   *
+   * @param blockUnsafeRemoteUri whether to block unsafe remote URIs
+   */
+  public static void setBlockUnsafeRemoteUri(boolean blockUnsafeRemoteUri) {
```
roryqi: Is this test only?

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
```diff
@@ -0,0 +1,161 @@
...
+   * @param fileUri the source URI; a missing scheme is treated as {@code file}
+   * @param destFile the local destination file
+   * @param timeoutMs the connect/read timeout in milliseconds, applied to remote (http/https/ftp)
+   *     downloads
+   * @param hadoopConf an {@code org.apache.hadoop.conf.Configuration} instance, required only for
+   *     the {@code hdfs} scheme; may be {@code null} when no {@code hdfs} URI is fetched
+   * @return the absolute path of {@code destFile}
+   * @throws IOException if the file cannot be fetched
+   */
+  public static String fetchFileFromUri(
```
roryqi: We can add a parameter `blockUnsafeAddress`?

roryqi: Do we need to consider thread safety?

docs/apache-hive-catalog.md
yuqi1129: I removed some empty lines in the table by the way.

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
yuqi1129: It can. Both is okay for me.

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
yuqi1129: No. We used this method to set it as false(The default is true)

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
yuqi1129: I won't include it in the PR since the content is the same as the previous one. If it indeed a problem, I would rather we use another PR to solve concurrent problem all.

yuqi1129: Passing `blockUnsafeAddress` explicitly would make `FetchFileUtils` stateless. However, the catalog Kerberos clients currently have no direct access to this server-level configuration, and I do not think each caller should read `ServerConfig` directly. Supporting this cleanly would require propagating the value through the catalog initialization path and storing it in each owning component. Since this option is process-wide and initialized once during server startup, I prefer keeping the current approach in this PR unless we want to include that broader configuration plumbing.

roryqi: This is weird for me to use a static variable.

yuqi1129: `fetchFileFromUri` will only be called once at start. If this is not so elegant, let me try to optimize it.

roryqi: You can put it to GravitinoEnv? It is weird now.

yuqi1129: This is a process-wide server configuration. The Gravitino server sets it once before initializing components, while the standalone Iceberg REST server runs in a separate JVM. Therefore, no components with different values compete in the same process. To clarify my previous comment, the configuration is initialized once; `fetchFileFromUri` itself may be called multiple times by `JobManager` and Kerberos credential refreshes.

Making `FetchFileUtils` stateless would require propagating this server-level option through the catalog initialization path because the Kerberos clients currently receive only catalog properties. I think that broader dependency-injection change should be handled separately rather than partially adding a parameter here.

yuqi1129: > You can put it to GravitinoEnv? It is weird now.

Okay, let me take a try.

yuqi1129: done

roryqi (dismissed):

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -545,6 +545,15 @@ public boolean cacheEnabled() {
     return config == null || config.get(Configs.CACHE_ENABLED);
   }
 
+  /**
+   * Returns whether remote URIs that resolve to unsafe addresses should be blocked.
+   *
+   * @return whether unsafe remote URIs should be blocked
+   */
+  public boolean blockUnsafeRemoteUri() {
+    return config == null || !Boolean.FALSE.equals(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
```
jerryshao: I'm curious why can't we use `return config == null || config.get(Configs.BLOCK_UNSAFE_REMOTE_URI);`

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
```diff
@@ -1386,7 +1386,10 @@ FileSystem getFileSystem(Path path, Map<String, String> config) throws IOExcepti
             () -> {
               if (scheme.equals(SCHEME_HDFS)) {
                 return new ImpersonationHDFSFileSystemProxy(
-                        path, config, PrincipalUtils::getCurrentUserName)
+                        path,
+                        config,
+                        PrincipalUtils::getCurrentUserName,
+                        GravitinoEnv.getInstance().blockUnsafeRemoteUri())
```
jerryshao: `GravitinoEnv.getInstance().blockUnsafeRemoteUri()` will not be changed in the runtime, so I think we can call it once when FetchFileUtils is initialized. We don't have to pass this configuration each time when we call the method, WDYT?

common/src/main/java/org/apache/gravitino/utils/FetchFileUtils.java
jerryshao: Please change it.

catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
yuqi1129: @jerryshao 
The initial code is what you suggested. see https://github.com/apache/gravitino/pull/11354#discussion_r3395339830

@roryqi  He would like to make `blockUnsafeRemoteUri` a method and believes using a static field to store it is not a good idea.  Frankly, these two are also acceptable to me; the current option( method parameter) is consistent with other configurations like `cacheEnabled`, but it seems to be tedious. The option to make it as `static` field is not so elegant, but it did make the code simple.

roryqi: Rarely see a Utils class with a state.

roryqi: Maybe we can change this to singleton factory.

yuqi1129: `GravitinoEnv` is now a singleton instance, and it may be quite similar to what you mentioned.

roryqi: Actually i mean like
```
FileFectcher fetcher = FileFetcher.get();
fetcher.initiailize();
```

yuqi1129: OK, that means it won't be a utils class but a singleton instance.

yuqi1129: This seems nicer than the current one.

roryqi (approved):

PR #11355, opened 2026-06-02
[#11353] fix(hive): fix three ClassLoader/Kerberos bugs causing HMS connection failure

### LINKED ISSUE #3: [EPIC] Design and implement the basic metadata spec
Metadata schema system is the core system in Unified Catalog to describe the entities, including memory structure, on-wire format and storage layout. 

This umbrella issue tracks the whole design and implementation of Unified Catalog' schema system, including:

1. Metadata entity design & implement.
2. Entity serialization and deserialization protocol design & implement.
3. Entity storage layout design & implement.

Subtasks are listed here:

- [x] Metadata and Type Spec design ( lakehouse, zone, table, column and other basic entities), including memory structure, on-wire and storage layout design. #11 
- [x] Metadata and Type Spec implementation (memory structure). #12 
- [x] Metadata on-wire protocol implement. #16 
- [x] Medata storage layout implement. #21

### LINKED ISSUE #11353: [Bug] HiveClientFactory: three bugs cause HMS connection failure when Kerberos is enabled without impersonation
### Environment

- Gravitino version: main / v0.9.x (confirmed on v1.2.0-rc8)
- Hive Metastore version: HMS with Kerberos authentication
- Configuration: `hive.metastore.sasl.enabled=true`, impersonation **disabled**

### Problem

When a Hive catalog is configured with Kerberos authentication and impersonation is **not** enabled, connecting to HMS fails with:

```
GSS initiate failed
No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)
```

or

```
java.lang.NoSuchMethodException: HiveClientImpl.<init>(HiveVersion, Properties)
```

Root cause analysis reveals **three independent bugs** in `HiveClientFactory`, all in the same code path.

---

### Bug 1 — `baseLoader` uses TCCL, causing UGI class to be loaded twice

**Location**: `HiveClientFactory.java`, `createHiveClientWithBackend()`, lines 121 and 143

**Code**:
```java
// Current (buggy)
classloader = HiveClientClassLoader.createLoader(HIVE3,
    Thread.currentThread().getContextClassLoader());  // TCCL is not stable
```

`HiveClientClassLoader` uses `baseLoader` to determine which classes are "shared" (delegated upward via `isSharedClass`). This includes `UserGroupInformation` (UGI) and the `HiveClient` interface.

Because `Thread.currentThread().getContextClassLoader()` (TCCL) is **not stable** across threads:

- `KerberosClient.login()` runs on thread A → UGI loaded by TCCL-A → TGT stored in TCCL-A's UGI static state
- HMS connection runs on thread B → UGI loaded by TCCL-B → different `Class` object → static state is empty → TGT not found → Kerberos handshake fails

The same `UserGroupInformation` class ends up loaded by two different ClassLoaders, so their static state is not shared.

**Fix**: Use `HiveClientFactory.class.getClassLoader()` as `baseLoader` — it is stable and consistent:
```java
ClassLoader factoryCl = HiveClientFactory.class.getClassLoader();
classloader = HiveClientClassLoader.createLoader(HIVE3, factoryCl);
```

---

### Bug 2 — `HiveVersion` enum type mismatch causes `NoSuchMethodException`

**Location**: `HiveClientFactory.java`, `createHiveClientImpl()`, lines 165–167

**Code**:
```java
// Current (buggy)
Constructor<?> hiveClientImplCtor = hiveClientImplClass.getConstructor(
    HiveClientClassLoader.HiveVersion.class,  // ← system CL's HiveVersion
    Properties.class);
```

`hiveClientImplClass` is loaded from the **isolated** `HiveClientClassLoader` (it is a barrier class redefined via `defineClass`). Its constructor expects a `HiveVersion` parameter belonging to the **isolated** ClassLoader's type space.

Passing `HiveClientClassLoader.HiveVersion.class` (system CL) causes `getConstructor` to find no matching constructor → `NoSuchMethodException`.

**Fix**: Load `HiveVersion` from the isolated ClassLoader before constructing:
```java
Class<?> hiveVersionInIsolated =
    classloader.loadClass(HiveClientClassLoader.HiveVersion.class.getName());
Object isolatedVersion =
    Enum.valueOf((Class<? extends Enum>) hiveVersionInIsolated, version.name());
Constructor<?> ctor =
    hiveClientImplClass.getConstructor(hiveVersionInIsolated, Properties.class);
return (HiveClient) ctor.newInstance(isolatedVersion, properties);
```

---

### Bug 3 — Non-impersonation Kerberos path never binds the JAAS Subject to the current thread

**Location**: `HiveClientFactory.java`, `createHiveClientInternal()`, lines 205–206

**Code**:
```java
// Current (buggy) — non-impersonation branch
} else {
    return createHiveClientImpl(classloader.getHiveVersion(), properties, classloader);
}
```

`KerberosClient.login()` calls `UserGroupInformation.loginUserFromKeytab(...)`, which stores the TGT in `loginUser.subject`. However, the HMS Thrift client performs a SASL/GSSAPI handshake at connection time. GSSAPI looks for Kerberos credentials in the **current thread's JAAS Subject context** (`AccessControlContext`). Since `javax.security.auth.useSubjectCredsOnly=true` by default in the JVM, if no Subject is bound to the current thread, GSSAPI immediately fails.

`loginUser.subject` is **not** automatically bound to the current thread. It must be explicitly injected via `Subject.doAs(subject, action)` or `ugi.doAs(action)`.

The impersonation path (`createProxyHiveClientImpl`) correctly uses `ugi.doAs(...)` — the non-impersonation Kerberos path is missing the equivalent.

**Fix**:
```java
} else if (enableKerberos) {
    UserGroupInformation realUgi = kerberosClient.getRealLoginUgi();
    final HiveClientClassLoader.HiveVersion hiveVersion = classloader.getHiveVersion();
    return realUgi.doAs(
        (PrivilegedExceptionAction<HiveClient>)
            () -> createHiveClientImpl(hiveVersion, properties, classloader));
}
```

This is consistent with how `ImpalaEngineAdapter` and `ProxyHiveClientImpl` already handle the Kerberos + `doAs` pattern.

---

### Summary

| # | Root cause | Effect | Fix |
|---|-----------|--------|-----|
| 1 | `baseLoader` uses unstable TCCL | UGI loaded twice; TGT not shared across threads | Use `HiveClientFactory.class.getClassLoader()` |
| 2 | `getConstructor` passes system CL's `HiveVersion` | `NoSuchMethodException` on every `createHiveClientImpl` call | Load `HiveVersion` from isolated CL |
| 3 | Non-impersonation Kerberos path missing `doAs` | GSSAPI can't find TGT → `No valid credentials` | Wrap in `realUgi.doAs(...)` |

All three bugs exist in the current `main` branch. I am happy to submit a PR with the fixes and corresponding tests.

JoegenUSTC:

This PR fixes three independent bugs in `HiveClientFactory` that together prevent Hive Metastore
connections from working when Kerberos authentication is enabled and impersonation is **disabled**.

**Changes:**

1. **`createHiveClientWithBackend()`** – Use `HiveClientFactory.class.getClassLoader()` as
   `baseLoader` instead of `Thread.currentThread().getContextClassLoader()` (TCCL). TCCL is
   unstable across threads; using it causes `UserGroupInformation` to be loaded by two different
   ClassLoaders, making the TGT stored after `login()` invisible at connection time.

2. **`createHiveClientImpl()`** – Load `HiveVersion` from the isolated `HiveClientClassLoader`
   before calling `getConstructor()`. `HiveClientImpl` is a barrier class redefined via
   `defineClass` inside the isolated ClassLoader; its constructor expects the `HiveVersion` type
   from the isolated ClassLoader scope. Passing the system ClassLoader's `HiveVersion.class`
   causes `NoSuchMethodException`.

3. **`createHiveClientInternal()`** – Add a `realUgi.doAs()` wrapper for the non-impersonation
   Kerberos branch. The JVM default `javax.security.auth.useSubjectCredsOnly=true` means GSSAPI
   only finds credentials in the current thread's JAAS Subject context. `KerberosClient.login()`
   stores the TGT in `realLoginUgi.subject` but does not bind it to the current thread;
   `ugi.doAs()` is required to do so. The impersonation branch already does this correctly via
   `createProxyHiveClientImpl`; this fix brings the non-impersonation path in line.

Also adds `KerberosClient.getRealLoginUgi()` as the public accessor needed by fix #3, and adds
`"Cannot find Hive jar directory"` to the Hive2 fallback condition to handle environments where
HIVE3 libs directory is absent.

Without these fixes, any Hive catalog configured with Kerberos authentication and impersonation
disabled fails to connect to HMS with one or more of:

```
GSS initiate failed
No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)
```
```
java.lang.NoSuchMethodException: HiveClientImpl.<init>(HiveVersion, Properties)
```

All three bugs exist on the same code path and must be fixed together for the feature to work.

Fix: #11353

No API or configuration changes. This is a bug fix for an existing feature
(Kerberos HMS authentication without impersonation) that was silently broken.

Verified on an internal deployment with a real Kerberos-secured HMS (non-impersonation mode):
- `GET /api/metalakes/{metalake}/catalogs/{catalog}/schemas` returns schema list correctly after fix
- `GET /api/metalakes/{metalake}/catalogs/{catalog}/schemas/{schema}/tables` returns table list correctly after fix
- No `GSSException` or `NoSuchMethodException` in server logs

Unit tests in `hive-metastore-common` all pass (`./gradlew :catalogs:hive-metastore-common:test -PskipITs`).

Note: a full Kerberos integration test requires a KDC environment which is not available in
standard CI. A follow-up issue can track adding a Docker-based KDC integration test.


yuqi1129:
@JoegenUSTC 
Can you add some UTs and ITs to cover and verify your changes here?

JoegenUSTC:
> @JoegenUSTC Can you add some UTs and ITs to cover and verify your changes here?

Thanks for the suggestion!

I've added tests in a follow-up commit (`9ec3efcf`):

**Unit tests** (`TestHiveClientFactoryClassLoader`, no external dependencies, all 4 pass locally):
- Two tests for Bug 1: one proves `HiveClientFactory.class.getClassLoader()` is stable across threads (the rationale for the fix), one contrast test shows TCCL is thread-dependent and therefore an unreliable `baseLoader`.
- Two tests for Bug 2: one verifies that `createHiveClientImpl()` calls `loadClass(HiveVersion)` on the isolated ClassLoader before `getConstructor()` (so no `NoSuchMethodException`), one contrast test reproduces the original failure by passing a mismatched CL's `HiveVersion` to `getConstructor()` and asserting it throws `NoSuchMethodException`.

**Integration test** (`TestHive2HMSWithKerberosNoImpersonation`, `@Tag("gravitino-docker-test")`):
- Extends the existing `TestHive2HMSWithKerberos` but sets `authentication.impersonation-enable=false`. This exercises exactly the non-impersonation Kerberos path fixed by Bug 3 (the missing `realUgi.doAs()`). The existing test only covered `impersonation=true`.
- Requires a live KDC (Docker), so it runs under the `gravitino-docker-test` tag only.

yuqi1129:
The CI fails. Can you take time to fix it? @JoegenUSTC

JoegenUSTC:
> The CI fails. Can you take time to fix it? @JoegenUSTC
@yuqi1129 
Thanks for flagging this! I've fixed the CI failure in commit `11fbb32`.

The root cause was that `TestHive2HMSWithKerberosNoImpersonation.startHiveContainer()` was
redundantly constructing a second `HiveClientFactory` after `super.startHiveContainer()`. Since
`TestHive2HMSWithKerberos.startHiveContainer()` already calls `createHiveProperties()` via
polymorphism, the overridden `createHiveProperties()` (with `impersonation-enable=false`) was
already in effect for the initial `hiveClient` creation. The redundant second construction failed
in the CI Kerberos environment, leaving `hiveClient=null` and causing NPE in all test methods.

The fix removes the `startHiveContainer()` override entirely — only `createHiveProperties()`
needs to be overridden.

All tests pass locally (`./gradlew :catalogs:hive-metastore-common:test -PskipITs`).
Could you please approve the CI workflows to run? Thank you!

yuqi1129:
> > The CI fails. Can you take time to fix it? @JoegenUSTC
> > @yuqi1129
> > Thanks for flagging this! I've fixed the CI failure in commit `11fbb32`.
> 
> The root cause was that `TestHive2HMSWithKerberosNoImpersonation.startHiveContainer()` was redundantly constructing a second `HiveClientFactory` after `super.startHiveContainer()`. Since `TestHive2HMSWithKerberos.startHiveContainer()` already calls `createHiveProperties()` via polymorphism, the overridden `createHiveProperties()` (with `impersonation-enable=false`) was already in effect for the initial `hiveClient` creation. The redundant second construction failed in the CI Kerberos environment, leaving `hiveClient=null` and causing NPE in all test methods.
> 
> The fix removes the `startHiveContainer()` override entirely — only `createHiveProperties()` needs to be overridden.
> 
> All tests pass locally (`./gradlew :catalogs:hive-metastore-common:test -PskipITs`). Could you please approve the CI workflows to run? Thank you!

Done.

JoegenUSTC:
> > > The CI fails. Can you take time to fix it? @JoegenUSTC
> > > @yuqi1129
> > > Thanks for flagging this! I've fixed the CI failure in commit `11fbb32`.
> > 
> > 
> > The root cause was that `TestHive2HMSWithKerberosNoImpersonation.startHiveContainer()` was redundantly constructing a second `HiveClientFactory` after `super.startHiveContainer()`. Since `TestHive2HMSWithKerberos.startHiveContainer()` already calls `createHiveProperties()` via polymorphism, the overridden `createHiveProperties()` (with `impersonation-enable=false`) was already in effect for the initial `hiveClient` creation. The redundant second construction failed in the CI Kerberos environment, leaving `hiveClient=null` and causing NPE in all test methods.
> > The fix removes the `startHiveContainer()` override entirely — only `createHiveProperties()` needs to be overridden.
> > All tests pass locally (`./gradlew :catalogs:hive-metastore-common:test -PskipITs`). Could you please approve the CI workflows to run? Thank you!
> 
> Done.

@yuqi1129 
All three CI failures are transient infrastructure issues, unrelated to this PR's changes:

1. **`catalog-fileset:test`** — Job was canceled by the CI runner (timeout), not a test failure.
2. **`catalog-glue:test`** — Docker image pull timeout (`motoserver/moto:5.1.4` not fetched within 2 minutes); `trino-connector-435-439:compileJava` — JDK toolchain download timeout (`SocketTimeoutException`). Both are network issues on the CI runner.
3. **`trino-connector-473-478`** — Job was canceled by the CI runner (timeout), not a test failure.

Could you please re-run the failed jobs? Thank you!

JoegenUSTC:
@yuqi1129 Thank you for reopening! It seems the CI workflows are still awaiting approval. Could you please approve them to run? Thank you!

JoegenUSTC:
@yuqi1129  All CI checks have passed (23 successful). Could you please review and approve when you have time? Thank you!

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/client/HiveClientFactory.java
```diff
@@ -162,10 +164,18 @@ public static HiveClient createHiveClientImpl(
...
+    // the two Class objects are not identical even though they have the same name.
+    Class<?> hiveVersionInIsolated =
+        classloader.loadClass(HiveClientClassLoader.HiveVersion.class.getName());
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    Object isolatedVersion =
+        Enum.valueOf((Class<? extends Enum>) hiveVersionInIsolated, version.name());
     Constructor<?> hiveClientImplCtor =
-        hiveClientImplClass.getConstructor(
-            HiveClientClassLoader.HiveVersion.class, Properties.class);
-    return (HiveClient) hiveClientImplCtor.newInstance(version, properties);
```
diqiu50, line 168: `HiveClientClassLoader.HiveVersion.class` is shared class.  We can pass to the `classloader` directly

catalogs/hive-metastore-common/src/test/java/org/apache/gravitino/hive/client/TestHiveClientFactoryClassLoader.java
```diff
@@ -0,0 +1,208 @@
...
+ * </ul>
+ *
+ * <p>Bug 3 (missing {@code doAs} in non-impersonation Kerberos path) requires a live KDC and is
+ * covered by the integration test {@link TestHive2HMSWithKerberosNoImpersonation}.
+ */
+public class TestHiveClientFactoryClassLoader {
+
+  // ---------------------------------------------------------------------------
+  // Bug 1: factory CL is stable; TCCL is not — contrast/rationale tests
+  // ---------------------------------------------------------------------------
```
diqiu50: Tests for TCCL are not necessary.

catalogs/hive-metastore-common/src/test/java/org/apache/gravitino/hive/client/TestHiveClientFactoryClassLoader.java
```diff
@@ -0,0 +1,208 @@
...
+   * NoSuchMethodException}), proving that the {@code HiveVersion} lookup on the isolated CL
+   * succeeded before reaching the constructor step.
+   *
+   * <p>If the bug were present, {@code getConstructor} would receive the system CL's {@code
+   * HiveVersion} and throw {@link NoSuchMethodException} instead.
+   */
+  @Test
+  @SuppressWarnings({"unchecked", "rawtypes"})
+  public void testCreateHiveClientImplLoadsHiveVersionFromIsolatedClassLoader() throws Exception {
+    ClassLoader mockCl = mock(ClassLoader.class);
```
diqiu50: HiveVersion is a shared class. This is not a problem

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/client/HiveClientFactory.java
```diff
@@ -202,6 +212,17 @@ private HiveClient createHiveClientInternal(HiveClientClassLoader classloader) {
...
+      } else if (enableKerberos) {
+        // UGI is a shared class (org.apache.hadoop.* delegated to baseLoader), so the system CL
+        // and HiveClientClassLoader share the same UGI static state. The TGT is already stored in
+        // realLoginUgi.subject by kerberosClient.login(). The only thing needed is to bind that
+        // Subject to the current thread so GSSAPI can find the TGT during the HMS Thrift handshake.
+        // UGI.doAs() wraps Subject.doAs() internally — same pattern as ImpalaEngineAdapter.
+        UserGroupInformation realUgi = kerberosClient.getRealLoginUgi();
+        final HiveClientClassLoader.HiveVersion hiveVersion = classloader.getHiveVersion();
+        return realUgi.doAs(
+            (java.security.PrivilegedExceptionAction<HiveClient>)
```
diqiu50: FQN

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/client/HiveClientFactory.java
JoegenUSTC: You are absolutely right, thank you for the correction!

I re-read `HiveClientClassLoader.isSharedClass()` carefully and found the rule:

    if (name.startsWith("org.apache.gravitino.")) { return true; }

`HiveVersion` is `org.apache.gravitino.hive.client.HiveClientClassLoader$HiveVersion`, 
which falls under this rule. The isolated classloader delegates it to the base 
classloader, so both sides hold the same `Class` object. `getConstructor(HiveVersion.class, 
Properties.class)` works correctly without any workaround.

I've simplified `createHiveClientImpl()` to pass `HiveClientClassLoader.HiveVersion.class` 
directly and removed the unnecessary `Enum.valueOf` reflection code.

catalogs/hive-metastore-common/src/test/java/org/apache/gravitino/hive/client/TestHiveClientFactoryClassLoader.java
JoegenUSTC: Agreed, removed.

`testFactoryClassLoaderIsStableAcrossThreads` and `testTCCLIsUnstableAcrossThreads` were 
testing Java language guarantees (a class's ClassLoader never changes; TCCL can be set per 
thread) rather than any behaviour specific to our code. They don't add regression value 
and have been deleted.

catalogs/hive-metastore-common/src/test/java/org/apache/gravitino/hive/client/TestHiveClientFactoryClassLoader.java
JoegenUSTC: Correct, the Bug 2 fix was based on a wrong assumption. Since `HiveVersion` is a shared 
class (covered by the `org.apache.gravitino.*` rule in `isSharedClass()`), the isolated 
classloader never redefines it — both the factory and the isolated classloader see the 
same `Class` object.

The corresponding tests (`testCreateHiveClientImplLoadsHiveVersionFromIsolatedClassLoader` 
and `testSystemClHiveVersionCausesNoSuchMethodException`) have been removed along with the 
production code workaround.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/client/HiveClientFactory.java
JoegenUSTC: Fixed. Added `import java.security.PrivilegedExceptionAction;` and replaced the inline 
FQN cast per the project coding standards.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/client/HiveClientFactory.java
```diff
@@ -202,6 +212,17 @@ private HiveClient createHiveClientInternal(HiveClientClassLoader classloader) {
...
+                () -> createHiveClientImpl(hiveVersion, properties, classloader));
```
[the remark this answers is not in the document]
JoegenUSTC: Good catch. Added `Preconditions.checkState(realLoginUgi != null, 
"KerberosClient.login() has not been called")` inside `getRealLoginUgi()` itself 
(following the suggested changeset) so callers fail fast with a clear error message 
rather than an NPE. Updated the Javadoc to document the `@throws IllegalStateException` 
contract accordingly.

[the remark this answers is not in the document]
JoegenUSTC: Fixed, same as the FQN comment above — added the import and removed the inline FQN.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/KerberosClient.java
```diff
@@ -198,4 +198,15 @@ public void close() {
...
+  /**
+   * Returns the real (non-proxy) {@link UserGroupInformation} that was obtained after Kerberos
+   * login via keytab. Used by callers that need to bind the JAAS Subject to the current thread
+   * (e.g., via {@code ugi.doAs(...)}) before performing a Kerberos-protected RPC call.
+   *
+   * @return the real login UGI, or {@code null} if {@link #login()} has not been called yet.
+   */
+  public UserGroupInformation getRealLoginUgi() {
+    return realLoginUgi;
+  }
```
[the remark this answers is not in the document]
JoegenUSTC: Implemented as suggested. The check is placed inside `getRealLoginUgi()` rather than 
at the call site in `HiveClientFactory`, which makes the contract clearer and avoids 
having every caller guard against null independently.

diqiu50 (commented):
LGTM , @yuqi1129 Do you have any ideas

JoegenUSTC:
@diqiu50 Thanks for the thorough review! I've addressed all the feedback in the latest 
commit (`4502c3a`). Here's a summary of what was changed:

**Simplified `createHiveClientImpl()` (re: "HiveVersion is a shared class")**  
Removed the unnecessary `Enum.valueOf` reflection workaround. You are correct — 
`HiveClientClassLoader.isSharedClass()` has an explicit rule covering all 
`org.apache.gravitino.*` classes, so `HiveVersion` is always delegated to the base 
classloader. Both sides hold the same `Class` object and we can pass 
`HiveVersion.class` directly to `getConstructor()`.

**Removed `TestHiveClientFactoryClassLoader` (re: "Tests for TCCL are not necessary" and "HiveVersion is a shared class. This is not a problem")**  
Deleted the entire test file. The two TCCL tests were verifying Java language 
guarantees rather than our code's behaviour, and the two Bug 2 tests were built 
on the incorrect assumption that `HiveVersion` is not a shared class.

**Replaced FQN with a proper import in `HiveClientFactory.java` (re: "FQN")**  
Added `import java.security.PrivilegedExceptionAction;` at the top of the file 
per the project coding standards.

**Strengthened `KerberosClient.getRealLoginUgi()` contract (Copilot suggestion)**  
Added `Preconditions.checkState(realLoginUgi != null, ...)` inside the method so 
callers fail fast with a clear error message if `login()` was not called. Updated 
the Javadoc to document the `@throws IllegalStateException` contract.

All changes compile cleanly and the unit tests (`hive-metastore-common:test`) and 
spotless format check pass locally. Please take another look when you have a chance!

yuqi1129:
> LGTM , @yuqi1129 Do you have any ideas

It also LGTM.

yuqi1129 (approved):

JoegenUSTC:
@yuqi1129 CI is green now, could you merge this when you have a chance? Thanks!

yuqi1129:
@JoegenUSTC Merge it, thanks for your hard work.

PR #11359, opened 2026-06-02
[#11352] fix(glue): improve error logging for Iceberg metadata loading

### LINKED ISSUE #11352: [Bug report] Glue catalog logs a misleading warning when Iceberg metadata file is missing
### Version

main branch

### Describe what's wrong

When loading an Iceberg table from the Glue catalog, Gravitino logs a warning saying that partitioning and sort order information may be incomplete.

However, the actual root cause is that the Iceberg metadata JSON file referenced by the Glue table does not exist in S3. The current warning is misleading because it makes the issue look like a partition/sort recovery problem, while the table metadata file itself is missing.

```text
WARN GlueCatalogOperations.loadTable - Failed to load Iceberg metadata for table <database>.<table>. Partitioning and sort order information may be incomplete.

org.apache.iceberg.exceptions.NotFoundException: Location does not exist:
s3://<bucket>/<table>/metadata/<version>.metadata.json

Caused by: software.amazon.awssdk.services.s3.model.NoSuchKeyException:
The specified key does not exist. (Service: S3, Status Code: 404)
```

### How to reproduce

1. Configure a Gravitino Glue catalog with an Iceberg table.
2. Make the Glue table reference an Iceberg metadata location whose S3 object no longer exists.
3. Load the table through Gravitino, or trigger a path that calls `GlueCatalogOperations.loadTable`.
4. Check the server log.

Expected behavior: the log should clearly indicate that the Iceberg metadata file is missing or cannot be loaded.

Actual behavior: the log says partitioning and sort order information may be incomplete, which hides the real cause.

### Additional context

The warning is emitted from `GlueCatalogOperations.loadTable` after `GlueIcebergTableHelper.loadTable` fails. The catch block handles all exceptions with the same partition/sort warning, so missing metadata files are reported with an inaccurate message.

xxubai:

This PR updates the Glue catalog warning when loading Iceberg metadata fails.

The warning no longer says that partitioning and sort order information may be incomplete. Instead, it reports the Iceberg metadata load failure directly and keeps the original exception in the stacktrace, so missing `metadata.json` files are easier to diagnose.

This PR also adds a regression test to ensure that a table returned after Iceberg metadata loading fails still keeps its Glue operation context.

The previous warning was misleading when the real root cause was a missing Iceberg metadata file, for example an S3 `NoSuchKeyException` for `metadata/*.metadata.json`. It made the issue look like a partition or sort-order compatibility problem.

Fix: #11352

Yes. The Glue catalog warning message for Iceberg metadata loading failures is changed.

No API or configuration changes are introduced.

- `./gradlew :catalogs:catalog-glue:test --tests org.apache.gravitino.catalog.glue.TestGlueCatalogOperationsForIceberg -PskipITs`
- `./gradlew :catalogs:catalog-glue:test -PskipITs`

### DIFF
--- a/catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
+++ b/catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
@@ -209,7 +209,7 @@ class TestGlueCatalogOperationsForIceberg {
   }
 
   @Test
-  void testLoadTable_icebergMetadataLoadFailureKeepsGlueOpsContext() {
+  void testLoadTableWithIcebergMetadataLoadFailure() {
     software.amazon.awssdk.services.glue.model.Table rawTable =
         software.amazon.awssdk.services.glue.model.Table.builder()
             .name(TABLE)
--- a/catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
+++ b/catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
@@ -20,6 +20,7 @@ package org.apache.gravitino.catalog.glue;
 
 import static org.apache.gravitino.catalog.glue.GlueConstants.ICEBERG_TABLE_TYPE_VALUE;
 import static org.apache.gravitino.catalog.glue.GlueConstants.TABLE_TYPE_PARAM;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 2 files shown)


catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
```diff
@@ -203,6 +208,33 @@ void testCreateTable_icebergMissingLocationThrows() {
                 Indexes.EMPTY_INDEXES));
   }
 
+  @Test
+  void testLoadTable_icebergMetadataLoadFailureKeepsGlueOpsContext() {
```
diqiu50: The function name is no match the code style

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
```diff
@@ -203,6 +208,33 @@ void testCreateTable_icebergMissingLocationThrows() {
...
+    software.amazon.awssdk.services.glue.model.Table rawTable =
```
diqiu50, line 213: FQN problems

xxubai: Already import Iceberg table class: `import org.apache.iceberg.Table`

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueCatalogOperationsForIceberg.java
xxubai: The other test functions use the same naming convention. But I update this unit test funtion name

diqiu50 (approved):
LGTM

xxubai:
Thanks @jerryshao @diqiu50

PR #11361, opened 2026-06-02
[#11360] fix(server): Ignore quote-only custom filter entries

### LINKED ISSUE #11360: [Bug report] Quote-only customFilters value breaks Jetty WebAppContext startup
### Version

main branch

### Describe what's wrong

When `gravitino.server.webserver.customFilters` is configured as a quote-only value such as `""`, Gravitino treats it as a custom servlet filter class name.

Jetty then tries to load a filter class named `""`, which fails during WebAppContext startup. The server connector may still start and log that Gravitino has started, but the web context is marked `UNAVAILABLE`.

```text
java.lang.ClassNotFoundException: ""
Failed startup of context ... gravitino-web ... UNAVAILABLE
javax.servlet.UnavailableException: Class loading error for holder ""...
```

### How to reproduce

1. Use the main branch.
2. Configure Gravitino with:

```properties
gravitino.server.webserver.customFilters = ""
```

3. Start the Gravitino server.
4. Observe that Jetty fails to start the `gravitino-web` WebAppContext because it attempts to load `""` as a filter class.

### Additional context

`JettyServerConfig` currently splits `customFilters` directly and preserves quote-only entries. The server should defensively ignore blank and quote-only filter entries before registering custom filters.

xxubai:

This PR hardens the parsing of `gravitino.server.webserver.customFilters`.

The parsing now trims comma-separated entries and ignores invalid empty entries, including blank values, `""`, and `''`, before exposing the custom filter set to Jetty.

It also extends `TestJettyServerConfig` to cover quote-only and mixed custom filter values.

If `gravitino.server.webserver.customFilters` is rendered or configured as `""`, Gravitino currently preserves that value as a filter class name. Jetty then attempts to load a class named `""`, causing the WebAppContext startup to fail with `ClassNotFoundException`.

The server should tolerate this malformed empty configuration and treat it as no custom filter.

Fix: #11360

No new configuration keys or APIs are introduced.

This only makes the existing `gravitino.server.webserver.customFilters` parsing more tolerant by ignoring quote-only empty values.

```text
./gradlew spotlessApply
./gradlew :server-common:test --tests org.apache.gravitino.server.web.TestJettyServerConfig -PskipITs -PskipDockerTests=false
git diff --check
```


server-common/src/main/java/org/apache/gravitino/server/web/JettyServerConfig.java
```diff
@@ -377,7 +379,15 @@ private JettyServerConfig(Map<String, String> configs) {
     this.customFilters =
         internalConfig
             .get(CUSTOM_FILTERS)
```
jerryshao: You can update the `CUSTOM_FILTERS` to use `toSequence()`, which will convert the string conf to list conf

xxubai: Updated

jerryshao (approved):

PR #11392, opened 2026-06-03
[#10464] improvement(iceberg): Decouple listener events from Iceberg APIs

### LINKED ISSUE #10464: [Improvement] Free Iceberg related listener from Iceberg dependencies

Iceberg releated classes like `IcebergCreateNamespaceEvent` can be relocated to another module like `iceberg-listener`. At the same time, we need to remove using field `createNamespaceRequest` and `createNamespaceRequest` to get rid of iceberg API dependencies. 

### How should we improve?

_No response_

yuqi1129:

- Add a new `iceberg-listener` module for Iceberg listener event APIs.
- Move Iceberg listener events out of `iceberg-rest-server`.
- Replace Iceberg REST request/response objects in listener events with lightweight JSON payloads.
- Keep Iceberg request/response serialization inside `iceberg-rest-server`.
- Add tests to guard the listener module from Iceberg/Servlet dependencies.

Iceberg listener events are expected to be consumed by modules like search, but the current event classes live in `iceberg-rest-server` and expose Iceberg REST API classes. This introduces unnecessary dependencies and can cause class conflicts.

Fix: #10464

No end-user behavior change. This changes the developer-facing Iceberg listener event payload API by replacing Iceberg REST objects with lightweight payload wrappers.

- `./gradlew :iceberg:iceberg-listener:test :iceberg:iceberg-rest-server:test -PskipITs`
- `git diff --check`
- Verified `iceberg-listener/src/main/java` has no `org.apache.iceberg`, `org.apache.gravitino.iceberg`, or `javax.servlet` imports.


yuqi1129:
Close it.

PR #11404, opened 2026-06-03
[#11403] improvement(authz): use per-request cache for group owner check in JcasbinAuthorizer

### LINKED ISSUE #11403: [Improvement] JcasbinAuthorizer group owner check bypasses per-request cache causing redundant DB queries

In `JcasbinAuthorizer.ownerMatchesUserOrGroups()`, when the owner type is `GROUP`, the method calls `resolveCurrentUserGroups(metalake, entityStore)` which directly invokes `entityStore.batchGet(groupIdents, GROUP, GroupEntity.class)`. This call is not protected by either the per-request cache (`requestContext.groupInfoCache`) or the shared Caffeine cache (`groupRoleCache`), so every GROUP owner check issues a DB round trip regardless of how warm the caches are.

The problem is amplified in `hasSetOwnerPermission`, which walks the full parent-chain in a loop calling `isOwner` repeatedly — meaning multiple DB queries for group entities can fire within a single request when the owner is a group.

The method also loads full `GroupEntity` objects, but only uses `groupEntity.id()` for comparison, making it heavier than necessary.

### How should we improve?

Replace the `resolveCurrentUserGroups()` call with `currentPrincipalGroupNames()` + `loadGroupInfo()`, which routes through `requestContext.groupInfoCache` (per-request dedup) and only queries `group_meta.group_id / updated_at` when the cache is cold. Remove the now-unused `resolveCurrentUserGroups()` method.

yuqi1129:

Replace `resolveCurrentUserGroups(metalake, entityStore)` in `ownerMatchesUserOrGroups()` with `currentPrincipalGroupNames()` + `loadGroupInfo()`. The new path routes through `requestContext.groupInfoCache` (per-request dedup) before falling back to a lightweight `group_meta` query via `GroupMetaMapper#getGroupUpdatedAt`, and removes the need to load full `GroupEntity` objects. The now-unused `resolveCurrentUserGroups()` method and its `GroupEntity` import are removed.

The old path bypassed all caches and issued an `entityStore.batchGet(GROUP)` DB call on every GROUP owner check. In `hasSetOwnerPermission`, which walks the full parent chain calling `isOwner` in a loop, this could fire multiple times per request when the owner is a group. The fix ensures group identity is resolved through the same version-validated cache path as all other identity lookups in the authorizer.

Fix: #11403

No.

- Updated `testAuthorizeByGroupOwner` to mock `groupMetaMapper.getGroupUpdatedAt` instead of `entityStore.batchGet(GROUP)`, and added a `verify` assertion that `entityStore.batchGet(GROUP)` is never called.
- Added `testGroupOwnerCheckDeduplicatesGroupInfoWithinRequest` which calls `isOwner` twice within the same `AuthorizationRequestContext` and asserts that `groupMetaMapper.getGroupUpdatedAt` is invoked exactly once (per-request cache dedup).

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -890,8 +890,8 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
     if (!Entity.EntityType.GROUP.name().equalsIgnoreCase(ownerInfo.getOwnerType())) {
       return false;
     }
-    for (String groupname : currentPrincipalGroupNames()) {
-      Optional<GroupUpdatedAt> groupInfo = loadGroupInfo(metalake, groupname, requestContext);
+    for (String groupName : currentPrincipalGroupNames()) {
+      Optional<GroupUpdatedAt> groupInfo = loadGroupInfo(metalake, groupName, requestContext);
       if (groupInfo.isPresent() && groupInfo.get().getGroupId() == ownerInfo.getOwnerId()) {
         return true;
       }
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -1045,7 +1045,7 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
     if (groups.isEmpty()) {
       return new ArrayList<>();
     }
-    return groups.stream().map(UserGroup::getGroupname).collect(Collectors.toList());
+    return groups.stream().map(UserGroup::getGroupName).collect(Collectors.toList());
   }
 
   private void versionCheckAndLoadRoles(
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
@@ -55,7 +55,6 @@ import org.apache.gravitino.authorization.Privilege;
 import org.apache.gravitino.authorization.SecurableObject;
 import org.apache.gravitino.cache.CaffeineGravitinoCache;
 import org.apache.gravitino.cache.GravitinoCache;
-import org.apache.gravitino.meta.GroupEntity;
 import org.apache.gravitino.meta.RoleEntity;
 import org.apache.gravitino.server.authorization.MetadataIdConverter;
 import org.apache.gravitino.storage.relational.SupportsEntityChangeLog;
@@ -871,6 +870,8 @@ public class JcasbinAuthorizer implements GravitinoAuthorizer {
    * Returns true when the cached owner type and ID match the given principal or one of the
    * principal's groups. The user id is resolved via the version-validated {@link #loadUserInfo}
    * cache so back-to-back ownership checks in the same request do not re-query {@code user_meta}.
+   * Group ids are resolved via {@link #loadGroupInfo}, which deduplicates within the request via
+   * {@code requestContext.groupInfoCache} and avoids loading full group entity objects.
    */
   private boolean ownerMatchesUserOrGroups(
       Optional<OwnerInfo> owner,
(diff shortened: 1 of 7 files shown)


server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -889,9 +890,9 @@ private boolean ownerMatchesUserOrGroups(
...
       return false;
     }
-    EntityStore entityStore = GravitinoEnv.getInstance().entityStore();
-    for (GroupEntity groupEntity : resolveCurrentUserGroups(metalake, entityStore)) {
-      if (Objects.equals(groupEntity.id(), ownerInfo.getOwnerId())) {
+    for (String groupname : currentPrincipalGroupNames()) {
+      Optional<GroupUpdatedAt> groupInfo = loadGroupInfo(metalake, groupname, requestContext);
+      if (groupInfo.isPresent() && groupInfo.get().getGroupId() == ownerInfo.getOwnerId()) {
         return true;
       }
```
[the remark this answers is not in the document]
yuqi1129: Fixed. GROUP owner checks now derive group names from the provided `principal` via `principalGroupNames(principal)`, while the existing current-principal helper remains for the normal authorization flow. Added `testGroupOwnerCheckUsesProvidedPrincipalGroups` to cover the explicit-principal case.

yuqi1129:
@roryqi 
Can you help to review this?

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -889,9 +890,9 @@ private boolean ownerMatchesUserOrGroups(
     if (!Entity.EntityType.GROUP.name().equalsIgnoreCase(ownerInfo.getOwnerType())) {
       return false;
     }
-    EntityStore entityStore = GravitinoEnv.getInstance().entityStore();
-    for (GroupEntity groupEntity : resolveCurrentUserGroups(metalake, entityStore)) {
-      if (Objects.equals(groupEntity.id(), ownerInfo.getOwnerId())) {
+    for (String groupname : principalGroupNames(principal)) {
```
jerryshao: `groupName`,

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -1033,36 +1034,18 @@ private List<Long> loadGroupRoles(
...
     List<UserGroup> groups = ((UserPrincipal) principal).getGroups();
     if (groups.isEmpty()) {
       return new ArrayList<>();
     }
-    List<NameIdentifier> groupIdents =
-        groups.stream()
-            .map(g -> NameIdentifierUtil.ofGroup(metalake, g.getGroupname()))
-            .collect(Collectors.toList());
-    return entityStore.batchGet(groupIdents, Entity.EntityType.GROUP, GroupEntity.class);
+    return groups.stream().map(UserGroup::getGroupname).collect(Collectors.toList());
```
jerryshao: Shall we change to `getGroupName`?

yuqi1129: This is a previous naming problem introduced by others, and I will change it, by the way.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -889,9 +890,9 @@ private boolean ownerMatchesUserOrGroups(
     if (!Entity.EntityType.GROUP.name().equalsIgnoreCase(ownerInfo.getOwnerType())) {
       return false;
     }
-    EntityStore entityStore = GravitinoEnv.getInstance().entityStore();
-    for (GroupEntity groupEntity : resolveCurrentUserGroups(metalake, entityStore)) {
-      if (Objects.equals(groupEntity.id(), ownerInfo.getOwnerId())) {
+    for (String groupName : principalGroupNames(principal)) {
+      Optional<GroupUpdatedAt> groupInfo = loadGroupInfo(metalake, groupName, requestContext);
```
roryqi, line 894: Do we need to batch get groups for the first operation?

yuqi1129: We have used the method `prefetchUserAndGroupInfo` to batch fetch groups, so we can get the information from cache.

roryqi: Ok for me.

roryqi (approved):
LGTM.

PR #11417, opened 2026-06-04
  [#11401] fix(trino-connector): drop_catalog falls back to server when local cache misses

### LINKED ISSUE #11401: [Bug report]  drop_catalog fails with "not exists" when catalog exists on Gravitino server
### Version

main branch

### Describe what's wrong

drop_catalog fails with "not exists" when catalog exists on Gravitino server but not in Trino connector's local cache

`CALL gravitino.system.drop_catalog('gt_glue_xxx1')` fails with "Catalog trino_connector_metalake.gt_glue_xxx1 not exists", but immediately calling `create_catalog` for the same catalog fails with "already exists". The catalog is stuck in an inconsistent state where it cannot be dropped or re-created.

The root cause is in `DropCatalogStoredProcedure.dropCatalog()`: it checks catalog existence via `catalogConnectorManager.getCatalogConnector()`, which only queries the Trino connector's **local in-memory cache** (the `catalogConnectors` map). If the catalog exists on the Gravitino server but was never successfully loaded into the connector (e.g., connector creation failed, cache was evicted, or Trino was restarted), the drop incorrectly reports "not exists" without ever consulting the Gravitino server.

The correct behavior should be: when the local connector is not found, fall back to querying the Gravitino server to determine whether the catalog actually exists, and if so, proceed with the server-side drop.

```
trino> CALL gravitino.system.drop_catalog('gt_glue_xxx1');
Query 20260602_164118_00410_q78qz failed: Drop catalog failed. Catalog trino_connector_metalake.gt_glue_xxx1 not exists.

io.trino.spi.TrinoException: Drop catalog failed. Catalog trino_connector_metalake.gt_glue_xxx1 not exists.
	at org.apache.gravitino.trino.connector.system.storedprocedure.DropCatalogStoredProcedure.dropCatalog(DropCatalogStoredProcedure.java:123)
	at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:735)
	at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:760)
	at io.trino.execution.CallTask.execute(CallTask.java:210)
	...
Caused by: io.trino.spi.TrinoException: Catalog trino_connector_metalake.gt_glue_xxx1 not exists.
	at org.apache.gravitino.trino.connector.system.storedprocedure.DropCatalogStoredProcedure.dropCatalog(DropCatalogStoredProcedure.java:99)
	... 15 more
```

Immediately after:
```
trino> CALL gravitino.system.create_catalog('gt_glue_xxx1', 'glue', ...);
-- fails with: Catalog trino_connector_metalake.gt_glue_xxx1 already exists in the server.
```

### How to reproduce

1. Deploy Gravitino + Trino with the Gravitino Trino connector (Trino 478).
2. Create a catalog `gt_glue_xxx1` via `create_catalog` stored procedure with properties that will cause connector instantiation to fail (e.g., an invalid placeholder like `${?aws_glue_endpoint}` for the endpoint URL). The catalog metadata is persisted to the Gravitino server, but the Trino connector fails to load.
3. The background scheduler (`gravitino-connector-schedule-0`) also attempts to load the catalog and fails — so no `CatalogConnectorContext` is ever stored in the local cache.
4. Call `CALL gravitino.system.drop_catalog('gt_glue_xxx1')` → fails with "not exists" because `getCatalogConnector()` returns null.
5. Call `CALL gravitino.system.create_catalog('gt_glue_xxx1', ...)` → fails with "already exists" because the Gravitino server still holds the metadata.
6. The catalog is now in a zombie state: cannot be dropped, cannot be re-created.

### Additional context

The problematic code is in `DropCatalogStoredProcedure.java` lines 91-99:

```java
CatalogConnectorContext catalogConnector =
    catalogConnectorManager.getCatalogConnector(
        catalogConnectorManager.getTrinoCatalogName(metalake, catalogName));
if (catalogConnector == null) {
    if (ignoreNotExist) {
        return;
    }
    throw new TrinoException(
}
```

**Suggested fix**: When `getCatalogConnector()` returns null, query the Gravitino server directly (via the Gravitino client API) to check if the catalog exists. If it does, issue the server-side drop. Only throw "not exists" if the catalog is absent from both the local cache and the server.

**Workaround**: Use `CALL gravitino.system.drop_catalog('gt_glue_xxx1', true)` (with `ignoreNotExist=true`) to skip the local cache check, though this may not actually remove the catalog from the Gravitino server either.

lhjchn:

  Previously `DropCatalogStoredProcedure` only checked the Trino connector's local cache (the `catalogConnectors` map) to decide whether a catalog exists. When the local connector was null, it directly threw "not exists"
  without checking the Gravitino server.

  This PR changes the null branch to call `metalake.dropCatalog(catalogName, true)` on the server:

  1. If the server returns true, the catalog is dropped on the server successfully.
  2. If the server returns false, the catalog does not exist on the server either. We then respect the `ignore_not_exist` flag: return silently if true, throw `GRAVITINO_CATALOG_NOT_EXISTS` if false.

  The branch where the local connector is found stays the same as before.

  Also adds a unit test class `TestDropCatalogStoredProcedure` for the new code path and updates the Javadoc.
  
  Fix: #11401
  
  The local cache may not have the catalog even though it exists on the Gravitino server. This can happen when:

  1. The catalog is created with invalid properties and the connector fails to load.
  2. The cache entry is evicted.
  3. Trino is restarted before the background scheduler reloads it.

  In this state the catalog cannot be dropped (local check returns null, `drop_catalog` throws "not exists") and cannot be recreated either (the server still holds the metadata, `create_catalog` throws "already exists"). The
  user has no way to recover without restarting Trino or calling the server-side API directly.

  By falling back to the server-side drop when the local cache misses, the user can clean up the catalog with the same `drop_catalog` procedure.

  No new API, configuration, or property key.
  
  The behavior of `drop_catalog(catalog, ignore_not_exist)` changes only for the bug case: calls that previously failed with "not exists" while the catalog actually existed on the server now succeed. The other cases (catalog
  truly does not exist anywhere; catalog exists in the local cache) behave the same as before.

  Added a new unit test class `TestDropCatalogStoredProcedure` with three cases:

  1. The local cache does not contain the catalog but the server does. The procedure should call the server-side drop and complete without error.
  2. The catalog is absent from both the local cache and the server, with `ignoreNotExist=false`. The procedure should throw `GRAVITINO_CATALOG_NOT_EXISTS`.
  3. Same as case 2 but with `ignoreNotExist=true`. The procedure should return silently and must not trigger a local cache reload.

  Ran `./gradlew :trino-connector:trino-connector:test -PskipITs` and `./gradlew :trino-connector:trino-connector:spotlessApply`, both pass.


jerryshao:
@yuqi1129 please help to review this PR, thanks.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/system/storedprocedure/DropCatalogStoredProcedure.java
```diff
@@ -79,25 +79,48 @@ public Procedure createStoredProcedure() throws NoSuchMethodException, IllegalAc
...
           catalogConnectorManager.getCatalogConnector(
               catalogConnectorManager.getTrinoCatalogName(metalake, catalogName));
+
+      // The local in-memory cache may not have an entry even if the catalog exists on
+      // the Gravitino server (e.g. the connector failed to load at startup, or the cache was
+      // evicted). In that case the catalog would otherwise be stuck in a zombie state:
+      // it cannot be dropped here because the local check would say "not exists", and it
+      // cannot be re-created either because the server still holds the metadata.
+      // Fall back to the server-side drop so the server is the source of truth.
       if (catalogConnector == null) {
```
yuqi1129, line 94: 1. The catalogs cache implementation is a map, see: https://github.com/apache/gravitino/blob/c4259fccdc35aae856a47100357a68a6b7d4cb80/trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorManager.java#L71-L72,  I can't get why it will be evicted.
2. `the connector failed to load at startup`, I'm afraid that we can't see the catalog as it fails to be loaded at startup, so can we remove it without being aware of its existence?

lhjchn: Thanks for your review @yuqi1129 
  1. You're right, I'll remove the "cache was evicted" wording.
  2. The user sees the catalog via CLI/REST/UI, just not from Trino. create_catalog with bad props succeeds on server but fails to load locally, so drop_catalog says "not exists" while create_catalog says "already exists".

lhjchn: I noticed PR #11432 covers the same fix with a more concise change.Would it make sense to keep this PR instead? It already includes unit tests for the fallback path and updated Javadoc. I'm happy to simplify the main code to match the shorter style in #11432 if preferred.

yuqi1129: You can go forward with your PR as you raise it ahead of #11432

diqiu50: I’ve closed #11432. You can go ahead with your PR.

lhjchn: @yuqi1129 Hi, do you think the conditional logic should be simplified further?

yuqi1129: The current change is okay for me, and let us merge it.

yuqi1129 (approved):

yuqi1129:
@lhjchn 
Can you also raise a PR to fix the problem in branch-1.3? I forgot to add the tag `branch-1.3`, which will be used to cherry-pick to `branch-1.3`, thanks.

lhjchn:
@yuqi1129 Sure, I'll create a PR for branch-1.3 .

Issue #11402, opened 2026-06-03
[Bug report] create_catalog leaves zombie catalog on Gravitino server when Trino connector creation fails (no rollback)


danhuawang:
### Version

main branch

### Describe what's wrong

When `CALL gravitino.system.create_catalog(...)` is executed, the catalog metadata is first persisted to the Gravitino server successfully, then the Trino connector attempts to instantiate the catalog locally. If the connector creation fails (e.g., due to invalid configuration properties), the catalog metadata remains on the Gravitino server with **no rollback**. This leaves the catalog in a "zombie" state:

- The catalog **exists** on the Gravitino server (so `create_catalog` reports "already exists" on retry).
- The catalog **does not exist** in the Trino connector's local cache (so `drop_catalog` reports "not exists" — see related Bug 1).
- The background scheduler repeatedly tries and fails to load the catalog, logging errors every cycle.

The correct behavior should be: if the Trino connector fails to instantiate the catalog, the system should either roll back the server-side metadata or provide a clear mechanism for the user to clean up the failed catalog.

### Error message and/or stacktrace

Initial `create_catalog` call (or background scheduler attempting to load the catalog):

```
2026-06-03 07:44:34 ERROR [Query-20260603_074434_00288_rkud3-393] CatalogConnectorManager:428 - Failed to create connector: gt_glue_xxx1
io.trino.spi.TrinoException: Failed to create connector lakehouse
  at org.apache.gravitino.trino.connector.GravitinoConnectorPluginManager.createConnector(GravitinoConnectorPluginManager.java:299)
  at org.apache.gravitino.trino.connector.catalog.CatalogConnectorContext$Builder.build(CatalogConnectorContext.java:273)
  at org.apache.gravitino.trino.connector.catalog.CatalogConnectorManager.createCatalogConnectorContext(CatalogConnectorManager.java:422)
  ...
Caused by: io.airlift.bootstrap.ApplicationConfigurationException: Configuration errors:
1) Error: Invalid value '${?aws_glue_endpoint}' for type URI (property 'hive.metastore.glue.endpoint-url')
   in order to call [public GlueHiveMetastoreConfig GlueHiveMetastoreConfig.setGlueEndpointUrl(URI)]
```

Background scheduler continuously retrying and failing:

```
2026-06-03 07:44:34 ERROR [gravitino-connector-schedule-0] CatalogRegister:159 - Failed to register catalog gt_glue_xxx1
[...]

PR #11432, opened 2026-06-04
[#11401] fix(trino-connector): drop_catalog falls back to server when local cache misses

### LINKED ISSUE #11401: [Bug report]  drop_catalog fails with "not exists" when catalog exists on Gravitino server

diqiu50:

When `drop_catalog` is called for a catalog that exists on the Gravitino server
but not in the Trino connector's local cache (zombie catalog), fall back to a
direct server-side drop instead of reporting "not exists". The `ignoreNotExist`
flag behavior is preserved.

`drop_catalog` only checked the local connector cache. If a catalog was persisted
to the Gravitino server but the connector failed to instantiate it, the catalog
was stuck — it could neither be dropped nor re-created.

Fix: #11401 #11402

`drop_catalog` can now clean up zombie catalogs that exist on the server but are
absent from the local connector cache.

Manual verification against the scenario described in #11401.

### DIFF
--- a/trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/system/storedprocedure/DropCatalogStoredProcedure.java
+++ b/trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/system/storedprocedure/DropCatalogStoredProcedure.java
@@ -90,13 +90,16 @@ public class DropCatalogStoredProcedure extends GravitinoStoredProcedure {
           catalogConnectorManager.getCatalogConnector(
               catalogConnectorManager.getTrinoCatalogName(metalake, catalogName));
       if (catalogConnector == null) {
-        if (ignoreNotExist) {
-          return;
+        // Local cache miss — the catalog may still exist on the Gravitino server (e.g., connector
+        // failed to load after creation). Fall back to a direct server-side drop so zombie catalogs
+        // can be cleaned up.
+        if (!catalogConnectorManager.getMetalake(metalake).dropCatalog(catalogName, true)
+            && !ignoreNotExist) {
+          throw new TrinoException(
+              GravitinoErrorCode.GRAVITINO_CATALOG_NOT_EXISTS,
+              "Catalog " + NameIdentifier.of(metalake, catalogName) + " not exists.");
         }
-
-        throw new TrinoException(
-            GravitinoErrorCode.GRAVITINO_CATALOG_NOT_EXISTS,
-            "Catalog " + NameIdentifier.of(metalake, catalogName) + " not exists.");
+        return;
       }
       // Always ignore "ignoreNotExist" inside Metalake.dropCatalog()
       // because we already handled the null check above.


yuqi1129:
There already exists a fix PR: https://github.com/apache/gravitino/pull/11417

PR #11461, opened 2026-06-05
[#11122 ][#11112] docs(upgrade): Improve PostgreSQL backup format and add rollback instructions

### LINKED ISSUE #11112: [Improvement] how-to-upgrade.md: Step 3 dump format is binary and cannot be diffed in Step 4

In Step 3, the PostgreSQL dump command uses -F c (custom binary format):
`
pg_dump -U username -h hostname -d database_name -n schema_name -s -F c -f schema-x.y.z-postgresql.sql
`
However, Step 4 asks users to compare the dumped schema against the official schema-x.y.z-postgresql.sql file in the repository. This comparison is impossible because -F c produces a binary file, not plain SQL text. Running diff on it yields:
`
Binary files schema-x.y.z-postgresql_dump.sql and schema-x.y.z-postgresql.sql differ
`
Additionally, Step 4 says "compare the contents" but gives no concrete instructions on how to actually perform the comparison — no command, no tool, no guidance on what differences are acceptable vs. problematic.

### How should we improve?

1. In Step 3, remove -F c from the PostgreSQL dump command so it outputs plain SQL text:
`
pg_dump -U username -h hostname -d database_name -n schema_name -s -f schema-x.y.z-postgresql.sql
`
2. In Step 4, add concrete comparison instruction

### LINKED ISSUE #11122: [Improvement] Add doc: how to rollback with the backup file in how-to-upgrade.md

The current how-to-upgrade.md
 describes how to back up the database before upgrading (Step 2), but there are no instructions on how to use that backup to roll back if the upgrade fails. Users are left without guidance when something goes wrong during the upgrade process.

### How should we improve?

Add an "Upgrade Failed" section at the end of how-to-upgrade.md with rollback instructions for each supported database type:

MySQL: Restore from the mysqldump backup file.
H2: Replace the database directory with the copy made during backup.
PostgreSQL: Document the correct rollback steps using the backup created in Step 2.

danhuawang:

This PR improves the `docs/how-to-upgrade.md` documentation with the following changes:

1. Changed the PostgreSQL data backup command to use `-Fc` (custom compressed binary format) instead of `-F c`, producing a `.dump` file instead of `.sql`, and added a note explaining the custom format.
2. Changed the PostgreSQL schema dump command to use plain-text format (`-f`) instead of custom format (`-F c`), making it easier to diff against official schema files.
3. Improved the wording in Step 4 (schema comparison) to clarify which differences are acceptable (e.g., object ordering, comments) and which need manual resolution (e.g., table structures, column types, missing tables/indexes).
4. Added a new "Rollback if Upgrade Fails" section with restore instructions for MySQL, H2, and PostgreSQL backends.

1. The previous PostgreSQL backup command used `-F c` (custom format) but saved to a `.sql` extension, which is misleading. The custom format should use a `.dump` extension and be restored with `pg_restore`.
2. The schema dump should be in plain-text format so users can directly diff it against the official schema SQL files.
3. The original documentation lacked rollback guidance, leaving users without a clear recovery path if an upgrade fails.

Fix: #11112
Fix: #11122

No. This is a documentation-only change.

Documentation-only change. Verified the commands are correct by reviewing PostgreSQL and MySQL official documentation.


docs/how-to-upgrade.md
```diff
@@ -49,8 +49,10 @@ by creating a copy of the directory containing your database.
 For PostgreSQL, you can use the following command to backup your database:
 
 ```shell
-pg_dump -U username -h hostname -d database_name -n schema_name -a -F c -f data_backup.sql
+pg_dump -U username -h hostname -d database_name -n schema_name -Fc -f data_backup.dump
```
yuqi1129, line 52: The `-a` (data-only) flag was silently dropped here. The original command backed up data only; this new command backs up both schema and data. That's actually the right behaviour for rollback purposes, but it should be explicitly documented — the added note only mentions the format change. Please add a sentence explaining this semantic difference, e.g.: "Unlike the previous command, this backup includes both schema and data, which is necessary for a full rollback."

docs/how-to-upgrade.md
```diff
@@ -206,4 +212,43 @@ Confirm the running image tag matches the target version:
 kubectl get pods -n <namespace> -o jsonpath='{.items[*].spec.containers[*].image}'
 ```
 
+## Rollback if Upgrade Fails
```
yuqi1129, line 221: The rollback section is missing a critical prerequisite: the Gravitino server must be stopped before any restore is performed, otherwise concurrent writes can corrupt the database. Please add a note at the top of this section, e.g.:\n\n> **Important:** Stop the Gravitino server before restoring the database to avoid data corruption from concurrent writes.

docs/how-to-upgrade.md
```diff
@@ -206,4 +212,43 @@ Confirm the running image tag matches the target version:
...
+```
+
+### PostgreSQL
+
+Use `pg_restore` with `--clean` to drop existing objects before
+restoring them in one transaction. This is safer than manually
+dropping the schema first:
+
+```shell
+pg_restore -U username -h hostname -d database_name --clean --if-exists --single-transaction data_backup.dump
```
yuqi1129: Two issues with this restore command:\n\n1. **Missing schema filter**: the backup was taken with `-n schema_name`, so the restore should include `--schema=schema_name` (or `-n schema_name`) for symmetry. Without it, the behaviour depends on what is in the dump and may confuse users.\n\n2. **`--clean` limitation**: `pg_restore --clean` only drops objects that are present in the dump file. If the upgrade scripts added new tables or sequences not present in the backup, those objects will remain after the restore and may leave the database in an inconsistent state. Worth adding a note such as: "Note: objects added by the upgrade scripts that are not in the backup will not be removed automatically; verify the schema matches the expected state after restoring."\n\nSuggested command:\n```shell\npg_restore -U username -h hostname -d database_name -n schema_name --clean --if-exists --single-transaction data_backup.dump\n```

yuqi1129 (commented):
Thanks for the improvements! A few things to address before merging.

**Note (not inline-commentable — line not in diff):** The IRC backup command at the line below the main backup still uses the old `-a -F c` format and saves to a `.sql` extension. It should be updated consistently with the main backup: change to `-Fc -f data_backup_irc.dump` and drop the `-a` flag for the same reasons.

docs/how-to-upgrade.md
```diff
@@ -98,8 +100,12 @@ Gravitino. The files in this directory with names like
 corresponding to each of the released versions of Gravitino. You can
 determine differences between your schema and the official schema
 by comparing the contents of the official dump with the schema dump
-you created in the previous step. Some differences are acceptable and will not interfere
-with the upgrade process, but others need to be resolved manually
+you created in the previous step.
+
```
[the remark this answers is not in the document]
danhuawang: I tried using the diff command, but it generates too much noise, like differences caused purely by line ordering.I suggest avoiding specific tool recommendations here to keep the guidance general.

docs/how-to-upgrade.md
danhuawang: Fixed

docs/how-to-upgrade.md
danhuawang: Fixed.

docs/how-to-upgrade.md
danhuawang: Fixed.

docs/how-to-upgrade.md
```diff
@@ -49,8 +49,10 @@ by creating a copy of the directory containing your database.
 For PostgreSQL, you can use the following command to backup your database:
 
 ```shell
-pg_dump -U username -h hostname -d database_name -n schema_name -a -F c -f data_backup.sql
+pg_dump -U username -h hostname -d database_name -n schema_name -Fc -f data_backup.dump
 ```
+The `-Fc` option produces a compressed binary dump in PostgreSQL's
+custom format, which can be restored with `pg_restore`.
```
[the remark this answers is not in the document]
danhuawang: Fixed

docs/how-to-upgrade.md
```diff
@@ -206,4 +212,43 @@ Confirm the running image tag matches the target version:
...
+restore your database from the backup created in Step 2.
+
+### MySQL
+
+Use the backup file to restore your database. The `mysqldump`
+backup contains `DROP TABLE` statements by default (via `--opt`),
+so it will replace the upgraded tables automatically:
+
+```shell
+mysql -u username -h hostname <db_name> < backup.sql
```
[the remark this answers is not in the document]
danhuawang: Fixed.

yuqi1129 (approved):

PR #11464, opened 2026-06-05
[#7189] feat(client): Add password support to simple authentication

### LINKED ISSUE #7189: [Improvement] client should support password for simple auth

The Gravitino client currently supports simple authentication but does not allow setting passwords for simple auth. 

### How should we improve?

We should add the functionality to set passwords on the client side.

shahina-akd:

Added a new `withSimpleAuth(String userName, String password)` overload to `GravitinoClientBase.Builder` and a corresponding `SimpleTokenProvider(String, String)` constructor, allowing callers to supply an explicit password when using simple authentication.

Previously, `withSimpleAuth` always encoded `username:dummy` as the Basic auth credential, with no way to pass a real password.

The simple auth path hardcoded `"dummy"` as the password in `SimpleTokenProvider.buildToken()`. There was no API to provide an actual password on the client side, which blocked use cases where the server or a proxy expects a real credential in the Basic auth header.

Fix: #7189

Yes.
* New builder method: `GravitinoClientBase.Builder.withSimpleAuth(String userName, String password)`
* Existing `withSimpleAuth()` and `withSimpleAuth(String userName)` overloads are unchanged and continue to use `"dummy"` as the password for backwards compatibility.

Unit tests added to `TestSimpleTokenProvider`:
* `testAuthenticationWithPassword` verifies new `SimpleTokenProvider("alice", "secret")` encodes `alice:secret` 
* `testSingleArgConstructorUsesDummyPassword` verifies the existing single-arg path still encodes `bob:dummy`

Run with: 
`./gradlew :clients:client-java:test -PskipITs`

### DIFF
--- a/clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoClientBase.java
+++ b/clients/client-java/src/main/java/org/apache/gravitino/client/GravitinoClientBase.java
@@ -277,6 +277,20 @@ public abstract class GravitinoClientBase implements Closeable {
     }
 
     /**
+     * Sets the simple mode authentication for Gravitino with a specific username and password.
+     *
+     * @param userName The username of the user.
+     * @param password The password of the user.
+     * @return This Builder instance for method chaining.
+     */
+    public Builder<T> withSimpleAuth(String userName, String password) {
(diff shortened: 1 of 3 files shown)


roryqi:
We have added the basic authentication. We don't need this improvement.

PR #11473, opened 2026-06-08
[#11469] fix(catalog-glue): Treat empty catalog ID as unset

### LINKED ISSUE #11469: [Bug report] Glue catalog treats empty catalog ID as unset
### Version

main branch

### Describe what's wrong

When `aws-glue-catalog-id` is set to an empty string, the Glue catalog passes it
as a valid catalog ID to Glue API requests and Iceberg catalog properties instead
of ignoring it.

No exception is thrown, but Glue API requests are sent with an empty catalogId,
which may result in unexpected API errors.

### How to reproduce

1. Create a Glue catalog with property `aws-glue-catalog-id` = `""`
2. Perform any schema or table operation
3. The empty string is propagated to Glue API calls instead of being ignored

### Additional context

N/A

diqiu50:

Normalize empty `aws-glue-catalog-id` to `null` at initialization so downstream
code only needs to check for `null`.

Fix: #11469

An empty string catalog ID should be treated the same as an unset one.

No.

Existing unit tests cover the catalog ID propagation logic.

### DIFF
--- a/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogOperations.java
+++ b/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogOperations.java
@@ -135,7 +135,7 @@ public class GlueCatalogOperations implements CatalogOperations, SupportsSchemas
       throws RuntimeException {
     this.glueClient = GlueClientProvider.buildClient(config);
     String rawCatalogId = config.get(GlueConstants.AWS_GLUE_CATALOG_ID);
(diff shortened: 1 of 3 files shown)


catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogOperations.java
mchades, line 1: do we need to update the user doc for this change?

diqiu50: This change does not impact users. It is intended to prevent errors when the frontend passes an empty string ("").

mchades (approved):

PR #11484, opened 2026-06-08
[#11480] fix(build): bump slf4j-api to 2.0.17 to eliminate duplicate jars in distribution

### LINKED ISSUE #11480: [Improvement] Duplicate slf4j-api jars in distribution libs (2.0.16 and 2.0.17)

`distribution/package/libs/` contains two versions of `slf4j-api` simultaneously:

```
slf4j-api-2.0.16.jar
slf4j-api-2.0.17.jar
```

**Root cause**: the version catalog pins `slf4j = "2.0.16"`, but `log4j-slf4j2-impl:2.25.4` (included in the `bundles.log4j` bundle used by `core`, `server`, and `server-common`) declares a transitive dependency on `slf4j-api:2.0.17`.

The root `assembleDistribution` task collects `runtimeClasspath` from each subproject independently. Modules that include `bundles.log4j` resolve `slf4j-api` to `2.0.17` via conflict resolution; modules without it resolve to the catalog-pinned `2.0.16`. Both end up in the same output directory because they have different filenames, bypassing `DuplicatesStrategy.EXCLUDE`.

Affected paths (verified on current `main`):
- `distribution/package/libs/`
- `distribution/package-all/libs/`

### How should we improve?

Fix the inconsistency so all modules resolve `slf4j-api` to a single version. Possible approaches:

1. **Bump the version catalog** entry `slf4j = "2.0.16"` → `"2.0.17"` to align with what `log4j-slf4j2-impl:2.25.4` already requires (cleanest).
2. **Add a global constraint** `resolutionStrategy.force("org.slf4j:slf4j-api:2.0.17")` in the root build so all subprojects converge at resolution time.
3. **Exclude the old jar** in the copy task with `exclude("slf4j-api-2.0.16.jar")` (fragile — version-hardcoded, not recommended).

Option 1 or 2 addresses the root cause.

yuqi1129:

- Bumps `slf4j = "2.0.16"` → `"2.0.17"` in `gradle/libs.versions.toml`
- Removes the previous `resolutionStrategy.force` workaround from `build.gradle.kts`

`distribution/package/libs/` contained two versions of `slf4j-api` simultaneously:
- `slf4j-api-2.0.16.jar` — from modules that only see the catalog-pinned version
- `slf4j-api-2.0.17.jar` — from modules using `bundles.log4j` (`core`, `server`, `server-common`), where `log4j-slf4j2-impl:2.25.4` declares a transitive dependency on `slf4j-api:2.0.17`

Because the `assembleDistribution` copy task collects `runtimeClasspath` from each subproject independently, and the two versions have different filenames, `DuplicatesStrategy.EXCLUDE` cannot deduplicate them.

Bumping the catalog entry to `2.0.17` (which `log4j-slf4j2-impl:2.25.4` already requires) makes all subprojects naturally converge to the same version through Gradle's normal conflict resolution, so only `slf4j-api-2.0.17.jar` ends up in the distribution.

Fix: #11480

No.

- `./gradlew compileJava -PskipITs --parallel` — BUILD SUCCESSFUL (all 456 tasks)
- `./gradlew test -PskipITs --parallel` — BUILD SUCCESSFUL (all 763 tasks, 29 min)
- Verified `distribution/package/libs/` now contains only `slf4j-api-2.0.17.jar`
- Verified `:common:dependencies` resolves directly to `2.0.17` without any `->` upgrade


build.gradle.kts
```diff
@@ -336,12 +336,16 @@ subprojects {
...
+  // Force upgrade commons-beanutils/snappy-java/slf4j-api for all subprojects to resolve outdated transitive versions
   // commons-beanutils: pulled by Hadoop, Hive, Spark, Flink, etc.
   // snappy-java: pulled by Hadoop, Kafka, Iceberg, etc.
+  // slf4j-api: log4j-slf4j2-impl:2.25.4 requires 2.0.17, while the catalog pins 2.0.16; without
+  //   this force different subprojects resolve to different versions and both jars end up in the
+  //   distribution lib directory simultaneously. Force to 2.0.17 so all subprojects converge.
   configurations.all {
     resolutionStrategy.force("commons-beanutils:commons-beanutils:$commonsBeanutilsVersion")
     resolutionStrategy.force("org.xerial.snappy:snappy-java:$snappyJavaVersion")
+    resolutionStrategy.force("org.slf4j:slf4j-api:2.0.17")
```
jerryshao: 1) Can you check which module introduced the 2.0.16 jar? 2) Use the variable, don't hardcode the version number here.

yuqi1129: > Can you check which module introduced the 2.0.16 jar?

slf4j-api-2.0.16.jar — from modules that only see the catalog-pinned version. see https://github.com/apache/gravitino/blob/d8e65d749fe3ee31562e204d70a29378525b4ad8/gradle/libs.versions.toml#L31
slf4j-api-2.0.17.jar — from modules using bundles.log4j (core, server, server-common), where log4j-slf4j2-impl:2.25.4 declares a transitive dependency on slf4j-api:2.0.17

> Use the variable, don't hardcode the version number here.

I will change it.

jerryshao: Why don't you bump the version of slf4j in the versions file?

yuqi1129: There already exists the following content. Do you mean that we can 

```
slf4j = "2.0.16"
slf4j-api = { group = "org.slf4j", name = "slf4j-api", version.ref = "slf4j" }
```
Use this format, or change the version of `slf4j` to `2.0.17` directly?

```
slf4j_2_0_16 = "2.0.16"
slf4j_2_0_17 = "2.0.17"
....
```

jerryshao: I mean that we can change to `slf4j = "2.0.17"` in the version.toml if possible.

yuqi1129: I'm hesitant to change it globally to `2.0.17` as it will affect all modules. Let me carefully verify it.

yuqi1129: Changed to `2.0.17` globally and verified it.

jerryshao (approved):

PR #11493, opened 2026-06-08
[#11434] fix(idp): Validate name length at API layer before DB persist

### LINKED ISSUE #11434: [Improvement] IDP should validate username length at API layer instead of relying on DB constraint

When creating an IDP user with a username exceeding 128 characters, the server returns HTTP 500 with a `RuntimeException` wrapping a `PSQLException: ERROR: value too long for type character varying(128)`. The database internal error and full stack trace are exposed in the API response.

**Current behavior:**

- Request: `POST /api/idp/users` with a 256-character username
- Response: HTTP 500, `code: 1002`, `type: RuntimeException`
- Error message leaks DB schema details: `value too long for type character varying(128)`

**Expected behavior:**

- Response: HTTP 400, `ILLEGAL_ARGUMENTS_CODE`
- Error message: a user-friendly message like `"Username must not exceed 128 characters"`

This affects both security (DB internals should not be exposed) and user experience (400 with a clear message is far more actionable than 500 with a stack trace).

### How should we improve?

Add username length validation at the API/service layer before persisting to the database:

1. In `IdpUserGroupManager.addUser()` (or the REST layer `IdpUserOperations.addUser()`), add a precondition check:

```java
if (username == null || username.length() > 128) {
  throw new IllegalArgumentException("Username must not exceed 128 characters");
}
```

2. The existing exception handling should map `IllegalArgumentException` to HTTP 400 with `ILLEGAL_ARGUMENTS_CODE`.

3. Consider applying the same pattern to group names if they share a similar DB column constraint.

This ensures:
- Users get clear, actionable error messages
- Database schema details are not leaked in API responses
- Validation is consistent regardless of the backend storage engine

lasdf1234:

- Add 128-character length validation for IDP usernames and group names in `IdpCredentialValidator` before persistence.
- Route group name validation through `AddGroupRequest.validate()` via `validateGroupName()`.
- Add unit and integration tests for over-length username/group name requests (HTTP 400, `ILLEGAL_ARGUMENTS_CODE`).
- Align IDP OpenAPI examples and responses with actual REST behavior (403, per-operation examples, descriptions).

When creating an IDP user or group with a name longer than 128 characters, the server previously hit the database constraint and returned HTTP 500 with internal DB error details exposed. Validation at the API layer returns HTTP 400 with a clear message instead.

Fix: #11434

1. `POST /api/idp/users` and `POST /api/idp/groups` now return HTTP 400 with `code: 1001` when the name exceeds 128 characters, instead of HTTP 500.
2. IDP OpenAPI documentation updated for error/success examples and 403 responses.

- `./gradlew spotlessApply`
- `./gradlew :plugins:idp-basic:test -PskipITs -PskipDockerTests=true`
- `./gradlew :docs:lintIdpOpenAPI`


lasdf1234:
@roryqi Could you help me to review.THX.

docs/open-api/idp/idp.yaml
```diff
@@ -465,7 +490,16 @@ components:
         "usersToRemove": ["carol"]
       }
 
-    IdpUserResponse:
+    IdpUserAddResponse:
```
roryqi, line 493: Why do we return the response? Is it aligned to your implementation?Doest it meet our API design convention?

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/web/IdpRESTUtils.java
```diff
@@ -69,11 +69,10 @@ public static Response doAs(
 
   public static Response handleException(
       String resourceType, IdpOperationType op, String name, Exception e) {
-    String formatted = StringUtils.isBlank(name) ? "" : " [" + name + "]";
     String errorMsg =
         String.format(
-            "Failed to operate built-in IdP %s %s operation [%s], reason [%s]",
-            resourceType, formatted, op.name(), e.getMessage());
+            "Failed to operate built-in IdP %s [%s] operation [%s], reason [%s]",
+            resourceType, StringUtils.defaultString(name), op.name(), e.getMessage());
```
roryqi, line 75: What's the behaviour of `StringUtils.defaultString`?

lasdf1234: behaviour:
null -> ""
Other String -> Return as is.

docs/open-api/idp/idp.yaml
lasdf1234: IdpUserAddResponse is only an OpenAPI example name; the response schema is still IdpUserResponse.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/web/IdpRESTUtils.java
roryqi: Thanks. Nice to know this point. We can use this to optimize our code.

roryqi (approved):
LGTM.

PR #11497, opened 2026-06-08
[#11409] fix(idp): Validate incompatible simple mode at startup

### LINKED ISSUE #11409: [Improvement] Web v2: adapt the configuration `gravitino.server.rest.extensionPackages = org.apache.gravitino.idp.web.rest.feature` for local idp

Image

### How should we improve?

N/A

lasdf1234:

- Add `IdpConfigurationValidator` in `org.apache.gravitino.idp.config` to validate server configuration before the built-in IdP plugin starts.
- Reject startup when `gravitino.authorization.enable=true` and `gravitino.authenticators` includes `simple` (including the default value).
- Invoke validation from `IdpRESTFeature.configure()` before IdP registers HTTP Basic authentication.
- Add unit tests covering explicit `simple`, default authenticators, and compatible OAuth configuration.

Enabling built-in IdP (`gravitino.server.rest.extensionPackages = org.apache.gravitino.idp.web.rest.feature`) together with Simple authentication and authorization leads to broken Web UI login: the IdP plugin injects HTTP Basic authentication ahead of Simple, while Web v2 sends username-only Basic credentials.

This change fails fast at startup with a clear error instead of leaving operators with a running server and a non-functional UI.

Fix: #11409

1. Server startup now fails with `IllegalStateException` when built-in IdP is enabled with authorization and Simple authentication (explicit or default).
2. Operators must remove `simple` from `gravitino.authenticators` or disable `gravitino.authorization.enable`.

- `./gradlew spotlessApply`
- `./gradlew :plugins:idp-basic:test -PskipITs -PskipDockerTests=true --tests org.apache.gravitino.idp.config.TestIdpConfigurationValidator`


lasdf1234:
@roryqi Could you please review this PR for me? Thank you.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/config/IdpConfigurationValidator.java
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.gravitino.idp.config;
+
+import org.apache.gravitino.Config;
+import org.apache.gravitino.Configs;
+import org.apache.gravitino.auth.AuthenticatorType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Validates server configuration before the built-in IdP plugin starts. */
+public final class IdpConfigurationValidator {
```
roryqi: The logic seems simple. We don't need this class.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/config/IdpConfigurationValidator.java
```diff
@@ -0,0 +1,55 @@
...
+
+  private IdpConfigurationValidator() {}
+
+  /**
+   * Validates that the server configuration is compatible with the built-in IdP plugin.
+   *
+   * @param config The server configuration.
+   */
+  public static void validate(Config config) {
+    if (!config.get(Configs.ENABLE_AUTHORIZATION)) {
```
roryqi: Why do we need this logic? Authentication is different from authorization.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/config/IdpConfigurationValidator.java
lasdf1234: Got this class has been removed.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/config/IdpConfigurationValidator.java
lasdf1234: Only when "Simple + authorization" is used will the basic header be sent. If only "simple" is configured but authorization is not enabled, the "basic" header will not be sent, it is still usable.

roryqi: No, it will send basic header only if we use the simple mode.

lasdf1234: Only when "Simple + authorization" is used and user login graivitno through WEB-UI, front end will send the basic header.
If only "simple" is configured but authorization is not enabled, front end will not send the basic header.

Only front end send the basic header ,the IDP's filter to perform the verification, resulting in an error.

roryqi: Why does front behavior decide backend behavior?

lasdf1234: It is not the front end that determines the behavior of the back end.

No matter which mode enables authentication, it is necessary to inform the backend which user invoked the interface. In the simple mode, if authentication is enabled, it will use the same format as the basic mode and send a header containing the basic information. This will cause the bug to occur.

roryqi: Authorization isn't different from authentication.
Basic is an authentication mode. We can use authentication when we don't enable authorization.
I will suggest removing the authorization check here.

lasdf1234: Got.

plugins/idp-basic/src/test/java/org/apache/gravitino/idp/integration/test/IdpRESTApiIT.java
```diff
@@ -60,6 +62,10 @@
  */
 public class IdpRESTApiIT extends BaseIT {
 
+  /** RSA public key used only to satisfy OAuth authenticator initialization in this IT. */
+  private static final String OAUTH_PUBLIC_SIGN_KEY =
```
roryqi: Too long.

lasdf1234: Got resolved.

roryqi (approved):
LGTM.

PR #11499, opened 2026-06-08
[#11498] improvement(core): Reuse catalog cache when resolving properties

### LINKED ISSUE #11498: [Improvement] Reuse the catalog cache when resolving catalog properties

`CatalogManager#getResolvedProperties` creates a new `IsolatedClassLoader` and `BaseCatalog` for every catalog whenever `listCatalogsInfo` is called. The temporary catalog and class loader are closed immediately after resolving the catalog properties.

This repeatedly performs catalog initialization and class-loader resource creation even though `CatalogManager` already maintains a catalog wrapper cache through `loadCatalogAndWrap`.

### How should we improve?

Resolve catalog properties through `loadCatalogAndWrap` so the first list call loads uncached catalogs into the existing catalog cache and subsequent calls reuse the cached wrappers.

Keep the existing cache expiration and invalidation behavior for catalog alter, drop, enable, disable, and remote entity changes.

Add regression coverage verifying that `listCatalogsInfo` loads uncached catalogs into the cache and reuses the same wrappers on subsequent calls.

yuqi1129:

Resolve catalog properties through `loadCatalogAndWrap` so that `listCatalogsInfo` loads uncached catalogs into the existing catalog cache and reuses cached wrappers on subsequent calls.

`getResolvedProperties` currently creates and closes a temporary catalog and isolated classloader for every catalog on every `listCatalogsInfo` call. Reusing the catalog cache avoids repeated catalog and classloader initialization.

Fix: #11498

No.

- Added coverage verifying that uncached catalogs are loaded into the cache.
- Verified that subsequent list calls reuse the same cached wrappers.
- Ran `TestCatalogManager`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
@@ -1145,9 +1145,13 @@ public class CatalogManager implements CatalogDispatcher, Closeable {
    * @return The resolved properties.
    */
   private Map<String, String> getResolvedProperties(CatalogEntity entity) {
-    CatalogWrapper catalogWrapper = loadCatalogAndWrap(entity.nameIdentifier());
-    return catalogWrapper.classLoader.withClassLoader(
-        cl -> catalogWrapper.catalog.properties(), RuntimeException.class);
+    Map<String, String> conf = entity.getProperties();
+    String provider = entity.getProvider();
+
+    try (IsolatedClassLoader classLoader = createClassLoader(provider, conf)) {
+      BaseCatalog<?> catalog = createBaseCatalog(classLoader, entity);
+      return classLoader.withClassLoader(cl -> catalog.properties(), RuntimeException.class);
+    }
   }
 
   private BaseCatalog<?> createBaseCatalog(IsolatedClassLoader classLoader, CatalogEntity entity) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/catalog/CatalogManager.java
```diff
@@ -1145,13 +1145,8 @@ CatalogWrapper createCatalogWrapper(
...
   private Map<String, String> getResolvedProperties(CatalogEntity entity) {
-    Map<String, String> conf = entity.getProperties();
-    String provider = entity.getProvider();
-
-    try (IsolatedClassLoader classLoader = createClassLoader(provider, conf)) {
-      BaseCatalog<?> catalog = createBaseCatalog(classLoader, entity);
-      return classLoader.withClassLoader(cl -> catalog.properties(), RuntimeException.class);
-    }
+    BaseCatalog<?> catalog = loadCatalogAndWrap(entity.nameIdentifier()).catalog();
+    return catalog.properties();
```
mchades: do we need `classLoader.withClassLoader(cl -> catalog.properties(), RuntimeException.class)`?

yuqi1129: done

yuqi1129: fix

mchades (approved):

PR #11502, opened 2026-06-08
[#11501] test(iceberg-rest): Add Trino IT for Iceberg REST authorization in auxMode

### LINKED ISSUE #11501: [Improvement] Add Trino integration test for Iceberg REST Catalog authorization in auxiliary mode

Gravitino has `IcebergAuthorizationIT` which verifies that Gravitino authorization is enforced when the Iceberg REST Catalog (IRC) is served in auxiliary mode (`auxMode`) — but it only exercises this path through **Spark**.

There is no equivalent coverage for **Trino's native `iceberg` REST connector**. Trino accesses the IRC differently from Spark (it cannot send Gravitino simple-auth/Basic identity and only supports OAuth2 `Bearer` tokens), so the Spark test does not cover the Trino code path. We should add a Trino integration test that confirms Gravitino privilege grant/deny is enforced end-to-end when an IRC is accessed via Trino.

### How should we improve?

Add a test-only module that boots the **full Gravitino server in deploy mode** (required because Gravitino is compiled against Iceberg 1.10.1 while Trino bundles its own Iceberg — they cannot coexist in one classloader) with the `iceberg-rest` auxiliary service and authorization enabled, then drives it with an in-process Trino query runner (`AbstractTestQueryFramework`, Trino 478 / Java 24 toolchain).

Authentication is OAuth-only (RS256-signed JWTs), since Trino's iceberg REST connector cannot send Gravitino Basic auth. The test asserts that a `NORMAL_USER` can/cannot create schemas/tables and `SELECT`/`INSERT` depending on the privileges granted/denied by `SUPER_USER`, mirroring a subset of the existing Iceberg authorization ITs.

roryqi:

Add a new test-only module `iceberg/iceberg-rest-trino-it` that verifies Gravitino's authorization is enforced when an Iceberg REST Catalog served in auxiliary mode (`auxMode`) is accessed through Trino's native `iceberg` REST connector. This is the Trino analog of the Spark-based `IcebergAuthorizationIT`.

Specifically:
- **`build.gradle.kts` + `settings.gradle.kts`** — scaffold the module on a Java 24 toolchain with `trino-iceberg:478` / `trino-testing:478`, mirroring `trino-connector-473-478` (Error Prone / JaCoCo disabled for JDK 24, JUnit aligned to Trino's 6.0.0 BOM, jjwt api+impl only).
- **`MockOAuthTokenServer`** (+ `TestMockOAuthTokenServer`) — a minimal RS256 client-credentials OAuth2 token endpoint bound to `127.0.0.1`, used by the IRC dynamic config provider's service identity. Covered by a unit test asserting a valid signed JWT is returned.
- **`TrinoIcebergRestAuthorizationITBase` / `TrinoIcebergRestAuthorizationIT` / `TrinoIcebergRestSmokeIT`** — boot the full Gravitino server in deploy mode with the `iceberg-rest` aux service + authorization + OAuth authenticator, and drive an in-process Trino query runner to assert privilege grant/deny.

The existing `IcebergAuthorizationIT` only covers the Spark path. Trino accesses the IRC over a different authentication path (OAuth2 bearer only, no Gravitino Basic auth), which was previously untested. This adds end-to-end coverage that Gravitino authorization is enforced for Trino IRC clients.

Fix: #11501

No. Test-only changes; no new APIs or property keys.

This change *is* an integration test. Run with the built distribution in deploy mode:

```
./gradlew compileDistribution
./gradlew :iceberg:iceberg-rest-trino-it:test -PskipDockerTests=false -DtestMode=deploy
```

`MockOAuthTokenServer` is additionally covered by `TestMockOAuthTokenServer`.


iceberg/iceberg-rest-trino-it/build.gradle.kts
```diff
@@ -0,0 +1,109 @@
...
+  testImplementation(libs.testcontainers)
+  testImplementation(libs.testcontainers.postgresql)
+
+  // jjwt is resolved to 0.13.x on this module's test classpath (Trino forces it up). Only pull
+  // api + impl; the JSON serializer comes from jjwt-jackson:0.13.x which Trino already provides.
+  // Do NOT pull jjwt-gson:0.11.x (the bundle's default) — its stale 0.11.x serializer service is
+  // incompatible with the 0.13.x runtime and makes signWith().compact() fail.
+  testImplementation(libs.jwt.api)
+  testRuntimeOnly(libs.jwt.impl)
+  testRuntimeOnly("io.jsonwebtoken:jjwt-jackson:0.13.0")
```
diqiu50: Should this be placed in the libs directory?

iceberg/iceberg-rest-trino-it/src/test/java/org/apache/gravitino/iceberg/integration/test/trino/TrinoIcebergRESTAuthorizationIT.java
```diff
@@ -0,0 +1,79 @@
...
+    catalogAsSuper.asSchemas().createSchema("db1", "", new HashMap<>());
+    startTrino();
+    sql(SUPER_CATALOG, "CREATE TABLE db1.shared (id integer)");
+    sql(SUPER_CATALOG, "INSERT INTO db1.shared VALUES (1)");
+  }
+
+  @Test
+  @Order(1)
+  public void testNormalUserDeniedWithoutPrivileges() {
+    assertThrows(RuntimeException.class, () -> sql(NORMAL_CATALOG, "SELECT * FROM db1.shared"));
```
diqiu50: Can you check the specific error message instead of throwing a RuntimeException?

iceberg/iceberg-rest-trino-it/src/test/java/org/apache/gravitino/iceberg/integration/test/trino/TrinoIcebergRESTAuthorizationITBase.java
```diff
@@ -0,0 +1,249 @@
...
+        LOG.warn("Failed to drop test metalake", e);
+      }
+    } finally {
+      if (tokenServer != null) {
+        tokenServer.close();
+      }
+      try {
+        super.stopIntegrationTest();
+      } finally {
+        containerSuite.close();
```
diqiu50: Closing ContainerSuite might affect other tests

iceberg/iceberg-rest-trino-it/build.gradle.kts
```diff
@@ -0,0 +1,109 @@
...
+    "com.github.docker-java:docker-java-transport:3.4.2",
+    "com.github.docker-java:docker-java-transport-zerodep:3.4.2"
+  )
+  exclude(group = "io.jsonwebtoken", module = "jjwt-gson")
+  exclude(group = "org.apache.logging.log4j", module = "log4j-to-slf4j")
+}
+
+// Test against the highest supported Trino version.
+val minTrinoVersion = 473
+val maxTrinoVersion = 478
```
diqiu50: Do we need to support multiple versions here?

iceberg/iceberg-rest-trino-it/build.gradle.kts
```diff
@@ -0,0 +1,109 @@
...
+  toolchain.languageVersion.set(JavaLanguageVersion.of(24))
+}
+
+dependencies {
+  testImplementation(project(":api"))
+  testImplementation(project(":catalogs:catalog-common"))
+  testImplementation(project(":common"))
+  testImplementation(project(":clients:client-java"))
+  testImplementation(project(":core"))
+  testImplementation(project(":server-common"))
```
diqiu50, line 55: Some dependencies may not be used.

iceberg/iceberg-rest-trino-it/build.gradle.kts
```diff
@@ -0,0 +1,109 @@
...
+    ?.isEnabled = false
+}
+
+tasks.test {
+  useJUnitPlatform()
+  // These ITs require the built distribution (deploy mode) and a server subprocess.
+  val skipITs = providers.gradleProperty("skipITs").map(String::toBoolean).orElse(false)
+  val skipTests = providers.gradleProperty("skipTests").map(String::toBoolean).orElse(false)
+  onlyIf { !skipITs.get() && !skipTests.get() }
+}
```
diqiu50, line 110: Skipping here is inconsistent with other modules, which use project.hasProperty("skipITs").

iceberg/iceberg-rest-trino-it/build.gradle.kts
roryqi: Some dependencies are required for BaseIT.

iceberg/iceberg-rest-trino-it/build.gradle.kts
roryqi: Good catch. It turned out the explicit declaration was redundant: `jjwt-jackson:0.13.0` is already provided transitively by `trino-iceberg`/`trino-main`/`trino-testing` (verified with `dependencyInsight` that it stays on the test runtime classpath). So I removed the line entirely — no hardcoded version and no new catalog entry needed. Separately, the remaining version literals in this module (`testcontainers`, `docker-java`, and the Trino version) are now centrally managed in `gradle/libs.versions.toml`.

iceberg/iceberg-rest-trino-it/src/test/java/org/apache/gravitino/iceberg/integration/test/trino/TrinoIcebergRESTAuthorizationIT.java
roryqi: Done. Added `assertAuthorizationDenied`, which walks the exception cause chain and asserts the IRC 403 message `"is not authorized to perform operation"` (produced by `BaseMetadataAuthorizationMethodInterceptor`), applied to both denial tests.

iceberg/iceberg-rest-trino-it/src/test/java/org/apache/gravitino/iceberg/integration/test/trino/TrinoIcebergRESTAuthorizationITBase.java
roryqi: Agreed. Removed `containerSuite.close()` from teardown — it is a process-wide singleton whose PostgreSQL container may be reused by other tests in the same JVM. This matches the Spark analog `IcebergAuthorizationIT`, which does not close it either.

iceberg/iceberg-rest-trino-it/build.gradle.kts
roryqi: No — this is a test-only module that boots an in-process Trino query runner, so it only needs a single version. Removed the min/max range + `-PtrinoVersion` machinery and pinned the version, now sourced from the catalog (`libs.versions.trino-iceberg-it`).

iceberg/iceberg-rest-trino-it/build.gradle.kts
roryqi: Removed `awaitility` and `commons-lang3` (no references). I also checked `:common`, but it is required at compile time (`ConfigEntry`/`MetalakeDTO` show up in the signatures of `Configs`/`createMetalake`), so it is kept.

iceberg/iceberg-rest-trino-it/build.gradle.kts
roryqi: Done. Switched to `project.hasProperty("skipITs")` + `exclude("**/integration/test/**")`, consistent with `iceberg-rest-server`. This also fixes a latent issue: the old `onlyIf` skipped these ITs under `-PskipTests`, when they should still run.

iceberg/iceberg-rest-trino-it/build.gradle.kts
roryqi: Correction: CI showed `awaitility` and `commons-lang3` are actually **not** unused — the inherited `BaseIT`/`ITUtils` harness (from the `integration-test-common` testArtifacts dep) uses them at runtime, and `testArtifacts` does not carry transitive runtime deps, so removing them failed the ITs with `NoClassDefFoundError: org/awaitility/Awaitility`. I restored both (as `testRuntimeOnly`, since this module's own sources do not reference them) in 63ac498. `:common` remains required at compile time. Net result: the declared dependencies are all needed.

diqiu50 (approved):
LGTM

PR #11506, opened 2026-06-08
[#10457][followup]fix(deps): Exclude log4j 1.x globally (CVE-2020-9493)

bharos:

Add a global `exclude(group = "log4j", module = "log4j")` in the root `configurations.all` block to prevent the vulnerable Log4j 1.x artifact from being resolved anywhere in the project.

Hive 2.3.9 and Hadoop 2.10.2 transitively pull in `log4j:log4j:1.2.17`, which is vulnerable to [CVE-2020-9493](https://nvd.nist.gov/vuln/detail/CVE-2020-9493).

The safe `log4j-1.2-api` bridge from Log4j 2.x is already included in the log4j bundle (`libs.versions.toml`), so this exclusion does not break any logging functionality. No source files in the project import `org.apache.log4j`.

PR #10465 already excludes `log4j` from `hive-metastore2-libs` and `hive-metastore3-libs`, but `log4j:log4j:1.2.17` is still resolved as a transitive **test** dependency in `catalog-hive` and `catalog-lakehouse-paimon` (via Hadoop's `slf4j-log4j12` → `log4j:log4j`). It does not end up in the distribution package, but it does get downloaded to `~/.gradle/caches/` during builds, triggering vulnerability scanners on developer machines.

This global exclusion provides defense-in-depth: it removes the vulnerable JAR from all configurations (including test) and prevents future modules from accidentally pulling it in.

Followup to #10465.

No. This only removes a vulnerable transitive dependency that is already replaced by the `log4j-1.2-api` bridge.

1. Verified `./gradlew :catalogs:hive-metastore-common:compileJava :catalogs:hive-metastore2-libs:compileJava :catalogs:hive-metastore3-libs:compileJava :catalogs:catalog-hive:compileJava` passes (BUILD SUCCESSFUL).
2. Confirmed `log4j:log4j` no longer appears in the dependency tree.
3. Confirmed zero source imports of `org.apache.log4j` across the entire project.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -342,6 +342,10 @@ subprojects {
   configurations.all {
     resolutionStrategy.force("commons-beanutils:commons-beanutils:$commonsBeanutilsVersion")
     resolutionStrategy.force("org.xerial.snappy:snappy-java:$snappyJavaVersion")
+
+    // Exclude log4j 1.x (CVE-2020-9493, CVSS 9.8) pulled transitively by Hive and Hadoop.
+    // The safe log4j-1.2-api bridge from Log4j 2.x is already included in the log4j bundle.
+    exclude(group = "log4j", module = "log4j")
   }
 
   repositories {


jerryshao:
@yuqi1129 can you please review this PR?

yuqi1129:
@bharos 
I could not see the jar you are intending to exclude from the distribution package based on the current main branch, so if I'm not wrong, this PR will not affect the final result of the distribution package. Is that true?

bharos:
> @bharos I could not see the jar you are intending to exclude from the distribution package based on the current main branch, so if I'm not wrong, this PR will not affect the final result of the distribution package. Is that true?

You're right that log4j:log4j does not end up in the distribution package , it's not in the runtimeClasspath of any module.

However, it is still resolved on the current main branch as a transitive test dependency:
```
 ./gradlew :catalogs:catalog-lakehouse-paimon:dependencies --configuration testRuntimeClasspath | grep "log4j:log4j:"
|         |    |    |    |    |         |    \--- log4j:log4j:1.2.17
(base) ╭─bmurali@FVFHP28PQ05N ~/Code/gravitino  ‹main*› 
╰─➤  ./gradlew :catalogs:catalog-hive:dependencies --configuration testRuntimeClasspath | grep "log4j:log4j:"

|    |    \--- log4j:log4j:1.2.17
|    |    |    |    |    |    \--- log4j:log4j:1.2.17
|    |    |    |    |    +--- log4j:log4j:1.2.17
```

This means the vulnerable JAR still gets downloaded to ~/.gradle/caches/ during builds and test runs, which triggers security scanners on developer machines (which is how I found it - our vulnerability scanner flagged it repeatedly on my laptop).

This global exclusion serves as defense-in-depth: it prevents the JAR from being resolved anywhere (including test configurations), and ensures that if a future module introduces a new transitive dependency on log4j 1.x, it will be caught automatically rather than requiring per-module exclusions.

jerryshao (approved):
The change LGTM.

yuqi1129 (approved):

PR #11510, opened 2026-06-09
[#11513] fix(catalog-mysql): Fix MySQL catalog incompatible with MariaDB due to case-sensitive CURRENT_TIMESTAMP check

### LINKED ISSUE #11513: [Bug report] MySQL catalog is incompatible with MariaDB due to case-sensitive CURRENT_TIMESTAMP check
### Version

main branch

### Describe what's wrong

When using the MySQL catalog to connect to a MariaDB database, columns with `CURRENT_TIMESTAMP` as the default value are not correctly parsed.

In `MysqlColumnDefaultValueConverter`, the `CURRENT_TIMESTAMP` check uses `String.startsWith()`, which is case-sensitive. MariaDB returns `current_timestamp` (lowercase) in the column metadata, while MySQL returns `CURRENT_TIMESTAMP` (uppercase). This causes the default value to be incorrectly handled when the catalog is connected to a MariaDB instance.

### How to reproduce

1. Set up a MariaDB instance and create a table with a `TIMESTAMP` column that has `CURRENT_TIMESTAMP` as the default value:

```sql
CREATE TABLE test_table (
  id INT PRIMARY KEY,
  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```
2. Configure a MySQL catalog in Gravitino pointing to the MariaDB database.
3. Load the table metadata through the Gravitino MySQL catalog.
4. Observe that the default value of created_at is not correctly parsed (it returns the raw string instead of being recognized as DEFAULT_VALUE_OF_CURRENT_TIMESTAMP).

### Additional context

_No response_

a19920714liou:
Fix https://github.com/apache/gravitino/issues/11513

Changed the `CURRENT_TIMESTAMP` comparison in `MysqlColumnDefaultValueConverter` from case-sensitive `startsWith` to case-insensitive `StringUtils.startsWithIgnoreCase`.

MariaDB returns default values with different casing (e.g., `current_timestamp` in lowercase) compared to MySQL (e.g., `CURRENT_TIMESTAMP` in uppercase). The existing `String.startsWith()` check is case-sensitive, which causes it to fail when connecting to a MariaDB database via the MySQL catalog, resulting in incorrect default value parsing.

No user-facing API changes. This fix ensures MariaDB users can correctly use the MySQL catalog with proper default value handling.

Manually tested by connecting the MySQL catalog to a MariaDB instance and verifying that columns with `CURRENT_TIMESTAMP` as default value are correctly parsed.

### DIFF
--- a/catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/converter/MysqlColumnDefaultValueConverter.java
+++ b/catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/converter/MysqlColumnDefaultValueConverter.java
@@ -25,6 +25,7 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.format.DateTimeParseException;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcColumnDefaultValueConverter;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcTypeConverter;
 import org.apache.gravitino.rel.expressions.Expression;
@@ -50,7 +51,7 @@ public class MysqlColumnDefaultValueConverter extends JdbcColumnDefaultValueConv
     }
 
     if (isExpression) {
-      if (columnDefaultValue.startsWith(CURRENT_TIMESTAMP)) {
+      if (StringUtils.startsWithIgnoreCase(columnDefaultValue, CURRENT_TIMESTAMP)) {
         return DEFAULT_VALUE_OF_CURRENT_TIMESTAMP;
       }
       // The parsing of MySQL expressions is complex, so we are not currently undertaking the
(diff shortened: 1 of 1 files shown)


roryqi:
If this is a fix, we should have an issue. Could u create an issue first?

a19920714liou:
> If this is a fix, we should have an issue. Could u create an issue first?

No, I just encountered this issue and made the necessary changes before submitting it directly

a19920714liou:
> If this is a fix, we should have an issue. Could u create an issue first?

Do I need to add an issue？

roryqi:
> > If this is a fix, we should have an issue. Could u create an issue first?
> 
> Do I need to add an issue？

 yes.

a19920714liou:
> > > If this is a fix, we should have an issue. Could u create an issue first?
> > 
> > 
> > Do I need to add an issue？
> 
> yes.

https://github.com/apache/gravitino/issues/11513

yuqi1129 (approved):
LGTM

yuqi1129:
@a19920714liou 
Thanks for your contribution, and welcome to join the Gravitno community!!!

PR #11558, opened 2026-06-10
[#11553] feat(lance): add health check endpoints for Lance REST servers

### LINKED ISSUE #11553: [FEATURE] Add health check endpoints for Lance REST servers
### Describe the feature

Add health check endpoints to the Lance REST server so that Kubernetes probes and monitoring systems and ha systems can check service status without credentials.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

flaming-archer:

This PR adds health check endpoints and a custom authentication filter for the Lance REST server, aligning it with the Iceberg REST server implementation.

1、LanceAuthenticationFilter — A new subclass of AuthenticationFilter that whitelists /lance/health and /lance/health/* paths to bypass authentication, in addition to the default health paths (/health, /api/health). This allows Kubernetes probes and monitoring systems to reach the Lance health endpoints without credentials.

2、LanceHealthOperations — A new JAX-RS resource at /health/live that returns a MicroProfile Health-compatible liveness response (200 OK with UP status).

3、LanceRESTService — Updated to inject LanceAuthenticationFilter into the Jetty server via createAuthenticationFilter() override.

4、build.gradle.kts — Added testImplementation(libs.junit.jupiter.params) dependency for @ParameterizedTest support.

- Health check endpoints are required for Kubernetes liveness probes and monitoring integration.
- Without LanceAuthenticationFilter, all Lance REST endpoints (including health checks) require authentication, making health probes fail.

Fixed: #11553 

Yes. The Lance REST server now exposes:

- GET /lance/health/live — Liveness probe endpoint (no authentication required).

Added unit tests:

- TestLanceAuthenticationFilter — Verifies health paths bypass auth and non-health paths require auth.
- TestLanceHealthOperations — Verifies the liveness endpoint returns 200 with UP status.


flaming-archer:
@jerryshao Could you please take a look? You wrote the Iceberg version, so your input would be really helpful.​

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
```diff
@@ -28,13 +28,17 @@
...
 import org.apache.gravitino.lance.common.ops.NamespaceWrapper;
+import org.apache.gravitino.lance.service.LanceAuthenticationFilter;
 import org.apache.gravitino.listener.EventBus;
 import org.apache.gravitino.listener.api.event.EventSource;
 import org.apache.gravitino.metrics.MetricsSystem;
 import org.apache.gravitino.metrics.source.MetricsSource;
 import org.apache.gravitino.server.web.HttpAuditFilter;
 import org.apache.gravitino.server.web.HttpServerMetricsSource;
-import org.apache.gravitino.server.web.JettyServer;
+import org.apache.gravitino.server.web.
```
yuqi1129: This import is currently syntactically broken, so the PR cannot pass formatting/compilation. I verified the PR head with `./gradlew :lance:lance-rest-server:compileJava -PskipITs`, and `:lance:lance-rest-server:spotlessJava` fails with `38:41: error: <identifier> expected`. Please restore the intended import, e.g. `org.apache.gravitino.server.web.JettyServer`.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
```diff
@@ -66,7 +70,13 @@ public void serviceInit(Map<String, String> properties, boolean auxMode) {
...
     JettyServerConfig serverConfig = JettyServerConfig.fromConfig(lanceConfig);
 
-    server = new JettyServer();
+    server =
+        new JettyServer() {
+          @Override
+          protected javax.servlet.Filter createAuthenticationFilter() {
+            return new LanceAuthenticationFilter();
+          }
+        };
```
yuqi1129: Should Lance also register root-level health aliases, matching Gravitino and Iceberg REST? Iceberg adds `new HealthAliasServlet("/iceberg")` for `/health/*` and `/health.html`, so common probes can use `/health`, `/health/live`, `/health/ready`, and `/health.html`. This PR only exposes `/lance/health*`; adding `HealthAliasServlet("/lance")` after registering the Lance servlet would make Lance consistent with the existing health endpoint behavior.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
```diff
@@ -66,7 +70,13 @@ public void serviceInit(Map<String, String> properties, boolean auxMode) {
     LanceConfig lanceConfig = new LanceConfig(properties);
     JettyServerConfig serverConfig = JettyServerConfig.fromConfig(lanceConfig);
 
-    server = new JettyServer();
+    server =
+        new JettyServer() {
+          @Override
+          protected javax.servlet.Filter createAuthenticationFilter() {
+            return new LanceAuthenticationFilter();
```
yuqi1129: When adding a Lance-specific auth filter for `/lance/health*`, please also pass the same Lance health matcher to `HttpAuditFilter` below. Iceberg REST uses `IcebergHealthCheckPathMatcher` for both auth and audit filtering. Without the audit-side matcher, `/lance/health/ready` returning 503 will be treated as a normal request failure and emit audit/failure events, because the default matcher only recognizes `/health*` and `/api/health*`.

lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/TestLanceAuthenticationFilter.java
```diff
@@ -0,0 +1,135 @@
...
+        "/lance/health/live",
+        "/lance/health/ready",
+        "/health",
+        "/health/live",
+        "/health/ready",
+        "/health.html",
+        "/api/health",
+        "/api/health/live",
+        "/api/health/ready"
+      })
```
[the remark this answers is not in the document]
flaming-archer: Let's align with Iceberg and leave it as is for now.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/LanceAuthenticationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+ *   <li>allows health check endpoints to bypass authentication via {@link
+ *       LanceHealthCheckPathMatcher};
+ *   <li>returns Lance-compatible JSON error responses on authentication failure instead of the
+ *       default HTML error pages.
+ * </ul>
+ */
+public class LanceAuthenticationFilter extends AuthenticationFilter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LanceAuthenticationFilter.class);
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
jerryshao: **Style: Use the project-shared `ObjectMapper` instead of a raw `new ObjectMapper()`.**

The base `AuthenticationFilter` uses `ObjectMapperProvider.objectMapper()`, which carries project-wide serialization config (module registration, null handling, date formats, etc.). `IcebergAuthenticationFilter` similarly uses `IcebergObjectMapper.getInstance()`. A bare `new ObjectMapper()` bypasses all of that — use whichever shared mapper is appropriate for the Lance serialization context.

jerryshao (commented):
Code review covering correctness, alignment with the Iceberg pattern, and minor style nits.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/LanceAuthenticationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+    int status;
+    String message;
+    if (exception instanceof UnauthorizedException) {
+      status = HttpServletResponse.SC_UNAUTHORIZED;
+      message = exception.getMessage();
+      if (message == null || message.isEmpty()) {
+        message = "Authentication failed";
+      }
+    } else {
+      status = HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
```
jerryshao, line 70: **Bug: `ForbiddenException` maps to 500 instead of 403.**

The `else` branch catches every non-`UnauthorizedException` — including `ForbiddenException` — and returns 500. The base `AuthenticationFilter.sendAuthErrorResponse` has an explicit `ForbiddenException` → 403 branch. This override silently drops that case.

Add an `else if (exception instanceof ForbiddenException)` branch mapping to `SC_FORBIDDEN` (403), and add a corresponding test to `TestLanceAuthenticationFilter`.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/LanceHealthCheckPathMatcher.java
```diff
@@ -0,0 +1,44 @@
...
+ * constructing the Lance REST server so that both filters agree on which paths are probe traffic.
+ */
+public class LanceHealthCheckPathMatcher extends HealthCheckPathMatcher {
+
+  @Override
+  public boolean isHealthCheckPath(String path) {
+    if (super.isHealthCheckPath(path)) {
+      return true;
+    }
+    if (path == null) {
```
jerryshao: **Minor: redundant `null` guard — `super.isHealthCheckPath(path)` already handles `null`.**

`HealthCheckPathMatcher.isHealthCheckPath` returns `false` immediately when `path == null`, so `null` can never reach this second check. The guard is harmless but misleading — it implies `null` could slip through the `super` call. Safe to remove.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceHealthOperations.java
```diff
@@ -0,0 +1,143 @@
...
+   * Liveness probe. Returns 200 as long as the HTTP thread can respond.
+   *
+   * @return 200 OK with an UP {@link HealthResponse}
+   */
+  @GET
+  @Path("/live")
+  @Timed(name = "lance.health.live." + MetricNames.HTTP_PROCESS_DURATION, absolute = true)
+  @ResponseMetered(name = "lance.health.live", absolute = true)
+  public Response live() {
+    HealthCheckDTO check = up(CHECK_HTTP_SERVER, Collections.<String, String>emptyMap());
```
jerryshao: **Minor: unnecessary explicit type witness on `Collections.emptyMap()`.**

`Collections.<String, String>emptyMap()` — Java infers the type parameter from context, so the `<String, String>` witness is redundant. `IcebergHealthOperations` uses the inferred form (`Collections.emptyMap()`) throughout. Same applies to the identical calls on lines 105 and 124.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/rest/LanceHealthOperations.java
```diff
@@ -0,0 +1,143 @@
...
+    return overall == HealthCheckDTO.Status.UP ? Utils.ok(body) : Utils.serviceUnavailable(body);
+  }
+
+  private HealthCheckDTO checkNamespaceWrapper() {
+    NamespaceWrapper wrapper = getNamespaceWrapper();
+    if (wrapper == null) {
+      return down(CHECK_NAMESPACE_WRAPPER, "reason", "namespace wrapper not initialized");
+    }
+    try {
+      wrapper.asNamespaceOps();
```
jerryshao: **Functional concern: readiness probe triggers lazy initialization and can block indefinitely.**

```java
wrapper.asNamespaceOps();   // calls initIfNeeded() → synchronized initAll() → initialize()
```

`NamespaceWrapper.asNamespaceOps()` lazily initializes the backend (the comment in that class says "it may block the startup"). A Kubernetes readiness probe hitting `/ready` before initialization completes will block inside the `synchronized` block until `initAll()` finishes (or hangs), which can cause probe timeouts and false pod restarts.

Compare with `IcebergHealthOperations.checkCatalogWrapperManager()` — it simply checks `if (getCatalogWrapperManager() == null)` with no side effects.

`NamespaceWrapper` already has a `private volatile boolean initialized` field. The cleanest fix is to expose a `public boolean isInitialized()` accessor and use that here instead of calling `asNamespaceOps()`.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
```diff
@@ -66,7 +69,13 @@ public void serviceInit(Map<String, String> properties, boolean auxMode) {
     LanceConfig lanceConfig = new LanceConfig(properties);
     JettyServerConfig serverConfig = JettyServerConfig.fromConfig(lanceConfig);
 
-    server = new JettyServer();
+    server =
```
jerryshao: **Style: FQN in method signature — use an import instead.**

```java
protected javax.servlet.Filter createAuthenticationFilter() {
```

Per the project coding standards, FQNs inline in code should be replaced with a regular `import javax.servlet.Filter;` declaration. The return type should just be `Filter`.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
```diff
@@ -66,7 +69,13 @@ public void serviceInit(Map<String, String> properties, boolean auxMode) {
     LanceConfig lanceConfig = new LanceConfig(properties);
     JettyServerConfig serverConfig = JettyServerConfig.fromConfig(lanceConfig);
```
jerryshao, line 69: **Minor: consider a named class instead of an anonymous `JettyServer` subclass.**

Anonymous inner classes in field initializers are harder to test and document in isolation. Since `LanceAuthenticationFilter` already exists as a named class, a small named `LanceJettyServer` (even package-private) would be consistent with the rest of the module and easier to reference in tests or javadoc.

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/service/LanceAuthenticationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+ * An {@link AuthenticationFilter} subclass for the Lance REST server that:
+ *
+ * <ul>
+ *   <li>allows health check endpoints to bypass authentication via {@link
+ *       LanceHealthCheckPathMatcher};
+ *   <li>returns Lance-compatible JSON error responses on authentication failure instead of the
+ *       default HTML error pages.
+ * </ul>
+ */
+public class LanceAuthenticationFilter extends AuthenticationFilter {
```
jerryshao, line 43: Do we support authenticaton for now for Lance rest?

lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java
```diff
@@ -66,7 +69,13 @@ public void serviceInit(Map<String, String> properties, boolean auxMode) {
...
 
-    server = new JettyServer();
+    server =
+        new JettyServer() {
+          @Override
+          protected javax.servlet.Filter createAuthenticationFilter() {
```
jerryshao: Don't use the full qualified package name.

yuqi1129 (approved):
It looks good to me.

flaming-archer:
Hi [@jerryshao](https://github.com/jerryshao), @yuqi1129 

Since all checks have passed and [@yuqi1129](https://github.com/yuqi1129) has approved the changes, could you please take another look and merge this PR when you have a moment?

Thank you!

yuqi1129:
@jerryshao 

Do you have any further comments on this PR?

PR #11559, opened 2026-06-10
[#11534] fix(spark-connector): GravitinoGlueCatalog does not invalidate Iceberg SparkCatalog cache after table mutations

### LINKED ISSUE #11534: [Bug report] GravitinoGlueCatalog does not invalidate Iceberg SparkCatalog cache after ALTER TABLE, causing stale schema on read
### Version

main branch

### Describe what's wrong

When using the Gravitino Spark connector with a Glue catalog, `ALTER TABLE ADD COLUMNS` on an Iceberg table succeeds on the server side, but the subsequent `SELECT` that references the new column fails with `IllegalStateException: Couldn't find <column> in [<old columns>]`.

The root cause is that `GravitinoGlueCatalog` maintains two separate catalogs:
- `sparkCatalog` → `HiveTableCatalog` (for non-Iceberg tables)
- `icebergGlueCatalog` → Iceberg `SparkCatalog` (for Iceberg tables, lazily created)

`BaseCatalog.alterTable` only calls `sparkCatalog.invalidateTable(ident)`, which invalidates the HiveTableCatalog cache. The Iceberg `SparkCatalog`'s internal `CachingCatalog` (enabled by default) is never invalidated. On the next `loadSparkTable`, the Iceberg catalog returns a stale table with the pre-ALTER schema. `SparkIcebergTable.schema()` then reports the updated schema (from the fresh Gravitino table), but `newScanBuilder()` delegates to the stale Iceberg `SparkTable` with the old schema, causing Spark's `BindReferences` to fail.

The same issue affects `dropTable`, `purgeTable`, `renameTable`, and `invalidateTable` for Iceberg tables in the Glue catalog.

```
java.lang.IllegalStateException: Couldn't find age#240 in [id#238,name#239]
at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1.applyOrElse(BoundAttribute.scala:80)
at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1.applyOrElse(BoundAttribute.scala:73)
at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$transformDownWithPruning$1(TreeNode.scala:461)
at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(origin.scala:76)
at org.apache.spark.sql.catalyst.trees.TreeNode.transformDownWithPruning(TreeNode.scala:461)
at org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:437)
at org.apache.spark.sql.catalyst.trees.TreeNode.transform(TreeNode.scala:405)
at org.apache.spark.sql.catalyst.expressions.BindReferences$.bindReference(BoundAttribute.scala:73)
at org.apache.spark.sql.catalyst.expressions.BindReferences$.$anonfun$bindReferences$1(BoundAttribute.scala:94)
at scala.collection.immutable.List.map(List.scala:297)
at org.apache.spark.sql.catalyst.expressions.BindReferences$.bindReferences(BoundAttribute.scala:94)
at org.apache.spark.sql.execution.ProjectExec.doConsume(basicPhysicalOperators.scala:69)
```

### How to reproduce

1. Create a Gravitino Glue catalog with Iceberg support via the Spark connector
2. Create an Iceberg table: `CREATE TABLE catalog.db.t (id INT, name STRING) USING iceberg`
3. Insert a row: `INSERT INTO catalog.db.t VALUES (1, 'Alice')`
4. Add a column: `ALTER TABLE catalog.db.t ADD COLUMNS (age INT)`
5. Query the new column: `SELECT id, name, age FROM catalog.db.t`
6. Step 5 fails with `IllegalStateException: Couldn't find age#<N> in [id#<N>,name#<N>]`

This reproduces on the main branch with Spark 3.5 and Iceberg 1.11.0.

### Additional context

The regular `GravitinoIcebergCatalog` does not have this bug because its `sparkCatalog` IS the Iceberg `SparkCatalog`, so `BaseCatalog.alterTable` → `sparkCatalog.invalidateTable(ident)` correctly evicts the cache. The Glue catalog is unique in that it routes Iceberg loads through a second, separate catalog instance that is never invalidated.

Suggested fix: Override `invalidateTable` in `GravitinoGlueCatalog` to also call `icebergGlueCatalog.invalidateTable(ident)` when the table is an Iceberg table (or unconditionally, since invalidation on a non-existent entry is a no-op).

diqiu50:

Override `invalidateTable` in `GravitinoGlueCatalog` to also clear `icebergGlueCatalog`'s cache. Change `BaseCatalog` mutation methods to call `invalidateTable(ident)` via virtual dispatch so the override takes effect.

`icebergGlueCatalog`'s internal `CachingCatalog` was never invalidated after table mutations, causing stale schema reads and `IllegalStateException: Couldn't find <col> in [...]` after `ALTER TABLE ADD COLUMNS`.

Fix #11534

Yes. Stale schema error on Iceberg Glue tables after ALTER/DROP/PURGE/RENAME is fixed.

- Unit tests in `TestGravitinoGlueCatalog`
- Integration test `testIcebergAlterTableAddColumnCacheInvalidation` in `SparkGlueCatalogIT`, verified against real AWS Glue + S3 (Spark 3.5)

### DIFF
--- a/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
+++ b/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/catalog/BaseCatalog.java
@@ -279,7 +279,7 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
             .map(sparkTableChangeConverter::toGravitinoTableChange)
             .toArray(org.apache.gravitino.rel.TableChange[]::new);
     try {
-      sparkCatalog.invalidateTable(ident);
+      invalidateTable(ident);
       org.apache.gravitino.rel.Table gravitinoTable =
           gravitinoCatalogClient
               .asTableCatalog()
@@ -301,7 +301,7 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
 
   @Override
   public boolean dropTable(Identifier ident) {
-    sparkCatalog.invalidateTable(ident);
+    invalidateTable(ident);
     return gravitinoCatalogClient
         .asTableCatalog()
         .dropTable(NameIdentifier.of(getDatabase(ident), ident.name()));
@@ -309,7 +309,7 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
 
   @Override
   public boolean purgeTable(Identifier ident) {
-    sparkCatalog.invalidateTable(ident);
+    invalidateTable(ident);
     return gravitinoCatalogClient
         .asTableCatalog()
         .purgeTable(NameIdentifier.of(getDatabase(ident), ident.name()));
@@ -354,7 +354,7 @@ public abstract class BaseCatalog implements TableCatalog, SupportsNamespaces, F
     org.apache.gravitino.rel.TableChange rename =
         org.apache.gravitino.rel.TableChange.rename(newIdent.name());
     try {
-      sparkCatalog.invalidateTable(oldIdent);
+      invalidateTable(oldIdent);
       gravitinoCatalogClient
           .asTableCatalog()
           .alterTable(NameIdentifier.of(getDatabase(oldIdent), oldIdent.name()), rename);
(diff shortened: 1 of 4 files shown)


roryqi (approved):
LGTM.

PR #11579, opened 2026-06-10
[#11578] fix(lance): Gravitino Lance REST does not support describeTable(version), causing Spark Lance write failures

### LINKED ISSUE #11578: [Bug report] Gravitino Lance REST does not support describeTable(version), causing Spark Lance write failures in spark lance connector 0.5.x
**Version**
main branch

**Describe what's wrong**
When Spark Lance uses Gravitino Lance REST (`impl=rest`) against a namespace-backed table, the write/commit path can call `describeTable(version=N)` instead of latest-only `describeTable()`.

Gravitino Lance REST currently rejects that request with:
`Unsupported: Describing specific table version is not supported.`

As a result, INSERT/commit on existing tables fails even though `describeTable()` without a version works.

**Error message and/or stacktrace**
`Unsupported: Describing specific table version is not supported.`

**How to reproduce**
1. Configure a Lance catalog in Gravitino with `impl=rest` and create a table.
2. Use Spark Lance against the Gravitino REST endpoint with a namespace-backed table.
3. Run `INSERT INTO` on an existing table.
4. During commit, Lance Spark calls `namespace.describeTable(version=1)` and the request fails.

**Additional context**
Please see more details in https://github.com/lance-format/lance-spark/issues/609 , this seems more like a bug or improvement in lance side, but to keep compatibility with current lance connectors and considering most lance namespace impls ignore version when describe table, Suggest Gravitino lance side ignore version too.

FANNG1:
Remove the version guard from Lance table describe handling so `describeTable(version=...)` is treated the same as the latest-table describe path. Add a regression assertion that versioned describe succeeds and returns the same metadata as the latest describe call.

Spark Lance can send a versioned describe request during write/commit flows. Gravitino Lance REST currently rejected that request, which breaks namespace-backed writes even though latest-only describe works.

Fix: #11578

Versioned Lance `describeTable` requests are now accepted instead of failing. The latest describe response shape is unchanged.

Ran the Lance REST server integration test class covering `describeTable` behavior.

### DIFF
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -104,12 +104,6 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
       Optional<Long> version,
       boolean checkDeclared,
       boolean loadDetailedMetadata) {
-    if (!version.isEmpty()) {
-      throw new UnsupportedOperationException(
-          "Describing specific table version is not supported. It should be null to indicate the"
-              + " latest version.");
-    }
-
     ObjectIdentifier nsId = ObjectIdentifier.of(tableId, Pattern.quote(delimiter));
     Preconditions.checkArgument(
         nsId.levels() == 3, "Expected at 3-level namespace but got: %s", nsId.levels());
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -97,6 +97,13 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
     this.namespaceWrapper = namespaceWrapper;
   }
 
+  /**
+   * {@inheritDoc}
+   *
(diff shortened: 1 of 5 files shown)


yuqi1129:
Is this problem an urgent one? Should we include it in 1.3.0?

FANNG1:
> Is this problem an urgent one? Should we include it in 1.3.0?

Yes，Gravitino lance REST server couldn't run with  spark lance connector 0.5.x, I have not checked whether it affects other spark lance versions.

lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
yuqi1129, line 132: You may need to add some comments that `version` does not take effect, as Gravitino only provides the latest schema.

yuqi1129:
Several suggestions:
1. Since this problem exists in `0.5.x` or main, please update the Lance REST supported matrix to support lance-spark 0.5.1
2. It's also preferable to update `lanceSparkMatrixTest` to support 0.5.1

FANNG1:
Done. Both suggestions addressed:
1. Updated the Lance REST supported matrix in `docs/lance-rest-integration.md` to include `lance-spark` 0.5.1 for Gravitino 1.3.0.
2. Updated `lanceSparkMatrixTest` defaults in `.github/workflows/lance-compatibility-matrix-test.yml` to include 0.5.1 in both the `workflow_dispatch` input default and the `LANCE_SPARK_VERSIONS` env var.

yuqi1129 (approved):

mchades:
please update the PR title to meet the project template

PR #11597, opened 2026-06-11
[#10909] fix(client): Fix list/load inconsistency for schema and table names containing dots

### LINKED ISSUE #10909: [Improvement] Support dotted object names consistently across list/load operations
Please improve consistency for object names containing dots (`.`), so list and load operations behave consistently.

Current behavior (verified with MySQL underlying names containing dots):
- A schema/database name like `db_xxx.nested` can appear in listing results.
- But loading the same object can fail in subsequent operations.
- Similar inconsistency is also observable for table loading when schema name contains dots.

This breaks the expected contract that objects returned by `list*` should be loadable by corresponding `load*` calls.

### How should we improve?
Expected behavior:
- If an object appears in `list*` results, loading that same object with `load*` should succeed.
- Dots inside a single logical name segment should not be treated as hierarchy delimiters.

Suggested scope:
- Cover schema and table paths first.
- Review related object paths (topic/fileset/model/function, etc.) for the same consistency requirement.

Reproduction reference (MySQL-based):
1. Create schema/database with dot in name: `CREATE DATABASE `db_xxx.nested``
2. `listSchemas()` includes `db_xxx.nested`
3. Loading the same schema fails
4. Creating/loading table under that schema can also hit inconsistency

mchades:

Fix the `list*`/`load*` inconsistency for MySQL schemas and tables whose names contain dots (e.g. a MySQL database named `db.v2` created externally with `` CREATE DATABASE `db.v2` ``).

- `GenericSchema`: replace `MetadataObjects.of(String parent, String name, Type)` with `MetadataObjects.of(List, Type)`, consistent with `GenericFileset`, `GenericTopic`, and `GenericModel`.
- `RelationalTable`: replace `MetadataObjects.parse(tableFullName(...), TABLE)` with `MetadataObjects.of(List, TABLE)`, and remove the now-unused `tableFullName()` helper and `DOT_JOINER` field.

Fix: #10909

`MetadataObjects.of(String parent, String name, Type)` joins parent and name with `.` then re-parses by splitting on `.`. When `name` itself contains a dot (e.g. `"db.v2"`), the split produces too many segments and the length check throws `IllegalArgumentException`:

```
IllegalArgumentException: If the type is SCHEMA, the length of names must be 2
  at GenericSchema.<init>(GenericSchema.java:48)
  at BaseSchemaCatalog.loadSchema(BaseSchemaCatalog.java:216)
```

`listSchemas()` had already returned the dotted name successfully, so `loadSchema()` crashing breaks the `list*`/`load*` contract.

No API changes. Bug fix only: `loadSchema` and `loadTable` now correctly handle names containing dots, consistent with what `listSchemas`/`listTables` already returns.

Added `testObjectNamesWithDots()` to `CatalogMysqlIT` (`@Tag("gravitino-docker-test")`). The test:
1. Creates a MySQL database with a dot in its name directly via JDBC.
2. Asserts `listSchemas()` returns that schema.
3. Asserts `loadSchema()` succeeds for the same name.
4. Creates a table in that schema directly via JDBC.
5. Asserts `listTables()` returns the table.
6. Asserts `loadTable()` succeeds for the same name.

### DIFF
--- a/catalogs/catalog-jdbc-mysql/src/test/java/org/apache/gravitino/catalog/mysql/integration/test/CatalogMysqlIT.java
+++ b/catalogs/catalog-jdbc-mysql/src/test/java/org/apache/gravitino/catalog/mysql/integration/test/CatalogMysqlIT.java
@@ -2380,4 +2380,50 @@ public class CatalogMysqlIT extends BaseIT {
       }
     }
   }
+
+  @Test
+  void testObjectNamesWithDots() {
+    // Create a MySQL database with a dot in its name directly (bypassing Gravitino)
+    String dottedSchemaName = GravitinoITUtils.genRandomName("db") + ".nested";
+    mysqlService.executeQuery("CREATE DATABASE `" + dottedSchemaName + "`");
+    try {
+      // listSchemas should include the schema with dot in name
+      String[] listedSchemas = catalog.asSchemas().listSchemas();
+      Set<String> schemaNames = Sets.newHashSet(listedSchemas);
+      Assertions.assertTrue(
+          schemaNames.contains(dottedSchemaName),
+          "listSchemas should return schema with dot in name: " + dottedSchemaName);
+
+      // loadSchema must succeed for a name that appeared in listSchemas
+      Schema loadedSchema = catalog.asSchemas().loadSchema(dottedSchemaName);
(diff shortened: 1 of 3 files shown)


roryqi (approved):
LGTM

PR #11617, opened 2026-06-12
[#11582] feat(iceberg): Support ETag / If-None-Match conditional loading on loadView endpoint

### LINKED ISSUE #11582: [Improvement] Support ETag / If-None-Match conditional loading on loadView endpoint
`loadTable` implements ETag-based conditional loading (`If-None-Match`, `generateETag`, `buildResponseWithETag`, returning `304 Not Modified`), but `loadView` returns the full response unconditionally with no ETag support. This is an inconsistency in spec coverage and misses a caching/bandwidth optimization that the Iceberg REST spec allows for views.

### How should we improve?
Add ETag generation and `If-None-Match` handling to `loadView`, reusing the existing helpers used by `loadTable` where possible. Return `304 Not Modified` when the client's ETag matches, and include the `ETag` header on `200` responses. Add tests for match/no-match/absent-header cases.

Related: spec-compliance epic #10547.

laserninja:

Add ETag / If-None-Match conditional loading support to the Iceberg REST loadView, createView, and replaceView endpoints, extending the existing ETag pattern from loadTable to views.

**Changes:**
- **loadView**: Generate ETag from view metadata file location, accept `If-None-Match` header, return `304 Not Modified` when ETag matches
- **createView / replaceView**: Return `ETag` header in response
- **IcebergRESTUtils**: Add `buildViewResponseWithETag()` helpers and extract `etagMatches()` as a shared public utility (previously private in `IcebergTableOperations`)
- **IcebergTableOperations**: Refactored private `etagMatches()` to delegate to `IcebergRESTUtils.etagMatches()`

The loadTable endpoint already supports ETag-based conditional loading (returning 304 Not Modified), but the loadView endpoint does not. This causes unnecessary data transfer when clients poll view metadata that has not changed.

Fix: #11582

Yes. The loadView endpoint now:
- Returns an `ETag` header derived from the view metadata file location
- Accepts `If-None-Match` header and returns `304 Not Modified` when the ETag matches

The createView and replaceView endpoints also now return `ETag` headers in their responses.

Added 8 new parameterized test methods to `TestIcebergViewOperations`:
- `testLoadViewReturnsETag` — verifies ETag header is present on load
- `testLoadViewReturns304WhenETagMatches` — verifies 304 with matching If-None-Match
- `testLoadViewReturns200WhenETagDoesNotMatch` — verifies 200 with non-matching ETag
- `testLoadViewETagConsistency` — verifies same ETag on repeated loads
- `testLoadViewETagChangesAfterReplace` — verifies ETag changes after view replace
- `testCreateViewReturnsETag` — verifies ETag on create response
- `testCreateViewETagMatchesLoadViewETag` — verifies create and load ETags match
- `testReplaceViewReturnsETag` — verifies ETag on replace response

All 36 tests pass (0 failures, 0 errors, 0 skipped). Verified with:
```
./gradlew :iceberg:iceberg-rest-server:test --tests "org.apache.gravitino.iceberg.service.rest.TestIcebergViewOperations" -PskipITs
```


roryqi:
Iceberg rest spec seems that view doesn't contain `etag`.

laserninja:
True that the Iceberg REST spec currently defines ETag/If-None-Match only for loadTable, not for loadView.

This PR extends the same pattern to views as a Gravitino-level enhancement. The rationale:

1. Views have the same metadataFileLocation property that tables use for ETag generation, so the mechanism is identical
2. ETag/If-None-Match is a standard HTTP caching mechanism (RFC 7232) that can apply to any resource, it's not table-specific semantics
3. It's fully backward-compatible, clients that don't send If-None-Match get the exact same response as before
4. It reduces unnecessary data transfer for clients polling view metadata that hasn't changed

That said, if the project prefers to stay strictly within the Iceberg REST spec for views, I understand. Let me know how you'd like to proceed, happy to adjust.

roryqi:
We should follow the Iceberg spec. Although we implement the feature, the open source clients like Spark connector can't benefit this feature. It's not meaningful.

laserninja:
Sounds good, will close it

PR #11618, opened 2026-06-12
[#11230] feat(server): Add HTTP-level audit logging via HttpAuditFilter

### LINKED ISSUE #11230: [Improvement] Support HTTP access log in Gravitino server

Currently, Gravitino's audit log is dispatched at the application (dispatcher) layer, which means any HTTP request that fails before reaching Jersey — malformed JSON, unknown routes, authentication rejection at the Jetty level — produces no record whatsoever. There is no way to observe these requests after the fact.

An HTTP-level access log (similar to Nginx/Apache access logs) would provide a complete, unfiltered record of every request the server receives, regardless of whether it reaches the application layer. This should be supported across all three servers: the main Gravitino server, the Iceberg REST server, and the Lance REST server.

### How should we improve?

Jetty 9.4 (already used by all three servers via `JettyServer` in `server-common`) natively supports access logging via `NCSARequestLog`, shipped in `jetty-server` with no additional dependencies required.

The implementation involves:

1. Add config entries to `JettyServerConfig` under `accessLog.*`:
   - `enabled` (default: `false`)
   - `path` — log file path supporting date-rolling pattern (e.g. `gravitino-access-yyyy_mm_dd.log`)
   - `retainDays` (default: `30`)

2. In `JettyServer.initialize()`, wire up Jetty's built-in `NCSARequestLog` via `server.setRequestLog(...)` when enabled.

Since all three servers each own a separate `JettyServer` instance and independently load their own `JettyServerConfig`, each server can be configured and enabled independently. The change is made once in `JettyServer` and applies to all three.

The access log complements the existing audit log: the audit log provides rich business context (operation type, resource, result count) while the access log provides complete HTTP-level coverage including requests that never reach the audit framework.

jerryshao:

Adds an `HttpAuditFilter` servlet filter registered on the main Gravitino server
(`/api/*`), Iceberg REST server (`/iceberg/*`), and Lance REST server (`/lance/*`).
The filter emits `HttpRequestFailureEvent` for every HTTP 4xx/5xx response when no
operation-layer failure event was already dispatched on the same thread, closing the
audit coverage gap for authentication failures, malformed requests, and unknown routes.

A new `AuthorizationDenialFailureEvent` is dispatched by
`GravitinoInterceptionService` for executor denials and `ForbiddenException`,
carrying the authenticated user, resource identifier, intercepted method name, and
authorization expression for rich audit attribution. `NoSuchMetalakeException` (a
resource-not-found masquerading as 403) intentionally falls through to the generic
HTTP-level event to avoid information disclosure.

Double-logging prevention: `EventBus.dispatchFailureEvent()` sets a
`RequestContext.operationFailureFired` flag for every non-`HttpRequestFailureEvent`
failure; `HttpAuditFilter` checks the flag in its `finally` block and skips emission
if already set.

Fix: #11230

HTTP-layer failures (401 auth rejections, 400 malformed JSON, 404 unknown routes,
403 authorization denials) currently produce no audit record. Enterprise users running
Gravitino in regulated environments require a complete audit trail — including failed
access attempts — to satisfy compliance requirements.

- New audit log entries for all HTTP 4xx/5xx failures with fields:
  `user`, `opType` (`UNKNOWN` or `AUTHORIZATION_DENIAL`), `identifier`, `status`,
  `source` (`GRAVITINO_SERVER` / `GRAVITINO_ICEBERG_REST_SERVER` / `GRAVITINO_LANCE_REST_SERVER`),
  `remoteAddr`, and `customInfo` (`http.method`, `http.uri`, `http.status` for HTTP
  events; `auth.method`, `auth.expression` for authorization denial events).
- New `OperationType.AUTHORIZATION_DENIAL` and `AuditLog.Operation.AUTHORIZATION_DENIAL`
  enum values in the public audit API.
- New `EventSource.GRAVITINO_LANCE_REST_SERVER` enum value.
- New public event classes: `HttpRequestFailureEvent`, `AuthorizationDenialFailureEvent`.

- Unit tests: `TestHttpAuditFilter` (6 cases), `TestEventBus` (2 new flag-marking
  cases), `TestGravitinoInterceptionService` (3 new cases for executor denial,
  `ForbiddenException`, and `NoSuchMetalakeException` paths).
- Coverage tests: `TestOperation`, `TestCompatibilityUtils`, `TestSimpleAuditLogV2`,
  `TestAuthorizationDenialFailureEvent`, `TestHttpRequestFailureEvent`.
- Integration tests: `HttpAuditFilterIT`, `HttpAuditAuthzDenialIT`.
- Manual end-to-end: all 18 test cases (TC-A1–A4, TC-B1–B5, TC-C1–C8, TC-D1–D2)
  verified against the running distribution.


server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
```diff
@@ -0,0 +1,275 @@
...
+  /**
+   * Constructs an {@code HttpAuditFilter}.
+   *
+   * @param eventBus the event bus used to dispatch {@link HttpRequestFailureEvent}s; may be {@code
+   *     null}, in which case the filter is a pass-through no-op (useful when no audit listener is
+   *     configured).
+   * @param eventSource identifies which server this filter instance is installed on; included in
+   *     every emitted {@link HttpRequestFailureEvent}.
+   */
+  public HttpAuditFilter(@Nullable EventBus eventBus, EventSource eventSource) {
```
yuqi1129, line 105: Use optional instead?

server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
```diff
@@ -0,0 +1,275 @@
...
+  private String resolveUser(HttpServletRequest request) {
+    Object principalObj =
+        request.getAttribute(AuthConstants.AUTHENTICATED_PRINCIPAL_ATTRIBUTE_NAME);
+    if (principalObj instanceof Principal) {
+      return ((Principal) principalObj).getName();
+    }
+    // The attribute is only set when authentication succeeds (AuthenticationFilter calls
+    // request.setAttribute before chain.doFilter). Absent attribute means auth was rejected
+    // before any principal was established. PrincipalUtils.getCurrentUserName() must not be used
+    // here — it returns "anonymous" when no Subject.doAs is active, making a rejected request
```
yuqi1129, line 225: The comments says `It returns anonymous` but we got `unknown`.

yuqi1129 (dismissed):

yuqi1129:
Sorry, I mistakenly clicked the "approve" button instead of the "comment" button.

server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
```diff
@@ -0,0 +1,275 @@
...
+      return;
+    }
+
+    HttpServletRequest httpRequest = (HttpServletRequest) request;
+    // Defensive cleanup at request entry in case a pooled thread leaked stale state.
+    RequestContext.resetOperationFailureFired();
+    if (isHealthCheckRequest(httpRequest)) {
+      chain.doFilter(request, response);
+      return;
+    }
```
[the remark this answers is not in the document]
jerryshao: Fixed — the health-check early-return path is now wrapped in a `try/finally` that calls `RequestContext.resetOperationFailureFired()`, matching the cleanup performed by the main path.

core/src/main/java/org/apache/gravitino/listener/EventBus.java
```diff
@@ -129,17 +131,28 @@ private void dispatchPostEvent(Event postEvent) {
...
   /**
-   * Dispatches a failure failureEvent to listeners, swallowing all exceptions to preserve the
-   * original error.
+   * Dispatches a failure event to listeners, swallowing all exceptions to preserve the original
+   * error.
    *
    * <p>When the primary operation fails, we want to notify listeners, but listener exceptions must
    * NOT propagate and mask the original failure. This method catches and logs all exceptions
    * (including {@link RuntimeException}, {@link Error}, etc.) to ensure the original exception can
    * be properly thrown to the caller.
```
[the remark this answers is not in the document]
jerryshao: Fixed — the Javadoc now says 'catches and logs all `Exception`s' to match the actual `catch (Exception e)` in the implementation. Broadening to `Throwable`/`Error` is intentionally not done here: `Error` indicates a JVM-level problem that should propagate up rather than be silently swallowed.

server/src/test/java/org/apache/gravitino/server/web/filter/TestGravitinoInterceptionService.java
```diff
@@ -250,6 +267,180 @@ public void testEmptyExpressionSkipsAuthorization() throws Throwable {
...
+      Response response = (Response) interceptor.invoke(invocation);
+
+      assertEquals(Response.Status.FORBIDDEN.getStatusCode(), response.getStatus());
+      // Event dispatched
+      ArgumentCaptor<org.apache.gravitino.listener.api.event.Event> captor =
+          ArgumentCaptor.forClass(org.apache.gravitino.listener.api.event.Event.class);
+      verify(mockEventBus).dispatchEvent(captor.capture());
+      Assertions.assertInstanceOf(AuthorizationDenialFailureEvent.class, captor.getValue());
+      AuthorizationDenialFailureEvent event = (AuthorizationDenialFailureEvent) captor.getValue();
+      assertEquals("tester", event.user());
```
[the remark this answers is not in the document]
jerryshao: Fixed — both captors are now typed directly to `ArgumentCaptor<AuthorizationDenialFailureEvent>`, removing the FQN and the redundant `assertInstanceOf` + cast.

server/src/test/java/org/apache/gravitino/server/web/filter/TestGravitinoInterceptionService.java
```diff
@@ -250,6 +267,180 @@ public void testEmptyExpressionSkipsAuthorization() throws Throwable {
...
+      Assertions.assertFalse(RequestContext.isOperationFailureFired());
+      Response response = (Response) interceptor.invoke(invocation);
+
+      assertEquals(Response.Status.FORBIDDEN.getStatusCode(), response.getStatus());
+      ArgumentCaptor<org.apache.gravitino.listener.api.event.Event> captor =
+          ArgumentCaptor.forClass(org.apache.gravitino.listener.api.event.Event.class);
+      verify(mockEventBus).dispatchEvent(captor.capture());
+      Assertions.assertInstanceOf(AuthorizationDenialFailureEvent.class, captor.getValue());
+      AuthorizationDenialFailureEvent event = (AuthorizationDenialFailureEvent) captor.getValue();
+      assertEquals("outsider", event.user());
```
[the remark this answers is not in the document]
jerryshao: Fixed — same as above.

core/src/main/java/org/apache/gravitino/listener/api/event/server/AuthorizationDenialFailureEvent.java
```diff
@@ -0,0 +1,142 @@
...
+        user,
+        accessMetadataName,
+        new ForbiddenException(
+            "Authorization denied for user '%s' on operation '%s'", user, methodName));
+    this.methodName = methodName;
+    this.expression = expression != null ? expression : "";
+    this.explicitEventSource = eventSource;
+  }
+
+  /** Returns {@link OperationType#AUTHORIZATION_DENIAL}. */
```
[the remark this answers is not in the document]
jerryshao: Fixed — replaced `ForbiddenException` with a lightweight `AuthorizationDenialException` inner class (modelled after `HttpRequestException` in `HttpRequestFailureEvent`) that suppresses stack trace collection via the 4-arg `RuntimeException` constructor. Authorization denials can be high-volume so the overhead saving is meaningful.

server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
jerryshao: Fixed — the `@Nullable EventBus` field is now `Optional<EventBus>`, set in the constructor via `Optional.ofNullable(eventBus)`. The null-check in `doFilter` is replaced with `eventBus.isEmpty()` and the dispatch uses `eventBus.get()`.

server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
jerryshao: Fixed — the comment now correctly says 'Returns `unknown` instead' to match what the code actually returns when no principal attribute is set on the request.

server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
```diff
@@ -0,0 +1,295 @@
...
+      chainException = t;
+      // Promote to 500 so the finally block emits an event for this escaped exception.
+      if (wrappedResponse.getCapturedStatus() < 400) {
+        wrappedResponse.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+      }
+    } finally {
+      try {
+        if (!RequestContext.isOperationFailureFired()) {
+          int status = wrappedResponse.getCapturedStatus();
+          if (status >= 400) {
```
yuqi1129, line 173: You have increased the status that less than `400`, so when the condition wil be `false`?

jerryshao:
@mchades @yuqi1129 can you please help to trigger the copilot review. Mine just ran out of quota.

server-common/src/main/java/org/apache/gravitino/server/web/HttpAuditFilter.java
jerryshao: What's the meaning of this problem, can you describe more? To promote to 500 is only for the failure case, if the operation succeeds, the http code will be less than 400, and it will not trigger an event.

yuqi1129: I see.

yuqi1129 (commented):
I have no more comments

mchades:
**[P2] Custom filters can bypass audit trail**

**Files affected:**
- `server/src/main/java/org/apache/gravitino/server/GravitinoServer.java` line 190-193
- `iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/RESTService.java` line 208-212  
- `lance/lance-rest-server/src/main/java/org/apache/gravitino/lance/LanceRESTService.java` line 95-98

**Issue:** HttpAuditFilter is registered AFTER `server.addCustomFilters()`. A custom filter that returns 401/403/5xx without calling `chain.doFilter()` can bypass HttpAuditFilter entirely, emitting no audit event.

**Recommended fix:** Reorder filters so HttpAuditFilter comes BEFORE custom filters.

For the main server (GravitinoServer), suggested order:
1. RequestContextFilter (ThreadLocal initialization)
2. HttpAuditFilter  
3. Custom filters
4. VersioningFilter
5. System filters

Iceberg and Lance REST servers should follow the same pattern: HttpAuditFilter before custom filters.

mchades:
**Residual risk: Missing integration test for uncaught servlet errors**

**Location:** `server-common/src/test/java/org/apache/gravitino/server/web/TestHttpAuditFilter.java` around line 408

**Issue:** While the unit tests cover most scenarios, there's no integration test verifying correct audit event handling for:
- Uncaught servlet errors (DispatcherType.ERROR)
- Exception escaping without being caught by filters
- Ensuring no duplicate or missing 500 audit events during error dispatch

**Recommendation:** Add an integration test (`HttpAuditFilterIT` or similar) that:
1. Triggers a true servlet error (e.g., JSP compilation error, uncaught exception in servlet initialization)
2. Verifies exactly one audit event is emitted
3. Confirms DispatcherType.ERROR requests don't bypass or double-log events

This prevents the audit trail from missing or duplicating events when errors bubble up from the servlet layer.

jerryshao:
OK, I will fix these two issues soon.

jerryshao:
@mchades please help to review again.

mchades (approved):

PR #11620, opened 2026-06-12
[#11565][#11566] feat(mcp-server): add --token CLI param and inject Bearer token into Gravitino REST calls

### LINKED ISSUE #11565: [Subtask] feat(mcp-server): add --token CLI parameter and auth fields to Setting
### Describe the subtask

Add authentication configuration to `Setting` and the CLI so the MCP server can accept a Bearer token at startup.

**Changes:**
- `mcp_server/core/setting.py`: add `token: str = ""` field to `Setting` dataclass
- `mcp_server/main.py`: add `--token` CLI argument (also readable from env var `GRAVITINO_TOKEN`); mask token in log output

**Acceptance:**
- `uv run mcp_server --metalake ml --token abc123` starts without error
- Log output does not contain the token value

### Parent issue

TBD (EPIC to be linked after creation)

### LINKED ISSUE #11566: [Subtask] feat(mcp-server): inject Bearer token into every Gravitino REST call
### Describe the subtask

Propagate the authenticated principal's Bearer token from the MCP server configuration through to every outbound HTTP call to Gravitino, replacing the current anonymous `httpx.AsyncClient`.

**Changes:**
- `mcp_server/client/plain/plain_rest_client_operation.py`: accept `token` parameter; set `Authorization: Bearer <token>` header on the shared `httpx.AsyncClient` when token is non-empty
- `mcp_server/client/factory.py`: pass `token` through `create_rest_client`
- `mcp_server/core/context.py`: read `setting.token` and pass to factory; reject startup (raise) if token is required but absent

**Acceptance:**
- A request with a valid token reaches Gravitino with `Authorization: Bearer <token>` header
- A missing token when auth is required raises an error at startup rather than silently running as anonymous

### Parent issue

TBD (EPIC to be linked after creation)

yuqi1129:

- Add a `token` field to `Setting`, masking its value in `__str__` to avoid leaking it into logs.
- Add a `--token` CLI argument to `main.py`, falling back to the `GRAVITINO_TOKEN` environment variable.
- `PlainRESTClientOperation` accepts a token and sets the `Authorization: Bearer` header on the httpx client when non-empty.
- `RESTClientFactory.create_rest_client` passes the token through, and `GravitinoContext` propagates `setting.token` to the factory.
- Update `MockOperation` to accept the token parameter for test compatibility.

The MCP server currently has no way to pass an authentication token to Gravitino REST calls, so it cannot work against a Gravitino server with auth enabled.

Fix: #11565
Fix: #11566

Yes. Adds a new `--token` CLI option (and `GRAVITINO_TOKEN` env var fallback) to the MCP server.

Added 7 unit tests in `mcp-server/tests/unit/test_auth_flow.py` covering token injection, masking, and context propagation.

### DIFF
--- a/mcp-server/mcp_server/client/factory.py
+++ b/mcp-server/mcp_server/client/factory.py
@@ -29,7 +29,7 @@ class RESTClientFactory:
 
     @classmethod
     def create_rest_client(
-        cls, metalake_name: str, uri: str
+        cls, metalake_name: str, uri: str, token: str = ""
     ) -> "PlainRESTClientOperation":
         """
         Create a new rest client instance with the specified parameters.
@@ -37,11 +37,12 @@ class RESTClientFactory:
         Args:
             metalake_name: Name of the metalake
             uri: URI of the Gravitino server endpoint
+            token: Bearer token for Authorization header (empty = anonymous)
 
         Returns:
             New instance of the configured rest client class
         """
-        return cls._rest_client_class(metalake_name, uri)
+        return cls._rest_client_class(metalake_name, uri, token)
 
     @classmethod
     def set_rest_client(cls, rest_client_class: type) -> None:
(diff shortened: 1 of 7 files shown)


yuqi1129:
Folding this into #11622. The token support here is reworked there to forward the raw `Authorization` header (supporting both OAuth2 `Bearer` and Gravitino simple-auth `Basic`) instead of assuming Bearer, so the MCP auth work now lives as a single foundational PR (#11622) with the write tools stacked on top in #11621. Closing in favor of #11622.

PR #11621, opened 2026-06-12
[#11567][#11574] feat(mcp-server): write tools for tag, catalog, schema, table, model, topic, fileset

yuqi1129:
### What changes were proposed in this pull request?

This is the **write/modify** PR (second in the stack), adding governed write tools to the MCP server.

- **#11567**: enable tag write tools (create/alter/delete/associate), protected by Gravitino authorization.
- **#11574**: implement write operations for catalog, schema, table, model, topic, and fileset (create/alter/drop, plus register/link/delete for models).

Write tools are exposed by default; authorization is enforced by Gravitino core.

> **Stacked on #11622.** This PR depends on the authentication/authorization/audit
> foundation in #11622. Until #11622 merges, this PR's diff will also show those
> commits; review after #11622 or focus on the two write commits here. Merge order:
> #11622 first, then this PR.

### Why are the changes needed?

The MCP server previously exposed only read-only tools. These write tools let agents manage Gravitino metadata through MCP, with per-request authorization (from #11622) ensuring each principal's permissions are enforced.

Fix: #11567
Fix: #11574

### Does this PR introduce _any_ user-facing change?

Yes. New MCP write tools for tag, catalog, schema, table, model, topic, and fileset.

### How was this patch tested?

`./gradlew :mcp-server:formatCheckPython :mcp-server:pylint` clean (pylint 10.00/10) and 123 unit tests pass.

PR #11622, opened 2026-06-12
[#11565][#11566][#11568][#11572][#11575] feat(mcp-server): authentication, per-request authorization and audit logging

### LINKED ISSUE #11565: [Subtask] feat(mcp-server): add --token CLI parameter and auth fields to Setting

### LINKED ISSUE #11566: [Subtask] feat(mcp-server): inject Bearer token into every Gravitino REST call

### LINKED ISSUE #11568: [Subtask] feat(mcp-server): structured per-tool-call audit logging attributed to principal
### Describe the subtask

Emit a structured audit record for every MCP tool invocation, capturing the calling principal, tool name, and allow/deny outcome. Records are written to `gravitino-mcp-audit.log` in JSON format and are integrated with the existing Gravitino audit subsystem so there is a single audit trail.

**Changes:**
- New `mcp_server/core/audit.py`: `AuditLogger` class; fields: `timestamp`, `principal`, `tool`, `outcome` (`allow`/`deny`), `error_type`
- `mcp_server/server.py`: add `AuditMiddleware` to the middleware chain; extract principal from `Authorization` header; record outcome after each tool call

**Acceptance:**
- A successful read tool call produces a JSON audit record with `outcome: allow` and the correct principal
- A denied write call produces a record with `outcome: deny`

### Parent issue

TBD (EPIC to be linked after creation)

### LINKED ISSUE #11572: [Subtask] feat(mcp-server): per-request token isolation for HTTP transport (multi-principal)
### Describe the subtask

In HTTP transport mode, extract the Bearer token from each incoming MCP HTTP request's `Authorization` header and use it for that request's Gravitino calls. Concurrent sessions with different principals must be fully isolated — one principal's identity must never leak into another's calls.

FastMCP 3.2.0 provides `get_http_request()` from `fastmcp.server.dependencies` which enables header access in middleware.

**Changes:**
- `mcp_server/server.py`: in HTTP transport mode, extract per-request `Authorization` header via `get_http_request()` and override the token used for that request's Gravitino calls
- `mcp_server/core/context.py`: support a per-request token that takes priority over the static startup token

**Acceptance:**
- Two concurrent HTTP sessions with different tokens each reach Gravitino with their own correct token
- Removing the per-request header falls back to the static startup token (if configured)

### Parent issue

TBD (EPIC to be linked after creation)

### LINKED ISSUE #11575: [Subtask] test(mcp-server): integration tests validating per-user authorization with live Gravitino
### Describe the subtask

With identity propagation (Task 1/2) and write tools enabled (Task 3/7), add integration tests that spin up a real Gravitino instance and verify that different principals get correctly differentiated responses through MCP — covering the three demo acceptance moments:

1. Two principals issue the same list call and get differently scoped results
2. A read-only principal attempts a write and is denied by Gravitino authorization
3. Both the reads and the denied write appear as audit records attributed to the correct principal

**Test approach:**

These tests require a running Gravitino server, so they should be tagged `@Tag("gravitino-docker-test")` (or the Python equivalent) and run in CI via Docker Compose. The recommended structure:

```
tests/integration/
├── conftest.py          # start MCP server in HTTP mode against a live Gravitino instance
├── test_authz_scoping.py  # Moment 1: two principals, different list results
├── test_write_denial.py   # Moment 2: read-only principal denied on write tool
└── test_audit_trail.py    # Moment 3: audit log records attributed to correct principal
```

**CI integration:**

Add a GitHub Actions job (or extend the existing docker-test matrix) that:
1. Starts Gravitino via Docker Compose with authorization enabled (Ranger or built-in policy engine)
2. Seeds two roles: `admin-role` (full access) and `readonly-role` (list-only on a specific catalog)
3. Creates two service tokens bound to each role
4. Starts the MCP server in HTTP transport mode with `--transport http`
5. Runs `pytest tests/integration/ -m gravitino_docker`

**Acceptance:**
- `test_authz_scoping`: `GET /catalogs` with admin token returns N catalogs; same call with readonly token returns only the permitted subset
- `test_write_denial`: `create_tag` with readonly token returns HTTP 403 from Gravitino, propagated as a MCP tool error
- `test_audit_trail`: `gravitino-mcp-audit.log` contains one `allow` record attributed to admin principal and one `deny` record attributed to readonly principal

### Parent issue

https://github.com/apache/gravitino/issues/11573

yuqi1129:

This is the **authentication / authorization / audit** foundation for the MCP server (the first PR in the stack). It folds in the earlier token PR (#11620) and reworks it so the Authorization header is forwarded **verbatim**, supporting both OAuth2 Bearer and Gravitino simple-auth (`Basic`).

- **#11565 / #11566**: `--token` CLI param (with `GRAVITINO_TOKEN` env fallback); token masked in `Setting.__str__` and kept out of `__repr__`.
- **#11575**: forward the raw `Authorization` header to Gravitino instead of assuming `Bearer`, so `Basic <base64(user:dummy)>` simple auth also works; plus a live-Gravitino authorization integration test wired into CI.
- **#11572**: per-request token isolation for HTTP transport (each request's identity is forwarded, not a shared server token); accept the `streamable-http` transport alias and serve the endpoint over TLS.
- **#11568**: structured per-tool-call audit logging attributed to the principal.

This PR contains **no write tools** — those live in the follow-up PR #11621, which is stacked on this one.

The MCP server had no way to authenticate to a Gravitino server with auth enabled, and no per-request identity propagation for multi-user HTTP deployments. The original Bearer-only approach (#11620) could not support Gravitino simple authentication; forwarding the raw header fixes that.

Fix: #11565
Fix: #11566
Fix: #11568
Fix: #11572
Fix: #11575

Yes:
1. New `--token` CLI option (and `GRAVITINO_TOKEN` env var).
2. HTTP transport forwards per-request `Authorization` headers (Bearer/Basic/Negotiate).
3. New `streamable-http` transport alias and TLS options.
4. Structured audit logs per tool call.

`./gradlew :mcp-server:formatCheckPython :mcp-server:pylint` clean (pylint 10.00/10) and 103 unit tests pass, plus a live-Gravitino authorization integration test running in CI.


mchades (commented):
can we split this PR into small ones?

yuqi1129:
> can we split this PR into small ones?

The core change is focus on the following sections:

image


Others are UT/IT or CI pipelines.

mcp-server/mcp_server/client/plain/plain_rest_client_operation.py
```diff
@@ -62,8 +62,23 @@
...
-        _rest_client = httpx.AsyncClient(base_url=uri)
+    def __init__(self, metalake_name: str, uri: str, authorization: str = ""):
+        """Create a REST client for one identity.
+
+        Args:
+            metalake_name: Name of the metalake.
+            uri: Gravitino server URI.
+            authorization: Full ``Authorization`` header value forwarded verbatim
+                on every request (for example ``"Bearer <token>"`` for OAuth2 or
+                ``"Basic <base64(user:dummy)>"`` for simple auth). Empty string
```
mchades: this also for basic auth, which we recently supported for local IDP, not only simple auth.

mcp-server/tests/integration/test_authz_e2e.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+async def _list_catalog_names(principal: str, mcp_url: str) -> set:
+    async with _client_for(principal, mcp_url) as client:
+        result = await client.call_tool("get_list_of_catalogs")
+    payload = json.loads(result.content[0].text)
+    return {entry["name"] for entry in payload}
+
+
+def test_moment1_authorization_scoped_discovery(
```
mchades: is the `moment1` in the method name is necessary?

yuqi1129: removed.

mchades (approved):

PR #11626, opened 2026-06-12
[#11588] fix(lance): Return leaf table names from REST ListTables

### LINKED ISSUE #11588: [Bug report] Lance REST server ListTables returns fully-qualified names, polluting Spark SHOW TABLES output
### Version

main branch

### Describe what's wrong

The Lance REST server's `ListTables` endpoint returns **fully-qualified** table names (`catalog{delimiter}schema{delimiter}table`) instead of the **leaf table names** required by the Lance Namespace spec.

As a result, a Spark client using the Lance namespace connector shows polluted table names in `SHOW TABLES` — each table name embeds the catalog and schema, e.g. `catalog.schema.my_table` instead of `my_table`.

### Root cause

`lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java` (`listTables`):

```java
List<String> tables =
    Arrays.stream(catalog.asTableCatalog().listTables(Namespace.of(schemaName)))
        .map(ident -> Joiner.on(delimiter).join(catalogName, schemaName, ident.name())) // <-- returns full qualified name
        .sorted()
        .collect(Collectors.toList());
```

The parent namespace is already conveyed by the request `id` (`catalog`, `schema`), so the response must only contain the child table names.

### Why it surfaces in Spark

The Lance Spark connector (`BaseLanceNamespaceSparkCatalog.listTables`) trusts the response strings as leaf names and wraps them directly:

```java
for (String table : response.getTables()) {
  identifiers.add(Identifier.of(namespace, table)); // table = "catalog.schema.tbl"
}
```

Spark then renders `Identifier.name()`, i.e. the full string returned by the server, producing the `catalog.schema.table` display.

### Spec & reference implementations

- Spec `ListTables` (`docs/src/spec.yaml`): *"List all child table **names** of the parent namespace `id`."*
- Reference impls return only the leaf name, e.g. Glue: `.forEach(t -> tables.add(t.name()))`.

(Note: the shared `ListTablesResponse` schema description mentioning a "full identifier in string form" refers to the recursive `/v1/table` (list-all-tables) endpoint, not the per-namespace `ListTables`.)

### How to reproduce

1. Start the Gravitino Lance REST server over a lakehouse catalog with a schema containing one or more tables.
2. Configure a Spark `lance` namespace catalog (`impl=rest`) pointing at the server.
3. Run `SHOW TABLES`.
4. Observe table names appear as `catalog.schema.table` instead of `table`.

### Expected behavior

`ListTables` should return only the leaf table names, so `SHOW TABLES` shows `table`.

### Additional context

Suggested fix: change the mapping to `.map(ident -> ident.name())` (and preserve sort order / use a proper list for the paged result).

bbiiaaoo:

This PR fixes the Lance REST per-namespace `ListTables` implementation to return child table names instead of fully-qualified `catalog.schema.table` strings.

It also tightens the related REST and Spark integration tests:
- `LanceRESTServiceIT` now expects `ListTables` to return only the leaf table name.
- `LanceSparkRESTServiceIT` now validates the raw `SHOW TABLES` table name instead of stripping namespace prefixes.

The Lance Namespace `ListTables` API lists child table names under the requested parent namespace. The current implementation returns fully-qualified names, causing Spark `SHOW TABLES` to display polluted table names such as `catalog.schema.table` instead of `table`.

Fix: #11588

Yes. The Lance REST per-namespace `ListTables` endpoint now returns leaf table names in the `tables` field.

The REST response shape is unchanged.

Ran:

```bash
./gradlew :lance:lance-rest-server:testClasses
./gradlew :lance:lance-rest-server:test --tests "org.apache.gravitino.lance.integration.test.LanceRESTServiceIT" -PskipDockerTests=false
./gradlew :lance:lance-rest-server:lanceSparkMatrixTest -PskipDockerTests=false

### DIFF
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceNameSpaceOperations.java
@@ -19,13 +19,13 @@
 
 package org.apache.gravitino.lance.common.ops.gravitino;
 
-import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -443,17 +443,14 @@ public class GravitinoLanceNameSpaceOperations implements LanceNamespaceOperatio
     String schemaName = nsId.levelAtListPos(1);
     List<String> tables =
         Arrays.stream(catalog.asTableCatalog().listTables(Namespace.of(schemaName)))
-            .map(ident -> Joiner.on(delimiter).join(catalogName, schemaName, ident.name()))
+            .map(ident -> ident.name())
             .sorted()
             .collect(Collectors.toList());
 
     PageUtil.Page page = PageUtil.splitPage(tables, pageToken, PageUtil.normalizePageSize(limit));
-    ListNamespacesResponse response = new ListNamespacesResponse();
-    response.setNamespaces(Sets.newHashSet(page.items()));
+    ListTablesResponse response = new ListTablesResponse();
+    response.setTables(new LinkedHashSet<>(page.items()));
     response.setPageToken(page.nextPageToken());
-
-    return new ListTablesResponse()
-        .tables(response.getNamespaces())
-        .pageToken(response.getPageToken());
+    return response;
   }
 }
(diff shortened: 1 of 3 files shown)


bbiiaaoo:
Hi @FANNG1  @yuqi1129 , I have opened this PR for #11588.

The fix is ready and the related REST/Spark tests passed. Could you please help review it when you have time? Thanks!

FANNG1:
Optional nit: in `LanceRESTServiceIT`, the size + contains assertions could be simplified to a direct expected-set comparison, e.g. compare `listResponse.getTables()` with `Set.of("table")`. The current assertions are also fine, so this is non-blocking.

FANNG1 (approved):

yuqi1129 (approved):

PR #11629, opened 2026-06-12
docs(auth): Add StarRocks Authorization Pushdown Design

zhaohehuhu:

The design proposes adding StarRocks authorization pushdown support by introducing a new authorization-starrocks provider. The proposed plugin would extend the existing JDBC authorization framework, translate Gravitino users, roles, securable objects, and privileges into StarRocks native authorization SQL, and define object/privilege mappings for catalog, schema, and table-level permissions.

Gravitino currently supports StarRocks metadata management through jdbc-starrocks, but authorization changes in Gravitino are not pushed into StarRocks.

The design is needed to align Gravitino’s unified authorization model with StarRocks’ native RBAC system.

No. It's a design doc.

no needed.

### DIFF
--- a/design-docs/starrocks-authorization-pushdown.md
+++ b/design-docs/starrocks-authorization-pushdown.md
@@ -162,7 +162,7 @@ authorization resources.
 
 ### 4.2 Ownership
 
-Gravitino supports object ownership, where the owner of an object implicitly has administrative privileges such as ALTER, DROP, and GRANT.
+Gravitino supports object ownership, where the owner of an object implicitly has administrative privileges such as ALTER and DROP.
 
 Since StarRocks does not provide a native object ownership model, the plugin should map Gravitino ownership to StarRocks privileges by granting ALL privileges on the corresponding resource to the owner. 
 This approach preserves the ownership semantics in Gravitino and ensures that owners retain full administrative control over their objects in StarRocks
--- a/design-docs/starrocks-authorization-pushdown.md
+++ b/design-docs/starrocks-authorization-pushdown.md
@@ -1,0 +1,248 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Design: StarRocks Authorization Pushdown
+
+## 1. Problem Statement and Goals
+
+### 1.1 Problem
+
+Apache Gravitino supports metadata management for StarRocks through the `jdbc-starrocks`
+catalog, but it does not yet push Gravitino authorization changes into StarRocks native
+authorization. When users create roles, grant privileges or assign roles to users or groups, 
+StarRocks does not automatically receive the corresponding users, roles, grants, revokes, 
+or ownership changes.
+
+This creates a gap between the Gravitino authorization model and the access control that
+StarRocks enforces when users query StarRocks directly. StarRocks already has a SQL-based
+authorization model with users, roles, catalog privileges, database privileges, and table
+privileges. Gravitino should translate supported Gravitino authorization operations into
+StarRocks SQL statements.
+
+### 1.2 Goals
+
+1. Implement a StarRocks authorization pushdown plugin based on the existing JDBC
+   authorization plugin framework.
+2. Translate supported Gravitino securable objects and privileges into StarRocks native
+   authorization objects and privileges.
+3. Keep the implementation aligned with the existing authorization provider model.
+4. Support user, group, role, privilege grant, privilege revoke pushdown for StarRocks catalogs.
+
+### 1.3 Non-goals
+
+1. This design does not add new Gravitino privilege types.
+2. This design does not implement StarRocks column-level privileges.
+3. This design does not implement StarRocks view, materialized view, resource, storage
+   volume, system, or function privileges.
+4. This design does not push down Gravitino deny privileges.
+5. This design does not manage StarRocks external catalogs.
+
+## 2. Background
+
+### 2.1 Gravitino Authorization Pushdown
+
+Gravitino's authorization pushdown mechanism lets a catalog load an `AuthorizationPlugin`
+based on the catalog property `authorization-provider`. The provider is loaded from an
+isolated package, and `BaseAuthorization.newPlugin(metalake, catalogProvider, config)`
+creates a catalog-specific plugin instance.
+
+The existing JDBC authorization base class, `JdbcAuthorizationPlugin`, already implements
+the common hook flow:
+
+1. Create or drop users.
+2. Create or drop roles.
+3. Grant or revoke roles to users and groups.
+4. Translate Gravitino securable objects into JDBC authorization objects.
+5. Generate and execute SQL statements for grants, revokes relative changes.
+
+
+### 2.2 StarRocks Authorization Model
+
+StarRocks 3.0 and later provide a role-based authorization model. The objects relevant to
+this design are:
+
+| StarRocks object | Relevant privileges |
+|------------------|---------------------|
+| Catalog          | `USAGE`, `CREATE DATABASE` |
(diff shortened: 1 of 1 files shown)


design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,248 @@
...
+
+| Gravitino object | StarRocks resource |
+|------------------|--------------------|
+| Metalake         | `default_catalog`, all databases, all tables |
+| Catalog          | `default_catalog`, all databases, all tables |
+| Schema           | Database `{schema}`, all tables in `{schema}` |
+| Table            | Table `{schema}.{table}` |
+
+
+### 4.2 Supported Privileges
```
roryqi: `Gravitino` has the concept owner. How to handle it?

zhaohehuhu: StarRocks does't have the concept of owner. @roryqi

zhaohehuhu: Could we grant all privileges (read and write) to the owner in StarRocks ?

design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,248 @@
...
+
+| Gravitino object | Gravitino privilege | StarRocks privilege                             | StarRocks resource                 |
+|------------------|---------------------|-------------------------------------------------|------------------------------------|
+| Catalog          | `USE_CATALOG`       | `USAGE`                                         | `default_catalog`                  |
+| Catalog          | `USE_SCHEMA`        | `USAGE`                                         | `default_catalog`                  |
```
roryqi, line 198: Why does use_schema map to `USAGE`?

design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,248 @@
...
+
+```sql
+REVOKE SELECT ON TABLE `db1`.`tbl1` FROM ROLE `catalog_user`;
+REVOKE INSERT ON ALL TABLES IN DATABASE `db1` FROM ROLE `catalog_user`;
+REVOKE CREATE TABLE ON DATABASE `db1` FROM ROLE `catalog_user`;
+REVOKE CREATE DATABASE ON CATALOG `default_catalog` FROM ROLE `catalog_user`;
+REVOKE USAGE ON CATALOG `default_catalog` FROM ROLE `catalog_user`;
+```
+
+## 6. References
```
roryqi, line 272: `Gravitino` supports Group now.

yuqi1129: I think we can explicitly throw an exception that `StarRocks` does not support the `group` concept.

design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,248 @@
...
+### 1.1 Problem
+
+Apache Gravitino supports metadata management for StarRocks through the `jdbc-starrocks`
+catalog, but it does not yet push Gravitino authorization changes into StarRocks native
+authorization. When users create roles, grant privileges or assign roles to users or groups, 
+StarRocks does not automatically receive the corresponding users, roles, grants, revokes, 
+or ownership changes.
+
+This creates a gap between the Gravitino authorization model and the access control that
+StarRocks enforces when users query StarRocks directly. StarRocks already has a SQL-based
```
yuqi1129, line 33: We may need to consider that users may change privileges without using Gravitino since they can query data from StarRocks directly, so how do we synchronize the privileges data?

design-docs/starrocks-authorization-pushdown.md
zhaohehuhu: OK.  We can throw an exception since StarRocks does not support the group concept.

design-docs/starrocks-authorization-pushdown.md
zhaohehuhu: The mapping should be removed or marked as unsupported, because StarRocks does not have corresponding schema-level USAGE privilege.

design-docs/starrocks-authorization-pushdown.md
zhaohehuhu: StarRocks is the single source of truth for authorization, while Gravitino periodically synchronizes privilege information from StarRocks. As a result, the two systems are eventually consistent.

zhaohehuhu: Since Gravitino leverages Ranger for authorization pushdown, we could consider adopting a similar design to address this scenario, if feasible.

design-docs/starrocks-authorization-pushdown.md
roryqi: You would better clarify these in the documents.

zhaohehuhu: Got it. Thanks!

design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,248 @@
...
+| Catalog          | `CREATE_TABLE`      | `CREATE TABLE`                                  | `default_catalog.*.*`              |
+| Catalog          | `SELECT_TABLE`      | `SELECT`                                        | `default_catalog.*.*`              |
+| Catalog          | `MODIFY_TABLE`      | `SELECT`, `INSERT`, `UPDATE`, `DELETE`, `ALTER` | `default_catalog.*.*`              |
+| Schema           | `USE_SCHEMA`        |                                                 |                                    |
+| Schema           | `CREATE_TABLE`      | `CREATE TABLE`                                  | `default_catalog.{schema}.*`       |
+| Schema           | `SELECT_TABLE`      | `SELECT`                                        | `default_catalog.{schema}.*`       |
+| Schema           | `MODIFY_TABLE`      | `SELECT`, `INSERT`, `UPDATE`, `DELETE`, `ALTER` | `default_catalog.{schema}.*`       |
+| Table            | `SELECT_TABLE`      | `SELECT`                                        | `default_catalog.{schema}.{table}` |
+| Table            | `MODIFY_TABLE`      | `SELECT`, `INSERT`, `UPDATE`, `DELETE`, `ALTER` | `default_catalog.{schema}.{table}` |
```
roryqi, line 208: Which privilege can drop the table?

roryqi: Which privilege can drop the schema?

zhaohehuhu: I don't see any DROP privilege mentioned in the Gravitino documentation. It appears that Gravitino uses ownership to control administrative operations such as dropping objects, rather than defining a dedicated DROP privilege.

zhaohehuhu: Added an Ownership section(4.2) to the document to support DROP permissions. @roryqi

design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,260 @@
...
+
+
+### 4.4 Deny Privileges
+
+The StarRocks pushdown plugin should only push down `Privilege.Condition.ALLOW`.
```
roryqi, line 213: You should throw exception for the unsupported privileges.

design-docs/starrocks-authorization-pushdown.md
roryqi: OK.

design-docs/starrocks-authorization-pushdown.md
```diff
@@ -0,0 +1,260 @@
...
+
+### 4.2 Ownership
+
+Gravitino supports object ownership, where the owner of an object implicitly has administrative privileges such as ALTER and DROP.
+
+Since StarRocks does not provide a native object ownership model, the plugin should map Gravitino ownership to StarRocks privileges by granting ALL privileges on the corresponding resource to the owner. 
+This approach preserves the ownership semantics in Gravitino and ensures that owners retain full administrative control over their objects in StarRocks
+
+| Gravitino object | StarRocks privilege | StarRocks resource                 |
+|------------------|---------------------|------------------------------------|
```
roryqi: Could u explain more about ownership? You can give a concrete example.

design-docs/starrocks-authorization-pushdown.md
zhaohehuhu: Added. Thanks.

design-docs/starrocks-authorization-pushdown.md
zhaohehuhu: Done. Could you plz review again ? @roryqi

roryqi: If Alice is the catalog owner and schema owner, when we change schema owner to other, what will it happen?

zhaohehuhu: When the schema owner changes from `alice` to `others`, the plugin should revoke only the schema-level owner privileges on database `sales`, then grant those schema-level privileges to `others`. The catalog-level owner privileges for `alice` should remain unchanged because Alice is still the catalog owner. @roryqi  Is it ok？ Thanks!

PR #11634, opened 2026-06-13
[#11584] feat(iceberg-rest): Gate config endpoints by backend capabilities

### LINKED ISSUE #11584: [Improvement] Advertise /v1/config endpoints based on actual catalog backend capabilities
`IcebergConfigOperations` advertises a hardcoded `DEFAULT_ENDPOINTS` list (including `V1_SUBMIT_TABLE_SCAN_PLAN`) for every catalog, regardless of whether the configured backend actually supports those operations. A client that trusts the advertised `endpoints` may call an operation the backend cannot serve, producing confusing runtime errors.

### How should we improve?
Derive the advertised endpoint set from the backing catalog's capabilities (e.g., gate optional endpoints like scan planning behind backend support) rather than a static list, while keeping the always-supported core endpoints. Add tests verifying the advertised set reflects backend capability.

Related: IRC-at-scale epic #11131; spec-compliance epic #10547.

laserninja:

Derive the endpoint set advertised by `/v1/config` from the backing catalog's capabilities rather than a hardcoded static list. `V1_SUBMIT_TABLE_SCAN_PLAN` is gated behind `CatalogWrapperForREST.supportsScanPlanOperations()`:

- `CatalogWrapperForREST` returns `true`. Gravitino plans scans locally on loaded table metadata, which works for every backend it handles directly (Hive, JDBC, Memory, Custom).
- `FederatedCatalogWrapper` overrides it to report what the remote catalog advertises in its own `/v1/config`, since it delegates `planTableScan` upstream rather than planning locally. The result is cached for the wrapper's lifetime, so the remote is queried at most once instead of on every local `/v1/config` call.

A remote that omits `endpoints` is treated as not supporting scan planning, matching the Iceberg client's fallback to an endpoint set that predates scan planning. A failed lookup is not cached and does not advertise the endpoint, so it can resolve once the remote recovers.

Also extracts the auth manager / HTTP client / auth session lifecycle shared by the federated credential, scan-plan and config calls into `callRemoteCatalog`, rather than adding a third copy of that boilerplate.

`IcebergConfigOperations` advertises a hardcoded `DEFAULT_ENDPOINTS` list (including `V1_SUBMIT_TABLE_SCAN_PLAN`) for every catalog, regardless of whether the configured backend actually supports those operations. A client that trusts the advertised `endpoints` may call an operation the backend cannot serve, producing confusing runtime errors.

Fix: #11584

A federated (REST backend) catalog advertises the scan plan endpoint only when the remote catalog it federates advertises it. Clients that rely on the advertised endpoints will see the operations their catalog can actually serve. Non-REST backends are unaffected.

- `TestCatalogWrapperForREST` gains coverage against an embedded `/v1/config` server: support reported when the remote advertises the endpoint, not reported when it omits the endpoint or sends no `endpoints` at all, the remote queried only once, `warehouse` forwarded and URL-encoded, and a failed lookup not cached.
- `TestIcebergConfigEndpointGating` and `TestIcebergConfig` cover the `/v1/config` wiring in both directions.
- `:iceberg:iceberg-rest-server:test` is green (404 tests, 0 failures).


iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/IcebergRestTestUtil.java
```diff
@@ -139,9 +148,15 @@ public static ResourceConfig getIcebergResourceConfig(
       IcebergConfigProvider configProvider = IcebergConfigProviderFactory.create(catalogConf);
       configProvider.initialize(catalogConf);
       // used to override register table interface
-      IcebergCatalogWrapperManager icebergCatalogWrapperManager =
-          new IcebergCatalogWrapperManagerForTest(
-              catalogConf, configProvider, false, configProvider.getMetalakeName());
+      IcebergCatalogWrapperManager icebergCatalogWrapperManager;
```
roryqi: Why do we use reflection here?

laserninja: Reflection was used here to keep the overload generic, it accepts any Class<? extends IcebergCatalogWrapperManager> so tests can plug in different wrapper manager subclasses (e.g., NoScanPlanWrapperManager) without adding a new overload for each one.

That said, I can replace the reflection with a Supplier<IcebergCatalogWrapperManager> or a simple factory lambda instead, that's cleaner and avoids the unchecked reflection. Would you prefer that approach, or would you rather I just pass in an already-constructed instance?

roryqi: We should avoid reflection asap. Because reflection lacks necessary check in the compile stage. We can only find the issues in the runtime stage.

laserninja: Makes sense, agreed. I'll replace the reflection with a factory lambda (Function<..., IcebergCatalogWrapperManager>) so the constructor call is checked at compile time. Will push the update shortly.

laserninja:
@roryqi when you have time, can you check this please

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/IcebergRestTestUtil.java
laserninja: Done - the reflection is gone. `IcebergRestTestUtil` now takes a `WrapperManagerFactory` functional interface, so the wrapper manager construction is resolved at compile time:

```java
@FunctionalInterface
public interface WrapperManagerFactory {
  IcebergCatalogWrapperManager create(
      Map<String, String> properties,
      IcebergConfigProvider configProvider,
      boolean auxMode,
      String metalakeName);
}
```

Callers pass a constructor reference (`IcebergCatalogWrapperManagerForTest::new`, `NoScanPlanWrapperManager::new`), so a signature mismatch is a compile error instead of a runtime one.

I've also merged the latest main into the branch. `:iceberg:iceberg-rest-server:test` is green (397 tests, 0 failures). The earlier red `build (17)` check was an unrelated OceanBase testcontainer startup failure.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -181,6 +181,20 @@ public LoadCredentialsResponse getTableCredentials(
...
+   * Whether this catalog wrapper supports server-side scan planning.
+   *
+   * <p>Gravitino implements scan planning locally on top of loaded table metadata. For non-REST
+   * backends (Hive, JDBC, Memory, Custom) this always works. For REST backends the upstream catalog
+   * owns scan planning; advertising the endpoint here could be misleading when the upstream does
+   * not support it, so we do not advertise it.
+   *
+   * @return {@code true} if the scan-plan endpoint should be advertised
+   */
+  public boolean supportsScanPlanOperations() {
```
roryqi, line 194: Have we supported remote catalog scan planning now?

laserninja: yes we do now, and this gating is stale as a result.

FederatedCatalogWrapper.planTableScan delegates to the remote catalog (POST {table}/plan, forwarding X-Iceberg-Access-Delegation for vended credentials and rewriting the credential refresh endpoints back to this IRC). That landed in [#10800](https://github.com/apache/gravitino/pull/10800) on Jul 13, a month after this PR was opened, and only reached this branch when I merged main yesterday.

Since FederatedCatalogWrapper extends CatalogWrapperForREST and doesn't override supportsScanPlanOperations(), the inherited !isRESTCatalog() returns false for exactly the catalogs that now support scan planning. So this would hide V1_SUBMIT_TABLE_SCAN_PLAN from /v1/config for REST backends that can actually serve it. The javadoc rationale is wrong too: we don't leave scan planning to the upstream, we proxy to it.

Two ways forward, and I'd like your preference:

Drop the scan-plan gating. Correct today, since every backend supports scan planning. But it leaves [#11584](https://github.com/apache/gravitino/issues/11584) with nothing to gate (view gating already existed before this PR), so this would effectively be closed as obsolete.
Have FederatedCatalogWrapper override supportsScanPlanOperations() to report whether the remote catalog advertises V1_SUBMIT_TABLE_SCAN_PLAN in its own /v1/config. That keeps the capability-derived intent of [#11584](https://github.com/apache/gravitino/issues/11584) and genuinely fixes misleading advertisement for the federation case. Iceberg 1.11's RESTCatalog exposes no accessor for the remote endpoint set, so the wrapper would need its own cached /v1/config fetch; it already builds an HTTP client for the credentials and plan paths.

I lean toward 2, but happy to go with 1 and close this out if you don't think the capability gating justifies the extra remote fetch.

roryqi: Agree with you. Option 2 seems better. You can continue.

laserninja: Done, pushed in `92edf5b`.

`CatalogWrapperForREST.supportsScanPlanOperations()` now returns `true` (Gravitino plans locally for every backend it handles directly), and `FederatedCatalogWrapper` overrides it to report whether the remote advertises `V1_SUBMIT_TABLE_SCAN_PLAN` in its own `/v1/config`. Cached for the wrapper's lifetime so the remote is queried at most once, not per local `/v1/config` call.

Two behaviours worth calling out:

- A remote that omits `endpoints` is treated as not supporting scan planning. That matches the Iceberg client, which falls back to a default endpoint set predating scan planning when the field is absent.
- A failed lookup is not cached and does not advertise the endpoint, so it can resolve once the remote recovers.

One thing outside the strict scope, so say the word if you'd rather I split it: `getRESTTableCredentials` and `getRESTTablePlanScan` each carried their own copy of the auth manager / client / session lifecycle, and the config fetch would have been a third. I pulled it into `callRemoteCatalog` instead. It's a separate commit-sized change if you prefer it out of this PR.

iceberg/iceberg-rest-server/src/test/java/org/apache/gravitino/iceberg/service/rest/TestIcebergConfigEndpointGating.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Tests that the /v1/config endpoint gates optional endpoints (like scan planning) based on the
+ * catalog backend's capabilities.
+ */
+public class TestIcebergConfigEndpointGating extends IcebergTestBase {
+
+  @Override
+  protected Application configure() {
+    return IcebergRestTestUtil.getIcebergResourceConfig(
```
roryqi, line 41: Nit: please avoid the FQN here and add a normal `import java.util.Collections;` instead, consistent with the project style.

laserninja: Fixed in `97c84fe` - now a normal `import java.util.Collections;`.

I also scanned the rest of the diff for the same pattern and this was the only FQN the PR introduced.

laserninja: done

roryqi (dismissed):

laserninja:
@bharos can you review this please

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/FederatedCatalogWrapper.java
```diff
@@ -188,12 +198,109 @@ public PlanTableScanResponse planTableScan(
...
+  /**
+   * Fetches the remote catalog's {@code /v1/config} response.
+   *
+   * <p>The {@code warehouse} query parameter is forwarded when configured, so a remote serving
+   * several warehouses returns the endpoint set for the one this catalog federates.
+   *
+   * @return the remote catalog's config response.
+   */
+  @VisibleForTesting
+  ConfigResponse fetchRemoteConfig() {
```
bharos: Two small things on this method, neither blocking:

1. Can this be `private`? `fetchRemoteConfig` is only called from `supportsScanPlanOperations` — the new tests inject through the `getCatalog()` override rather than calling this directly, so `@VisibleForTesting` isn't quite accurate. (The import can go too if nothing else in the file uses it.)

2. Worth one line of javadoc on *why* it refetches. `RESTCatalog.initialize()` already does a `GET /v1/config` against this same remote, but `RESTSessionCatalog` keeps `endpoints` in a private field with no accessor, so it can't be read back — the refetch is the only option. Without that noted, this reads as a redundant round trip and someone will eventually try to remove it. Something like:

```java
 * <p>{@code RESTCatalog} already fetched this at init but keeps the endpoint set private, so it
 * has to be re-fetched here.
```

laserninja: Both done in 57ccb8ab6.

1. `fetchRemoteConfig` is now `private`. The `@VisibleForTesting` import stays since it is still used elsewhere in the file.
2. Added the javadoc paragraph you suggested, near-verbatim.

bharos (approved):
LGTM

roryqi (approved):

bharos:
Thanks for the review @roryqi

PR #11647, opened 2026-06-15
[#11646] fix(ci): Vendor dorny/paths-filter to comply with Apache CI allowlist

### LINKED ISSUE #11646: [Bug report] CI fails due to dorny/paths-filter not allowed in Apache org
**Describe what's wrong:**

`dorny/paths-filter` is not on Apache's GitHub Actions allowlist, causing all CI workflows to fail.

**Error message:**

```
The action dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 is not allowed in apache/gravitino because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: ...
```

**How to reproduce:**

Open any pull request — all CI checks using `dorny/paths-filter` will fail with the above error.

**Additional context:**

Fix: vendor the compiled artifact of `dorny/paths-filter@de90cc6` (MIT License) into `.github/actions/paths-filter/` and replace all 12 workflow references with `uses: ./.github/actions/paths-filter`. Local actions are always allowed and behavior is identical since it is the same compiled code.

diqiu50:

Vendor `dorny/paths-filter@de90cc6` (`dist/index.js`, MIT License) into `.github/actions/paths-filter/` and replace all 12 workflow references with the local action. Behavior is identical.

Fix: #11646

`dorny/paths-filter` is not on Apache's GitHub Actions allowlist, causing CI to fail on every PR. Local actions under `./.github/actions/` are always permitted.

No.

CI on this PR.


yuqi1129:
@diqiu50 
Why do we need to introduce such a large change, please see: https://github.com/apache/gravitino/pull/11648

diqiu50:
> @diqiu50 Why do we need to introduce such a large change, please see: #11648

It seems switching to another version would resolve the issue. I vendored the previous version into the repo.
Let’s go ahead with your approach.

PR #11649, opened 2026-06-15
[#11615] fix(idp): Disable idp plugin when simple is listed in authenticators

### LINKED ISSUE #11615: [Improvement] Built-in IdP should automatically exclude simple authenticator when enabled

When the `idp-basic` plugin is enabled, users are currently required to manually ensure that `gravitino.authenticators` does not include `simple`. While there is already a startup validation that catches this misconfiguration and fails fast, the user experience can be further improved.

From a user perspective, enabling a built-in IdP should be a single-step operation. Requiring users to understand the incompatibility relationship and manually remove `simple` from the authenticators list creates unnecessary configuration friction, especially since `simple` is the default authenticator. Users expect that enabling a more secure authentication mechanism automatically supersedes the less secure default.

### How should we improve?

When `idp-basic` is enabled, the system should **automatically exclude `simple` from the effective authenticators list** at startup, without requiring the user to explicitly remove it from the configuration.

Specifically:
- If `gravitino.authenticators` contains both `simple` and `oauth2` (with `idp-basic` enabled), the system should automatically drop `simple` from the active authenticator chain and log a warning (e.g., `"simple authenticator automatically disabled because idp-basic plugin is active"`).
- The user's configuration file does not need to be modified — this is a runtime behavior adjustment.
- This removes the current "configuration trap" where users enable IdP but forget to remove the default `simple` authenticator.

This approach is more user-friendly than the current fail-fast validation because it allows the system to do the right thing automatically while informing the user via logs.

lasdf1234:

When `simple` is listed in `gravitino.authenticators`, the `idp-basic` plugin now logs a warning and skips IdP initialization instead of calling `System.exit(1)`. The server continues to start and requests use `simple`. IdP REST resources (`/api/idp/*`), service-admin initialization, and built-in Basic authenticator registration are not performed in this case.

Docs are updated to describe the runtime behavior and the requirement to configure `gravitino.authenticators` without `simple` to enable built-in IdP.

Issue #11615 highlights friction when built-in IdP is enabled while `simple` (the default authenticator) remains configured. The previous fail-fast behavior blocked server startup. This change allows coexistence of the extension package with default/simple auth by disabling IdP at startup when the conflict is detected.

Note: #11615 proposes automatically excluding `simple` from the active authenticator chain. This PR takes the alternative approach of keeping `simple` and disabling the `idp-basic` plugin, because both mechanisms handle `Authorization: Basic` headers.

Fix: #11615

1. Server startup no longer fails when `idp-basic` is enabled and `simple` is listed in `gravitino.authenticators`.
2. A WARN log is emitted and built-in IdP APIs are not registered in that configuration.
3. Documentation updates for built-in IdP setup and OpenAPI description.

1. `./gradlew :plugins:idp-basic:test -PskipITs` 
2. `./gradlew :docs:build`
3. `./gradlew spotlessApply`

### DIFF
--- a/design-docs/gravitino-local-authentication.md
+++ b/design-docs/gravitino-local-authentication.md
@@ -319,8 +319,9 @@ fresh Gravitino deployment.
    gravitino.authorization.serviceAdmins=admin1,admin2
    ```
 
-   Built-in IdP is incompatible with the `simple` authenticator. When the `idp-basic` plugin is
-   enabled, `gravitino.authenticators` must not include `simple`.
+   Built-in IdP is incompatible with the `simple` authenticator. If `simple` is listed in
+   `gravitino.authenticators`, the `idp-basic` plugin is disabled at startup. Configure
+   `gravitino.authenticators` without `simple` to enable the plugin.
 
 2. Export the initial service admin password before starting Gravitino:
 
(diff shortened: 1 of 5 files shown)


lasdf1234:
@roryqi Could you help me to review code. THX.

roryqi:
I don't think this is a good solution. It may make users confused and need user to see the logs to know why the configuration doesn't take effect.

PR #11658, opened 2026-06-15
[#11639] fix(server): inherit tags/policies through multi-level schemas

### LINKED ISSUE #11639: [Improvement] Multi-level schema should inherit tags/policies from parent schema

In the current implementation, when using multi-level (nested) schemas, tags and policies assigned to a parent schema are not automatically inherited by child schemas or the objects within them. This forces users to manually assign the same tags/policies at every level of the schema hierarchy, which is repetitive, error-prone, and does not match user expectations for hierarchical metadata management.

For example, if a user assigns a "PII" tag or an access control policy to a top-level schema, they expect all sub-schemas, tables, and views under that schema to implicitly inherit that tag/policy unless explicitly overridden.

### How should we improve?

Implement tag/policy inheritance along the schema hierarchy:

1. When resolving tags for a metadata object (schema, table, view, column), walk up the schema hierarchy and merge inherited tags from ancestor schemas.
2. Allow explicit override at lower levels — a child schema or table can override or suppress an inherited tag/policy.
3. Provide an API option to distinguish between "directly assigned" and "inherited" tags when listing tags for an object.
4. Document the inheritance semantics clearly for users.

roryqi:

Tag/policy inheritance for a metadata object is computed in the REST layer
(`MetadataObjectTagOperations` / `MetadataObjectPolicyOperations`), which walks
ancestors via `MetadataObjects.parent()` and marks each with `inherited=true`.

This PR adds `MetadataObjectUtil.getParentMetadataObjects()`, a shared helper that
returns the ancestor chain and **expands multi-level (hierarchical) schema
ancestors** using the configured `gravitino.schema.separator`. All four
parent-walks (list + get, for tags and policies) now use it. Docs for tag/policy
inheritance are updated to describe the hierarchical-schema semantics.

For a hierarchical schema such as `a:b:c`, the parent walk jumped straight from
the schema to the catalog, **skipping the intermediate parent schemas `a:b` and
`a`**. As a result, tags/policies assigned to a parent schema were not inherited
by its child schemas or the tables/views within them, which does not match users'
expectations for hierarchical metadata management.

Fix: #11639

Yes. Listing tags/policies of an object under a multi-level schema now also returns
the tags/policies of the intermediate parent schemas (marked `inherited=true`).
No API or property-key changes.

- Unit tests `TestMetadataObjectUtil` for the new helper (flat schema, hierarchical
  schema, column-under-hierarchical cases).
- `TestMetadataObjectTagOperations` / `TestMetadataObjectPolicyOperations`: new tests
  asserting a table under `hcat.a:b:c.tbl` inherits from `a:b:c`, `a:b`, `a`, and the catalog.
- New integration test `HierarchicalSchemaTagPolicyIT` (embedded server + Iceberg/H2
  catalog with separator `:`) verifying end-to-end tag and policy inheritance through
  the hierarchical schema chain. Both IT tests pass.


clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
```diff
@@ -0,0 +1,231 @@
...
+ * a multi-level (hierarchical) schema hierarchy.
+ *
+ * <p>A hierarchical schema {@code A:B:C} (using the configured {@code ":"} separator) has the
+ * intermediate schemas {@code A:B} and {@code A} as ancestors. Listing the tags/policies of {@code
+ * A:B:C} must therefore include the tags/policies assigned to {@code A:B}, {@code A} and the
+ * catalog as inherited (see issue #11639). The catalog is an Iceberg catalog backed by an in-memory
+ * H2 database, because {@code ":"} hierarchical schema names are only supported by Iceberg catalogs
+ * accessed through the Gravitino REST server with a configured schema separator.
+ */
+@org.junit.jupiter.api.Tag("gravitino-docker-test")
```
jerryshao: Don't use full qualified package name.

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
```diff
@@ -0,0 +1,231 @@
...
+    catalog
+        .asSchemas()
+        .loadSchema(SCHEMA_AB)
+        .supportsTags()
+        .associateTags(null, new String[] {midTag});
+    leafSchema.supportsTags().associateTags(null, new String[] {leafTag});
+    for (String tag : new String[] {catalogTag, rootTag, midTag, leafTag}) {
+      metalake.deleteTag(tag);
+    }
+  }
```
jerryshao, line 202: You should also test creating tables under different schemas, to make sure tables / columns can correctly inherit the tags and policies.

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
jerryshao: This test uses an in-memory H2 backend and doesn't actually need a Docker container. Tagging it `gravitino-docker-test` means it's gated behind `-PskipDockerTests=false`, excluding it from standard CI runs that could otherwise exercise this fix. Consider a lighter IT tag (or whatever marker runs the embedded-server ITs without Docker).

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
```diff
@@ -0,0 +1,231 @@
...
+            CATALOG, Catalog.Type.RELATIONAL, "lakehouse-iceberg", "comment", catalogProperties);
+
+    // Creating "A:B:C" auto-creates the parent chain "A" and "A:B".
+    catalog.asSchemas().createSchema(SCHEMA_ABC, "hierarchical schema", new HashMap<>());
+    Assertions.assertEquals(SCHEMA_AB, catalog.asSchemas().loadSchema(SCHEMA_AB).name());
+    Assertions.assertEquals(ROOT_A, catalog.asSchemas().loadSchema(ROOT_A).name());
+  }
+
+  @AfterAll
+  public void tearDown() {
```
jerryshao, line 130: Missing `super.stopIntegrationTest()` looks like a lifecycle bug but isn't: JUnit 5 invokes `@AfterAll` methods from both the subclass and the parent class, so `BaseIT.stopIntegrationTest()` (which stops `miniGravitino` and clears `customConfigs`) still runs. Setting `client = null` here prevents the parent from double-closing it. Worth a short comment here so future readers don't add the super call thinking it was accidentally omitted.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
```diff
@@ -139,6 +142,62 @@ public static NameIdentifier toEntityIdent(String metalakeName, MetadataObject m
...
+   * additionally expands hierarchical (multi-level) schemas: a schema name such as {@code a:b:c}
+   * (using the configured schema separator) has the intermediate schemas {@code a:b} and {@code a}
+   * as ancestors. Without this expansion, inheritance would jump directly from {@code a:b:c} to the
+   * catalog and skip the intermediate parent schemas, so tags/policies assigned to a parent schema
+   * would not be inherited by its child schemas and the objects within them.
+   *
+   * @param object The metadata object
+   * @return The ancestor metadata objects, nearest first
+   */
+  public static List<MetadataObject> getParentMetadataObjects(MetadataObject object) {
```
jerryshao, line 168: The public overload resolves the separator by calling `HierarchicalSchemaUtil.schemaSeparator()`, which reads `GravitinoEnv.getInstance().config()`. When `config()` is null (e.g., in unit tests that don't boot the server), `schemaSeparator()` silently falls back to the default `":"`. The new tests in `TestMetadataObjectTagOperations` and `TestMetadataObjectPolicyOperations` rely on this fallback: they use `":"` in schema names and call the public overload, which works only because the default matches. Either add a comment here linking to `schemaSeparator()`'s null-safe contract, or have the REST-layer unit tests call the package-private overload with an explicit `":"` to make the dependency visible.

server/src/test/java/org/apache/gravitino/server/web/rest/TestMetadataObjectTagOperations.java
```diff
@@ -380,6 +380,60 @@ public void testListTagsForObject() {
     Assertions.assertFalse(resultTags8.get("tag0").inherited().get());
   }
 
+  @Test
```
jerryshao, line 383: The mocked objects (`schemaA`, `schemaB`, `schemaC`) are constructed with `":"` as separator. `MetadataObjectUtil.getParentMetadataObjects(object)` (the public overload used in the production code under test) resolves the separator from `GravitinoEnv`, which is null here, so it defaults to `":"`. The test passes incidentally because the default matches. See the note on `MetadataObjectUtil.getParentMetadataObjects(MetadataObject)` — same concern applies to the policy-operations sibling test.

jerryshao (commented):
Overall the fix is correct and well-tested — the ancestor-expansion logic in `getParentMetadataObjects` is sound and the unit tests verify every case. A few minor nits inline.

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
roryqi: Resolved. The class no longer carries any `@Tag` annotation at all — the `gravitino-docker-test` tag was dropped per the later comment — so the fully-qualified name is gone. (d22f3ee6d)

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
roryqi: Added in 673f096f4: `testTagInheritanceForTableAndColumnUnderHierarchicalSchema` and `testPolicyInheritanceForTableUnderHierarchicalSchema` create a table under `A:B:C` and one under `A:B`, asserting tables/columns inherit through the whole chain and that a table under `A:B` does not pick up the `A:B:C`-only tag/policy. Note: `Column` supports tags but not policies, so column coverage is tag-only.

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
roryqi: Removed the `gravitino-docker-test` tag in d22f3ee6d; the IT now runs in the standard embedded integration-test runs (verified it executes without `-PskipDockerTests=false`).

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
roryqi: Added a comment in d22f3ee6d explaining that JUnit 5 still invokes `BaseIT.stopIntegrationTest()` and that closing/nulling the client here prevents BaseIT from double-closing it.

core/src/main/java/org/apache/gravitino/utils/MetadataObjectUtil.java
roryqi: Documented the null-safe separator fallback on the public `getParentMetadataObjects(MetadataObject)` overload in d22f3ee6d, pointing to `HierarchicalSchemaUtil.schemaSeparator()` and the package-private overload for deterministic separators.

server/src/test/java/org/apache/gravitino/server/web/rest/TestMetadataObjectTagOperations.java
roryqi: Added a note in both REST-layer hierarchical tests (tag + policy) that they rely on the default `:` separator because `GravitinoEnv` config isn't booted. (d22f3ee6d)

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
```diff
@@ -0,0 +1,231 @@
...
+ *
+ * <p>A hierarchical schema {@code A:B:C} (using the configured {@code ":"} separator) has the
+ * intermediate schemas {@code A:B} and {@code A} as ancestors. Listing the tags/policies of {@code
+ * A:B:C} must therefore include the tags/policies assigned to {@code A:B}, {@code A} and the
+ * catalog as inherited (see issue #11639). The catalog is an Iceberg catalog backed by an in-memory
+ * H2 database, because {@code ":"} hierarchical schema names are only supported by Iceberg catalogs
+ * accessed through the Gravitino REST server with a configured schema separator.
+ */
+@org.junit.jupiter.api.Tag("gravitino-docker-test")
+public class HierarchicalSchemaTagPolicyIT extends BaseIT {
```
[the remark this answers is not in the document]
roryqi: Removed the Docker tag; the test now runs by default in embedded IT runs. (d22f3ee6d)

clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/HierarchicalSchemaTagPolicyIT.java
```diff
@@ -0,0 +1,231 @@
...
+
+    // Create an Iceberg catalog backed by an in-memory H2 database; it is lightweight, needs no
+    // extra container, and supports ':' hierarchical schema names. H2 is NOT a supported
+    // production Iceberg backend.
+    Map<String, String> catalogProperties = new HashMap<>();
+    catalogProperties.put("catalog-backend", "jdbc");
+    catalogProperties.put("warehouse", "/tmp/gravitino-it-hierarchical-tag-policy");
+    catalogProperties.put(
+        "uri", "jdbc:h2:mem:gravitino-it-hierarchical-tag-policy;DB_CLOSE_DELAY=-1;MODE=MYSQL");
+    catalogProperties.put("jdbc-driver", "org.h2.Driver");
```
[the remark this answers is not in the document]
roryqi: The warehouse path and in-memory H2 database name are now derived from the randomized metalake name to avoid cross-run interference. (d22f3ee6d)

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectTagOperations.java
```diff
@@ -114,11 +115,13 @@ public Response getTagForObject(
             Optional<Tag> tag = getTagForObject(metalake, object, tagName);
             Optional<TagDTO> tagDTO = tag.map(t -> DTOConverters.toDTO(t, Optional.of(false)));
 
-            MetadataObject parentObject = MetadataObjects.parent(object);
-            while (!tag.isPresent() && parentObject != null) {
+            for (MetadataObject parentObject :
+                MetadataObjectUtil.getParentMetadataObjects(object)) {
+              if (tag.isPresent()) {
+                break;
+              }
```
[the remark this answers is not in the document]
roryqi: Added `testGetTagForObjectUnderHierarchicalSchema` covering GET `/tags/{tag}` for a table under `a:b:c` inheriting from `a:b` and `a`. (d22f3ee6d)

server/src/main/java/org/apache/gravitino/server/web/rest/MetadataObjectPolicyOperations.java
```diff
@@ -110,11 +111,13 @@ public Response getPolicyForObject(
             Optional<PolicyDTO> policyDTO =
                 policyEntity.map(t -> PolicyOperations.toDTO(t, Optional.of(false)));
 
-            MetadataObject parentObject = MetadataObjects.parent(object);
-            while (!policyEntity.isPresent() && parentObject != null) {
+            for (MetadataObject parentObject :
+                MetadataObjectUtil.getParentMetadataObjects(object)) {
+              if (policyEntity.isPresent()) {
+                break;
+              }
```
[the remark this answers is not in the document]
roryqi: Added `testGetPolicyForObjectUnderHierarchicalSchema` covering GET `/policies/{policy}` for a table under `a:b:c` inheriting from `a:b` and `a`. (d22f3ee6d)

mchades (approved):

PR #11664, opened 2026-06-16
[#11672] improvement(optimizer): optimize trigger-expr evaluation with table metadata & partition short-circuit

### LINKED ISSUE #11672: [Improvement] Optimize optimizer trigger-expr evaluation with table metadata & partition short-circuit

The `maintenance/optimizer` recommender's trigger-expr / score-expr evaluation has two limitations:

1. **Limited evaluation context** — Trigger expressions can currently only reference partition/table statistics. They cannot reference table metadata (column count, partition count, sort-order count, table properties), which restricts the kinds of rules users can write.
2. **Inefficient partitioned-table evaluation** — Every partition is re-evaluated even when a table-level expression already decides the outcome, and the table-level context is rebuilt for every partition. This is costly for large partitioned tables. Additionally, `QLExpressionEvaluator` calls `Pattern.compile` on every evaluation.

### How should we improve?

1. Extend the trigger-expr context to expose `column_count`, `partition_count`, `sort_order_count`, and table properties (numeric values parsed to `long`, others kept as `string`), alongside existing partition/table statistics. Both partitioned and non-partitioned tables evaluate against partition statistics (when present), table statistics, and table metadata. The trigger-expr string representation stays unchanged.
2. Speed up partitioned-table evaluation (port of Pinterest gravitino-pinterest#249):
   - Short-circuit: evaluate with table-level context only; if it resolves without referencing partition variables, skip the per-partition loop.
   - Precompute the table-level context once per `initialize()`.
   - Cache compiled hyphen-to-underscore regex patterns in `QLExpressionEvaluator`.
   - Add `ExpressionEvaluator#tryToEvaluateBool` returning `Optional<Boolean>`.

roryqi:

Two related optimizations to the `maintenance/optimizer` recommender's trigger-expr / score-expr evaluation:

1. **Extend evaluation context with table metadata** — the trigger-expr context now exposes `column_count`, `partition_count`, `sort_order_count`, and table properties (numeric values parsed to `long`, others kept as `string`), in addition to partition and table statistics. Both partitioned and non-partitioned tables now evaluate against partition statistics (when present), table statistics, and table metadata. The trigger-expr string representation is unchanged.
2. **Speed up partitioned-table evaluation** (port of Pinterest gravitino-pinterest#249):
   - Short-circuit: evaluate the expression with table-level context only; if it resolves without referencing partition variables, skip the per-partition loop (relies on the QL engine's left-to-right `&&` / `||` short-circuiting).
   - Precompute the table-level context once per `initialize()` instead of rebuilding it for every partition.
   - Cache compiled hyphen-to-underscore regex patterns in `QLExpressionEvaluator` to avoid `Pattern.compile` on every evaluation.
   - Adds `ExpressionEvaluator#tryToEvaluateBool` returning `Optional<Boolean>`.

Trigger expressions previously could only reference partition/table statistics, limiting the rules users can write. They also re-evaluated every partition even when a table-level expression already decided the outcome, which is costly for large partitioned tables.

Fixes #11672

No API changes. Trigger-expr authors gain new referenceable variables (`column_count`, `partition_count`, `sort_order_count`, and table properties).

New/extended unit tests: `TestTableMetadataTriggerExpressionUtils`, `TestQLExpressionEvaluator`, and `TestCompactionStrategyHandler`. `./gradlew :maintenance:optimizer:test` passes locally.

### DIFF
--- a/maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/handler/BaseExpressionStrategyHandler.java
+++ b/maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/handler/BaseExpressionStrategyHandler.java
@@ -24,6 +24,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.PriorityQueue;
 import java.util.stream.Collectors;
 import lombok.Value;
@@ -66,6 +67,8 @@ public abstract class BaseExpressionStrategyHandler implements StrategyHandler {
   private Map<PartitionPath, List<StatisticEntry<?>>> partitionStatistics;
   private Table tableMetadata;
   private NameIdentifier nameIdentifier;
+  // Cached table-level context (table stats + metadata + rules), computed once per initialize().
+  private Map<String, Object> tableLevelContext;
 
   /** Create a handler that evaluates expressions with the default QL evaluator. */
   protected BaseExpressionStrategyHandler() {
@@ -80,6 +83,7 @@ public abstract class BaseExpressionStrategyHandler implements StrategyHandler {
     this.strategy = context.strategy();
     this.tableStatistics = context.tableStatistics();
     this.partitionStatistics = context.partitionStatistics();
+    this.tableLevelContext = buildTableLevelContext();
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/util/TableMetadataTriggerExpressionUtils.java
```diff
@@ -0,0 +1,89 @@
...
+        TableMetadataTriggerExpression.SORT_ORDER_COUNT.getName(),
+        tableMetadata.sortOrder() != null ? tableMetadata.sortOrder().length : 0);
+
+    if (tableMetadata.properties() != null) {
+      tableMetadata
+          .properties()
+          .forEach(
+              (k, v) -> {
+                try {
+                  context.put(k, Long.parseLong(v));
```
yuqi1129: I would suggest we use `NumberUtils.isCreatable` to replace this one.

maintenance/optimizer/src/main/java/org/apache/gravitino/maintenance/optimizer/recommender/handler/BaseExpressionStrategyHandler.java
```diff
@@ -219,10 +238,44 @@ private boolean evaluateBool(String expression, List<StatisticEntry<?>> statisti
...
-      Strategy strategy, List<StatisticEntry<?>> statistics) {
-    Map<String, Object> context = new HashMap<>();
+  private Optional<Boolean> tryToEvaluateBool(
+      String expression, List<StatisticEntry<?>> statistics) {
+    Map<String, Object> context = buildExpressionContext(statistics);
+    try {
+      return expressionEvaluator.tryToEvaluateBool(expression, context);
+    } catch (RuntimeException e) {
+      LOG.warn("Failed to evaluate expression '{}' with context {}", expression, context, e);
+      return Optional.of(false);
```
yuqi1129: It should be `Optional.of(null)` according to the Java docs in expressionEvaluator.

yuqi1129 (approved):

Issue #11682, opened 2026-06-16
[EPIC] Web UI support for built-in IdP Basic login


lasdf1234:
### Describe the proposal

Gravitino built-in IdP supports HTTP Basic authentication on the server (`idp-basic` plugin), but the Web UI does not provide a username/password login flow today. Login pages only handle `simple` and `oauth` modes; see [How to use built-in IDP](https://gravitino.apache.org/docs/security/how-to-use-built-in-idp#web-ui).

This epic tracks end-to-end work to let operators sign in to the Web UI with built-in IdP Basic credentials when the server is configured for IdP Basic authentication (`gravitino.authenticators` configured without `simple`).

**Scope**
- Backend: expose whatever the Web UI needs to detect Basic/IdP mode and authenticate API calls after login.
- Frontend: add Basic login UX in Web UI (v1 and/or v2 as agreed by maintainers).

**Out of scope (for this epic)**
- Replacing enterprise IdPs (Keycloak, Azure AD, etc.).
- Changing connector/client Basic auth flows (REST API, Spark/Flink/Trino connectors already documented separately).

**Related**
- Built-in IdP docs: `docs/security/how-to-use-built-in-idp.md`
- Authentication docs: `docs/security/how-to-authenticate.md#basic-mode`

### Task list

- [x] #11680
- [x] #11681

Octavi00:
Hi @lasdf1234, I've made a lot of progress on the login flow and features, but I had some important questions about the expectations of the flow. Currently, I have maybe 90% complete with a custom login page for basic auth, using a proper user and pw allows the user to access the data page, but I haven't properly resolved improper inputs yet. 

My main concerns were about how to handle the user returning to the login page by typing the url in. Since they are already logged in, should a feature that auto-redirects them back to the metadata page be implemented? Currently, if they go back to the login page and enter an improper User/Pw, they get 'logged out' in a sense. I'm unsure of how deep that really is, but they are forced to log back in to access metadata once more.

Secondly, if a user is on the metadata page and reloads the page, the logout button disappears. I'm assuming it may have something to do with me removing the expiredIn timer for basicAuth, but since it appears during the first redirection, I feel it may be an issue for all authenticators?

Now, some less important issues I noticed:
- I noticed the constructor in factory.js didn't contain this.providerType = null for its standard initialization. initializeProvider() always handles that, but since configPromise and currentProvider are initialized to null here, I felt that it was missing.
[...]

Issue #11681, opened 2026-06-16
[Subtask] Frontend: add Basic login support in Web UI


lasdf1234:
### Describe the subtask

Add username/password login to the Gravitino Web UI when the server uses built-in IdP Basic authentication, including login UX, API authentication integration, error handling, documentation, and tests.

### Parent issue

#11682

Octavi00:
Hi @lasdf1234, I saw the backend component has been completed for the Parent issue, so I'd like to ask if I can tackle the remaining Front End portion described here?

lasdf1234:
> Hi [@lasdf1234](https://github.com/lasdf1234), I saw the backend component has been completed for the Parent issue, so I'd like to ask if I can tackle the remaining Front End portion described here?

The "Basic auth login" is the issue that gravitino1.3 aims to address. If you have time, please complete this part as soon as possible. Thank you.

PR #11688, opened 2026-06-16
[#11680] feat(idp-basic): Add basic authenticator type for Web UI login contract

### LINKED ISSUE #11680: [Subtask] Backend: adapt server for built-in IdP Basic login in Web UI
### Describe the subtask

Enable the Gravitino server side to support Web UI login with built-in IdP Basic authentication, including configuration exposure, authentication contract for the UI, documentation, and tests.

### Parent issue

#11682

lasdf1234:

This PR implements the **server-side** work for built-in IdP Basic authentication in the Web UI (#11680, subtask of #11682):

- Add `basic` to `AuthenticatorType` and treat it as a configuration token; `BasicAuthenticator` is still registered by the `idp-basic` plugin at Jersey startup.
- Add `IdpBasicAuthConfigValidator` and validate configuration during `GravitinoServer` initialization and `IdpRESTFeature` startup:
  - `basic` and `gravitino.server.rest.extensionPackages=org.apache.gravitino.idp.web.rest.feature` must be configured **together**; startup fails if only one is set.
  - `basic` and `simple` cannot be configured together.
- Expose `basic` through the existing `/configs` response (`gravitino.authenticators`) so the Web UI can detect the auth type.
- Update operator documentation and OpenAPI description for the Basic auth + Web UI contract.

**Out of scope (tracked in #11681):** Web UI login form, axios Basic header handling, and session routing for `authType === 'basic'`.

Built-in IdP previously relied on implicit OAuth-only wiring when the extension package was enabled, and documentation stated that the Web UI could not use Basic login. #11680 requires an explicit `basic` authenticator type and strict startup rules so operators and the Web UI share a clear server contract before frontend work lands in #11681.

Fix: #11680

Yes.

1. New authenticator token: `basic` in `gravitino.authenticators`.
2. Startup validation: enabling the idp-basic extension package without `basic` (or listing `basic` without the extension package) now fails fast with a clear error.
3. `basic` + `simple` together now fails at startup.
4. Documentation updates describe Basic auth configuration and Web UI auth type detection (UI form implementation is #11681).

1. Added unit tests: `TestIdpBasicAuthConfigValidator`, updated `TestIdpRESTFeature`.
2. Updated idp-basic integration test configs to use `basic` (`BasicAuthOperationsIT`, `IdpRESTApiIT`).
3. Ran:
   - `./gradlew :core:test --tests "org.apache.gravitino.auth.TestIdpBasicAuthConfigValidator" -PskipITs`
   - `./gradlew :plugins:idp-basic:test --tests "org.apache.gravitino.idp.web.rest.feature.TestIdpRESTFeature" -PskipITs`


lasdf1234:
@roryqi  Could you help me to review this code. Thanks!

roryqi (dismissed):
LGTM.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/auth/BasicAuthenticator.java
```diff
@@ -62,10 +68,31 @@ public Principal authenticateToken(byte[] tokenData) {
...
+    boolean idpExtensionEnabled =
+        config.get(Configs.REST_API_EXTENSION_PACKAGES).stream()
+            .anyMatch(
+                pkg -> IdpRESTFeature.IDP_REST_EXTENSION_PACKAGE.equalsIgnoreCase(pkg.trim()));
+    if (!idpExtensionEnabled) {
+      LOG.error(
+          "'basic' in gravitino.authenticators requires gravitino.server.rest.extensionPackages "
+              + "to include {}.",
+          IdpRESTFeature.IDP_REST_EXTENSION_PACKAGE);
+      System.exit(1);
```
roryqi: You should throw an exception here instead of exit system call.

lasdf1234: Got resolved. I deployed gravitino for testing and confirmed that the plan was feasible. 
Code has been modified.

roryqi: You should modify the IDP initialization if possible, too.

lasdf1234: Jersey throws an exception during Servlet initialization. However, Jetty catches the exception and still starts the connector, and the process remains alive.

BasicAuthenticator →  It would be better to change it to throw 
The verification is executed in the `initialize()` function, and the exception is passed to `main()`, causing the process to exit.
This effect is the same as the original `System.exit(1)`, which can prevent startup and makes the unit tests simpler.

`IdpRESTFeature` → Simply changing it to `throw` is not enough 
The verification is carried out in the Jersey Feature.configure(). In Jetty, the exception is swallowed and the process does not exit; instead, 8090 becomes 503.
Originally, using System.exit(1) could actually terminate the process immediately.

roryqi (approved):

roryqi:
Because you modified the document. I labeled branch 1.3.

PR #11697, opened 2026-06-16
[#11696] improvement(helm): Make init-postgresql secret name configurable via postgresql.existingSecretName

### LINKED ISSUE #11696: [Improvement] `init-postgresql` init container hardcodes secret name, incompatible with external PostgreSQL / CNPG
### Description

 

The Helm chart's `init-postgresql` init container (in `templates/deployment.yaml`) hardcodes the Kubernetes secret name for the PostgreSQL password as `{{ .Release.Name }}-postgresql` (line 169). This assumes the Bitnami PostgreSQL subchart is always used and that the secret is always named according to the Bitnami convention.

 

This breaks compatibility when:

- Using an external PostgreSQL provider (e.g., CloudNativePG, CrunchyData, AWS RDS)

- Using the `cluster` subchart dependency with CNPG, where the secret name is user-defined (e.g., `gravitino-cnpg-secret`)

 

### Current Code

 

```yaml

env:

  - name: POSTGRES_USER

    value: {{ .Values.postgresql.auth.username }}

  - name: GRAVITINO_DB

    value: {{ .Values.postgresql.auth.database }}

  - name: POSTGRES_PASSWORD

    valueFrom:

      secretKeyRef:

        name: {{ .Release.Name }}-postgresql

        key: password

```

 

`templates/deployment.yaml`, lines ~163-169

 

### Proposed Fix

 

Make the secret name configurable via a values field (e.g., `postgresql.existingSecretName`) while keeping backward compatibility:

 

**values.yaml** — Add a new field:

```yaml

postgresql:

  enabled: false

  auth:

    username: gravitino

    password: gravitino

    database: gravitino

    existingSecret: ""

  existingSecretName: ""     # <-- new: override for the init container secret name

```

 

**deployment.yaml** — Use the value with a fallback to the Bitnami default:

```yaml

            - name: POSTGRES_PASSWORD

              valueFrom:

                secretKeyRef:

                  name: {{ default (printf "%s-postgresql" .Release.Name) .Values.postgresql.existingSecretName }}

                  key: password

```

 

**Behaviour:**

- When `postgresql.existingSecretName` is **not set** (default `""`): falls back to `{{ .Release.Name }}-postgresql` — fully backward compatible with Bitnami PostgreSQL.

- When `postgresql.existingSecretName` is **set** (e.g., `gravitino-cnpg-secret`): uses the user-defined secret name.

 

### Additional Context

 

The same pattern is already used in the `gravitino.conf` template for the JDBC host name (line 46 of `resources/config/gravitino.conf`):

 

```

jdbc:postgresql://{{ .Release.Name }}-cnpg-rw:5432/{{ .Values.postgresql.auth.database }}

```

 

This already assumes a CNPG-like naming convention (`-cnpg-rw`) for the host when PostgreSQL is enabled, so adding a configurable secret name would make the Helm chart consistently support both Bitnami and external/CNPG PostgreSQL setups.

 

### Environment

 

- Chart version: 1.3.0

- PostgreSQL backend: CloudNativePG (CNPG) 17

- Workaround currently: patch the deployment template to reference the correct secret name

### How should we improve?

_No response_

dennismdejong:

Make the Kubernetes secret name used by the `init-postgresql` init container configurable via a new `postgresql.auth.existingSecretName` values field, falling back to the Bitnami default `{{ .Release.Name }}-postgresql` when not set.

The Helm chart's `init-postgresql` init container hardcodes the secret name as `{{ .Release.Name }}-postgresql`, which assumes the Bitnami PostgreSQL subchart is always used. This breaks compatibility with external PostgreSQL providers (CloudNativePG, CrunchyData, AWS RDS) where the secret name is user-defined.

Fix: https://github.com/apache/gravitino/issues/11696

Yes, adds `postgresql.auth.existingSecretName` (default: `""`) to `values.yaml`.

- Existing helm unit tests pass (31 tests, 5 suites)
- New test case: default behavior produces `gravitino-postgresql`
- New test case: `existingSecretName=gravitino-cnpg-secret` produces that name

### DIFF
--- a/dev/charts/gravitino/templates/deployment.yaml
+++ b/dev/charts/gravitino/templates/deployment.yaml
@@ -172,7 +172,7 @@ spec:
             - name: POSTGRES_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ .Release.Name }}-postgresql
+                  name: {{ default (printf "%s-postgresql" .Release.Name) .Values.postgresql.existingSecretName }}
                   key: password
           volumeMounts:
             - mountPath: /scripts
--- a/dev/charts/gravitino/tests/deployment_test.yaml
+++ b/dev/charts/gravitino/tests/deployment_test.yaml
@@ -265,3 +265,36 @@ tests:
             name: GRAVITINO_DB
             value: gravitino
           any: true
+      - contains:
+          path: spec.template.spec.initContainers[?(@.name == "init-postgresql")].env
+          content:
+            name: POSTGRES_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: gravitino-postgresql
+                key: password
+          any: true
+
+  - it: renders custom PostgreSQL secret name when existingSecretName is set
+    template: deployment.yaml
+    release:
+      name: gravitino
+    set:
+      postgresql:
+        enabled: true
+        existingSecretName: gravitino-cnpg-secret
+    asserts:
+      - contains:
+          path: spec.template.spec.initContainers
+          content:
+            name: init-postgresql
+          any: true
+      - contains:
+          path: spec.template.spec.initContainers[?(@.name == "init-postgresql")].env
+          content:
+            name: POSTGRES_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: gravitino-cnpg-secret
+                key: password
+          any: true
--- a/dev/charts/gravitino/values.yaml
+++ b/dev/charts/gravitino/values.yaml
@@ -114,6 +114,13 @@ postgresql:
     ## The value is evaluated as a template.
     ##
     existingSecret: ""
+    ## @param auth.existingSecretName Name of existing secret to use in the init container's secretKeyRef
+    ## When set, this overrides the default `{{ .Release.Name }}-postgresql` secret name used by
+    ## the init-postgresql init container to retrieve the PostgreSQL password.
+    ## Useful when using an external PostgreSQL provider (e.g., CloudNativePG, AWS RDS) where the
+    ## secret name does not follow the Bitnami convention.
+    ##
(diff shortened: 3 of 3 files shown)


danhuawang (commented):
Thanks for the PR! I see the motivation — making the init container work with external PostgreSQL providers (CloudNativePG, CrunchyData, AWS RDS) is a legitimate need. However, there's a structural issue that makes this change ineffective in that scenario.

The `init-postgresql` init container is gated on `{{ if .Values.postgresql.enabled }}` ([deployment.yaml L128](https://github.com/apache/gravitino/blob/main/dev/charts/gravitino/templates/deployment.yaml#L128)). This means:

- **`postgresql.enabled=true`** → The Bitnami subchart is deployed, it creates the secret `{{ .Release.Name }}-postgresql` automatically, and the hardcoded name already works. The override has no use here.
- **`postgresql.enabled=false`** → The init container is **not rendered at all**, so the new `existingSecretName` field is never evaluated — this is precisely the scenario where an external PostgreSQL provider would be in play.

In other words, the override can never fire in the use case it's designed for.

To properly support external PostgreSQL providers, the init container's condition would need to be decoupled from `postgresql.enabled` — for example, rendering it whenever the Gravitino entity store targets PostgreSQL, regardless of whether the Bitnami subchart is active. That's a broader change (likely involving an `externalPostgresql` config section, similar to how many Bitnami-based charts handle this pattern).

I'd suggest either:
1. Expanding the scope to decouple the init container guard (e.g. `{{- if or .Values.postgresql.enabled .Values.externalPostgresql.enabled }}`), with a corresponding external-postgres values section that holds the secret name/key; or
2. Closing this in favor of a design discussion on the linked issue about how external PostgreSQL should be supported end-to-end in the chart.

Issue #11297, opened 2026-05-29
[Bug report] `listBindingRoleNames` returns stale role after `revokePrivilegesFromRole` removes last privilege


danhuawang:
### Version

main branch

### Describe what's wrong

After calling `revokePrivilegesFromRole` to remove the last (and only) privilege from a role's securable object on a schema, `listBindingRoleNames()` on that schema still returns the revoked role. The role should no longer appear because the securable object row has been soft-deleted in the database.

This is a **general issue** affecting all schema types (both flat and hierarchical). It is not specific to hierarchical/nested namespace schemas.

The root cause appears to be a cache invalidation gap: `RelationalEntityStore.update` invalidates the role entity cache (`cache.invalidate(roleIdent, ROLE)`) after the revoke, but the `METADATA_OBJECT_ROLE_REL` relation cache — keyed by the schema's identifier — is not properly invalidated. As a result, `listBindingRoleNames` reads stale cached relation data that still includes the revoked role.

### Error message and/or stacktrace

```
org.opentest4j.AssertionFailedError: Flat schema should no longer have role after revoke,
got: [hs_d2286d6c_role_flat] ==> expected: <false> but was: <true>
  at org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:235)
  at org.apache.gravitino.integration.test.schema.TestHierarchicalSchemaIT
      .testRoleBindingRevokeOnFlatSchema(TestHierarchicalSchemaIT.java:1021)
```

The same failure occurs for hierarchical schemas:

```
org.opentest4j.AssertionFailedError: Level1 schema should no longer have role1 after revoke
==> expected: <false> but was: <true>
  at org.apache.gravitino.integration.test.schema.TestHierarchicalSchemaIT
      .testRoleBindingForHierarchicalSchema(TestHierarchicalSchemaIT.java:965)
```

### How to reproduce

1. Use Gravitino main branch with any catalog (Iceberg, Hive, etc.).
2. Create a schema and a role:
   ```java
   icebergCatalog.asSchemas().createSchema("test_schema", "comment", Map.of());
   metalake.createRole("test_role", Map.of(), List.of());
   ```
3. Grant a privilege to the role on the schema:
   ```java
   MetadataObject schemaObject = MetadataObjects.of(catalogName, "test_schema", MetadataObject.Type.SCHEMA);
   metalake.grantPrivilegesToRole("test_role", schemaObject, Set.of(Privileges.UseSchema.allow()));
   ```
[...]

PR #11702, opened 2026-06-17
[#11701] fix(core): invalidate METADATA_OBJECT_ROLE_REL cache on role grant/override/create

### LINKED ISSUE #11701: [Bug report] Newly granted role missing from listBindingRoleNames until relation cache TTL elapses
### Version

main branch

### Describe what's wrong

After `grantPrivilegesToRole` (or `overridePrivilegesInRole` adding a new object, or `createRole` with securable objects) grants a role access to a metadata object that was already queried before, `listBindingRoleNames()` on that object does **not** return the newly granted role until the `METADATA_OBJECT_ROLE_REL` relation cache entry expires (default `expireAfterAccess` 1h). Continuously polling the object keeps the stale entry alive even longer.

This is the symmetric **"add"** side of #11297 (which fixed the **"remove"/revoke** side via the reverse-index patch in `ROLE_SECURABLE_OBJECT_REVERSE_RULE`). That patch only establishes the `roleIdent:ROLE -> objectKey` reverse mapping when a role is *already cached as a binding role* — so a role that is newly granted access (and was therefore never cached for that object) is unreachable by the role-side invalidation.

Root cause: `RelationalEntityStore.update()` / `put()` only call `cache.invalidate(roleIdent, ROLE)` after the write. That is a role-side BFS invalidation; reaching the object-keyed `METADATA_OBJECT_ROLE_REL` entry requires the reverse index to already map `roleIdent -> objectKey`, which does not exist for a not-yet-cached role.

```
org.opentest4j.AssertionFailedError: grant must be immediately visible via listBindingRoleNames
  ==> expected: <[roleA, roleB]> but was: <[roleA]>
    at org.apache.gravitino.storage.TestEntityStorageRelationCache
        .testGrantPrivilegeInvalidatesMetadataObjectRoleRelCache
```

### How to reproduce

1. Use Gravitino main branch.
2. Create metalake/catalog/schema; create `roleA` bound to the schema and `roleB` with no securable objects.
3. `listBindingRoleNames(schema)` → `[roleA]`. This warms the `METADATA_OBJECT_ROLE_REL` cache; the reverse index maps `roleA -> schemaKey` but **not** `roleB`.
4. Grant `roleB` a privilege on the schema:
   `metalake.grantPrivilegesToRole("roleB", schemaObject, Set.of(Privileges.UseSchema.allow()))`.
5. `listBindingRoleNames(schema)` again:
   - **Actual**: still `[roleA]` (missing `roleB`).
   - **Expected**: `[roleA, roleB]`.
6. The new role only appears after the cache entry's TTL (~1h; longer if polled).

### Additional context

**Affected code path**:
- `RelationalEntityStore.update()` / `put()` — invalidates only `cache.invalidate(roleIdent, ROLE)`.
- `CaffeineEntityCache.invalidate(ident, type)` BFS — relies on the reverse index; `ROLE_SECURABLE_OBJECT_REVERSE_RULE` (`ReverseIndexRules.java:121`) only adds `roleIdent -> objectKey` when the role is already cached as a binding role.
- The DB layer is correct (`SecurableObjectMapper` inserts the new row; `listRolesByMetadataObject` SQL would return `roleB` on a fresh query) — the stale read comes from the cached relation result.

**Why revoke/delete (#11297) is not affected**: when removing a role, that role was previously cached as a binding role of the object, so the reverse index already holds `role -> objectKey` and the role-side BFS reaches it. Invariant: `objectKey contains role ⟺ reverse index has role->objectKey`.

**Suggested fix**: in `RelationalEntityStore.update()` and `put()`, after the write, explicitly invalidate the `METADATA_OBJECT_ROLE_REL` cache entry keyed by each securable object of the (new) `RoleEntity`.

sunyuhan1998:

After writing a `RoleEntity`, `RelationalEntityStore.update()` and `put()` now invalidate the object-keyed `METADATA_OBJECT_ROLE_REL` cache entry for each of the role's securable objects (private helper `invalidateMetadataObjectRoleRelationCache`; no-op for non-ROLE entities or empty securable objects). Covers grant / override-add / createRole.

The invalidation uses a new `SupportsRelationEntityCache.invalidateRelationEntry`, which drops only the cached relation result and its index entry **without cascading through the reverse index**. This matters because the reverse index is shared across all roles bound to one object; a full `invalidate(ident, type, relType)` would BFS through it and evict the other roles' mappings (which would break `testInvalidRelationCache`).

Fixes #11701: for a metadata object that was already queried, `listBindingRoleNames()` returns stale data (missing the newly granted role) after `grantPrivilegesToRole` (or `overridePrivilegesInRole` adding an object, or `createRole` with objects), until the relation cache TTL (default 1h) elapses.

Root cause: the write path only calls `cache.invalidate(roleIdent, ROLE)` (a role-side BFS). Reaching the object-keyed `METADATA_OBJECT_ROLE_REL` entry requires the reverse index to already map `roleIdent -> objectKey`, which is only established when the role has previously been cached as a binding role of that object (the #11297 patch). A newly granted role was never cached there, so the invalidation never reaches the entry. The subtraction side (revoke / override-remove / deleteRole) is already covered by the same role-side invalidation via the reverse index (the removed role was previously cached), so this PR does not touch it.

Fix: #11701

No API or configuration change. Behavioral fix: "list binding roles by object" reflects grant/override/create immediately instead of being delayed until the cache entry expires.

Added 5 parameterized tests in `TestEntityStorageRelationCache` (h2/mysql/postgresql × enableCache true/false):
- `testGrantPrivilegeInvalidatesMetadataObjectRoleRelCache` — reproduces the bug (red before / green after);
- `testCreateRoleWithSecurableObjectsInvalidatesMetadataObjectRoleRelCache` — covers the createRole (`store.put`) path;
- `testRevokeAllPrivileges…` / `testOverrideRemoveObject…` / `testDeleteRole…` — three subtraction guards (no stale residue; pin coverage so a future reverse-index rule change cannot silently regress).

Ran: `./gradlew :core:test --tests '*TestEntityStorageRelationCache.test{Invalid,Grant,Revoke,RevokeAll,OverrideRemove,Delete,Create}*' -PskipDockerTests=false -PskipITs` → BUILD SUCCESSFUL, `tests=42 skipped=0 failures=0 errors=0` (mysql ~10s/case, postgresql ~3s, h2 ~0.1–2s; all three backends actually executed). `:core:spotlessCheck` passes.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -204,12 +204,14 @@ public class CaffeineEntityCache extends BaseEntityCache {
     return segmentedLock.withLock(
         key,
         () -> {
-          // Drop only the cached relation result and its index entry. Do NOT cascade through the
-          // reverse index: it is shared across entities (e.g. all roles bound to one metadata
-          // object), and evicting it here would drop other entities' reverse-index mappings. An
-          // explicit cacheData.invalidate does not trigger the expiry listener, so the reverse
-          // index is left intact.
+          // Drop the cached relation result, its index entry, and the reverse-index bookkeeping
+          // for this relation key only. Do NOT cascade through the reverse index to other
+          // entities: the reverse index is shared (e.g. all roles bound to one metadata object),
+          // and a BFS cascade would evict their mappings. cacheData.invalidate is explicit so it
+          // bypasses the removal listener; reverseIndex.remove(key) then cleans up only this
+          // entry's own bookkeeping (entityToReverseIndexMap + reverseIndex references to it).
           cacheData.invalidate(key);
+          reverseIndex.remove(key);
           cacheIndex.remove(key.toString());
           return true;
         });
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -197,6 +197,26 @@ public class CaffeineEntityCache extends BaseEntityCache {
 
   /** {@inheritDoc} */
   @Override
+  public boolean invalidateRelationEntry(
+      NameIdentifier ident, Entity.EntityType type, SupportsRelationOperations.Type relType) {
+    checkArguments(ident, type, relType);
+    EntityCacheRelationKey key = EntityCacheRelationKey.of(ident, type, relType);
+    return segmentedLock.withLock(
+        key,
+        () -> {
+          // Drop only the cached relation result and its index entry. Do NOT cascade through the
+          // reverse index: it is shared across entities (e.g. all roles bound to one metadata
+          // object), and evicting it here would drop other entities' reverse-index mappings. An
+          // explicit cacheData.invalidate does not trigger the expiry listener, so the reverse
+          // index is left intact.
+          cacheData.invalidate(key);
+          cacheIndex.remove(key.toString());
+          return true;
+        });
+  }
+
+  /** {@inheritDoc} */
+  @Override
   public boolean invalidate(NameIdentifier ident, Entity.EntityType type) {
     checkArguments(ident, type);
     return segmentedLock.withLock(
--- a/core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
@@ -57,6 +57,24 @@ public interface SupportsRelationEntityCache {
       NameIdentifier ident, Entity.EntityType type, SupportsRelationOperations.Type relType);
 
   /**
+   * Invalidates only the cached relation result for the given key, without cascading through the
+   * reverse index.
+   *
+   * <p>Unlike {@link #invalidate(NameIdentifier, Entity.EntityType,
+   * SupportsRelationOperations.Type)}, this does not evict the reverse-index mappings (which are
+   * shared across entities, e.g. all roles bound to one metadata object) or other entities' caches.
+   * Use it when a relation result is known to be stale and the next read must re-query the backend,
+   * but the shared reverse index must be preserved.
+   *
+   * @param ident the name identifier
+   * @param type the entity type
+   * @param relType the relation type
+   * @return true if the cache entry was removed
+   */
+  boolean invalidateRelationEntry(
+      NameIdentifier ident, Entity.EntityType type, SupportsRelationOperations.Type relType);
+
+  /**
    * Checks whether an entity with the given name identifier, type, and relation type is present in
    * the cache.
    *
--- a/core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
@@ -57,14 +57,16 @@ public interface SupportsRelationEntityCache {
       NameIdentifier ident, Entity.EntityType type, SupportsRelationOperations.Type relType);
 
   /**
-   * Invalidates only the cached relation result for the given key, without cascading through the
-   * reverse index.
+   * Invalidates the cached relation result for the given key and cleans up the reverse-index
+   * bookkeeping for that entry, without cascading through the reverse index to other entities.
    *
    * <p>Unlike {@link #invalidate(NameIdentifier, Entity.EntityType,
-   * SupportsRelationOperations.Type)}, this does not evict the reverse-index mappings (which are
-   * shared across entities, e.g. all roles bound to one metadata object) or other entities' caches.
-   * Use it when a relation result is known to be stale and the next read must re-query the backend,
-   * but the shared reverse index must be preserved.
+   * SupportsRelationOperations.Type)}, this does NOT perform a BFS cascade: it leaves other
+   * entities' caches and the reverse-index mappings of the entities referenced by this relation
+   * (e.g. all roles bound to one metadata object) intact. It only drops this relation result and
+   * the reverse-index references that point at this relation key, so the entry is rebuilt on the
+   * next read. Use it when a relation result is known to be stale and a full cascade would
+   * incorrectly evict other entities' state.
    *
    * @param ident the name identifier
    * @param type the entity type
--- a/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
@@ -49,10 +49,13 @@ import org.apache.gravitino.cache.EntityCacheKey;
 import org.apache.gravitino.cache.EntityCacheRelationKey;
 import org.apache.gravitino.cache.NoOpsCache;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
+import org.apache.gravitino.meta.GroupEntity;
 import org.apache.gravitino.meta.RoleEntity;
+import org.apache.gravitino.meta.UserEntity;
 import org.apache.gravitino.storage.relational.service.EntityIdService;
 import org.apache.gravitino.utils.Executable;
 import org.apache.gravitino.utils.MetadataObjectUtil;
+import org.apache.gravitino.utils.NamespaceUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -147,7 +150,7 @@ public class RelationalEntityStore
       throws IOException, EntityAlreadyExistsException {
     backend.insert(e, overwritten);
     cache.put(e);
-    invalidateMetadataObjectRoleRelationCache(e);
+    invalidateAggregatedRoleRelationCache(e);
   }
 
   @Override
@@ -156,7 +159,7 @@ public class RelationalEntityStore
       throws IOException, NoSuchEntityException, EntityAlreadyExistsException {
     E updatedEntity = backend.update(ident, entityType, updater);
     cache.invalidate(ident, entityType);
-    invalidateMetadataObjectRoleRelationCache(updatedEntity);
+    invalidateAggregatedRoleRelationCache(updatedEntity);
     return updatedEntity;
   }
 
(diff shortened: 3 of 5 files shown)


sunyuhan1998:
In the last CI, an integration test failed, but based on my analysis of the error, I feel it seems unrelated to the current PR changes. I tried rebasing onto the latest main and force-pushing to retrigger the CI.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -451,4 +456,36 @@ private <E extends Entity & HasIdentifier> void batchPopulateRelationCache(
...
+    String metalake = ((RoleEntity) entity).namespace().level(0);
+    for (SecurableObject securableObject : securableObjects) {
+      // Drop only the relation result entry for this object, not the shared reverse index. The
+      // reverse index is shared across all roles bound to the object; a full invalidate would
+      // cascade through it and evict the other roles' mappings. The next listRolesByObject
+      // re-queries the backend and rebuilds both the entry and the reverse index.
+      cache.invalidateRelationEntry(
+          MetadataObjectUtil.toEntityIdent(metalake, securableObject),
+          MetadataObjectUtil.toEntityType(securableObject.type()),
+          SupportsRelationOperations.Type.METADATA_OBJECT_ROLE_REL);
```
yuqi1129: Does this problem only exist for `METADATA_OBJECT_ROLE_REL`?

core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
```diff
@@ -56,6 +56,24 @@ <E extends Entity & HasIdentifier> Optional<List<E>> getIfPresent(
...
+   * shared across entities, e.g. all roles bound to one metadata object) or other entities' caches.
+   * Use it when a relation result is known to be stale and the next read must re-query the backend,
+   * but the shared reverse index must be preserved.
+   *
+   * @param ident the name identifier
+   * @param type the entity type
+   * @param relType the relation type
+   * @return true if the cache entry was removed
+   */
+  boolean invalidateRelationEntry(
```
yuqi1129, line 76: Can you try to check whether `invalidateOnKeyChange` can solve the problem?  I am cautious about the newly added interfaces.

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -195,6 +195,26 @@ public boolean invalidate(
...
+        key,
+        () -> {
+          // Drop only the cached relation result and its index entry. Do NOT cascade through the
+          // reverse index: it is shared across entities (e.g. all roles bound to one metadata
+          // object), and evicting it here would drop other entities' reverse-index mappings. An
+          // explicit cacheData.invalidate does not trigger the expiry listener, so the reverse
+          // index is left intact.
+          cacheData.invalidate(key);
+          cacheIndex.remove(key.toString());
+          return true;
```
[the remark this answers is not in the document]
sunyuhan1998: Fixed in the latest push: `invalidateRelationEntry` now also calls `reverseIndex.remove(key)` to clean up this relation key's own reverse-index bookkeeping (`entityToReverseIndexMap` + the `reverseIndex` references pointing at it) — no BFS cascade, so other entities' mappings are preserved.

On the "leak over time" severity: it's bounded in practice (one metadata object maps to a single relation key, rebuilt/merged on the next read), but cleaning it up is clearly more correct, so I adopted it.

On `cacheData.asMap().remove` for an accurate boolean: I kept `cacheData.invalidate` to stay consistent with `invalidate(ident,type,relType)`, which also returns `true` unconditionally and whose callers don't inspect the return value.

core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
```diff
@@ -56,6 +56,24 @@ <E extends Entity & HasIdentifier> Optional<List<E>> getIfPresent(
...
+   * SupportsRelationOperations.Type)}, this does not evict the reverse-index mappings (which are
+   * shared across entities, e.g. all roles bound to one metadata object) or other entities' caches.
+   * Use it when a relation result is known to be stale and the next read must re-query the backend,
+   * but the shared reverse index must be preserved.
+   *
+   * @param ident the name identifier
+   * @param type the entity type
+   * @param relType the relation type
+   * @return true if the cache entry was removed
+   */
```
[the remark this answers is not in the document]
sunyuhan1998: Updated — the Javadoc now states the precise semantics: no BFS cascade to other entities, but the invalidated relation key's own reverse-index bookkeeping is cleaned up and the entry is rebuilt on the next read.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -451,4 +456,36 @@ private <E extends Entity & HasIdentifier> void batchPopulateRelationCache(
...
+    List<SecurableObject> securableObjects = ((RoleEntity) entity).securableObjects();
+    if (securableObjects == null || securableObjects.isEmpty()) {
+      return;
+    }
+    String metalake = ((RoleEntity) entity).namespace().level(0);
+    for (SecurableObject securableObject : securableObjects) {
+      // Drop only the relation result entry for this object, not the shared reverse index. The
+      // reverse index is shared across all roles bound to the object; a full invalidate would
+      // cascade through it and evict the other roles' mappings. The next listRolesByObject
+      // re-queries the backend and rebuilds both the entry and the reverse index.
```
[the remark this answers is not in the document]
sunyuhan1998: Reworded the inline comment to match (no BFS cascade; entry rebuilt on next read).

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
sunyuhan1998: Good question — it's not unique to `METADATA_OBJECT_ROLE_REL`. I wrote an H2 probe and confirmed `user.roleNames` (`grantRolesToUser` → `ROLE_USER_REL`) and `group.roleNames` (`grantRolesToGroup` → `ROLE_GROUP_REL`) have the same defect: an aggregated field mutated via `store.update`, while only the entity itself is invalidated, so the counterpart-keyed relation cache goes stale when the freshly granted grantee was never cached against that role.

The latest commit generalizes the helper to all three (`role.securableObjects`, `user.roleNames`, `group.roleNames`). `tag`/`policy`/`owner` relations are unaffected — they're written via `insertRelation`/`updateEntityRelations`, which already invalidate both ends.

core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
sunyuhan1998: I checked `invalidateOnKeyChange`; unfortunately it can't cover this case:

1. It's only invoked from `cache.put(entity)` (i.e. `store.put` / `store.get` cache-miss backfill). But `grantPrivilegesToRole` / `overridePrivilegesInRole` / `grantRolesToUser` / `grantRolesToGroup` go through `store.update`, which calls `cache.invalidate(ident, type)` — **not** `cache.put` — so `invalidateOnKeyChange` never fires for them, and those are exactly the main trigger paths.
2. Even where it does fire, it uses `invalidate(ident, type)` (BFS), which cascades through the reverse index and evicts other roles' mappings — the same issue `testInvalidRelationCache` caught.

Dropping a single relation entry without a BFS cascade needs a dedicated method (the existing `invalidate(ident,type,relType)` is also BFS), so `invalidateRelationEntry` is the minimal necessary addition. Its Javadoc spells out the difference from `invalidate`.

sunyuhan1998:
@yuqi1129  I have fixed all currently known issues. Please help review it again when you have time. Thank you!

sunyuhan1998:
The CI failure seems to be caused by network fluctuation (unable to connect to the Maven central repository).

core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
yuqi1129: > It's only invoked from cache.put(entity) (i.e. store.put / store.get cache-miss backfill). But grantPrivilegesToRole / overridePrivilegesInRole / grantRolesToUser / grantRolesToGroup go through store.update, which calls cache.invalidate(ident, type) — not cache.put — so invalidateOnKeyChange never fires for them, and those are exactly the main trigger paths.

For the first one, I think you can call `invalidateOnKeyChange` in the `store.update` as what you did in this PR. 
If I missed something, please correct me.

> Dropping a single relation entry without a BFS cascade needs a dedicated method (the existing invalidate(ident,type,relType) is also BFS), so invalidateRelationEntry is the minimal necessary addition. Its Javadoc spells out the difference from invalidate.

What's the problem of invaliding entity with BFS? Can you provide me with more details?

sunyuhan1998: **On (1) — calling `invalidateOnKeyChange` from `store.update`**: you're right — that works. The blocker is (2): what `invalidateOnKeyChange` does internally.

 **On (2) — the BFS problem.** The key point is: `invalidate(ident, type, relType)` does **not** just drop the one relation entry. It calls `invalidateEntities`, which is a BFS over the **shared** reverse index + cacheIndex:


https://github.com/apache/gravitino/blob/4ada1e9b6644aa4691a54f8350cfc80c1a6889fa/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java#L457-L496

 ```
...
 cacheData.invalidate(currentKey); //evicts entity *cache*, not just the relation row
...
 ```

 Starting from the object, the walk follows `reverseIndex.getValuesForKeysStartingWith(objectIdent)` — i.e. **every role bound to that object** — evicts each role's own entity cache, and removes the whole `reverseIndex[objectIdent]` entry. So one grant reaches far beyond the single relation row.

 To make it concrete, this is exactly what `testInvalidRelationCache` exercises, and I verified it empirically :
 - `readRole` and `writeRole` both bind `fileset`. Putting each one builds `reverseIndex["fileset:FILESET"] = [readRole, writeRole]` and caches each role's own entity.
 - After `delete(readRole)`, that reverse-index entry is `[writeRole]` (the test asserts `size == 1` here).
 - `store.put(readRole, true)` then re-adds readRole, so the entry is back to `[writeRole, readRole]`.
 - Now if the helper calls `invalidate(fileset, FILESET, METADATA_OBJECT_ROLE_REL)`, the BFS walks the reverse index from `fileset`, evicts **both** roles' entity caches, and deletes the whole `reverseIndex["fileset:FILESET"]` entry.
 - The test then does `store.get(readRole)`, which re-loads only readRole, so the entry becomes `[readRole]` — writeRole is never re-loaded. The next assertion expects `size == 2` but gets `1`, hence the failure.

 So BFS is not a query-correctness bug per se (a re-query rebuilds), but it (1) breaks the existing `testInvalidRelationCache` reverse-index invariant, and (2) over-invalidates: granting one role evicts the entity caches of **every** role bound to the same object — extra DB round-trips for them, and their reverse-index mappings are gone until re-loaded.

 `invalidateRelationEntry` avoids this precisely because it does **not** call `invalidateEntities`: it does only `cacheData.invalidate(relationKey)` + `reverseIndex.remove(relationKey)` (which trims just this key's own bookkeeping in `entityToReverseIndexMap`/`reverseIndex`, no walking) + `cacheIndex.remove`. No cascade → other roles' caches and the shared reverse-index entry stay intact.

yuqi1129 (approved):

PR #9460, opened 2025-12-11
[#9459] improve(catalog-hive/hudi) Refactor the Hive and Hudi catalogs to use the HiveClient

diqiu50:
### What changes were proposed in this pull request?

Refactor the Hive and Hudi catalogs to use the HiveClient, enabling support for both Hive2 and Hive3 Metastore.

### Why are the changes needed?

Fix: #9459 

### Does this PR introduce _any_ user-facing change?

Update docs

### How was this patch tested?

UT and ITs

PR #11705, opened 2026-06-17
[#11704] fix(core): add org.apache.gravitino.hive.* prefix to IsolatedClassLoader.isCatalogClass()

### LINKED ISSUE #11704: [Bug report] bug(catalog-hive): IsolatedClassLoader fails to recognize org.apache.gravitino.hive.* as catalog classes after #9460 refactoring
### Version

main branch

### Bug Description

Since PR #9460 refactored the Hive catalog to use `HiveClient`, several classes were
moved from `org.apache.gravitino.catalog.hive.*` to a new package
`org.apache.gravitino.hive.*` (e.g., `HiveExceptionConverter`, `HiveClientPool`).

However, `IsolatedClassLoader.isCatalogClass()` was not updated to recognize the new
package prefix. As a result, these classes are incorrectly treated as **shared classes**
and delegated to the server classloader for loading.

### Root Cause

`isCatalogClass()` only checks for `org.apache.gravitino.catalog.*` sub-packages:

```java
private boolean isCatalogClass(String name) {
    return name.startsWith("org.apache.gravitino.catalog")
        && (name.startsWith("org.apache.gravitino.catalog.hive.")
            || ...);
}
```

Classes under `org.apache.gravitino.hive.*` are not matched, so `isSharedClass()`
returns `true` for them. The server classloader then tries to load these classes but
cannot find them in the server classpath (they only exist under
`/gravitino/catalogs/hive/libs/`).

In practice, the server classloader falls back to `super.loadClass()` (URLClassLoader
over execJars) and succeeds for the class itself. But the **compiler-generated synthetic
class `$1`** (produced by `switch(enum)` statements) is loaded through the defining
classloader of the enclosing class. If the enclosing class happens to be loaded by the
server classloader in a specific timing window, `$1` is also requested from the server
classloader — which cannot find it. **The JVM permanently caches this load failure**,
causing all subsequent calls to throw `NoClassDefFoundError` until the process restarts.

### Impact

- `NoClassDefFoundError: org/apache/gravitino/hive/client/HiveExceptionConverter$1`
- Affects any API that triggers `HiveExceptionConverter.toNoSuchObjectException()`
  (e.g., loading a non-existent table via Hive catalog)
- Failure is **permanent** within the process lifetime — not recoverable without restart
- Appears intermittent because it depends on classloader initialization timing

### Fix

Add `org.apache.gravitino.hive.` as a recognized catalog class prefix in
`IsolatedClassLoader.isCatalogClass()`:

```java
private boolean isCatalogClass(String name) {
    return name.startsWith("org.apache.gravitino.hive.")   // ← add this
        || (name.startsWith("org.apache.gravitino.catalog")
            && (name.startsWith("org.apache.gravitino.catalog.hive.")
                || ...));
}
```

### Steps to Reproduce

1. Deploy Gravitino with Hive catalog
2. Call an API that triggers `HiveExceptionConverter` with a non-existent resource
   (e.g., `loadTable` on a table that does not exist in Hive metastore)
3. Under specific classloader initialization timing, observe
   `NoClassDefFoundError: org/apache/gravitino/hive/client/HiveExceptionConverter$1`
4. All subsequent calls to the same catalog on this process instance fail with the
   same error until restart

### Related

- Introduced by: #9460

java.lang.NoClassDefFoundError: org/apache/gravitino/hive/client/HiveExceptionConverter$1
        at org.apache.gravitino.hive.client.HiveExceptionConverter.toNoSuchObjectException(HiveExceptionConverter.java:233)
        at org.apache.gravitino.hive.client.HiveExceptionConverter.toGravitinoException(HiveExceptionConverter.java:77)
        at org.apache.gravitino.catalog.hive.HiveCatalogOperations.loadTable(HiveCatalogOperations.java:...)
        at org.apache.gravitino.catalog.SchemaOperationDispatcher.lambda$...
        at org.apache.gravitino.utils.IsolatedClassLoader.withClassLoader(IsolatedClassLoader.java:86)
        ...

### How to reproduce

+ Which Gravitino version to use
0.9.x / main (after #9460 merged)

+ Steps
1. Deploy Gravitino with a Hive catalog configured
2. Under specific JVM classloader initialization timing (e.g. first
   request to a catalog that has never been accessed, or after a
   long-running process where classloader init races with a request),
   call any API that triggers HiveExceptionConverter internally —
   for example, load a table that does not exist in Hive Metastore:

   GET /api/metalakes/{metalake}/catalogs/{hive-catalog}/schemas/{schema}/tables/{non-existent-table}

3. Observe NoClassDefFoundError: org/apache/gravitino/hive/client/HiveExceptionConverter$1
4. All subsequent requests to the same catalog on this process instance
   fail with the same error — the JVM permanently caches the load
   failure until the process is restarted.

+ Note
The bug is intermittent because it depends on whether
IsolatedClassLoader is initialized before or during the first request.
Once triggered, it is permanent within the process lifetime.
Introduced by #9460 which moved HiveExceptionConverter to
org.apache.gravitino.hive.client.* without updating isCatalogClass().

### Additional context

+ Which Gravitino version to use
0.9.x / main (after #9460 merged)

+ Steps
1. Deploy Gravitino with a Hive catalog configured
2. Under specific JVM classloader initialization timing (e.g. first
   request to a catalog that has never been accessed, or after a
   long-running process where classloader init races with a request),
   call any API that triggers HiveExceptionConverter internally —
   for example, load a table that does not exist in Hive Metastore:

   GET /api/metalakes/{metalake}/catalogs/{hive-catalog}/schemas/{schema}/tables/{non-existent-table}

3. Observe NoClassDefFoundError: org/apache/gravitino/hive/client/HiveExceptionConverter$1
4. All subsequent requests to the same catalog on this process instance
   fail with the same error — the JVM permanently caches the load
   failure until the process is restarted.

+ Note
The bug is intermittent because it depends on whether
IsolatedClassLoader is initialized before or during the first request.
Once triggered, it is permanent within the process lifetime.
Introduced by #9460 which moved HiveExceptionConverter to
org.apache.gravitino.hive.client.* without updating isCatalogClass().

JoegenUSTC:

Add `org.apache.gravitino.hive.` as a recognized catalog class prefix in
`IsolatedClassLoader.isCatalogClass()`, and add unit tests to cover the
classification of all hive-related packages.

Fix: #11704

PR #9460 refactored the Hive catalog and moved classes such as
`HiveExceptionConverter` from `org.apache.gravitino.catalog.hive.*` to
`org.apache.gravitino.hive.*`. However, `isCatalogClass()` was not updated,
so classes under `org.apache.gravitino.hive.*` are incorrectly treated as
shared classes and delegated to the server classloader.

In certain classloader initialization timing windows, the server classloader
becomes the defining classloader for `HiveExceptionConverter`. Its
compiler-generated synthetic class `$1` (produced by `switch`-on-enum) is
then requested from the server classloader, which cannot find it in the server
classpath. **The JVM permanently caches this load failure for the process
lifetime**, causing all subsequent calls to throw `NoClassDefFoundError` until
the process is restarted.

No. This is an internal classloader fix with no API or behavior change.

Added `TestIsolatedClassLoader` with four test methods covering:
- `org.apache.gravitino.hive.*` classes are recognized as catalog classes
- `org.apache.gravitino.catalog.hive.*` classes continue to be recognized
- Other catalog packages (`lakehouse`, `jdbc`, `kafka`, etc.) are unaffected
- Server-side shared classes are correctly excluded

### DIFF
--- a/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
+++ b/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
@@ -240,16 +240,23 @@ public class IsolatedClassLoader implements Closeable {
    * @return true if the class is a catalog class, false otherwise.
    */
   private boolean isCatalogClass(String name) {
-    return name.startsWith("org.apache.gravitino.catalog")
-        && (name.startsWith("org.apache.gravitino.catalog.hive.")
-            || name.startsWith("org.apache.gravitino.catalog.lakehouse.")
-            || name.startsWith("org.apache.gravitino.catalog.jdbc.")
-            || name.startsWith("org.apache.gravitino.catalog.mysql.")
-            || name.startsWith("org.apache.gravitino.catalog.postgresql.")
-            || name.startsWith("org.apache.gravitino.catalog.doris.")
-            || name.startsWith("org.apache.gravitino.catalog.fileset.")
-            || name.startsWith("org.apache.gravitino.catalog.model.")
-            || name.startsWith("org.apache.gravitino.catalog.kafka."));
+    // org.apache.gravitino.hive.* covers classes moved to the shared hive-metastore-common
+    // module by the HiveClient refactoring (e.g. HiveExceptionConverter). Without this prefix
+    // those classes are treated as shared and loaded by the server classloader; their
+    // compiler-generated synthetic classes (e.g. $1 from switch-on-enum) are then requested
+    // from the server classloader which cannot find them, causing a permanent
+    // NoClassDefFoundError that is cached by the JVM for the lifetime of the process.
+    return name.startsWith("org.apache.gravitino.hive.")
+        || (name.startsWith("org.apache.gravitino.catalog")
+            && (name.startsWith("org.apache.gravitino.catalog.hive.")
+                || name.startsWith("org.apache.gravitino.catalog.lakehouse.")
+                || name.startsWith("org.apache.gravitino.catalog.jdbc.")
+                || name.startsWith("org.apache.gravitino.catalog.mysql.")
+                || name.startsWith("org.apache.gravitino.catalog.postgresql.")
+                || name.startsWith("org.apache.gravitino.catalog.doris.")
+                || name.startsWith("org.apache.gravitino.catalog.fileset.")
+                || name.startsWith("org.apache.gravitino.catalog.model.")
+                || name.startsWith("org.apache.gravitino.catalog.kafka.")));
   }
 
   /**
--- a/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
+++ b/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
@@ -247,16 +247,15 @@ public class IsolatedClassLoader implements Closeable {
     // from the server classloader which cannot find them, causing a permanent
     // NoClassDefFoundError that is cached by the JVM for the lifetime of the process.
     return name.startsWith("org.apache.gravitino.hive.")
-        || (name.startsWith("org.apache.gravitino.catalog")
-            && (name.startsWith("org.apache.gravitino.catalog.hive.")
-                || name.startsWith("org.apache.gravitino.catalog.lakehouse.")
-                || name.startsWith("org.apache.gravitino.catalog.jdbc.")
-                || name.startsWith("org.apache.gravitino.catalog.mysql.")
-                || name.startsWith("org.apache.gravitino.catalog.postgresql.")
-                || name.startsWith("org.apache.gravitino.catalog.doris.")
-                || name.startsWith("org.apache.gravitino.catalog.fileset.")
-                || name.startsWith("org.apache.gravitino.catalog.model.")
-                || name.startsWith("org.apache.gravitino.catalog.kafka.")));
+        || name.startsWith("org.apache.gravitino.catalog.hive.")
+        || name.startsWith("org.apache.gravitino.catalog.lakehouse.")
+        || name.startsWith("org.apache.gravitino.catalog.jdbc.")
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
```diff
@@ -240,16 +240,23 @@ private boolean isSharedClass(String name) {
...
-            || name.startsWith("org.apache.gravitino.catalog.kafka."));
+    // org.apache.gravitino.hive.* covers classes moved to the shared hive-metastore-common
+    // module by the HiveClient refactoring (e.g. HiveExceptionConverter). Without this prefix
+    // those classes are treated as shared and loaded by the server classloader; their
+    // compiler-generated synthetic classes (e.g. $1 from switch-on-enum) are then requested
+    // from the server classloader which cannot find them, causing a permanent
+    // NoClassDefFoundError that is cached by the JVM for the lifetime of the process.
+    return name.startsWith("org.apache.gravitino.hive.")
+        || (name.startsWith("org.apache.gravitino.catalog")
+            && (name.startsWith("org.apache.gravitino.catalog.hive.")
```
diqiu50: The condition `name.startsWith("org.apache.gravitino.catalog")` is redundant.
Removing it would make the code more readable.

JoegenUSTC: @diqiu50 
Thanks for the review! You're right — the outer name.startsWith("org.apache.gravitino.catalog") is indeed redundant since every inner sub-prefix (e.g. org.apache.gravitino.catalog.hive.) already implies it. I've removed the wrapping AND and flattened the conditions into a straightforward OR chain. The semantics remain identical.

diqiu50 (approved):
LGTM

Issue #11590, opened 2026-06-11
[Improvement] Upgrade Gravitino Doris catalog compatibility from 1.2.x to 3.0.x / 4.0.x


jiangxt2:
**Version**
main branch

**Describe the improvement**

Gravitino's Doris catalog declares support for Doris 1.2.x, which has been EOL for over two years. This upgrade targets compatibility with **Doris 4.0.6** (current stable) and **Doris 3.0.6.2** (latest LTS patch), bringing Gravitino's Doris support in line with what users are actually running in production.

**Compatibility testing results**

Gravitino 1.2.0 was tested against Doris 3.0.x and the following incompatibilities were identified (testing performed by @HaozeeWu):

| # | Issue | Affected Version | Impact | Severity |
|---|-------|-----------------|--------|----------|
| 1 | `DorisUtils.extractPartitionInfoFromSql()` regex fails on Doris 3.0+ `SHOW CREATE TABLE` output for LIST partitions (space between `LIST` and `(`) | 3.0.x+ | LIST partition operations broken — partitioned tables misidentified as non-partitioned; RANGE partitions are not affected | High |
| 2 | `DorisTableOperations.appendIndexesSql()` generates `INDEX PRIMARY (id)` syntax invalid in Doris 3.0+ (`PRIMARY` is reserved word) | 3.0.x+ | CREATE TABLE / ALTER TABLE with indexes fails | High |
| 3 | `DATE` type disabled in Doris 4.0.x (`disable_datev1=true`), `fromGravitino(DateType)` outputs `"date"` | 4.0.x | CREATE TABLE with date columns fails on 4.0.x | High |
| 4 | `DATETIMEV2` precision lost — `COLUMN_SIZE`/`DECIMAL_DIGITS` return null in JDBC, `calculateDatetimePrecision()` returns 0 | 3.0.x+ (always reproducible on 4.0.x) | `datetimev2(3)` read back as `datetime(0)` | Medium |
| 5 | Auto Increment blocked — `validateIncrementCol()` hard-throws for all cases | 2.1+ | Cannot create tables with auto-increment columns | Medium |

**Proposed tasks**

- [ ] Fix partition regex for Doris 3.0+ SHOW CREATE TABLE format (#1)
- [ ] Fix INDEX syntax for Doris 3.0+ compatibility (#2)
- [ ] Fix `date` → `datev2` for 4.0.x compatibility (#3)
- [ ] Fix DATETIME precision loss (#4)
[...]

PR #11715, opened 2026-06-17
[#11590] improvement(doris): upgrade Doris catalog compatibility from 1.2.x to 3.0.x / 4.0.x

jiangxt2:

**Updated 2026-06-18**: Split into 6 focused PRs per reviewer feedback. This PR now contains only type system changes (PR 1 of 6).

### Type Converter Enhancement (`DorisTypeConverter.java`)

- **datev2 fix**: `fromGravitino(DateType)` now returns `"datev2"` instead of `"date"` for Doris 4.0.x compatibility (`disable_datev1=true`)
- **Type mapping expansion**: Added mappings for binary, varbinary, json, variant, ipv4, ipv6, largeint, bitmap, hll → `BinaryType` / `ExternalType`
- **Complex type parsing**: Recursive parsing of `array<int(11)>`, `map<varchar(20),int(11)>`, `struct<x:int(11),y:int(11)>` → `ListType` / `MapType` / `StructType`

### DATETIME Precision Fix (`DorisTableOperations.java`)

- `calculateDatetimePrecision()` now parses precision from `"datetime(N)"` type string instead of relying on `COLUMN_SIZE - 20` (DATETIMEV2 returns `COLUMN_SIZE=0` in Doris 4.0.x)

### Parameter Type Parsing Fix (`DorisTypeConverter.java`)

- `toGravitino()` extracts base type name from parameterized types (`"int(11)"` → `"int"`, `"decimal(10,2)"` → `"decimal"`) before switch matching, preventing scalar types with parameters from falling into `ExternalType` fallback

No. Type mapping improvements are transparent to users.

## How was this patch tested?

Unit tests:
- `TestDorisTypeConverter`: datev2, DATETIME precision (0/3/6), binary, json/variant/ip/largeint/bitmap/hll, standalone parameterized types `int(11)`/`decimal(10,2)`/`varchar(100)`/`datetime(3)`, nested complex types
- `TestDorisTableOperations`: `calculateDatetimePrecision` for DATE/DATETIME/VARCHAR

Integration tests (Doris 4.0.6 and 3.0.6.2):
- datev2 table creation verified
- DATETIME precision (0/3/6) readback verified
- Type mapping for all new types verified

Related to #11590

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
@@ -798,7 +798,8 @@ public class DorisTableOperations extends JdbcTableOperations {
     String upperTypeName = typeName.toUpperCase();
 
     // Check driver version compatibility first
-    boolean isDatetimeType = "DATETIME".equals(upperTypeName);
+    boolean isDatetimeType =
+        "DATETIME".equals(upperTypeName) || upperTypeName.startsWith("DATETIME(");
 
     if (isDatetimeType) {
       String driverVersion = getMySQLDriverVersion();
@@ -809,6 +810,17 @@ public class DorisTableOperations extends JdbcTableOperations {
             driverVersion,
             upperTypeName,
             driverVersion);
+        return null;
+      }
+    }
+
+    // Handle datetime(N) format from SHOW CREATE TABLE
+    if (upperTypeName.startsWith("DATETIME(") && upperTypeName.endsWith(")")) {
+      try {
+        String precisionStr = upperTypeName.substring(9, upperTypeName.length() - 1);
+        return Integer.parseInt(precisionStr);
+      } catch (NumberFormatException e) {
+        LOG.warn("Failed to parse datetime precision from type: {}", typeName);
         return null;
       }
     }
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/converter/DorisTypeConverter.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/converter/DorisTypeConverter.java
@@ -18,7 +18,8 @@
  */
 package org.apache.gravitino.catalog.doris.converter;
 
-import java.util.Optional;
+import java.util.ArrayList;
+import java.util.List;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcTypeConverter;
 import org.apache.gravitino.rel.types.Type;
 import org.apache.gravitino.rel.types.Types;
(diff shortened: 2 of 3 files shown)


FANNG1:
Thanks for tackling the Doris 3.0/4.0 compatibility work — this is valuable. One process suggestion: I think this PR is worth splitting.

It currently bundles ~6–7 independent concerns (+1185/−84). They're cleanly decoupled, and mixing the low-risk changes (type/enum mapping, removing a validation) with the one stateful, higher-risk piece (the `SHOW CREATE TABLE` cache + generated-column parsing) means the whole PR is gated on its riskiest part and is hard to review thoroughly in a single pass.

Suggested split, ordered by dependency/risk:

1. **Type system** — `DorisTypeConverter`: `datev2` + new type mappings + `datetime(N)` precision. Pure functions, easy to review/merge first.
2. **Index syntax** — `DorisTableOperations`: `USING` clause, `UNIQUE KEY` model, `mapDorisIndexType`, ADD/DROP back-quoting. Independent, low risk.
3. **Partition compatibility** — `DorisUtils`: multi-line / space-tolerant regex + LIST value assignments. Self-contained.
4. **Complex / generated column loading** — `DorisTableOperations`: `getColumnBuilder` override + `SHOW CREATE TABLE` cache + generated-column extraction. This is the only part that introduces cross-call state, so isolating it lets reviewers focus on cache lifecycle and parsing edge cases, and lets it be reverted independently. (Depends on 1 for type parsing.)
5. **Auto-increment enablement** — tiny; standalone or folded into 2.

Dependencies: 4 depends on 1; the other three are independent and can proceed in parallel.

I also did a review pass, and the correctness issues I found cluster exactly in 3 and 4 — for example, the generated-column splitter increments angle-bracket depth on every `<` (so a `<` inside an `AS (...)` expression breaks column splitting and drops later generated columns), and the `SHOW CREATE TABLE` cache is never invalidated, so `loadTable` after an `ALTER` can return stale column types. That clustering is the main reason I'd isolate those two. Happy to post the detailed findings as inline comments if helpful.

FANNG1 (changes requested):
Thanks for splitting this PR down to the type-system scope. I found several issues that should be addressed before merge:

1. `fromGravitino()` now emits Doris `array/map/struct` types, but the existing Doris table operation test still treats `ListType`, `MapType`, and `StructType` as unsupported. Please update the test expectation or clarify whether creating tables with complex types is in scope for this PR.

2. The complex type parser is only one level deep. Doris supports nested complex types such as `array<array<int>>`, `map<string,array<int>>`, and `struct<a:map<string,int>>`, but `parseSimpleType()` falls back to `ExternalType` for nested `array/map/struct` values. This can make loaded schemas differ from the original Gravitino schema.

3. `MapType` should validate that the key type is scalar. Doris does not support complex map keys, but the current `fromGravitino(MapType)` can generate unsupported SQL such as `map<array<int>,int>`.

Also, please add reverse/round-trip tests for `array/map/struct`, nested complex type tests, invalid map key tests, and a direct test for `calculateDatetimePrecision("DATETIME(N)", ...)`.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -813,6 +814,17 @@ public Integer calculateDatetimePrecision(String typeName, int columnSize, int s
       }
     }
 
+    // Handle datetime(N) format from SHOW CREATE TABLE
+    if (upperTypeName.startsWith("DATETIME(") && upperTypeName.endsWith(")")) {
+      try {
+        String precisionStr = upperTypeName.substring(9, upperTypeName.length() - 1);
```
FANNG1, line 820: Nit: please replace the hard-coded `9` with `"DATETIME(".length()` or a named constant so the substring boundary is self-documenting.

jiangxt2:
Closing this PR to resubmit with all reviewer feedback addressed. Will open a new PR shortly.

PR #11717, opened 2026-06-17
[MINOR] fix(oceanbase): Avoid MySQL driver version check

jarredhj0214:

This PR introduces `OceanBaseCatalogOperations` for the OceanBase JDBC catalog and makes `OceanBaseCatalog` use it instead of `MySQLProtocolCompatibleCatalogOperations`.

The new operations class extends the generic `JdbcCatalogOperations`, so OceanBase catalog initialization no longer applies the MySQL Connector/J major-version check. It still shuts down MySQL Connector/J cleanup resources when present, which keeps the MySQL protocol driver path covered.

A unit test is added to verify that OceanBase catalog uses `OceanBaseCatalogOperations` and does not use `MySQLProtocolCompatibleCatalogOperations`.

OceanBase catalog supports both MySQL protocol URLs and OceanBase protocol URLs. Users may configure either MySQL Connector/J or the OceanBase official JDBC driver.

Currently, OceanBase catalog reuses `MySQLProtocolCompatibleCatalogOperations`, which rejects drivers whose major version is lower than 8. This can incorrectly reject OceanBase JDBC drivers, for example `oceanbase-client-2.4.18.jar`, during catalog creation with an error like:

`Mysql catalog does not support the jdbc driver version 2.4.18, minimal required version is 8.0`

This issue is not specific to `2.4.18`; OceanBase JDBC driver 4.x can also be incorrectly rejected because its major version is still lower than 8.

Fix: N/A

Yes. Creating an OceanBase catalog with the OceanBase JDBC driver is no longer incorrectly blocked by the MySQL Connector/J version check.

No public API or catalog property is added or removed.

Added unit coverage in `TestOceanBaseCatalogOperations`.

`git diff --check` passed.

Gradle tests were not run locally because the current machine only has JDK 8 installed, while Gravitino requires JDK 17 to build.

### DIFF
--- a/catalogs-contrib/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/OceanBaseCatalogOperations.java
+++ b/catalogs-contrib/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/OceanBaseCatalogOperations.java
@@ -18,24 +18,20 @@
  */
 package org.apache.gravitino.catalog.oceanbase;
 
-import org.apache.gravitino.catalog.jdbc.JdbcCatalogOperations;
+import org.apache.gravitino.catalog.jdbc.MySQLProtocolCompatibleCatalogOperations;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcColumnDefaultValueConverter;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcExceptionConverter;
 import org.apache.gravitino.catalog.jdbc.converter.JdbcTypeConverter;
 import org.apache.gravitino.catalog.jdbc.operation.JdbcDatabaseOperations;
 import org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Operations for interacting with the OceanBase catalog in Apache Gravitino.
  *
- * <p>OceanBase supports both MySQL Connector/J and the OceanBase JDBC driver, so it should not use
- * {@code MySQLProtocolCompatibleCatalogOperations}, which rejects non-MySQL Connector/J drivers by
- * major version during catalog initialization.
+ * <p>OceanBase supports both MySQL Connector/J and the OceanBase JDBC driver, so it should skip the
+ * MySQL Connector/J major-version check during catalog initialization.
  */
-public class OceanBaseCatalogOperations extends JdbcCatalogOperations {
-  private static final Logger LOG = LoggerFactory.getLogger(OceanBaseCatalogOperations.class);
+public class OceanBaseCatalogOperations extends MySQLProtocolCompatibleCatalogOperations {
 
   /**
    * Constructs a new instance of {@link OceanBaseCatalogOperations}.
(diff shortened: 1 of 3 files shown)


catalogs-contrib/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/OceanBaseCatalogOperations.java
```diff
@@ -0,0 +1,80 @@
...
+  public void close() {
+    super.close();
+
+    try {
+      Class.forName("com.mysql.cj.jdbc.AbandonedConnectionCleanupThread")
+          .getMethod("uncheckedShutdown")
+          .invoke(null);
+      LOG.info("AbandonedConnectionCleanupThread has been shutdown...");
+    } catch (ClassNotFoundException e) {
+      LOG.debug("MySQL AbandonedConnectionCleanupThread was not found", e);
```
geyanggang: since the ClassNotFoundException is completely expected when OceanBase's own driver is used (no MySQL classes on classpath), logging the full stack trace at DEBUG level is unnecessary noise. Consider dropping the exception parameter:

} catch (ClassNotFoundException e) {
    LOG.debug("MySQL AbandonedConnectionCleanupThread not found on classpath, skipping shutdown");
}

catalogs-contrib/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/TestOceanBaseCatalogOperations.java
```diff
@@ -0,0 +1,52 @@
...
+
+    Assertions.assertDoesNotThrow(operations::checkJDBCDriverVersion);
+  }
+
+  private static class TestableOceanBaseCatalog extends OceanBaseCatalog {
+    @Override
+    protected CatalogOperations newOps(Map<String, String> config) {
+      return super.newOps(config);
+    }
+  }
```
geyanggang: newOps is protected and the test class is in the same package as OceanBaseCatalog, so you can call it directly without the TestableOceanBaseCatalog subclass.

catalogs-contrib/catalog-jdbc-oceanbase/src/test/java/org/apache/gravitino/catalog/oceanbase/TestOceanBaseCatalogOperations.java
```diff
@@ -0,0 +1,52 @@
...
+    Assertions.assertFalse(operations instanceof MySQLProtocolCompatibleCatalogOperations);
+  }
+
+  @Test
+  public void testCheckJDBCDriverVersionDoesNotRejectOceanBaseDriverVersion() {
+    OceanBaseCatalogOperations operations =
+        new OceanBaseCatalogOperations(null, null, null, null, null);
+
+    Assertions.assertDoesNotThrow(operations::checkJDBCDriverVersion);
+  }
```
geyanggang, line 44: Mind adding a quick comment here to explain what this test is for?

jarredhj0214:
Thanks for the review. I have addressed all three comments:
- Removed the exception parameter from the expected ClassNotFoundException debug log.
- Removed the unnecessary TestableOceanBaseCatalog subclass and call newOps directly.
- Added a short comment explaining the JDBC driver version test.
@geyanggang

catalogs-contrib/catalog-jdbc-oceanbase/src/main/java/org/apache/gravitino/catalog/oceanbase/OceanBaseCatalogOperations.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Operations for interacting with the OceanBase catalog in Apache Gravitino.
+ *
+ * <p>OceanBase supports both MySQL Connector/J and the OceanBase JDBC driver, so it should not use
+ * {@code MySQLProtocolCompatibleCatalogOperations}, which rejects non-MySQL Connector/J drivers by
+ * major version during catalog initialization.
+ */
+public class OceanBaseCatalogOperations extends JdbcCatalogOperations {
```
yuqi1129: You can directly extend `MySQLProtocolCompatibleCatalogOperations` and overwrite `checkJDBCDriverVersion` to simply the code.

jarredhj0214: Thanks for the suggestion. I have updated `OceanBaseCatalogOperations` to extend `MySQLProtocolCompatibleCatalogOperations` directly and override `checkJDBCDriverVersion()` to skip the MySQL Connector/J major-version check for OceanBase catalog initialization.

yuqi1129 (approved):

PR #11727, opened 2026-06-18
[#11194] docs: Add design doc for iceberg expire-snapshots maintenance job

### LINKED ISSUE #11194: [FEATURE] Add builtin-iceberg-expire-snapshots maintenance job
### Describe the feature

Add a new built-in Iceberg maintenance job `builtin-iceberg-expire-snapshots` that expires old snapshots from Iceberg tables. This would use Iceberg's `ExpireSnapshots` action (via Spark's `expire_snapshots` procedure) to remove snapshot metadata and associated data files that are no longer needed, preventing unbounded metadata growth.

### Motivation

Without periodic snapshot expiration, Iceberg table metadata grows indefinitely, accumulating snapshot JSON files and manifest lists that slow down table operations and waste storage. The existing built-in jobs (`builtin-iceberg-rewrite-data-files` and `builtin-iceberg-update-stats`) cover data compaction and metrics but don't address metadata cleanup. This is one of the most critical Iceberg housekeeping operations.

PR #10500 (merged) already added Trino-side delegation for `expire_snapshots` as a procedure, but there is no server-side built-in job that can be triggered automatically via the Table Maintenance Service (Optimizer) policies.

### Describe the solution

- Create `IcebergExpireSnapshotsJob` in `maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/iceberg/` following the same pattern as `IcebergRewriteDataFilesJob`
- Register it in `BuiltInJobTemplateProvider` as `builtin-iceberg-expire-snapshots`
- Support configurable parameters: `older_than` (timestamp), `retain_last` (number of snapshots to keep), `stream_results`
- Optionally add a corresponding policy type (e.g., `system_iceberg_snapshot_expiration`) similar to `system_iceberg_compaction`

### Additional context

Related: #10280 (closed: Trino-side procedure delegation), #8864 (umbrella: add built-in jobs for table maintenance)

laserninja:

Add a design document for the built-in Iceberg expire-snapshots maintenance job (`design-docs/iceberg-expire-snapshots-maintenance-job.md`).

The design covers the full end-to-end architecture:
- **Layer 1 (Policy):** New `IcebergSnapshotExpirationContent` policy content class and `ICEBERG_SNAPSHOT_EXPIRATION` built-in type
- **Layer 2 (Strategy):** `SnapshotExpirationStrategyHandler` operating at the table level (no partition scoring)
- **Layer 3 (Adapter):** `GravitinoSnapshotExpirationJobAdapter` converting strategy results to job config
- **Layer 4 (Job):** `IcebergExpireSnapshotsJob` executing via Spark (implementation in PR #11206)

Key design decisions:
- All expiration parameters (`olderThanDays`, `retainLast`, `streamResults`) are **independently optional** — users can specify any combination
- Snapshot expiration operates at the table level (no partition statistics needed)
- Trigger based on `custom-snapshot-count` statistic

As requested by reviewers in PR #11206, the design document should be submitted as a dedicated PR before the implementation PRs.

Fix: #11194

No. Documentation only.

N/A — documentation only.

### DIFF
--- a/design-docs/iceberg-expire-snapshots-maintenance-job.md
+++ b/design-docs/iceberg-expire-snapshots-maintenance-job.md
@@ -1,0 +1,516 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Design: Built-in Iceberg Expire Snapshots Maintenance Job
+
+| Field   | Value                                                                |
+| ------- | -------------------------------------------------------------------- |
+| Status  | Draft                                                                |
+| Authors | @laserninja                                                          |
+| Created | 2026-05-26                                                           |
+| Issue   | [#11194](https://github.com/apache/gravitino/issues/11194)           |
+| Module  | `api`, `maintenance/jobs`, `maintenance/optimizer`                   |
+
+---
+
+## 1. Background
+
+Without periodic snapshot expiration, Iceberg table metadata grows
+indefinitely — accumulating snapshot JSON files and manifest lists that
+slow down table operations and waste storage. The existing built-in
+maintenance jobs (`builtin-iceberg-rewrite-data-files` for data compaction
+and `builtin-iceberg-update-stats` for metrics) address data file
+optimization but do not cover metadata cleanup.
+
+PR [#10500](https://github.com/apache/gravitino/pull/10500) added
+Trino-side delegation for `expire_snapshots` as a procedure, but there is
+no server-side built-in job that can be triggered automatically via the
+Table Maintenance Service (Optimizer) policies.
+
+This design proposes adding full end-to-end support for Iceberg snapshot
+expiration: from policy definition through strategy evaluation to Spark
+job execution.
+
+---
+
+## 2. Goals
+
+1. Add a new built-in policy type `system_iceberg_snapshot_expiration` for
+   declarative snapshot expiration configuration.
+2. Add a strategy handler that evaluates when snapshot expiration should
+   run based on table statistics (snapshot count, age).
+3. Add a job adapter that converts strategy evaluation results into job
+   configurations.
+4. Add the Spark job that executes Iceberg's `expire_snapshots` procedure.
+5. Ensure the full flow works end-to-end: policy → strategy → job
+   submission → Spark execution.
+
+---
+
+## 3. Non-Goals
+
+- Orphan file removal (separate Iceberg procedure, separate issue).
+- Automatic policy creation — users must explicitly create and attach
+  policies.
+- Changes to the Optimizer scheduling framework itself.
+
+---
+
+## 4. Existing Architecture Overview
+
+The Gravitino maintenance module follows a layered architecture for
+automated table maintenance. The existing Iceberg compaction flow
+establishes the pattern:
+
+```
+Policy Creation (REST API)
+    ↓
+GravitinoStrategyProvider  (loads policies as strategies)
+    ↓
+CompactionStrategyHandler  (evaluates trigger / score expressions)
+    ↓
+CompactionJobContext → GravitinoCompactionJobAdapter  (converts to job config)
+    ↓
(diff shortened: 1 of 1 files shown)


design-docs/iceberg-expire-snapshots-maintenance-job.md
```diff
@@ -0,0 +1,516 @@
...
+specify only `olderThanDays`, only `retainLast`, or both — the job adapter
+will only pass set parameters to the Spark procedure. This matches the
+Iceberg `expire_snapshots` procedure semantics, where each parameter is
+independent.
+
+| Field             | Type      | Required | Default | Description                                                  |
+| ----------------- | --------- | -------- | ------- | ------------------------------------------------------------ |
+| `olderThanDays`   | `Long`    | No       | 5       | Expire snapshots older than this many days                    |
+| `retainLast`      | `Long`    | No       | 1       | Minimum number of snapshots to always retain                  |
+| `streamResults`   | `Boolean` | No       | false   | Whether to stream deletion results (useful for large tables)  |
```
roryqi: Do we have benefits if streamResults is false?

design-docs/iceberg-expire-snapshots-maintenance-job.md
```diff
@@ -0,0 +1,516 @@
...
+| `minSnapshotCount`| `long`    | No       | 10      | Trigger threshold — only run when snapshot count exceeds this |
+
+**Usage scenarios:**
+
+- **Only `olderThanDays`:** Expire snapshots older than N days, keep the
```
roryqi: Should we provide the default value for `olderThanDays` and `retainLast`? For some users, one condition may be enough.

design-docs/iceberg-expire-snapshots-maintenance-job.md
```diff
@@ -0,0 +1,516 @@
...
+
+---
+
+## 6. Statistics Requirements
+
+The strategy handler needs a **snapshot count** statistic to evaluate
+whether expiration should be triggered.
+
+| Statistic Name          | Type   | Source                 | Description                      |
+| ----------------------- | ------ | ---------------------- | -------------------------------- |
```
roryqi: Could we use the table metadata directly?

design-docs/iceberg-expire-snapshots-maintenance-job.md
laserninja: Not much benefit. `stream_results=false` (the procedure default) collects the whole deleted-file list to the driver in one result set; `true` streams it by RDD partition, which the Iceberg docs recommend to avoid driver OOM on large deletions. This job only consumes the summary counts (`deleted_data_files_count` etc.), never the individual rows, so materializing the full list buys nothing - the only cost of streaming is slight per-partition overhead on small tables, negligible next to the expiration itself.

Since the job runs unattended on tables of unknown size, I flipped the default to `streamResults = true` (deliberately differing from the procedure default) and kept the field configurable. Rationale is in §5.2.4.

design-docs/iceberg-expire-snapshots-maintenance-job.md
laserninja: Agreed, Gravitino shouldn't supply its own defaults here - removed them. When a field is unset the adapter omits the argument and Iceberg falls back to table properties: `older_than` -> `history.expire.max-snapshot-age-ms` (5 days), `retain_last` -> `history.expire.min-snapshots-to-keep` (1).

So a policy that sets only `retainLast` no longer silently applies age-based expiration as well, and a per-table `history.expire.*` setting isn't overridden by the policy. Table and usage scenarios updated in §5.2.3.

design-docs/iceberg-expire-snapshots-maintenance-job.md
laserninja: Yes, and it's the better source - switched to it. Snapshot count is pure metadata with no scan, so routing it through a statistic means the trigger can be stale by up to one `builtin-iceberg-update-stats` interval for a value that's free to read.

One gap to close: the optimizer sees the Gravitino `Table` from `GravitinoTableMetadataProvider`, and `IcebergTablePropertiesUtil.buildReservedProperties` currently exposes only `current-snapshot-id`, `format`, `provider`, `format-version`, `location`, `sort-order`, `identifier-fields` - no snapshot count. So the design now adds a reserved read-only `snapshot-count` property computed from `TableMetadata.snapshots()`, which is already in memory when those properties are built.

That removes `custom-snapshot-count` and `TABLE_STATISTICS` from the design: `dataRequirements()` is just `{TABLE_METADATA}`, and the handler implements `StrategyHandler` directly rather than extending `BaseExpressionStrategyHandler` (the base class builds its expression context from statistics plus numeric rules, so a metadata value isn't addressable from `trigger-expr` / `score-expr`, and a single table-level condition doesn't need the expression machinery). Details in §6; §8 has the extra PR for the property.

laserninja:
@roryqi let me know if you have more comments.

roryqi:
@lasdf1234 Could u take a look again?

lasdf1234:
> @lasdf1234 Could u take a look again?

OK

lasdf1234:
LGTM

lasdf1234 (approved):

roryqi (approved):

PR #11730, opened 2026-06-18
[#11716] docs: Document Web UI OIDC setup with Keycloak (secure context, public client, issuer alignment)

### LINKED ISSUE #11716: [Improvement] Documentation gaps: v2 Web UI with Keycloak OIDC

There are three documentation gaps in the v2 Web UI OIDC setup, confirmed empirically on the OSS build (branch-1.3, 1.3.1-SNAPSHOT) with `oauth` authenticator (`JwksTokenValidator`), `authorization.enable = true`, and Keycloak. The primary page to correct is the security how-to-authenticate guide; a couple also belong on the Web UI v2 page.

**Gap 1: Browser OIDC login requires a secure context (HTTPS or localhost)**

Affected page: security / how-to-authenticate, Web UI section.

The v2 UI OIDC client uses the browser Web Crypto API for PKCE, which the browser only exposes in a secure context. When the UI is served over plain HTTP on a non-localhost host, login fails with the browser console error "Failed to get access token: Error: OIDC login requires the UI to run on HTTPS or localhost. Current origin: http://HOST:PORT," surfaced in the page as "Failed to get UserManager from OIDC provider." Reaching the same server over a localhost origin (e.g. through an SSH or SSM tunnel) succeeds, as does serving it over HTTPS.

The secure-context requirement is not stated anywhere in the setup.

**Gap 2: The UI needs a dedicated public Keycloak client, separate from the confidential machine client**

Affected page: security / how-to-authenticate, Keycloak example.

The browser cannot hold a client secret, so the UI client must be a public client with client authentication off and the standard flow enabled. The confidential client used by machine flows and engines (e.g. Trino) relies on client-credentials and must stay confidential. The two roles require two separate clients in the same realm. The UI client also needs valid redirect URIs set to `UI_ORIGIN/*` and web origins set to `UI_ORIGIN`.

The Keycloak example walks through one client without distinguishing the public UI client from the confidential engine client.

**Gap 3: Issuer must be unified when browser and backend reach Keycloak by different hostnames**

Affected page: security / how-to-authenticate.

Once `gravitino.authenticator.oauth.authority` is set for the UI, the validator requires the token `iss` claim to match that authority. In deployments where the browser reaches Keycloak by one hostname and engines reach it by another, the issuer differs and tokens minted against the other endpoint are rejected with "JWT iss claim value rejected." Setting the realm Frontend URL to a single canonical value makes Keycloak stamp one issuer regardless of the connecting hostname.

The interaction between `authority` and the token issuer is not documented, and there is no guidance for split-hostname deployments.

### How should we improve?

1. **Gap 1**: Add a prominent note that Web UI OIDC login requires a secure context (HTTPS or localhost origin). Recommend HTTPS for shared deployments and a localhost tunnel for local testing.

2. **Gap 2**: In the Keycloak setup, call out two clients explicitly — a confidential client for engines/scripts using client-credentials, and a separate public client for the Web UI with client authentication off, standard flow on, PKCE, redirect URIs of `UI_ORIGIN/*`, and web origins of `UI_ORIGIN`.

3. **Gap 3**: Note that `authority` makes the validator enforce the token issuer, and that all token sources must present the same issuer. Recommend setting the Keycloak realm Frontend URL to one canonical hostname and warn that mixing internal/external hostnames without this will reject otherwise valid tokens.

roryqi:

Fill three documentation gaps in `docs/security/how-to-authenticate.md` for the v2 Web UI OIDC login with Keycloak:

1. **Secure context requirement** — A note in the *OAuth Mode* section explaining that Web UI OIDC login uses the Authorization Code flow with PKCE, which depends on the browser Web Crypto API and therefore requires a secure context (HTTPS or `localhost`). Plain HTTP on a non-`localhost` host fails.
2. **Separate client configurations** — A new *Enable Web UI OIDC login with Keycloak* subsection that distinguishes the confidential client (machine/engine flows) from a public client (`Client authentication` off, `Standard flow` on) required by the browser UI, with the matching `oidc`-provider configuration.
3. **Token issuer alignment** — Guidance to set the Keycloak realm Frontend URL so the `iss` claim matches across split-hostname deployments, avoiding `JWT iss claim value rejected`.

Users setting up the v2 Web UI with Keycloak OIDC hit undocumented failures: login breaks over plain HTTP, a single confidential client does not work for the browser, and split-hostname setups get token rejections. These were reported in #11716.

Fix: #11716

No code change. Documentation only.

Documentation-only change. Verified with the repo docs audit scripts (title/intro/heading/spacing and internal link/anchor checks) — all pass.

### DIFF
--- a/docs/security/how-to-authenticate.md
+++ b/docs/security/how-to-authenticate.md
@@ -107,6 +107,15 @@ To enable OAuth mode:
 - For **static key validation**: Configure `gravitino.authenticator.oauth.defaultSignKey`, `gravitino.authenticator.oauth.serverUri` and `gravitino.authenticator.oauth.tokenPath`.
 - For **JWKS validation**: Configure `gravitino.authenticator.oauth.jwksUri` and `gravitino.authenticator.oauth.tokenValidatorClass=org.apache.gravitino.server.authentication.JwksTokenValidator`. You can use either `gravitino.authenticator.oauth.provider=default` or `gravitino.authenticator.oauth.provider=oidc` depending on whether you want Web UI OIDC login flow.
 - For **Web UI OIDC authentication**: Set `gravitino.authenticator.oauth.provider=oidc` and configure `gravitino.authenticator.oauth.clientId`, `gravitino.authenticator.oauth.authority`, and `gravitino.authenticator.oauth.scope`. These settings are exposed to the Web UI via the `/configs` endpoint to enable OAuth login flow. Configure your OAuth provider with the callback redirect URI: `https://your-gravitino-server/ui/oauth/callback`.
+
+  :::note
+  The Web UI OIDC login uses the Authorization Code flow with PKCE, which depends on the browser
+  [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API). Browsers only
+  expose this API in a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts),
+  that is, when the Web UI is served over HTTPS or from `localhost`. If you open the Web UI over plain
+  HTTP on a non-`localhost` host, login fails because the crypto primitives required for PKCE are
+  unavailable. Serve the Web UI over HTTPS (or access it via `localhost`) for OIDC login to work.
+  :::
 - Next, for the client side, users can enable `OAuth` mode by the following code:
 
 ```java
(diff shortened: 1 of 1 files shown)


mchades:
@danhuawang could you please help to check if the doc is clear?

docs/security/how-to-authenticate.md
```diff
@@ -653,3 +662,64 @@ For Gravitino Iceberg REST Service, the 'Accept: application/vnd.gravitino.v1+js
...
+| `gravitino-ui`     | Public       | Browser Web UI OIDC login                                     | *Client authentication* off, *Standard flow* enabled                                |
+
+To register the public client for the Web UI:
+
+* Click *Clients* in the `gravitinorealm` realm, then *Create client*.
+* Fill in *Client type*: `OpenID Connect`, *Client ID*: `gravitino-ui`, then click *Next*.
+* Turn *Client authentication* **off** (this makes it a public client) and enable *Standard flow*. Click *Next*.
+* Set *Valid redirect URIs* to `https://your-gravitino-server/ui/oauth/callback`.
+* Set *Web origins* to `https://your-gravitino-server`.
+* Click *Save*.
```
danhuawang, line 686: Missing set `Valid post logout redirect URIs` to `https://your-gravitino-server/*` for log out before saving.

docs/security/how-to-authenticate.md
```diff
@@ -653,3 +662,64 @@ For Gravitino Iceberg REST Service, the 'Accept: application/vnd.gravitino.v1+js
...
+gravitino.authenticators = oauth
+gravitino.authenticator.oauth.provider = oidc
+gravitino.authenticator.oauth.clientId = gravitino-ui
+gravitino.authenticator.oauth.authority = http://localhost:8080/realms/gravitinorealm
+gravitino.authenticator.oauth.scope = openid profile email
+gravitino.authenticator.oauth.jwksUri = http://localhost:8080/realms/gravitinorealm/protocol/openid-connect/certs
+gravitino.authenticator.oauth.tokenValidatorClass = org.apache.gravitino.server.authentication.JwksTokenValidator
+gravitino.authenticator.oauth.serviceAudience = account
+gravitino.authenticator.oauth.principalFields = preferred_username,email,sub
+```
```
danhuawang, line 700: When do the user need to set the following items? 
```
gravitino.authenticator.oauth.defaultSignKey = ******
gravitino.authenticator.oauth.serverUri = http://192.168.97.2:30080/
gravitino.authenticator.oauth.tokenPath = /realms/myrealm/protocol/openid-connect/token
```

docs/security/how-to-authenticate.md
roryqi: Good catch. Added a step to set *Valid post logout redirect URIs* to `https://your-gravitino-server/*` before saving, so logout can redirect back to the Web UI.

docs/security/how-to-authenticate.md
roryqi: Those static-key settings (`defaultSignKey`/`serverUri`/`tokenPath`) are only needed when using `StaticSignKeyValidator`. This Web UI example uses JWKS-based validation (`jwksUri` + `JwksTokenValidator`), where the server fetches Keycloak's public keys automatically, so they are not required. I added a paragraph after the config block clarifying this and noting the same server config validates both Web UI and machine tokens.

danhuawang:
LGTM

jerryshao (approved):

Issue #3272, opened 2024-05-06
[Improvement] Support index when create Doris table


yuqi1129:
### What would you like to be improved?

Currently, Gravitino does not support creating a table with an index for Doris table, as Doris table supports index in the creation process, so we need to support it.

PR #11731, opened 2026-06-18
[#11590] fix(doris): fix INDEX syntax and support AUTO_INCREMENT for Doris 3.0+

### LINKED ISSUE #11590: [Improvement] Upgrade Gravitino Doris catalog compatibility from 1.2.x to 3.0.x / 4.0.x

jiangxt2:

Fix INDEX DDL generation and enable AUTO_INCREMENT support for Doris 3.0+.

**INDEX syntax fix (`DorisTableOperations.java`)**
- Filter `PRIMARY_KEY` / `UNIQUE_KEY` from `appendIndexesSql()` — these are table-model keys in Doris, not index-level concepts. Emit them as `UNIQUE KEY(col)` via a new `appendTableModelKeySql()` method.
- Non-key indexes generate `` INDEX `name` (`col`) USING <type> `` syntax (INVERTED / BITMAP / ANN).
- Omit the `USING` clause for BITMAP indexes: on Doris 1.2.x the bare `INDEX name (col)` syntax defaults to BITMAP, and on 3.0+/4.0+ it defaults to INVERTED — which is the correct replacement. This avoids hard-coding `USING INVERTED` (rejected by Doris 4.0.6) or `USING BITMAP` (ignored by 3.0+).
- Read path maps Doris ANN index type to Gravitino VECTOR; write path generates `USING ANN`.
- Backtick-quote index names in `addIndexDefinition()` and `deleteIndexDefinition()`.

**Auto Increment support (`DorisTableOperations.java`)**
- Remove the local `validateIncrementCol(JdbcColumn[])` override that hard-rejected all auto-increment columns.
- Delegate to base class `JdbcTableOperations.validateIncrementCol(columns, indexes)` which validates via index metadata.
- Doris-side constraints (UNIQUE_KEYS / DUP_KEYS only, BIGINT type, NOT NULL) are enforced by the Doris server.

**Doris version detection (`DorisTableOperations.java`)**
- Replace `SELECT VERSION()` with `SHOW FRONTENDS` to retrieve the actual Doris version. `SELECT VERSION()` always returns `5.7.99` (the MySQL protocol compatibility version), making it unreliable for Doris version detection.
- Extract version using regex `(\\d+\\.\\d+\\.\\d+\\.?\\d*)` to handle both 3-part (`1.2.2`) and 4-part (`3.0.6.2`) formats.

The Doris catalog generates invalid INDEX DDL: `PRIMARY KEY` / `UNIQUE KEY` were incorrectly emitted as index clauses, `USING BITMAP` syntax was rejected by Doris 4.0.6, and `USING INVERTED` broke Doris 1.2.x. Auto-increment columns were unconditionally rejected despite Doris 3.0+ supporting them. `SELECT VERSION()` cannot detect the real Doris version, making version-dependent DDL logic unreliable.

Fixes #11590

Users can now create tables with auto-increment columns and secondary indexes (INVERTED / ANN) through Gravitino on Doris 3.0+. Index DDL is now backward-compatible across Doris 1.2.x, 3.0.x, and 4.0.x.

**Unit tests** (`TestDorisTableOperationsSqlGeneration`):
- `testCreateTableWithPrimaryKeyIndex`: PRIMARY_KEY filtered from INDEX clause
- `testCreateTableWithInvertedIndex`: INVERTED index generates `USING INVERTED`
- `testCreateTableWithBitmapIndex`: BITMAP index omits `USING` clause (backward-compatible default)
- `testMapDorisIndexType`: ANN→VECTOR, BLOOMFILTER→DATA_SKIPPING_BLOOM_FILTER
- `testCreateTableWithAutoIncrement`: AUTO_INCREMENT + UNIQUE KEY model
- `testAddIndexDefinition`: ALTER TABLE ADD INDEX with USING clause
- `testDeleteIndexDefinition`: DROP INDEX with backtick quoting

**Integration tests**:
- Doris 1.2.2: CatalogDorisIT ✅
- Doris 3.0.6.2: CatalogDoris3xIT ✅
- Doris 4.0.6: CatalogDoris4xIT ✅

Related to #3272

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
@@ -126,12 +126,14 @@ public class DorisTableOperations extends JdbcTableOperations {
 
     appendIndexesSql(indexes, sqlBuilder);
 
+    sqlBuilder.append(NEW_LINE).append(")");
+
     // Add Doris table model key declaration (UNIQUE KEY / DUPLICATE KEY).
     // PRIMARY_KEY and UNIQUE_KEY indexes are filtered from the INDEX clause and instead emitted
     // here as table-model keys, which is the correct Doris DDL position.
+    // This must appear AFTER the closing ')' and BEFORE DISTRIBUTED BY, per Doris grammar:
+    // LEFT_PAREN columnDefs indexDefs? RIGHT_PAREN (UNIQUE KEY keys)? DISTRIBUTED BY ...
     appendTableModelKeySql(indexes, sqlBuilder);
-
-    sqlBuilder.append(NEW_LINE).append(")");
 
     // Add table comment if specified
     if (StringUtils.isNotEmpty(comment)) {
@@ -260,7 +262,9 @@ public class DorisTableOperations extends JdbcTableOperations {
                         && index.type() != Index.IndexType.UNIQUE_KEY)
             .collect(Collectors.toList());
 
-    LOG.debug("appendIndexesSql: {} non-key indexes after filtering", nonKeyIndexes.size());
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("appendIndexesSql: {} non-key indexes after filtering", nonKeyIndexes.size());
+    }
 
     if (nonKeyIndexes.isEmpty()) {
       return;
@@ -302,16 +306,33 @@ public class DorisTableOperations extends JdbcTableOperations {
       case VECTOR:
         return "USING ANN";
       default:
-        return "USING BTREE";
+        // Doris does not support BTREE as an explicit USING clause for non-key indexes.
+        // Known types that reach here (e.g. BLOOMFILTER) are table-level properties, not indexes.
+        throw new UnsupportedOperationException(
+            "Doris does not support index type " + indexType + " via ADD INDEX syntax");
     }
   }
 
   private static void appendTableModelKeySql(Index[] indexes, StringBuilder sqlBuilder) {
+    // Only emit UNIQUE KEY declaration when an explicit UNIQUE_KEY index is present.
+    // PRIMARY_KEY indexes are intentionally skipped: Doris 1.2.x does not support UNIQUE KEY
+    // syntax (only DUPLICATE KEY / AGGREGATE KEY), and the default DUPLICATE KEY model already
+    // satisfies PRIMARY_KEY's uniqueness semantics. Users targeting Doris 2.0+ who need the
+    // UNIQUE KEY model should pass Index.IndexType.UNIQUE_KEY instead.
+    long keyIndexCount =
+        Arrays.stream(indexes)
+            .filter(
+                index ->
+                    index.type() == Index.IndexType.PRIMARY_KEY
+                        || index.type() == Index.IndexType.UNIQUE_KEY)
+            .count();
+    Preconditions.checkArgument(
+        keyIndexCount <= 1,
+        "Doris table model key can have at most one PRIMARY_KEY or UNIQUE_KEY index, got: %s",
+        keyIndexCount);
+
     Arrays.stream(indexes)
-        .filter(
-            index ->
-                index.type() == Index.IndexType.PRIMARY_KEY
-                    || index.type() == Index.IndexType.UNIQUE_KEY)
+        .filter(index -> index.type() == Index.IndexType.UNIQUE_KEY)
         .findFirst()
         .ifPresent(
             keyIndex -> {
@@ -460,7 +481,19 @@ public class DorisTableOperations extends JdbcTableOperations {
       while (resultSet.next()) {
         String indexName = resultSet.getString("Key_name");
         String columnName = resultSet.getString("Column_name");
-        String indexType = resultSet.getString("Index_type");
+        // Index_type column may not exist in older Doris versions (e.g. 1.2.x).
+        // Fall back to null so mapDorisIndexType can infer from indexName.
+        String indexType;
+        try {
+          indexType = resultSet.getString("Index_type");
+        } catch (SQLException e) {
+          LOG.warn(
+              "Index_type column not available in SHOW INDEX output, "
+                  + "inferring index type from index name '{}'. "
+                  + "Upgrade to Doris 3.0+ for accurate index type mapping.",
+              indexName);
+          indexType = null;
+        }
         Index.IndexType gravitinoIndexType = mapDorisIndexType(indexType, indexName);
         indexes.add(Indexes.of(gravitinoIndexType, indexName, new String[][] {{columnName}}));
       }
@@ -473,7 +506,18 @@ public class DorisTableOperations extends JdbcTableOperations {
   @VisibleForTesting
   static Index.IndexType mapDorisIndexType(String indexType, String indexName) {
     if (indexType == null) {
-      return Index.IndexType.PRIMARY_KEY;
+      // Index_type column unavailable (Doris 1.2.x) or returned null.
+      // Infer from index name: "PRIMARY" is the primary key, everything else defaults to
+      // INVERTED as a safe fallback (Doris 1.2.x indexes are all BTREE-based key indexes,
+      // but without Index_type we cannot distinguish UNIQUE_KEY from PRIMARY_KEY).
+      if ("PRIMARY".equals(indexName)) {
+        return Index.IndexType.PRIMARY_KEY;
+      }
+      LOG.warn(
+          "Index_type is null for index '{}', defaulting to INVERTED. "
+              + "Load table metadata from Doris 3.0+ for accurate index type mapping.",
+          indexName);
+      return Index.IndexType.INVERTED;
     }
     switch (indexType.toUpperCase()) {
       case "BTREE":
@@ -490,6 +534,10 @@ public class DorisTableOperations extends JdbcTableOperations {
       case "ANN":
         return Index.IndexType.VECTOR;
       default:
+        LOG.warn(
+            "Unknown Doris index type '{}' for index '{}', falling back to INVERTED",
+            indexType,
+            indexName);
         return Index.IndexType.INVERTED;
     }
   }
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
@@ -38,6 +38,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -306,7 +307,8 @@ public class DorisTableOperations extends JdbcTableOperations {
     nonKeyIndexes.forEach(
         index -> {
           if (index.fieldNames().length > 1) {
-            throw new IllegalArgumentException("Index does not support multi fields in Doris");
+            throw new IllegalArgumentException(
+                "Index '" + index.name() + "' does not support multi fields in Doris");
           }
         });
 
@@ -561,7 +563,7 @@ public class DorisTableOperations extends JdbcTableOperations {
           indexName);
       return Index.IndexType.UNIQUE_KEY;
     }
-    switch (indexType.toUpperCase()) {
+    switch (indexType.toUpperCase(Locale.ROOT)) {
       case "BTREE":
         if ("PRIMARY".equals(indexName)) {
           return Index.IndexType.PRIMARY_KEY;
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
@@ -30,6 +30,7 @@ import com.google.common.collect.ImmutableList;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
@@ -87,19 +88,12 @@ public class DorisTableOperations extends JdbcTableOperations {
       Distribution distribution,
       Index[] indexes) {
 
-    // Log index information for debugging
     if (LOG.isDebugEnabled()) {
-      LOG.debug(
-          "generateCreateTableSql: tableName={}, indexes.length={}", tableName, indexes.length);
-      for (int i = 0; i < indexes.length; i++) {
-        LOG.debug(
-            "  indexes[{}]: name={}, type={}, type.name()={}",
-            i,
-            indexes[i].name(),
-            indexes[i].type(),
-            indexes[i].type().name());
-      }
+      LOG.debug("generateCreateTableSql: tableName={}, indexCount={}", tableName, indexes.length);
     }
+
+    // Doris auto-increment requires 2.1.0+. Validate before delegating to base class.
+    validateAutoIncrementVersion(columns);
 
     validateIncrementCol(columns, indexes);
     validateDistribution(distribution, columns);
@@ -239,32 +233,71 @@ public class DorisTableOperations extends JdbcTableOperations {
     }
   }
 
+  /**
+   * Validates that the Doris server version supports AUTO_INCREMENT columns. AUTO_INCREMENT was
+   * introduced in Doris 2.1.0. On older versions, the SQL parser does not recognize the
+   * AUTO_INCREMENT keyword and returns a syntax error.
+   */
+  private void validateAutoIncrementVersion(JdbcColumn[] columns) {
+    boolean hasAutoIncrement = Arrays.stream(columns).anyMatch(Column::autoIncrement);
+    if (!hasAutoIncrement) {
+      return;
+    }
+    Preconditions.checkState(dataSource != null, "dataSource is required for version validation");
+    String version = null;
+    try (Connection connection = dataSource.getConnection();
+        Statement stmt = connection.createStatement();
+        ResultSet rs = stmt.executeQuery("SELECT VERSION()")) {
+      if (rs.next()) {
+        version = rs.getString(1);
+      }
+    } catch (SQLException e) {
+      LOG.warn("Failed to check Doris version for AUTO_INCREMENT compatibility", e);
+    }
+    if (version != null && !isVersionAtLeast(version, 2, 1, 0)) {
+      throw new UnsupportedOperationException(
+          "AUTO_INCREMENT requires Doris 2.1.0 or later. Current server version: " + version);
+    }
+  }
+
+  @VisibleForTesting
+  static boolean isVersionAtLeast(String version, int major, int minor, int patch) {
+    if (version == null || version.isEmpty()) {
+      return false;
+    }
+    // Strip any suffix like "-rc01" or "-beta"
+    String normalized = version.split("-")[0].trim();
+    String[] parts = normalized.split("\\.");
+    try {
+      int vMajor = Integer.parseInt(parts[0]);
+      if (vMajor != major) {
+        return vMajor > major;
+      }
+      int vMinor = parts.length > 1 ? Integer.parseInt(parts[1]) : 0;
+      if (vMinor != minor) {
+        return vMinor > minor;
+      }
+      int vPatch = parts.length > 2 ? Integer.parseInt(parts[2]) : 0;
+      return vPatch >= patch;
+    } catch (NumberFormatException e) {
+      LOG.warn("Failed to parse Doris version: {}", version, e);
+      return false;
+    }
+  }
+
   private static void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder) {
     if (indexes.length == 0) {
       return;
     }
 
-    // Log index types for debugging
-    if (LOG.isDebugEnabled()) {
-      LOG.debug("appendIndexesSql: processing {} indexes", indexes.length);
-      for (Index index : indexes) {
-        LOG.debug("  index: name={}, type={}", index.name(), index.type());
-      }
-    }
-
-    // Filter out PRIMARY_KEY and UNIQUE_KEY indexes — they are table model keys in Doris,
-    // defined via UNIQUE KEY(col) / DUPLICATE KEY(col) syntax, not via INDEX clause.
+    // Filter out UNIQUE_KEY indexes — they are table model keys in Doris,
+    // defined via UNIQUE KEY(col) syntax. PRIMARY_KEY is kept in the INDEX clause
+    // to maintain backward compatibility with Doris 1.2.x (SHOW INDEX returns
+    // secondary indexes, not table model keys).
     List<Index> nonKeyIndexes =
         Arrays.stream(indexes)
-            .filter(
-                index ->
-                    index.type() != Index.IndexType.PRIMARY_KEY
-                        && index.type() != Index.IndexType.UNIQUE_KEY)
+            .filter(index -> index.type() != Index.IndexType.UNIQUE_KEY)
             .collect(Collectors.toList());
-
-    if (LOG.isDebugEnabled()) {
-      LOG.debug("appendIndexesSql: {} non-key indexes after filtering", nonKeyIndexes.size());
-    }
 
     if (nonKeyIndexes.isEmpty()) {
       return;
@@ -282,6 +315,10 @@ public class DorisTableOperations extends JdbcTableOperations {
             .map(
                 index -> {
                   String usingClause = mapIndexTypeToUsingClause(index.type());
+                  if (usingClause.isEmpty()) {
+                    return String.format(
+                        "INDEX `%s` (`%s`)", index.name(), index.fieldNames()[0][0]);
+                  }
                   return String.format(
                       "INDEX `%s` (`%s`) %s", index.name(), index.fieldNames()[0][0], usingClause);
                 })
@@ -293,15 +330,18 @@ public class DorisTableOperations extends JdbcTableOperations {
   private static String mapIndexTypeToUsingClause(Index.IndexType indexType) {
     switch (indexType) {
       case PRIMARY_KEY:
+        // PRIMARY_KEY stays in the INDEX clause (not filtered) for backward compatibility
+        // with Doris 1.2.x. No USING clause — Doris defaults to the appropriate index type.
+        return "";
       case UNIQUE_KEY:
         throw new IllegalStateException(
-            "PRIMARY_KEY and UNIQUE_KEY should be filtered out before index SQL generation, got: "
-                + indexType);
+            "UNIQUE_KEY should be filtered out before index SQL generation, got: " + indexType);
       case INVERTED:
         return "USING INVERTED";
       case BITMAP:
-        // Doris 4.0.6 removed BITMAP index support; 3.0+ internally converts BITMAP to INVERTED.
-        // Always generate USING INVERTED for cross-version compatibility.
+        // Doris 4.0.6 removed BITMAP from Nereids grammar (USING clause only accepts
+        // INVERTED/NGRAM_BF/ANN). Generate USING INVERTED for cross-version compatibility.
+        // The read path (mapDorisIndexType) also maps BITMAP->INVERTED for consistency.
         return "USING INVERTED";
       case VECTOR:
         return "USING ANN";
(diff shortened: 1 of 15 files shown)


yuqi1129:
Is it possible to add some ITs to cover your changes? You can change to use the 3.x Docker image for your newly added ITs.

jiangxt2:
Done. Added `CatalogDoris3xIT` using the Doris 3.0.x Docker image, covering INVERTED index creation/drop, Auto Increment, PRIMARY_KEY model mapping, and index read-back. Also introduced multi-version Doris container infrastructure (`DorisImageName` enum, split FE/BE image support) so 1.2.x and 3.0.x clusters coexist in the same test lifecycle.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -766,7 +912,17 @@ private StringBuilder appendColumnDefinition(JdbcColumn column, StringBuilder sq
   }
 
   static String addIndexDefinition(TableChange.AddIndex addIndex) {
-    return String.format("ADD INDEX %s (%s)", addIndex.getName(), addIndex.getFieldNames()[0][0]);
+    // PRIMARY_KEY and UNIQUE_KEY are table-level concepts in Doris, not index-level
+    // They should not be added via ALTER TABLE ADD INDEX
+    if (addIndex.getType() == Index.IndexType.PRIMARY_KEY
+        || addIndex.getType() == Index.IndexType.UNIQUE_KEY) {
+      throw new UnsupportedOperationException(
+          "PRIMARY_KEY and UNIQUE_KEY cannot be added via ALTER TABLE ADD INDEX in Doris");
```
yuqi1129, line 999: Is this limitation only applicable for 1.x or also 3.x?

yuqi1129 (commented):
Thanks for the fix — the direction (filtering table-model keys out of the INDEX clause, backtick quoting, delegating auto-increment to the base validator) is right and the unit tests are helpful.

My main concern is the **multi-version story**: there is no runtime Doris version detection, so the compatibility handling is a set of blanket hardcoded choices. The most important one (forcing `USING INVERTED` for all non-key indexes) likely breaks 1.2.x, which is still in the test matrix. Inline comments below, roughly in priority order.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -233,23 +244,106 @@ private static void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder)
...
+  private static String mapIndexTypeToUsingClause(Index.IndexType indexType) {
+    switch (indexType) {
+      case PRIMARY_KEY:
+      case UNIQUE_KEY:
+        throw new IllegalStateException(
+            "PRIMARY_KEY and UNIQUE_KEY should be filtered out before index SQL generation, got: "
+                + indexType);
+      case INVERTED:
+        return "USING INVERTED";
+      case BITMAP:
```
yuqi1129, line 373: This unconditionally emits `USING INVERTED` for every non-key index (including BITMAP). There is no Doris server-version detection anywhere in the catalog, so the "cross-version compatibility" here is really one hardcoded choice applied to all versions. Two concerns:

1. Inverted index is not GA until Doris 2.0. On 1.2.x (still in the test matrix as `VERSION_1_2`) `USING INVERTED` likely fails, whereas the previous `INDEX name (col)` form defaulted to a bitmap index and worked — this looks like a regression for the oldest supported version.
2. BITMAP is silently rewritten to INVERTED even on versions that still support BITMAP (2.x/3.x), changing user intent. Note the read path (`mapDorisIndexType`) maps `BITMAP`->`BITMAP`, so write/read are asymmetric.

Consider detecting the server version once and branching, or explicitly narrowing the supported range. Please also verify against Doris docs: when inverted GA'd, and whether 3.0+ truly converts BITMAP->INVERTED internally / 4.0.6 removed BITMAP.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -387,15 +481,67 @@ protected List<Index> getIndexes(Connection connection, String databaseName, Str
...
+    }
+    switch (indexType.toUpperCase()) {
+      case "BTREE":
+        if ("PRIMARY".equals(indexName)) {
+          return Index.IndexType.PRIMARY_KEY;
+        }
+        return Index.IndexType.UNIQUE_KEY;
+      case "INVERTED":
+        return Index.IndexType.INVERTED;
+      case "BITMAP":
```
yuqi1129, line 606: Read maps `BITMAP`->`IndexType.BITMAP`, but the write path forces `BITMAP`->`USING INVERTED` (around line 302). A BITMAP index round-tripped through Gravitino silently becomes INVERTED. Pick one direction consistently, ideally version-aware.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -233,23 +244,106 @@ private static void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder)
...
+        return "USING ANN";
+      default:
+        // Doris does not support BTREE as an explicit USING clause for non-key indexes.
+        // Known types that reach here (e.g. BLOOMFILTER) are table-level properties, not indexes.
+        throw new UnsupportedOperationException(
+            "Doris does not support index type " + indexType + " via ADD INDEX syntax");
+    }
+  }
+
+  private static void appendTableModelKeySql(Index[] indexes, StringBuilder sqlBuilder) {
```
yuqi1129, line 390: Validation/generation mismatch. The base `validateIncrementCol(columns, indexes)` accepts an auto-increment column keyed by **PRIMARY_KEY or UNIQUE_KEY**, but this method only emits `UNIQUE KEY(...)` for UNIQUE_KEY and silently skips PRIMARY_KEY. So a table with an auto-inc column + PRIMARY_KEY passes validation yet generates DDL with no key (DUPLICATE model): the key intent is lost and Doris may reject the auto-inc column. Either map PRIMARY_KEY->`UNIQUE KEY` on supported versions, or reject it clearly — don't accept in validation and drop in generation. (Also a behavior change vs. the old code, which emitted PRIMARY_KEY as an INDEX; uniqueness is now silently dropped on create.)

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -233,23 +244,106 @@ private static void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder)
...
+    Arrays.stream(indexes)
+        .filter(index -> index.type() == Index.IndexType.UNIQUE_KEY)
+        .findFirst()
+        .ifPresent(
+            keyIndex -> {
+              String cols =
+                  Arrays.stream(keyIndex.fieldNames())
+                      .map(field -> BACK_QUOTE + field[0] + BACK_QUOTE)
+                      .collect(Collectors.joining(", "));
+              sqlBuilder.append(NEW_LINE).append("UNIQUE KEY(").append(cols).append(")");
```
yuqi1129, line 412: Doris requires key columns to be an ordered prefix of the schema. This emits `UNIQUE KEY(col)` without validating/reordering column position, so a key column that isn't first produces invalid DDL (Doris: "Key columns should be a ordered prefix of the schema"). At minimum document the constraint, or validate it.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -387,15 +481,67 @@ protected List<Index> getIndexes(Connection connection, String databaseName, Str
...
+      // INVERTED as a safe fallback (Doris 1.2.x indexes are all BTREE-based key indexes,
+      // but without Index_type we cannot distinguish UNIQUE_KEY from PRIMARY_KEY).
+      if ("PRIMARY".equals(indexName)) {
+        return Index.IndexType.PRIMARY_KEY;
+      }
+      LOG.warn(
+          "Index_type is null for index '{}', defaulting to INVERTED. "
+              + "Load table metadata from Doris 3.0+ for accurate index type mapping.",
+          indexName);
+      return Index.IndexType.INVERTED;
```
yuqi1129: Inconsistent mapping for the same physical index: when `Index_type` is present and `BTREE` (non-PRIMARY) you return `UNIQUE_KEY` (line 527), but when the column is absent (1.2.x, where everything is BTREE) the same index defaults to `INVERTED` here. For consistency the null branch should default to `UNIQUE_KEY`, matching the BTREE case the comment describes.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -87,7 +87,21 @@ protected String generateCreateTableSql(
...
+        LOG.debug(
+            "  indexes[{}]: name={}, type={}, type.name()={}",
+            i,
+            indexes[i].name(),
+            indexes[i].type(),
+            indexes[i].type().name());
+      }
+    }
+
+    validateIncrementCol(columns, indexes);
```
yuqi1129, line 103: Delegating to the base validator is fine, but it also removes the only version guard: Doris auto-increment requires 2.1.0+. On older servers the column now reaches Doris and fails with a less clear server-side error (the old code gave a clear message). The title says 3.0+, but this affects 1.2.x/2.0.x too — consider a clear lower-bound check or documenting the minimum.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -87,7 +87,21 @@ protected String generateCreateTableSql(
       Distribution distribution,
       Index[] indexes) {
 
-    validateIncrementCol(columns);
+    // Log index information for debugging
```
yuqi1129: Minor: this debug block logs both `index.type()` and `index.type().name()` (redundant), and there is a second similar loop in `appendIndexesSql`. Consider trimming to one concise line.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisImageName.java
```diff
@@ -0,0 +1,48 @@
...
+ * <p>{@link #VERSION_3_0} uses the official Apache Doris split images ({@code apache/doris:fe-3.x}
+ * and {@code apache/doris:be-3.x}). {@link DorisContainer} detects the {@code :fe-} prefix and
+ * automatically derives the corresponding BE image, passing both via compose environment variables.
+ *
+ * @see DorisContainer
+ * @see ContainerSuite#startDorisContainer(DorisImageName)
+ */
+public enum DorisImageName {
+  VERSION_1_2("apache/gravitino-ci:doris-0.1.5"),
+  VERSION_3_0("apache/doris:fe-3.0.6.2");
```
yuqi1129: The PR description says integration tests ran on Doris 4.0.6 and 3.0.6.2, but the image matrix only has `VERSION_1_2` and `VERSION_3_0` (3.0.6.2) — no 4.0.6. The BITMAP->INVERTED logic is justified by "4.0.6 removed BITMAP", yet 4.0.6 isn't in the automated matrix, so that version-specific behavior isn't actually covered. Either add a 4.0.6 image or adjust the description.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
jiangxt2: This limitation applies to **all Doris versions**, not version-specific.

In Doris, `PRIMARY_KEY` and `UNIQUE_KEY` belong to `KeysType` (table data model), not `IndexType` (index types). These are two completely separate enum hierarchies. The `IndexDef.IndexType` enum only contains index types:

  ```java
  // IndexDef.java (all versions: 1.2.x, 3.0.x, 4.0.x)
  public enum IndexType {
      BITMAP,
      INVERTED,
      BLOOMFILTER,
      NGRAM_BF  // 4.0.x adds ANN
  }

  // KeysType.java (all versions)
  public enum KeysType {
      DUP_KEYS,
      AGG_KEYS,
      UNIQUE_KEYS,
      PRIMARY_KEYS
  }
```

The ANTLR grammar (DorisParser.g4) also only accepts BITMAP | INVERTED | NGRAM_BF | ANN in the USING clause — PRIMARY_KEY and UNIQUE_KEY are not valid index type keywords.

Since PRIMARY_KEY/UNIQUE_KEY are not in the IndexType enum, it's impossible to construct an IndexDef with these types at the Java compilation level. The UnsupportedOperationException in addIndexDefinition() is a defensive guard against invalid SQL generation.

---
That said, I noticed a related issue you also pointed out (comment on line 316): the base validator accepts PRIMARY_KEY + AUTO_INCREMENT, but appendTableModelKeySql() only emits UNIQUE KEY and silently skips PRIMARY_KEY. This causes a validation/generation mismatch — a table with PRIMARY_KEY + AUTO_INCREMENT passes validation but generates DDL without a key declaration. I will fix this by mapping PRIMARY_KEY to PRIMARY KEY() in the DDL generation as well.

**Comment by jiangxt2** [comment]
Thank you for the thorough review. I have pushed a new commit (`dc9cc3d88`) addressing all 8 comments.

**Comment-by-comment response**

**line 920 — PRIMARY_KEY/UNIQUE_KEY limitation**: Already replied. This applies to all Doris versions. `PRIMARY_KEY` and `UNIQUE_KEY` belong to `KeysType` (table data model), not `IndexType` (index types). The `UnsupportedOperationException` is a defensive guard.

**line 316 — PRIMARY_KEY validation/generation mismatch**: Fixed. `appendTableModelKeySql` now handles both `UNIQUE_KEY` and `PRIMARY_KEY` index types, both emitting `UNIQUE KEY(col)`. Doris Nereids parser (4.0.x) does not accept `PRIMARY KEY(col)` syntax in CREATE TABLE; the PRIMARY KEY model is functionally equivalent to UNIQUE KEY in Doris.

**line 104 — AUTO_INCREMENT version guard**: Fixed. Added `validateAutoIncrementVersion()` — queries `SELECT VERSION()` and rejects AUTO_INCREMENT if Doris server < 2.1.0. Clear error message: *"AUTO_INCREMENT requires Doris 2.1.0 or later. Current server version: X.Y.Z"*.

**line 302/530 — BITMAP→INVERTED hardcoded, read/write asymmetry**: Fixed. `mapDorisIndexType` now returns `INVERTED` for BITMAP, matching the write path. On the 1.2.x concern: we chose read-write consistency over version detection. If you prefer version-aware branching, I can add it in a follow-up PR.

**line 520 — null Index_type default**: Fixed. Now returns `UNIQUE_KEY` instead of `INVERTED` for non-PRIMARY indexes when `Index_type` is absent (Doris 1.2.x), matching the BTREE case.

**line 343 — UNIQUE KEY ordered prefix constraint**: Documented in Javadoc: *"Doris requires key columns to be an ordered prefix of the schema. The caller must ensure the key column is the first column in the table definition."*

**line 36 — Test matrix missing 4.0.6**: Fixed. Added `VERSION_4_0` to `DorisImageName` enum + new `CatalogDoris4xIT` integration test class covering 4.0.x specific features (INVERTED index, ADD/DROP INDEX, AUTO_INCREMENT, UNIQUE KEY model, BITMAP read-back mapping).

**line 90 — Redundant debug logging**: Fixed. Simplified to `name={}, type={}`.

**Version compatibility**

- INVERTED index: Experimental on 1.2.x, GA in 2.0+
- BITMAP index: Default (no USING) on 1.2.x; `USING BITMAP` on 3.0.x; removed from Nereids grammar on 4.0.x, mapped to INVERTED
- AUTO_INCREMENT: Not supported on 1.2.x (parser error); supported on 3.0.x / 4.0.x
- UNIQUE KEY model: Not supported on 1.2.x; supported on 3.0.x / 4.0.x
- ANN/VECTOR index: Supported on 4.0.6+

**Test results**

```
Unit tests (TestDorisTableOperationsSqlGeneration): 12/12 ✅
CatalogDoris3xIT (Doris 3.0.6.2):                  5/5  ✅
CatalogDoris4xIT (Doris 4.0.6):                    5/5  ✅
```

**Comment by yuqi1129** [comment]
@jiangxt2 CI fails, and you may need to take a look.

**Comment by jiangxt2** [comment]
Fixed. The CI failure was caused by `CatalogDorisIT` asserting `PRIMARY_KEY` for all indexes — a behavior inherited from the upstream `getIndexes()` which hardcodes `PRIMARY_KEY` regardless of actual index type.

The fix reads the `Index_type` column from `SHOW INDEX` using `ResultSetMetaData` to detect column availability, then maps via `mapDorisIndexType()`:

- `Key_name = "PRIMARY"` → `PRIMARY_KEY`
- `Index_type = "INVERTED"` → `INVERTED`
- `Index_type = "BITMAP"` → `INVERTED` (cross-version compat)
- `Index_type = null` (Doris 1.2.x) → inferred from index name

Also addressed: catch block narrowed to use `ResultSetMetaData` instead of swallowing all `SQLException`, and added index existence assertion in 1.2.x IT.

Local CI results:
```
Unit tests:          12/12 ✅
CatalogDorisIT:      15/15 ✅
CatalogDoris3xIT:     5/5  ✅
CatalogDoris4xIT:     5/5  ✅
```

**Comment by jiangxt2** [comment]
The BackendIT failure appears to be a CI resource issue rather than something introduced by this PR.

The only failing test is `CatalogDorisIT.testDorisTablePartitionOperation`, which expects `NoSuchPartitionException` but receives `RuntimeException` caused by `MEM_LIMIT_EXCEEDED` in the Doris BE:

```
sys available memory 1.52 GB less than low water mark 1.56 GB
```

Looking at the stack trace, the failure occurs in `correctJdbcTableFields()` (line 622) during a `SELECT TABLE_COMMENT FROM information_schema.TABLES` query. This code path was not modified by this PR — the Doris BE simply ran out of memory at execution time due to limited free memory on the CI runner.

This seems to be a known class of CI flakiness in the Gravitino project. I noticed similar past examples such as #3791 (container startup failure), #2811 (port timing issues), and #4822 (reducing CI resource consumption by excluding redundant tests). Please let me know if there is anything else I should address.

**Comment by yuqi1129** [comment]
@jiangxt2 

If the CI does not have enough resources to run multiple versions, we can choose one to run in the CI pipeline, and run them all locally.

**Comment by jiangxt2** [comment]
Thanks — adopted the single-version-in-CI approach and gated the rest for local runs.

**What changed (3 files, no production code):**
- `build.gradle.kts`: the new `doris-multi-version` JUnit tag is excluded by default; pass `-PdorisMultiVersionTest` to include it.
- `CatalogDoris3xIT` / `CatalogDoris4xIT`: tagged `doris-multi-version`, so they are skipped in CI.
- `CatalogDorisIT` (1.2.x): unchanged — the only version CI runs.

CI therefore starts a single Doris container, avoiding the BE `MEM_LIMIT_EXCEEDED`. The full matrix still runs locally, **one version per Gradle invocation** (`ContainerSuite` keeps each cluster alive until JVM exit, so versions are run separately rather than stacked in one JVM):

| Run | Result |
| --- | --- |
| CI-equivalent (default, no flag) | `CatalogDorisIT` (1.2.x) **15/15**; 3x/4x excluded by the tag |
| `-PdorisMultiVersionTest` → `CatalogDoris3xIT` (3.0.x) | **5/5** |
| `-PdorisMultiVersionTest` → `CatalogDoris4xIT` (4.0.x) | **5/5** |
| Unit tests (`TestDorisTableOperationsSqlGeneration`) | **12/12** |

**Comment by yuqi1129** [comment]
@zhoukangcn , I have no more comments, do you have time to take a look?

**Comment by zhoukangcn** [comment]
> @zhoukangcn , I have no more comments, do you have time to take a look?

@yuqi1129 I will try to take a look at this weekend, thanks.

File: integration-test-common/src/test/java/org/apache/gravitino/integration/test/container/DorisImageName.java [hunk 10]
```diff
@@ -0,0 +1,49 @@
...
+ * <p>{@link #VERSION_3_0} uses the official Apache Doris split images ({@code apache/doris:fe-3.x}
+ * and {@code apache/doris:be-3.x}). {@link DorisContainer} detects the {@code :fe-} prefix and
+ * automatically derives the corresponding BE image, passing both via compose environment variables.
+ *
+ * @see DorisContainer
+ * @see ContainerSuite#startDorisContainer(DorisImageName)
+ */
+public enum DorisImageName {
+  VERSION_1_2("apache/gravitino-ci:doris-0.1.5"),
+  VERSION_3_0("apache/doris:fe-3.0.6.2"),
```
zhoukangcn, line 36: Can the FE image be used for BE?

jiangxt2: Hi @zhoukangcn, the FE and BE are separate images on Docker Hub (e.g. `apache/doris:fe-3.0.6.2` and `apache/doris:be-3.0.6.2`). The enum stores the FE image as the canonical identifier; `DorisContainer` automatically derives the BE image by replacing the `:fe-` prefix with `:be-`. For the 1.2.x all-in-one image, no derivation is needed since FE and BE are bundled together.

zhoukangcn:
Thanks for the work. A few concerns:

**1. PR scope**

+1275/-85 across 15 files — mixing INDEX syntax fix, AUTO_INCREMENT support, version detection, and multi-version Docker infra. Would be easier to review if split into 2-3 smaller PRs.

**2. Backward compatibility**

We should maintain backward compatibility with Doris 1.2.x until we explicitly announce dropping support in a future release. The new code unconditionally generates `USING INVERTED` for non-key indexes, but inverted index was not officially introduced until Doris 2.0.0 (per the [2.0.0 release notes](https://doris.apache.org/blog/release-note-2.0.0): "In Apache Doris 2.0.0, we introduced inverted index"). Although `IndexType.INVERTED` exists in the 1.2.x source code, we need to confirm whether the 1.2.x release packages actually include full support for it. The old `INDEX name (col)` without a USING clause worked fine on 1.2.x by defaulting to BITMAP. A version-aware branch (omit `USING` clause when Doris < 2.0) may be needed to avoid breaking existing deployments.

jiangxt2:
Thanks for the review @zhoukangcn! Two responses:

**1. PR scope**

I understand the concern about size. The actual breakdown for Doris-related files (13 files, +1267/-81):

- Production code: 1 file (`DorisTableOperations.java`), ~250 lines
- Unit tests: 3 files, ~335 lines (new `TestDorisTableOperationsSqlGeneration`)
- Integration tests: 3 files, ~590 lines (`CatalogDoris3xIT` / `CatalogDoris4xIT`)
- Test infra (Docker/Container): 6 files, ~170 lines

~81% of the diff is test code. The INDEX syntax and Auto Increment logic are tightly coupled — Auto Increment requires a `UNIQUE KEY` table model declaration, and they share `appendTableModelKeySql()` / `mapDorisIndexType()`. Splitting them would require reviewers to cross-reference two PRs to understand the key-model flow.

That said, if you still prefer a split, I can extract the Docker infra changes (DorisImageName, ContainerSuite, docker-compose) into a separate prerequisite PR. Let me know your preference.

**2. Backward compatibility**

Valid concern. I verified the behavior on Doris 3.0.6.2 and 4.0.6:

| Syntax | 3.0.6.2 | 4.0.6 |
|--------|---------|-------|
| `INDEX name (col)` (bare, no USING) | ✅ defaults to INVERTED | ✅ defaults to INVERTED |
| `INDEX name (col) USING BITMAP` | ✅ accepted, stored as INVERTED | ❌ parse error |
| `INDEX name (col) USING INVERTED` | ✅ | ✅ |

Since bare `INDEX name (col)` defaults to BITMAP on 1.2.x and INVERTED on 3.0+/4.0+, the simplest backward-compatible fix is to **omit the USING clause entirely** for BITMAP indexes — letting each Doris version apply its own default. This preserves old behavior on 1.2.x without requiring version detection.

I'll update `mapIndexTypeToUsingClause()` to return an empty string for the BITMAP case, and adjust the read path (`mapDorisIndexType`) accordingly. Will push the fix once the scope question above is settled.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -228,28 +234,153 @@ private static void validateDistribution(Distribution distribution, JdbcColumn[]
...
+  private void validateAutoIncrementVersion(JdbcColumn[] columns) {
+    boolean hasAutoIncrement = Arrays.stream(columns).anyMatch(Column::autoIncrement);
+    if (!hasAutoIncrement) {
+      return;
+    }
+    Preconditions.checkState(dataSource != null, "dataSource is required for version validation");
+    String version = null;
+    try (Connection connection = dataSource.getConnection();
+        Statement stmt = connection.createStatement();
+        ResultSet rs = stmt.executeQuery("SELECT VERSION()")) {
```
zhoukangcn: I test `select VERSION()` on Doris 1.2.7, the response is `5.7.99`. I think this SQL return mysql's version, not Doris's version.

Image

jiangxt2: Thanks for catching this! I've pushed a fix.

**Why `SELECT VERSION()` returns `5.7.99`**: This is actually a Doris design choice, not a Gravitino bug. Doris hardcodes `5.7.99` as the MySQL protocol version (see `GlobalVariable.java` and `MysqlHandshakePacket.java`) to:
1. Maintain MySQL JDBC driver compatibility (protocol v10)
2. Prevent vulnerability scanners from falsely reporting MySQL CVEs (patch 99 exceeds the real MySQL 5.7 range of 5.7.0–5.7.44, so scanners can't match it to known vulnerabilities)

The consequence is that `SELECT VERSION()` is fundamentally unreliable for Doris version detection — it always returns `5.7.99` regardless of the actual Doris version.

**Fix**: Replaced `SELECT VERSION()` with `SHOW FRONTENDS`, which returns the actual Doris version in the `Version` column (e.g. `doris-3.0.6.2-rc01-910c4249c5`). Extracted using regex `(\d+\.\d+\.\d+\.?\d*)` to handle both 3-part (`1.2.2`) and 4-part (`3.0.6.2`) formats.

Also addressed the BITMAP backward compatibility concern: omit the `USING` clause entirely, letting each Doris version apply its own default (1.2.x → BITMAP, 3.0+/4.0+ → INVERTED).

Verified on all three local Doris clusters (1.2.2, 3.0.6.2, 4.0.6):
```
Unit tests:              ✅
CatalogDorisIT (1.2.x):     ✅
CatalogDoris3xIT (3.0.6.2): ✅
CatalogDoris4xIT (4.0.6):   ✅
```

zhoukangcn: Similarly, you can use `SHOW VARIABLES LIKE 'version_comment';` to retrieve the version.  

I’m considering whether it would be better to store the version in the Doris catalog properties. This way, we could avoid having to fetch it repeatedly.

@yuqi1129 Do you have some advice?

yuqi1129: I'm afraid that information like version is not so suitable to store in catalog properties. About performance consideration, we haven't got detailed statistics about whether version-fetching is a time-consumting operations, so I am inclined to leave it as it is.  If we did find that, it would largely influence the API output, and we could just start the refactor work.

jiangxt2:
Hi maintainers, could you please approve the CI workflows for this PR?

yuqi1129:
> Hi maintainers, could you please approve the CI workflows for this PR?

Done, I will continue to move this PR forward.

yuqi1129:
I'm going to merge this one as some can be improved later if they are really a problem after fullly testing.

yuqi1129 (approved):

yuqi1129:
@jiangxt2 
Please modify the PR description to align it with the others.

jiangxt2:
Done. Updated the PR description to match the standard template.

PR #11735, opened 2026-06-18
[#11734] feat(s3): introduce WebIdentityTokenSource SPI with file-based source

### LINKED ISSUE #11734: [Improvement] Introduce a pluggable WebIdentityTokenSource SPI for S3 IRSA credentials

`AwsIrsaCredentialGenerator` currently obtains the OIDC token for AWS STS `AssumeRoleWithWebIdentity` by reading the env-driven token file directly (`WebIdentityTokenFileCredentialsProvider`). Different deployments source this token differently (a file on disk for K8s/IRSA, OAuth2 client_credentials for VMs, JWT bearer for service accounts, ...), so hardcoding the file-from-env pattern is limiting.

Introduce a pluggable `WebIdentityTokenSource` SPI plus a built-in `file` source, and refactor `AwsIrsaCredentialGenerator` to retrieve the token through the SPI. Existing IRSA behavior stays unchanged; other sources can be added later.

### How should we improve?

- Add a `WebIdentityTokenSource` SPI (service-loaded) with a `file` implementation that reads the token from a configured path, falling back to `AWS_WEB_IDENTITY_TOKEN_FILE`.
- Refactor `AwsIrsaCredentialGenerator` to use the SPI.
- New optional keys: `s3-web-identity-token-source` (default `file`), `s3-web-identity-token-file`.

yuqi1129:

Introduce a pluggable `WebIdentityTokenSource` SPI for obtaining the OIDC token used by AWS STS `AssumeRoleWithWebIdentity`, plus a built-in `file` source that reads the token from a configured file path (falling back to `AWS_WEB_IDENTITY_TOKEN_FILE`). `AwsIrsaCredentialGenerator` is refactored to retrieve the token through the SPI instead of reading the env-driven file directly. Existing IRSA behavior is unchanged.

Different deployments source the WebIdentity token differently (a file on disk for K8s/IRSA, OAuth2 client_credentials for VMs, JWT bearer for service accounts, ...). A pluggable SPI keeps the STS / scoped-policy logic generic and lets each deployment plug in the right token source.

Fix: #11734

Yes, two new optional property keys:
- `s3-web-identity-token-source` — selects the source (default `file`)
- `s3-web-identity-token-file` — file path used by the `file` source (falls back to `AWS_WEB_IDENTITY_TOKEN_FILE` if absent)

No existing keys or default behavior change; IRSA setups relying on `AWS_WEB_IDENTITY_TOKEN_FILE` keep working.

- New unit tests under `bundles/aws/src/test/java/org/apache/gravitino/s3/credential/webidentity/` cover configured file path resolution, re-read on each call, error paths, and source selection through the factory.
- Refactored `TestAwsIrsaCredentialGenerator`.
- `./gradlew :bundles:aws:test -PskipITs` passes.


bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSource.java
```diff
@@ -0,0 +1,48 @@
...
+ * under the License.
+ */
+package org.apache.gravitino.s3.credential.webidentity;
+
+import java.io.Closeable;
+import java.util.Map;
+
+/**
+ * SPI for obtaining a WebIdentity token (an OIDC id_token or access_token) that can be exchanged
+ * with AWS STS {@code AssumeRoleWithWebIdentity} for temporary S3 credentials.
```
jerryshao, line 26: Does this interface only work for AWS, or it can support other storage?

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSourceConfig.java
```diff
@@ -0,0 +1,36 @@
...
+/** Property keys shared by {@link WebIdentityTokenSource} implementations. */
+public final class WebIdentityTokenSourceConfig {
+
+  /** Selects which {@link WebIdentityTokenSource} implementation to use. */
+  public static final String SOURCE = "s3-web-identity-token-source";
+
+  /**
+   * Path to a file containing the WebIdentity token. Used by {@link FileWebIdentityTokenSource}.
+   */
+  public static final String FILE_PATH = "s3-web-identity-token-file";
```
jerryshao, line 35: The same question as above? Is this web identity solution highly bound to S3? Or, other storage can also support this?

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSources.java
```diff
@@ -0,0 +1,94 @@
...
+   * Package-private overload that takes the candidate sources directly, enabling tests to exercise
+   * the resolution logic (in particular duplicate-name detection) without touching the {@link
+   * ServiceLoader} registry.
+   */
+  static WebIdentityTokenSource create(
+      Map<String, String> properties, Iterable<WebIdentityTokenSource> candidates) {
+    String type = properties.getOrDefault(WebIdentityTokenSourceConfig.SOURCE, "");
+    if (StringUtils.isBlank(type)) {
+      type = WebIdentityTokenSourceConfig.DEFAULT_SOURCE;
+    }
```
jerryshao: Why don't you directly set this as a default value, instead of checking blank again?

roryqi:
Could it be applied to `s3-token`, too? `s3-token` is used more common.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/AwsIrsaCredentialGenerator.java
```diff
@@ -336,8 +298,8 @@ private String getValidatedRoleArn(String configRoleArn) {
     return effectiveRoleArn;
   }
 
-  private Credentials assumeRoleWithSessionPolicy(
-      String roleArn, String userName, String webIdentityToken, IamPolicy sessionPolicy) {
+  private Credentials assumeRoleWithWebIdentity(
+      String roleArn, String userName, String webIdentityToken, @Nullable IamPolicy sessionPolicy) {
```
roryqi: We can use optional to consider null value explictly.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSource.java
yuqi1129: Currently, I intend to use it only for S3; other storage, like GCS, OSS, is not planned for the STS service now.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSourceConfig.java
yuqi1129: dito

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSource.java
yuqi1129: Add docs to address that it only for S3.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSources.java
yuqi1129: changed.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/AwsIrsaCredentialGenerator.java
yuqi1129: Change.

yuqi1129:
> Could it be applied to `s3-token`, too? `s3-token` is used more common.

Yeah, this PR only includes the interface, and supporting ' s3-token` will be in the next PRs.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/WebIdentityTokenSource.java
yuqi1129: Add https://github.com/apache/gravitino/issues/11758 to track it, and I will not do this in this PR.

bundles/aws/src/main/java/org/apache/gravitino/s3/credential/webidentity/FileWebIdentityTokenSource.java
```diff
@@ -0,0 +1,90 @@
...
+      throw new IllegalStateException("Failed to read WebIdentity token file: " + tokenFile, e);
+    }
+  }
+
+  @Override
+  public void close() {}
+
+  String readEnv(String name) {
+    return System.getenv(name);
+  }
```
jerryshao: It is useless to have a such simple helper method. You can directly use the method from `System`.

yuqi1129: Changed to call `System.getenv` directly and removed the test-only override hook.

jerryshao (approved):

PR #11738, opened 2026-06-18
[#11923] docs(core): add multi-node entity store cache invalidation design

### LINKED ISSUE #11923: [Subtask] Add design document about entity store cache improvement in multi-node deployment.
### Describe the subtask

This issue tracks the design document for entity store cache that works in the multi-node deployment.

### Parent issue

#11737

yuqi1129:

Adds a design doc, `design-docs/entity-cache-multinode-changelog-design.md`, proposing how to make the entity store cache (`gravitino.cache.enabled`) usable in multi-node deployments by invalidating it across nodes via the existing `entity_change_log` and `EntityChangeLogPoller`. It presents two options (coarse by-metalake vs. precise per-endpoint) with a recommendation, plus a consistency model and a coverage audit.

Today the entity store cache only invalidates locally, so multi-node deployments must disable it, which hurts performance (Iceberg especially). This is design-only; no code behavior changes yet.

Fix: #11923

No. Documentation and design only.

No code changes; not applicable. The doc passes the repo docs-style audit scripts.


design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,349 @@
...
+
+So caching user, group, and role buys almost nothing (authorization does not depend on it) and costs a lot (the reverse-lookup problem comes back). tag and policy are safe to cache, but they are low-volume governance objects, not the read-heavy metadata the cache exists for.
+
+**Conclusion: cache only the basic metadata objects** — metalake, catalog, schema, table, topic, model, view, fileset. In code this is a small allowlist of entity types in the cache. Every cached entity is then self-contained and one-to-one, so cross-node invalidation stays precise with no reverse lookups. As a bonus, these entities already write ALTER/DROP rows to `entity_change_log`, so the writer side needs no change at all.
+
+The cache now holds only self-contained metadata objects. The next question is how to tell the other nodes to drop a key when it changes.
```
yuqi1129, line 135: There has been a relatively long-standing problem. Currently, due to the introduction of relation in the entity store cache, the overall design has deviated from the original intention of being simple and pluggable. It has become so complex, with the addition of prefix indexes and reverse indexes to maintain the complex relation model. It is impossible for users to introduce a new reliable implementation in a short period of time. So it is inevitable to return to a simple design; otherwise, it would be impossible to achieve plug-and-pull functionality.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,350 @@
...
+- **Authorization checks** (`JcasbinAuthorizer`) read roles straight from mappers (`RoleMetaMapper.listRolesByUserId` / `listRolesByGroupId`), owner from its own `ownerRelCache` (which polls `owner_meta`), and id from its own `metadataIdCache`. It does not use the entity store relation cache at all. Relation caching was first added with authorization in mind, but authorization no longer uses it.
+
+**Cost of removing it.** Each call becomes one indexed query — not an N+1 (one query per item). No list endpoint loads relations per item in a loop. The one place a single request makes several queries is **tag / policy inheritance**: listing an object's tags also reads its parents (the object plus up to ~3 ancestors), so a tag lookup is up to ~4 single-row queries; the cache mainly saves re-reading popular parents. This is the only workload where removing the cache adds real DB load, and it can be measured today — those calls are already `@Monitored` (`baseMetricName = "listTagsForMetadataObject"`, `"getOwner"`, `"listPoliciesForMetadataObject"`).
```
jerryshao, line 97: Is it possible to reduce to 1 query for tag operation?

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,350 @@
...
+
+So caching user, group, and role buys almost nothing (authorization does not depend on it) and costs a lot (the reverse-lookup problem comes back). tag and policy are safe to cache, but they are low-volume governance objects, not the read-heavy metadata the cache exists for.
+
+**Conclusion: cache only the basic metadata objects** — metalake, catalog, schema, table, topic, model, view, fileset. In code this is a small allowlist of entity types in the cache. Every cached entity is then self-contained and one-to-one, so cross-node invalidation stays precise with no reverse lookups. As a bonus, these entities already write ALTER/DROP rows to `entity_change_log`, so the writer side needs no change at all.
```
jerryshao: Should you consider function?

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,350 @@
...
+# Implementation A — Caffeine (`caffeine`, `LOCAL_PER_NODE`, default)
+
+With only self-contained metadata objects cached, "entity X changed" names exactly the key to drop, and **any node can drop it** — no reverse key, no fan-out. The channel to carry "entity X changed" to the other nodes already exists and is already used by other caches.
+
+<!-- diagram source: diagrams/entity-key-invalidation.mmd (regenerate with mermaid-cli) -->
+![Entity-key invalidation via the shared change-log](images/entity-key-invalidation.png)
+
+### Reuse the existing transport
+
+`entity_change_log` and `EntityChangeLogPoller` already exist and are already used by two consumers:
```
jerryshao, line 219: Do you still need a db version checkpoint mechanism to keep the consistency for this change log poll solution?

design-docs/gravitino-entity-cache-multinode-design.md
yuqi1129: Let me analyze optimize it.

design-docs/gravitino-entity-cache-multinode-design.md
yuqi1129: Yeah, it should also be considered, sorry to omit it.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,350 @@
...
+
+A new `EntityChangeLogListener` for the entity store cache runs on the **same poller** and replays each row as a direct invalidation:
+
+```
+onEntityChange(rows):
+    for row in rows:
+        cache.invalidate(identOf(row), row.entity_type)   # drop exactly this entity key
+```
+
+Dropping a container (for example a schema) must also drop its cached children. That uses the **forward** prefix scan the cache already has (`cacheIndex`), on the local node — the reverse index is not involved. This is the only remaining cascade.
```
jerryshao, line 250: How do you guarantee the strong consistency by using polling mechanism? My feeling is that you can only achieve the final consistency with this change log solution, right?

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,350 @@
...
+
+### A third consumer of the existing feed
+
+The entity store cache becomes a third consumer of the poller, next to the catalog cache and the jcasbin id-mapping cache. All three read the same structural ALTER/DROP rows that already exist — the entity store cache clears its entity key, the catalog cache clears its catalog, jcasbin clears its id mapping. There is no new row type, so the existing consumers do not change.
+
+### Consistency
+
+- **Other nodes:** eventual — an entity key is dropped within one poll interval of the change.
```
jerryshao: Is this eventual consistency acceptable?

design-docs/gravitino-entity-cache-multinode-design.md
yuqi1129: Yeah, using the change log can only achieve eventual consistency. The conventional way can only guarantee an eventual consistency model.  If a strong model is required, then a shared cache like Redis should be used, or some ways like version check,  but the benefit of version check is unclear, and in other ways is unworthy to some extent.

design-docs/gravitino-entity-cache-multinode-design.md
yuqi1129: Function and review should both be added.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,395 @@
...
+| 1 | Node A ALTERs a table (add column); B GETs it                  | B serves the old schema for up to one poll interval                               | Accept within the SLA window; a caller that needs an exact schema uses a consistent (cache-bypass) read   |
+| 2 | Node A DROPs a table; B GETs it                                | B still reports it as present (cache hit) until the next poll                     | Accept — any write to it on B is re-validated against the DB and rejected as not-found, so no corruption  |
+| 3 | RENAME (the old name behaves like a DROP); B GETs the old name | B keeps resolving the old name until the next poll; the new name misses → DB load | Same as DROP — bounded by the poll interval                                                               |
+| 4 | A property changes (e.g. Iceberg metadata-location); B GETs it | B returns the old location → a query on B may read an old snapshot/path           | Highest-value case: offer a consistent (cache-bypass) read for location-sensitive callers, or use `redis` |
```
yuqi1129: If Iceberg supports changing the metadata location, this is really a big problem.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -252,10 +252,55 @@ The entity store cache becomes a third consumer of the poller, next to the catal
...
+| 9  | Poller thread dies / DB briefly unreachable                                                               | That node serves stale data until the TTL                 | TTL backstop + monitor poller liveness/lag and alert                                                                         |
+| 10 | Hierarchical DROP: only the container (schema) emits a row; children are cleared by a forward prefix scan | Correct as long as the prefix scan covers every child key | Cover with the hierarchical-drop test; confirm no child is written in a separate transaction that outlives the container row |
+
+**5. Idempotency (no issue).** Replaying ALTER then DROP, in any order or duplicated, is safe: invalidation is idempotent, so out-of-order or repeated replay has no effect.
+
+#### Summary: main risks, resolutions, and the staleness SLA
+
+- **Correctness is never at stake.** The DB plus the optimistic version lock stay authoritative; the worst a stale read can do is surface a bounded-old value (scenarios 1–4) or an unexpected conflict (scenario 7), never a lost update or corruption.
```
jerryshao: One scenario is the modification scenario. If we read the stale data, and this stale data will not raise the exception when querying, and will implicitly return a wrong value, this could be a big problem.

If the modification throws an exception when querying, then the stale data should be safe.

I think you need to think about this problem, and clarify it as much as you can.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -252,10 +252,55 @@ The entity store cache becomes a third consumer of the poller, next to the catal
...
+| 1 | Node A ALTERs a table (add column); B GETs it                  | B serves the old schema for up to one poll interval                               | Accept within the SLA window; a caller that needs an exact schema uses a consistent (cache-bypass) read   |
+| 2 | Node A DROPs a table; B GETs it                                | B still reports it as present (cache hit) until the next poll                     | Accept — any write to it on B is re-validated against the DB and rejected as not-found, so no corruption  |
+| 3 | RENAME (the old name behaves like a DROP); B GETs the old name | B keeps resolving the old name until the next poll; the new name misses → DB load | Same as DROP — bounded by the poll interval                                                               |
+| 4 | A property changes (e.g. Iceberg metadata-location); B GETs it | B returns the old location → a query on B may read an old snapshot/path           | Highest-value case: offer a consistent (cache-bypass) read for location-sensitive callers, or use `redis` |
```
jerryshao: I think you should consider all the supported "alter" scenarios for all the cached entities. To see if such a case will lead to the unexpected scenario. If so, we should introduce db version checking like authorization cache.

yuqi1129: Okay, I will check it on a case-by-case basis.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,412 @@
...
+| metalake — disable                                                    | a node with a stale copy still thinks the metalake is on and **lets operations run** on a metalake that was turned off. (Turning it back on is safe: a stale node only throws "in use" by mistake.) |
+
+#### What we do about it
+
+We keep the simple design and remove the danger at the source:
+
+- **Do not cache the model area (model, model version) or functions.** The cache exists for the read-heavy path, which today is tables; model and function reads are a small share of traffic, so reading them from the DB every time costs almost nothing. Leaving these areas out removes four of the five dangerous cases outright and keeps the rule simple (an entity type is either in or out — no half-cached entity with a version-checked pointer). These reads are then always fresh, on every node. If model or function traffic grows later, we can add them back with a proper freshness check.
+- **Read the metalake on/off flag straight from the DB, not the cache.** This is a cheap single-row read on the guard check only, and it removes the last dangerous case.
```
yuqi1129: I'm not so sure about this point. Reading it from databases every time is not so efficient, and we need to come up with a better plan for it.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -252,54 +254,68 @@ The entity store cache becomes a third consumer of the poller, next to the catal
...
+| model version — update / add / remove URI                             | the URI points to the model files. An old URI silently loads the **wrong model files**.                                                                                                             |
+| model version — update aliases                                        | an alias silently points to the **wrong version**.                                                                                                                                                  |
+| model — latest version (after a new version is added on another node) | "get the latest version" silently returns an **old version**.                                                                                                                                       |
+| function — add / update / remove implementation or definition         | the implementation is the code the function runs. An old copy silently runs the **wrong function code**.                                                                                            |
```
jerryshao, line 297: Will you do the cache for these scenarios?

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -252,54 +254,68 @@ The entity store cache becomes a third consumer of the poller, next to the catal
...
+| metalake — disable                                                    | a node with a stale copy still thinks the metalake is on and **lets operations run** on a metalake that was turned off. (Turning it back on is safe: a stale node only throws "in use" by mistake.) |
```
jerryshao, line 298: How about others like tag, policy?

yuqi1129: Tags and policies are all self-managed and will also have consistency problems during the fresh window, so for the entity store cache, I plan to omit them temporarily. 

I noticed that we are developing a feature: tag-based access control. I'm not sure whether the performance can be satisfied when we disable the cache for tags and policies.

design-docs/gravitino-entity-cache-multinode-design.md
yuqi1129: For the model, functions, potential consistency problems will happen, so I'm going to disable them temporarily, and I could not find a good solution to avoid the defect during the refresh window.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -0,0 +1,412 @@
...
+
+#### What we do about it
+
+We keep the simple design and remove the danger at the source:
+
+- **Do not cache the model area (model, model version) or functions.** The cache exists for the read-heavy path, which today is tables; model and function reads are a small share of traffic, so reading them from the DB every time costs almost nothing. Leaving these areas out removes four of the five dangerous cases outright and keeps the rule simple (an entity type is either in or out — no half-cached entity with a version-checked pointer). These reads are then always fresh, on every node. If model or function traffic grows later, we can add them back with a proper freshness check.
```
yuqi1129: @jerryshao 
This is the final decision, and we will exclude entities of those types shortly.

design-docs/gravitino-entity-cache-multinode-design.md
jerryshao: Yes, that's what I concerned about. We can start without tag. But we need to think about it.

design-docs/gravitino-entity-cache-multinode-design.md
jerryshao: OK, then we should define an explicit scope in the code to let developers know that what is cached and what is not. Besides, we should document it.

design-docs/gravitino-entity-cache-multinode-design.md
```diff
@@ -225,16 +225,17 @@ Each row is `{metalake, entity_type, full_name, operate_type (ALTER | DROP), cre
...
+| Change                                                | Cached? | Writes a change-log row               |
+| ----------------------------------------------------- | ------- | ------------------------------------- |
+| structural ALTER/DROP (table, schema, catalog, …)     | yes     | yes — already emitted                 |
+| tag / policy ALTER/DROP                               | yes     | **add a new emit point** (none today) |
+| user / group / role ALTER/DROP                        | no      | not needed (not cached)               |
+| model / model version / function ALTER/DROP           | no      | not needed (not cached)               |
+| relation change (grant, set owner, attach tag/policy) | no      | not needed (not cached)               |
 
-The existing structural rows are written in the same transaction as the entity write, so nothing new is added here.
+The existing structural rows are written in the same transaction as the entity write; the new tag/policy rows must be written the same way, in the same transaction as the tag/policy write.
```
jerryshao, line 238: The key thing is to cache the relation between the tag and entity. The relation is used much more frequently when we do tag-based access control.

yuqi1129: I see. Since the work on tag-based access control is on the way, we need to rethink whether it's proper to use all relations from the entity cache.  If we indeed decide to drop them all, alternatives should be provided for tag relations.

yuqi1129: Thinking a bit. 

The principal that the entity store cache will only handle single entities remains unchanged. If the performance of tag relations is poor during authorization,  I will cooperate with the guy who is responsible for tag-based access control to optimize it. Entity store cache will only be used in basic API operations like load/alter/list tags.

jerryshao:
What is the status for now?

design-docs/gravitino-entity-cache-multinode-design.md
roryqi: @yuqi1129 Could u sync this place final conclusion to me? What's the current state? What's the final state?

yuqi1129: @roryqi

For the entity store cache, the tag and policy will be cached as designed. If strong consistency is required for access control, I will modify it in the future.

roryqi: From my view, eventual consistence is accepted. Performance is more important.

yuqi1129: One thing that needs to be noted is that the entity cache only stores a single entity, and relations will be removed as design. It will greatly impact tag/policy operations if it's called frequently. @roryqi

yuqi1129:
> What is the status for now?

The design is stable, and I'm actively working on its subtasks.

yuqi1129:
@jerryshao 
If the document is generally okay with you, can you merge it when you have time, and then I can continue to push implementation PRs to merge.

jerryshao:
> @jerryshao If the document is generally okay with you, can you merge it when you have time, and then I can continue to push implementation PRs to merge.

OK, I will review it today.

jerryshao (approved):

PR #11739, opened 2026-06-18
[#11736] fix(core): retry failed entity change log listeners and do not remove logs when polling.

### LINKED ISSUE #11736: [Improvement] EntityChangeLogPoller can permanently miss cache invalidations in multi-node deployments
**Version**: main branch

### Describe what's wrong

The shared `EntityChangeLogPoller` (consumed today by jcasbin caches, and planned to back entity-store cache multi-node invalidation) has three independent paths that cause a change-log row's invalidation to be **permanently lost**, leaving stale data on remote nodes with no self-healing.

**1. Auto-increment id cursor has a commit-ordering gap.**
The cursor is `WHERE id > highWater ORDER BY id` (`EntityChangeLogBaseSQLProvider#selectEntityChanges`). Auto-increment ids are assigned at INSERT time but become visible at COMMIT time, so a lower id can commit *after* a higher id:

```
Txn A inserts change id=100 (not yet committed)
Txn B inserts change id=101, commits first
poller sees 101 -> advances highWater to 101
Txn A commits id=100
next poll: WHERE id > 101 -> row 100 is never read -> its invalidation is lost
```

**2. Cursor advances even when a listener throws.**
In `EntityChangeLogPoller#doPollChanges`, listener exceptions are caught and `entityPollHighWaterId` is advanced unconditionally. A single transient listener failure permanently drops that batch's invalidations.

**3. `expireAfterAccess` turns a missed invalidation into permanent staleness.**
`CaffeineEntityCache` builds the cache with `expireAfterAccess` (TTL resets on every read). A continuously-read hot key whose invalidation was lost via (1) or (2) never expires -> stale indefinitely. Cold keys self-heal; hot keys do not.

No crash/stacktrace -- silent stale reads. Observable as a remote node returning an outdated entity/relation after another node committed an ALTER/DROP/grant/revoke/setOwner.

### How to reproduce

- Version: main branch, multi-node deployment sharing one backend DB.
- (1): drive concurrent writes from multiple nodes so change-log inserts commit out of id order; observe a remote node never invalidating one of the changes.
- (2): make a registered `EntityChangeLogListener` throw on one batch; observe the cursor advancing past it and the invalidation never being retried.
- (3): keep reading a hot key on a remote node after its invalidation was missed; observe it never expiring despite the configured TTL.

### Additional context

Suggested fixes (one PR can cover all three):
1. Cursor by `created_at` with a trailing overlap window + id-dedup, or a lagging high-water mark, to close the commit-ordering gap.
2. Do not advance the high-water past a batch whose listener failed (or retry that batch); keep listeners idempotent.
3. Use `expireAfterWrite` (alone or in addition to `expireAfterAccess`) so a missed invalidation degrades to bounded staleness instead of permanent.

Found during design of entity-store cache multi-node invalidation; problems (1) and (2) already affect the existing jcasbin cache path.

yuqi1129:

This PR makes `EntityChangeLogPoller` retry a batch when a listener fails, with a bounded number of retries:

- Keep the fetched batch until every registered listener applies it.
- Retry only listeners that have not succeeded, and stop waiting for one that is unregistered.
- Advance the high-water cursor after all pending listeners succeed.
- Bound the retries with `gravitino.entityChangeLog.listenerMaxRetries` (default 10). Once a listener exhausts them, `gravitino.entityChangeLog.listenerFailureAction` decides what happens: `EXIT` (default) stops this server, because its local caches are known to be stale; `SKIP` drops the batch for that listener and keeps serving. `EXIT` goes through `System.exit`, so the server shutdown hook still stops the server gracefully.
- Keep `CatalogChangeLogListener` swallowing its own failures instead of propagating them. `CatalogManager.consumeLocalMutation` is single-shot, so re-delivering an already-applied batch would invalidate a catalog this process mutated itself and close its in-use `IsolatedClassLoader`. A dropped invalidation only costs staleness bounded by the catalog cache TTL.
- Increase the fetch batch size from 500 to 2000.
- Log poller startup, consumed ID ranges, cursor advancement, retries, and listener failures.
- Move cleanup out of the polling path into a dedicated single-threaded cleaner, and run its first pass after a short randomized delay so that a frequently restarted server still prunes and HA nodes do not delete the same rows at once.
- Retain change logs for 30 days by default and run cleanup daily.
- Use database time for both insertion and expiration, and commit each 1000-row cleanup batch.
- Close every entity store component even if an earlier one fails or was never created.

This PR intentionally does not address the commit-ordering gap or cache-expiration behavior tracked by #11736.

Previously, the poller advanced its cursor even when a listener threw, permanently dropping that batch for the failed listener. Retrying without a bound has the opposite problem: a permanently failing listener would freeze cache invalidation for the whole process, so the retries are now bounded and the outcome is explicit. Cleanup was coupled to polling, used a non-committing session (so deletions were never committed), and calculated expiration from each server's JVM clock.

Fixed: #11736.

Yes.

- New config `gravitino.entityChangeLog.listenerMaxRetries` (default `10`).
- New config `gravitino.entityChangeLog.listenerFailureAction` (default `EXIT`). With the default, a server whose change log listener keeps failing after its retries stops itself instead of serving from caches it knows to be stale. Set it to `SKIP` to prefer availability.
- The default change-log retention increases from 1 day to 30 days, and the default cleanup interval changes from 1 hour to 1 day.

Both new configs are documented in `docs/gravitino-server-config.md`.

- `./gradlew :core:test -PskipITs -PskipDockerTests=false`
- Added coverage for failed-listener retries, the retry bound under both `EXIT` and `SKIP`, a 2001-record backlog, a paused batch blocking new fetches, catalog-listener failure containment, the cleaner lifecycle and its initial delay, independent batched cleanup, database-time SQL, and committed cleanup visibility from a new SQL session.
- The pruning statements are exercised against H2. The PostgreSQL variant is only asserted at the SQL-string level, matching the existing PostgreSQL provider tests in `core`; verifying it against a real PostgreSQL would need a new docker-tagged integration test.


yuqi1129:
Thanks for the review — addressed in e85067982:

1. **Cross-field validation (EntityChangeLogPoller):** the constructor now rejects `retentionMs <= pollLagMs` when cleanup is enabled (`retentionMs > 0`), so a row can never be pruned before it becomes eligible for polling. Added `testRejectsRetentionNotGreaterThanLag`.
2. **Config / Javadoc / docs wording (Configs, EntityChangeLogBaseSQLProvider, gravitino-server-config.md):** dropped the inaccurate "all transactions that started before a consumed row have committed" claim. The wording now states the actual guarantee: a smaller-id row is inserted no later than a larger-id row, so waiting until a row is `pollLagMs` old gives that row's *inserting* transaction time to commit before the cursor advances past it — preventing the commit-ordering gap as long as the lag exceeds the longest write transaction.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -38,8 +38,18 @@
  *
  * <p>The poller owns the single high-water mark for a Gravitino server process and dispatches each
  * consumed batch to registered listeners. Listeners should only perform idempotent local cache
- * invalidation. The cursor always advances after dispatch regardless of individual listener
- * failures, so a faulty listener cannot block other listeners or prevent pruning.
+ * invalidation. A failing listener does not block the other listeners in the same cycle, but the
+ * cursor is <b>not</b> advanced past a batch whose dispatch had any listener failure, so the batch
+ * is retried on the next cycle until every listener succeeds. Because invalidation is idempotent,
+ * re-dispatching an already-applied batch to a healthy listener is harmless.
```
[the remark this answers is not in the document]
yuqi1129: Clarified the wording in `EntityChangeLogPoller`: a failed listener does not prevent other listeners from being invoked in the same cycle, but the shared cursor pauses forward progress until the failed batch is retried successfully. I also bounded the retry wording by the retention window to match the cleanup behavior.

jerryshao (commented):
Overall the fix is sound and well-tested. Two bugs (commit-ordering gap and listener-failure cursor advance) are addressed correctly. Comments below are one functional concern, two style/usability nits, and one minor test coverage gap.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -68,6 +83,7 @@ public EntityChangeLogPoller(long pollIntervalSecs) {
         pollIntervalSecs,
         TimeUnit.DAYS.toMillis(1),
         TimeUnit.HOURS.toMillis(1),
+        TimeUnit.SECONDS.toMillis(1),
         System::currentTimeMillis);
```
jerryshao: **Style: hardcoded `1` should reference `DEFAULT_ENTITY_CHANGE_LOG_POLL_LAG_SECS`.**

This repeats the magic value that is also defined in `Configs.DEFAULT_ENTITY_CHANGE_LOG_POLL_LAG_SECS`. If the default is ever changed in `Configs`, this single-arg constructor is silently left inconsistent.

```java
// Prefer
TimeUnit.SECONDS.toMillis(DEFAULT_ENTITY_CHANGE_LOG_POLL_LAG_SECS),
```

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -183,22 +215,31 @@ private synchronized void doPollChanges() {
     }
 
     List<EntityChangeRecord> dispatchedChanges = Collections.unmodifiableList(changes);
+    boolean allListenersSucceeded = true;
     for (EntityChangeLogListener listener : listeners) {
       try {
         listener.onEntityChange(dispatchedChanges);
       } catch (Exception e) {
+        allListenersSucceeded = false;
         LOG.warn("Entity change listener {} failed", listener.getClass().getName(), e);
```
jerryshao: **Concern: `LOG.warn` is likely too quiet for a persistent listener failure.**

With the new cursor semantics, a listener failure blocks forward progress for ALL new change rows. If a listener fails on every cycle, the cache silently serves stale data until it recovers — or until retention cleanup prunes the stuck rows, at which point the invalidations are permanently lost. A single `WARN` per cycle is easy to miss in production.

Consider escalating to `LOG.error`, or tracking a consecutive-failure counter per listener and switching to `LOG.error` after N failures. At that point the operator has a closing window (bounded by `retentionMs / pollIntervalMs` cycles) before invalidations are permanently dropped.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
```diff
@@ -77,20 +93,36 @@ public EntityChangeLogPoller(long pollIntervalSecs) {
...
    * @param cleanupIntervalMs interval between successive cleanup attempts in milliseconds
+   * @param pollLagMs lag in milliseconds; only rows older than this (by the DB clock) are consumed,
+   *     or 0 to disable the lag
    */
-  public EntityChangeLogPoller(long pollIntervalSecs, long retentionMs, long cleanupIntervalMs) {
-    this(pollIntervalSecs, retentionMs, cleanupIntervalMs, System::currentTimeMillis);
+  public EntityChangeLogPoller(
+      long pollIntervalSecs, long retentionMs, long cleanupIntervalMs, long pollLagMs) {
+    this(pollIntervalSecs, retentionMs, cleanupIntervalMs, pollLagMs, System::currentTimeMillis);
   }
```
jerryshao, line 102: **Minor: Preconditions error message omits the actual values.**

```java
"retentionMs must be greater than pollLagMs when cleanup is enabled"
```

When this fires at startup the operator sees an `IllegalArgumentException` without knowing which values triggered it. Prefer:

```java
String.format(
    "retentionMs (%d) must be greater than pollLagMs (%d) when cleanup is enabled",
    retentionMs, pollLagMs)
```

core/src/test/java/org/apache/gravitino/storage/relational/mapper/provider/base/TestEntityChangeLogMapper.java
```diff
@@ -150,12 +150,29 @@ void testEntityChangeLogSameTimestampOrderedById() throws SQLException {
...
     Assertions.assertTrue(rows.get(0).getId() < rows.get(1).getId());
     Assertions.assertTrue(rows.get(1).getId() < rows.get(2).getId());
   }
 
+  @Test
+  void testEntityChangeLogLagExcludesFreshRows() throws SQLException {
+    // A freshly written row is younger than the lag window, so it must not be consumed yet.
+    entityChangeLogMapper.insertEntityChange(
+        "metalake1", "TABLE", "metalake1.cat.schema.fresh", OperateType.ALTER);
+    List<EntityChangeRecord> withLag = entityChangeLogMapper.selectEntityChanges(0L, 60_000L, 100);
```
jerryshao: **Minor: `lagMs = 0` ("disable lag") path not explicitly asserted in this new test.**

Since `lagMs = 0` is the documented contract for users who set `pollLagSecs = 0`, a brief assertion that a fresh row IS immediately returned with `lagMs = 0` would pin the zero-lag SQL behaviour at the mapper level:

```java
// lagMs = 0 disables the filter — fresh rows must be consumable immediately
List<EntityChangeRecord> noLag = entityChangeLogMapper.selectEntityChanges(0L, 0L, 100);
Assertions.assertFalse(noLag.isEmpty(), "lagMs=0 should return the fresh row");
```

diqiu50 (commented):
The lag is only applied on the regular polling path. `start()` still initializes `entityPollHighWaterId` from unlagged `selectMaxChangeId()`, so a restart can still skip a lower id that was inserted but uncommitted while a higher id was already committed. Once the cursor is initialized past that lower id, the lagged polling query cannot recover it. The startup cursor should be initialized using the same lagged/eligible boundary.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: OK, it makes sense.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: I'm hesitant about these two levels and adopt the suggestion.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogPoller.java
yuqi1129: OK

core/src/test/java/org/apache/gravitino/storage/relational/mapper/provider/base/TestEntityChangeLogMapper.java
yuqi1129: Changed.

yuqi1129:
> The lag is only applied on the regular polling path. `start()` still initializes `entityPollHighWaterId` from unlagged `selectMaxChangeId()`, so a restart can still skip a lower id that was inserted but uncommitted while a higher id was already committed. Once the cursor is initialized past that lower id, the lagged polling query cannot recover it. The startup cursor should be initialized using the same lagged/eligible boundary.

Changed `selectMaxChangeId`.

common/src/main/java/org/apache/gravitino/config/ConfigConstants.java
```diff
@@ -92,6 +92,9 @@ private ConfigConstants() {}
   /** The version number for the 1.3.0 release. */
   public static final String VERSION_1_3_0 = "1.3.0";
 
+  /** The version number for the 1.4.0 release. */
+  public static final String VERSION_1_4_0 = "1.4.0";
```
jerryshao: Our next release will be 2.0.0, you probably need to update this.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
```diff
@@ -35,14 +35,31 @@ public class EntityChangeLogBaseSQLProvider {
    * #selectMaxChangeId()} because its cache starts empty and it does not need historical
    * invalidations. Re-consuming a row on an existing instance is acceptable: entity DROP/ALTER
    * handling only invalidates cache keys, and invalidation is idempotent.
+   *
+   * <p>The {@code created_at <= dbNowMs - lagMs} predicate applies a lagging high-water mark.
+   * Auto-increment ids are assigned at INSERT time but become visible at COMMIT time, so a lower id
+   * can commit after a higher id; advancing the cursor purely by the largest committed id can step
```
jerryshao: How could this be happened, is this a auto incremental id?

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
```diff
@@ -35,14 +35,31 @@ public class EntityChangeLogBaseSQLProvider {
...
+   * over a not-yet-committed lower id and lose its invalidation forever. A smaller-id row was
+   * necessarily inserted no later than a larger-id row, so by only consuming rows whose {@code
+   * created_at} is at least {@code lagMs} in the past (measured by the database clock, so there is
```
jerryshao: How do you know that 1 sec lag is enough? If for example, we hit minor or major gc in the Gravitino side, the id committed should be late. In this case, is it enough for just 1 second by default?

My feeling is that instead of adding a lag. We should make sure that the id we insert is incremental. And what the reader gets is always sequential.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
yuqi1129: Yes — id is BIGINT AUTO_INCREMENT (MySQL) / BIGSERIAL (PG). The id is set when the INSERT runs, but the row is only visible after COMMIT. The change-log insert commits together with the rest of the metadata write (the soft-deletes etc.), so a lower id can become visible after a higher id that committed first. An id-only cursor can then jump over the lower id before it shows up and drop its invalidation for good. So the gap is real.

The following is an example:

```text
Transaction A starts → INSERT gets id=10 (slow transaction, late commit)
Transaction B starts → INSERT gets id=11, fast COMMIT
Poller runs: sees id=11 (committed), cursor advances to 11
Transaction A finally COMMIT, id=10 becomes visible
Poller next check WHERE id > 11 → id=10 never seen ❌ Failure information permanently lost
```

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/base/EntityChangeLogBaseSQLProvider.java
yuqi1129: The lag is just a best-effort guess, not a guarantee. It has to be bigger than the longest time between a row's INSERT and its COMMIT, and a GC pause, a lock wait, a slow disk write, or a long transaction can easily push that past 1s. When it does, the gap comes back and the invalidation is lost silently. So a fixed 1s default is fragile.

I will try to think about it again to gain a better solution.

yuqi1129: I reworked it. The cursor is now created_at instead of id, and every cycle it re-reads a small window before the cursor:

```sql
  WHERE created_at > #{watermarkMs} - #{overlapMs} ORDER BY created_at, id
```

So if a row commits late (its created_at is behind the cursor), the next poll re-reads that window and picks it up, instead of skipping it forever. Re-reading a row we already handled is safe because invalidation is idempotent.

About your "make the ids the reader sees always in order" idea: I did not go that way because auto-increment ids are not reused after a rollback, so a rolled-back id never shows up and the reader would wait on it forever. The overlap window avoids that problem.

One thing I want to be clear about: the overlap does not add delay. A new change is still applied within one poll interval. The overlap only re-scans the past window to catch late commits, it does not hold back new rows
(this is the difference from the old lag).

Config renamed pollLagSecs -> pollOverlapSecs (default 30s, should be larger than the longest INSERT-to-COMMIT time). Done in the latest commit.

common/src/main/java/org/apache/gravitino/config/ConfigConstants.java
yuqi1129: modified.

yuqi1129:
@roryqi 

roryqi:
> @roryqi Can you help to review this one?

I  will take a  look.

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/EntityChangeLogPostgreSQLProvider.java
```diff
@@ -26,6 +27,13 @@
 
 public class EntityChangeLogPostgreSQLProvider extends EntityChangeLogBaseSQLProvider {
 
+  /**
+   * PostgreSQL flavour of the DB-side "now in milliseconds" expression. Insertion and expiration
+   * share it, so retention is measured entirely with the database clock.
+   */
+  private static final String CURRENT_TIME_MILLIS_SQL =
```
roryqi: It would better to put it a better place. Other class can reuse this constant easily.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogCleaner.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import org.apache.gravitino.storage.relational.mapper.EntityChangeLogMapper;
+import org.apache.gravitino.storage.relational.utils.SessionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Periodically removes expired rows from {@code entity_change_log} on a dedicated thread.
```
roryqi, line 33: If we use unreasonable parameter, it may cause the changes lost
Could u add the parameter for the similar cases?

core/src/main/java/org/apache/gravitino/storage/relational/mapper/provider/postgresql/EntityChangeLogPostgreSQLProvider.java
yuqi1129: Good point, done. I moved both flavours into `org.apache.gravitino.storage.relational.mapper.provider.DatabaseTimeSQL`, which sits in the package shared by the `base`, `h2` and `postgresql` provider packages, so any provider can reuse them: `DatabaseTimeSQL.MYSQL` and `DatabaseTimeSQL.POSTGRESQL`. The entity change log providers now reference the constants.

I did not migrate the other providers in this PR on purpose. The same literals are inlined 114 times across 32 files (MySQL flavour) and 107 times across 31 files (PostgreSQL flavour), so replacing them all here would bury this PR's actual fix under a large mechanical diff. Filed #12236 to do that separately.

core/src/main/java/org/apache/gravitino/storage/relational/EntityChangeLogCleaner.java
yuqi1129: Agreed, that failure mode is silent, which makes it worse. `EntityChangeLogCleaner` now takes the poll interval and validates the retention against it at startup:

```
retentionMs == 0 (cleanup disabled), or retentionMs >= 10 * pollIntervalMs
```

Anything shorter throws `IllegalArgumentException` during initialization instead of silently pruning records before every node consumed them. The bound is 10 poll cycles because a record has to outlive more than a single cycle: a node can miss cycles while restarting, during a long GC pause, or while the poller is paused retrying a failed listener. The comparison uses division rather than multiplication so a large configured poll interval cannot overflow.

Covered by `TestEntityChangeLogCleaner#testRejectsRetentionShorterThanTheConsumptionWindow`, and the constraint is documented for `gravitino.entityChangeLog.retentionSecs` in `docs/gravitino-server-config.md`.

roryqi (approved):
LGTM.

PR #11087, opened 2026-05-14
[#11086] feat(catalog-glue): Improve Glue Iceberg table support via Iceberg SDK

diqiu50:
### What changes were proposed in this pull request?

Improve Iceberg table lifecycle management in the Glue catalog by delegating operations to the Iceberg SDK GlueCatalog:

- createTable: use Iceberg SDK for non-register-mode tables to write metadata.json to S3
- loadTable: recover partitioning and sort orders from Iceberg metadata (Glue partitionKeys() is empty for Iceberg tables)
- alterTable: use Iceberg SDK UpdateSchema/UpdateProperties; fall back to native Glue SDK for register-mode tables
- Merge GlueIcebergHelper and GlueIcebergCatalogHelper into GlueIcebergTableHelper
- Support partition transforms (year, month, day, hour, bucket, truncate)

### Why are the changes needed?

Without Iceberg SDK integration, Glue skips writing metadata.json, making tables unreadable by Iceberg-native engines. loadTable also returns empty partitioning for Iceberg tables.

Fix: #11086

### Does this PR introduce _any_ user-facing change?

Yes — users can now create and alter Iceberg tables in Glue via the Iceberg SDK.

### How was this patch tested?

- TestGlueIceberg — 10 unit tests
- MotoGlueCatalogIT — 24 integration tests (Moto mock)
- AwsGlueCatalogIT — 24 integration tests (real AWS Glue)

PR #11750, opened 2026-06-22
[MINOR ]fix(license): upgrade jcasbin version to fix Aviator license

mchades:

upgrade jcasbin version to fix Aviator license

see:https://lists.apache.org/thread/fsd27p87178ygvo6cp2kcd1m99pvsrpn

> - In the binary NOTICE, the required notices for bundled Apache Arrow and Apache Kafka are not propagated. In the source NOTICE, the RC4 LICENSE additions were not added: Apache Paimon and Apache Doris should be added (Iceberg is already present, and Lance Namespace does not need anything, as it ships no NOTICE file). [1]
> - Category X dependency: Aviator (LGPL-3.0) in the binaries. aviator-5.4.3.jar is bundled in the main binary (libs/and iceberg-rest-server/libs/) and in the standalone iceberg-rest-server binary. Its licence is LGPL-3.0, confirmed from Aviator's own upstream licenses.txt. LGPL is Category X and must not be distributed in an Apache release. As this is not a new issue, it might be worth discussing it on legal-discuss to determine whether other releases need to be corrected/removed. [2]
> 1. https://infra.apache.org/licensing-howto.html#alv2-dep
> 3. https://www.apache.org/legal/resolved.html#prohibited

no

CI pass


NOTICE
```diff
@@ -13,6 +13,12 @@ for it's contents.
 Apache Iceberg
 Copyright 2017-2022 The Apache Software Foundation
 
+Apache Paimon
+Copyright 2023-2026 The Apache Software Foundation
+
+Apache Doris
+Copyright 2018-2026 The Apache Software Foundation
```
yuqi1129, line 20: So this problem should have existed for a long time?

mchades: Partly. It's two different cases:

- Paimon / Doris (this file, source NOTICE): these go with code added during 1.3
  (PaimonViewTestCatalogHelper #11013, GravitinoGlueCredentialsProvider #11087),
  so the gap is new to this release - v1.2.1's NOTICE didn't have them either
  because the attributions themselves are new.

- Arrow / Kafka (NOTICE.bin): yes, that one is long-standing. Arrow and Kafka
  have been bundled for a while and were already in the source NOTICE, but the
  binary NOTICE was never kept in sync. SCANOSS/Justin's review is what surfaced
  it.

So this PR fixes both the new (Paimon/Doris) and the pre-existing (Arrow/Kafka in
the binary) gaps.

yuqi1129: Yeah. If possible, please just contact @justinmclean to make sure there are no similar problems. The current changes LGTM.

yuqi1129 (approved):

PR #11765, opened 2026-06-22
[#11486] refactor(catalog): Add shared Hadoop Kerberos auth module

### LINKED ISSUE #11486: [EPIC] Unify Kerberos keytab file fetching and configuration
### Describe the proposal

PR #11354 added remote URI validation to the Kerberos keytab fetching implementations. The repository currently contains multiple similar `FetchFileUtils` implementations across Hadoop, Hive, Iceberg, and Paimon modules.

These implementations have different dependency and behavior requirements:

- Hadoop, Hive, and Iceberg support HDFS and depend on Hadoop APIs.
- Paimon does not currently support HDFS keytab fetching.
- Hive provides additional concurrent and atomic symlink handling.
- Configuration keys and prefixes differ between catalogs and services.

A large refactoring close to a release would introduce unnecessary regression risk. Keep the current implementations and configuration keys stable for this release, then consolidate them in the next development iteration.

The refactoring must remain backward compatible with all configuration keys published by PR #11354. Existing configurations must continue to work after the consolidation. If a canonical key is introduced, legacy keys must be supported as aliases with documented precedence and deprecation behavior.

Related PR: #11354

### Task list

- [x] #11487
- [x] #11488
- [x] #11489
- [x] #11490
- [x] #11491
- [x] #11492

yuqi1129:

This PR adds a new `catalogs:hadoop-auth` module to centralize shared Hadoop Kerberos authentication logic, including keytab fetching, krb5.conf setup, principal validation, login, and ticket refresh.

It updates the Kerberos clients in hadoop-common, Hive metastore common, Paimon, and Iceberg common to delegate common logic to the new module while preserving their existing catalog-specific behavior.

Multiple catalogs had duplicated Kerberos authentication logic. Centralizing the shared Hadoop-dependent logic reduces duplication and makes future fixes easier to apply consistently across catalogs.

Fix: #11486

No.

- `./gradlew :catalogs:hadoop-auth:test :catalogs:hadoop-common:compileJava :iceberg:iceberg-common:compileJava :catalogs:catalog-lakehouse-paimon:compileJava :catalogs:hive-metastore-common:compileJava -PskipITs`
- `./gradlew compileDistribution -x test`
- Kerberos integration tests:
  - Hive HMS Kerberos tests
  - Fileset `HDFSKerberosIT`
  - Hudi Kerberos Hive embedded/deploy
  - Paimon Kerberos filesystem embedded/deploy
  - Iceberg catalog Kerberos Hive embedded/deploy
  - Iceberg REST Kerberos Hive embedded tests, including user impersonation
- `git diff --check`


catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,204 @@
...
+
+  private static ThreadFactory daemonThreadFactory(String threadNameFormat) {
+    AtomicInteger threadId = new AtomicInteger(0);
+    return runnable -> {
+      Thread thread = new Thread(runnable);
+      thread.setDaemon(true);
+      thread.setName(String.format(threadNameFormat, threadId.getAndIncrement()));
+      return thread;
+    };
+  }
```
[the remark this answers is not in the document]
yuqi1129: Fixed. The shared helper now takes a thread name prefix and appends the numeric id directly, so callers do not pass `String.format` patterns.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/HmsKerberosClient.java
```diff
@@ -126,35 +122,19 @@ public UserGroupInformation login() throws Exception {
...
-            } catch (Exception e) {
-              LOG.error("Fail to refresh ugi token: ", e);
-            }
-          },
-          checkInterval,
-          checkInterval,
-          TimeUnit.SECONDS);
+      checkTgtExecutor =
+          KerberosAuthUtils.startTicketRefresh(loginUgi, checkInterval, "check-tgt-%d", LOG);
     }
```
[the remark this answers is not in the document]
yuqi1129: Fixed. Hive Kerberos login now shuts down any existing ticket refresh executor before starting a new one, and the shared Kerberos client does the same on repeated login calls.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,204 @@
...
+
+    String[] principalComponents = principal.split("@", -1);
+    if (principalComponents.length != 2
+        || principalComponents[0].isEmpty()
+        || principalComponents[1].isEmpty()) {
+      throw new IllegalArgumentException("The principal has the wrong format");
+    }
+
+    return principalComponents[1];
+  }
```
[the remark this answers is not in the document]
yuqi1129: Fixed. Principal validation now normalizes the principal once with `trim()` before splitting, so the returned realm is normalized as well.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,204 @@
...
+      return UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, keytabFilePath);
+    }
+
+    UserGroupInformation.loginUserFromKeytab(principal, keytabFilePath);
+    if (loginMode == LoginMode.CURRENT_USER) {
+      return UserGroupInformation.getCurrentUser();
+    }
+
+    return UserGroupInformation.getLoginUser();
+  }
```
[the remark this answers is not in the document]
yuqi1129: Added tests for krb5.conf system property wiring, non-positive refresh interval rejection, blank thread name prefix rejection, daemon refresh thread naming, and trimmed principal realm parsing. I did not add static mocking for `UserGroupInformation` login modes because the existing Kerberos integration tests cover the real login paths without introducing brittle static mocks.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosClient.java
```diff
@@ -0,0 +1,181 @@
...
+      return this;
+    }
+
+    /** Builds the client. */
+    public KerberosClient build() {
+      Preconditions.checkArgument(
+          principal != null && !principal.trim().isEmpty(), "principal can't be blank");
+      Preconditions.checkArgument(hadoopConf != null, "hadoopConf can't be null");
+      return new KerberosClient(this);
+    }
```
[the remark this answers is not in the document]
yuqi1129: Fixed in b59367232. `KerberosClient.Builder.build()` now validates `loginMode`, the refresh interval/prefix when refresh is enabled, and paired krb5.conf keys before constructing the client.

catalogs/hadoop-auth/src/test/java/org/apache/gravitino/catalog/hadoop/auth/TestKerberosAuthUtils.java
```diff
@@ -0,0 +1,142 @@
...
+  public void testStartTicketRefreshCreatesDaemonThreadWithPrefix() {
+    UserGroupInformation ugi = Mockito.mock(UserGroupInformation.class);
+    ScheduledThreadPoolExecutor executor =
+        KerberosAuthUtils.startTicketRefresh(
+            ugi, 60, "check-test-tgt-", LoggerFactory.getLogger(getClass()));
+
+    try {
+      Thread thread = executor.getThreadFactory().newThread(() -> {});
+      Assertions.assertTrue(thread.isDaemon());
+      Assertions.assertEquals("check-test-tgt-1", thread.getName());
```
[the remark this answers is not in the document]
yuqi1129: Fixed in b59367232. The test now checks the actual scheduled refresh thread named with the 0-based suffix (`...-0`) instead of creating a second thread from the executor factory.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,241 @@
...
+   * @param hadoopKrb5ConfKey Hadoop configuration key that stores the krb5.conf path
+   * @param systemKrb5ConfKey JVM system property key for the krb5.conf path
+   */
+  public static void configureKrb5Conf(
+      Configuration hadoopConf, String hadoopKrb5ConfKey, String systemKrb5ConfKey) {
+    String krb5Config = hadoopConf.get(hadoopKrb5ConfKey);
+    if (krb5Config != null) {
+      System.setProperty(systemKrb5ConfKey, krb5Config);
+    }
+  }
```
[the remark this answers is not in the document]
yuqi1129: Fixed in b59367232. `configureKrb5Conf()` now rejects null/blank Hadoop and system keys explicitly, and the builder also rejects invalid krb5.conf key pairs.

jerryshao:
Can we do this in hadoop-common module?

yuqi1129:
> Can we do this in hadoop-common module?

I think keeping this as a separate hadoop-auth module is a cleaner boundary.

hadoop-common is currently the filesystem/provider common module and is used by catalogs, connectors, and bundle paths. It also declares Hadoop dependencies as implementation, not compileOnly. If we put the shared Kerberos helpers there, non-filesystem modules such as hive-metastore-common and Paimon would need to depend on hadoop-common just to reuse Kerberos login/keytab utilities, and they would also inherit hadoop-common's broader runtime dependency surface.

The code here is Hadoop/Kerberos auth related, but not HDFS filesystem specific. It is shared by Hive, Iceberg, Paimon, and hadoop-common, so a small hadoop-auth module keeps the dependency boundary narrower.

One concern is that hadoop-auth currently uses compileOnly Hadoop dependencies, so each consumer must provide Hadoop runtime dependencies explicitly. I will double-check those consumer dependencies and make that contract clear, but I would prefer not to merge this into hadoop-common just to solve that.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/KerberosClient.java
```diff
@@ -126,35 +122,20 @@ public UserGroupInformation login() throws Exception {
 
     // Check the principal and keytab file
     String catalogPrincipal = kerberosConfig.getPrincipalName();
-    Preconditions.checkArgument(
-        StringUtils.isNotBlank(catalogPrincipal), "The principal can't be blank");
-    List<String> principalComponents = Splitter.on('@').splitToList(catalogPrincipal);
-    Preconditions.checkArgument(
-        principalComponents.size() == 2, "The principal has the wrong format");
+    KerberosAuthUtils.checkPrincipalAndGetRealm(catalogPrincipal);
```
diqiu50: `KerberosAuthUtils.login` call the function  `checkPrincipalAndGetRealm` too

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
```diff
@@ -273,10 +275,24 @@ private void closeInternalClientPool() {
...
+              .loginMode(KerberosAuthUtils.LoginMode.CURRENT_USER)
+              .checkIntervalSec(kerberosConfig.getCheckIntervalSec())
+              .threadNamePrefix("check-Iceberg-Hive-tgt-")
+              .build();
       // catalog_uuid always exists for Gravitino managed catalogs, `0` is just a fallback value.
       String catalogUUID = properties().getOrDefault("catalog_uuid", "0");
-      File keytabFile = kerberosClient.saveKeyTabFileFromUri(Long.parseLong(catalogUUID));
+      File keytabFile =
+          new File(
+              String.format(KerberosConfig.GRAVITINO_KEYTAB_FORMAT, Long.parseLong(catalogUUID)));
```
diqiu50: Long.parseLong(catalogUUID) is redundant.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,251 @@
...
+   * @return the login UGI
+   * @throws IOException if login fails
+   */
+  public static UserGroupInformation login(
+      String principal, String keytabFilePath, Configuration hadoopConf, LoginMode loginMode)
+      throws IOException {
+    checkPrincipalAndGetRealm(principal);
+    UserGroupInformation.setConfiguration(hadoopConf);
+
+    if (loginMode == LoginMode.RETURN_UGI) {
```
diqiu50: Use a switch statement to handle the enum and avoid invalid fallbacks.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/KerberosClient.java
diqiu50: Why we keep this KerberosClient

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,251 @@
...
+    if (checkIntervalSec <= 0) {
+      throw new IllegalArgumentException("The check interval must be positive");
+    }
+    if (threadNamePrefix == null || threadNamePrefix.trim().isEmpty()) {
+      throw new IllegalArgumentException("The thread name prefix can't be blank");
+    }
+
+    ScheduledThreadPoolExecutor executor =
+        new ScheduledThreadPoolExecutor(1, daemonThreadFactory(threadNamePrefix));
+    executor.scheduleAtFixedRate(
```
diqiu50, line 219: We should use a global thread pool to handle tick refreshes.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/KerberosClient.java
yuqi1129: Fixed in 699e38cfc. The Hive HMS `login()` path now relies on `KerberosAuthUtils.login()` for principal validation, and `loginProxyUser()` reuses `checkPrincipalAndGetRealm()` for realm parsing.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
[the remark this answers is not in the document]
yuqi1129: This was handled earlier in b59367232 and is still covered in 699e38cfc: `configureKrb5Conf()` rejects null/blank keys, and `KerberosClient.Builder.build()` rejects incomplete krb5.conf key pairs.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableHiveCatalog.java
yuqi1129: Fixed in 699e38cfc. The Iceberg keytab path now formats `catalogUUID` directly without `Long.parseLong(catalogUUID)`.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/KerberosClient.java
yuqi1129: Kept this class as the HMS-specific adapter. The shared `catalogs:hadoop-auth` client now handles common keytab login and TGT refresh, while this Hive Metastore client still owns proxy-user creation, HMS delegation-token retrieval, HiveClient wiring, and local keytab cleanup. I added class Javadoc in 699e38cfc to make that boundary explicit.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
yuqi1129: Fixed in 699e38cfc. `KerberosAuthUtils.login()` now uses an explicit `switch` over `LoginMode`, including a null check before the switch.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
yuqi1129: Fixed in 699e38cfc. Ticket refresh now uses a shared daemon scheduler in `KerberosAuthUtils`; callers keep only their own `ScheduledFuture` and cancel that task on close or repeated login.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -182,30 +188,35 @@ public static UserGroupInformation login(
       String principal, String keytabFilePath, Configuration hadoopConf, LoginMode loginMode)
       throws IOException {
     checkPrincipalAndGetRealm(principal);
-    UserGroupInformation.setConfiguration(hadoopConf);
-
-    if (loginMode == LoginMode.RETURN_UGI) {
-      return UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, keytabFilePath);
+    if (loginMode == null) {
+      throw new IllegalArgumentException("loginMode can't be null");
     }
```
diqiu50: Using Preconditions.checkArgument is better

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -239,6 +242,27 @@ private static boolean isHdfsUri(String keytabUri) {
...
+    ScheduledThreadPoolExecutor executor =
+        new ScheduledThreadPoolExecutor(1, daemonThreadFactory(TICKET_REFRESH_THREAD_NAME_PREFIX));
+    executor.setRemoveOnCancelPolicy(true);
+    return executor;
+  }
+
+  private static void refreshTicket(
+      UserGroupInformation loginUgi, String refreshTaskName, Logger log) {
+    Thread currentThread = Thread.currentThread();
+    String originalThreadName = currentThread.getName();
```
diqiu50: refreshTicket runs quickly, so switching the thread name is not practical.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/KerberosClient.java
diqiu50: Rename to HmsKerberosClient is better

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
yuqi1129: Do you mean that there is no need to switch thread name?

yuqi1129: changed.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
yuqi1129: changed.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
diqiu50: Yes

diqiu50 (dismissed):
LGTM

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,249 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.slf4j.Logger;
+
+/** Shared Kerberos helpers for Hadoop-aware Gravitino modules. */
+public final class KerberosAuthUtils {
+
+  private static final String TICKET_REFRESH_THREAD_NAME_PREFIX = "kerberos-ticket-refresh-";
+  private static final ScheduledThreadPoolExecutor TICKET_REFRESH_EXECUTOR =
+      newTicketRefreshExecutor();
```
jerryshao: **Regression: static single-threaded executor shared by all catalog instances**

`TICKET_REFRESH_EXECUTOR` has `corePoolSize=1` and is shared by every Kerberos-enabled catalog in the process. Before this PR each catalog had an isolated executor, so a slow or unreachable KDC only blocked that one catalog's refresh. Now, if any catalog's `checkTGTAndReloginFromKeytab()` blocks (e.g. a 30–90 s TCP timeout to the KDC), every other catalog's TGT renewal queues behind it on the same thread — potentially causing cascading auth failures across all catalogs simultaneously.

Consider either sizing the pool relative to the expected number of concurrent Kerberos catalogs, or keeping per-instance executors and using the shared class only for the refresh lambda.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -0,0 +1,249 @@
...
+
+  private static boolean isHdfsUri(String keytabUri) {
+    try {
+      URI uri = new URI(keytabUri.trim());
+      return "hdfs".equalsIgnoreCase(uri.getScheme());
+    } catch (URISyntaxException e) {
+      return keytabUri.trim().startsWith("hdfs");
+    }
+  }
+
```
jerryshao, line 232: **`TICKET_REFRESH_EXECUTOR` is never shut down**

`setRemoveOnCancelPolicy(true)` prevents queue memory growth when tasks are cancelled, but the backing daemon thread itself persists after all catalogs are closed. There is no shutdown hook or lifecycle method. In test JVMs, Gradle workers, or OSGi containers, the thread keeps static references alive and emits log errors for catalogs that no longer exist.

Consider `Runtime.getRuntime().addShutdownHook(new Thread(TICKET_REFRESH_EXECUTOR::shutdownNow))`, or a package-private `shutdown()` for test cleanup.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/HmsKerberosClient.java
```diff
@@ -164,38 +143,36 @@ public File saveKeyTabFileFromUri(String path) throws IOException {
...
-      if (checkTgtExecutor != null) {
-        checkTgtExecutor.shutdown();
-      }
+      cancelTicketRefreshTask();
 
       Files.deleteIfExists(Paths.get(keytabFilePath));
     } catch (IOException e) {
       LOG.warn("Failed to delete keytab file: {}", keytabFilePath, e);
     }
   }
```
jerryshao, line 163: **`cancel(true)` interrupts a running TGT renewal mid-flight**

The old `close()` called `checkTgtExecutor.shutdown()`, which lets an in-progress task run to completion. The new `cancelTicketRefreshTask()` calls `checkTgtRefreshTask.cancel(true)`, which immediately sets the thread interrupt flag. If `checkTGTAndReloginFromKeytab()` is interrupted after the old TGT is marked for replacement but before the new one is committed to the UGI Subject, the process-wide `loginUser` could briefly be left without valid credentials. HMS uses `LoginMode.LOGIN_USER` (process-wide UGI), so the impact is not limited to a single catalog.

Consider `cancel(false)` to let an in-flight relogin complete gracefully before teardown.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosClient.java
```diff
@@ -0,0 +1,184 @@
...
+          KerberosAuthUtils.startTicketRefresh(loginUser, checkIntervalSec, LOG);
+    }
+    return loginUser;
+  }
+
+  /** Returns the realm parsed from the principal, available after {@link #login(String)}. */
+  public String getRealm() {
+    return realm;
+  }
+
```
jerryshao, line 93: **`getRealm()` and `getLoginUser()` can return `null` before `login()` is called; missing `@Nullable`**

Both `realm` and `loginUser` are `null` until `login()` completes. `ClosableHiveCatalog.doKerberosOperations()` calls `kerberosClient.getRealm()` and feeds the result into a `String.format` to construct a proxy principal — producing `"user@null"` if there is any ordering issue (e.g. login failed silently or was skipped).

CLAUDE.md requires `@Nullable` on return types that can be null. Add `@Nullable` to both `getRealm()` and `getLoginUser()`, or add `Preconditions.checkState(realm != null, "login() has not been called")` guards.

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/HmsKerberosClient.java
```diff
@@ -164,38 +143,36 @@ public File saveKeyTabFileFromUri(String path) throws IOException {
...
 
     File keytabsDir = new File("keytabs");
     if (!keytabsDir.exists()) {
       keytabsDir.mkdir();
     }
     File keytabFile = new File(path);
     int fetchKeytabFileTimeout = kerberosConfig.getFetchTimeoutSec();
-    FileFetcher.get()
-        .fetchFileFromUri(keyTabUri, keytabFile, fetchKeytabFileTimeout * 1000, hadoopConf);
+    KerberosAuthUtils.fetchKeytabFromUri(
```
jerryshao, line 149: **Redundant `keytabsDir.mkdir()` — `fetchKeytabFromUri` already calls `parentFile.mkdirs()`**

`KerberosAuthUtils.fetchKeytabFromUri` calls `parentFile.mkdirs()` before invoking `FileFetcher`, so the three lines creating `keytabsDir` here are dead code. The manual `mkdir()` also uses the single-level form, which would silently fail if any parent of `keytabs/` was missing — the exact case that `fetchKeytabFromUri`'s `mkdirs()` already handles correctly.

These three lines can be removed.

jerryshao (commented):
This PR cleanly centralizes four near-identical Kerberos auth implementations into a shared `hadoop-auth` module. The `KerberosAuthUtils` / `KerberosClient` split is a good abstraction and the builder pattern is easy to extend. Five issues below, ranked by severity.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
yuqi1129: Modified. `startTicketRefresh` now creates a per-caller single-thread daemon scheduler and returns it.

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
yuqi1129: Modified. Each client owns its executor and calls shutdown() in close()

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/HmsKerberosClient.java
yuqi1129: Modified. replaced cancel(true) with executor.shutdown()

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosClient.java
yuqi1129: Done, added Preconditions.checkState(... != null, "login() has not been called")

catalogs/hive-metastore-common/src/main/java/org/apache/gravitino/hive/kerberos/HmsKerberosClient.java
yuqi1129: Done.  removed the 3 dead lines; fetchKeytabFromUri already does parentFile.mkdirs().

catalogs/hadoop-auth/src/main/java/org/apache/gravitino/catalog/hadoop/auth/KerberosAuthUtils.java
```diff
@@ -198,19 +197,28 @@ public static UserGroupInformation login(
...
-  public static ScheduledFuture<?> startTicketRefresh(
+  public static ScheduledExecutorService startTicketRefresh(
       UserGroupInformation loginUgi, int checkIntervalSec, Logger log) {
     Preconditions.checkArgument(checkIntervalSec > 0, "The check interval must be positive");
 
-    return TICKET_REFRESH_EXECUTOR.scheduleAtFixedRate(
+    ScheduledExecutorService executor =
+        Executors.newSingleThreadScheduledExecutor(
+            daemonThreadFactory(TICKET_REFRESH_THREAD_NAME_PREFIX));
+    executor.scheduleAtFixedRate(
```
diqiu50, line 219: If you prefer to keep one Kerberos client per executor, I’d suggest moving the executor creation and shutdown logic into HmsKerberosClient. In general, it’s better to manage resource creation and cleanup in a single place.

yuqi1129: I'd like to keep the current code, because:

1. Each catalog already has its own HmsKerberosClient, and the executor is a per-instance field. It's created in login() and shut down in close(), both inside the client. Nothing is shared across catalogs, so there's no leak.
2. Only the construction (newSingleThreadScheduledExecutor + scheduleAtFixedRate) is in KerberosAuthUtils.startTicketRefresh, because it's shared by both KerberosClient and HmsKerberosClient. Moving it into HmsKerberosClient would duplicate this code in both clients.

So I prefer not to inline it. Let me know if you see a real lifecycle problem.

yuqi1129: @diqiu50

diqiu50: That's OK

diqiu50 (approved):

PR #11778, opened 2026-06-23
[#11775] improvement(authz): Short-circuit list authorization via parent-scope grant

### LINKED ISSUE #11775: [Improvement] List tables with Authorization enabled has N+1 performance issue

When Authorization is enabled, the "List tables" operation on Iceberg REST service exhibits severe performance degradation due to per-table permission checks.

**Performance test results** (30 concurrent users, schema `load3000` with 10,496 tables):

| Configuration | Samples | RPS | P50 | P95 |
|---|---|---|---|---|
| Auth:OFF, Cache:ON, Cred:ON | 420 | 14.03 | 3,033ms | 4,001ms |
| Auth:ON, Cache:ON, Cred:ON | 5 | 0.20 | 134,000ms | 136,000ms |

Authorization causes a **~70x performance drop** for list operations. In contrast, single-table "Load table" operations show only a ~2% RPS decrease with Authorization enabled, confirming the issue is specific to list operations that return many results.

The root cause appears to be that the authorization layer checks permissions on each table individually when filtering list results, resulting in ~10,496 permission checks per single list request.

### Test Environment Background

- **Metalake:** `irc_test`
- **Catalog:** `iceberg` (Iceberg REST service)
- **Schema:** `load3000` containing **10,496 tables**
- **Concurrent users:** 30 (user1 - user30)
- **Role setup:** Each user is assigned an individual role (`role1` - `role30`) plus a shared `role_all`. The `role_all` role provides full access to the metalake (matching all per-user role permissions combined).
- **Role privileges:** Roles are created with empty `securableObjects` at creation time, then granted metalake-level access. The `role_all` role effectively gives full access, so every table passes the authorization check — the bottleneck is the **cost of checking**, not the filtering result.
- **List table count:** Each list request returns all **10,496 tables** in the schema (no server-side pagination), and authorization is evaluated per table in the result set.

### How should we improve?

Potential solutions:
1. **Batch authorization check** — Check permissions for multiple resources in a single call rather than one-by-one.
2. **Schema-level short circuit** — If a user/group has access to the entire schema (or parent catalog/metalake), skip per-table authorization checks for list operations.
3. **Authorization result caching** — Cache authorization decisions for list-filter operations so repeated checks on the same scope are fast.
4. **Pagination with lazy auth** — Only check permissions on the page of results being returned, not the entire table list.

yuqi1129:

When authorization is enabled, list operations evaluated one per-object permission check per result, producing an N+1 explosion. This PR evaluates the **ancestor-only portion** of the filter expression **once** per list request: if a privilege is granted (or the object is owned) at a parent scope (metalake/catalog/schema), every child is visible unless an object-level deny overrides it.

- New parent-scope expressions in `AuthorizationExpressionConstants` for `TABLE`/`SCHEMA`/`CATALOG`.
- `MetadataAuthzHelper.filterByExpression(…, NameIdentifier[])` attempts the short-circuit before the per-object loop. One insertion point covers both Gravitino REST (`listTables`/`listSchemas`/`listCatalogs`) and Iceberg REST (`listTable`/`listNamespace`), since they all route through this helper.
- New SPI method `GravitinoAuthorizer#hasDenyPolicyOnType` (conservative default `true`); `JcasbinAuthorizer` implements it as an in-memory scan of the current user's deny-enforcer policies. The short-circuit is taken only when no object-level deny may exist for the relevant type/privileges, otherwise it falls back to the existing per-object filtering.

Per #11775, with authorization on, listing a schema of ~10,496 tables took ~134s (vs ~4s with auth off) — a ~70x regression — because permission was checked once per table. Owner/grant at a parent scope makes every child visible (OR semantics); only an object-level deny can subtract, so the ancestor part is a loop-invariant that can be lifted out of the per-object loop. Ancestor-level denies are already handled inside the parent-scope expression's `!ANY(DENY_..., METALAKE, CATALOG, SCHEMA)` terms.

Fix: #11775

No. The new method on the `GravitinoAuthorizer` SPI is a backward-compatible default method (default returns `true`, i.e. no short-circuit), and authorization decisions are unchanged.

New unit tests:
- `TestMetadataAuthzHelper`: parent grant + no deny returns the whole list; parent grant + possible deny falls back and excludes the denied table; schema list short-circuits via a catalog-level grant; catalog list short-circuits via a metalake-level grant.
- `TestJcasbinAuthorizer#testHasDenyPolicyOnType`: detects an object-level deny for the matching type/privilege and ignores non-matching type/privilege.

Full `:server-common` test suite passes; `:server`, `:iceberg:iceberg-rest-server`, and `:core` compile clean.

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
@@ -166,6 +166,44 @@ public class AuthorizationExpressionConstants {
                   ANY_MODIFY_TABLE
                   """;
 
+  /**
+   * Parent-scope expression used to short-circuit table-list authorization. It is the ancestor-only
+   * portion of {@link #FILTER_TABLE_AUTHORIZATION_EXPRESSION}: ownership of, or a SELECT/MODIFY
+   * grant at, the metalake/catalog/schema scope means every table in the schema is visible unless a
+   * table-level deny overrides it. Evaluated once per list request instead of once per table.
+   */
+  public static final String TABLE_LIST_PARENT_SCOPE_AUTHORIZATION_EXPRESSION =
+      """
+              ANY(OWNER, METALAKE, CATALOG, SCHEMA) ||
+              (ANY(SELECT_TABLE, METALAKE, CATALOG, SCHEMA)
+                  && !ANY(DENY_SELECT_TABLE, METALAKE, CATALOG, SCHEMA)) ||
+              (ANY(MODIFY_TABLE, METALAKE, CATALOG, SCHEMA)
+                  && !ANY(DENY_MODIFY_TABLE, METALAKE, CATALOG, SCHEMA))
+              """;
+
+  /**
+   * Parent-scope expression used to short-circuit schema-list authorization. Ancestor-only portion
+   * of {@link #FILTER_SCHEMA_AUTHORIZATION_EXPRESSION}: ownership of, or a USE_SCHEMA grant at, the
+   * metalake/catalog scope means every schema in the catalog is visible unless a schema-level deny
+   * overrides it.
+   */
+  public static final String SCHEMA_LIST_PARENT_SCOPE_AUTHORIZATION_EXPRESSION =
+      """
+              ANY(OWNER, METALAKE, CATALOG) ||
+              (ANY(USE_SCHEMA, METALAKE, CATALOG) && !ANY(DENY_USE_SCHEMA, METALAKE, CATALOG))
+              """;
+
+  /**
+   * Parent-scope expression used to short-circuit catalog-list authorization. Ancestor-only portion
+   * of {@link #LOAD_CATALOG_AUTHORIZATION_EXPRESSION}: metalake ownership, or a USE_CATALOG grant
+   * at the metalake scope, means every catalog is visible unless a catalog-level deny overrides it.
+   */
+  public static final String CATALOG_LIST_PARENT_SCOPE_AUTHORIZATION_EXPRESSION =
+      """
+              METALAKE::OWNER ||
+              (ANY(USE_CATALOG, METALAKE) && !ANY(DENY_USE_CATALOG, METALAKE))
+              """;
+
   public static final String FILTER_VIEW_AUTHORIZATION_EXPRESSION =
       """
                   ANY(OWNER, METALAKE, CATALOG, SCHEMA, VIEW) ||
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
@@ -169,39 +169,43 @@ public class AuthorizationExpressionConstants {
   /**
    * Parent-scope expression used to short-circuit table-list authorization. It is the ancestor-only
    * portion of {@link #FILTER_TABLE_AUTHORIZATION_EXPRESSION}: ownership of, or a SELECT/MODIFY
-   * grant at, the metalake/catalog/schema scope means every table in the schema is visible unless a
-   * table-level deny overrides it. Evaluated once per list request instead of once per table.
+   * grant at, the metalake/catalog/schema scope means every table in the schema is visible.
+   * Evaluated once per list request instead of once per table.
+   *
+   * <p>This expression only checks the positive parent-scope grant; deny coverage (at any scope) is
+   * handled separately by {@link
+   * org.apache.gravitino.authorization.GravitinoAuthorizer#hasDenyPolicy} so the short-circuit
+   * falls back to per-object authorization whenever a relevant deny may exist.
    */
   public static final String TABLE_LIST_PARENT_SCOPE_AUTHORIZATION_EXPRESSION =
       """
               ANY(OWNER, METALAKE, CATALOG, SCHEMA) ||
-              (ANY(SELECT_TABLE, METALAKE, CATALOG, SCHEMA)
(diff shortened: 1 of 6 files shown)


jerryshao:
@roryqi please help to review.

roryqi:
> @roryqi please help to review.

I will.

server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
```diff
@@ -166,6 +166,44 @@ public class AuthorizationExpressionConstants {
...
+  /**
+   * Parent-scope expression used to short-circuit schema-list authorization. Ancestor-only portion
+   * of {@link #FILTER_SCHEMA_AUTHORIZATION_EXPRESSION}: ownership of, or a USE_SCHEMA grant at, the
+   * metalake/catalog scope means every schema in the catalog is visible unless a schema-level deny
+   * overrides it.
+   */
+  public static final String SCHEMA_LIST_PARENT_SCOPE_AUTHORIZATION_EXPRESSION =
+      """
+              ANY(OWNER, METALAKE, CATALOG) ||
+              (ANY(USE_SCHEMA, METALAKE, CATALOG) && !ANY(DENY_USE_SCHEMA, METALAKE, CATALOG))
```
roryqi: Our code will rewrite the ANY expressions. So we shouldn't add deny expression here.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
```diff
@@ -363,6 +363,41 @@ public boolean isMetalakeUser(String metalake, AuthorizationRequestContext reque
...
+    // below sees every policy the user can carry. Idempotent within a request.
+    loadRolePrivilege(metalake, principal.getName(), userId, userInfo, requestContext);
+
+    Set<String> privilegeNames = privileges.stream().map(Enum::name).collect(Collectors.toSet());
+    String typeName = type.name();
+    String userIdStr = String.valueOf(userId);
+    // p-row layout: [sub(roleId), metadataType, metadataId, act(privilege), eft]. Filtering by
+    // (roleId, metadataType) keeps the scan bounded by the user's role/policy count, never by the
+    // number of listed objects.
+    for (String roleId : denyEnforcer.getRolesForUser(userIdStr)) {
```
roryqi, line 403: Why don't u reuse the enforce logic here?

yuqi1129: The existing `enforce` path answers "is *this specific* object id authorized?" — it needs a concrete `metadataId`. Here we need a different question: "does the user hold *any* deny policy for these privileges, on any object?" There's no single id, so reusing `enforce` would force iterating every listed object, which defeats the short-circuit (it's meant to be O(roles), not O(objects)).

I've kept the existence-query semantics but routed it through named policy-field constants (`POLICY_SUBJECT_FIELD_INDEX` / `POLICY_ACTION_FIELD_INDEX`) instead of the positional `policy.get(3)` / `size() > 3`.

server-common/src/main/java/org/apache/gravitino/server/authorization/expression/AuthorizationExpressionConstants.java
yuqi1129: Good point — the converter already injects the deny half for the `ANY_*` macros. I've dropped the hand-written `!ANY(DENY_*, ...)` from the parent-scope expressions and kept only the positive parent grant.

All deny coverage (parent **and** object scope) now lives in `hasDenyPolicy` (renamed from `hasDenyPolicyOnType`, type param dropped), which scans the user's deny policies by privilege regardless of scope — so a deny at any level forces the per-object fallback.

server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/JcasbinAuthorizer.java
roryqi: I have a little concern about this logic. This isn't easy to understand.

yuqi1129: I have added more comments and tests to make it more readable. If you still have questions about it, feel free to left you problem here.

roryqi: Could we use expression evaluation here? Maybe we can use property expression.

yuqi1129: I looked into this, but I think expression evaluation does not fit here.

The expression evaluator and the enforce path both match on a concrete metadataId (the matcher is r.metadataId == p.metadataId). They answer "is this object allowed?".

But here the question is different: "does the user have any deny for these privileges, on any object?". There is no single object id to pass in. So to use expression evaluation, I would have to either:

  1. run it once per listed object (back to O(objects), which removes the whole point of the short-circuit), or
  2. add a new "no object id" evaluation path that only matches by privilege — which is basically the same scan as now, just hidden behind the expression layer.

So I kept the direct scan. It stays O(roles), not O(objects). To make it easier to read I used named field constants (POLICY_SUBJECT_FIELD_INDEX /
  POLICY_ACTION_FIELD_INDEX), added comments explaining the policy row layout, and added tests (including a group-inherited deny case).

If you still prefer to wrap it in an expression primitive, I'm open to it, but I think it adds more code without changing the core logic. Let me know what you think.

roryqi: If we don't have better solution, it is ok for me.

roryqi (approved):
LGTM.

PR #11783, opened 2026-06-24
[#11782] feat(lance): support Hive Metastore (hive2) namespace backend for Lance REST server

### LINKED ISSUE #11782: [FEATURE] Support Hive Metastore namespace backend for Lance REST server
### Describe the feature

Add a `hive` namespace backend to the Lance REST server so Lance tables can be managed in a Hive Metastore (HMS), the same way the Iceberg REST server already supports HMS.

A namespace would map to a Hive database (1 level) and a table to `db.table` (2 levels). Lance tables are registered as `EXTERNAL_TABLE`s carrying `table_type=lance`. The backend acts purely as a pointer registry (`managedVersioning=false`) — it never writes Lance data — which matches the existing Gravitino backend and the Lance namespace reference implementations.

### Motivation

Today Lance tables can only be cataloged through the Gravitino backend. Many deployments already operate a Hive Metastore as their central metadata service; a Hive backend lets them register and discover Lance tables there without standing up additional infrastructure.

### Describe the solution

Register a new `HIVE` value in the existing `LanceNamespaceBackend` enum plug-point (no REST-server wiring changes), backed by a `HiveLanceNamespaceWrapper` that talks to HMS through a pooled `IMetaStoreClient` (compiled against Hive 2.3.9). Add config keys `hive-metastore-uris`, `hive-warehouse`, and `hive-client-pool-size`.

### Additional context

This is the first of two steps; a follow-up will swap in an mTLS Hive client at the `HiveClientPool.newClient()` seam.

puchengy:

Adds a `hive2` namespace backend to the Lance REST server so Lance tables can be managed in a Hive Metastore (HMS), mirroring how the Iceberg REST server supports HMS.

The backend type is named `hive2` (rather than `hive`) after the `lance-namespace-hive2` implementation it wraps, since Lance also ships a separate `lance-namespace-hive3` implementation — this leaves room to add a `hive3` backend later without a confusing rename. The `hive-*` config keys are shared Hive Metastore settings (not tied to the backend name), so a future `hive3` backend can reuse them as-is.

- Registers `HIVE2("hive2", HiveLanceNamespaceWrapper.class)` in the existing `LanceNamespaceBackend` enum plug-point — no REST-server wiring changes needed.
- The backend is built on the official **`org.lance:lance-namespace-hive2`** implementation (`Hive2Namespace`, which implements `LanceNamespace`). Thin adapter classes (`HiveLanceNamespaceOperations`, `HiveLanceTableOperations`) translate Gravitino's flat-parameter operation interfaces onto the Lance Namespace request/response model; `HiveLanceNamespaceWrapper` builds the Hadoop `Configuration` from `LanceConfig` and owns the delegate.
- Namespace model: a namespace is a Hive database (1 level); a table is `db.table` (2 levels), registered as an `EXTERNAL_TABLE` with `table_type=lance`. It is a pointer registry (`managedVersioning=false`) — no Lance data is written by the catalog.
- New config keys on `LanceConfig`: `hive-metastore-uris`, `hive-warehouse`, `hive-client-pool-size` (default 3).
- `Hive2Namespace` implements `Closeable` (as of `lance-namespace-hive2` 0.4.1); `HiveLanceNamespaceWrapper#close()` closes the delegate so the pooled Hive metastore connections are released on shutdown instead of leaked (covered by `TestHiveLanceNamespaceWrapperClose`).
- `describeTable` always requests the lightweight pointer-only describe — the HMS backend stores only the table pointer (location + properties); the schema/stats live in the Lance dataset on storage.

**Dependency handling.** The backend uses **`lance-namespace-hive2:0.4.1`**, which is built against `lance-namespace 0.7.7` and pulls Hadoop 2.8.5 + an embedded-metastore stack. The build:
- excludes its Hadoop 2.8.5 / Derby / DataNucleus so the module runs on Gravitino's Hadoop 3 metastore client;
- excludes the transitive `lance-namespace-core`/`apache-client` 0.7.7 so **0.7.5** (the version Gravitino standardizes on, declared via `libs.lance.namespace.core`) remains the single Lance Namespace version on the classpath (verified on `runtimeClasspath`), and pins Arrow to 18.x;
- drops legacy Netty fat-jars (`netty-all` 4.0.x / `netty` 3.x) that otherwise shadow `netty-buffer` 4.1.x and break `arrow-memory-netty` initialization on JDK 17;
- drops the legacy servlet stack that `hive-common` drags in transitively (`jetty-all` 7.6.0 / mortbay / orbit Jetty, `javax.servlet:servlet-api:2.4`, `jsp-api:2.0`) from both `lance-common` and `lance-rest-server` — packaged `libs/` jars load alphabetically, so these otherwise shadow Gravitino's Jetty 9.4 / Servlet 3.1 and crash the standalone server (`IncompatibleClassChangeError` at startup, or `NoSuchMethodError: HttpServletRequest.getDispatcherType()` on the first request). Guarded by `TestHiveClasspathHygiene`.

Lance tables currently can only be cataloged through the Gravitino backend. Many deployments already run a Hive Metastore as their central metadata service; this lets them register and discover Lance tables there without standing up extra infrastructure. Building on the official `lance-namespace-hive2` keeps behavior aligned with the upstream Lance ecosystem.

Fix: #11782

Yes — new property keys:
- `gravitino.lance-rest.namespace-backend=hive2`
- `gravitino.lance-rest.hive-metastore-uris`
- `gravitino.lance-rest.hive-warehouse`
- `gravitino.lance-rest.hive-client-pool-size`

New dependency version entry: `lance-namespace-impls = 0.4.1` (`lance-namespace` stays at 0.7.5).

Note: with the Hive backend, table creation is performed via **`declareTable`** (a metastore pointer). `createTable`, `registerTable`, and column alteration are reported as unsupported by the underlying `lance-namespace-hive2` implementation, and `describeTable` returns the pointer-only metadata (the schema lives in the Lance dataset on storage, not the catalog).

- Unit tests (mocked `Hive2Namespace`) covering the namespace/table operation adapters, identifier parsing, config/backend wiring, wrapper close behavior, and classpath hygiene: `:lance:lance-common:test` and `:lance:lance-rest-server:test` (with `-PskipITs`) — all pass on latest `main`.
- A Docker-gated integration test `LanceRESTHiveCatalogIT` (`@Tag("gravitino-docker-test")`) that launches the embedded Gravitino server with the Lance REST aux-service on the `hive` backend against a live Hive Metastore container and exercises the namespace/table lifecycle. (Requires Docker; run with `-PskipDockerTests=false`.)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -33,6 +33,7 @@ log4j = "2.25.4"
 arrow = "18.3.0"
 lance = "6.0.0"
 lance-namespace = "0.7.5"
+lance-namespace-impls = "0.4.1"
 delta-kernel = "3.3.0"
 jetty = "9.4.58.v20250814"
 jersey = "2.41"
@@ -205,6 +206,7 @@ log4j-layout-template-json = { group = "org.apache.logging.log4j", name = "log4j
 arrow-vector = { group = "org.apache.arrow", name = "arrow-vector", version.ref = "arrow" }
 lance = { group = "org.lance", name = "lance-core", version.ref = "lance" }
 lance-namespace-core = { group = "org.lance", name = "lance-namespace-core", version.ref = "lance-namespace" }
+lance-namespace-hive2 = { group = "org.lance", name = "lance-namespace-hive2", version.ref = "lance-namespace-impls" }
 delta-kernel = { group = "io.delta", name = "delta-kernel-api", version.ref = "delta-kernel" }
 delta-kernel-defaults = { group = "io.delta", name = "delta-kernel-defaults", version.ref = "delta-kernel" }
 jakarta-validation-api = { group = "jakarta.validation", name = "jakarta.validation-api", version.ref = "jakarta-validation" }
--- a/lance/lance-common/build.gradle.kts
+++ b/lance/lance-common/build.gradle.kts
@@ -24,7 +24,50 @@ plugins {
   id("idea")
 }
 
+// Pin the Lance Namespace API to the version Gravitino standardizes on (0.7.5). lance-namespace-hive2
+// 0.4.1 is built against lance-namespace 0.7.7 and drags its 0.7.7 core/apache-client in
+// transitively; those are excluded from the hive2 dependency below so the whole module resolves to
+// the single, consistent 0.7.5 Lance Namespace version Gravitino declares.
+configurations.all {
+  // Legacy Netty fat-jars (netty-all 4.0.x, netty 3.x) arrive transitively via Hadoop/Hive/
+  // lance-core and shadow the modern netty-buffer 4.1.x that Arrow's allocation manager requires,
+  // breaking arrow-memory-netty initialization on JDK 17. Arrow uses netty-buffer/netty-common
+  // directly, so drop the fat-jars everywhere in this module.
+  exclude(group = "io.netty", module = "netty-all")
+  exclude(group = "io.netty", module = "netty")
+  // Hive (hive-common 2.3.9, pulled transitively by lance-namespace-hive2) drags in an ancient
+  // Jetty (org.eclipse.jetty.aggregate:jetty-all 7.6.0, where Container is a *class*) plus
+  // mortbay/orbit Jetty. These land in the packaged libs/, and because the distribution globs
+  // jars alphabetically, jetty-all-7.6.0 loads before Gravitino's Jetty 9.4.x at runtime ->
+  // "class org.eclipse.jetty.server.Connector can not implement ...Container" -
+  // IncompatibleClassChangeError that crashes the standalone server at startup. The Hive metastore
+  // client needs no servlet container, so drop all legacy Jetty everywhere in this module.
+  exclude(group = "org.eclipse.jetty.aggregate")
+  exclude(group = "org.eclipse.jetty.orbit")
+  exclude(group = "org.mortbay.jetty")
+  // Hive/Hadoop also drag in the ancient Servlet 2.4 / JSP 2.0 APIs (javax.servlet:servlet-api:2.4,
+  // javax.servlet:jsp-api:2.0). These land in libs/ and, loaded alphabetically before the modern
+  // javax.servlet:javax.servlet-api:3.1.0 that Jetty 9.4 requires, shadow it at runtime ->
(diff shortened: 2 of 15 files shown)


puchengy:
@Jackeyzhe @jerryshao Hi both, I have a few open questions:
1. we should make use of https://github.com/lance-format/lance-namespace-impls/tree/main/java/lance-namespace-hive2 instead of rebuilding lance hive table operations classes? hive2 depends on hadoop2 and is conflicting with gravitino common's hadoop3 dependency.
2. How do we want to support hive 3/ 4, do we want a different profile or do we want to use hive2 for hive3/4 client? Some basic reasearch shows that hive2 usage here will be compatible with hive3/4.
Appreciate your response, thanks

yuqi1129:
@puchengy 
Would you like to push forward with this one as discussed offline? 

> we should make use of [lance-format/lance-namespace-impls@main/java/lance-namespace-hive2](https://github.com/lance-format/lance-namespace-impls/tree/main/java/lance-namespace-hive2?rgh-link-date=2026-06-24T04%3A42%3A33Z) instead of rebuilding lance hive table operations classes? hive2 depends on hadoop2 and is conflicting with gravitino common's hadoop3 dependency

Since Gravitino has already defined a common backend API like `NamespaceWrapper`, I believe it's okay to rebuild Lance hive table operations classes. For Hive2, Gravitino depends on Hive2 and Hive3 at the same time, this should not be a big problem.


> How do we want to support hive 3/ 4, do we want a different profile or do we want to use hive2 for hive3/4 client? Some basic reasearch shows that hive2 usage here will be compatible with hive3/4.

The API should be generally compatible with most API for Hive2/Hive3/Hive4, we can resolve them case by case if we did plan to support those versions.

puchengy:
Reopened with the branch rebased onto latest `main` and squashed into a single commit. Compared with the earlier revision, this now also releases the pooled Hive metastore connections on shutdown (`Hive2Namespace` is `Closeable` in `lance-namespace-hive2` 0.4.0), always uses the pointer-only `describeTable`, and excludes the legacy Jetty/servlet jars that `hive-common` drags in so the standalone server starts cleanly. Unit tests pass on latest `main`.

puchengy:
@yuqi1129 @romanhorilyi @lixmgl @EDCRLF PTAL

PR #11785, opened 2026-06-24
[#11784] improvement(common): harden remote file fetch against SSRF DNS-rebinding

### LINKED ISSUE #11784: [Improvement] Harden server-side remote file fetch against SSRF DNS-rebinding

`FileFetcher` validates a remote URI's host against an SSRF denylist before downloading keytabs/jars, but the download step then re-resolves the hostname when it opens the connection. The DNS answer can change between the two steps (DNS rebinding / TOCTOU), so a host that passed validation can still connect to an internal or cloud-metadata address. A few related gaps make it worse:

- the denylist misses some address forms (0.0.0.0/8, CGNAT 100.64/10, the Alibaba/Oracle metadata IPs, broadcast, IPv6 unique-local, and IPv4 addresses embedded in IPv6 literals);
- FTP's PASV data channel connects to an address the server chooses, which can't be pinned;
- credential-bearing URIs (userinfo, presigned query tokens) are logged verbatim.

### How should we improve?

- Resolve the host once, validate every resolved address, and pin the download to the validated address so the hostname is never looked up again.
- Extend the denylist to the forms above, including IPv4-compatible/mapped/NAT64/6to4/ISATAP embeddings.
- Reject FTP when unsafe-URI blocking is on (the default); operators can opt out for trusted sources.
- Strip userinfo and query strings from URIs before they reach logs or error messages.

LuciferYang:

`FileFetcher` already checked a remote URI's host against an SSRF denylist, but the download step then let the JDK re-resolve the hostname when it connected. That leaves a DNS-rebinding window: the host can look safe during validation and resolve to an internal address when the download actually connects.

The fix resolves the host once in `RemoteUriValidator`, validates every address it resolves to, and returns one for the caller to pin. `RemoteFileDownloader` is a small HTTP/1.0 client that connects straight to that pinned `InetAddress` and never looks the hostname up again, while still using the original hostname for TLS/SNI verification. It follows no redirects and accepts only Content-Length or chunked bodies, so a truncated download can't be mistaken for a complete one.

Along the way it also:

- widens the denylist to 0.0.0.0/8, CGNAT 100.64/10 (covers the Alibaba metadata IP), the Oracle metadata IP, broadcast, IPv6 unique-local, and the IPv4-compatible/mapped/NAT64/6to4/ISATAP forms that hide an IPv4 in an IPv6 literal;
- rejects `ftp://` when blocking is on, because its PASV data channel connects to an address the server picks and can't be pinned;
- adds `SafeUri.redact` to strip userinfo and query strings from URIs before they reach logs or error messages.

The server downloads keytabs and jars from operator-supplied URIs. Without pinning, someone who controls DNS for one of those hosts can pass validation and still steer the connection to a cloud metadata endpoint or an internal service. The extra denylist entries and the FTP restriction remove the other ways to reach those addresses.

Fix: #11784

No new config or API. One behavior change: with `gravitino.fetchFile.blockUnsafeRemoteUri` enabled (the default), `ftp://` URIs are now rejected. Set it to false to allow them from a trusted source.

Added 48 unit tests across four classes covering address classification (including the IPv6-embedded forms), address pinning, redirect/unframed/oversized/truncated responses, chunked decoding, FTP rejection, and URI redaction. `./gradlew :common:test` and `:common:spotlessCheck` pass.

### DIFF
--- a/common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
+++ b/common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
@@ -1,0 +1,377 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.utils;
+
+import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
+/**
+ * Downloads an {@code http} or {@code https} file while pinning the TCP connection to a
+ * pre-validated address.
+ *
+ * <p>This closes the DNS-rebinding TOCTOU window between SSRF validation and the actual fetch: the
+ * host is resolved and validated once by {@link RemoteUriValidator#resolveAndValidate}, and the
+ * resulting {@link InetAddress} is connected to directly here, so the hostname is never
+ * re-resolved. A minimal HTTP/1.0 client is used so the connection can be pinned at the socket
+ * level (the JDK exposes no per-connection address override for plain {@code HttpURLConnection},
+ * and silently drops a {@code Host} header set via {@code setRequestProperty}). The original
+ * hostname is still used for the HTTP {@code Host} header and, for {@code https}, for TLS SNI and
+ * certificate-hostname verification, so virtual hosting and certificate validation keep working.
+ * Redirects are not followed, because a redirect target would re-resolve to an unvalidated address.
+ *
+ * <p>{@code ftp} is intentionally not handled here: the FTP data channel is opened to an address
+ * the server chooses in its {@code PASV}/{@code EPSV} reply, which cannot be pinned, so {@link
+ * FileFetcher} rejects {@code ftp} on the SSRF-blocking path instead.
+ *
+ * <p>Only responses framed by {@code Content-Length} or chunked transfer-encoding are accepted; a
+ * connection-close-delimited body is rejected because a premature close cannot be told apart from a
+ * complete one. Combined with streaming to a sibling temporary file that is atomically moved into
+ * place only on success, this means a connection that drops mid-transfer never leaves a truncated
+ * destination behind. Response and header sizes are bounded to prevent a malicious pinned host from
+ * exhausting disk or memory.
+ */
+final class RemoteFileDownloader {
+
+  private static final int DEFAULT_HTTP_PORT = 80;
+  private static final int DEFAULT_HTTPS_PORT = 443;
+  private static final int MAX_HEADER_BYTES = 64 * 1024;
+  private static final long MAX_BODY_BYTES = 2L * 1024 * 1024 * 1024;
+  private static final int COPY_BUFFER_BYTES = 8192;
+
+  private RemoteFileDownloader() {}
+
+  /**
+   * Downloads {@code uri} into {@code destFile}, connecting only to {@code pinnedAddress}.
+   *
+   * @param uri the source URI; must use the {@code http} or {@code https} scheme
+   * @param pinnedAddress the validated address to connect to
+   * @param destFile the local destination file
+   * @param timeoutMs the connect/read timeout in milliseconds
+   * @throws IOException if the file cannot be fetched or the response is not successful
+   */
+  static void download(URI uri, InetAddress pinnedAddress, File destFile, int timeoutMs)
+      throws IOException {
+    String scheme = Optional.ofNullable(uri.getScheme()).orElse("").toLowerCase(Locale.ROOT);
+    if (!scheme.equals("http") && !scheme.equals("https")) {
+      throw new IllegalArgumentException(
+          String.format("Pinned download does not support scheme '%s'", scheme));
+    }
+
+    Path destPath = destFile.toPath().toAbsolutePath();
+    Path tempPath =
+        Files.createTempFile(destPath.getParent(), destPath.getFileName() + ".", ".tmp");
+    try {
+      httpDownload(uri, pinnedAddress, tempPath, timeoutMs, scheme.equals("https"));
+      Files.move(tempPath, destPath, StandardCopyOption.REPLACE_EXISTING);
+    } catch (Throwable e) {
+      // Catch every throwable (incl. Error/OutOfMemoryError) so the temp file is never orphaned in
+      // the staging directory. Precise rethrow keeps the declared `throws IOException`.
+      try {
+        Files.deleteIfExists(tempPath);
+      } catch (IOException suppressed) {
+        e.addSuppressed(suppressed);
+      }
+      throw e;
+    }
+  }
+
+  private static void httpDownload(
+      URI uri, InetAddress pinnedAddress, Path destPath, int timeoutMs, boolean tls)
+      throws IOException {
+    String host = uri.getHost();
+    int port = uri.getPort() != -1 ? uri.getPort() : (tls ? DEFAULT_HTTPS_PORT : DEFAULT_HTTP_PORT);
+
+    Socket socket = new Socket();
+    try {
+      socket.connect(new InetSocketAddress(pinnedAddress, port), timeoutMs);
+      socket.setSoTimeout(timeoutMs);
+      if (tls) {
+        socket = startTls(socket, host, port);
+      }
+
+      sendGetRequest(socket.getOutputStream(), requestTarget(uri), hostHeader(host, uri.getPort()));
+
+      InputStream in = new BufferedInputStream(socket.getInputStream());
+      int status = readStatusCode(in);
+      if (status < 200 || status >= 300) {
+        throw new IOException(
+            String.format(
+                "Unexpected HTTP status %d fetching pinned URI '%s'", status, SafeUri.redact(uri)));
+      }
+      Map<String, String> headers = readHeaders(in);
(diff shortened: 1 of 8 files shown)


LuciferYang:
cc @yuqi1129 @roryqi FYI

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
```diff
@@ -0,0 +1,377 @@
...
+      String value = line.substring(colon + 1).trim();
+      headers.put(name, value);
+    }
+    throw new IOException("Unexpected end of stream while reading HTTP headers from pinned host");
+  }
+
+  private static void writeBody(InputStream in, Map<String, String> headers, OutputStream out)
+      throws IOException {
+    String transferEncoding = headers.get("transfer-encoding");
+    if (transferEncoding != null && transferEncoding.toLowerCase(Locale.ROOT).contains("chunked")) {
```
yuqi1129: `contains("chunked")` also matches multi-coding values like `Transfer-Encoding: gzip, chunked`, which then route to `copyChunked`. That de-chunks but does **not** gunzip, so the still-compressed bytes get written as the final file — a silently corrupt download. `Transfer-Encoding: xchunked` would mis-dispatch on the substring too. Suggest requiring the final (or sole) transfer-coding token to be exactly `chunked` and rejecting any other coding.

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
```diff
@@ -0,0 +1,377 @@
...
+  private static void httpDownload(
+      URI uri, InetAddress pinnedAddress, Path destPath, int timeoutMs, boolean tls)
+      throws IOException {
+    String host = uri.getHost();
+    int port = uri.getPort() != -1 ? uri.getPort() : (tls ? DEFAULT_HTTPS_PORT : DEFAULT_HTTP_PORT);
+
+    Socket socket = new Socket();
+    try {
+      socket.connect(new InetSocketAddress(pinnedAddress, port), timeoutMs);
+      socket.setSoTimeout(timeoutMs);
```
yuqi1129, line 122: `soTimeout` bounds each individual `read()`, not the whole transfer. A pinned host that drips one byte just inside each timeout window can keep the socket + temp file + worker thread alive for a body up to `MAX_BODY_BYTES` (2 GiB) — a slow-loris-style resource exhaustion. Not a regression (`FileUtils.copyURLToFile` had the same property), but since this is now a bespoke client an overall wall-clock deadline would be cheap insurance.

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
```diff
@@ -0,0 +1,377 @@
...
+    sslSocket.startHandshake();
+    return sslSocket;
+  }
+
+  private static void sendGetRequest(OutputStream out, String target, String hostHeader)
+      throws IOException {
+    // HTTP/1.0 with "Connection: close" requests a non-persistent connection. The response must be
+    // framed by Content-Length or chunked encoding; an unframed (close-delimited) body is rejected
+    // by writeBody because its completeness cannot be verified.
+    String request =
```
yuqi1129, line 167: With an HTTP/1.0 request a compliant origin may legally frame the body by connection close. `writeBody` (~L240) rejects close-delimited bodies, so such servers now hard-fail. This is intentional for truncation safety, but worth confirming the keytab/jar sources always send `Content-Length` (most static-file servers do); otherwise consider HTTP/1.1 to bias servers toward `Content-Length`.

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+    try {
+      return Integer.parseInt(parts[1]);
+    } catch (NumberFormatException e) {
+      throw new IOException("Malformed HTTP status line: " + statusLine, e);
+    }
+  }
+
+  private static Map<String, String> readHeaders(InputStream in) throws IOException {
+    Map<String, String> headers = new HashMap<>();
```
yuqi1129, line 198: Headers go into a `HashMap`, so duplicate headers silently overwrite (last wins) rather than being rejected. Two conflicting `Content-Length` values should be rejected per RFC 7230 §3.3.2. Low risk over a single pinned connection with no intermediary, but cheap hardening while here.

common/src/main/java/org/apache/gravitino/utils/RemoteUriValidator.java
```diff
@@ -71,19 +74,119 @@ private static boolean isUnsafeAddress(InetAddress address) {
...
+    }
+    if (highTenZero) {
+      boolean compatible = bytes[10] == 0 && bytes[11] == 0;
+      // Defensive: the JDK normally collapses IPv4-mapped addresses to a 4-byte Inet4Address.
+      boolean mapped = (bytes[10] & 0xFF) == 0xFF && (bytes[11] & 0xFF) == 0xFF;
+      if (compatible || mapped) {
+        return lowestFourBytes(bytes);
+      }
+    }
+    // NAT64 well-known prefix 64:ff9b::/96 (RFC 6052): the IPv4 is the low 32 bits.
```
yuqi1129: NAT64 matching only covers the well-known `64:ff9b::/96` prefix. RFC 6052 network-specific prefixes (and RFC 8215 `64:ff9b:1::/48`) embedding e.g. `169.254.169.254` would still pass as global-unicast. Only exploitable where such a prefix is actually routed, and full coverage isn't practical without config — worth a comment noting the limitation rather than a code change.

yuqi1129 (commented):
Reviewed the SSRF DNS-rebinding hardening. The pinning, TLS SNI/cert verification, truncation/early-EOF handling, temp-file and socket cleanup, `Content-Length`/chunk-size parsing and bounds, line-folding rejection, and CRLF-injection defenses all look correct, and validate-all-then-pin-`addresses[0]` is sound. No high-severity bypass found — the notes below are hardening/compat items, none blocking.

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
LuciferYang: Good catch. `writeBody` now requires the `Transfer-Encoding` to be exactly `chunked` (trimmed, case-insensitive); `gzip, chunked`, `xchunked`, etc. are rejected with an `Unsupported Transfer-Encoding` error rather than de-chunked into corrupt bytes. Added tests for both `gzip, chunked` and `xchunked`. (335f6aea3)

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
LuciferYang: You're right that the per-read `soTimeout` doesn't bound total wall-clock — on the chunked path a server dripping one byte just inside each window can stretch a transfer toward `MAX_BODY_BYTES`. Two notes, then a proposal:

- The fetch is only triggered when an operator configures a catalog with a keytab/jar URI (a privileged action), so the blast radius is a single stuck worker thread per misconfiguration rather than an unauthenticated DoS — which is why it wasn't already a hard default.
- The API currently carries only a per-operation timeout, so an overall budget is a new constant that has to stay generous enough not to abort a legitimately large/slow jar fetch.

That said, I'm happy to add it as cheap insurance: an overall deadline checked in the copy loops, either a generous fixed constant or wired to config so it's tunable. Do you have a preference on the value / whether it should be configurable? I'll add it in this PR.

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
LuciferYang: Switched the request to HTTP/1.1 (still `Connection: close`) to bias origins toward `Content-Length`/chunked framing. Close-delimited bodies are still rejected for truncation safety (intentional — documented in the class Javadoc). Added a test asserting the request line is HTTP/1.1. (335f6aea3)

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
LuciferYang: Fixed: conflicting duplicate `Content-Length`/`Transfer-Encoding` headers are now rejected (RFC 7230 §3.3.2/§3.3.3) instead of silently last-winning; identical duplicates are tolerated. Added tests. (335f6aea3)

common/src/main/java/org/apache/gravitino/utils/RemoteUriValidator.java
LuciferYang: Added a comment noting the limitation: only the well-known `64:ff9b::/96` prefix is matched; network-specific prefixes (RFC 6052 §2.2, RFC 8215 `64:ff9b:1::/48`) are site-defined and routable only where deployed, so detecting them would need configuration. (335f6aea3)

jerryshao:
@yuqi1129 can you please help to review again?

common/src/main/java/org/apache/gravitino/utils/RemoteUriValidator.java
```diff
@@ -71,19 +74,122 @@ private static boolean isUnsafeAddress(InetAddress address) {
...
+    int b2 = bytes[2] & 0xFF;
+    int b3 = bytes[3] & 0xFF;
+
+    // 0.0.0.0/8 "this network" (RFC 1122). isAnyLocalAddress only covers the single 0.0.0.0.
+    if (b0 == 0) {
+      return true;
+    }
+    // 100.64.0.0/10 carrier-grade NAT / shared address space (RFC 6598). This range also covers
+    // the Alibaba Cloud metadata endpoint 100.100.100.200. Not caught by isSiteLocalAddress.
+    if (b0 == 100 && (b1 & 0xC0) == 0x40) {
```
yuqi1129, line 184: Why will we still need to `& oxC0`? Can we use `b1 == 0x40` directly?

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
```diff
@@ -0,0 +1,396 @@
...
+ * place only on success, this means a connection that drops mid-transfer never leaves a truncated
+ * destination behind. Response and header sizes are bounded to prevent a malicious pinned host from
+ * exhausting disk or memory.
+ */
+final class RemoteFileDownloader {
+
+  private static final int DEFAULT_HTTP_PORT = 80;
+  private static final int DEFAULT_HTTPS_PORT = 443;
+  private static final int MAX_HEADER_BYTES = 64 * 1024;
+  private static final long MAX_BODY_BYTES = 2L * 1024 * 1024 * 1024;
```
yuqi1129, line 73: Better to add comments like `we can change this value if a larger file is needed`

common/src/main/java/org/apache/gravitino/utils/RemoteUriValidator.java
LuciferYang: `100.64.0.0/10` (RFC 6598) has a 10-bit prefix, so its range is `100.64.0.0`–`100.127.255.255` — the second octet runs 64–127, not just 64. `(b1 & 0xC0) == 0x40` tests the top two bits to cover that whole range; `b1 == 0x40` would only match `100.64.0.0/16`.

The address that motivated this entry — the Alibaba Cloud ECS metadata IP `100.100.100.200` ([Alibaba Cloud docs](https://help.aliyun.com/en/ecs/user-guide/view-instance-metadata/)) — has second octet 100, so `b1 == 0x40` would let it through. Added a comment making the 10-bit prefix explicit.

common/src/main/java/org/apache/gravitino/utils/RemoteFileDownloader.java
LuciferYang: Added a comment noting the cap can be raised for larger files.

yuqi1129:
I have no more valuable comments.

LuciferYang:
Thank you @yuqi1129

yuqi1129 (approved):

PR #11788, opened 2026-06-24
[#11781] fix(idp-basic): Cache Basic auth group names by username

### LINKED ISSUE #11781: [Improvement] Group-based authorization adds 2-3x latency overhead compared to user-based authorization

Group-based authorization (built-in IDP with group→role mapping) introduces significant latency overhead compared to direct user-based authorization for single-table operations.

**Performance test results** (30 concurrent users, Auth:ON, Cache:ON, CredentialVending:ON, schema `load3000` with 10,496 tables):

| Service | Auth Mode | Samples | RPS | P50 | P75 | P95 |
|---------|-----------|---------|-----|-----|-----|-----|
| Iceberg REST | User Auth | 4001 | 13.36 | 130ms | 140ms | 180ms |
| Iceberg REST | Group Auth | 3673 | 12.27 | 300ms | 360ms | 510ms |
| Gravitino | User Auth | 4142 | 13.83 | 65ms | 75ms | 110ms |
| Gravitino | Group Auth | 3834 | 12.81 | 210ms | 270ms | 430ms |

Key observations:
- **Group Auth adds ~130-170ms extra latency per request** compared to User Auth on the same operation.
- P50 latency increases by **130% (Iceberg REST)** and **223% (Gravitino)** when switching from User Auth to Group Auth.
- Tail latency (P95) is disproportionately affected — P95/P50 ratio grows from 1.38x to 1.70x (Iceberg) and 1.69x to 2.05x (Gravitino), suggesting occasional cache misses or extra lookups in the group→role resolution path.
- RPS impact is moderate (~8% drop), but the latency increase is significant for latency-sensitive workloads.

The extra latency likely comes from the additional group→role resolution step required on each request when using group-based authorization.

### Test Environment Background

- **Metalake:** `irc_test`
- **Catalog:** `iceberg` (Iceberg REST service at port 9001, Gravitino service at port 8090)
- **Schema:** `load3000` containing **10,496 tables**
- **Concurrent users:** 30 (user1 - user30)
- **Group Auth setup:** All 30 users belong to a single group `perf_test_group`, which is granted `role_all` (full metalake access). The built-in IDP extension (`org.apache.gravitino.idp.web.rest.feature`) is enabled with `gravitino.authenticators = basic`.
- **User Auth setup:** Each user is directly assigned `role_all` plus an individual role (`role1` - `role30`).
- **Operation:** Load single table (random table from `load3000` schema)

### How should we improve?

Potential solutions:
1. **Cache group→role mapping** — Cache the resolution from group membership to role assignments so it does not require a lookup on every request.
2. **Reduce IDP lookup round-trips** — Batch or pre-resolve user→group→role chain at authentication time rather than at each authorization check.
3. **Session-level authorization context** — Resolve the full permission set once per session/token and reuse it for subsequent requests within the same session.

lasdf1234:

Add `IdpUserGroupsCache` to cache built-in IdP group names keyed by username during Basic authentication, reducing relational store lookups on repeated requests. Password verification still uses `getIdpUserByUsername` on every request; only group membership is cached. Invalidate cached group names when users are removed, group membership changes, or a group is force-deleted. Add server configs `gravitino.authenticator.basic.groupsCacheTtlSecs` (default 3600) and `gravitino.authenticator.basic.groupsCacheSize` (default 10000), documented in the built-in IDP guide.

Basic authentication previously loaded full user metadata (including group membership) from the relational store on every request. Group authorization adds extra DB work compared to user-only checks. This change caches group names per username while preserving per-request password verification, as discussed in #11781.

Fix: #11781

Yes, two new optional configuration keys:

- `gravitino.authenticator.basic.groupsCacheTtlSecs` — TTL in seconds for the per-username group names cache (default 3600)
- `gravitino.authenticator.basic.groupsCacheSize` — maximum number of cached usernames (default 10000)

No existing keys or default authentication behavior change.

- `./gradlew :plugins:idp-basic:test -PskipITs`
- `TestIdpUserGroupsCache` — cache hit/miss and invalidation
- `TestIdpUserGroupManager` — auth with groups, wrong/old password, cache invalidation on membership change and user removal
- `TestBasicAuthenticator` and `TestBasicAuthenticationIntegration`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/Configs.java
+++ b/core/src/main/java/org/apache/gravitino/Configs.java
@@ -245,6 +245,30 @@ public class Configs {
               ConfigConstants.NOT_BLANK_ERROR_MSG)
           .createWithDefault(Lists.newArrayList("simple"));
 
+  public static final long DEFAULT_BASIC_AUTHENTICATION_GROUPS_CACHE_TTL_SECS = 3600L;
+
+  public static final ConfigEntry<Long> BASIC_AUTHENTICATION_GROUPS_CACHE_TTL_SECS =
+      new ConfigBuilder("gravitino.authenticator.basic.groupsCacheTtlSecs")
+          .doc(
+              "The expiration time in seconds for the built-in IdP Basic authentication "
+                  + "group-name cache keyed by username")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .longConf()
+          .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(DEFAULT_BASIC_AUTHENTICATION_GROUPS_CACHE_TTL_SECS);
+
+  public static final long DEFAULT_BASIC_AUTHENTICATION_GROUPS_CACHE_SIZE = 10000L;
+
+  public static final ConfigEntry<Long> BASIC_AUTHENTICATION_GROUPS_CACHE_SIZE =
+      new ConfigBuilder("gravitino.authenticator.basic.groupsCacheSize")
+          .doc(
+              "The maximum number of cached usernames for the built-in IdP Basic authentication "
+                  + "group-name cache")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .longConf()
+          .checkValue(value -> value > 0, ConfigConstants.POSITIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(DEFAULT_BASIC_AUTHENTICATION_GROUPS_CACHE_SIZE);
+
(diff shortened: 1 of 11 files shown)


lasdf1234:
@yuqi1129 Could you help me to review this code?THX~

yuqi1129:
We need to consider the cache consistency in a multiple-node deployment.

roryqi:
We should avoid using cache if possible. We should optimize the SQL query to select complete information once.

lasdf1234:
> We should avoid using cache if possible. We should optimize the SQL query to select complete information once.

SQL has been optimized to handle data with the least amount of I/O operations.

lasdf1234:
After conducting tests, I found that the bottleneck was not in the database query. This PR has been temporarily converted to a draft PR.

PR #11790, opened 2026-06-24
[#10894] docs: Add design doc for role assumption (SET ROLE) permission narrowing

bharos:

Adds a design doc at `design-docs/gravitino-role-assumption.md` proposing **role assumption (SET ROLE)** — letting a caller narrow the active role set for a request via an `X-Gravitino-Active-Role` header. The narrowing is **subtractive-only** (validated against the caller's roles; can never escalate) and applies consistently across access checks, list filtering, and credential vending on the native authorization path (Iceberg REST + native Gravitino API).

The doc covers the header grammar, how the server enforces narrowing on the existing Casbin model without a model change, transport per engine (Spark works today, Trino needs a small setting), security/trust boundaries (cooperative header vs. token-bound), prior art (Snowflake/Hive/Postgres), options considered, and a phased rollout.

A workload that only needs a few tables still runs with everything its identity can reach; misbehavior fails *open* and silent. Role assumption gives workloads (and AI agents) real runtime least-privilege and a clean audit signal, and restores `SET ROLE` parity for teams migrating Hive SQL-Standard-Auth tables to Iceberg behind Gravitino.

This is a **proposal for discussion** (Status: Draft). Key open decisions are listed in §12 — most importantly the ownership-vs-narrowing semantics (§5). Feedback welcome inline.

Discussion: apache/gravitino#10894

No — documentation only.

N/A — design document only.

### DIFF
--- a/design-docs/gravitino-role-assumption.md
+++ b/design-docs/gravitino-role-assumption.md
@@ -1,0 +1,340 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Design: Role Assumption (SET ROLE) — Narrowing Effective Permissions
+
+| Field | Value |
+| ---------- | ------------------------------------------------------------------------- |
+| Status     | Draft |
+| Authors    | @bharos |
+| Created    | 2026-06-24 |
+| Discussion | [#10894](https://github.com/apache/gravitino/discussions/10894) |
+| Scope      | Native authorization path (Iceberg REST catalog + native Gravitino API) |
+
+---
+
+## 1. Summary
+
+When a user holds multiple roles, Gravitino always enforces the union of all of them. A workload has no
+way to run with a narrower subset of the access its identity could reach.
+
+This document proposes **role assumption** — the analog of Snowflake's `USE ROLE` and Hive's `SET ROLE`.
+A caller declares which role(s) should be active for a request, Gravitino verifies the caller actually
+holds them, and authorization is evaluated against only that active set. The feature can only *reduce* a
+caller's effective permissions, never expand them. If no role is declared, behavior is exactly as it is
+today.
+
+The mechanism is intentionally simple: a request header, `X-Gravitino-Active-Role`, carries the active
+role(s), and the server narrows enforcement to match — across access checks, list results, and credential
+vending alike. Section 3 defines the header and its values; Section 4 explains how the server enforces
+it. Transport is the easy part — the common instinct is to treat this as "just a Trino change," but the
+substance is server-side.
+
+---
+
+## 2. Motivation
+
+### 2.1 The problem
+
+A pipeline or AI agent that only needs 5 tables still runs with access to every table (say 105) its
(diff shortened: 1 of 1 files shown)


design-docs/gravitino-role-assumption.md
```diff
@@ -0,0 +1,334 @@
...
+- Write/`CREATE` semantics (which role owns newly created objects) — flagged for forward-compat only.
+
+---
+
+## 3. The active-role header
+
+The interface is a single request header:
+
+```
+X-Gravitino-Active-Role: <value>
```
roryqi: Should it be active roles?

design-docs/gravitino-role-assumption.md
```diff
@@ -0,0 +1,334 @@
...
+
+## 7. Transport — how each engine sends the active role
+
+The mechanism is a request header. Engines differ only in whether they can forward it:
+
+| Access path | Works today? | What's needed |
+|---|---|---|
+| Direct Gravitino API | ✅ Yes | Server-side change only |
+| **Spark** (via Iceberg REST) | ✅ Yes | No code change — a catalog config setting |
+| **Trino** (via Iceberg REST) | ❌ No | A small connector setting, *or* the token path (no Trino change) |
```
roryqi: New version Trino starts to support configure custom HTTP header.

design-docs/gravitino-role-assumption.md
```diff
@@ -0,0 +1,334 @@
...
+- **Option A — ownership always applies.** Narrowing affects role-granted privileges only; you keep
+  access to objects you own. Simplest and least surprising for existing deployments, but a narrowed
+  workload could still reach objects it owns outside its declared scope.
+- **Option B — ownership is narrowed too.** When an active set is declared, ownership-derived access is
+  honored only if it is reachable through the active roles. A stronger least-privilege guarantee, but a
+  larger behavioral change.
+- **Option C — make it explicit.** A reserved value in the header grammar lets the caller opt ownership
+  in or out, mirroring Snowflake's primary-vs-secondary distinction.
+
+Proposed default: ship **Option A** for v1 — it delivers the read-narrowing that motivates the feature
```
roryqi, line 234: OK for me.

design-docs/gravitino-role-assumption.md
```diff
@@ -0,0 +1,334 @@
...
+
+A value is either a role name, a comma-separated list of role names, or one of the reserved keywords
+`ALL` / `NONE`:
+
+| Value | Meaning |
+|---|---|
+| `<role>` (e.g. `analyst`) | Activate a single named role. |
+| `<role>,<role>` (e.g. `analyst,reader`) | Activate a list of roles; effective access is the union of just these. |
+| `ALL` | Activate every role the caller holds — identical to today's behavior. |
+| `NONE` | Activate no roles; all role-derived access is denied. |
```
roryqi, line 113: Why do we need `NONE`?

bharos: NONE is the only way to request the empty active set: an absent header already means ALL, so without an explicit keyword there's no way to say "use none of my roles." It's useful for owner-only execution (a job/agent that should touch nothing its roles grant) and for verifying deny paths, and costs nothing — an empty set just means no role policies are consulted. 
I see similar pattern in snowflake USE SECONDARY ROLES as well:
https://docs.snowflake.com/en/sql-reference/sql/use-secondary-roles

This is not a top-priority to support for now, but we can do it as it's not that hard to support as well.

design-docs/gravitino-role-assumption.md
bharos: Nice, I see trino upstream issues like https://github.com/trinodb/trino/issues/24236 which likely means this is supported, at least at a static level with the connector config

design-docs/gravitino-role-assumption.md
bharos: Fixed

roryqi:
How to handle not-existed roles or other illegal roles?

bharos:
> How to handle not-existed roles or other illegal roles?

Good question  , I added a "Validation and error handling" subsection. Two failure classes:

Malformed value → 400: empty entries, an unknown keyword, or ALL/NONE combined with anything (they're exclusive).
Unauthorized role → 403: well-formed but names a role you don't currently hold. A non-existent role and a not-granted role are treated the same (both 403) so the response can't be used to probe which roles exist.

roryqi (approved):
LGTM.

bharos:
Thanks @roryqi for the review!

PR #11797, opened 2026-06-25
[#5869] test(client-python): add special character and encoding tests

### LINKED ISSUE #5869: [Subtask] Handle file name encoding and special characters.
### Describe the subtask

In different client environments, file name encoding and the special characters within them need to be handled to ensure correct file operations and avoid garbled text.

### Parent issue

#5504

mariamjafri:

Added a new unit-test file: `tests/unittests/test_gvfs_filename_encoding.py` which tests file-name encodings that contain special characters in the GVFS path-conversion layer. Particularly testing the following functions: 
`actual_path_to_gvfs_path`, `actual_info_to_gvfs_info`, and the round-trip through `extract_identifier` / `get_sub_path_from_virtual_path` / `to_gvfs_path_prefix`.

Tests include spaces and URL-significant ASCII (`#`, `%`, `&`, `?`, `+`, `=`, `@`, `!`), multi-byte Unicode (CJK, accented Latin, Cyrillic, RTL), emoji, and path-structure characters (encoded slash `%2F`, leading/trailing dots, very long names) placed in filenames, nested directories, fileset-name segments, and sub-paths.

The tests confirm all of these convert and round-trip correctly. One issue found is documented via `@unittest.expectedFailure`: the all-occurrences `str.replace` in `actual_path_to_gvfs_path` corrupts paths when the
storage prefix recurs as a later directory segment. 

Fix: #5869

Per the above issue, GVFS special-character / encoding behavior was untested and its correctness
was uncertain. These tests verify and document it.

No. 

`pytest tests/unittests/test_gvfs_filename_encoding.py -v` → 5 passed, 1 xfailed
100 subtests passed. 
`ruff check` and `ruff format --check` pass clean.

### DIFF
--- a/clients/client-python/tests/unittests/test_gvfs_filename_encoding.py
+++ b/clients/client-python/tests/unittests/test_gvfs_filename_encoding.py
@@ -1,0 +1,188 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


yuqi1129:
@mariamjafri 
Please help to review this one.

diqiu50:
@mariamjafri Thanks for your contribution.
The goal of this issue is to investigate and resolve filename encoding issues in GVFS FUSE, rather than address issues in the GVFS Python client.

PR #11806, opened 2026-06-26
[#11802] fix(clickhouse): support user-defined GRANULARITY and validate shard key type

### LINKED ISSUE #11802: [Bug] ClickHouse catalog: GRANULARITY hardcoded and shard key validation missing

There are two issues in the ClickHouse catalog's index and distributed table handling. Both are in the same file and involve pending TODO items from the same author, so I'm filing them together. Happy to split if preferred.

**1. Index GRANULARITY hardcoded with wrong default for bloom_filter**

`ClickHouseTableOperations.java` hardcodes GRANULARITY values in three locations:
- `appendIndexesSql()` (line 486): minmax → `GRANULARITY 1`
- `appendIndexesSql()` (line 495): bloom_filter → `GRANULARITY 3`
- `addIndexDefinition()` (line 841, 845): same hardcoded values

The bloom_filter default of `3` is incorrect — ClickHouse's actual default is `1`. This means tables created through Gravitino have different index behavior than tables created directly via SQL.

Additionally, `getSecondaryIndexes()` (line 1243) queries `system.data_skipping_indices` but does not read the `granularity` column, losing round-trip fidelity.

The TODO comments at lines 484 and 491 acknowledge this gap: `// TODO(yuqi) add a properties field to Index to support user defined GRANULARITY value.`

The `Index.properties()` API (Index.java line 52) and `Indexes.of(type, name, fields, properties)` already exist — the infrastructure is ready.

**2. Shard key validation only checks column existence, not type or nullability**

`ClickHouseTableOperations.java` `handleDistributeTable()` (line 348-352) has a TODO: `// TODO(yuqi) WE need to check the columns in shard keys should be integer and not nullable`.

Current validation (lines 359-369) only checks that the column name exists. ClickHouse itself rejects non-integer and nullable columns with `TYPE_MISMATCH`, but the error message is opaque. Client-side validation would provide clearer feedback.

Verified on a local ClickHouse cluster:
- `Nullable(Int64)` as shard key → server error: `Sharding expression has type Nullable(Int64), but should be one of integer type`
- `String` as shard key → server error: `Sharding expression has type String, but should be one of integer type`

### How should we improve?

**For GRANULARITY (3 locations in ClickHouseTableOperations.java):**
- Read `index.properties().getOrDefault("granularity", "1")` instead of hardcoded values
- In `getSecondaryIndexes()`, add `granularity` to the SELECT query and pass it via `Indexes.of(type, name, fields, Map.of("granularity", value))`
- Fix bloom_filter default from `3` to `1` (ClickHouse actual default)

**For shard key validation (handleDistributeTable):**
- After finding the column, add:
  ```java
  Preconditions.checkArgument(!col.nullable(), 
      "Sharding key column '%s' must not be nullable", columnName);
  Preconditions.checkArgument(isIntegerType(col.dataType()), 
      "Sharding key column '%s' must be an integer type", columnName);
  ```

jiangxt2:
#### GRANULARITY round-trip (`ClickHouseTableOperations.java`)

- **Read path**: `getSecondaryIndexes()` queries the `granularity` column from `system.data_skipping_indices` and stores it in `Index.properties()`. Default values (1) are omitted so that indexes created without explicit GRANULARITY have empty properties and match the original creation state (avoids false index-change diffs).
- **Write path**: `appendIndexesSql()` passes `resolveGranularity(index.properties(), 1)` to `buildDataSkippingIndexDdl`, replacing the old hardcoded defaults.

#### Shard key validation (`ClickHouseTableOperations.java`)

- Extracted `validateShardKeyColumns()` shared method covering column existence, nullable check, and integer type check. Called from both `create()` (empty columns path) and `handleDistributeTable()` (explicit columns path).
- Added `isIntegerType()`: accepts Gravitino `IntegralType` (Int8-Int64, UInt8-UInt64) and ClickHouse wide integers (Int128/256, UInt128/256) mapped to `ExternalType`, using regex for future-proof coverage.
- Function-wrapped shard keys (e.g. `cityHash64(string_col)`) skip type/nullability checks — only column existence is verified.
- **Empty columns path**: New `create()` override fetches remote table columns via `fetchRemoteColumns()` and validates shard keys before `super.create()`, covering the case where `handleDistributeTable` is skipped (distributed table using `AS remote_table`).

#### `getColumns()` override (`ClickHouseTableOperations.java`)

- Parent `JdbcTableOperations.getColumns()` uses `connection.getSchema()` as the schema pattern, which is incorrect for ClickHouse when the target database differs from the connection's default database. The override passes `databaseName` directly to `metaData.getColumns()`.

ClickHouse data skipping indexes support a `GRANULARITY` parameter, but the existing code hardcoded default values. Users who create indexes with custom GRANULARITY would lose that information on round-trip.

Distributed table shard keys in ClickHouse must be non-nullable integer columns. Previously this was documented in a TODO comment — invalid shard keys reached the ClickHouse server and produced opaque error messages instead of clear client-side validation.

1. Custom GRANULARITY for data skipping indexes via `Index.properties()` — round-trip (create → load → recreate) preserves non-default values.
2. Nullable or non-integer shard keys are rejected at the Gravitino client level with clear `IllegalArgumentException` messages.
3. Int128/UInt128/Int256/UInt256 columns are now accepted as shard keys.

**Unit tests** (`-PskipITs`):

`TestClickHouseTableOperations`:
- `testGranularityNormalizedToCanonicalForm` — leading zero / whitespace trimming

`TestClickHouseTableOperationsCluster`:
- `testIndexNonNumericGranularityRejected`, `testIndexZeroGranularityRejected`, `testIndexZeroAfterParsingGranularityRejected`, `testIndexNegativeGranularityRejected` — invalid GRANULARITY
- `testShardingKeyNullableColumnRejected`, `testShardingKeyNonIntegerColumnRejected` — shard key rejected
- `testShardingKeyFunctionWithNonIntegerColumnAccepted` — function-wrapped key
- `testShardingKeyInt128ColumnAccepted`, `testShardingKeyUInt128ColumnAccepted` — wide integer keys
- `testAlterTableAddIndexDefaultGranularity`, `testAlterTableAddIndexWithClusterProperties`, `testAlterTableAddIndexWithNullProperties` — ALTER TABLE ADD INDEX

**Docker integration tests** (`-PskipDockerTests=false`):

`CatalogClickHouseIT` (35 tests, 0 failures):
- `testCreateAndLoadWithCustomGranularity`, `testCreateAndLoadWithDefaultGranularity`, `testLoadSqlCreatedTableWithCustomGranularity` — GRANULARITY round-trip
- `testAlterTableAddIndexUsesDefaultGranularity` — ALTER TABLE ADD INDEX end-to-end

`CatalogClickHouseClusterIT` (17 tests, 0 failures):
- `testDistributedTableNullableShardKeyRejected`, `testDistributedTableNonIntegerShardKeyRejected` — shard key rejected
- `testDistributedTableIntegerShardKeyAccepted`, `testDistributedTableWideIntegerShardKeyAccepted`, `testDistributedTableFunctionShardKeyAccepted` — shard key accepted

Run commands:
```bash
./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test -PskipITs
./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test --tests "CatalogClickHouseIT" -PskipDockerTests=false
./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test --tests "CatalogClickHouseClusterIT" -PskipDockerTests=false
```

Closes #11802.

### DIFF
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
@@ -42,6 +42,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.regex.Matcher;
@@ -63,6 +64,7 @@ import org.apache.gravitino.catalog.jdbc.JdbcTable;
 import org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations;
 import org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils;
 import org.apache.gravitino.exceptions.NoSuchTableException;
+import org.apache.gravitino.exceptions.TableAlreadyExistsException;
 import org.apache.gravitino.rel.Column;
 import org.apache.gravitino.rel.TableChange;
 import org.apache.gravitino.rel.expressions.Expression;
@@ -79,11 +81,16 @@ import org.apache.gravitino.rel.expressions.transforms.Transform;
 import org.apache.gravitino.rel.expressions.transforms.Transforms;
 import org.apache.gravitino.rel.indexes.Index;
 import org.apache.gravitino.rel.indexes.Indexes;
+import org.apache.gravitino.rel.types.Type;
+import org.apache.gravitino.rel.types.Types;
 
 public class ClickHouseTableOperations extends JdbcTableOperations {
 
   private static final String CLICKHOUSE_NOT_SUPPORT_NESTED_COLUMN_MSG =
       "Clickhouse does not support nested column names.";
+  /** Default GRANULARITY for data skipping indexes, matching ClickHouse's own default. */
+  private static final long DEFAULT_INDEX_GRANULARITY = 1;
+
   private static final Pattern ORDER_BY_PATTERN =
       Pattern.compile(
           "(?is)\\bORDER\\s+BY\\s*(.+?)(?=\\bPARTITION\\s+BY\\b|\\bPRIMARY\\s+KEY\\b|\\bSAMPLE\\s+BY\\b|\\bTTL\\b|\\bSETTINGS\\b|\\bCOMMENT\\b|$)");
@@ -95,6 +102,8 @@ public class ClickHouseTableOperations extends JdbcTableOperations {
   private static final Pattern DISTRIBUTED_ENGINE_PATTERN =
       Pattern.compile(
           "(?i)^Distributed\\(([^,]+),\\s*([^,]+),\\s*([^,]+),\\s*(.+)\\)$", Pattern.DOTALL);
+  /** Matches ClickHouse wide integer type names (Int128/256, UInt128/256, and future variants). */
+  private static final Pattern WIDE_INTEGER_PATTERN = Pattern.compile("^U?INT\\d+$");
 
   private static final String QUERY_INDEXES_SQL =
       """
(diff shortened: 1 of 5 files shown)


catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -502,6 +519,42 @@ private void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder) {
...
+   */
+  private static String resolveGranularity(Index index) {
+    String granularity =
+        index.properties().getOrDefault(INDEX_GRANULARITY_KEY, DEFAULT_INDEX_GRANULARITY);
+    Preconditions.checkArgument(
+        StringUtils.isNumeric(granularity) && Long.parseLong(granularity) >= 1,
+        "Index %s granularity must be a positive integer (>= 1), but got %s",
+        index.name(),
+        granularity);
+    return granularity;
```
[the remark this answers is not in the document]
jiangxt2: The `NumberFormatException` is already caught at line 542 and re-thrown as `IllegalArgumentException` with a clear message including the index name and invalid value. The catch block handles all `NumberFormatException` cases, including overflow.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -502,6 +519,52 @@ private void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder) {
...
+          granularity);
+      return String.valueOf(value);
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Index %s granularity must be a positive integer (>= 1), but got '%s'",
+              index.name(), granularity),
+          e);
+    }
+  }
```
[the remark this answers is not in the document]
jiangxt2: Good catch. I've added a `Preconditions.checkArgument` guard for null/blank values before parsing, which throws `IllegalArgumentException` with a clear message instead of an NPE.

docs/jdbc-clickhouse-catalog.md
```diff
@@ -240,8 +240,9 @@ If you need Gravitino to manage an existing cluster database or table, recreate
 
 - `PRIMARY_KEY`
 - Data-skipping indexes:
-  - `DATA_SKIPPING_MINMAX` (`GRANULARITY` fixed to 1)
-  - `DATA_SKIPPING_BLOOM_FILTER` (`GRANULARITY` fixed to 3)
+  - `DATA_SKIPPING_MINMAX` (default `GRANULARITY` 1, configurable via `Index.properties()` key `granularity`)
+  - `DATA_SKIPPING_BLOOM_FILTER` (default `GRANULARITY` 1, configurable via `Index.properties()` key `granularity`)
+  - Note: `ALTER TABLE ADD INDEX` always uses the default `GRANULARITY` because the `TableChange.AddIndex` API has no `properties` field.
```
yuqi1129: Can you add create an issue to track this problem?

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -504,6 +519,56 @@ private void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder) {
...
+  private static String resolveGranularity(Index index) {
+    String granularity =
+        index.properties().getOrDefault(INDEX_GRANULARITY_KEY, DEFAULT_INDEX_GRANULARITY);
+    Preconditions.checkArgument(
+        granularity != null && !granularity.isBlank(),
+        "Index %s granularity must not be null or blank",
+        index.name());
+    try {
+      long value = Long.parseLong(granularity);
+      Preconditions.checkArgument(
```
yuqi1129: Can you try to use `NumberUtils.isCreatable` to simplify the code?

docs/jdbc-clickhouse-catalog.md
jiangxt2: Created #11946 to track this.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Thanks for the suggestion! I looked into NumberUtils.isCreatable — it accepts hex (0xFF), decimals (3.14), scientific notation (1.2e3), and negative numbers, which are all invalid for granularity (must be a positive integer ≥ 1). So I think the current Long.parseLong + range check is more appropriate here.

yuqi1129: I see.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -112,6 +121,98 @@ ARRAY JOIN arrayZip(splitByChar(',', primary_key), arrayEnumerate(splitByChar(',
...
+        databaseName,
+        tableName,
+        columns,
+        comment,
+        properties,
+        partitioning,
+        distribution,
+        indexes,
+        sortOrders);
+  }
```
yuqi1129:
```suggest

if ( ArrayUtils.isNotEmpty(columns)) {
      super.create(
        databaseName,
        tableName,
        columns,
        comment,
        properties,
        partitioning,
        distribution,
        indexes,
        sortOrders);
}

....
```

Another thing is, why don't you add that check logic in the following code block

https://github.com/apache/gravitino/blob/8a07043b9edeb372d53c403b99ba94279c31d6a2/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java#L431-L434


This makes me feel a bit out of place

jiangxt2: Good point. The reason it's here rather than in `handleDistributeTable` is an architectural constraint: `generateCreateTableSql` (and its callees including `handleDistributeTable`) is a pure SQL generation method with no `Connection` parameter. When `columns` is empty, we need a JDBC connection to fetch the remote table's column metadata for shard key validation.

Moving the fetch into `handleDistributeTable` would require either changing the base class signature (breaks all catalog impls) or passing state implicitly (hurts readability).

I agree it's not ideal to check Distributed engine in two places though. Let me see if there's a cleaner way to restructure this — will follow up.

jiangxt2: Followed up as promised. After evaluating several restructuring approaches (fetching columns then passing down, internal property markers, ThreadLocal), they all introduce either implicit state or side effects in the pure generation path.

My suggestion is to keep the current structure — the split is intentional separation of concerns: `create()` as the orchestration layer handles IO-dependent pre-validation, while `handleDistributeTable` handles validation when columns are already available.

If the base class evolves to support a `Connection`-aware generation path in the future, we can consolidate both checks into `handleDistributeTable` at that point.

What do you think?

yuqi1129: I see. Consideration limitations mentioned above, it's acceptable to keep as it is. 

You can still use the following codes to reduce code(brackets) cascading. 

```
if ( ArrayUtils.isNotEmpty(columns)) {
     return super.create(
        databaseName,
        tableName,
        columns,
        comment,
        properties,
        partitioning,
        distribution,
        indexes,
        sortOrders);
}

// special logic compare to 
```

jiangxt2: Done — refactored to use the early return pattern as suggested.

yuqi1129 (approved):

PR #11807, opened 2026-06-26
[#11803] fix(clickhouse): add ON CLUSTER support for ALTER TABLE

### LINKED ISSUE #11803: [Bug] ClickHouse catalog: ALTER TABLE missing ON CLUSTER for clustered tables

`ClickHouseTableOperations.generateAlterTableSql()` (line 821-823) generates `ALTER TABLE %s \n%s;` without `ON CLUSTER` clause. For tables created `ON CLUSTER`, this causes schema divergence — the ALTER only affects the local node.

In contrast, `generateDropTableSql()` (line 687-699) correctly reads cluster info from properties and generates `DROP TABLE ... ON CLUSTER ... SYNC`.

**Reproduction steps:**

1. Create a table on cluster:
   ```sql
   CREATE TABLE test ON CLUSTER 'cluster' (id Int64, val String) ENGINE = MergeTree() ORDER BY id;
   ```
2. Execute `ALTER TABLE test ADD COLUMN new_col Int64` (without ON CLUSTER)
3. Check all nodes: `new_col` exists on the local node only, other nodes don't have it

Verified on a local ClickHouse cluster: one node added `new_col`, the other nodes did not have it.

### How should we improve?

Follow the same pattern as `generateDropTableSql()`:
- Read `ClusterConstants.CLUSTER_NAME` from `lazyLoadTable.properties()`
- If present, inject `ON CLUSTER <clusterName>` into the ALTER TABLE SQL template

Location: `ClickHouseTableOperations.java`, `generateAlterTableSql()` method, line 822.

jiangxt2:

- `generateAlterTableSql()` now checks `ON_CLUSTER` and `CLUSTER_NAME` table properties and injects `ON CLUSTER <clusterName>` into the ALTER TABLE SQL when both are set.
- Adds `getOrCreateTable()` call for change types that do not otherwise require loading the table, so cluster metadata is available.
- Adds null-safety for `properties()` access, consistent with `generateDropTableSql()`.
- Omits `SYNC` (unlike `generateDropTableSql()`) because ClickHouse does not support synchronous ALTER operations.

Fix: #11803

Previously, `generateAlterTableSql()` always generated `ALTER TABLE ...` without `ON CLUSTER`. For tables created with `ON CLUSTER`, subsequent ALTER operations only affected the local node, causing schema divergence across cluster nodes.

Yes. ALTER TABLE operations on clustered tables now propagate to all nodes in the cluster, preventing schema divergence.

- `./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test --tests "org.apache.gravitino.catalog.clickhouse.operations.TestClickHouseTableOperationsCluster" -PskipITs`
- `./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test --tests "org.apache.gravitino.catalog.clickhouse.integration.test.CatalogClickHouseClusterIT" -PskipDockerTests=false`


yuqi1129 (commented):
LGTM

yuqi1129:
Could you also add some test cases in `CatalogClickHouseClusterIT`? @jiangxt2

jiangxt2:
Thanks for the suggestion! I've added `testAlterTableOnClusterSqlGeneration()` in `CatalogClickHouseClusterIT`.

**What it verifies**: queries `system.query_log` after ALTER TABLE operations to confirm:
- ALTER on a cluster table (ON_CLUSTER=true) → SQL contains `ON CLUSTER` clause ✅
- ALTER on a non-cluster table → SQL does NOT contain `ON CLUSTER` ✅

**Test result**:
```
CatalogClickHouseClusterIT > testAlterTableOnClusterSqlGeneration() PASSED (0.52s)
Tests: 1 passed, 0 failed, 0 skipped
```

This differs from the existing `testAlterTableBranchCoverageInCluster` which validates ALTER functionality (add/modify columns etc.). The new test specifically proves the ON CLUSTER clause is injected into the executed SQL — something that cannot be verified through the Gravitino API alone, since single-node clusters succeed regardless of ON CLUSTER presence.

yuqi1129 (approved):

yuqi1129:
@jiangxt2 
I'm going to merge this one. Could you try to modify the PR description, the following is a good example:
image
see: https://github.com/apache/gravitino/pull/11809

jiangxt2:
Thanks @yuqi1129! 
I've updated the PR description to follow the standard template. PTAL, thanks!

PR #11809, opened 2026-06-26
[#11059] feat(iceberg-rest): Support overwrite on registerTable endpoint

### LINKED ISSUE #11059: [Improvement] Support overwrite on Iceberg REST registerTable endpoint

The Iceberg REST `POST /v1/.../namespaces/{ns}/register` endpoint in Gravitino supports first-time registration only. There is no way to re-point an existing table at a new metadata.json in a single atomic operation.

The current workaround — dropTable(purge=false) followed by registerTable — is non-atomic (a query between the two steps sees the table missing) and corrupts Gravitino's entity bindings. `IcebergTableHookDispatcher.dropTable` unconditionally deletes the TableEntity (does not gate on purgeRequested), so the next `registerTable` mints a new table_id and orphans every role/owner/tag/policy binding keyed on the old table_id.

This blocks control-plane use cases that need atomic pointer updates: Iceberg metadata replication, external sync, DR/restore from a known-good metadata.json.

## Considered alternatives

- **Wait for Iceberg release with `Catalog.registerTable(id, location, overwrite)`
  in a release**: feasible, but the convenience method is sugar over primitives
  already public in 1.10.1. Same atomicity, same backend dispatch. Composition can
  simplify to a one-line delegate when the Iceberg pin moves.

- **Add this to Gravitino's native admin API (`/api/metalakes/.../tables/register`)
  instead of extending IRS**: arguably a better fit for control-plane operations,
  since IRS positions itself as Iceberg-spec-compliant. This issue focuses on the IRS-side change because it's the     smaller scoped extension and aligns with consumers already using IRS. Open to this approach too, but a bigger change.

### How should we improve?

Add overwrite to the registerTable flow:

overwrite=false (default): existing behavior; AlreadyExistsException if the table already exists.
overwrite=true: atomic in-place update of the metadata pointer. 
Gravitino entity preserved — table_id unchanged, role/owner/tag/policy bindings intact, only audit fields (last_modifier, last_modified_time) updated.
This can land on the current Iceberg 1.10.1 pin without bumping. The 3-arg Catalog.registerTable(id, location, overwrite) exists on Iceberg main but is not in a release yet; the primitives it composes — tableExists, loadTable, HasTableOperations.operations(), TableOperations.refresh(), TableMetadataParser.read(), TableOperations.commit(base, target) — are all public API in 1.10.1. Atomicity is provided by TableOperations.commit(base, target).

Changes affect IcebergCatalogWrapper (compose primitives), IcebergNamespaceOperationExecutor (pass overwrite through), and IcebergNamespaceHookDispatcher (preserve entity on overwrite path; update audit only). When the Iceberg pin moves to a release containing the convenience signature, the composition can simplify to a one-line delegate without API or behavior change.

API surface: Iceberg main has standardized overwrite as a body field on RegisterTableRequest. This proposal mirrors that design — extend Gravitino's request type with the same overwrite body field. Standard Iceberg REST clients (Spark, Trino) on 1.10.1 don't set the field, so default false preserves current behavior. When Gravitino's Iceberg pin moves to a release containing the standardized field, the Gravitino-side extension can be replaced by Iceberg's native type with no API surface change for callers.

registerView is a related but separate gap (no REST handler exists in Gravitino today; Iceberg main exposes it on a distinct /v1/.../namespaces/{ns}/register-view endpoint, not yet in a release) — should be tracked as its own issue.

lasdf1234:

Implement Iceberg REST `POST .../namespaces/{ns}/register` support for `RegisterTableRequest.overwrite=true`:

- Add `RegisterTableOverwrite` with CAS metastore pointer swap for JDBC, memory, and Hive backends (does not write a new metadata file via `TableOperations.commit`).
- Wire 3-arg `registerTable(..., overwrite)` through `JdbcCatalogWithMetadataLocationSupport`, `MemoryCatalogWithMetadataLocationSupport`, and `HiveCatalogWithMetadataLocationSupport`.
- Pass `request.overwrite()` through `FederatedCatalogWrapper` (federation path previously ignored the flag).
- In `IcebergNamespaceHookDispatcher`, skip Gravitino entity import and ownership assignment when `overwrite=true` so existing `table_id` and role/owner/tag/policy bindings are preserved.

Fix: #11059

Without `overwrite=true`, re-registering an existing table requires drop + register, which is non-atomic and can mint a new Gravitino table entity, orphaning bindings keyed on the old `table_id`.

Yes. IRC clients can now pass `overwrite=true` on registerTable to atomically repoint an existing table registration at a new metadata.json location.

- `./gradlew :iceberg:iceberg-common:test --tests org.apache.iceberg.jdbc.TestJdbcCatalogWithMetadataLocationSupport --tests org.apache.iceberg.memory.TestMemoryCatalogWithMetadataLocationSupport -PskipITs`
- `./gradlew :iceberg:iceberg-rest-server:test --tests org.apache.gravitino.iceberg.service.TestCatalogWrapperForREST --tests org.apache.gravitino.iceberg.service.dispatcher.TestIcebergNamespaceHookDispatcher -PskipITs`


lasdf1234:
@roryqi  Could help me to review this code? Thans.

roryqi:
Maybe we can use the CatalogHandlers.registerTable(overwrite = true) directly.

lasdf1234:
> Maybe we can use the CatalogHandlers.registerTable(overwrite = true) directly.

The CatalogWrapperForREST meets the requirements. The FederatedCatalogWrapper has already modified the code.

roryqi:
> > Maybe we can use the CatalogHandlers.registerTable(overwrite = true) directly.
> 
> The CatalogWrapperForREST meets the requirements. The FederatedCatalogWrapper has already modified the code.

You can only change the FederatedCatalogWrapper code.

lasdf1234:
> > > Maybe we can use the CatalogHandlers.registerTable(overwrite = true) directly.
> > 
> > 
> > The CatalogWrapperForREST meets the requirements. The FederatedCatalogWrapper has already modified the code.
> 
> You can only change the FederatedCatalogWrapper code.

Got code has been modified.

lasdf1234:
@The code has been revised. If you have time, please review it. Thank you.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergNamespaceHookDispatcher.java
```diff
@@ -183,6 +183,12 @@ public LoadTableResponse registerTable(
       RegisterTableRequest registerTableRequest) {
     LoadTableResponse response = dispatcher.registerTable(context, namespace, registerTableRequest);
 
+    if (registerTableRequest.overwrite()) {
+      // overwrite=true updates the Iceberg metadata pointer in place. The Gravitino table entity,
+      // table_id, and existing role/owner/tag/policy bindings must be preserved.
+      return response;
```
roryqi: How to handle this case if table doesn't exist and overwrite is true?

iceberg/iceberg-common/src/main/java/org/apache/iceberg/MetastoreRegisterTableUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    if (!catalog.tableExists(identifier)) {
+      return catalog.registerTable(identifier, metadataFileLocation);
+    }
+
+    TableOperations ops = catalog.newTableOps(identifier);
+    Preconditions.checkArgument(
+        ops instanceof BaseMetastoreTableOperations,
+        "Cannot register table %s: table operations are not metastore-backed",
+        identifier);
+    BaseMetastoreTableOperations metastoreOps = (BaseMetastoreTableOperations) ops;
```
roryqi, line 80: Should u use TableOperations to commit new location here?

roryqi:
Why do u add the label `1.3`?

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergNamespaceHookDispatcher.java
lasdf1234: Got resolved.

iceberg/iceberg-common/src/main/java/org/apache/iceberg/MetastoreRegisterTableUtils.java
lasdf1234: Using the commit method will generate a new metadata.json file.
Overwrite should be avoided so that no new metadata.json file is created.

lasdf1234:
> Why do u add the label `1.3`?

Got remove it

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/dispatcher/IcebergNamespaceHookDispatcher.java
```diff
@@ -183,6 +183,21 @@ public LoadTableResponse registerTable(
       RegisterTableRequest registerTableRequest) {
     LoadTableResponse response = dispatcher.registerTable(context, namespace, registerTableRequest);
 
+    if (registerTableRequest.overwrite()) {
+      TableDispatcher tableDispatcher = GravitinoEnv.getInstance().internalTableDispatcher();
+      if (tableDispatcher != null
```
roryqi: Could u avoid using tableDispatcher? You can use Iceberg related classes.
`exists` will always return `true`.

lasdf1234: Got. Use EntityStore to check if table exists.

roryqi (approved):
LGTM.

PR #11811, opened 2026-06-26
[#11810] fix(core): treat authorization plugin classes as isolated in IsolatedClassLoader

### LINKED ISSUE #11810: [Bug report] Authorization plugin classes loaded by server classloader cause permanent NoClassDefFoundError
### Version

main branch

### Describe what's wrong

`IsolatedClassLoader.isCatalogClass()` does not recognize the authorization plugin packages (`org.apache.gravitino.authorization.{ranger,chain,jdbc,common}.*`). As a result these plugin classes are classified as *shared* and delegated to the server classloader. Under certain classloader-initialization timing windows, the server classloader becomes the defining classloader of a plugin class (e.g. `RangerPrivileges`), and its nested/synthetic classes (e.g. the inner enum `RangerPrivileges$RangerHdfsPrivilege`) cannot be resolved from the server classpath. The JVM permanently caches this load failure for the lifetime of the process, so every subsequent role/authorization operation throws `NoClassDefFoundError` until restart.

This is the same class of bug as #11704 (fixed for the hive packages in #11705), now affecting the authorization plugins. It surfaces intermittently in `RangerHiveE2EIT` and is unrelated to the PR under test.

```
java.lang.NoClassDefFoundError: org/apache/gravitino/authorization/ranger/RangerPrivileges$RangerHdfsPrivilege
    at org.apache.gravitino.authorization.ranger.RangerPrivileges.<clinit>(RangerPrivileges.java:145)
    at org.apache.gravitino.authorization.ranger.RangerAuthorizationPlugin.doAddSecurableObject(RangerAuthorizationPlugin.java:864)
    ...
    at org.apache.gravitino.authorization.RoleManager.createRole(RoleManager.java:83)
Caused by: java.lang.ClassNotFoundException: Class not found
```

### How to reproduce

Intermittent. Run `RangerHiveE2EIT` (e.g. `testSelectModifyTableWithTableLevelRole`) in CI; depending on classloader init ordering, `RangerPrivileges` gets defined by the server classloader and the nested enum fails to resolve. Example failing run: https://github.com/apache/gravitino/actions/runs/28228153382/job/83626344824

### Additional context

Root cause is in `core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java` `isCatalogClass()`. Fix: whitelist the authorization plugin sub-packages (but NOT the bare `org.apache.gravitino.authorization.*`, which holds server-side classes like `RoleManager`).

yuqi1129:

Add the authorization plugin packages `org.apache.gravitino.authorization.{ranger,chain,jdbc,common}.` to `IsolatedClassLoader.isCatalogClass()` so these plugin classes are loaded by the isolated classloader instead of being delegated to the server classloader. The bare `org.apache.gravitino.authorization.*` prefix is intentionally left out so server-side classes (e.g. `RoleManager`, `AuthorizationUtils`) remain shared.

Tests: add `TestIsolatedClassLoader.testAuthorizationPluginPackagesRecognizedAsCatalogClass` and extend the negative test to assert server-side authorization classes stay shared.

Authorization plugin classes were classified as *shared* and delegated to the server classloader. Under certain classloader-initialization timing windows the server classloader becomes the defining classloader of a plugin class (e.g. `RangerPrivileges`), and its nested/synthetic classes (e.g. the inner enum `RangerPrivileges$RangerHdfsPrivilege`) cannot be resolved from the server classpath. The JVM permanently caches this load failure for the lifetime of the process, so every subsequent role/authorization operation throws `NoClassDefFoundError` until restart. This surfaces intermittently in `RangerHiveE2EIT`.

This is the same class of bug as #11704 (fixed for the hive packages in #11705), now applied to the authorization plugins.

Fix: #11810

No. This is an internal classloader fix with no API or behavior change.

`./gradlew :core:test --tests "org.apache.gravitino.utils.TestIsolatedClassLoader"` — passes (4 test methods, including the new authorization-plugin coverage and the server-side-stays-shared assertions).

### DIFF
--- a/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
+++ b/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
@@ -246,7 +246,20 @@ public class IsolatedClassLoader implements Closeable {
     // compiler-generated synthetic classes (e.g. $1 from switch-on-enum) are then requested
     // from the server classloader which cannot find them, causing a permanent
     // NoClassDefFoundError that is cached by the JVM for the lifetime of the process.
+    // org.apache.gravitino.authorization.{ranger,chain,jdbc,common}.* are authorization plugin
+    // classes loaded through this IsolatedClassLoader. They must be treated as isolated (catalog)
+    // classes; otherwise they are delegated to the server classloader, and their nested or
+    // compiler-generated synthetic classes (e.g. the inner enum
+    // RangerPrivileges$RangerHdfsPrivilege)
+    // cannot be resolved from the server classpath, producing a permanently cached
+    // NoClassDefFoundError. Note org.apache.gravitino.authorization.* (without these sub-packages)
+    // holds server-side classes (e.g. RoleManager) that must stay shared, so only the plugin
+    // sub-packages are whitelisted here.
     return name.startsWith("org.apache.gravitino.hive.")
+        || name.startsWith("org.apache.gravitino.authorization.ranger.")
+        || name.startsWith("org.apache.gravitino.authorization.chain.")
+        || name.startsWith("org.apache.gravitino.authorization.jdbc.")
+        || name.startsWith("org.apache.gravitino.authorization.common.")
         || name.startsWith("org.apache.gravitino.catalog.hive.")
         || name.startsWith("org.apache.gravitino.catalog.lakehouse.")
         || name.startsWith("org.apache.gravitino.catalog.jdbc.")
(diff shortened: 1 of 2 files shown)


yuqi1129:
Closing: this was based on an incorrect hypothesis. The `RangerPrivileges` NoClassDefFoundError is not caused by an `isCatalogClass()` whitelist gap. Root cause is elsewhere (the entity-change-log poller invalidating an in-use catalog and hard-closing its IsolatedClassLoader), so this change does not fix the issue. Will track the real fix separately.

PR #11812, opened 2026-06-27
Web UI support for built-in IdP Basic login

### LINKED ISSUE #11681: [Subtask] Frontend: add Basic login support in Web UI

Octavi00:
Web UI support for built-in IdP Basic login [#11682] EPIC

- Fix: #11681: Added Basic login functionality following front-end session logic as requested + supporting UI modifications.

### Files Changed + why are the changes needed?

1. server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java (Feature)
-     Adds web login header check to skip the unwanted browser basic auth popup

2. web/web/src/app/login/components/BasicLogin.js (New Component)
- Basic Login Page setup
- Handles user login input + error catching and message upon invalid credentials
- (Possibly add a more descriptive error message for invalid login?)
- Prompts UI error popup upon failed login

3. web/web/src/app/login/page.js (Feature)
- When visited + logged in (confirmed with an access  token) immediately redirects to metalakes
- Replaced if-else logic with switch case to handle different login components (oidc, basic, and default)

4. web/web/src/app/page.js (Feature)
- Replaced empty web page with auto redirect to real main page /metalakes as described to be the home page in the issue thread.

5. web/web/src/lib/api/auth/index.js (API Token)
- Added new 'basicToken'
- Allows simple token-accept functionalities, mainly 'X-Gravitino-Web-Login' which is useful later on for stopping unwanted page refresh upon invalid logins

6. web/web/src/lib/auth/providers/factory.js (Feature)
- Added new variable 'authenticators' to read from the current config file to allow for setting 'providerType' and returning early before unsuitable OAuth behaviour, which isn't wanted for Basic Auth
- Added missing 'this.providerType = providerType' line
- Added null catch case for provider now that basic doesn't carry one
- Fixed getProvider type to prevent possible null issues

7. web/web/src/lib/provider/session.js (Feature)
- Added else-if to redirect user to the metalake (home) page if already logged in otherwise forced into the login page.

8. web/web/src/lib/store/auth/index.js (Feature)
- Added basicLoginAction to replicate loginAction with custom Basic Auth logic
- Added specific variable retention during refresh in authSlice. Prevents logout button from disappearing during refresh
- Added authType to extraReducers for extra security with 'basic' functionality (previously only for oauth)

9. web/web/src/lib/utils/axios/index.js (Feature)
- Added isFullAuthHeader to set tokens relative to Basic Login auth while also maintaining jwt flow in the subsequent line
- Added 'isFailedWebUILoginRequest' to prevent page refresh upon failed logins (this is a feature meant for OAuth which should be disabled for basic auth)
- Added catch for 401 API pop-up caused by failed logins

  1. Change in user-facing APIs
- Added new basicLoginApi with basicToken.
- Contains X-Gravitino-Web-Login

       
  1. Enabled basic auth mode using given instructions, and manually tested
  3. Ensured additions didn't cause any new unit test errors
  4. Manually reverted to the default conf file to ensure default behaviours unaffected


LauraXia123:
@Octavi00 Please implement this UI support onto web-v2, as we will no longer be maintaining the legacy web version moving forward. Thx

web/web/src/lib/auth/providers/factory.js
```diff
@@ -65,6 +65,18 @@ class OAuthProviderFactory {
...
+      const authenticators = config['gravitino.authenticators'] || []
+
+      // Authenticator logic
+      if (authenticators.includes('org.apache.gravitino.idp.auth.BasicAuthenticator')) {
+        // If BasicAuthenticator is present, set provider type to 'basic' and do not initialize any OAuth provider
+        this.providerType = 'basic'
+        this.currentProvider = null
+
+        return null
+      }
```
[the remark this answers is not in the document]
LauraXia123: This file handles OAuth. Why is Basic authentication logic added here?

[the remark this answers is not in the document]
Octavi00: Hi Laura, I chose to use this file to avoid creating more auth files, as I felt I could easily keep the current auth flow by making this file a little more flexible to both auth types. I've made all other requested changes + tested to ensure it's working on my end, so I wanted to confirm if you'd like me to make a whole new file for basic Auth? Thank you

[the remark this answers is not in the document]
LauraXia123: In the case where authType === 'basic', the OAuth factory will not be called.

LauraXia123:
> @Octavi00 Please implement this UI support onto web-v2, as we will no longer be maintaining the legacy web version moving forward. Thx

@Octavi00  Please take a look at this comment. In addition, in the web v2 version, if authentication is enabled in simple mode, there will also be a login page, which is in the DefaultLogin component. So we need to first decouple the simple authentication login from the OAuth default provider login interface. Therefore, the login page should first be divided by oauthType into simple/basic/oauth, and then further divided by the providerType under the OAuth mode. The providerType is determined by the config['gravitino.authenticator.oauth.provider'] configuration.

LauraXia123:
@Octavi00 Hi, will you continue working on this PR? All changes need to support web v2, and any changes specific to the old web need to be removed.

Octavi00:
@LauraXia123 Yes, I assumed all original changes would want to be removed, but I'm currently working on transferring my additions from v1 to v2 right now. In doing so, I have noticed a few differences that may need to be addressed, but I will hold off until I get to a more complete state.

Octavi00:
Hi @LauraXia123, I've spent some time adapting my V1 changes into the V2 web, and it actually turned out to be a pretty smooth process. I used an ant design in my BasicLoginPage.js to avoid making an unnecessary decision, like installing another library in order to reuse my V1 page. I got the ui page to show a proper error message upon failed login attempts, and also included the redirection from root (8090) to homepage (ui/metalakes) as I mentioned in the v1 rendition. I avoided some unnecessary changes for v2, and I included the decoupling method you were looking for in the login/page.js. I separated auth from oidcs or default into being derived based on authType, as well as providerType if it's Oauth. Also, would you like me to rename the default variable to something like defaultOAuth? Apart from that, all changes are very similar, and I would like to request a copilot review, as was done previously, to help make sure I'm not overlooking something. 
Thank you

web-v2/web/src/app/login/page.js
```diff
@@ -28,36 +29,47 @@ import { useEffect, useState, Suspense } from 'react'
...
 
   useEffect(() => {
-    const detectProviderType = async () => {
-      try {
-        const detectedType = await oauthProviderFactory.getProviderType()
-        setProviderType(detectedType)
-      } catch (error) {
-        setProviderType('default') // fallback to default provider
+    const detectLoginType = async () => {
+      const [, resAuthConfigs] = await to(dispatch(getAuthConfigs()))
```
LauraXia123: In the initial stage of the project, getAuthConfigs() is called by default. You can see this in the initAuth function inside the session file. Therefore, you don’t need to call it again; just directly use authType from the store.auth object.

web-v2/web/src/app/page.js
```diff
@@ -19,6 +19,16 @@
...
+import { useEffect } from 'react'
+
+import { useRouter } from 'next/navigation'
+
 export default function Home() {
-  return <></>
+  const router = useRouter()
+
+  useEffect(() => {
+    router.replace('/metalakes')
```
LauraXia123: In the session provider, if there is a token, it should redirect to goToMetalakeListPage(); if there is no token, it should go to the Login page. It should not default to redirecting to the Metalake page here.

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
```diff
@@ -41,15 +41,21 @@
...
    */
   protected HealthCheckPathMatcher healthCheckMatcher = new HealthCheckPathMatcher();
 
+  private final List<Authenticator> filterAuthenticators;
+  // Skip HTTP challenge for Web UI login requests to avoid the browser's native Basic auth popup.
+  private boolean isWebLoginRequest(ServletRequest request) {
+    return request instanceof HttpServletRequest
+        && "true".equalsIgnoreCase(((HttpServletRequest) request).getHeader(WEB_LOGIN_HEADER));
+  }
+
```
LauraXia123: On the web side, we decide whether to call basicLoginApi based on whether authType is 'basic'. Is it necessary to add the header here? Or is this requirement due to some conflict on the backend? @lasdf1234

lasdf1234: @Octavi00 if "basic" is set as the first option, it will automatically enable the basic login function.What is this WEB_LOGIN_HEADER for?

lasdf1234:
@Octavi00 We shouldn't modify java code.

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
Octavi00: Im using isWebLoginRequest, as a way to avoid an unwanted web popup when an incorrect passsword is entered. Maybe I could add more description to these changes, but as of right now, removing the addition added to line 117 would recreate my issue. 


Image

web-v2/web/src/app/page.js
Octavi00: This was more of an issue with the URL adjustment that I had asked about previously. Basically, this is unrelated to my login function, and more so UX. If a user loads the main page, http://localhost:8090, they are greeted with a blank white page. Since I was previously explained that the main page for gravitino is the metalakes page, I had decided that it would be nice to include a redirect to the metalake page, which is the functional main page. This would happen regardless of authentication being enabled.

The nice thing about this change is that when auth is enabled, if the user is sent to metalakes page, they are then sent to login due to the metalakes page being locked without a token.

web-v2/web/src/app/login/page.js
Octavi00: Thanks for the input, I will apply your suggested changes. I hadn't realized that authType was loaded by default, so my adjustment was pretty redundant.

web-v2/web/src/lib/store/auth/index.js
```diff
@@ -144,25 +174,29 @@ export const logoutAction = createAsyncThunk(
...
+
+    dispatch(clearIntervalId())
+    dispatch(setAuthToken(''))
 
     // Always clear authUser in Redux and sessionStorage on logout
     // This ensures consistent behavior for both OAuth and simple auth
     dispatch(setAuthUser(null))
     sessionStorage.removeItem('simpleAuthToken')
 
-    // Clear persisted authType to avoid stale auth mode on next visit
```
Octavi00, line 163: I removed these lines due to logout -> login issues. Basically, the page would need to be refreshed to rerun initAuth and reset authType now that im no longer pulling Auth conf in longin.js

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
lasdf1234: It doesn't matter if a default browser pop-up appears. Don't modify any backend code for now.

web-v2/web/src/app/page.js
LauraXia123: The initAuth function in session.js already fully covers the redirect logic for all authType values:
* simple: Unauthenticated → /login; Authenticated → goToMetalakeListPage() → /metalakes
* basic: No token → /login; Has token → goToMetalakeListPage() → /metalakes
* oauth: No token → /login; Has token → goToMetalakeListPage() → /metalakes
The goToMetalakeListPage() function (defined at session.js:54-70) already executes router.replace('/metalakes') when the current pathname is /.
Since AuthProvider wraps all pages in layout.js, when a user visits the root path /:
1. AuthProvider's useEffect triggers initAuth, which handles the redirect based on auth state
2. page.js's useEffect also triggers router.replace('/metalakes')
These two redirects execute simultaneously, making the one in page.js completely redundant — it is entirely covered by initAuth's logic.
Conclusion: The router.replace('/metalakes') in page.js is unnecessary and can be safely removed, as initAuth already handles all redirect scenarios for every authType.

web-v2/web/src/app/login/components/BasicLogin.js
```diff
@@ -0,0 +1,106 @@
...
+
+        <Form.Item label='Password' validateStatus={errors.password ? 'error' : ''} help={errors.password?.message}>
+          <Controller
+            name='password'
+            control={control}
+            render={({ field }) => <Input.Password {...field} placeholder='' />}
+          />
+        </Form.Item>
+
+        <Button block size='large' type='primary' htmlType='submit'>
```
LauraXia123: Can u add a loading status for the login button?

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
lasdf1234: @Octavi00 After my research, I found that a better solution is that for basic authentication, if the status code is 401, then there should be no headers in the response.
Can you modify the Java code according to this logic?

web-v2/web/src/app/login/page.js
```diff
@@ -94,7 +99,15 @@ const LoginContent = () => {
...
-        {useOidcLogin ? <OidcLogin /> : <DefaultLogin />}
+        {authType === null ? null : authType === 'basic' ? (
+          <BasicLogin />
+        ) : authType === 'simple' ? (
+          <DefaultLogin />
+        ) : authType === 'oauth' && providerType === null ? null : authType === 'oauth' && providerType === 'oidc' ? (
+          <OidcLogin />
+        ) : authType === 'oauth' ? (
+          <DefaultLogin />
+        ) : null}
```
LauraXia123: I suggest splitting the simple login form and the OAuth login form within the DefaultLogin component into two separate components, so that loading components based on authType becomes clearer.

web-v2/web/src/lib/store/auth/index.js
```diff
@@ -175,6 +209,11 @@ export const logoutAction = createAsyncThunk(
 )
 
 export const setIntervalIdAction = createAsyncThunk('auth/setIntervalIdAction', async (expiredIn, { dispatch }) => {
+  // If the auth type is basic, we don't need to set an interval for token refresh
+  if (localStorage.getItem('authType') === 'basic') {
+    return { intervalId: null }
+  }
+
```
LauraXia123: setIntervalIdAction is never called for basic auth — the guard is unreachable dead code

web-v2/web/src/app/login/components/BasicLogin.js
```diff
@@ -0,0 +1,106 @@
...
+    defaultValues: Object.assign({}, defaultValues),
+    mode: 'onChange',
+    resolver: yupResolver(schema)
+  })
+
+  useEffect(() => {
+    if (store.intervalId) {
+      dispatch(clearIntervalId())
+    }
+  }, [store.intervalId, dispatch])
```
LauraXia123: intervalId is always null when BasicLogin is mounted; logoutAction already clears it

web-v2/web/src/app/login/components/BasicLogin.js
```diff
@@ -0,0 +1,106 @@
...
+
+import { useRouter } from 'next/navigation'
+import { useEffect } from 'react'
+import { Button, Form, Input } from 'antd'
+import * as yup from 'yup'
+import { useForm, Controller } from 'react-hook-form'
+import { yupResolver } from '@hookform/resolvers/yup'
+
+import { useAppDispatch, useAppSelector } from '@/lib/hooks/useStore'
+import { basicLoginAction, setIntervalIdAction, clearIntervalId } from '@/lib/store/auth'
```
LauraXia123: `setIntervalIdAction` It's imported but never used，should remove

web-v2/web/src/lib/provider/session.js
```diff
@@ -100,6 +100,16 @@ const AuthProvider = ({ children }) => {
           dispatch(setAuthUser(sessionUser))
           goToMetalakeListPage()
         }
+      } else if (authType === 'basic') {
+        const tokenToUse = sessionStorage.getItem('accessToken')
+
+        if (tokenToUse) {
+          basicAuthToken && dispatch(setAuthToken(basicAuthToken))
+          basicAuthUser && dispatch(setAuthUser(basicAuthUser))
+          goToMetalakeListPage()
```
LauraXia123, line 107: `basicAuthToken` & `basicAuthUser` These two variables are not declared or defined anywhere.

lasdf1234:
Another issue is, In order to avoid certain boundary value exceptions, could you modify the following code?:
Coluld you modify IdpUserGroupManager?
public IdpUser authenticate(String username, String password) {
  try {
    IdpUser user = USER_SERVICE.getIdpUser(username);
    if (user.passwordHash() == null || !passwordHasher.verify(password, user.passwordHash())) {
      throw new UnauthorizedException("Invalid username or password", ...);
    }
    return user;
  } catch (NotFoundException e) {
    throw new UnauthorizedException("Invalid username or password", ...);
  }
}

Could you use Exception instead of NotFoundException?

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
```diff
@@ -41,15 +41,14 @@
 import org.apache.gravitino.utils.PrincipalUtils;
 
 public class AuthenticationFilter implements Filter {
-
-  private final List<Authenticator> filterAuthenticators;
-
```
lasdf1234, line 46: Do not change the position of filterAuthenticators.

lasdf1234:
@Octavi00 The modified Java code has no logical issues.
Don't forget to modify the IdpUserGroupManager class.

Octavi00:
@lasdf1234, I made your requested change to the IdpUserGroupManager class, and I also wanted to ask if my decision to add a const isBasicLoginRequest inside the axios Index.js file was okay. I know a few days ago you mentioned ignoring the browser popups for a trade off in avoiding header responses on failed logins. Is this a good alternative, or should I also avoid this until I find a better fix / altogether?

lasdf1234:
> I made your requested change to the IdpUserGroupManager class, and I also wanted to ask if my decision to add a const isBasicLoginRequest inside the axios Index.js file was okay. I know a few days ago you mentioned ignoring the browser popups for a trade off in avoiding header responses on failed logins. Is this a good alternative, or should I also avoid this until I find a better fix / altogether?

If such changes are made, without modifying the backend code, it is possible to avoid the pop-up window. I think this is feasible.

plugins/idp-basic/src/main/java/org/apache/gravitino/idp/IdpUserGroupManager.java
```diff
@@ -158,7 +158,7 @@ public IdpUser authenticate(String username, String password) {
             "Invalid username or password", AuthConstants.AUTHORIZATION_BASIC_HEADER.trim());
       }
       return user;
-    } catch (NotFoundException e) {
+    } catch (Exception e) {
       throw new UnauthorizedException(
           "Invalid username or password", AuthConstants.AUTHORIZATION_BASIC_HEADER.trim());
     }
```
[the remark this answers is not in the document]
Octavi00: Not applicable, I was directly requested to swap into Exception from previous NFE to strengthen catch.

server-common/src/main/java/org/apache/gravitino/server/authentication/AuthenticationFilter.java
```diff
@@ -113,7 +113,9 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         // to let client to create correct authenticated request.
         // Refer to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate
         for (String challenge : ue.getChallenges()) {
-          resp.setHeader(AuthConstants.HTTP_CHALLENGE_HEADER, challenge);
+          if (!challenge.toLowerCase().startsWith("basic")) {
+            resp.setHeader(AuthConstants.HTTP_CHALLENGE_HEADER, challenge);
+          }
```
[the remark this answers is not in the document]
Octavi00: NA, was advised against using headers during 401 responses

web-v2/web/src/app/login/components/SimpleLogin.js
```diff
@@ -0,0 +1,61 @@
...
+  const router = useRouter()
+  const dispatch = useAppDispatch()
+  const store = useAppSelector(state => state.auth)
+  const [form] = Form.useForm()
+
+  useEffect(() => {
+    if (store.intervalId) {
+      clearIntervalId()
+    }
+  }, [store.intervalId])
```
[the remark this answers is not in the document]
Octavi00: Yeah, since it is a redux this is necessary.

web-v2/web/src/app/login/page.js
```diff
@@ -42,22 +44,45 @@ const LoginContent = () => {
...
+        return <SimpleLogin />
+
+      case 'oauth':
+        if (providerType === null) return null
+        if (providerType === 'oidc') return <OidcLogin />
+
+        return <DefaultLogin />
+
+      default:
+        return null
```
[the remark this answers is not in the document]
Octavi00: I like this idea, but I will instead use a spin from antd just for the sake of a nicer ui

web-v2/web/src/lib/api/auth/index.js
```diff
@@ -44,3 +44,17 @@ export const loginApi = (url, params) => {
...
+
+export const basicLoginApi = basicToken => {
+  return defHttp.get(
+    {
+      url: '/api/version',
+      headers: {
+        Authorization: basicToken,
+        Accept: 'application/vnd.gravitino.v1+json',
+        'Content-Type': 'application/json'
+      }
```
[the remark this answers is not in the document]
Octavi00: Irrelevant, im no longer implementing headers for 401 responses, and this was included to do just that. Content type shouldnt cause any issues, so i'd rather leave it than remove unless specifically requested

web-v2/web/src/lib/utils/axios/index.js
```diff
@@ -244,7 +253,7 @@ const transform = {
     }
 
     try {
-      if (code === 'ECONNABORTED' && message.indexOf('timeout') !== -1) {
+      if (code === 'ECONNABORTED' && message?.indexOf('timeout')) {
```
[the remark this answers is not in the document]
Octavi00: This was a JS oversight on my part. It's definitely correct to change, otherwise will always be true unless 0.

Octavi00:
> If such changes are made, without modifying the backend code, it is possible to avoid the pop-up window. I think this is feasible.

Great, I had already implemented such changes, and they have proven to be successful on my end.

web-v2/web/src/app/login/components/SimpleLogin.js
```diff
@@ -0,0 +1,61 @@
...
+  const router = useRouter()
+  const dispatch = useAppDispatch()
+  const store = useAppSelector(state => state.auth)
+  const [form] = Form.useForm()
+
+  useEffect(() => {
+    if (store.intervalId) {
+      dispatch(clearIntervalId())
+    }
+  }, [dispatch, store.intervalId])
```
LauraXia123: intervalId is always null when SimpleLogin is mounted

web-v2/web/src/lib/store/auth/index.js
```diff
@@ -98,6 +98,36 @@ export const loginAction = createAsyncThunk('auth/loginAction', async ({ params,
   return { token: access_token, expired: expires_in }
 })
 
+export const basicLoginAction = createAsyncThunk(
+  'auth/basicLoginAction',
+  async ({ username, password, router }, { dispatch }) => {
+    const basicToken = `Basic ${btoa(`${username}:${password}`)}`
+
+    const [err, res] = await to(basicLoginApi(basicToken))
```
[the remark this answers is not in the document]
Octavi00: @LauraXia123 I have started working on a test file, but I wanted to ask if I was dropping it in the right spot? Basically oidc test file was placed right in the provider folder alongside it, but since basicLoginAction is simply a function, would it make sense to write a index.test.js file in the store/auth folder? I get that the loginAction doesn't particularly need a test file because it's handled by the oidc provider test, but what about the simple auth case? Would that also need a test?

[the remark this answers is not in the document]
LauraXia123: Leave it for now

LauraXia123:
@Octavi00 Thank for your contribution, LGTM

LauraXia123 (approved):

PR #11816, opened 2026-06-29
[#10172] fix(server): avoid catch-path request dereference masking or…

ethannguyen128:
## What does this PR do?

This PR Fixes a silent exception-masking bug across 15 REST handler classes in the server/module. Each affected handler logged request.get*() inside its catch (Exception e) block without first checking whether request was null. When a null request body caused a primary exception, the catch path's dereference threw a second NullPointerException that replaced the original failure in the stack trace which makes the real root cause invisible to operators and impossible to diagnose from logs alone. The fix precomputes a null-safe name string before the try block so the catch path always has something safe to log, and the original exception is always the one that reaches ExceptionHandlers and the client.

## Why was this PR needed?

Issue #10172 identified that sending a null or empty request body to any of the affected endpoints would produce a misleading NullPointerException response pointing at the catch block and not at the actual failure site. For example, a deserialization failure in CatalogOperations.createCatalog would be silently swallowed, and the client would receive an NPE originating from request.getName() in the catch path instead of the real error. This made triage significantly harder because the original exception context was destroyed before it could be logged or marshalled. Precomputing the identifier is the minimal, zero-API-change fix: it keeps all ExceptionHandlers routing intact and doesn't alter any response contracts.

## What are the relevant issue numbers?

 Number #10172

### DIFF
--- a/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java
+++ b/server/src/main/java/org/apache/gravitino/server/web/rest/CatalogOperations.java
@@ -141,6 +141,7 @@ public class CatalogOperations {
       @PathParam("metalake") @AuthorizationMetadata(type = Entity.EntityType.METALAKE)
           String metalake,
       CatalogCreateRequest request) {
+    String catalogName = request != null ? request.getName() : "unknown";
     LOG.info("Received create catalog request for metalake: {}", metalake);
     try {
       return Utils.doAs(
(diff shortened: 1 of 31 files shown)


ethannguyen128:
Hi @justinmclean  — this is my first contribution to this project. I've fixed #10172 by by precomputing a null-safe identifier before each try block in the affected REST handlers, so that when a null request body triggers an exception, the catch path no longer masks the original failure with a secondary NullPointerException. Would appreciate a review when you have time!

PR #11843, opened 2026-07-01
[#11836] fix(server): Return visible service admins as list

### LINKED ISSUE #11836: [Bug report] Inconsistent return type of visibleConfigs field in /configs API
**Describe what's wrong:**
When `gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins` is set in the configuration file, the `/configs` API returns the `visibleConfigs` field as a concatenated string (e.g., `"gravitino.authorization.serviceAdmins"`). When `visibleConfigs` is not configured, the field returns an empty array `[]`. This results in inconsistent behavior that may cause client‑side parsing errors.

**Error message / stack trace:**
No error message, but the API response format is inconsistent.

**How to reproduce:**
1. Set `gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins` in the configuration file.
2. Start the Gravitino server.
3. Call the `/configs` API.
4. Observe that the `visibleConfigs` field value is a string.
5. Remove (or comment out) the `visibleConfigs` setting in the configuration file.
6. Restart the server and call the `/configs` API again.
7. Observe that the `visibleConfigs` field value is an array.

**Additional context:**
The issue may stem from the configuration parsing logic: `Configs.VISIBLE_CONFIGS` is defined as `stringConf().toSequence()`, which is expected to return a list. However, when the configured value contains only a single item, it might be incorrectly parsed as a string instead of a single‑element list.

Looking at the code in `ConfigServlet.java` line 79-86:
```java
List<String> visibleConfigs = serverConfig.get(Configs.VISIBLE_CONFIGS);
for (String config : visibleConfigs) {
    String configValue = serverConfig.getRawString(config);
    if (configValue != null) {
        configs.put(config, configValue);
    }
}
```
The issue might be in how `serverConfig.get(Configs.VISIBLE_CONFIGS)` parses the configuration value when it's a single string vs. when it's not set.

roryqi:

This PR updates the config servlet to return `gravitino.authorization.serviceAdmins` as a list when it is exposed through `gravitino.server.visibleConfigs`.

It also adds regression tests for:
- a single configured service admin returned as a JSON array
- `visibleConfigs` including service admins while the service admins value is not configured

When `gravitino.server.visibleConfigs=gravitino.authorization.serviceAdmins` is configured, `/configs` currently returns `gravitino.authorization.serviceAdmins` as a raw string for a single value. This is inconsistent with the typed config value and may break clients expecting an array.

Fix: #11836

Yes. `/configs` now returns `gravitino.authorization.serviceAdmins` as a JSON array when it is exposed by `gravitino.server.visibleConfigs`.

- `./gradlew :server:test --tests org.apache.gravitino.server.web.TestConfigServlet`
- `git diff --check`

### DIFF
--- a/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
+++ b/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
@@ -107,12 +107,12 @@ public class TestConfigServlet {
     ServerConfig serverConfig = new ServerConfig(false);
     Properties properties = new Properties();
     properties.setProperty(Configs.VISIBLE_CONFIGS.getKey(), Configs.SERVICE_ADMINS.getKey());
-    properties.setProperty(Configs.SERVICE_ADMINS.getKey(), "admin1");
+    properties.setProperty(Configs.SERVICE_ADMINS.getKey(), "admin1,admin2");
     serverConfig.loadFromProperties(properties);
 
     Map<String, Object> configs = fetchConfigs(serverConfig);
     Assertions.assertEquals(
-        Lists.newArrayList("admin1"), configs.get(Configs.SERVICE_ADMINS.getKey()));
+        Lists.newArrayList("admin1", "admin2"), configs.get(Configs.SERVICE_ADMINS.getKey()));
   }
 
   @Test
--- a/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
+++ b/server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
@@ -30,6 +30,7 @@ import com.google.common.collect.Lists;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.Map;
+import java.util.Properties;
 import javax.servlet.http.HttpServletResponse;
 import org.apache.gravitino.Configs;
 import org.apache.gravitino.auth.AuthenticatorType;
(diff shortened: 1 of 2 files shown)


roryqi:
@yuqi1129 Could u help me review this pull request?

server/src/test/java/org/apache/gravitino/server/web/TestConfigServlet.java
```diff
@@ -101,6 +102,44 @@ public void testConfigServletWithVisibleConfigs() throws Exception {
     Assertions.assertEquals("test", configs.get(customConfig.getKey()));
   }
 
+  @Test
+  public void testConfigServletWithVisibleServiceAdminsFromProperties() throws Exception {
+    ServerConfig serverConfig = new ServerConfig(false);
+    Properties properties = new Properties();
+    properties.setProperty(Configs.VISIBLE_CONFIGS.getKey(), Configs.SERVICE_ADMINS.getKey());
+    properties.setProperty(Configs.SERVICE_ADMINS.getKey(), "admin1");
```
yuqi1129: Can we set the value `admin1, admin2` to verify it works with an array whose size is greater than 1?

roryqi: Updated the properties-based test to use two service admins and assert both values are returned as a list.

yuqi1129 (approved):

PR #11877, opened 2026-07-01
[#11829] fix(doris): support replication allocation property

### LINKED ISSUE #11829: [Bug report]jdbc-doris catalog does not support `replication_allocation` table property (Doris 2.1+ resource tag)
### Version

1.1.0 branch

## Describe what's wrong

When creating a table through the Gravitino `jdbc-doris` catalog, passing
`replication_allocation` as a table property (e.g. `"replication_allocation" = "tag.location.default: 1"`)
fails with `Unknown properties: {replication_allocation=...}`, even though the same
DDL works fine when executed directly against Doris (via a MySQL client on port 9030).

The JDBC catalog appears to be unaware of the `replication_allocation` property
introduced in Doris 2.1, and instead force-injects the legacy `replication_num`
property — which then conflicts with the user-supplied `replication_allocation`
on the Doris FE side.

**Expected behavior:** The `jdbc-doris` catalog should either:
- (a) accept `replication_allocation` and forward it to Doris, OR
- (b) detect it and transparently downgrade to `replication_num` (with a warning)
- (c) reject it early with a clear validation error message instead of letting the
  raw `Unknown properties` error bubble up from Doris FE.

**Root cause:** Two pieces of evidence in the `catalog-jdbc-doris` module point to it.

### (1) The property whitelist only registers `replication_num`

`catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/DorisTablePropertiesMetadata.java`

```java
public class DorisTablePropertiesMetadata extends JdbcTablePropertiesMetadata {

  public static final String REPLICATION_FACTOR = "replication_num";
  public static final int DEFAULT_REPLICATION_FACTOR = 1;
  public static final int DEFAULT_REPLICATION_FACTOR_IN_SERVER_SIDE = 3;

  private static final Map<String, PropertyEntry<?>> PROPERTIES_METADATA;

  static {
    List<PropertyEntry<?>> propertyEntries = new ArrayList<>();
    propertyEntries.add(
        PropertyEntry.integerOptionalPropertyEntry(
            REPLICATION_FACTOR,                     // <-- only "replication_num"
            "The number of replications for the table. ...",
            false /* immutable */,
            DEFAULT_REPLICATION_FACTOR,
            false /* hidden */));
    // ... no entry for "replication_allocation" anywhere

    PROPERTIES_METADATA = Maps.uniqueIndex(propertyEntries, PropertyEntry::getName);
  }
}
```

`replication_allocation` is not part of the registered property entries, so it
will not be validated/typed by the catalog and is passed through to the
underlying JDBC driver unchanged.

### (2) `DorisTableOperations` force-injects `replication_num` during `createTable`

`catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java`

```java
// If the backend server is less than DEFAULT_REPLICATION_FACTOR_IN_SERVER_SIDE (3), we need to
// set the property 'replication_num' to 1 explicitly.
if (!resultMap.containsKey(REPLICATION_FACTOR)) {
  String query = "show backends";
  try (Connection connection = dataSource.getConnection(); ...) {
    ...
    if (backendCount < DEFAULT_REPLICATION_FACTOR_IN_SERVER_SIDE) {
      resultMap.put(
          REPLICATION_FACTOR,                                          // = "replication_num"
          DORIS_TABLE_PROPERTIES_META
              .propertyEntries()
              .get(REPLICATION_FACTOR)
              .getDefaultValue()
              .toString());
    }
  } catch (Exception e) {
    throw new RuntimeException("Failed to get the number of backend servers", e);
  }
}
```

So the generated DDL ends up containing both `replication_num` (auto-injected)
and the user-supplied `replication_allocation`, which Doris FE rejects because
the two are mutually exclusive.

## Error message and/or stacktrace

**HTTP response (Gravitino REST server):**

```json
{
  "code": 500,
  "message": "Failed to operate object [test_repl] operation [CREATE] under [...], reason [errCode = 2, detailMessage = Unknown properties: {replication_allocation=tag.location.default: 1}]"
}
```

**Server-side stack trace:**

```
java.sql.SQLException: errCode = 2, detailMessage = Unknown properties: {replication_allocation=tag.location.default: 1}
  at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:130)
  at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
  at com.mysql.cj.jdbc.StatementImpl.executeUpdateInternal(StatementImpl.java:1335)
  ...
  at org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils.executeUpdate(JdbcConnectorUtils.java:42)
  at org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations.create(JdbcTableOperations.java:111)
  at org.apache.gravitino.catalog.jdbc.JdbcCatalogOperations.createTable(JdbcCatalogOperations.java:478)
  at org.apache.gravitino.catalog.doris.converter.DorisExceptionConverter.toGravitinoException(DorisExceptionConverter.java:132)
  ...
  at org.apache.gravitino.server.web.rest.TableOperations.createTable(TableOperations.java:133)
```

## How to reproduce

1. Start a Gravitino server with a `jdbc-doris` catalog pointing to a Doris 2.1+ cluster
2. Create a metalake / catalog / schema as usual
3. Issue a `createTable` request with the following body:

```json
{
  "name": "test_repl",
  "columns": [
    { "name": "id", "dataType": "integer", "nullable": false },
    { "name": "name", "dataType": "string", "nullable": true }
  ],
  "distribution": { "strategy": "HASH", "fields": ["id"], "number": 3 },
  "properties": {
    "replication_allocation": "tag.location.default: 1"
  }
}
```

4. The same DDL, executed directly via `mysql` client, succeeds:

```sql
CREATE TABLE test.t (
  id INT,
  name VARCHAR(64)
)
DUPLICATE KEY(id)
DISTRIBUTED BY HASH(id) BUCKETS 3
PROPERTIES ("replication_allocation" = "tag.location.default: 1");
```

5. As a control, swapping `replication_allocation` for `replication_num="1"` against
   the same endpoint succeeds — proving the bridge / request format is fine and
   the failure is specific to the `replication_allocation` key.

## Additional context

**Environment:**
- Doris cluster: 1 BE (single-backend test environment)
- Doris version: 2.1+ (returns `5.7.99` to MySQL handshake)
- Gravitino: 1.1.0
- MySQL JDBC driver: `com.mysql.cj.jdbc.Driver`
- Test JDBC URL: `jdbc:mysql://xxx:9030`

**Related issues:**
- #6722 — `[Improvement] An error occurred when Doris2.1 was creating a table, reporting "Failed to get the number of backend servers"`. This was closed but the underlying Doris 2.1 compatibility gaps were not fully addressed.
- #8278 — `[Bug report] Failed to create table in starrock catalog due to GravitinoRuntimeException: Table replication num should be less than or equal to the number of available backends`. Same `replication_num` force-injection pattern, suggesting the StarRocks catalog has the same issue.

A search for `replication_allocation` in the Gravitino issue tracker returns 0
matching results — this is a blind spot, not a known-and-deferred issue.

**Suggested fix:** Two parts, plus a bonus.

### A. Catalog property metadata (`DorisTablePropertiesMetadata`)

Add `replication_allocation` as a recognised property. Register it as a
`stringOptionalPropertyEntry` and pass it through verbatim when present:

```java
public static final String REPLICATION_ALLOCATION = "replication_allocation";

propertyEntries.add(
    PropertyEntry.stringOptionalPropertyEntry(
        REPLICATION_ALLOCATION,
        "Doris 2.1+ resource-tag based replication allocation, "
            + "e.g. \"tag.location.default: 3\" or \"tag.zone1: 2, tag.zone2: 1\"",
        false /* immutable */,
        null /* default value */,
        false /* hidden */));
```

### B. Table operations (`DorisTableOperations`)

Treat `replication_allocation` and `replication_num` as **mutually exclusive** and
skip the `replication_num` force-injection logic if the user has already supplied
`replication_allocation`:

```java
if (!resultMap.containsKey(REPLICATION_FACTOR)
    && !resultMap.containsKey(REPLICATION_ALLOCATION)) {   // <-- new check
  // ... existing show-backends / force-inject logic
}
```

Either property alone is acceptable; both at the same time is not.

hutiefang76:
### What changed

This PR lets the JDBC Doris catalog accept the Doris 2.1+ `replication_allocation` table property.

It also keeps the existing single-BE fallback behavior for old-style `replication_num`, but skips that automatic `replication_num=1` injection when the user has already provided `replication_allocation`. Doris treats those two properties as mutually exclusive, so adding both can make table creation fail.

### Why

`replication_allocation` is valid Doris table syntax, but Gravitino did not register it in the Doris table property metadata. In single-backend environments, Gravitino could also add `replication_num` automatically, which conflicts with the user-supplied allocation policy.

### Tests

```bash
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :catalogs:catalog-jdbc-doris:test --tests org.apache.gravitino.catalog.doris.TestDorisCatalog --tests org.apache.gravitino.catalog.doris.operation.TestDorisTableOperationsSqlGeneration -PskipITs
git diff --check
```

Closes #11829


yuqi1129:
@hutiefang76 
Can you add some ITs to cover your changes?

hutiefang76:
Added a Doris integration test covering this path.

The test creates a table through Gravitino with `replication_allocation=tag.location.default: 1`, reloads it from Doris, verifies that the allocation is preserved, and verifies that no conflicting `replication_num` property is injected.

I ran the targeted Docker-backed IT against a real Doris cluster after rebasing onto the latest `main`:

```
CatalogDorisIT.testCreateTableWithReplicationAllocation
1 test, 0 failures, 0 errors
```

The related Doris unit tests and Spotless checks also pass.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -180,7 +181,8 @@ Map<String, String> appendNecessaryProperties(Map<String, String> properties) {
 
     // If the backend server is less than DEFAULT_REPLICATION_FACTOR_IN_SERVER_SIDE (3), we need to
     // set the property 'replication_num' to 1 explicitly.
-    if (!resultMap.containsKey(REPLICATION_FACTOR)) {
+    if (!resultMap.containsKey(REPLICATION_FACTOR)
+        && !resultMap.containsKey(REPLICATION_ALLOCATION)) {
```
yuqi1129, line 193: What if the user set `replication_num` and `replication_allocation` at the same time?

hutiefang76: Doris does not define a successful precedence for this combination: it consumes `replication_num` and later rejects the remaining `replication_allocation` as an unknown property. I updated Gravitino to fail fast with a clear error when both are present, rather than silently choosing one. The change includes a regression test for the conflicting pair and documents the mutual-exclusion rule. I also rebased the PR onto the current `main`; the full `catalog-jdbc-doris` non-Docker test suite and Spotless check pass locally.

yuqi1129:
LGTM, let's wait for the CI to pass.

yuqi1129 (approved):

PR #11878, opened 2026-07-01
[#11841] fix ClickHouse partition expression round trip

### LINKED ISSUE #11841: [Bug report] ClickHouse partition expression round-trip broken for day transforms and nested functions
### Version

main branch

### Describe what's wrong

Two problems in the ClickHouse partition expression handling:

**1. Day transform maps to `toDate()` instead of `toYYYYMMDD()`.**

`toPartitionExpression` outputs `toDate(col)` for the `day` transform. `toDate()` returns a `Date` type, but ClickHouse's integer day-partitioned tables use `toYYYYMMDD()` which returns `UInt32` (e.g. `20230622`). This mismatch means `loadTable()` + `createTable()` does not produce an equivalent table — the partition expression type differs from what ClickHouse actually stores.

**2. Nested partition function calls are not handled.**

`parsePartitionExpression` uses `normalizeIdentifier()` to strip backticks from the inner argument of `toYear()`, `toYYYYMM()`, and `toDate()`. This fails on nested expressions like `toYYYYMM(toDate(dt))` because `normalizeIdentifier("toDate(dt)")` returns `"toDate(dt)"`, which contains parentheses and is not a simple identifier. The function should recursively peel wrapper functions to find the innermost column name.

For problem 1: no error, but `loadTable` followed by `createTable` produces a table with `PARTITION BY toDate(col)` instead of `PARTITION BY toYYYYMMDD(col)`, meaning the round-trip does not preserve the original partition semantics.

For problem 2: `IllegalArgumentException: Only simple identifier is supported` when loading a table whose partition key contains nested function wrappers.

### How to reproduce

1. Create a ClickHouse table with `PARTITION BY toYYYYMMDD(dt)`.
2. Load it via Gravitino → partition is parsed as `day("dt")`.
3. Create a new table with the same partition transform → Gravitino generates `PARTITION BY toDate(dt)` instead of `toYYYYMMDD(dt)`.

For nested functions:
1. Create a ClickHouse table with `PARTITION BY toYYYYMM(toDate(event_time))`.
2. Load it via Gravitino → `parsePartitionExpression` fails on `toDate(event_time)`.

### Additional context

Both fixes are in `ClickHouseTableSqlUtils`:
- Change `toPartitionExpression` to output `toYYYYMMDD` for the `day` transform.
- Add `extractInnermostField` to recursively unwrap nested date/time functions (with max depth guard).
- Add `TO_DAY_PATTERN` so `toYYYYMMDD(...)` is recognized during parsing.

hutiefang76:

Fix ClickHouse partition expression round-tripping for day transforms and nested date wrappers:

- format Gravitino day transforms as `toYYYYMMDD(column)` instead of `toDate(column)`
- parse `toYYYYMMDD(column)` back to a day transform
- unwrap a supported `toDate(...)` wrapper when parsing expressions such as `toYYYYMM(toDate(event_time))`
- update the ClickHouse catalog documentation for supported partition functions

Fixes #11841.

Without this change, a ClickHouse partition expression like `toYYYYMM(toDate(event_time))` is parsed with `toDate(event_time)` as the field name. Also, day transforms are formatted as `toDate(...)`, which does not round-trip cleanly with the ClickHouse day partition form described in the issue.

Yes. ClickHouse day partition transforms are now generated as `toYYYYMMDD(column)`. Existing `toDate(column)` partition expressions are still accepted when reading metadata.

- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test --tests org.apache.gravitino.catalog.clickhouse.operations.TestClickHouseTableOperationsPartitioning -PskipITs`
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:spotlessCheck -PskipITs`
- `git diff --check`

### DIFF
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableSqlUtils.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableSqlUtils.java
@@ -35,6 +35,8 @@ final class ClickHouseTableSqlUtils {
 
   private static final Pattern TO_DATE_PATTERN =
       Pattern.compile("toDate\\((.+)\\)", Pattern.CASE_INSENSITIVE);
+  private static final Pattern TO_DAY_PATTERN =
+      Pattern.compile("toYYYYMMDD\\((.+)\\)", Pattern.CASE_INSENSITIVE);
   private static final Pattern TO_YEAR_PATTERN =
       Pattern.compile("toYear\\((.+)\\)", Pattern.CASE_INSENSITIVE);
   private static final Pattern TO_MONTH_PATTERN =
@@ -76,7 +78,7 @@ final class ClickHouseTableSqlUtils {
           .formatted(quoteIdentifier(partitionFieldName(transform)));
       case Transforms.NAME_OF_MONTH -> "toYYYYMM(%s)"
           .formatted(quoteIdentifier(partitionFieldName(transform)));
-      case Transforms.NAME_OF_DAY -> "toDate(%s)"
+      case Transforms.NAME_OF_DAY -> "toYYYYMMDD(%s)"
           .formatted(quoteIdentifier(partitionFieldName(transform)));
       default -> throw new IllegalArgumentException(
           "Unsupported partition transform: " + transform.name());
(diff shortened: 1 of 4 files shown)


hutiefang76:
I noticed #11842 is already covering the same issue and direction. Closing this PR to avoid duplicate review work. Thanks.

PR #11883, opened 2026-07-02
[MINOR] fix(aws): Support MinIO role ARNs in IRSA credential vending

yuqi1129:

- Accept `arn:minio:` role ARNs in `AwsIrsaCredentialGenerator` when a custom STS endpoint is configured (in addition to `arn:aws` ARNs).
- Fall back to the default web identity token source when the `SOURCE` property is present but blank.
- Fix table formatting in `docs/security/credential-vending.md`.

IRSA credential vending previously rejected MinIO-style role ARNs, preventing its use against MinIO STS endpoints. A blank source property should behave the same as an unset one.

No.

Existing unit tests.


diqiu50 (approved):
LGTM

PR #11886, opened 2026-07-02
[#11881] fix(clickhouse): distinguish MATERIALIZED/ALIAS from DEFAULT via default_kind

### LINKED ISSUE #11881: [Bug report] ClickHouse catalog: column default value handling issues with MATERIALIZED and isExpression
### Version

main branch

### Describe what's wrong

ClickHouse distinguishes column default value kinds via `system.columns.default_kind`:
DEFAULT, MATERIALIZED, and ALIAS. The ClickHouse catalog does not fetch this field, so
MATERIALIZED and ALIAS columns are indistinguishable from DEFAULT columns on round-trip.

Additionally, the ClickHouse JDBC driver's `getColumns()` implementation hardcodes `'NO'` as
`IS_GENERATEDCOLUMN` in its SQL query (verified in clickhouse-jdbc 0.7.1 source), independent
of the actual column definition. This is the same class of driver limitation as the one already
worked around in `getIndexes()` (clickhouse-java#1625). The code at
`ClickHouseColumnDefaultValueConverter.java`:109 has a TODO acknowledging this limitation, and
relies solely on the hardcoded `IS_GENERATEDCOLUMN` to determine `isExpression`, causing
expression-based columns to be treated as literal values.

N/A. Found via code review. No runtime error — MATERIALIZED/ALIAS columns are silently
downgraded to DEFAULT.

### How to reproduce

1. Create a table in ClickHouse with all three default kinds:
   ```sql
   CREATE TABLE test_default_kind (
       id Int64,
       created_date Date DEFAULT today(),
       computed_date Date MATERIALIZED today(),
       alias_date Date ALIAS today()
   ) ENGINE=MergeTree ORDER BY id;
   ```
2. Verify ClickHouse stores the distinction:
   ```sql
   SELECT name, default_kind, default_expression
   FROM system.columns
   WHERE database = 'default' AND table = 'test_default_kind';
   ```
   Returns: `created_date` → DEFAULT, `computed_date` → MATERIALIZED, `alias_date` → ALIAS
3. Load the table via Gravitino REST API and inspect column `defaultValue` fields
4. Expected: columns should have distinct default value kinds (DEFAULT / MATERIALIZED / ALIAS)
5. Actual: all three columns have identical `defaultValue: {type: "unparsed", unparsedExpression: "today()"}`
   — no way to distinguish MATERIALIZED or ALIAS from DEFAULT

### Additional context

- `ClickHouseColumnDefaultValueConverter.java`:109 — TODO acknowledging the hardcoded IS_GENERATEDCOLUMN limitation
- `JdbcTableOperations.getBasicJdbcColumnInfo()` — uses `IS_GENERATEDCOLUMN` only, never queries `default_kind`
- `ClickHouseTableOperations` does not override `getColumns()` to query `system.columns`
- `getIndexes()` in `ClickHouseTableOperations.java` — existing workaround for a similar driver limitation (clickhouse-java#1625), establishing the pattern of querying system tables directly
- Fix: override `getColumns()` in `ClickHouseTableOperations` to query `system.columns` (including `default_kind`), following the same pattern as `getIndexes()` which already bypasses the JDBC driver. Then pass `default_kind` to the default value converter to correctly distinguish MATERIALIZED and ALIAS from DEFAULT.

jiangxt2:

Fix column default value handling for MATERIALIZED and ALIAS columns in the ClickHouse catalog.

ClickHouse distinguishes column default value kinds via `system.columns.default_kind` (DEFAULT,
MATERIALIZED, ALIAS). The JDBC driver`getColumns()` hardcodes `'NO'` as `IS_GENERATEDCOLUMN`
(clickhouse-java#1625 pattern), so the catalog cannot distinguish these kinds through standard
JDBC metadata.

Changes:
- Added `getDefaultKinds()` method that queries `system.columns` for `default_kind`, following
  the same pattern as `getIndexes()` which already bypasses the JDBC driver for similar limitations
- `load()` now calls `getDefaultKinds()` and uses the result to correct column metadata inline
  during the column loading loop, without exposing instance state
- Fixed `ClickHouseContainer.getJdbcUrl()` to use `localhost` + `getMappedPort()` instead of
  `getContainerIpAddress()`, consistent with how `DorisContainer` handles macOS Docker Desktop
  compatibility

Thread safety: `ClickHouseTableOperations` is a shared singleton (cached by CatalogManager's
Caffeine cache, accessed concurrently via JettyServer's QueuedThreadPool). The `default_kinds`
map is a local variable in `load()`, not an instance field, making the method reentrant.

Fix: #11881

MATERIALIZED and ALIAS columns were indistinguishable from DEFAULT columns on round-trip,
because the catalog never fetched `default_kind` from ClickHouse system tables.

Yes. MATERIALIZED and ALIAS columns now have correct default value kinds in Gravitino metadata,
enabling consumers to distinguish them from DEFAULT columns.

**Unit tests** (`TestClickHouseTableOperations`):
- `testLoadTableWithMaterializedAndAliasColumns`: creates table with all three default kinds
  via raw SQL, verifies all three columns return `UnparsedExpression("today()")` via their
  respective code paths

**Integration tests** (`CatalogClickHouseIT`):
- `testMATERIALIZEDAndALIASColumnDefaultKinds`: end-to-end test through Gravitino API covering
  DEFAULT/MATERIALIZED/ALIAS × multiple types (Date, Int64, String) × Nullable columns

Both test classes require `-PskipDockerTests=false` to run.

### DIFF
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
@@ -57,6 +57,7 @@ import org.apache.gravitino.catalog.clickhouse.ClickHouseTablePropertiesMetadata
 import org.apache.gravitino.catalog.clickhouse.ClickHouseTablePropertiesMetadata.ENGINE;
 import org.apache.gravitino.catalog.jdbc.JdbcColumn;
 import org.apache.gravitino.catalog.jdbc.JdbcTable;
+import org.apache.gravitino.catalog.jdbc.converter.JdbcTypeConverter;
 import org.apache.gravitino.catalog.jdbc.operation.JdbcTableOperations;
 import org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils;
 import org.apache.gravitino.exceptions.NoSuchTableException;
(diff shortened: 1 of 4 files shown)


yuqi1129:
I reviewed it and had the following concerns:
- Can we fully recognize different column default values
- Does Gravitino support creating a table with a MATERIZED/ALIAS column value?

jiangxt2:
> * Can we fully recognize different column default values

Yes — the getDefaultKinds() query to system.columns.default_kind covers all three kinds (DEFAULT, MATERIALIZED, ALIAS). For MATERIALIZED/ALIAS columns, we call toGravitino(typeBean, columnDef, true, nullable) with isExpression=true, which forces creation of an UnparsedExpression instead of attempting literal parsing. For DEFAULT columns, the existing code path through getBasicJdbcColumnInfo() is unchanged.

There is a fourth kind EPHEMERAL (ClickHouse 21.12+) that is not explicitly handled, but such columns are excluded from SELECT * and are extremely rare in practice. If needed, we can add it in a follow-up.

jiangxt2:
> * Does Gravitino support creating a table with a MATERIZED/ALIAS column value?

No — this PR only addresses the load path. Creating such columns would require API-level changes (the Column interface has no "default value kind" concept, and JdbcTableOperations.appendDefaultValue() always emits the DEFAULT keyword). This feels like a separate feature that needs its own API design discussion.

yuqi1129:
> > * Does Gravitino support creating a table with a MATERIZED/ALIAS column value?
> 
> No — this PR only addresses the load path. Creating such columns would require API-level changes (the Column interface has no "default value kind" concept, and JdbcTableOperations.appendDefaultValue() always emits the DEFAULT keyword). This feels like a separate feature that needs its own API design discussion.

I see. This should only guarantee that Gravitino can load tables with MATERIZED/ALIAS columns created before via Gravitino, and can't create such a table via the Gravitino API.

yuqi1129 (approved):

PR #11888, opened 2026-07-02
[#10474] docs: Add concurrency-control design for multi-node Gravitino (TreeLock)

yuqi1129:

Adds a design document (`design-docs/treelock-necessity-and-concurrency-design.md`, with rendered diagrams) that:

- Explains what `TreeLock` protects today and why it is per-JVM, so it stops protecting anything across servers under HA.
- Analyzes that every catalog metadata write touches two stores (the external catalog and the Gravitino store) with no shared transaction; for external-backed catalogs the external system is the source of truth and the Gravitino store self-fixes on read (import).
- Compares two directions — **(1)** remove the lock and move correctness into the shared database (OCC + conditional inserts) vs **(2)** a cross-node lock — on correctness, performance, maintainability, and operational cost, and reviews how comparable systems solve the same dual-write problem.
- Concludes with Direction 1 (remove/shrink TreeLock to a small in-process lock) and a phased plan.

Rendered diagrams are embedded under `design-docs/images/treelock-*.png`.

#10474 asks for a concrete proposal on TreeLock's limitations for HA. This document provides the analysis and a recommended direction before any code change.

Related to #10474

No. Documentation only.

Not applicable — documentation only. Diagrams were rendered from their `.mmd` sources with mermaid-cli and embedded as PNGs.


yuqi1129:
duplicated.

PR #11896, opened 2026-07-03
[#11893] fix(lance): Add external table guard to deregisterTable

### LINKED ISSUE #11893: [Bug] Lance deregisterTable relies on coincidence (all tables external) for correctness
### Version

main branch

### Describe what's wrong

`GravitinoLanceTableOperations.deregisterTable()` (line 284) has a TODO: "Support real deregister API." The current implementation calls `catalog.asTableCatalog().dropTable()`, which is the same path used by `dropTable`.

The interface contract (`LanceTableOperations.java:94-98`) defines:
- `deregisterTable`: "It will not delete the underlying lance data."
- `dropTable`: "It will delete the underlying lance data."

The downstream `LanceTableOperations.dropTable()` (catalog layer, lines 339-360) branches on `PROPERTY_EXTERNAL`:
- External tables: metadata removal only (correct for deregister)
- Managed tables: metadata removal + `dropLanceDataset()` (incorrect for deregister)

Since all Lance REST-created tables are forced to `PROPERTY_EXTERNAL=true` (lines 190, 221, 248), the current deregister accidentally produces correct behavior. But it relies on this implementation coincidence rather than explicit semantics.

### Why this matters

If a managed table path is ever introduced (e.g. Lance tables created via Spark catalog, or a future change to allow managed Lance tables), `deregisterTable` would silently delete physical data — a data loss bug that's hard to diagnose because the code "looks correct" at the call site.

### Proposed fix

The `TableCatalog` interface exposes only `dropTable` and `purgeTable` — no metadata-only path. Adding `deregisterTable` to the interface would require an API-level change and maintainer discussion, which is out of scope for a lightweight fix.

The pragmatic solution is a fast-failing external assertion: refuse to deregister non-external tables rather than risk silent data loss.

In `GravitinoLanceTableOperations.deregisterTable()`, add before the `dropTable` call (line 285):

```java
boolean external = Boolean.parseBoolean(
    properties.getOrDefault(Table.PROPERTY_EXTERNAL, "false"));
if (!external) {
    // deregister must NOT delete underlying data; for managed tables dropTable would,
    // and TableCatalog exposes no metadata-only path. Fail fast instead of silently deleting.
    throw new UnsupportedOperationException(
        "deregisterTable only supports external tables: " + tableId);
}
// External tables: dropTable removes catalog metadata only, preserving Lance data.
boolean result = catalog.asTableCatalog().dropTable(tableIdentifier);
```

This guard is not redundant with the downstream `dropTable` external check — the downstream check decides *how* to drop; this assertion decides *whether* to proceed at all.

### Testing

A unit test stubs `Table.properties()` (via the existing Mockito pattern in `TestGravitinoLanceTableOperations`) to return a map without `PROPERTY_EXTERNAL=true` and asserts `deregisterTable` throws `UnsupportedOperationException`. The external happy path (metadata removed, physical data preserved) is already covered by `LanceRESTServiceIT`.

### Future work

A proper `deregisterTable` on the `TableCatalog` interface (or a dedicated `LanceTableOperations.deregisterTable` exposed through a Lance-specific API) would allow metadata-only removal for any table type. This requires an API-level design discussion with maintainers.

### Additional context

- TODO: `lance/lance-common/.../GravitinoLanceTableOperations.java:284`
- Contract: `lance/lance-common/.../LanceTableOperations.java:94-98`
- Interface: `api/.../rel/TableCatalog.java:297` (`dropTable`), `312` (`purgeTable`)
- dropTable logic: `catalogs/catalog-lakehouse-generic/.../LanceTableOperations.java:339-360`

jiangxt2:

Add a guard in `GravitinoLanceTableOperations.deregisterTable()` that rejects
non-external (managed) tables with `UnsupportedOperationException` before calling
`dropTable`.

The guard reads `PROPERTY_EXTERNAL` from the loaded table properties using the same
`Optional.ofNullable` + `Boolean.parseBoolean` pattern as the downstream
`LanceTableOperations.dropTable()`, ensuring consistent behavior across both layers.

`deregisterTable()` delegates to `dropTable()`, which for managed tables deletes the
underlying Lance dataset — violating the interface contract ("It will not delete the
underlying lance data").

The current code is safe only because all REST-created tables are forced to
`PROPERTY_EXTERNAL=true` (lines 190, 221, 248). This is an implementation coincidence,
not a semantic guarantee. If a managed table path is ever introduced, `deregisterTable`
would silently delete physical data.

The guard converts this implicit assumption into an explicit invariant: refuse
non-external tables rather than risk silent data loss. This mirrors
`HiveCatalogOperations.purgeTable()`, which throws `UnsupportedOperationException` for
external tables — same exception type, opposite direction (purge rejects external
tables; deregister rejects managed tables).

Fix: #11893

`deregisterTable` now throws `UnsupportedOperationException` for managed tables, which
`LanceExceptionMapper` maps to HTTP 406 (Not Acceptable). Clients would receive an
explicit error instead of silent data loss. This path is currently unreachable via the
REST API (all REST tables are external), so no existing clients are affected.

- Unit test `testDeregisterTableRejectsManagedTable` mocks a managed table (empty
  properties) and verifies:
  - `UnsupportedOperationException` is thrown
  - `dropTable` is never called on the catalog
- Existing integration test `LanceRESTServiceIT` covers the external table happy path
  (deregister preserves physical data).

### DIFF
--- a/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
+++ b/lance/lance-common/src/main/java/org/apache/gravitino/lance/common/ops/gravitino/GravitinoLanceTableOperations.java
@@ -281,7 +281,21 @@ public class GravitinoLanceTableOperations implements LanceTableOperations {
           "Table not found: " + tableId, CommonUtil.formatCurrentStackTrace(), tableId);
     }
     Map<String, String> properties = t.properties();
-    // TODO Support real deregister API.
+
+    // Verify the table is external before deregistering. For non-external (managed) tables,
+    // dropTable would delete the underlying Lance data, violating the deregister contract
+    // ("It will not delete the underlying lance data"). The TableCatalog interface does not
+    // expose a metadata-only removal path, so we fail fast rather than risk silent data loss.
+    boolean external =
+        Optional.ofNullable(properties.get(Table.PROPERTY_EXTERNAL))
+            .map(Boolean::parseBoolean)
+            .orElse(false);
+    if (!external) {
+      throw new UnsupportedOperationException(
+          "deregisterTable only supports external tables, but table " + tableId + " is managed");
+    }
+
+    // External tables: dropTable removes catalog metadata only, preserving Lance data.
     boolean result = catalog.asTableCatalog().dropTable(tableIdentifier);
     if (!result) {
       throw new TableNotFoundException(
--- a/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestGravitinoLanceTableOperations.java
+++ b/lance/lance-rest-server/src/test/java/org/apache/gravitino/lance/service/rest/TestGravitinoLanceTableOperations.java
@@ -21,11 +21,17 @@ package org.apache.gravitino.lance.service.rest;
 
 import static org.apache.gravitino.lance.common.utils.LanceConstants.LANCE_TABLE_VERSION;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.apache.gravitino.Catalog;
+import org.apache.gravitino.NameIdentifier;
+import org.apache.gravitino.lance.common.ops.gravitino.GravitinoLanceNamespaceWrapper;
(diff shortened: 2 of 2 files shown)


FANNG1:
LGTM

FANNG1 (approved):

yuqi1129 (approved):

PR #11906, opened 2026-07-06
[#10805] Support bulk access control operations

### LINKED ISSUE #10805: [FEATURE] Support bulk deletion of users and roles in authorization REST APIs
## Background

When an external business system syncs a batch of users and roles into Gravitino's authorization module and the imported data turns out to be incorrect, there is no way to clean them up in bulk. Every entity must be removed individually via a separate HTTP request, which is inefficient and error-prone at scale.

## Current behavior

The existing REST endpoints in `UserOperations` and `RoleOperations` only support single-entity removal via path parameters:

```
DELETE /api/metalakes/{metalake}/users/{user}    // removes one user per call
DELETE /api/metalakes/{metalake}/roles/{role}    // removes one role per call
```

To remove N entities, N sequential or parallel HTTP calls are required. There is no batch endpoint.

## Proposed solution

Add bulk-remove endpoints for both users and roles. The request body pattern already used in `UserAddRequest` / `RoleCreateRequest` (POST handlers) can be followed for consistency.

**Bulk remove users:**

```http
DELETE /api/metalakes/{metalake}/users

{
  "userNames": ["alice", "bob", "charlie"]
}
```

**Bulk remove roles:**

```http
DELETE /api/metalakes/{metalake}/roles

{
  "roleNames": ["role_analyst", "role_viewer"]
}
```

**Suggested response** — mirrors the existing `RemoveResponse` / `DropResponse` pattern but extended for partial results:

```json
{
  "removed": ["alice", "bob"],
  "failed":  [{ "name": "charlie", "reason": "NoSuchUserException" }]
}
```

Authorization should reuse the existing permission expressions:

- Users: `METALAKE::OWNER || METALAKE::MANAGE_USERS`
- Roles: `METALAKE::OWNER || ROLE::OWNER`

The metalake-owner guard (checking whether a user is the metalake owner before removal) that exists in the current single-delete path should also be applied per-entity in the bulk path.

## Additional context

- Files to change: `UserOperations.java`, `RoleOperations.java`, and corresponding DTO / dispatcher interfaces
- A matching Java SDK / Python client method would also be appreciated
- Gravitino version: (please fill in)

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

jarredhj0214:

This PR adds bulk access control APIs for users, groups, and roles under a metalake:

- `POST /api/bulk/metalakes/{metalake}/users/add`
- `POST /api/bulk/metalakes/{metalake}/users/remove`
- `POST /api/bulk/metalakes/{metalake}/groups/add`
- `POST /api/bulk/metalakes/{metalake}/groups/remove`
- `POST /api/bulk/metalakes/{metalake}/roles/add`
- `POST /api/bulk/metalakes/{metalake}/roles/remove`

The implementation includes:

- New REST resources for bulk user, group, and role operations.
- Request DTOs for user names, group names, role names, and bulk role creation.
- A shared bulk operation response with `succeeded` and `failed` results.
- Request validation for empty input and duplicated names.
- OpenAPI documentation and examples for the new APIs.
- Unit tests for request validation and REST behavior.

Currently, users need to call single-resource APIs repeatedly to add or remove multiple users, groups, or roles. These new APIs provide a batch operation interface for access control management and reduce repeated client-side API calls.

This PR follows the behavior discussed in #10805:

- Validate the whole request before processing.
- Reject invalid requests, such as duplicated names, without processing any item.
- After validation succeeds, process each item independently.
- Return successful items in `succeeded` and per-item business failures in `failed`.

Closes #10805.

Yes. This PR introduces new REST APIs for bulk access control operations.

```bash
./gradlew :common:test --tests org.apache.gravitino.dto.requests.TestBulkRequests
./gradlew :server:test --tests org.apache.gravitino.server.web.rest.TestBulkOperations -PskipITs
./gradlew :docs:build


roryqi:
@lasdf1234 Could u help me review this pull request?

lasdf1234:
> @lasdf1234 Could u help me review this pull request?

No problem.

lasdf1234:
Could you add some content to the doc document to explain the usage scenarios of these APIs?

lasdf1234:
Is the bulk interface better when processing data one by one or in batches? Could you please share your research findings?

jarredhj0214:
> Is the bulk interface better when processing data one by one or in batches? Could you please share your research findings?

Thanks for the review. @lasdf1234 

For the first question:

I added a new section named "Bulk Access Control Operations" in `docs/security/access-control.md`.

The new documentation explains:

- Usage scenarios for these APIs, such as team onboarding/offboarding, synchronizing users and groups from an external identity provider, bootstrapping roles for a new environment, and cleaning up obsolete principals.
- Required privileges for each bulk API.
- Request body fields for users, groups, and roles.
- Shell examples for all six bulk APIs.
- The response format with `succeeded` and `failed`.
- The processing semantics: validate the whole request first, then process each item independently after validation succeeds.

For the second question:

I did some research on common batch/bulk API designs. There does not seem to be a single universal rule, but there are several common patterns.

1. Per-item result with partial success

- Microsoft Graph JSON batching returns one top-level batch response, but each individual request has its own status. A top-level 200 only means the batch payload was accepted; clients still need to inspect each item result.
  https://learn.microsoft.com/en-us/graph/json-batching

- Microsoft Graph throttling documentation also says requests in a batch are evaluated individually, and failed requests should be retried individually.
  https://learn.microsoft.com/en-us/graph/throttling

- Elasticsearch Bulk API returns a top-level `errors` flag and an `items` array. Each operation has its own result, and successful operations are not rolled back when other items fail.
  https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-bulk

- Amazon SQS `SendMessageBatch` returns `Successful` and `Failed` lists. AWS explicitly says the batch can contain both successful and unsuccessful actions, so clients should check batch errors even when HTTP status is 200.
  https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html

- Criteo documents a similar model: authentication, authorization, or deserialization errors return 4xx; otherwise the bulk call can return 200 regardless of individual operation success or failure, and clients should inspect the response errors.
  https://developers.criteo.com/marketing-solutions/docs/partial-success-for-bulk-operations

2. Batch request accepted, failed items returned for retry

- Amazon DynamoDB `BatchWriteItem` says each individual put/delete operation is atomic, but the batch as a whole is not. Failed operations are returned as `UnprocessedItems`, and clients should retry those items, usually with exponential backoff.
  https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html

3. Atomic bulk operation

- Some API guidelines recommend atomic bulk operations. For example, Adidas API guidelines say a bulk operation should be atomic and should not commit partial changes if any item fails.
  https://github.com/adidas/api-guidelines/blob/master/rest-api-guidelines/execution/batch-operations.md

For Gravitino access control APIs, I think the current design is closer to the first pattern and fits this feature better:

- We validate the whole request before processing, including empty input, duplicated names, and authorization.
- If request-level validation fails, no item is processed.
- After the request is valid, each user/group/role operation is processed independently.
- The response returns successful items in `succeeded` and per-item business failures in `failed`.

This design avoids partial execution for malformed or unauthorized requests, while still allowing clients to make progress and retry only failed items for business-level failures, such as removing a non-existent user.

The current implementation batches at the API contract level and reuses existing per-item access-control managers internally. This keeps the behavior consistent with existing single-item APIs, including existing authorization checks, events, and cache invalidation behavior. Internal storage-level batching can still be optimized later without changing the public API response contract.

jerryshao:
@jarredhj0214 can you split into several small PRs? The current PR is too large to review.

lasdf1234:
@roryqi Should the "bulk" scenario also consider the scenario where "externalId" is used as the association key?

roryqi:
> @roryqi Should the "bulk" scenario also consider the scenario where "externalId" is used as the association key?

I think yes.

lasdf1234:
> > @roryqi Should the "bulk" scenario also consider the scenario where "externalId" is used as the association key?
> 
> I think yes.

@jarredhj0214 Coluld you split PR? In addition, the implementation of the externalID scenario needs to be considered.

jarredhj0214:
> > > @roryqi Should the "bulk" scenario also consider the scenario where "externalId" is used as the association key?
> > 
> > 
> > I think yes.
> 
> @jarredhj0214 Coluld you split PR? In addition, the implementation of the externalID scenario needs to be considered.

Currently, the existing single user/group/role access-control APIs are name-based and do not support `externalId` as the association key. This PR follows the current single-operation API contract and keeps the bulk APIs name-based as well.

If we want to support `externalId`, I think it should be discussed as a separate API/model design because it may require changes to the user/group identity model, request DTOs, storage lookup, and OpenAPI contract, not only the bulk APIs.

jarredhj0214:
> @jarredhj0214 can you split into several small PRs? The current PR is too large to review.

I have split the original large PR into smaller PRs.

The first split PR will include bulk user operations and the common bulk infrastructure. The bulk group and bulk role operations will be submitted as follow-up PRs after the first one is reviewed/merged, so each PR remains easier to review.

For the `externalId` point: the existing single user/group/role access-control APIs are name-based and do not support `externalId` as the association key. The split PRs will stay aligned with the current name-based API contract. I think `externalId` support may require a separate API/model design and can be discussed as a follow-up.

lasdf1234:
> > @jarredhj0214 can you split into several small PRs? The current PR is too large to review.
> 
> I have split the original large PR into smaller PRs.
> 
> The first split PR will include bulk user operations and the common bulk infrastructure. The bulk group and bulk role operations will be submitted as follow-up PRs after the first one is reviewed/merged, so each PR remains easier to review.
> 
> For the `externalId` point: the existing single user/group/role access-control APIs are name-based and do not support `externalId` as the association key. The split PRs will stay aligned with the current name-based API contract. I think `externalId` support may require a separate API/model design and can be discussed as a follow-up.

You can update the code. The externalId method has already been implemented.

jarredhj0214:
The first split PR has been created:

- #11930

It includes bulk user operations and the common bulk infrastructure. I will submit the bulk group and bulk role operations as follow-up PRs after this one is reviewed/merged.

jarredhj0214:
> > > @jarredhj0214 can you split into several small PRs? The current PR is too large to review.
> > 
> > 
> > I have split the original large PR into smaller PRs.
> > The first split PR will include bulk user operations and the common bulk infrastructure. The bulk group and bulk role operations will be submitted as follow-up PRs after the first one is reviewed/merged, so each PR remains easier to review.
> > For the `externalId` point: the existing single user/group/role access-control APIs are name-based and do not support `externalId` as the association key. The split PRs will stay aligned with the current name-based API contract. I think `externalId` support may require a separate API/model design and can be discussed as a follow-up.
> 
> You can update the code. The externalId method has already been implemented.

I have updated this PR based on the latest `externalId` and `enabled` support. Bulk user add now uses `UserAddRequest[]`, so it supports `name`, `externalId`, and `enabled`, consistent with the single user add API. The OpenAPI docs, access-control docs, examples, and tests have also been updated.

The bulk remove API remains name-based, aligned with the existing single delete user API.

The latest changes have been pushed.

PR #11915, opened 2026-07-07
[#11910] fix(clickhouse): fix GraphiteMergeTree classification and add missing type mappings

### LINKED ISSUE #11910: [Bug] ClickHouse catalog: GraphiteMergeTree misclassified and multiple types fall through to ExternalType
### Version

main branch

### Describe what's wrong

The ClickHouse catalog has two categories of issues:

**1. GraphiteMergeTree incorrectly classified as not requiring ORDER BY**

In `ClickHouseTablePropertiesMetadata`, the `GRAPHITEMERGETREE` engine constant uses the single-argument constructor, which sets `requireOrderBy=false` and `acceptPartition=false`. All six sibling MergeTree-family engines (`MERGETREE`, `REPLACINGMERGETREE`, `SUMMINGMERGETREE`, `AGGREGATINGMERGETREE`, `COLLAPSINGMERGETREE`, `VERSIONEDCOLLAPSINGMERGETREE`) use the three-argument constructor `(value, true, true)`.

Because `requireOrderBy=false`, the write path (`appendOrderBy`) throws `UnsupportedOperationException("ORDER BY clause is not supported for engine: GraphiteMergeTree")` when creating a GraphiteMergeTree table with an ORDER BY clause. And because ClickHouse itself requires ORDER BY for MergeTree-family engines, omitting it is also rejected by the server. **This means GraphiteMergeTree tables cannot be created through Gravitino at all.**

**2. Multiple types fall through to ExternalType due to missing switch cases**

`ClickHouseTypeConverter` defines constants for several types but the `toGravitino` switch lacks corresponding `case` branches. All fall through to the default `ExternalType`:

| Type | Constant | Impact |
|------|----------|--------|
| Enum8/Enum16 | `ENUM = "Enum"` | ClickHouse always normalizes `Enum` to `Enum8`; the string `"Enum8(...)"` never matches the `"Enum"` case |
| Int128/Int256 | `INT128`, `INT256` | Wide integer columns become opaque ExternalType |
| UInt128/UInt256 | `UINT128`, `UINT256` | Same as above |
| BFloat16 | `BFLOAT16` | Google Brain 16-bit float (ML inference) becomes opaque ExternalType |

This follows the same pattern as #11879 (IPv4/IPv6/LowCardinality), fixed by #11884 in `ClickHouseTypeConverter`.

### How to reproduce

**GraphiteMergeTree (create path):**

```java
catalog.asTableCatalog().createTable(
    NameIdentifier.of("db_name", "t_graphite"),
    new Column[] { Column.of("id", Types.LongType.get(), false) },
    "comment",
    Map.of("engine", "GraphiteMergeTree"),                 // engine is a property, not positional
    new Transform[0],                                      // partitions
    Distributions.NONE,                                    // distribution
    new SortOrder[] { SortOrders.ascending(NamedReference.field("id")) },  // ORDER BY
    new Index[0]);
// → UnsupportedOperationException: "ORDER BY clause is not supported for engine: GraphiteMergeTree"
```

Note: the engine must be passed via `Map.of("engine", "GraphiteMergeTree")` in the properties map. Passing empty properties defaults to `MergeTree` and the bug does not manifest.

**Type conversion (load path):**

```sql
CREATE TABLE t_enum (id Int64, status Enum8('active'=1,'inactive'=2)) ENGINE=MergeTree ORDER BY id;
CREATE TABLE t_int128 (id Int64, big_val Int128) ENGINE=MergeTree ORDER BY id;
CREATE TABLE t_bfloat (id Int64, score BFloat16) ENGINE=MergeTree ORDER BY id;
```

Load each table via Gravitino REST API. Expected: `status` → enum type, `big_val` → integer type, `score` → float type. Actual: all columns return `ExternalType`.

### Additional context

- The type-switch fix is in `ClickHouseTypeConverter` — the same file and the same `toGravitino` switch that #11884 extended for IPv4/IPv6, so it follows established precedent
- The GraphiteMergeTree fix is a one-line change in `ClickHouseTablePropertiesMetadata` (a separate file, not touched by #11884)
- Enum fix: add `if (typeName.startsWith("Enum"))` guard before the switch, or strip parameters first
- Int128/Int256/UInt128/UInt256: add 4 switch cases (note: Gravitino `DecimalType` caps precision at 38, so Int256/UInt256 cannot map to Decimal — use `ExternalType` with correct `fromGravitino` restoration, or discuss with community)
- BFloat16: add `case BFLOAT16: return Types.FloatType.get()`

jiangxt2:
Fix: #11910

- Fix GraphiteMergeTree incorrectly classified as not requiring ORDER BY (single-arg → three-arg constructor)
- Add Enum8/Enum16 normalization via startsWith guard before the type switch
- Add type mappings for Int128/Int256/UInt128/UInt256 (→ ExternalType, preserves round-trip)
- Add BFloat16 → ExternalType mapping (no half-precision in Gravitino)
- Restore DATE32 → ExternalType mapping (Date32 has wider range than Date)

GraphiteMergeTree tables cannot be created through Gravitino at all — the engine requires ORDER BY but the catalog rejects it. Multiple ClickHouse types (Enum8/16, wide integers, BFloat16, Date32) fall through to generic ExternalType instead of being handled explicitly.

## Does this PR introduce _any_ user-facing change?

No. GraphiteMergeTree tables become creatable; other type mappings preserve existing ExternalType behavior.

## How was this patch tested?

- `./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test -PskipITs` — all 46 unit tests pass
- Type converter round-trip tests cover Enum8/Enum16, Int128/Int256/UInt128/UInt256, BFloat16, DATE32


catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/converter/ClickHouseTypeConverter.java
```diff
@@ -141,7 +158,9 @@ public Type toGravitino(JdbcTypeBean typeBean) {
         return Types.FixedCharType.of(typeBean.getColumnSize());
       case DATE:
         return Types.DateType.get();
-        // No type mapping for date32, we will use external type to handle it.
+      case DATE32:
+        // Date32 supports 1900-2299 vs Date's 1970-2149. Use ExternalType to preserve round-trip.
+        return Types.ExternalType.of(DATE32);
```
yuqi1129, line 166: We haven't handled most types that can't be converted to the Gravitino type, as engines like Trino will not recognize `externalType`.  Adding this type is designed to display types that are created by the catalog itself in Gravitino.  As for how to use this type uniformly in engines, there is no clear solution now. anyway, this PR is useful and valuable.

yuqi1129:
Adding some ITs is preferable.

yuqi1129:
@jiangxt2 
Is this one ready for review?

jiangxt2:
> @jiangxt2 Is this one ready for review?

Yes, ready now.

yuqi1129 (approved):

PR #11916, opened 2026-07-07
[#11912] feat(clickhouse): support set data skipping index type

### LINKED ISSUE #11912: feat: ClickHouse catalog: support set data skipping index type
### Version

main branch

### Describe the feature

The ClickHouse catalog currently supports only `minmax` and `bloom_filter` skip index types. The `set` index type — one of the most common data skipping indexes in ClickHouse — is not supported:

**Read path** — `getClickHouseIndexType()` in `ClickHouseTableOperations` handles only `minmax` and `bloom_filter`. Unrecognized types are silently dropped with a warning log — tables load successfully but index metadata is lost.

**Write path** — `appendIndexesSql()` and `addIndexDefinition()` generate DDL only for `minmax` and `bloom_filter`; all other types throw `IllegalArgumentException`.

### Proposed implementation

1. Add `DATA_SKIPPING_SET` to the `Index.IndexType` enum.
2. Extend the switch statements in `ClickHouseTableOperations` for both read and write paths.
3. The `set(N)` type requires a cardinality parameter, which is passed as the index expression.
4. `GRANULARITY` is fixed to 1 (same as `minmax`), since the `Index` API does not yet support custom GRANULARITY values.

Related: #10446 (GRANULARITY configuration)

jiangxt2:

- Add set data skipping index type to the Index API
- Support set index in CREATE TABLE, ALTER TABLE, and table loading
- Refactor index DDL generation into a shared helper

Fix: #11912

The ClickHouse catalog only supports minmax and bloom_filter skip indexes. Tables with set indexes load without error but index metadata is silently dropped, and explicit creation attempts fail. The set index type has no required parameters (unlike ngrambf_v1/tokenbf_v1), making it safe to add without a parameter-passing mechanism.

Yes. ClickHouse tables with set data skipping indexes can now be created and loaded through Gravitino with index metadata preserved.

**Unit tests** (`TestClickHouseTableOperations`):
- `testGetClickHouseIndexType`: verifies index type string-to-enum mapping (minmax, bloom_filter, set)

**Integration tests** (`CatalogClickHouseIT`):
- `testCreateAndLoadWithPartitionSortAndIndexes`: extended with SET index create/load round-trip
- `testAlterTableAddIndexWithSetIndex`: ALTER TABLE ADD INDEX with SET type

Both test classes require `-PskipDockerTests=false` to run.


api/src/main/java/org/apache/gravitino/rel/indexes/Index.java
```diff
@@ -122,12 +122,14 @@ enum IndexType {
     /** IVF_HNSW_PQ */
     IVF_HNSW_PQ,
 
-    // The following index types are data skipping indexes in ClickHouse, ngrambf_v1 and tokenbf_v1
```
yuqi1129, line 125: Skip indexes like ngrambf_v1 and tokenbf_v1 still not supported, we'd better not remove the comment here.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -479,23 +480,21 @@ private void appendIndexesSql(Index[] indexes, StringBuilder sqlBuilder) {
           sqlBuilder.append(" PRIMARY KEY (").append(fieldStr).append(")");
           break;
         case DATA_SKIPPING_MINMAX:
-          Preconditions.checkArgument(
-              StringUtils.isNotBlank(index.name()), "Data skipping index name must not be blank");
-          // The GRANULARITY value is always 1 here currently as we can't set it by Index: there is
```
yuqi1129, line 484: Comment like `The GRANULARITY value is always 1 here currently as we can't set it by Index` needs to be kept until we have already support it.

jiangxt2:
Thanks for the thorough review! Ive addressed all comments in `ea1d1aa`:

1. **Comment on `Index.java`** — Restored the note about `ngrambf_v1` and `tokenbf_v1` not yet supported. These require a parameter-passing mechanism (unlike `set`), so the comment is still relevant.

2. **Comment on `appendIndexesSql`** — Restored the GRANULARITY explanation comments for `minmax` (always 1) and `bloom_filter` (always 3), as the `Index` API doesnt support custom GRANULARITY yet.

3. **Double "INDEX" bug in `addIndexDefinition`** — Good catch by Copilot! The refactored code was producing `"ADD INDEX INDEX ..."` because `buildDataSkippingIndexDdl()` already includes the leading `INDEX` keyword. Fixed by changing `"ADD INDEX "` to `"ADD "`.

All tests pass: unit tests (25) + Docker IT — CatalogClickHouseIT (31) + CatalogClickHouseClusterIT (12), including the `testAlterIndexAndAutoIncrementBranches` and `testAlterAddIndexAndAutoIncrementInCluster` tests that exercise the ALTER TABLE ADD INDEX path.

yuqi1129:
@jiangxt2 
Since https://github.com/apache/gravitino/pull/11731 has been merged, we'd better add IT to cover the change in this PR and others.

jiangxt2:
Done. Added IT coverage for set index across all paths:
- `CatalogClickHouseIT`: CREATE + LOAD, ALTER ADD + DELETE, branch coverage
- `CatalogClickHouseClusterIT`: cluster ADD/DELETE with fieldNames verification

Also found and fixed a DDL bug: ClickHouse requires `set(N)` syntax, not bare `set`. The unit tests only checked string generation, but the IT tests caught the actual ClickHouse rejection. Fixed in `ClickHouseTableOperations` + synced unit test assertions.

Additionally hardened `getClickHouseIndexType` with `startsWith("set(")` fallback for versions returning parameterized type from `system.data_skipping_indices`.

yuqi1129:
@jiangxt2 
Could you also change the document `jdbc-clichouse-catalog` and add support for set data skipping index?

jiangxt2:
Sure, will update `jdbc-clickhouse-catalog.md` to document `DATA_SKIPPING_SET` — both in the feature overview table and the Table Indexes section.

jiangxt2:
Updated `jdbc-clickhouse-catalog.md` in `5aa3539`:

- Feature overview table (L175): added `DATA_SKIPPING_SET` to the supported data-skipping indexes list.
- Table Indexes section (L245): added `DATA_SKIPPING_SET` (`GRANULARITY` fixed to 1).

yuqi1129:
@jiangxt2 
Can you also update the related issues, since the PR is a feature but the issue says it's a bug?

yuqi1129 (approved):

jiangxt2:
Done, updated.

PR #11926, opened 2026-07-07
[#11925] feat(trino-connector): Support views in Trino connector

### LINKED ISSUE #11925: [FEATURE] Add view support to Trino connector
### Describe the feature

Support creating, reading, replacing, renaming, and dropping views via Trino SQL on Gravitino-backed catalogs.

### Motivation

Gravitino exposes a ViewCatalog API and the Flink connector already consumes it. The Trino connector should allow users to manage and query these views via Trino SQL.

### Describe the solution

Implement the view-related `ConnectorMetadata` methods in the Trino connector by delegating to the ViewCatalog API, and fix the Hive backend's rejection of the `trino` SQL dialect so Hive-backed views work end-to-end.

### Additional context

No additional context.

diqiu50:

Add view support to the Gravitino Trino connector:

- `HiveViewCatalogOperations`: accept the `trino` SQL dialect in view creation and loading.
- `CatalogConnectorMetadata`: delegate view CRUD to Gravitino's `ViewCatalog` API, only exposing views with a `trino` SQL representation.
- `GravitinoMetadata` (+ per-Trino-version modules): implement `listViews`, `getView`, `createView`, `dropView`, `renameView`, `isView`.

Fix: #11925

Yes — users can now `CREATE VIEW` / `CREATE OR REPLACE VIEW` / `SHOW CREATE VIEW` / `DROP VIEW` / `ALTER VIEW ... RENAME TO` via Trino SQL on Hive/Iceberg-backed catalogs, and views appear in `SHOW TABLES`.

- Unit tests: `TestCatalogConnectorMetadataView`, `TestGravitinoMetadataView`, Hive dialect tests in `TestHiveCatalogOperations`.
- Integration testsets: new view test cases for Hive and Iceberg catalogs.


trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +473,258 @@ public Function getFunction(String schemaName, String functionName) {
...
+      throw new TrinoException(
+          GravitinoErrorCode.GRAVITINO_UNSUPPORTED_OPERATION, "Catalog does not support views");
+    }
+    NameIdentifier identifier = NameIdentifier.of(view.getSchemaName(), view.getName());
+    SQLRepresentation[] representations = {
+      SQLRepresentation.builder().withDialect(Dialects.TRINO).withSql(view.getSql()).build()
+    };
+    try {
+      boolean exists = viewCatalog.viewExists(identifier);
+      if (exists && !getViewIfPresent(view.getSchemaName(), view.getName()).isPresent()) {
```
mchades: here call Gravitino twice, can we reduce that?

diqiu50: fixed

diqiu50: fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -368,15 +380,15 @@ private HiveView toHiveView(
     String detectedDialect = HiveView.detectDialect(representationSql, params);
     switch (detectedDialect.toLowerCase(Locale.ROOT)) {
       case Dialects.HIVE:
+      case Dialects.TRINO:
```
mchades, line 470: Could we handle native Trino HMS view encoding before treating `viewOriginalText` as SQL? Native Trino stores a Base64-encoded `ConnectorViewDefinition` wrapped in `/* Presto View: ... */`, together with a dummy HMS column. This code recognizes that marker as the Trino dialect but forwards the encoded text and dummy schema unchanged, so loading an existing native Trino view through Gravitino cannot reconstruct its SQL or columns.

Please either decode the native payload, including its columns and default catalog/schema, or use a distinct marker for Gravitino's raw-SQL format. Could we also add a test using an actual encoded Trino payload? The current `presto_view=true` plus `SELECT 1` fixture only covers the raw-SQL round trip introduced by this PR.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +474,265 @@ public Function getFunction(String schemaName, String functionName) {
...
+        }
+        if (!replace) {
+          throw new TrinoException(
+              GravitinoErrorCode.GRAVITINO_VIEW_ALREADY_EXISTS, "View already exists");
+        }
+        viewCatalog.alterView(
+            identifier,
+            ViewChange.replaceView(
+                view.getRawColumns(),
+                mergeWithNonTrinoRepresentations(existingView.representations(), representations),
```
mchades: The merge avoids deleting non-Trino representations, but it can leave them semantically stale. `CREATE OR REPLACE VIEW` also replaces the shared columns and default catalog/schema, while the Spark/Hive/Flink SQL is preserved unchanged. For example, Trino can add an output column while the retained Spark representation still produces the old schema.

Since the connector cannot regenerate arbitrary dialects, could we reject replacement when unrelated dialect representations are present instead of persisting an inconsistent logical view? Paimon's mandatory `query` representation needs an explicit policy: update it together with the Trino SQL, or mark Paimon replacement as unsupported.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +474,265 @@ public Function getFunction(String schemaName, String functionName) {
...
+          GravitinoErrorCode.GRAVITINO_UNSUPPORTED_OPERATION, "Catalog does not support views");
+    }
+    Preconditions.checkArgument(
+        view.getSql() != null,
+        "View %s.%s has no Trino dialect SQL representation",
+        view.getSchemaName(),
+        view.getName());
+    NameIdentifier identifier = NameIdentifier.of(view.getSchemaName(), view.getName());
+    SQLRepresentation[] representations = {
+      SQLRepresentation.builder().withDialect(Dialects.TRINO).withSql(view.getSql()).build()
```
mchades, line 611: This create path only supplies a `trino` representation, but Paimon requires a non-empty canonical `query` representation and rejects the request when it is missing. As a result, the documented `CREATE VIEW` support for Paimon fails before persistence.

Could we emit both `trino` and `query` representations for Paimon-backed catalogs, or remove Paimon from the supported list for now? Please also add a Paimon Trino integration test covering create and replace.

mchades (changes requested):
Thanks for adding the view support. The main flow is well covered, but the current implementation still has backend-specific correctness gaps that make the documented Paimon creation path fail and can misread or persist inconsistent multi-representation views. Please address the inline comments before merge.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Added a dedicated gravitino.view.trino_dialect marker so views without it (including genuine
  native Trino views) are treated as Hive dialect instead of having their encoded body mis-parsed as SQL. Verified with a live integration test creating a real native Trino view via a separate native_hive catalog.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
diqiu50: Replace now rejects with GRAVITINO_UNSUPPORTED_OPERATION when other-dialect representations exist, instead of merging them.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
diqiu50: Trino connector doesn't support Paimon catalogs yet (no registered provider), so this
path is currently unreachable. Let me know if you want it handled now or deferred.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
mchades: Thanks for adding the dedicated marker and the live test. The marker prevents the Gravitino Trino connector from reading the native payload as Trino SQL, but `HiveView.detectDialect()` now falls through to `hive`, and `toHiveView()` still publishes the encoded `/* Presto View: ... */` text and HMS columns as a Hive representation. Other consumers of the same Hive ViewCatalog, such as Spark and Flink, fall back to the Hive representation and will receive invalid SQL. The new IT only verifies that the view is hidden from this Trino connector. Could we detect `presto_view=true` or the native prefix and treat the view as unsupported (or decode it), rather than exposing it as Hive SQL? Please also add a catalog-hive test asserting that a native payload is not returned as a Hive representation.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
mchades: Agreed that the Paimon path is currently unreachable because no Paimon provider is registered in the Trino connector. In that case, I think implementation and the Paimon IT can be deferred, but the user-facing scope should be narrowed in this PR: `docs/trino-connector/sql-support.md` and the PR description currently state that view management supports Paimon. Could we remove Paimon from those support claims so users are not promised `CREATE VIEW` support for a provider that the connector rejects?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Fixed detectDialect no longer falls back to hive for unrecognized native Presto/Trino views

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
diqiu50: Remove docs

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -431,6 +443,22 @@ private SQLRepresentation validateSQLRepresentation(
             defaultSchema,
             ident);
         return selected;
+      case Dialects.TRINO:
+        // HMS has no field to persist a Trino view's default catalog/schema, so any value
```
mchades: **[P1] Please preserve the Trino resolution context instead of accepting and dropping it.** Trino supplies the session catalog/schema in `ConnectorViewDefinition`; a normal `USE gt_hive.db; CREATE VIEW v AS SELECT * FROM t` is valid at creation, but this branch stores neither value, so reload returns null defaults and Trino cannot resolve `t`. Lowering the log level does not change that behavior. Could we persist both values in reserved HMS properties and restore them in `toHiveView()` (or reject the definition), and add a Hive IT using `USE` plus an unqualified source table?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
```diff
@@ -148,6 +155,83 @@ public GravitinoTable createTable(ConnectorTableMetadata tableMetadata) {
...
+                    new ViewColumn(
+                        column.getName(),
+                        dataTypeTransformer.getTrinoType(column.getType()).getTypeId(),
+                        Optional.ofNullable(column.getComment())))
+            .collect(Collectors.toList());
+
+    return new ConnectorViewDefinition(
+        view.getSql(),
+        Optional.ofNullable(view.getDefaultCatalog()),
+        Optional.ofNullable(view.getDefaultSchema()),
```
mchades, line 209: **[P2] Avoid constructing an invalid schema-only `ConnectorViewDefinition`.** Iceberg view metadata can have a non-null default namespace while its default catalog is null, and `catalog-lakehouse-iceberg` currently supports that shape. This adapter converts it to an empty catalog plus a present schema, but the Trino constructor throws `IllegalArgumentException` because a schema requires a catalog. As a result, loading a Trino-representation view with that metadata fails before `SELECT` or `SHOW CREATE VIEW` can use it. Could we normalize the catalog to the current connector catalog, or hide/reject the view with a controlled error, and add a regression test?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
```diff
@@ -148,6 +155,83 @@ public GravitinoTable createTable(ConnectorTableMetadata tableMetadata) {
...
+        columns,
+        Optional.ofNullable(view.getComment()),
+        Optional.empty(),
+        true,
+        List.of());
```
mchades, line 214: **[P2] Please preserve or explicitly reject the view path.** Trino stores the session path in `ConnectorViewDefinition`, but `createView()` below ignores `definition.getPath()` and this load path always returns `List.of()`. A view that resolves unqualified functions through `SET PATH` can therefore fail or bind differently after reload. Could we round-trip the path through Gravitino metadata, or reject non-empty paths, and cover that behavior with a test?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
```diff
@@ -40,9 +40,15 @@
...
- * other views are treated as native Hive SQL views.
+ * from table properties: Trino views (written by this catalog) carry {@code
+ * gravitino.view.trino_dialect}, Spark views carry {@code spark.sql.create.version}, Flink views
+ * carry properties prefixed with {@code flink.}, and all other views are treated as native Hive SQL
+ * views.
+ *
+ * <p>Native Presto/Trino views created outside Gravitino use the {@code presto_view} HMS property
+ * and encode their body as a base64-wrapped comment rather than plain SQL text; since decoding that
+ * native format requires Trino's own serialization logic, this catalog does not attempt to read it
+ * and such views are treated as Hive dialect (i.e. not exposed as Trino-readable SQL).
```
mchades: **[P3] Keep the documented behavior in sync.** The latest implementation rejects native Presto/Trino views with `UnsupportedOperationException`; it no longer treats them as Hive dialect. Could we update this sentence so the class documentation matches `detectDialect()`?

mchades (commented):
Additional findings from the latest head review.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
mchades: Thanks, the repository documentation now correctly limits view support to Hive and Iceberg. The PR description still says `Hive/Iceberg/Paimon-backed catalogs`, though. Could you remove Paimon from that user-facing change statement as well? I am keeping this thread open until both support claims are aligned.

diqiu50: fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: fixed

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
diqiu50: fixed

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
diqiu50: fixed

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
diqiu50: fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
diqiu50: fixed

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
mchades: Thanks, this fixes the invalid constructor shape in single-metalake mode. In multi-metalake mode, however, `getView()` still passes `CatalogConnectorMetadata.getCatalogName()`, whose field is initialized from `catalogIdentifier.name()` (for example, `iceberg`). The actual Trino catalog name is produced by `getTrinoCatalogName(metalake, catalog)` (for example, `metalake.iceberg`). A schema-only Iceberg view is therefore normalized to the wrong catalog and cannot resolve during analysis. Could we retain and pass the actual `connectorName` supplied to `GravitinoConnectorFactory.create(...)`, and add a `use-single-metalake=false` regression test?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
mchades: Thanks, rejecting non-empty paths avoids the silent semantic change. One issue remains: `Preconditions.checkArgument` throws a plain `IllegalArgumentException`, and `createViewInternal()` does not translate it, so Trino reports this user-selected `SET PATH` case as `GENERIC_INTERNAL_ERROR`. Could we throw `TrinoException` with `GRAVITINO_UNSUPPORTED_OPERATION` (or `NOT_SUPPORTED`) instead, and assert that error code in the regression test?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
diqiu50: In multi-metalake mode, view handling is more complex — catalog names need quoting, and Trino's handling of quoted catalog names is inconsistent across different code paths. This PR won't support that case for now; the limitation is documented in the code and docs instead.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
diqiu50: fixed

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
mchades: Thanks for clarifying. Deferring multi-metalake view support is fine, but the current change only adds a source comment; the connector still returns an incorrect catalog, and the limitation is not documented for users. Could we explicitly reject this case with `GRAVITINO_UNSUPPORTED_OPERATION`, and add a regression test and docs update? I will keep this thread open until then.

diqiu50: fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveView.java
```diff
@@ -109,18 +134,23 @@ public AuditInfo auditInfo() {
...
 
   /**
-   * Detects the SQL dialect from HMS table properties and view text.
+   * Detects the SQL dialect from HMS table properties.
    *
-   * @param viewOriginalText The original view text from HMS.
    * @param parameters The HMS table parameters map.
    * @return The detected dialect string: "trino", "spark", "flink", or "hive".
+   * @throws UnsupportedOperationException if the view was created by a native Presto/Trino Hive
+   *     connector (bypassing Gravitino); its encoded body cannot be read as SQL by any engine.
```
mchades: we should decode the body and return the Trino dialect view SQL

diqiu50: Fixed

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadataAdapter.java
```diff
@@ -148,6 +157,116 @@ public GravitinoTable createTable(ConnectorTableMetadata tableMetadata) {
...
+      defaultCatalog = catalogName;
+    }
+
+    return new ConnectorViewDefinition(
+        view.getSql(),
+        Optional.ofNullable(defaultCatalog),
+        Optional.ofNullable(view.getDefaultSchema()),
+        columns,
+        Optional.ofNullable(view.getComment()),
+        Optional.empty(),
```
mchades, line 212: **[P1] Could we round-trip the native Trino security and resolution context through reserved internal view properties?** `createView()` drops the owner/security mode, native view loading also discards these fields and the SQL path, while this method always reconstructs an ownerless `SECURITY INVOKER` definition with an empty path. Trino creates views as `SECURITY DEFINER` by default, so this silently changes semantics after reload. Please namespace and protect these properties from direct mutation, use the decoded native payload as the source of truth on load, and cover DEFINER, INVOKER, and non-empty-path round trips.

trino-connector/integration-test/src/test/resources/trino-ci-testset/testsets/hive/00019_view.sql
```diff
@@ -0,0 +1,47 @@
...
+
+-- error cases: drop nonexistent view; create view colliding with existing table name (HMS natural rejection)
+DROP VIEW gt_hive.gt_hive_view_db.nonexistent_view;
+CREATE VIEW gt_hive.gt_hive_view_db.t01 AS SELECT 1;
+
+-- native Trino view interop: a view created directly through Trino's own native Hive connector
+-- (bypassing Gravitino) is encoded using Trino's own native "Presto View" format; Gravitino's
+-- Trino connector recognizes this format directly, so the view is visible and queryable through
+-- gt_hive too, without going through Gravitino at all to create it.
+CREATE VIEW native_hive.gt_hive_view_db.native_v01 AS SELECT id, name FROM native_hive.gt_hive_view_db.t01;
```
mchades, line 41: **[P2] This covers native Hive → Gravitino, but not the reverse direction.** Could we also create a view through `gt_hive`, then run `SHOW CREATE VIEW` and `SELECT` through `native_hive`, including a `timestamp(3)` output column? Otherwise the new encoder is only validated by Gravitino’s own decoder, so a native-incompatible payload could still pass this test.

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +498,297 @@ public Function getFunction(String schemaName, String functionName) {
...
+        }
+        List<ViewChange> changes = new ArrayList<>();
+        changes.add(
+            ViewChange.replaceView(
+                view.getRawColumns(),
+                representations,
+                view.getDefaultCatalog(),
+                view.getDefaultSchema(),
+                view.getComment()));
+        changes.addAll(computePropertyChanges(existingView.properties(), view.getProperties()));
```
mchades, line 646: **[P2] For Trino 445+, how can a non-empty view-property map reach this code from SQL?** No connector implementation exposes `getViewProperties()`, so Trino rejects `CREATE VIEW ... WITH (...)` before `createViewInternal()`, while the current unit test injects the property map directly. Could we delegate the internal connector’s view-property metadata in the applicable version shim(s), splitting the 440–445 range if necessary, and add an SQL-level IT?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +498,297 @@ public Function getFunction(String schemaName, String functionName) {
...
+                        ((SQLRepresentation) representation).dialect()));
+  }
+
+  /**
+   * Computes the {@link ViewChange} operations needed to apply {@code desired} on top of a view's
+   * current properties, so that {@code CREATE OR REPLACE VIEW ... WITH (...)} applies the new
+   * properties instead of silently leaving them unset ({@link ViewChange#replaceView} does not
+   * affect properties).
+   *
+   * <p>Only additions/updates are applied, never removals: {@code current} is the view's full
```
mchades, line 774: **[P2] `CREATE OR REPLACE VIEW` should replace caller-managed properties, but this only adds or updates them, so a property omitted from the new `WITH (...)` clause remains stale.** Could we preserve backend/internal reserved keys while emitting `RemoveProperty` for omitted caller-owned keys, and test removing a user property while retaining reserved Hive and Trino metadata?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +498,297 @@ public Function getFunction(String schemaName, String functionName) {
...
+      NameIdentifier[] views = viewCatalog.listViews(Namespace.of(schemaName));
+      return Arrays.stream(views)
+          .map(NameIdentifier::name)
+          .filter(viewName -> getViewIfPresent(schemaName, viewName).isPresent())
+          .toList();
+    } catch (UnsupportedOperationException e) {
+      LOG.debug(
+          "Catalog {} does not support listing views for schema {}", catalogName, schemaName, e);
+      return List.of();
+    } catch (NoSuchSchemaException e) {
```
mchades, line 583: **[P2] `ConnectorMetadata.listViews()` must return an empty list when the requested schema does not exist, but this converts `NoSuchSchemaException` into an error.** Could we return `List.of()` for this case and add a missing-schema regression test?

trino-connector/trino-connector/src/main/java/org/apache/gravitino/trino/connector/catalog/CatalogConnectorMetadata.java
```diff
@@ -454,4 +498,297 @@ public Function getFunction(String schemaName, String functionName) {
...
+      return;
+    }
+    try {
+      viewCatalog.alterView(
+          NameIdentifier.of(oldViewName.getSchemaName(), oldViewName.getTableName()),
+          ViewChange.rename(newViewName.getTableName()));
+    } catch (NoSuchViewException e) {
+      throw new TrinoException(
+          GravitinoErrorCode.GRAVITINO_VIEW_NOT_EXISTS, "View does not exist", e);
+    } catch (ViewAlreadyExistsException e) {
```
mchades, line 741: **[P2] This catch is not reached for Iceberg rename collisions because `IcebergViewCatalogOperations` lets `org.apache.iceberg.exceptions.AlreadyExistsException` escape instead of translating it to Gravitino’s `ViewAlreadyExistsException`.** Could we normalize that exception in the Iceberg adapter and add an occupied-target regression test?

Issue #11949, opened 2026-07-08
[FEATURE] Native variant type support for Iceberg V3 in Gravitino


nevzheng:
### Describe the feature

Add a first-class `variant` type to Gravitino's unified type model so Iceberg V3 `variant` columns are represented natively — as `Types.VariantType`, not the `ExternalType("variant")` stopgap.

Following from #11927, #11928, #11932, and the discussion in #11929, **this issue is the canonical thread tracking the implementation of Iceberg V3 `variant` support in Gravitino.** Detailed design and review discussion should happen here; the prior threads remain as background.

### Motivation

Loading an Iceberg V3 table with a `variant` column through the native metadata API fails with `UnsupportedOperationException: Unsupported type: variant` (#11927). The immediate options were:

- **`ExternalType` stopgap (#11928)** — maps `variant` → `external(variant)`. Loads the table, but `variant` stays an opaque string with no identity: nothing can branch on it, and external types have caused downstream problems (`external(serial)` unqueryable via Trino #10957; `catalogString()` written verbatim into DDL #11805).
- **Native `VariantType` (#11932)** — gives `variant` a real, modeled identity so each connector maps it deliberately.

Per discussion in #11929 (@jerryshao), we're going with native support and will land it in a later release rather than shipping the interim `ExternalType` patch now. #11928 is being closed in favor of #11932.

### Describe the solution

Model `variant` as a parameterless leaf (same shape as `UUIDType`):

- **api** — `Type.Name.VARIANT` + `Types.VariantType`.
- **common** — `JsonUtils` serializes/parses it as the token `"variant"`.
- **catalog-lakehouse-iceberg** — map Iceberg `VariantType` ⇄ Gravitino `VariantType`.
- **Python client** — mirror `Name.VARIANT`, `Types.VariantType`, and the serde registry.
- **docs** — unified type reference + per-connector type-mapping tables.
- **connector propagation** — native mapping where the engine has Open Variant (Iceberg, Paimon, Doris); reject-with-test where it's absent or a different concept, flippable as engines ship the standard.

No formal design doc — small, well-scoped feature (per @jerryshao in #11929).

Implemented by #11932.

### Additional context

- #11927 — the load bug that surfaced this.
- #11928 — the `ExternalType` stopgap (closing in favor of #11932).
- #11929 — `[DISCUSS]` thread; decision to go native.
- #11932 — the native implementation.
- #10994 — variant on the IRC/JDBC backend (already handled).
[...]

PR #11932, opened 2026-07-08
[#11949] feat(iceberg): native variant type support

nevzheng:
### What changes were proposed in this pull request?

Add a first-class `variant` type to Gravitino's unified type model and wire it through the Iceberg catalog and the Python client. At the schema level Iceberg V3's `variant` is an opaque, parameterless leaf, so `VariantType` is a singleton `PrimitiveType` like `UUIDType`.

- `api`: `Type.Name.VARIANT` + `Types.VariantType`.
- `common`: `JsonUtils` serializes/parses it as the token `"variant"`.
- `catalog-lakehouse-iceberg`: map Iceberg `VariantType` ⇄ Gravitino `VariantType`.
- Python client: mirror `Name.VARIANT`, `Types.VariantType`, and the serde registry, so a Python client loads a variant column as `VariantType` rather than `UnparsedType`.
- Docs: unified type reference, Iceberg type-mapping table, OpenAPI examples.

Fixes #11949

**Follow-ups:** this is the first of a short series, split for reviewability. Later PRs add native `variant` support for the other engines that have Open Variant (Paimon, Doris), then test coverage locking in reject behavior for the engines that don't. They'll go up one at a time as each merges.

### Why are the changes needed?

Loading an Iceberg V3 table with a `variant` column through the native metadata API failed with `UnsupportedOperationException: Unsupported type: variant` (#11927). Modeling variant natively rather than as an opaque `ExternalType` string gives it a stable identity so each connector maps it deliberately. Discussed in #11929; supersedes the `ExternalType` stopgap (#11928).

### Does this PR introduce any user-facing change?

Yes — Iceberg V3 tables with a `variant` column now load through the native API as a `variant` type. Other connectors are unchanged in this PR.

### How was this patch tested?

- Unit: `TestTypes` (type contract), `TestJsonUtils` (JSON round-trip), `TestConvertUtil` (Iceberg converter, both directions).
- Python unit: variant type contract + serde round-trip.
- Docker IT: `CatalogIcebergBaseIT` — create a format-version-3 table via the raw Iceberg client, load it through the native API, and assert variant resolves to the native type.

PR #11956, opened 2026-07-09
[#11954] feat(iceberg): default and validate the table format-version property

### LINKED ISSUE #11954: [FEATURE] Support creating Iceberg format-version 3 tables through the Gravitino API
### Describe the feature

Allow the Iceberg table **format version** to be selected (`2` or `3`) when creating a table through the Gravitino relational API, with validation and a sensible default. Today `format-version` is an unvalidated pass-through property: the effective default is `2`, there is no first-class/validated way to ask for `3`, and misconfigurations surface as low-level Iceberg errors rather than clear Gravitino ones.

Format version **3** is a prerequisite for Iceberg's V3 types (e.g. `variant`; see the [Iceberg table spec](https://iceberg.apache.org/spec/) — "Version 3"). Without a supported way to create v3 tables, those types can't be written through Gravitino.

### Motivation

Creating a table that needs v3 fails with a downstream Iceberg error, because the format version defaults to `2`. For example, a `variant` column (a V3 type) rejected at v2:

```
Failed to operate object [t_variant] operation [CREATE] under [db],
reason [Invalid schema for v2:
- Invalid type for payload: variant is not supported until v3]
```

And because the property is an unvalidated pass-through, an empty or invalid value also leaks a raw Iceberg/parse error instead of a clear message:

```
Failed to operate table(s) [t] operation [CREATE] under schema [db],
reason [For input string: ""]
```

Users have no documented, validated way to select v3, and no clear signal about which versions are supported.

### Describe the solution

Make `format-version` a first-class, validated Iceberg table property:

- Accept only `2` or `3`; reject other values at the Gravitino layer with a clear message, e.g. `Invalid value: '4' for property: 'format-version'`.
- Default to `2` when unset or empty (`empty -> 2`, `2 -> 2`, `3 -> 3`), so existing behavior is unchanged and v3 is opt-in.
- Document the property and add create examples for both v2 and v3.
- Add unit tests and REST/IRC integration tests that create a v3 table (incl. a `variant` column) through the Gravitino API and read it back.

Example — creating a v3 table via REST:

```bash
curl -X POST .../schemas/db/tables -H 'Content-Type: application/json' -d '{
  "name": "t_v3",
  "columns": [{"name": "id", "type": "integer", "nullable": false},
              {"name": "payload", "type": "variant", "nullable": true}],
  "properties": {"format-version": "3"}
}'
```

### Additional context

- Related to #11949 (native `variant` type — the primary V3 type motivating this) and its implementation in #11932. This issue is the general "select/enable format version 3" mechanism; it is not limited to variant.
- Iceberg format versions: v3 support lands in Iceberg 1.9+, and Gravitino currently builds against Iceberg 1.11.0.

nevzheng:
Surface `format-version` as a first-class, validated Iceberg table property (builds on the native `variant` type from #11932):
- Typed as an integer restricted to `{2, 3}`; other values are rejected at the Gravitino layer with a clear message.
- Gravitino owns the default and stamps `2` when the value is unset or empty (`empty → 2`, `2 → 2`, `3 → 3`). Setting `3` is what enables V3 types such as `variant`.
- Unit tests (property validation + create-property resolution) and REST/IRC integration tests that create a `variant` column through the Gravitino API at `format-version` 3 and read it back.
- Docs: the catalog property table plus two OpenAPI create examples (`IcebergTableCreate` v2, `IcebergVariantTableCreate` v3).

Creating a `variant` column (the V3 type added in #11932) requires the table at Iceberg format-version 3. `format-version` was previously an unvalidated free-form string with no default, so a variant create failed with a low-level Iceberg error, and nothing covered the write path to a REST/IRC backend.

Fixes #11954. Related to #11949 (the umbrella tracking issue for the whole variant feature).

- The `format-version` Iceberg table property is now validated — allowed values `2`/`3`, unset/empty defaults to `2`, others rejected with a clear error.
- New OpenAPI examples for creating Iceberg tables.

- Unit tests: property validation (default, valid/invalid, empty) and create-property mapping (`absent/""/2/3`).
- Integration tests (REST/IRC): variant column created via the Gravitino API at v3 and read back; variant-without-v3 rejected; unset/`2`/empty resolve to `2`.
- Verified end-to-end in the Gravitino playground against an Iceberg REST catalog (read back over the `/iceberg` endpoint).
- `./gradlew :docs:build` validates the OpenAPI spec.

### DIFF
--- a/catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTable.java
+++ b/catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTable.java
@@ -77,6 +77,16 @@ public class IcebergTable extends BaseTable {
   private IcebergTable() {}
 
   public static Map<String, String> rebuildCreateProperties(Map<String, String> createProperties) {
+    // Gravitino owns the default Iceberg table format version: when it is not explicitly set, stamp
+    // ICEBERG_DEFAULT_FORMAT_VERSION rather than relying on the Iceberg library's version-dependent
+    // default or letting a blank value fail to parse downstream.
+    String formatVersion = createProperties.get(IcebergTablePropertiesMetadata.FORMAT_VERSION);
+    if (formatVersion == null || formatVersion.trim().isEmpty()) {
+      createProperties.put(
+          IcebergTablePropertiesMetadata.FORMAT_VERSION,
+          String.valueOf(IcebergTablePropertiesMetadata.ICEBERG_DEFAULT_FORMAT_VERSION));
+    }
+
     String provider = createProperties.get(PROP_PROVIDER);
     if (ICEBERG_PARQUET_FILE_FORMAT.equalsIgnoreCase(provider)) {
       createProperties.put(DEFAULT_FILE_FORMAT, ICEBERG_PARQUET_FILE_FORMAT);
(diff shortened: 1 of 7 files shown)


nevzheng:
Reopened as #11957 — this PR closed automatically when its head branch was renamed to `iceberg-format-version-v3`. Same commit, same content; please review #11957 instead.

PR #11957, opened 2026-07-09
[#11954] feat(iceberg): enable creating format-version 3 (V3) tables via the Gravitino API

### LINKED ISSUE #11954: [FEATURE] Support creating Iceberg format-version 3 tables through the Gravitino API

nevzheng:
Enable creating Iceberg **format-version 3 (V3)** tables through the Gravitino relational API — the format version required for Iceberg's V3 types such as `variant` (added in #11932). This makes `format-version` a first-class, validated table property; everything else follows from that:

- **Select the version** — Gravitino accepts creating tables at versions **1–4** (the range the bundled Iceberg 1.11.0 can write); other values are rejected at the Gravitino layer with a clear message instead of a low-level Iceberg error.
- **Sensible default** — unset or empty resolves to `2` (`empty → 2`, `2 → 2`, `3 → 3`), stamped by Gravitino at create time, so existing behavior is unchanged and v3 is opt-in.
- **Tests** — unit (validation over `1–4` + create-property resolution) and REST/IRC integration: create a `variant` column via the Gravitino API at v3 and read it back.
- **Docs** — catalog property table + two OpenAPI create examples (`IcebergTableCreate` v2, `IcebergVariantTableCreate` v3).

Per-version support: **v1/v2** are fully supported (v2 is the community default since Iceberg 1.4.0); **v3** is required for V3 types and its type coverage is being added incrementally (`variant` now); **v4** is creatable but not yet a finalized Iceberg spec.

Format version 3 is a prerequisite for Iceberg's V3 types (e.g. `variant`; see the [Iceberg table spec](https://iceberg.apache.org/spec/) — "Version 3"). Before this, `format-version` was an unvalidated free-form string with no default or documentation, so creating a v3 table was not a supported, discoverable operation — attempts that needed v3 failed with confusing downstream Iceberg errors, and nothing covered the write path to a REST/IRC backend.

Fixes #11954. Related to #11949 (the umbrella tracking issue for the whole variant feature).

- You can now create Iceberg tables at **format versions 1–4** through the Gravitino API (default `2`) — e.g. `"format-version": "3"` for V3 types like `variant`.
- `format-version` is now validated: allowed values `1`/`2`/`3`/`4`, unset/empty defaults to `2`, other values rejected with a clear error.
- New OpenAPI examples for creating Iceberg tables.

- Unit tests: property validation (default; valid `""`/`1`/`2`/`3`/`4`; invalid `0`/`5`/`100`/`-1`/`INT_MAX`/`INT_MIN`/non-numeric) and create-property mapping (`absent/""/2/3`).
- Integration tests (REST/IRC): variant column created via the Gravitino API at v3 and read back; variant-without-v3 rejected; unset/`2`/empty resolve to `2`.
- Verified end-to-end in the Gravitino playground against an Iceberg REST catalog (read back over the `/iceberg` endpoint); confirmed Iceberg 1.11 creates v1–v4 and rejects v5.
- `./gradlew :docs:build` validates the OpenAPI spec.

### DIFF
--- a/catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTablePropertiesMetadata.java
+++ b/catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTablePropertiesMetadata.java
@@ -21,10 +21,14 @@ package org.apache.gravitino.catalog.lakehouse.iceberg;
 import static org.apache.gravitino.connector.PropertyEntry.stringImmutablePropertyEntry;
 import static org.apache.gravitino.connector.PropertyEntry.stringReservedPropertyEntry;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.connector.BasePropertiesMetadata;
 import org.apache.gravitino.connector.PropertyEntry;
 import org.apache.iceberg.TableProperties;
@@ -41,6 +45,23 @@ public class IcebergTablePropertiesMetadata extends BasePropertiesMetadata {
   public static final String FORMAT = IcebergConstants.FORMAT;
   public static final String FORMAT_VERSION = IcebergConstants.FORMAT_VERSION;
   public static final String DISTRIBUTION_MODE = TableProperties.WRITE_DISTRIBUTION_MODE;
+
+  /**
+   * The default Iceberg table format version Gravitino applies when {@link #FORMAT_VERSION} is not
+   * explicitly set. Gravitino owns this default rather than deferring to the Iceberg library's own
+   * version-dependent default, and stamps it onto the table at creation.
+   */
+  public static final int ICEBERG_DEFAULT_FORMAT_VERSION = 2;
+
+  /**
+   * The Iceberg table format versions Gravitino allows creating: {@code 1}–{@code 4}, the range the
+   * bundled Iceberg version (1.11.0) can write. Gravitino is not more restrictive than Iceberg for
+   * the create passthrough; each version's feature set is defined by the Iceberg spec (v1/v2/v3 are
+   * adopted, v3 is required for V3 types such as {@code variant}, and v4 is under active
+   * development and not yet finalized). An unset (empty) value defaults to {@link
+   * #ICEBERG_DEFAULT_FORMAT_VERSION}. Extend this set as newer Iceberg writer versions ship.
+   */
+  public static final Set<Integer> SUPPORTED_FORMAT_VERSIONS = ImmutableSet.of(1, 2, 3, 4);
 
   private static final Map<String, PropertyEntry<?>> PROPERTIES_METADATA;
 
@@ -65,8 +86,7 @@ public class IcebergTablePropertiesMetadata extends BasePropertiesMetadata {
             stringReservedPropertyEntry(
                 IDENTIFIER_FIELDS, "The identifier field(s) for defining the table", false),
             stringReservedPropertyEntry(DISTRIBUTION_MODE, "Write distribution mode", false),
-            stringImmutablePropertyEntry(
-                FORMAT_VERSION, "The Iceberg table format version, ", false, null, false, false),
+            formatVersionPropertyEntry(),
             stringImmutablePropertyEntry(
                 PROVIDER,
                 "Iceberg provider for Iceberg table fileFormat, such as Parquet, Orc, Avro, or Iceberg",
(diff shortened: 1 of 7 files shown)


catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTablePropertiesMetadata.java
```diff
@@ -42,6 +45,22 @@ public class IcebergTablePropertiesMetadata extends BasePropertiesMetadata {
...
+  public static final int ICEBERG_DEFAULT_FORMAT_VERSION = 2;
+
+  /**
+   * The explicit Iceberg table format versions Gravitino accepts for {@link #FORMAT_VERSION}. An
+   * unset (empty) value is also allowed and defaults to {@link #ICEBERG_DEFAULT_FORMAT_VERSION}.
+   * Version 3 is required for V3 types such as {@code variant}; extend this set as Gravitino adopts
+   * newer Iceberg writer versions. Note this is a Gravitino policy: Iceberg 1.11.0 itself supports
+   * up to format version 4.
+   */
+  public static final Set<Integer> SUPPORTED_FORMAT_VERSIONS = ImmutableSet.of(2, 3);
```
roryqi: Why don't we support version 1? Could we create a v1 table by Iceberg REST API?

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTablePropertiesMetadata.java
```diff
@@ -81,4 +99,52 @@ public class IcebergTablePropertiesMetadata extends BasePropertiesMetadata {
...
+   * must be an integer in {@link #SUPPORTED_FORMAT_VERSIONS}.
+   *
+   * @param value the raw property value.
+   * @return the parsed format version, or {@link #ICEBERG_DEFAULT_FORMAT_VERSION} when the value is
+   *     unset.
+   * @throws IllegalArgumentException if the value is neither blank nor a version in {@link
+   *     #SUPPORTED_FORMAT_VERSIONS}.
+   */
+  private static Integer decodeFormatVersion(String value) {
+    if (value == null || value.trim().isEmpty()) {
```
roryqi: Could u use `StringUtils.isBlank`?

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTablePropertiesMetadata.java
nevzheng: I updated the accepted range to be 1-4, defaulting to (2). 1-2 should be fully supported, with 3 on the way. and 4 as experimental.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergTablePropertiesMetadata.java
nevzheng: Done! applied in similar places.

roryqi (approved):
LGTM. Thanks for your contribution! Welcome to join the Apache Gravitino community! Enjoy your open source trip!

PR #11959, opened 2026-07-09
[#11943] improvement(iceberg,core): validate warehouse at catalog creation and fail fast with an actionable error

### LINKED ISSUE #11943: [Improvement] Document and guard the URI-shaped warehouse quirk on the Iceberg REST catalog backend

**Purpose of this issue: document a known quirk and open discussion on how (or whether) to guard against it — not to mandate a specific fix.**

**Suggested priority: P3 (usability / `minor`).** No data loss, no blocked functionality, and a trivial workaround exists (remove the property). The cost is purely a confusing first-run experience. Could be argued up to P2 given it's demo-visible and hits anyone copying a hive/jdbc example.

On the `lakehouse-iceberg` catalog with `catalog-backend=rest`, the `warehouse` property is a **catalog selector (a name)** — it's forwarded to the remote IRC as `GET /v1/config?warehouse=<value>`. On the `hive`/`jdbc` backends the same property is a **storage-location URI**. Because the name means two different things by backend, users copying a hive/jdbc example naturally set a URI like `warehouse: "s3://warehouse/"` on a REST catalog.

The catalog is then **created without complaint** (the value is accepted, never validated), and the failure is deferred to the first table/namespace operation:

```
org.apache.iceberg.exceptions.NoSuchWarehouseException:
Couldn't find Iceberg configuration for catalog s3://warehouse/
```

The error gives no hint that the fix is to remove the property (or set it to a catalog *name*). The value is forwarded verbatim: the REST branch of the required-check is skipped ([`IcebergCatalogWrapper.java:93`](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java#L93)), and the receiving IRC treats it as a catalog name ([`IcebergConfigOperations.java:128`](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/rest/IcebergConfigOperations.java#L128) -> [`IcebergCatalogWrapperManager.java:123`](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/IcebergCatalogWrapperManager.java#L123)).

#### From the end-user's perspective (UX matrix)

The confusion is best seen as a matrix over two dimensions — **backend** (`rest` vs `hive`/`jdbc`) and whether `warehouse` **is a URI, a name, or omitted** — comparing what the user *intends* against the *outcome*:

| Backend | `warehouse` | User *intends* | What it *means* here | Outcome |
|---------|-------------|----------------|----------------------|---------|
| `hive`/`jdbc` | `s3://warehouse/` (URI) | "where data lives" | storage location ✅ *matches* | ✅ Works — correct usage |
| `hive`/`jdbc` | *(omitted)* | "use a default" | required, no default | ❌ Rejected **at create**, clear msg: `The 'warehouse' parameter must have a value.` |
| `rest` | *(omitted)* | "just connect" | **default catalog** | ✅ Works — recommended REST setup |
| `rest` | `sales` (a name) | "use my *sales* catalog" | catalog selector ✅ *matches* | ✅ Works — multi-catalog usage |
| `rest` | `s3://warehouse/` (URI) | "where data lives" *(copied from hive/jdbc)* | catalog **name** to look up ✗ *mismatch* | ❌ **The trap.** Create → `200`; first op → `NoSuchWarehouseException: Couldn't find Iceberg configuration for catalog s3://warehouse/` |
| `rest` | `""` (empty) | (accidental) | blank → default catalog | ✅ Works |

The whole issue is **one cell** — `rest` + URI. It's dangerous because on `hive`/`jdbc` that same URI value is *required and correct*, so the user's instinct ("warehouse = my S3 path") is right everywhere except here — and unlike the hive/jdbc mistake, this one is silently accepted at create and only fails later.

#### Flow (what actually happens)

```mermaid
flowchart TD
    U(["User creates a lakehouse-iceberg catalog<br/>with a 'warehouse' value"]) --> B{catalog-backend?}

    B -->|hive / jdbc| H["'warehouse' = STORAGE LOCATION"]
    H --> HV{value provided?}
    HV -->|"s3://warehouse/ (URI)"| HOK["✅ used as the data location — matches intent"]
    HV -->|omitted| HERR["❌ rejected at CREATE, clear message:<br/>'The warehouse parameter must have a value'"]

    B -->|rest| R["Gravitino forwards the value verbatim:<br/>GET /v1/config?warehouse=VALUE"]
    R --> IRC["IRC interprets VALUE as a CATALOG NAME to look up"]
    IRC --> F{name resolves?}
    F -->|"omitted → default, or real name e.g. 'sales'"| ROK["✅ 200 ConfigResponse — catalog selected"]
    F -->|"s3://warehouse/ (URI) — no such name"| RERR["❌ 404 → NoSuchWarehouseException<br/>CREATE returned 200; surfaces only on first use"]

    classDef good fill:#e6ffed,stroke:#2ea043,color:#003300;
    classDef bad fill:#ffebe9,stroke:#cf222e,color:#330000;
    class HOK,ROK good;
    class HERR,RERR bad;
```

The same `s3://warehouse/` edge lands on **green on the hive/jdbc side** and **red on the rest side** — that mirror is the bug in one picture.

#### Reproduction (Docker quickstart)

Environment: `apache/gravitino:latest` (native API on `:8090`) + `apache/gravitino-iceberg-rest:latest` (IRC on `:9001`), as reported on the `gravitino-irc-quickstart` setup.

```bash
# Register a REST-backend catalog with a URI-shaped warehouse (the natural mistake
# when copying a jdbc/hive example). This SUCCEEDS — nothing rejects it.
curl -s -X POST http://localhost:8090/api/metalakes/v3check/catalogs \
  -H 'Accept: application/vnd.gravitino.v1+json' \
  -d '{
    "name": "irc_probe",
    "type": "RELATIONAL",
    "provider": "lakehouse-iceberg",
    "comment": "probe with URI warehouse",
    "properties": {
      "catalog-backend": "rest",
      "uri": "http://gravitino-irc:9001/iceberg",
      "warehouse": "s3://warehouse/"
    }
  }'
# -> HTTP 200, catalog created.

# Any subsequent operation triggers the IRC handshake and fails:
curl -s http://localhost:8090/api/metalakes/v3check/catalogs/irc_probe/schemas
# -> {"code":1006,"type":"NoSuchWarehouseException",
#     "message":"... Couldn't find Iceberg configuration for catalog s3://warehouse/"}

# Removing 'warehouse' (or setting it to a real catalog name) works:
curl -s -X POST http://localhost:8090/api/metalakes/v3check/catalogs \
  -H 'Accept: application/vnd.gravitino.v1+json' \
  -d '{
    "name": "irc_probe_ok",
    "type": "RELATIONAL",
    "provider": "lakehouse-iceberg",
    "properties": {
      "catalog-backend": "rest",
      "uri": "http://gravitino-irc:9001/iceberg"
    }
  }'
curl -s http://localhost:8090/api/metalakes/v3check/catalogs/irc_probe_ok/schemas   # -> 200
```

#### Tests that expose the issue

The server-side behavior is already covered — a URI `warehouse` is just another unknown catalog name and returns `404` (which the Iceberg REST client rethrows as `NoSuchWarehouseException`). Adding the URI case to the existing parameterized test makes it explicit:

```java
// iceberg/iceberg-rest-server/.../rest/TestIcebergConfig.java  (testConfigWithNonExistentWarehouses)
@ValueSource(strings = {"invalid-catalog", "warehouse_123", "s3://warehouse/"})  // s3:// -> 404
```

Verified locally: the added `"s3://warehouse/"` case passes (returns 404) unchanged, confirming the value is treated as a catalog name, not a location.

Related: #5756 made `warehouse` optional for REST (fixed); #9810 proposes hiding it in the Web UI; PR #9811 documented the distinction. This issue covers the remaining gap — the **REST-API / catalog-init path gives no runtime hint** when a URI-shaped value is used.

### How should we improve?

Filing to surface the quirk and gather maintainer input. Some options, in ascending effort / intrusiveness:

- **Option A — Improve the error message only.** Catch `NoSuchWarehouseException` on the REST path and append: "on the REST backend, `warehouse` selects a catalog by name; remove it or set a catalog name." Low effort, but the hint arrives late (first operation, not creation) and threads through Iceberg's exception.
- **Option B — WARN at catalog init when `warehouse` is URI-shaped.** In the branch that already special-cases REST ([`IcebergCatalogWrapper.java:93`](https://github.com/apache/gravitino/blob/main/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ops/IcebergCatalogWrapper.java#L93)):

  ```java
  } else if (IcebergCatalogBackend.REST.equals(catalogBackend)) {
    String warehouse = icebergConfig.get(IcebergConfig.CATALOG_WAREHOUSE);
    if (StringUtils.isNotBlank(warehouse) && warehouse.contains("://")) {
      LOG.warn(
          "The 'warehouse' value '{}' looks like a storage-location URI, but for the REST "
              + "catalog backend 'warehouse' selects a catalog by name. It is sent to the Iceberg "
              + "REST server as a catalog selector and will likely fail with "
              + "NoSuchWarehouseException. Remove 'warehouse' to use the default catalog, or set "
              + "it to the target catalog's name.",
          warehouse);
    }
  }
  ```

  Non-breaking, fires at creation, correctly scoped (hive/jdbc take the other branch). Pairs with a unit test asserting construction succeeds (no throw) plus the URI case in `TestIcebergConfig`.
- **Option C — Reject a URI-shaped `warehouse` at creation.** Fail fast with a clear message. Best UX for the common case, **but** `warehouse` is a server-defined opaque identifier per the Iceberg REST spec, and some non-Gravitino IRC servers legitimately use URI/ARN-shaped identifiers — a hard reject risks breaking valid federation.
- **Option D — Hide/omit `warehouse` for REST in the Web UI.** Tracked separately in #9810; complements but doesn't cover the REST-API path.

**Leaning toward Option B** as the smallest non-breaking change that fires early and gives the exact hint the raw error omits — while avoiding Option C's risk to the legitimate opaque-identifier case. But the intent here is to document the quirk and discuss; happy to go with whatever the maintainers prefer (including doc-only, since PR #9811 already covers part of it).

*Reported by Mark Hoerth; filed on his behalf.*

nevzheng:

This PR fixes the visible configuration error reported in #11943 for a `lakehouse-iceberg` catalog using `catalog-backend=rest`.

For the Iceberg REST backend, `warehouse` is forwarded to `GET /v1/config?warehouse=...` as a server-side catalog selector. It is not a storage location. Previously, a catalog could be created successfully with an unresolvable value such as `s3://warehouse/`, then fail cryptically on first use.

The changes are:

- **Validate the REST `warehouse` selector at creation time.** `IcebergCatalog` opts into create-time initialization only when the backend is REST and `warehouse` is nonblank. Hive, JDBC, and REST catalogs without `warehouse` keep their existing creation behavior.
- **Return an actionable configuration error.** If the REST server rejects the selector with `NoSuchWarehouseException`, creation fails with `IllegalArgumentException` / HTTP 400 and explains that `warehouse` should be removed to use the default catalog or set to a catalog name the REST server recognizes.
- **Keep dependency failures distinct from invalid configuration.** Because the new validation contacts the remote REST service, a transport failure or HTTP 503 during REST initialization is returned as `ConnectionFailedException` / HTTP 502 instead of being presented as an invalid warehouse or a generic internal error. With a nonblank `warehouse`, the failed creation is rolled back and no catalog is persisted.
- **Preserve the exception across Gravitino boundaries.** The catalog classloader, server handler, and Java client preserve `IllegalArgumentException` and `ConnectionFailedException` so callers receive the intended 400 or 502 result.
- **Clarify the property in OpenAPI examples.** The examples contrast storage backends, where `warehouse` is a location, with the REST backend, where it is a catalog selector and may be omitted.

#### Scope and future work

The primary scope of this PR is the `NoSuchWarehouseException` case from #11943. The adjacent transport/503 mapping is included only so the new create-time validation does not confuse a temporarily unavailable REST dependency with an invalid warehouse value.

This PR does **not** attempt to normalize every response or exception produced by the Iceberg REST client. Iceberg also has other response types and endpoint-specific errors, including bad-request, authorization, forbidden, service-failure, and operation-time failures after initialization. Mapping those consistently into Gravitino's error taxonomy should be handled as future work with dedicated scenarios and tests.

On a lakehouse-iceberg catalog with `catalog-backend=rest`, the `warehouse` property is a catalog selector sent to the Iceberg REST server, not a storage path. A user copying a Hive or JDBC example may set it to a storage URI such as `s3://warehouse/`. Before this change, catalog creation returned success and every later attempt to use the catalog failed with a cryptic `NoSuchWarehouseException`.

The catalog should reject that specific misconfiguration at creation time with a useful message. At the same time, a temporarily unavailable remote server must remain distinguishable from an invalid selector.

Fix: #11943

- Creating a REST-backed lakehouse-iceberg catalog with a nonblank `warehouse` that the remote server cannot resolve now fails at creation with HTTP 400 and an actionable message.
- If the remote service is unreachable or returns HTTP 503 while REST initialization is running, the failure is returned as HTTP 502 (`ConnectionFailedException`, error code `1007`). This can occur during creation when `warehouse` is configured, or during lazy initialization on first use when it is omitted.
- Creation behavior for REST catalogs without `warehouse` is unchanged: initialization remains deferred until first use.
- Hive and JDBC behavior is unchanged.
- No property keys are added or removed.

**Unit tests** (`./gradlew :core:test :catalogs:catalog-lakehouse-iceberg:test -PskipITs`):

- `TestCatalogManager#testCreateCatalogValidatesBackendConnection` verifies opt-in create-time validation, rollback after validation failure, and that catalogs which do not opt in are not eagerly initialized.
- `TestIcebergCatalog#testShouldValidateWarehouseProperty` covers REST with a value, missing/blank warehouse, JDBC, and a null backend.
- `TestIcebergCatalogOperations#testHandleRestExceptionTranslatesUnavailableToConnectionFailed` verifies the scoped initialization mapping for a transport failure and HTTP 503.
- `TestIcebergCatalogOperations#testTestConnectionNonRestWarehouseHasNoHint` verifies that the REST-specific warehouse guidance is not shown for other backends.

**Integration tests** (`CatalogIcebergRestIT`, against the embedded Iceberg REST service):

- `testCreateRestCatalogWithUnresolvableWarehouseFailsAtCreate` verifies that an invalid selector returns the actionable error and leaves no catalog behind.
- `testCreateRestCatalogWithValidWarehouseSucceeds` verifies that a named selector recognized by the REST server creates successfully and can list schemas.
- `testCreateRestCatalogWithUnreachableServerAtCreate` verifies that a dependency outage during create-time validation returns `ConnectionFailedException` and persists nothing.
- `testOperationOnUnreachableRestCatalogFailsWithConnectionError` verifies the lazy-initialization path: creation is allowed without `warehouse`, and the first operation reports the unavailable REST dependency.

Full `CatalogIcebergRestIT` run: 43 tests, 1 skipped, 0 failures.

Manual verification covered the three warehouse-selector cases:

- Recognized named selector → HTTP 200 and usable catalog.
- Unrecognized storage-URI-shaped selector → HTTP 400 with the selector-versus-location explanation.
- Omitted selector → HTTP 200 using the server's default catalog.


catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -68,6 +68,24 @@ public ViewCatalog asViewCatalog() {
...
+   * storage-location-shaped URI mistakenly copied from a hive/jdbc example) into a fast, actionable
+   * failure. An omitted {@code warehouse} (the server's default REST catalog) needs no resolution,
+   * so creation is not forced to touch the backend in that case.
+   *
+   * @return {@code true} if a non-blank {@code warehouse} is configured, {@code false} otherwise.
+   */
+  @Override
+  public boolean shouldValidateConnectionForCreate() {
+    Map<String, String> properties = entity().getProperties();
+    return properties != null && StringUtils.isNotBlank(properties.get(IcebergConstants.WAREHOUSE));
```
yuqi1129: I'm not very clear about why we need to validate the connection if `warehouse` is not empty. According to the Java documentation, do you only need to check the connection when `type` is `rest`?  I remember for `jdbc`, `hive`, it's only not empty, so we will change them all?

nevzheng: ack. I updated shouldValidateForConnection adding a filter for REST + connection not empty

nevzheng:
Ready for review thanks! changes: tried to cleanup code / style / comments.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
```diff
@@ -68,6 +68,22 @@ public ViewCatalog asViewCatalog() {
     return (ViewCatalog) ops();
   }
 
+  @Override
+  public boolean shouldValidateConnectionForCreate() {
```
roryqi: Federated REST catalog should say no here. It won't block the creation if remote server is down.

nevzheng: Rory, can you elaborate on your concern?

Here's what I think you mean correct me if I'm off: Gravitino persists the catalog into its own metastore on create, and today create never contacts the remote server. Your concern is that by validating against the remote at create time, we'd now fail the create if the remote REST server is down and registering a federated catalog shouldn't require the remote to be reachable at that exact moment.

My response: I'd like to propose a principle, Gravitino's metastore should only store data known to be valid at create time. [#11943](https://github.com/nevzheng/gravitino/issues/11943) shows the cost of not having it: we accept a catalog that can never work, return 200, and it fails later with a confusing NoSuchWarehouseException. That's a change from today's behavior, and this PR implements it (I've updated it to make the intent clearer). For now it's a focused change, Iceberg REST warehouse, but in principle the same "validate before persist" idea extends to other backends and other create-time checks.

How do you feel about that principle?

Conversely, if we decide the metastore may store unvalidated data (federation pointers we haven't checked), that's a legitimate data-model choice too, but we should document it explicitly. And under that model, [#11943](https://github.com/nevzheng/gravitino/issues/11943) is actually WAI: it's the expected consequence of storing without validating, not a bug.

Either way, I think the useful outcome is to pick the principle deliberately and write it down, rather than leave it implicit.

nevzheng:
Following up on Rory's point that a federated REST catalog shouldn't be treated as broken just because its remote is momentarily unreachable — I've split the failure modes into a proper taxonomy:

- **Bad `warehouse` value** (server reachable, rejects the selector) → `IllegalArgumentException` / **HTTP 400**. User misconfiguration, unchanged.
- **Backend unreachable or 503** (create-time *or* later operations) → `ConnectionFailedException` / **HTTP 502 Bad Gateway** (error code 1007). A downstream-dependency outage, not a Gravitino bug and not a user error.
- **Anything else** → still 500.

So create-time validation no longer conflates "your config is wrong" with "the remote happens to be down" — the latter is now a distinct, honest 502. The translation happens once in the Iceberg layer (`RESTException` → `ConnectionFailedException`); the server maps the type to 502 and the Java client reconstructs it from code 1007, so both the create and operation paths report it consistently.

Covered by an IT for each path (unreachable-at-create and operation-on-dead-backend) plus a unit test on the translation. Full `CatalogIcebergRestIT` run is green.

Note: the base `RESTException` (bare transport failure) and `ServiceUnavailableException` (503) both map to 502; other REST subtypes (4xx caller errors, `ServiceFailureException`) still pass through — I've filed a follow-up to sweep those mappings across all catalog backends rather than expand scope here.

server/src/main/java/org/apache/gravitino/server/web/rest/ExceptionHandlers.java
```diff
@@ -1095,6 +1095,14 @@ public Response handle(OperationType op, String object, String parent, Exception
 
       String errorMsg =
           getBaseErrorMsg(formattedObject, op.name(), formattedParent, getErrorMsg(e));
+
+      // A backend a catalog federates to being unreachable is a downstream-dependency failure, not
+      // an internal Gravitino error: surface it as 502 Bad Gateway so callers can tell a dependency
+      // outage from a server bug.
+      if (e instanceof ConnectionFailedException) {
+        return Utils.connectionFailed(errorMsg, e);
+      }
```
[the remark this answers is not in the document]
nevzheng: Good catch — fixed in c8d65acd4. Added `LOG.warn(errorMsg, e)` before the 502 return, mirroring the sibling `CatalogExceptionHandler`. WARN rather than ERROR: a downstream dependency outage isn't a Gravitino bug, but it should still leave a server-side trace.

clients/client-java/src/main/java/org/apache/gravitino/client/ErrorHandlers.java
```diff
@@ -1374,6 +1374,9 @@ public ErrorResponse parseResponse(int code, String json, ObjectMapper mapper) {
 
     @Override
     public void accept(ErrorResponse errorResponse) {
+      if (errorResponse.getCode() == ErrorConstants.CONNECTION_FAILED_CODE) {
+        throw new ConnectionFailedException("%s", formatErrorMessage(errorResponse));
+      }
```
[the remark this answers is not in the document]
nevzheng: Kept the `"%s"` here deliberately. The plain-message form is only free inside handlers that carry a class-level `@SuppressWarnings("FormatStringAnnotation")` — as `CatalogErrorHandler` does (line 550), which is why its `new ConnectionFailedException(errorMessage)` compiles. `RestErrorHandler` is the un-suppressed base, so passing a non-constant string as the sole format argument to the `@FormatMethod` constructor trips error-prone.

So the choice is: this literal `"%s"` (lint-clean, no suppression), or the plain form plus a class-level suppression on the base handler that disables the check for every throw in it. I'd rather keep the lint active than trade it for cosmetics — and `"%s"` matches the `"Unable to process: %s"` line just below.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
roryqi: Yes, it should throw an exception if the parameter is error.
We should consider the case which the remote catalog is temporarily down. Users will expect federated down won't influence the other.

roryqi:
LGTM.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
nevzheng: Following up with the final implementation: create-time validation is now enabled only for the Iceberg REST backend when `warehouse` is nonblank. Hive and JDBC retain their existing behavior. `TestIcebergCatalog#testShouldValidateWarehouseProperty` covers the REST, missing/blank warehouse, JDBC, and null-backend cases, and `TestCatalogManager#testCreateCatalogValidatesBackendConnection` verifies that catalogs which do not opt in skip validation during creation. This addresses the scope concern.

catalogs/catalog-lakehouse-iceberg/src/main/java/org/apache/gravitino/catalog/lakehouse/iceberg/IcebergCatalog.java
nevzheng: Following up with the implemented behavior: I separated invalid configuration from downstream dependency failures.

- An invalid REST `warehouse` selector returns HTTP 400.
- An unreachable backend or HTTP 503 returns `ConnectionFailedException` / HTTP 502.
- With a nonblank `warehouse`, an outage fails creation cleanly and nothing is persisted.
- Without `warehouse`, creation remains allowed and the dependency failure is reported when the catalog is first used.

This is covered by `testCreateRestCatalogWithUnreachableServerAtCreate`, `testOperationOnUnreachableRestCatalogFailsWithConnectionError`, and `testHandleRestExceptionTranslatesUnavailableToConnectionFailed`.

This resolves the configuration-versus-dependency distinction. Since you subsequently said LGTM, I’m marking the thread resolved.

nevzheng:
@roryqi @yuqi1129 Thanks for the feedback. I’ve addressed and resolved the review threads and updated the PR description to clarify the scope:

- The primary fix is the `NoSuchWarehouseException` case from #11943.
- Invalid REST warehouse selectors now fail at creation with an actionable HTTP 400.
- Transport/503 handling is limited to REST initialization so a temporary dependency outage is not confused with invalid configuration.
- Broader Iceberg REST response mapping is explicitly left as future work.

The targeted unit and integration coverage is included. Could you please take a final look and formally approve if the behavior and scope look good? CI is still running, with no failures currently.

roryqi (approved):
LGTM, thanks @nevzheng for work, thanks @yuqi1129 for review.

PR #11964, opened 2026-07-09
[#11963] fix(cli): support parsing the variant column type

### LINKED ISSUE #11963: [Improvement] CLI cannot parse the variant column type

The native `variant` type was added to Gravitino's type system (#11932), but the CLI's string→type parsing was never updated to recognize it. Running the CLI with `--datatype variant` fails:

```
gravitino table create ... --columns col --datatype variant
# → IllegalArgumentException: Unknown or unsupported type: variant
```

Every other primitive (`string`, `binary`, `uuid`, …) is accepted, so `variant` is an inconsistent gap in the "specify a column by typing its type string" flow.

### How should we improve?

Map `"variant"` to `Types.VariantType.get()` in `TypeConverter.convert(String)` (`clients/cli/`), so the CLI accepts `variant` as a column datatype (case-insensitive). `variant` is a parameterless primitive, so parsing is lossless. Add unit tests covering `TypeConverter.convert("variant")` and `ParseType.toType("variant")`.

nevzheng:

Adds a `variant` case to the CLI's `TypeConverter.convert(String)` so a user can specify a `variant` column from the terminal via the type string (e.g. `--datatype variant`). Previously this threw `IllegalArgumentException: Unknown or unsupported type: variant`, even though `Types.VariantType` exists in the type system.

- `TypeConverter.java`: map `"variant"` → `Types.VariantType.get()`.
- Added unit tests in `TestTypeConverter` and `TestParseType` covering both the direct converter and the full `ParseType.toType("variant")` entry point.

The native `variant` type (#11932) was wired into the type system but not into the CLI's string→type parsing. As a result the CLI could not create or alter a column of type `variant`, leaving a gap in the "specify a column by typing its type string" flow. `variant` is a parameterless primitive, so parsing is lossless.

Fix: #11963

Yes. The CLI now accepts `variant` as a column datatype (case-insensitive), e.g.:

```
gravitino table create ... --columns col --datatype variant
```

Added and ran unit tests:

```
./gradlew :clients:cli:test --tests 'org.apache.gravitino.cli.TestTypeConverter' \
                            --tests 'org.apache.gravitino.cli.TestParseType' -PskipITs
```

Both pass. Tests assert `TypeConverter.convert("variant")` and `ParseType.toType("variant")` return `Types.VariantType`.

### DIFF
--- a/clients/cli/src/main/java/org/apache/gravitino/cli/TypeConverter.java
+++ b/clients/cli/src/main/java/org/apache/gravitino/cli/TypeConverter.java
@@ -76,6 +76,8 @@ public class TypeConverter {
         return Types.StringType.get();
       case "binary":
         return Types.BinaryType.get();
+      case "variant":
+        return Types.VariantType.get();
       default:
         throw new IllegalArgumentException("Unknown or unsupported type: " + typeName);
     }
--- a/clients/cli/src/test/java/org/apache/gravitino/cli/TestParseType.java
+++ b/clients/cli/src/test/java/org/apache/gravitino/cli/TestParseType.java
@@ -47,6 +47,12 @@ public class TestParseType {
   }
 
   @Test
+  public void testParseTypeVariant() {
+    Type type = ParseType.toType("variant");
+    assertThat(type, instanceOf(Types.VariantType.class));
+  }
+
+  @Test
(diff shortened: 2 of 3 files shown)


mchades (approved):
lgtm

Issue #11966, opened 2026-07-09
[Subtask] Parse & validate the X-Gravitino-Active-Roles header


bharos:
### Describe the subtask

Add the header grammar and request-scoped transport for role assumption. A parser produces an `ActiveRoles` model (`ALL` / `NONE` / named list; entries trimmed and de-duplicated; role names case-sensitive; `ALL` and `NONE` are reserved and must appear alone). Malformed value → `400`; a role the caller does not hold → `403`. The validated assertion is carried on the request principal context so it reaches direct access checks, list filtering, and credential vending (including the parallel authorization path). No behavior change when the header is absent.

### Parent issue

#11965

PR #11970, opened 2026-07-09
[#11966] feat(auth): add X-Gravitino-Active-Roles header grammar and parser

bharos:

First step of role assumption (SET ROLE): the header grammar and parser only, with no wiring and no behavior change yet.

- Add the `X-Gravitino-Active-Roles` header constant to `AuthConstants`.
- Add `ActiveRoles`, an immutable value type modeling the declared active roles as `ALL` / `NONE` / `NAMED` (ordered, de-duplicated role names).
- Add `ActiveRolesParser`, which parses a raw header value into `ActiveRoles`:
  - absent / blank value maps to `ALL` (today's behavior);
  - entries are trimmed and de-duplicated; role names are case-sensitive;
  - `ALL` and `NONE` are reserved keywords (exact upper-case) and must each appear alone;
  - a malformed value (empty entry, or a reserved keyword combined with anything) raises `IllegalActiveRolesException` (extends `IllegalArgumentException`).
- Unit tests covering the full grammar, malformed inputs, and the value type.

This is intentionally scoped to the model plus parser so it can be reviewed on its own. Carrying the value on the request, the `400`/`403` mapping, and enforcement follow in later PRs.

Part of #11966.

Implements Phase 1 of the approved design [design-docs/gravitino-role-assumption.md](https://github.com/apache/gravitino/blob/main/design-docs/gravitino-role-assumption.md) (discussion #10894). A caller needs a way to declare which of its roles should be active for a request so authorization can be narrowed; this PR establishes the header grammar that the rest of the feature builds on.

Not yet. It adds the `X-Gravitino-Active-Roles` header name and its grammar as internal building blocks, but nothing reads the header or changes authorization behavior in this PR.

New unit tests (JUnit 5): `TestActiveRolesParser` covers single role, comma-separated lists, whitespace trimming, duplicate collapsing, case-sensitivity, `ALL`/`NONE`, absent/blank values, and malformed inputs; `TestActiveRoles` covers the value type contract. Verified locally:

- `./gradlew :common:test --tests "org.apache.gravitino.auth.TestActiveRolesParser" --tests "org.apache.gravitino.auth.TestActiveRoles"`
- `./gradlew :common:spotlessCheck :common:javadoc`

### DIFF
--- a/common/src/main/java/org/apache/gravitino/auth/ActiveRoleParser.java
+++ b/common/src/main/java/org/apache/gravitino/auth/ActiveRoleParser.java
@@ -24,27 +24,24 @@ import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
- * Parses the value of the {@link AuthConstants#X_GRAVITINO_ACTIVE_ROLES_HEADER} header into an
- * {@link ActiveRoles} declaration.
+ * Parses the {@link AuthConstants#X_GRAVITINO_ACTIVE_ROLES_HEADER} header value into an {@link
+ * ActiveRoles} declaration.
  *
  * <p>Accepted grammar:
  *
  * <ul>
- *   <li>absent, empty, or whitespace-only value &rarr; {@link ActiveRoles#all()} (today's
- *       behavior);
- *   <li>{@link #ALL_KEYWORD} &rarr; {@link ActiveRoles#all()};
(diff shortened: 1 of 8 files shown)


roryqi (approved):
LGTM.

Issue #5820, opened 2024-12-11
[Improvement] Optimize Python client ITs


xunliu:
### What would you like to be improved?

Currently, Gravitino Python client running every integration test will restart Gravitino server in the GitHub Action CI.
It wastes a lot of time and increases risk  

### How should we improve?

1. Only running Gravitino Python client ITs in the IDE, every IT unit will start and stop Gravitino server.
2. In the GitHub Action CI, I will run the Gravitino server through Gradle script. The IT unit don't start and stop Gravitino Server.

PR #11977, opened 2026-07-10
[#5820] test(client-python): Improve IT server lifecycle

henry3260:

This PR improves the Gravitino server lifecycle used by Python client integration tests.

The Gradle `integrationTest` task now uses dedicated `startGravitinoServer` and `stopGravitinoServer` tasks, with `stopGravitinoServer` registered through `finalizedBy` so Gradle attempts to stop the server even when the integration test task fails.

When `START_EXTERNAL_GRAVITINO=true` and the expected external Gravitino server is unavailable, fail the integration test immediately with a clear exception instead of only logging an error and continuing.

This PR also centralizes the `START_EXTERNAL_GRAVITINO` check in `IntegrationTestEnv.use_external_gravitino()` and updates several authorization-related integration tests to reuse that helper.

The Python client integration test setup already supports using an externally started Gravitino server in CI. However, the server lifecycle logic was still embedded directly inside the `integrationTest` task, and the external-server environment check was duplicated across multiple test classes.

Separating start and stop into dedicated Gradle tasks makes the lifecycle easier to understand and safer on failure. Centralizing the environment check also keeps related integration tests consistent.

Failing fast makes a missing or unreachable externally managed server immediately visible to CI and developers, avoiding follow-up failures with less actionable error messages.

Related to #5820.

No. This change only affects Python client integration test infrastructure.

- `git diff --check`

### DIFF
--- a/clients/client-python/build.gradle.kts
+++ b/clients/client-python/build.gradle.kts
@@ -186,11 +186,22 @@ tasks {
     args = listOf("scripts/generate_version.py")
   }
 
-  val integrationTest by registering(VenvTask::class) {
-    doFirst {
+  val stopGravitinoServer by registering {
+    doLast {
+      gravitinoServer("stop")
+    }
+  }
+
+  val startGravitinoServer by registering {
+    finalizedBy(stopGravitinoServer)
+    doLast {
       gravitinoServer("start")
     }
+  }
 
+  val integrationTest by registering(VenvTask::class) {
+    dependsOn(startGravitinoServer)
+    finalizedBy(stopGravitinoServer)
     venvExec = "coverage"
     args = listOf("run", "--branch", "-m", "unittest")
     workingDir = projectDir.resolve("./tests/integration")
@@ -213,10 +224,6 @@ tasks {
       "PYTHONPATH" to "${project.rootDir.path}/clients/client-python"
     ))
     environment = envMap
-
-    doLast {
-      gravitinoServer("stop")
-    }
   }
 
   val unitCoverageReport by registering(VenvTask::class){
--- a/clients/client-python/tests/integration/integration_test_env.py
+++ b/clients/client-python/tests/integration/integration_test_env.py
@@ -67,15 +67,19 @@ class IntegrationTestEnv(unittest.TestCase):
     gravitino_startup_script = None
     gravitino_admin_client: GravitinoAdminClient = None
 
+    @staticmethod
+    def use_external_gravitino() -> bool:
+        return os.environ.get("START_EXTERNAL_GRAVITINO", "").lower() == "true"
+
     @classmethod
     def setUpClass(cls):
-        if (
-            os.environ.get("START_EXTERNAL_GRAVITINO") is not None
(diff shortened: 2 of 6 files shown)


clients/client-python/build.gradle.kts
```diff
@@ -186,11 +186,21 @@ tasks {
...
+
+  val stopGravitinoServer by registering {
+    doLast {
+      gravitinoServer("stop")
+    }
+  }
 
+  val integrationTest by registering(VenvTask::class) {
+    dependsOn(startGravitinoServer)
+    finalizedBy(stopGravitinoServer)
```
jerryshao, line 204: `finalizedBy` only runs if `integrationTest` itself is scheduled to execute. If `startGravitinoServer` (a `dependsOn`) fails — e.g. the server never comes up — `integrationTest` is skipped entirely and `stopGravitinoServer` never fires, leaving a partially-started server process orphaned. I verified this with a minimal Gradle repro (dependency task fails -> finalizer of the dependent task never runs). Not a regression vs. the old single-task version (same gap existed there), but it does mean this only covers *test* failures, not *server-start* failures — might be worth a note in the description, or also finalizing `startGravitinoServer` with the stop task.

clients/client-python/tests/integration/integration_test_env.py
```diff
@@ -67,15 +67,18 @@ class IntegrationTestEnv(unittest.TestCase):
...
     def setUpClass(cls):
-        if (
-            os.environ.get("START_EXTERNAL_GRAVITINO") is not None
-            and os.environ.get("START_EXTERNAL_GRAVITINO").lower() == "true"
-        ):
+        if cls.use_external_gravitino():
             # Maybe Gravitino server already startup by Gradle test command or developer manual startup.
             if not check_gravitino_server_status():
-                logger.error("ERROR: Can't find online Gravitino server!")
+                raise GravitinoRuntimeException(
```
jerryshao, line 79: This changes behavior from logging-and-continuing to failing fast when an external server isn't reachable. Looks like a deliberate improvement, but it's a functional behavior change beyond the stated 'centralize the check' refactor — might be worth calling out explicitly in the PR description.

jerryshao (commented):
Two things worth a look — not blocking, mostly for awareness.

clients/client-python/build.gradle.kts
henry3260: > `finalizedBy` only runs if `integrationTest` itself is scheduled to execute. If `startGravitinoServer` (a `dependsOn`) fails — e.g. the server never comes up — `integrationTest` is skipped entirely and `stopGravitinoServer` never fires, leaving a partially-started server process orphaned. I verified this with a minimal Gradle repro (dependency task fails -> finalizer of the dependent task never runs). Not a regression vs. the old single-task version (same gap existed there), but it does mean this only covers _test_ failures, not _server-start_ failures — might be worth a note in the description, or also finalizing `startGravitinoServer` with the stop task.

Good catch!

jerryshao (approved):

PR #11978, opened 2026-07-10
[#11976] fix(core): refresh built-in job template when content drifts even if version is unchanged

### LINKED ISSUE #11976: [Bug report] Built-in job template run fails with "Source file does not exist" after Gravitino version upgrade
### Version

main branch

### Describe what's wrong

After upgrading Gravitino across versions (e.g. via a rolling update on Kubernetes), running a built-in job template (e.g. `builtin-iceberg-rewrite-data-files`) fails because the server still tries to load the old-version jar path (`gravitino-jobs-<old-version>.jar`) that no longer exists in the container.

The root cause is in `BuiltInJobTemplateEventListener.reconcileBuiltInJobTemplates()` (`core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java`) — it only compares the template's `version` custom field (e.g. `"v1"`) to decide whether to refresh the persisted record. Since the built-in job's logic didn't change, `version` stays at `"v1"` across a Gravitino release bump, so the `executable` path — which is resolved dynamically at startup from the on-disk jar location via `BuiltInJob.resolveExecutable()` (`maintenance/jobs/src/main/java/org/apache/gravitino/maintenance/jobs/BuiltInJob.java`) — is never refreshed in the database, even though the actual jar filename changed with the release.

```
{
  "code": 1002,
  "type": "RuntimeException",
  "message": "Failed to operate object operation [RUN] under [test], reason [Failed to fetch file from URI /opt/gravitino/auxlib/gravitino-jobs-1.4.0-SNAPSHOT.jar]",
  "stack": [
    "java.lang.RuntimeException: Failed to fetch file from URI /opt/gravitino/auxlib/gravitino-jobs-1.4.0-SNAPSHOT.jar",
    "at org.apache.gravitino.job.JobManager.fetchFileFromUri(JobManager.java:812)",
    "at org.apache.gravitino.job.JobManager.createRuntimeJobTemplate(JobManager.java:677)",
    "at org.apache.gravitino.job.JobManager.runJob(JobManager.java:438)",
    "...",
    "Caused by: java.io.IOException: Source file does not exist: /opt/gravitino/auxlib/gravitino-jobs-1.4.0-SNAPSHOT.jar",
    "at org.apache.gravitino.utils.FileFetcher.linkLocalFile(FileFetcher.java:129)",
    "at org.apache.gravitino.utils.FileFetcher.fetchFileFromUri(FileFetcher.java:107)",
    "at org.apache.gravitino.job.JobManager.fetchFileFromUri(JobManager.java:806)"
  ]
}
```

Container actual state after upgrade:

```
$ ls /opt/gravitino/auxlib/
gravitino-cli-2.0.0-SNAPSHOT.jar  gravitino-jobs-2.0.0-SNAPSHOT.jar
```

### How to reproduce

1. Deploy a Gravitino build with built-in job templates enabled (e.g. via Helm chart on Kubernetes)
2. Run any built-in job (e.g. `builtin-iceberg-rewrite-data-files`) — this succeeds and persists the template to the database with `executable = /opt/gravitino/auxlib/gravitino-jobs-<old-version>.jar`
3. Upgrade the Gravitino image to a new version (rolling update), which ships `gravitino-jobs-<new-version>.jar`
4. Run the same built-in job again — fails with the error above

### Additional context

`reconcileBuiltInJobTemplates()` should also detect changes to the `executable` path (and other mutable fields like arguments/configs), not rely solely on the `version` string, when deciding whether to refresh the persisted template. `JobTemplate` already implements a full `equals()`, so the fix can compare `existing.toJobTemplate().equals(newTemplate)` in addition to the version check.

jerryshao:

`BuiltInJobTemplateEventListener.reconcileBuiltInJobTemplates()` decided whether to refresh a persisted built-in job template by comparing only the `version` custom field. This PR makes it also compare the full template content (via `JobTemplate.equals()`) and refresh the persisted record whenever it drifts, regardless of the version field. Also added a regression test, `testReconcileBuiltInJobTemplatesUpdateWhenExecutableChangedButVersionSame`.

The `executable` path of a built-in job template is resolved from the on-disk jar location at server startup (`BuiltInJob.resolveExecutable()`), so it changes across a Gravitino release upgrade even though the job logic version (`"v1"`) stays the same. Since reconciliation only checked `version`, the stale jar path from the old release was never refreshed in the database, causing built-in job runs to fail after upgrade with `Source file does not exist: /opt/gravitino/auxlib/gravitino-jobs-<old-version>.jar`.

Fix: #11976

No user-facing API changes. Built-in job templates will now self-heal their `executable` path (and other fields) on server startup after an upgrade, instead of requiring manual intervention.

Added a unit test verifying that when `executable` differs but `version` is unchanged, the persisted template is updated. Ran `TestBuiltInJobTemplateEventListener` (14 tests, all passing).

### DIFF
--- a/core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
+++ b/core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
@@ -269,7 +269,12 @@ public class BuiltInJobTemplateEventListener implements EventListenerPlugin {
           } else {
             int existingVersion = version(existing.templateContent().customFields());
             int newVersion = version(newTemplate.customFields());
-            if (newVersion > existingVersion) {
+            // The version field only tracks changes to job logic, but fields like `executable`
+            // can drift independently (e.g. the jar path changes with the Gravitino release
+            // version even though the job logic version stays the same), so also refresh the
+            // persisted template whenever its content no longer matches, regardless of version.
+            boolean contentChanged = !existing.toJobTemplate().equals(newTemplate);
+            if (newVersion > existingVersion || contentChanged) {
               updateBuiltInJobTemplate(metalake, existing, newTemplate);
             } else {
               LOG.info("Built-in job template {} under metalake {} is up to date", name, metalake);
--- a/core/src/test/java/org/apache/gravitino/job/TestBuiltInJobTemplateEventListener.java
+++ b/core/src/test/java/org/apache/gravitino/job/TestBuiltInJobTemplateEventListener.java
@@ -416,10 +416,11 @@ public class TestBuiltInJobTemplateEventListener {
     String metalakeName = "test_metalake";
     Map<String, JobTemplate> builtInTemplates = new HashMap<>();
 
+    // Same version and identical content as the entity created by createJobTemplateEntity(...)
     ShellJobTemplate template =
         ShellJobTemplate.builder()
             .withName("builtin-existing")
-            .withComment("same version")
+            .withComment("test")
             .withExecutable("/bin/echo")
             .withCustomFields(Collections.singletonMap("version", "v1"))
             .build();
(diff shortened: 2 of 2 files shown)


jerryshao:
@yuqi1129 can you please help to review?

core/src/main/java/org/apache/gravitino/job/BuiltInJobTemplateEventListener.java
```diff
@@ -269,7 +269,12 @@ void reconcileBuiltInJobTemplates(String metalake, Map<String, JobTemplate> buil
           } else {
             int existingVersion = version(existing.templateContent().customFields());
             int newVersion = version(newTemplate.customFields());
-            if (newVersion > existingVersion) {
+            // The version field only tracks changes to job logic, but fields like `executable`
+            // can drift independently (e.g. the jar path changes with the Gravitino release
+            // version even though the job logic version stays the same), so also refresh the
+            // persisted template whenever its content no longer matches, regardless of version.
+            boolean contentChanged = !existing.toJobTemplate().equals(newTemplate);
+            if (newVersion > existingVersion || contentChanged) {
```
yuqi1129, line 277: Do we allow downgrading the version here if the content has been changed?

jerryshao: No. I don't think it is a valid case. We should always increase the version number if the content is changed.

yuqi1129: Okay, let's leave it as it is.

yuqi1129 (approved):

PR #11984, opened 2026-07-11
[#11983] test(iceberg): add end-to-end coverage for round-trip of remaining V2 column types

### LINKED ISSUE #11983: [Improvement] Add end-to-end integration test coverage for Iceberg V2 column types
What would you like to be improved?

On #11927, Mark asked for test coverage of adjacent Iceberg types (V2 UUID; struct/array/map), verified across both the IRC/REST and native interfaces — https://github.com/apache/gravitino/issues/11927#issuecomment-4909915786

Following up on that: the Iceberg catalog integration test (`CatalogIcebergBaseIT`, run by both `CatalogIcebergRestIT` and `CatalogIcebergHiveIT`) only round-trips a subset of V2 column types end-to-end through a live backend — `int`, `long`, `date`, `string`, `struct`, `time`, `timestamp`.

The remaining V2 types — `boolean`, `float`, `double`, `uuid`, `fixed`, `binary`, `decimal`, `list`, `map` — are unit-tested at the converter level (`TestConvertUtil`) but are never created and loaded through the real REST/IRC and Hive backends. That leaves the backend/serialization path for these types (notably the parameterized/nested ones — `decimal` precision/scale, `fixed` length, `list`, `map`) unverified end-to-end.

How should we improve?

Add an integration test to `CatalogIcebergBaseIT` that creates a table with the remaining V2 types through the native metadata API, loads it back, and asserts both that the Gravitino types round-trip unchanged and that the underlying Iceberg schema stores the expected Iceberg types. As all are valid V2 types, it runs on both backends (REST/IRC and Hive) with no assumptions.

nevzheng:

Adds `testCreateAndLoadIcebergTableV2Types` to `CatalogIcebergBaseIT`. It creates a table containing the V2 column types that were not previously exercised end-to-end — `boolean`, `float`, `double`, `uuid`, `fixed`, `binary`, `decimal`, `list`, `map` — through the native metadata API, loads it back, and asserts that (a) the Gravitino types round-trip unchanged and (b) the underlying Iceberg schema stores the expected Iceberg types. Because these are all valid V2 types, the test runs on both backends via the existing subclasses (`CatalogIcebergRestIT`, `CatalogIcebergHiveIT`) with no assumptions.

`CatalogIcebergBaseIT` only round-tripped a subset of V2 types through a live backend (`int`, `long`, `date`, `string`, `struct`, `time`, `timestamp`). The converter mappings for the remaining types are unit-tested in `TestConvertUtil`, but were never verified through the real REST/IRC and Hive backends — so backend/serialization regressions for `decimal` (precision/scale), `fixed` (length), `list`, and `map` in particular could slip through. This closes that gap.

Fix: #11983

No. Test-only.

This PR adds the test (`testCreateAndLoadIcebergTableV2Types`), and it passes. Run locally on both backends via `-PskipTests -PskipDockerTests=false`:

- `CatalogIcebergRestIT.testCreateAndLoadIcebergTableV2Types` — passed (1 test, 0 failures).
- `CatalogIcebergHiveIT.testCreateAndLoadIcebergTableV2Types` — passed (1 test, 0 failures).


nevzheng:
@mchades @roryqi, could you please take a look when you have a chance? This PR is ready for review. Thanks!

catalogs/catalog-lakehouse-iceberg/src/test/java/org/apache/gravitino/catalog/lakehouse/iceberg/integration/test/CatalogIcebergBaseIT.java
```diff
@@ -632,6 +632,86 @@ void testTimestampTypeConversion() {
...
+    // so this runs unconditionally under both subclasses of this base class.
+    Column[] columns =
+        new Column[] {
+          Column.of("c_boolean", Types.BooleanType.get(), "boolean col"),
+          Column.of("c_float", Types.FloatType.get(), "float col"),
+          Column.of("c_double", Types.DoubleType.get(), "double col"),
+          Column.of("c_uuid", Types.UUIDType.get(), "uuid col"),
+          Column.of("c_fixed", Types.FixedType.of(16), "fixed col"),
+          Column.of("c_binary", Types.BinaryType.get(), "binary col"),
+          Column.of("c_decimal", Types.DecimalType.of(10, 2), "decimal col"),
```
roryqi, line 650: Why do we need to support struct?

nevzheng: Could you clarify what you’re asking?

- If you mean, “Why isn’t `struct` included here?”: it is already covered by `testCreateAndLoadIcebergTable` through `createColumns()`. This test covers V2 types not otherwise tested end to end.
- If you mean, “Why should we support `decimal`?”: it is a supported Iceberg V2 type without existing end-to-end coverage. This test verifies that its precision and scale survive the live backend create-and-load path.

Refs: https://iceberg.apache.org/spec/#primitive-types

mchades (approved):
lgtm. @roryqi do you have more comments?

roryqi (approved):

catalogs/catalog-lakehouse-iceberg/src/test/java/org/apache/gravitino/catalog/lakehouse/iceberg/integration/test/CatalogIcebergBaseIT.java
roryqi: Thanks.

Issue #11737, opened 2026-06-18
[Improvement] Make entity store cache usable in multi-node deployments


yuqi1129:
### What would you like to be improved?

The entity store cache (`gravitino.cache.enabled`, default `true`) only invalidates **locally**. In a multi-node deployment it must be set to `false`, which significantly hurts some catalogs (Iceberg especially). The jcasbin cache already solves cross-node invalidation via `entity_change_log` and `EntityChangeLogPoller`; the entity store cache is simply not wired into that channel.

This tracks a design to make `enabled=true` usable in multi-node by invalidating the entity store cache across nodes through the existing change-log mechanism.

### How should we improve?

A design doc is proposed (see linked PR). It covers:

- **Consistency model**: writes stay strongly consistent (DB plus optimistic lock); single-entity and relation reads become cross-node eventually consistent (within one poll interval).
- **The relation-cache difficulty**: relations are cached bidirectionally, so a remote node cannot always derive the reverse keys to invalidate.
- **Two options**: (A, main) coarse invalidation by metalake, a small change that reuses the schema; (B, alternative) precise per-endpoint invalidation, more code but no stampede.
- **Coverage gap**: auth and metadata entities (role/user/group/tag/policy) emit no change-log today and must gain emit points.

Related: #11736 (poller reliability issues found while designing this).

yuqi1129:
Tracking refinement after reviewing the design and the current implementation.

New subtasks:

- #12149 defines a shared, default-deny cache eligibility policy and audits action-sensitive and connector-backed entities.
- #12150 adds explicit fresh/cache-bypass reads for consistency-sensitive operations.
- #12151 makes change-log emission transactionally complete across cacheable entity types and mutation paths.
- #12152 batches tag/policy inheritance relation reads and adds executable database-performance acceptance criteria.

Existing implementation subtasks remain:

- #11961 removes generic relation-list caching.
- #12018 implements an entity-type allowlist. Its decisions must be reconciled with the broader policy audit in #12149 before either issue closes, especially for metalake state, policy/job data, and connector-backed schema/table data.
- #12019 wires Caffeine to cross-node change-log invalidation.
- #12020 provides the optional Redis implementation, subject to an explicit DB/cache failure contract.

Correctness dependencies:

- #11736 must not be treated as fully resolved by listener retry alone. PR #11739 explicitly leaves the auto-increment commit-order gap as a known limitation; the entity-cache design needs either prevention or a documented fail-safe/bounded-staleness contract.
[...]

PR #12006, opened 2026-07-13
[#11961] refactor(core): Simplify EntityCache and remove relation data from CaffeineEntityCache

### LINKED ISSUE #11961: [Subtask] Simplify EntityCache and remove relation data from CaffeineEntityCache
### Describe the subtask

Simplify the `EntityCache` interface and the default `CaffeineEntityCache` implementation by removing relation-list caching from `CaffeineEntityCache`.

Currently `CaffeineEntityCache` stores both single-entity entries and relation query results. To keep relation data consistent, it introduces relation keys, prefix indexes, reverse indexes, and bidirectional invalidation logic. This makes both reads and writes hard to reason about, and makes it nearly impossible for users to implement another `EntityStore`-compatible storage/cache path based on the current interfaces.

There are still a small number of APIs that directly depend on relation operations, so this change must also validate the runtime impact before it can be merged. In particular, after removing relation data from the cache, we need to verify that the affected APIs do not suffer a significant performance regression in real workloads or representative benchmarks.

This subtask should:

- Keep the default cache focused on single-entity lookup entries.
- Remove relation data storage from `CaffeineEntityCache`.
- Simplify or remove relation-specific cache APIs and indexes that are no longer needed by the default implementation.
- Let relation/list operations fall back to `EntityStore` unless a future cache implementation explicitly supports them.
- Update related tests and benchmarks to match the simplified behavior.
- Identify the remaining APIs that directly use relation operations and validate their performance after relation cache removal.

The goal is to make the cache SPI easier to implement and maintain, and to unblock the broader multi-node entity cache work tracked in #11737.

### Parent issue

yuqi1129:

- Remove relation-list caching from the entity cache: `EntityCache` now extends only `SupportsEntityStoreCache`; `SupportsRelationEntityCache`, `EntityCacheRelationKey`, `ReverseIndexCache` and `ReverseIndexRules` are deleted.
- `CaffeineEntityCache` stores one entity per `EntityCacheKey`. The radix prefix index is kept only for hierarchical cascade invalidation, and child matching now uses `identifier + "."` so siblings sharing a name prefix (`catalog1` vs `catalog10`) are no longer evicted together.
- `USER` / `GROUP` / `ROLE` entities are no longer cached: their materialized form embeds relation-derived data (securable objects, role names) that only the removed reverse index could keep consistent.
- Relation operations in `RelationalEntityStore` (`listEntitiesByRelation`, `batchListEntitiesByRelation`, `getEntityByRelation`) always fall back to the backend; relation writes invalidate the single-entity entries of the identifiers involved.
- Update unit tests, jcstress tests and JMH benchmarks accordingly.

The relation keys, prefix/reverse indexes and bidirectional invalidation made the cache hard to reason about and the SPI nearly impossible to re-implement. Keeping the default cache focused on single-entity entries unblocks the multi-node entity cache work (#11737).

APIs that now always hit the DB for relation lists: `TagManager`, `PolicyManager`, `FutureGrantManager`, `RoleManager`, `OwnerManager` (one indexed query per call, same as the previous cache-miss path). Performance validation notes are tracked in the issue.

Fix: #11961

No API or configuration changes. Behavior change: relation query results and role/user/group entities are no longer served from cache.

- Rewritten `TestCaffeineEntityCacheInvalidation` (hierarchy cascade, sibling-prefix safety, non-cacheable types, model-version invalidation).
- Updated `TestEntityStorageRelationCache` to assert store-level freshness (grant/revoke/rename visibility) instead of cache internals; updated `TestRelationalEntityStore`, `TestCacheConfig`, `TestCacheIndex`, `TestEntityCacheKey`.
- `./gradlew :core:test -PskipITs` (h2 variants and all cache/storage tests pass; mysql/postgresql parameterized variants fail in `init()` due to local Docker environment, unrelated to this change).


core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -400,8 +318,13 @@ public void insertRelation(
       boolean override)
       throws IOException {
     backend.insertRelation(relType, srcIdentifier, srcType, dstIdentifier, dstType, override);
-    cache.invalidate(srcIdentifier, srcType, relType);
-    cache.invalidate(dstIdentifier, dstType, relType);
+    // Defensive: relation results are not cached, and no currently cached entity type embeds
+    // relation-derived data (the only types that do — USER/GROUP/ROLE — are excluded from the
+    // cache), so these invalidations are a no-op today. They are kept so that if a cached type ever
+    // starts materializing this relation, its stale single-entity entry is dropped on a relation
+    // write.
```
diqiu50: I don't think this comment is accurate anymore. Specifically, *"so these invalidations are a no-op today"* doesn't appear to be correct.

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -80,33 +85,34 @@ public class CaffeineEntityCache extends BaseEntityCache {
...
-  private final Cache<EntityCacheRelationKey, List<Entity>> cacheData;
-
-  /** Cache reverse index structure. */
-  private ReverseIndexCache reverseIndex;
+  private final Cache<EntityCacheKey, Entity> cacheData;
 
-  /** Cache Index structure. */
-  private RadixTree<EntityCacheRelationKey> cacheIndex;
+  /** Prefix index over cache keys, used for cascading removal of descendant entries. */
+  private RadixTree<EntityCacheKey> cacheIndex;
```
diqiu50: cacheIndex is handled outside the lock in size().

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -30,34 +30,39 @@
...
+ * entity also drops every cached descendant entry (e.g. invalidating a catalog drops the cached
+ * schemas and tables under it).
+ *
+ * <p>Relation query results are NOT cached by this implementation; relation and list operations
+ * always fall back to the {@code EntityStore}. Entity types whose materialized form embeds
+ * relation-derived data ({@code USER}, {@code GROUP}, {@code ROLE}) are excluded from caching
+ * entirely, because without relation tracking their entries could not be invalidated when the
+ * referenced entities change.
+ */
 public class CaffeineEntityCache extends BaseEntityCache {
```
diqiu50, line 64: The function of withMultipleKeyCacheLock  is unused

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
```diff
@@ -80,33 +85,34 @@ public class CaffeineEntityCache extends BaseEntityCache {
...
+   * embedded data also goes stale through a mutation on a different entity. For example, deleting
+   * or renaming a securable object changes a role's materialized form, and deleting or renaming a
+   * role changes a user's/group's role names. Such a mutation touches neither this entity's own key
+   * nor any hierarchy ancestor of it, so neither the write-path invalidation nor the prefix cascade
+   * in {@link #invalidateHierarchy} would evict it; only the (now removed) reverse index could.
+   * Caching them would therefore serve stale authorization data.
+   */
+  private static final Set<Entity.EntityType> NON_CACHEABLE_TYPES =
+      Sets.immutableEnumSet(
+          Entity.EntityType.USER, Entity.EntityType.GROUP, Entity.EntityType.ROLE);
```
diqiu50: NON_CACHEABLE_TYPES (USER/GROUP/ROLE) is a correctness constraint — caching them serves stale authorization data. But it's a private field in CaffeineEntityCache, while SupportsEntityStoreCache#put documents nothing about it. Since this PR aims to make the SPI re-implementable for #11737, suggest moving the check to BaseEntityCache.put and documenting it on the interface.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
yuqi1129: Yeah, you are right, and I will change it.

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
yuqi1129: I will try to make it in a lock/synchronize code block

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
yuqi1129: OK

core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
yuqi1129: Indeed. I will move `NON_CACHEABLE_TYPES` to the right position.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -396,37 +339,12 @@ public <E extends Entity & HasIdentifier> E getEntityByRelation(
...
-
-          return r;
-        });
+    List<E> backendEntities = backend.listEntitiesByRelation(relType, srcIdentifier, srcType, true);
+    return backendEntities.stream()
+        .filter(e -> e.nameIdentifier().equals(destEntityIdent))
+        .findFirst()
+        .orElseThrow(
+            () -> new NoSuchEntityException("No such entity with ident: %s", destEntityIdent));
   }
```
lasdf1234, line 388: Could we remove     return backendEntities.stream()
        .filter(e -> e.nameIdentifier().equals(destEntityIdent))
        .findFirst()
        .orElseThrow(
            () -> new NoSuchEntityException("No such entity with ident: %s", destEntityIdent)); ?
Directly call the method of the backend?

yuqi1129:
@diqiu50 
Please help to take a look again.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
yuqi1129: Changed.

yuqi1129:
@jerryshao 
This is first PR about entity store cache.

core/src/main/java/org/apache/gravitino/cache/BaseEntityCache.java
```diff
@@ -30,8 +32,31 @@
...
+   * embedded data also goes stale through a mutation on a different entity. For example, deleting
+   * or renaming a securable object changes a role's materialized form, and deleting or renaming a
+   * role changes a user's/group's role names. Such a mutation touches neither this entity's own key
+   * nor any hierarchy ancestor of it, so neither the write-path invalidation nor a prefix cascade
+   * over the entity hierarchy would evict it. Caching them would therefore serve stale
+   * authorization data.
+   */
+  private static final Set<Entity.EntityType> NON_CACHEABLE_TYPES =
+      Sets.immutableEnumSet(
+          Entity.EntityType.USER, Entity.EntityType.GROUP, Entity.EntityType.ROLE);
```
jerryshao: My feeling is that to deliberately define the cacheable types, rather than the non-cacheable types would be better.

Your solution will blindly cache the new type entities; if that entity doesn't fit your current design, it will lead to an issue. So I think an explicit list is better. After we fully validate the new type, we can add it to the list explicitly.

roryqi (approved):
LGTM.

diqiu50 (approved):
LGTM

PR #12007, opened 2026-07-13
[MINOR] fix(jdbc): Quote comments in generated DDL

yuqi1129:

- Add a shared utility for escaping SQL string literal content.
- Escape table, column, and schema comments in generated JDBC DDL.
- Apply the escaping to MySQL, PostgreSQL, Doris, and StarRocks operations.
- Add regression tests covering quotes, backslashes, semicolons, and SQL comment tokens.

JDBC catalog comments were inserted directly into generated DDL statements. Comments containing quotes could produce malformed SQL or change the structure of the generated statement.

The change ensures comment content remains inside its SQL string literal.

Yes. JDBC table, column, and PostgreSQL schema comments containing quotes or backslashes are now handled correctly.

There is no API or configuration change.

- `./gradlew spotlessApply`
- `env SKIP_DOCKER_TESTS=true ./gradlew :catalogs:catalog-jdbc-common:test :catalogs:catalog-jdbc-mysql:test :catalogs:catalog-jdbc-postgresql:test :catalogs:catalog-jdbc-doris:test :catalogs:catalog-jdbc-starrocks:test -PskipITs -PskipDockerTests=true`

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
@@ -22,6 +22,7 @@ import static org.apache.gravitino.catalog.doris.DorisCatalog.DORIS_TABLE_PROPER
 import static org.apache.gravitino.catalog.doris.DorisTablePropertiesMetadata.DEFAULT_REPLICATION_FACTOR_IN_SERVER_SIDE;
 import static org.apache.gravitino.catalog.doris.DorisTablePropertiesMetadata.REPLICATION_FACTOR;
 import static org.apache.gravitino.catalog.doris.utils.DorisUtils.generatePartitionSqlFragment;
+import static org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils.escapeSqlLiteral;
 import static org.apache.gravitino.rel.Column.DEFAULT_VALUE_NOT_SET;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -136,7 +137,7 @@ public class DorisTableOperations extends JdbcTableOperations {
 
     // Add table comment if specified
     if (StringUtils.isNotEmpty(comment)) {
-      sqlBuilder.append(" COMMENT \"").append(comment).append("\"");
+      sqlBuilder.append(" COMMENT \"").append(escapeSqlLiteral(comment, '"')).append("\"");
     }
 
     // Add Partition Info
@@ -796,7 +797,7 @@ public class DorisTableOperations extends JdbcTableOperations {
           newComment = StringIdentifier.addToComment(identifier, newComment);
         }
       }
-      alterSql.add("MODIFY COMMENT \"" + newComment + "\"");
+      alterSql.add("MODIFY COMMENT \"" + escapeSqlLiteral(newComment, '"') + "\"");
     }
 
     if (CollectionUtils.isEmpty(alterSql)) {
@@ -848,7 +849,8 @@ public class DorisTableOperations extends JdbcTableOperations {
     }
     String col = updateColumnComment.fieldName()[0];
 
-    return String.format("MODIFY COLUMN `%s` COMMENT '%s'", col, newComment);
+    return String.format(
+        "MODIFY COLUMN `%s` COMMENT '%s'", col, escapeSqlLiteral(newComment, '\''));
   }
 
   private String addColumnFieldDefinition(TableChange.AddColumn addColumn) {
@@ -873,7 +875,10 @@ public class DorisTableOperations extends JdbcTableOperations {
     }
     // Append comment if available
     if (StringUtils.isNotEmpty(addColumn.getComment())) {
-      columnDefinition.append("COMMENT '").append(addColumn.getComment()).append("' ");
+      columnDefinition
+          .append("COMMENT '")
+          .append(escapeSqlLiteral(addColumn.getComment(), '\''))
+          .append("' ");
     }
 
     // Append position if available
@@ -985,7 +990,7 @@ public class DorisTableOperations extends JdbcTableOperations {
 
     // Add column comment if specified
     if (StringUtils.isNotEmpty(column.comment())) {
-      sqlBuilder.append("COMMENT '").append(column.comment()).append("' ");
+      sqlBuilder.append("COMMENT '").append(escapeSqlLiteral(column.comment(), '\'')).append("' ");
     }
     return sqlBuilder;
   }
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
@@ -19,6 +19,7 @@
 package org.apache.gravitino.catalog.doris.utils;
 
 import static org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils.escapeSqlLiteral;
+import static org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils.unescapeSqlLiteral;
 
 import com.google.common.collect.ImmutableList;
 import java.util.Arrays;
@@ -95,8 +96,10 @@ public final class DorisUtils {
       if (isProperties) {
         final Matcher matcherProperties = patternProperties.matcher(line);
         if (matcherProperties.find()) {
-          final String key = matcherProperties.group(1).trim();
-          String value = matcherProperties.group(2).trim();
+          // generatePropertiesSql escapes keys and values for double-quoted literals, and SHOW
+          // CREATE echoes them escaped; unescape both so callers observe the original text.
+          final String key = unescapeSqlLiteral(matcherProperties.group(1).trim(), '"');
+          String value = unescapeSqlLiteral(matcherProperties.group(2).trim(), '"');
           properties.put(key, value);
         }
       }
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
@@ -18,6 +18,8 @@
  */
 package org.apache.gravitino.catalog.doris.utils;
 
+import static org.apache.gravitino.catalog.jdbc.utils.JdbcConnectorUtils.escapeSqlLiteral;
+
 import com.google.common.collect.ImmutableList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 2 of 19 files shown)


catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/operations/StarRocksTableOperations.java
```diff
@@ -481,7 +482,10 @@ private String addColumnFieldDefinition(TableChange.AddColumn addColumn) {
 
     // Append comment if available
     if (StringUtils.isNotEmpty(addColumn.getComment())) {
-      columnDefinition.append("COMMENT '").append(addColumn.getComment()).append("' ");
+      columnDefinition
+          .append("COMMENT '")
+          .append(escapeSqlLiteral(addColumn.getComment(), '\''))
+          .append("' ");
```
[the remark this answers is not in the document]
yuqi1129: Added a focused TableChange.addColumn SQL-generation regression test covering single/double quotes and backslashes in 41967b107d.

catalogs/catalog-jdbc-postgresql/src/main/java/org/apache/gravitino/catalog/postgresql/operation/PostgreSqlTableOperations.java
```diff
@@ -689,7 +697,7 @@ private List<String> addColumnFieldDefinition(
               + col
               + PG_QUOTE
               + IS
-              + addColumn.getComment()
+              + escapeSqlLiteral(addColumn.getComment(), '\'')
               + "';");
```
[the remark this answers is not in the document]
yuqi1129: Added a focused TableChange.addColumn SQL-generation regression test covering single/double quotes and backslashes in 41967b107d.

catalogs/catalog-jdbc-mysql/src/main/java/org/apache/gravitino/catalog/mysql/operation/MysqlTableOperations.java
```diff
@@ -460,7 +461,10 @@ private String addColumnFieldDefinition(TableChange.AddColumn addColumn) {
     }
     // Append comment if available
     if (StringUtils.isNotEmpty(addColumn.getComment())) {
-      columnDefinition.append("COMMENT '").append(addColumn.getComment()).append("' ");
+      columnDefinition
+          .append("COMMENT '")
+          .append(escapeSqlLiteral(addColumn.getComment(), '\''))
+          .append("' ");
```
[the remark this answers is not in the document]
yuqi1129: Added a focused TableChange.addColumn SQL-generation regression test covering single/double quotes and backslashes in 41967b107d.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -873,7 +875,10 @@ private String addColumnFieldDefinition(TableChange.AddColumn addColumn) {
     }
     // Append comment if available
     if (StringUtils.isNotEmpty(addColumn.getComment())) {
-      columnDefinition.append("COMMENT '").append(addColumn.getComment()).append("' ");
+      columnDefinition
+          .append("COMMENT '")
+          .append(escapeSqlLiteral(addColumn.getComment(), '\''))
+          .append("' ");
```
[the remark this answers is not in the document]
yuqi1129: Added a focused TableChange.addColumn SQL-generation regression test covering single/double quotes and backslashes in 41967b107d.

catalogs/catalog-jdbc-doris/src/test/java/org/apache/gravitino/catalog/doris/utils/TestDorisUtils.java
```diff
@@ -92,6 +97,12 @@ public void testExtractTablePropertiesFromSql() {
     result = DorisUtils.extractPropertiesFromSql(createTableSql);
     assertEquals("value1", result.get("property1"));
     assertEquals("comment", result.get("comment"));
+
+    createTableSql =
+        "CREATE DATABASE `test`\nPROPERTIES (\n"
+            + "\"key\"name\" = \"owner's \"comment\" D:\\data; --\"\n)";
+    result = DorisUtils.extractPropertiesFromSql(createTableSql);
+    assertEquals("owner's \"comment\" D:\\data; --", result.get("key\"name"));
```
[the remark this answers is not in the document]
yuqi1129: The test string does compile (the double quotes are escaped as \"), but the round-trip point is valid: the write path now escapes while the read path returned raw captured text. Fixed in 89687d9f71 — added `JdbcConnectorUtils.unescapeSqlLiteral` (inverse of `escapeSqlLiteral`, also tolerating backslash-style escapes), used it in `DorisUtils.extractPropertiesFromSql` for both key and value, and updated this test to feed escaped SHOW CREATE output and assert the original text comes back. StarRocks' private comment unescaper now delegates to the shared helper.

catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/utils/StarRocksUtils.java
```diff
@@ -67,7 +68,9 @@ public class StarRocksUtils {
           "(?:^|\\s|\\))DISTRIBUTED\\s+BY\\s+(?:RANDOM\\b|\\w+\\s*\\()", Pattern.CASE_INSENSITIVE);
 
   private static final Pattern TABLE_COMMENT_PATTERN =
-      Pattern.compile("COMMENT\\s*\"([^\\(]+?)\\s*\\(From Gravitino,.*\\)\"");
+      Pattern.compile(
+          "COMMENT\\s*\"((?:\\\\.|\"\"|[^\"\\\\])*)\\s+"
+              + "\\(From Gravitino, DO NOT EDIT: gravitino\\.v\\d+\\.uid-?\\d+\\)\"");
```
diqiu50, line 73: The regex is now stricter. How can we remain compatible with existing data?

catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/operations/StarRocksTableOperations.java
```diff
@@ -170,7 +171,17 @@ protected String generateAlterTableSql(
...
         String newComment = updateComment.getNewComment();
-        alterSql.add("MODIFY COMMENT \"" + newComment + "\"");
+        if (StringIdentifier.fromComment(newComment) == null) {
+          lazyLoadTable = getOrCreateTable(databaseName, tableName, lazyLoadTable);
+          StringIdentifier identifier = StringIdentifier.fromComment(lazyLoadTable.comment());
+          if (identifier != null) {
+            newComment = StringIdentifier.addToComment(identifier, newComment);
+          }
+        }
+        if (StringUtils.isNotEmpty(newComment)) {
```
diqiu50, line 181: We need to check the mark is exist

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
diqiu50, line 88: Doris's property-parsing regex "(.*)"\s*=\s*"(.*)" (in DorisUtils.java, line ~86) is greedy and escape-unaware. The PR now lets values legally contain " and = (via escaping), but only added unescapeSqlLiteral after the match — the regex itself was never updated. If a value contains both " and =, the regex can split at the wrong point, silently storing a corrupted key/value pair with no error.

catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/utils/StarRocksUtils.java
yuqi1129: Good question. The new regex stays compatible with existing data:

1. **Marker part** — `(From Gravitino, DO NOT EDIT: gravitino.v1.uid...)` matches exactly the format `StringIdentifier` writes, which is the only marker format that has ever been persisted (`CURRENT_FORMAT_VERSION = 1`). So every comment Gravitino has written is still matched.
2. **Comment body** — the new group `(?:\\.|""|[^"\\])*` is actually *more* tolerant than the old `[^\(]+?`: the old pattern failed whenever the comment itself contained a `(`, while the new one handles that fine.
3. The only stricter requirement is that quotes must be escaped — but a comment containing a raw `"` already produced malformed SQL under the previous write path, so there is no valid stored data that relied on the old, unescaped form.

So for anything Gravitino previously wrote, the new regex parses it the same or better. The round-trip stays covered by the escaping/unescaping tests.

catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java
yuqi1129: Thanks — I looked into this carefully, and unfortunately an escape-aware regex here actually breaks the round-trip against real Doris.

The key point is what `SHOW CREATE DATABASE` echoes back. Doris does **not** re-emit the value with escaped/doubled quotes; it echoes the comment with **bare (un-escaped) double quotes**. For example, a comment `owner's "comment" C:\tmp; --` comes back as:

```
"comment" = "owner's "comment" C:\tmp; --"
```

I verified this against the `TestDorisDatabaseOperations` docker IT: switching to an escape-aware, non-greedy group made the value truncate at the first inner quote (`owner's`), because that group stops at the first bare `"`.

The greedy `"(.*)"..."(.*)"` pattern is what makes this work: it captures the value up to the **last** quote on the line, which is exactly right for Doris's one-property-per-line, bare-quote output. Because Doris echoes bare quotes, the value is inherently ambiguous, and "value = everything up to the last quote on this line" is the only sane heuristic.

So the escaping on the write path is still needed (to send valid SQL to Doris), but on the read path there is nothing to disambiguate via escaping. The only residual limitation is a value that itself contains a literal `"="`, which we accept as a known edge case rather than regress the normal round-trip. I have reverted the regex change and kept the greedy pattern.

catalogs/catalog-jdbc-starrocks/src/main/java/org/apache/gravitino/catalog/starrocks/operations/StarRocksTableOperations.java
yuqi1129: I looked into this more closely against the StarRocks docker IT, and it turns out the second `addToComment` should stay unconditional — guarding it actually breaks the round-trip.

The reason is the asymmetry between write and read:

- **Read** (`StarRocksUtils.extractTableCommentFromSql`) strips exactly **one** (outermost) identifier marker from the `SHOW CREATE TABLE` output.
- **Write** must therefore append one **sacrificial** marker on top of the preserved identifier. `generateCreateTableSql` already does this unconditionally (it always calls `addToComment(DUMMY_ID, comment)` even when the comment already carries a real identifier), and `UpdateComment` mirrors it.

So the stored comment is `body (uid42) (uid-1)`, and after the read strips the outer `(uid-1)`, the real identifier `(uid42)` survives — which is what the JDBC layer needs to correlate the table with its Gravitino entity.

I verified this with `TestStarRocksTableOperations.testTableCommentWithSqlLiteralCharacters`: adding an `fromComment(...) == null` guard made the write store only `body (uid42)`, the read then stripped `(uid42)`, and the loaded comment lost its identifier entirely. I have reverted that change, so the behavior is unchanged from before. Thanks for the careful look here.

diqiu50 (approved):
LGTM

PR #12010, opened 2026-07-13
[#12009] improvement(core): Support custom case sensitivity normalization in Capability

### LINKED ISSUE #12009: [Improvement] Support custom case sensitivity normalization in Capability

`CapabilityHelpers` currently normalizes names using a hardcoded rule: if `caseSensitiveOnName(scope)` is unsupported, it lowercases the name. This doesn't work for catalogs with different native case sensitivity behavior — e.g. a catalog that normalizes to uppercase by default, or one that supports both a case-sensitive quoted form and a case-insensitive unquoted form (such as Oracle).

### How should we improve?

Add a `default String normalizeName(Scope scope, String name)` method to `Capability`, defaulting to the existing lowercase-normalization behavior for backward compatibility. `CapabilityHelpers` should call `normalizeName` instead of re-implementing the normalization logic inline, so catalogs can override it with custom case sensitivity rules.

diqiu50:
### What type of PR is it?
Improvement

### What does this PR do?
Adds a `default String normalizeName(Scope scope, String name)` method to `Capability`, and updates `CapabilityHelpers` to call it instead of hardcoding a lowercase fold. This lets catalogs override normalization for case sensitivity behavior that isn't a simple lowercase fold (e.g. uppercase-by-default catalogs, or ones supporting both quoted case-sensitive and unquoted case-insensitive forms).

### Related issues
Fix: #12009

No

`./gradlew :core:test --tests "org.apache.gravitino.connector.capability.TestCapability"`


core/src/main/java/org/apache/gravitino/connector/capability/Capability.java
```diff
@@ -73,6 +74,22 @@ default CapabilityResult caseSensitiveOnName(Scope scope) {
...
+   * a catalog needs custom folding logic (for example, a catalog whose native folding direction is
+   * uppercase, or one that supports a case-sensitive quoted form alongside a case-insensitive
+   * unquoted form).
+   *
+   * @param scope The scope of the capability.
+   * @param name The name to normalize.
+   * @return The normalized, canonical name.
+   */
+  default String normalizeName(Scope scope, String name) {
+    return caseSensitiveOnName(scope).supported() ? name : name.toLowerCase(Locale.ROOT);
```
mchades, line 93: I recommend following the existing code pattern and utilizing `DEFAULT.normalizeName(scope, name)`.

diqiu50: fixed

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueCatalogCapability.java
```diff
@@ -76,4 +76,12 @@ public CapabilityResult caseSensitiveOnName(Scope scope) {
...
     }
   }
+
+  @Override
+  public String normalizeName(Scope scope, String name) {
+    if (name == null) {
+      return null;
+    }
+    return caseSensitiveOnName(scope).supported() ? name : name.toLowerCase(Locale.ROOT);
+  }
```
mchades: why not use the default behavior? It's the same

diqiu50: Tried this, but it breaks correctness: the default normalizeName logic relies on caseSensitiveOnName(scope) resolving polymorphically on this. Delegating to DEFAULT binds that call to DefaultCapability instead, so HiveCatalogCapability/GlueCatalogCapability (which override caseSensitiveOnName but not normalizeName) would silently lose their case-folding while still compiling fine. Kept the interface default self-contained (dispatch on this + null short-circuit) so any current/future implementation stays correct without needing a duplicate override.

mchades (approved):

PR #12013, opened 2026-07-14
[#11946] improvement(api): Add properties field to TableChange.AddIndex for custom index parameters

### LINKED ISSUE #11946: [Improvement] Add properties field to TableChange.AddIndex for custom index parameters

`TableChange.AddIndex` (`api/src/main/java/org/apache/gravitino/rel/TableChange.java`) only carries three fields:

- `Index.IndexType type`
- `String name`
- `String[][] fieldNames`

There is no `properties` field. This means the `ALTER TABLE ADD INDEX` path cannot pass any index-specific parameters to catalog implementations.

The `Index` interface already supports `properties()` (`Index.java`), and `Indexes.of(type, name, fieldNames, properties)` works correctly in the CREATE TABLE path. But when a catalog processes `TableChange.AddIndex` in `alterTable()`, properties are unavailable — it must fall back to hardcoded defaults.

**Concrete example:** PR #11806 added customizable GRANULARITY for ClickHouse data-skipping indexes via `Index.properties()`. CREATE TABLE respects user-specified GRANULARITY, but ALTER TABLE ADD INDEX always generates `GRANULARITY 1` because `AddIndex` has no way to carry the value.

This gap was identified during review of #11806 (https://github.com/apache/gravitino/pull/11806#discussion_r3544289778).

### How should we improve?

Add a `Map<String, String> properties` field to `TableChange.AddIndex`:

1. **AddIndex class** (`TableChange.java`):
   - Add `private final Map<String, String> properties` field
   - Add constructor overload accepting properties
   - Update `equals()` / `hashCode()` to include properties

2. **Factory method** (`TableChange.java`):
   - Add overload: `addIndex(IndexType type, String name, String[][] fieldNames, Map<String, String> properties)`
   - Existing `addIndex(IndexType, String, String[][])` delegates with `Collections.emptyMap()`

3. **REST API serialization** (TableChangeDTO or equivalent):
   - Add optional `properties` field; absent/null treated as empty map for backward compatibility

4. **Catalog implementations**:
   - Update ClickHouse `addIndexDefinition()` to read `addIndex.properties()` when generating ALTER TABLE ADD INDEX (falling back to default when absent)
   - Other catalogs remain unaffected (empty map = use catalog default)

### Related

-  #11806
- `Index.properties()`: `api/src/main/java/org/apache/gravitino/rel/indexes/Index.java`
- `Indexes.of()` with properties: `api/src/main/java/org/apache/gravitino/rel/indexes/Indexes.java`

jiangxt2:

Add a `Map<String, String> properties` field to `TableChange.AddIndex`, enabling catalog implementations to accept custom index parameters (e.g., GRANULARITY for ClickHouse data-skipping indexes) through the ALTER TABLE ADD INDEX path.

The change touches four layers:

- **API** (`TableChange.java`): New `addIndex(type, name, fieldNames, properties)` factory method; existing no-properties method delegates with `Collections.emptyMap()`. Defensive copy + null safety in constructor. `equals`/`hashCode` updated.
- **REST DTO** (`TableUpdateRequest.java`): Pass `index.properties()` through deserialization.
- **Core** (`ManagedTableOperations.java`): Pass `addIndex.getProperties()` when building the new `Index` object.
- **ClickHouse catalog**: Both CREATE TABLE and ALTER TABLE paths resolve GRANULARITY and `set(N)` max values from `index.properties()`, falling back to sensible defaults when absent. Constants defined for property keys (`granularity`, `set_max_values`).

`TableChange.AddIndex` currently carries only `type`, `name`, and `fieldNames` — no `properties`. This means the ALTER TABLE ADD INDEX path cannot pass index-specific parameters to catalog implementations, forcing them to use hardcoded defaults.

Concrete example: the ClickHouse data-skipping index support (see #11806) allows custom GRANULARITY via `Index.properties()`. CREATE TABLE respects user-specified GRANULARITY, but ALTER TABLE ADD INDEX was stuck at `GRANULARITY 1` because `AddIndex` had no way to carry the value. This PR closes that gap for both CREATE TABLE and ALTER TABLE paths.

Fix: #11946

Yes.

- New public API: `TableChange.addIndex(IndexType, String, String[][], Map<String, String>)`. The existing 3-parameter overload is unchanged and fully backward compatible.
- **Behavior change (ClickHouse catalog)**: bloom_filter default GRANULARITY is unified to `1` (previously `3` in CREATE TABLE, `1` in ALTER TABLE). This aligns with the ClickHouse server default.

- **Unit tests**: `TestTableChange` — 6 new tests covering properties, null safety, equals/hashCode. `TestClickHouseTableOperations` — SQL generation for custom properties in both CREATE TABLE and ALTER TABLE paths, plus negative tests for invalid input.
- **Docker integration test**: `CatalogClickHouseIT.testAlterTableAddIndexWithCustomGranularity` — end-to-end ALTER TABLE ADD INDEX with GRANULARITY=5, verified via `SHOW CREATE TABLE`.

Run commands:
- `./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test -PskipITs` (unit tests)
- `./gradlew :catalogs-contrib:catalog-jdbc-clickhouse:test --tests "CatalogClickHouseIT" -PskipDockerTests=false` (Docker integration test)

### DIFF
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
@@ -21,6 +21,8 @@ package org.apache.gravitino.catalog.clickhouse.operations;
 import static org.apache.gravitino.catalog.clickhouse.ClickHouseConstants.IndexConstants.DATA_SKIPPING_BLOOM_FILTER;
 import static org.apache.gravitino.catalog.clickhouse.ClickHouseConstants.IndexConstants.DATA_SKIPPING_MINMAX_VALUE;
 import static org.apache.gravitino.catalog.clickhouse.ClickHouseConstants.IndexConstants.DATA_SKIPPING_SET;
+import static org.apache.gravitino.catalog.clickhouse.ClickHouseConstants.IndexConstants.GRANULARITY;
+import static org.apache.gravitino.catalog.clickhouse.ClickHouseConstants.IndexConstants.SET_MAX_VALUES;
 import static org.apache.gravitino.catalog.clickhouse.ClickHouseTablePropertiesMetadata.CLICKHOUSE_ENGINE_KEY;
 import static org.apache.gravitino.catalog.clickhouse.ClickHouseTablePropertiesMetadata.ENGINE_PROPERTY_ENTRY;
 import static org.apache.gravitino.catalog.clickhouse.ClickHouseTablePropertiesMetadata.GRAVITINO_ENGINE_KEY;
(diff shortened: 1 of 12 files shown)


catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -897,6 +919,56 @@ private String addIndexDefinition(JdbcTable table, TableChange.AddIndex addIndex
...
+  }
+
+  /**
+   * Resolves the set(N) max unique values from index properties.
+   *
+   * @param properties the index properties map
+   * @return the resolved max unique values (0 = unlimited); must be non-negative
+   * @throws IllegalArgumentException if the value is present but not a non-negative integer
+   */
+  private int resolveSetMaxValues(Map<String, String> properties) {
```
yuqi1129, line 957: These two methods are mostly the same; you'd better extract a common one.

yuqi1129:
Can you also update the document of clickhouse catalog about the change?

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Done.

jiangxt2:
Updated.

yuqi1129 (approved):

PR #12015, opened 2026-07-14
[#12014] fix(core): Don't wrap plain ClassNotFoundException in IsolatedClassLoader

### LINKED ISSUE #12014: [Bug report] IsolatedClassLoader fails to resolve java.lang simple names for generated code
### Version

main branch

### Describe what's wrong

When generated Java code is compiled or resolved inside Gravitino's isolated catalog classloader, simple names from `java.lang`, such as `Override`, may be requested directly from the classloader.

`IsolatedClassLoader` currently treats names without a package as normal class names and fails to resolve them as `java.lang` classes. This can break codegen-based libraries that rely on the context classloader, such as Janino usage triggered by generated-code paths.

The generated-code path may fail with `ClassNotFoundException` for simple `java.lang` names, for example:

```text
ClassNotFoundException: Class not found Override
```

### How to reproduce

Use a catalog/library path that performs Java code generation under Gravitino's isolated classloader and asks the context classloader to resolve simple `java.lang` annotation names.

A minimal reproduction is loading `Override` through `IsolatedClassLoader`:

```java
classLoader.withClassLoader(cl -> cl.loadClass("Override"));
```

Before the fix this fails; after the fix it resolves to `java.lang.Override`.

### Additional context

This is a compatibility issue in `IsolatedClassLoader` class resolution. The fix should not declare support for a new Paimon version; it only makes the isolated classloader correctly handle simple `java.lang` names used by generated-code / Janino paths.

jarredhj0214:

This PR updates `IsolatedClassLoader.CustomURLClassLoader#loadClass` to rethrow a
`ClassNotFoundException` from `doLoadClass` as-is, instead of wrapping it in a new
`ClassNotFoundException` with a cause. Other exceptions are still wrapped as before.

It also adds a unit test verifying that a class-miss surfaces as a plain
`ClassNotFoundException` with a `null` cause.

Probe-driven callers such as Janino (used by codegen paths, e.g. Paimon) try a series of
candidate class names (`Override`, then `java.lang.Override`, ...) and rely on the standard
classloader contract: a `ClassNotFoundException` with a `null` cause means "class does not
exist, try the next candidate", while a non-null cause means a fatal loading failure.

`IsolatedClassLoader` previously wrapped every miss in a `ClassNotFoundException` with a
cause, which turned an ordinary probe miss into a fatal failure and broke such codegen
paths under the isolated catalog classloader.

Note: the classloader does not (and should not) resolve simple names like `Override` to
`java.lang.Override` itself; with this fix, probing callers can handle that themselves.

Fix: #12014

No.

Added `testClassNotFoundExceptionDoesNotWrapProbeMiss` in `TestIsolatedClassLoader`,
which fails against the old wrapping behavior and passes with this change.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
+++ b/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
@@ -172,8 +172,13 @@ public class IsolatedClassLoader implements Closeable {
 
       try {
         return clazz == null ? doLoadClass(name, resolve) : clazz;
+      } catch (ClassNotFoundException e) {
+        // Probe-driven callers, such as Janino, use a plain ClassNotFoundException as a
+        // signal that the current candidate name missed and they should try the next one.
+        // Wrapping it with a cause changes that signal into a fatal loading failure.
+        throw e;
       } catch (Exception e) {
-        throw new ClassNotFoundException("Class not found " + name, e);
+        throw new ClassNotFoundException("Failed to load " + name, e);
       }
     }
 
--- a/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
+++ b/core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
@@ -172,8 +172,10 @@ public class IsolatedClassLoader implements Closeable {
 
       try {
         return clazz == null ? doLoadClass(name, resolve) : clazz;
+      } catch (ClassNotFoundException e) {
+        throw e;
       } catch (Exception e) {
-        throw new ClassNotFoundException("Class not found " + name, e);
+        throw new ClassNotFoundException("Failed to load " + name, e);
       }
     }
 
--- a/core/src/test/java/org/apache/gravitino/utils/TestIsolatedClassLoader.java
+++ b/core/src/test/java/org/apache/gravitino/utils/TestIsolatedClassLoader.java
@@ -45,6 +45,18 @@ public class TestIsolatedClassLoader {
   }
 
   @Test
+  public void testClassNotFoundExceptionDoesNotWrapProbeMiss() {
+    // Probe-driven callers such as Janino expect a plain ClassNotFoundException for misses
+    // so they can continue trying the next candidate name.
+    ClassNotFoundException exception =
+        Assertions.assertThrows(
+            ClassNotFoundException.class,
+            () -> classLoader.withClassLoader(cl -> cl.loadClass("UnknownProbeClass")));
+
+    Assertions.assertNull(exception.getCause());
+  }
+
+  @Test
   public void testHivePackageRecognizedAsCatalogClass() throws Exception {
     // org.apache.gravitino.hive.* was moved from catalog.hive.* by HiveClient refactoring.
     // These must be treated as catalog classes so they are loaded by the IsolatedClassLoader,


core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
```diff
@@ -178,6 +178,14 @@ protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundE
     }
 
     private Class<?> doLoadClass(String name, boolean resolve) throws Exception {
+      if (!name.contains(".")) {
+        try {
+          return getRootClassLoader().loadClass("java.lang." + name);
```
yuqi1129: I recommend that you come up with a thorough resolution to fix this problem. The fix only works for classes that are in the package `java.lang`. What if the class is in the package `java.utils` or another?

yuqi1129:
The `java.lang` mapping fixes the reported symptom, but the root cause is broader: Janino probes name candidates one by one and relies on `ClassNotFoundException` **without a cause** to mean "class doesn't exist, try the next candidate" (see `ClassLoaderIClassLoader#findIClass`). Our `loadClass()` wraps every failure into a CNFE **with a cause**, so any benign probe miss is treated as a fatal load error.

That's why this fix only covers names that happen to exist in `java.lang`: e.g. generated code with `import java.util.*;` using a bare `List` still aborts — the bare-name probe fails with a caused CNFE before Janino ever tries `java.util.List`.

Suggested change in `CustomURLClassLoader.loadClass()` — keep pure not-found cause-less, wrap only real loading failures:

```java
try {
  return clazz == null ? doLoadClass(name, resolve) : clazz;
} catch (ClassNotFoundException e) {
  throw e; // benign miss: keep it cause-less for probe-driven callers (Janino)
} catch (Exception e) {
  throw new ClassNotFoundException("Failed to load " + name, e);
}
```

With this, the `java.lang` special case becomes unnecessary (Janino falls back to `java.lang.Override` by itself via the shared `java.` path). If you keep it as a compatibility shim, note it currently shadows any legitimate default-package class whose name collides with a `java.lang` one (`Compiler`, `Process`, `Record`, ...), since it runs before the normal lookup. A regression test with `import java.util.*;` + bare simple name would lock in the general fix.

jarredhj0214:
> The `java.lang` mapping fixes the reported symptom, but the root cause is broader: Janino probes name candidates one by one and relies on `ClassNotFoundException` **without a cause** to mean "class doesn't exist, try the next candidate" (see `ClassLoaderIClassLoader#findIClass`). Our `loadClass()` wraps every failure into a CNFE **with a cause**, so any benign probe miss is treated as a fatal load error.
> 
> That's why this fix only covers names that happen to exist in `java.lang`: e.g. generated code with `import java.util.*;` using a bare `List` still aborts — the bare-name probe fails with a caused CNFE before Janino ever tries `java.util.List`.
> 
> Suggested change in `CustomURLClassLoader.loadClass()` — keep pure not-found cause-less, wrap only real loading failures:
> 
> ```java
> try {
>   return clazz == null ? doLoadClass(name, resolve) : clazz;
> } catch (ClassNotFoundException e) {
>   throw e; // benign miss: keep it cause-less for probe-driven callers (Janino)
> } catch (Exception e) {
>   throw new ClassNotFoundException("Failed to load " + name, e);
> }
> ```
> 
> With this, the `java.lang` special case becomes unnecessary (Janino falls back to `java.lang.Override` by itself via the shared `java.` path). If you keep it as a compatibility shim, note it currently shadows any legitimate default-package class whose name collides with a `java.lang` one (`Compiler`, `Process`, `Record`, ...), since it runs before the normal lookup. A regression test with `import java.util.*;` + bare simple name would lock in the general fix.

Thanks for the detailed explanation. I updated the PR to follow this direction.

Instead of resolving `java.lang` simple names in `IsolatedClassLoader`, the latest change preserves the original `ClassNotFoundException` for normal class probe misses and only wraps non-CNFE loading failures. This keeps Janino's probing behavior intact, so it can continue trying candidate names such as `java.lang.Override` or imported classes like `java.util.List`.

I also removed the `java.lang` simple-name shortcut to avoid shadowing legitimate default-package classes, and added a regression test to verify that probe misses are not wrapped with a cause.

yuqi1129:
@jarredhj0214 

Have you verified whether the following code works in your environment?

```java
try {
  return clazz == null ? doLoadClass(name, resolve) : clazz;
} catch (ClassNotFoundException e) {
  throw e; // benign miss: keep it cause-less for probe-driven callers (Janino)
} catch (Exception e) {
  throw new ClassNotFoundException("Failed to load " + name, e);
}
```

jarredhj0214:
Yes, I verified it with a Paimon 1.4.1 codegen path close to the original failure.

The verification loads `paimon-bundle-1.4.1.jar` through two isolated `URLClassLoader` variants and triggers the same codegen path:

- `UserDefinedSeqComparator.create(...)`
- `CodeGenUtils.newRecordComparator(...)`
- `GeneratedClass.compile(CodeGenUtils.class.getClassLoader())`
- Paimon's shaded Janino compile

With the old behavior, where class loading misses are wrapped into a caused `ClassNotFoundException`, the compile fails with `ClassNotFoundException: Override`.

With the updated behavior, where plain `ClassNotFoundException` is preserved for probe misses, the same Paimon 1.4.1 codegen path succeeds and returns a `UserDefinedSeqComparator`.

So this change fixes the Janino probing behavior needed by the Paimon 1.4.1 codegen path.

core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
```diff
@@ -172,8 +172,10 @@ protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundE
 
       try {
         return clazz == null ? doLoadClass(name, resolve) : clazz;
+      } catch (ClassNotFoundException e) {
+        throw e;
```
yuqi1129, line 179: Can you add some comments here about why we need to treat `ClassNotFoundException` differently?  Then users can get more details when they review the code.

yuqi1129:
> Yes, I verified it with a Paimon 1.4.1 codegen path close to the original failure.
> 
> The verification loads `paimon-bundle-1.4.1.jar` through two isolated `URLClassLoader` variants and triggers the same codegen path:
> 
> * `UserDefinedSeqComparator.create(...)`
> * `CodeGenUtils.newRecordComparator(...)`
> * `GeneratedClass.compile(CodeGenUtils.class.getClassLoader())`
> * Paimon's shaded Janino compile
> 
> With the old behavior, where class loading misses are wrapped into a caused `ClassNotFoundException`, the compile fails with `ClassNotFoundException: Override`.
> 
> With the updated behavior, where plain `ClassNotFoundException` is preserved for probe misses, the same Paimon 1.4.1 codegen path succeeds and returns a `UserDefinedSeqComparator`.
> 
> So this change fixes the Janino probing behavior needed by the Paimon 1.4.1 codegen path.

Okay.  LGTM except a minor one.

core/src/main/java/org/apache/gravitino/utils/IsolatedClassLoader.java
jarredhj0214: Added comments to explain why `ClassNotFoundException` needs to be treated differently here.

yuqi1129 (approved):

PR #12016, opened 2026-07-14
[#11972] fix(clickhouse): strip PROJECTION blocks before extracting sort key from SHOW CREATE TABLE

### LINKED ISSUE #11972: [BUG] ClickHouse catalog: PROJECTION ORDER BY corrupts sort key extraction
### Describe the bug

When a ClickHouse table contains a Normal-type `PROJECTION` that includes an internal `ORDER BY` clause, loading the table through the ClickHouse catalog produces incorrect sort order metadata. The sort-order extraction from the table DDL mistakenly matches the PROJECTION's internal `ORDER BY` instead of the table's actual sort key. The returned sort orders contain garbled PROJECTION query fragments rather than the table's sort columns.

Aggregate-type `PROJECTION`s (defined with `GROUP BY`, no `ORDER BY`) are not affected.

### Motivation

This is a correctness bug — downstream tools that rely on Gravitino's sort order metadata for query planning or schema validation will make decisions based on wrong information. The issue is silent: no warning or error is produced. The incorrect sort orders look plausible at a glance (they contain valid column references), making the problem hard to detect.

### Describe the solution you'd like

Ensure the sort-key extraction logic does not match into `PROJECTION` bodies, so that a PROJECTION's internal `ORDER BY` is not mistaken for the table-level sort key.

### Test plan

| # | Scenario | Expected after fix |
|---|----------|-------------------|
| 1 | Normal PROJECTION with `ORDER BY` — loadTable | sortOrders = table's `ORDER BY` columns, not PROJECTION fragments |
| 2 | Aggregate PROJECTION with `GROUP BY` (no `ORDER BY`) — loadTable | sortOrders = table's `ORDER BY` columns (already correct, must not regress) |
| 3 | Multiple PROJECTIONs on same table | sortOrders = table's `ORDER BY` columns |
| 4 | Table without PROJECTION — loadTable | sortOrders unchanged (no regression) |
| 5 | Table with PROJECTION + existing INDEX | sortOrders + indexes both correct, no interference |
| 6 | Normal PROJECTION at all positions (inline, ALTER-added) | consistent behavior regardless of DDL position |

### Additional context

**Test results** (ClickHouse + Gravitino 1.2.0):

| # | Scenario | Result before fix |
|---|----------|:-----------------:|
| 1 | Normal PROJECTION w/ `ORDER BY` → loadTable | ❌ sortOrders = `dt\n)...ORDER BY (id, dt)` (garbled) |
| 2 | Aggregate PROJECTION (no `ORDER BY`) → loadTable | ✅ sortOrders = `(id, dt)` (correct) |
| 3 | Multiple Normal PROJECTIONs → loadTable | ❌ all PROJECTION bodies consumed into sortOrders |
| 4 | No PROJECTION → loadTable | ✅ normal (no regression risk confirmed) |
| 5 | PROJECTION added via ALTER TABLE | ❌ same corruption as inline |
| 6 | PROJECTION with SECONDARY INDEX on same table | ❌ sortOrders corrupted, INDEX unaffected |

A follow-up feature request will track full PROJECTION definition support for round-trip preservation.

I'd like to pick this up. Please let me know if you have any feedback.

jiangxt2:

When a ClickHouse table contains a PROJECTION with its own ORDER BY clause, Gravitino parses the wrong ORDER BY from the SHOW CREATE TABLE output and returns incorrect sort keys.

This PR strips PROJECTION definition blocks from the DDL before applying the existing regex patterns, so the table-level ORDER BY / PARTITION BY / SETTINGS clauses are matched correctly.

A PROJECTION block sits inside the column-definition body of the DDL:

```sql
CREATE TABLE t (
    `id` Int64,
    PROJECTION p_normal ( SELECT * ORDER BY dt ),
    `dt` Date
)
ENGINE = MergeTree
ORDER BY (id, dt)
```

The projection-internal `ORDER BY dt` appears before the table-level `ORDER BY (id, dt)`, so `find()` matches the wrong one. This affects any table with a Normal projection (projections with no ORDER BY, like Aggregate projections with only GROUP BY, are not affected).

Fixes #11972

No. Tables without projections are unaffected. For tables with projections, `sortOrder()` now returns the correct table-level sort keys instead of garbled projection-internal content.

- Unit tests: 19 new tests covering Normal/Aggregate/Multiple projections, INDEX coexistence, projection at different positions, string literals containing "PROJECTION", and escaped string literals inside projection bodies.
- Docker integration tests: 11 tests on real ClickHouse 24.8 covering inline CREATE, ALTER TABLE ADD PROJECTION, PARTITION BY + PRIMARY KEY, ReplacingMergeTree, complex sort keys, and projections at all positions.
- Existing tests (SETTINGS, INDEX, type mapping) continue to pass.


catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -1376,23 +1376,174 @@ Transform[] parsePartitioning(String partitionKey) {
...
+            if (i < len && createSql.charAt(i) == '(') {
+              int depth = 1;
+              i++;
+              while (i < len && depth > 0) {
+                char bodyCh = createSql.charAt(i);
+                if (bodyCh == '\'') {
+                  // skip string literal inside projection body
+                  i++;
+                  while (i < len) {
+                    if (createSql.charAt(i) == '\'') {
```
yuqi1129: ClickHouse `SHOW CREATE TABLE` uses backslash escapes for string literals. For example, input `it''s` is returned as `it\'s`. This code treats the quote after `\` as the end of the string. It may then skip the rest of the DDL, including the table-level `ORDER BY`.

The current test uses `''`, but that is not the actual `SHOW CREATE TABLE` output. Could we handle backslash escapes in both string loops and add a test with real ClickHouse output?

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -1376,23 +1376,174 @@ Transform[] parsePartitioning(String partitionKey) {
...
+                    }
+                    i++;
+                  }
+                } else {
+                  if (bodyCh == '(') depth++;
+                  else if (bodyCh == ')') depth--;
+                  i++;
+                }
+              }
+              // Skip trailing whitespace and optional comma
```
yuqi1129: Newer ClickHouse versions allow `PROJECTION p (...) WITH SETTINGS (...)`. This code stops after the projection body and leaves `WITH SETTINGS (...)` in `cleanedSql`. Then `SETTINGS_PATTERN.find()` may read projection settings as table settings.

Could we remove the full projection declaration, including `WITH SETTINGS`, or avoid parsing projection syntax here? See the current ClickHouse parser: https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/ParserCreateQuery.cpp#L267-L334

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -1376,23 +1376,174 @@ Transform[] parsePartitioning(String partitionKey) {
...
+   *
+   * <p>A projection block has the form {@code PROJECTION name ( SELECT ... )} and sits inside the
+   * column-definition body of the DDL. This method removes every such block including the optional
+   * trailing comma, while preserving string literals and respecting nested parentheses.
+   *
+   * @param createSql raw {@code SHOW CREATE TABLE} output
+   * @return the DDL with all PROJECTION blocks removed, or the original string if none are found
+   */
+  @VisibleForTesting
+  String stripProjections(String createSql) {
```
yuqi1129: Could we avoid parsing projection definitions here? ClickHouse already provides the table sort expression in `system.tables.sorting_key`. This class also reads `partition_key` from `system.tables`, and `getTableProperties()` already queries the same table. Reusing `sorting_key` would let us remove this long parser and avoid more ClickHouse syntax edge cases.

yuqi1129 (commented):
Thanks for the fix. I found two cases that still need attention, and I also left one suggestion to simplify the code.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Thanks for catching this. I removed the projection scanner in 78e2a1c4f. Sort orders now come from system.tables.sorting_key, so SHOW CREATE TABLE escaping no longer affects this path. The projection IT is retained, and all 46 CatalogClickHouseIT tests pass.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: I avoided parsing projection syntax: table Settings now come only from system.tables.engine_full, and SETTINGS_PATTERN is no longer applied to the full SHOW CREATE TABLE output. Since engine_full excludes projection declarations, projection-level WITH SETTINGS cannot contaminate table Settings.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Implemented in 78e2a1c4f using an exact prepared query for sorting_key and engine_full, filtered by both database and table name. A missing or invisible row now throws NoSuchTableException. I kept this separate from getTableProperties() because that existing method filters only by table name; broader metadata consolidation is tracked separately.

jiangxt2:
Hi @yuqi1129, could you please take another look when you have time? Thanks!

yuqi1129:
I have no more comments, let run the CI again.

yuqi1129 (dismissed):

yuqi1129:
@jiangxt2 
Please resolve the conflict if you have time.

jiangxt2:
> @jiangxt2 Please resolve the conflict if you have time.

I've resolved the code conflicts.

yuqi1129 (approved):

PR #12025, opened 2026-07-15
[#12024] improvement(common): harden JDBC unsafe-parameter detection against connectionProperties smuggling

### LINKED ISSUE #12024: [Improvement] Harden JdbcUrlUtils unsafe-parameter detection against connectionProperties smuggling

`JdbcUrlUtils.validateJdbcConfig` rejects dangerous JDBC connection parameters (MySQL `autoDeserialize`, `queryInterceptors`, `allowLoadLocalInfile`, …; PostgreSQL `socketFactory`, …) that can lead to RCE when a catalog is created/altered. The current check has two gaps on the catalog datasource path (`DataSourceUtils` → DBCP2 `BasicDataSourceFactory`):

1. **It inspects config *values*, not *names*.** `containsValueIgnoreCase` matches a param name against config values, but the map *keys* are never checked. An unsafe parameter supplied as a config key is not detected.
2. **It ignores DBCP2's `connectionProperties`.** DBCP2 recognizes a `connectionProperties` key whose value is a `;`-delimited list of `name=value` pairs that it forwards verbatim to the JDBC driver. A user can reach it via `gravitino.bypass.connectionProperties=autoDeserialize=true` (the `gravitino.bypass.` prefix is stripped before the map reaches the datasource), smuggling an unsafe parameter straight to the driver and fully bypassing the check.

The URL substring check `lowerUrl.contains(lowerParam)` is also locale-sensitive (`toLowerCase()` without `Locale.ROOT`).

### How should we improve?

- Inspect the parameter **names** the driver actually receives: every config **key**, plus each name embedded in the `connectionProperties` value.
- Parse `connectionProperties` **exactly as DBCP2 does** (`Properties.load` after replacing `;`→`\n`) so escape/newline/`\uXXXX` parser-differential smuggling can't slip through; reject a malformed value fail-closed.
- Use `Locale.ROOT` for all case folding.

LuciferYang:

- Replace the config-**value** scan in `JdbcUrlUtils` with a config-**name** scan: every config key, plus each parameter name embedded in DBCP2's `connectionProperties` value.
- Parse `connectionProperties` exactly as DBCP2 does (`Properties.load` after replacing `;`→`\n`); reject a malformed value fail-closed.
- Use `Locale.ROOT` for all case folding.

`JdbcUrlUtils.validateJdbcConfig` guards against dangerous JDBC parameters (e.g. MySQL `autoDeserialize`, an RCE enabler) but only matched config *values* and never inspected DBCP2's `connectionProperties`. An unsafe parameter could be smuggled to the driver as a config key, or inside `connectionProperties` (reachable via `gravitino.bypass.connectionProperties=autoDeserialize=true`, whose prefix is stripped before the map reaches the datasource), fully bypassing the check.

Fix: #12024

No. A configuration that was already unsafe is now rejected earlier with a clear error. The exception message wording changes from "...detected in JDBC URL" to "...detected in JDBC configuration" to reflect that config keys/`connectionProperties` are inspected too, not only the URL.

New unit tests in `TestJdbcUrlUtils` (connectionProperties for MySQL/MariaDB/PostgreSQL, config-key detection, newline / `\uXXXX` / double-URL-encoding smuggling, whole-name-vs-substring and value-vs-name negative controls, Turkish-locale folding, malformed-value rejection) and an end-to-end `TestDataSourceUrlValidation` case through `DataSourceUtils.createDataSource`. `./gradlew :common:test :catalogs:catalog-jdbc-common:test -PskipITs` and spotless pass.

### DIFF
--- a/common/src/main/java/org/apache/gravitino/utils/JdbcUrlUtils.java
+++ b/common/src/main/java/org/apache/gravitino/utils/JdbcUrlUtils.java
@@ -20,10 +20,17 @@
 package org.apache.gravitino.utils;
 
 import com.google.common.base.Preconditions;
+import java.io.IOException;
+import java.io.StringReader;
 import java.net.URLDecoder;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.gravitino.exceptions.GravitinoRuntimeException;
 
@@ -34,6 +41,11 @@ import org.apache.gravitino.exceptions.GravitinoRuntimeException;
  * safe to use.
  */
 public class JdbcUrlUtils {
+
+  // The DBCP2 connection-pool property whose value is a list of connection properties forwarded
+  // verbatim to the JDBC driver. Unsafe parameters smuggled here would otherwise evade a check
+  // that only inspects the URL string and the raw config values.
+  private static final String CONNECTION_PROPERTIES_KEY = "connectionProperties";
 
   // Unsafe parameters for MySQL and MariaDB, other parameters like
   // trustCertificateKeyStoreUrl, serverTimezone, characterEncoding,
@@ -73,7 +85,7 @@ public class JdbcUrlUtils {
     Preconditions.checkArgument(StringUtils.isNotBlank(driver), "Driver class name can't be blank");
     Preconditions.checkArgument(StringUtils.isNotBlank(url), "JDBC URL can't be blank");
 
-    String lowerUrl = url.toLowerCase();
+    String lowerUrl = url.toLowerCase(Locale.ROOT);
     String decodedUrl = recursiveDecode(lowerUrl);
 
     if (decodedUrl.startsWith("jdbc:mysql")) {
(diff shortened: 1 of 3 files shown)


LuciferYang:
cc @yuqi1129 @roryqi

common/src/main/java/org/apache/gravitino/utils/JdbcUrlUtils.java
```diff
@@ -88,13 +100,20 @@ public static void validateJdbcConfig(String driver, String url, Map<String, Str
...
 
-    String lowerUrl = url.toLowerCase();
+    // Percent-decoding in recursiveDecode can reintroduce upper-case characters (e.g. "%4a" ->
+    // 'J'), so lower-case again here rather than relying on the pre-decode lower-casing.
+    String lowerUrl = url.toLowerCase(Locale.ROOT);
+
+    // Parameter names that reach the JDBC driver through the config map: the config keys
+    // themselves (defense in depth) plus any names embedded in the DBCP2 "connectionProperties"
+    // value, which is forwarded verbatim to the driver.
+    Set<String> configParamNames = collectConfigParameterNames(config);
```
yuqi1129, line 110: [P1] This still leaves an unsafe effective URL path unvalidated. `JdbcCatalogOperations` strips arbitrary `gravitino.bypass.*` keys, and `DataSourceUtils` passes the resulting map to `BasicDataSourceFactory.createDataSource` before applying the canonical URL/driver/user/password. With DBCP properties such as `url`, `driverClassName`, and `initialSize > 0`, the factory initializes the pool before returning; after that, the later setters have no effect. I reproduced this with the pinned DBCP 2.11.0: after factory initialization and `setUrl(canonical)`, `getConnection().getMetaData().getURL()` still returned the bootstrap URL. Because this method only collects config/property names, the DBCP `url` value can still carry one of the denied parameters (and can bypass the H2 guard) without being checked. Please prevent factory initialization until the canonical connection fields are applied—for example, hold/remove `initialSize` and override sensitive DBCP properties before invoking the factory—and add a regression test for this ordering.

LuciferYang: Good catch, and thanks for the DBCP 2.11.0 repro — I reproduced the same: with a raw `url` + `initialSize > 0`, `BasicDataSourceFactory` initializes the pool on the bootstrap URL before the later `setUrl(canonical)` takes effect, so `getConnection().getMetaData().getURL()` returns the attacker URL.

Since the root cause is `DataSourceUtils`'s factory-init ordering (not the `JdbcUrlUtils` name detection this PR is about), I've addressed it in the companion PR #12036, which rejects unsafe DBCP pool properties before the config reaches the factory. Following your comment I expanded that blocklist to also cover the raw `url`, `username`, and `password` keys — the canonical values are applied from `jdbc-url`/`jdbc-user`/`jdbc-password` via explicit setters, so the raw DBCP keys are never legitimate — alongside `driverClassName`/`initialSize`. I added a regression test that a raw-`url` + `initialSize` config is rejected before factory init, plus a positive control asserting a legitimate config's live connection uses the canonical URL. Could we continue this thread on #12036?

LuciferYang:
friendly ping @yuqi1129 and @roryqi

yuqi1129 (approved):
LGTM

LuciferYang:
Thank you @roryqi @yuqi1129

PR #12030, opened 2026-07-15
[#12029] improvement(openapi): Add generated default OpenAPI bundle

### LINKED ISSUE #12029: [Improvement] Publish the bundled OpenAPI specification as a versioned artifact

Gravitino owns the OpenAPI source under `docs/open-api/`, but downstream consumers do not have a stable, versioned artifact containing the bundled specification.

The OpenAPI workflow proposed in #11986 creates `openapi.yaml` and `openapi.json` and uploads them as a workflow-run artifact. However, the documentation site needs a release-oriented artifact that it can retrieve deterministically. Re-bundling or copying the source files in `gravitino-site` could allow the rendered reference and published raw specification to drift.

This is needed by:

- https://github.com/apache/gravitino-site/issues/138
- https://github.com/apache/gravitino/issues/11985
- https://github.com/apache/gravitino/pull/11986

### How should we improve?

Provide one supported CLI/build entry point that bundles `docs/open-api/openapi.yaml` into self-contained YAML and JSON files. This could be a Gradle task such as:

    ./gradlew :docs:bundleOpenApi

or a documented equivalent backed by `dev/openapi`.

Publish the resulting bundle from the build/release workflow using an explicit, versioned artifact name and a location that another repository can retrieve reliably.

Acceptance criteria:

- A documented command produces self-contained `openapi.yaml` and `openapi.json`.
- The build/release workflow publishes those files as a clearly named, versioned artifact.
- The artifact can be retrieved by an exact Gravitino release or documentation version.
- The published files are generated from the same commit used for the release.
- CI validates the bundle before publication.
- `gravitino-site` can consume the artifact for its Docusaurus OpenAPI reference and raw-spec link without re-bundling it.

nevzheng:

- Add `:docs:bundleOpenApi` to bundle the default OpenAPI specification.
- Write generated bundles to `docs/open-api/default/openapi.json` and
  `docs/open-api/default/openapi.yaml`.
- Regenerate both bundles as part of `:docs:build`.
- Extend the existing build workflow to fail when the committed bundles are stale or missing.
- Document that OpenAPI source changes must include the regenerated JSON and YAML.
- Update the Stage 1 release flow to regenerate both bundles after setting the final OpenAPI
  version and before creating the release commit and tag.

This provides self-contained OpenAPI JSON and YAML documents that documentation and other
downstream consumers can retrieve from a pinned Gravitino revision. For this temporary solution,
the exact Git commit or release tag is the version identifier and the tracked files are the
published artifact contract.

CI enforces that committed bundles match the OpenAPI source during normal development. The release
script enforces the same invariant after replacing the snapshot version with the final release
version, so the release tag cannot contain stale bundles.

This PR is intentionally a small solution that is easier and faster to land than the broader
OpenAPI pipeline proposed in #11986.

Fixes #12029.

### Follow-up work

- `apache/gravitino-site#138` will pin a Gravitino commit or release tag, use the pinned
  `default/openapi.json` for Docusaurus rendering, and expose the same file as the raw download.
- Attaching the bundles directly to GitHub Releases is an optional enhancement; it is not required
  by this temporary Git revision/tag contract.
- If #11986 is merged, its tooling must preserve `docs/open-api/default/openapi.json` and
  `docs/open-api/default/openapi.yaml` so existing consumers do not break.

The repository now provides bundled OpenAPI JSON and YAML documents at known, revision-addressable
paths. There are no runtime API changes.

- `./gradlew :docs:bundleOpenApi --rerun-tasks`
- Verified repeated generation produces identical SHA-256 checksums for both formats.
- Verified both bundles contain no external file references.
- `./gradlew :docs:spotlessApply :docs:build`
- `./gradlew :docs:spotlessCheck`
- Parsed the updated GitHub Actions workflow as YAML.
- Ran the exact CI freshness command against the committed bundles.
- Verified the root CI Gradle command schedules `:docs:bundleOpenApi` before the freshness check.
- `bash -n dev/release/release-tag.sh`
- Simulated the release transition from `2.0.0-SNAPSHOT` to `2.0.0`, regenerated both bundles, and
  verified the source, JSON, and YAML all report `2.0.0`.
- `git diff --check`


nevzheng:
@roryqi Could you please review this PR? Thanks!

roryqi:
> @roryqi Could you please review this PR? Thanks!

cc @mchades

docs/open-api/default/openapi.json
mchades, line 1: It is recommended to mark the file with `linguist-generated=true` in the `.gitattributes` file to reduce review noise.

nevzheng:
@mchades I updated it to mark the generated bundles as generated. Thanks! PTAL! Thx

PR #12036, opened 2026-07-16
[#12035] improvement(catalog-jdbc): block DBCP2 class-loading properties in datasource config

### LINKED ISSUE #12035: [Improvement] Block DBCP2 class-loading properties in JDBC catalog datasource configuration

`DataSourceUtils.createDataSource` hands the entire catalog config map to DBCP2's `BasicDataSourceFactory`. Several DBCP2 pool properties cause the factory to load and instantiate an arbitrary class by name via reflection (`Class.forName` + `newInstance`), which is a remote-code-execution vector when a catalog is created/altered:

- `connectionFactoryClassName` and `evictionPolicyClassName` — values are class names instantiated at pool init;
- `driverClassName` — value is `Class.forName`'d and instantiated by DBCP2's `DriverFactory` (the legitimate driver comes from the `jdbc-driver` property via an explicit setter, so the raw key is never needed);
- `initialSize` > 0 — makes the factory eagerly open a connection during `createDataSource`, which triggers the reflective driver load *inside* the factory call, before Gravitino's own `setDriverClassName` override runs.

A user reaches these via `gravitino.bypass.<key>=...` (the `gravitino.bypass.` prefix is stripped before the map reaches the datasource). This is the same catalog-creator to server-code-execution threat class already guarded against for unsafe JDBC URL parameters.

### How should we improve?

Reject these DBCP2 pool properties from catalog configuration before the config is handed to `BasicDataSourceFactory` (case-insensitive whole-key match), with a clear error. They are never needed for a Gravitino JDBC catalog.

LuciferYang:

Add `rejectUnsafePoolProperties` in `DataSourceUtils`, rejecting the DBCP2 pool properties `connectionFactoryClassName`, `evictionPolicyClassName`, `driverClassName`, and `initialSize` (case-insensitive whole-key match) before the config map is handed to `BasicDataSourceFactory`.

DBCP2's factory loads and instantiates the class named by the first three via reflection (`Class.forName` + `newInstance`), and `initialSize` > 0 forces an eager connection — triggering that driver load — during `createDataSource`, before Gravitino's explicit `setDriverClassName` override. Reachable via `gravitino.bypass.<key>=...`, this lets a catalog creator run arbitrary classes on the server (RCE).

Fix: #12035

No. These properties are never used by a legitimate Gravitino JDBC catalog (the driver comes from `jdbc-driver`; pool sizing from `jdbc.pool.*`). Supplying one now fails fast with a clear message.

New unit tests in `TestDataSourceUrlValidation` covering each rejected property (including case-insensitive variants) and a whole-key negative control proving a similarly-named legitimate property still builds. `./gradlew :catalogs:catalog-jdbc-common:test -PskipITs` and spotless pass.

### DIFF
--- a/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/utils/DataSourceUtils.java
+++ b/catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/utils/DataSourceUtils.java
@@ -41,13 +41,18 @@ public class DataSourceUtils {
   private static final String POOL_TEST_QUERY = "SELECT 1";
 
   // DBCP2 connection-pool properties that must never come from catalog configuration. The whole
-  // config map is handed to BasicDataSourceFactory, so any of these would let a catalog creator run
-  // arbitrary classes on the server (remote code execution):
+  // config map is handed to BasicDataSourceFactory, so allowing these would either run arbitrary
+  // code or let a raw property override the validated canonical connection fields:
   //   - connectionFactoryClassName / evictionPolicyClassName / driverClassName: their values are
   //     class names the factory loads and instantiates via reflection (Class.forName + newInstance)
-  //     when the pool creates connections. The legitimate driver is set separately from the
-  //     "jdbc-driver" property via an explicit setter, so the raw "driverClassName" key is never
-  //     needed.
+  //     when the pool creates connections (remote code execution). The legitimate driver is set
+  //     separately from the "jdbc-driver" property via an explicit setter, so the raw
+  //     "driverClassName" key is never needed.
+  //   - url / username / password: the connection identity is validated (JdbcUrlUtils, the H2
+  //     guard) and applied from the canonical "jdbc-url"/"jdbc-user"/"jdbc-password" properties via
+  //     explicit setters. A raw DBCP key here would reach the factory unvalidated and, once the
+  //     pool initializes, take effect before those setters run — e.g. a raw "url" could smuggle an
+  //     H2 or otherwise-denied URL past the guards.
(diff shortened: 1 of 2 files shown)


catalogs/catalog-jdbc-common/src/main/java/org/apache/gravitino/catalog/jdbc/utils/DataSourceUtils.java
```diff
@@ -92,6 +115,30 @@ private static Properties getProperties(JdbcConfig jdbcConfig) {
...
+   * @throws GravitinoRuntimeException if an unsafe connection-pool property is present
+   */
+  private static void rejectUnsafePoolProperties(Map<String, String> config) {
+    if (config == null) {
+      return;
+    }
+    for (String key : config.keySet()) {
+      for (String unsafe : UNSAFE_POOL_PROPERTIES) {
+        if (unsafe.equalsIgnoreCase(key)) {
+          throw new GravitinoRuntimeException(
```
yuqi1129, line 143: A minor suggestion: Since you use `equalsIgnoreCase` here, and `DRIVERCLASSNAME` and `driverClassName` will be treated as equal, I suggest you use the user value in the error message.

LuciferYang: Done in b92808dd0, the message now reports the user-supplied key (e.g. `DRIVERCLASSNAME`) rather than the canonical driverClassName, and the case-insensitive tests assert the as-typed key. Thanks!

yuqi1129 (approved):

LuciferYang:
Thank you @yuqi1129

PR #12042, opened 2026-07-16
[#12018] refactor(core): cache only self-contained entity types via an allowlist

### LINKED ISSUE #12018: [Subtask] Restrict the entity cache to the entity types that are safe to cache per node
### Describe the subtask

Once relation data is removed from the cache (#11961), the cache holds only single-entity entries. This subtask restricts *which* entity types the cache keeps, so that every cached entity is safe to serve from a per-node copy.

Cache the self-contained metadata objects — metalake, catalog, schema, table, topic, view, fileset — plus tag, policy, and the job entity. A stale read of any of these is at worst cosmetic (an old comment, property, or job status), never a wrong pointer. Metalake is cached like the rest: disabling or deleting a metalake is a rare, tenant-level admin action, so the brief window until other nodes refresh (one poll interval) is acceptable, and no special handling is needed for the in-use flag.

Do not cache model, model version, or function: each holds a load-bearing pointer (a model version URI, the latest version, or the function implementation) that would be silently wrong on another node during the staleness window, so they read straight from the `EntityStore`. (A shared cache such as redis has no staleness window and can cache them; that is handled in the redis subtask.) User, group, and role also stay out, because their derived fields (`roleNames`, `securableObjects`) would need a reverse lookup.

This subtask should:

- Add an entity-type allowlist to the cache write path: cache only metalake, catalog, schema, table, topic, view, fileset, tag, policy, and job.
- Skip caching for model, model version, function, user, group, and role; route their reads to the `EntityStore`.
- Add unit tests for the allowlist (cached vs. not-cached types).

The cross-node invalidation for these types, and the change-log emit point that tag, policy, and job still need, are handled in the caffeine multi-node subtask (#12019). This keeps every cached entity safe to serve per node, which is a prerequisite for that work under #11737.

### Parent issue

yuqi1129:

Restrict `CaffeineEntityCache` to an explicit allowlist of self-contained entity types, replacing the previous exclude-list of `{USER, GROUP, ROLE}`.

- `CACHEABLE_TYPES = {METALAKE, CATALOG, SCHEMA, TABLE, TOPIC, VIEW, FILESET, TAG, POLICY, JOB}`; `put()` skips any type not in the allowlist, so reads for other types fall through to the `EntityStore`.
- Model, model version, and function are no longer cached: each carries a load-bearing pointer (a version URI, the latest version, the function implementation) that would be silently wrong if served stale on another node.
- Removed the now-dead `invalidateOnKeyChange` model-version → model branch (every cacheable entity is self-contained, so inserting one never invalidates a different key), dropped the unused imports, and updated the class javadoc.
- Tests: replaced `testPutModelVersionInvalidatesModel` with `testModelAndModelVersionAreNotCached`; added `testTagIsCached`.

Stacked on #12006; please merge after it.

With relation caching removed (#11961) the cache holds only single-entity entries, but not every type is safe to serve from a per-node copy. Types whose value embeds a load-bearing pointer (model / model version URI and latest version, function implementation) would return a silently wrong answer on a node with a stale copy; relation-derived types (user/group/role) cannot be invalidated without a reverse index. An explicit allowlist keeps only the self-contained types that are safe per node, and makes any newly added type default to not-cached. Prerequisite for the cross-node invalidation work under #11737.

Fix: #12018

No. Internal caching behavior only; uncached types are read straight from the `EntityStore`, so results are unchanged.

- `./gradlew :core:test --tests "org.apache.gravitino.cache.*"` — all pass, including the new `testModelAndModelVersionAreNotCached` and `testTagIsCached`.
- `TestRelationalEntityStore` (4/4) and the H2 matrix of `TestEntityStorageRelationCache` pass.


yuqi1129:
Consolidated into #12111 after merging the latest main branch. The cache type allowlist changes are now part of the combined PR.

PR #12043, opened 2026-07-16
[MINOR] improvement(cache): Add logs for Caffeine cache removals

yuqi1129:

Add DEBUG-level removal logs to all production Caffeine removal listeners. The logs include cache-specific keys or identifiers and the removal cause, while avoiding cached values that may be sensitive or large.

Cache expiration, eviction, replacement, and explicit invalidation were not consistently observable across modules. Consistent DEBUG-level logs make cache lifecycle and resource cleanup easier to diagnose.

Fix: N/A

Yes. Cache removal logs now consistently use DEBUG level and include cache keys or identifiers and removal causes. There are no API or configuration changes.

- Ran `./gradlew spotlessApply`.
- Ran targeted unit tests for the affected core, server-common, catalog-fileset, hive-metastore-common, filesystem-hadoop3, iceberg-common, and iceberg-rest-server modules.
- After standardizing the log level, reran targeted tests for core, server-common, catalog-fileset, and filesystem-hadoop3.
- Attempted `./gradlew test -PskipITs`; an unrelated `TestJDBCBackendBatchGet` PostgreSQL case failed because the container reported `No space left on device`.

### DIFF
--- a/catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
+++ b/catalogs/catalog-fileset/src/main/java/org/apache/gravitino/catalog/fileset/FilesetCatalogOperations.java
@@ -282,7 +282,7 @@ public class FilesetCatalogOperations extends ManagedSchemaOperations
               .removalListener(
                   (key, value, cause) -> {
                     FileSystemCacheKey cacheKey = (FileSystemCacheKey) key;
-                    LOG.info(
+                    LOG.debug(
                         "Removing FileSystem from cache: scheme={}, authority={}, user={}, cause={}",
                         cacheKey == null ? null : cacheKey.scheme,
                         cacheKey == null ? null : cacheKey.authority,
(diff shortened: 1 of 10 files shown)


diqiu50 (commented):
Would logging this at the INFO level generate too much log output?

yuqi1129:
> Would logging this at the INFO level generate too much log output?

It should be fine as long as they are not operating a performance test; in this way, the log message will be large enough. For the normal case, I think it's acceptable.

diqiu50:
These logs are primarily intended for debugging. Users can choose whether to enable them through their Log4j configuration. Logging this kind of information at the INFO level doesn't provide much value and would only add unnecessary noise to the logs.

yuqi1129:
> These logs are primarily intended for debugging. Users can choose whether to enable them through their Log4j configuration. Logging this kind of information at the INFO level doesn't provide much value and would only add unnecessary noise to the logs.

OK, I changed them to debug level. In fact, the first edition is debug level.

diqiu50 (approved):
LGTM

Issue #11965, opened 2026-07-09
[EPIC] Role assumption (SET ROLE): narrow effective permissions via X-Gravitino-Active-Roles


bharos:
### Describe the proposal

Implement Phase 1 of the approved design [design-docs/gravitino-role-assumption.md](https://github.com/apache/gravitino/blob/main/design-docs/gravitino-role-assumption.md) (discussion #10894): a request header `X-Gravitino-Active-Roles` that lets a caller narrow the active role set for a request on the native authorization path (Iceberg REST + native API). The server validates that the caller holds the declared role(s) and evaluates authorization against only those roles. Subtractive only; an absent header behaves exactly as today.

Agreed semantics:
- Narrowing removes **allows** only. **Denies are global** — a deny on any role the caller holds always applies, regardless of the active set (keeps narrowing strictly subtractive).
- Ownership always applies (Option A).
- Grammar: single role, comma-separated list, or the reserved keywords `ALL` / `NONE`.

### Task list

- [x] #11966 — Parse & validate the `X-Gravitino-Active-Roles` header
- [x] #11967 — Narrow enforcement in `JcasbinAuthorizer` (allow-only; deny global)
- [x] #12095 — Wire `X-Gravitino-Active-Roles` into the request path (parse + 400/403)
- [x] #11968 — Apply narrowing to list filtering + credential vending, with ITs & docs

bharos:
Phase 1 is complete: all four subtasks are merged and closed, so closing this epic.

PR #12044, opened 2026-07-16
[#11967] feat(auth): narrow allow enforcement to active roles (deny stays global)

### LINKED ISSUE #11967: [Subtask] Narrow enforcement in JcasbinAuthorizer (allow-only; deny global)
### Describe the subtask

When active roles are declared, evaluate allow policies against only those roles instead of the caller's full role union. Deny evaluation and deny-based list short-circuits continue to use the full role set, so a deny on any held role still applies regardless of the active set. Ownership checks are unchanged. Includes unit tests for allow-narrowing, deny-stays-global, `NONE`, and group-inherited roles.

### Parent issue

#11965

bharos:

Phase 1 role-assumption **enforcement** in `JcasbinAuthorizer`. When a request declares a subset of the caller's roles as active, allow policies are evaluated against only that subset; deny stays global.

- `AuthorizationRequestContext`: carry an immutable `ActiveRoles` (defaults to `ALL`, so an unset value is byte-for-byte today's behavior).
- `JcasbinAuthorizer`: only the allow enforcer narrows. When a subset is declared, `enforceNarrowed`:
  - applies deny **globally** first (over the caller's full role union) — a deny on any held role always applies, keeping narrowing subtractive;
  - then ORs allow across the active set = `(declared names) ∩ (caller's effective roles)`.
  - `NONE` activates no role; `ALL` / absent is unchanged.
- Ownership is deliberately not narrowed: access a caller derives from *owning* an object comes from the ownership grant, not a role, so `isOwner` is untouched and owners keep that access regardless of which roles are active. This is the design's default (Option A in `design-docs/gravitino-role-assumption.md` §5).

This is intentionally scoped to enforcement so it can be reviewed on its own. Carrying the value from the `X-Gravitino-Active-Roles` header onto the request and the `400`/`403` mapping follow in later PRs.

Implements the enforcement step of the approved design [`design-docs/gravitino-role-assumption.md`](https://github.com/apache/gravitino/blob/main/design-docs/gravitino-role-assumption.md) (discussion [#10894](https://github.com/apache/gravitino/discussions/10894)). A caller needs authorization to be narrowed to a declared subset of its roles; this PR is the server-side narrowing the rest of the feature builds on.

Part of #11965. Closes #11967.

No. The narrowing path only activates when an `ActiveRoles` subset is set on the request; nothing populates it yet, so behavior is unchanged.

New unit tests in `TestJcasbinAuthorizer`: allow-narrowing to a named role, `NONE`, deny-stays-global when the allow is narrowed, group-inherited active role, and unheld-role-activates-nothing (subtractive). Verified locally:

- `./gradlew :server-common:test --tests "org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer"`
- `./gradlew :core:test --tests "org.apache.gravitino.authorization.TestAuthorizationRequestContext"`
- `./gradlew :server-common:spotlessCheck :core:spotlessCheck`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
+++ b/core/src/main/java/org/apache/gravitino/authorization/AuthorizationRequestContext.java
@@ -30,6 +30,7 @@ import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.apache.gravitino.MetadataObject;
+import org.apache.gravitino.auth.ActiveRoles;
 import org.apache.gravitino.storage.relational.po.auth.GroupUpdatedAt;
 import org.apache.gravitino.storage.relational.po.auth.OwnerInfo;
 import org.apache.gravitino.storage.relational.po.auth.RoleUpdatedAt;
@@ -83,6 +84,12 @@ public class AuthorizationRequestContext {
   private volatile Map<Long, RoleUpdatedAt> prefetchedRoleVersions;
 
   private volatile String originalAuthorizationExpression;
+
+  /**
+   * The roles the caller has declared active for this request (role assumption). Defaults to {@link
+   * ActiveRoles#all()}, which evaluates every role the caller holds (no narrowing).
+   */
+  private volatile ActiveRoles activeRoles = ActiveRoles.all();
 
   /**
    * check allow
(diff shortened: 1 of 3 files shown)


roryqi (commented):
Could u resolve the conflicts? The pull request looks good to me.

roryqi (approved):
LGTM.

PR #12045, opened 2026-07-16
[MINOR] fix (trino-connector): Trino connector oauth2 user forwarding

markhoerth:
### Summary
Adds per-user authorization and credential vending support to the Gravitino Trino connector for Iceberg catalogs backed by the Iceberg REST Catalog (IRC).
Changes
Two commits:

* oauth2 per-user token forwarding — Extends the connector's session forwarding (forwardUser=true) to support authType=oauth2, not just simple. When enabled, the end user's forwarded IdP token is presented to the Gravitino server, so authorization runs against the real end user instead of a shared service identity. Adds StaticUserTokenProvider, widens the auth guards in GravitinoConnector and GravitinoAuthProvider, and keys the per-user client cache by auth type.
* per-user credential vending for IRC-backed Iceberg — Configures the generated Trino Iceberg REST catalog so that an IRC-backed Gravitino catalog requests per-user, per-table vended credentials. Sets security=OAUTH2, session=USER, and vended-credentials-enabled=true, and maps the warehouse through, so Trino's Iceberg REST client forwards the end-user token to the IRC and obtains scoped storage credentials over the REST protocol.

### Testing
Unit tests added for the oauth2 forwarding path (TestGravitinoAuthProvider) and the REST backend config mapping (TestIcebergCatalogPropertyConverter). End-to-end verification against a live IRC is in progress.

### Motivation
Trino is a primary query engine for enterprise customers who require real per-user authorization and scoped credentials rather than a shared service identity.


roryqi (dismissed):
LGTM.

roryqi (approved):

roryqi:
Could u modify the pull request title?

PR #12046, opened 2026-07-17
[MINOR] fix(web-v2): Normalize service admins configuration

gaohong:

This PR adds a shared `normalizeServiceAdmins` utility and uses it in the Metalake list page and user settings component.

The utility normalizes `serviceAdmins` into an array and supports both formats returned by the configuration API:

- An array of user names
- A comma-separated string of user names

Unsupported values are normalized to an empty array.

`serviceAdmins` can be returned as either an array or a comma-separated string. The previous duplicated page-level logic could call `.split(',')` on a non-string value, causing a frontend runtime error.

This change centralizes the normalization logic and prevents the UI from failing when the configuration value has an unexpected type.

Fix: N/A

Yes. The UI now handles both supported `serviceAdmins` formats consistently and no longer fails for unsupported values.

No user-facing API or configuration property is added or changed.

- Ran `git diff --check upstream/main...HEAD`.
- Verified that both existing `serviceAdmins` consumers use the shared utility:
  - `web-v2/web/src/app/rootLayout/UserSetting.js`
  - `web-v2/web/src/app/metalakes/page.js`

No automated frontend tests were run.


yuqi1129:
@LauraXia123 
Please help to review this PR, thanks.

LauraXia123 (approved):
LGTM

PR #12047, opened 2026-07-17
[#12038] improvement(scripts): increase relational store column comment length limit to 4096

### LINKED ISSUE #12038: [Bug report] Importing tables fails when column comment exceeds 256 characters
### Version

main branch

### Describe what's wrong

When importing an external table, Gravitino stores column metadata in `table_column_version_info`.

The `column_comment` column is currently defined as `VARCHAR(256)` in the relational store schema. If an external table contains a column comment longer than 256 characters, the import fails with a MySQL data truncation error.

Column comments are part of table schema metadata and may come from external systems where the comment length is not limited to 256 characters. The current limit is too restrictive and can block table import.

Caused by: java.io.IOException: com.mysql.cj.jdbc.exceptions.MysqlDataTruncation: Data truncation: Data too long for column 'column_comment' at row 29
    at org.apache.gravitino.storage.relational.converters.MySQLExceptionConverter.toGravitinoException(MySQLExceptionConverter.java:44)
    at org.apache.gravitino.storage.relational.utils.ExceptionUtils.checkSQLException(ExceptionUtils.java:33)
    at org.apache.gravitino.storage.relational.service.TableMetaService.insertTable_aroundBody6(TableMetaService.java:153)
    at org.apache.gravitino.storage.relational.service.TableMetaService.insertTable(TableMetaService.java:1)
    at org.apache.gravitino.storage.relational.JDBCBackend.insert(JDBCBackend.java:185)
    at org.apache.gravitino.storage.relational.RelationalEntityStore.put(RelationalEntityStore.java:151)
    at org.apache.gravitino.catalog.TableOperationDispatcher.importTable(TableOperationDispatcher.java:521)

### How to reproduce

Use Gravitino with MySQL as the relational store.
Prepare an external table that contains at least one column whose comment is longer than 256 characters.
Import the table into Gravitino.
The import fails when inserting column metadata into table_column_version_info.
The related schema definition is:
`column_comment` VARCHAR(256) DEFAULT '' COMMENT 'column comment'

### Additional context

table_column_version_info is used to store versioned column metadata and reconstruct table schema snapshots by table version. Therefore, column_comment is necessary in this table, but the current VARCHAR(256) length is too small for imported external metadata.
This issue may affect imported tables from external systems whose column comments are longer than 256 characters. Since Gravitino preserves column metadata in `table_column_version_info`, silently truncating the comment may lose metadata, so increasing the storage length would be preferable.

jarredhj0214:

This PR increases the `table_column_version_info.column_comment` length from `VARCHAR(256)` to `VARCHAR(4096)` for all relational store backends:

- MySQL
- H2
- PostgreSQL

It updates both schema initialization scripts and upgrade scripts so newly initialized databases and upgraded existing databases use the same column comment length.

Importing an external table can fail when a column comment is longer than 256 characters because Gravitino stores column metadata in `table_column_version_info.column_comment`.

Increasing the column length to 4096 follows the discussion in #12038 and avoids import failures for most practical column comments. For comments larger than 4096, users can still customize their database schema definition.

Fix: #12038

The relational store schema now supports column comments up to 4096 characters in `table_column_version_info`.

- Verified all `table_column_version_info.column_comment` definitions in MySQL, H2, and PostgreSQL schema scripts are updated to `VARCHAR(4096)`.
- Verified related upgrade scripts include column type migration statements.
- Ran `git diff --check`.


scripts/h2/schema-0.7.0-h2.sql
```diff
@@ -95,7 +95,7 @@ CREATE TABLE IF NOT EXISTS `table_column_version_info` (
     `column_name` VARCHAR(128) NOT NULL COMMENT 'column name',
     `column_position` INT UNSIGNED NOT NULL COMMENT 'column position, starting from 0',
     `column_type` CLOB NOT NULL COMMENT 'column type',
-    `column_comment` VARCHAR(256) DEFAULT '' COMMENT 'column comment',
+    `column_comment` VARCHAR(4096) DEFAULT '' COMMENT 'column comment',
```
yuqi1129: We should not change this file, only a new `schema-2.0.0.sql` and `upgrade-1.3.0-2.0.0.sql` are needed.

jarredhj0214: Thanks for the reminder. I have reverted the changes to historical schema/upgrade scripts and kept this change only in the 2.0.0 schema files and the 1.3.0-to-2.0.0 upgrade scripts.

yuqi1129:
Would you like to modify the table comments as well?

jarredhj0214:
> Would you like to modify the table comments as well?

I checked the table comment definitions. The table comments are already stored as MEDIUMTEXT for MySQL, CLOB for H2, and TEXT for PostgreSQL in table_version_info, so there is no 256-character limit there.

yuqi1129 (approved):

Issue #12095, opened 2026-07-20
[Subtask] Wire X-Gravitino-Active-Roles into the request path with validation


bharos:
### Describe the subtask

Parse the `X-Gravitino-Active-Roles` header on each request and make the declared active roles available to the authorization path, so role narrowing takes effect. Validate the header value:

- Malformed value → `400 Bad Request`.
- Well-formed value naming a role the caller does not hold → `403 Forbidden`.

An absent header behaves as `ALL` (no narrowing), preserving current behavior.

### Parent issue

#11965

PR #12096, opened 2026-07-20
[#12095] feat(auth): parse X-Gravitino-Active-Roles into the request path

bharos:

Parse the `X-Gravitino-Active-Roles` header and carry the declared active roles on the `UserPrincipal`, so the narrowing from #11967 takes effect.

- `UserPrincipal`: carry an immutable `ActiveRoles` (defaults to `ALL`) alongside the existing per-request `accessToken`; `withActiveRoles()` attaches the parsed value.
- `AuthenticationFilter`: parse the header after authentication and attach it to the principal. A malformed value maps to `400` (the Lance filter maps the same).
- `AuthorizationRequestContext`: read `activeRoles` from the current `UserPrincipal`, so every authorization decision uses the request's declaration.

Membership validation (`403` for an unheld role) is metalake-scoped and follows in a later PR. Part of #11965.

#11967 added the narrowing enforcement, but nothing populated the active roles, so it never activated. This wires the header through the request path.

Yes, additive and backward compatible. A request may send `X-Gravitino-Active-Roles: <role>[,<role>] | ALL | NONE` to narrow the roles used for authorization; a malformed value returns `400`. An absent header behaves as before (`ALL`).

New unit tests for `UserPrincipal.withActiveRoles`, the filter's header parsing (valid / absent / malformed → `400`), and `AuthorizationRequestContext` reading the active roles from the principal. Verified with the module `test` and `spotlessCheck` tasks for `core`, `server-common`, and `lance-rest-server`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/UserPrincipal.java
+++ b/core/src/main/java/org/apache/gravitino/UserPrincipal.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import javax.annotation.Nullable;
+import org.apache.gravitino.auth.ActiveRoles;
 
 /**
  * A simple implementation of Principal that holds a username, optional group membership, and
@@ -38,6 +39,7 @@ public class UserPrincipal implements Principal {
   private final String username;
   private final List<UserGroup> groups;
   @Nullable private final String accessToken;
+  private final ActiveRoles activeRoles;
 
   /**
    * Constructs a UserPrincipal with the given username.
@@ -80,6 +82,14 @@ public class UserPrincipal implements Principal {
    */
   public UserPrincipal(
       final String username, final List<UserGroup> groups, @Nullable final String accessToken) {
+    this(username, groups, accessToken, ActiveRoles.all());
+  }
+
+  private UserPrincipal(
+      final String username,
+      final List<UserGroup> groups,
+      @Nullable final String accessToken,
+      final ActiveRoles activeRoles) {
     Preconditions.checkArgument(username != null, "UserPrincipal must have the username");
     this.username = username;
     this.groups =
@@ -87,6 +97,7 @@ public class UserPrincipal implements Principal {
             ? Collections.unmodifiableList(new ArrayList<>(groups))
             : Collections.emptyList();
     this.accessToken = accessToken;
+    this.activeRoles = Objects.requireNonNull(activeRoles, "activeRoles must not be null");
   }
 
   /**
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/apache/gravitino/auth/ActiveRolesContext.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Objects;
+
+/**
+ * Request-scoped holder for the caller's declared {@link ActiveRoles} (role assumption).
+ *
+ * <p>Backed by a {@link ThreadLocal}; the value must be {@link #clear() cleared} when the request
+ * completes so it never leaks to the next request on a pooled thread.
+ */
+public final class ActiveRolesContext {
```
roryqi: I prefer moving this into UserPrincipal instead of thread local variable.

bharos: Done , moved active roles onto UserPrincipal instead of a thread-local. Since UserPrincipal is immutable, `withActiveRoles(...)` returns a copy with the roles attached, and the filter binds that copy before doAs. Hope creating and returning a copy is OK here ? Otherwise please let me know if you think of any better way.

roryqi: OK for me.

core/src/main/java/org/apache/gravitino/UserPrincipal.java
```diff
@@ -113,6 +124,27 @@ public List<UserGroup> getGroups() {
     return groups;
   }
 
+  /**
+   * Returns the roles the caller declared active for this request (role assumption); {@link
+   * ActiveRoles#all()} when none was declared.
+   *
+   * @return the active-role declaration
+   */
+  public ActiveRoles getActiveRoles() {
```
roryqi, line 133: Do we need to override the hashCode and equals?

bharos: I think that no override is needed because equals/hashCode intentionally cover only identity (username + groups).
I see accessToken is also left out of equals/hashCode as it is also a per-request variable (similar to activeRoles), whereas the equality is built on the static user identity.

Adding activeRoles would also be fine since a principal only exists per request, but I'm inclined to follow the same convention as accessToken and skip it. Let me know if you'd prefer it included (either ways looks ok to me)

roryqi: OK.

roryqi (approved):
LGTM.

PR #12098, opened 2026-07-20
[#12097] improvement(core): Validate names before normalizing in CapabilityHelpers

### LINKED ISSUE #12097: [Improvement] Validate names before normalizing in CapabilityHelpers to support context-dependent name folding

`CapabilityHelpers.applyCapabilities`/`applyCapabilitiesOnName` currently validates a name (`applyNameSpecification`) after it has already been normalized (`applyCaseSensitiveOn*`). This works for catalogs whose folding is context-independent (e.g. always upper/lower-case), but breaks down for catalogs whose folding depends on the name's original form (e.g. a quoted identifier keeps case/spaces, an unquoted one gets folded) — validation ends up checking the normalized form instead of what the caller actually supplied.

Similarly, `SchemaNormalizeDispatcher#listSchemas` and `TableNormalizeDispatcher#listTables` re-apply `normalizeName` to identifiers the underlying catalog already returned in canonical form, which isn't guaranteed to be idempotent for such catalogs.

### How should we improve?

Validate the original name before normalizing it in `CapabilityHelpers`. Drop the redundant re-normalization of catalog-returned identifiers in `listSchemas`/`listTables`, trusting the catalog to already return canonical names.

diqiu50:

Reorder validation and normalization in `CapabilityHelpers` so `applyNameSpecification`
runs on the original name before `applyCaseSensitiveOn*`/`normalizeName`. Remove the
redundant re-normalization of identifiers already returned by `listSchemas`/`listTables`
in `SchemaNormalizeDispatcher`/`TableNormalizeDispatcher`.

For catalogs whose name folding depends on the name's original form (e.g. a quoted
identifier keeps case/spaces, an unquoted one gets folded), validating after
normalization checks the wrong form of the name. Similarly, re-applying `normalizeName`
to identifiers a catalog already returned in canonical form isn't guaranteed to be
idempotent for such catalogs and can corrupt an already-correct name.

Fix: #12097

No.

Added a unit test; existing core module tests pass.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
@@ -56,8 +56,8 @@ public class SchemaNormalizeDispatcher implements SchemaDispatcher {
               NameIdentifier.of(namespace.levels())));
     }
 
-    // The catalog guarantees listSchemas() already returns names in their canonical, legal
-    // form, so no re-normalization is needed here.
+    // listSchemas() is assumed to already return names in their canonical, legal form, so no
+    // re-normalization is applied here.
     return dispatcher.listSchemas(namespace);
   }
 
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
@@ -33,6 +33,10 @@ import org.apache.gravitino.exceptions.NoSuchSchemaException;
 import org.apache.gravitino.exceptions.NonEmptySchemaException;
 import org.apache.gravitino.exceptions.SchemaAlreadyExistsException;
 
+/**
+ * Note on list operations: names returned by list methods (e.g. {@link #listSchemas(Namespace)})
+ * are assumed to already be in their canonical, legal form and are not re-normalized here.
+ */
 public class SchemaNormalizeDispatcher implements SchemaDispatcher {
   private final CatalogManager catalogManager;
   private final SchemaDispatcher dispatcher;
@@ -56,8 +60,6 @@ public class SchemaNormalizeDispatcher implements SchemaDispatcher {
               NameIdentifier.of(namespace.levels())));
     }
 
-    // listSchemas() is assumed to already return names in their canonical, legal form, so no
-    // re-normalization is applied here.
     return dispatcher.listSchemas(namespace);
   }
 
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
@@ -23,7 +23,6 @@ import static org.apache.gravitino.catalog.CapabilityHelpers.applyCaseSensitive;
 import static org.apache.gravitino.catalog.CapabilityHelpers.getCapability;
 
 import java.util.Map;
-import org.apache.commons.lang3.ArrayUtils;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.Namespace;
 import org.apache.gravitino.Schema;
@@ -57,10 +56,9 @@ public class SchemaNormalizeDispatcher implements SchemaDispatcher {
               NameIdentifier.of(namespace.levels())));
     }
 
-    NameIdentifier[] identifiers = dispatcher.listSchemas(namespace);
-    // The constraints of the name spec may be more strict than underlying catalog,
-    // and for compatibility reasons, we only apply case-sensitive capabilities here.
-    return normalizeCaseSensitive(identifiers);
+    // The catalog guarantees listSchemas() already returns names in their canonical, legal
+    // form, so no re-normalization is needed here.
+    return dispatcher.listSchemas(namespace);
   }
 
   @Override
@@ -109,14 +107,5 @@ public class SchemaNormalizeDispatcher implements SchemaDispatcher {
   private NameIdentifier normalizeCaseSensitive(NameIdentifier schemaIdent) {
     Capability capabilities = getCapability(schemaIdent, catalogManager);
     return applyCaseSensitive(schemaIdent, Capability.Scope.SCHEMA, capabilities);
-  }
-
-  private NameIdentifier[] normalizeCaseSensitive(NameIdentifier[] schemaIdents) {
-    if (ArrayUtils.isEmpty(schemaIdents)) {
-      return schemaIdents;
-    }
-
-    Capability capabilities = getCapability(schemaIdents[0], catalogManager);
-    return applyCaseSensitive(schemaIdents, Capability.Scope.SCHEMA, capabilities);
   }
 }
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/apache/gravitino/catalog/TableNormalizeDispatcher.java
```diff
@@ -52,8 +51,9 @@ public NameIdentifier[] listTables(Namespace namespace) throws NoSuchSchemaExcep
     // The constraints of the name spec may be more strict than underlying catalog,
     // and for compatibility reasons, we only apply case-sensitive capabilities here.
     Namespace caseSensitiveNs = normalizeCaseSensitive(namespace);
-    NameIdentifier[] identifiers = dispatcher.listTables(caseSensitiveNs);
-    return normalizeCaseSensitive(identifiers);
+    // The catalog guarantees listTables() already returns names in their canonical, legal
+    // form, so no re-normalization is needed here.
+    return dispatcher.listTables(caseSensitiveNs);
```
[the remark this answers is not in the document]
diqiu50: Not need test that case

core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
```diff
@@ -57,10 +56,9 @@ public NameIdentifier[] listSchemas(Namespace namespace) throws NoSuchCatalogExc
...
     }
 
-    NameIdentifier[] identifiers = dispatcher.listSchemas(namespace);
-    // The constraints of the name spec may be more strict than underlying catalog,
-    // and for compatibility reasons, we only apply case-sensitive capabilities here.
-    return normalizeCaseSensitive(identifiers);
+    // The catalog guarantees listSchemas() already returns names in their canonical, legal
+    // form, so no re-normalization is needed here.
+    return dispatcher.listSchemas(namespace);
   }
```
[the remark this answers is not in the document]
diqiu50: same

core/src/test/java/org/apache/gravitino/catalog/TestCapabilityHelpers.java
```diff
@@ -92,4 +127,19 @@ void testApplyCaseSensitiveOnNameAllowsNullNameToPassThrough() {
...
+    // form no longer matches the plain-word pattern. specificationOnName must be checked against
+    // the original name, not the already-normalized one, or this would be wrongly rejected.
+    NameIdentifier quotedIdent =
+        NameIdentifier.of(Namespace.of("metalake", "catalog", "schema"), "\"My Table\"");
+
+    NameIdentifier result =
+        CapabilityHelpers.applyCapabilities(
+            quotedIdent, Capability.Scope.TABLE, QUOTE_AWARE_CAPABILITY);
+
+    Assertions.assertEquals("My Table", result.name());
```
mchades, line 143: This only tests the one-way conversion `"My Table" -> My Table`. Please add a dispatcher-level round-trip test such as `listTables() -> loadTable()`. It currently fails because `My Table` is normalized again to `MY TABLE`.

core/src/main/java/org/apache/gravitino/catalog/SchemaNormalizeDispatcher.java
```diff
@@ -57,10 +56,9 @@ public NameIdentifier[] listSchemas(Namespace namespace) throws NoSuchCatalogExc
               NameIdentifier.of(namespace.levels())));
     }
 
-    NameIdentifier[] identifiers = dispatcher.listSchemas(namespace);
-    // The constraints of the name spec may be more strict than underlying catalog,
-    // and for compatibility reasons, we only apply case-sensitive capabilities here.
-    return normalizeCaseSensitive(identifiers);
+    // listSchemas() is assumed to already return names in their canonical, legal form, so no
+    // re-normalization is applied here.
+    return dispatcher.listSchemas(namespace);
```
mchades, line 63: We should not return the underlying catalog result directly here. `Capability` defines the naming rules exposed by Gravitino, so list results must still be normalized at the core boundary. Otherwise, backend-native names may leak through and become inconsistent with other schema operations. Please keep `normalizeCaseSensitive(identifiers)`.

diqiu50: Skipping re-normalization in listSchemas()/listTables() is intentional here, because the name returned by the backend catalog is already correct.

For example, when creating a table with the quoted name "Aaa", Oracle stores the physical table name as Aaa (case preserved due to quoting). listTables() then returns Aaa, which is already correct. If we re-apply normalizeCaseSensitive() on top of that, Oracle's folding rule (unquoted names are upper-cased) would turn Aaa into AAA — silently corrupting a name that was already right.

This only breaks for catalogs like Oracle where folding depends on the name's original form (quoted vs. unquoted), so normalizeName isn't idempotent. For catalogs with simple, context-independent folding (always upper/lower-case), re-normalizing is harmless since it's already idempotent — but we can't tell the two cases apart here, so the safer fix is to trust the catalog's result as-is.

core/src/test/java/org/apache/gravitino/catalog/TestCapabilityHelpers.java
diqiu50: Added

diqiu50: In this scenario, the double quotes " are just a marker telling Oracle how to fold the name — they're not part of the physical name itself. listTables() returns the physical name without the quotes (e.g. My Table). If you want to loadTable() that same case-sensitive table again, you need to add the quotes back ("My Table"), the same way you'd write it in SQL.

I added a test (testCreateTableListTablesLoadTableRoundTrip) covering the full createTable("\"My Table\"") → listTables() → loadTable("\"My Table\"") round trip, confirming the physical name is preserved correctly throughout when used this way.

core/src/main/java/org/apache/gravitino/catalog/TableNormalizeDispatcher.java
```diff
@@ -52,8 +51,9 @@ public NameIdentifier[] listTables(Namespace namespace) throws NoSuchSchemaExcep
     // The constraints of the name spec may be more strict than underlying catalog,
     // and for compatibility reasons, we only apply case-sensitive capabilities here.
     Namespace caseSensitiveNs = normalizeCaseSensitive(namespace);
-    NameIdentifier[] identifiers = dispatcher.listTables(caseSensitiveNs);
-    return normalizeCaseSensitive(identifiers);
+    // listTables() is assumed to already return names in their canonical, legal form, so no
+    // re-normalization is applied here.
+    return dispatcher.listTables(caseSensitiveNs);
```
mchades, line 58: If catalog-returned identifiers must not be re-normalized because `normalizeName()` may be non-idempotent, the same applies to `listViews()`, `listFilesets()`, `listFunctions()`, `listModels()`, and `listTopics()`, which still normalize their results. Please update these paths consistently.

diqiu50: fixed

mchades: Thanks. One remaining path has the same issue: `PartitionNormalizeDispatcher#listPartitionNames()` and `#listPartitions()` still re-normalize catalog-returned partition names. With the quote-aware capability, `My Partition` is still changed to `MY PARTITION`. Please update both paths and add coverage.

diqiu50: fixed

mchades (approved):

PR #12101, opened 2026-07-20
[MINOR] improvement(authz): Add debug logging for list-authorization short-circuit

yuqi1129:

Adds `DEBUG`-level logging around the list-authorization parent-scope short-circuit (`MetadataAuthzHelper.filterByExpression` / `allVisibleViaParentScope`, introduced in #11778) so it is possible to tell, per request and per principal, whether the short-circuit fired and — when it did not — exactly which condition disqualified it:

- **HIT**: every listed object is returned via a parent-scope grant and the per-object authorization loop is skipped.
- **MISS / skipped**, with the reason: no short-circuit spec registered / filter-expression mismatch, listed objects span multiple parent namespaces, the principal has no parent-scope grant, or a deny policy exists on the relevant privileges.

Every log line now includes the current principal, the entity type, and the metalake.

This is purely diagnostic — no behavior change.

When a `list` call (e.g. `list catalogs`) is slow, there is currently no way to observe whether the #11778 short-circuit applied. The `verbose` (`details=true`) path and finer-grained or denied grants fall back to per-object authorization; these logs make that visible and greatly speed up diagnosing slow-list reports.

Fix: N/A (minor diagnostic improvement)

No. Debug logging only.

No behavior change (only added `LOG.debug` calls and an extracted `hasDeny` local; control flow is unchanged). Existing tests cover the short-circuit logic. Verified `:server-common:compileJava` succeeds.

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/MetadataAuthzHelper.java
@@ -206,13 +206,29 @@ public class MetadataAuthzHelper {
       String expression,
       Entity.EntityType entityType,
       NameIdentifier[] nameIdentifiers) {
-    if (enableAuthorization()
-        && nameIdentifiers.length > 0
-        && allVisibleViaParentScope(metalake, expression, entityType, nameIdentifiers)) {
-      // A privilege granted at a parent scope (metalake/catalog/schema) makes every object in the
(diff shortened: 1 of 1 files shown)


roryqi (approved):
LGTM.

roryqi:
Do u need to cherry pick to branch 1.3?

yuqi1129:
> Do u need to cherry pick to branch 1.3?

I would be better to cherry-pick to branch-1.3

PR #12105, opened 2026-07-20
[#12104] fix(spark-connector): Map ExternalType to StringType to prevent UnsupportedOperationException

### LINKED ISSUE #12104: fix(spark-connector): Map ExternalType to StringType to prevent UnsupportedOperationException
### Describe what's wrong

When loading a JDBC catalog table through the Spark connector, if any column is mapped to `Types.ExternalType` in Gravitino metadata, `loadTable()` throws `UnsupportedOperationException` and the entire table becomes unusable.

Root cause: `SparkTypeConverter.toSparkType()` handles ~20 Gravitino native types but does not have a branch for `Types.ExternalType`. Any column with this type falls through to the final `throw` statement.

The problem affects all JDBC catalogs where `toGravitino()` produces `ExternalType` for database-specific types:

| Catalog | Types producing ExternalType |
|---------|------------------------------|
| ClickHouse | IPv4, IPv6, Date32, DateTime64, Decimal(>38), LowCardinality, Map, Tuple, Array, JSON, Enum, Variant, Geo, Nested, Dynamic, AggregateFunction |
| Doris | LARGEINT, BITMAP, HLL, JSON, Variant, IPV4, IPV6 |
| PostgreSQL | interval, custom types, and unknown-type catch-all |
| MySQL | TINYTEXT, MEDIUMTEXT, YEAR, ENUM, SET, JSON, GEOMETRY and other types |
| OceanBase | Unknown-type catch-all |
| Hologres | Unknown-type catch-all |
| Glue (via Hive connector) | Unknown Hive types |

The Glue catalog uses `SparkHiveTypeConverter` which also delegates unknown types to the base class `SparkTypeConverter.toSparkType()`, triggering the same crash.

```
java.lang.UnsupportedOperationException: Not support ExternalType: IPv4
    at org.apache.gravitino.spark.connector.SparkTypeConverter.toSparkType(SparkTypeConverter.java:196)
    at org.apache.gravitino.spark.connector.jdbc.SparkJdbcTypeConverter.toSparkType(SparkJdbcTypeConverter.java:39)
    at org.apache.gravitino.spark.connector.utils.GravitinoTableInfoHelper.getSchema(GravitinoTableInfoHelper.java:89)
```

### How to reproduce

1. Register a ClickHouse catalog in Gravitino with a table containing an `IPv4` column:
   ```sql
   CREATE TABLE test.ipv4_table (id Int32, ip IPv4) ENGINE = MergeTree ORDER BY id;
   ```

2. Load the table through the Spark connector:
   ```sql
   SELECT * FROM clickhouse_catalog.test.ipv4_table;
   ```

3. Result: `UnsupportedOperationException` — the entire table is inaccessible.

### Proposed fix

Add an `ExternalType → DataTypes.StringType` mapping in `SparkTypeConverter.toSparkType()` before the final `throw`. This follows the existing pattern used for `UUID` (also mapped to `StringType` because Spark has no native UUID type). Placing the fix in the base class automatically covers all subclasses (JDBC, Hive, and Glue connectors across all Spark versions 3.3/3.4/3.5).

jiangxt2:

Add `ExternalType → DataTypes.StringType` mapping in `SparkTypeConverter.toSparkType()` base class. This prevents tables with `ExternalType` columns from throwing `UnsupportedOperationException` during `loadTable()` in the Spark connector.

Placing the fix in the base class covers all subclasses across Spark versions 3.3/3.4/3.5 (JDBC, Hive, and Glue connectors) with a single change.

`SparkTypeConverter.toSparkType()` handles ~20 Gravitino native types but has no branch for `Types.ExternalType`. When a JDBC catalog table contains database-specific types (e.g. ClickHouse `IPv4`/`IPv6`, Doris `LARGEINT`/`BITMAP`), the Gravitino server maps them to `ExternalType`, which then falls through to the final `throw` in `toSparkType()` — crashing the entire table load.

Affected catalogs include ClickHouse (16+ types), Doris (7 types), MySQL, PostgreSQL, OceanBase, Hologres, and Glue.

The fix follows the existing pattern used for `UUID` (also mapped to `StringType` because Spark has no native UUID type).

Fix: #12104

Yes — tables with `ExternalType` columns are now loadable in Spark. The columns are presented as `StringType`, and users can `CAST` to the desired type as needed, e.g. `CAST(ip AS STRING)`.

- Unit tests: `TestSparkJdbcTypeConverter.testConvertExternalTypeToSparkString` (spark-common) + `TestSparkJdbcTypeConverter34` (v3.4), covering LARGEINT/BITMAP/IPv4
- Docker integration tests:
  - `CatalogClickHouseIT.testIPv4ColumnIsExternalType` — verifies ClickHouse IPv4 columns are exposed as ExternalType in Gravitino metadata
  - `CatalogDoris4xIT.testLargeintColumnIsExternalType` — verifies Doris LARGEINT columns round-trip correctly

### DIFF
--- a/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/SparkTypeConverter.java
+++ b/spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/SparkTypeConverter.java
@@ -192,6 +192,11 @@ public class SparkTypeConverter {
       return DataTypes.createStructType(fields);
     } else if (gravitinoType instanceof Types.NullType) {
       return DataTypes.NullType;
+    } else if (gravitinoType instanceof Types.ExternalType) {
+      // ExternalType represents types with no Gravitino-native mapping (e.g. ClickHouse
+      // IPv4/IPv6, Doris LARGEINT/BITMAP, Glue unknown types). Map to StringType so that
+      // tables containing these columns remain loadable in Spark. Users can CAST as needed.
+      return DataTypes.StringType;
     }
     throw new UnsupportedOperationException("Not support " + gravitinoType.toString());
   }
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/test/java/org/apache/gravitino/catalog/clickhouse/integration/test/CatalogClickHouseIT.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/test/java/org/apache/gravitino/catalog/clickhouse/integration/test/CatalogClickHouseIT.java
@@ -2805,4 +2805,28 @@ public class CatalogClickHouseIT extends BaseIT {
     // SHOW CREATE TABLE output, so one settings key is expected even without explicit SETTINGS.
     Assertions.assertEquals(1, settingsCount);
   }
+
+  @Test
+  void testIPv4ColumnIsExternalType() {
+    // Verify that ClickHouse IPv4 columns are exposed as ExternalType in Gravitino metadata.
+    // This is the first half of the Spark crash path: the Gravitino metadata
+    // layer must return ExternalType for ClickHouse-specific types, which then reaches
+    // SparkTypeConverter.toSparkType() — the crash itself is covered by unit tests
+    // (TestSparkJdbcTypeConverter.testConvertExternalTypeToSparkString).
+    String tableName = GravitinoITUtils.genRandomName("test_ipv4_type_");
+    clickhouseService.executeQuery(
+        String.format(
+            "CREATE TABLE `%s`.`%s` (id Int32, ip IPv4 COMMENT 'IPv4 address') "
+                + "ENGINE = MergeTree ORDER BY id",
+            schemaName, tableName));
+
+    Table loaded = catalog.asTableCatalog().loadTable(NameIdentifier.of(schemaName, tableName));
+
+    Column ipCol = loaded.columns()[1];
+    Assertions.assertEquals("ip", ipCol.name());
+    Assertions.assertTrue(
+        ipCol.dataType() instanceof Types.ExternalType,
+        "ClickHouse IPv4 must be ExternalType — otherwise Spark crash path is not triggered");
+    Assertions.assertEquals("IPv4", ((Types.ExternalType) ipCol.dataType()).catalogString());
+  }
 }
--- a/catalogs/catalog-jdbc-doris/src/test/java/org/apache/gravitino/catalog/doris/integration/test/CatalogDoris4xIT.java
+++ b/catalogs/catalog-jdbc-doris/src/test/java/org/apache/gravitino/catalog/doris/integration/test/CatalogDoris4xIT.java
@@ -27,6 +27,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import com.google.common.collect.Maps;
 import java.io.IOException;
@@ -64,6 +65,7 @@ import org.apache.gravitino.rel.partitions.ListPartition;
 import org.apache.gravitino.rel.partitions.Partition;
 import org.apache.gravitino.rel.partitions.Partitions;
 import org.apache.gravitino.rel.partitions.RangePartition;
+import org.apache.gravitino.rel.types.Type;
 import org.apache.gravitino.rel.types.Types;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.AfterAll;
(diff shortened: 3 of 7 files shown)


spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/SparkTypeConverter.java
```diff
@@ -192,6 +192,11 @@ public DataType toSparkType(Type gravitinoType) {
       return DataTypes.createStructType(fields);
     } else if (gravitinoType instanceof Types.NullType) {
       return DataTypes.NullType;
+    } else if (gravitinoType instanceof Types.ExternalType) {
+      // ExternalType represents types with no Gravitino-native mapping (e.g. ClickHouse
+      // IPv4/IPv6, Doris LARGEINT/BITMAP, Glue unknown types). Map to StringType so that
+      // tables containing these columns remain loadable in Spark. Users can CAST as needed.
+      return DataTypes.StringType;
```
yuqi1129, line 199: Can you add some Spark IT to verify whether it works? I'm also curious about how we handle gravitino external types and convert them to their inner type

jiangxt2: I've added a Spark IT in this update — SparkJdbcMysqlCatalogIT.testExternalTypeEnumColumn() — which verifies the full path end-to-end using a MySQL ENUM column.

Regarding how we handle ExternalType, here's the conversion flow using ClickHouse IPv4 as an example:

1. JDBC catalog layer: ClickHouseTypeConverter.toGravitino("IPv4") finds no native mapping → wraps it as Types.ExternalType.of("IPv4"). The original type name is preserved via catalogString().
2. Spark connector layer: SparkTypeConverter.toSparkType(ExternalType) → returns DataTypes.StringType. This follows the same pattern as UUID (which also has no native Spark type).
3. User experience in Spark: The column appears as STRING. The JDBC driver returns the value in its string representation (e.g. "192.168.1.1" for IPv4), so SELECT ip FROM table works directly — no explicit CAST is needed in most cases.

This approach ensures tables with database-specific types (ClickHouse IPv4/IPv6, Doris LARGEINT/BITMAP, MySQL ENUM, etc.) remain loadable in Spark without throwing UnsupportedOperationException, while preserving the original type name in Gravitino metadata for catalog-level queries.

yuqi1129: It seems that `The column appears as STRING` is rather tricky. It is indeed that most data types can be converted to varchar/string type, but it will lose precision, and the efficiency is not good.

@diqiu50 What's your thought on this point?

jiangxt2: Thanks for the feedback! I agree that blindly converting everything to STRING has drawbacks — precision loss and efficiency concerns are both valid.

However, I'd like to clarify the scope and rationale of this PR:

**1. This is a last-resort fallback, not a general-purpose conversion strategy.**

ExternalType only appears when a type has *no* Gravitino-native mapping whatsoever (IPv4, BITMAP, ENUM, etc.). The alternative today is `UnsupportedOperationException` during `loadTable()` — which makes the *entire table* inaccessible in Spark, even for columns that have perfectly valid types.

**2. This follows established precedents in both Trino connector and Spark itself:**

- Gravitino's own Trino connector: `MySQLExternalDataType` maps unknown types to `VARCHAR` with an explicit comment: *"For unsupported data types, standardize mapping to varchar for enhanced usability. Equivalent to default configuration of the unsupported_type_handling is CONVERT_TO_VARCHAR"*
- Spark's built-in PostgreSQL JDBC dialect: `case Types.OTHER => Some(StringType)` — same pattern, same rationale.
- The existing `UUID → StringType` mapping in this same `SparkTypeConverter` follows the same philosophy.

**3. For types where a better mapping exists, we can improve incrementally:**

Some types like `MEDIUMINT` could map to `IntegerType` (as the Trino connector already does). But for types like `LARGEINT` (128-bit signed, max 39 digits — exceeding Spark's max decimal precision of 38), there's genuinely no lossless numeric type in Spark — STRING is the only option that preserves the full value. Similarly, `BITMAP`/`HLL` are opaque aggregate types with no meaningful numeric or binary representation in Spark's type system.

**4. The data path is consistent with how Spark reads any VARCHAR column:**

When Spark reads a StringType column, it calls `ResultSet.getString()`, which the JDBC driver natively supports for all types. The data path is identical to how Spark reads any VARCHAR column — no custom serialization layer is introduced.

**Summary:** This PR addresses an availability issue (table completely inaccessible → table loadable with some columns degraded to STRING). A configurable approach like `gravitino.spark.unsupported-type-handling = CONVERT_TO_VARCHAR | IGNORE | FAIL` (similar to Trino's `unsupported_type_handling`) could be a follow-up improvement to give users more control.

jerryshao:
@mchades can you please also take a look?

spark-connector/spark-common/src/main/java/org/apache/gravitino/spark/connector/SparkTypeConverter.java
mchades: I checked the Spark JDBC implementation for the Spark versions supported here. Spark derives the table schema from `ResultSetMetaData`, consulting `JdbcDialect.getCatalystType` first and then its generic JDBC mapping ([`JdbcUtils.getSchema`](https://github.com/apache/spark/blob/v3.5.6/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala#L262-L312)). For example, JDBC `BLOB` is mapped to `BinaryType`, not `StringType` ([generic mapping](https://github.com/apache/spark/blob/v3.5.6/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala#L178-L220)). `JDBCTable` then uses this resolved schema for both scan and write builders ([source](https://github.com/apache/spark/blob/v3.5.6/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/jdbc/JDBCTable.scala#L43-L52)).

This means `ExternalType` does not necessarily imply that Spark has no suitable type: Gravitino may preserve only the external type name while the Spark JDBC dialect/driver can still resolve a native Spark type. A global `ExternalType -> StringType` mapping discards that information; for `BLOB`, it also changes the Spark JDBC access path from bytes to strings.

Could we avoid this global mapping? `SparkJdbcTable` already receives the loaded `JDBCTable`, so we could pass `jdbcTable.schema()` to `GravitinoTableInfoHelper` and reuse the matching JDBC field type only when the Gravitino type is `ExternalType`, while retaining the Gravitino field nullability, comment, and metadata. If no matching field exists, it would be safer to fail explicitly rather than silently coercing it to `StringType`.

Please also add coverage for at least `ENUM -> StringType` and `BLOB -> BinaryType`.

jiangxt2: @mchades Thanks for the thorough analysis — you are right. I traced the `JDBCTableCatalog.loadTable()` → `JDBCRDD.resolveTable()` → `JdbcUtils.getSchema()` chain and confirmed that Spark JDBC independently resolves types from `ResultSetMetaData` with dialect-specific overrides. For example, `BLOB → BinaryType` and `ENUM → StringType` are correctly resolved by Spark JDBC without any Gravitino involvement.

I will update the approach: instead of a global `ExternalType → StringType` mapping in `toSparkType()`, I will leverage `jdbcTable.schema()` (already available in `SparkJdbcTable` constructor) to fall back to the JDBC-resolved type when Gravitino type is `ExternalType`. If no matching field can be found, fail explicitly. I will also add IT coverage for both `ENUM → StringType` and `BLOB → BinaryType` paths.

Will push the update once ready. Also curious about @diqiu50 thoughts on this direction.

diqiu50: I think the current behavior of falling back to StringType for ExternalType is a reasonable default for the connector. The primary goal is to keep queries working instead of failing because of unsupported types.

Leveraging the JDBC-resolved schema in Spark is a good optimization and can provide more accurate type mappings where available. However, I don't think we should fail the query if a matching JDBC field cannot be found. Falling back to StringType is a more user-friendly behavior than failing explicitly.

mchades: I understand the availability concern, but keeping queries running should not be the primary goal. Since `SparkJdbcTable` is built on Spark JDBC, its default behavior should remain consistent with the native connector.

A query that succeeds with silently changed semantics can be worse than an explicit failure. For example, mapping `BLOB` to `StringType` changes `getBytes`/`setBytes` to `getString`/`setString`. Exposing different schemas for the same table also increases debugging and explanation costs.

Therefore, `jdbcTable.schema()` should be authoritative. If the schemas cannot be reconciled, we should fail explicitly by default; a `StringType` fallback could be offered as an opt-in policy with a warning.

jiangxt2:
After reconsidering the design, I agree that applying an `ExternalType -> StringType` fallback in the shared `SparkTypeConverter` is too broad. 

The fallback should be handled in each engine-specific Spark Connector path, where the engine's physical schema and read semantics are available. For example, Doris-specific external types should be handled by the Doris-specific connector path rather than by a global conversion rule. 

Since this PR implements the global fallback, it no longer matches the intended design. I will close this PR for now and revisit the required handling as part of the corresponding engine-specific connector implementations.

PR #12111, opened 2026-07-20
[#12018, #12019] feat(core): add safe-type allowlist and cross-node cache invalidation

### LINKED ISSUE #12018: [Subtask] Restrict the entity cache to the entity types that are safe to cache per node

### LINKED ISSUE #12019: [Subtask] Make the caffeine entity cache multi-node correct via cross-node change-log invalidation
### Describe the subtask

Today the entity store cache only invalidates on the node that made the change, so other nodes keep serving stale data until their entries expire. This subtask makes the default `caffeine` cache correct across nodes by reusing the existing `entity_change_log` and `EntityChangeLogPoller` that the jcasbin and catalog caches already use, so no new infrastructure is added.

It also adds a coherence marker to the cache SPI so the write path can tell a per-node cache (which must be told about changes made on other nodes) from a shared cache (which does not). The later Redis implementation relies on the same marker.

This subtask should:

- Add `EntityCache.coherence()` returning `LOCAL_PER_NODE | SHARED`, implement `LOCAL_PER_NODE` for `CaffeineEntityCache`, and add the write-path coherence gate.
- Add an entity-store `EntityChangeLogListener` that replays each ALTER/DROP row as an `invalidate(ident, type)`, with a container drop cascading to cached children through the existing forward prefix scan (`cacheIndex`).
- Register the listener as a third consumer of the existing poller, next to the catalog cache and the jcasbin id-mapping cache; the structural entities (metalake, catalog, schema, table, topic, view, fileset, model) already emit rows.
- Add a change-log emit point for **tag and policy** (their MetaServices emit no row today), written in the same transaction as the tag/policy write, so a per-node cache can refresh them across nodes.
- Reuse the lag-safe, id-based polling built for the change log (see #11736) and keep the cache TTL as a backstop.
- Add multi-node integration tests: an ALTER/DROP on node A is reflected on node B within one poll interval (table / schema / catalog and tag / policy), a schema drop clears the child tables on the other node, and adding the new consumer does not change the two existing ones.

This is the main Phase 1 deliverable of #11737: `gravitino.cache.enabled=true` becomes safe in multi-node deployments with the default cache.

### Parent issue

yuqi1129:

Make the default `caffeine` entity store cache safe for multi-node deployments by restricting it to explicitly approved entity types and replaying the existing `entity_change_log` through `EntityChangeLogPoller`.

- **Cache eligibility**: cache only metalake, catalog, schema, table, topic, view, fileset, tag, policy, and job. Model, model version, function, relation-derived entities, and newly added entity types fall through to the `EntityStore`.
- **Coherence marker**: add `EntityCache.coherence()` with `LOCAL_PER_NODE` and `SHARED`; `BaseEntityCache` defaults to `LOCAL_PER_NODE`.
- **Reader side**: add `EntityCacheChangeLogListener` to replay ALTER/DROP rows as cache invalidations, including hierarchical child invalidation through the existing prefix index. Register it only for enabled `LOCAL_PER_NODE` caches.
- **Writer side**: emit transactional change-log rows for tag, policy, and job update/delete paths that previously had none.

The entity cache previously admitted types that could not safely be served from a per-node copy, and only the writing node invalidated its local entries. In a multi-node deployment, another node could therefore continue serving stale metadata until expiration.

This change limits the cache to the reviewed type set and reuses the existing change-log infrastructure to invalidate those entries across nodes without adding a new dependency or schema.

Fix: #12018

Fix: #12019

No. This changes internal caching behavior. Excluded types are read directly from the `EntityStore`.

- `./gradlew :core:spotlessApply --max-workers=2`
- Cache, listener, cross-node invalidation, change-log service, and relational entity store tests.
- `TestEntityChangeLogPoller`, `TestTagMetaService`, `TestPolicyMetaService`, and `TestJobMetaService` regression tests.


yuqi1129:
No needed.

PR #12120, opened 2026-07-21
[#12093] fix(doris): Load partition assignments

### LINKED ISSUE #12093: [Bug report] Doris list/range partition assignments are empty when loading a table
## Describe the bug

When creating a Doris table with list partitions, the create response correctly returns `partitioning.assignments` with the partition data. However, when loading the same table, the `partitioning.assignments` field returns an empty array `[]`.

The same issue also affects range partition assignments.

## To Reproduce

1. Create a Doris table with list partition via Gravitino REST API
2. Observe the create response - `partitioning.assignments` correctly contains the partition data
3. Load the same table via GET API
4. Observe the load response - `partitioning.assignments` is empty `[]`

### Create response (correct)
```json
"partitioning": [
  {
    "strategy": "list",
    "fieldNames": [["col"]],
    "assignments": [
      {
        "type": "list",
        "name": "p1",
        "lists": [
          [{"type": "literal", "dataType": "integer", "value": "1"}],
          [{"type": "literal", "dataType": "integer", "value": "2"}],
          [{"type": "literal", "dataType": "integer", "value": "3"}]
        ],
        "properties": null
      }
    ]
  }
]
```

### Load response (assignments empty)
```json
"partitioning": [
  {
    "strategy": "list",
    "fieldNames": [["col"]],
    "assignments": []
  }
]
```

## Expected behavior

Loading a Doris table with list/range partitions should return the same `partitioning.assignments` data as when the table was created.

## Root cause

In `DorisUtils.extractPartitionInfoFromSql()` (`catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/utils/DorisUtils.java:99-125`), the method only parses the `PARTITION BY LIST(col)` clause from `SHOW CREATE TABLE` output to extract column names, but does **not** parse the individual partition definitions (`PARTITION p1 VALUES IN (...)`).

As a result, it calls `Transforms.list(filedNames)` (line 114) which is the overload **without** assignments, instead of `Transforms.list(filedNames, assignments)`.

The same issue exists for range partitions at line 116: `Transforms.range(new String[] {columns[0]})` is called without assignments.

### Relevant code

```java
// DorisUtils.java:111-117
if (LIST_PARTITION.equals(partitionType)) {
    String[][] filedNames =
        Arrays.stream(columns).map(s -> new String[] {s}).toArray(String[][]::new);
    return Optional.of(Transforms.list(filedNames));  // <-- missing assignments
} else if (RANGE_PARTITION.equals(partitionType)) {
    return Optional.of(Transforms.range(new String[] {columns[0]}));  // <-- missing assignments
}
```

### Suggested fix

The `extractPartitionInfoFromSql` method (or `getTablePartitioning` in `DorisTableOperations`) should also parse the partition definitions from the DDL and construct the appropriate `ListPartition[]` / `RangePartition[]` assignments to pass to `Transforms.list(filedNames, assignments)` / `Transforms.range(fieldName, assignments)`.

Alternatively, the existing `DorisTablePartitionOperations.fromDorisPartition()` method already has the logic to parse partition data from `SHOW PARTITIONS` output. This could be leveraged during table load to populate assignments.

## Additional context

- Gravitino version: current main branch
- Catalog: jdbc-doris

yuqi1129:

Load Doris list and range partition assignments from `SHOW PARTITIONS` when loading table metadata.

Reuse the existing Doris partition conversion logic for both table loading and partition operations, and add coverage for loaded list/range assignments.

Doris table loading currently restores only partition fields from `SHOW CREATE TABLE`. As a result, list and range partition assignments are returned as empty arrays.

Fix: #12093

Yes. Loading a Doris table now returns its list or range partition assignments in `partitioning.assignments`.

- Added assertions for loaded range partition assignments.
- Added assertions for loaded multi-column list partition assignments.
- Ran:

  `./gradlew :catalogs:catalog-jdbc-doris:spotlessApply :catalogs:catalog-jdbc-doris:test -PskipITs -PskipDockerTests=true`

The Docker-tagged Doris integration tests were not executed locally because Docker was unavailable.

### DIFF
--- a/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
+++ b/catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
@@ -642,28 +642,25 @@ public class DorisTableOperations extends JdbcTableOperations {
       Transform partitioning = transform.get();
       Map<String, Type> columnTypes =
           DorisTablePartitionOperations.getColumnTypes(connection, tableName, typeConverter);
-      List<Partition> assignments = new ArrayList<>();
       String showPartitionsSql = String.format("SHOW PARTITIONS FROM `%s`", tableName);
-      try (Statement partitionStatement = connection.createStatement();
-          ResultSet partitions = partitionStatement.executeQuery(showPartitionsSql)) {
-        while (partitions.next()) {
-          assignments.add(
-              DorisTablePartitionOperations.fromDorisPartition(
-                  partitions, partitioning, columnTypes));
-        }
-      }
+      List<Partition> assignments =
+          DorisTablePartitionOperations.loadPartitions(
+              connection, showPartitionsSql, partitioning, columnTypes);
 
       if (partitioning instanceof Transforms.ListTransform) {
         String[][] fieldNames = ((Transforms.ListTransform) partitioning).fieldNames();
         return new Transform[] {
           Transforms.list(fieldNames, assignments.toArray(new ListPartition[0]))
         };
+      } else if (partitioning instanceof Transforms.RangeTransform) {
+        String[] fieldName = ((Transforms.RangeTransform) partitioning).fieldName();
+        return new Transform[] {
+          Transforms.range(fieldName, assignments.toArray(new RangePartition[0]))
+        };
       }
 
-      String[] fieldName = ((Transforms.RangeTransform) partitioning).fieldName();
-      return new Transform[] {
-        Transforms.range(fieldName, assignments.toArray(new RangePartition[0]))
-      };
+      throw new UnsupportedOperationException(
+          String.format("%s is not a supported partition transform", partitioning));
     }
   }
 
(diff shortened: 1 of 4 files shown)


catalogs/catalog-jdbc-doris/src/main/java/org/apache/gravitino/catalog/doris/operation/DorisTableOperations.java
```diff
@@ -633,7 +635,32 @@ protected Transform[] getTablePartitioning(
...
+        return Transforms.EMPTY_TRANSFORM;
+      }
+
+      Transform partitioning = transform.get();
+      Map<String, Type> columnTypes =
+          DorisTablePartitionOperations.getColumnTypes(connection, tableName, typeConverter);
+      String showPartitionsSql = String.format("SHOW PARTITIONS FROM `%s`", tableName);
+      List<Partition> assignments =
+          DorisTablePartitionOperations.loadPartitions(
+              connection, showPartitionsSql, partitioning, columnTypes);
```
diqiu50, line 648: Table loading will fail if loadPartition throws an exception, for example, for tables with multi-column range partitions.

Why do we load all partitions when loading a table? This can be expensive for tables with a large number of partitions.

yuqi1129:
Closing this PR based on the review discussion. As @diqiu50 pointed out, loading all partition assignments during `loadTable` is too expensive for tables with many partitions (e.g., Doris dynamic partitioning can produce thousands), and a partition parsing failure (e.g., multi-column range partitions) would break table loading entirely — a regression from the current behavior where assignments are simply empty.

We'll keep the current semantics: `loadTable` returns only the partition definition, and partition assignments should be fetched lazily via `supportPartitions().listPartitions()`. See #12093 for the full reasoning.

PR #12121, opened 2026-07-21
[#12051] fix(flink-connector): Move Paimon Hadoop options to Hadoop conf

### LINKED ISSUE #12051: [Bug report] Flink Paimon connector may expose sensitive catalog properties in logs
### Version

main branch

### Describe what's wrong

When using the Gravitino Flink connector with a Paimon catalog, some catalog properties are passed through to the underlying Paimon/Flink runtime configuration.

If the catalog contains storage-related credential properties, these properties may appear in Flink job logs because Paimon logs the full dynamic table options at INFO level.

Examples of sensitive property keys include:

- `gravitino.bypass.hadoop.fs.oss.accessKeyId`
- `gravitino.bypass.hadoop.fs.oss.accessKeySecret`
- `gravitino.bypass.hadoop.fs.bos.access.key`
- `gravitino.bypass.hadoop.fs.bos.secret.access.key`

This can expose sensitive configuration values in application logs.

A Flink job log contains an INFO message from Paimon's Flink table factory similar to:

```text
INFO org.apache.paimon.flink.AbstractFlinkTableFactory - Loading dynamic table options for <table_name> in table config: { ... fs.oss.accessKeyId=<masked>, fs.oss.accessKeySecret=<masked>, fs.bos.access.key=<masked>, fs.bos.secret.access.key=<masked>, ... }

### How to reproduce

1. Create a Gravitino Paimon catalog.
2. Configure the catalog with storage-related bypass properties that include sensitive credentials, such as OSS/BOS access keys.
3. Use the Gravitino Flink connector and enable the Gravitino catalog store.
4. Run a Flink job that reads or writes a Paimon table.
5. Check the Flink job logs.
The Paimon Flink table factory may print the full table configuration, including sensitive catalog-derived properties.

### Additional context

This log is emitted from the Flink job process, not from the Gravitino server process.
A short-term workaround is to raise the log level of the relevant Paimon Flink logger to WARN in the Flink job's log4j configuration.
Potential improvements:
Avoid exposing sensitive catalog properties through loggable runtime configuration where possible.
Mask sensitive property values before logging configuration maps.
Add masking for sensitive catalog property values in the Gravitino Web UI, especially for keys containing terms such as secret, password, token, credential, accessKey, or access.key.
Consider applying the same masking policy to catalog property API responses if backward compatibility allows it, or provide a safe display representation for UI usage.

### Possible solutions and trade-offs

There may be multiple layers where this can be improved:

1. Mask sensitive values in Paimon/Flink logs

   - Pros:
     - Addresses the direct log exposure shown in this issue.
     - Protects not only Gravitino-provided properties, but also any sensitive options passed to Paimon from other sources.
   - Cons:
     - This likely needs to be fixed in Paimon rather than Gravitino.
     - It may take longer if it requires coordination with the Paimon community.

2. Avoid passing sensitive properties through general-purpose Flink/Paimon option maps where possible

   - Pros:
     - Reduces the chance of sensitive values being exposed by downstream components.
     - Can improve the boundary between internal credentials and user-visible/runtime options.
   - Cons:
     - Some properties, especially Hadoop filesystem credentials, are required by Paimon/Hadoop at runtime.
     - It may not be possible to completely avoid passing them through configuration maps without larger design changes.

3. Mask sensitive catalog property values in the Gravitino Web UI

   - Pros:
     - Prevents accidental exposure when users view or edit catalog properties in the UI.
     - This is relatively self-contained and may be easier to implement in Gravitino.
   - Cons:
     - It does not solve log exposure in Flink/Paimon runtime.
     - The UI still needs a careful design for edit/reveal behavior.

4. Provide a safe display representation in Gravitino APIs

   - Pros:
     - Protects all API consumers, including the Web UI and third-party tools.
     - Establishes a consistent masking policy in one place.
   - Cons:
     - May affect backward compatibility if existing clients expect raw catalog property values.
     - Might require separate APIs or flags to distinguish "display" responses from operational metadata.

5. Document a short-term log4j workaround

   - Pros:
     - Easy for users to apply immediately.
     - Can stop this specific INFO log by setting `org.apache.paimon.flink.AbstractFlinkTableFactory` to `WARN`.
   - Cons:
     - This is only a workaround.
     - It does not prevent other components from logging sensitive configuration values.

### PR DESCRIPTION
**Pull request by jarredhj0214**

Move Paimon Flink connector Hadoop filesystem options (`hadoop.*`, `fs.*`, and `dfs.*`) from Paimon catalog options into Hadoop `Configuration` before creating the inner Paimon catalog.

This creates the inner Paimon catalog with a `CatalogContext` that carries the sanitized Paimon options plus Hadoop configuration, and keeps vended Paimon S3/OSS/JDBC credentials available as Paimon options.

Paimon can log dynamic/catalog options and expose sensitive filesystem credentials such as OSS/BOS access keys when they remain in Paimon options. Moving Hadoop filesystem credentials to Hadoop configuration follows the Flink Hive connector pattern and avoids passing these sensitive Hadoop properties through Paimon options.

Fix: #12051

No API changes. Existing Hadoop filesystem catalog properties continue to be accepted, but the Flink Paimon connector forwards them via Hadoop `Configuration` instead of Paimon catalog options.

Added unit coverage for:
- moving `hadoop.*`/`fs.*` filesystem options into Hadoop configuration
- keeping vended S3/OSS credentials in Paimon options for native Paimon FileIOs
- keeping JDBC credentials in Paimon options

Local checks:
- `git diff --check`

Could not run Gradle tests locally because only JDK 8 is installed and the build requires JDK 17.

### CODE REVIEW

File: flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java [hunk 1]
```diff
@@ -129,43 +134,75 @@ public void open() throws CatalogException {
       super.open();
       return;
     }
+    Map<String, String> paimonOptions = new HashMap<>(mutableOptions);
+    Configuration hadoopConf =
+        HadoopUtils.getHadoopConfiguration(Options.fromMap(withoutHadoopOptions(paimonOptions)));
+    moveHadoopOptionsToConf(paimonOptions, hadoopConf);
     try {
       CredentialPropertyUtils.applyPaimonCredentials(
-          CredentialPropertyUtils.getCredentials(catalog()), mutableOptions);
```
diqiu50, line 134: `CredentialPropertyUtils.applyPaimonCredentials` may put credentials to  mutableOptions. 

We need to confirm one thing: should these properties be removed from `paimonOptions`?

jarredhj0214: > `CredentialPropertyUtils.applyPaimonCredentials` may put credentials to mutableOptions.
> 
> We need to confirm one thing: should these properties be removed from `paimonOptions`?

Good point. These two sets of properties are intentionally handled differently.

The Hadoop filesystem options from catalog properties (`hadoop.*`, `fs.*`, and `dfs.*`) are removed from `paimonOptions` and moved into Hadoop `Configuration`.

The credentials injected by `CredentialPropertyUtils.applyPaimonCredentials` are vended credentials, and they need to stay in `paimonOptions` because Paimon native S3/OSS FileIO loaders declare these keys as required options. If we remove them from `paimonOptions`, native Paimon S3/OSS FileIO may fail to initialize.

I added a comment to make this distinction explicit.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
```diff
@@ -139,6 +139,8 @@ public void open() throws CatalogException {
         HadoopUtils.getHadoopConfiguration(Options.fromMap(withoutHadoopOptions(paimonOptions)));
     moveHadoopOptionsToConf(paimonOptions, hadoopConf);
     try {
+      // Keep vended credentials as Paimon options because Paimon native S3/OSS FileIO loaders
+      // declare them as required options.
```
diqiu50: If the credentials are added back here, then this issue isn't actually resolved, right?

jarredhj0214: > If the credentials are added back here, then this issue isn't actually resolved, right?

You're right. I reworked this so storage credentials are not added back to paimonOptions.

Now Hadoop-style options are moved into Hadoop Configuration, Paimon native S3 options are translated to Hadoop S3A configuration and removed from paimonOptions, and vended S3/OSS credentials are also written to Hadoop Configuration. Only JDBC backend credentials remain in paimonOptions.

I also added/updated tests to verify that storage credentials are absent from paimonOptions and still available from Hadoop Configuration. This should address the logging leak without dropping the credentials from the filesystem access path.

lasdf1234:
1、Would these three methods(toHadoopConfKey、withoutHadoopOptions、moveHadoopOptionsToConf) be more suitable to be placed in PropertyUtils?
2、withoutHadoopOptions + moveHadoopOptionsToConf may be can mrege into one method better: 
static Configuration extractHadoopConfiguration(Map<String, String> options) {
  Map<String, String> hadoopProps = new HashMap<>();
  options.entrySet().removeIf(e -> {
    String hadoopKey = toHadoopConfKey(e.getKey()); // private
    if (hadoopKey == null) return false;
    hadoopProps.put(hadoopKey, e.getValue());
    return true;
  });
  Configuration conf = HadoopUtils.getHadoopConfiguration(Options.fromMap(options));
  hadoopProps.forEach(conf::set);
  return conf;
}

Make toHadoopConfKey become a private method in PropertyUtils.

3、May be method applyCredentialsToCatalog move into CredentialPropertyUtils better?
@jarredhj0214

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
```diff
@@ -129,43 +143,124 @@ public void open() throws CatalogException {
...
+   * @param hadoopConf Hadoop configuration carrying filesystem credentials and Hadoop options.
+   * @return the created inner Paimon catalog.
+   */
+  @VisibleForTesting
+  protected AbstractCatalog createInnerCatalog(
+      Map<String, String> paimonOptions, Configuration hadoopConf) {
+    CatalogContext catalogContext =
+        CatalogContext.create(
+            Options.fromMap(paimonOptions), hadoopConf, new FlinkFileIOLoader(), null);
+    return (AbstractCatalog)
```
diqiu50, line 174: `new FlinkFileIOLoader()` should last parameter in the function

jarredhj0214:
> 1、Would these three methods(toHadoopConfKey、withoutHadoopOptions、moveHadoopOptionsToConf) be more suitable to be placed in PropertyUtils? 2、withoutHadoopOptions + moveHadoopOptionsToConf may be can mrege into one method better: static Configuration extractHadoopConfiguration(Map<String, String> options) { Map<String, String> hadoopProps = new HashMap<>(); options.entrySet().removeIf(e -> { String hadoopKey = toHadoopConfKey(e.getKey()); // private if (hadoopKey == null) return false; hadoopProps.put(hadoopKey, e.getValue()); return true; }); Configuration conf = HadoopUtils.getHadoopConfiguration(Options.fromMap(options)); hadoopProps.forEach(conf::set); return conf; }
> 
> Make toHadoopConfKey become a private method in PropertyUtils.
> 
> 3、May be method applyCredentialsToCatalog move into CredentialPropertyUtils better? @jarredhj0214

Thanks for the suggestions. I moved the Hadoop option extraction logic into PropertyUtils and reused CredentialPropertyUtils.applyPaimonCredentials to simplify the catalog code.

flink-connector/flink-common/src/main/java/org/apache/gravitino/flink/connector/paimon/GravitinoPaimonCatalog.java
jarredhj0214: > `new FlinkFileIOLoader()` should last parameter in the function

Good catch. I verified the CatalogContext signature in Paimon 1.2.0 and 1.4.1, and moved FlinkFileIOLoader to the fallback loader parameter.

diqiu50 (approved):
LGTM

PR #12124, opened 2026-07-21
[#12123] fix(catalog): shut down MySQL AbandonedConnectionCleanupThread and broaden ThreadLocal cleanup to prevent Metaspace leak

### LINKED ISSUE #12123: [Bug report] Dropping a MySQL-backed catalog leaks its ClassLoader via AbandonedConnectionCleanupThread and non-webserver ThreadLocals
### Version

main branch

### Describe what's wrong

When a catalog that used the MySQL Connector/J driver is dropped or reloaded, its isolated ClassLoader is not garbage collected, so Metaspace grows on every reload and eventually hits `OutOfMemoryError: Metaspace`.

Two references keep the ClassLoader alive after cleanup runs:

1. MySQL's `AbandonedConnectionCleanupThread` (`mysql-cj-abandoned-connection-cleanup`) holds the driver's ClassLoader as its context ClassLoader. A plain thread interrupt does not release it; only `uncheckedShutdown()` clears the tracked-connection map so the thread and ClassLoader can be collected.
2. `ClassLoaderResourceCleanerUtils.clearThreadLocalMap` only sweeps `Gravitino-webserver-*` threads. ThreadLocals that reference the dropped ClassLoader can also sit on other threads (Caffeine ForkJoinPool workers, catalog-cleaner, Hadoop daemons), and those are never cleared.

`closeClassLoaderResource` is the shared routine that releases a catalog ClassLoader's resources, but it releases neither of the above references, so a MySQL-backed catalog leaks its ClassLoader whenever cleanup runs.

`java.lang.OutOfMemoryError: Metaspace` after repeatedly creating and dropping MySQL-backed catalogs (Metaspace committed size grows monotonically with reload count).

### How to reproduce

1. Gravitino version: main branch
2. Create a catalog that loads the MySQL driver (a `jdbc-mysql` catalog, or an Iceberg JDBC catalog on MySQL).
3. Drop the catalog, then repeat create/drop in a loop.
4. Observe with `jcmd <pid> VM.metaspace` (or a heap dump) that the isolated ClassLoader and the `mysql-cj-abandoned-connection-cleanup` thread survive each drop; Metaspace committed size keeps climbing.

### Additional context

The fix is scoped to `ClassLoaderResourceCleanerUtils`: shut down the MySQL `AbandonedConnectionCleanupThread` (guarded so it only acts when the target ClassLoader owns the driver class), and broaden the ThreadLocal sweep beyond webserver threads.

LuciferYang:

Two additions to `ClassLoaderResourceCleanerUtils.closeClassLoaderResource`:

- Shut down MySQL's `AbandonedConnectionCleanupThread` via `uncheckedShutdown()`, which also clears the tracked-connection map so the driver's ClassLoader can be collected. The call is guarded by `isOwnedByClassLoader`, matching the other cleanup steps in this class, so it never stops a cleanup thread owned by a parent or shared ClassLoader (for example a MySQL entity store loaded on the app ClassLoader, whose driver class delegates up to the app loader).
- Broaden the ThreadLocal sweep from `Gravitino-webserver-*` threads to all application threads, skipping only the JVM `system` thread group. Only entries whose value was loaded by the target ClassLoader are cleared, so ThreadLocals owned by other ClassLoaders are left alone.

After a MySQL-backed catalog is dropped, its isolated ClassLoader stays reachable through the `mysql-cj-abandoned-connection-cleanup` thread's context ClassLoader and through ThreadLocals on non-webserver threads (Caffeine ForkJoinPool, catalog-cleaner, Hadoop daemons). `closeClassLoaderResource` releases neither today, so the ClassLoader leaks on every reload and Metaspace grows until `OutOfMemoryError: Metaspace`.

Fix: #12123

No.

Added `TestClassLoaderResourceCleanerUtils` cases (`mysql-connector-j` added as a test-only dependency):

- `testShutdownMySQLAbandonedConnectionCleanupThreadStopsThread`: loading `AbandonedConnectionCleanupThread` in a child `URLClassLoader` starts the cleanup thread bound to that loader; after `shutdownMySQLAbandonedConnectionCleanupThread` the thread is gone. It loads the cleanup-thread class directly rather than the driver, so the test does not register a driver into the JVM-global `DriverManager`.
- `testShutdownMySQLAbandonedConnectionCleanupThreadSkipsParentOwnedClass`: when the class resolves to a parent loader, a child-scoped cleanup leaves the parent's thread alive; an owner-scoped cleanup stops it.
- `testClearThreadLocalMapClearsOnlyTargetClassLoaderValues`: on a non-webserver thread, a ThreadLocal whose value is loaded by the target ClassLoader is cleared, while one loaded by a different (non-null) ClassLoader is kept.
- `testClearThreadLocalMapSkipsSystemThreadGroup`: a ThreadLocal on a `system`-group thread is left untouched.

Ran `./gradlew :catalogs:catalog-common:test --tests "org.apache.gravitino.utils.TestClassLoaderResourceCleanerUtils"`: 7/7 green.

### DIFF
--- a/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
+++ b/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
@@ -31,9 +31,8 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Utility class to clean up resources related to a specific class loader to prevent memory leaks.
- * Gravitino will create a new class loader for each catalog and release it when there exist any
- * changes to the catalog. So, it's important to clean up resources related to the class loader to
- * prevent memory leaks.
+ * Gravitino uses isolated class loaders for catalog implementations, and these class loaders must
+ * be properly cleaned up when no longer needed to prevent Metaspace leaks.
  */
 public class ClassLoaderResourceCleanerUtils {
 
@@ -42,7 +41,13 @@ public class ClassLoaderResourceCleanerUtils {
   private ClassLoaderResourceCleanerUtils() {}
 
   /**
-   * Close all resources related to the given class loader to prevent memory leaks.
+   * Close all resources related to the given class loader to prevent memory leaks. This includes
+   * stopping threads, clearing ThreadLocals, shutting down MySQL's
+   * AbandonedConnectionCleanupThread, and releasing various logging and cloud SDK resources.
+   *
+   * <p><b>Warning:</b> This method performs destructive cleanup that affects all code sharing the
+   * ClassLoader. It must only be called when the ClassLoader is about to be destroyed and no
+   * catalog is still using it, i.e. from a catalog's {@code close()} path.
    *
    * @param classLoader the classloader to be closed
    */
@@ -57,12 +62,9 @@ public class ClassLoaderResourceCleanerUtils {
     executeAndCatch(
         ClassLoaderResourceCleanerUtils::closeStatsDataClearerInFileSystem, classLoader);
 
-    // Stop all threads with the current class loader and clear their threadLocal variables for
-    // jetty threads that are loaded by the current class loader.
-    // For example, thread local `threadData` in FileSystem#StatisticsDataCleaner is created
-    // within jetty thread with the current class loader. However, there are clear by
-    // `catalog.close` in ForkJoinPool in CaffeineCache, in this case, the thread local variable
-    // will not be cleared, so we need to clear them manually here.
+    // Stop threads that run under the target class loader and clear ThreadLocals that reference it.
+    // Such ThreadLocals can sit on any thread (Jetty webserver, Caffeine ForkJoinPool,
+    // catalog-cleaner) and keep the ClassLoader from being collected.
     executeAndCatch(
         ClassLoaderResourceCleanerUtils::stopThreadsAndClearThreadLocalVariables, classLoader);
 
@@ -77,6 +79,10 @@ public class ClassLoaderResourceCleanerUtils {
     executeAndCatch(ClassLoaderResourceCleanerUtils::closeResourceInAzure, classLoader);
 
     executeAndCatch(ClassLoaderResourceCleanerUtils::clearShutdownHooks, classLoader);
+
+    executeAndCatch(
+        ClassLoaderResourceCleanerUtils::shutdownMySQLAbandonedConnectionCleanupThread,
+        classLoader);
   }
 
   /**
@@ -146,7 +152,7 @@ public class ClassLoaderResourceCleanerUtils {
     for (Thread thread : threads) {
       // First clear thread local variables
       clearThreadLocalMap(thread, classLoader);
-      // Close all threads that are using the FilesetCatalogOperations class loader
+      // Stop all threads that are using the target class loader
       if (runningWithClassLoader(thread, classLoader)) {
         LOG.debug("Interrupting thread: {}", thread.getName());
         thread.setContextClassLoader(null);
@@ -178,8 +184,25 @@ public class ClassLoaderResourceCleanerUtils {
     return threads;
   }
 
-  private static void clearThreadLocalMap(Thread thread, ClassLoader targetClassLoader) {
-    if (thread == null || !thread.getName().startsWith("Gravitino-webserver-")) {
+  @VisibleForTesting
+  static void clearThreadLocalMap(Thread thread, ClassLoader targetClassLoader) {
+    if (thread == null) {
+      return;
+    }
+
+    // Sweep every application thread, not only the Gravitino-webserver-* ones: a ThreadLocal
+    // pointing at the target ClassLoader can live on any thread, such as a Caffeine ForkJoinPool
+    // worker, a catalog-cleaner thread, or a Hadoop daemon. The check below clears only entries
+    // whose value was loaded by the target ClassLoader, so ThreadLocals owned by other
+    // ClassLoaders are left alone.
+    //
+    // Skip threads whose immediate group is the JVM "system" group (Reference Handler, Finalizer,
+    // Signal Dispatcher, and the like): they hold no catalog ClassLoader references, and reflecting
+    // into their ThreadLocals is best avoided. This checks the immediate group only, so threads in
+    // a sub-group of system (such as InnocuousThreadGroup for common-pool workers) are still swept.
+    // That is intended: ForkJoinPool threads can hold catalog ThreadLocals.
+    ThreadGroup group = thread.getThreadGroup();
+    if (group != null && "system".equals(group.getName())) {
       return;
     }
 
@@ -256,7 +279,7 @@ public class ClassLoaderResourceCleanerUtils {
       LOG.debug("HTrace is not used, skipping release of HTrace LogFactory...");
     }
 
-    // Release the LogFactory for the FilesetCatalogOperations class loader
+    // Release the LogFactory for the target class loader
     Class<?> logFactoryClass =
         Class.forName("org.apache.commons.logging.LogFactory", true, currentClassLoader);
     MethodUtils.invokeStaticMethod(logFactoryClass, "release", currentClassLoader);
@@ -353,6 +376,36 @@ public class ClassLoaderResourceCleanerUtils {
   @VisibleForTesting
   static boolean isOwnedByClassLoader(Class<?> clazz, ClassLoader classLoader) {
     return classLoader != null && clazz.getClassLoader() == classLoader;
+  }
+
+  /**
+   * Shutdown MySQL's AbandonedConnectionCleanupThread to prevent it from holding a reference to the
+   * ClassLoader. Unlike simple thread interruption, {@code uncheckedShutdown()} cleans up the
+   * tracked connections map, releasing references to classes loaded by the target ClassLoader.
+   *
+   * @param classLoader the classloader where the MySQL driver is loaded
+   */
+  @VisibleForTesting
+  static void shutdownMySQLAbandonedConnectionCleanupThread(ClassLoader classLoader)
+      throws Exception {
+    // Load with initialize=false: when this class is owned by the target ClassLoader the MySQL
+    // driver has already been used, so the class is initialized and its cleanup thread started.
+    Class<?> clazz =
+        Class.forName("com.mysql.cj.jdbc.AbandonedConnectionCleanupThread", false, classLoader);
+    // uncheckedShutdown() is a JVM-global static that stops the single cleanup thread tied to this
+    // driver class. If the driver was resolved from a parent or shared ClassLoader (for example a
+    // MySQL entity store loaded on the app ClassLoader), stopping it here would break
+    // abandoned-connection cleanup for the whole JVM. Only act when the target ClassLoader owns the
+    // class, the same guard the other cleanup steps in this class use.
+    if (!isOwnedByClassLoader(clazz, classLoader)) {
+      LOG.debug(
+          "AbandonedConnectionCleanupThread is owned by {}, not the target classloader {}; skipping shutdown",
+          clazz.getClassLoader(),
+          classLoader);
+      return;
+    }
+    MethodUtils.invokeStaticMethod(clazz, "uncheckedShutdown");
+    LOG.info("AbandonedConnectionCleanupThread has been shutdown.");
   }
 
   @FunctionalInterface
--- a/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
+++ b/catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
@@ -33,6 +33,13 @@ import org.slf4j.LoggerFactory;
  * Utility class to clean up resources related to a specific class loader to prevent memory leaks.
  * Gravitino uses isolated class loaders for catalog implementations, and these class loaders must
  * be properly cleaned up when no longer needed to prevent Metaspace leaks.
+ *
+ * <p>Several steps here are per-vendor special cases (AWS, GCP, Azure, commons-logging, MySQL).
+ * Three things hold for all of them, and for any added later: guard with {@link
+ * #isOwnedByClassLoader} so a library resolved from a parent or shared ClassLoader is left alone;
+ * prefer the library's own shutdown API over interrupting its thread; and run the step before
+ * {@link #stopThreadsAndClearThreadLocalVariables}, which interrupts any thread whose context
+ * ClassLoader is the target and would otherwise get there first.
  */
 public class ClassLoaderResourceCleanerUtils {
 
@@ -58,9 +65,30 @@ public class ClassLoaderResourceCleanerUtils {
       return;
     }
 
+    runCleanupSteps(classLoader);
+  }
+
+  /**
+   * The cleanup sequence itself, without the test-environment guard above. Split out so a test can
+   * drive the whole sequence: unit tests run with {@code GRAVITINO_TEST} set, which would otherwise
+   * make {@link #closeClassLoaderResource} return before doing anything. The order of the steps
+   * matters, see the class javadoc.
+   *
+   * @param classLoader the classloader to be closed
+   */
+  @VisibleForTesting
+  static void runCleanupSteps(ClassLoader classLoader) {
     // Clear statics threads in FileSystem and close all FileSystem instances.
     executeAndCatch(
         ClassLoaderResourceCleanerUtils::closeStatsDataClearerInFileSystem, classLoader);
+
+    // Ahead of the generic thread sweep below: the MySQL driver sets its cleanup thread's context
+    // ClassLoader to the loader that defined the driver, so the sweep would match that thread and
+    // interrupt it. Interrupting it lands in referenceQueue.remove() rather than the driver's own
+    // shutdown path, so let the driver stop the thread itself first.
+    executeAndCatch(
+        ClassLoaderResourceCleanerUtils::shutdownMySQLAbandonedConnectionCleanupThread,
+        classLoader);
 
     // Stop threads that run under the target class loader and clear ThreadLocals that reference it.
     // Such ThreadLocals can sit on any thread (Jetty webserver, Caffeine ForkJoinPool,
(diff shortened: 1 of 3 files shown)


catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
```diff
@@ -178,8 +184,25 @@ private static Thread[] getAllThreads() {
...
+    // whose value was loaded by the target ClassLoader, so ThreadLocals owned by other
+    // ClassLoaders are left alone.
+    //
+    // Skip threads whose immediate group is the JVM "system" group (Reference Handler, Finalizer,
+    // Signal Dispatcher, and the like): they hold no catalog ClassLoader references, and reflecting
+    // into their ThreadLocals is best avoided. This checks the immediate group only, so threads in
+    // a sub-group of system (such as InnocuousThreadGroup for common-pool workers) are still swept.
+    // That is intended: ForkJoinPool threads can hold catalog ThreadLocals.
+    ThreadGroup group = thread.getThreadGroup();
+    if (group != null && "system".equals(group.getName())) {
```
yuqi1129, line 205: Have you tested and verified that it works or not? I'm not sure whether removing logic `Gravitino-webserver` and replacing it with the thread `group` can function well when the catalog is closed.

LuciferYang: Good question, I verified it with a test rather than reasoning about it.

Added `testClearThreadLocalMapLetsDroppedClassLoaderBeCollected`: a long-lived, non-`Gravitino-webserver-*` worker thread holds a `ThreadLocal` whose value is loaded by the target ClassLoader, which is what pins a dropped catalog's ClassLoader in production. After `clearThreadLocalMap` runs and the test drops its own references, a `WeakReference` to the ClassLoader clears once GC runs. Under the old `Gravitino-webserver-` name filter this exact test fails (the thread is skipped, the ClassLoader stays reachable), so it both proves the fix and guards against a regression.

Two things worth clarifying about the scope of the change:

- It only changes *which threads' ThreadLocals are swept*, not *which threads are stopped*. The interrupt path in `stopThreadsAndClearThreadLocalVariables` is still guarded by `runningWithClassLoader` (contextClassLoader == target), unchanged. And an entry is only cleared when `value.getClass().getClassLoader() == targetClassLoader`, so ThreadLocals owned by other ClassLoaders are never touched. The worst case of the wider scope is visiting a thread and skipping it.
- The reflective `ThreadLocalMap` access needs `--add-opens java.base/java.lang=ALL-UNNAMED`, which `bin/gravitino.sh.template` already sets, so it works at runtime.

`./gradlew :catalogs:catalog-common:test --tests "*TestClassLoaderResourceCleanerUtils"`: 8/8 green.

yuqi1129:
I need to take time to verify the effect carefully, so this one will not be merged in the short term.

LuciferYang:
No problem, take your time.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
```diff
@@ -355,6 +378,36 @@ static boolean isOwnedByClassLoader(Class<?> clazz, ClassLoader classLoader) {
...
+    // abandoned-connection cleanup for the whole JVM. Only act when the target ClassLoader owns the
+    // class, the same guard the other cleanup steps in this class use.
+    if (!isOwnedByClassLoader(clazz, classLoader)) {
+      LOG.debug(
+          "AbandonedConnectionCleanupThread is owned by {}, not the target classloader {}; skipping shutdown",
+          clazz.getClassLoader(),
+          classLoader);
+      return;
+    }
+    MethodUtils.invokeStaticMethod(clazz, "uncheckedShutdown");
```
yuqi1129, line 407: I verified the mechanism this is built on, and it holds. Notes for whoever reviews next.

`mysql-connector-j` 8.0.33 (the version pinned in `libs.versions.toml`) creates the cleanup thread through a factory that explicitly pins the loader:

```
16: ldc           class com/mysql/cj/jdbc/AbandonedConnectionCleanupThread
18: invokevirtual Class.getClassLoader()
32: invokevirtual Thread.setContextClassLoader(ClassLoader)
37: putstatic     Field threadRef:Ljava/lang/Thread;
```

So a live cleanup thread pins whichever loader defined the driver class, and the static `threadRef` keeps the thread reachable. Note it is `implements Runnable`, not a `Thread` subclass, so the pin is through the Runnable and the TCCL rather than the thread's own class.

I also ran an independent reclaim probe with a control group:

```
[CONTROL]  child loader that never touches MySQL   -> collected: true
[NO FIX]   load the driver class, drop the loader  -> collected: false
[WITH FIX] same, then uncheckedShutdown()          -> collected: true
```

The control matters: without it, `collected: false` could just mean the WeakReference test was written badly. With it, the leak and the fix are both real.

The `isOwnedByClassLoader` guard here is also correct, and I want to say that explicitly so nobody "fixes" it later: two isolated loaders each holding their own driver copy have separate classes and separate statics, so there is no cross-talk; a shared or parent-loaded driver is skipped; and `ClassLoaderPool` only calls this at refcount zero.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
```diff
@@ -77,6 +79,10 @@ public static void closeClassLoaderResource(ClassLoader classLoader) {
     executeAndCatch(ClassLoaderResourceCleanerUtils::closeResourceInAzure, classLoader);
 
     executeAndCatch(ClassLoaderResourceCleanerUtils::clearShutdownHooks, classLoader);
+
+    executeAndCatch(
+        ClassLoaderResourceCleanerUtils::shutdownMySQLAbandonedConnectionCleanupThread,
```
yuqi1129, line 84: **Ordering looks wrong.** `runningWithClassLoader` matches on `thread.getContextClassLoader() == targetClassLoader`, and the bytecode above shows the MySQL cleanup thread's TCCL is set to exactly the loader that defined the driver, i.e. the target loader.

So by the time this new step runs, `stopThreadsAndClearThreadLocalVariables` (six steps earlier) has already matched that thread and done `setContextClassLoader(null)` + `interrupt()` on it. MySQL's `run()` blocks in `referenceQueue.remove()`, so it gets an out-of-band `InterruptedException` instead of the driver's own shutdown path, and whether `connectionFinalizerPhantomRefs` is left populated at that point is not something the driver promises.

The end state is probably still correct, because `uncheckedShutdown()` clears the map afterwards, but that is the second step cleaning up after the first, which is fragile.

Worth being explicit about the limits of my verification: my probe called `shutdownMySQLAbandonedConnectionCleanupThread` **directly**, not through `closeClassLoaderResource`, so I confirmed the method works but did **not** exercise this ordering.

Minimal fix: move this `executeAndCatch` above the `stopThreadsAndClearThreadLocalVariables` call. The driver then exits through its own API and the generic sweep finds no thread to interrupt.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
yuqi1129, line 205: This widens a genuinely unsafe operation. `clearThreadLocalMap` reflects into another thread's `ThreadLocalMap.table` and writes `entry.value = null` from a foreign thread with no synchronisation. Until now that was limited to `Gravitino-webserver-*`, which are mostly idle between requests. Skipping only the immediate `system` group means we now also sweep `InnocuousThreadGroup` (ForkJoinPool common-pool workers) and Hadoop daemons, and the comment says that is deliberate — those are threads actively running work.

Structural corruption is not the risk (rehash moves `Entry` references, so writing `value` still lands on the right object). The risk is nulling a value another thread is using right now, producing an NPE in unrelated code at catalog-drop time, which is close to impossible to attribute afterwards.

One thing that makes me more comfortable with the change than I expected: I checked the blame, and the `Gravitino-webserver-` restriction has been there since the class was introduced in #8252, so this is extending an incomplete fix rather than reversing a deliberate safety decision. Worth saying so in the PR description.

Suggestion: keep the widening but bound it — an allowlist of the thread names actually implicated (`Gravitino-webserver-*`, `ForkJoinPool*`, `catalog-cleaner*`, Hadoop's statistics cleaner) reads better than "everything except the system group", and it documents what we are actually chasing. Skipping threads currently in `RUNNABLE` would be another way.

catalogs/catalog-common/src/main/java/org/apache/gravitino/utils/ClassLoaderResourceCleanerUtils.java
```diff
@@ -355,6 +378,36 @@ static boolean isOwnedByClassLoader(Class<?> clazz, ClassLoader classLoader) {
...
 
+  /**
+   * Shutdown MySQL's AbandonedConnectionCleanupThread to prevent it from holding a reference to the
+   * ClassLoader. Unlike simple thread interruption, {@code uncheckedShutdown()} cleans up the
+   * tracked connections map, releasing references to classes loaded by the target ClassLoader.
+   *
+   * @param classLoader the classloader where the MySQL driver is loaded
+   */
+  @VisibleForTesting
+  static void shutdownMySQLAbandonedConnectionCleanupThread(ClassLoader classLoader)
```
yuqi1129, line 389: Minor, maintainability. This class is accumulating vendor special cases (AWS, GCP, Azure, commons-logging, now MySQL), and the same shape exists for other JDBC drivers we ship catalogs for.

No need to abstract anything now, but a line in the class javadoc recording the pattern would help the next person: guard with `isOwnedByClassLoader`, use the driver's own shutdown API rather than an interrupt, and run it **before** the generic thread sweep.

catalogs/catalog-common/src/test/java/org/apache/gravitino/utils/TestClassLoaderResourceCleanerUtils.java
```diff
@@ -68,4 +78,288 @@ void testIsOwnedByClassLoaderReturnsFalseForBootstrapLoadedClass() {
...
+   * End-to-end proof that broadening the sweep lets a dropped catalog's ClassLoader be collected. A
+   * long-lived worker thread that is not a {@code Gravitino-webserver-*} thread holds a ThreadLocal
+   * whose value was loaded by the target ClassLoader, which is exactly what pins the ClassLoader
+   * after a catalog is dropped. After {@code clearThreadLocalMap} runs and the test drops its own
+   * references, a WeakReference to the ClassLoader must clear once GC runs. The webserver-only
+   * filter used before would skip this thread and the ClassLoader would stay reachable.
+   */
+  @Test
+  @SuppressWarnings("ThreadLocalUsage") // a per-thread ThreadLocal is the leak this test reproduces
+  void testClearThreadLocalMapLetsDroppedClassLoaderBeCollected() throws Exception {
```
yuqi1129, line 240: This is a good test and it covers the ThreadLocal half of the issue end to end.

The MySQL half has no equivalent: `testShutdownMySQLAbandonedConnectionCleanupThreadStopsThread` asserts the thread is gone, but not that the ClassLoader then becomes collectable, which is the property the issue is actually about. I wrote that test as a throwaway while verifying this PR and it passes, so it is worth having here:

```
[CONTROL]  child loader that never touches MySQL   -> collected: true
[NO FIX]   load the driver class, drop the loader  -> collected: false
[WITH FIX] same, then uncheckedShutdown()          -> collected: true
```

The control case is the part worth copying — a bare `WeakReference` + `System.gc()` assertion is unreliable enough that without it a green test does not mean much.

catalogs/catalog-common/src/test/java/org/apache/gravitino/utils/TestClassLoaderResourceCleanerUtils.java
```diff
@@ -20,17 +20,27 @@
...
 
+import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 
 class TestClassLoaderResourceCleanerUtils {
```
yuqi1129, line 36: Test-structure point: every case here calls `shutdownMySQLAbandonedConnectionCleanupThread` or `clearThreadLocalMap` directly, so nothing exercises `closeClassLoaderResource` itself. That is why the step-ordering problem I flagged on the main file is invisible to this suite.

One integration case through `closeClassLoaderResource` would cover it. Note it early-returns when the `GRAVITINO_TEST` environment variable is set, so such a test has to clear that first.

yuqi1129 (commented):
I said earlier I wanted to verify the effect before merging this, so here is the result. **The leak is real**, on three independent levels, and the fix does release the ClassLoader.

1. Code: `clearThreadLocalMap` still returns early for anything that is not `Gravitino-webserver-*`, and `closeClassLoaderResource` has no MySQL handling. Both claims in #12123 match the code.
2. Bytecode: in `mysql-connector-j` 8.0.33 the cleanup thread's factory explicitly calls `AbandonedConnectionCleanupThread.class.getClassLoader()` and `Thread.setContextClassLoader(...)`, and keeps the thread in a static `threadRef`. Details on the line comment.
3. Measured: an independent reclaim probe, with a control group, shows the loader is not collected without the fix and is collected with it.

One thing the description should say, because it decides whether this fix does anything for a given deployment: `IsolatedClassLoader.isSharedClass()` treats `com.mysql.*` as shared and delegates to the app loader first. If an operator has the driver on the server classpath (needed for a MySQL entity store) the driver is app-loaded, there is no isolated-loader leak, and this code correctly skips. We do not ship the driver ourselves (`libs.mysql.driver` is `testImplementation` only, and `jdbc-mysql.conf` tells users to supply it), so the documented layout puts it in `catalogs/jdbc-mysql/libs/` only, which is the case that leaks. The guard covers both, which is right.

Two things to fix before merge, both small:

- The new MySQL shutdown runs **after** the generic thread sweep, which will already have interrupted that same thread, because the sweep matches on context ClassLoader and MySQL sets it to the driver's loader. Move it earlier.
- Decide how far the ThreadLocal sweep should reach. Widening it is defensible (I checked, the webserver-only limit was the original scope in #8252, not a deliberate narrowing after an incident), but "every thread outside the system group" now includes actively-running ForkJoinPool workers, and we are writing into their ThreadLocalMap from another thread.

Tests: the ThreadLocal half has a proper reclaim test, the MySQL half does not, and nothing in the suite goes through `closeClassLoaderResource`, which is exactly why the ordering issue is invisible today.

Nice work on the `isOwnedByClassLoader` guard, by the way — I went looking for a hole in the JVM-global `uncheckedShutdown()` and there isn't one.

LuciferYang:
@yuqi1129 Thanks for verifying it rather than taking the issue's word for it. The ordering call was right, both fixes are in, and there is one suggestion I decided against.

**Ordering.** The MySQL shutdown now runs above `stopThreadsAndClearThreadLocalVariables`, with the reason inline: the driver's factory sets the cleanup thread's context ClassLoader to the loader that defined the driver, which is exactly what the generic sweep matches on. The driver now leaves through its own API and the sweep finds no thread to interrupt.

**Sweep scope: kept broad, allowlist declined.** What bounds this sweep is not the thread set, it is the value check. An entry is cleared only when `value.getClass().getClassLoader() == targetClassLoader`, so visiting a ForkJoinPool worker or a Hadoop daemon touches nothing unless that thread is holding a value loaded by the catalog being dropped. The allowlist would put the thread set back in charge of correctness, and every name nobody thought of becomes a silent miss with the loader still pinned and no signal anywhere. That is the failure mode #8252 shipped, and it took an issue to find. A `RUNNABLE` check has a different problem: the state can change between the check and the write, so it buys the appearance of safety rather than safety.

The residual risk you named is real and I am not claiming otherwise. A value belonging to the departing catalog can be nulled while a thread is still using it. That work is against a catalog being dropped and fails either way, and the value check means no other ClassLoader's entries are ever in reach, but the window exists. The comment on `clearThreadLocalMap` now says that instead of implying the sweep is safe because of who it skips.

**Class javadoc** records the vendor pattern for whoever adds the next one: guard with `isOwnedByClassLoader`, prefer the library's own shutdown API over interrupting its thread, and run the step before the generic sweep.

**Tests.** Two, both mutation checked.

The first is the reclaim test for the MySQL half, with your control group, because you are right that a bare `WeakReference` plus `System.gc()` proves little on its own. It asserts the control loader is collectable, then loads the driver in a child loader, drops every strong reference and asserts the loader is *not* collected, recovers it through `ref.get()`, runs the shutdown, and asserts it becomes collectable. Deleting the shutdown call fails it.

The second drives the step sequence, which nothing did before. `closeClassLoaderResource` returns immediately when `GRAVITINO_TEST` is set and the root build sets that on every test task, so clearing it in-process would mean reflecting into `ProcessEnvironment` and mutating state shared with every other test in the JVM. Instead the step list is now a package-private `runCleanupSteps` and the test drives that. Removing the MySQL step from the sequence fails it.

One thing I will not overclaim: that second test does not catch the ordering bug. Under the old order the sweep interrupts the thread and `uncheckedShutdown()` still clears the map afterwards, so the end state matches, exactly as you expected. What it buys is that the sequence is now exercised at all, so a future reordering that does change the outcome fails instead of passing quietly. The ordering fix rests on the mechanism you traced in the bytecode, not on a failing test.

**Two points for the description**, both going in.

`com.mysql.*` is a shared prefix: `isSharedClass` returns true for it, so `IsolatedClassLoader` asks `baseClassLoader` first and only falls through to `super.loadClass` when the app loader misses. A driver on the server classpath is therefore app-loaded, nothing is pinned to the isolated loader, and this step correctly skips through the `isOwnedByClassLoader` guard. The documented layout, driver in `catalogs/jdbc-mysql/libs/` only, is the one that takes the fall-through and leaks. So whether this fix does anything depends on where the operator put the jar, and the guard is what makes both layouts safe.

The `Gravitino-webserver-` restriction arrived with the class in #8252, so this widens a fix that was incomplete from the start rather than reversing a decision someone made after an incident. Worth having in the description so the next reader does not have to run blame to find that out.

LuciferYang:
Closing this as superseded. `main` now independently implements both changes this PR proposed: `ClassLoaderResourceCleanerUtils` shuts down MySQL's `AbandonedConnectionCleanupThread` (via `shutdownMysqlConnectionCleanup`) and broadens the ThreadLocal cleanup in `clearThreadLocalMap`. After merging the latest `main`, this branch had no remaining source change (only an unused test dependency), so there is nothing left for it to contribute. Thanks to whoever landed the equivalent fix.

PR #12128, opened 2026-07-21
[#12099] feat(mcp-server): add read tools for table partitions

### LINKED ISSUE #12099: [Subtask] feat(mcp-server): read tools for table partitions (list/get)
### Describe the subtask

Follow-up to #11574 / #11693. The MCP server already exposes
`list_statistics_for_partition`, but there is **no tool to discover partition
names/infos**, so an agent has no way to obtain the `from`/`to` partition inputs
that tool requires. Add read-only partition tools following the established
pattern (interface → plain REST client impl → FastMCP tool → mock → unit tests),
with authorization enforced by Gravitino.

Proposed tools:
- `list_of_partitions(catalog, schema, table, details=false)` — `GET .../tables/{table}/partitions` (names, or full infos when `details=true`).
- `get_partition(catalog, schema, table, partition)` — `GET .../tables/{table}/partitions/{partition}`.

Scope: **read-only** (list/get). Partition add/drop (write) can be a later subtask.

The backend REST already exists (`PartitionOperations`, both GETs gated by
`LOAD_TABLE_AUTHORIZATION_EXPRESSION`), so authorization-scoped discovery is
inherited automatically — no filtering logic added in the MCP layer.

### Parent issue
#11573

bharos:

Add two read-only MCP tools for table partitions:

- `list_of_partitions(catalog, schema, table, details=false)` — returns partition
  names by default, or full partition metadata when `details=true`.
- `get_partition(catalog, schema, table, partition)` — returns the metadata of a
  single partition.

They follow the existing metadata-type pattern (interface → plain REST client →
FastMCP tool → factory wiring → mock + unit tests). All user-supplied path
segments are URL-encoded, and authorization is enforced by Gravitino
(`LOAD_TABLE`), so no filtering logic is added in the MCP layer.

The server already exposes `list_statistics_for_partition`, which requires
partition names as input, but there was no tool to discover those names. These
tools close that gap and round out read coverage for partitioned tables.

Fix: #12099

Yes — two new MCP tools (`list_of_partitions`, `get_partition`), tagged
`partition`.

Added unit tests for the tools (`test_partition.py`) and for URL-encoding /
query-param handling (`test_url_encoding.py`). Full unit suite passes
(163 tests); `black` and `pylint` (10/10) are clean.

### DIFF
--- a/docs/gravitino-mcp-server.md
+++ b/docs/gravitino-mcp-server.md
@@ -124,6 +124,8 @@ Gravitino MCP server supports the following tools, and you could export tool by 
 | `list_policies_for_metadata`        | List all policies associated with a specific metadata item.                    | `policy`     | 1.0.0         |
 | `list_metadata_by_policy`           | List all metadata items associated with a specific policy.                     | `policy`     | 1.0.0         |
 | `get_policy_for_metadata`           | Get a policy associated with a specific metadata item.                         | `policy`     | 1.0.0         |
+| `list_of_partitions`                | Retrieve partitions for a table. Only for catalogs with a partition API.       | `partition`  | 2.0.0         |
+| `get_partition`                     | Retrieve a partition's metadata. Only for catalogs with a partition API.       | `partition`  | 2.0.0         |
 
 
 ## Configuration
--- a/mcp-server/tests/unit/tools/test_partition.py
+++ b/mcp-server/tests/unit/tools/test_partition.py
@@ -1,0 +1,92 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import asyncio
+import unittest
+
+from fastmcp import Client
+
+from mcp_server.client.factory import RESTClientFactory
+from mcp_server.core import Setting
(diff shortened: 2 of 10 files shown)


docs/gravitino-mcp-server.md
```diff
@@ -124,6 +124,8 @@ Gravitino MCP server supports the following tools, and you could export tool by
 | `list_policies_for_metadata`        | List all policies associated with a specific metadata item.                    | `policy`     | 1.0.0         |
 | `list_metadata_by_policy`           | List all metadata items associated with a specific policy.                     | `policy`     | 1.0.0         |
 | `get_policy_for_metadata`           | Get a policy associated with a specific metadata item.                         | `policy`     | 1.0.0         |
+| `list_of_partitions`                | Retrieve a list of partitions for a specific table. Only for catalogs that support the partition API. | `partition`  | 2.0.0         |
```
yuqi1129: Align the table column

mcp-server/tests/unit/tools/test_partition.py
```diff
@@ -0,0 +1,92 @@
...
+
+from fastmcp import Client
+
+from mcp_server.client.factory import RESTClientFactory
+from mcp_server.core import Setting
+from mcp_server.server import GravitinoMCPServer
+from tests.unit.tools import MockOperation
+
+
+class TestPartitionTool(unittest.TestCase):
```
yuqi1129, line 29: Have you leveraged AI to run real tests to verify the effect? Strictly, we lack ITs to cover the MCP operations.

bharos: I tested the tools end-to-end against a live Gravitino through the MCP HTTP server. I only have a `iceberg` catalog available (no Hive catalog), so the calls exercised the full path and returned the expected "Table does not support partition operations" result for Iceberg (plus clean NoSuchCatalog/NoSuchSchema errors for bad inputs), all surfaced correctly as MCP isError responses. I couldn't exercise the happy path (actual partition listing) since that needs a Hive catalog.

I can try to bring up one and check with hive catalog

bharos: @yuqi1129 Done.
Posted here : https://github.com/apache/gravitino/issues/12099#issuecomment-5055315090

yuqi1129: Yeah, let me check it.

yuqi1129 (approved):

bharos:
Thanks @yuqi1129 for the review

Issue #12131, opened 2026-07-22
[EPIC] Add provider-neutral KMS interfaces


nevzheng:
### Describe the proposal

Add server-private, provider-neutral KMS interfaces and named-source routing to Apache Gravitino.

This epic includes:

- typed references that identify a configured KMS source and provider-native key;
- stable JSON persistence for key references and canonical provider API values;
- normalized key properties and capabilities;
- provider-facing client and factory contracts;
- named-source configuration, discovery, routing, and lifecycle management; and
- reusable fake implementations and contract tests.

KMS credentials and key material remain server-private. Concrete provider implementations, provider integration tests, provider documentation, and distribution packaging are outside this epic.

### Task list

- [x] [#12132 Define provider-neutral KMS contracts and canonical key-reference JSON](https://github.com/apache/gravitino/pull/12132)
- [x] [#12133 Add named-source configuration, routing, and lifecycle management](https://github.com/apache/gravitino/pull/12133)

nevzheng:
done: Both scoped KMS foundation PRs are merged with green CI. Concrete provider implementations remain tracked in the enterprise provider stack.

PR #12132, opened 2026-07-22
[#12131] feat(kms): Define key inspection API

nevzheng:

Add provider-neutral Java contracts for server-private KMS key inspection:

- `api`: `KmsReference` (api / source / keyId) and `KmsApiIdentifiers` validation helper
- `common`: `KmsClient` / `KmsClientFactory` SPI, `KmsKeyProperties`, config/auth exceptions, `KmsReferenceDTO`, reusable fake + contract tests

Key identity rules:
- `api` is a `String` matched exactly (no trim/lowercase normalization)
- identifiers must be lowercase kebab-case with no padding (e.g. `aws-kms`), enforced by `KmsApiIdentifiers`
- `source` is trimmed; provider-native `keyId` is preserved as-is
- unknown/custom APIs are allowed at the contract layer; dispatch remains SPI-based

Inspection semantics:
- `KmsClient.getKeyProperties` returns `Optional<KmsKeyProperties>`
- `Optional.empty()` means authoritative key-not-found
- configuration vs authentication failures remain distinct exception types

KMS integrations need a common contract that does not expose credentials or key material. Using string API identifiers (instead of a closed enum) keeps the Developer API easier to extend through SPI, consistent with catalog/credential provider patterns.

This PR defines only the provider-neutral API/SPI. Named-source routing follows in #12133. Concrete providers, provider ITs, server configuration, and cryptographic operations are outside this PR.

Part of #12131.

Yes. It adds new `@DeveloperApi` Java contracts. It does not add REST APIs, server configuration, concrete providers, or cryptographic operations.

- `./gradlew :api:test :common:test`
- `./gradlew :api:spotlessCheck :common:spotlessCheck`
- `./gradlew :common:testFixturesJar`
- `./gradlew :api:javadoc :common:javadoc`

### Series

- Epic: #12131
- Position: 1 of 2
- Previous: none
- Next: #12133
- Full stack: #12132 → #12133


nevzheng:
@lasdf1234, could you also review this KMS API series? The full series is tracked in #12131.

nevzheng:
Hi @roryqi and @lasdf1234 — please review this KMS API change. The full series and review scope are tracked in #12131.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsApi.java
```diff
@@ -0,0 +1,85 @@
...
+package org.apache.gravitino.encryption.kms;
+
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.stream.Collectors;
+import org.apache.gravitino.annotation.DeveloperApi;
+
+/** The KMS product API that defines key identifier and operation semantics. */
+@DeveloperApi
+public enum KmsApi {
```
roryqi: If we want to use enum here and add a new KMS service, it will be difficult.
I prefer using SPI patten. Maybe you can refer to catalog provider and credential provider. 
I feel that it would be better to use String.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsKeyProperties.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.apache.gravitino.annotation.DeveloperApi;
+
+/**
+ * Common properties reported for a KMS key.
+ *
+ * <p>Capabilities describe downstream use; this API does not perform cryptographic operations.
+ */
+@DeveloperApi
+public interface KmsKeyProperties {
```
roryqi, line 31: I don't know the concrete scenario? I feel a weird. Because it will common to use a capabilities class instead of properties class. You can refer to our CatalogCapabiities class.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsAuthenticationException.java
```diff
@@ -0,0 +1,50 @@
...
+   * Creates an authentication exception with the specified detail message.
+   *
+   * @param message the detail message
+   * @param args the arguments to the message
+   */
+  @FormatMethod
+  public KmsAuthenticationException(@FormatString String message, Object... args) {
+    super(message, args);
+  }
+}
```
jerryshao, line 50: The API module is mainly the user-facing APIs. These interfaces are more developer-facing APIs. We should not put it here.

We can either put them in `connector` module. Or, we can put them in the place where we use it.

lasdf1234:
@nevzheng All the java.class files are located in a single package path. Can different functions of java.class be placed in different packages?

common/src/main/java/org/apache/gravitino/encryption/kms/KmsAuthenticationException.java
roryqi: KMS should be used for server and client. Maybe we could consider putting it into `common` module, too.

nevzheng:
Added canonical JSON persistence for KmsReference and KmsApi plus round-trip coverage. Persisted key references now retain their canonical API, source, and key ID. #12133 has been restacked on this head.

nevzheng:
@lasdf1234 Done in 391ec3d14. I separated the functions by Gradle module rather than creating several small Java subpackages: the user-facing reference contract remains in `api`, while the provider SPI, exceptions, and reusable test contracts moved to `common`. The Java package remains cohesive, following the existing credential precedent where `Credential` is in `api` and `CredentialProvider` is in `common` under the same `org.apache.gravitino.credential` package. Concrete providers can use provider-specific subpackages.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsAuthenticationException.java
nevzheng: Done in 391ec3d14. `api` now retains only the user-facing `KmsReference` and `KmsApi` data contracts. `KmsClient`, `KmsClientFactory`, key properties, exceptions, the fake client, and shared contract-test fixtures now live in `common`. Validation passed with `:api:test`, `:common:test`, both modules’ Spotless checks, the `common` test-fixtures JAR, and API/common Javadocs.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsApi.java
nevzheng: Done in 10fb53ada. The closed enum is removed: `KmsReference.api()` and `KmsClientFactory.api()` now use stable string identifiers, and references reject null/blank values and canonicalize by trimming and lowercasing. The identifier is only a registry key; it is never used as a class name or reflection input. API/common tests, JSON round-trip/canonicalization coverage, Spotless, test-fixture packaging, and Javadocs all pass.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsKeyProperties.java
nevzheng: The proposed `CatalogCapability` model is not applicable here. `KmsKeyProperties` describes a specific key’s presence, enabled state, and supported operations. These values vary by key; they are not capabilities of `KmsClient`.

Client- or provider-level capability discovery is a separate API. It is out of scope and can be added later when there is a concrete requirement.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsKeyProperties.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package org.apache.gravitino.encryption.kms;
+
+import org.apache.gravitino.annotation.DeveloperApi;
+
+/**
+ * Common properties reported for a KMS key.
+ *
+ * <p>Capabilities describe downstream use; this API does not perform cryptographic operations.
+ */
```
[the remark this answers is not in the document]
nevzheng: Addressed in cf4a8c976 by making presence structural: `KmsClient.getKeyProperties` now returns `Optional<KmsKeyProperties>`, and `present()` is removed. `Optional.empty()` is reserved for authoritative provider not-found, while authentication, authorization, timeout, availability, and other indeterminate failures remain exceptions. `KmsKeyProperties` therefore only describes a located key, and its Javadocs now define the lifecycle and structural-capability semantics.

api/src/testFixtures/java/org/apache/gravitino/encryption/kms/TestKmsClientContract.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Test
+  void testReportsMissingKey() {
+    KmsReference reference = missingKey();
+    KmsKeyProperties properties = client().getKeyProperties(reference);
+
+    Assertions.assertEquals(reference, properties.reference());
+    Assertions.assertFalse(properties.present());
+  }
```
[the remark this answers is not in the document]
nevzheng: Addressed in cf4a8c976. The reusable provider contract now requires a non-null `Optional`, asserts `Optional.of` for a found key, and asserts `Optional.empty` for authoritative not-found. The fake-client tests also cover a disabled-but-present key, so lifecycle state is independent from presence. API/common tests, Spotless, test-fixture packaging, and Javadocs pass.

nevzheng:
Thanks for the review feedback. I’ve completed the author-side revisions on the current PR:

- Moved the provider SPI, exceptions, and reusable test contracts from `api` to `common`; `api` retains the public `KmsReference` value.
- Replaced the closed `KmsApi` enum with canonical string identifiers so independently loaded providers can extend the SPI safely.
- Clarified that `KmsKeyProperties` describes one inspected key, not client- or provider-level capabilities.
- Changed key lookup to return `Optional<KmsKeyProperties>`: empty means authoritative not-found, while authentication, authorization, availability, and indeterminate failures remain exceptions.
- Documented that lifecycle state is point-in-time and may apply to the selected/current key version; operation support does not guarantee authorization or future success.
- Added contract coverage for canonical identifiers, found/missing/disabled keys, invalid provider results, and distinct failure types.

Local API/common tests, Spotless, test-fixture packaging, and Javadocs pass. GitHub CI currently has no failures; the remaining integration jobs are still running.

Current head: `cf4a8c976`

@roryqi @lasdf1234 — please take another look and either approve or identify any remaining concrete concerns.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsKeyProperties.java
roryqi: OK, I got it.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsReference.java
```diff
@@ -0,0 +1,112 @@
...
+   * @param api explicitly selected KMS API identifier
+   * @param source configured KMS client-instance name
+   * @param keyId provider-native key identifier
+   * @throws IllegalArgumentException if any argument is null or blank
+   */
+  @JsonCreator
+  public KmsReference(
+      @JsonProperty("api") String api,
+      @JsonProperty("source") String source,
+      @JsonProperty("keyId") String keyId) {
```
jerryshao: We typically don't add JSON annotations in the API definition. Instead, we have a DTO for this. This leaves us room to separate the API definition from the wire protocol.

common/src/testFixtures/java/org/apache/gravitino/encryption/kms/FakeKmsClient.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Creates an empty fake client.
+   *
+   * @param api canonical KMS API identifier accepted by the client
+   * @param source configured source accepted by the client
+   */
+  public FakeKmsClient(String api, String source) {
+    if (api == null || api.trim().isEmpty()) {
+      throw new IllegalArgumentException("KMS API cannot be blank");
```
lasdf1234: May be StringUtils is better?

common/src/main/java/org/apache/gravitino/encryption/kms/KmsAuthenticationException.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.apache.gravitino.encryption.kms;
+
+import com.google.errorprone.annotations.FormatMethod;
+import com.google.errorprone.annotations.FormatString;
+import org.apache.gravitino.exceptions.ConnectionFailedException;
+
+/** Indicates that a KMS backend rejected or could not resolve configured credentials. */
+public class KmsAuthenticationException extends ConnectionFailedException {
+
```
lasdf1234, line 27: May be '*Exception' class shoud be placed in Api module and  exceptions package just like other 'Exception class' ?

nevzheng: Nack. These exceptions are part of the provider-facing KMS SPI, not the external user/client API. Following the earlier review, we moved the SPI into `common`; moving only its exceptions back to `api` would split one contract across modules.

They should remain beside the KMS code that throws and consumes them. If these failures later cross a REST or client boundary, we can map them to separate public exceptions there.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsReference.java
nevzheng: Done in 54adf1693. `KmsReference` is now an annotation-free API value, while `KmsReferenceDTO` in `common` owns the Jackson wire shape and bidirectional conversion. `testKmsReferenceDtoSerde` verifies the exact JSON shape, DTO round-trip, conversion back to the API value, and API identifier canonicalization. The complete API/common tests, Spotless checks, test-fixture packaging, and Javadocs pass.

common/src/testFixtures/java/org/apache/gravitino/encryption/kms/FakeKmsClient.java
nevzheng: Done in 54adf1693. The fake client now uses `StringUtils.isBlank(api)`, with constructor coverage for null, empty, and whitespace-only values. `commons-lang3` is scoped to the reusable test-fixtures implementation, and the complete common test suite passes.

nevzheng:
@jerryshao @roryqi — the latest review updates are pushed in `54adf1693`:

- separated the `KmsReference` API value from its Jackson wire DTO;
- added exact DTO serialization, round-trip, conversion, and canonicalization coverage; and
- adopted the requested `StringUtils.isBlank` validation in the fake provider fixture.

The complete API/common tests, Spotless checks, fixture packaging, and Javadocs pass. Please take another look.

common/build.gradle.kts
jerryshao, line 40: Can you remove these dependencies? I don't think API module depends on the jackson module, nor should we.

jerryshao:
LGTM, just some minor issues. @nevzheng

lasdf1234:
@nevzheng Check and remove unnecessary dependencies. Resolve the copilot's comment. Beside, I have no more comment.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsApi.java
nevzheng: I reconsidered this and restored the string-backed `KmsApi` enum in 64c5bdc29. This is an intentional judgment call: Gravitino must prevalidate KMS API identifiers against the set it explicitly supports. A domain enum provides that guarantee and avoids a stringly typed contract where unsupported values fail only during runtime dispatch. Adding a new KMS API should be an explicit API change; the SPI remains the extension point for implementations of those approved APIs.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsReference.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public KmsReference(String api, String source, String keyId) {
+    Preconditions.checkArgument(StringUtils.isNotBlank(api), "KMS API cannot be blank");
+    Preconditions.checkArgument(StringUtils.isNotBlank(source), "KMS source cannot be blank");
+    Preconditions.checkArgument(StringUtils.isNotBlank(keyId), "KMS key ID cannot be blank");
+
+    this.api = api.trim().toLowerCase(Locale.ROOT);
+    this.source = source;
+    this.keyId = keyId;
+  }
```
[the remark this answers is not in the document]
nevzheng: Done in b09ef2351. `KmsReference` now trims the logical source name while preserving its case and the provider-native `keyId` verbatim. `TestKmsReference` covers the normalized source and opaque key behavior. API/common tests, Spotless, fixture packaging, and Javadocs pass.

nevzheng:
@lasdf1234 Addressed the `KmsReference.source` normalization feedback in b09ef2351 and resolved that Copilot thread. API/common tests, Spotless, fixture packaging, and Javadocs pass. I am continuing through the remaining minor review items.

common/src/testFixtures/java/org/apache/gravitino/encryption/kms/FakeKmsClient.java
```diff
@@ -0,0 +1,141 @@
...
+   * @param api canonical KMS API identifier accepted by the client
+   * @param source configured source accepted by the client
+   */
+  public FakeKmsClient(String api, String source) {
+    if (StringUtils.isBlank(api)) {
+      throw new IllegalArgumentException("KMS API cannot be blank");
+    }
+    this.api = api.trim().toLowerCase(Locale.ROOT);
+    this.source = source;
+  }
```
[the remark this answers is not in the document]
nevzheng: Done in 14367a6cb. This is intentionally scoped to the reusable test fixture: `FakeKmsClient` now rejects null/blank source names and trims valid names, with focused constructor and normalization coverage. No production-client or external-API validation was added. Common tests, Spotless, fixture packaging, and Javadocs pass.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsClient.java
```diff
@@ -0,0 +1,53 @@
...
+   *
+   * <p>An empty result means the provider authoritatively reported that the key does not exist.
+   * Authentication, authorization, timeout, availability, and other indeterminate failures are
+   * reported as exceptions, never as an empty result.
+   *
+   * @param reference key to inspect
+   * @return normalized key properties, or empty when the key authoritatively does not exist; never
+   *     null
+   * @throws IllegalArgumentException if the reference does not belong to this client
+   * @throws ConnectionFailedException if the provider cannot be queried
```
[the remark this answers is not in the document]
nevzheng: Done in f4c0e16af. `KmsClient.getKeyProperties` now documents the non-null reference requirement and that null or a reference owned by another API/source produces `IllegalArgumentException`, matching the reusable contract tests. Common tests, Spotless, and Javadocs pass.

common/build.gradle.kts
nevzheng: @jerryshao, this cleanup request is reasonable. I removed the unnecessary test-fixture dependency introduced by this PR and replaced its only use with JDK code.

The remaining selected Jackson dependencies predate this change and are used broadly throughout `common`. Removing them would require auditing and refactoring existing consumers, reconsidering the shared JSON/DTO module boundary, and validating downstream dependency metadata and builds. That is worthwhile, but outside the scope of this KMS API PR.

Tracked separately in [#12193](https://github.com/apache/gravitino/issues/12193).

common/src/testFixtures/java/org/apache/gravitino/encryption/kms/FakeKmsClient.java
nevzheng: Superseded by 64c5bdc29 and 99b5defeb. `api` is now the closed `KmsApi` enum, so the fake client no longer performs string validation for it. The remaining fixture-only source check uses JDK validation, which lets us remove the fixture Commons Lang dependency. Fixture compilation, packaging, and tests pass. Resolving this thread as complete.

api/src/main/java/org/apache/gravitino/encryption/kms/KmsApi.java
nevzheng: Thanks for raising this. I considered the string/SPI alternative and am resolving this in favor of the string-backed enum. `KmsApi` represents the finite set of KMS APIs Gravitino explicitly supports, so unsupported identifiers should be rejected at the API boundary rather than accepted as arbitrary runtime strings. The SPI remains the extension point for implementations of those supported APIs; adding a new KMS API is intentionally an explicit contract change.

This is the design judgment for this PR, so I am closing the thread.

common/src/main/java/org/apache/gravitino/encryption/kms/KmsAuthenticationException.java
nevzheng: Closing as addressed. These exceptions are part of the provider-facing KMS SPI in `common`; moving them into `api` would split the contract across modules. If these failures later cross a REST or client boundary, we can introduce separate public exception mappings there.

nevzheng:
@jerryshao @roryqi @lasdf1234 — this review round is complete.

Changes made from the feedback:
- separated the public `KmsReference` / `KmsApi` values from the developer-facing SPI, DTOs, and exceptions in `common`;
- restored the string-backed `KmsApi` enum so supported APIs are validated at the boundary while implementations remain SPI-extensible;
- modeled authoritative key-not-found with `Optional.empty()` and clarified key-state and client preconditions;
- normalized logical source names while preserving provider-native key IDs;
- added DTO serialization, provider-contract, and fake-client coverage; and
- removed the unnecessary test-fixture utility dependency and filed #12193 for the broader Jackson dependency audit.

Focused API/common tests, Spotless, fixture packaging, and Javadocs pass. Current CI has no failures, with the remaining jobs still running. All review threads are resolved.

@roryqi @lasdf1234, please re-review when ready.

roryqi:
> @jerryshao @roryqi @lasdf1234 — this review round is complete.
> 
> Changes made from the feedback:
> 
> * separated the public `KmsReference` / `KmsApi` values from the developer-facing SPI, DTOs, and exceptions in `common`;
> * restored the string-backed `KmsApi` enum so supported APIs are validated at the boundary while implementations remain SPI-extensible;
> * modeled authoritative key-not-found with `Optional.empty()` and clarified key-state and client preconditions;
> * normalized logical source names while preserving provider-native key IDs;
> * added DTO serialization, provider-contract, and fake-client coverage; and
> * removed the unnecessary test-fixture utility dependency and filed [[Improvement] Audit and narrow Jackson dependency boundaries in api and common #12193](https://github.com/apache/gravitino/issues/12193) for the broader Jackson dependency audit.
> 
> Focused API/common tests, Spotless, fixture packaging, and Javadocs pass. Current CI has no failures, with the remaining jobs still running. All review threads are resolved.
> 
> @roryqi @lasdf1234, please re-review when ready.

If you use enum for  KmsApi and add want to add a new KmsApi, you must modify the code. right? It's not convenient.

nevzheng:
Yes — adding a new `KmsApi` requires a core API change, and that’s intentional. `KmsApi` is the closed set of APIs Gravitino supports (same idea as `JobType`). The SPI remains the extension point for implementations of those APIs.

I’m keeping the enum for this PR. Further debate on open strings vs enum isn’t material to the rest of the series, so I’d rather not bikeshed it here.

Ref: Effective Java (3rd ed.), Item 34 — prefer enums over string constants for a fixed set of values known at compile time.

LLMs make code modifications inexpensive these days. I'd like to focus on getting the well tested, well designed, well specified, well functioning code possible. "Insist on highest standards"

@roryqi

roryqi:
> Yes — adding a new `KmsApi` requires a core API change, and that’s intentional. `KmsApi` is the closed set of APIs Gravitino supports (same idea as `JobType`). The SPI remains the extension point for implementations of those APIs.
> 
> I’m keeping the enum for this PR. Further debate on open strings vs enum isn’t material to the rest of the series, so I’d rather not bikeshed it here.
> 
> Ref: Effective Java (3rd ed.), Item 34 — prefer enums over string constants for a fixed set of values known at compile time.
> 
> LLMs make code modifications inexpensive these days. I'd like to focus on getting the well tested, well designed, well specified, well functioning code possible. "Insist on highest standards"
> 
> @roryqi

@jerryshao WDYT?

api/src/main/java/org/apache/gravitino/encryption/kms/KmsApi.java
nevzheng: @roryqi Thanks for the sync — we agreed to prefer a `String` for the KMS API identifier so the contract stays easier to extend.

Changes in the follow-up:
- Removed the closed `KmsApi` enum
- Restored string-backed `KmsReference.api` / `KmsClientFactory.api` with exact matching (no trim/lowercase normalization)
- Documented and enforced lowercase kebab-case identifiers with no padding via `KmsApiIdentifiers`
- Shared that validation between `KmsReference` and `FakeKmsClient`
- Added `TestKmsReferenceDTO` coverage, including custom identifiers like `acme-kms-v2`
- Kept `source` trimming and left `keyId` untouched

Dispatch remains SPI-based.

nevzheng:
@roryqi @lasdf1234 @jerryshao I updated the PR description to match the current state.

Rory and I synced and decided on strings for flexibility.

Thanks for your feedback so far.

roryqi (approved):

roryqi:
Thanks @nevzheng for contribution! Thanks @jerryshao @lasdf1234 for review.

PR #12133, opened 2026-07-22
[#12131] feat(kms): Configure and resolve named sources

nevzheng:

Add server-side configuration, resolution, and lifecycle management for named KMS sources:

- parse and validate `gravitino.kms.*` source configuration;
- discover `KmsClientFactory` implementations with `ServiceLoader`;
- create and own one reusable client per configured source during environment initialization;
- resolve each `KmsReference` to its configured source and API;
- close clients in reverse order, including after partial startup failure; and
- define provider-client concurrency and ownership expectations.

`KmsClientFactory.create(...)` validates local configuration and constructs a reusable client without contacting the configured KMS. Network, authentication, and availability failures are reported by client operations and do not invalidate the reusable client.

This builds on the provider-neutral KMS contracts merged in #12132. Concrete provider implementations remain outside this PR.

A deployment can use multiple named KMS instances, including multiple instances of the same provider. Gravitino needs one server-side resolver and lifecycle owner so callers do not construct provider SDK or HTTP clients per request or depend on provider-specific configuration.

Creating clients during environment initialization validates the complete local configuration deterministically without making startup depend on KMS availability.

Part of #12131.

Yes. It introduces these server configuration properties:

- `gravitino.kms.sources`
- `gravitino.kms.source.<name>.api`
- `gravitino.kms.source.<name>.<provider-property>`

Provider properties remain server-private. The registry is initialized with both base and full Gravitino environments. Client construction does not connect to the configured KMS.

- `./gradlew :core:test --tests org.apache.gravitino.encryption.kms.TestKmsConfig --tests org.apache.gravitino.encryption.kms.TestKmsClientRegistry --tests org.apache.gravitino.TestGravitinoEnvKmsClientRegistry -PskipITs -PskipDockerTests=true -PskipWeb=true`
- `./gradlew :common:spotlessCheck :core:spotlessCheck :common:javadoc :core:javadoc -PskipITs -PskipDockerTests=true -PskipWeb=true`

### Review size

This PR contains 1,129 additions: 434 production lines and 695 test lines. Configuration parsing, client resolution, eager lifecycle management, and server wiring form one usable unit. Splitting them would leave unused internal code or separate lifecycle tests from the behavior they verify.

### Series

- Epic: #12131
- Previous: #12132 (merged)
- Position: 2 of 2

### DIFF
--- a/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
+++ b/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
@@ -210,13 +210,7 @@ public class GravitinoEnv {
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
     this.manageFullComponents = false;
-    this.kmsClientRegistry = new KmsClientRegistry(config);
-    try {
-      initBaseComponents();
-    } catch (RuntimeException | Error e) {
-      closeKmsClientRegistry();
-      throw e;
-    }
+    initBaseComponents();
     LOG.info("Gravitino base environment is initialized.");
   }
 
@@ -230,14 +224,8 @@ public class GravitinoEnv {
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
     this.manageFullComponents = true;
-    this.kmsClientRegistry = new KmsClientRegistry(config);
-    try {
-      initBaseComponents();
-      initGravitinoServerComponents();
-    } catch (RuntimeException | Error e) {
-      closeKmsClientRegistry();
-      throw e;
-    }
+    initBaseComponents();
+    initGravitinoServerComponents();
     LOG.info("Gravitino full environment is initialized.");
   }
 
@@ -613,59 +601,56 @@ public class GravitinoEnv {
   public void shutdown() {
     LOG.info("Shutting down Gravitino Environment...");
 
-    try {
-      if (entityStore != null) {
-        try {
-          entityStore.close();
-        } catch (Exception e) {
-          LOG.warn("Failed to close EntityStore.", e);
-        }
+    if (entityStore != null) {
+      try {
+        entityStore.close();
+      } catch (Exception e) {
+        LOG.warn("Failed to close EntityStore.", e);
       }
-
-      if (catalogManager != null) {
-        catalogManager.close();
-      }
-
-      if (auxServiceManager != null) {
-        try {
-          auxServiceManager.serviceStop();
-        } catch (Exception e) {
-          LOG.warn("Failed to stop AuxServiceManager", e);
-        }
-      }
-
-      if (metricsSystem != null) {
-        metricsSystem.close();
-      }
-
-      if (eventListenerManager != null) {
-        eventListenerManager.stop();
-      }
-
-      if (metalakeManager != null) {
-        metalakeManager.close();
-      }
-
-      if (jobOperationDispatcher != null) {
-        try {
-          jobOperationDispatcher.close();
-          jobOperationDispatcher = null;
-        } catch (Exception e) {
-          LOG.warn("Failed to close JobOperationDispatcher", e);
-        }
-      }
-
-      if (statisticDispatcher != null) {
-        try {
-          statisticDispatcher.close();
-        } catch (Exception e) {
-          LOG.warn("Failed to close StatisticDispatcher", e);
-        }
-      }
-
-    } finally {
-      closeKmsClientRegistry();
     }
+
+    if (catalogManager != null) {
+      catalogManager.close();
+    }
+
+    if (auxServiceManager != null) {
+      try {
+        auxServiceManager.serviceStop();
+      } catch (Exception e) {
+        LOG.warn("Failed to stop AuxServiceManager", e);
+      }
+    }
+
+    if (metricsSystem != null) {
+      metricsSystem.close();
+    }
+
+    if (eventListenerManager != null) {
+      eventListenerManager.stop();
+    }
+
+    if (metalakeManager != null) {
+      metalakeManager.close();
+    }
+
+    if (jobOperationDispatcher != null) {
+      try {
+        jobOperationDispatcher.close();
+        jobOperationDispatcher = null;
+      } catch (Exception e) {
+        LOG.warn("Failed to close JobOperationDispatcher", e);
+      }
+    }
+
+    if (statisticDispatcher != null) {
+      try {
+        statisticDispatcher.close();
+      } catch (Exception e) {
+        LOG.warn("Failed to close StatisticDispatcher", e);
+      }
+    }
+
+    closeKmsClientRegistry();
 
     LOG.info("Gravitino Environment is shut down.");
   }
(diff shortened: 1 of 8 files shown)


nevzheng:
Hi @roryqi and @lasdf1234 — please review the second and final PR in the KMS API series. Previous: #12132. Epic: #12131.

nevzheng:
Restacked on #12132 after its canonical key-reference JSON persistence fix. This PR functional scope is unchanged.

nevzheng:
Restacked onto the current #12132 head (`cf4a8c976`) and propagated its key-not-found contract. The registry now preserves `Optional.empty()` only for an authoritative provider not-found result, while rejecting null provider results and mismatched key references. Focused API/common/core tests, Spotless, and Javadocs pass locally. Current head: `235980fc5`.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -586,58 +613,75 @@ public void start() {
   public void shutdown() {
     LOG.info("Shutting down Gravitino Environment...");
 
-    if (entityStore != null) {
-      try {
-        entityStore.close();
-      } catch (Exception e) {
-        LOG.warn("Failed to close EntityStore.", e);
+    try {
```
nevzheng: issue: Wrapping the entire existing shutdown sequence in `try`/`finally` creates a large diff in unrelated component lifecycle code and gives KMS cleanup a stronger failure guarantee than the other components currently receive. Please leave the existing shutdown flow unchanged and close the registry once at the end. A holistic always-close lifecycle refactor should be separate.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -207,7 +210,13 @@ public void initializeBaseComponents(Config config) {
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
     this.manageFullComponents = false;
-    initBaseComponents();
+    this.kmsClientRegistry = new KmsClientRegistry(config);
+    try {
```
nevzheng: suggestion: Can we simplify how `KmsClientRegistry` is added to `GravitinoEnv`? Registering one owned component now duplicates `try`/`catch` rollback blocks in both initialization paths, which feels disproportionate and unlike the nearby component initialization. Please keep the lifecycle change consistent with the existing environment pattern and minimize KMS-specific control flow.

core/src/main/java/org/apache/gravitino/encryption/kms/KmsConfig.java
```diff
@@ -0,0 +1,152 @@
...
+final class KmsConfig {
+
+  static final String KMS_CONFIG_PREFIX = "gravitino.kms.";
+  static final String KMS_SOURCES = KMS_CONFIG_PREFIX + "sources";
+
+  private static final String SOURCES = "sources";
+  private static final String SOURCE_PREFIX = "source.";
+  private static final String API = "api";
+  private static final Pattern SOURCE_NAME_PATTERN = Pattern.compile("[A-Za-z0-9][A-Za-z0-9_-]*");
+  private static final Pattern SOURCE_PROPERTY_PATTERN =
```
nevzheng: suggestion: Can we simplify this parser? The grammar is only `sources` plus `source.<name>.<property>`, but the regex-based structural parsing and multiple passes make it harder to follow. Please keep the small source-name validation, parse hierarchical keys with `startsWith`/`indexOf`/`substring`, and build each source property map in one pass while preserving malformed and unlisted-key checks.

nevzheng:
@roryqi @lasdf1234 PTAL when you have a chance. I’ve updated the implementation and PR description to reflect the current KMS configuration and lifecycle design. Thanks!

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
nevzheng: Done in `0cd36a9b8`. Restored the existing shutdown flow, close the KMS registry once at the end, and removed the test that asserted a stronger KMS-only cleanup guarantee. The full core test suite passes.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
nevzheng: Done in `0cd36a9b8`. The registry now initializes once in `initBaseComponents()`, and the duplicated KMS-specific rollback blocks have been removed from both public initialization paths.

core/src/main/java/org/apache/gravitino/encryption/kms/KmsConfig.java
nevzheng: Done in `0cd36a9b8`. Removed the structural regex and separate validation pass. Configuration keys are now parsed in one pass with `startsWith`, `indexOf`, and `substring`, with added malformed-key coverage.

roryqi (dismissed):
LGTM.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -635,10 +650,26 @@ public void shutdown() {
       }
     }
 
+    closeKmsClientRegistry();
+
     LOG.info("Gravitino Environment is shut down.");
   }
 
+  private void closeKmsClientRegistry() {
+    KmsClientRegistry registry = kmsClientRegistry;
```
lasdf1234: According to the code conventions, is this method not going to be used at all?

lasdf1234:
@nevzheng There is only one minor issue that needs to be corrected. I have no other comments.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
nevzheng: note: I removed the single-use helper and inlined registry shutdown into `GravitinoEnv.shutdown()`. `KmsClientRegistry` now owns its closed state, so `GravitinoEnv` no longer duplicates that lifecycle state.

question: Could you clarify which code convention you mean? I am not sure whether your concern was the single-use helper, the field lifecycle, or something else.

lasdf1234 (approved):
LGTM

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
lasdf1234: Since this method is only used once—and the other methods in the class follow the same pattern—it can be inlined.

nevzheng:
note: Changes since the previous review:

- simplified registry ownership and shutdown;
- made empty KMS configuration produce a valid empty registry without enumerating providers; and
- strengthened configuration, routing, partial-initialization, and close-behavior tests.

The public key-inspection contract is unchanged.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
nevzheng: done: Inlined the single-use `KmsClientRegistry` close directly into `GravitinoEnv.shutdown()`.

lasdf1234:
> note: Changes since the previous review:
> 
> * simplified registry ownership and shutdown;
> * made empty KMS configuration produce a valid empty registry without enumerating providers; and
> * strengthened configuration, routing, partial-initialization, and close-behavior tests.
> 
> The public key-inspection contract is unchanged.

Got it. I've approved. But I dont't have privilege to merge. When the ci is completed, you can ask Rory to review and merge.

roryqi (approved):
LGTM, wait for CI.

PR #12146, opened 2026-07-22
[#12145] fix(test): prevent overlapping embedded server lifecycle

### LINKED ISSUE #12145: [Bug report] MiniGravitino shutdown can close the next server ClassLoaderPool
### Version

main branch

### Describe what's wrong

In embedded integration tests, `MiniGravitino.stop()` returns after the old HTTP port becomes unavailable, without waiting for the server task and `GravitinoEnv.shutdown()` to finish.

The next test class can start another embedded server in the same JVM while the previous shutdown is still running. Because `GravitinoEnv` is a singleton with mutable component fields, the old shutdown can close the newly initialized `CatalogManager`. Since #10480, this permanently closes the new manager's `ClassLoaderPool`, and subsequent catalog creation fails.

`MetalakeIT.testMetalakeAvailable` failed in CI:

```
java.lang.IllegalStateException: ClassLoaderPool is already closed
  at org.apache.gravitino.utils.ClassLoaderPool.acquire(ClassLoaderPool.java:67)
  at org.apache.gravitino.catalog.CatalogManager.createCatalogWrapper(CatalogManager.java:1198)
```

Failed run:
https://github.com/apache/gravitino/actions/runs/29865361861/job/88752222436?pr=12096

### How to reproduce

Run client-java integration tests in embedded mode so multiple test classes repeatedly start and stop `MiniGravitino`.

The failure requires this ordering:

1. Server A begins shutdown.
2. Its HTTP port closes, so `MiniGravitino.stop()` returns.
3. Server B initializes the singleton `GravitinoEnv`.
4. Server A's unfinished shutdown closes Server B's `CatalogManager`.
5. Server B attempts to create a catalog.

### Additional context

PR #10480 exposed the race by introducing the terminal `ClassLoaderPool.close()` state. PR #12096 is unrelated.

`MiniGravitino.stop()` should wait for the server task to terminate completely. The environment lifecycle should also prevent overlapping initialize/shutdown operations on the singleton.

yuqi1129:

Wait for the embedded Gravitino server task to terminate completely in
`MiniGravitino.stop()` instead of treating an unavailable HTTP port as
completion.

Always close the REST client and delete the temporary configuration directory,
even when server termination times out or is interrupted. Preserve the original
shutdown failure and attach REST client cleanup failures as suppressed exceptions.

Embedded integration tests repeatedly start and stop Gravitino servers in the
same JVM. The HTTP port can close before `GravitinoEnv.shutdown()` finishes,
allowing the next server to initialize the singleton environment while the old
server is still shutting down.

The old shutdown can then close the new server catalog manager, leaving its
`ClassLoaderPool` permanently closed and causing catalog creation to fail with:

`IllegalStateException: ClassLoaderPool is already closed`

Fix: #12145

No.

- `./gradlew :integration-test-common:check :integration-test-common:spotlessCheck -PskipITs`
- `./gradlew :core:test --tests org.apache.gravitino.catalog.TestClassLoaderPoolIntegration :integration-test-common:compileTestJava -PskipITs`
- `./gradlew :clients:client-java:test --tests org.apache.gravitino.client.integration.test.MetalakeIT -PskipDockerTests=true`
- `git diff --check`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
+++ b/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
@@ -202,7 +202,7 @@ public class GravitinoEnv {
    *
    * @param config The configuration object to initialize the environment.
    */
-  public void initializeBaseComponents(Config config) {
+  public synchronized void initializeBaseComponents(Config config) {
     LOG.info("Initializing Gravitino base environment...");
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
@@ -216,7 +216,7 @@ public class GravitinoEnv {
    *
    * @param config The configuration object to initialize the environment.
    */
-  public void initializeFullComponents(Config config) {
+  public synchronized void initializeFullComponents(Config config) {
     LOG.info("Initializing Gravitino full environment...");
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
@@ -583,7 +583,7 @@ public class GravitinoEnv {
   }
 
   /** Shutdown the Gravitino environment. */
-  public void shutdown() {
+  public synchronized void shutdown() {
     LOG.info("Shutting down Gravitino Environment...");
 
     if (entityStore != null) {
--- a/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
+++ b/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
@@ -202,7 +202,7 @@ public class GravitinoEnv {
    *
    * @param config The configuration object to initialize the environment.
    */
-  public synchronized void initializeBaseComponents(Config config) {
+  public void initializeBaseComponents(Config config) {
     LOG.info("Initializing Gravitino base environment...");
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
@@ -216,7 +216,7 @@ public class GravitinoEnv {
    *
    * @param config The configuration object to initialize the environment.
    */
-  public synchronized void initializeFullComponents(Config config) {
+  public void initializeFullComponents(Config config) {
     LOG.info("Initializing Gravitino full environment...");
     this.config = config;
     FileFetcher.get().initialize(config.get(Configs.BLOCK_UNSAFE_REMOTE_URI));
@@ -583,7 +583,7 @@ public class GravitinoEnv {
   }
 
   /** Shutdown the Gravitino environment. */
-  public synchronized void shutdown() {
+  public void shutdown() {
     LOG.info("Shutting down Gravitino Environment...");
 
     if (entityStore != null) {
--- a/integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
+++ b/integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
@@ -24,6 +24,7 @@ import static org.apache.gravitino.lance.common.config.LanceConfig.LANCE_CONFIG_
 import static org.apache.gravitino.lance.common.config.LanceConfig.NAMESPACE_BACKEND;
 import static org.apache.gravitino.lance.common.config.LanceConfig.NAMESPACE_BACKEND_URI;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/gravitino/GravitinoEnv.java
```diff
@@ -583,7 +583,7 @@ public void start() {
   }
 
   /** Shutdown the Gravitino environment. */
-  public void shutdown() {
+  public synchronized void shutdown() {
```
LuciferYang: I think what actually fixes this issue here is `awaitTermination` in `MiniGravitino.stop()`; the `synchronized` half doesn't match what the description claims. `synchronized` only prevents init/shutdown from running concurrently — it doesn't stop an old shutdown that has fully completed from acting on the next generation's fields. `shutdown()` reads the singleton's current `catalogManager` at execution time (and never nulls it), so if B's init takes the lock and sets the field first, then A's shutdown takes the lock afterwards, `catalogManager.close()` still closes B's manager. `start()` also isn't synchronized and so isn't mutually exclusive with `shutdown()`, so this set of locks doesn't really establish a "one lifecycle op at a time" invariant. The lock is harmless in a single-JVM production run, so no code change is needed, but I'd reword the "synchronized prevents an unfinished shutdown from closing a new server's components" line to reflect reality: the real serialization comes from `awaitTermination`, and `synchronized` is just extra defense.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
```diff
@@ -220,6 +220,13 @@ public void stop() throws IOException, InterruptedException {
     sleepUninterruptibly(500, TimeUnit.MILLISECONDS);
     executor.shutdownNow();
 
+    // The HTTP port may be closed before GravitinoServer.main() finishes shutting down the
+    // singleton GravitinoEnv. Wait for the server task to terminate so the next embedded server
+    // cannot initialize the same environment while this shutdown is still in progress.
+    if (!executor.awaitTermination(3, TimeUnit.MINUTES)) {
```
LuciferYang: The new `awaitTermination` check sits before `restClient.close()` and `deleteDirectory(mockConfDir)`, so when it times out (RuntimeException) or the wait is interrupted (InterruptedException), both cleanup steps are skipped and the restClient and temp dir leak. The previous `if (started) throw` ran after cleanup, so this is introduced by this PR. Moving the restClient/temp-dir cleanup into a `finally` covers it; note `restClient.close()` needs its own try/catch inside the finally, otherwise the IOException it may throw would replace the in-flight RuntimeException/InterruptedException (`deleteDirectory` already has a try/catch(ignore), so it's fine).

LuciferYang (approved):
Some minor issues

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
yuqi1129: Fixed in 73460081d. Resource cleanup now runs in finally. The original timeout or interruption remains the primary failure, and an IOException from closing the REST client is attached as a suppressed exception. Added tests for both timeout cleanup and interrupted cleanup.

core/src/main/java/org/apache/gravitino/GravitinoEnv.java
yuqi1129: Agreed. I removed the synchronized changes from GravitinoEnv and updated the PR description to make awaitTermination the lifecycle barrier.

integration-test-common/src/test/java/org/apache/gravitino/integration/test/MiniGravitino.java
```diff
@@ -216,31 +233,39 @@ public void start() throws Exception {
...
-    try {
-      FileUtils.deleteDirectory(mockConfDir);
-    } catch (Exception e) {
-      // Ignore
-    }
+      long beginTime = System.currentTimeMillis();
+      boolean started = true;
 
-    if (started) {
-      throw new RuntimeException("Can not stop Gravitino server");
```
diqiu50, line 243: executor.shutdown() guarantees that the min server has already been stopped. The following code is redundant

yuqi1129: Yeah, the following three can guarantee that the pool has been shutdown. 

```java
executor.shutdown();
executor.shutdownNow();
executor.awaitTermination()
```

diqiu50 (commented):
LGTM

diqiu50 (approved):

PR #12160, opened 2026-07-22
[#12154] fix(core): Clean up orphaned metadata object relations

### LINKED ISSUE #12154: [Subtask] Background GC for orphaned metadata-object relation rows
### Describe the subtask

Relation tables (`owner_meta`, `tag_relation_meta`, `policy_relation_meta`, `statistic_meta`, `role_meta_securable_object`) reference an entity via a plain `metadata_object_id (+ type)` column with no foreign key. The normal delete path already cascades these in one transaction (e.g. `TableMetaService.deleteTable`), but rows can still be orphaned when an entity row is lost out-of-band or re-imported under a new id (see #12153). Nothing actively removes such orphaned relations today.

Pre-existing data-hygiene gap, independent of the TreeLock work.

### How to fix

Add a pass to `RelationalGarbageCollector` that soft-deletes relation rows whose `metadata_object_id` has no live entity of that type — one statement per relation table, e.g.:

```sql
UPDATE owner_meta SET deleted_at = ?
WHERE deleted_at = 0 AND metadata_object_type = 'TABLE'
  AND NOT EXISTS (SELECT 1 FROM table_meta WHERE table_id = metadata_object_id AND deleted_at = 0);
```

- Iterate over each relation table and each metadata object type.
- Run on the existing GC schedule; keep it idempotent and rate-limited.
- Tests: orphaned owner/tag/policy/statistic rows are collected; live relations are untouched.

### Parent issue

yuqi1129:

Add a garbage collection pass to soft-delete metadata object relation rows whose referenced entity no longer has a live record.

The cleanup:

- covers owner, tag, policy, statistic, and role securable-object relations;
- iterates over supported metadata object types;
- processes orphaned object IDs in bounded batches;
- runs on the existing relational garbage collector schedule;
- remains idempotent.

Metadata object relation tables reference entities using plain object ID and type columns without foreign keys. If an entity row is lost outside the normal transactional deletion path, its relation rows remain orphaned indefinitely.

This change provides background cleanup for those orphaned rows while leaving relations belonging to live entities untouched.

Fix: #12154

No. This change only adds background cleanup for orphaned relational-store data.

Added tests covering:

- cleanup of orphaned owner, tag, policy, statistic, and securable-object relations;
- preservation of relations referencing live entities;
- idempotent repeated cleanup;
- repeated batch execution by the relational garbage collector.

Ran:

`./gradlew :core:test -PskipITs -PskipDockerTests=true`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/JDBCBackend.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/JDBCBackend.java
@@ -95,7 +95,7 @@ import org.slf4j.LoggerFactory;
  * syntax, please implement the SQL statements and methods in MyBatis Mapper separately and switch
  * according to the {@link Configs#ENTITY_RELATIONAL_JDBC_BACKEND_URL_KEY} parameter.
  */
-public class JDBCBackend implements RelationalBackend {
+public class JDBCBackend implements RelationalBackend, SupportsOrphanedRelationCleanup {
 
   private static final Logger LOG = LoggerFactory.getLogger(JDBCBackend.class);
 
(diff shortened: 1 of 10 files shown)


roryqi:
Do we need to consider cache about this improvement?

yuqi1129:
> Do we need to consider cache about this improvement?

Relation data will not be cached when https://github.com/apache/gravitino/pull/12006 is merged. So, it's okay to leave the cache unchaged.

roryqi (approved):
LGTM

PR #12162, opened 2026-07-22
[#12163] feat(api): Add tag assignment value APIs

### LINKED ISSUE #12163: [Subtask] Add tag assignment value APIs
### Describe the subtask

Track the API part of tag assignment values for PR #12162.

This subtask adds:
- Public `TagValue` API model for valued tag assignments.
- `allowedValues` exposure on tag APIs.
- `TagValue[]` association overloads while preserving existing valueless `String[]` APIs.
- Bean-style JSON deserialization support without `JsonProperty`, `JsonCreator`, or `ConstructorProperties`.

PR: https://github.com/apache/gravitino/pull/12162

### Parent issue

#12127

roryqi:

Add the public tag assignment value APIs with `TagValue`, tag allowed values, and valued tag association overloads.

This is the API layer for supporting multiple assignment values for a tag.

Fix: #12163

Yes. It adds `TagValue`, exposes tag allowed values, and adds valued tag association APIs while keeping existing valueless tag association APIs.

`./gradlew :api:spotlessApply`
`./gradlew :api:test --tests org.apache.gravitino.tag.TestTagValue -PskipITs -PskipDockerTests=false`

### DIFF
--- a/api/src/main/java/org/apache/gravitino/tag/SupportsTags.java
+++ b/api/src/main/java/org/apache/gravitino/tag/SupportsTags.java
@@ -54,17 +54,30 @@ public interface SupportsTags {
   Tag getTag(String name) throws NoSuchTagException;
 
   /**
-   * Associate tags to the specific object. The tagsToAdd will be added to the object, and the
-   * tagsToRemove will be removed from the object. Note that: 1) Adding or removing tags that are
-   * not existed will be ignored. 2) If the same name tag is in both tagsToAdd and tagsToRemove, it
-   * will be ignored. 3) If the tag is already associated with the object, it will throw {@link
-   * TagAlreadyAssociatedException}
+   * Associate tags with no value to the specific object. The tagsToAdd will be added to the object,
+   * and the tagsToRemove will be removed from the object. Missing tags are ignored. If the same tag
+   * is in both tagsToAdd and tagsToRemove, it will be ignored. Repeated existing assignments are
+   * idempotent.
    *
-   * @param tagsToAdd The arrays of tag name to be added to the object.
-   * @param tagsToRemove The array of tag name to be removed from the object.
+   * @param tagsToAdd The array of tag names to be added to the object.
+   * @param tagsToRemove The array of tag names to be removed from the object.
    * @return The array of tag names that are associated with the object.
-   * @throws TagAlreadyAssociatedException If the tag is already associated with the object.
+   * @throws TagAlreadyAssociatedException If the association cannot be applied.
    */
   String[] associateTags(String[] tagsToAdd, String[] tagsToRemove)
       throws TagAlreadyAssociatedException;
+
+  /**
+   * Associate tag values to the specific object.
+   *
+   * @param tagsToAdd The tag values to be added to the object.
+   * @param tagsToRemove The tag values to be removed from the object.
+   * @return The array of tag names that are associated with the object.
+   * @throws TagAlreadyAssociatedException If the association cannot be applied.
+   */
+  default String[] associateTags(TagValue[] tagsToAdd, TagValue[] tagsToRemove)
+      throws TagAlreadyAssociatedException {
+    throw new UnsupportedOperationException(
+        "The associateTags(TagValue[], TagValue[]) method is not supported.");
+  }
 }
--- a/api/src/main/java/org/apache/gravitino/tag/Tag.java
+++ b/api/src/main/java/org/apache/gravitino/tag/Tag.java
@@ -55,6 +55,26 @@ public interface Tag extends Auditable {
   Map<String, String> properties();
 
   /**
+   * @return The assignment value constraint of the tag. The default means the tag has no value
+   *     restriction and may be assigned without values or with any non-empty assignment value.
+   */
+  default TagValueConstraint valueConstraint() {
+    return TagValueConstraint.anyValue();
+  }
+
+  /**
+   * Returns the tag assignment in the current metadata-object context.
+   *
+   * <p>The return value is present only when the tag is returned for a metadata object. A present
+   * assignment without values means the tag is directly or effectively assigned without values.
+   *
+   * @return The assignment in context, or empty when no assignment context exists.
+   */
+  default Optional<TagAssignment> assignment() {
+    return Optional.empty();
+  }
+
+  /**
    * Check if the tag is inherited from a parent object or not. If the tag is inherited, it will
    * return true, if it is owned by the object itself, it will return false.
    *
@@ -88,5 +108,15 @@ public interface Tag extends Auditable {
      * @return The list of objects that are associated with this tag.
      */
     MetadataObject[] objects();
+
+    /**
+     * List objects associated with this tag and assignment value.
+     *
+     * @param value The exact assignment value to filter by.
+     * @return The list of objects that are associated with this tag and value.
+     */
+    default MetadataObject[] objects(String value) {
+      throw new UnsupportedOperationException("The objects(value) method is not supported.");
+    }
   }
 }
--- a/api/src/main/java/org/apache/gravitino/tag/TagAssignment.java
+++ b/api/src/main/java/org/apache/gravitino/tag/TagAssignment.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.tag;
+
+import com.google.common.base.Preconditions;
+import java.util.Arrays;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.annotation.Evolving;
+
+/** Represents a tag assignment in a metadata-object context. */
+@Evolving
+public final class TagAssignment {
+
+  private static final TagAssignment NO_VALUE = new TagAssignment(new String[0]);
+
+  private final String[] values;
+
+  private TagAssignment(String[] values) {
+    this.values = values.clone();
+  }
+
+  /**
+   * Creates an assignment with no value.
+   *
+   * @return The tag assignment.
+   */
+  public static TagAssignment noValue() {
+    return NO_VALUE;
+  }
+
+  /**
+   * Creates an assignment with values.
+   *
+   * @param values The assignment values.
+   * @return The tag assignment.
+   */
+  public static TagAssignment ofValues(String... values) {
+    Preconditions.checkArgument(
+        values != null && values.length > 0, "Assignment values must not be null or empty");
+    for (String value : values) {
+      Preconditions.checkArgument(
+          StringUtils.isNotBlank(value), "Assignment values must not contain blank values");
+    }
+
+    return new TagAssignment(values);
+  }
+
+  /**
+   * @return True if the assignment has one or more values.
+   */
+  public boolean hasValues() {
+    return values.length > 0;
+  }
+
+  /**
+   * @return The assignment values. Empty when this assignment has no value.
+   */
+  public String[] values() {
+    return values.clone();
+  }
+
+  /**
+   * Compares this assignment with another object.
+   *
+   * @param o The object to compare.
+   * @return True if the object is equal to this assignment.
+   */
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof TagAssignment)) {
+      return false;
+    }
+
+    TagAssignment that = (TagAssignment) o;
+    return Arrays.equals(values, that.values);
+  }
+
+  /**
+   * @return The hash code of this assignment.
+   */
+  @Override
+  public int hashCode() {
+    return Arrays.hashCode(values);
+  }
+
+  /**
+   * @return The string representation of this assignment.
+   */
+  @Override
+  public String toString() {
+    return "TagAssignment{" + "values=" + Arrays.toString(values) + "}";
+  }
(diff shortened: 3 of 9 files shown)


api/src/main/java/org/apache/gravitino/tag/Tag.java
```diff
@@ -54,6 +54,26 @@ public interface Tag extends Auditable {
    */
   Map<String, String> properties();
 
+  /**
+   * @return The allowed assignment values of the tag. {@link Optional#empty()} means the tag is
+   *     unrestricted. A present empty array means the tag only accepts valueless assignments.
+   */
+  default Optional<String[]> allowedValues() {
+    return Optional.empty();
+  }
```
jerryshao, line 63: I think the definition is vague and hard to use. You use `none`, empty arrays, and value arrays to cover three scenarios.

We'd better have a clear/separate definition for 3 scenarios.

api/src/main/java/org/apache/gravitino/tag/TagOperations.java
```diff
@@ -70,6 +70,22 @@ public interface TagOperations {
...
+   * @param comment The comment of the tag.
+   * @param properties The properties of the tag.
+   * @param allowedValues The allowed assignment values of the tag.
+   * @return The created tag.
+   * @throws TagAlreadyExistsException If the tag already exists.
+   */
+  default Tag createTag(
+      String name, String comment, Map<String, String> properties, String[] allowedValues)
+      throws TagAlreadyExistsException {
+    return createTag(name, comment, properties);
```
jerryshao: Are you going to change this in the next PR?

api/src/main/java/org/apache/gravitino/tag/Tag.java
```diff
@@ -54,6 +54,26 @@ public interface Tag extends Auditable {
...
+
+  /**
+   * Returns the assignment values of the tag in the current metadata-object context.
+   *
+   * <p>The return value is present only when the tag is returned for a metadata object. A present
+   * empty array means the tag is directly or effectively assigned without values.
```
jerryshao: Also here. We'd better have a clear definition to separate different scenarios.

api/src/main/java/org/apache/gravitino/tag/TagValue.java
```diff
@@ -0,0 +1,151 @@
...
+  public TagValue() {}
+
+  /**
+   * Creates a tag assignment value.
+   *
+   * @param name The tag name.
+   * @param value The assignment value. Null means a valueless tag assignment.
+   */
+  public TagValue(String name, @Nullable String value) {
+    this.name = name == null ? "" : name;
```
jerryshao: In which scenario will the name be empty?

api/src/main/java/org/apache/gravitino/tag/TagValue.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /**
+   * Sets the tag name.
+   *
+   * @param name The tag name.
+   */
+  public void setName(String name) {
+    this.name = name == null ? "" : name;
+  }
```
jerryshao: Any reason to make this object mutable?

api/src/main/java/org/apache/gravitino/tag/TagValue.java
```diff
@@ -0,0 +1,151 @@
...
+  public String getValue() {
+    return value;
+  }
+
+  /**
+   * @return The optional assignment value.
+   */
+  public Optional<String> value() {
+    return Optional.ofNullable(value);
+  }
```
jerryshao: What's the purpose of having two different getters?

roryqi:
@jerryshao I have addressed the comments. Could u take a look again?

api/src/main/java/org/apache/gravitino/tag/Tag.java
roryqi: Updated this API to use TagValueConstraint instead of Optional<String[]>. It now models the three definition-side cases explicitly as ANY_VALUE, WITHOUT_VALUES, and ALLOWED_VALUES, so callers no longer need to infer semantics from Optional.empty() or empty arrays.

api/src/main/java/org/apache/gravitino/tag/Tag.java
roryqi: Updated this part as Tag.assignment(), returning Optional<TagAssignment>. Optional.empty() means the tag has no metadata-object assignment context; TagAssignment.withoutValues() means assigned without values; TagAssignment.ofValues(...) means assigned with one or more values.

api/src/main/java/org/apache/gravitino/tag/TagOperations.java
roryqi: Changed in this PR rather than deferring it. TagOperations.createTag now accepts TagValueConstraint instead of a raw String[]; the default implementation only delegates for null/anyValue and rejects non-default constraints unless the concrete implementation supports them.

api/src/main/java/org/apache/gravitino/tag/TagValue.java
roryqi: There is no valid scenario for an empty name. The empty default was only to make the first draft easy to deserialize, but this API type no longer handles JSON deserialization directly. TagValue now uses a private constructor and static factories that validate non-blank tag names.

api/src/main/java/org/apache/gravitino/tag/TagValue.java
roryqi: No reason to keep it mutable. I changed TagValue to an immutable API value object: final fields, private constructor, static factories, and no setters. JSON deserialization is handled by the request DTO layer instead of this public API type.

api/src/main/java/org/apache/gravitino/tag/TagValue.java
roryqi: Removed the duplicate bean-style getters/setters from the API type. TagValue now exposes only name() and value(); the DTO request wrapper owns the Jackson-facing shape.

api/src/main/java/org/apache/gravitino/tag/TagValueConstraint.java
```diff
@@ -0,0 +1,149 @@
...
+    return WITHOUT_VALUES;
+  }
+
+  /**
+   * Creates a constraint that only accepts values from the allowed value list.
+   *
+   * @param allowedValues The allowed assignment values.
+   * @return The value constraint.
+   */
+  public static TagValueConstraint ofAllowedValues(String... allowedValues) {
```
jerryshao, line 87: How about `withAllowedValues`?

roryqi: I prefer keeping `ofAllowedValues` here because this is a static factory for an immutable value object, not a builder-style setter. `withAllowedValues` matches the DTO/PO builder pattern better, while `ofAllowedValues` is consistent with `TagValue.of(...)` and `TagAssignment.ofValues(...)`.

api/src/main/java/org/apache/gravitino/tag/Tag.java
```diff
@@ -54,6 +54,25 @@ public interface Tag extends Auditable {
    */
   Map<String, String> properties();
 
+  /**
+   * @return The assignment value constraint of the tag.
+   */
+  default TagValueConstraint valueConstraint() {
+    return TagValueConstraint.anyValue();
+  }
```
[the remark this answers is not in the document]
roryqi: The default is intentional. `ANY_VALUE` means no assignment-value restriction, so it allows both valueless assignments and valued assignments. It does not mean the tag requires a value. I clarified the Javadocs to make this explicit.

api/src/main/java/org/apache/gravitino/tag/SupportsTags.java
```diff
@@ -54,17 +54,30 @@ public interface SupportsTags {
...
   /**
-   * Associate tags to the specific object. The tagsToAdd will be added to the object, and the
-   * tagsToRemove will be removed from the object. Note that: 1) Adding or removing tags that are
-   * not existed will be ignored. 2) If the same name tag is in both tagsToAdd and tagsToRemove, it
-   * will be ignored. 3) If the tag is already associated with the object, it will throw {@link
-   * TagAlreadyAssociatedException}
+   * Associate valueless tags to the specific object. The tagsToAdd will be added to the object, and
+   * the tagsToRemove will be removed from the object. Missing tags are ignored. If the same tag is
+   * in both tagsToAdd and tagsToRemove, it will be ignored. Repeated existing assignments are
+   * idempotent.
```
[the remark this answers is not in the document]
roryqi: This Javadoc describes the new API contract. The follow-up storage/server implementation skips already-active assignment values before insert, so repeated assignments are idempotent. The existing String[] API is converted to valueless TagValue and follows the same path.

api/src/main/java/org/apache/gravitino/tag/TagOperations.java
```diff
@@ -70,6 +70,31 @@ public interface TagOperations {
...
+   */
+  default Tag createTag(
+      String name,
+      String comment,
+      Map<String, String> properties,
+      TagValueConstraint valueConstraint)
+      throws TagAlreadyExistsException {
+    if (valueConstraint == null || TagValueConstraint.anyValue().equals(valueConstraint)) {
+      return createTag(name, comment, properties);
+    }
```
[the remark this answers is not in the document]
roryqi: Same intent here: the 3-arg createTag has no value constraint, so it maps to the unrestricted constraint (`anyValue`). `withoutValues()` is a stricter new constraint and should only be supported by implementations that override this overload.

api/src/main/java/org/apache/gravitino/tag/TagValue.java
```diff
@@ -0,0 +1,114 @@
...
+public final class TagValue {
+
+  private final String name;
+
+  @Nullable private final String value;
+
+  private TagValue(String name, @Nullable String value) {
+    this.name = name;
+    this.value = value;
+  }
```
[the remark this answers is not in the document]
roryqi: `TagValue` is not intended to be the Jackson DTO. It is the public API value object. The common/server PR uses a request DTO wrapper for the JSON shape and converts between `{name, value}` and `TagValue`, so keeping `TagValue` immutable with a private constructor is intentional.

api/src/main/java/org/apache/gravitino/tag/TagAssignment.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.common.base.Preconditions;
+import java.util.Arrays;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.annotation.Evolving;
+
+/** Represents a tag assignment in a metadata-object context. */
+@Evolving
+public final class TagAssignment {
+
+  private static final TagAssignment WITHOUT_VALUES = new TagAssignment(new String[0]);
```
jerryshao: How about using `NO_VALUE` for simplicity?

api/src/main/java/org/apache/gravitino/tag/TagAssignment.java
```diff
@@ -0,0 +1,113 @@
...
+  private TagAssignment(String[] values) {
+    this.values = values.clone();
+  }
+
+  /**
+   * Creates an assignment without values.
+   *
+   * @return The tag assignment.
+   */
+  public static TagAssignment withoutValues() {
```
jerryshao: `noValue`?

api/src/main/java/org/apache/gravitino/tag/TagValue.java
```diff
@@ -0,0 +1,114 @@
...
+    return new TagValue(name, value);
+  }
+
+  /**
+   * Creates a valueless tag assignment value.
+   *
+   * @param name The tag name.
+   * @return The tag value.
+   */
+  public static TagValue valueless(String name) {
```
jerryshao: Valueless means "no worth, not important", it doesn't mean no value. You can change to `noValue` for simplicity. or `ofNoValue`.

api/src/main/java/org/apache/gravitino/tag/TagValueConstraint.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.gravitino.annotation.Evolving;
+
+/** Describes what assignment values a tag accepts. */
+@Evolving
+public final class TagValueConstraint {
+
+  private static final TagValueConstraint ANY_VALUE =
+      new TagValueConstraint(Type.ANY_VALUE, new String[0]);
+
+  private static final TagValueConstraint WITHOUT_VALUES =
```
jerryshao: Also here, how about using `NO_VALUE`?

api/src/main/java/org/apache/gravitino/tag/TagValueConstraint.java
```diff
@@ -0,0 +1,151 @@
...
+   * @return The string representation of this constraint.
+   */
+  @Override
+  public String toString() {
+    return "TagValueConstraint{"
+        + "type="
+        + type
+        + ", allowedValues="
+        + Arrays.toString(allowedValues)
+        + "}";
```
jerryshao, line 156: Allowed values will be empty for other two cases, you'd better polish the string to better differentiate different cases.

jerryshao (approved):

api/src/main/java/org/apache/gravitino/tag/TagValueConstraint.java
roryqi: Fixed. Renamed the constraint case from `WITHOUT_VALUES` to `NO_VALUE` and the factory to `noValue()`.

api/src/main/java/org/apache/gravitino/tag/TagAssignment.java
roryqi: Fixed. Renamed the factory to `noValue()`.

api/src/main/java/org/apache/gravitino/tag/TagValue.java
roryqi: Fixed. Renamed `TagValue.valueless(name)` to `TagValue.noValue(name)` and updated the wording to avoid the overloaded meaning of "valueless".

api/src/main/java/org/apache/gravitino/tag/TagValueConstraint.java
roryqi: Fixed. `toString()` now prints only the type for `ANY_VALUE` and `NO_VALUE`; `ALLOWED_VALUES` is the only case that includes the `allowedValues` list.

api/src/main/java/org/apache/gravitino/tag/TagAssignment.java
roryqi: Fixed. Renamed the singleton/type wording to `NO_VALUE`.

PR #12167, opened 2026-07-23
[#12095] feat(auth): reject assumption of unheld active roles with 403

### LINKED ISSUE #12095: [Subtask] Wire X-Gravitino-Active-Roles into the request path with validation

bharos:

Validate the `X-Gravitino-Active-Roles` header against the roles the caller actually holds. A request that names an active role the caller does **not** hold (directly or via a group) is now rejected with `403 Forbidden` before authorization, instead of the previous behavior of silently dropping the unheld name. This validates the *declaration* only (like SQL `SET ROLE`): you can activate only roles you were granted — the object-level access decision is unchanged.

- `GravitinoAuthorizer`: add `findUnheldRoles(...)`, defaulting to empty so authorizers without role membership (e.g. `PassThroughAuthorizer` when authorization is disabled) never block role assumption.
- `JcasbinAuthorizer`: resolve the caller's held role ids (direct + group-inherited) and return the declared names not held; fails closed if membership can't be resolved.
- `GravitinoInterceptionService` (native REST) and `BaseMetadataAuthorizationMethodInterceptor` (Iceberg REST): reject `NAMED` declarations with unheld roles (`403`). `ALL`/`NONE` name nothing and need no check.

#11967 added the narrowing enforcement and #12096 wired the header onto the principal, but a declared role the caller does not hold was silently dropped. This completes subtask #12095 by rejecting it with `403`.

Yes. Sending `X-Gravitino-Active-Roles` with a role the caller does not hold now returns `403 Forbidden` (previously ignored). An absent header, `ALL`, and `NONE` are unaffected; when authorization is disabled, role assumption is never rejected.

New unit tests:
- `TestJcasbinAuthorizer`: `findUnheldRoles` — all declared roles held (direct + group-inherited), a declared-but-unassigned role, and a non-existent role name.
- `TestGravitinoInterceptionService` and `TestIcebergMetadataAuthorizationMethodInterceptor`: a `NAMED` declaration with an unheld role returns `403` and does not proceed to the method.

Verified `test` + `spotlessCheck` for `core`, `server-common`, `server`, and `iceberg-rest-server`.

Part of #11965.


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/server/web/filter/BaseMetadataAuthorizationMethodInterceptor.java
```diff
@@ -158,6 +162,28 @@ public Object invoke(MethodInvocation methodInvocation) throws Throwable {
...
           }
+
+          // Role assumption: reject a NAMED declaration that names roles the caller does not hold
+          // (403); ALL/NONE need no membership check.
+          ActiveRoles activeRoles = authorizationRequestContext.getActiveRoles();
+          if (activeRoles.mode() == ActiveRoles.Mode.NAMED) {
+            Set<String> unheldRoles =
+                GravitinoAuthorizerProvider.getInstance()
+                    .getGravitinoAuthorizer()
+                    .findUnheldRoles(
```
roryqi, line 173: Will it bring extra IO operations?

bharos: Updated the code now.
`findUnheldRoles` now reuses the existing `prefetchUserAndGroupInfo` which is the same single `batchGetAuthSubjectsForUser` query the authorize path already issues. The interceptor now passes one `AuthorizationRequestContext` to both the membership check and the expression evaluator (it previously created a fresh one), so the prefetch is shared and the second call hits the per-request cache.

NAMED header + expression: 0 extra queries.
NAMED header, no expression: 1 query - intrinsic, we have to read role bindings to verify the claim.
ALL / NONE / no header: 0 extra queries, returns immediately.

`testFindUnheldRolesEmptyWhenAllRolesHeld` asserts exactly one `batchGetAuthSubjectsForUser` and zero `MetadataIdConverter.getID` calls across the membership check plus authorize.

roryqi (approved):
LGTM.

Issue #11891, opened 2026-07-03
[Bug report] Lance loadTable throws "Failed to repair table" when concurrent repair-on-load loses the optimistic-lock CAS


yuqi1129:
### Version

main (Lance catalog; also affects branch-1.3)

### Describe what's wrong

`LanceTableOperations.loadTable` repairs a table's stored schema from the underlying Lance dataset on every load via `repairTableMetadata` / `recordCheckedEmptyVersion`, which do an optimistic-locked `EntityStore.update` (CAS on the entity's `currentVersion`). When two loads repair the same table concurrently (e.g. Spark issues parallel `LOAD` during planning + execution), the slower CAS matches zero rows and `TableMetaService.updateTable` throws `IOException("Failed to update the entity")`. `repairTableMetadata` rethrows it as a fatal `RuntimeException("Failed to repair table ...")` → HTTP 500, even though the concurrent update already left the table consistent.

### Error message and/or stacktrace

```
java.lang.RuntimeException: Failed to repair table ...orders...
    at ...LanceTableOperations.repairTableMetadata(LanceTableOperations.java:546)
    at ...LanceTableOperations.loadTable(LanceTableOperations.java:222)
Caused by: java.io.IOException: Failed to update the entity: ...orders...
    at ...TableMetaService.updateTable(TableMetaService.java:239)   // updateResult == 0 (lost CAS)
    at ...RelationalEntityStore.update(RelationalEntityStore.java:160)
    at ...LanceTableOperations.repairTableMetadata(LanceTableOperations.java:530)
```

### How to reproduce

Concurrently load / `SELECT` the same Lance table so two repair-on-load paths race. Deterministic unit reproduction: stub `EntityStore.update` to throw `IOException("Failed to update the entity")` (exactly what a lost CAS produces) and call `loadTable` — it throws `RuntimeException("Failed to repair table")`. Seen in CI: `LanceSparkRESTServiceIT.testSelectFromEmptyTableViaSpark` (and the drop-table case) fail intermittently.

### Proposed fix

Wrap the repair `store.update` (in both `repairTableMetadata` and `recordCheckedEmptyVersion`) in a bounded CAS-retry loop: on the conflict `IOException`, re-read and re-apply the idempotent updater against the latest version, returning a usable table instead of failing.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

PR #12168, opened 2026-07-24
[#12166] improvement(core): enable version-CAS OCC for entity updates (alter half)

### LINKED ISSUE #12166: [Subtask] Implement version-CAS OCC for entity update and soft-delete
### Describe the subtask

Implement real optimistic concurrency control (version CAS) on the Gravitino store's entity write paths, per the OCC requirements in the design doc (PR #12157). Today no entity does a clean version CAS: `alter` (UPDATE) carries `current_version` in its `WHERE` but **freezes** the version for metalake/catalog/schema/topic (only `table` increments) and leans on a fragile full-row compare; `drop` (soft-delete) has **no** version predicate at all (`WHERE id=? AND deleted_at=0`, and the mapper takes only the id).

### Scope

1. **alter — always raise the version.** Every OCC-covered entity's successful UPDATE must set `current_version = old + 1`, and the UPDATE predicate is reduced to `id = ? AND current_version = ? AND deleted_at = 0` (drop the `audit_info`/name/... full-row compare). Frozen today: metalake, catalog, schema, topic; conditional: fileset, policy; already increments: table (still has a full-row `WHERE` to slim).

2. **drop — CAS delete.** Add an expected-version parameter to the soft-delete path; change the SQL to `UPDATE ... SET deleted_at=<now> WHERE id=? AND current_version=? AND deleted_at=0`. (Today `softDelete...ById(Long id)` takes only the id.)

3. **drop — report 0 vs 1 rows.** The soft-delete returns the affected-row count so the service can tell **1 = deleted** from **0 = version changed / already gone**. Map 0 rows by re-reading: a version mismatch -> OCC conflict (retry -> HTTP 409); an already-deleted row -> `NoSuchEntity` / idempotent `false`, per the API contract.

4. **Backward compatibility with existing data.**
   - **No schema migration**: `current_version`/`last_version` already exist (DEFAULT 1). Existing rows keep their value; the first OCC write reads-then-CAS's on it (frozen rows sitting at `current_version=1` bump to 2 on first write — no backfill needed).
   - **Rolling upgrade / mixed nodes**: an old-code node's write does **not** bump the version, so a new-code node's CAS would not detect it. Gate the change (feature flag, or a single-release cutover with a documented note) so a cluster is never half-CAS / half-frozen.
   - **Ordering**: land "always increment version" first, then slim the `WHERE`; never remove the full-row compare while any path still depends on it.
   - **Out of scope of CAS**: `put(overwrite=true)` on create/import/reconcile stays a blind upsert — CAS applies only to update and soft-delete.

### Acceptance criteria

- alter/alter lost-update: two concurrent alters — one wins (`N->N+1`), the other matches 0 rows -> OCC conflict.
- drop CAS: stale-version drop returns 0 rows -> reported as conflict; current-version drop returns 1 row -> success.
- alter/drop: current-version drop wins; the concurrent alter then matches 0 rows (via `deleted_at`) — no lost write, no stale row.
- change-then-change-back (A->B->A) is caught by version, not by JSON byte equality.
- existing rows (`current_version=1`) upgrade cleanly with no migration.
- tests pass on MySQL and PostgreSQL.

### Relevant code

- `POConverters.update*POWithVersion` — version freeze/increment
- `*MetaBaseSQLProvider.update*Meta` (alter `WHERE`) and `softDelete*ById` (drop `WHERE`)
- `*MetaService.delete*` — pass the version, interpret the row count

### Parent issue

yuqi1129:

Makes the Gravitino store's entity **update** path a real optimistic-concurrency (version CAS) operation. Three parts:

1. **Always raise `current_version` on update** for every entity that was freezing it (metalake, catalog, schema, topic, tag, user, group, role). table/view/function already did; fileset/policy stay conditional (they version-history on change).
2. **Slim the `UPDATE ... WHERE`** to `id + current_version + deleted_at` for the always-increment entities, dropping the fragile full-row / JSON-byte compare. fileset/policy keep the full-row compare (their conditional version cannot guard an audit-only concurrent update — proven by an existing OCC test).
3. **model migration**: `model_meta` had no version columns, so it could not do CAS. Adds `current_version`/`last_version` (mysql/postgresql/h2 base schema + 1.3.0->2.0.0 upgrade scripts, `DEFAULT 1`), raises them on `updateModel`, and makes a model-version add also bump `current_version` (adding a version modifies the model).

Today no entity does a clean version CAS: updates froze the version and leaned on a byte-for-byte row compare, which is fragile and, under HA, cannot reliably detect concurrent writes. This is the "alter" half of #12166.

Fix: #12166

No API change. A write that loses a concurrent race now matches 0 rows (surfaced as the existing update-failure path). Storage: two additive columns on `model_meta` (with an upgrade script); no data migration.

New unit tests pin post-update version increments across all entities; a DB-level CAS test (topic) asserts stale-version -> 0 rows, current -> 1 row + bump; a model test asserts insert=1 -> updateModel=2 -> add-version=3. Full `storage.relational.*` suite green on H2.

### Note

This is the **alter** half. A follow-up PR (drop-side version-checked soft-delete) will be based on this branch.

### DIFF
--- a/api/src/main/java/org/apache/gravitino/exceptions/OptimisticLockException.java
+++ b/api/src/main/java/org/apache/gravitino/exceptions/OptimisticLockException.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.exceptions;
+
+import com.google.errorprone.annotations.FormatMethod;
+import com.google.errorprone.annotations.FormatString;
+
+/**
+ * Exception thrown when an optimistic-lock update fails because the entity changed after it was
+ * read.
+ */
+public class OptimisticLockException extends GravitinoRuntimeException {
+
+  /**
+   * Constructs a new exception with the specified detail message.
+   *
+   * @param message the detail message.
+   * @param args the arguments to the message.
+   */
+  @FormatMethod
+  public OptimisticLockException(@FormatString String message, Object... args) {
+    super(message, args);
+  }
+
+  /**
+   * Constructs a new exception with the specified detail message and cause.
+   *
+   * @param cause the cause.
+   * @param message the detail message.
+   * @param args the arguments to the message.
+   */
+  @FormatMethod
+  public OptimisticLockException(Throwable cause, @FormatString String message, Object... args) {
+    super(cause, message, args);
+  }
+}
--- a/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
+++ b/catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
@@ -50,6 +50,7 @@ import org.apache.gravitino.connector.SupportsSchemas;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
 import org.apache.gravitino.exceptions.NoSuchSchemaException;
 import org.apache.gravitino.exceptions.NoSuchTableException;
+import org.apache.gravitino.exceptions.OptimisticLockException;
 import org.apache.gravitino.exceptions.TableAlreadyExistsException;
 import org.apache.gravitino.lance.common.ops.gravitino.LanceDataTypeConverter;
 import org.apache.gravitino.lance.common.utils.LanceConstants;
@@ -65,7 +66,6 @@ import org.apache.gravitino.rel.expressions.sorts.SortOrder;
 import org.apache.gravitino.rel.expressions.transforms.Transform;
 import org.apache.gravitino.rel.indexes.Index;
 import org.apache.gravitino.storage.IdGenerator;
-import org.apache.gravitino.storage.relational.service.TableMetaService;
 import org.apache.gravitino.utils.PrincipalUtils;
 import org.lance.Dataset;
 import org.lance.ReadOptions;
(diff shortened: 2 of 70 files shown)


roryqi:
It looks good to me overall. The structure of storage isn't good. It is oriented process programming. We can't define interface to extract common code. There are fragiled and duplicated code. It would better that we can have object hierarchy layer. Now they are tool classes and we lacked constraints.
I think it would be better to have some interfaces at least. We shouldn't be stuck in the old design.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
```diff
@@ -564,27 +564,18 @@ private Table repairTableMetadata(NameIdentifier ident, Column[] columns, long d
...
+   * serialization errors, etc.) are not conflicts and fail fast.
    */
   private TableEntity updateTableWithCasRetry(
       NameIdentifier ident, Function<TableEntity, TableEntity> updater) throws IOException {
-    IOException lastConflict = null;
+    OptimisticLockException lastConflict = null;
     for (int attempt = 1; attempt <= REPAIR_UPDATE_MAX_ATTEMPTS; attempt++) {
       try {
         return store.update(ident, TableEntity.class, Entity.EntityType.TABLE, updater);
-      } catch (IOException e) {
```
lasdf1234, line 579: I have a question. Why does Lance need to retry? Then, in other cases, would there be a need for retrying? What are the criteria for determining a retry?

api/src/main/java/org/apache/gravitino/exceptions/OptimisticLockException.java
```diff
@@ -16,35 +16,37 @@
...
 import com.google.errorprone.annotations.FormatString;
-import org.apache.gravitino.exceptions.ConnectionFailedException;
 
-/** Indicates that a KMS backend rejected or could not resolve configured credentials. */
-public class KmsAuthenticationException extends ConnectionFailedException {
+/**
+ * Exception thrown when an optimistic-lock update fails because the entity changed after it was
+ * read.
+ */
+public class OptimisticLockException extends GravitinoRuntimeException {
```
lasdf1234, line 29: Adding this class is merely to better reflect errors or to avoid certain issues? So, if it's an error of this kind, are there any further steps to take?

yuqi1129:
> The structure of storage isn't good. It is oriented process programming. We can't define interface to extract common code. There are fragiled and duplicated code. It would better that we can have object hierarchy layer. Now they are tool classes and we lacked constraints.
> I think it would be better to have some interfaces at least. We shouldn't be stuck in the old design.

This should be another problem. `EntityStore` is designed for key-value storage specifically originally. The current JDBC entity store was introduced later, but the interface has not been adjusted accordingly. 

I'm not entirely sure whether we should do this refactor in this release.  @roryqi do you have such a plan?

roryqi:
> > The structure of storage isn't good. It is oriented process programming. We can't define interface to extract common code. There are fragiled and duplicated code. It would better that we can have object hierarchy layer. Now they are tool classes and we lacked constraints.
> > I think it would be better to have some interfaces at least. We shouldn't be stuck in the old design.
> 
> This should be another problem. `EntityStore` is designed for key-value storage specifically originally. The current JDBC entity store was introduced later, but the interface has not been adjusted accordingly.
> 
> I'm not entirely sure whether we should do this refactor in this release. @roryqi do you have such a plan?

We can define some interfaces first. We don't need to refactor totally now.  Can we define a update interface for every service. Extract the common logic.
We can do the work step by step.

yuqi1129:
> > > The structure of storage isn't good. It is oriented process programming. We can't define interface to extract common code. There are fragiled and duplicated code. It would better that we can have object hierarchy layer. Now they are tool classes and we lacked constraints.
> > > I think it would be better to have some interfaces at least. We shouldn't be stuck in the old design.
> > 
> > 
> > This should be another problem. `EntityStore` is designed for key-value storage specifically originally. The current JDBC entity store was introduced later, but the interface has not been adjusted accordingly.
> > I'm not entirely sure whether we should do this refactor in this release. @roryqi do you have such a plan?
> 
> We can define some interfaces first. We don't need to refactor totally now. Can we define a update interface for every service. Extract the common logic. We can do the work step by step.

I will create an issue to track the problem you mentioned.

api/src/main/java/org/apache/gravitino/exceptions/OptimisticLockException.java
yuqi1129: The client can judge the detailed problem by the exception and choose to retry or just to stop the operations. For example, if a user wants to update a table and gets an `OptimisticLockException`, then he can reload the table and then retry.

catalogs/catalog-lakehouse-generic/src/main/java/org/apache/gravitino/catalog/lakehouse/lance/LanceTableOperations.java
yuqi1129: Good question. The retry behavior in Lance predates this PR and was introduced to fix #11891/#11892.
 
LanceTableOperations.loadTable may perform an internal repair-on-load update. When multiple clients load the same table concurrently, they can read the same metadata version and attempt the same repair. One update wins the CAS, while the others receive an optimistic-lock conflict even though the metadata has already been repaired correctly.
  
Retrying is safe here because the repair updater is idempotent and each retry re-reads the latest entity before reapplying the update. It also prevents an ordinary table load from failing because of an internal metadata-repair race.
  
We should not retry every optimistic-lock conflict globally. A retry is appropriate only when the operation is idempotent, is recomputed from the latest state, has no non-idempotent external side effects, and uses a
bounded retry/backoff policy. User-initiated alter operations generally do not meet these conditions, so we return HTTP 409 and let the caller reload the latest state and decide whether to retry. Other internal operations can use the same approach if they satisfy these criteria

roryqi:
> > > > The structure of storage isn't good. It is oriented process programming. We can't define interface to extract common code. There are fragiled and duplicated code. It would better that we can have object hierarchy layer. Now they are tool classes and we lacked constraints.
> > > > I think it would be better to have some interfaces at least. We shouldn't be stuck in the old design.
> > > 
> > > 
> > > This should be another problem. `EntityStore` is designed for key-value storage specifically originally. The current JDBC entity store was introduced later, but the interface has not been adjusted accordingly.
> > > I'm not entirely sure whether we should do this refactor in this release. @roryqi do you have such a plan?
> > 
> > 
> > We can define some interfaces first. We don't need to refactor totally now. Can we define a update interface for every service. Extract the common logic. We can do the work step by step.
> 
> I will create an issue to track the problem you mentioned.

Could u create some utils or helper to reduce duplicated code for this pull request?

yuqi1129:
@jerryshao This is the first part for OCC improvement.

jerryshao:
I would suggest that instead of fixing all the updates in one PR, it is better to fix one by one entity. For example, fixing all the metalake-related operations in one PR to make sure all the metalake operations are transactional without treelock. Then, move on to another entity type.

That will be easier to review and understand the context. What do you think?

yuqi1129:
> I would suggest that instead of fixing all the updates in one PR, it is better to fix one by one entity. For example, fixing all the metalake-related operations in one PR to make sure all the metalake operations are transactional without treelock. Then, move on to another entity type.
> 
> That will be easier to review and understand the context. What do you think?

It makes sense. The only problem is that there will be several PRs to be raised. I will try to split into small ones to make them easier to merge.

yuqi1129:
Close this one as it has been split into several parts.

PR #12171, opened 2026-07-24
[#12170] fix(core): don't hold cache locks across the backend DB call in batchListEntitiesByRelation

### LINKED ISSUE #12170: [Bug report] Mixed list+create workload collapses: batchListEntitiesByRelation holds cache locks across the backend DB call
### Version

main (also affects 1.3.x)

### Describe what's wrong

With authorization and the entity cache enabled, running `listTables` and `createTable` concurrently on a large schema makes creates very slow. Each workload on its own is fine — only the mix is affected.

Measured on a ~10k-table schema (weighted mixed load), `create_table` p50:

| Concurrency | expected (1.2) | observed |
| --- | --- | --- |
| 30 users | ~1.5s | 37s |
| 50 users | ~1.1s | 54s |
| 100 users | ~1.0s | 204s |

Under a heavier run (25 concurrent listers + 25 creators), creates stop completing entirely — they block server-side, waiting on a cache lock that the lists hold. Reads (`loadTable`) stay fast throughout.

### Root cause

`RelationalEntityStore.batchListEntitiesByRelation` runs the backend DB query **inside** `cache.withMultipleKeyCacheLock(...)`, so it holds the cache segment locks for the whole batch **across the DB round-trip**.

List authorization preloads owners for the entire page (`MetadataAuthzHelper.preloadOwner` → this method), which grabs most of the cache's segment locks for the full DB latency. Meanwhile `createTable` → `insertRelation` → `CaffeineEntityCache.invalidate` needs one of those locks, so every create waits for the list's entire DB call. Raising `gravitino.cache.lockSegments` does not help (a larger list just grabs proportionally more segments).

### How to reproduce

1. Enable authorization and the entity cache.
2. Create a schema with a large number of tables.
3. Run `listTables` and `createTable` concurrently against it.
4. `create_table` latency spikes (and hangs under sustained lists) while lists are in flight.

### Potential fix

Don't hold a cache lock across the DB call. Read cache-aside instead: check the cache lock-free, run the backend query with no lock held, then populate (each `put` takes its own short per-key lock). This keeps individual list/get paths correct while letting concurrent creates proceed.

Draft PR: #12171.

### Willingness to contribute

Yes — PR attached.

bharos:

Fixes #12170.

`batchListEntitiesByRelation` runs the backend batch-list inside `cache.withMultipleKeyCacheLock(...)`, so it holds the cache segment locks across the DB call. List authorization preloads owners for the whole page, so one list grabs a near-global set of locks for the full DB latency, and a concurrent `createTable` invalidation that needs one of those locks blocks until the list's DB call finishes (and can be starved under sustained lists).

This switches the method to cache-aside: read lock-free with `getIfPresent`, run the backend query with no lock held, then populate each key under its own short per-key lock (guarded against overwriting a concurrent invalidation — see Cache safety). The multi-key lock and the now-unused `EntityCacheKey` import are removed.

### Cache safety

- **Unchanged:** writes still invalidate after committing; single-key reads (`get`, single `listEntitiesByRelation`, `getEntityByRelation`) still load under their per-key lock; `put` stays atomic per key.
- **Guarded:** the batch read populates cache-aside (not atomically), but a per-stripe generation guard prevents a stale write. The reader snapshots the relation's generation right after the miss (before the backend call) and, under the per-key lock, skips the `put` if the generation changed during the backend read — leaving the slot absent to be rebuilt on the next read (a miss, never a stale hit). Two cases:
  - **Direct invalidate** (the key `K` itself is invalidated): the generation bump runs under `K`'s stripe lock — the same lock the populate takes — so they can't interleave. Fully closed, zero window.
  - **Cascade invalidate** (`K` is evicted as a side-effect of invalidating a *related* key — e.g. an owner change evicts every co-owned table's owner entry through the reverse index): that eviction runs under the *related* key's lock, not `K`'s, so it can still race `K`'s populate. The guard bumps `K`'s generation on eviction too, closing all but the narrow gap between the generation check and the write. This race isn't introduced here — that cascade path never held `K`'s populate lock before this PR either; the guard only shrinks it.

  The counter is per-stripe (not per-key) to reuse the existing striping and stay bounded; a shared-stripe collision can only cause a benign false-skip (a miss), never a stale hit.
- **Still useful:** only the lock scope changed, so cache hits and population are the same.

No.

New test blocks the backend mid-call and asserts a concurrent same-key `invalidate` still completes — it fails on the old code and passes here. A second test reproduces the exact reader/writer ordering from review (block the backend, invalidate the key mid-call, release, and assert the stale result is not subsequently cached), plus a control that a normal populate still caches; `SegmentedLock` gets unit tests for the generation counter. Also load-tested on a large schema (25 concurrent listers + 25 creators): `create_table` went from 0 completions in 5 min to ~1,450, with list latency unchanged.

A separate, pre-existing list-latency item (the per-object DB work list authorization does even without concurrency) is out of scope here.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
@@ -102,4 +102,40 @@ public interface SupportsRelationEntityCache {
       Entity.EntityType type,
       SupportsRelationOperations.Type relType,
       List<E> entities);
+
+  /**
+   * Returns an opaque generation token for the cache slot that holds the given relation. Callers
+   * that populate the cache from a backend using a cache-aside pattern should snapshot this token
+   * before reading the backend (right after observing a cache miss) and pass it to {@link
+   * #putIfNotInvalidatedSince}, so a concurrent invalidation that happens during the backend read
+   * is detected and the possibly-stale result is not written.
+   *
+   * @param ident the name identifier
+   * @param type the entity type
+   * @param relType the relation type
+   * @return the current generation token for the relation's cache slot
+   */
+  long relationGeneration(
+      NameIdentifier ident, Entity.EntityType type, SupportsRelationOperations.Type relType);
+
+  /**
+   * Puts a list of related entities into the cache only if the relation has not been invalidated
+   * since the given generation token was snapshotted. If the relation was invalidated in the
+   * meantime, the put is skipped so a stale backend result cannot overwrite a concurrent
+   * invalidation; the entry is left absent and rebuilt on the next read.
+   *
+   * @param ident The name identifier of the entity to cache the related entities for
+   * @param type The type of the entity to cache the related entities for
+   * @param relType The relation type to cache the related entities for
+   * @param entities The list of related entities to cache
+   * @param sinceGeneration The generation token snapshotted via {@link #relationGeneration} before
+   *     the backend read
+   * @param <E> The class of the related entities
+   */
+  <E extends Entity & HasIdentifier> void putIfNotInvalidatedSince(
+      NameIdentifier ident,
+      Entity.EntityType type,
+      SupportsRelationOperations.Type relType,
+      List<E> entities,
+      long sinceGeneration);
 }
--- a/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
+++ b/core/src/main/java/org/apache/gravitino/cache/CaffeineEntityCache.java
@@ -211,6 +211,7 @@ public class CaffeineEntityCache extends BaseEntityCache {
           // bypasses the removal listener; reverseIndex.remove(key) then cleans up only this
           // entry's own bookkeeping (entityToReverseIndexMap + reverseIndex references to it).
           cacheData.invalidate(key);
+          segmentedLock.incrementGeneration(key);
           reverseIndex.remove(key);
           cacheIndex.remove(key.toString());
           return true;
@@ -293,6 +294,7 @@ public class CaffeineEntityCache extends BaseEntityCache {
     segmentedLock.withGlobalLock(
         () -> {
           cacheData.invalidateAll();
+          segmentedLock.incrementAllGenerations();
         });
   }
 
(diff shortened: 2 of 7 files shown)


nevzheng:
@bharos Thanks for the PR. supportive of the direction.

Could you update the PR description (or add a short comment) with a cache safety argument covering:

1. Current state. how batchListEntitiesByRelation uses the cache/locks today
2. Problem — why that blocks writers under mixed list+create
3. Proposed solution — cache-aside, no lock across the backend call
4. Safety — why this is safe given our consistency model (what we keep / what we relax; any residual races)
5. Efficacy — why the cache remains useful / evidence from the regression test + load results

That’ll help reviewers reason about the architecture change, not just the lock removal. Thanks

bharos:
Thanks @nevzheng. I've added a **Cache safety argument** section to the PR description covering current state, problem, solution, safety (what we keep/relax + the bounded per-key staleness window and why it stays within the existing consistency model), and efficacy (regression test + load results). Also fixed the spotless formatting that was failing CI. PTAL.

nevzheng (approved):
lgtm

nevzheng (changes requested):
I considered this a bit more, and there’s one specific concurrency case I’d like you to consider in the safety/consistency analysis. Could you explain how the current design would behave in this situation?
```
Batch reader                    Concurrent writer
------------                    -----------------
Cache miss for key K
Read old owner A from DB
                                Change owner A → B
                                Commit
                                Invalidate K
Populate K with owner A
```
Here’s my specific concern: the batch reader could repopulate stale owner A after the corresponding invalidation has already completed. A later single-key listEntitiesByRelation call would then return A from the cache instead of reading owner B from the backend. With the default TTL, this could persist for up to one hour, and another invalidation is not guaranteed. Since OWNER_REL participates in authorization, this may have consistency and access-control implications.
A possible solution would be conditional cache population—for example, using a per-key generation/version check—so the backend result is cached only if the key has not been invalidated since the initial cache miss. This would retain the main benefit of avoiding cache locks across backend I/O while preventing a concurrent invalidation from being overwritten.
It would also be helpful to extend the regression test to cover this exact ordering: allow invalidation to complete while the backend call is blocked, release the backend, and verify that the stale result is not subsequently cached.

diqiu50 (commented):
LGTM

bharos:
> Also load-tested on a large schema (25 concurrent listers + 25 creators): `create_table` went from 0 completions in 5 min to ~1,450, with list latency unchanged.

Thanks @nevzheng for the review. addressed in the latest commit, and updated the Cache safety section accordingly.

SegmentedLock now keeps a per-stripe generation counter, bumped on every relation-cache eviction. The batch read snapshots the relation's generation right after the miss (before the backend call) and, under the per-key lock, skips the cache.put if the generation changed ,  so a concurrent invalidation can't be overwritten. On a skip the slot is left absent and rebuilt on the next read (a miss, never a stale hit), so a later single-key listEntitiesByRelation reads fresh from the backend instead of returning stale A.

On your exact ordering:

Direct invalidate (your table): the generation bump runs under the same stripe lock as the populate, so they're fully serialized : the populate either runs before the invalidate (which then evicts it) or sees the changed generation and skips. 

Cascade invalidate (entity K is evicted as a side-effect of invalidating a related key - e.g. an owner change evicts every co-owned table's owner entry through the reverse index): that eviction runs under the related key's lock, not K's, so it can still race K's populate. The guard bumps K's generation on eviction too, closing all but the narrow gap between the guard's generation check and its write. This race isn't introduced here , that cascade path never held K's populate lock on main either; the guard only shrinks it.

Went per-stripe rather than per-key to reuse the existing striping and avoid an unbounded, non-self-pruning per-key map; the only cost is a benign false-skip when unrelated keys share a stripe (a miss, never a stale hit).

Added the ordering test you asked for testBatchListSkipsPopulateWhenInvalidatedDuringBackendCall

PTAL

nevzheng (approved):
LGTM thanks!

core/src/main/java/org/apache/gravitino/cache/SupportsRelationEntityCache.java
```diff
@@ -102,4 +102,40 @@ <E extends Entity & HasIdentifier> void put(
...
+   *
+   * @param ident The name identifier of the entity to cache the related entities for
+   * @param type The type of the entity to cache the related entities for
+   * @param relType The relation type to cache the related entities for
+   * @param entities The list of related entities to cache
+   * @param sinceGeneration The generation token snapshotted via {@link #relationGeneration} before
+   *     the backend read
+   * @param <E> The class of the related entities
+   */
+  <E extends Entity & HasIdentifier> void putIfNotInvalidatedSince(
```
yuqi1129, line 135: It seems those methods are not general enough; a developer trying to implement another entity store cache will have trouble understanding and using them. I have raised a PR to remove all relations from the cache and simplify it; in this way, the entity cache SPI can be a real interface that can be implemented without too much effort. 

So @bharos, if it's convenient, can you help to verify the imperfomance reach your expectations after #12006

bharos: Sure, I will check

PR #12174, opened 2026-07-24
[#12173] improvement(auth): use per-role policy index on JCasbin authorization hot path

### LINKED ISSUE #12173: [Improvement] Use per-role policy index on JCasbin authorization hot path to avoid O(total_policies) enforce

The JCasbin authorization hot path calls `enforcer.enforce(...)` for every privilege probe (`JcasbinAuthorizer#authorizeByJcasbin`, `enforceNarrowed`, `hasDenyPolicy`). In jcasbin 1.99.0, `enforce` **linearly scans every `p` policy line** loaded in the enforcer and evaluates the aviator matcher per line — there is no index by subject/object. Cost is therefore `O(total_policies_in_enforcer)` per probe, and the enforcer accumulates policies for up to `GRAVITINO_AUTHORIZATION_ROLE_CACHE_SIZE` (default 10000) roles across all metalakes on a node.

A standalone benchmark against the real `jcasbin_model.conf` (a single user holding one role, targeting one object) confirms the scan is over **all** policies regardless of the user's role bindings:

| roles | p-rules | `enforce` avg | indexed lookup |
|---|---|---|---|
| 100 | 6,000 | ~1.0 ms | 0.06 us |
| 1,000 | 60,000 | ~12.8 ms | 0.015 us |
| 5,000 | 300,000 | ~51 ms | 0.015 us |
| 10,000 | 600,000 | **~106 ms** | 0.015 us |

One `authorize` call can trigger many probes (schema inheritance chain, per-object list checks, allow + deny enforcers, per-active-role narrowing), multiplying the cost.

### How should we improve?

Use a per-role policy index on the hot path instead of `enforce`, reducing per-probe cost from `O(total_policies)` to `O(roles_per_user)` hash probes. The same approach was previously demonstrated in #10908 / #10930.

Concretely, on the current architecture:

- Extend the `loadedRoles` cache value to carry, alongside the `role_meta.updated_at` version sentinel, a `Map<PolicyKey, Effect>` index (`PolicyKey = type + metadataId + privilege`), built in `loadPolicyByRoleEntity` with DENY-beats-ALLOW within a role.
- Replace the three `enforce` call sites (`authorizeByJcasbin`, `enforceNarrowed`, `hasDenyPolicy`) with index lookups over the user's / active roles; keep cross-role DENY precedence and the OWNER short-circuit to the owner cache.
- Preserve the current version-validation (`role_meta.updated_at`) and active-role narrowing semantics.
- Add a regression/benchmark test guarding the complexity.

Related: #10907, #10908, #10930.

yuqi1129:

Privilege probes on the JCasbin authorization hot path no longer call `enforcer.enforce`. They resolve against a per-role index instead:

- `loadedRoles` now caches `roleId → CachedRolePolicies` (a `role_meta.updated_at` version sentinel plus a `Map<PolicyKey, Effect>` index), built in `loadPolicyByRoleEntity`.
- `authorizeByJcasbin` / the new `resolveRoleEffect`, the active-role narrowing path, and `hasDenyPolicy` all resolve via index lookups over the roles the caller holds (`indexEffect`).
- The enforcer keeps only the user/group→role grouping graph (`g` rows); it no longer stores privilege (`p`) policies. The now-redundant second (`deny`) enforcer, `clearRolePolicies`, and the removal listener in `JcasbinLoadedRolesCache` are removed.
- New types: `PolicyKey`, `Effect`, `CachedRolePolicies`.

Semantics preserved: DENY wins within a role and globally across the caller's roles; active-role narrowing (role assumption) still narrows allows while deny stays global; `role_meta.updated_at` version validation is unchanged; OWNER still short-circuits to the owner cache.

#### Why `JcasbinLoadedRolesCache` no longer needs a removal listener

Policy ownership changes in this PR:

| | Before this PR | After this PR |
|---|---|---|
| `loadedRoles` value | `roleId → updatedAt` | `roleId → CachedRolePolicies(updatedAt, policyIndex)` |
| Privilege-policy storage | JCasbin enforcer `p` rows | `CachedRolePolicies.index` |
| Enforcer responsibility | `g` bindings and `p` policies | `g(user/group, role)` bindings only |
| Work on cache eviction | Delete the separate `p` rows | None; the index is removed with the cache entry |

Previously, evicting the version-only `loadedRoles` entry left a second copy of the role's policies in the enforcers. The removal listener therefore had to clear those `p(roleId, ...)` rows. The recently merged fix correctly changed that cleanup from `deleteRole`, which also removes independently managed `g` bindings, to policy-only cleanup.

This PR removes enforcer `p` rows entirely. The complete privilege index and its version sentinel now live in the same `CachedRolePolicies` cache value, so eviction atomically discards the policy data itself. On the next request, `versionCheckAndLoadRoles` observes the cache miss, reloads the role from the database, and rebuilds the index. There is no second policy copy for a listener to clean.

The enforcer's remaining `g` bindings have a separate lifecycle managed by the user/group role caches. Deleting them when a policy-index entry expires would be incorrect. A residual `g` binding cannot grant access without a corresponding `loadedRoles` index: a missing index resolves to no privilege and is reloaded or rejected by version validation before authorization.

In jcasbin 1.99.0, `enforce` linearly scans **every** policy line loaded in the enforcer and evaluates the aviator matcher per line — `O(total_policies)` per probe — and the enforcer accumulates policies for up to `roleCacheSize` (default 10000) roles across all metalakes on a node. A benchmark against the real `jcasbin_model.conf` (one user, one role, one object) shows the scan is over all policies regardless of the user's bindings:

| roles | p-rules | `enforce` avg |
|---|---|---|
| 1,000 | 60,000 | ~12.8 ms |
| 10,000 | 600,000 | ~106 ms |

The index makes each probe `O(roles_per_user)` hash lookups (sub-microsecond, flat in total policy count). One `authorize` triggers many probes (schema chain, per-object list checks, allow+deny, per-active-role narrowing), so the saving compounds.

Fix: #12173

No.

`./gradlew :server-common:test --tests "org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer" -PskipITs` — 57 tests pass.

- `testPrivilegeProbesUseIndexNotEnforcerPolicies` asserts that the index carries the granted privilege while the enforcer contains no `p` policies.
- `testRoleCacheInvalidationPreservesUserRoleBindings` asserts that invalidating the policy-index entry removes it without deleting the enforcer's independently managed `g` binding.
- The deleted-role/version-miss test asserts that a residual `g` binding grants nothing after its `loadedRoles` index is evicted.

### Optimization approach and scope

Authorization no longer calls JCasbin enforce for privilege probes. allowEnforcer retains only the user/group-to-role (g) relationship graph. Privilege policies are stored in a per-role PolicyKey-to-Effect index in loadedRoles. A request first resolves the roles held by the user, then performs hash lookups only in those roles' indexes; DENY takes precedence globally, while active-role narrowing applies to ALLOW decisions.

This primarily addresses nodes with many cached roles and policies, where the old JCasbin path scanned all loaded p policy lines for every probe. It does not eliminate the cost of a single very large role: that role's complete policy index still needs to be loaded and retained. Likewise, a role shared by many users still has one shared index and the user-role (g) relationship remains a separate scale consideration. These cases are outside this PR's scope.

### DIFF
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/CachedRolePolicies.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/CachedRolePolicies.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.server.authorization.jcasbin;
+
+import java.util.Map;
+
+/**
+ * Cached, indexed privileges of a single role. The {@code updatedAt} timestamp corresponds to the
+ * {@code role_meta.updated_at} column and is used as a version sentinel: if the DB value is newer,
+ * the cached index is stale and must be rebuilt from the role entity.
+ *
+ * <p>The {@code index} maps each {@link PolicyKey} the role grants or denies to its {@link Effect},
+ * with {@link Effect#DENY} already taking precedence over {@link Effect#ALLOW} within the role. A
+ * privilege probe becomes a single {@link Map#get(Object)} instead of a linear scan over every
+ * jcasbin policy line.
+ */
+final class CachedRolePolicies {
+
+  private final long updatedAt;
+  private final Map<PolicyKey, Effect> index;
+
+  CachedRolePolicies(long updatedAt, Map<PolicyKey, Effect> index) {
+    this.updatedAt = updatedAt;
+    this.index = index;
+  }
+
+  long getUpdatedAt() {
+    return updatedAt;
+  }
+
+  Map<PolicyKey, Effect> getIndex() {
+    return index;
+  }
+}
--- a/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/Effect.java
+++ b/server-common/src/main/java/org/apache/gravitino/server/authorization/jcasbin/Effect.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino.server.authorization.jcasbin;
+
+/**
+ * The effect of a single privilege rule in the per-role policy index. {@link #DENY} beats {@link
+ * #ALLOW} both within a single role and across the roles a user holds, mirroring the jcasbin {@code
+ * policy_effect} {@code some(where (p.eft == allow)) && !some(where (p.eft == deny))}.
+ */
+enum Effect {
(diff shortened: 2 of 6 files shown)


yuqi1129:
@roryqi 

roryqi:
I would suggest removing enforcer directly. We can implement it more efficient.

yuqi1129:
@roryqi 
This may need a thorough refactoring. Have you got a plan for this in this release?

roryqi:
> @roryqi This may need a thorough refactoring. Have you got a plan for this in this release?

This improvement seems sufficient. Just need to remove some extra code.

yuqi1129:
> > @roryqi This may need a thorough refactoring. Have you got a plan for this in this release?
> 
> This improvement seems sufficient. Just need to remove some extra code.

Okay. I was a little concerned before that a modification of this magnitude would affect the correctness of the authz. If it was planned in this version, we can just start it.

roryqi:
> > > @roryqi This may need a thorough refactoring. Have you got a plan for this in this release?
> > 
> > 
> > This improvement seems sufficient. Just need to remove some extra code.
> 
> Okay. I was a little concerned before that a modification of this magnitude would affect the correctness of the authz. If it was planned in this version, we can just start it.

I don't think that it will bring extra danger to remove enforcer code.

yuqi1129:
Pending this PR, I will take time to remove the enforcer as dicussed.

yuqi1129:
Close it as  #12282 will solve the problem thoroughly.

PR #12175, opened 2026-07-24
[#11968] test(auth): add integration test for X-Gravitino-Active-Roles

### LINKED ISSUE #11968: [Subtask] Apply role narrowing to list filtering + credential vending, with ITs & docs
### Describe the subtask

Verify and test that list filtering (`listTables` / `listNamespaces`) and Iceberg credential vending honor the active role set, including the `WRITE`→`READ` credential downgrade and deny / `NONE` cases. Adds integration tests and user documentation covering the grammar, subtractive + deny-global semantics, ownership behavior, scope limits (native path only; not Hive/JDBC/Ranger pushdown), and transport notes for Spark/Trino.

### Parent issue

#11965

bharos:

Adds `RoleAssumptionAuthorizationIT`, an end-to-end integration test for the `X-Gravitino-Active-Roles` header.

The test user holds two roles, each granting only `USE_CATALOG` on a different catalog, so the result of `listCatalogs` is a direct observation of which roles the server activated for the request. The catalogs are `MODEL` catalogs, so no external metastore or container is required.

| Header value | Expectation |
| --- | --- |
| absent | all held roles active, both catalogs visible |
| `ALL` | same as absent |
| `role_catalog1` | only `catalog1` visible |
| `role_catalog2` | only `catalog2` visible |
| `role_catalog1,role_catalog2` | both catalogs visible |
| `NONE` | no catalog visible |
| `role_not_held` | `403`, message names the unheld role |
| `ALL,role_catalog1` | `400`, message names the header |

The `400` case is asserted over raw HTTP (`java.net.http.HttpClient`) so that the status code itself is verified rather than the client-side exception mapping.

`X-Gravitino-Active-Roles` is covered by unit tests, but nothing exercised the full path from client header through the authentication filter, role validation, and privilege filtering. This closes that gap.

Part of #11968. This PR covers the list-filtering half of the subtask; the Iceberg credential-vending ITs and the user documentation follow in separate PRs, so the issue stays open.

No. Test-only.

The new integration test, verified green in CI on all three `BackendIT` legs (h2/embedded, mysql/deploy, postgresql/deploy). Tests are order-independent, and every client the test creates is closed via try-with-resources.

### DIFF
--- a/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/RoleAssumptionAuthorizationIT.java
+++ b/clients/client-java/src/test/java/org/apache/gravitino/client/integration/test/authorization/RoleAssumptionAuthorizationIT.java
@@ -1,0 +1,171 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


roryqi (approved):
LGTM.

bharos:
Merged the PR. @roryqi thanks for the review

PR #12126, opened 2026-07-21
[#12125] feat(design-docs): Add tag assignment values design

roryqi:
### What changes were proposed in this pull request?

Add a design document for tag assignment values.

The design covers:
- multiple values per tag assignment
- allowedValues on tag creation
- one row per assignment value in tag_relation_meta
- pair-array assignment updates through tagsToAdd and tagsToRemove
- exact value lookup through GET /tags/{tag}/objects?value=...

### Why are the changes needed?

Tag assignments are currently boolean and cannot carry object-specific values. This design defines the API, storage, client, cache, authorization, event, and test plan for assignment-level tag values.

Fix: #12125

### Does this PR introduce _any_ user-facing change?

No code or API behavior changes. This PR only adds a design document.

### How was this patch tested?

- git diff --check apache/main...HEAD
- Checked the design document for stale standalone search API references and placeholders.

PR #12177, opened 2026-07-24
[#12176] feat(design-docs): add policy-on-tag design

### LINKED ISSUE #12176: [FEATURE] Introduce policy-on-tag governance model
### Describe the feature

Introduce the policy-on-tag governance model in Gravitino.

The target model is:

```text
Policy -> Tag -> Metadata Object
```

Policies remain first-class metalake-scoped governance objects. Tags become the object-side governance attachment point. Object policies are read-only derived results computed from effective tags and policy-to-tag relations.

### Motivation

The current direct object policy model requires users to manage tags and policies separately on the same metadata object. This does not scale well across catalogs, schemas, tables, and columns.

Policy-on-tag aligns governance behavior with classification and provides a shared selection layer for TMS and future ABAC use cases.

### Describe the solution

Add a design document under `design-docs/` describing the policy-on-tag model, including target model, APIs, data model, authorization, migration, TMS integration, conflict handling, and task breakdown.

### Additional context

Design doc PR will add `design-docs/policy-on-tag.md`.

roryqi:

Add `design-docs/policy-on-tag.md`, a design document for the policy-on-tag governance model.

The design covers:

1. Target model: `Policy -> Tag -> Metadata Object`.
2. Effective tag based object policy resolution.
3. Policy-to-tag relation storage.
4. REST and client API changes.
5. Authorization, visibility, and audit semantics.
6. Row filter and column mask conflict handling.
7. Tag value overwrite semantics.
8. TMS integration, migration, ABAC evolution, and task breakdown.

The current direct object policy model requires users to manage tags and policies separately on the same metadata object. This does not scale well for governance across catalogs, schemas, tables, and columns.

Policy-on-tag aligns policy assignment with classification and provides a shared selection layer for TMS and future ABAC use cases.

Fix: #12176

No code or API behavior changes. This is a design documentation-only change.

- `git diff --cached --check`
- `git diff --check apache/main...HEAD`
- `./gradlew :docs:build`


design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+
+---
+
+## Goals
+
+1. **Single Object-Side Attachment Point**: Metadata objects receive governance behavior only
+   through tags, not through direct policy attachment.
+2. **Reusable Policy Lifecycle**: Policies remain first-class objects with typed content, enabled
+   state, audit information, and metalake-scoped lifecycle operations.
+3. **Immutable Policy Definitions**: Policy definitions are not modified in place. To change policy
```
jerryshao: Can you explain more about it? I don't think it relates to the topic here.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+3. If a child object has a direct assignment for a tag name, that direct tag becomes the effective
+   source for that tag and overrides inherited assignments with the same tag name.
+4. Policy-on-tag uses tag presence for policy resolution. It does not need assignment values to
+   resolve the phase-1 policy set.
+5. The effective tag set is the de-duplicated result of walking from the object to its ancestors.
+6. Policies bound to effective tags become object policy candidates.
+
+### Policy Supported Object Types
+
+`PolicyContent.supportedObjectTypes()` should mean "the metadata object types on which this policy
```
jerryshao: I think we can deprecate this field. This seems not so useful if we change to policy-tag relationship.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+```
+
+### Policy Mutability
+
+Policy definitions are immutable after creation. Users can create, view, delete, enable, disable,
+and associate policies according to authorization rules, but they cannot update an existing policy
+definition in place.
+
+If policy behavior needs to change, administrators create a new policy and update the relevant
+policy-to-tag association.
```
jerryshao: @mchades can you please confirm this?

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+does not exist.
+
+#### New: `POST /api/metalakes/{metalake}/tags/{tag}/policies`
+
+**Request:**
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `policiesToAdd` | array of string | no | Policy names to associate with the tag. |
+| `policiesToRemove` | array of string | no | Policy names to disassociate from the tag. |
```
jerryshao: Shall we follow or not follow the tag's design principle? I don't think we will have lots of policy to add or remove/

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+POST /api/metalakes/{metalake}/objects/{type}/{fullName}/policies
+```
+
+**Old behavior:** These APIs get or associate policies directly on metadata objects.
+
+**New behavior:** Object policies are derived from tags and are only exposed through the object
+policy lookup API.
+
+**Migration impact:** Callers must move direct policy association workflows to tag assignment and
+policy-to-tag association workflows.
```
jerryshao: What is the behavior now for these two APIs?

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+
+Policy-on-tag makes tags part of the governance control plane. Assigning a high-impact tag can
+change security, maintenance, or retention behavior.
+
+Add two read-only visibility privileges:
+
+```text
+VIEW_TAG
+VIEW_POLICY
+```
```
jerryshao: I don't think we have these two privileges.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,669 @@
...
+
+Rules:
+
+1. If no system iceberg compaction policy appears in the object policy result, TMS does not generate
+   a compaction strategy from policy-on-tag.
+2. If a system iceberg compaction policy appears in the object policy result, TMS uses its content.
+3. TMS does not read direct object policy relations.
+
+### User Process
```
jerryshao, line 772: Where do you mention the inheritance behavior?

design-docs/policy-on-tag.md
mchades: I checked the current implementation. Policy definitions are mutable today: [`PolicyChange.updateContent()`](https://github.com/apache/gravitino/blob/2214a29dcc3d6091b52f0ee4ad8eedee92615393/api/src/main/java/org/apache/gravitino/policy/PolicyChange.java#L51-L59) and [`PolicyOperations.alterPolicy()`](https://github.com/apache/gravitino/blob/2214a29dcc3d6091b52f0ee4ad8eedee92615393/api/src/main/java/org/apache/gravitino/policy/PolicyOperations.java#L90-L100) allow updating an existing policy. So making policy definitions immutable would be a breaking change. If this is intentional, we should explicitly mention how these APIs will be handled.

roryqi:
@jerryshao @mchades Could u help me review this pull request?

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+| Combine row filters with OR semantics | BigQuery combines multiple row-level access policies with OR semantics. | Flexible, but can broaden access and is risky as a default for tag-driven governance. | Rejected |
+| Priority-based resolution | Some systems can choose a winning policy by priority. | Flexible, but effective access becomes harder to reason about and easier to misconfigure. | Rejected |
+| Fail closed on ambiguity | Databricks ABAC blocks access when multiple distinct row filters or column masks apply to the same target. | Safest default, but administrators must fix overlapping tags or policy associations. | **Chosen** |
+
+### Tag Value Semantics
+
+| Approach | Pros | Cons | Decision |
+|----------|------|------|----------|
+| Append values | Can express multiple values for the same tag key on one object. | Makes policy selection ambiguous and can trigger multiple policies for one logical classification dimension. | Rejected |
+| Overwrite values | Keeps each tag key single-valued for one object; aligns with common tag and label systems. | Administrators must use separate tag keys for separate dimensions. | **Chosen** |
```
bharos: Since https://github.com/apache/gravitino/pull/12126 is merged and implementing , and it says tags support multiple values,  could this doc defer to it and drop its own value-semantics rule? Otherwise two designs prescribe incompatible behaviour for single/multiple values for tags ?

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+```text
+ObjectPolicyResolver
+  input: metalake, metadata object
+  output: object policies
+```
+
+Algorithm:
+
+1. Validate that the metadata object exists.
+2. Load effective tags for the object using current inheritance semantics.
```
bharos: ObjectPolicyResolver treats effective-tag resolution as an internal step, but that merge (direct tags + ancestor walk) currently only exists inlined in the REST layer — MetadataObjectTagOperations L186–L203 — so nothing in `core` module can reuse it and the resolver would end up reimplementing it.

Could we extract it into a shared core component instead? Three efforts need it now: this PR, tag assignment values (#12126 already changes that same dedup), and tag-based authorization. Separate copies risk the tag-listing API and policy resolution disagreeing on inherited tags, which is hard to spot.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+such as policy type and `enforced=true`. Users with `VIEW_POLICY` may receive policy names, types,
+definitions, and other policy details. Source tag information should only be returned when the user
+also has `VIEW_TAG`.
+
+Audit events should cover policy lifecycle changes, policy-to-tag relation changes, and
+tag-to-object assignment changes.
+
+### Row Filter and Column Mask Conflict Handling
+
+For row filter and column mask policies, if multiple distinct effective policies of the same kind
```
mchades: Could we clarify which layer owns this fail-closed behavior? Phase 1 is described as policy resolution and leaves engine-specific enforcement out of scope, while the task list includes fail-closed validation.

Does `ObjectPolicyResolver` detect conflicting policies and return a conflict result, or does it return all resolved policies and leave the downstream enforcement layer to deny access? Clarifying this boundary would help avoid implementing the same conflict logic differently in each consumer.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+```json
+{
+  "names": ["iceberg_compaction_standard"]
+}
+```
+
+**Behavior:** Atomically updates policy associations for one tag. The request supports adding and
+removing multiple policies so callers can change the complete relation set without issuing one
+request per policy or exposing an intermediate partial state. The tag and all added policies must
+exist in the same metalake. Adding an already associated policy returns a duplicate association
```
mchades: The atomic batch rationale is clear, but the duplicate-add rule still makes retries unsafe.

For example, a client sends:

```json
{
  "policiesToAdd": ["policyA"],
  "policiesToRemove": []
}
```

The server successfully creates the association, but the `200 OK` response is lost. The client then retries the same request. Under the current contract, the retry returns a duplicate-association error even though the requested state has already been achieved.

In the earlier design review, we agreed that adding an already-associated policy and removing a non-associated policy should both be no-ops, so retries are safe. Could we update the normative behavior here accordingly? Atomicity and idempotency are separate guarantees.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+6. Object policy lookup requires access to the metadata object.
+7. Detailed policy content requires `VIEW_POLICY`; source tag details require `VIEW_TAG`.
+8. `APPLY_POLICY` satisfies policy visibility checks, and `APPLY_TAG` satisfies tag visibility
+   checks, preserving existing role behavior.
+
+Policy enforcement must be independent from policy visibility. A user does not need `VIEW_POLICY`
+for row filters or column masks to take effect. The trusted server-side enforcement path must always
+resolve and enforce applicable policies, even when the end user cannot see policy details.
+
+For object policy APIs, users without `VIEW_POLICY` may receive only a redacted governance summary,
```
mchades: The redaction goal makes sense, but could we make the object-policy response contract normative here? The REST section currently defines `details=false` as policy names and `details=true` as policy objects with content/source information. This paragraph introduces a third redacted-summary shape for callers without `VIEW_POLICY`, but “may receive” and “such as” do not define when it is returned or its exact contents.

For example, what should the same request return for:

- object access only;
- `VIEW_POLICY` without `VIEW_TAG`;
- `VIEW_TAG` without `VIEW_POLICY`;
- both privileges?

Could we define, for each privilege combination and each `details` value, whether policies are returned in full, redacted, filtered out, or rejected, together with the exact fields exposed?

JoegenUSTC:
@roryqi 
Thanks for this design. Two questions about column masking under the phase-1 model:

**1. Can masking vary by user/role?**
Phase-1 resolves policies by tag presence only, so a column's object policy is the same for everyone who queries it. But column masking is normally per-principal (Ranger/Snowflake/Databricks show the same column differently depending on the querying role). Does per-user/role masking only come with the future tag-expression ABAC step, or is there a way to make masking depend on the principal already under policy-on-tag?

**2. What does a masking policy's content look like?**
The doc defines the content for `system_iceberg_compaction`. Is there a planned content shape for a column-mask policy type (mask type, arguments, etc.), or is that intentionally left out for now?

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+**Response:** `200 OK`
+
+```json
+{
+  "policies": [
+    {
+      "name": "iceberg_compaction_standard",
+      "policyType": "system_iceberg_compaction",
+      "enabled": true,
+      "sourceTag": "maintenance_standard",
```
mchades: The response currently has only one `sourceTag`/`sourceObject`, but a single policy can be associated with multiple tags because uniqueness is only defined for each `(policy_id, tag_id)` pair.

For example, if `policyA` is associated with both `pii` and `finance`, and an object has both effective tags, the resolver reaches the same policy through two sources. After deduplication, which source should be returned?

Could we define whether “identical” means the same policy entity or different policies with equivalent content, and preserve all contributing provenance, for example through a deterministically ordered `sources[]`? Otherwise the response may hide a valid association or expose a source that varies with load order, which conflicts with the traceability goal.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+catalog iceberg
+  direct tags: [maintenance_standard]
+
+table iceberg.db.orders
+  direct tags: []
+  inherited tags: [maintenance_standard]
+  effective tags: [maintenance_standard]
+  object policies: [iceberg_compaction_standard]
+  policy source: CATALOG iceberg through tag maintenance_standard
+```
```
jerryshao, line 211: How do you support tag with value feature for policy here? I don't see how you mention this so far. Your example should always reflect this tag with value new feature.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+
+```text
+policy_tag_relation_meta
+  policy_id
+  tag_id
+  audit_info
+  current_version
+  last_version
+  deleted_at
+```
```
jerryshao, line 297: How do you migrate the existing users?

jerryshao: You should document the migration steps.

design-docs/policy-on-tag.md
jerryshao: Clarify this behavior here? @roryqi

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+{
+  "policies": [
+    {
+      "name": "iceberg_compaction_standard",
+      "policyType": "system_iceberg_compaction",
+      "enabled": true,
+      "sourceTag": "maintenance_standard",
+      "sourceTagInherited": true,
+      "sourceObjectType": "CATALOG",
+      "sourceObjectName": "iceberg"
```
jerryshao: What's the meaning of `sourceXXX`? The name is not so clear and not well documented here. We should have a better name.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+
+Algorithm:
+
+1. Validate that the metadata object exists.
+2. Load effective tags for the object using current inheritance semantics.
+3. Load policies associated with those tags in batch.
+4. Drop disabled policies.
+5. Return the object policies and their source tag information.
+
+### Authorization, Visibility, and Audit
```
jerryshao: We should consider the event listener part in the doc.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+| `APPLY_POLICY` | `METALAKE`, `POLICY` | Associate a policy with a tag. |
+
+Recommended authorization rules:
+
+1. Creating and deleting policies keeps existing policy privileges.
+2. Creating and altering tags keeps existing tag privileges.
+3. Associating a policy with a tag is both a policy operation and a tag operation. It requires
+   permission on both the policy and the tag, or metalake owner privilege.
+4. Tag assignment permission alone is not enough to attach a policy to a tag. Policy permission
+   alone is not enough to bind the policy to arbitrary tags.
```
jerryshao: Clarify which privileges should be used for policy association.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+### Row Filter and Column Mask Conflict Handling
+
+For row filter and column mask policies, if multiple distinct effective policies of the same kind
+apply to the same evaluation target, access should be denied. The system should not merge them,
+union them, or choose one by priority in this design.
+
+Multiple identical policies may be deduplicated and enforced once. Administrators must consolidate
+tags or policy associations before access is allowed.
+
+### Tag Value Semantics
```
jerryshao: This part is important, and directly affect how user uses the policy tag feature. I think the behavior should be clarified: policy on tag name, or policy on tag value?

I don't want to defer this to the phase 2. This is a behavior decision; we should think about this. This will affect the whole design.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,692 @@
...
+Recommended evolution:
+
+1. `Policy -> Tag -> Metadata Object`: fixed policy-to-tag relations.
+2. Tag assignment values: object tag assignments can carry values such as `data_domain = finance`.
+3. Tag-expression policies: policies match expressions such as `hasTag("pii")` or
+   `tagValue("data_domain") == "finance"`.
+4. Explainability: APIs and UI show which selector matched and which tag caused the match.
+5. Type-specific validation: future ABAC or policy consumers may reject unsupported combinations of
+   matching policies. This phase returns the derived object policy set and defines fail-closed
+   behavior for ambiguous row filter and column mask enforcement.
```
jerryshao: This is unrelated here, remove this chapter.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: added EffectiveTagResolver to the implementation flow and documented that object tag listing, object policy resolution, tag assignment values, and tag-based authorization should share the same effective tag resolver.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: clarified the boundary. ObjectPolicyResolver selects/deduplicates/preserves provenance only; row-filter and column-mask enforcement consumers own fail-closed decisions through a shared conflict helper.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: the doc now defers tag value storage/update semantics to tag-assignment-values.md and makes phase-1 policy matching explicitly keyed by tag definition, not tag value.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: updated the POST behavior to make adding an already-associated policy and removing a missing association idempotent no-ops. A policy in both arrays now returns 400 Bad Request.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: made the object-policy response contract normative for details=false/details=true and for object-only, policy-read, tag-read, and policy+tag-read authorization combinations.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: examples now include valued tag assignments, and the effective-tag section states that values are preserved as source context but are not phase-1 policy match keys.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: replaced the single source fields with deterministic sources[] and defined deduplication by policy entity/current version while preserving all contributing tag sources.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: expanded the migration process to inventory direct policy relations, map policies to tags, handle valued tag assignments, verify resolved object policies, and retire direct relations after the compatibility window.

roryqi: Addressed in 3f4a238104: documented concrete migration steps, including inventory, policy-to-tag mapping, valued tag assignment decisions, verification, consumer upgrade, and retirement of direct object policy relations.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: clarified the update behavior as atomic and idempotent for retry safety, with 400 Bad Request when the same policy appears in both add/remove arrays.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: renamed sourceXXX to sources[] with fields tagName, tagValues, inherited, objectType, and objectName, and documented the meaning/order of the provenance entries.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: added an Event Listener section for policy-to-tag association pre/success/failure events and clarified that derived object-policy lookup does not create relation events.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: made the behavior explicit. Phase 1 is policy on tag definition/name, not tag value. Tag assignment values are carried in effective tag source context but do not select policies.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: removed the standalone ABAC Evolution chapter and kept only the minimal phase-1/future-match-key notes needed by this design.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: clarified that policy-to-tag association requires both policy-side and tag-side authorization, equivalent to APPLY_POLICY on the policy and APPLY_TAG on the tag, or metalake owner.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: documented PolicyContent.supportedObjectTypes() as deprecated for policy-on-tag and stated that object policy resolution does not filter by it.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: kept the batch tag-scoped association API and clarified why it is atomic and retry-safe with idempotent add/remove semantics.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: expanded Effective Tag Semantics and User Process to describe inherited tags and direct-assignment override behavior, including assignment values from the winning tag assignment.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: added a Policy Mutability section clarifying that existing policy lifecycle/mutation APIs, including alterPolicy and PolicyChange.updateContent, remain supported; only derived object policies are read-only.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: expanded the direct object policy API section with the current list/get/post behavior and the new derived lookup behavior.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: removed the proposed VIEW_TAG/VIEW_POLICY model from the design and task list. The doc now uses existing tag/policy read authorization concepts for response redaction.

design-docs/policy-on-tag.md
roryqi: Addressed in 3f4a238104: removed the separate visibility-privilege goal and folded response visibility into the authorization section using the existing authorization model.

roryqi:
@JoegenUSTC Addressed in 3f4a238104. For phase 1, policy resolution is keyed by tag definition/name and is not principal-dependent. Principal-specific row filter or column mask behavior belongs to the downstream enforcement/policy content design, while ObjectPolicyResolver only selects/deduplicates policies and preserves provenance. The doc now also defines the response visibility matrix and conflict boundary. The exact column-mask content shape remains out of scope for this policy-on-tag design and should be defined with the row-filter/column-mask policy type design.

design-docs/policy-on-tag.md
roryqi: Follow-up in e51e637351: restored the ABAC evolution content as a concise resolver-boundary section instead of the previous broad standalone chapter, so the design still captures the future tag-expression/principal-aware path.

roryqi:
Follow-up pushed in e51e637351: restored ABAC evolution as a concise resolver-boundary section. It keeps the future path for tag values, tag-expression selectors, principal-aware selectors, and explainable provenance without reintroducing the broader standalone ABAC chapter.

roryqi:
Follow-up pushed in e05de8c916: removed the ABAC Evolution Boundary section and the matching task item. The design now keeps policy-on-tag focused on the phase-1 model.

roryqi:
Follow-up pushed in 7491cb7ded: updated the design to make tag value selector part of the policy-to-tag relation. The doc now defines selector types (ANY, VALUE_IN, VALUE_ABSENT), selector storage in policy_tag_relation_meta, REST request/response payloads with valueSelector, resolver selector filtering, source provenance with matchedValues/valueSelector, migration mapping, and task breakdown updates. Verified with ./gradlew :docs:build.

roryqi:
Follow-up pushed in 90e5741542: simplified the first version of tag value matching to one optional `tagValue` field on the policy-to-tag relation. Omitted/null `tagValue` matches by tag presence; non-null `tagValue` matches when the effective tag values contain it. Removed the selector JSON object, selector type enum, and multi-value selector list from the first version.

roryqi:
Thanks for the questions @JoegenUSTC.

This design does not cover ABAC or principal-aware masking. The phase-1 scope is only policy selection through `Policy -> Tag -> Metadata Object`, including tag presence and simple tag value selector matching. The resolved object policy set is the same for the same metadata object and effective tag assignments; it is not varied by querying user or role in this design.

Column-mask policy content shape is also intentionally out of scope here. It should be defined in a separate design for the column-mask policy type and its enforcement integration.

design-docs/policy-on-tag.md
roryqi: Policy lifecycle and mutability are separate from policy-on-tag. Policy definitions remain mutable through the existing lifecycle APIs. Only object policies derived from effective tags are read-only.

design-docs/policy-on-tag.md
roryqi: `PolicyContent.supportedObjectTypes()` is deprecated in the policy-on-tag model. Policy-to-tag relations target tags, and object policy resolution does not filter by this field. Type-specific consumers decide which policy types they consume.

design-docs/policy-on-tag.md
roryqi: Policy definitions remain mutable. Existing operations such as `PolicyOperations.alterPolicy()` and `PolicyChange.updateContent()` continue to be supported. Only derived object policies are read-only.

design-docs/policy-on-tag.md
roryqi: The design uses a tag-scoped batch API so relation updates are atomic and retry-safe. Add and remove operations apply to `(policy, selector)` relations. Duplicate additions and missing removals are no-ops, while the same pair in both lists is rejected.

design-docs/policy-on-tag.md
roryqi: The current list API reads direct relations from the requested object and its parents and marks parent results as inherited. The current get API checks the direct relation first and then parent objects. The current post API atomically updates direct object-policy relations. In the target model, list becomes derived lookup and the direct get and mutation APIs are removed.

design-docs/policy-on-tag.md
roryqi: The design introduces `VIEW_POLICY` and `VIEW_TAG` as explicit privileges on metalake and entity scopes. For backward compatibility, effective `APPLY_POLICY` and `APPLY_TAG` grants imply their corresponding view privileges.

design-docs/policy-on-tag.md
roryqi: Effective tags are resolved by walking the metadata object hierarchy. A direct assignment overrides inherited assignments with the same tag name, including their values. Policy selectors are evaluated against the winning effective assignment.

design-docs/policy-on-tag.md
roryqi: Tag value storage, allowed values, search, and update semantics are defined by `tag-assignment-values.md`. This design consumes the resulting effective values: a null selector matches tag presence, while `TAG_VALUE` matches when the effective assignment contains the configured value.

design-docs/policy-on-tag.md
roryqi: `EffectiveTagResolver` is a shared core component used by object tag listing, object policy resolution, tag assignment values, and tag-based authorization. This keeps inheritance, direct-assignment override, value grouping, and ordering consistent across consumers.

design-docs/policy-on-tag.md
roryqi: `ObjectPolicyResolver` owns selector evaluation, disabled-policy filtering, and policy deduplication. It does not make allow or deny decisions. Row-filter and column-mask consumers fail closed on conflicting policies through a shared conflict helper.

design-docs/policy-on-tag.md
roryqi: The batch operation is idempotent. Adding an existing `(policy, selector)` relation and removing a missing relation are no-ops. Supplying the same pair in both add and remove lists returns `400 Bad Request`.

design-docs/policy-on-tag.md
roryqi: Object access is required for object-policy lookup. Policies without `VIEW_POLICY` are filtered from both response shapes. With `VIEW_POLICY`, `details=false` returns names and `details=true` returns full policy content. `VIEW_TAG` does not affect this response because object-policy results do not expose tag sources. Trusted enforcement always consumes the unfiltered resolver result.

design-docs/policy-on-tag.md
roryqi: Per-object source provenance is not part of the object-policy response. The resolver deduplicates by policy entity and current version, so multiple matching relations that reach the same policy produce one result. Policy-to-tag associations remain inspectable through the dedicated relation APIs.

design-docs/policy-on-tag.md
roryqi: Value-aware matching is represented by the policy-to-tag relation selector. A null selector matches by tag presence. A `TAG_VALUE` selector matches when the effective tag assignment contains its value. The examples cover both inherited valued tags and direct-value override.

design-docs/policy-on-tag.md
roryqi: Migration inventories existing direct relations, maps each policy to administrator-approved tags and selectors, assigns the required tag values, verifies derived object-policy results, upgrades consumers, and retires direct relations after the compatibility window.

design-docs/policy-on-tag.md
roryqi: Source fields are not included in the object-policy response. The response contains only visible policy data, while policy-to-tag relations and selectors are inspected through the dedicated association APIs.

design-docs/policy-on-tag.md
roryqi: Policy-to-tag association has pre-event, success-event, and failure-event listener semantics. Event payloads include the metalake, tag, added and removed policy-selector pairs, actor, request context, and final relations. Derived lookup does not emit relation-change events.

design-docs/policy-on-tag.md
roryqi: Policy-to-tag association requires metalake ownership or both `APPLY_TAG` on the target tag and `APPLY_POLICY` on every referenced policy. `VIEW_TAG` and `VIEW_POLICY` are read-only and do not authorize association changes.

design-docs/policy-on-tag.md
roryqi: A policy-to-tag relation is bound to a tag definition and may carry one selector. A null selector matches by tag presence, while `TAG_VALUE` narrows matching to one effective tag value. The selector is stored in one JSON `selector` column.

design-docs/policy-on-tag.md
roryqi: General ABAC evolution is outside this design. Phase 1 covers fixed policy-to-tag relations with optional tag-presence or `TAG_VALUE` matching. Principal-aware matching, general expressions, and engine-specific enforcement remain out of scope.

design-docs/policy-on-tag.md
roryqi: The design uses single-relation APIs rather than a batch update. `PUT /tags/{tag}/policies/{policy}` creates the relation or updates its selector, and `DELETE` on the same path removes the relation. Both operations are idempotent.

design-docs/policy-on-tag.md
roryqi: The design no longer uses a batch request. Repeating the same single-relation PUT is a no-op, and deleting a missing relation is also a no-op, so both operations are safe to retry.

design-docs/policy-on-tag.md
roryqi: Each single relation create, selector update, and delete operation has pre-event, success-event, and failure-event semantics. The payload identifies one policy, one tag, the previous and requested selector as applicable, the actor, and the request context.

design-docs/policy-on-tag.md
roryqi: Creating, updating, or deleting one policy-to-tag relation requires metalake ownership or both `APPLY_TAG` on the target tag and `APPLY_POLICY` on the target policy. `VIEW_TAG` and `VIEW_POLICY` remain read-only.

design-docs/policy-on-tag.md
```diff
@@ -0,0 +1,777 @@
...
+  CATALOG:iceberg:
+    tags:
+      - name: data_domain
+        values:
+          - finance
+```
+
+---
+
+## Task Breakdown
```
nevzheng, line 856: suggestion. non blocking. Could we organize it into milestones? It could make progress more legible over time. I like framing miltones in terms of outcomes. For example if a group of tasks is complete X/Y/Z will be true. AND XYZ matter for the overall project goal OKR / CUJ.

if we just have a list of tasks, it becomes harder to understand at a glance or high level where the project is.

nevzheng (approved):
LGTM.

Overall the design is reasonable and solid at a high level. We have enough to start .I'm happy to help w/ reviews, organization, design discussion for points of contention. I don't see anything immediately blocking / friction.

design-docs/policy-on-tag.md
roryqi: OK, we can modify this. It will be used for AI coding. It's not suitable for human reviewing. We can modify our design document skills, too.

jerryshao (approved):

PR #12184, opened 2026-07-24
[#12166] improvement(core): OCC via version CAS for entity delete

### LINKED ISSUE #12166: [Subtask] Implement version-CAS OCC for entity update and soft-delete

yuqi1129:
> This PR is the **drop (delete)** half of #12166. The **alter (update)** half is #12168 — this branch is stacked on it, so until #12168 merges the diff here also includes the alter commits; the review scope of *this* PR is the version-checked soft-delete.

**Drop — version-checked soft-delete (this PR)**

- Add a `current_version` parameter to the direct `softDelete<Entity>By...Id` and check it in the `WHERE` (base + PostgreSQL) for 12 entities: topic, table, fileset, view, function, model, metalake, catalog, user, group, role, schema (non-cascade). Each delete reads the PO first, then passes its version, so the CAS protects the read->delete window.
- Semantics: 0 rows -> `return false`; cascade / relation cleanup only when the affected row count > 0.
- Excluded by design: **tag / policy** (single atomic by-name delete, no read->delete window, already count-based -> version CAS adds no value and would change not-found semantics) and **schema's hierarchical cascade** (deletes multiple descendant schemas via a name-batch, so a single-version CAS does not apply; only the non-cascade single-schema delete is versioned).

The **alter half** (always-raise `current_version`, slim `UPDATE ... WHERE` to a version CAS, `model_meta` version-column migration, and dependent-write rollback on conflict) is in **#12168** and appears here only as the branch base.

Deletes previously ignored the version entirely, so a soft-delete removed whatever live row existed regardless of concurrent modification. Under HA this cannot detect a concurrent write in the read->delete window. This is the drop half of #12166.

Fix: #12166

No REST/API change. A delete that loses a concurrent race matches 0 rows and is surfaced through the existing `delete...` -> `false` path; cascade/relation cleanup runs only when the entity row was actually deleted. No storage/schema change in the drop half (the `model_meta` columns are added by the alter half, #12168).

DB-level CAS tests assert stale version -> 0 rows (entity stays) and current version -> 1 row (entity gone) at the mapper level, plus a model service test asserting insert=1 -> updateModel=2 -> add-version=3. The full `storage.relational.*` suite passes on H2.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/CatalogMetaMapper.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/CatalogMetaMapper.java
@@ -89,7 +89,8 @@ public interface CatalogMetaMapper {
   @UpdateProvider(
       type = CatalogMetaSQLProviderFactory.class,
       method = "softDeleteCatalogMetasByCatalogId")
-  Integer softDeleteCatalogMetasByCatalogId(@Param("catalogId") Long catalogId);
+  Integer softDeleteCatalogMetasByCatalogId(
+      @Param("catalogId") Long catalogId, @Param("currentVersion") Long currentVersion);
 
   @UpdateProvider(
       type = CatalogMetaSQLProviderFactory.class,
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/CatalogMetaSQLProviderFactory.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/CatalogMetaSQLProviderFactory.java
@@ -109,8 +109,9 @@ public class CatalogMetaSQLProviderFactory {
     return getProvider().updateCatalogMeta(newCatalogPO, oldCatalogPO);
   }
 
-  public static String softDeleteCatalogMetasByCatalogId(@Param("catalogId") Long catalogId) {
-    return getProvider().softDeleteCatalogMetasByCatalogId(catalogId);
+  public static String softDeleteCatalogMetasByCatalogId(
+      @Param("catalogId") Long catalogId, @Param("currentVersion") Long currentVersion) {
+    return getProvider().softDeleteCatalogMetasByCatalogId(catalogId, currentVersion);
   }
 
   public static String softDeleteCatalogMetasByMetalakeId(@Param("metalakeId") Long metalakeId) {
--- a/core/src/main/java/org/apache/gravitino/storage/relational/mapper/FilesetMetaMapper.java
+++ b/core/src/main/java/org/apache/gravitino/storage/relational/mapper/FilesetMetaMapper.java
@@ -246,7 +246,8 @@ public interface FilesetMetaMapper {
   @UpdateProvider(
       type = FilesetMetaSQLProviderFactory.class,
       method = "softDeleteFilesetMetasByFilesetId")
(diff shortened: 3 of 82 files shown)


yuqi1129:
Close it as we have split it into several parts.

PR #12185, opened 2026-07-24
[Cherry-pick to branch-1.3] [#12024] improvement(common): harden JDBC unsafe-parameter detection against connectionProperties smuggling (#12025)

LuciferYang:

- Replace the config-**value** scan in `JdbcUrlUtils` with a config-**name** scan: every config key, plus each parameter name embedded in DBCP2's `connectionProperties` value.
- Parse `connectionProperties` exactly as DBCP2 does (`Properties.load` after replacing `;`→`\n`); reject a malformed value fail-closed.
- Use `Locale.ROOT` for all case folding.

`JdbcUrlUtils.validateJdbcConfig` guards against dangerous JDBC parameters (e.g. MySQL `autoDeserialize`, an RCE enabler) but only matched config *values* and never inspected DBCP2's `connectionProperties`. An unsafe parameter could be smuggled to the driver as a config key, or inside `connectionProperties` (reachable via `gravitino.bypass.connectionProperties=autoDeserialize=true`, whose prefix is stripped before the map reaches the datasource), fully bypassing the check.

Fix: #12024

No. A configuration that was already unsafe is now rejected earlier with a clear error. The exception message wording changes from "...detected in JDBC URL" to "...detected in JDBC configuration" to reflect that config keys/`connectionProperties` are inspected too, not only the URL.

New unit tests in `TestJdbcUrlUtils` (connectionProperties for MySQL/MariaDB/PostgreSQL, config-key detection, newline / `\uXXXX` / double-URL-encoding smuggling, whole-name-vs-substring and value-vs-name negative controls, Turkish-locale folding, malformed-value rejection) and an end-to-end `TestDataSourceUrlValidation` case through `DataSourceUtils.createDataSource`. `./gradlew :common:test :catalogs:catalog-jdbc-common:test -PskipITs` and spotless pass.

---

Cherry-pick of #12025 to `branch-1.3`; replaces the auto-generated #12183 (its GitHub auto cherry-pick committed conflict markers). The only manual resolution was in `TestDataSourceUrlValidation.java`, taking #12025's strengthened cause-message assertions over the older type-only `assertThrows` on `branch-1.3`.

### DIFF
--- a/catalogs/catalog-jdbc-common/src/test/java/org/apache/gravitino/catalog/jdbc/utils/TestDataSourceUrlValidation.java
+++ b/catalogs/catalog-jdbc-common/src/test/java/org/apache/gravitino/catalog/jdbc/utils/TestDataSourceUrlValidation.java
@@ -53,8 +53,16 @@ public class TestDataSourceUrlValidation {
     properties.put(JdbcConfig.USERNAME.getKey(), "test");
     properties.put(JdbcConfig.PASSWORD.getKey(), "test");
 
-    Assertions.assertThrows(
-        GravitinoRuntimeException.class, () -> DataSourceUtils.createDataSource(properties));
+    GravitinoRuntimeException gre =
+        Assertions.assertThrows(
+            GravitinoRuntimeException.class, () -> DataSourceUtils.createDataSource(properties));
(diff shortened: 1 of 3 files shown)


LuciferYang:
cc @yuqi1129

roryqi (approved):

PR #11635, opened 2026-06-13
[#11583] feat(iceberg-rest): Implement fetch and cancel endpoints for async scan planning lifecycle

laserninja:
### What changes were proposed in this pull request?

Implement `fetchPlanningResult` (`GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{planId}`) and `cancelPlanning` (`DELETE .../plan/{planId}`) endpoints for the Iceberg REST server, wired through the full dispatcher chain.

**Files changed:**
- **`IcebergTableOperations`**: Two new JAX-RS endpoints with metrics, authorization, and logging
- **`IcebergTableOperationDispatcher`**: Interface methods for `fetchPlanningResult` and `cancelPlanning`
- **`IcebergTableOperationExecutor`**: Delegates to `CatalogWrapperForREST`
- **`IcebergTableEventDispatcher`**: Pre/post/failure event dispatch
- **`IcebergTableHookDispatcher`**: Pass-through delegation (no hooks needed)
- **`CatalogWrapperForREST`**: Throws `NoSuchPlanIdException` (sync-only planning)
- **`IcebergExceptionMapper`**: Maps `NoSuchPlanIdException` → 404
- **`OperationType`**: Adds `FETCH_PLANNING_RESULT` and `CANCEL_PLANNING`
- **6 new event classes**: Pre/post/failure events for both operations

### Why are the changes needed?

The Iceberg REST spec defines endpoints for fetching and cancelling async scan plans. Even though Gravitino currently only supports synchronous scan planning (always returns `COMPLETED`), these endpoints must exist to be spec-compliant. They correctly return 404 since no async plan IDs are retained, and provide a clear extension point for future async support.

Fix: #11583

### Does this PR introduce _any_ user-facing change?

Yes — two new REST API endpoints:
- `GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{planId}` — returns 404 (no async plans exist)
- `DELETE /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{planId}` — returns 404 (no async plans exist)

### How was this patch tested?

- Added `testFetchPlanningResultNotFound` and `testCancelPlanningNotFound` in `TestIcebergTableOperations`
- Tests verify 404 response status and correct event dispatch (`PreEvent` + `FailureEvent`)
- All existing tests continue to pass: `./gradlew :iceberg:iceberg-rest-server:test -PskipITs`

PR #12194, opened 2026-07-25
[#11284] feat(iceberg-rest): Implement fetch scan tasks endpoint

### LINKED ISSUE #11284: [Improvement] Support `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` endpoint for server-side scan planning

Gravitino Iceberg REST server supports the `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan` endpoint for initiating server-side scan planning, but does not support the `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` endpoint for fetching scan task results.

The Iceberg REST scan planning protocol is a two-step process:

1. **Submit scan plan**: `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan` — returns a plan task token
2. **Fetch scan tasks**: `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` — uses the plan task token to retrieve the actual scan tasks

Gravitino currently advertises the `/plan` endpoint in its `/v1/config` response:
```json
"endpoints": [
  ...
  "POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan",
  ...
]
```

But it does **not** advertise or implement the `/tasks` endpoint. When a client (e.g., pyiceberg with `scan-planning-mode=server`) successfully submits a scan plan and then attempts to fetch the tasks, it fails with:

```
NotImplementedError: Server does not support endpoint: POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks
```

This makes the server-side scan planning feature incomplete and unusable for clients that follow the two-step protocol.

### Steps to reproduce

1. Configure Gravitino Iceberg REST server with JDBC backend and S3 storage
2. Configure pyiceberg client with `scan-planning-mode=server`
3. Create a table and append data
4. Attempt to scan the table using server-side planning:

```python
from pyiceberg.catalog import load_catalog

catalog = load_catalog(
    "local",
    **{
        "type": "rest",
        "s3.endpoint": "http://localhost:9000",
        "s3.access-key-id": "admin",
        "s3.secret-access-key": "password",
        "scan-planning-mode": "server",
        "auth": {"type": "noop"},
    },
)

table = catalog.load_table("default.my_table")
tasks = list(table.scan().plan_files())
# Raises: NotImplementedError: Server does not support endpoint:
#   POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks
```

### Environment

- Gravitino version: main branch (`datastratosandbox/gravitino:1.3.0-SNAPSHOT-20260527024943`)
- Client: pyiceberg (latest)
- Configuration: JDBC catalog backend with S3 storage

### How should we improve?

Implement the `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` endpoint in the Gravitino Iceberg REST server and advertise it in the `/v1/config` response.

The endpoint should accept a plan task token (returned by the `/plan` endpoint) and return the corresponding scan tasks (file scan tasks with data file paths, partition info, etc.).

Alternatively, if Gravitino's scan planning implementation returns all tasks in a single response from the `/plan` endpoint (without requiring a separate `/tasks` call), the `/plan` endpoint response should indicate completion so that clients do not attempt to call `/tasks`.

Reference: [Apache Iceberg REST Catalog Spec - Scan Planning](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml)

laserninja:

Implement `fetchScanTasks` (`POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks`) for the Iceberg REST server, wired through the full dispatcher chain, and advertise it in `/v1/config`.

Scan planning stays synchronous (status is always `COMPLETED`, never `SUBMITTED`), but a plan is now handed to the client in batches of at most `scan-plan-task-batch-size` file scan tasks (new config, default 100, matching the Iceberg side). The first batch is returned inline by `POST .../plan`; each remaining batch is offered as a `plan-task` token that the client exchanges for its tasks through `POST .../tasks`, the second step of the Iceberg REST scan planning protocol. A plan that fits in one batch is byte-identical to before and carries no `plan-task` tokens, so most scans still complete in a single round trip.

**Plan-task tokens are self-describing.** Each token carries the scan request it was planned from - with the snapshot pinned at planning time - plus the range of tasks it covers. Nothing is stored server side between the two calls, so a token stays redeemable after a server restart and can be redeemed by any Gravitino instance serving the same catalog. Redeeming a token serves the plan from the scan plan cache when it is still cached, and replans the pinned snapshot otherwise, so enabling `scan-plan-cache-impl` is recommended when planning large tables. Planned tasks are sorted by (file location, start, length) so a token's positions resolve to the same tasks on a later replan; Iceberg plans manifests in parallel, so the raw order is not reproducible. A token this server did not issue, a token issued for another table, and a token whose plan can no longer be reproduced (for example its snapshot expired) all return `404` with `NoSuchPlanTaskException`.

**Files changed:**
- **`IcebergTableOperations`**: New JAX-RS endpoint with metrics, authorization, and logging
- **`IcebergTableOperationDispatcher`** / **`IcebergTableOperationExecutor`** / **`IcebergTableEventDispatcher`** / **`IcebergTableHookDispatcher`**: `fetchScanTasks` through the dispatcher chain, with pre/post/failure events
- **`CatalogWrapperForREST`**: batches a plan into `plan-tasks`, resolves a token back to its tasks, pins the planned snapshot, and orders planned tasks so token positions are stable
- **`PlanTaskToken`** (new): the token codec
- **`FederatedCatalogWrapper`**: forwards `POST .../tasks` to the remote catalog, since plan tasks are minted upstream in a federated setup; the HTTP client lifecycle shared by the three remote calls is factored into one helper
- **`IcebergConfig`** / **`IcebergConstants`**: new `scan-plan-task-batch-size`
- **`IcebergExceptionMapper`**: Maps `NoSuchPlanTaskException` -> 404
- **`IcebergConfigOperations`**: Advertises `Endpoint.V1_FETCH_TABLE_SCAN_PLAN_TASKS`
- **`OperationType`** / **`AuditLog`** / **`CompatibilityUtils`**: Add `FETCH_SCAN_TASKS`
- **3 new event classes**: Pre/post/failure events
- **`docs/iceberg-rest-service.md`**: Document batched planning, the token semantics, and the new config

Server-side scan planning is a two-step protocol in the Iceberg REST spec: `POST .../plan` to submit, `POST .../tasks` to collect batched results. Gravitino implemented only the first step, so `/tasks` returned 404 purely because the route did not exist.

Clients gate on the advertised endpoint set rather than probing. pyiceberg refuses to use server-side scan planning at all unless `/tasks` appears in `/v1/config`, raising `NotImplementedError`. Implementing both steps completes the protocol surface and unblocks those clients.

Batching is also what the second step exists for: without it, a plan of a large table has to be returned as one unbounded response.

Fix: #11284

> **Note on overlap with #11635:** these are different endpoints for different issues. #11635 adds `GET`/`DELETE .../plan/{planId}` (the async plan lifecycle); this PR adds `POST .../tasks` (the batched-results step). No behavioral overlap, but they touch several of the same files additively, so whichever merges second will need a trivial rebase.

Yes:
- **New REST endpoint**: `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` - returns the file scan tasks a `plan-task` token covers; 404 `NoSuchPlanTaskException` for a token this server did not issue or that is no longer resolvable; 404 `NoSuchTableException` if the table does not exist.
- **`/v1/config`** now advertises `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks`.
- **`POST .../plan`** returns at most `scan-plan-task-batch-size` file scan tasks inline and offers the rest as `plan-tasks`. Plans with fewer tasks than the batch size are unaffected. Set `gravitino.iceberg-rest.scan-plan-task-batch-size=0` to restore the previous always-inline behavior.
- **New config** `gravitino.iceberg-rest.scan-plan-task-batch-size` (default 100).
- **New audit operation** `FETCH_SCAN_TASKS`, and three new `@DeveloperApi` event classes for event listeners.
- **Incidental fix to an existing endpoint**: `POST .../plan` with a missing or empty request body previously returned **500** (unmapped `NullPointerException`); it now returns **400**. This also makes the empty-body case consistent with `{}` and `{"plan-task": null}`, which Jackson already rejected with 400.

New tests:
- `TestScanPlanTaskBatching` - plan splits into inline tasks plus plan tasks, each token resolves to its own tasks, every planned file is reachable exactly once, tokens resolve against the pinned snapshot after the table changes, the cached-plan path, batching disabled, a merge-on-read batch carrying the delete files its own tasks reference, and rejection of unknown / foreign / stale tokens.
- `TestPlanTaskToken` - token round trip (snapshot, filter, projection, stats fields, range), rejection of tokens for another table, malformed and unsupported-version payloads, URL safety.
- `TestIcebergFetchScanTasksEndpoint` - end-to-end over HTTP with a batch size of 1: `/plan` advertises two `plan-tasks` and each `POST .../tasks` returns 200 with its file scan task.
- `TestCatalogWrapperForREST.testFederatedFetchScanTasksDelegatesToRemote` - the federated path posts to the remote `.../tasks` with the token forwarded untouched.
- `TestIcebergTableOperations.testFetchScanTasksUnknownPlanTask` / `testFetchScanTasksTableNotFound` - 404 with the right error payload type, plus `PreEvent` + `FailureEvent` dispatch.
- `TestIcebergTableOperations.testScanPlanningEndpointsRejectMissingRequestBody` - the empty-body 400 for both `/plan` and `/tasks`.
- `TestIcebergConfig.testConfigAdvertisesFetchScanTasksEndpoint` - `/v1/config` advertises the `/tasks` path.

```bash
./gradlew :iceberg:iceberg-rest-server:check :iceberg:iceberg-common:check :catalogs:catalog-common:check :core:test -PskipITs
```


lasdf1234:
@roryqi This is a relatively simple PR. I have nothing else to add. From my perspective, everything looks fine. You can take a look at it sometime as well. Thank you.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -484,6 +492,46 @@ public PlanTableScanResponse planTableScan(
...
+    // Validate the table exists first, so a bad table reports 404 for the table rather than
+    // masking it as an unknown plan task. Consistent with planTableScan behavior.
+    getCatalog().loadTable(tableIdentifier);
+
+    LOG.info(
+        "Rejecting unknown plan task '{}' for table {}: scan planning is synchronous and does not "
+            + "issue plan-task tokens.",
+        request.planTask(),
+        tableIdentifier);
+    throw new NoSuchPlanTaskException(
```
roryqi: You should implement this method.

lasdf1234: This method will not be invoked in the current version of iceberg 1.11.0. Ninja implemented this function merely to fill the gap in the interface.

roryqi: If we don't implement this method, we can't say that we have supported this interface.

lasdf1234: @laserninja May be you should implement this method and update your PR's description.

laserninja: @roryqi @lasdf1234 fair point - it is implemented now, and the PR description is updated.

`POST .../plan` hands a plan to the client in batches of at most `scan-plan-task-batch-size` file scan tasks (new config, default 1000): the first batch inline, each remaining batch as a `plan-task` token that `POST .../tasks` exchanges for its tasks. Plans that fit in one batch are unchanged and carry no tokens, so the common case is still a single round trip.

Tokens are self-describing rather than server-side state: each carries the scan request it was planned from, with the snapshot pinned at planning time, plus the range of tasks it covers. That way a token stays redeemable after a restart and on any instance serving the catalog - it is resolved from the scan plan cache when the plan is still cached, and by replanning the pinned snapshot otherwise (so enabling `scan-plan-cache-impl` is worthwhile for large tables). Planned tasks are sorted by (file location, start, length) because Iceberg plans manifests in parallel, so positions stay stable across replans. Unknown, foreign and no-longer-resolvable tokens still return 404 `NoSuchPlanTaskException`.

For federated catalogs the tokens are minted by the remote catalog, so `FederatedCatalogWrapper` forwards `POST .../tasks` upstream untouched.

Two things worth flagging while reviewing:
- Batching is on by default, so `/plan` on a table with more than 1000 file scan tasks now returns tokens instead of everything inline. `scan-plan-task-batch-size=0` restores the old behavior.
- Scan responses now list the delete files their tasks reference. Iceberg serializes `delete-file-references` as indexes into that list, so before this change a plan of a merge-on-read table could not be serialized at all.

Covered by `TestScanPlanTaskBatching`, `TestPlanTaskToken`, `TestIcebergFetchScanTasksEndpoint` (end-to-end over HTTP), and a federated delegation test.

docs/iceberg-rest-service.md
```diff
@@ -697,13 +697,20 @@ Gravitino provides the build-in `org.apache.gravitino.iceberg.common.cache.Local
...
+| `gravitino.iceberg-rest.scan-plan-cache-impl`              | The implementation of the scan plan cache.                                                                                                                                           | (none)        | No       | 1.2.0         |
+| `gravitino.iceberg-rest.scan-plan-cache-capacity`          | The capacity of the scan plan cache.                                                                                                                                                 | 200           | No       | 1.2.0         |
+| `gravitino.iceberg-rest.scan-plan-cache-expire-minutes`    | The expiration time (in minutes) of the scan plan cache.                                                                                                                             | 60            | No       | 1.2.0         |
+| `gravitino.iceberg-rest.scan-plan-task-batch-size`         | Maximum number of file scan tasks returned inline by one scan planning response. Tasks beyond this limit are offered as `plan-task` tokens. Set to 0 to always return every task inline. | 1000          | No       | 1.3.0         |
```
lasdf1234: Why is it 1000? Is this value setting not quite reasonable? The implementation of iceberg itself is 100. @roryqi WDYT

roryqi: It's ok to align to Iceberg side.

lasdf1234:
@laserninja Could you modify 'gravitino.iceberg-rest.scan-plan-task-batch-size' default value to 100?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/IcebergConfig.java
```diff
@@ -341,6 +341,18 @@ public class IcebergConfig extends Config implements OverwriteDefaultConfig {
...
+          .doc(
+              "Maximum number of file scan tasks returned inline by one scan planning response. "
+                  + "Tasks beyond this limit are handed out as plan-task tokens that clients "
+                  + "exchange for the remaining tasks. Set to 0 to disable batching and always "
+                  + "return every task inline.")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .intConf()
+          .checkValue(value -> value >= 0, ConfigConstants.NON_NEGATIVE_NUMBER_ERROR_MSG)
+          .createWithDefault(1000);
+
```
lasdf1234, line 355: May be 100 is better.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -484,6 +479,232 @@ public PlanTableScanResponse planTableScan(
...
+
+  /**
+   * Pins the snapshot a scan request plans against, so that plan-task tokens issued for the
+   * resulting plan keep resolving to the same snapshot after the table changes.
+   *
+   * <p>Requests that already name a snapshot, incremental requests (which pin a snapshot range) and
+   * scans of a table without a current snapshot are returned unchanged.
+   */
+  private static PlanTableScanRequest pinSnapshot(Table table, PlanTableScanRequest scanRequest) {
+    boolean isIncremental =
```
lasdf1234: This method can be omitted (as it was only used once), and the logic can be incorporated into the original method.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
```diff
@@ -82,9 +89,21 @@ public class CatalogWrapperForREST extends IcebergCatalogWrapper {
...
+  private final int scanPlanTaskBatchSize;
+
   private static final String DATA_ACCESS_VENDED_CREDENTIALS = "vended-credentials";
   private static final String DATA_ACCESS_REMOTE_SIGNING = "remote-signing";
 
+  /**
+   * Total order over file scan tasks, so that plan-task tokens, which address tasks by position,
+   * keep pointing at the same tasks when a plan is recomputed.
+   */
+  private static final Comparator<FileScanTask> FILE_SCAN_TASK_ORDER =
```
lasdf1234: FILE_SCAN_TASK_ORDER is only used once — consider inlining the comparator at the sort call site to avoid the extra constant.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/PlanTaskCodec.java
```diff
@@ -0,0 +1,166 @@
...
+ * Nothing about the token is stored server side, so it stays valid across server restarts and can
+ * be redeemed by any Gravitino instance that serves the same catalog.
+ *
+ * <p>The token is opaque to clients but is not a capability: it grants no access on its own. {@code
+ * POST .../tasks} authorizes the table in the request path and rejects a token minted for a
+ * different table, so a forged token can at most express a scan the caller could already submit
+ * through {@code POST .../plan}.
+ */
+class PlanTaskToken {
+
```
lasdf1234, line 52: This is a very good design. But I think there are some properties that don't need to be done.May be 'table‘ + 'limit' + 'offset' + 'scan' can meet the requirements.

lasdf1234:
@laserninja 
The overall approach is feasible. 
However, this is a rather touching project. Especially, a new type of token has been added. So, could you please write a design document? Thank you very much. I am more than willing to review this document.

This design document should include the token and change interfaces as well as the added interfaces.

docs/iceberg-rest-service.md
laserninja: Aligned - the default is 100 now, in the config, the docs table and the PR description. 1000 was my own guess at "big enough that most scans stay single round trip"; matching the Iceberg side is the better default, and operators who want the old behaviour can set `scan-plan-task-batch-size=0` to keep every task inline.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/IcebergConfig.java
laserninja: Changed to 100.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
laserninja: Done - the snapshot pinning is now inline in `planTableScan`, so `pinSnapshot` is gone.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
laserninja: Done - the comparator is now built at the `fileScanTasks.sort(...)` call site and the constant is gone.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/PlanTaskCodec.java
laserninja: Agreed, dropped `version`. The token is now exactly `table` + `offset` + `limit` + `scan`. Version was only there for a future format change, which `decode` already handles: anything that does not decode into those four fields is treated as a token this server did not issue and reported as `404 NoSuchPlanTaskException`, so an old server meeting a future token behaves the same way with or without the field.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/CatalogWrapperForREST.java
laserninja: Correction to something I claimed earlier in this PR: I said scan responses had to list the delete files their tasks reference, and that a merge-on-read plan could not be serialized without it. That is wrong, and the extra code is now removed (3e239e24f).

Iceberg 1.11's `BaseScanTaskResponse.Builder.withFileScanTasks` already derives the response's `delete-files` from the tasks it is given, which is exactly why `withDeleteFiles` is deprecated. I verified it by serializing a plan of a table with a position delete file without touching delete files at all: the response carries `"delete-files":[…]` and `"delete-file-references":[0]` either way. So `POST .../plan` was never broken for merge-on-read tables, and the deprecated calls I had added were redundant.

What batching does still have to guarantee is that a task and the delete files it references end up in the *same* response, since the references are indexes into that response. That holds because each batch is built through `withFileScanTasks(batch)`, and the test now asserts it for a fetched batch instead of restating Iceberg's behaviour. The PR description is updated to drop the incorrect claim.

laserninja:
@lasdf1234 design document written up as a separate PR so it can be reviewed on its own: https://github.com/apache/gravitino/pull/12241

It covers the `plan-task` token (payload, encoding, validation, and why it is not a capability), the new `POST .../tasks` interface, what changes about `POST .../plan` and the migration impact per client, the internal dispatcher interfaces added, and the determinism the token depends on - the planned snapshot is pinned and planned tasks are totally ordered, because Iceberg plans manifests in parallel and an index range only identifies tasks if the plan is reproducible.

It also records the alternatives I rejected, with reasons: an in-memory plan-task store keyed by a random id (a token dies with the process and is meaningless on another replica, and the Iceberg client fetches plan tasks concurrently, so a scan would fail for timing reasons), and asynchronous planning with server-side plan state (that is #11635's scope and does not remove the need for `.../tasks`).

Three questions in §8 are genuinely open and I would value your opinion, since they are design calls rather than implementation details:

1. Should batching require the scan plan cache to be enabled? With the cache off (today's default) an N-batch plan costs N full plans, since a token is resolved by replanning its pinned snapshot. The alternative is to keep everything inline unless a cache is configured.
2. Is 100 the right default? It matches the Iceberg side, but it makes a 100,000-file scan 1,000 requests. Sizing batches by serialized bytes rather than task count may serve large tables better.
3. Should tokens carry an expiry? Today a token is valid as long as its snapshot exists.

Happy to fold any of these back into #12194.

nevzheng:
@laserninja 
1. please break [#12194](https://github.com/apache/gravitino/pull/12194) into smaller, reviewable PRs. We trust your judgment on how to divide it holistically.
2. when you split the PRs, please tie each one to [#11284](https://github.com/apache/gravitino/issues/11284) so we can track the stack from the epic.



Refs:
* [small CLs](https://google.github.io/eng-practices/review/developer/small-cls.html);
* [CONTRIBUTING](https://github.com/apache/gravitino/blob/main/CONTRIBUTING.md). 

Thanks,
Nevin
Sent from my 🤖 (Cursor)

laserninja:
@nevzheng agreed, this PR is too big to review as one change. Here is how I propose to split it, all five tied to #11284. Splitting it also surfaced something I had not noticed: this PR bundles credential vending for `POST .../plan`, which has nothing to do with the `/tasks` endpoint and belongs on its own.

| # | PR | What it contains | Size |
| - | -- | ---------------- | ---- |
| 1 | #12411 | Order planned file scan tasks deterministically | ~190 lines, 2 files |
| 2 | to open | Credential vending for `POST .../plan` (`X-Iceberg-Access-Delegation`), and `400` instead of `500` for an empty request body | ~200 lines |
| 3 | to open | The `POST .../tasks` endpoint and its plumbing: dispatcher chain, events, audit operation type, `NoSuchPlanTaskException` → `404`. **Not** advertised in `/v1/config` yet | ~350 lines |
| 4 | to open | Federated catalogs forward `POST .../tasks` to the remote catalog | ~200 lines |
| 5 | this PR | Batching a plan into `plan-tasks`, `PlanTaskCodec`, `scan-plan-task-batch-size`, advertising the endpoint in `/v1/config`, user docs | ~700 lines |

Two things about the ordering that are deliberate rather than arbitrary.

**The `/v1/config` advertisement lands last, in PR 5.** Clients gate on the advertised endpoint set rather than probing: pyiceberg enables server-side scan planning when it sees `POST .../tasks` in `/v1/config` and fails the scan if the endpoint then does not serve tasks. So advertising it before it can return anything would be worse than not having it at all, and PR 3 deliberately ships the endpoint unadvertised. That also answers @roryqi's point on this PR - PR 3 makes no claim to support the interface; PR 5 is where the claim and the working implementation arrive together.

**Federation comes before batching.** A federated catalog delegates `/plan` upstream, so the plan tasks a client receives are the remote catalog's, and forwarding them is correct and testable before Gravitino issues any of its own. Doing it after PR 5 would leave a window where a federated deployment advertises the endpoint and 404s on it.

PR 1 is open now. I will open each of the others as its predecessor merges, so that every diff you see is against a base that already has what it depends on, rather than four PRs showing cumulative diffs. This PR stays open as the last one and gets rebased down to slice 5 once 1 to 4 land - the review history here is worth keeping attached to the part it argued about, including the naming discussion with @lasdf1234, the ordering bug @roryqi found, and the manifest entry-range design in #12241.

Each PR will be green on `./gradlew :iceberg:iceberg-rest-server:check -PskipITs` on its own, not only at the end of the stack.

If you would rather see a different division - for example credential vending folded back in, or the endpoint and batching kept together so the interface is never dormant in `main` - say so and I will reshape it before opening more.

nevzheng:
@laserninja this split plan looks reasonable. Please proceed.

When you're ready add me @lasdf1234 , and @roryqi to the R line.

laserninja:
Correction to the plan above: **slice 2 is not needed, credential vending for `POST .../plan` is already in `main`.** I had been comparing against a stale fork main, 132 commits behind, which is also what broke CI on #12411 - the merge with the real `main` placed a hunk outside the scope it needed, because `planTableScan` has since been restructured. Rebased and fixed there.

So the stack is four, not five:

| # | PR | What it contains | State |
| - | -- | ---------------- | ----- |
| 1 | #12411 | Order planned file scan tasks deterministically | open, CI green |
| 2 | #12412 | The `POST .../tasks` endpoint and its plumbing: dispatcher chain, events, audit operation type, `NoSuchPlanTaskException` → `404`. Not advertised in `/v1/config` yet | open |
| 3 | to open | Federated catalogs forward `POST .../tasks` to the remote catalog | - |
| 4 | this PR | Batching a plan into `plan-tasks`, `PlanTaskCodec`, `scan-plan-task-batch-size`, advertising the endpoint in `/v1/config`, user docs | open |

The ordering constraints are unchanged: the `/v1/config` advertisement lands last so the endpoint is never advertised before it can serve tasks, and federation lands before batching so a federated deployment never advertises an endpoint it would 404 on.

Two notes on #12412 now that it is open. It is based directly on `main` rather than on #12411, so the two can be reviewed and merged in either order. And it carries one piece that was originally part of the dropped credential vending slice and did not merge with the rest of it: returning `400` instead of `500` when a scan planning endpoint is called with an empty request body. `/tasks` needed that guard anyway, so applying the same three lines to `/plan` seemed better than leaving the neighbouring endpoint wrong - happy to split it out if you would rather keep #12412 single-purpose.

#12412 is where the endpoint interface gets decided, so if shipping it dormant is the wrong call, that is the place to say so. Reshaping it now is cheap; I have deliberately not built slice 3 yet.

This PR is rebased onto current `main`, so what you see here is a diff against the base it will actually merge into.

docs/iceberg-rest-service.md
```diff
@@ -697,13 +697,20 @@ Gravitino provides the build-in `org.apache.gravitino.iceberg.common.cache.Local
...
+| `gravitino.iceberg-rest.scan-plan-cache-impl`              | The implementation of the scan plan cache.                                                                                                                                           | (none)        | No       | 1.2.0         |
+| `gravitino.iceberg-rest.scan-plan-cache-capacity`          | The capacity of the scan plan cache.                                                                                                                                                 | 200           | No       | 1.2.0         |
+| `gravitino.iceberg-rest.scan-plan-cache-expire-minutes`    | The expiration time (in minutes) of the scan plan cache.                                                                                                                             | 60            | No       | 1.2.0         |
+| `gravitino.iceberg-rest.scan-plan-task-batch-size`         | Maximum number of file scan tasks returned inline by one scan planning response. Tasks beyond this limit are offered as `plan-tasks`. Set to 0 to always return every task inline. | 100           | No       | 1.3.0         |
```
lasdf1234, line 714: @roryqi  Should it be 2.0.0?

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/IcebergConfig.java
```diff
@@ -341,6 +341,18 @@ public class IcebergConfig extends Config implements OverwriteDefaultConfig {
...
 
+  public static final ConfigEntry<Integer> SCAN_PLAN_TASK_BATCH_SIZE =
+      new ConfigBuilder(IcebergConstants.SCAN_PLAN_TASK_BATCH_SIZE)
+          .doc(
+              "Maximum number of file scan tasks returned inline by one scan planning response. "
+                  + "Tasks beyond this limit are handed out as plan-tasks that clients "
+                  + "exchange for the remaining tasks. Set to 0 to disable batching and always "
+                  + "return every task inline.")
+          .version(ConfigConstants.VERSION_1_3_0)
+          .intConf()
```
lasdf1234, line 352: VERSION_2_0_0 ?

docs/iceberg-rest-service.md
laserninja: You are right, fixed in both places - `VERSION_2_0_0` in `IcebergConfig` and `2.0.0` in the docs table.

I had copied 1.3.0 from the neighbouring entries in `IcebergConfig`, but all of those were added before `gradle.properties` was bumped to `2.0.0-SNAPSHOT` in ee6b17409 (2026-06-29). Every config added since then uses `VERSION_2_0_0`, so this one should too.

iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/IcebergConfig.java
laserninja: You are right, fixed in both places - `VERSION_2_0_0` in `IcebergConfig` and `2.0.0` in the docs table.

I had copied 1.3.0 from the neighbouring entries in `IcebergConfig`, but all of those were added before `gradle.properties` was bumped to `2.0.0-SNAPSHOT` in ee6b17409 (2026-06-29). Every config added since then uses `VERSION_2_0_0`, so this one should too.

lasdf1234:
Another issue is that this PR seems to need to be split into two parts, as each PR is theoretically around 1000 lines.
Apart from this, should planTask use a new structure instead of limit and offset?

laserninja:
@lasdf1234 both fair. Taking them in turn.

**Splitting.** Already in progress - see [the plan above](https://github.com/apache/gravitino/pull/12194#issuecomment-5246735447). Two of the four are open: #12411 (189 lines) and #12412 (387 lines). Measured against a tree with both merged, what is left here is 1156 insertions across 14 files, of which 424 are production code and 732 are tests, plus 200 lines of federation that becomes its own PR.

If 1156 is still too much, the natural next cut is three:

| PR | Contents | Size |
| -- | -------- | ---- |
| a | `PlanTaskCodec` and its round-trip tests. Encoding only, nothing calls it yet | 292 |
| b | Batching in `planTableScan`, redemption in `fetchScanTasks`, `scan-plan-task-batch-size`, batching and E2E tests | 828 |
| c | Advertise `POST .../tasks` in `/v1/config`, user docs | 36 |

Splitting (c) out is worth it on its own: it is the commit where the feature becomes visible to clients, so it is the one to revert if anything goes wrong, and it is small enough to review in a minute. Say the word and I will cut it this way.

**`plan-task` structure.** Yes, and that is your manifest entry-range proposal from [#12241](https://github.com/apache/gravitino/pull/12241#issuecomment-5178530018) - I worked it through in §8.5 of the design doc. It is the better end state: naming manifest entry ranges lets redemption read those manifest slices instead of re-planning the snapshot, so cost scales with the batch rather than the plan, and it retires the total-ordering requirement entirely, since a position inside an immutable manifest is already stable.

What keeps it out of this PR is delete attachment. `DeleteFileIndex`, which matches delete files to data files by partition and sequence number, is package-private in `iceberg-core`, so entry ranges would mean reimplementing those rules - equality deletes, positional deletes, deletion vectors - where a subtle mistake silently returns deleted rows. The staged path in §8.5.4 is to use entry ranges when the pinned snapshot has no delete manifests and fall back to re-planning when it does, which covers copy-on-write tables.

The reason offset/limit now does not lock us in: a `plan-task` is opaque to clients and is never persisted server side, so the payload can change in any later release with no migration and no compatibility window. The only thing that has to be right now is that the string stays opaque, which it is.

roryqi:
@lasdf1234 @nevzheng Could u help me review this pull request?

PR #12197, opened 2026-07-25
[#12196] fix(python): Deserialize datetime fields in AuditDTO

### LINKED ISSUE #12196: [Bug] Python client AuditDTO fails to deserialize datetime fields and violates Audit interface
### Version

main branch

### Describe what's wrong

Currently, the Audit interface in clients/client-python/gravitino/api/audit.py defines the create_time and last_modified_time methods as returning datetime objects:   
 
 @abstractmethod
    def create_time(self) -> datetime:
        pass

However, AuditDTO in clients/client-python/gravitino/dto/audit_dto.py, which implements Audit, returns a str instead. There are explicit TODO comments indicating that datetime deserialization from JSON is currently missing:

    _create_time: Optional[str] = field(
        default=None, metadata=config(field_name="createTime")
    )  # TODO: Can't deserialized datetime from JSON

Why it needs to be improved: This causes a type mismatch at runtime. Any user relying on the Python client's Audit interface expects to work with datetime objects, but will surprisingly receive string representations.

# No explicit crash, but causes unexpected typing behavior in downstream code relying on datetime methods:
audit_info = get_audit()
time = audit_info.create_time()

# This will fail at runtime because time is a str, not a datetime:
print(time.year) 
# AttributeError: 'str' object has no attribute 'year'

### How to reproduce

1. Use the python client to interact with the Gravitino API and retrieve an entity containing Audit metadata (e.g., getting a Metalake).
2. Call .audit().create_time() on the retrieved object.
3. Observe that the returned type is a raw ISO-8601 string rather than a Python datetime object, contrary to the Audit interface definition.

### Additional context

_No response_

niraj-mx07:

1. Updated `AuditDTO` to use `datetime` instead of `str` for `_create_time` and `_last_modified_time`.
2. Configured custom `decoder` and `encoder` functions in `dataclasses_json` metadata to properly parse the ISO-8601 strings into `datetime` objects.
3. Updated the return types of `create_time()` and `last_modified_time()` methods to properly return `datetime`.
4. Modified python unit tests that incorrectly tested for strings to compare against `datetime` objects.

The python client's `AuditDTO` was failing to deserialize `datetime` fields from JSON, keeping them as raw ISO-8601 strings (and marked with `TODO: Can't deserialized datetime from JSON`). This violated the `Audit` interface, which defines `create_time()` and `last_modified_time()` to return `datetime` objects. This change fixes the type mismatch at runtime.

Fix: #12196

Yes, `AuditDTO.create_time()` and `AuditDTO.last_modified_time()` in the Python client will now correctly return `datetime` objects instead of strings, aligning with the expected interface.

- Ran and updated existing python unit tests (`test_relational_table.py`, `test_table_dto.py`, `test_responses.py`) which now correctly assert against `datetime` objects.


jerryshao:
@niraj-mx07 can you please remove the changelog file, I think it doesn't relate to this PR.

jerryshao:
@niraj-mx07 can you please fix some lint issues reported in CI?

niraj-mx07:
Thanks for the feedback! I've addressed the reported lint issues and pushed an update. Could you please take another look when you have time?

README.md
```diff
@@ -106,6 +106,11 @@ See: [Iceberg REST catalog service](https://gravitino.apache.org/docs/latest/ice
 Gravitino provides a native Lance REST catalog service.  
 See: [Lance REST catalog service](https://gravitino.apache.org/docs/latest/lance-rest-service/)
 
+## 📋 Changelog
+
+A detailed release history is available in [CHANGELOG.md](CHANGELOG.md).  
+For the latest release notes, see [GitHub Releases](https://github.com/apache/gravitino/releases).
+
```
jerryshao, line 113: Can you revert this change?

jerryshao:
I think you pulled in lots of unnecessary code, can you please fix it?

jerryshao:
Hi @niraj-mx07 this issue has already been fixed by other PRs. Thanks a lot for your contributions.

Issue #11748, opened 2026-06-20
[Subtask]  Support Tencent Cloud COS as fileset storage backend (Java side)


whua3:
### Describe the subtask

Sub-task of #6490.

This sub-task introduces Java-side support for Tencent Cloud Object Storage (COS) in the fileset catalog.

### Scope

- Add a new `bundles/tencent` module that provides `COSFileSystemProvider`, registering the `cosn://` scheme (the standard scheme used by `hadoop-cos`) so the fileset catalog can manage COS-backed filesets.
- Add a `bundles/tencent-bundle` module that produces a shaded runtime jar (bundling `hadoop-cos`, `cos_api-bundle`, etc.) for server-side classpath.
- Add `COSProperties` under `catalogs/catalog-common` to declare the set of COS-related property keys (`cos-region`, `cos-endpoint`, `cos-access-key-id`, `cos-secret-access-key`, …) used by the catalog and credential layer.
- Add `COSSecretKeyCredential` / `COSSecretKeyProvider` for static AK/SK credential vending; STS support is left to a follow-up sub-task.
- Add user-facing documentation under `docs/` describing how to configure a COS-backed fileset catalog.
- Add unit tests and an opt-in integration test (`FilesetCOSCatalogIT`) gated by environment variables (`COS_*`), so CI is unaffected when credentials are absent.

### Pull Request

- PR: #11713

### Parent issue

https://github.com/apache/gravitino/issues/6490

PR #11713, opened 2026-06-17
 [#11748] feat(catalog-hadoop): support Tencent Cloud COS as fileset storage backend (Java side)

whua3:
### What changes were proposed in this pull request?

Add Tencent Cloud Object Storage (COS) support for the fileset catalog. The layout follows the existing OSS / S3 / GCS / Azure modules.

- `bundles/tencent`: source module containing
  - `COSFileSystemProvider`: builds a Hadoop `CosNFileSystem` for the `cosn://` scheme.
  - `COSCredentialsProvider`: Hadoop-side provider that bridges Gravitino-vended `COSSecretKey` credentials to `hadoop-cos`.
  - `COSSecretKeyProvider`: server-side `CredentialProvider` that returns long-lived COS Secret Key credentials.
  - `COSUtils`: helper that maps Gravitino properties to Hadoop `Configuration`.
- `bundles/tencent-bundle`: shaded fat-jar (`gravitino-tencent-bundle-<version>.jar`) that bundles `hadoop-cos` and `cos_api-bundle` with relocated packages.
- `api`: new `COSSecretKeyCredential` type.
- `catalogs/catalog-common`: `COSProperties` and `COSCredentialConfig`.
- `clients/filesystem-hadoop3-runtime`: include `bundles/tencent` in the GVFS runtime jar.
- `docs`: new `fileset-catalog-with-cos.md` and a cross-link from `fileset-catalog.md`.
- Tests: `TestCOSFileSystemProvider`, `TestCOSCredentialProvider`, and `FilesetCOSCatalogIT`. The IT is skipped on CI when no real COS credentials are provided, the same way `FilesetOSSCatalogIT` is.

### Why are the changes needed?

#6490 tracks adding Tencent COS support. Without it, users on Tencent Cloud have to either rely on S3-compatible workarounds, which do not work with Gravitino's credential vending, or maintain their own fork.

This PR is the first of a chain of PRs against `main`:

- PR-A (this PR): Java side, with the static `cos-secret-key` credential.
- PR-B: server-side STS credential vending (`cos-token`).
- PR-C: Python GVFS support (`COSStorageHandler` and Python-side `COSSecretKeyCredential`).
- PR-D: Python `COSTokenCredential` wiring. Can be folded into PR-C if reviewers prefer fewer PRs.

Part of #6490
Fix: #11748 

### Does this PR introduce _any_ user-facing change?

Yes, additive only. Existing fileset / catalog behavior is unchanged. New surface:

[...]

Issue #6490, opened 2025-02-20
[EPIC] Support COS(Tencent cloud storage) for Hadoop catalog


yuqi1129:
### Describe the proposal

Gravitino has support S3, GCS, Azure Blob Storage and OSS,  as many Chinese users use COS, we'd support it also
 '

### Task list

- [ ] Support OSS for the Hadoop Catalog and Java GVFS client.
- [ ] Support Python GVFS client for COS

yuqi1129:
> I'm very interested in this task. Our company could also use this feature, but I haven't found an existing Python COS (Cloud Object Storage) FileSystem package that meets the requirements.

Thanks for your enthusiasm. Sure, please just go ahead.

> but I haven't found an existing Python COS (Cloud Object Storage) FileSystem package that meets the requirements.

Is it possible to access COS using S3, as it is S3-compatible?  I have not verified it, but you may give it a try.

241600489:
ok ,i‘ll try it

whua3:
Hi @yuqi1129 @mchades , I would like to pick this one up if nobody's working on it yet.

Plan is to follow roughly the same path the OSS support took (#5174 → #5701 → #5239 → #5777), splitting the work into the following PRs — open to reordering or merging based on review:

- **PR-A** — Java side: `bundles/cos` + `bundles/cos-bundle` + Hadoop catalog wiring + Java GVFS, static AK/SK only. (mirrors #5174)
- **PR-B** — Server-side STS credential vending: `COSTokenProvider` + `COSTokenGenerator` + policy DSL. Depends on PR-A. (mirrors #5701)
- **PR-C** — Python GVFS support: `COSStorageHandler` in `gvfs_storage_handler.py`, plus `COSSecretKeyCredential`. Depends on PR-A. (mirrors #5239)
- **PR-D** — Python `COSTokenCredential` wiring. Depends on PR-B + PR-C. Could fold into PR-C if you'd prefer fewer PRs — happy either way.

Iceberg / Paimon support I'd leave for a separate follow-up issue, otherwise the scope gets too big.

A few things I'd like to confirm before starting:

1. **Scheme** — planning to use `cosn://` since that's what `hadoop-cos` registers natively. Let me know if there's a reason to prefer something else.
2. **AppID** — Tencent recommends embedding it into the bucket name (`my-bucket-1259999999`), so I wasn't going to add a separate `cos-app-id` property. Sound reasonable?
[...]

PR #12199, opened 2026-07-25
[#12198] feat(credential): support Tencent Cloud COS credential vending

### LINKED ISSUE #12198: [Subtask] Server-side STS credential vending for Tencent Cloud COS (cos-token)
### Describe the subtask

Sub-task of #6490.

This sub-task introduces server-side STS credential vending for Tencent Cloud Object Storage (COS), the follow-up to #11748 that was scoped out at the time. With this change, a fileset catalog can be configured with `credential-providers=cos-token` so that the Gravitino server calls Tencent Cloud CAM `AssumeRole` on behalf of the client and vends a short-lived STS triple (`TmpSecretId` / `TmpSecretKey` / `SessionToken`) scoped down to the paths of the requesting fileset. This brings COS to parity with the existing S3 / OSS token providers.

### Parent issue

https://github.com/apache/gravitino/issues/6490

whua3:

Add credential vending support for the Tencent Cloud COS fileset backend introduced in #11713. This is a follow-up subtask under #6490.

- Add `COSTokenCredential` to the public credential API, alongside the existing `S3TokenCredential`, `OSSTokenCredential`, `GCSTokenCredential`, and `ADLSTokenCredential`.
- Add `cos-token` credential provider implementation (`COSTokenProvider` + `COSTokenGenerator`) that calls Tencent Cloud STS `AssumeRole` and returns temporary credentials scoped to the requested table path via a generated CAM policy.
- Add lightweight CAM policy model classes (`Policy`, `Statement`, `Effect`, `Condition`, `StringLike`) shared by the token generator and covered by unit tests.
- Reuse the existing `cos-secret-key` provider so the two providers share the same `cos-access-key-id` / `cos-secret-access-key` configuration surface.
- Register both providers via `META-INF/services` so they are picked up by `CredentialProviderFactory` and by `CredentialFactory` on the client side.
- Wire `COSCredentialsProvider` (the Hadoop `AWSCredentialsProvider`) to consume vended `COSTokenCredential`s, in addition to the static secret-key credential.
- Update `bundles/tencent` and `bundles/tencent-bundle` to include the new sources and the Tencent STS SDK dependency.
- Update `CredentialConstants`, `COSCredentialConfig`, and `COSProperties` to expose the new provider name and configuration keys (`cos-role-arn`, `cos-region`, `cos-app-id`, `cos-external-id`, `cos-token-expire-in-secs`).

Tests:
- Unit tests for `COSTokenGenerator`, `COSCredentialConfig`, `COSCredentialProvider` SPI registration, and `TestCredentialFactory` round-trip.
- Integration test `GravitinoVirtualFileSystemCOSCredentialIT` exercising the end-to-end vending path against a real COS bucket through GVFS.

Documentation:
- Add a `## COS` section to `docs/security/credential-vending.md`, documenting `cos-token` and `cos-secret-key`, their properties, and the required CAM trust and permission policies.
- Update `docs/fileset-catalog-with-cos.md` with credential-vending configuration examples.

#11713 added COS as a fileset storage backend but only supported the static access-key path. Without credential vending, every client that loads a fileset receives long-lived, coarse-grained Tencent Cloud credentials, which is inconsistent with how Gravitino already handles S3, OSS, GCS, and ADLS. This PR closes that gap so COS fileset deployments can hand out temporary, table-scoped STS credentials.

Yes. Two new credential providers, `cos-token` and `cos-secret-key`, are now available for catalogs using the COS storage backend, together with the associated `cos-role-arn` / `cos-region` / `cos-app-id` / `cos-external-id` / `cos-token-expire-in-secs` catalog properties. Existing catalogs continue to work unchanged.

- New unit tests under `bundles/tencent` and `catalogs/catalog-common`.
- New integration test `GravitinoVirtualFileSystemCOSCredentialIT` run locally against a Tencent Cloud COS bucket with an assumable CAM role, covering both read and write paths through GVFS with vended credentials.

Fix: #12198
Related: #6490, #11713, #11748

### DIFF
--- a/bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
+++ b/bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
@@ -1,0 +1,311 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.gravitino.cos.credential;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Preconditions;
+import com.tencentcloudapi.common.Credential;
+import com.tencentcloudapi.common.exception.TencentCloudSDKException;
+import com.tencentcloudapi.sts.v20180813.StsClient;
+import com.tencentcloudapi.sts.v20180813.models.AssumeRoleRequest;
+import com.tencentcloudapi.sts.v20180813.models.AssumeRoleResponse;
+import java.io.IOException;
+import java.net.URI;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.gravitino.cos.credential.policy.Effect;
+import org.apache.gravitino.cos.credential.policy.Policy;
+import org.apache.gravitino.cos.credential.policy.Statement;
+import org.apache.gravitino.credential.COSTokenCredential;
+import org.apache.gravitino.credential.CredentialContext;
+import org.apache.gravitino.credential.CredentialGenerator;
+import org.apache.gravitino.credential.PathBasedCredentialContext;
+import org.apache.gravitino.credential.config.COSCredentialConfig;
+
+/** Generates Tencent Cloud COS STS tokens scoped to the requested fileset paths. */
+public class COSTokenGenerator implements CredentialGenerator<COSTokenCredential> {
+
+  private static final String POLICY_VERSION = "2.0";
+
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  private String accessKeyId;
+  private String secretAccessKey;
+  private String roleArn;
+  private String externalId;
+  private String region;
+  private String appId;
+  private int tokenExpireSecs;
+
+  @Override
+  public void initialize(Map<String, String> properties) {
+    COSCredentialConfig config = new COSCredentialConfig(properties);
+    this.accessKeyId = config.accessKeyID();
+    this.secretAccessKey = config.secretAccessKey();
+    this.roleArn = config.cosRoleArn();
+    this.externalId = config.externalID();
+    this.region = config.region();
+    this.appId = config.appID();
+    this.tokenExpireSecs = config.tokenExpireInSecs();
+  }
+
+  @Override
+  public COSTokenCredential generate(CredentialContext context) throws Exception {
+    if (!(context instanceof PathBasedCredentialContext)) {
+      return null;
+    }
+
+    PathBasedCredentialContext pathContext = (PathBasedCredentialContext) context;
+
+    AssumeRoleResponse response =
+        callAssumeRole(
+            pathContext.getReadPaths(), pathContext.getWritePaths(), pathContext.getUserName());
+
+    com.tencentcloudapi.sts.v20180813.models.Credentials credentials = response.getCredentials();
+    Long expiredTime = response.getExpiredTime();
+    Preconditions.checkState(
+        credentials != null && expiredTime != null,
+        "Tencent STS AssumeRole returned an incomplete response, requestId: %s",
+        response.getRequestId());
+    // Tencent STS returns ExpiredTime in seconds; the Credential contract uses ms.
+    long expireTimeInMs = expiredTime * 1000L;
+    return new COSTokenCredential(
+        credentials.getTmpSecretId(),
+        credentials.getTmpSecretKey(),
+        credentials.getToken(),
+        expireTimeInMs);
+  }
+
+  private AssumeRoleResponse callAssumeRole(
+      Set<String> readLocations, Set<String> writeLocations, String userName)
+      throws TencentCloudSDKException {
+    Credential cred = new Credential(accessKeyId, secretAccessKey);
+    StsClient client = new StsClient(cred, region);
+
+    AssumeRoleRequest request = new AssumeRoleRequest();
+    request.setRoleArn(roleArn);
+    request.setRoleSessionName(getRoleSessionName(userName));
+    request.setDurationSeconds((long) tokenExpireSecs);
+    if (StringUtils.isNotBlank(externalId)) {
+      request.setExternalId(externalId);
+    }
+    request.setPolicy(buildPolicy(readLocations, writeLocations));
+
+    return client.AssumeRole(request);
+  }
+
+  private String buildPolicy(Set<String> readLocations, Set<String> writeLocations) {
+    Preconditions.checkArgument(
+        !readLocations.isEmpty() || !writeLocations.isEmpty(),
+        "COS token generator requires at least one read or write location");
+    Policy.Builder policyBuilder = Policy.builder().version(POLICY_VERSION);
+
+    Statement.Builder readObjectStatement =
+        Statement.builder()
+            .effect(Effect.ALLOW)
+            .addAction("cos:GetObject")
+            .addAction("cos:HeadObject");
+
+    // LinkedHashMap keeps the emitted statements in a deterministic order.
+    Map<String, Statement.Builder> bucketListStatements = new LinkedHashMap<>();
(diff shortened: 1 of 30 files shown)


bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/policy/Condition.java
```diff
@@ -0,0 +1,57 @@
...
+      return this;
+    }
+
+    public Condition build() {
+      return new Condition(this);
+    }
+  }
+
+  @SuppressWarnings("unused")
+  public StringLike getStringLike() {
```
yuqi1129: Why can't we just remove it?

catalogs/catalog-common/src/main/java/org/apache/gravitino/credential/config/COSCredentialConfig.java
```diff
@@ -58,11 +59,63 @@ public class COSCredentialConfig extends Config {
           .checkValue(StringUtils::isNotBlank, ConfigConstants.NOT_BLANK_ERROR_MSG)
           .create();
 
+  public static final ConfigEntry<String> COS_ROLE_ARN =
+      new ConfigBuilder(COSProperties.GRAVITINO_COS_ROLE_ARN)
+          .doc("The CAM role ARN that the server assumes when issuing STS temporary credentials")
```
yuqi1129: What's the meaning of `CAM` here?

docs/security/credential-vending.md
```diff
@@ -11,16 +11,16 @@ Gravitino credential vending is used to generate temporary or static credentials
...
-| Catalog type | Vends                           |
-|--------------|---------------------------------|
-| Fileset      | S3, OSS, GCS, ADLS              |
-| Hive         | S3, OSS, GCS, ADLS              |
-| Iceberg      | S3, OSS, GCS, ADLS              |
-| Glue         | S3                              |
-| JDBC         | JDBC user and password          |
-| Paimon       | S3, OSS, JDBC user and password |
+| Catalog type | Vends                                |
+|--------------|--------------------------------------|
```
yuqi1129: Remove redundant space in the table.

yuqi1129:
Good job. Only some minors were found in the first round of review, and I will start the next round of review within a few days.

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/policy/StringLike.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+/**
+ * Tencent Cloud CAM {@code string_like} condition value. The {@code cos:prefix} key is used to
+ * restrict {@code cos:GetBucket}/{@code ListObjects} to a specific prefix inside the bucket.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class StringLike {
+
+  @JsonProperty("cos:prefix")
```
yuqi1129, line 36: Value "cos:prefix" is not so common in Java. Is there any special reason that we use this format?

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
```diff
@@ -0,0 +1,354 @@
...
+   * <p>hadoop-cos may pass the fileset prefix itself (e.g. {@code xxx/}) or the prefix plus a
+   * sub-path (e.g. {@code xxx/foo/}) as the {@code prefix} of a list request. Tencent Cloud CAM's
+   * {@code string_like} matcher requires the pattern to be {@code xxx*} (no trailing slash) to
+   * cover both forms; the alternative {@code xxx/*} fails to match {@code xxx/} and returns
+   * AccessDenied. The pattern is still anchored to the fileset prefix, so it cannot broaden the
+   * scope of the token.
+   */
+  private Condition buildPrefixCondition(URI uri) {
+    return Condition.builder()
+        .stringLike(StringLike.builder().addPrefix(trimLeadingSlash(uri.getPath()) + "*").build())
```
yuqi1129: Appending "*" directly will cause the following problem:
```
.../orders -> .../order*  // it will make get the priviledge for paths like `orders_backup/..`
```

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/policy/Condition.java
whua3: done

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/policy/StringLike.java
whua3: cos:prefix is the literal condition key defined by Tencent Cloud CAM policy language, analogous to AWS IAM's s3:prefix and Alibaba Cloud OSS's oss:prefix. It must be serialized verbatim into the policy JSON — Jackson uses the string as-is via @JsonProperty, so Java-style naming does not apply here. I've added a Javadoc note on the class explaining this.

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
whua3: done

whua3:
Thanks for review. @yuqi1129 I've been busy recently, sorry for the delay.

whua3:
Clipboard_Screenshot_1787928992
Clipboard_Screenshot_1787929005
Clipboard_Screenshot_1787929015
Clipboard_Screenshot_1787929023
Clipboard_Screenshot_1787929029

yuqi1129:
I have no more comments, @mchades Would you like to take a look?

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
```diff
@@ -0,0 +1,303 @@
...
+   * xxx_backup/}.
+   */
+  private Condition buildPrefixCondition(URI uri) {
+    String prefix = trimLeadingSlash(uri.getPath());
+    if (!prefix.endsWith("/")) {
+      prefix = prefix + "/";
+    }
+    return Condition.builder()
+        .stringLike(StringLike.builder().addPrefix(prefix).addPrefix(prefix + "*").build())
+        .build();
```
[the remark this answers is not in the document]
whua3: Good catch. done.

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+    if (!writeLocations.isEmpty()) {
+      Statement.Builder writeObjectStatement =
+          Statement.builder()
+              .effect(Effect.ALLOW)
+              .addAction("cos:PutObject")
+              .addAction("cos:DeleteObject")
+              .addAction("cos:InitiateMultipartUpload")
+              .addAction("cos:UploadPart")
+              .addAction("cos:CompleteMultipartUpload")
+              .addAction("cos:AbortMultipartUpload");
```
yuqi1129, line 178: The generated write policy is missing `cos:ListParts`. The pinned `hadoop-cos` 3.3.0-8.3.23 calls `listParts(key, uploadId)` when `UploadPart` returns 409 to determine whether the part actually succeeded; with this session policy, that recovery request gets 403 and turns a recoverable multipart upload into a failure. Tencent Cloud also lists `cos:ListParts` as required for resumable multipart uploads: https://cloud.tencent.com/document/product/436/47231. Please add this action and assert the complete multipart action set in the policy test.

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+              // CAM requires different resource ARN forms per action: cos:GetBucket needs
+              // bucket/*, whereas cos:HeadBucket / cos:GetBucketLocation need bucket/.
+              bucketListStatements.computeIfAbsent(
+                  bucketWildcardResource,
+                  key ->
+                      Statement.builder()
+                          .effect(Effect.ALLOW)
+                          .addAction("cos:GetBucket")
+                          .addResource(key)
+                          .condition(buildPrefixCondition(uri)));
```
yuqi1129: This `computeIfAbsent` captures only the first URI for a bucket. If the context contains `cosn://bucket/path-a` and `cosn://bucket/path-b`, the object statements include both paths, but the single `GetBucket` statement retains only the `path-a` prefix condition, so listing `path-b` is denied. `CredentialOperationDispatcher` deliberately merges multiple `PathContext`s of the same credential type into one `PathBasedCredentialContext`, so this is a supported input shape. Please accumulate all prefixes per bucket (or emit one statement per prefix) and add regression coverage for two paths in the same bucket, including a read/write combination.

whua3: Fixed by accumulating cos:prefix per URI, following the same pattern as S3TokenGenerator.

bundles/tencent/src/main/java/org/apache/gravitino/cos/credential/COSTokenGenerator.java
whua3: Good catch — added cos:ListParts to the write policy.

yuqi1129 (approved):

yuqi1129:
@whua3 
Thanks for your hard work.

whua3:
> @whua3 Thanks for your hard work.

@yuqi1129 Thanks a lot for your review and many valuable suggestions.

PR #12202, opened 2026-07-26
[#12201] refactor(core): inject table hook dependencies

### LINKED ISSUE #12201: [Improvement] Remove GravitinoEnv service lookup from TableHookDispatcher

`TableHookDispatcher` decorates table operations with ownership and authorization lifecycle behavior. Although its underlying `TableDispatcher` is supplied through its constructor, it still obtains `OwnerDispatcher` and `CatalogManager` at runtime through `GravitinoEnv.getInstance()`.

This makes those dependencies implicit and couples unit tests to shared singleton state. Tests must modify private `GravitinoEnv` fields through reflection and restore them afterward. This is an internal code-quality and test-isolation improvement; there is no known user-facing behavior defect.

`GravitinoEnv` should remain the composition root, while `TableHookDispatcher` receives its collaborators explicitly.

### How should we improve?

Update the `TableHookDispatcher` constructor to receive:

- `Supplier<OwnerDispatcher>` for lazy resolution, because the owner dispatcher is constructed later and may be `null` when authorization is disabled.
- `CatalogManager` for catalog capability handling.

Wire these dependencies from `GravitinoEnv`, propagate the constructor change to every construction site, and update the affected tests to instantiate `TableHookDispatcher` directly with mocks.

The change should preserve behavior and should not introduce a dependency-injection framework. Deleting `GravitinoEnv` or its `getInstance()` accessor is out of scope.

nevzheng:

This pull request:

- Injects `Supplier<OwnerDispatcher>` and `CatalogManager` into `TableHookDispatcher`.
- Wires both dependencies from `GravitinoEnv`.
- Replaces singleton- and reflection-based test setup with focused dispatcher unit tests.
- Adds focused `AuthorizationUtils` tests for table rename and removal payloads.
- Covers normalized ownership, disabled authorization, privilege removal, privilege rename, result propagation, and owner-assignment failure.

GravitinoEnv becomes composition-root-only in the touched areas; not deleted.

`TableHookDispatcher` currently resolves collaborators through
`GravitinoEnv.getInstance()`. This makes its dependencies implicit and forces
tests to mutate shared singleton state.

Constructor injection makes the dependency relationship explicit while
preserving lazy owner resolution and authorization-disabled behavior.

#### Test design rationale

The previous `TestTableHookDispatcher` was an over-scoped hybrid rather than a
focused unit test or a true end-to-end test. It combined hook orchestration with
real table and schema operations, shared singleton mutation, lock setup, and
mocked authorization collaborators.

Reviewing its assertions exposed two fidelity problems:

- The drop test mocked `AuthorizationUtils` and did not verify that privilege
  removal was requested.
- The rename test only verified `onMetadataUpdated(any())`; it did not verify
  the metadata change type, old and new table identities, or locations.

The replacement separates those responsibilities:

- `TestTableHookDispatcher` verifies delegation, returned-result propagation,
  owner handling, and the exact authorization helper calls and arguments.
- `TestAuthorizationUtils` executes the real rename and removal helpers and
  verifies that the authorization plugin receives exactly one
  `RenameMetadataObject` or `RemoveMetadataObject` with the expected table
  identity and locations.
- Existing operation-level tests cover real table behavior:
  `TestTableOperationDispatcher#testCreateAndDropTable` and
  `TestManagedTableOperations#testAlterTable`.
- Existing Docker-backed Ranger tests provide broader integration coverage:
  `RangerHiveHdfsE2EIT#testRenameTable` exercises the client, REST resource,
  `TableHookDispatcher`, `AuthorizationUtils`, and Ranger plugin, then verifies
  permissions at the renamed HDFS path.
  `RangerBaseE2EIT#testRenameMetadataObjectPrivilege` renames through the same
  server path and verifies operations on the renamed table.
- `RangerHiveIT#testMetadataObjectChangeRenameTable` and
  `#testMetadataObjectChangeRemoveTable` directly verify the resulting Ranger
  policy changes.

##### Coverage map

| Previous test or behavior | Coverage after this change |
| --- | --- |
| `testDropAuthorizationPrivilege` | The rewritten test verifies the delegated result and exact remove-helper arguments. `TestAuthorizationUtils` now verifies the real `RemoveMetadataObject` payload. Real create/drop behavior remains in `TestTableOperationDispatcher#testCreateAndDropTable`. |
| `testCreateTableSetsOwnerWithNormalizedIdentifier` | The same owner call and normalized table/schema assertions are retained using injected mocks, and delegate-result propagation is now asserted. |
| `testCreateTableThrowsWhenSetOwnerFails` | The same exception type and message are retained. The unrelated real table-create path remains covered by table-operation tests. |
| `testRenameAuthorizationPrivilege` | The rewritten hook test verifies non-rename behavior, exact rename-helper arguments, and result propagation. `TestAuthorizationUtils` verifies the real rename payload, while the Ranger tests retain broader integration coverage. |
| Inherited `testGetCatalogIdentifier` | It still runs through `TestTableOperationDispatcher` and other `TestOperationDispatcher` subclasses; removing inheritance here only removes a duplicate execution. |
| Authorization disabled | A new test verifies that owner/catalog work is skipped while the delegated table result is preserved. |

This gives each layer a focused contract while retaining broader end-to-end
coverage.

Fixes #12201

No. This changes internal server wiring and tests only. It does not change
public APIs, REST contracts, or configuration.

- `./gradlew spotlessApply`
- `./gradlew :core:test :server:test -PskipITs`

The Docker-backed Ranger tests listed above are pre-existing coverage and were
not executed by the `-PskipITs` verification command.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
+++ b/core/src/main/java/org/apache/gravitino/GravitinoEnv.java
@@ -711,7 +711,8 @@ public class GravitinoEnv {
         new TableNormalizeDispatcher(internalTableOperationDispatcher, catalogManager);
     TableEventDispatcher tableEventDispatcher =
         new TableEventDispatcher(eventBus, tableNormalizeDispatcher);
-    this.tableDispatcher = new TableHookDispatcher(tableEventDispatcher);
+    this.tableDispatcher =
+        new TableHookDispatcher(tableEventDispatcher, this::ownerDispatcher, catalogManager);
 
     // TODO: We can install hooks when we need, we only supports ownership post hook,
     //  partition doesn't have ownership, so we don't need it now.
--- a/core/src/main/java/org/apache/gravitino/hook/TableHookDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/hook/TableHookDispatcher.java
@@ -20,14 +20,15 @@ package org.apache.gravitino.hook;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Supplier;
 import org.apache.gravitino.Entity;
-import org.apache.gravitino.GravitinoEnv;
 import org.apache.gravitino.NameIdentifier;
 import org.apache.gravitino.Namespace;
 import org.apache.gravitino.authorization.AuthorizationUtils;
 import org.apache.gravitino.authorization.Owner;
 import org.apache.gravitino.authorization.OwnerDispatcher;
 import org.apache.gravitino.catalog.CapabilityHelpers;
+import org.apache.gravitino.catalog.CatalogManager;
 import org.apache.gravitino.catalog.TableDispatcher;
 import org.apache.gravitino.connector.capability.Capability;
 import org.apache.gravitino.exceptions.NoSuchSchemaException;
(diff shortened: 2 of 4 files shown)


roryqi (approved):
LGTM, elegant solution!

PR #12214, opened 2026-07-27
[MINOR] fix(auth): Handle missing parent catalog during metadata id resolution

### LINKED ISSUE #12216: [Bug report] Revoking privileges can fail with internal error when parent catalog is missing
### Version

main branch

### Describe what's wrong

Revoking privileges from a role for a child metadata object can fail with an internal authorization error when the parent catalog no longer exists.

During authorization, `MetadataIdConverter` resolves the requested metadata object into an internal metadata id. For child objects such as filesets, schemas, tables, or topics, it first loads the parent catalog to apply the catalog case-sensitivity normalization. If the parent catalog has already been dropped, the lookup throws `NoSuchCatalogException`, which bubbles up as a system internal error before the revoke operation reaches the business logic.

```text
System internal error during authorization - Operation: revokePrivilegeFromRole
Caused by: org.apache.gravitino.exceptions.NoSuchCatalogException:
Catalog <metalake>.<catalog> does not exist
  at org.apache.gravitino.catalog.CatalogManager.loadCatalogInternal
  at org.apache.gravitino.catalog.CapabilityHelpers.getCapability
  at org.apache.gravitino.server.authorization.MetadataIdConverter.normalizeCaseSensitive
  at org.apache.gravitino.server.authorization.MetadataIdConverter.getID
  at org.apache.gravitino.server.authorization.jcasbin.JcasbinAuthorizationLookups.loadMetadataId

### How to reproduce

Use the main branch.
Create a role and grant privileges on a child metadata object, such as a schema, table, fileset, or topic.
Drop the parent catalog.
Try to revoke the privilege from the role for that child metadata object.
The revoke operation can fail during authorization metadata id resolution with an internal error.

### Additional context

A missing parent catalog means the requested child metadata object cannot be resolved. MetadataIdConverter#getID should treat this as missing metadata and return Optional.empty() instead of propagating NotFoundException.

### PR DESCRIPTION
**Pull request by jarredhj0214**

This PR updates `MetadataIdConverter` to gracefully handle missing parent metadata during metadata id resolution.

When resolving metadata ids for objects such as filesets, schemas, tables, or topics, `MetadataIdConverter` normalizes the object identifier according to the parent catalog's case-sensitivity capability before looking up the entity id. If that parent catalog no longer exists, the target metadata cannot be resolved, so the converter now returns `Optional.empty()` instead of propagating `NotFoundException`.

A unit test was added to cover the missing parent catalog case.

Revoking privileges from a role for a metadata object whose parent catalog no longer exists can fail during authorization expression evaluation with an internal error.

For example, the failure can look like:

```text
System internal error during authorization - Operation: revokePrivilegeFromRole
Caused by: org.apache.gravitino.exceptions.NoSuchCatalogException:
Catalog <metalake>.<catalog> does not exist
  at org.apache.gravitino.catalog.CatalogManager.loadCatalogInternal
  at org.apache.gravitino.catalog.CapabilityHelpers.getCapability
  at org.apache.gravitino.server.authorization.MetadataIdConverter.normalizeCaseSensitive
  at org.apache.gravitino.server.authorization.MetadataIdConverter.getID
  at org.apache.gravitino.server.authorization.jcasbin.JcasbinAuthorizationLookups.loadMetadataId
```

This happens before the revoke operation reaches the business logic. During authorization, the requested metadata object is converted to an internal metadata id. For child metadata objects, this conversion first loads the parent catalog to apply the correct case-sensitivity normalization. If the parent catalog has already been dropped, that lookup throws `NoSuchCatalogException`, which currently bubbles up through the authorization expression evaluator as an internal error.

Returning `Optional.empty()` is consistent with the existing contract of `MetadataIdConverter#getID`: if the metadata object cannot be resolved, the converter should return empty. A missing parent catalog means the requested child metadata object cannot exist or be resolved in Gravitino, so treating it as missing metadata is safer and more accurate than failing authorization evaluation with an internal exception.

Fix: #12216

Ran:

```shell
./gradlew :server-common:test \
  --tests org.apache.gravitino.server.authorization.TestMetadataIdConverter \
  --tests org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizationLookups \
  -PskipITs -PskipDockerTests=false
```


roryqi:
Could u create an issue first? The fix type pull request should have an issue to track the fixed version.

jarredhj0214:
> Could u create an issue first? The fix type pull request should have an issue to track the fixed version.

ok

jarredhj0214:
> Could u create an issue first? The fix type pull request should have an issue to track the fixed version.

The tracking issue requested above has been created and linked: #12216.

Could someone please help take another look when available?

Thanks!

yuqi1129 (approved):

PR #12215, opened 2026-07-27
[#12218] feat(secret): Add GravitinoSecretProvider SPI and in-memory provider

### LINKED ISSUE #12218: [FEATURE] Add GravitinoSecretProvider SPI for entity connection secrets
### Describe the feature

Introduce a secret-provider SPI (`GravitinoSecretProvider`) so Gravitino can
store and resolve entity connection secrets through pluggable backends.

Initial scope:
- SPI lifecycle and CRUD-style secret operations (`initialize` / `type` /
  `writeSecret` / `readSecret` / `deleteSecret` / `close`)
- process-local `InMemorySecretsProvider` for development and unit tests
- minimal supporting types (`SecretWriteContext`, `SecretUrn`, `SecretConstants`)

Out of scope for the first PR: provider registry wiring, REST/create-alter-drop
flows, and external secret references.

### Motivation

Entity connection properties (for example JDBC passwords) should not stay as
plaintext in metadata. A provider SPI is the OSS foundation for write-through
secrets, resolve/omit-on-read, and later KMS-backed or external backends.

### Describe the solution

Add `org.apache.gravitino.secret` types in `core`, with default no-op
`initialize`/`close` so simple providers stay constructor-only, and a Base64
in-memory provider marked non-production / non-encryption.

### Additional context

Follow-up work can add registry configuration, create-time write-through,
resolve/omit helpers, and REST exposure.

lasdf1234:

Add the secret-provider SPI (`GravitinoSecretProvider`) with:
- `initialize` / `type` / `writeSecret` / `readSecret` / `deleteSecret` / `close`
- a process-local `InMemorySecretsProvider` for development and unit tests
- minimal supporting types used by the SPI (`SecretWriteContext`, `SecretUrn`,
  `SecretConstants`)

This PR intentionally does **not** wire create/alter/drop flows, REST APIs,
provider registry configuration, or external secret references.

To land the OSS foundation for entity connection-secret backends. Later PRs can
add provider registry, create-time write-through, resolve/omit-on-read, and REST.

Fix: #12218

- No user-facing REST/config change in this PR.
- Adds internal SPI types under `org.apache.gravitino.secret`.

- `./gradlew :core:test --tests 'org.apache.gravitino.secret.TestInMemorySecretsProvider' --tests 'org.apache.gravitino.secret.TestSecretUrn' -PskipITs`


roryqi:
Will  this  interface be  used in the  client module? If yes, you shouldn't put it into the code module. Connector can't reuse the code in the core module.

lasdf1234:
> Will this interface be used in the client module? If yes, you shouldn't put it into the code module. Connector can't reuse the code in the core module.

At the moment, I haven't considered the specific scenarios for the client. However, for better compatibility, I have decided to place it in the "common" module as per your suggestion, (keeping it consistent with the implementation of KMS)

lasdf1234:
@nevzheng Rory has no more issues with my code. If you have time, please take a look at it. Thank you.

common/src/main/java/org/apache/gravitino/secret/GravitinoSecretProvider.java
```diff
@@ -0,0 +1,78 @@
...
+   * Initializes this provider after construction.
+   *
+   * <p>The default implementation is a no-op so providers that need no configuration (for example
+   * in-memory) can keep a no-arg constructor only.
+   *
+   * @param name the configured provider instance name
+   * @param config provider-specific configuration (without the {@code gravitino.secret.provider.
+   *     <name>.} prefix)
+   */
+  default void initialize(String name, Map<String, String> config) {}
```
nevzheng: suggestion (blocking): Please remove the default no-op initialize/close.

Each implementer should explicitly consider lifecycle (an empty body is fine). Silent defaults make it too easy to skip cleanup.

common/src/main/java/org/apache/gravitino/secret/memory/InMemorySecretsProvider.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.gravitino.secret.SecretWriteContext;
+
+/**
+ * In-memory secret provider for development and unit tests.
+ *
+ * <p>Secrets are stored in process memory only and are lost on restart. Values are Base64-encoded
+ * for opaque storage, which is <strong>not</strong> encryption. Do not use this provider in
+ * production.
+ */
+public class InMemorySecretsProvider implements GravitinoSecretProvider {
```
nevzheng: issue (blocking): close() does not release stored secrets.

suggestion: Override close() and call secrets.clear(), and assert that in the unit test. As written, close() leaves secrets reachable for the process lifetime.

nevzheng: note (non-blocking, security): Map.clear() only drops references.

The JVM does not securely zero String/heap contents when values become unreachable, so secret material can linger until GC (and still isn’t guaranteed wiped). Fine for this test-only provider if documented; production backends should use wipeable byte[]/char[] with explicit zeroing.

nevzheng (changes requested):

common/src/main/java/org/apache/gravitino/secret/GravitinoSecretProvider.java
lasdf1234: Thanks for the suggestion. I've removed the default no-op implementations of initialize and close so they are abstract, and each provider must override them explicitly (an empty body is fine when there is nothing to set up or clean up). The in-memory provider now overrides both methods accordingly.

common/src/main/java/org/apache/gravitino/secret/memory/InMemorySecretsProvider.java
lasdf1234: Good catch. InMemorySecretsProvider.close() now clears the in-memory map via secrets.clear(), and I added a unit test (testCloseClearsStoredSecrets) to verify secrets are no longer readable after close.

lasdf1234: Agreed. Since this provider is intended for development and unit tests only, I kept String/Base64 storage and documented that close() clears map references but does not securely zero heap contents (String is not wipeable). Production providers should use wipeable byte[]/char[] and explicitly zero them on delete/close.

nevzheng (approved):
LGTM!

nevzheng (approved):

common/src/main/java/org/apache/gravitino/secret/GravitinoSecretProvider.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package org.apache.gravitino.secret;
+
+import java.util.Map;
+import org.apache.gravitino.annotation.DeveloperApi;
+
+/** Service provider interface for Gravitino secret backends. */
+@DeveloperApi
+public interface GravitinoSecretProvider {
```
roryqi: Could u rename the key word `Gravitino`?

common/src/main/java/org/apache/gravitino/secret/SecretWriteContext.java
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Creates a secret write context.
+   *
+   * @param providerName the configured secret provider name used in the URN
+   * @param entityType the entity type (catalog, schema, or fileset)
+   * @param entityId the entity identifier
+   * @param propertyKey the property key holding the secret
+   */
+  public SecretWriteContext(
```
roryqi: Entity is the core module concept. We shouldn't put it into the common module. Maybe you can use securable object type and objectUUID.

common/src/main/java/org/apache/gravitino/secret/GravitinoSecretProvider.java
lasdf1234: What name would you like to change it to?This is exactly how my design document is written.

roryqi: Sorry, I mean that we should remove the keyword `Gravitino`.

lasdf1234: Got. Rename to SecretProvider.

common/src/main/java/org/apache/gravitino/secret/SecretWriteContext.java
lasdf1234: Got I remove this class and place it in api module.

common/src/main/java/org/apache/gravitino/secret/SecretProvider.java
```diff
@@ -0,0 +1,84 @@
...
+   */
+  String writeSecret(String plaintext, Map<String, String> attributes);
+
+  /**
+   * Reads a secret by URN.
+   *
+   * @param urn the secret URN
+   * @return the secret plaintext
+   */
+  String readSecret(String urn);
```
roryqi: Should we use SecretUrn urn here?

common/src/main/java/org/apache/gravitino/secret/SecretProvider.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Deletes a secret by URN.
+   *
+   * @param urn the secret URN
+   */
+  void deleteSecret(String urn);
```
roryqi: Should we use `SecretUrn` here?

common/src/main/java/org/apache/gravitino/secret/SecretProvider.java
lasdf1234: Got resolved.

common/src/main/java/org/apache/gravitino/secret/SecretProvider.java
lasdf1234: Got resolved.

roryqi (approved):
LGTM.

Issue #12221, opened 2026-07-27
[EPIC] Add Iceberg REST soft-delete lifecycle and recovery APIs


nevzheng:
### Describe the proposal

This is the Iceberg REST API child epic of #12219. It depends on the relational foundation in #12220.

The first API milestone is intentionally small:

```text
DELETE table → retain and hide the original row
GET tables?deleted=true → list authorized retained names
POST table/undrop → reactivate the selected original row
```

### DELETE

Reuse the existing Iceberg REST table DELETE endpoint.

- When soft deletion is enabled, both `purgeRequested` values retain the original `table_meta` row and reserve its name.
- Return `204 No Content` only after the metadata transaction commits.
- A repeated DELETE of the retained generation returns `204` without allocating another action.
- Never unregister Iceberg metadata or delete files on the request thread.
- When soft deletion is disabled, preserve existing behavior.

### Deleted-table discovery

Reuse the existing table collection resource:

`GET /v1/{prefix}/namespaces/{namespace}/tables?deleted=true`

- Return the ordinary `ListTablesResponse`, containing authorized retained table names only.
- Reuse existing `pageToken` and `pageSize` behavior.
- Filter authorization before pagination and disclosure.
- Do not expose deletion IDs, lifecycle details, metadata locations, credentials, cleanup fields, or deletion-action ETags.
- Omitted `deleted` and `deleted=false` keep live-table behavior unchanged.

### UNDROP

Add an explicit, empty-body action:

`POST /management/v1/{prefix}/namespaces/{namespace}/tables/{table}/undrop`

- Resolve and authorize the retained row by name, then carry its exact internal `{table_id, deletion_id}` into the transaction.
- Do not require a public deletion ID, deletion-action ETag, or `If-Match`.
- Require the action to remain `DELETED`, unexpired, and unattached to a purge job.
- Reactivate the original `table_meta` row and consume the active action in one transaction.
- Return the ordinary live table response and live-table ETag.
- Use concealed `404` for missing or unauthorized targets, `409` for a conflicting state, and `410` after the UNDROP boundary.

## Shared invariants

- The retained row reserves its original name; create, register, and rename-to that name fail until restore or purge.
- DELETE and UNDROP use fine-grained row coordination, not table-wide locks.
- The internal deletion ID fences the exact generation but is not part of the public API.
- UNDROP directly reactivates the row. It does not create a replacement table, register Iceberg metadata, touch files, or run a saga.

## After this epic

[...]

PR #12225, opened 2026-07-27
[#12221] feat(iceberg): Add Iceberg REST soft-delete lifecycle

nevzheng:

> **Iceberg REST API stack (Stack 20), review 1 of 4.** Depends on #12223. Because this PR targets `main`, Files changed is cumulative until the storage stack lands. Review only [`3f7ed54ac`](https://github.com/apache/gravitino/pull/12225/commits/3f7ed54ac8871243c1e792097ae5462e86667f49) and [`8806e3255`](https://github.com/apache/gravitino/pull/12225/commits/8806e32556367de2ebf28c136acf3dee2c4ea980).

This PR adds the configuration and internal Iceberg REST lifecycle coordinator:

1. Add `gravitino.iceberg-rest.soft-delete.enabled` (default `false`).
2. Add `gravitino.iceberg-rest.soft-delete.retention-ms` (default 24 hours; range 0–90 days).
3. Resolve the existing live table to its immutable `table_id`.
4. Lock that table row with fine-grained `SELECT ... FOR UPDATE`.
5. In one relational transaction, allocate the opaque deletion ID, insert the `DELETED` action, stamp `deleted_at` and `deletion_id`, and append the DROP change event.
6. Keep identity, name, ownership, version, and related metadata on the original table row.

Both `purgeRequested` values use retention when soft deletion is enabled. When it is disabled or unavailable, existing routing remains unchanged.

**After this PR:** an internal coordinator can durably retain one existing table generation. The public DELETE endpoint does not use it until #12228.

Not in this PR: DELETE wiring, name-reservation enforcement, discovery, UNDROP, Iceberg unregister/register, file deletion, purge jobs, sagas, or audit storage.

Review order: #12222 → #12223 → **#12225** → #12228 → #12226 → #12227.

The API needs one short transaction boundary that retains the original row before DELETE, discovery, and UNDROP can be exposed.

Part of #12221.

Yes. It adds the two opt-in Iceberg REST soft-delete configuration properties. DELETE behavior is connected in #12228.

Focused tests cover configuration bounds, row locking, transaction rollback, repeated and concurrent deletion, missing tables, fixed retention deadlines, and preservation of the original table identity.


iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/deletion/IcebergDeletionAuthorization.java
```diff
@@ -0,0 +1,66 @@
...
+   * Checks current {@code DROP_TABLE} semantics against a live or retained table generation.
+   *
+   * <p>The retained-owner fallback is fenced by the immutable table ID and deletion ID, and still
+   * requires current parent-use privileges.
+   *
+   * @param identifier full Gravitino table identifier
+   * @param deletion exact retained root and action, or {@code null} for a live or missing table
+   * @return whether the current principal may manage the deletion
+   */
+  public static boolean canDrop(
```
roryqi: Why do we need to check permission here? Usually, we check it in the IncepterService.

nevzheng:
Stack update: this PR's code is unchanged.

We simplified the downstream API after review. The retained deletion coordinator still preserves the existing table row, creates the internal deletion action transactionally, and uses fine-grained row locking.

The final Stack 20 review order is #12225 → #12228 → #12226 → #12227. PR #12229 is no longer required.

iceberg/iceberg-rest-server/src/main/java/org/apache/gravitino/iceberg/service/deletion/IcebergDeletionAuthorization.java
nevzheng: Hi Rory — agreed. I removed the standalone `IcebergDeletionAuthorization` check from this PR. Authorization now uses the existing `IcebergRESTAuthInterceptionService` path in [#12228](https://github.com/apache/gravitino/pull/12228): the endpoint declares `MANAGE_TABLE_DELETION`, the interceptor selects `IcebergTableDeletionAuthzHandler`, and the endpoint receives the already-authorized `{tableId, deletionId}` target. The lifecycle coordinator in this PR is now authorization-neutral. Thanks!

PR #12226, opened 2026-07-27
[#12221] feat(iceberg): List soft-deleted Iceberg REST tables

nevzheng:

> **Iceberg REST API stack (Stack 20), review 3 of 4.** Depends on #12228. Files changed is cumulative until earlier stack layers land. Review only [`c7b0cba36`](https://github.com/apache/gravitino/pull/12226/commits/c7b0cba3683a7350fb0d8e0e0c70c446b2f74cb9).

This PR adds deleted-table discovery to the existing table collection API:

`GET /v1/{prefix}/namespaces/{namespace}/tables?deleted=true`

- Return the ordinary `ListTablesResponse`, containing retained table identifiers only.
- Reuse existing `pageToken` and `pageSize` behavior.
- Authorize each retained row locally by immutable `table_id` before pagination and disclosure.
- Mark deleted-only responses `Cache-Control: private, no-store`.
- Do not load the live Iceberg catalog for deleted-only requests.
- Keep omitted `deleted` and `deleted=false` behavior unchanged.
- Do not add a specialized deletion route or expose deletion IDs, state, expiry, metadata locations, credentials, cleanup fields, or deletion-action ETags.

**After this PR:** authorized clients can find retained table names through the existing paginated table API. #12227 restores the selected name.

The POC deliberately authorizes the namespace's retained set before applying in-memory pagination. Database-level filtered pagination is future optimization work.

Review order: #12222 → #12223 → #12225 → #12228 → **#12226** → #12227.

Clients need a small discovery surface for choosing a retained table to restore. Reusing the existing list resource keeps naming and pagination in one API.

Part of #12221.

Yes. `deleted=true` returns authorized retained table names in the ordinary table-list response.

Focused tests cover deleted-only response shape, authorization before pagination, two-page pagination, missing namespaces, cache control, live-catalog isolation, and unchanged omitted/false behavior.


nevzheng:
This PR was rewritten to expose only retained table names through the existing table-list API:

`GET /v1/{prefix}/namespaces/{namespace}/tables?deleted=true`

It preserves existing pagination, filters authorization before pagination, avoids loading the live Iceberg catalog, and returns the ordinary `ListTablesResponse`. It exposes no deletion ID, lifecycle record, specialized detail route, or deletion-specific ETag.

This keeps discovery inside the existing table API and gives clients exactly what they need to select a table for UNDROP.

PR #12227, opened 2026-07-27
[#12221] feat(iceberg): Add UNDROP for soft-deleted Iceberg REST tables

nevzheng:

> **Iceberg REST API stack (Stack 20), review 4 of 4.** Depends on #12226. Files changed is cumulative until earlier stack layers land. Review only [`36206cc5f`](https://github.com/apache/gravitino/pull/12227/commits/36206cc5f7600942971f5f9365704aedfec7e147).

This PR adds the name-only UNDROP action:

`POST /management/v1/{prefix}/namespaces/{namespace}/tables/{table}/undrop`

### Request and authorization

- Use an empty request body with no public deletion ID, deletion-action ETag, or `If-Match`.
- Reuse the existing Iceberg authorization interceptor.
- Resolve and authorize the retained row by immutable `table_id`.
- Carry the exact internal `{table_id, deletion_id}` target into the restore path.
- Fail closed when authorization is enabled but the exact request-scoped target is unavailable.

### Restore transaction

- Require the action to remain `DELETED`, unexpired, and unattached to a purge job.
- Lock the exact action first and its retained table row second.
- Set `deleted_at=0`, clear `deletion_id`, consume the active action, and append the ALTER change event in one relational transaction.
- Preserve the original `table_id` and all table-owned metadata.
- Invalidate caches after commit and return the ordinary live `LoadTableResponse` and live-table ETag.

HTTP outcomes are concealed `404` for missing, unauthorized, wrong-generation, or already-restored targets; `409` for a conflicting current state; and `410` after expiry or purge ownership.

UNDROP directly reactivates the retained row. It never creates a replacement table, registers or unregisters Iceberg metadata, touches files, or runs a saga. A repeated request after success performs no mutation and normally observes the now-live name as `404`.

**After this PR:** Stack 20's API user journey is complete: DELETE retains the original row, `deleted=true` lists authorized retained names, and UNDROP reactivates the selected row. Physical purge remains separate work.

Review order: #12222 → #12223 → #12225 → #12228 → #12226 → **#12227**.

This provides recovery through one short relational transaction while fencing the exact deletion generation selected during authorization.

Part of #12221.

Yes. Authorized clients can restore a retained table by name using an empty POST. Success returns the normal live-table representation and ETag.

Focused tests cover bodyless success, exact-generation authorization, row-locked restoration, rollback and retry, concurrent UNDROP, expiry and purge boundaries, `404`/`409`/`410` mapping, action consumption, change-log commit, and cache invalidation.


nevzheng:
This PR was rewritten as name-only UNDROP:

`POST /management/v1/{prefix}/namespaces/{namespace}/tables/{table}/undrop`

The request has no body, deletion ID, or `If-Match` requirement. The server resolves the retained row by schema and table name, captures its internal deletion ID, and carries that exact generation through authorization and the row-locked restore transaction.

UNDROP directly reactivates the original row, clears its deletion fields, consumes the action, appends the metadata change, invalidates caches, and returns the ordinary live table response. It performs no Iceberg registration, synchronous cleanup, or saga.

PR #12228, opened 2026-07-28
[#12221] feat(iceberg): Enable soft deletion through Iceberg REST DELETE

nevzheng:

> **Iceberg REST API stack (Stack 20), review 2 of 4.** Depends on #12225. Files changed is cumulative until earlier stack layers land. Review only [`8d13b833b`](https://github.com/apache/gravitino/pull/12228/commits/8d13b833b00f1b60a14358889c0431fb35af80ef) and [`3ee6f7cd3`](https://github.com/apache/gravitino/pull/12228/commits/3ee6f7cd361104027d57d7edc5aa3d3d0e0643a8); merge commit `01e4efdc0` only synchronizes current upstream `main`.

This PR connects the coordinator from #12225 to the existing Iceberg REST DELETE endpoint.

### DELETE behavior

- When soft deletion is enabled, route DELETE through the retained-row transaction and return `204` after commit.
- Preserve existing behavior when soft deletion is disabled.
- Treat DELETE of an already retained generation as an idempotent `204` without allocating another action or redispatching by name.
- Never unregister Iceberg metadata or delete files on the request thread.

### Authorization and generation safety

- Reuse the existing Iceberg authorization interceptor with a deletion-aware handler.
- Authorize a retained row by immutable `table_id` using the existing DROP permission semantics.
- Carry the exact internal `{table_id, deletion_id}` target in request scope.
- Perform local authorization for this local relational operation even when ordinary REST-backed catalog operations delegate authorization.
- Conceal unauthorized retained names as missing.

### Name reservation and cache coherence

- Hide retained names from ordinary load, list, exists, update, scan, credentials, and rename-from operations.
- Reject create, register, and rename-to operations for a reserved name.
- Invalidate the local table cache after commit.
- Use the existing entity change log to invalidate corresponding peer-server cache entries.
- Cache eviction remains best effort after the durable metadata commit; failure does not turn a committed DELETE into a failed response.

**After this PR:** the existing DELETE endpoint durably retains and hides the original row, reserves its name, and keeps local and peer caches coherent. Discovery and UNDROP follow in #12226 and #12227.

Not in this PR: public deletion IDs, deletion-action ETags, `If-Match`, physical purge, or job progress.

Review order: #12222 → #12223 → #12225 → **#12228** → #12226 → #12227.

The retained-row transaction must be connected to the public DELETE path with authorization, name reservation, and cache coherence before recovery can be exposed safely.

Part of #12221.

Yes. With soft deletion enabled, DELETE retains and hides the existing table row and reserves its name. Disabled behavior is unchanged.

Focused tests cover DELETE routing and `204`, idempotent retries, exact-target authorization, REST-backend delegation, name reservation across table operations, local and peer cache invalidation, concurrency, and legacy-cleanup compatibility.


nevzheng:
Stack update: this PR's code is unchanged.

DELETE still retains and hides the original row, reserves its name, and performs no synchronous Iceberg cleanup. Only the downstream discovery and UNDROP APIs were simplified.

The final Stack 20 review order is #12225 → #12228 → #12226 → #12227.

PR #12229, opened 2026-07-28
[#12221] feat(iceberg): Add retained deletion read model

nevzheng:

This is Stack 20 review unit 3 of 5. It depends on #12228 and should be reviewed before #12226.

Review the commits in order:

1. `feat(core): join retained table deletion records`
   - Adds a read-only projection pairing the retained `table_meta` row with its exact `entity_deletion` action.
   - Implements one-statement joins for an exact deleted table and all deleted tables in a schema.
   - Avoids N+1 action lookups and does not duplicate table identity or name in `entity_deletion`.
2. `feat(iceberg): add retained deletion read model`
   - Maps the joined storage projection into a safe Iceberg deletion representation.
   - Produces a strong ETag from the public deletion representation for conditional UNDROP.
   - Keeps metadata locations, FileIO configuration, credentials, and cleanup internals out of the public model.

**After this PR:** the server can resolve an exact retained table generation and derive its strong ETag without loading the live Iceberg catalog. The public `deleted=true` table API is added by #12226.

Review order: #12222 → #12223 → #12225 → #12228 → **#12229** → #12226 → #12227.

Incremental diff:
https://github.com/nevzheng/gravitino/compare/codex/iceberg-rest-delete-20-01b-delete-api-cache...codex/iceberg-rest-delete-20-01c-deleted-read-model

Deleted-table APIs and conditional UNDROP need one generation-safe storage/read boundary that joins the retained table identity with its deletion action and supplies a stable public precondition token.

Related: #12221

No. This PR adds the internal joined read model and ETag derivation. The existing table API exposes it in #12226.

The review-unit head passes focused joined-row storage, exact-name lookup, lifecycle mapping, null-retention, safe representation, and strong ETag tests.


nevzheng:
This PR is superseded by the simplified implementation in #12226.

Deleted-table discovery now reuses the existing paginated table-list API with `deleted=true` and returns ordinary table identifiers. We intentionally removed the specialized retained-deletion read model, public deletion ID, detailed deleted-table representation, and deletion-specific ETag because a name list is sufficient for the current UNDROP workflow.

The final Stack 20 review order is #12225 → #12228 → #12226 → #12227.

PR #12231, opened 2026-07-28
[#12230] fix(auth): stop a lost JCasbin role policy load from denying forever

### LINKED ISSUE #12230: [Bug report] A single node denies forever when a JCasbin role policy load is lost
### Version

main branch

### Describe what's wrong

On a multi-node deployment with authorization enabled, a single node can permanently deny a user that its peers authorize correctly. The node keeps serving other requests normally, and only a restart clears it.

The node reaches a state where `loadedRoles` records a role as loaded while the JCasbin enforcer holds none of that role's `p` rows. `versionCheckAndLoadRoles` decides whether to reload by comparing `role_meta.updated_at` against `loadedRoles` only — it never inspects the enforcer. Since `role_meta.updated_at` does not move, the reload is skipped on every subsequent request and the missing privileges never come back.

Three defects combine to produce and then pin that state:

1. **A partial load is recorded as complete.** `loadPolicyByRoleEntity` silently skips a securable object whose metadata id cannot be resolved, but `versionCheckAndLoadRoles` calls `loadedRoles.put(roleId, dbUpdatedAt)` unconditionally afterwards. A role whose objects failed to resolve — even once, even transiently — is marked as loaded with zero or partial policies.

2. **Policy mutations are not serialized against cache eviction.** `SyncedEnforcer` makes each call atomic, but the `clearRolePolicies` → re-add sequence is not, and neither is it ordered against the `JcasbinLoadedRolesCache` removal listener, which clears policies from whichever thread drains Caffeine's maintenance queue. An eviction landing between another thread's policy writes and its `loadedRoles` update erases the rows that thread just wrote while leaving the marker claiming they are loaded.

3. **The TTL cannot expire the bad state away.** `JcasbinLoadedRolesCache` uses `expireAfterAccess`, unlike every other authorization cache. `versionCheckAndLoadRoles` probes the entry on every request carrying the role, so each *denied* request renews the very entry that suppresses the reload. On a node under steady traffic the entry never expires.

No exception is thrown. The authorization check simply returns `false`, and no role reload is attempted, so there is no `listSecurableObjectsByRoleId` query around the failure either.

### How to reproduce

Not reproducible on demand — it depends on a metadata-id lookup failing during a role load, or on an eviction interleaving with a concurrent load. It can be reproduced deterministically at the unit level:

- Stub `MetadataIdConverter.getID` to return `Optional.empty()`, invoke `versionCheckAndLoadRoles`, and observe that `loadedRoles` records the role although the enforcer holds no policy for it.
- Hold a lock around the enforcer mutations and observe that `clearRolePolicies` proceeds concurrently.
- Read a `JcasbinLoadedRolesCache` entry more often than its TTL and observe that it never expires.

### Additional context

Same area as #12169, but a distinct defect: #12169 fixed the removal listener deleting the role's grouping rows along with its policies, whereas this one is about the enforcer losing the policies while `loadedRoles` still claims they are present.

Observed on a three-pod deployment: one pod denied `USE_CATALOG` on a metalake six times over an hour while the other two pods hit the same role's policy throughout. The failing pod never logged a role reload; a peer recovered on its own only after its cache entry went idle long enough to expire.

yuqi1129:

Ensures a role can never be recorded as loaded unless the enforcer actually holds its policies, and that nothing can erase those policies behind the loader's back.

1. **Record a role as loaded only when its policies fully loaded.** `loadPolicyByRoleEntity` is split into `resolveRolePolicies` (resolves every securable object to a metadata id, may hit the DB) and `applyRolePolicies` (writes rows into the enforcers, pure in-memory). The intermediate `ResolvedRolePolicies` carries the objects that failed to resolve, and `loadedRoles.put` now runs only when that list is empty. Incomplete loads are retried instead of being pinned, throttled by a short backoff (`partialRoleLoadBackoff`, 10s) so that a permanently unresolvable object — a role still referencing a dropped table, say — cannot turn into a DB round-trip per request. `handleRolePrivilegeChange` drops the backoff so an explicit privilege change is retried immediately.

2. **Serialize role policy mutations with `rolePolicyLock`.** The lock covers the clear → apply → record sequence and `clearRolePolicies` itself, which the `JcasbinLoadedRolesCache` removal listener calls from arbitrary threads. Metadata ids are resolved before the lock is taken, so no DB call ever runs inside the critical section. A single reentrant lock is used rather than per-role striping: a `loadedRoles` write performed under the lock can evict a *different* role and re-enter `clearRolePolicies`, which striped locks would expose to lock-order inversion.

   The clear before a reload is now unconditional. It was previously skipped when `loadedRoles` had no entry, but rows can exist without an entry after an incomplete load, and re-adding over them would leave stale rows behind.

3. **`JcasbinLoadedRolesCache` switches to `expireAfterWrite`**, matching every other authorization cache.

Denials also report the enforcer state behind them, escalating to `WARN` when a role is recorded as loaded but carries no allow policy. The scan is gated on `DEBUG` because denials are an ordinary outcome on the request path; enable `DEBUG` for `JcasbinAuthorizer` on the node under investigation and the diagnosis itself still arrives at `WARN`.

A node could reach a state where `loadedRoles` claims a role is loaded while the enforcer holds none of its policies. The version check consults `loadedRoles` alone, so it never notices and keeps skipping the reload; every request carrying that role is denied on that node while peers serve the same user correctly. See the issue for the three defects that combine to create and then pin that state.

Fix: #12230

No. No API change and no new configuration property.

`./gradlew :server-common:test -PskipITs` — 245 tests, 0 failures, including the 52 existing `TestJcasbinAuthorizer` cases.

Four new tests, each verified to fail when the corresponding fix is reverted:

- `TestJcasbinAuthorizer#testPartialPolicyLoadIsNotRecordedAsLoaded` — a role whose securable object does not resolve is not recorded as loaded, the backoff suppresses the immediate retry, and the role loads and is recorded once the object resolves again. Reverting fix 1 fails it on *"a role whose policies could not be loaded must not be recorded as loaded"*.
- `TestJcasbinAuthorizer#testClearRolePoliciesSerializesWithRolePolicyLock` — `clearRolePolicies` blocks while the mutation lock is held and proceeds once it is released. Reverting fix 2 fails it on *"clearRolePolicies must not proceed while the role policy lock is held"*.
- `TestJcasbinLoadedRolesCache#testTtlIsWriteBasedSoReadsCannotKeepAnEntryAlive` — reads more frequent than the TTL do not keep an entry alive, and expiry still clears the role's policies. Reverting fix 3 fails it.
- `TestJcasbinLoadedRolesCache#testReplacingAnEntryDoesNotClearPolicies` — guards the existing behaviour that a refresh must not be treated as a removal.


yuqi1129:
Convert to draft as it's still under verification.

roryqi:
Thanks for the fix. I reviewed the role-policy loading path and the new tests; the main approach looks sound to me: resolving metadata ids before taking `rolePolicyLock`, then serializing `clear -> apply -> record`, closes the loaded-marker/enforcer-state race without holding the lock across DB lookups.

I only noticed two non-blocking follow-ups:

1. `authorizeByJcasbin` returns early for active-role narrowing (`narrowByActiveRoles && !activeRoles.isAll()`), so denied narrowed requests do not call `diagnoseDenial`. That does not affect the core fix, but it means the new denial diagnostics will miss one request path.

2. The cache architecture comment in `JcasbinAuthorizer` still says loaded role policies use access-based TTLs through `JcasbinLoadedRolesCache`; this PR changes that cache to write-based TTL, so the comment should be updated.

I also verified the focused tests locally on the PR head:

```bash
./gradlew :server-common:test \
  --tests 'org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer' \
  --tests 'org.apache.gravitino.server.authorization.jcasbin.TestJcasbinLoadedRolesCache' \
  -PskipITs
```

Result: `BUILD SUCCESSFUL`.

roryqi:
One more correctness concern about the partial-load semantics:

This PR intentionally keeps resolved policies in the enforcer even when the role is incomplete, and only avoids recording the role in `loadedRoles`. That preserves availability for resolved privileges, but it may weaken DENY semantics.

Example:

```text
Role A grants ALLOW SELECT on table1.
Role B grants DENY SELECT on table1.
User has both roles.

During Role B load, table1 cannot be resolved, so the DENY row is not added to denyEnforcer.
Role B is not marked loaded, but partialRoleLoadBackoff suppresses immediate retry.

A later SELECT table1 request resolves table1 successfully.
allowEnforcer has Role A allow, denyEnforcer is missing Role B deny.
Result may become allowed, although DENY should have won.
```

So the current partial-load behavior is not strictly fail-closed. Missing ALLOW rows usually cause false denies, but missing DENY rows can cause false allows.

Can we clarify the intended invariant here? If DENY must remain conservative, I think an incomplete role should probably not apply any ALLOW rows until the role is fully resolved, or partial loads should otherwise fail closed for the whole role.

yuqi1129:
> Thanks for the fix. I reviewed the role-policy loading path and the new tests; the main approach looks sound to me: resolving metadata ids before taking `rolePolicyLock`, then serializing `clear -> apply -> record`, closes the loaded-marker/enforcer-state race without holding the lock across DB lookups.
> 
> I only noticed two non-blocking follow-ups:
> 
> 1. `authorizeByJcasbin` returns early for active-role narrowing (`narrowByActiveRoles && !activeRoles.isAll()`), so denied narrowed requests do not call `diagnoseDenial`. That does not affect the core fix, but it means the new denial diagnostics will miss one request path.
> 2. The cache architecture comment in `JcasbinAuthorizer` still says loaded role policies use access-based TTLs through `JcasbinLoadedRolesCache`; this PR changes that cache to write-based TTL, so the comment should be updated.
> 
> I also verified the focused tests locally on the PR head:
> 
> ```shell
> ./gradlew :server-common:test \
>   --tests 'org.apache.gravitino.server.authorization.jcasbin.TestJcasbinAuthorizer' \
>   --tests 'org.apache.gravitino.server.authorization.jcasbin.TestJcasbinLoadedRolesCache' \
>   -PskipITs
> ```
> 
> Result: `BUILD SUCCESSFUL`.

Modified.

roryqi (approved):
LGTM.

PR #12234, opened 2026-07-28
[#12233] refactor(core): Support relation targets

### LINKED ISSUE #12233: [Subtask] Store tag assignment values
### Describe the subtask

Store tag assignment values in the relational storage layer.

### Parent issue

https://github.com/apache/gravitino/issues/12127

roryqi:

Introduce object-based relation query and update requests for relation-edge metadata.

This PR adds `RelationQuery`, `RelationUpdate`, and `RelationEdgeTarget` as the canonical generic relation APIs. Existing primitive `SupportsRelationOperations` methods are kept as compatibility adapters, while relation-value reads and value-carrying updates now evolve through the request/update objects instead of tag-specific or value-specific overloads.

`RelationalEntityStore` delegates the object APIs, validates relation target entity types before backend writes, and invalidates relation caches after backend writes using the same derived target type.

This is the first split from the original tag assignment value storage PR. The schema and tag storage implementation are kept in `tag-assignment-values-storage-next` for the follow-up PR.

Tag assignment values need relation-level metadata, but the store contract should stay relation-generic instead of adding tag-specific methods.

Fix: #12233

No. This is internal core/store infrastructure only.

`./gradlew :core:spotlessApply`

`./gradlew :core:test --tests org.apache.gravitino.TestSupportsRelationOperations --tests org.apache.gravitino.storage.relational.TestRelationalEntityStore -PskipITs -PskipDockerTests=false`

### DIFF
--- a/core/src/main/java/org/apache/gravitino/RelationUpdate.java
+++ b/core/src/main/java/org/apache/gravitino/RelationUpdate.java
@@ -22,8 +22,13 @@ import com.google.common.base.Preconditions;
 import java.util.Arrays;
 
 /**
- * Describes a relation update from a source entity to target endpoints. Target endpoint identity
- * and relation-edge attributes are carried by {@link RelationEdgeTarget}.
+ * Describes an immutable relation update from a source entity to target endpoints. Target endpoint
+ * identity and relation-edge attributes are carried by immutable {@link RelationEdgeTarget}
+ * instances.
+ *
+ * <p>The target endpoint arrays are copied when the update is created and when the targets are
+ * returned. The copy protects the array container from external mutation; the elements themselves
+ * are immutable.
  */
 public final class RelationUpdate {
 
@@ -93,14 +98,14 @@ public final class RelationUpdate {
   }
 
   /**
-   * @return A defensive copy of target endpoints to associate with the source entity.
+   * @return A copy of the target endpoint array to associate with the source entity.
    */
   public RelationEdgeTarget[] targetsToAdd() {
     return targetsToAdd.clone();
   }
 
   /**
-   * @return A defensive copy of target endpoints to disassociate from the source entity.
+   * @return A copy of the target endpoint array to disassociate from the source entity.
    */
   public RelationEdgeTarget[] targetsToRemove() {
     return targetsToRemove.clone();
--- a/core/src/main/java/org/apache/gravitino/RelationUpdate.java
+++ b/core/src/main/java/org/apache/gravitino/RelationUpdate.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.gravitino;
+
+import com.google.common.base.Preconditions;
+import java.util.Arrays;
+
+/**
+ * Describes a relation update from a source entity to target endpoints. Target endpoint identity
+ * and relation-edge attributes are carried by {@link RelationEdgeTarget}.
+ */
+public final class RelationUpdate {
+
+  private static final RelationEdgeTarget[] EMPTY_TARGETS = new RelationEdgeTarget[0];
+
+  private final SupportsRelationOperations.Type relationType;
+  private final NameIdentifier sourceIdentifier;
+  private final Entity.EntityType sourceEntityType;
+  private final RelationEdgeTarget[] targetsToAdd;
+  private final RelationEdgeTarget[] targetsToRemove;
+
+  private RelationUpdate(
+      SupportsRelationOperations.Type relationType,
+      NameIdentifier sourceIdentifier,
+      Entity.EntityType sourceEntityType,
+      RelationEdgeTarget[] targetsToAdd,
+      RelationEdgeTarget[] targetsToRemove) {
+    this.relationType = relationType;
+    this.sourceIdentifier = sourceIdentifier;
+    this.sourceEntityType = sourceEntityType;
+    this.targetsToAdd = copyTargets(targetsToAdd, "targetsToAdd");
+    this.targetsToRemove = copyTargets(targetsToRemove, "targetsToRemove");
+  }
+
+  /**
+   * Creates a relation update.
+   *
+   * @param relationType The type of relation.
+   * @param sourceIdentifier The identifier of the source entity whose relations are being updated.
+   * @param sourceEntityType The source entity type.
+   * @param targetsToAdd Target endpoints to associate with the source entity.
+   * @param targetsToRemove Target endpoints to disassociate from the source entity.
+   * @return A relation update.
+   */
+  public static RelationUpdate of(
+      SupportsRelationOperations.Type relationType,
+      NameIdentifier sourceIdentifier,
+      Entity.EntityType sourceEntityType,
+      RelationEdgeTarget[] targetsToAdd,
+      RelationEdgeTarget[] targetsToRemove) {
+    Preconditions.checkArgument(relationType != null, "relationType must not be null");
+    Preconditions.checkArgument(sourceIdentifier != null, "sourceIdentifier must not be null");
+    Preconditions.checkArgument(sourceEntityType != null, "sourceEntityType must not be null");
+    return new RelationUpdate(
+        relationType, sourceIdentifier, sourceEntityType, targetsToAdd, targetsToRemove);
+  }
+
+  /**
+   * @return The type of relation.
+   */
+  public SupportsRelationOperations.Type relationType() {
+    return relationType;
+  }
+
+  /**
+   * @return The source entity identifier.
+   */
+  public NameIdentifier sourceIdentifier() {
+    return sourceIdentifier;
+  }
+
+  /**
+   * @return The source entity type.
+   */
+  public Entity.EntityType sourceEntityType() {
+    return sourceEntityType;
+  }
+
+  /**
+   * @return A defensive copy of target endpoints to associate with the source entity.
+   */
+  public RelationEdgeTarget[] targetsToAdd() {
+    return targetsToAdd.clone();
+  }
+
+  /**
+   * @return A defensive copy of target endpoints to disassociate from the source entity.
+   */
+  public RelationEdgeTarget[] targetsToRemove() {
+    return targetsToRemove.clone();
+  }
+
+  /**
+   * @return Whether any target endpoint carries relation values.
+   */
+  public boolean hasRelationValues() {
+    return Arrays.stream(targetsToAdd).anyMatch(target -> target.relationValue().isPresent())
+        || Arrays.stream(targetsToRemove).anyMatch(target -> target.relationValue().isPresent());
+  }
+
+  private static RelationEdgeTarget[] copyTargets(
+      RelationEdgeTarget[] targets, String parameterName) {
+    if (targets == null) {
+      return EMPTY_TARGETS;
+    }
+
+    for (RelationEdgeTarget target : targets) {
+      Preconditions.checkArgument(target != null, "%s must not contain null", parameterName);
+    }
+
+    return targets.clone();
+  }
+}
--- a/core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
+++ b/core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
@@ -19,6 +19,7 @@
 package org.apache.gravitino;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
 
@@ -60,7 +61,8 @@ public interface SupportsRelationOperations {
   }
 
   /**
-   * List the entities according to a given entity in a specific relation.
+   * List the entities according to a given entity in a specific relation. Compatibility adapter for
+   * {@link #listEntitiesByRelation(RelationQuery)}.
    *
    * @param <E> the type of entities returned.
    * @param relType The type of relation.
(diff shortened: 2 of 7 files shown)


jerryshao:
Can we split this PR? This PR seems too big to review.

core/src/main/java/org/apache/gravitino/meta/AssignedTagEntity.java
```diff
@@ -0,0 +1,133 @@
...
+import org.apache.gravitino.Entity;
+import org.apache.gravitino.Field;
+import org.apache.gravitino.HasIdentifier;
+import org.apache.gravitino.Namespace;
+import org.apache.gravitino.tag.Tag;
+import org.apache.gravitino.tag.TagAssignment;
+import org.apache.gravitino.tag.TagValueConstraint;
+
+/** A tag entity in a metadata-object relation context, with assignment values from the relation. */
+public final class AssignedTagEntity implements Tag, Entity, Auditable, HasIdentifier {
```
jerryshao: Can we directly update the `TagEntity` class, instead of creating a new one?

scripts/mysql/schema-2.0.0-mysql.sql
```diff
@@ -299,14 +300,16 @@ CREATE TABLE IF NOT EXISTS `tag_relation_meta` (
     `tag_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'tag id',
     `metadata_object_id` BIGINT(20) UNSIGNED NOT NULL COMMENT 'metadata object id',
     `metadata_object_type` VARCHAR(64) NOT NULL COMMENT 'metadata object type',
+    `tag_value` VARCHAR(256) DEFAULT NULL COMMENT 'tag relation value',
+    `value_order` INT UNSIGNED NOT NULL DEFAULT 0 COMMENT 'tag relation value order',
```
jerryshao: Why do we need this field?

roryqi: To give the values a stable order. I can remove this field. Because it makes others confused.

roryqi:
Split this PR as requested. This PR now only contains the generic relation-target infrastructure:

- `RelationTarget`
- `SupportsRelationOperations` overloads for exact relation-value listing and target-based relation updates
- `RelationalEntityStore` delegation and cache invalidation for relation targets
- focused tests for the new generic behavior

The tag schema and storage implementation has been moved to `tag-assignment-values-storage-next` and will be opened as the follow-up PR after this one. The old inline comments are now outdated against this smaller diff; in the follow-up branch I also removed `value_order` and documented that `AssignedTagEntity.fields()` exposes only tag definition fields while assignment values are exposed via `Tag.assignment()`.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -490,6 +491,41 @@ public <E extends Entity & HasIdentifier> List<E> updateEntityRelations(
...
+      RelationTarget[] destEntitiesToAdd,
+      RelationTarget[] destEntitiesToRemove)
+      throws IOException, NoSuchEntityException, EntityAlreadyExistsException {
+    List<E> result =
+        backend.updateEntityRelations(
+            relType, srcEntityIdent, srcEntityType, destEntitiesToAdd, destEntitiesToRemove);
+
+    cache.invalidate(srcEntityIdent, srcEntityType, relType);
+    invalidateRelationTargetCache(relType, destEntitiesToAdd);
+    invalidateRelationTargetCache(relType, destEntitiesToRemove);
```
jerryshao: Do we need lock for such operations?

jerryshao:
@yuqi1129 can you please help to review?

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
```diff
@@ -490,6 +491,41 @@ public <E extends Entity & HasIdentifier> List<E> updateEntityRelations(
     return result;
   }
 
+  @Override
+  public <E extends Entity & HasIdentifier> List<E> listEntitiesByRelation(
+      Type relType,
+      NameIdentifier nameIdentifier,
+      Entity.EntityType identType,
+      boolean allFields,
+      String relationValue)
```
jerryshao: Is value "string" enough?

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
```diff
@@ -92,6 +94,55 @@ <E extends Entity & HasIdentifier> List<E> listEntitiesByRelation(
       Type relType, NameIdentifier nameIdentifier, Entity.EntityType identType, boolean allFields)
       throws IOException;
 
+  /**
+   * List the entities according to a given entity in a specific relation, optionally filtering by
+   * an exact value carried by the relation row.
+   *
+   * @param <E> the type of entities returned.
+   * @param relType The type of relation.
+   * @param nameIdentifier The given entity identifier.
```
yuqi1129: Is this parameter the name of the from point or the end point? I think you need to clarify it.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
```diff
@@ -92,6 +94,55 @@ <E extends Entity & HasIdentifier> List<E> listEntitiesByRelation(
...
+   * @param identType The entity type of parameter nameIdentifier represents.
+   * @param relationValue The exact relation value to match, or null to return all relations.
```
yuqi1129: What's the meaning of the relation value here? More context is needed here.

core/src/main/java/org/apache/gravitino/RelationTarget.java
```diff
@@ -0,0 +1,77 @@
...
+
+import com.google.common.base.Preconditions;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * Represents a target entity in a relation update, with optional metadata carried by the relation
+ * row.
+ */
+public final class RelationTarget {
```
yuqi1129: Does this represent the point of edges?

yuqi1129: I would suggest making `RelationTarget` a subclass of `NameIdentifier` or extracting an interface to get `NameIdentifier`. Adding a new method in `SupportsRelationOperations` is not elegant...

roryqi: Let me think about it.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: For the current use case, yes. The relation value maps to the tag assignment value, which the public API models as a String and the relational storage stores as a VARCHAR. I clarified the Javadocs to call this an exact string value carried by the relation edge. If a future relation needs structured relation metadata, we should introduce a separate typed payload instead of overloading this field.

core/src/main/java/org/apache/gravitino/storage/relational/RelationalEntityStore.java
roryqi: I do not think we need an extra store-level cache lock here. The backend update owns the transactional write, and this path does not read/populate the cache. The invalidation calls already take the cache segmented lock internally. I kept the same post-backend invalidation ordering as the existing update path, and also fixed the legacy NameIdentifier[] overload to invalidate the destination relation cache with the actual destination type.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
roryqi: Clarified in the Javadocs. This parameter is the anchor entity for the relation lookup; it can be either endpoint of the relation. `identType` tells which endpoint it represents, and the returned entities are from the opposite endpoint.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
roryqi: Added more context in the Javadocs. The relation value is metadata stored on the relation edge, not on either endpoint entity. For this feature it is the tag assignment value; null means no value filter and delegates to the existing relation listing behavior.

core/src/main/java/org/apache/gravitino/RelationTarget.java
roryqi: Yes. I clarified `RelationTarget` as the target endpoint of a relation edge. The source endpoint is passed separately to `updateEntityRelations`, and the optional relation value belongs to the edge.

roryqi: I checked this direction. `NameIdentifier` has a private constructor and represents only the name path, while this object also needs the entity type and optional edge value. Making it a subclass would require widening `NameIdentifier` and would make relation-edge metadata look like an identifier property. I kept `RelationTarget` as a wrapper so the existing `NameIdentifier[]` overload stays intact, and value-carrying updates use an explicit target endpoint object. The Javadocs now call this out as the target endpoint of the relation edge.

yuqi1129: Thanks. I agree that `RelationTarget` should not extend `NameIdentifier`; relation-edge metadata should not become identifier state. However, that only addresses the inheritance option and does not address my main concern about API extensibility.

A single optional `relationValue` already doubles the `listEntitiesByRelation` overloads and adds a parallel `updateEntityRelations` overload. The suggestion that a future structured payload would require another type indicates that this pattern will keep expanding `SupportsRelationOperations`.

I understand that the existing `NameIdentifier[]` method needs to remain for compatibility. Keeping it as a compatibility adapter is fine, but we should introduce one canonical, extensible abstraction—for example, `RelationQuery` for reads and `RelationUpdate`/`RelationEdgeTarget` with relation attributes for writes—so future relation metadata evolves inside those objects rather than through more overloads.

An interface exposing `NameIdentifier` may still help with conversion, but the more important point is to separate endpoint identity from edge attributes and define which method is the canonical SPI for implementers.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
```diff
@@ -220,4 +283,55 @@ default <E extends Entity & HasIdentifier> List<E> updateEntityRelations(
...
+      Entity.EntityType srcEntityType,
+      RelationTarget[] destEntitiesToAdd,
+      RelationTarget[] destEntitiesToRemove)
+      throws IOException, NoSuchEntityException, EntityAlreadyExistsException {
+    if (hasRelationValues(destEntitiesToAdd) || hasRelationValues(destEntitiesToRemove)) {
+      throw new UnsupportedOperationException(
+          "updateEntityRelations with relation values is not supported by this implementation");
+    }
+
+    return updateEntityRelations(
```
yuqi1129: The compatibility path drops `RelationTarget.entityType()` when converting targets to `NameIdentifier[]`, but `RelationalEntityStore` later uses the original entity type for cache invalidation. Since the target type is not validated here, the backend can update the actual relation while the store invalidates a different cache key, leaving the reverse-side relation cache stale.

For example, a `TAG_METADATA_OBJECT_REL` target incorrectly marked as `TABLE` can still be written through the legacy backend method, while `(tagIdent, TABLE, relType)` is invalidated instead of the actual `(tagIdent, TAG, relType)` cache entry. Please validate or derive the destination type before the backend write and use the same validated type for invalidation.

roryqi: Fixed in `3a301a664`. `RelationalEntityStore` now converts legacy `NameIdentifier[]` updates into `RelationUpdate` with the derived target type, validates `RelationEdgeTarget.entityType()` before the backend write, and invalidates target caches using that same derived type. Added regression coverage for mismatched target type before backend interaction.

core/src/main/java/org/apache/gravitino/RelationTarget.java
roryqi: Fixed in `3a301a664`. I replaced the value-specific list overload and the parallel target-array update overload with object-based APIs: `RelationQuery` for reads and `RelationUpdate`/`RelationEdgeTarget` for writes. The existing primitive overloads are kept as compatibility adapters, and future edge metadata can now evolve inside those objects instead of adding more overloads.

jerryshao:
@yuqi1129 , please help to review this PR. Thanks.

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
```diff
@@ -217,7 +237,55 @@ default <E extends Entity & HasIdentifier> List<E> updateEntityRelations(
...
+    Entity.EntityType targetEntityType = relationDestinationType(relType, srcEntityType);
+    return Arrays.stream(nameIdentifiers)
+        .map(nameIdentifier -> RelationEdgeTarget.of(nameIdentifier, targetEntityType, null))
+        .toArray(RelationEdgeTarget[]::new);
+  }
+
+  private static Entity.EntityType relationDestinationType(
+      Type relType, Entity.EntityType srcEntityType) {
+    switch (relType) {
+      case POLICY_METADATA_OBJECT_REL:
```
yuqi1129, line 290: I wonder what the result of the method would be if we want to query all metadata objects that have a specific policy or tag? Is it the parameter still `POLICY_METADATA_OBJECT_REL` or `TAG_METADATA_OBJECT_REL`.

core/src/main/java/org/apache/gravitino/RelationUpdate.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  public RelationEdgeTarget[] targetsToAdd() {
+    return targetsToAdd.clone();
+  }
+
+  /**
+   * @return Target endpoints to disassociate from the source entity.
+   */
+  public RelationEdgeTarget[] targetsToRemove() {
+    return targetsToRemove.clone();
```
yuqi1129, line 111: What's the benefit of cloning the array?

core/src/main/java/org/apache/gravitino/SupportsRelationOperations.java
roryqi: Clarified in `1e1f55001`. Reverse lookup still uses the same relation type: querying metadata objects for a policy uses `POLICY_METADATA_OBJECT_REL` with the policy as the `RelationQuery` anchor, and querying metadata objects for a tag uses `TAG_METADATA_OBJECT_REL` with the tag as the anchor. The helper was renamed to `relationUpdateTargetType` to make clear it is only for the legacy update adapter target inference, not query result typing.

core/src/main/java/org/apache/gravitino/RelationUpdate.java
roryqi: Clarified in `1e1f55001`. The clone is a defensive copy: `RelationUpdate` is intended to be immutable, but Java arrays are mutable, so the constructor and accessors copy arrays to prevent callers from mutating the stored update after creation or through a returned array.

yuqi1129: A shadow copy for an array will only create a new array, but the elements in the array will remain pointed to the old elements. If the elements change in the original array, the elements in the new one will also change.

roryqi: Correct, this is a shallow array copy. For this API that is sufficient because `RelationEdgeTarget` is immutable: it is final, has private final fields, a private constructor, no mutators, and only contains immutable endpoint/value data. I clarified the Javadocs in `0662d7171` to say the copy protects the array container from external mutation, and added a unit test that mutating the input arrays or returned arrays does not affect the stored `RelationUpdate`.

yuqi1129:
I have no more comments, and it looks good to me.

yuqi1129 (approved):
Just a leave me

PR #12235, opened 2026-07-28
[#12232] fix(core): Preserve metadata during concurrent drop and rename

### LINKED ISSUE #12232: [Bug report] Concurrent drop and rename on an external-backed table silently deletes the surviving table's owner/tag/policy/role relations
### Version

main branch

### Describe what's wrong

On an external-backed catalog (Hive, Glue, JDBC, Iceberg, …), `dropTable` deletes the store row unconditionally, even when the external drop found nothing (`TableOperationDispatcher.java:378-386`, `// In all situations, we try to delete the table from the store`). A `false` return means "the external system has no such table", which in a concurrent rename is not "nothing to clean up" but "the table is still there under a new name".

With two servers (or after TreeLock is removed), `rename t1 -> t2` on A and `drop t1` on B can interleave so that:

1. A's external rename lands first, so t2 exists in the external catalog;
2. B's external drop finds no t1 and returns `false`, but still runs `store.delete(t1)`;
3. A's store update for t1 -> t2 then matches 0 rows and writes nothing.

The damage is not limited to the entity row. `TableMetaService.deleteTable` cascades, soft-deleting the owner relation, the securable-object (role/privilege) relations, tag relations, policy relations, statistics and table-version rows for that table id, and appends a `DROP` to `entity_change_log`.

`importTable` (`TableOperationDispatcher.java:474-525`) restores only the entity row and its columns in a single `store.put`, so re-import on the next `loadTable(t2)` does **not** bring any of that back — **not even when the id is preserved** (Hive/Iceberg carry `gravitino.identifier` in table properties and rename carries them to t2, so the id is reused but the relation rows stay soft-deleted). The loss is silent and has no recovery path, and it changes authorization state.

The caller is not told either: the failed store update throws `NoSuchEntityException`, which `OperationDispatcher.operateOnEntity` (`OperationDispatcher.java:204-226`) swallows with a `LOG.error`, while the REST response is built from the connector result — so the rename is reported as successful.

No exception reaches the client. The only trace is a server-side log line from `operateOnEntity`:

```
ERROR ... Entity of {} does not exist   // FormattedErrorMessages.ENTITY_NOT_FOUND
```

### How to reproduce

- Gravitino main branch, an external-backed catalog (e.g. Hive), two servers sharing one metadata store (or a single server with the alter/drop dispatcher paths driven concurrently — the TreeLock write lock only serializes them within one JVM).
- Create table `db.t1`, set an owner and attach a tag/policy to it.
- Concurrently: server A `alterTable(db.t1, rename -> t2)`, server B `dropTable(db.t1)`.
- With the interleaving above: `db.t2` exists in Hive, `loadTable(db.t2)` succeeds (re-imported), but its owner, tags and policies are gone, and no error was returned to either caller.

### Additional context

Proposed fix, two parts:

1. Do not touch the store when the external drop returned `false`. The behaviour this gives up — cleaning a store row after an out-of-band drop — is already argued to be harmless elsewhere: `loadTable` asks the external system first so a stale row is never trusted, and a later create overwrites it with `overwrite = true`. Leaving those rows to the reconcile path trades a benign stale row for an irreversible loss of Gravitino-only metadata.
2. Make the failure observable: a rename whose store update matches 0 rows should not be reduced to a log line — surface it (metric/warning) so operators can detect "renamed externally, registration not updated".

Two related hardenings are tracked separately and are **not** part of this issue: deleting by `(id, current_version)` rather than by name (which fixes the drop/re-create ABA but does not fix this race), and moving the relation cascade to a deferred GC so a wrong delete stays recoverable.

yuqi1129:

- Delete unmanaged table, view, and schema registrations only when the external catalog confirms the drop or purge.
- Surface store failures around external table and view renames instead of reporting success.
- Add regression tests for the related drop, purge, and rename paths.

A concurrent drop and rename can cause the drop to return `false` because the object has already been renamed externally. Deleting the old stored registration in this case removes metadata associated with the surviving object.

The same two-phase external-catalog and Gravitino-store behavior exists for views, and an out-of-band schema rename has the corresponding stale-name drop risk. Table purge also followed the unsafe deletion path.

The rename paths could additionally perform an external rename when the stored registration could not be read or updated, leaving inconsistent state while reporting success.

Fix: #12232

Yes. For unmanaged tables, views, and schemas, a drop or purge that returns `false` no longer deletes the stored registration. Table and view renames now return an error when Gravitino cannot safely read or update the stored registration.

No API or configuration changes are introduced.

- Added regression coverage to `TestTableOperationDispatcher`, `TestViewOperationDispatcher`, and `TestSchemaOperationDispatcher`.
- Ran `./gradlew spotlessApply`.
- Ran `./gradlew :core:check -PskipITs -PskipDockerTests=true`.
- Ran `git diff --check`.

### DIFF
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
@@ -339,20 +339,17 @@ public class SchemaOperationDispatcher extends OperationDispatcher implements Sc
             return droppedFromCatalog;
           }
 
-          // For the unmanaged schema, it could happen that the schema:
-          // 1. It's not found in the catalog (dropped directly from underlying sources)
-          // 2. It's found in the catalog but not in the store (not managed by Gravitino)
-          // 3. It's found in the catalog and the store (managed by Gravitino)
-          // 4. Neither found in the catalog nor in the store.
-          // In all situations, we try to delete the schema from the store, but we don't take the
-          // return value of the store operation into account. We only take the return value of the
-          // catalog into account.
-          try {
-            store.delete(ident, SCHEMA, true);
-          } catch (NoSuchEntityException e) {
-            LOG.warn("The schema to be dropped does not exist in the store: {}", ident, e);
-          } catch (Exception e) {
-            throw new RuntimeException(e);
+          // A false result can mean that the schema was renamed directly in the external catalog.
+          // Only remove the stored registration after the catalog confirms that this drop deleted
+          // the schema and its descendants.
+          if (droppedFromCatalog) {
+            try {
+              store.delete(ident, SCHEMA, true);
+            } catch (NoSuchEntityException e) {
+              LOG.warn("The schema to be dropped does not exist in the store: {}", ident, e);
+            } catch (Exception e) {
+              throw new RuntimeException(e);
+            }
           }
 
           SchemaEntityCleaner.deleteOrphanedSchemaEntities(
--- a/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
@@ -339,9 +339,10 @@ public class SchemaOperationDispatcher extends OperationDispatcher implements Sc
             return droppedFromCatalog;
           }
 
-          // A false result can mean that the schema was renamed directly in the external catalog.
-          // Only remove the stored registration after the catalog confirms that this drop deleted
-          // the schema and its descendants.
+          // A false result is ambiguous: the external schema may have been renamed or dropped out
+          // of band. Preserve the registration because deleting it after a rename would lose
+          // Gravitino-only metadata. A true out-of-band drop can therefore leave a stale
+          // registration that requires separate cleanup.
           if (droppedFromCatalog) {
             try {
               store.delete(ident, SCHEMA, true);
--- a/core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
+++ b/core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
@@ -46,6 +46,7 @@ import org.apache.gravitino.Namespace;
 import org.apache.gravitino.StringIdentifier;
 import org.apache.gravitino.connector.HasPropertyMetadata;
 import org.apache.gravitino.connector.capability.Capability;
+import org.apache.gravitino.exceptions.GravitinoRuntimeException;
 import org.apache.gravitino.exceptions.NoSuchEntityException;
 import org.apache.gravitino.exceptions.NoSuchSchemaException;
 import org.apache.gravitino.exceptions.NoSuchTableException;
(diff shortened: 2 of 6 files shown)


yuqi1129:
@roryqi 
Could you help to review it？

core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
```diff
@@ -339,20 +339,17 @@ public boolean dropSchema(NameIdentifier ident, boolean cascade) throws NonEmpty
...
-          try {
-            store.delete(ident, SCHEMA, true);
-          } catch (NoSuchEntityException e) {
-            LOG.warn("The schema to be dropped does not exist in the store: {}", ident, e);
-          } catch (Exception e) {
-            throw new RuntimeException(e);
+          // A false result can mean that the schema was renamed directly in the external catalog.
+          // Only remove the stored registration after the catalog confirms that this drop deleted
+          // the schema and its descendants.
+          if (droppedFromCatalog) {
```
diqiu50, line 346: Once the external table is gone, dropTable always returns false, so the stored registration is never deleted — and there is no API left that can clean it up.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -374,20 +395,17 @@ public boolean dropTable(NameIdentifier ident) {
...
-          try {
-            store.delete(ident, TABLE);
-          } catch (NoSuchEntityException e) {
-            LOG.warn("The table to be dropped does not exist in the store: {}", ident, e);
-          } catch (Exception e) {
-            throw new RuntimeException(e);
+          // A false result can mean that a concurrent rename already moved the external table.
+          // Only remove the stored registration after the catalog confirms that this drop deleted
+          // the table.
+          if (droppedFromCatalog) {
```
diqiu50, line 404: The same issue

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
```diff
@@ -338,6 +350,15 @@ public Table alterTable(NameIdentifier ident, TableChange... changes)
                   "UPDATE",
                   tableId);
 
+          if (isRenameTable && updatedTableEntity == null) {
```
diqiu50, line 355: Is using updatedTableEntity == null to determine whether the store update failed accurate?

core/src/main/java/org/apache/gravitino/catalog/SchemaOperationDispatcher.java
yuqi1129: Thanks for pointing this out. This thread is on `dropSchema`, but the same ambiguity exists for tables and views. A `false` result only tells us that the old name was absent when the external drop ran; it cannot distinguish a true out-of-band drop from a concurrent or out-of-band rename. Deleting the stored registration in the latter case irreversibly removes Gravitino-only metadata, so this PR deliberately preserves it as the safer default.

You are right that a true external drop can therefore leave a stale registration, and there is currently no generic API to explicitly remove it. The stale row is not returned by normal `list` or `load` operations because they consult the external catalog first, but explicit registration cleanup or reconciliation remains a separate limitation. We should not describe `false` as meaning that the object was dropped; it means only that this call did not find the old name to drop.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
yuqi1129: Yes, the same ambiguity applies to `dropTable` and `purgeTable`. If `false` is caused by a concurrent rename, deleting the old registration recreates the metadata-loss race fixed here. If it is caused by a true out-of-band drop, preserving it can leave stale metadata. Since the boolean result does not let us distinguish those cases, this PR chooses the non-destructive behavior. The lack of an explicit cleanup or reconciliation API is a real follow-up limitation rather than a reason to restore unconditional deletion.

core/src/main/java/org/apache/gravitino/catalog/TableOperationDispatcher.java
yuqi1129: `EntityStore.update` is contractually expected to return the updated entity or throw. `operateOnEntity` converts `NoSuchEntityException`, other store failures, and an ID mismatch into `null`, so this check means that no valid, matching updated registration was obtained.

That said, `null` does not identify only one failure mode, and the ID mismatch is currently detected after the store call. Therefore it is accurate as a consistency check, but less precise if described narrowly as determining whether the physical update ran. A stricter rename-specific update path that validates the expected ID inside the updater would make the semantics clearer.

roryqi (approved):
LGTM.

roryqi:
Do u need to cherry pick to branch 1.3?

yuqi1129:
> Do u need to cherry pick to branch 1.3?

I prefer to keep it in the main branch.

PR #12240, opened 2026-07-28
[#12237] fix(iceberg): de-flake IcebergClassLoaderPoolIT by tolerating concurrent JdbcCatalog view-migration on a shared iceberg_tables

### LINKED ISSUE #12237: [Bug report] IcebergClassLoaderPoolIT flaky: JdbcCatalog V1 view-migration re-runs "ADD COLUMN iceberg_type" and fails with "Duplicate column name"
### Version

main branch (also affects any deployment with two Iceberg JDBC catalogs on the same backend `uri`)

### Describe what's wrong

`IcebergClassLoaderPoolIT.testDropOneSharedCatalogKeepsSiblingUsable()` fails intermittently in the MySQL backend integration test.

The test deliberately places two Iceberg JDBC catalogs on the **same** backend `uri` (DB_A), so they share one `iceberg_tables` control table. Gravitino defaults `jdbc.schema-version=V1` (`IcebergCatalogUtil#loadJdbcCatalog`, `putIfAbsent(..., "V1")`), so every `JdbcCatalog.initialize` runs Iceberg's V1 view-support migration. That migration probes for the column with `getColumns(null, null, "iceberg_tables", "iceberg_type")` and, if the probe reports it missing, runs a non-idempotent `ALTER TABLE iceberg_tables ADD COLUMN iceberg_type` (no `IF NOT EXISTS`). When the column already exists on the shared/persisted table but the probe misses it, MySQL rejects the ALTER.

This is not test-only: any real deployment that loads two Iceberg JDBC catalogs pointing at the same JDBC `uri` shares `iceberg_tables` and can hit the same non-idempotent migration.

```
IcebergClassLoaderPoolIT > testDropOneSharedCatalogKeepsSiblingUsable() FAILED
  java.lang.RuntimeException: Failed to operate object operation [LIST] under [clp_drop_2_...],
    reason [Cannot check and eventually update SQL schema]
  Caused by: java.sql.SQLSyntaxErrorException: Duplicate column name 'iceberg_type'
    at org.apache.iceberg.jdbc.JdbcCatalog.executeV1CatalogUpdate(JdbcCatalog.java:276)
    at org.apache.iceberg.jdbc.JdbcCatalog.updateSchemaIfRequired(JdbcCatalog.java:235)
    at org.apache.iceberg.jdbc.JdbcCatalog.initialize(JdbcCatalog.java:157)
    at org.apache.gravitino.iceberg.common.ClosableJdbcCatalog.initialize(ClosableJdbcCatalog.java:76)
```

### How to reproduce

1. Run the MySQL backend integration test `:catalogs:catalog-lakehouse-iceberg:test` (deploy mode, `JDK17-deploy-mysql`).
2. `IcebergClassLoaderPoolIT` creates multiple Iceberg JDBC catalogs on the same MySQL `uri`; the shared `iceberg_tables` in DB_A persists across test methods/runs (cleanup only drops metalakes, never the control table).
3. On a subsequent init the V1 migration re-fires the `ADD COLUMN iceberg_type` against a table that already has the column, producing `Duplicate column name`.

### Additional context

- Iceberg 1.11.0; MySQL backend. Migration precondition is met because Gravitino defaults `jdbc.schema-version=V1`.
- The migration is Iceberg-side and non-idempotent; the practical fix belongs in Gravitino's JDBC catalog init (`ClosableJdbcCatalog#initialize`) — treat "column already exists" during the V1 view-migration as benign (schema already migrated by another catalog on the same `uri`), rather than failing init.
- Surfaced by the newly added `IcebergClassLoaderPoolIT` (PR #10480), which is why it appears now.

yuqi1129:

De-flake `IcebergClassLoaderPoolIT` and harden Iceberg JDBC catalog initialization when several logical catalogs share one JDBC metadata database:

- Detect the narrow V1 migration conflict where `JdbcCatalog.initialize` fails because the `iceberg_type` column already exists.
- Close the failed catalog's partially initialized JDBC connection pool and `FileIO` before discarding it. Iceberg creates these resources before the schema migration but does not register them with its `CloseableGroup` until the migration succeeds.
- Initialize a fresh catalog once after the conflict. The shared schema has already been migrated, so the retry can complete normally.
- Recognize MySQL, PostgreSQL, and SQLite duplicate-column error wordings in `isConcurrentViewMigrationConflict`.
- Keep the existing access-denied conversion null-safe and simplify the authentication branch without changing its behavior.

Iceberg JDBC catalogs that use the same backend `uri` share the physical `iceberg_tables` and `iceberg_namespace_properties` control tables. Different `catalog-backend-name` values isolate their logical namespaces and tables through the `catalog_name` column, so this is a valid, although relatively narrow, deployment topology.

Gravitino defaults `jdbc.schema-version` to `V1`. Iceberg creates `iceberg_tables` without `iceberg_type`, checks for that column with JDBC metadata, and then runs a non-idempotent migration:

```sql
ALTER TABLE iceberg_tables ADD COLUMN iceberg_type ...
```

The conflict can occur when:

1. Two JDBC catalog instances point to the same backend `uri`.
2. They initialize concurrently while the shared control table is being migrated to V1, or a JDBC metadata probe temporarily reports the already-added column as missing.
3. Both instances attempt the same `ALTER TABLE`; the losing instance fails with a duplicate-column error such as:

```text
Caused by: java.sql.SQLSyntaxErrorException: Duplicate column name 'iceberg_type'
  at org.apache.iceberg.jdbc.JdbcCatalog.executeV1CatalogUpdate(JdbcCatalog.java:276)
  at org.apache.iceberg.jdbc.JdbcCatalog.updateSchemaIfRequired(JdbcCatalog.java:235)
```

`IcebergClassLoaderPoolIT` deliberately creates multiple catalogs on the same MySQL JDBC URL because it verifies that catalogs with the same isolation key share a ClassLoader and that dropping one catalog does not break its sibling. This supported topology exposed the migration race in `JDK17-deploy-mysql`.

The duplicate-column error means another initializer has completed the shared migration. This patch therefore closes the losing catalog's partial resources and retries initialization once with a fresh catalog. Other SQL failures are still propagated unchanged.

Fix: #12237

No.

- Added classifier tests for MySQL, PostgreSQL, and SQLite duplicate-column messages, plus unrelated and null-message cases.
- Added a deterministic failed-migration test that simulates the duplicate `iceberg_type` error and verifies that the JDBC pool, underlying connection, and `FileIO` are closed.
- `./gradlew :iceberg:iceberg-common:test :iceberg:iceberg-common:spotlessCheck` — all 64 tests pass.
- `IcebergClassLoaderPoolIT` continues to cover the end-to-end shared-URI scenario in the deploy/MySQL CI.

### DIFF
--- a/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableJdbcCatalog.java
+++ b/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/ClosableJdbcCatalog.java
@@ -23,6 +23,7 @@ import java.io.Closeable;
 import java.util.Map;
 import java.util.function.Function;
 import javax.annotation.Nullable;
+import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.gravitino.catalog.hadoop.auth.KerberosClient;
 import org.apache.gravitino.iceberg.common.authentication.AuthenticationConfig;
 import org.apache.gravitino.iceberg.common.authentication.SupportsKerberos;
@@ -73,7 +74,12 @@ public class ClosableJdbcCatalog extends JdbcCatalog implements Closeable, Suppo
    */
   @Override
   public void initialize(String inputName, Map<String, String> properties) {
-    super.initialize(inputName, properties);
+    try {
+      super.initialize(inputName, properties);
+    } catch (RuntimeException e) {
+      closePartiallyInitializedResources(e);
+      throw e;
+    }
 
     AuthenticationConfig authenticationConfig = new AuthenticationConfig(properties);
     if (authenticationConfig.isKerberosAuth()) {
@@ -104,5 +110,25 @@ public class ClosableJdbcCatalog extends JdbcCatalog implements Closeable, Suppo
   @Override
   public <R> R doKerberosOperations(Executable<R> executable) throws Throwable {
     return KerberosCatalogUtils.doKerberosOperations(this.properties(), kerberosClient, executable);
+  }
+
+  private void closePartiallyInitializedResources(RuntimeException initializationException) {
+    // JdbcCatalog creates these resources before updating the JDBC schema, but registers them with
+    // its CloseableGroup only after the update succeeds. Its close() method therefore cannot clean
+    // them up when initialization fails during the schema update.
+    closePartiallyInitializedResource("connections", initializationException);
+    closePartiallyInitializedResource("io", initializationException);
+  }
+
+  private void closePartiallyInitializedResource(
+      String fieldName, RuntimeException initializationException) {
+    try {
+      Closeable resource = (Closeable) FieldUtils.readField(this, fieldName, true);
+      if (resource != null) {
+        resource.close();
+      }
+    } catch (Exception closeException) {
+      initializationException.addSuppressed(closeException);
+    }
   }
 }
--- a/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/utils/IcebergCatalogUtil.java
+++ b/iceberg/iceberg-common/src/main/java/org/apache/gravitino/iceberg/common/utils/IcebergCatalogUtil.java
@@ -57,6 +57,12 @@ public class IcebergCatalogUtil {
 
   private static final Logger LOG = LoggerFactory.getLogger(IcebergCatalogUtil.class);
 
+  /**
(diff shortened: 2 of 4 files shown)


yuqi1129:
@roryqi 
Can you help to review this PR?

roryqi (approved):
LGTM

Issue #11583, opened 2026-06-10
[Improvement] Implement fetch and cancel endpoints for async scan planning lifecycle


laserninja:
### What would you like to be improved?
Only the scan-plan submit endpoint (`planTableScan`) is implemented. The async lifecycle defined by the Iceberg REST spec is incomplete: when `planTableScan` returns a `submitted` status with a plan id, clients have no way to poll the result via `GET .../plan/{plan-id}` or cancel it via `DELETE .../plan/{plan-id}`. The sibling `POST .../tasks` endpoint is already tracked in #11284.

### How should we improve?
Implement `fetchPlanningResult` (`GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}`) and `cancelPlanning` (`DELETE .../plan/{plan-id}`), wiring them through the table operation dispatcher and `IcebergExceptionMapper`. Advertise the corresponding endpoints in `/v1/config` and add tests.

Related: scan-planning epic #7668 / #7261; sibling endpoint #11284; spec-compliance epic #10547.

PR #12241, opened 2026-07-28
[#11284] docs: Add design doc for Iceberg REST server-side scan planning results

### LINKED ISSUE #11284: [Improvement] Support `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` endpoint for server-side scan planning

laserninja:

Adds `design-docs/iceberg-rest-server-side-scan-planning.md`, the design for completing the Iceberg REST two-step scan planning protocol in the Gravitino Iceberg REST server, as requested in [#12194](https://github.com/apache/gravitino/pull/12194#issuecomment-5102634079).

The document covers:

- **Background**: `POST .../plan` is implemented, `POST .../tasks` is not, so clients that gate on the advertised endpoint set (pyiceberg with `scan-planning-mode=server`) cannot use server-side planning at all, and a plan of a large table has to be returned in one unbounded response.
- **Solution Investigations**: four token representations compared - always-inline (status quo), an in-memory plan-task store with random ids, a self-describing token (chosen), and asynchronous planning with server-side plan state - each with the concrete reason it was accepted or rejected, plus the cost of the chosen option stated in a table.
- **Proposal**:
  - the new `POST .../tasks` interface (request fields, response schema, behavior, and the full error table),
  - the changed `POST .../plan` interface (old vs. new response, per-client migration impact),
  - the `plan-task` token: payload fields, encoding, validation, forward compatibility, and why the token is not a capability,
  - determinism: why the planned snapshot is pinned and why planned tasks are ordered totally,
  - interaction with the existing scan plan cache,
  - the internal dispatcher interfaces added, events and audit operation type,
  - federated catalogs, delete files and merge-on-read tables, configuration, user process, and implementation flow.
- **Backward compatibility** and **Testing** summaries.
- **Open questions** the implementation deliberately leaves to review: whether batching should require the scan plan cache to be enabled, whether 100 is the right default batch size, and whether tokens should carry an expiry.
- **Task Breakdown**: what #12194 delivers, and the follow-ups that should be their own issues.

The implementation in #12194 introduces a new token type and changes what `POST .../plan` returns, and reviewers asked for the design to be written down and reviewed separately from the code.

Fix: #11284

No. Documentation only; the design document describes changes proposed in #12194.

No code change. The document was checked against the implementation in #12194: every interface, config default, error status and behavior it describes was verified against the code on that branch.


design-docs/iceberg-rest-server-side-scan-planning.md
```diff
@@ -0,0 +1,655 @@
...
+
+| Field   | Value                                                                                     |
+| ------- | ----------------------------------------------------------------------------------------- |
+| Status  | In review                                                                                 |
+| Authors | @laserninja                                                                               |
+| Created | 2026-07-28                                                                                |
+| Issue   | [#11284](https://github.com/apache/gravitino/issues/11284)                                |
+| PR      | [#12194](https://github.com/apache/gravitino/pull/12194)                                  |
+| Module  | `iceberg/iceberg-rest-server`, `iceberg/iceberg-common`, `catalogs/catalog-common`, `core` |
+
```
lasdf1234, line 26: # Design: Server-Side Scan Planning Results (`plan-tasks`) for the Gravitino Iceberg REST Server  

I think this part can be remove

lasdf1234:
@laserninja 
Thank you very much for your design document. Especially the open-ended questions are very profound. Your meticulous and striving spirit is truly admirable.

Regarding the three open-ended questions, I will provide the answers in the next session.

1. Should batching depend on the scan plan cache?
This is a very good question. I think for this design document, we can temporarily not address the changes related to caching. You can create a new issue to discuss whether caching should be enabled by default. (I prefer to enable caching. If an issue is raised, it can be discussed in that issue.)

2. Is 100 the right default?
In my opinion, the design of the interface should take into account the actual usage scenarios. For many scenarios, the default size of the gateway/LB is 1MB. 100 items would result in approximately 0.3MB of data being returned. So for now, I think 100 is a feasible option. If issues are encountered during actual scenario testing (where the default value of 100 causes many problems), I believe we can open another issue.

3. Should tokens carry an expiry?
In the source code and specifications of iceberg, there is no explanation for this thing called "token". Instead, it is more regarded as an opaque string. (The open API of iceberg: An opaque string provided by the REST server that represents a unit of work for generating file scan tasks for scan planning.) Therefore, I think this class should be named PlanTaskCodec instead of planTaskToken. This string is just an ordinary parameter (like tableName and catalogName, etc.), so I don't think there is a need to consider the issue of expiration.

I'm not sure if these answers are appropriate. I'm still willing to continue the discussion with you. If you agree with these responses, you can rewrite the design document based on this content. Don't forget to create a new issue: Should the cache be set as the default value. 

Thank you.

design-docs/iceberg-rest-server-side-scan-planning.md
laserninja: Removed. Two of the eight docs in `design-docs/` carry that table and six do not, so dropping it is the more common style; the issue, PR and module list now sit in one line under the title instead.

If you meant something else by "this part" - the title itself, or the whole header block including the line I kept - say so and I will adjust.

laserninja:
Thank you for the thorough answers, and for the pointer to how Iceberg itself frames a plan task. All three are folded into the document, which now records them as decisions taken in review rather than open questions (§8), and the implementation follows in #12194.

**1. Batching and the cache.** Agreed - caching is out of scope for this design. Batching always applies, and I filed https://github.com/apache/gravitino/issues/12254 to decide whether the scan plan cache should be enabled by default. I put the numbers, the memory-footprint question and the multi-replica caveat in that issue so the discussion has something concrete to work from.

**2. The default of 100.** Your reasoning is better than mine, and I have adopted it: gateways and load balancers commonly cap a body at about 1 MB, and 100 file scan tasks land around 0.3 MB, so 100 leaves comfortable headroom. That rationale is now in the document next to the config table (§5.12, §8.2) instead of the bare "matches Iceberg" note, and revisiting it is a follow-up gated on real workloads rather than an open question.

**3. Expiry and naming.** You are right that the specification treats it as an opaque string, not a token - "an opaque string provided by the REST server that represents a unit of work for generating file scan tasks for scan planning" - and that framing settles the expiry question: an ordinary request parameter has no lifetime of its own. No expiry, and a plan task now stays redeemable exactly as long as the snapshot it pins exists.

I also took the naming point further than the class name, since the token framing had leaked into comments, log messages, the user documentation and the test names. In #12194 (b3560127b): `PlanTaskToken` is now `PlanTaskCodec`, with `PlanTaskCodec.PlanTask` as the decoded unit of work, and "token" is gone from the Iceberg REST server, `docs/iceberg-rest-service.md` and this design document.

One thing I would like your read on: `PlanTaskCodec` is a static encode/decode pair, and the thing `decode` returns is the decoded unit of work, which I made a nested `PlanTaskCodec.PlanTask`. The alternative is two top-level classes, `PlanTaskCodec` plus a `PlanTask` value type. I went with the nested class to keep it to one file, but if you prefer the split, it is a small change.

lasdf1234:
@laserninja I don't have many questions left. I think others can also review this document.

lasdf1234:
@roryqi  @yuqi1129  Please review the document, thank you.

lasdf1234:
> Thank you for the thorough answers, and for the pointer to how Iceberg itself frames a plan task. All three are folded into the document, which now records them as decisions taken in review rather than open questions (§8), and the implementation follows in #12194.
> 
> **1. Batching and the cache.** Agreed - caching is out of scope for this design. Batching always applies, and I filed #12254 to decide whether the scan plan cache should be enabled by default. I put the numbers, the memory-footprint question and the multi-replica caveat in that issue so the discussion has something concrete to work from.
> 
> **2. The default of 100.** Your reasoning is better than mine, and I have adopted it: gateways and load balancers commonly cap a body at about 1 MB, and 100 file scan tasks land around 0.3 MB, so 100 leaves comfortable headroom. That rationale is now in the document next to the config table (§5.12, §8.2) instead of the bare "matches Iceberg" note, and revisiting it is a follow-up gated on real workloads rather than an open question.
> 
> **3. Expiry and naming.** You are right that the specification treats it as an opaque string, not a token - "an opaque string provided by the REST server that represents a unit of work for generating file scan tasks for scan planning" - and that framing settles the expiry question: an ordinary request parameter has no lifetime of its own. No expiry, and a plan task now stays redeemable exactly as long as the snapshot it pins exists.
> 
> I also took the naming point further than the class name, since the token framing had leaked into comments, log messages, the user documentation and the test names. In #12194 ([b356012](https://github.com/apache/gravitino/commit/b3560127b3bd9943cc038670b7afad6ffbb535f4)): `PlanTaskToken` is now `PlanTaskCodec`, with `PlanTaskCodec.PlanTask` as the decoded unit of work, and "token" is gone from the Iceberg REST server, `docs/iceberg-rest-service.md` and this design document.
> 
> One thing I would like your read on: `PlanTaskCodec` is a static encode/decode pair, and the thing `decode` returns is the decoded unit of work, which I made a nested `PlanTaskCodec.PlanTask`. The alternative is two top-level classes, `PlanTaskCodec` plus a `PlanTask` value type. I went with the nested class to keep it to one file, but if you prefer the split, it is a small change.

For the last question, I don't have many thoughts on it. I will make sure that more people pay attention to this issue during the code review.

roryqi:
Apache Polaris is developing the plan API. Could u compare with their solution?

roryqi:
Thanks for writing this down. The stateless/self-describing `plan-task` direction looks workable for multi-replica deployments because a task issued by one Gravitino instance can be decoded and re-planned by another instance without node-local state.

Could you add a short section that makes the multi-replica behavior and constraints explicit?

The main risk I see is not correctness of decoding across nodes, but load amplification. With the scan plan cache disabled by default, or with only node-local caches, `/plan` may run on replica A while concurrent `/tasks` requests land on replicas B/C/D. Each cache miss then re-plans the full pinned snapshot, reads manifests again, sorts the full task list, and only then slices one batch. For large tables this could multiply planning/object-store load across replicas. It would be useful to document the expected production posture here, for example whether operators should enable a cache, whether a shared cache is expected later, and whether any rate/concurrency limits are needed.

I also think the determinism requirement needs one more sentence of proof. The design depends on `(data file location, start, length)` being a true total order across independent re-plans on different nodes. If Iceberg can ever produce two semantically distinct file scan tasks with the same triple, different replicas could slice batches differently. Either the document should cite why that cannot happen, or the implementation should add another deterministic tie-breaker and a test for it.

One related point: if we later decide to make `plan-task` signed/HMACed to distinguish "issued by this server" from a forged but well-formed string, the signing key must be shared across all replicas. Without a shared secret, signing would reintroduce the same cross-replica failure mode that this design avoids.

lasdf1234:
> Thanks for writing this down. The stateless/self-describing `plan-task` direction looks workable for multi-replica deployments because a task issued by one Gravitino instance can be decoded and re-planned by another instance without node-local state.
> 
> Could you add a short section that makes the multi-replica behavior and constraints explicit?
> 
> The main risk I see is not correctness of decoding across nodes, but load amplification. With the scan plan cache disabled by default, or with only node-local caches, `/plan` may run on replica A while concurrent `/tasks` requests land on replicas B/C/D. Each cache miss then re-plans the full pinned snapshot, reads manifests again, sorts the full task list, and only then slices one batch. For large tables this could multiply planning/object-store load across replicas. It would be useful to document the expected production posture here, for example whether operators should enable a cache, whether a shared cache is expected later, and whether any rate/concurrency limits are needed.
> 
> I also think the determinism requirement needs one more sentence of proof. The design depends on `(data file location, start, length)` being a true total order across independent re-plans on different nodes. If Iceberg can ever produce two semantically distinct file scan tasks with the same triple, different replicas could slice batches differently. Either the document should cite why that cannot happen, or the implementation should add another deterministic tie-breaker and a test for it.
> 
> One related point: if we later decide to make `plan-task` signed/HMACed to distinguish "issued by this server" from a forged but well-formed string, the signing key must be shared across all replicas. Without a shared secret, signing would reintroduce the same cross-replica failure mode that this design avoids.

He opened another PR and enabled the cache.

roryqi:
> > Thanks for writing this down. The stateless/self-describing `plan-task` direction looks workable for multi-replica deployments because a task issued by one Gravitino instance can be decoded and re-planned by another instance without node-local state.
> > Could you add a short section that makes the multi-replica behavior and constraints explicit?
> > The main risk I see is not correctness of decoding across nodes, but load amplification. With the scan plan cache disabled by default, or with only node-local caches, `/plan` may run on replica A while concurrent `/tasks` requests land on replicas B/C/D. Each cache miss then re-plans the full pinned snapshot, reads manifests again, sorts the full task list, and only then slices one batch. For large tables this could multiply planning/object-store load across replicas. It would be useful to document the expected production posture here, for example whether operators should enable a cache, whether a shared cache is expected later, and whether any rate/concurrency limits are needed.
> > I also think the determinism requirement needs one more sentence of proof. The design depends on `(data file location, start, length)` being a true total order across independent re-plans on different nodes. If Iceberg can ever produce two semantically distinct file scan tasks with the same triple, different replicas could slice batches differently. Either the document should cite why that cannot happen, or the implementation should add another deterministic tie-breaker and a test for it.
> > One related point: if we later decide to make `plan-task` signed/HMACed to distinguish "issued by this server" from a forged but well-formed string, the signing key must be shared across all replicas. Without a shared secret, signing would reintroduce the same cross-replica failure mode that this design avoids.
> 
> He opened another PR and enabled the cache.

My main concern is multiple node deployment.

lasdf1234:
> > > Thanks for writing this down. The stateless/self-describing `plan-task` direction looks workable for multi-replica deployments because a task issued by one Gravitino instance can be decoded and re-planned by another instance without node-local state.
> > > Could you add a short section that makes the multi-replica behavior and constraints explicit?
> > > The main risk I see is not correctness of decoding across nodes, but load amplification. With the scan plan cache disabled by default, or with only node-local caches, `/plan` may run on replica A while concurrent `/tasks` requests land on replicas B/C/D. Each cache miss then re-plans the full pinned snapshot, reads manifests again, sorts the full task list, and only then slices one batch. For large tables this could multiply planning/object-store load across replicas. It would be useful to document the expected production posture here, for example whether operators should enable a cache, whether a shared cache is expected later, and whether any rate/concurrency limits are needed.
> > > I also think the determinism requirement needs one more sentence of proof. The design depends on `(data file location, start, length)` being a true total order across independent re-plans on different nodes. If Iceberg can ever produce two semantically distinct file scan tasks with the same triple, different replicas could slice batches differently. Either the document should cite why that cannot happen, or the implementation should add another deterministic tie-breaker and a test for it.
> > > One related point: if we later decide to make `plan-task` signed/HMACed to distinguish "issued by this server" from a forged but well-formed string, the signing key must be shared across all replicas. Without a shared secret, signing would reintroduce the same cross-replica failure mode that this design avoids.
> > 
> > 
> > He opened another PR and enabled the cache.
> 
> My main concern is multiple node deployment.

If it's a batch scenario involving multiple nodes, the issue is mainly about performance (in my opinion, this is inevitable). Do you think there is a problem with accuracy?

roryqi:
It is a performance issue instead of correctness issue.

lasdf1234:
> It is a performance issue instead of correctness issue.

Yes, I also agree with your opinion. 
Enabling caching by default is a solution.
@laserninja Can you think of any other ways to improve performance?

roryqi:
> > It is a performance issue instead of correctness issue.
> 
> Yes, I also agree with your opinion. Enabling caching by default is a solution. @laserninja Can you think of any other ways to improve performance?

Internal cache can't solve multiple node deployment performance issue.

lasdf1234:
> > > It is a performance issue instead of correctness issue.
> > 
> > 
> > Yes, I also agree with your opinion. Enabling caching by default is a solution. @laserninja Can you think of any other ways to improve performance?
> 
> Internal cache can't solve multiple node deployment performance issue.

It can alleviate the problem to some extent, but it cannot be completely cured. Nevertheless, enabling caching by default can still improve performance.

laserninja:
@roryqi thank you - the determinism point was not a theoretical one, and chasing it found a correctness bug. All three of your asks are addressed; the document is updated and #12194 has the fix.

**The total order was not total.** You asked me to either cite why two semantically distinct tasks cannot share `(data file location, start, length)`, or add a tie-breaker. It cannot be cited, because they can: appending one data file path twice leaves two manifest entries that tie on all three fields while carrying different sequence numbers, and potentially different delete files. Planning such a table shows it directly:

```
location=…/dup.parquet start=0 length=10  dataSeq=2  manifest=…-m0.avro pos=0
location=…/dup.parquet start=0 length=10  dataSeq=1  manifest=…-m0.avro pos=0
```

Sorting is stable, so tied tasks keep the order Iceberg planned them in - precisely the order that is not reproducible. Two of them either side of a batch boundary can swap between the plan and a later re-plan, and the client then receives one twice and never sees the other. That is a wrong result, not a slow one, so thank you for pushing on it.

Fixed in df9a2c065 by sorting on `(location, start, length, data sequence number, file sequence number, manifest location, entry position)`. A manifest entry is unique within a snapshot, so the order is now total for any two tasks that are not interchangeable; sequence numbers come first because they survive a rewrite into new manifests. The regression test builds a table holding the same path twice and asserts which entry each batch gets - it fails on the old comparator (`expected: <[2]> but was: <[1]>`) and passes on the new one.

**Multi-replica behaviour is now its own section (§5.15).** It separates the two halves of your point: correctness does not depend on which replica serves a request, and cost does. It states the worst case (`ceil(N/B)` full plans spread across replicas, since the cache is node-local and the Iceberg client fetches plan tasks concurrently), the production posture as a table (enable a cache, route a scan to one replica, raise the batch size, coalesce concurrent redemptions on a replica), and explicitly that those shrink the multiplier without removing it. Your signing observation is recorded there too: if we ever sign a plan task, the key must be shared by every replica, or verification succeeds only on the issuing replica and clients see `404`s that depend on routing.

**On removing the amplification rather than shrinking it (§8.5).** I agree with your reply to @lasdf1234 that an internal cache cannot cure the multi-node case. The way to cure it is to change what a plan task *is*: name a data manifest, or a manifest plus an entry range, instead of an offset into a global task list. Redemption then reads one manifest plus the delete manifests, so its cost is independent of plan size, on any replica, with no shared cache - and it makes the whole ordering question moot, because there is no global index to stabilise. I did not propose it for this PR because Iceberg does not expose enough to build it: `ManifestGroup` and `DeleteFileIndex`, which do residual evaluation and delete attachment, are package-private in `iceberg-core`, so it needs an upstream API or a careful port. It is written up as a follow-up.

**Polaris comparison (§8.4).** I looked, and there is nothing to compare against yet: Polaris excludes `.../plan`, `.../plan/{plan-id}` and `.../tasks` from the catalog spec it generates its API from, marked "Not implemented in Polaris", and omits them from its advertised `/v1/config` endpoints, so clients fall back to client-side planning. Their tracking issue, apache/polaris#966, has been open since February 2025 with no comments, through spec refreshes up to Iceberg 1.11. So they have neither problem you raised, for the reason that they do not plan server side.

The implementation that does exist is Iceberg's own `CatalogHandlers` with `RESTCatalogAdapter`, which keeps plans in a static `InMemoryPlanningState` and resolves a plan task by map lookup. That is the exact inverse trade-off: redemption is free, but a plan task is meaningless after a restart or on another process. I would rather pay re-planning cost than ship that, but the section states both sides so a reader can disagree with the choice.

lasdf1234:
# Manifest entry-range splitting (follow-up)

A `plan-task` is one batch of manifest entry ranges to read.
The opaque string is `base64url(JSON array)`: the root is the **array itself** — no `ranges` wrapper, and no embedded scan/snapshot.

## `planTaskKey` (decoded)

```json
[
  {
    "manifest": "s3://wh/db/t/metadata/snap-42-m1.avro",
    "entry-start": 40,
    "entry-end": 60
  },
  {
    "manifest": "s3://wh/db/t/metadata/snap-42-m2.avro",
    "entry-start": 0,
    "entry-end": 60
  },
  {
    "manifest": "s3://wh/db/t/metadata/snap-42-m3.avro",
    "entry-start": 0,
    "entry-end": 20
  }
]
```

| Field | Meaning |
| --- | --- |
| `manifest` | Data manifest path |
| `entry-start` / `entry-end` | Half-open range `[start, end)` within that manifest |
| Array length | `1` = single manifest; `≥ 2` = this batch spans multiple manifests |

## Setup (example)

```text
batch-size = 100

m0 → 60 entries
m1 → 60 entries
m2 → 60 entries
m3 → 70 entries
```

```text
Batch 0 (100):  m0[0,60) + m1[0,40)            → inlined on /plan
Batch 1 (100):  m1[40,60) + m2[0,60) + m3[0,20) → plan-task
Batch 2 (50):   m3[20,70)                        → plan-task
```

## First `POST .../plan` response

```json
{
  "plan-status": "completed",
  "file-scan-tasks": [
    { "data-file": { "file-path": "s3://wh/db/t/data/m0-e00.parquet" } },
    { "data-file": { "file-path": "s3://wh/db/t/data/m1-e39.parquet" } }
  ],
  "plan-tasks": [
    "W3sibWFuaWZlc3QiOiJzMzovL3doL2RiL3QvbWV0YWRhdGEvc25hcC00Mi1tMS5hdnJvIiwiZW50cnktc3RhcnQiOjQwLCJlbnRyeS1lbmQiOjYwfSx7Im1hbmlmZXN0IjoiczM6Ly93aC9kYi90L21ldGFkYXRhL3NuYXAtNDItbTIuYXZybyIsImVudHJ5LXN0YXJ0IjowLCJlbnRyeS1lbmQiOjYwfSx7Im1hbmlmZXN0IjoiczM6Ly93aC9kYi90L21ldGFkYXRhL3NuYXAtNDItbTMuYXZybyIsImVudHJ5LXN0YXJ0IjowLCJlbnRyeS1lbmQiOjIwfV0=",
    "W3sibWFuaWZlc3QiOiJzMzovL3doL2RiL3QvbWV0YWRhdGEvc25hcC00Mi1tMy5hdnJvIiwiZW50cnktc3RhcnQiOjIwLCJlbnRyeS1lbmQiOjcwfV0="
  ],
  "specs-by-id": {
    "0": { "spec-id": 0, "fields": [] }
  }
}
```

`file-scan-tasks`: Batch 0 — 100 tasks (only head/tail shown).

`plan-tasks[0]` decodes to Batch 1:

```json
[
  { "manifest": "s3://wh/db/t/metadata/snap-42-m1.avro", "entry-start": 40, "entry-end": 60 },
  { "manifest": "s3://wh/db/t/metadata/snap-42-m2.avro", "entry-start": 0,  "entry-end": 60 },
  { "manifest": "s3://wh/db/t/metadata/snap-42-m3.avro", "entry-start": 0,  "entry-end": 20 }
]
```

`plan-tasks[1]` decodes to Batch 2:

```json
[
  { "manifest": "s3://wh/db/t/metadata/snap-42-m3.avro", "entry-start": 20, "entry-end": 70 }
]
```

## `POST .../tasks`

### Request

```json
{
  "plan-task": "W3sibWFuaWZlc3QiOiJzMzovL3doL2RiL3QvbWV0YWRhdGEvc25hcC00Mi1tMS5hdnJvIiwiZW50cnktc3RhcnQiOjQwLCJlbnRyeS1lbmQiOjYwfSx7Im1hbmlmZXN0IjoiczM6Ly93aC9kYi90L21ldGFkYXRhL3NuYXAtNDItbTIuYXZybyIsImVudHJ5LXN0YXJ0IjowLCJlbnRyeS1lbmQiOjYwfSx7Im1hbmlmZXN0IjoiczM6Ly93aC9kYi90L21ldGFkYXRhL3NuYXAtNDItbTMuYXZybyIsImVudHJ5LXN0YXJ0IjowLCJlbnRyeS1lbmQiOjIwfV0="
}
```

Server: decode the array → read each slice in order → return that batch’s `file-scan-tasks` (no full-table re-plan).

### Response

```json
{
  "file-scan-tasks": [
    { "data-file": { "file-path": "s3://wh/db/t/data/m1-e40.parquet" } },
    { "data-file": { "file-path": "s3://wh/db/t/data/m2-e00.parquet" } },
    { "data-file": { "file-path": "s3://wh/db/t/data/m3-e00.parquet" } }
  ],
  "plan-tasks": [],
  "specs-by-id": {
    "0": { "spec-id": 0, "fields": [] }
  }
}
```

## Compared with the current PR design

| | PR (offset) | This approach |
| --- | --- | --- |
| Decoded `plan-task` | `{ table, offset, limit, scan }` | JSON **array** `[{ manifest, entry-start, entry-end }, …]` |
| `/tasks` | May fully re-plan, then slice | Read only those manifest slices |
| Multi-manifest batch | Crossed via global indexes | Multiple elements in the array |


@roryqi @laserninja In my opinion, this approach is quite good. Because when the /plan interface is called for the first time, a complete splitting plan can be returned. For subsequent calls to the tasks interface, there is no need to recalculate what files should be read. What do you think?

laserninja:
@lasdf1234 I like this, and I think it is the right end state - it is the direction §8.5 sketches, and your framing is sharper than mine: the `/plan` call has already computed the whole split, so redeeming a batch should never recompute it. It also removes the ordering problem entirely, because manifests are immutable and a position inside one is fixed, so §5.6 stops being load-bearing.

Before we commit to the payload, four things I ran into when checking it against the Iceberg 1.11 APIs. Three of them come from the same decision - "no embedded scan/snapshot" - and I think that decision has to be reversed.

**1. Without the snapshot, delete files cannot be attached.** Delete manifests are reachable only through a snapshot (`Snapshot.deleteManifests(io)`); a manifest path on its own does not say which snapshot referenced it. The same applies to sequence numbers: entries inherit them from the `ManifestFile` in the snapshot's manifest list, and `ManifestFiles.read` needs that `ManifestFile`, not a path. So a merge-on-read table would return tasks with no deletes attached, and the client would read rows that have been deleted. Silently wrong results are the failure mode I would most want to avoid here.

**2. Without the scan, the ranges do not reproduce and residuals cannot be computed.** `ManifestReader.iterator()` yields entries *after* `filterRows`/`filterPartitions`/`caseSensitive` are applied, so "entry 40" means one thing with the client's filter and another without it. And `residual-filter`, which the engine applies on top of each data file, is derived from the scan filter (`ResidualEvaluator.of(spec, filter, caseSensitive)`); with no filter we would have to emit always-true residuals, which again changes query results rather than just performance.

**3. Without the table, `/tasks` would read whatever path it is handed.** Authorization is evaluated from the table in the request path, but the manifest paths come from the request body. A caller authorized on table A could send a plan task naming a manifest of table B and get back B's data file paths and column statistics. We would need to validate every manifest against the pinned snapshot's manifest list - which needs the snapshot again.

**4. Payload size moves from bounded to unbounded.** Your example already spans three manifests in one batch. Tables written by frequent small commits have many small manifests, so a 100-task batch can name dozens, at roughly 100 bytes of S3 path each. The plan tasks then grow to kilobytes each, and `/plan` carries all of them, which erodes the bounded-response property batching exists for. Today a plan task is a few hundred bytes whatever the plan size. Capping the manifests per plan task fixes it, at the price of uneven batches.

So the payload I would suggest is yours plus the three fields it dropped:

```json
{
  "table": "db.t",
  "snapshot-id": 42,
  "scan": { "filter": …, "case-sensitive": true, "select": [], "stats-fields": [] },
  "ranges": [
    { "manifest": "s3://wh/db/t/metadata/snap-42-m1.avro", "entry-start": 40, "entry-end": 60 }
  ]
}
```

That keeps validation, residuals and delete attachment correct, and still gets the property we are after: redemption cost is proportional to the batch, not to the plan, on any replica, with no shared cache.

**The piece that needs care is delete attachment.** `ManifestFiles.read`, `ManifestReader.filterRows/select/caseSensitive` and `ResidualEvaluator` are all public, so the data side is buildable. `DeleteFileIndex` is not - it is package-private in `iceberg-core` - so matching deletes to data files by partition and sequence number, including equality deletes and v3 deletion vectors, would have to be reimplemented. That is the part where a subtle mistake returns deleted rows.

A way to get most of the win without that risk: use the manifest-scoped path only when the pinned snapshot has no delete manifests, and fall back to the current re-plan when it does. Copy-on-write tables - the common case, and the ones with the largest plans - get redemption proportional to the batch immediately, and merge-on-read keeps today's correctness until either Iceberg exposes delete indexing or we port it deliberately with tests.

On sequencing, one point that I think takes the pressure off: a plan task is opaque and nothing persists it, so its encoding can change in any later release without breaking a client or needing a migration. Changing it is not an API change. So this does not have to block #12194 - and @nevzheng made the same point about not blocking it on the cache discussion in #12254.

I will write this up in §8.5 either way, with your payload as the starting point, so the follow-up has a design to implement rather than a paragraph. Two questions for you and @roryqi:

1. Do you want the manifest-scoped payload in #12194 now, or as the follow-up it is written up as? I lean follow-up, because of the delete-index work, but I am happy either way and will do it now if you would rather not merge the offset form at all.
2. For the follow-up, is the copy-on-write-only first cut acceptable, or would you rather wait and do both paths at once?

Two small corrections for whoever implements it: the plan response field is `status`, not `plan-status`, and a redeemed plan task should omit `plan-tasks` rather than send `[]`.

nevzheng:
Hi all. Thanks @laserninja, @lasdf1234, and @roryqi. I have reviewed the design and discussion and would like to share my review, verdict, and proposed next steps for this effort.

**Overall:** LGTM on Phase-1 design direction. **Defer (blocking follow-up):** Phase-2 ticket/design (ranges §8.5 + caching from [#12254](https://github.com/apache/gravitino/issues/12254)).

### Next steps

**Phase 1 design: LGTM** (see understanding + scoping below).

**P0 — Deliver Phase 1**
1. **@laserninja** — proceed with Phase-1 implementation (#12194).

**P1 — Close this design PR / park Phase 2**
2. **@lasdf1234 @roryqi** — close out open threads on this PR; migrate any remaining topics or open discussions into the Phase-2 ticket below. If we still need @yuqi1129, please track him down and ask — fine if he’s busy; we just need this thread closed.
3. **@laserninja** — open a Phase-2 follow-up ticket / design (ranges §8.5, caching starting from [#12254](https://github.com/apache/gravitino/issues/12254), related cost work). **Defer (blocking follow-up)** into that track — not a gate on this design PR. I’m happy to help review and give input.

**P2 — Clarity (non-blocking)**
4. **@laserninja** — **suggestion (strong, non-blocking):** add sequence diagrams for the `/plan` → `plan-task` → `/tasks` flow, and explicitly list which CUJs Phase 1 vs Phase 2 cover.

**P3 — Impl review**
5. **@lasdf1234** — smaller impl-detail questions (e.g. `PlanTaskCodec` nesting) → hash out in #12194 implementation review, not on this design doc.

## Scope — Phase 1

### IRC spec — scan planning compliance

**Understanding**
- Complete Iceberg REST Catalog scan planning: `POST .../plan` + `POST .../tasks`, opaque `plan-task`, advertise `/tasks` in `/v1/config`.
- We’re following the IRC spec for server-side planning, so there aren’t many API-shape decisions to make here.

**Verdict:** LGTM
- **note:** Wire protocol is Iceberg’s; our decisions are mostly behind the opaque `plan-task`, not the API shape.

### State model

**Understanding**
- Stateless: compute on every `/plan` and `/tasks` call.
- Continuation lives in the opaque `plan-task` (no plan DB / shared plan-id session).
- Caching is optional for *correctness* (not the source of truth).

**Verdict:** LGTM · **Defer (blocking follow-up)** — caching (#12254, Phase-2 track)
- **issue (non-blocking):** Stateless recompute without a real caching design is an operational hazard (staleness, multi-replica, invalidation, load amplification). Does **not** block merging this design doc.
- **suggestion (non-blocking):** @laserninja @lasdf1234 @roryqi — **Defer (blocking follow-up)** into the Phase-2 track: own a caching design/process from [#12254](https://github.com/apache/gravitino/issues/12254). Do not treat “turn the cache on” as a small toggle. I’m happy to help review.

### Metadata / continuation model (`plan-task`)

**Understanding**
- Opaque `plan-task` = replay params (table + pinned scan/snapshot + offset/limit).
- No stored task list; `/tasks` recomputes and slices.
- Encoding can change later without an IRC API break.

**Verdict:** LGTM w/ non-blocking comments · **Defer (blocking follow-up)** — caching (#12254, Phase-2 track)
- **note:** This interaction model is fairly opaque; sequence diagrams help a lot. I trust the owners to implement it.
- **issue (non-blocking):** Recompute-heavy redeem + weak caching ⇒ perf / cost / operational risk (self-DoS-shaped load under concurrent `/tasks`). Does **not** block merging this design doc.
- **suggestion (non-blocking):** Mitigate in layers: (1) **Defer (blocking follow-up)** caching/#12254 on the Phase-2 track (@laserninja @lasdf1234 @roryqi); (2) **Defer** Phase 2 §8.5 manifest-range plan-tasks (recompute less); (3) **Defer** later shared cache / coalesce as needed.

### Correctness bar

**Understanding**
Under Phase-1 recompute, paging is only safe if all of the following hold
(design §5.2, §5.5.4, §5.6, §5.11; pin + total order also stressed in Rory’s thread):

1. **Snapshot pinning** — `plan-task` embeds the planned snapshot so later commits don’t move the scan.
2. **Total order** — tasks are totally ordered before offset/limit so independent recomputes slice the same way (incl. dup-path fix).
3. **Authz on POST** — `POST .../plan` and `POST .../tasks` authorize from the **URL table**; plan-task is **not a capability**.
4. **MoR deletes (Phase 1)** — full planner + per-response `delete-files` for that page’s tasks (no cross-page delete indexes).

**Verdict:** LGTM
- **praise:** Snapshot pinning and total order — the right correctness foundation for stateless redeem; good catch on the sort key in review.
- **praise:** Path-based authz + not-a-capability is solid.
- **note:** Phase-1 MoR path looks correct; the harder delete-index problem belongs to Phase 2 ranges (§8.5), not this bar.

### Config

**Understanding**
- Principle: **good defaults + configurability**, with a clear mechanism to change behavior.
- Phase-1 paging knob: `gravitino.iceberg-rest.scan-plan-task-batch-size`
  - **default `100`** — page size for file-scan-tasks (~0.3MB vs common ~1MB LB caps; Iceberg-aligned)
  - **`0`** — escape hatch back to all-inline `/plan`
  - other values — operators can tune page size
- Batching is always applied (not gated on cache). Cache on/off is separate (#12254).
- Mechanism: standard Gravitino Iceberg REST config property (documented; not a hidden constant).

**Verdict:** LGTM w/ follow-up requested
- **note:** Confirming **good defaults + configurability**: default `100` looks right; `0` and tunable values are the right escape/tune mechanism via `scan-plan-task-batch-size`.
- **suggestion (non-blocking):** Follow-up — revisit the default under real workloads if 100 causes too many `/tasks` RPCs (§8.2); keep the property as the single knob (don’t add parallel undocumented ways to set page size).

Best,  
Nevin

PS — I will write up thoughts on Phase 2.

nevzheng:
### Phase 2 — initial thoughts

@laserninja — I recommend landing the proposed Phase 1 scope. Then we can address the proposed Phase 2 scope below.

**P0 — Caching (blocking):** @laserninja @lasdf1234 @roryqi — we need a strong caching story here: a solid design doc / plan (starting from [#12254](https://github.com/apache/gravitino/issues/12254)), not just enabling the existing cache. Cover at least cache keys, invalidation / staleness, multi-replica behavior, and failure modes. **Shared cache across replicas** is something we should try to address — a detail under this general theme.

**Metadata / `plan-task` structure:**
1. I agree with the direction / evolution of the metadata (§8.5) — it’s solid. Document and discuss in Phase 2; don’t forget diagrams for the flow / delta. Fields under discussion: keep `table` / `snapshot-id` / `scan`; add `ranges[]` (`manifest`, `entry-start`, `entry-end`) instead of Phase-1 `offset` / `limit`.
2. Open question: is this **(a)** a versioned encoding / compatibility concern, or **(b)** just an implementation milestone (opaque string; encoding can change without an IRC API break)?

**Load amplification:** Discussion / doc (§4, §5.15) revealed key pain points to resolve in Phase 2:
1. **Recompute-per-redeem** — each `/tasks` may re-plan the full snapshot; cost scales with plan size, not batch
2. **Multi-replica concurrent fetch** — Iceberg client fetches plan-tasks in parallel; requests land on different replicas by design
3. **Node-local / cold cache** — today’s cache is per-replica and often off by default, so (1)+(2) become the common production posture

Mitigations already on the table: **caching** (P0), **metadata / ranges**, **coalesce** concurrent redemptions on one replica, plus other ops levers. The fact that we’re talking about this many mitigations is itself a signal — we should discuss and design for load amplification thoroughly in Phase 2.

**MoR delete attachment under ranges:**
1. **Problem** — once `/tasks` only reads named manifest slices, delete attachment is no longer “free” from the planner. `DeleteFileIndex` is package-private; a subtle mistake returns deleted rows (§8.5.4)
2. **We should address this** — don’t leave MoR correctness implicit when ranges land
3. **Sizing** — may be a **Phase 3** item depending on sizing; discuss further (e.g. CoW-first + MoR re-plan fallback vs full MoR in Phase 2)

**Production tuning:** If we follow **good defaults + configurability**, this should stay easy — e.g. `scan-plan-task-batch-size` (default 100). Let’s make sure Phase 1 and Phase 2 keep this in mind: revisit under real load; don’t bury tunables as hidden constants.

**Other notes (P2/P3):**
1. **Async `plan/{id}`:** Settled for now (defer; sync Option C / [#11635](https://github.com/apache/gravitino/issues/11635)). I think this could be a strong basis for a **stateful** design where Gravitino maintains plan state. No further comment at the moment — notes only; document for later.
2. **LB affinity:** Finer-grained routing detail (stick one scan to one replica) for P2/P3 — not its own workstream.

### Next steps

@laserninja @lasdf1234 @roryqi — please open a Phase-2 follow-up ticket and migrate the concerns / discussion above into it. Don’t overthink Phase 2 right now — **focus on landing Phase 1**.

Best,  
Nevin

nevzheng:
@laserninja please let us know if a Google Meet sync would be useful so we can align and drive this project effort forward. Overall I think we are aligned.

@roryqi @lasdf1234 if that sounds useful please coordinate with @laserninja on timing and send a calendar invite with a Google Meet link. Feel free to add me.

Thanks,
Nevin

nevzheng (approved):
LGTM. see my other comments in thread.

lasdf1234:
@laserninja Perhaps we could hold a Google Meeting. As for the design document, I think it's fine too. LGTM.

lasdf1234 (approved):
LGTM.

laserninja:
Thanks @nevzheng, @lasdf1234, @roryqi for the reviews and the approvals. Taking the assigned items in order.

**Doc updates before this merges (P2 #4).** I will push to this PR:

- a sequence diagram for `/plan` → `plan-task` → `/tasks`, covering the federated path and the terminal batch,
- an explicit CUJ table saying which journeys Phase 1 serves and which are Phase 2, and
- a rewrite of §9 Task Breakdown, which is now stale. Per the split agreed in #12194, Phase 1 ships as four PRs: #12411 (deterministic task order), #12412 (the `/tasks` endpoint and its plumbing, unadvertised), federation forwarding (to open), then #12194 (batching, `PlanTaskCodec`, `scan-plan-task-batch-size`, the `/v1/config` advertisement, user docs). §9 currently reads as if all of it is one PR.

**Phase-2 ticket (P1 #3).** I will open it with the discussion here migrated into it rather than restated, sectioned as: manifest entry-range plan tasks (§8.5); the caching design starting from #12254 - keys, invalidation and staleness, multi-replica behavior, failure modes, sizing; load amplification - recompute-per-redeem, concurrent redemptions landing on different replicas, cold node-local cache - together with the mitigation layers (cache, ranges, coalescing on a replica); and MoR delete attachment once `/tasks` reads named manifest slices (§8.5.4), which may fall to Phase 3 depending on sizing. #11583 and LB affinity go in as notes, not as work.

**Your Phase-2 open question, (a) versioned encoding or (b) implementation milestone: (b).** A plan task is opaque, nothing persists it, and none outlives the scan that issued it, so replacing offset/limit with ranges is not an IRC API change and needs neither a migration nor a version field (§5.5.3, §8.5.5). The one case that is not free is a rolling deploy, where a plan task minted by an old replica is redeemed by a new one. Phase 1 already gives that the right shape: a payload the decoder does not understand becomes a `404`, not a guess, so the client replans rather than receiving a wrong slice. The Phase-2 doc should state that ranges keep that property instead of leaving it implied.

**The cache default (#12254 / #12320).** @lasdf1234 asked on the issue for a PR that flips the default, and #12320 is that and nothing more: `scan-plan-cache-impl` defaults to `LocalScanPlanCache`, an empty string still disables it, capacity 200 and expiry 60 unchanged. @nevzheng, that is deliberately the toggle and not the caching story you asked for - it settles none of keys, invalidation, multi-replica or sizing, all of which stay in the Phase-2 ticket. If you would rather it wait behind that design, say so and I will hold it. The argument for landing it now is that batched redemption replans on every miss, so the toggle is worth more with Phase 1 in than without it.

**Meeting.** Yes, useful, mainly to settle where the Phase-1/Phase-2 line sits. I am on US Pacific; the overlap with Beijing that is not unkind to anyone is 17:00-18:00 PT / 09:00-10:00 CST the next day. @lasdf1234 @roryqi any day this week works - send an invite and add @nevzheng.

PR #12256, opened 2026-07-29
[#12255] fix(lance): authenticate the Lance REST service to the Gravitino server

### LINKED ISSUE #12255: [Bug report] Lance REST service cannot operate against a Gravitino server with authentication enabled
### Version

main branch

## Describe what's wrong

The Lance REST service constructs its `GravitinoClient` with no `AuthDataProvider`, so the requests it makes to the Gravitino server carry no `Authorization` header.

`GravitinoLanceNamespaceWrapper.initialize()`:

```java
this.client =
    GravitinoClient.builder(uri)
        .withMetalake(metalakeName)
        .withClientConfig(clientProperties)
        .build();
```

There is no configuration property to change this. `LanceConfig` defines only `namespace-backend`, `gravitino-metalake`, and `gravitino-uri`.

Two consequences follow, depending on how `gravitino.authenticators` is set on the Gravitino server.

**With `simple`.** `SimpleAuthenticator.supportsToken(null)` returns `true` and `authenticateToken(null)` returns `ANONYMOUS_PRINCIPAL`, so the requests are accepted. Every Lance operation executes in Gravitino as the anonymous user rather than as any identity attributable to the service.

**With `oauth2` or `kerberos`.** `OAuth2TokenAuthenticator.supportsToken` requires a Bearer prefix and returns `false` for a null token. No configured authenticator accepts the request, so `AuthenticationFilter` throws and every Lance operation fails.

Because `gravitino.authenticators` is a single server-wide list, there is no per-service override. The only configuration in which the Lance REST service works alongside OAuth2 is one that also lists `simple`, and in that configuration any caller of the Gravitino server who omits the `Authorization` header is likewise accepted as anonymous. The authentication documentation already identifies that pattern as unsafe for Basic auth, where it states that `gravitino.authenticators` must include `basic` and must not include `simple`.

| `gravitino.authenticators` | Lance REST service | Other callers of the Gravitino server |
|----------------------------|---------------------------------|----------------------------------------|
| `simple`                   | Works, as anonymous             | Authenticated, or anonymous if no header |
| `oauth2`                   | Every operation fails           | Authenticated                          |
| `oauth2,simple`            | Works, as anonymous             | Anonymous accepted for any caller omitting the header |

The net effect is that the Lance REST service is unusable in a deployment with authentication enforced, and usable only in a deployment that accepts anonymous requests.

The Iceberg REST service does not have this problem. It exposes `gravitino.iceberg-rest.gravitino-auth-type` with `simple` and `oauth2` values and a service user name defaulting to `iceberg-rest-server`, and as an auxiliary service it uses internal interfaces rather than HTTP, so it does not make this call at all.

With `gravitino.authenticators = oauth`, any Lance REST operation returns a 500 and the server logs:

```
org.apache.gravitino.exceptions.UnauthorizedException: The provided credentials did not support
	at org.apache.gravitino.server.authentication.AuthenticationFilter.doFilter(AuthenticationFilter.java:...)
```

The relevant code paths:

- `GravitinoLanceNamespaceWrapper.initialize()` builds the client with `authDataProvider` unset.
- `AuthenticationFilter.doFilter` iterates the configured authenticators and throws `UnauthorizedException("The provided credentials did not support")` when none returns a principal.
- `SimpleAuthenticator.supportsToken` returns `true` when `tokenData == null`; `authenticateToken(null)` returns `ANONYMOUS_PRINCIPAL`.
- `OAuth2TokenAuthenticator.supportsToken` returns `false` when `tokenData == null`.

### How to reproduce

Gravitino version: main branch.

1. Configure the Gravitino server for OAuth2 per [How to authenticate](https://gravitino.apache.org/docs/latest/security/how-to-authenticate), so that `gravitino.authenticators = oauth` with no `simple` entry.
2. Enable the Lance REST service as an auxiliary service:

```properties
gravitino.auxService.names = lance-rest
gravitino.lance-rest.classpath = lance-rest-server/libs
gravitino.lance-rest.namespace-backend = gravitino
gravitino.lance-rest.gravitino-uri = http://localhost:8090
gravitino.lance-rest.gravitino-metalake = test_metalake
```

3. Start the server and create a metalake named `test_metalake`.
4. Call any Lance REST operation with a valid Bearer token, so that authentication at the Lance REST service itself succeeds:

```shell
curl -X POST http://localhost:9101/lance/v1/namespace/lance_catalog/create \
  -H "Authorization: Bearer ${VALID_TOKEN}" \
  -d '{"id": ["lance_catalog"], "mode": "create"}'
```

The call fails. The service authenticated the caller and then made its own unauthenticated request to the Gravitino server, which rejected it.

5. Add `simple` to `gravitino.authenticators`, restart, and repeat. The call now succeeds, and a request to the Gravitino server with no `Authorization` header at all also succeeds as anonymous.

### Additional context

Two related gaps are out of scope for this issue and worth tracking separately.

**Caller identity is not forwarded.** Even once the service authenticates to Gravitino under its own identity, it does not act on behalf of the caller, so Gravitino privileges cannot be evaluated per caller on the Lance path. The Iceberg REST service supports access control in auxiliary mode by sending the authenticated user identity to the Gravitino server for authorization. Achieving the same for Lance is a larger change of the same shape as the user forwarding added to the Trino connector.

**Credential vending is not wired into the Lance path.** Storage access is configured with static `lance.storage.*` properties and the resolved values, including access keys and secrets, are returned to clients in the `storageOptions` of `CreateTable` and `DescribeTable` responses. Neither the Lance REST service nor the Generic Lakehouse catalog requests vended credentials, so short-lived scoped credentials are not available on either Lance path.

A fix for the issue described above is prepared and adds `gravitino.lance-rest.gravitino-auth-type` with `simple` and `oauth2` support, mirroring the Iceberg REST service properties and defaulting to `simple` with the user name `lance-rest-server` so that existing deployments are unaffected.

markhoerth:

Adds `gravitino.lance-rest.gravitino-auth-type` with `simple` and `oauth2`
support, plus the supporting `gravitino-simple.user-name` and
`gravitino-oauth2.*` properties, and uses them when constructing the
`GravitinoClient` in `GravitinoLanceNamespaceWrapper`. The property names and
the dispatch logic mirror `DynamicIcebergConfigProvider.createGravitinoClient`.

The builder configuration is split into `newClientBuilder` so it can be tested
without opening a connection. Runtime behavior of `initialize()` is unchanged
apart from the credentials now being attached.

The Lance REST service built its client with no `AuthDataProvider`, so its
requests to the Gravitino server carried no `Authorization` header. Under the
`simple` authenticator those requests are accepted as the anonymous user.
Under `oauth2` or `kerberos` no authenticator accepts a null token, so every
Lance operation fails. Since `gravitino.authenticators` is server-wide, the
only configuration where the service works alongside OAuth2 also lists
`simple`, which leaves an anonymous bypass open for all callers of the server.

Fix: #12255

Yes, six new configuration properties, documented in
`docs/lance-rest-service.md`. The default is `simple` with user name
`lance-rest-server`, so existing deployments continue to work and gain a named
identity in place of anonymous.

Documentation here is limited to a config table for the six new properties. A
broader rewrite of `lance-rest-service.md` will follow in a separate PR.

Unit tests in TestGravitinoLanceClientAuth cover the default auth type, a configured simple user name, case-insensitive parsing, both missing-required-OAuth2-property cases, and rejection of an unsupported auth type.

LanceRESTServiceAuthIT covers the behavior end to end: with a configured simple user name, a catalog created through the Lance REST service records that user as its creator. Against the current code the test fails with:

```
expected: <lance_rest_service_user> but was: <anonymous>
```

The OAuth2 path is covered at the configuration level only, since end-to-end coverage would need a token issuer in the test harness.

### Target branch

The config entries are marked `1.3.0` because this fixes behavior in the 1.3
line: the Lance REST service cannot operate against a server with
authentication enabled. Requesting the branch-1.3 cherry-pick label.

### DIFF
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -141,14 +141,14 @@ credentials, otherwise a Gravitino server configured with an authenticator other
 rejects them and every Lance operation fails. Configure the auth type to match the Gravitino
 server:
 
-| Configuration Property                            | Description                                                                       | Default Value      | Required          | Since Version |
-|---------------------------------------------------|-----------------------------------------------------------------------------------|--------------------|-------------------|---------------|
-| `gravitino.lance-rest.gravitino-auth-type`        | Auth type used to reach the Gravitino server. Supported values: `simple`, `oauth2` | `simple`           | No                | 1.3.0         |
-| `gravitino.lance-rest.gravitino-simple.user-name` | User name presented when the auth type is `simple`                                | `lance-rest-server` | No                | 1.3.0         |
-| `gravitino.lance-rest.gravitino-oauth2.server-uri`| OAuth2 server URI                                                                 | (none)             | Yes, for `oauth2` | 1.3.0         |
-| `gravitino.lance-rest.gravitino-oauth2.credential`| Credential used to request the OAuth2 token                                       | (none)             | Yes, for `oauth2` | 1.3.0         |
-| `gravitino.lance-rest.gravitino-oauth2.token-path`| Path on the OAuth2 server used to request the token                               | (none)             | Yes, for `oauth2` | 1.3.0         |
-| `gravitino.lance-rest.gravitino-oauth2.scope`     | Scope of the requested OAuth2 token                                               | (none)             | Yes, for `oauth2` | 1.3.0         |
+| Configuration Property                             | Description                                                                        | Default Value       | Required          | Since Version |
+|----------------------------------------------------|------------------------------------------------------------------------------------|---------------------|-------------------|---------------|
+| `gravitino.lance-rest.gravitino-auth-type`         | Auth type used to reach the Gravitino server. Supported values: `simple`, `oauth2` | `simple`            | No                | 2.0.0         |
+| `gravitino.lance-rest.gravitino-simple.user-name`  | User name presented when the auth type is `simple`                                 | `lance-rest-server` | No                | 2.0.0         |
+| `gravitino.lance-rest.gravitino-oauth2.server-uri` | OAuth2 server URI                                                                  | (none)              | Yes, for `oauth2` | 2.0.0         |
+| `gravitino.lance-rest.gravitino-oauth2.credential` | Credential used to request the OAuth2 token                                        | (none)              | Yes, for `oauth2` | 2.0.0         |
+| `gravitino.lance-rest.gravitino-oauth2.token-path` | Path on the OAuth2 server used to request the token                                | (none)              | Yes, for `oauth2` | 2.0.0         |
+| `gravitino.lance-rest.gravitino-oauth2.scope`      | Scope of the requested OAuth2 token                                                | (none)              | Yes, for `oauth2` | 2.0.0         |
 
 This setting controls how the service authenticates to the Gravitino server. It does not change how
 callers authenticate to the Lance REST service itself.
--- a/docs/lance-rest-service.md
+++ b/docs/lance-rest-service.md
@@ -134,6 +134,25 @@ To enable the Lance REST service within Gravitino server, configure the followin
 | `gravitino.lance-rest.gravitino-uri`      | Gravitino server URI (required when namespace-backend is `gravitino`)        | http://localhost:8090   | Yes      | 1.1.0         |
 | `gravitino.lance-rest.gravitino-metalake` | Gravitino metalake name (required when namespace-backend is `gravitino`)     | (none)                  | Yes      | 1.1.0         |
 
+**Authentication to the Gravitino Server**
+
+The Lance REST service makes its own requests to the Gravitino server. Those requests must carry
+credentials, otherwise a Gravitino server configured with an authenticator other than `simple`
+rejects them and every Lance operation fails. Configure the auth type to match the Gravitino
+server:
+
+| Configuration Property                            | Description                                                                       | Default Value      | Required          | Since Version |
+|---------------------------------------------------|-----------------------------------------------------------------------------------|--------------------|-------------------|---------------|
+| `gravitino.lance-rest.gravitino-auth-type`        | Auth type used to reach the Gravitino server. Supported values: `simple`, `oauth2` | `simple`           | No                | 1.3.0         |
+| `gravitino.lance-rest.gravitino-simple.user-name` | User name presented when the auth type is `simple`                                | `lance-rest-server` | No                | 1.3.0         |
+| `gravitino.lance-rest.gravitino-oauth2.server-uri`| OAuth2 server URI                                                                 | (none)             | Yes, for `oauth2` | 1.3.0         |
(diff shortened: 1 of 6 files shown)


jerryshao:
CC @yuqi1129 please help to push this forward.

docs/lance-rest-service.md
```diff
@@ -134,6 +134,25 @@ To enable the Lance REST service within Gravitino server, configure the followin
...
+**Authentication to the Gravitino Server**
+
+The Lance REST service makes its own requests to the Gravitino server. Those requests must carry
+credentials, otherwise a Gravitino server configured with an authenticator other than `simple`
+rejects them and every Lance operation fails. Configure the auth type to match the Gravitino
+server:
+
+| Configuration Property                            | Description                                                                       | Default Value      | Required          | Since Version |
+|---------------------------------------------------|-----------------------------------------------------------------------------------|--------------------|-------------------|---------------|
+| `gravitino.lance-rest.gravitino-auth-type`        | Auth type used to reach the Gravitino server. Supported values: `simple`, `oauth2` | `simple`           | No                | 1.3.0         |
```
yuqi1129: Format the code.

docs/lance-rest-service.md
```diff
@@ -134,6 +134,25 @@ To enable the Lance REST service within Gravitino server, configure the followin
...
+| `gravitino.lance-rest.gravitino-simple.user-name` | User name presented when the auth type is `simple`                                | `lance-rest-server` | No                | 1.3.0         |
+| `gravitino.lance-rest.gravitino-oauth2.server-uri`| OAuth2 server URI                                                                 | (none)             | Yes, for `oauth2` | 1.3.0         |
+| `gravitino.lance-rest.gravitino-oauth2.credential`| Credential used to request the OAuth2 token                                       | (none)             | Yes, for `oauth2` | 1.3.0         |
+| `gravitino.lance-rest.gravitino-oauth2.token-path`| Path on the OAuth2 server used to request the token                               | (none)             | Yes, for `oauth2` | 1.3.0         |
+| `gravitino.lance-rest.gravitino-oauth2.scope`     | Scope of the requested OAuth2 token                                               | (none)             | Yes, for `oauth2` | 1.3.0         |
```
yuqi1129: The version should be `2.0.0` not `1.3.0`

yuqi1129 (commented):
Some minor issues.

yuqi1129:
@markhoerth 
If you do not have free time to push this one forward, I would like help to work on it.

markhoerth:
That's great Qi, I see you merged one change.  Anything else?

yuqi1129:
> That's great Qi, I see you merged one change. Anything else?

I could not see any problems, and will merge it today.

yuqi1129 (approved):

Issue #12220, opened 2026-07-27
[EPIC] Add durable deletion-generation storage


nevzheng:
### Describe the proposal

This is the relational-storage child epic of #12219. It owns only the metadata representation and transactions required by soft deletion.

## Data model

- The existing `table_meta.table_id` remains the stable table identity.
- `table_meta.deleted_at` records when the row was deleted.
- Nullable `table_meta.deletion_id` is a logical pointer to the current active deletion action.
- `entity_deletion` stores only the opaque `deletion_id`, lifecycle `state`, fixed `retention_expires_at`, and optional future `purge_job_id`.
- Identity, parent, name, ownership, and deletion time remain authoritative on `table_meta`; they are not duplicated into the action.
- Table-owned child metadata remains attached through `table_id` and receives no duplicate deletion pointer.

## Transaction invariants

- DELETE inserts the action and stamps the exact live table root with `deleted_at` and `deletion_id` in one transaction. A losing root update rolls back the action insert.
- RESTORE locks the exact action and retained root, verifies that the action is recoverable, sets `deleted_at=0`, clears `deletion_id`, and removes the active action in one transaction.
- Coordination is limited to the affected rows; there are no table-wide database locks.
- Successful restore consumes the action. The storage model does not retain a terminal `RESTORED` receipt.
- Legacy metadata garbage collection skips table roots carrying an active `deletion_id`; future purge work owns their final removal.

## After this epic

The relational layer can represent, atomically retain, inspect, and reactivate an existing table row. No public API behavior or physical Iceberg cleanup is introduced here.

## Out of scope

- Iceberg REST DELETE, discovery, and UNDROP routes
- Same-name create/register enforcement
- Physical Iceberg cleanup
- Purge scheduling, workers, job progress, or operator APIs
- Deletion-specific audit storage

### Task list

- [ ] #12222 — schema, migrations, and persistent models
- [ ] #12223 — active deletion and restore transactions

Review order: #12222 → #12223. API integration continues with #12225 under #12221.

roryqi:
@nevzheng Could  u raise the  design document  pull request  first?

nevzheng:
@roryqi 
> [@nevzheng](https://github.com/nevzheng) Could u raise the design document pull request first?

https://github.com/apache/gravitino/pull/12257

Issue #12219, opened 2026-07-27
[EPIC] Support durable soft deletion and recovery


nevzheng:
### Describe the proposal

Add durable soft deletion and recovery to Gravitino, beginning with Iceberg REST tables.

A soft-deleted table is hidden from normal use while its original metadata row is retained for a configured recovery period. During that period, an authorized user can restore the same table identity. Physical cleanup happens asynchronously after the recovery period in a later delivery phase.

## Product invariants

- The original metadata row remains the source of identity, name, parent, ownership, and related metadata while retained.
- Each active deletion generation has an opaque `deletion_id` referenced by the source row.
- The deleted name remains reserved until the table is restored or finally purged.
- Restoration reactivates the original object; it does not create a replacement.
- DELETE never removes Iceberg files on the request thread.
- Existing behavior remains unchanged when soft deletion is disabled.
- Additional entity tables may later adopt the same opaque deletion-pointer pattern. Entity-specific purge context and handler dispatch remain pending design.
- Terminal history belongs in emitted audit/change events, not permanent `entity_deletion` receipts or a new deletion-specific audit table.

## Delivery plan

| Area | Tracking | Milestone |
|---|---|---|
| Relational storage | #12220 | Represent, retain, and reactivate the original table row transactionally. |
| Iceberg REST API | #12221 | DELETE, deleted-name discovery, name reservation, and name-only UNDROP. |
| Durable purge jobs | #12258 | Claim expired deletions and remove Iceberg resources asynchronously with progress and retry. |
| Integration and scale validation | Pending | Exercise concurrency, restart, large-table cleanup, and the required scale scenarios. |
| Additional entity types | Pending | Apply the deletion-pointer pattern beyond Iceberg REST tables. |

## First milestone

After #12220 and #12221 land, the retained-metadata user journey works end to end: DELETE retains and hides the original table, `deleted=true` discovers its name, and UNDROP reactivates that same row. Physical purge, durable job progress, terminal audit events, and scale validation remain separate work.

Related work:

- #12187 tracks broader management-API semantics for metadata undelete.
- #11265 introduced asynchronous hard deletion for the Iceberg REST catalog.

### Task list

- [ ] #12220 — relational deletion-generation storage
- [ ] #12221 — Iceberg REST soft-delete and recovery APIs
- [ ] #12258 — durable asynchronous purge execution
- [ ] Cleanup progress, observability, and operator controls — pending design
- [ ] Integration, concurrency, restart, and scale validation — pending design
[...]

PR #12257, opened 2026-07-29
[#12219] docs: add design doc for soft deletion of Iceberg tables

nevzheng:

Adds a design document for soft deletion of Iceberg tables: a deletion record that makes a dropped table restorable for a bounded window, the API for discovering and undeleting it, and the purge that runs when the window closes.

- §5.3 Metadata model → apache/gravitino#12220 (relational storage epic)
- §5.2 API → apache/gravitino#12221 (Iceberg REST API epic)

**Reviewers:** @roryqi (primary); @lasdf1234, @jerryshao (secondary)

**Review focus:** §5.2 and §5.3 are what I'd most like scrutinized. §5.6 (purge) is offered as a proposal and needs further review — the claim and finalization transactions are new, while scheduling, retry tuning, and operator repair are deliberately deferred.

Follows up on Non-Goal 1 of the merged async hard-deletion design, which defers soft delete and undrop to a separate document.

A dropped Iceberg table is unrecoverable today across all three drop paths. There is no way to see what was dropped or get it back, and the name frees up immediately. This proposes a recoverable window with the name reserved until purge.

Part of apache/gravitino#12219

No — documentation only. The design it describes would add `?deleted=true` discovery, a `POST …/undrop` endpoint, and the `gravitino.entity.soft-delete.*` property keys.

Not applicable — design document only. §8 defines the test plan for the implementation.


jerryshao:
@roryqi and @lasdf1234 please help to review.

lasdf1234:
> @roryqi and @lasdf1234 please help to review.

Got it. I will review this document as soon as possible.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+
+Dropping an Iceberg table is terminal — no window exists in which a mistaken drop
+can be undone.
+
+| Request                                       | What happens                                  | Recoverable |
+| --------------------------------------------- | --------------------------------------------- | ----------- |
+| `DELETE …` (no purge)                         | Registration removed; data files orphaned     | No          |
+| `DELETE …?purgeRequested=true` (synchronous)  | Files deleted on the request thread           | No          |
+| `DELETE …?purgeRequested=true` (asynchronous) | A cleanup job deletes files in the background | No          |
+
```
lasdf1234, line 49: The second and third rows in the table look the same. Could some additional information be included to better distinguish between synchronous and asynchronous processes?

design-docs/deletion-v2/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+   scheduling parameters, retry tuning, and operator repair tooling are deferred.
+2. **Changing existing drop or purge behavior**: the synchronous and asynchronous
+   hard-delete paths are untouched, including their defaults and `purgeRequested`.
+3. **Non-Iceberg connectors**: JDBC, Kafka, and Paimon drops destroy the object at
+   the source. Iceberg is recoverable because a saved metadata pointer can be
+   re-registered.
+4. **Namespace, schema, and view recovery**: different containment rules; a follow-up.
+5. **Per-catalog retention**: phase 1 ships one server-level window.
+6. **Trash / recycle-bin UX**: no Web UI work.
+
```
lasdf1234, line 142: I think Web UI work needs to be synchronized with Mark.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+All new surface extends the **Gravitino metadata routes**. No new route tree, and no
+change to the Iceberg REST wire protocol.
+
+| Operation    | Route                                                                      |
+| ------------ | -------------------------------------------------------------------------- |
+| Drop         | `DELETE /api/metalakes/{m}/catalogs/{c}/schemas/{s}/tables/{t}`             |
+| List deleted | `GET  /api/metalakes/{m}/catalogs/{c}/schemas/{s}/tables?deleted=true`      |
+| Load deleted | `GET  /api/metalakes/{m}/catalogs/{c}/schemas/{s}/tables/{t}?deleted=true`  |
+| Restore      | `POST /api/metalakes/{m}/catalogs/{c}/schemas/{s}/tables/{t}/undrop`        |
+
```
lasdf1234, line 341: | Restore      | `POST /api/metalakes/{m}/catalogs/{c}/schemas/{s}/tables/{t}/undrop`        | 

"undrop" seems a little strange. Maybe a better name could be chosen?

design-docs/deletion-v2/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+   the source. Iceberg is recoverable because a saved metadata pointer can be
+   re-registered.
+4. **Namespace, schema, and view recovery**: different containment rules; a follow-up.
+5. **Per-catalog retention**: phase 1 ships one server-level window.
+6. **Trash / recycle-bin UX**: no Web UI work.
+
+---
+
+## 4. Solution Investigations
+
```
lasdf1234, line 146: Perhaps it would be more beneficial to include here the implementation plans of other manufacturers, such as Polaris and UC ....., for a comparison.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+### 5.8 Configuration
+
+| Key                                         | Default    | Description                                    |
+| ------------------------------------------- | ---------- | ---------------------------------------------- |
+| `gravitino.entity.soft-delete.enabled`      | `false`    | Opt-in. Off means today's behavior exactly.    |
+| `gravitino.entity.soft-delete.retention-ms` | `86400000` | Window length (24h). Valid range 0 to 90 days. |
+
+Phase 1 honors these for Iceberg tables only. Retention `0` means the record is
+immediately expired — no usable window.
+
```
lasdf1234, line 509: Perhaps a catalog-level configuration would be better?

design-docs/deletion-v2/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+Where the feature was never enabled the table is empty and every read is an indexed
+miss.
+
+### 5.8 Configuration
+
+| Key                                         | Default    | Description                                    |
+| ------------------------------------------- | ---------- | ---------------------------------------------- |
+| `gravitino.entity.soft-delete.enabled`      | `false`    | Opt-in. Off means today's behavior exactly.    |
+| `gravitino.entity.soft-delete.retention-ms` | `86400000` | Window length (24h). Valid range 0 to 90 days. |
+
```
lasdf1234, line 506: It feels like 24 hours is a bit too short. Would a week be a more appropriate duration? Could you conduct a study on other systems?

design-docs/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+| Column                 | Purpose                                                                                            |
+| ---------------------- | -------------------------------------------------------------------------------------------------- |
+| `deletion_id`          | Opaque key for one drop. Identifies the generation, not the name.                                  |
+| `entity_type`          | `TABLE` in phase 1. Lets a future purge resolve a target without scanning every entity table.       |
+| `entity_id`            | The original table id, so restore returns the same identity.                                       |
+| `state`                | `DELETED` (recoverable until the deadline) or `PURGING` (claimed, never recoverable).                |
+| `retention_expires_at` | Computed once at drop time; later configuration changes affect only later drops.                     |
+| `purge_requested`      | Whether purge deletes files, or only removes metadata (§5.6).                                      |
+| `purge_job_id`         | Set when a purge claims the record; `NULL` while recoverable.                                       |
+
```
lasdf1234: I think it would be better to expand the table_meta table.  Perhaps we could refer to the solutions of other products.

design-docs/iceberg-table-soft-deletion.md
```diff
@@ -0,0 +1,479 @@
...
+```mermaid
+erDiagram
+    table_meta {
+        bigint table_id PK
+        varchar table_name
+        bigint deleted_at "stays 0 while retained"
+        varchar deletion_id FK "nullable, unique"
+    }
+    entity_deletion {
+        varchar deletion_id PK
```
roryqi: I have some concern about deletion_id. soft deletion is required for IRC. Do we need to modify the schema of the database? Could we simplify our design?

nevzheng: - We do need state. with soft delete. Deleted is a state that needs to be recorded, and manipulated. This will require schema modification and migration.
- How we store it is something we can discuss.
  - I like storing deletion_id + deletion_entity.
  
- Please elaborate on your concerns. IMO the metadata here seems fairly simple. we can discuss it.

design-docs/iceberg-table-soft-deletion.md
nevzheng: - happy to do a comparative study. 
- I feel that this is strong design. inlining this data into the existing table meta doesn't make sense because the the row exposes fields that don't make sense in typical operation. Being deleted is a separate state.
- having a deletion id + deletion entity separates concerns.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: - we can copy the default retention period from the existing design.
- the key is that its configurable.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: - agreed. set retention times at the catalog level. 
- Future work. Policy Tags to set retention

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: Can we be specific about the what we'd like to cover in the competetive analysis.

I'm hearing:
- API comparison
- Metadata Implemenation

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: - this was specified in the PRD (mailed it to you)
- we can do a comparison to other systems
- UNDROP make sense because its the opposite of DROP table.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: - ack. lets defer this. we have a long way before web ui work is done.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: - synchronous is the current version (v1)
- asynchronu(v2)
- I'm not focusing on the exact purge mechanism in this review. but in S5.6 I outline what it would look like.

nevzheng: if you're talking about adding a sync=True or async=true, I'd rather not because that adds complexity. I recommend defaulting to asynchronous w/ immediate expiry. and relying on one system.

nevzheng:
@roryqi @lasdf1234

Pushed an update to the design doc.

The main addition is **Appendix A**, a focused Unity Catalog and Apache Polaris comparison covering their public recovery/drop APIs and the metadata/lifecycle backing available from published docs or source.

I also made the proposal clearer near the top:

- §2 states the preferred V2 deletion-entity model and its API contract.
- §4.1 is the preferred separate `entity_deletion` record, with lifecycle state separate from `table_meta`.
- §4.2 documents the “glob everything onto the current table row” alternative and why I do not recommend it.
- §6 spells out the recommended full V1 → V2 cutover versus the dual-path fallback.

This needs another cleanup pass, but the intended ideas and tradeoffs should now be visible for discussion. `./gradlew :docs:build` passes.

jerryshao:
@lasdf1234 @roryqi please continue to review this doc, thanks.

lasdf1234:
> @lasdf1234 @roryqi please continue to review this doc, thanks.

Got, I will review it again and ensure that each issue has my own solution suggestion.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
lasdf1234: OK. If this synchronous and asynchronous feature is not included in this design, I think this table can be removed because I would get confused when looking at it.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
lasdf1234: May be restore is better? How do you think? @roryqi

design-docs/deletion-v2/iceberg-table-soft-deletion.md
lasdf1234: The overall design scheme of soft deletion can be presented in a table. The table shows the implementation of UC and Polaris, and each row represents a different comparison item.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
lasdf1234: Got.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
lasdf1234: Soft deletion and hard deletion are different scenarios. I feel that the default duration for soft deletion could be longer. @roryqi How do you think?

jerryshao:
Hi @nevzheng , I did a cursory review about the current design doc. Looks like your solution is to introduce a generic soft deletion for all the entities, not just for Iceberg alone. Also, your design is focus on the Gravitino API, not the IRC API. I have several questions:

1. Is it meaningful to add soft deletion support to some data source catalogs that don't have a soft deletion mechanism? For example, like the Hive catalog, what is the behavior after we support the Hive table's soft deletion? I can think of one solution is to only make it invisible from Gravitino side. But it also leaves an inconsistency from Hive's perspective.
2. If we want to support generic soft deletion, shall we expand the scope to not just the table, but also other entities?

These two are fundamental decision questions that we should think carefully about. From my point, I think it is not so necessary to extend to non-IRC scope unless we have a clear requirement. We can focus on IRC alone to implement an IRC-only feature like what we did before for async hard deletion.

design-docs/iceberg-table-soft-deletion.md
lasdf1234: Can we refer to the implementation of hard deletion and create a new table to achieve soft deletion instead of extending the table_meta table? For the scenario of IRC, this design should be better because the metadata may not be registered in the table_meta.

nevzheng:
@jerryshao 

# PR #12257 Review Responses

## Draft response to Jerry — scope and API focus

Jerry, thank you for the review. I heard three related concerns. Please correct me if I
have misunderstood any of them.

### A. Scope of the deletion model

**What I heard:** The model appears general. Is that intentional, and if so, should it cover
other Gravitino entity types too?

**Response:** The model is extensible, but V1 is intentionally narrow: it supports only
`ICEBERG_TABLE` objects through IRC. It defines retention, restore, and purge for that case;
it does not enable other entity types or catalog backends. I support expanding later, but each
type needs its own design for source-system semantics, recovery guarantees, and purge jobs.

I chose this shape because it extends the established
[Iceberg REST asynchronous hard-delete design](async-iceberg-rest-hard-deletion.md) authored
by Rory. We add a recoverable deletion action ahead of its cleanup path: the action owns
retention, recovery, and audit; the existing `iceberg_cleanup_job` owns execution, retries,
and worker progress after expiry. This does not require `table_meta` to be the source of truth.

### B. IRC-only versus broader catalog scope

**What I heard:** Why not make this IRC-only, like the existing asynchronous hard-delete work?

Soft deletion for a source catalog without equivalent semantics—for example Hive—could make
an object invisible through Gravitino while it remains visible in the source catalog. The
first version should therefore be scoped to the IRC/Iceberg use case, analogous to the
existing asynchronous hard-deletion feature. It does not define soft-delete behavior for
Hive, JDBC, or other non-Iceberg catalog types.

### C. API focus and IRC compatibility

**What I heard:** The proposal currently describes the native Gravitino API more than the IRC
API. Do we intend to add `UNDROP` to IRC, or should this be a Gravitino API feature?

**Response:** Great question. I re-read Mark's PRD: it asks for the Iceberg IRC
`/management/...` API (A), so we can meet the immediate requirement through that route. The
upstream IRC specification does not define `UNDROP`, so it would be a documented Gravitino
extension.

**B is also a valid option.** I believe soft deletion in the general Gravitino API is valuable:
it gives users one predictable recovery experience across metadata types and avoids creating a
parallel Iceberg-only recovery surface that we would later duplicate. This is primarily a product
and UX decision, not just an implementation choice.

If we choose B, we should roll it out carefully: Iceberg first, then enable each additional type
only after defining its restore and purge semantics, implementing its handler, documenting its
limits, and adding targeted tests. Both options should call the same deletion record and control
plane. I recommend B as the longer-term direction for those reasons.

I do not want to block the Iceberg work on choosing B. We can carry both options through review
and resolve the public API shape there.

## Next steps

1. **Decide the API shape — @jerryshao @roryqi @lasdf1234 and Mark:** Mark's PRD advocates an
   Iceberg IRC `/management/...` API (A). Nevin is asking whether we should instead establish a
   generalized Gravitino soft-delete/purge API (B), with Iceberg as the first supported type and
   a staged rollout for additional entity types. For now, carry both API shapes against the same
   internal metadata model and control plane; select the initial public API as this review
   resolves. Nevin favors B, provided each type adds its own restore and purge semantics before
   it is enabled. (todo(team)) 
2. Add an appendix that explains the upstream IRC surface, the two management API choices, and
   their shared deletion record and control plane. (done)

nevzheng:
Hi @roryqi @lasdf1234 @jerryshao — I updated the Iceberg soft-deletion design.

Main changes:

- Added the Gravitino Iceberg REST `/management/v1/...` API option alongside the native Gravitino API.
- Added Appendix B covering the Iceberg REST Catalog surface and why deleted-table discovery and UNDROP are Gravitino management extensions.

PTAL, thanks!

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: purgeRequested is part of the API, so I'd like to keep the table here. It's intentionally light — the mechanism gets a proper writeup in a later doc.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: done. See appendix A.

nevzheng:
@lasdf1234 Cleaned up Appendix A into the requested Unity Catalog / Polaris comparison table, with comparison items as rows. I also clarified the V1 versus V2 purge behavior; the exact V2 worker mechanism remains follow-up design work. PTAL, thanks!

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: nack. non-substantive naming debate. undrop is in the PRD, and is also a route used in other systems see appendix a.

review comments should try to make the design substantativelty better, debating naming is typically bike shedding.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: if you accepted the change, you can just say. ack or LGTm and close the thread.

design-docs/deletion-v2/iceberg-table-soft-deletion.md
nevzheng: done I changed the default to two weeks. its adjustable so I wouldn't worry about it too much.

`gravitino.entity.soft-delete.retention-ms` | `1209600000` | Two-week recovery window; valid range 0–90 days.

design-docs/iceberg-table-soft-deletion.md
nevzheng: done. writing appendix A, I didn't see a specific general trend or winner. I think we should choose something reasonable for our requirements.

LMK if you have a alt proposal.

design-docs/iceberg-table-soft-deletion.md
nevzheng: This does create a new table. we add a foreign key into a entity deletion table.

Can you elaborate, or have an alternative idea.

nevzheng:
replied + resolved comments and am ready for review.

Topics to prioritize aligning on:
- Discussing + Accepting the API
- Discussing + Accepting the Metadata model

design-docs/iceberg-table-soft-deletion.md
lasdf1234: I don't think the table_meta table needs to be modified, because for the IRC scenario, it is very likely that table_meta does not store the metadata of iceberg table. So, a new table is added to store the iceberg tables that have been softly deleted.

lasdf1234: This is just a suggestion. The specific table structure may still require your further confirmation:
entity_deletion (or iceberg_table_deletion)
  - deletion_id
  - catalog_key / prefix          # which IRC catalog
  - namespace
  - table_name
  - metadata_location             # needed for undrop / purge
  - deleted_at / retention_expires_at / state / deleted_by
  .............

nevzheng: question: @lasdf1234 I’m not understanding the specific concern or requested change yet. Could you elaborate on the exact requirement, user scenario, or failure in the current design that this alternative is intended to address?

My current understanding is that an Iceberg table may be visible through IRC without a `table_meta` projection. You are therefore proposing an IRC-only deletion record and no `table_meta.deletion_id`.

If that is the concern, this is a scope decision rather than only a schema simplification: should V1 support every backing-catalog table, including unprojected tables, or only Gravitino-known tables?

The current recommendation chooses Gravitino-known tables. It already adds `entity_deletion`; `table_meta.deletion_id` is only the relationship and transactional anchor. That allows exact, atomic restoration of the existing Gravitino metadata tree and preserves the option to expose the same lifecycle through the native Gravitino API later.

Unless the unprojected-table user path is an explicit V1 requirement, I recommend retaining the current model. Please confirm whether I have understood your concern correctly, or describe the scenario you want us to evaluate.

lasdf1234: @nevzheng Thanks — your reading of my concern is largely right; this is mainly a V1 scope decision.

Proposal: Soft delete (including undrop / purge) should apply to every Iceberg table visible through IRC, whether or not Gravitino already has a table_meta projection.

So V1 should cover both:

tables Gravitino already stores metadata for, and
tables that are IRC-visible but not projected into table_meta.
Under that scope, the deletion lifecycle should not depend on extending table_meta with deletion_id. An independent deletion record can be the source of truth for both cases; when table_meta exists it can still be linked/updated for restore, but absence of table_meta must not block soft delete.

roryqi:
Since this feature is scoped to Iceberg REST Catalog, I wonder whether we need a generic `entity_deletion` model and a new `table_meta.deletion_id` column.

I would prefer keeping the Entity Store metadata and its relations unchanged during the recovery window, and extending the existing `iceberg_cleanup_job` instead:

- add `entity_id`, `retention_expires_at`, and a `RETAINED` state;
- use the cleanup-job ID as the immutable `deletionId`;
- while `RETAINED`, use the row as the durable tombstone to hide the table and reserve its identifier, as the current cleanup job already does for name reuse;
- on undrop, consume the retained job and expose the unchanged Entity Store metadata again;
- on expiry or explicit purge, transition `RETAINED -> PENDING`, then reuse the existing worker, heartbeat, CAS, and retry flow.

This avoids reversing the current Entity Store `deleted_at` cascade for columns, owners, tags, policies, and other relations. It also avoids introducing a generic deletion abstraction and changing `table_meta` for an IRC-only feature.

Could the design explain which concrete requirement cannot be satisfied by this approach? In particular, if the native Gravitino table API and non-Iceberg entities are outside the product scope, it is not clear what requires the separate `entity_deletion` record plus the relationship pointer in `table_meta`.

lasdf1234:
> Since this feature is scoped to Iceberg REST Catalog, I wonder whether we need a generic `entity_deletion` model and a new `table_meta.deletion_id` column.
> 
> I would prefer keeping the Entity Store metadata and its relations unchanged during the recovery window, and extending the existing `iceberg_cleanup_job` instead:
> 
> * add `entity_id`, `retention_expires_at`, and a `RETAINED` state;
> * use the cleanup-job ID as the immutable `deletionId`;
> * while `RETAINED`, use the row as the durable tombstone to hide the table and reserve its identifier, as the current cleanup job already does for name reuse;
> * on undrop, consume the retained job and expose the unchanged Entity Store metadata again;
> * on expiry or explicit purge, transition `RETAINED -> PENDING`, then reuse the existing worker, heartbeat, CAS, and retry flow.
> 
> This avoids reversing the current Entity Store `deleted_at` cascade for columns, owners, tags, policies, and other relations. It also avoids introducing a generic deletion abstraction and changing `table_meta` for an IRC-only feature.
> 
> Could the design explain which concrete requirement cannot be satisfied by this approach? In particular, if the native Gravitino table API and non-Iceberg entities are outside the product scope, it is not clear what requires the separate `entity_deletion` record plus the relationship pointer in `table_meta`.

I agree with your answer. Expanding the "iceberg_cleanup_job" might be a better option.This will make the entire system more simpler and clearer.

nevzheng:
note: Hi @roryqi and @lasdf1234 — thanks for the suggestion. This is a material alternative
architecture, not a small revision to the current proposal.

### Recommended V2 model

I recommend retaining the current V2 model. It is grounded in prior research, API and metadata
experimentation, and a tested POC.

### Principles and design choices

- **Separation of concerns and clarity.** A deletion record captures retention, audit, restore,
  and purge eligibility. A cleanup job claims and executes physical purge work. Reusing the
  existing purge worker is valuable, but making its job row also serve as the retained deletion
  record couples two state machines with different lifetimes and responsibilities.
- **Narrow initial scope with deliberate extensibility.** This V2 proposal initially applies only
  to Iceberg tables exposed through IRC. I chose a flexible, more separated data model that can
  extend soft deletion to other object types through later, separately reviewed work. I recognize
  that this goes beyond the narrowest IRC-only implementation; I believe the product optionality
  is valuable, and ask @markhoerth and @jerryshao to weigh in on that trade-off.
- **Iterate before replacing.** I am confident that substantive concerns can be evaluated and
  addressed within the current model before we replace it with a different metadata model.

### Initial critique of the retained-job model

I do not yet have a complete alternative design to evaluate, so this is an initial architectural
objection rather than a claim that the retained-job approach cannot work. My concern is that it
models a logical deletion as a cleanup job.

- A deletion is a durable entity-lifecycle fact: the object was deleted, by whom, when it expires,
  whether it can be restored, and whether it is eligible for purge.
- A cleanup job is an operational fact: work has been scheduled or claimed, it is running or
  retrying, and it eventually succeeds or fails.
- A table should be deleted because it has a deletion record in a deleted state, not because a
  cleanup job happens to exist. The cleanup job should reference an eligible deletion when
  physical purge begins; it should not define the deletion itself.

Combining those concepts forces user-facing retention and recovery semantics into a worker state
machine, and makes audit, visibility, restore, and retry behavior depend on the lifecycle of an
execution row. I prefer the clearer relationship: a durable deletion record may later create or
link to a purge job.

### Alternative and decision

The retained-`iceberg_cleanup_job` model is a valid alternative, but it needs a comparable
metadata/API proposal before it can replace the current design. I cannot lead a parallel design.

If @roryqi or @lasdf1234 would like this option evaluated, please document its implementation
scope, data model, lifecycle, IRC visibility behavior, support for unprojected IRC-visible tables,
restore/purge behavior, and trade-offs.

### Next steps

- @markhoerth and @jerryshao: please weigh in on two linked decisions:
  1. Should V2 use a soft-delete model that is initially delivered only for Iceberg tables through
     IRC, but can be extended to other supported object types through later, separately reviewed
     work?
  2. Should durable deletion, recovery, retention, and audit metadata remain a clean, separate
     lifecycle model rather than being combined with purge-worker execution state in
     `iceberg_cleanup_job`?

  I recommend yes to both.
- @roryqi and @lasdf1234: if the retained-job model should be considered, please provide the
  comparable alternative proposal described above.
- I will update the design and implementation plan to align with the resulting decision.


### Design Review Feedback
note: @roryqi, for future design reviews, I recommend leading with the specific requirement,
correctness risk, or trade-off that the current design does not address, rather than immediately
asking the author to pivot to a different architecture. That lets us first test whether the
existing proposal can resolve the concern, and only then decide whether a material metadata-model
change is warranted.
This approach keeps review actionable, makes trade-offs legible, and lets us move forward as a
team without prematurely discarding completed design work.

jerryshao:
> Should V2 use a soft-delete model that is initially delivered only for Iceberg tables through
IRC, but can be extended to other supported object types through later, separately reviewed
work?
Should durable deletion, recovery, retention, and audit metadata remain a clean, separate
lifecycle model rather than being combined with purge-worker execution state in
iceberg_cleanup_job?

Yes for both. But I don't think the doc change reflects the questions we left before.

PR #12271, opened 2026-07-30
[#12270] fix(glue): Ignore blank AWS static credentials

### LINKED ISSUE #12270: [Bug report] Blank AWS credentials break Glue Iceberg metadata loading
### Version

main, 1.3.0 branch

### Describe what's wrong

When the optional `aws-access-key-id` and `aws-secret-access-key` properties are present as empty strings, `GlueIcebergTableHelper` treats them as configured static credentials.

This causes Iceberg metadata enrichment to fail instead of using the AWS default credential chain. Gravitino then falls back to the Glue schema, losing the accurate Iceberg schema, partition spec, and sort order.

```text
Failed to load Iceberg metadata for table <database>.<table>

java.lang.IllegalStateException: Access key ID is not set
    at org.apache.gravitino.catalog.glue.GravitinoGlueCredentialsProvider.resolveCredentials(...)
```

### How to reproduce

1. Create a Glue catalog using an IAM role or another AWS default credential provider.
2. Set the optional credential properties to empty strings:

   ```properties
   aws-region=us-east-1
   warehouse=s3://<bucket>/<path>
   aws-access-key-id=
   aws-secret-access-key=
   ```

3. Load an Iceberg table from the Glue catalog.
4. Observe that Iceberg metadata loading fails and Gravitino returns only the Glue-derived metadata.

Expected behavior: blank credential properties should be treated as absent, allowing Iceberg to use the AWS default credential chain.

### Additional context

`GlueIcebergTableHelper#createGlueCatalog` only checks whether the credential values are non-null. It should require both values to be non-blank, consistent with `GlueClientProvider`.

xxubai:

Treat blank Glue AWS access and secret keys as absent so that Iceberg uses the AWS default credential chain.

Validate that static access and secret keys are configured together, and add unit coverage for blank and partial credential values.

Optional Glue credential properties may be stored as empty strings. `GlueIcebergTableHelper` previously checked only for null values and installed `GravitinoGlueCredentialsProvider`, which then failed with `Access key ID is not set`.

This prevented Iceberg metadata enrichment and caused Gravitino to fall back to Glue-derived metadata.

Fix: #12270

Yes. Glue Iceberg catalogs with blank static credential properties now use the AWS default credential chain. A partially configured static credential pair fails fast.

No API or property keys are changed.

```bash
./gradlew :catalogs:catalog-glue:test \
  --tests org.apache.gravitino.catalog.glue.TestGlueIcebergTableHelper \
  -PskipITs
```

### DIFF
--- a/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueIcebergTableHelper.java
+++ b/catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueIcebergTableHelper.java
@@ -143,6 +143,15 @@ final class GlueIcebergTableHelper {
    * @throws IllegalArgumentException if {@code aws-region} or {@code warehouse} is not configured
    */
   static Catalog createGlueCatalog(Map<String, String> config) {
+    Map<String, String> icebergProps = buildIcebergCatalogProperties(config);
+
+    GlueCatalog glueCatalog = new GlueCatalog();
+    glueCatalog.initialize("gravitino-glue-iceberg", icebergProps);
+    LOG.info("Initialized Iceberg GlueCatalog for region {}", config.get(GlueConstants.AWS_REGION));
+    return glueCatalog;
+  }
+
+  static Map<String, String> buildIcebergCatalogProperties(Map<String, String> config) {
     String region = config.get(GlueConstants.AWS_REGION);
     Preconditions.checkArgument(region != null, "AWS region is required for Iceberg Glue catalog");
 
@@ -164,18 +173,14 @@ final class GlueIcebergTableHelper {
 
     String accessKey = config.get(GlueConstants.AWS_ACCESS_KEY_ID);
     String secretKey = config.get(GlueConstants.AWS_SECRET_ACCESS_KEY);
-    boolean hasAccessKey = StringUtils.isNotBlank(accessKey);
-    boolean hasSecretKey = StringUtils.isNotBlank(secretKey);
-    Preconditions.checkArgument(
-        hasAccessKey == hasSecretKey,
-        "Both '%s' and '%s' must be set together.",
-        GlueConstants.AWS_ACCESS_KEY_ID,
-        GlueConstants.AWS_SECRET_ACCESS_KEY);
-    if (hasAccessKey) {
+    boolean hasCredential = GlueClientProvider.hasAwsStaticCredentials(accessKey, secretKey);
+    if (hasCredential) {
       icebergProps.put(
           CLIENT_CREDENTIALS_PROVIDER, GravitinoGlueCredentialsProvider.class.getName());
       icebergProps.put(CLIENT_CREDENTIALS_PROVIDER_ACCESS_KEY_ID, accessKey);
       icebergProps.put(CLIENT_CREDENTIALS_PROVIDER_SECRET_ACCESS_KEY, secretKey);
+      icebergProps.put(IcebergConstants.ICEBERG_S3_ACCESS_KEY_ID, accessKey);
+      icebergProps.put(IcebergConstants.ICEBERG_S3_SECRET_ACCESS_KEY, secretKey);
     }
 
     String endpoint = config.get(GlueConstants.AWS_GLUE_ENDPOINT);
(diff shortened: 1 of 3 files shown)


catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueIcebergTableHelper.java
```diff
@@ -189,6 +190,20 @@ static Catalog createGlueCatalog(Map<String, String> config) {
     return glueCatalog;
   }
 
+  @VisibleForTesting
+  static boolean validateAwsStaticCredential(String accessKey, String secretKey) {
```
yuqi1129: We'd better improve the readability here. `validateAwsStaticCredential` means that we will check the validity of `accessKey` and `secretKey`. In the normal course of events, `true` means it's okay and `false` means it's invalid. However, when both `accessKey` and `secretKey` are empty, the check result is `true`, but the return value is `false`

xxubai: updated

yuqi1129:
LGTM. @diqiu50 
Please also take a look; it's your field.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueIcebergTableHelper.java
```diff
@@ -164,7 +164,14 @@ static Catalog createGlueCatalog(Map<String, String> config) {
 
     String accessKey = config.get(GlueConstants.AWS_ACCESS_KEY_ID);
     String secretKey = config.get(GlueConstants.AWS_SECRET_ACCESS_KEY);
-    if (accessKey != null && secretKey != null) {
+    boolean hasAccessKey = StringUtils.isNotBlank(accessKey);
+    boolean hasSecretKey = StringUtils.isNotBlank(secretKey);
+    Preconditions.checkArgument(
```
diqiu50: We'd better make the error message is the same as GlueClientProvider

diqiu50: The logic is the same in GlueClientProvider, Could you consider reuse it

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueIcebergTableHelper.java
```diff
@@ -235,4 +239,75 @@ void testLoadTableOverwritesColumnTypesFromIcebergSchema() {
     assertEquals(Types.TimeType.of(6), cols[1].dataType());
     assertEquals(Types.DecimalType.of(10, 2), cols[2].dataType());
   }
+
+  @Test
+  void testCreateGlueCatalogWithoutStaticCredentials() {
+    Map<String, String> config = new HashMap<>();
+    config.put(GlueConstants.AWS_REGION, "us-east-1");
+    config.put(GlueConstants.WAREHOUSE, "s3://test-bucket/warehouse");
+
```
diqiu50, line 250: The new blank-credential test doesn't actually cover the regression. createGlueCatalog() never threw on blank keys — the failure surfaces later in GravitinoGlueCredentialsProvider.resolveCredentials(). So testCreateGlueCatalogWithBlankStaticCredentials passes even without the fix.

catalogs/catalog-glue/src/main/java/org/apache/gravitino/catalog/glue/GlueIcebergTableHelper.java
xxubai: done

catalogs/catalog-glue/src/test/java/org/apache/gravitino/catalog/glue/TestGlueIcebergTableHelper.java
xxubai: Moved the Iceberg properties into a separate code block.

diqiu50 (approved):
LGTM

PR #12274, opened 2026-07-30
[#12273] fix(clickhouse): preserve MergeTree engine parameters through create-load round-trip

### LINKED ISSUE #12273: [Bug report] ClickHouse MergeTree engine parameters lost on table load
### Describe what's wrong

When loading a ClickHouse table that uses a MergeTree-family engine with
parameters — for example `ReplacingMergeTree(ts)` — Gravitino reads only the
engine name from `system.tables.engine` and discards the parameters in
parentheses.

After a create→load round-trip (or load of a pre-existing table), the engine
parameters are gone. `ReplacingMergeTree(ts)` becomes plain `ReplacingMergeTree`,
which changes the table's deduplication behaviour.

The `system.tables.engine_full` column already contains the full engine
declaration (e.g. `ReplacingMergeTree(ts) ORDER BY id SETTINGS ...`), but it is
only parsed for `Distributed` engines today.

No error is thrown. The engine parameters are silently dropped — the table loads
without them and subsequent DDL generation omits the parameters as well.

### How to reproduce

1. Start a Gravitino server with a ClickHouse catalog.
2. Create a table in ClickHouse with engine parameters:
   ```sql
   CREATE TABLE test.t (id Int32, ts Int64) ENGINE = ReplacingMergeTree(ts) ORDER BY id
   ```
3. Load the table via Gravitino:
   ```java
   Table t = catalog.asTableCatalog().loadTable(NameIdentifier.of("test", "t"));
   ```
4. Check `t.properties()` — no `engine_parameters` key exists, and the engine
   is recorded as plain `ReplacingMergeTree`.

### Additional context

The following MergeTree-family engines accept parameters that affect correctness
(not just performance):

| Engine | Example parameter | Lost semantic |
|---|---|---|
| ReplacingMergeTree | `ts` (version column) | dedup behaviour |
| CollapsingMergeTree | `sign` (sign column) | collapse behaviour |
| VersionedCollapsingMergeTree | `sign, ver` | versioned collapse |
| SummingMergeTree | `(a, b)` (tuple of columns) | auto-sum columns |
| GraphiteMergeTree | `'config_section'` | graphite rollup config |

`AggregatingMergeTree` takes no parameters and is unaffected.

jiangxt2:

Read the `engine_full` column from `system.tables` (already fetched via `SELECT *`) to extract engine parameters for non-Distributed MergeTree-family engines, store them as a new `engine_parameters` table property, and include them when generating `ENGINE = ...` clauses in `CREATE TABLE` DDL.

The `engine` column in `system.tables` only returns the engine name (e.g. `ReplacingMergeTree`) without parameters. After a create→load round-trip (or loading a pre-existing table), the parameters in parentheses are silently dropped. `ReplacingMergeTree(ts)` becomes plain `ReplacingMergeTree` — the deduplication version column is lost, changing the table's correctness semantics.

This affects five MergeTree-family engines:

- `ReplacingMergeTree` — version column for dedup
- `CollapsingMergeTree` — sign column for collapse
- `VersionedCollapsingMergeTree` — sign + version for versioned collapse
- `SummingMergeTree` — columns (including nested-paren tuples) for auto-sum
- `GraphiteMergeTree` — graphite rollup config

The `engine_full` column already contains the full engine DDL and is already parsed for `Distributed` engines; extending this to other engines reuses the existing query path.

Fix: #12273

A new optional table property `engine_parameters` is added for ClickHouse tables. Users can set it when creating tables (e.g. `engine_parameters = "ts"` for `ReplacingMergeTree(ts)`), and it will appear in `table.properties()` for loaded tables that have engine parameters.

No existing property keys are changed or removed.

- Unit tests: 9 test cases covering single-param, multi-param, nested-paren (`SummingMergeTree((a, b))`), no-param, blank/null input, engine name mismatch, `GraphiteMergeTree` quote preservation, and `AggregatingMergeTree` no-param.
- Docker IT: 5 round-trip tests on a real ClickHouse container covering `ReplacingMergeTree`, `SummingMergeTree` (nested parens), `VersionedCollapsingMergeTree` (multi-param), `CollapsingMergeTree` (load from existing table), and `MergeTree` (no params). Each test verifies both the Gravitino property value and the `SHOW CREATE TABLE` output.


catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -458,7 +458,12 @@ private ClickHouseTablePropertiesMetadata.ENGINE appendTableEngine(
       return engine;
     }
 
-    sqlBuilder.append("\n ENGINE = %s".formatted(engine.getValue()));
+    String engineParams = StringUtils.trim(properties.get(TableConstants.ENGINE_PARAMETERS));
```
yuqi1129: GraphiteMergeTree still cannot round-trip through the new property. The earlier Graphite branch reads only `graphite.config` and returns before this generic path, but the load path stores the extracted value only as `engine_parameters`. Therefore creating Graphite with the new public property fails, and properties returned by `loadTable` cannot recreate the same engine. Please keep Graphite on the dedicated `graphite.config` path by populating that property (unquoted) during load and excluding Graphite from generic extraction, or unify the two paths consistently. A SQL-generation round-trip unit test can cover this without requiring a configured Graphite server.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -1642,6 +1661,43 @@ private String buildDataSkippingIndexDdl(
...
+    if (paramsStart >= engineFull.length() || engineFull.charAt(paramsStart) != '(') {
+      return null;
+    }
+    // Bracket-depth counting handles nested parentheses (e.g. SummingMergeTree((a, b))).
+    int depth = 0;
+    for (int i = paramsStart + 1; i < engineFull.length(); i++) {
+      char ch = engineFull.charAt(i);
+      if (ch == '(') {
+        depth++;
+      } else if (ch == ')') {
```
yuqi1129: The depth scan is not quote-aware. For example, an engine parameter containing a backtick-quoted identifier whose name is `ver)` will terminate at the `)` inside the identifier and return a truncated value. Parentheses inside string literals have the same problem. Please track single-quoted/backtick-quoted and escaped state (or use a parser), and add regression tests for quoted identifiers and strings.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -705,6 +710,20 @@ protected Map<String, String> getTableProperties(Connection connection, String t
                       put(ClusterConstants.ON_CLUSTER, String.valueOf(false));
                     }
 
+                    // Extract engine parameters from engine_full for non-Distributed engines.
+                    // engine_full contains the full engine DDL (e.g.
+                    // "ReplacingMergeTree(ts) ORDER BY id SETTINGS ..."), while the engine
+                    // column only contains the engine name without parameters.
+                    // For Distributed engines, parameters are already stored as separate
+                    // distributed-table properties below.
+                    if (!StringUtils.equalsIgnoreCase(engine, ENGINE.DISTRIBUTED.getValue())) {
```
yuqi1129: The scope check does not match the property contract. This extracts parameters for every non-Distributed engine, including MySQL/JDBC/S3/Kafka/Join/Buffer/File. `engine_full` for integration engines contains connection arguments (and may contain or mask credentials), while `ENGINE_PARAMETERS_PROPERTY_ENTRY` is public and non-hidden; the create path also re-emits the value verbatim. This exposes engine configuration through `table.properties()` and silently expands CREATE support beyond the intended MergeTree family. Please gate both extraction and rendering on an explicit supported MergeTree engine set, handle Graphite separately, and add a non-MergeTree regression test.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Fixed in 7517295a. Graphite now uses only graphite.config; generic engine_parameters extraction excludes Graphite. Load and CREATE use symmetric SQL-literal escaping, with a round-trip unit test added.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Fixed in 7517295a. The scanner now handles quoted strings, quoted identifiers, escapes, and nested parentheses correctly, with corresponding regression tests.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
jiangxt2: Fixed in 7517295a. LOAD and CREATE now share an allowlist of four supported MergeTree engines. Graphite and Distributed use dedicated properties, while other engines neither expose nor accept engine_parameters.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/operations/ClickHouseTableOperations.java
```diff
@@ -705,6 +710,20 @@ protected Map<String, String> getTableProperties(Connection connection, String t
...
+                      String engineFull = resultSet.getString("engine_full");
+                      String engineParams = extractEngineParams(engine, engineFull);
+                      if (StringUtils.isNotBlank(engineParams)) {
+                        put(TableConstants.ENGINE_PARAMETERS, engineParams);
+                      }
+                    }
```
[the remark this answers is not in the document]
jiangxt2: Fixed in 7517295a. Graphite parameters are restored only as the unquoted graphite.config value and excluded from generic engine_parameters, so load-to-create round-tripping is consistent.

catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/ClickHouseTablePropertiesMetadata.java
```diff
@@ -99,6 +99,15 @@ public class ClickHouseTablePropertiesMetadata extends JdbcTablePropertiesMetada
...
           false);
 
+  public static final PropertyEntry<String> ENGINE_PARAMETERS_PROPERTY_ENTRY =
+      stringOptionalPropertyEntry(
+          TableConstants.ENGINE_PARAMETERS,
+          "Engine parameters extracted from engine_full (e.g. \"ts\" for ReplacingMergeTree(ts)). "
+              + "Only used for non-Distributed MergeTree-family engines.",
+          false,
+          "",
+          false);
```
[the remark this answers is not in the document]
jiangxt2: Fixed in 7517295a. The property description now states that engine_parameters can be supplied during CREATE and restored by loadTable, and is limited to the four supported MergeTree engines.

catalogs-contrib/catalog-jdbc-clickhouse/src/test/java/org/apache/gravitino/catalog/clickhouse/operations/TestClickHouseTableOperationsUnit.java
```diff
@@ -84,4 +84,92 @@ void testGetIndexesSqlEscapesSingleQuotes() throws Exception {
...
+    Assertions.assertNull(ClickHouseTableOperations.extractEngineParams("MergeTree", null));
+    Assertions.assertNull(
+        ClickHouseTableOperations.extractEngineParams(null, "MergeTree ORDER BY id"));
+    Assertions.assertNull(
+        ClickHouseTableOperations.extractEngineParams("", "MergeTree ORDER BY id"));
+  }
+
+  @Test
+  void testExtractEngineParamsEngineNameNotAtStart() {
+    // Engine name appears later in the string — should not match because '(' check fails
```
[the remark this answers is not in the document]
jiangxt2: Fixed in 7517295a. The comment now correctly states that the method returns null because the input does not start with the requested engine name.

jiangxt2:
Hi @yuqi1129, could you please take another look when you have time? Thanks!

yuqi1129 (approved):

PR #12275, opened 2026-07-30
[#12272] feat(clickhouse): map ClickHouse Date32 to Gravitino DateType and verify Enum round-trip

### LINKED ISSUE #12272: [Improvement] Map ClickHouse Date32 to native Gravitino DateType and verify Enum round-trip

ClickHouse Date32 (range 1900–2299) currently maps to ExternalType via the default branch in `ClickHouseTypeConverter.toGravitino()`. Gravitino already has a `Types.DateType` which represents a calendar date and is the closest native match.

Additionally, ClickHouse Enum8/Enum16 types are not covered by integration tests, leaving a gap in verifying their round-trip correctness through the Gravitino type system.

### How should we improve?

1. **Date32 → DateType**: Add `case DATE32: return Types.DateType.get()` in `toGravitino()`. This is a one-way downgrade — the read path maps Date32 to DateType, but the write path maps DateType back to `Date` (not `Date32`), because `fromGravitino` maps `DateType` to the `Date` string without preserving the source precision. This is consistent with the existing `DateTime64(3)` → `TimestampType.withoutTimeZone(3)` → `DateTime(0)` pattern, where `fromGravitino` maps `TimestampType` back to `DateTime` (dropping sub-second precision) rather than `DateTime64(3)`. Users who need the full 1900–2299 range can use ExternalType explicitly when creating tables.

2. **Enum IT test**: Add an integration test that:
   - Creates a ClickHouse table with Enum8 and Enum16 columns
   - Verifies they load as ExternalType with the full enum definition preserved (including spaces and equals signs)
   - Validates round-trip: recreating the table through Gravitino using the loaded schema and confirming the DDL output matches the original

jiangxt2:

1. **Date32 → DateType**: Add `case DATE32` in `ClickHouseTypeConverter.toGravitino()`, mapping ClickHouse Date32 to Gravitino `Types.DateType`. Previously Date32 fell through to the default branch and was mapped to `ExternalType`, despite being a calendar date type semantically identical to Date (just with a wider year range: 1900–2299 vs 1970–2149).

2. **Enum IT test**: Add `testEnumRoundTrip()` integration test verifying that Enum8/Enum16 round-trip correctly through the Gravitino type system via ExternalType. The enum definitions (including spaces and equals signs) are fully preserved across create → load → recreate cycles.

Date32 was the last user-visible ClickHouse type that has a Gravitino native counterpart (`Types.DateType`) but was not mapped to it. The `DateTime64(3)` → `TimestampType.withoutTimeZone(3)` → `DateTime(0)` pattern already establishes the precedent of one-way precision/range narrowing in the ClickHouse converter — Date32 follows the same pattern.

Mapping Date32 to `DateType` (instead of `ExternalType`) also keeps the column usable in downstream systems: the Gravitino Spark connector currently throws `UnsupportedOperationException` for `ExternalType` columns, making the entire table inaccessible (see #12104), and the proposed fix there maps them to `StringType`, losing the date semantics entirely. A native `DateType` mapping preserves the date semantics across Spark/Flink/Trino integrations.

Enum8/Enum16 already work correctly via ExternalType pass-through, but had no integration test coverage, leaving a gap in verifying their round-trip correctness. The test confirms no code change is needed for Enum support.

Fix: #12272

Yes, with a caveat. After this change:
- **Read path**: ClickHouse `Date32` columns are now loaded as Gravitino `DateType` (instead of `ExternalType("Date32")`)
- **Write path**: `DateType` is still mapped to ClickHouse `Date` (not `Date32`), so round-tripping a Date32 table through Gravitino will produce `Date` on the ClickHouse side, narrowing the supported range from 1900–2299 to 1970–2149-06-06

⚠️ **Silent saturation of out-of-range values** (verified against ClickHouse 25.3.2.39): ClickHouse does not reject dates outside the `Date` range — it silently saturates them. Writing `1900-01-01` to a `Date` column stores `1970-01-01`, and writing `2150-01-01` stores `2149-06-06`. This only affects values in 1900–1969 or after 2149-06-06, which are rare in practice.

Users who need the full 1900–2299 range should use `ExternalType("Date32")` explicitly when creating tables — the ExternalType pass-through path remains fully available.

This is consistent with the existing `DateTime64(3)` → `DateTime(0)` precision narrowing pattern.

- Unit tests: `TestClickHouseTypeConverter` — verified Date32 now maps to `DateType.get()` (previously `ExternalType.of("Date32")`)
- Docker integration tests: `CatalogClickHouseIT.testEnumRoundTrip` — verified Enum8/Enum16 round-trip correctness (create → load → verify → recreate → SHOW CREATE TABLE) against ClickHouse 25.3.2.39
- Static checks: `spotlessCheck`, `rat` (license headers), `gravitino-pr-precheck.py` — all passed

### DIFF
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/converter/ClickHouseTypeConverter.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/converter/ClickHouseTypeConverter.java
@@ -162,8 +162,11 @@ public class ClickHouseTypeConverter extends JdbcTypeConverter {
       case DATE:
         return Types.DateType.get();
       case DATE32:
-        // Date32 supports 1900-2299 vs Date's 1970-2149. Use ExternalType to preserve round-trip.
-        return Types.ExternalType.of(DATE32);
+        // Date32 supports 1900-2299 vs Date's 1970-2149-06-06. Map to DateType as a lossy
+        // downgrade: writing DateType yields Date (not Date32), and out-of-range Date32 values
+        // are silently saturated by ClickHouse (e.g. 1900-01-01 -> 1970-01-01). Use
+        // ExternalType("Date32") when lossless round-trip of the wider range is required.
+        return Types.DateType.get();
       case DATETIME:
         // Default is 0 precision
         return Types.TimestampType.withoutTimeZone(0);
--- a/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/converter/ClickHouseTypeConverter.java
+++ b/catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/converter/ClickHouseTypeConverter.java
@@ -162,11 +162,8 @@ public class ClickHouseTypeConverter extends JdbcTypeConverter {
       case DATE:
         return Types.DateType.get();
       case DATE32:
-        // Date32 supports 1900-2299 vs Date's 1970-2149-06-06. Map to DateType as a lossy
-        // downgrade: writing DateType yields Date (not Date32), and out-of-range Date32 values
-        // are silently saturated by ClickHouse (e.g. 1900-01-01 -> 1970-01-01). Use
-        // ExternalType("Date32") when lossless round-trip of the wider range is required.
-        return Types.DateType.get();
+        // Date32 has a wider range than Date; preserve its catalog type for round-trip.
+        return Types.ExternalType.of(DATE32);
       case DATETIME:
         // Default is 0 precision
         return Types.TimestampType.withoutTimeZone(0);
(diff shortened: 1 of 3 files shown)


catalogs-contrib/catalog-jdbc-clickhouse/src/main/java/org/apache/gravitino/catalog/clickhouse/converter/ClickHouseTypeConverter.java
```diff
@@ -162,8 +162,11 @@ public Type toGravitino(JdbcTypeBean typeBean) {
       case DATE:
         return Types.DateType.get();
       case DATE32:
-        // Date32 supports 1900-2299 vs Date's 1970-2149. Use ExternalType to preserve round-trip.
-        return Types.ExternalType.of(DATE32);
+        // Date32 supports 1900-2299 vs Date's 1970-2149-06-06. Map to DateType as a lossy
+        // downgrade: writing DateType yields Date (not Date32), and out-of-range Date32 values
+        // are silently saturated by ClickHouse (e.g. 1900-01-01 -> 1970-01-01). Use
+        // ExternalType("Date32") when lossless round-trip of the wider range is required.
+        return Types.DateType.get();
```
yuqi1129: The only problem is that we can't tell the exact type when converting Gravitino `DateType`  to Clickhouse type.

jiangxt2: Thanks for pointing this out. DateType cannot distinguish ClickHouse Date from Date32, so I kept Date32 as ExternalType("Date32") to preserve its type and range. The fix is in commit 7975bc455, and the relevant tests pass.

yuqi1129: OK

yuqi1129 (approved):

PR #12285, opened 2026-07-30
[MINOR] docs(lance): update the Lance and generic lakehouse catalog pages

markhoerth:

Updates the five Lance and generic lakehouse catalog documentation pages around the model they never stated, and corrects claims that turned out to be wrong when checked against source. 681 insertions, 1351 deletions.

#### Corrections verified against source

- Alter is partially supported (drop column, rename column, add index), not "Not support now" (`LanceTableOperations.handleLanceTableChange`)
- Index creation is supported through alter; the page listed indexes under "Feature Limitations" (same)
- Name `lance.schema-refresh-mode`, which the refresh section described at length but never identified (`LanceConstants`)
- Add `lance.declared` to the table properties, which was missing (`LanceTableDelegator.tablePropertyEntries`)
- `Map` is supported by Lance; the page said it was not (lance `docs/src/format/table/schema.md`)
- `Union` is not supported by Lance; the page implied it was (same)
- `Interval_year` is not supported because Lance has no interval type at all; `Interval_day` is fine because it maps to an Arrow `Duration` (same)
- Vector columns need a fixed-size list. Gravitino `List` produces a variable-length Arrow `List`, which cannot carry a vector index (`LanceDataTypeConverter`, lance schema spec)
- Mark `GET /v1/namespace/list` as a Gravitino extension, absent from the specification and previously listed alongside spec endpoints with no distinction (lance-namespace `v0.7.5` `openapi.yaml`)
- Add three endpoints missing from the operations table: `/v1/namespace/list`, `drop_columns`, `alter_columns` (`LanceNamespaceOperations`, `LanceTableOperations`)
- Delta `location` inherits from schema and catalog; resolution happens before the format delegator runs (`GenericCatalogOperations.calculateTableLocation`)
- Gravitino performs no Delta type conversion; the mapping table is guidance for declaring metadata, not a conversion it applies (no converter exists in the `delta` package)
- Document three Delta create-time rejections that were undocumented: distributions, sort orders, indexes (`DeltaTableOperations.createTable`)
- Fix the `lance-ray` example, which used `write_lance(namespace=...)`, a signature the page's own compatibility notes call unsupported (`clients/client-python/tests/integration/test_lance_ray.py`)
- `provider` is a field on the create request, not a catalog property as the page listed it (`GenericCatalogPropertiesMetadata`)
- Add `lance.storage.*` to the catalog properties table, although three pages tell readers to set it there (same)
- The generic catalog supports Lance and Delta, not "Lance, Iceberg, Hudi" as the page claimed (`META-INF/services`)

#### `lance-rest-service.md`

- Delete the Deployment Modes table; state the recommended deployment, then the alternative and its limitations
- New `Lance REST API` section: endpoint table, identifiers, calling details, limitations, storage options
- Fold `Namespace Model` into the API section as `Identifiers`; hierarchy and delimiter exist only to let a reader build a request
- Dissolve the orphan `Authentication` and `Storage Configuration` sections into `Configuration` and the API section
- State where configuration properties go, and that a separate process uses the same names in a different file
- Structure configuration by outcome rather than by property
- Cut `Engine Integration`; which engines can read Lance is the integration page's subject
- Cut the Iceberg REST counterpart paragraph
- Add a Quick Start, correct the specification links, consolidate the configuration tables
- Move the endpoint examples to the Lance tables page so both APIs get equal treatment
- Reorder the Quick Start so the server starts once; the previous ordering implied start, stop, edit config, start again
- Correct readiness. `NamespaceWrapper` initializes lazily, so `/health/ready` returns 503 until the first namespace or table call. The Quick Start previously checked readiness before the call that triggers initialization, so the check failed as written
- Create the metalake in the Quick Start, which the service requires but the page never mentioned
- Pass `location` and `lance.storage.*` in the Quick Start namespace create; it previously used an `s3://` path with no credentials, so it could not work
- Split `Gravitino-Specific Behavior` into `Calling the Operations` and `Limitations`; the old heading claimed everything under it was a deviation, but several rows were plain specification mechanics
- State that Gravitino implements a subset. The `v0.7.5` specification defines 29 table endpoints; Gravitino serves 9, with the entire data plane absent
- Document that `DropTable` purges. Every table this service creates is marked `external`, but `DropTable` removes the dataset files anyway
- Rewrite storage options to say the prefix is stripped: `lance.storage.region` reaches the client as `region`
- Note that health endpoints bypass the authentication filter

#### `lance-rest-integration.md`

- Reorganize per engine; version information now sits with the engine it applies to, instead of a shared matrix that implied the two libraries were coupled
- Promote the compatibility content out of a `:::note` block into real sections
- Cut `Rationale`, which claimed incompatible versions cause data corruption
- Cut `General Integration Pattern`, which was contentless
- Remove the duplicated MinIO catalog example
- Keep Table Location on the Lance path; the old page reached for a Gravitino API call to set properties the Lance `CreateNamespace` call accepts directly
- Correct the claim that an engine without Lance REST support needs its own credentials; `DescribeTable` returns storage options alongside the location
- Drop the Quick Start; with only two engines it privileged Spark arbitrarily, and each engine section is its own quick start
- Add a credentials section pointing at the service page

#### `lakehouse-generic-catalog.md`

- Rewrite the Overview around what distinguishes this catalog: Lance and Delta define no catalog of their own, so Gravitino holds the metadata directly rather than federating. Explains the word "generic", which the page previously asserted without grounding
- Split catalog creation per format; a Lance catalog carries `location` and `lance.storage.*`, a Delta catalog needs no properties at all
- Reorganize properties so format-neutral ones come first and Lance-only ones are separated
- Move location resolution next to the `location` rows it explains, and convert a 20-line ASCII block into a table plus one worked paragraph
- Document the `lance.storage.*` option names; four pages told readers to set them and none listed them
- Cut the Benefits list and both bullet lists of "Supported Operations"
- Remove `"owner": "sales-team"` from the schema example, which implied a property that does not exist

#### `lakehouse-generic-lance-table.md`

- Restructure to be path-neutral; the page is reachable from both APIs and previously read as Gravitino-API-only, giving a Lance REST reader instructions that do not apply
- Replace `Supported Operations` with a `Capabilities` table giving both APIs side by side
- Add a column to the table properties showing how the Lance REST service supplies each one, instead of marking five of eight as irrelevant
- Cover both APIs symmetrically in Examples, with the Lance examples moved here from the service page
- Add Troubleshooting with ten error strings taken verbatim from source
- Cut the Migration Guide, a bash `for` loop registering tables
- Preserve `#data-type-mappings`, which `manage-relational-metadata-using-gravitino.md` links into

#### `lakehouse-generic-delta-table.md`

- Lead the Overview with the constraint that matters: every Delta table is external and a create is a registration
- State the consequence the old page only implied: Gravitino and the `_delta_log` are two records of the same table, and Gravitino never reconciles them
- Align the `Capabilities` table with the Lance page's phrasing so the two formats can be compared
- Add Troubleshooting with eight verbatim error strings, three of which the old page did not mention
- Cut Planned Enhancements, which speculated about unreleased features
- Cut the 20-line Spark modification example and the read-via-Spark snippet, both of which belong to Delta Lake's documentation

#### Conventions applied across all five pages

- No `:::` admonition blocks, no `<Tabs>` / `<TabItem>`, no ✅ / ❌ in tables, no em dashes
- No `Since Version` columns or "Introduced in X" sentences, since the docs are version-selected
- No raw HTML in table cells; `<sup>` renders literally, so footnote markers use Unicode superscripts
- Title Case headings, `{snake_case}` placeholders, aligned table pipes
- Shell variables declared once per code block so each block stands alone
- Code lines kept short enough to avoid horizontal scroll
- No one-column tables and no sections with a single child

Five pages document Lance in Gravitino, and none of them states the model that makes the set coherent: Lance tables live in a `lakehouse-generic` catalog, and two APIs reach the same tables. Configure the Lance REST service and it creates the catalog when a client first connects, or create the catalog through the Gravitino REST API and skip the service entirely.

Because that was never written down, the pages were organized by artifact rather than by the model. Authentication appeared on two pages. Storage configuration lived on one page and engine configuration on another, with no cross-reference. A Lance REST user landing on the tables page got operations and properties that do not apply on their path. The word "generic" was never explained. Following the service page Quick Start in order did not work, because it checked readiness before the call that initializes the service.

The five files cannot be split into smaller PRs. Ten cross-page anchors point at headings that only exist after this rewrite, running in every direction between the files, so no ordering of smaller PRs avoids shipping broken links.

`lance-rest-server-chart.md` is deliberately not included. Its most useful content requires a chart change rather than a doc change, and its doc-only fixes will follow separately.

Documentation only. No API, property, or behavior changes.

The pages themselves change substantially, including headings. All inbound anchors from other documentation pages were checked and preserved, including `#data-type-mappings`, which `manage-relational-metadata-using-gravitino.md` links into.

Documentation only, so no code tests. Claims were checked as follows:

- Endpoint names, methods, and paths against lance-namespace `v0.7.5` `openapi.yaml`
- Type mappings against `LanceDataTypeConverter` and the Lance format schema specification
- Property names and defaults against `LanceConstants`, `GenericCatalogPropertiesMetadata`, `GenericTablePropertiesMetadata`, and `LanceTableDelegator`
- Operation support against `LanceTableOperations` and `DeltaTableOperations`
- Error strings copied verbatim from source
- The `lance-ray` example against `clients/client-python/tests/integration/test_lance_ray.py`
- All cross-page and in-page anchors verified to resolve

### DIFF
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -13,6 +13,7 @@ license: "This software is licensed under the Apache License version 2."
 ## Overview
 
 Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
+
 Each table names its format with the required `format` property, which is immutable once set, and Gravitino selects the matching implementation. Two are registered today.
 
 - **Lance**, as `format=lance`. Fully supported, and documented in [Lance Tables](./lakehouse-generic-lance-table.md).
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -14,7 +14,6 @@ license: "This software is licensed under the Apache License version 2."
 
 Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
-Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
 Each table names its format with the required `format` property, which is immutable once set, and Gravitino selects the matching implementation. Two are registered today.
 
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -12,7 +12,7 @@ license: "This software is licensed under the Apache License version 2."
 
 ## Overview
 
-Some table formats define a catalog of their own, and Apache Gravitino federates to it. 
+Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
 Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -12,7 +12,7 @@ license: "This software is licensed under the Apache License version 2."
 
 ## Overview
 
-Some table formats define a catalog of their own, and Apache Gravitino federates to it. Apache Iceberg is the example: the Iceberg catalog connects to an existing Iceberg REST catalog or Hive metastore, which remains the system of record.
+Some table formats define a catalog of their own, and Apache Gravitino federates to it. 
 
 Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -14,7 +14,6 @@ license: "This software is licensed under the Apache License version 2."
 
 Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
-
 Each table names its format with the required `format` property, which is immutable once set, and Gravitino selects the matching implementation. Two are registered today.
 
 - **Lance**, as `format=lance`. Fully supported, and documented in [Lance Tables](./lakehouse-generic-lance-table.md).
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -13,7 +13,6 @@ license: "This software is licensed under the Apache License version 2."
 ## Overview
 
 Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
-
 Each table names its format with the required `format` property, which is immutable once set, and Gravitino selects the matching implementation. Two are registered today.
 
 - **Lance**, as `format=lance`. Fully supported, and documented in [Lance Tables](./lakehouse-generic-lance-table.md).
--- a/docs/lakehouse-generic-catalog.md
+++ b/docs/lakehouse-generic-catalog.md
@@ -4,200 +4,166 @@ slug: "/lakehouse-generic-catalog"
 keywords:
   - lakehouse
   - lance
+  - delta
   - metadata
   - generic catalog
-  - file system
 license: "This software is licensed under the Apache License version 2."
 ---
 
-import Tabs from '@theme/Tabs';
-import TabItem from '@theme/TabItem';
-
 ## Overview
 
-The Generic Lakehouse Catalog is a Gravitino catalog implementation designed to seamlessly integrate with lakehouse storage systems built on file system-based architectures. This catalog enables unified metadata management for lakehouse tables stored on various storage backends, providing a consistent interface for data discovery, governance, and access control. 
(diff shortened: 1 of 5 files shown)


docs/lakehouse-generic-catalog.md
```diff
@@ -4,200 +4,166 @@ slug: "/lakehouse-generic-catalog"
...
 license: "This software is licensed under the Apache License version 2."
 ---
 
-import Tabs from '@theme/Tabs';
-import TabItem from '@theme/TabItem';
-
 ## Overview
 
-The Generic Lakehouse Catalog is a Gravitino catalog implementation designed to seamlessly integrate with lakehouse storage systems built on file system-based architectures. This catalog enables unified metadata management for lakehouse tables stored on various storage backends, providing a consistent interface for data discovery, governance, and access control. 
+Some table formats define a catalog of their own, and Apache Gravitino federates to it. Apache Iceberg is the example: the Iceberg catalog connects to an existing Iceberg REST catalog or Hive metastore, which remains the system of record.
```
markhoerth:
```suggestion
Some table formats define a catalog of their own, and Apache Gravitino federates to it. 
```

docs/lakehouse-generic-catalog.md
```diff
@@ -4,200 +4,166 @@ slug: "/lakehouse-generic-catalog"
...
 license: "This software is licensed under the Apache License version 2."
 ---
 
-import Tabs from '@theme/Tabs';
-import TabItem from '@theme/TabItem';
-
 ## Overview
 
-The Generic Lakehouse Catalog is a Gravitino catalog implementation designed to seamlessly integrate with lakehouse storage systems built on file system-based architectures. This catalog enables unified metadata management for lakehouse tables stored on various storage backends, providing a consistent interface for data discovery, governance, and access control. 
+Some table formats define a catalog of their own, and Apache Gravitino federates to it.
```
markhoerth:
```suggestion
Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
```

docs/lakehouse-generic-catalog.md
```diff
@@ -4,200 +4,166 @@ slug: "/lakehouse-generic-catalog"
...
+Some table formats define a catalog of their own, and Apache Gravitino federates to it. Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
 
-Gravitino fully supports the **Lance** lakehouse format, with plans to extend support to additional formats in the future.
+Lance and Delta define no catalog. A table in either format is a set of files, so the metadata needs a home. The Generic Lakehouse Catalog provides one, holding the metadata directly rather than federating. That is what makes it generic: a single catalog serves any format that arrives without one of its own.
```
markhoerth:
```suggestion
```

docs/lance-rest-integration.md
```diff
@@ -12,288 +12,183 @@ license: "This software is licensed under the Apache License version 2."
...
-
-This documentation assumes familiarity with the Lance REST service setup as described in the [Lance REST Service](./lance-rest-service) documentation.
-
-## Compatibility Matrix
-
-The following table outlines the tested compatibility between Gravitino versions and Lance connector versions:
-
-| Gravitino Version (Lance REST) | Supported lance-spark Versions | Supported lance-ray Versions                  |
-|--------------------------------|--------------------------------|-----------------------------------------------|
-| 1.1.1 - 1.2.1                  | 0.0.10 - 0.0.15                | 0.0.6 - 0.0.8                                 |
```
yuqi1129, line 25: I would suggest that we keep this compatibility matrix table to let users know about the historical compatibility status.

PR #12288, opened 2026-07-31
[#12287] feat(server): Add bulk user access-control APIs

### LINKED ISSUE #12287: [EPIC] Support bulk operations for access control entities
### Describe the proposal

Support best-effort bulk operations for access-control entities under a metalake, including users, groups, and roles.

This epic tracks the work to add REST APIs, request/response DTOs, server-side validation, authorization handling, tests, and OpenAPI documentation for bulk access-control operations.

### Task list

- [x] Add shared bulk infrastructure and bulk user APIs.
- [x] Add bulk group APIs.
- [ ] Add bulk role APIs, including reusable role request validation and per-item authorization for role removal.

jarredhj0214:

This PR adds the shared foundation for best-effort bulk access-control APIs and introduces bulk user operations.

Changes include:
- Add shared bulk request/response DTOs for item-level errors and summary counts.
- Add `gravitino.server.bulk.maxItems`, defaulting to 100.
- Add bulk user add/remove REST APIs.
- Register the bulk REST resource.
- Add OpenAPI definitions and access-control documentation for bulk user APIs.
- Add tests for bulk user best-effort behavior, duplicate-name validation, and request size limit validation.

Bulk access-control operations are needed to efficiently add or remove multiple users under a metalake while preserving item-level best-effort results.

Fix: #12287

Yes.

New REST APIs:
- `POST /api/bulk/metalakes/{metalake}/users/add`
- `POST /api/bulk/metalakes/{metalake}/users/remove`

New server configuration:
- `gravitino.server.bulk.maxItems`, default value: `100`

- `JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.18/libexec/openjdk.jdk/Contents/Home ./gradlew :server:spotlessApply :common:spotlessApply :server:test --tests org.apache.gravitino.server.web.rest.TestBulkOperations -PskipITs`
- `JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.18/libexec/openjdk.jdk/Contents/Home ./gradlew :docs:build`
- `git diff --check`


lasdf1234:
The main issues are as follows:
1. BulkOperations class: Only three methods can be retained: the constructor method, addUsers, and removeUsers.
2. In core module, add two class under org/apahce/gravitino/bulk:
BulkItemResult:

package org.apache.gravitino.bulk;

import java.util.Optional;
import javax.annotation.Nullable;

public final class BulkItemResult<T> {

  private final int index;
  private final String name;
  @Nullable private final T value;
  @Nullable private final Exception error;

  private BulkItemResult(int index, String name, @Nullable T value, @Nullable Exception error) {
    this.index = index;
    this.name = name;
    this.value = value;
    this.error = error;
  }

  public static <T> BulkItemResult<T> success(int index, String name, T value) {
    return new BulkItemResult<>(index, name, value, null);
  }

  /** For remove (or any op with no success payload). */
  public static <T> BulkItemResult<T> success(int index, String name) {
    return new BulkItemResult<>(index, name, null, null);
  }

  public static <T> BulkItemResult<T> failure(int index, String name, Exception error) {
    return new BulkItemResult<>(index, name, null, error);
  }

  public int index() {
    return index;
  }

  public String name() {
    return name;
  }

  public boolean succeeded() {
    return error == null;
  }

  public Optional<T> value() {
    return Optional.ofNullable(value);
  }

  public Optional<Exception> error() {
    return Optional.ofNullable(error);
  }
}

Another is BulkManager(The BulkManager is similar to MetaLakeManager. In the GravitinoEnv class, this BulkManager is created. (The BulkManager is similar to MetaLakeManager. In the GravitinoEnv class, this BulkManager is created. (The extra methods in the BulkOperations class can actually be moved into this class. Also, in the BulkOperations class, these can be removed:  
    Config config = GravitinoEnv.getInstance().config();
    this.bulkMaxItems =
        config == null
            ? ServerConfig.BULK_MAX_ITEMS.getDefaultValue()
            : config.get(ServerConfig.BULK_MAX_ITEMS);
)

3. In BulkOperations, Remove addUser method and AccessControlManager add three methods as follows:
import java.util.ArrayList;
import java.util.List;
import java.util.Optional;
import org.apache.gravitino.bulk.BulkItemResult;
import org.apache.gravitino.exceptions.NoSuchUserException;

List<BulkItemResult<User>> addUsers(String metalake, List<UserAdd> users) {
  return TreeLockUtils.doWithTreeLock(
      NameIdentifier.of(AuthorizationUtils.ofUserNamespace(metalake).levels()),
      LockType.WRITE,
      () -> {
        List<BulkItemResult<User>> results = new ArrayList<>(users.size());
        for (int i = 0; i < users.size(); i++) {
          UserAdd item = users.get(i);
          try {
            User user =
                item.hasExternalId()
                    ? userGroupExternalManager.addUser(
                        metalake,
                        item.name(),
                        item.externalId(),
                        item.enabled() == null ? true : item.enabled())
                    : userGroupManager.addUser(metalake, item.name());
            results.add(BulkItemResult.success(i, item.name(), user));
          } catch (Exception e) {
            results.add(BulkItemResult.failure(i, item.name(), e));
          }
        }
        return results;
      });
}

public List<BulkItemResult<Void>> removeUsers(
    String metalake, List<String> users, Optional<Owner> metalakeOwner) {
  return TreeLockUtils.doWithTreeLock(
      NameIdentifier.of(AuthorizationUtils.ofUserNamespace(metalake).levels()),
      LockType.WRITE,
      () -> {
        List<BulkItemResult<Void>> results = new ArrayList<>(users.size());
        for (int i = 0; i < users.size(); i++) {
          String name = users.get(i);
          try {
            ensureNotMetalakeOwner(metalakeOwner, metalake, name);
            boolean removed = userGroupManager.removeUser(metalake, name);
            if (!removed) {
              throw new NoSuchUserException("User does not exist: %s", name);
            }
            results.add(BulkItemResult.success(i, name));
          } catch (Exception e) {
            results.add(BulkItemResult.failure(i, name, e));
          }
        }
        return results;
      });
}

private static void ensureNotMetalakeOwner(
    Optional<Owner> metalakeOwner, String metalake, String user) {
  metalakeOwner.ifPresent(
      owner -> {
        if (owner.type() == Owner.Type.USER && owner.name().equals(user)) {
              String.format(
                  "Cannot remove user %s from metalake %s because the user is the owner of the metalake.",
                  user, metalake));
        }
      });
}

The only thing that needs to be considered is whether to create a new UserAdd class(addUsers method will use this class), or to change the input parameter to another existing User class? This needs to be evaluated.

You can modify the code according to this logic, and then I will continue to review it. Thank you. (The first PR has many details to be discussed and may need to be revised many times. The subsequent PRs will be merged quickly.)

jarredhj0214:
> Why is the input parameter of addUsers a List users? Your scenario should not include externalId. Let's first create a simpler version without externalId.
> 
> You can modify the code according to this logic, and then I will continue to review it. Thank you. (The first PR has many details to be discussed and may need to be revised many times. The subsequent PRs will be merged quickly.)

Thanks for the detailed suggestion.

Regarding the `addUsers` input, the current request body follows the approved design doc. In section 3.1, the bulk add users API is defined as `POST /api/bulk/metalakes/{metalake}/users/add`, and the request body is specified as a `users` array whose items contain `name`, `externalId`, and `enabled`:

```json
{
  "users": [
    { "name": "alice", "externalId": "ext-alice", "enabled": true },
    { "name": "bob", "enabled": true },
    { "name": "alice" }
  ]
}

**Comment by lasdf1234** [comment]
> > Why is the input parameter of addUsers a List users? Your scenario should not include externalId. Let's first create a simpler version without externalId.
> > You can modify the code according to this logic, and then I will continue to review it. Thank you. (The first PR has many details to be discussed and may need to be revised many times. The subsequent PRs will be merged quickly.)
> 
> Thanks for the detailed suggestion.
> 
> Regarding the `addUsers` input, the current request body follows the approved design doc. In section 3.1, the bulk add users API is defined as `POST /api/bulk/metalakes/{metalake}/users/add`, and the request body is specified as a `users` array whose items contain `name`, `externalId`, and `enabled`:
> 
> ```json
> {
>   "users": [
>     { "name": "alice", "externalId": "ext-alice", "enabled": true },
>     { "name": "bob", "enabled": true },
>     { "name": "alice" }
>   ]
> }
> ```

I have already corrected the comment.

**Comment by jarredhj0214** [comment]
> > > Why is the input parameter of addUsers a List users? Your scenario should not include externalId. Let's first create a simpler version without externalId.
> > > You can modify the code according to this logic, and then I will continue to review it. Thank you. (The first PR has many details to be discussed and may need to be revised many times. The subsequent PRs will be merged quickly.)
> > 
> > 
> > Thanks for the detailed suggestion.
> > Regarding the `addUsers` input, the current request body follows the approved design doc. In section 3.1, the bulk add users API is defined as `POST /api/bulk/metalakes/{metalake}/users/add`, and the request body is specified as a `users` array whose items contain `name`, `externalId`, and `enabled`:
> > ```json
> > {
> >   "users": [
> >     { "name": "alice", "externalId": "ext-alice", "enabled": true },
> >     { "name": "bob", "enabled": true },
> >     { "name": "alice" }
> >   ]
> > }
> > ```
> 
> I have already corrected the comment.

Thanks for the suggestion. I have updated this PR following your proposal.

The latest changes include:
- Kept `BulkOperations` thin with only the constructor, `addUsers`, and `removeUsers`.
- Added core-level `BulkManager` and `BulkItemResult`.
- Added `AccessControlManager#addUsers` and `AccessControlManager#removeUsers`, both using a single tree lock for the whole bulk request.
- Introduced a new `UserAdd` core object for bulk user creation instead of reusing REST DTOs in core.
- Kept `externalId` and `enabled` support for `addUsers` according to the design doc.
- Changed duplicate user names in `addUsers` to be reported as item-level errors instead of rejecting the whole request.

File: core/src/main/java/org/apache/gravitino/bulk/BulkManager.java [hunk 1]
```diff
@@ -0,0 +1,136 @@
...
+   * @param accessControlDispatcher The access-control dispatcher.
+   * @param ownerDispatcher The owner dispatcher.
+   */
+  public BulkManager(
+      Config config,
+      AccessControlDispatcher accessControlDispatcher,
+      OwnerDispatcher ownerDispatcher) {
+    this.accessControlDispatcher = accessControlDispatcher;
+    this.ownerDispatcher = ownerDispatcher;
+    this.bulkMaxItems =
```
lasdf1234: this.accessControlDispatcher = accessControlDispatcher;
    this.ownerDispatcher = ownerDispatcher;

These two parameters(accessControlDispatcher, ownerDispatcher) should stay in BulkOperations.

server/src/main/java/org/apache/gravitino/server/web/rest/BulkOperations.java
```diff
@@ -0,0 +1,159 @@
...
+@NameBindings.AccessControlInterfaces
+@Path("/bulk/metalakes/{metalake}")
+public class BulkOperations {
+
+  private final BulkManager bulkManager;
+
+  @Context private HttpServletRequest httpRequest;
+
+  /** Creates a new bulk operations resource. */
+  public BulkOperations() {
```
lasdf1234, line 75: this.accessControlManager = GravitinoEnv.getInstance().accessControlDispatcher();
    this.ownerManager = GravitinoEnv.getInstance().ownerDispatcher();

Add these in Constructor method

core/src/main/java/org/apache/gravitino/bulk/BulkManager.java
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Adds users in bulk.
+   *
+   * @param metalake The metalake name.
+   * @param users The users to add.
+   * @return The item-level results.
+   */
+  public List<BulkItemResult<User>> addUsers(String metalake, List<UserAdd> users) {
+    checkBulkSize("users", users.size());
```
lasdf1234: This method(addUsers) seems unnecessary. Could you delete addUsers?

core/src/main/java/org/apache/gravitino/bulk/BulkManager.java
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Removes users in bulk.
+   *
+   * @param metalake The metalake name.
+   * @param users The user names to remove.
+   * @return The item-level results.
+   */
+  public List<BulkItemResult<String>> removeUsers(String metalake, List<String> users) {
+    checkBulkSize("names", users.size());
```
lasdf1234: This method(removeUsers) seems unnecessary.Could you delete removeUsers?

core/src/main/java/org/apache/gravitino/bulk/BulkManager.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.gravitino.dto.responses.ErrorConstants;
+import org.apache.gravitino.exceptions.AlreadyExistsException;
+import org.apache.gravitino.exceptions.ForbiddenException;
+import org.apache.gravitino.exceptions.NotFoundException;
+import org.apache.gravitino.exceptions.NotInUseException;
+import org.apache.gravitino.metalake.MetalakeManager;
+
+/** Manages best-effort bulk operations. */
+public class BulkManager {
+
```
lasdf1234, line 32: This class can add a new method checkBulkSize and BulkOperations can use this method.

jarredhj0214: > This class can add a new method checkBulkSize and BulkOperations can use this method.

Sorry for the late response. I was on vacation last week.

I have fixed the review comments:
- Moved `accessControlDispatcher` and `ownerDispatcher` back to `BulkOperations`.
- Initialized them in the `BulkOperations` constructor.
- Removed `addUsers` and `removeUsers` from `BulkManager`.
- Kept `checkBulkSize` in `BulkManager` and now `BulkOperations` uses it.

I also rebased this PR on the latest `apache/main` and resolved the conflicts.

Tests passed:
```bash
./gradlew :server:test --tests org.apache.gravitino.server.web.rest.TestBulkOperations :core:test --tests org.apache.gravitino.authorization.TestAccessControlManager.testBulkAddUsers --tests org.apache.gravitino.authorization.TestAccessControlManager.testBulkRemoveUsers :docs:build -PskipITs

**Comment by lasdf1234** [comment]
@roryqi @yuqi1129  I have no more comment. Could you take a look?

File: core/src/main/java/org/apache/gravitino/authorization/AccessControlManager.java [hunk 6]
```diff
@@ -80,6 +84,34 @@ public User addUser(String metalake, String user, String externalId, boolean ena
...
 
+  @Override
+  public List<BulkItemResult<User>> addUsers(String metalake, List<UserAdd> users)
+      throws NoSuchMetalakeException {
+    return TreeLockUtils.doWithTreeLock(
+        NameIdentifier.of(AuthorizationUtils.ofUserNamespace(metalake).levels()),
+        LockType.WRITE,
+        () -> {
+          List<BulkItemResult<User>> results = Lists.newArrayListWithCapacity(users.size());
+          for (int index = 0; index < users.size(); index++) {
```
yuqi1129, line 95: What is the timeout setting between the server and client? I'm afraid that the client may receive an HTTP timeout before fetching the result.

core/src/main/java/org/apache/gravitino/authorization/AccessControlManager.java
```diff
@@ -88,6 +120,32 @@ public boolean removeUser(String metalake, String user) throws NoSuchMetalakeExc
...
+        LockType.WRITE,
+        () -> {
+          List<BulkItemResult<String>> results = Lists.newArrayListWithCapacity(users.size());
+          for (int index = 0; index < users.size(); index++) {
+            String user = users.get(index);
+            try {
+              ensureNotMetalakeOwner(metalakeOwner, metalake, user);
+              boolean removed = userGroupManager.removeUser(metalake, user);
+              if (!removed) {
+                throw new NoSuchUserException("User does not exist: %s", user);
```
yuqi1129: I think you can directly add a failed item to the result here.

core/src/main/java/org/apache/gravitino/bulk/BulkManager.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.gravitino.dto.responses.ErrorConstants;
+import org.apache.gravitino.exceptions.AlreadyExistsException;
+import org.apache.gravitino.exceptions.ForbiddenException;
+import org.apache.gravitino.exceptions.NotFoundException;
+import org.apache.gravitino.exceptions.NotInUseException;
+
+/** Manages best-effort bulk operations. */
+public class BulkManager {
+
+  private final int bulkMaxItems;
```
yuqi1129: Is `maxBulkItems` more proper?

docs/security/access-control.md
```diff
@@ -270,6 +270,15 @@ object: the owner of the table or view, plus `CREATE_TABLE` or `CREATE_VIEW` on
...
 | Job          |                         | Owner                                  | Owner           |                                           |
 
+Bulk user access-control APIs use the same privileges as the matching single-user operations. These
+bulk operations are authorized once before processing the request. Bulk user add requests report
+item-level failures in `errors`.
+
+| API                                                 | Required privilege                          | Request field |
+|-----------------------------------------------------|---------------------------------------------|---------------|
+| `POST /api/bulk/metalakes/{metalake}/users/add`     | `OWNER` of the metalake or `MANAGE_USERS`   | `users`       |
+| `POST /api/bulk/metalakes/{metalake}/users/remove`  | `OWNER` of the metalake or `MANAGE_USERS`   | `names`       |
```
yuqi1129: Can you add some examples, such as curl and Java examples

server/src/main/java/org/apache/gravitino/server/web/rest/BulkOperations.java
```diff
@@ -0,0 +1,178 @@
...
+  public Response addUsers(
+      @PathParam("metalake") @AuthorizationMetadata(type = Entity.EntityType.METALAKE)
+          String metalake,
+      BulkUserAddRequest request) {
+    try {
+      return Utils.doAs(
+          httpRequest,
+          () -> {
+            request.validate();
+            bulkManager.checkBulkSize("users", request.getUsers().length);
```
yuqi1129: I think the first parameter, of type string, is arbitrary, and we can pass any value without checking.

core/src/main/java/org/apache/gravitino/authorization/AccessControlManager.java
jarredhj0214: Good point. This PR keeps the bulk user APIs as synchronous best-effort APIs, so the timeout still depends on the existing client/server/proxy settings.

The request size is bounded by `gravitino.server.bulk.maxItems`, defaulting to 100. For larger inputs, clients should split them into multiple bulk requests. If we want to support much larger batches later, I think an async bulk job API would be a better follow-up.

server/src/main/java/org/apache/gravitino/server/web/rest/BulkOperations.java
jarredhj0214: Thanks for the review. I have fixed the other comments in the latest update:

- Directly added a failed item to the result when a user does not exist during bulk remove.
- Renamed `bulkMaxItems` to `maxBulkItems`.
- Replaced the arbitrary string parameter in `checkBulkSize` with a typed `BulkRequestField`.
- Added curl and Java examples for the bulk user APIs.

roryqi (commented):
I found a few things worth addressing:

1. `AccessControlEventDispatcher#addUsers` and `#removeUsers` currently just delegate to the underlying dispatcher. The single-user paths dispatch `AddUserPreEvent` / `AddUserEvent` / `AddUserFailureEvent` and `RemoveUserPreEvent` / `RemoveUserEvent` / `RemoveUserFailureEvent`, but the bulk paths skip those events entirely. This means audit/listener integrations will miss user changes made through the new bulk APIs. Could we either emit the existing per-user events for each item, or introduce explicit bulk events and cover them in tests?

2. `BulkRequestField` feels a bit over-engineered and also leaks REST request field names into the core `BulkManager`. Since it is only used to format an error message, I think this can be simplified to `checkBulkSize(String fieldName, int size)`, with the REST layer passing `"users"` / `"names"` directly or via private constants in `BulkOperations`. That keeps payload naming at the REST boundary and makes the helper more generic for future bulk APIs.

3. The new `gravitino.server.bulk.maxItems` config is added to `Configs` and `conf/gravitino.conf.template`, but it is missing from `docs/gravitino-server-config.md`. Please add it to the server configuration table so users can discover the setting from the main config reference.

4. Related to the new config: Docker config rewrite does not expose an env var mapping for it. If this server setting should be configurable the same way as other `gravitino.server.*` options in Docker, please add something like `GRAVITINO_SERVER_BULK_MAX_ITEMS -> server.bulk.maxItems` in `dev/docker/gravitino/rewrite_gravitino_server_config.py` and document it in the container config table.

jarredhj0214:
> I found a few things worth addressing:
> 
> 1. `AccessControlEventDispatcher#addUsers` and `#removeUsers` currently just delegate to the underlying dispatcher. The single-user paths dispatch `AddUserPreEvent` / `AddUserEvent` / `AddUserFailureEvent` and `RemoveUserPreEvent` / `RemoveUserEvent` / `RemoveUserFailureEvent`, but the bulk paths skip those events entirely. This means audit/listener integrations will miss user changes made through the new bulk APIs. Could we either emit the existing per-user events for each item, or introduce explicit bulk events and cover them in tests?
> 2. `BulkRequestField` feels a bit over-engineered and also leaks REST request field names into the core `BulkManager`. Since it is only used to format an error message, I think this can be simplified to `checkBulkSize(String fieldName, int size)`, with the REST layer passing `"users"` / `"names"` directly or via private constants in `BulkOperations`. That keeps payload naming at the REST boundary and makes the helper more generic for future bulk APIs.
> 3. The new `gravitino.server.bulk.maxItems` config is added to `Configs` and `conf/gravitino.conf.template`, but it is missing from `docs/gravitino-server-config.md`. Please add it to the server configuration table so users can discover the setting from the main config reference.
> 4. Related to the new config: Docker config rewrite does not expose an env var mapping for it. If this server setting should be configurable the same way as other `gravitino.server.*` options in Docker, please add something like `GRAVITINO_SERVER_BULK_MAX_ITEMS -> server.bulk.maxItems` in `dev/docker/gravitino/rewrite_gravitino_server_config.py` and document it in the container config table.

Thanks for the review. I have addressed the comments in the latest update:

- Updated `AccessControlEventDispatcher#addUsers` and `#removeUsers` to emit the existing per-user add/remove pre, success, and failure events for each bulk item, with tests added.
- Simplified `BulkManager#checkBulkSize` to take `String fieldName, int size`, and moved the REST request field names back to `BulkOperations`.
- Added `gravitino.server.bulk.maxItems` to `docs/gravitino-server-config.md`.
- Added the Docker env mapping `GRAVITINO_SERVER_BULK_MAX_ITEMS -> server.bulk.maxItems` and documented it in the container configuration table.

docs/security/access-control.md
```diff
@@ -270,6 +270,66 @@ object: the owner of the table or view, plus `CREATE_TABLE` or `CREATE_VIEW` on
 | Job template | `REGISTER_JOB_TEMPLATE` | `USE_JOB_TEMPLATE`                     | Owner           | Run a job: `RUN_JOB` and `USE_JOB_TEMPLATE` |
 | Job          |                         | Owner                                  | Owner           |                                           |
 
+Bulk user access-control APIs use the same privileges as the matching single-user operations. These
+bulk operations are authorized once before processing the request. Bulk user add requests report
+item-level failures in `errors`.
+
+| API                                                 | Required privilege                          | Request field |
```
roryqi: Why do we need request field? Is it a user oriented concepted?

docs/security/access-control.md
```diff
@@ -270,6 +270,66 @@ object: the owner of the table or view, plus `CREATE_TABLE` or `CREATE_VIEW` on
...
+curl -X POST "http://localhost:8090/api/bulk/metalakes/{metalake}/users/remove" \
+  -H "Authorization: Bearer $MANAGER_TOKEN" \
+  -H "Accept: application/vnd.gravitino.v1+json" \
+  -H "Content-Type: application/json" \
+  -d '{
+  "names": ["analyst", "developer"]
+}'
+```
+
+The same bulk add request can be sent from Java with `HttpClient`:
```
roryqi: We should use Gravitino Java client here. But we maybe don't need the Java client API here. Java client is used for the connector. It's ok not to add the API to the Gravitino Java client. 
But we should remove the document here.

docs/security/access-control.md
jarredhj0214: Removed the `Request field` column from the bulk user API privilege table.

docs/security/access-control.md
jarredhj0214: Removed the Java `HttpClient` example from this section.

jarredhj0214:
Thanks for the review. I updated the access-control documentation:

- Removed the `Request field` column from the bulk user API privilege table.
- Removed the Java `HttpClient` example from this section.
@roryqi

jarredhj0214:
Hi  @roryqi , friendly ping on this PR.

I have addressed the latest review comments and synced with `main`. Could you please help take another look when convenient? If there are still any concerns or changes needed, I’m happy to follow up. Thanks!

roryqi (approved):
LGTM.

PR #12291, opened 2026-07-31
[MINOR] docs: Update server configuration and relational backend storage pages

markhoerth:

Update `gravitino-server-config.md` and `how-to-use-relational-backend-storage.md`.

Corrections, each verified against source rather than carried over from the existing pages:

- `gravitino.server.visibleConfigs` was documented as restrictive ("properties the config endpoint is allowed to return"). `ConfigServlet` shows it is additive: the endpoint always publishes a fixed set, and each entry in `visibleConfigs` adds to what `GET /configs` exposes without authentication.
- `GRAVITINO_AUTHORIZATION_THREAD_POOL_SIZE` was documented as a container environment variable but is not in the allowlist in `rewrite_gravitino_server_config.py`, so setting it does nothing. The whole environment variable table is now generated from that file.
- The catalog provider table was missing `glue`, `lakehouse-generic`, `jdbc-starrocks`, `jdbc-clickhouse`, and `jdbc-hologres`, and listed a conf path for `jdbc-oceanbase` that exists only in the `-all` distribution package.
- `gravitino.eventListener.names` was marked required; `EventListenerConfig` gives it a default.
- `gravitino.auxService.names` was documented as defaulting to nothing and mentioning only `iceberg-rest`; the shipped `conf/gravitino.conf` sets `iceberg-rest,lance-rest`.
- Both JDBC driver links were stale: MySQL pointed at the retired `mysql:mysql-connector-java` artifact, and PostgreSQL pinned `postgresql-42.7.11.jar`.

Structural changes: the config page gains Quick Start profiles for development, production, and Docker, and its reference sections are grouped instead of running as a flat list of same-level headings. The relational page becomes one procedure per database rather than two near-identical copies. The event listener tables, which enumerated 345 class names, are replaced by the naming rule and a link to the package. The container section now documents that the entrypoint overwrites `conf/gravitino.conf` unconditionally for the properties it has defaults for, and the `SKIP_CONFIG_REWRITE` escape hatch.

These two pages are the entry point for configuring a server, and several statements in them would mislead an operator. The `visibleConfigs` inversion is the most serious: someone would reasonably set it believing it narrows what an unauthenticated endpoint exposes, when it widens it. The container rewrite behaviour is undocumented today, so a `gravitino.conf` mounted from a ConfigMap is silently overwritten with no indication why.

Documentation only. No property is added or removed. Previously undocumented but pre-existing properties and behaviours are now covered: `gravitino.authorization.serviceAdmins`, `impl`, `threadPoolSize`; `SKIP_CONFIG_REWRITE`; and the container's unconditional config rewrite.

Every property name, default, and behavioural claim traced to `Configs.java`, `JettyServerConfig`, `HealthOperations`, `ConfigServlet`, `AuditLogRedactor`, `FileAuditWriter`, `H2Database`, `conf/log4j2.properties.template`, `conf/gravitino.conf.template`, or the catalog `build.gradle.kts` files. All relative links and heading anchors resolved against the docs tree.

### DIFF
--- a/docs/gravitino-server-config.md
+++ b/docs/gravitino-server-config.md
@@ -8,360 +8,218 @@ license: "This software is licensed under the Apache License version 2."
 
 ## Introduction
 
-Apache Gravitino supports several configurations:
+The Apache Gravitino server reads `conf/gravitino.conf` at startup. Almost every property has a
+default, so the server starts with an empty file, and most deployments change only a handful of
+them. The exception is `gravitino.authorization.serviceAdmins`, which you must set once you turn
+authorization on.
 
-1. **Gravitino server configuration**: Used to start up the Gravitino server.
-2. **Gravitino catalog properties configuration**: Used to make default values for different catalogs.
-3. **Some other configurations**: Includes HDFS and other configurations.
+This page covers the server itself. Catalog properties, which configure an individual catalog
+rather than the server, are covered further down. Properties for the auxiliary services live with
+those services: see [Iceberg REST Catalog Service](iceberg-rest-service.md) and
+[Security](security/security.md).
 
-## Gravitino Server Configurations
+## Quick Start
 
-Customize the Gravitino server by editing the configuration file `gravitino.conf` in the `conf` directory. The default values are sufficient for most use cases.
-We strongly recommend that you read the following sections to understand the configuration file, so you can change the default values to suit your specific situation and usage scenario.
+### Development
 
-The `gravitino.conf` file lists the configuration items in the following table. It groups those items into the following categories:
+The defaults are already right for local work. The server listens on `0.0.0.0:8090` and keeps its
+metadata in an embedded H2 database, so no configuration is needed:
 
-### HTTP Server Configuration
-
-| Configuration item                                   | Description                                                                                                                                                                           | Default value                                                                | Required | Since version    |
-|------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------|----------|------------------|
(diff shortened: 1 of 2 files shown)


docs/gravitino-server-config.md
```diff
@@ -8,587 +8,662 @@ license: "This software is licensed under the Apache License version 2."
...
+`property = value` pair per line. The server reads the file once, at startup, so a change
+takes effect on the next restart. A Default Value of `(empty)` means the property exists with an
+empty string or list; `(none)` means it has no default at all.
+
+### Serving Requests
+
+#### HTTP Server
+
+| Configuration Item                                   | Description                                                                                                                                                                                  | Default Value                           |
+|------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------|
```
jerryshao, line 161: Hi Mark, why do we remove the `Since Version` column in the new doc? My feeling is that `Since Version` is useful to let users know which version we supported this configuration with. WDYT?

jerryshao (approved):

jerryshao:
Hi @markhoerth, this change is a behavior change. I will not merge it into branch-1.3; it will only happen in 2.0. Do you want it to be in 1.3?

markhoerth:
@jerryshao This is a documentation only change that does add or remove any configuration, so it should not be a behavior change.  Do you see something I missed?

jerryshao:
I feel this is quite a large change compared to 1.3.0. Users will be surprised that the doc in 1.3.1 is quite different from 1.3.0. But I don't have a strong opinion on it.

PR #12294, opened 2026-07-31
[#11926] Encode Trino dialect Hive views using Trino's native Presto View format

diqiu50:
Store and read Trino dialect views in the Hive catalog using Trino's own native "Presto View" Hive Metastore encoding (via a new `TrinoNativeViewCodec`), instead of the previous custom marker/prefix-based encoding.

So Hive-backed views created through Gravitino are interoperable with a native Trino/Presto Hive connector pointed at the same Hive Metastore, and vice versa. This is the catalog-side (non-Trino-connector) portion of #11926, split out for independent review.

No behavior change for existing non-Trino views; Trino dialect views stored via the Hive catalog now use Trino's native encoding format.

Existing and new unit tests in `catalog-hive` (`TestHiveCatalogOperations`, `TestTrinoNativeViewCodec`).

### DIFF
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
@@ -27,6 +27,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -194,6 +195,28 @@ class HiveViewCatalogOperations implements ViewCatalog {
       // here too, instead of being silently treated as a non-Trino view.
       boolean isTrinoView =
           Dialects.TRINO.equalsIgnoreCase(HiveView.detectDialect(currentHiveTable.properties()));
+      // Gravitino's view model has no owner/runAsInvoker/path concept, so replacing a native Trino
+      // view that carries a non-default value for any of them would silently discard it (e.g. a
+      // SECURITY DEFINER view with an owner would silently become an ownerless SECURITY INVOKER
+      // view). Reject the replace instead of doing that.
+      boolean currentTrinoViewHasUnrepresentableFields = false;
+      if (isTrinoView) {
+        TrinoNativeViewCodec.ViewDefinition currentDefinition;
+        try {
+          currentDefinition = TrinoNativeViewCodec.decode(currentHiveTable.viewOriginalText());
+        } catch (IllegalArgumentException e) {
+          throw new UnsupportedOperationException(
+              "View "
+                  + ident
+                  + " carries the presto_view marker but its payload cannot be "
+                  + "decoded",
+              e);
+        }
+        currentTrinoViewHasUnrepresentableFields =
+            currentDefinition.owner != null
+                || !currentDefinition.runAsInvoker
+                || !currentDefinition.path.isEmpty();
+      }
 
       String newViewName = currentHiveTable.name();
       String updatedViewOriginalText = currentHiveTable.viewOriginalText();
@@ -220,6 +243,12 @@ class HiveViewCatalogOperations implements ViewCatalog {
                       + "ReplaceView to change it, not SetProperty(comment)");
             }
             updatedComment = sp.getValue();
+          } else if (TrinoNativeViewCodec.PRESTO_VIEW_FLAG.equals(sp.getProperty())) {
+            throw new UnsupportedOperationException(
+                "Property '"
+                    + TrinoNativeViewCodec.PRESTO_VIEW_FLAG
+                    + "' is reserved for native Trino view storage and cannot be set directly; "
+                    + "use ReplaceView to change the view's dialect");
           } else {
             updatedProperties.put(sp.getProperty(), sp.getValue());
           }
@@ -232,10 +261,24 @@ class HiveViewCatalogOperations implements ViewCatalog {
                       + "ReplaceView to change it, not RemoveProperty(comment)");
             }
             updatedComment = null;
+          } else if (TrinoNativeViewCodec.PRESTO_VIEW_FLAG.equals(property)) {
+            throw new UnsupportedOperationException(
+                "Property '"
+                    + TrinoNativeViewCodec.PRESTO_VIEW_FLAG
+                    + "' is reserved for native Trino view storage and cannot be removed "
+                    + "directly; use ReplaceView to change the view's dialect");
           } else {
             updatedProperties.remove(property);
           }
         } else if (change instanceof ViewChange.ReplaceView) {
+          if (currentTrinoViewHasUnrepresentableFields) {
+            throw new UnsupportedOperationException(
+                "View "
+                    + ident
+                    + " is a native Trino view with a non-default owner, runAsInvoker, or SQL "
+                    + "path; Gravitino cannot represent these fields, so replacing it would "
+                    + "silently discard them");
+          }
           ViewChange.ReplaceView replace = (ViewChange.ReplaceView) change;
           SQLRepresentation sqlRepresentation =
               validateSQLRepresentation(
@@ -521,6 +564,12 @@ class HiveViewCatalogOperations implements ViewCatalog {
                 + "payload cannot be decoded on the next load",
             selected.dialect(),
             ident);
+        Preconditions.checkArgument(
+            defaultSchema == null || defaultCatalog != null,
+            "Dialect '%s' does not support a defaultSchema without a defaultCatalog for view "
+                + "%s, since Trino's native view format rejects such a payload",
+            selected.dialect(),
+            ident);
         return selected;
       case Dialects.FLINK:
         Preconditions.checkArgument(
@@ -597,7 +646,8 @@ class HiveViewCatalogOperations implements ViewCatalog {
                 viewColumns,
                 comment,
                 /* owner= */ null,
-                /* runAsInvoker= */ true));
+                /* runAsInvoker= */ true,
+                /* path= */ Collections.emptyList()));
       default:
         throw new UnsupportedOperationException(
             String.format(
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
@@ -189,10 +189,11 @@ class HiveViewCatalogOperations implements ViewCatalog {
         throw new NoSuchViewException("No view named %s (it is a table, not a view)", ident.name());
       }
 
+      // Reuse the same dialect detection as loadHiveView()/toHiveView() so that a presto_view
+      // entry that is not a plain Trino view (e.g. a Trino/Presto materialized view) is rejected
+      // here too, instead of being silently treated as a non-Trino view.
       boolean isTrinoView =
-          "true"
-              .equalsIgnoreCase(
-                  currentHiveTable.properties().get(TrinoNativeViewCodec.PRESTO_VIEW_FLAG));
+          Dialects.TRINO.equalsIgnoreCase(HiveView.detectDialect(currentHiveTable.properties()));
 
       String newViewName = currentHiveTable.name();
       String updatedViewOriginalText = currentHiveTable.viewOriginalText();
--- a/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
+++ b/catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
@@ -121,7 +121,7 @@ class HiveViewCatalogOperations implements ViewCatalog {
     Map<String, String> safeProperties = properties == null ? ImmutableMap.of() : properties;
     SQLRepresentation sqlRepresentation =
         validateSQLRepresentation(
-            representations, defaultCatalog, defaultSchema, safeProperties, ident);
+            representations, defaultCatalog, defaultSchema, safeProperties, columns, ident);
 
     try {
       Map<String, String> params = Maps.newHashMap(safeProperties);
@@ -242,6 +242,7 @@ class HiveViewCatalogOperations implements ViewCatalog {
                   replace.getDefaultCatalog(),
                   replace.getDefaultSchema(),
                   updatedProperties,
+                  replace.getColumns(),
                   ident);
           updatedColumns = hmsColumns(replace.getColumns(), sqlRepresentation.dialect());
           updatedViewOriginalText =
@@ -252,10 +253,9 @@ class HiveViewCatalogOperations implements ViewCatalog {
                   replace.getDefaultCatalog(),
                   replace.getDefaultSchema(),
                   ident);
+          isTrinoView = Dialects.TRINO.equalsIgnoreCase(sqlRepresentation.dialect());
           updatedComment =
-              Dialects.TRINO.equalsIgnoreCase(sqlRepresentation.dialect())
-                  ? TrinoNativeViewCodec.PRESTO_VIEW_COMMENT
-                  : replace.getComment();
+              isTrinoView ? TrinoNativeViewCodec.PRESTO_VIEW_COMMENT : replace.getComment();
           applyTrinoViewMarker(updatedProperties, sqlRepresentation.dialect());
         } else {
           throw new IllegalArgumentException(
@@ -485,6 +485,7 @@ class HiveViewCatalogOperations implements ViewCatalog {
       String defaultCatalog,
       String defaultSchema,
       Map<String, String> properties,
+      Column[] columns,
       NameIdentifier ident) {
     int representationCount = representations == null ? 0 : representations.length;
     Representation firstRepresentation =
@@ -513,6 +514,12 @@ class HiveViewCatalogOperations implements ViewCatalog {
         // The default catalog/schema are encoded into the Trino native view payload by
         // toHmsViewOriginalText() and restored in toHiveView(), so no value is required to be null
         // here.
+        Preconditions.checkArgument(
+            columns != null && columns.length > 0,
+            "Dialect '%s' requires at least one column for view %s; without it the encoded "
+                + "payload cannot be decoded on the next load",
+            selected.dialect(),
+            ident);
         return selected;
       case Dialects.FLINK:
         Preconditions.checkArgument(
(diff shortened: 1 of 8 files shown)


catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -177,6 +189,12 @@ public View alterView(NameIdentifier ident, ViewChange... changes)
         throw new NoSuchViewException("No view named %s (it is a table, not a view)", ident.name());
       }
 
+      // Reuse the same dialect detection as loadHiveView()/toHiveView() so that a presto_view
+      // entry that is not a plain Trino view (e.g. a Trino/Presto materialized view) is rejected
+      // here too, instead of being silently treated as a non-Trino view.
+      boolean isTrinoView =
+          Dialects.TRINO.equalsIgnoreCase(HiveView.detectDialect(currentHiveTable.properties()));
```
mchades, line 210: Could we also prevent `presto_view` from being modified as a regular view property in the `ViewChange.SetProperty` / `ViewChange.RemoveProperty` branches below?

This property is part of the native Trino view storage contract. I verified that removing it from a Trino view succeeds and is persisted; the returned and subsequently loaded view is then classified as a Hive view and exposes the encoded `/* Presto View: ... */` wrapper as its SQL.

Please treat this key as reserved, or restore it according to the final dialect after applying all changes. It would also be useful to cover both removing it from a Trino view and setting it on a non-Trino view.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -431,6 +511,17 @@ private SQLRepresentation validateSQLRepresentation(
             defaultSchema,
             ident);
         return selected;
+      case Dialects.TRINO:
```
mchades, line 570: Could we add the invariant `defaultSchema == null || defaultCatalog != null` here? This still permits both values to be null; it rejects only the schema-without-catalog combination.

Trino's `ConnectorViewDefinition` explicitly rejects a schema when the catalog is absent. I verified that Gravitino currently accepts this input and persists a payload that native Trino cannot decode. A regression test for the schema-only case would help.

Reference: https://github.com/trinodb/trino/blob/435/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorViewDefinition.java#L30-L65

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -457,27 +548,61 @@ private SQLRepresentation validateSQLRepresentation(
...
+                .collect(Collectors.toList());
+        return TrinoNativeViewCodec.encode(
+            new TrinoNativeViewCodec.ViewDefinition(
+                representation.sql(),
+                defaultCatalog,
+                defaultSchema,
+                viewColumns,
+                comment,
+                /* owner= */ null,
+                /* runAsInvoker= */ true));
```
mchades: A native Trino view definition also carries `owner`, `runAsInvoker`, and `path`. At present, `path` is not decoded, `toHiveView` drops the owner/security fields, and every replacement is encoded with `owner=null`, `runAsInvoker=true`, and an empty path.

I verified that replacing a native view with `owner="alice"` and `runAsInvoker=false` silently changes it to an invoker view with no owner. Since Trino uses `SECURITY DEFINER` by default, this can change both authorization and name-resolution behavior.

Could we preserve these fields end to end? If the Gravitino view model cannot represent them yet, rejecting unsupported native definitions or replacements would be safer than silently changing their semantics.

mchades (commented):
Thanks for the implementation. I found three interoperability/data-integrity issues in the native Trino view handling; details are in the inline comments.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -457,27 +597,62 @@ private SQLRepresentation validateSQLRepresentation(
...
+                "Hive catalog currently supports only [%s] view dialects, but got '%s' for view %s",
+                SUPPORTED_VIEW_DIALECTS, selected.dialect(), ident));
+    }
+  }
+
+  /**
+   * Sets or clears the {@code presto_view} marker in the given HMS property map, so that a Trino
+   * dialect view is recognized as a native Trino view (see {@link TrinoNativeViewCodec}).
+   */
+  private static void applyTrinoViewMarker(Map<String, String> params, String dialect) {
```
mchades: Is it necessary to use static?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -116,18 +122,25 @@ public View createView(
...
       Map<String, String> params = Maps.newHashMap(safeProperties);
       params.put(TABLE_TYPE, TableType.VIRTUAL_VIEW.name());
-      String viewOriginalText = toHmsViewOriginalText(sqlRepresentation, ident);
+      applyTrinoViewMarker(params, sqlRepresentation.dialect());
+      String viewOriginalText =
+          toHmsViewOriginalText(
+              sqlRepresentation, columns, comment, defaultCatalog, defaultSchema, ident);
+      String hmsComment =
+          Dialects.TRINO.equalsIgnoreCase(sqlRepresentation.dialect())
+              ? TrinoNativeViewCodec.PRESTO_VIEW_COMMENT
```
mchades: Could we add a create/load round-trip test with a non-null comment to verify that the original comment is preserved in the encoded payload?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
```diff
@@ -116,18 +122,25 @@ public View createView(
...
     SQLRepresentation sqlRepresentation =
         validateSQLRepresentation(
-            representations, defaultCatalog, defaultSchema, safeProperties, ident);
+            representations, defaultCatalog, defaultSchema, safeProperties, columns, ident);
 
     try {
       Map<String, String> params = Maps.newHashMap(safeProperties);
       params.put(TABLE_TYPE, TableType.VIRTUAL_VIEW.name());
-      String viewOriginalText = toHmsViewOriginalText(sqlRepresentation, ident);
+      applyTrinoViewMarker(params, sqlRepresentation.dialect());
```
mchades: Could we align on the ownership contract for Trino-dialect views? The caller should provide the logical SQL, output columns, user comment, default catalog/schema, and ordinary user properties, while Gravitino should derive the native HMS representation: `presto_view=true`, the HMS comment `Presto View`, a dummy HMS column, and the encoded `viewOriginalText`. 

If this is the intended contract, could we centralize that conversion for both create and replace, and register `presto_view` in `HiveTablePropertiesMetadata` as a reserved property? This would prevent callers from setting or removing the storage marker while allowing the connector to manage it from the final dialect.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
```diff
@@ -0,0 +1,435 @@
...
+import org.apache.gravitino.rel.types.Types;
+
+/**
+ * Encodes and decodes Trino/Presto's native "Presto View" HMS view format, so that views created by
+ * Gravitino's Trino dialect are readable by a native Trino Hive connector (and vice versa).
+ *
+ * <p>Trino recognizes a Hive Metastore VIRTUAL_VIEW table as a Trino view when it carries the
+ * {@code presto_view=true} table parameter and its {@code comment} table parameter equals {@code
+ * "Presto View"}; the view body is stored in {@code viewOriginalText} as {@code "/* Presto View: "
+ * + base64(json) + " * /"}. The JSON payload mirrors Trino's {@code ConnectorViewDefinition} wire
```
mchades, line 42: Was this codec adapted from Trino source? If so, please add the corresponding attribution for this file under `Trino` in the root `LICENSE`.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
mchades: The guard looks fixed. Could you also add the two regression tests mentioned above: removing `presto_view` from a Trino view and setting it on a non-Trino view?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
mchades: The invariant looks fixed. Could you also add the schema-without-catalog regression test mentioned above?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
mchades: The rejection looks fixed. Could you also add regression coverage for non-default `owner`, `runAsInvoker`, and `path`?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
diqiu50: No, this is a from-scratch implementation of the wire format, not copied from Trino's source — no io.trino.* imports, no dependency on Trino. The comments just document the format for interop purposes. I don't think it needs a LICENSE entry.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Add tests

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Added

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Removed static

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Added

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/HiveViewCatalogOperations.java
diqiu50: Fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
```diff
@@ -0,0 +1,435 @@
...
+      case DATE:
+        return "date";
+      case TIMESTAMP:
+        Types.TimestampType timestampType = (Types.TimestampType) type;
+        if (timestampType.hasTimeZone()) {
+          throw new UnsupportedOperationException(
+              "Unsupported conversion to Trino type: TIMESTAMP WITH TIME ZONE is not supported by "
+                  + "Hive-backed views");
+        }
+        return "timestamp(6)";
```
mchades: `TimestampType` carries an explicit precision, but this always writes `timestamp(6)`, and `fromTrinoTypeString` drops the encoded `p`. A `timestamp(3)` column therefore round-trips as `timestamp(6)`. Could we preserve the precision in both directions, use Trino’s default `3` only when it is unset, and add a `timestamp(3)` round-trip test?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
```diff
@@ -0,0 +1,435 @@
...
+        for (int i = 0; i < fields.length; i++) {
+          if (i > 0) {
+            row.append(',');
+          }
+          row.append(rowFieldName(fields[i].name()))
+              .append(' ')
+              .append(toTrinoTypeString(fields[i].type()));
+        }
+        return row.append(')').toString();
+      default:
```
mchades, line 284: Native Trino view column types live in the encoded payload; HMS stores only the dummy column, so Hive physical type support should not be the limiter here. Gravitino’s existing Trino transformer supports `time(p)`, `timestamp(p) with time zone`, and `uuid`, while this codec rejects them. Could we support these shared types, or narrow the documented bidirectional interoperability claim?

docs/apache-hive-catalog.md
```diff
@@ -241,9 +241,11 @@ Support for altering partitions is under development.
...
+- Supports creating views with the `hive`, `trino`, `flink`, or `spark` dialect.
+- When loading an existing HMS view, Gravitino automatically detects whether the view uses the `hive`, `trino`, `flink`, or `spark` dialect.
 - For the `hive` and `flink` dialects, `defaultCatalog` and `defaultSchema` must be `null`.
+- For the `trino` dialect, `defaultSchema` requires `defaultCatalog` to also be set (a schema without a catalog cannot be represented).
+- The `trino` dialect requires at least one output column, and is stored using Trino's own native "Presto View" Hive Metastore encoding, so a view created through Gravitino is interoperable with a native Trino/Presto Hive connector pointed at the same Hive Metastore, and vice versa. The HMS `presto_view` property this relies on is reserved and managed internally based on the view's dialect; it cannot be set or removed directly. Gravitino's view model cannot represent a native Trino view's owner, `runAsInvoker`, or SQL path, so replacing an existing native view that has a non-default value for any of them is rejected rather than silently discarding it.
```
mchades, line 248: Could we add an end-to-end IT under `trino-connector/integration-test` and wire it into the active `TrinoIT` runner? It should point a native `connector.name=hive` catalog and the Gravitino Hive catalog at the same HMS, then verify both directions: native Trino create → Gravitino REST load (SQL/columns/comment/defaults), and REST create → native Trino `SHOW CREATE VIEW`/`SELECT`. The unit tests exercise only our codec, not Trino’s actual encoder/decoder; a `timestamp(3)` column would also catch type-signature drift.

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
diqiu50: fixed

docs/apache-hive-catalog.md
diqiu50: Integration tests in the next PR #11926

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
diqiu50: fixed

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
```diff
@@ -0,0 +1,454 @@
...
+   * row(a integer,b varchar)}), for use in the encoded view's column list.
+   *
+   * @param type the Gravitino type
+   * @return the Trino type signature string
+   */
+  static String toTrinoTypeString(Type type) {
+    switch (type.name()) {
+      case BOOLEAN:
+        return "boolean";
+      case BYTE:
```
mchades, line 222: These mappings ignore `IntegralType.signed()`, including inside nested types. For example, `long unsigned` is persisted as `bigint`, while the existing Trino transformer uses `decimal(20,0)` to preserve its range; reloading also turns it into a signed `long`. Could we reject unsigned view columns or define and test the same widening contract rather than silently narrowing them?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
```diff
@@ -0,0 +1,454 @@
...
+        }
+        return row.append(')').toString();
+      default:
+        throw new UnsupportedOperationException("Unsupported conversion to Trino type: " + type);
+    }
+  }
+
+  /** Quotes a row field name if it is not a plain lowercase identifier Trino can parse bare. */
+  private static String rowFieldName(String name) {
+    if (SIMPLE_IDENTIFIER.matcher(name).matches()) {
```
mchades: `SIMPLE_IDENTIFIER` also accepts reserved words. For example, a Gravitino struct field named `select` is emitted as `row(select integer)`, which Trino cannot parse; Trino's own `NamedTypeSignature` quotes every named row field. Could we quote all field names and add a reserved-keyword test?

catalogs/catalog-hive/src/main/java/org/apache/gravitino/catalog/hive/TrinoNativeViewCodec.java
```diff
@@ -0,0 +1,454 @@
...
+      default:
+        break;
+    }
+    if (lower.startsWith("varchar(")) {
+      return Types.VarCharType.of(Integer.parseInt(innerContent(s)));
+    }
+    if (lower.startsWith("char(")) {
+      return Types.FixedCharType.of(Integer.parseInt(innerContent(s)));
+    }
+    if (lower.startsWith("time(")) {
```
mchades, line 342: `time(p) with time zone` also matches this branch and is silently converted to an unzoned `TimeType`. Trino 435 serializes `TIME WITH TIME ZONE` in exactly this form, while Gravitino has no zone-bearing `TimeType`. Could we explicitly reject it and add a regression test instead of changing the column semantics?

docs/apache-hive-catalog.md
mchades: Thanks. As discussed offline, the complete bidirectional interoperability IT will be added in a follow-up PR.

mchades (approved):

PR #12295, opened 2026-07-31
[#12218] feat(secret): Add secret provider registry and discovery API

### LINKED ISSUE #12218: [FEATURE] Add GravitinoSecretProvider SPI for entity connection secrets

lasdf1234:

- Add `SecretProviderRegistry` to load named `SecretProvider` instances from server conf.
- Own the registry in `GravitinoEnv` (same lifecycle pattern as `KmsClientRegistry`).
- Expose `GET /configs/secrets/providers` for safe provider discovery (`name` / `type` / optional `uri`).

Clients need to discover registered secret-provider instance names for create/alter bindings, and the server must keep providers loaded for the process lifetime before write/resolve paths land.

Fix: #12218

- New endpoint: `GET /configs/secrets/providers`
- New config keys: `gravitino.secret.providers`, `gravitino.secret.provider.<name>.className`, optional `gravitino.secret.provider.<name>.uri`

- `./gradlew :core:test --tests 'org.apache.gravitino.secret.TestSecretProviderRegistry' --tests 'org.apache.gravitino.TestGravitinoEnvSecretProviderRegistry' -PskipITs`
- `./gradlew :server:test --tests 'org.apache.gravitino.server.web.TestSecretProvidersConfigServlet' --tests 'org.apache.gravitino.server.TestGravitinoServer' -PskipITs`


common/src/main/java/org/apache/gravitino/secret/GravitinoSecretProvider.java
```diff
@@ -0,0 +1,79 @@
...
+  String type();
+
+  /**
+   * Writes a plaintext secret and returns its URN.
+   *
+   * @param plaintext the secret plaintext
+   * @param context the write context
+   * @return the secret URN
+   */
+  String writeSecret(String plaintext, SecretWriteContext context);
```
roryqi: SecretWriteContext used UUID. Client can't get the accurate UUID. It's hard for the client to reuse the code.

lasdf1234: Got resolved.

roryqi (approved):
LGTM.

